PR #70144, opened 2024-05-02
ref: use ARG instead of ENV for gosu / tini vars

asottile-sentry:
ENV persists into the image's runtime whereas ARG is build-only

### DIFF
--- a/self-hosted/Dockerfile
+++ b/self-hosted/Dockerfile
@@ -11,11 +11,10 @@ LABEL org.opencontainers.image.authors="oss@sentry.io"
 # add our user and group first to make sure their IDs get assigned consistently
 RUN groupadd -r sentry && useradd -r -m -g sentry sentry
 
-ENV GOSU_VERSION=1.17 \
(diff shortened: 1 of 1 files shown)


self-hosted/Dockerfile
```diff
@@ -24,11 +23,11 @@ RUN set -x \
   && apt-get update && apt-get install -y --no-install-recommends $buildDeps \
   && rm -rf /var/lib/apt/lists/* \
   # grab gosu for easy step-down from root
-  && wget -O /usr/local/bin/gosu "https://github.com/tianon/gosu/releases/download/$GOSU_VERSION/gosu-amd64" \
+  && wget --quiet -O /usr/local/bin/gosu "https://github.com/tianon/gosu/releases/download/$GOSU_VERSION/gosu-amd64" \
```
azaslavsky, line 26: What's the reasoning behind adding `--quiet` here? Just minimize log spam?

azaslavsky (approved):

self-hosted/Dockerfile
asottile-sentry: `wget` displays a progress bar which is meaningless in a docker build

PR #70145, opened 2024-05-02
feat(crons): Capture information from checkin ingestion to user for user feedback

wedamija:
This captures information about why a checkin failed to pass through ingestion. We don't display any of this to the user yet, Will tackle that in a follow up pr.

### DIFF
--- a/src/sentry/monitors/consumers/monitor_consumer.py
+++ b/src/sentry/monitors/consumers/monitor_consumer.py
@@ -1,13 +1,16 @@
 from __future__ import annotations
 
+import dataclasses
 import logging
 import uuid
 from collections import defaultdict
 from collections.abc import Mapping
 from concurrent.futures import ThreadPoolExecutor, wait
+from copy import deepcopy
 from datetime import datetime, timedelta
+from enum import Enum
 from functools import partial
-from typing import Literal
+from typing import Any, Literal
 
 import sentry_sdk
 from arroyo.backends.kafka.consumer import KafkaPayload
@@ -111,6 +114,15 @@ def _ensure_monitor_with_config(
             "errors": validator.errors,
         }
         logger.info("monitors.consumer.invalid_config", extra=extra)
+        if not monitor:
+            raise CheckinValidationError(
+                [
+                    ProcessingError(
+                        ProcessingErrorType.MONITOR_INVALID_CONFIG,
+                        {"errors": validator.errors},
+                    )
+                ]
+            )
         return monitor
 
     validated_config = validator.validated_data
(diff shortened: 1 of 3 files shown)


src/sentry/monitors/consumers/monitor_consumer.py
```diff
@@ -234,6 +246,54 @@ def transform_checkin_uuid(
...
+    MONITOR_ENVIRONMENT_RATELIMITED = 14
+    # This monitor environment is sending checkins too frequently
+    ORGANIZATION_KILLSWITCH_ENABLED = 15
+    # We have disabled checkin ingestion for this org. Contact support for details
+
+
+@dataclasses.dataclass(frozen=True)
+class ProcessingError:
+    type: ProcessingErrorType
+    data: dict[str, Any] = dataclasses.field(default_factory=dict)
```
evanpurkhiser, line 288: Wondering if it would make more sense to have a bunch of concrete dataclasses for each type, with clear data attribute typing

like

```py
@dataclass
class CheckinEnvironmnetMismatchError:
    type = ProcessingErrorType.CHECKIN_ENVIRONMENT_MISMATCH
   
    description = "The environment sent with the checkin update doesn't match the environment already associated with the check-in"

    expected_environment_name: string
    """
    The name of the environment we expected the check-in to be a part of.
    """

    got_environment_name: string
    """
    The name of the environment that was sent
    """

    @property
    def message(self):
        return f"Check-in was expected to be part of {self.expected_environment_name}. Got {self.got_environmnent_name}"
```

Idk just a thought

evanpurkhiser: this is fine tbh lol

src/sentry/monitors/consumers/monitor_consumer.py
```diff
@@ -234,6 +246,54 @@ def transform_checkin_uuid(
...
+    MONITOR_NOT_FOUND = 11
+    # Monitor with the provided slug doesn't exist, and either no or invalid upsert data provided
+    MONITOR_OVER_QUOTA = 12
+    # This monitor can't accept checkins and is over quota
+    MONITOR_ENVIRONMENT_LIMIT_EXCEEDED = 13
+    # The monitor has too many environments associated with it already, can't add another
+    MONITOR_ENVIRONMENT_RATELIMITED = 14
+    # This monitor environment is sending checkins too frequently
+    ORGANIZATION_KILLSWITCH_ENABLED = 15
+    # We have disabled checkin ingestion for this org. Contact support for details
```
evanpurkhiser, line 282: IMO we should put these in a diff file. This consumer is already getting long

src/sentry/monitors/consumers/monitor_consumer.py
```diff
@@ -305,6 +371,10 @@ def update_existing_check_in(
         )
 
     if not valid_duration(updated_duration):
+        processing_errors.append(
+            ProcessingError(ProcessingErrorType.CHECKIN_INVALID_DURATION),
+            {"duration": updated_duration},
+        )
```
evanpurkhiser, line 377: is this correct? shouldn't the `{"duration": updated_duration}` be in the `ProcessingError`

src/sentry/monitors/consumers/monitor_consumer.py
wedamija: I kind of thought about this but tbh I just wanted to keep it pretty simple

src/sentry/monitors/consumers/monitor_consumer.py
wedamija: Yeah makes sense

src/sentry/monitors/consumers/monitor_consumer.py
wedamija: Oops yeah, good catch

src/sentry/monitors/consumers/monitor_consumer.py
```diff
@@ -818,11 +898,29 @@ def process_checkin(item: CheckinItem):
             op="_process_checkin",
             name="monitors.monitor_consumer",
         ) as txn:
-            _process_checkin(item, txn)
+            _process_checkin(deepcopy(item), txn)
```
evanpurkhiser, line 901: Why do we need to do this?

tests/sentry/monitors/consumers/test_monitor_consumer.py
```diff
@@ -687,6 +827,24 @@ def test_monitor_invalid_config(self):
...
+                        {
+                            "errors": {
+                                "schedule": [
+                                    ErrorDetail(
+                                        string="Only 5 field crontab syntax is supported",
+                                        code="invalid",
+                                    )
+                                ]
+                            }
+                        },
```
evanpurkhiser, line 843: The reason I kind of think dataclasses per type may have made sense is that we're going to have to have code somewhere that coerces this down to something consistent for the API

src/sentry/monitors/consumers/monitor_consumer.py
```diff
@@ -234,6 +246,18 @@ def transform_checkin_uuid(
...
+@dataclasses.dataclass(frozen=True)
+class ProcessingError:
+    type: ProcessingErrorType
+    data: dict[str, Any] = dataclasses.field(default_factory=dict)
+
+
+class CheckinValidationError(Exception):
+    def __init__(self, processing_errors: list[ProcessingError], monitor: Monitor | None = None):
+        self.processing_errors = processing_errors
+        self.monitor = monitor
```
evanpurkhiser, line 258: Why include the monitor sometimes again? Should we capture in a comment?

src/sentry/monitors/consumers/monitor_consumer.py
```diff
@@ -305,6 +335,11 @@ def update_existing_check_in(
         )
 
     if not valid_duration(updated_duration):
+        processing_errors.append(
+            ProcessingError(
+                ProcessingErrorType.CHECKIN_INVALID_DURATION, {"duration": updated_duration}
```
evanpurkhiser, line 340: Super nit: may as well add the magic trailing comma to get black to newline this like you did up here https://github.com/getsentry/sentry/pull/70145/files#diff-107af27baa83506a99019b39a35654dcb0e19c2e5a3bef9181a0de9185c62560R281-R282

src/sentry/monitors/consumers/monitor_consumer.py
wedamija: We aren't always able to access it at the point that we failed, so sometimes we just have the slug

evanpurkhiser (approved):

tests/sentry/monitors/consumers/test_monitor_consumer.py
wedamija: This is just text returned from the validator, similar to what we'd return if they hit the api with the wrong info. We could define things more consistently if we need to, but I thought we'd just do something like
- Convert enum type to user friendly description
- Show blog of extra context to the user

wedamija: But if we want to render it better we can include more detail. Does having actual dataclasses help though? We basically just need to know which data will be present right?

PR #70148, opened 2024-05-02
ref: delete unused parse_bool

asottile-sentry:
noticed while cleaning up `cdc` stuff -- cc @JonasBa since you introduced this it seems 🤔


JonasBa:
Agh, good find @asottile-sentry, I forgot to remove it 🤦🏼

JonasBa (approved):

PR #70149, opened 2024-05-02
fix(hybrid-cloud): Adds composite index on region and control tombstones

GabeVillalobos:
Removes the `unique_together` section of the base tombstone class as it turns out this was being ignored entirely.

This also adds a composite index to both tombstone tables for the `object_identifer` and `table_name` columns.


src/sentry/migrations/0712_create_tombstone_compound_indexes.py
```diff
@@ -0,0 +1,40 @@
...
+        ("sentry", "0711_backfill_group_attributes_to_self_hosted"),
+    ]
+
+    operations = [
+        migrations.AddIndex(
+            model_name="controltombstone",
+            index=models.Index(
+                fields=["table_name", "object_identifier"], name="sentry_cont_table_n_940c6d_idx"
+            ),
+        ),
```
GabeVillalobos, line 33: Is this going to work in a split deployment? Both of these index operations target different silos and tables

markstory (approved):

src/sentry/models/tombstone.py
```diff
@@ -26,7 +26,6 @@ class TombstoneBase(Model):
 
     class Meta:
         abstract = True
-        unique_together = ("table_name", "object_identifier")
```
wedamija, line 29: Was this ignored because the subclasses define their own Meta?

GabeVillalobos: Yep, that appears to be the case. It's a bit unfortunate since this would ideally be a unique index, but I'm not sure it's worth the effort deduplicating rows to accomplish this now 😞

PR #70160, opened 2024-05-02
chore(issue-platform): Replace log with metric

snigdhas:
This isn't an actionable error, so converting it to a metric instead. 

Fixes SENTRY-1954


src/sentry/issues/status_change_consumer.py
```diff
@@ -145,16 +145,14 @@ def bulk_get_groups_from_fingerprints(
...
-                    "fingerprint": fingerprint,
-                },
-            )
+        metrics.incr(
+            "occurrence_ingest.grouphash.not_found",
+            sample_rate=1.0,
+            tags={
+                "project_id": project_id,
+                "fingerprint": fingerprint,
+            },
```
vartec, line 154: This will have too high cardinality, as a rule of thumb if there's more than 100 or so unique values it shouldn't be a tag in DataDog.

Project id might be OK-ish, but fingerprint for sure isn't.

src/sentry/issues/status_change_consumer.py
```diff
@@ -145,16 +145,14 @@ def bulk_get_groups_from_fingerprints(
...
-            logger.exception(
-                "grouphash.not_found",
-                extra={
-                    "project_id": project_id,
-                    "fingerprint": fingerprint,
-                },
-            )
+        metrics.incr(
+            "occurrence_ingest.grouphash.not_found",
+            sample_rate=1.0,
```
JoshFerge, line 150: do we need a 1.0 sample rate? or is the default fine

JoshFerge (approved):

src/sentry/issues/status_change_consumer.py
vartec: see: https://www.notion.so/sentry/Cardinality-made-easy-aba4f9cd0f5e4f3488c6d9413bc7239d

PR #70161, opened 2024-05-02
chore(trace): Don't call old trace endpoint from event details

wmak:
- As far as I can tell we don't use quickTrace at all in event details, and its 2 fairly expensive queries, removing them entirely


scttcper (changes requested):
DM'd you about the ANR component

scttcper (approved):
🙏

PR #70165, opened 2024-05-02
ref: ignore all of google.* third party namespaces

asottile-sentry:
due to the way google's namespace package is set up it sometimes thinks some of the modules exist and produces bogus errors

fixes spurious:
```
src/sentry/analytics/pubsub.py:8: error: Module "google.cloud" has no attribute "pubsub_v1"  [attr-defined]
```


evanpurkhiser (approved):

asottile-sentry:
this didn't have the effect I was expected :/  got tricked by the mypy cache

PR #70170, opened 2024-05-02
feat(traces): Use span search hints and values

k-fish:
### Summary
This adds span-specific search hints to the traces explorer search boxes. It (will) use the new span/fields endpoint to get example fields back for the span op in question. Also included is an example of omitting a span op when it doesn't make sense (eg. it's internal span only)

#### Screenshot
Screenshot 2024-05-02 at 3 35 30 PM

### DIFF
--- a/static/app/actionCreators/tags.tsx
+++ b/static/app/actionCreators/tags.tsx
@@ -150,3 +150,45 @@ export function fetchTagValues({
     query,
   });
 }
+
+export function fetchSpanFieldValues({
+  api,
+  orgSlug,
+  fieldKey,
+  endpointParams,
+  projectIds,
+  search,
+}: {
+  api: Client;
+  fieldKey: string;
+  orgSlug: string;
+  endpointParams?: Query;
+  projectIds?: string[];
+  search?: string;
+}): Promise<TagValue[]> {
+  const url = `/organizations/${orgSlug}/spans/fields/${fieldKey}/values/`;
+
+  const query: Query = {};
+  if (search) {
+    query.query = search;
+  }
+  if (projectIds) {
+    query.project = projectIds;
+  }
+  if (endpointParams) {
+    if (endpointParams.start) {
+      query.start = endpointParams.start;
+    }
+    if (endpointParams.end) {
+      query.end = endpointParams.end;
+    }
+    if (endpointParams.statsPeriod) {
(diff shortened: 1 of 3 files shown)


static/app/views/performance/traces/tracesSearchBar.tsx
```diff
@@ -39,6 +56,9 @@ export function TracesSearchBar({
               'Search for traces containing a span matching these attributes'
             )}
             organization={organization}
+            metricAlert={false}
```
k-fish, line 59: This unfortunately has to be declared explicitly as false because of a tri-state boolean in SearchBar.

static/app/views/performance/traces/tracesSearchBar.tsx
```diff
@@ -27,6 +42,8 @@ export function TracesSearchBar({
   const organization = useOrganization();
   const canAddMoreQueries = queries.length <= 2;
   const localQueries = queries.length ? queries : [''];
+  const supportedTags = getTracesSupportedTags();
```
Zylphrex, line 45: Did you want to memo this?

static/app/actionCreators/tags.tsx
```diff
@@ -150,3 +150,45 @@ export function fetchTagValues({
...
+  endpointParams?: Query;
+  projectIds?: string[];
+  search?: string;
+}): Promise<TagValue[]> {
+  const url = `/organizations/${orgSlug}/spans/fields/${fieldKey}/values/`;
+
+  const query: Query = {};
+  if (search) {
+    query.query = search;
+  }
```
Zylphrex, line 174: FYI this search doesn't do anything yet but not a big deal for the first pass I think

Zylphrex (approved):

static/app/views/performance/traces/tracesSearchBar.tsx
k-fish: nah it's a small dict comp, it shouldn't be a big deal

static/app/actionCreators/tags.tsx
k-fish: I know but I'm copying the form from tags values and I assume long-terms the apis will have similar shape. No point removing this just to add it in a month or two.

PR #70172, opened 2024-05-02
feat(api): Add `organization_id_or_slug` Support to Customer Domain Middleware

iamrajjoshi:
We are updating our APIs work with id and slug path params. We need to make sure that the middleware works with both type of path params. We need to make sure that if an orgid is passed, we don't redirect them.

For more context: https://sentry.slack.com/archives/C032E82D338/p1714605291155389

Example: `acme.sentry.io/organizations/12345/issues` to `http://12345.sentry.io/issues`.

### DIFF
--- a/src/sentry/middleware/customer_domain.py
+++ b/src/sentry/middleware/customer_domain.py
@@ -52,11 +52,8 @@ def _resolve_redirect_url(request, activeorg):
     if redirect_subdomain:
         redirect_url = generate_organization_url(activeorg)
     result = resolve(request.path)
-    org_slug_path_mismatch = (
-        result.kwargs
-        and (
-            "organization_slug" in result.kwargs and result.kwargs["organization_slug"] != activeorg
-        )
+    org_slug_path_mismatch = result.kwargs and (
+        ("organization_slug" in result.kwargs and result.kwargs["organization_slug"] != activeorg)
         or (
             "organization_id_or_slug" in result.kwargs
             and result.kwargs["organization_id_or_slug"] != activeorg
--- a/src/sentry/middleware/customer_domain.py
+++ b/src/sentry/middleware/customer_domain.py
@@ -54,14 +54,26 @@ def _resolve_redirect_url(request, activeorg):
     result = resolve(request.path)
     org_slug_path_mismatch = (
         result.kwargs
-        and "organization_slug" in result.kwargs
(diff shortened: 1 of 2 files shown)


src/sentry/middleware/customer_domain.py
```diff
@@ -52,16 +52,25 @@ def _resolve_redirect_url(request, activeorg):
     if redirect_subdomain:
         redirect_url = generate_organization_url(activeorg)
     result = resolve(request.path)
-    org_slug_path_mismatch = (
-        result.kwargs
-        and "organization_slug" in result.kwargs
-        and result.kwargs["organization_slug"] != activeorg
+    org_slug_path_mismatch = result.kwargs and (
```
sentaur-athena, line 55: To make this easier to understand, with this solution, if ID is passed in `organization_id_or_slug` and it is not the correct ID of the active org redirect_url stays activeorg url but `kwargs["organization_id_or_slug"]` remain the wrong ID so this leads to 404. I think it's expected but just clarifying.

sentaur-athena (commented):
Changes makes sense to be but a hybrid cloud stamp would be nice. @dashed ?

src/sentry/middleware/customer_domain.py
iamrajjoshi: Yep thats correct. In general, this scenario shouldn't be happening because the org_id should only be used in subdomain-less queries.

tests/sentry/middleware/test_customer_domain.py
```diff
@@ -343,6 +391,37 @@ def test_with_middleware_and_customer_domain(self):
...
+            assert "activeorg" in self.client.session
+            assert self.client.session["activeorg"] == "albertos-apples"
+
+            # No redirect for id
+            response = self.client.get(
+                reverse("org-events-endpoint-id-or-slug", kwargs={"organization_id_or_slug": 1234}),
+                data={"querystring": "value"},
+                HTTP_HOST="albertos-apples.testserver",
+                follow=True,
+            )
```
markstory, line 421: In this scenario in saas, we'll end up going through the API gateway middleware and being proxied to the organization's region based on the `1234` slug and the subdomain will be ignored. I think that is a reasonable outcome as all slug subdomains resolve to control silo instances.

markstory (approved):

PR #70173, opened 2024-05-02
fix(alerts): Handle None when parsing event exceptions

snigdhas:
Fixes SENTRY-Y88


JoshFerge:
do we know the root cause of this error, and would it be possible to fix upstream as well?

JoshFerge (approved):

snigdhas:
Hm, I'm not sure what could cause the exception interface to be empty

PR #70184, opened 2024-05-02
ci: Use eslint to track calls to @deprecated functions/components

ryan953:
This creates a new workflow specifically for tracking calls to `@deprecated` functions in javascript.

It's going to run on a schedule, or on demand. It's not needed to run on every PR because the expectation is that there will be a long list of deprecations, and we should work over time to fix them up. If this were running on every PR it would desensitize people to lint problems so quickly, a really bad place to be.

Followup tasks that we don't need to start with but would be amazing:
- Change the output so it's grouped by `@deprecated` method, so we can see all related things at once
- Connect it to a slackbot command so we can use/share the results with the team more easily, especially the count of errors.

Example output: 
https://github.com/getsentry/sentry/actions/runs/8943695929/job/24568982206?pr=70184


.eslintrc.js
```diff
@@ -3,14 +3,21 @@
 module.exports = {
   root: true,
   extends: ['sentry-app/strict'],
+
+  parserOptions: {
+    project: './tsconfig.json',
+  },
```
scttcper, line 9: eh i'm not sure its worth it to start using the tsc backed parser for a bunch of warnings

scttcper:
image

these will just appear on every PR

.eslintrc.js
ryan953: thats fair. i think ideally these aren't enabled on every PR, because folks will just get desensitized to it. Instead I'd like a way to generate the list on-demand, but also have a number reported every week or something so we can track progress.

.github/workflows/frontend-lint-burndown.yml
```diff
@@ -0,0 +1,38 @@
+name: frontend-lint-burndown
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 4 * * 1'
+  pull_request:
```
asottile-sentry, line 7: we can't really afford to add more workflows / jobs as we're already right at our parallelism limit where more jobs will result in queueing behind itself.  if you can add these commands to an existing job that's a better approach (such as the existing linting job)

ryan953: i put the `pull_request` trigger there just to kick off the build and test it out. I def don't want this to be part of each PR.

ryan953: also updated the description to capture more of the thought process.

package.json
```diff
@@ -190,6 +190,7 @@
     "benchmark": "^2.1.4",
     "eslint": "8.57.0",
     "eslint-config-sentry-app": "2.7.0",
+    "eslint-plugin-deprecation": "^2.0.0",
```
billyvg, line 193: Do you need this in here if you are adding it in GHA step?

billyvg (commented):
An idea here could be to output eslint results as JSON, then use the `upload-artifact` action to save to GH. Future slackbot could then pull artifact down to parse and output.

package.json
ryan953: oops, yeah meant to remove this and the lockfile changes.

billyvg (approved):

PR #70187, opened 2024-05-02
fix(slack): remove backticks from tag values

cathteng:
Tag values are wrapped in backticks to make them show up as markdown `code`. If the tag value itself has backticks then it will break the markdown code formatting.

To fix this we can remove the backticks from the tag in the message. It shouldn't be too much of a disturbance.

### DIFF
--- a/src/sentry/integrations/slack/message_builder/base/block.py
+++ b/src/sentry/integrations/slack/message_builder/base/block.py
@@ -52,6 +52,8 @@ class BlockSlackMessageBuilder(SlackMessageBuilder, ABC):
         for tag in tags:
             title = tag["title"]
             value = tag["value"]
+            # remove backticks from value, otherwise it will break the markdown
+            value = value.replace("`", "")
             text += f"{title}: `{value}`  "
         return {
             "type": "section",
--- a/tests/sentry/integrations/slack/test_message_builder.py
+++ b/tests/sentry/integrations/slack/test_message_builder.py
@@ -101,6 +101,7 @@ def build_test_message_blocks(
         for k, v in tags.items():
             if k == "release":
                 v = format_release_tag(v, group)
+            v = v.replace("`", "")
             tags_text += f"{k}: `{v}`  "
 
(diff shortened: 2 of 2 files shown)


iamrajjoshi (approved):

src/sentry/integrations/slack/message_builder/base/block.py
```diff
@@ -52,6 +52,8 @@ def get_tags_block(tags) -> SlackBlock:
         for tag in tags:
             title = tag["title"]
             value = tag["value"]
+            # remove backticks from value, otherwise it will break the markdown
+            value = value.replace("`", "")
```
ykamo001, line 56: what if it's a code block?

cathteng: the tag value is a code block? it will end up being a code block again because we put it inside backticks

cathteng: oh you mean multiple backticks? doesn't this remove them all

ykamo001: if it ends up being a code block again then that's great

ykamo001 (approved):

src/sentry/integrations/slack/message_builder/base/block.py
ykamo001: also do we want to only replace them in the start and end? or all in general?

cathteng: this replaces them all, if there's any in the middle it will still break the markdown code block

ykamo001: code block is triple, code text is single

cathteng: we just want code text... it won't preserve the code block

cathteng: but also who is sending code blocks as tags

PR #70193, opened 2024-05-02
chore(api): Update List Your Organizations docs

isabellaenriquez:
Updates the OrganizationIndexEndpoint to use the modern API documentation format instead of the legacy format + updates example with fields added since the API was first published + add query params not currently included

Completes https://www.notion.so/sentry/Update-documentation-on-OrganizationIndexEndpoint-GET-a3ace41e39904326bb6f0aa4d216d6a8?pvs=4

localhost_3000_api_organizations_list-your-organizations_ (1)


src/sentry/api/endpoints/organization_index.py
```diff
@@ -48,28 +54,32 @@ def validate_agreeTerms(self, value):
...
 
 
+@extend_schema(tags=["Organizations"])
 @region_silo_endpoint
 class OrganizationIndexEndpoint(Endpoint):
     publish_status = {
-        "GET": ApiPublishStatus.UNKNOWN,
-        "POST": ApiPublishStatus.UNKNOWN,
+        "GET": ApiPublishStatus.PUBLIC,
+        "POST": ApiPublishStatus.PRIVATE,
```
isabellaenriquez, line 62: POST is private because it requires user context

src/sentry/apidocs/parameters.py
```diff
@@ -109,6 +109,44 @@ class OrganizationParams:
...
 - `/?project=1234&project=56789`
 - `/?project=-1`
+""",
+    )
+    OWNER = OpenApiParameter(
+        name="owner",
+        location="query",
+        required=False,
+        type=bool,
+        description="""Specify `true` to restricts results to organizations in which you are an owner.""",
```
schew2381, line 119:
```suggestion
        description="""Specify `true` to restrict results to organizations in which you are an owner.""",
```

src/sentry/apidocs/parameters.py
```diff
@@ -109,6 +109,44 @@ class OrganizationParams:
...
+    )
+    QUERY = OpenApiParameter(
+        name="query",
+        location="query",
+        required=False,
+        type=str,
+        description="""The search filter for your query, read more about query syntax [here](https://docs.sentry.io/product/sentry-basics/search/).
```
schew2381, line 126:
```suggestion
        description="""Filters results by using [query syntax](https://docs.sentry.io/product/sentry-basics/search/).
```

src/sentry/apidocs/parameters.py
```diff
@@ -109,6 +109,44 @@ class OrganizationParams:
...
+
+Valid query fields include:
+- `id`: The organization ID
+- `slug`: The organization slug
+- `status`: The organization's current status (one of `active`, `pending_deletion`, or `deletion_in_progress`)
+- `email` or `member_id`: Filter your organizations by specific members included
+- `platform`: Filter your organizations for those with a project that uses this platform
```
schew2381, line 133:
```suggestion
- `platform`: Filter your organizations to those with at least one project using this platform
```

src/sentry/apidocs/parameters.py
```diff
@@ -109,6 +109,44 @@ class OrganizationParams:
...
+        location="query",
+        required=False,
+        type=str,
+        description="""The search filter for your query, read more about query syntax [here](https://docs.sentry.io/product/sentry-basics/search/).
+
+Valid query fields include:
+- `id`: The organization ID
+- `slug`: The organization slug
+- `status`: The organization's current status (one of `active`, `pending_deletion`, or `deletion_in_progress`)
+- `email` or `member_id`: Filter your organizations by specific members included
```
schew2381, line 132: What ID is `member_id` referring to in this case? Is it the org member id or the user id?

schew2381 (commented):
small nits, one question

src/sentry/apidocs/parameters.py
isabellaenriquez: should i update the other endpoint docs that use this language?

src/sentry/apidocs/parameters.py
isabellaenriquez: org member id, i can specify that (and maybe link the API docs for retrieve an org member since there isn't anywhere in the UI that we surface these)

schew2381: If you query by member_id that would only return a single org right? But if you did it by email, that could return multiple?

src/sentry/apidocs/parameters.py
schew2381: Yeah that's a good idea, we should try to consolidate text for similar endpoints

isabellaenriquez: i'm gonna update the other param that uses this

schew2381 (approved):

PR #70197, opened 2024-05-02
ref(feedback): move user report event linking to post_process

JoshFerge:
currently we link user reports to groups in [event_manager](https://github.com/getsentry/sentry/blob/026e3cc5a6a1fb25ac0a9f6e474d060e91a75421/src/sentry/event_manager.py#L572). we'd like to move this to post_process to get it out of the main ingestion path.

This PR simply copies the code over from the event manager.

We'll do a follow up PR to delete the code from event manager after this has fully deployed, as to not miss any links.

### DIFF
--- a/src/sentry/tasks/post_process.py
+++ b/src/sentry/tasks/post_process.py
@@ -1401,6 +1401,39 @@ def check_has_high_priority_alerts(job: PostProcessJob) -> None:
         )
 
 
+def link_event_to_user_report(job: PostProcessJob) -> None:
+    from sentry.models.userreport import UserReport
+
+    event = job["event"]
+    project = event.project
+    group = event.group
+
+    if features.has(
+        "organizations:user-feedback-event-link-ingestion-changes", project.organization
+    ):
+        metrics.incr("event_manager.save._update_user_reports_with_event_link")
+        event = job["event"]
+        project = event.project
+        user_reports_without_group = UserReport.objects.filter(
+            project_id=project.id,
+            event_id=event.event_id,
+            group_id__isnull=True,
+            environment_id__isnull=True,
+        )
+
+        user_reports_updated = user_reports_without_group.update(
+            group_id=group.id, environment_id=event.get_environment().id
+        )
+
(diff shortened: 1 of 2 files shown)


src/sentry/tasks/post_process.py
```diff
@@ -1401,6 +1401,39 @@ def check_has_high_priority_alerts(job: PostProcessJob) -> None:
...
+            group_id=group.id, environment_id=event.get_environment().id
+        )
+
+        if user_reports_updated:
+            metrics.incr("event_manager.save._update_user_reports_with_event_link_updated")
+
+    else:
+        UserReport.objects.filter(project_id=project.id, event_id=job["event"].event_id).update(
+            group_id=group.id, environment_id=event.get_environment().id
+        )
```
vartec, line 1434: If overwriting all is fine, then why not just do that instead of trying to get null ones? 
AFAIK an update to same value as already in row should be a no-op in PostgreSQL.

vartec (approved):

src/sentry/tasks/post_process.py
```diff
@@ -1401,6 +1401,39 @@ def check_has_high_priority_alerts(job: PostProcessJob) -> None:
...
+            group_id__isnull=True,
+            environment_id__isnull=True,
+        )
+
+        user_reports_updated = user_reports_without_group.update(
+            group_id=group.id, environment_id=event.get_environment().id
+        )
+
+        if user_reports_updated:
+            metrics.incr("event_manager.save._update_user_reports_with_event_link_updated")
```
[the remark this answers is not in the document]
JoshFerge: it should be synonymous with call count as the number of times that an event is linked to more than one report is near 0

src/sentry/tasks/post_process.py
JoshFerge: that is what we originally did, but am about to make an update where we don't simply want to overwrite all, as we want only the reports that do not currently have a group associated https://github.com/getsentry/sentry/pull/70200

PR #70117, opened 2024-05-02
fix(feedback): add feedback shimming to event manager

JoshFerge:
If we determine it is okay to do so, we shim user reports to user feedback in the event_manager. 

depends on https://github.com/getsentry/sentry/pull/70116

PR #70200, opened 2024-05-02
feat(feedback): shim reports to feedbacks in post_process

JoshFerge:
Shim user reports to feedback here in post process. a re-do of https://github.com/getsentry/sentry/pull/70117 but not in ingestion hotpath

Follow up: instead of calling shim directly, create a celery task and call the celery task instead. so that way when we enable spam detection, any latency there will not affect post process workers


bruno-garcia (approved):

bruno-garcia:
feature flagged and tested. thank you Josh!

wedamija (approved):

PR #69830, opened 2024-04-26
ref(rules): Fire delayed rules

ceorourke:
Follow up to https://github.com/getsentry/sentry/pull/69167 to actually fire the rules

Closes https://github.com/getsentry/team-core-product-foundations/issues/242

PR #70203, opened 2024-05-02
ref(rules): Don't fire delayed rules if they're snoozed

ceorourke:
Follow up to https://github.com/getsentry/sentry/pull/69830#discussion_r1586723946 to check the `RuleSnooze` table before firing a delayed rule, on the off chance it got muted in the < 1 minute it took to process. 

Closes https://github.com/getsentry/team-core-product-foundations/issues/307

### DIFF
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -14,6 +14,7 @@ from sentry.models.group import Group
 from sentry.models.grouprulestatus import GroupRuleStatus
 from sentry.models.project import Project
 from sentry.models.rule import Rule
+from sentry.models.rulesnooze import RuleSnooze
 from sentry.rules import history, rules
 from sentry.rules.conditions.event_frequency import (
     BaseEventFrequencyCondition,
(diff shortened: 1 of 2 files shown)


src/sentry/rules/processing/delayed_processing.py
```diff
@@ -288,10 +289,14 @@ def apply_delayed(project_id: int) -> None:
         )
     # Step 7: Fire the rule's actions
     now = timezone.now()
-    # TODO: check rulesnooze table again before firing
     parsed_rulegroup_to_event_data = parse_rulegroup_to_event_data(rulegroup_to_event_data)
-
+    snoozed_rules = RuleSnooze.objects.filter(
```
saponifi3d, line 293: thoughts on moving this condition to step 3? I was thinking it if we filter the snoozed alert rules there (around line 272) it would mean we don't have to process the alerts in the other steps as well.

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -269,7 +270,10 @@ def apply_delayed(project_id: int) -> None:
 
     # STEP 3: Fetch the Rule models we need to check
     alert_rules = Rule.objects.filter(id__in=list(rules_to_groups.keys()))
-
+    snoozed_rules = RuleSnooze.objects.filter(rule__in=alert_rules, user_id=None).values_list(
+        "rule", flat=True
+    )
```
wedamija, line 275: Nit: Let's convert this to a set?

wedamija (approved):

src/sentry/rules/processing/delayed_processing.py
ceorourke: I'm about to do follow ups in this file so I'll change it there!

PR #70204, opened 2024-05-03
fix(feedback): check for undefined location query

michellewzhang:
prevent this from happening:

SCR-20240502-phkq

### DIFF
--- a/static/app/components/feedback/list/useMailboxCounts.tsx
+++ b/static/app/components/feedback/list/useMailboxCounts.tsx
@@ -25,13 +25,13 @@ export default function useMailboxCounts({
   organization,
 }: Props): UseApiQueryResult<HookReturnType, RequestError> {
   const location = useLocation();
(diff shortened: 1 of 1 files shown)


JoshFerge (approved):

static/app/components/feedback/list/useMailboxCounts.tsx
```diff
@@ -25,13 +25,13 @@ export default function useMailboxCounts({
   organization,
 }: Props): UseApiQueryResult<HookReturnType, RequestError> {
   const location = useLocation();
+  const locationQuery = location.query.query ?? '';
```
ryan953, line 28: decodeScalar is a good one to use here, it will handle the case where someone has typed `?query=foo&query=bar`

https://github.com/getsentry/sentry/blob/d647fbb02240343f843c1a6d90dda749e39a43f0/static/app/utils/queryString.tsx#L89-L103

PR #70205, opened 2024-05-03
fix(feedback): check event tags for replay ID

michellewzhang:
need to check an extra spot for the replay ID - see https://github.com/getsentry/sentry/blob/1665fdf89d94e86554494d5443a35fcd8464927a/static/app/utils/replays/getReplayIdFromEvent.tsx#L4-L5

### DIFF
--- a/static/app/components/feedback/feedbackItem/feedbackReplay.tsx
+++ b/static/app/components/feedback/feedbackItem/feedbackReplay.tsx
@@ -21,7 +21,10 @@ export default function FeedbackReplay({eventData, feedbackItem, organization}: 
   const {feedbackHasReplay} = useReplayCountForFeedbacks();
   const hasReplayId = feedbackHasReplay(feedbackItem.id);
 
-  const replayId = eventData?.contexts?.feedback?.replay_id;
+  // replay ID can be found in two places
+  const replayId =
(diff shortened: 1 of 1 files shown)


c298lee (approved):

static/app/components/feedback/feedbackItem/feedbackReplay.tsx
```diff
@@ -21,7 +21,10 @@ export default function FeedbackReplay({eventData, feedbackItem, organization}:
   const {feedbackHasReplay} = useReplayCountForFeedbacks();
   const hasReplayId = feedbackHasReplay(feedbackItem.id);
 
-  const replayId = eventData?.contexts?.feedback?.replay_id;
+  // replay ID can be found in two places
+  const replayId =
+    eventData?.contexts?.feedback?.replay_id ??
+    eventData?.tags?.find(({key}) => key === 'replayId')?.value;
```
ryan953, line 27: similar but different to https://github.com/getsentry/sentry/blob/d647fbb02240343f843c1a6d90dda749e39a43f0/static/app/utils/replays/getReplayIdFromEvent.tsx#L3-L8

ryan953 (approved):

PR #70207, opened 2024-05-03
chore(slack): Remove block kit flag code

ceorourke:
Remove usages of the "organizations:slack-block-kit" flag - this doesn't completely remove the flag because we still need to update some getsentry tests.

### DIFF
--- a/src/sentry/integrations/slack/webhooks/action.py
+++ b/src/sentry/integrations/slack/webhooks/action.py
@@ -11,7 +11,7 @@ from rest_framework import serializers
 from rest_framework.request import Request
 from rest_framework.response import Response
 
-from sentry import analytics, features
+from sentry import analytics
 from sentry.api import client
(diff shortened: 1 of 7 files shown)


src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -639,22 +615,21 @@ def _handle_group_actions(
...
-                        "unresolved:ongoing",
-                    )  # TODO: delete the first two names when block kit is GA
+                if action.name in (
+                    "status",
+                    "ignored:forever",
+                    "ignored:until_escalating",
+                    "unresolved:ongoing",
                 ):
+                    # check to see which action names are in use
+                    logger.info("slack.action.name", extra={"name": action.name})
```
ceorourke, line 625: I wasn't really sure if it was safe to remove the ignored ones, so I'm adding a logger and we can check if it's ever sent.

PR #70213, opened 2024-05-03
temporary(dynamic-sampling): Log outgoing envelopes that belong to boost low volume projects

iambriccardo:
This PR adds logging to outgoing envelopes for specific transactions because we are investigating problems with trace headers.

### DIFF
--- a/src/sentry/utils/sdk.py
+++ b/src/sentry/utils/sdk.py
@@ -226,6 +226,13 @@ def before_send_transaction(event: Event, _: Hint) -> Event | None:
     ):
         return None
 
+    # This code is added only for debugging purposes, as such, it should be removed once the investigation is done.
+    if event.get("transaction") in {
+        "sentry.dynamic_sampling.boost_low_volume_projects_of_org",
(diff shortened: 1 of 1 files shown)


src/sentry/utils/sdk.py
```diff
@@ -226,6 +226,13 @@ def before_send_transaction(event: Event, _: Hint) -> Event | None:
     ):
         return None
 
+    # This code is added only for debugging purposes, as such, it should be removed once the investigation is done.
+    if event.get("transaction") in {
+        "sentry.dynamic_sampling.boost_low_volume_projects_of_org",
+        "sentry.dynamic_sampling.tasks.boost_low_volume_projects",
+    }:
```
iambriccardo, line 233: Regarding performance, both transactions happen every 10 minutes, so it should not be a concern.

antonpirker: Yes, this looks OK!

antonpirker (approved):
lgtm

PR #70217, opened 2024-05-03
feat(cardinality-limit): Improve validation

### LINKED ISSUE #69750: Validate project cardinality limit option using rust validation functions

ArthurKnaus:
Add upper limit for cardinality limit value.
Add tests that ensures that the generated limit is normalized.

Closes https://github.com/getsentry/sentry/issues/69750


tests/sentry/api/endpoints/test_project_details.py
```diff
@@ -762,6 +766,11 @@ def test_custom_metrics_cardinality_limit(self):
                 }
             }
         ]
+
+        limit = config[0]["limit"]
+        normalized_limit = normalize_cardinality_limit_config(limit)
+        assert normalized_limit == limit
```
shellmayr, line 772: If normalize_cardinality_limit_config does not change anything, and we assert that they are the same before and after, why do we even call it? If we know it doesn't do anything, why do we assert?

shellmayr (approved):
LGTM

src/sentry/api/endpoints/project_details.py
```diff
@@ -264,9 +264,18 @@ def validate_relayPiiConfig(self, value):
...
     def validate_relayCustomMetricCardinalityLimit(self, value):
-        if value is not None and value < 0:
+        if value is None:
+            return value
+
+        if value < 0:
             raise serializers.ValidationError("Cardinality limit must be a non-negative integer.")
 
+        # Value is stored as uint32 in relay
+        if value > 4_000_000_000:
```
iambriccardo, line 274: Why are we hardcoding this number here? Seems to be a place where divergence can happen very easily. Can't we validate the config entirely via the `sentry-relay` function in a place where we have access to the actual configuration we want to store?

iambriccardo, line 274: Why are we hardcoding this number here? Seems to be a place where divergence can happen very easily. Can't we validate the config entirely via the `sentry-relay` function in a place where we have access to the actual configuration we want to store?

iambriccardo (commented):
Left a minor comment

src/sentry/api/endpoints/project_details.py
ArthurKnaus: hmm, i get you point. Since the API consumer only sends a number I would not call the whole cardinality limit normalizer in the serializer. Creating the object is part of our endpoint logic and should be checked by our tests.

For validating the number, the cleanest option would probably be to export another validator / a constant for the max value from relay 🤔 

Is it fine for you to merge the current state with a TODO? I would like to sync on that with you when I am back from vacation.

src/sentry/api/endpoints/project_details.py
```diff
@@ -264,9 +264,19 @@ def validate_relayPiiConfig(self, value):
...
-        if value is not None and value < 0:
+        if value is None:
+            return value
+
+        if value < 0:
             raise serializers.ValidationError("Cardinality limit must be a non-negative integer.")
 
+        # Value is stored as uint32 in relay
+        # TODO: find a way to share this constant between relay and sentry
+        if value > 4_000_000_000:
```
asottile-sentry, line 275: why 4 million instead of the actual uint32 limit?

ArthurKnaus: Mostly for the error message being nice to read. But agree from a code perspective it is just more confusing.
Will change to the uint32 limit 👍

PR #70223, opened 2024-05-03
feat: Update onboarding docs for v8

mydea:
In preparation for v8.

Missing is angular, cc @Lms24 feel free to add this to this PR or make a dedicated one!

We should merge this once v8 is released (probably monday).


Lms24:
@mydea added Angular docs updates and pinned capacitor/ionic framework SDK versions to v7

static/app/gettingStartedDocs/javascript/angular.tsx
```diff
@@ -74,44 +49,45 @@ const getNextStep = (
...
 
-const onboarding: OnboardingConfig<PlatformOptions> = {
-  install: (params: Params) => [
+const onboarding: OnboardingConfig = {
+  install: () => [
     {
       type: StepType.INSTALL,
       description: tct(
-        'Add the Sentry SDK as a dependency using [codeNpm:npm] or [codeYarn:yarn]:',
+        'Add the Sentry SDK as a dependency using [codeNpm:npm], [codeYarn:yarn] or [codePnpm:pnpm]:',
```
s1gr1d, line 83: Should we add `pnpm` everywhere?

mydea: let's look at this later, but possibly!

Lms24: +1 to tackle this in a follow-up PR but I'd vote yes, generally

static/app/gettingStartedDocs/javascript/nextjs.tsx
```diff
@@ -182,8 +188,13 @@ const replayOnboarding: OnboardingConfig = {
         <Fragment>
           <TracePropagationMessage />
           {tct(
-            'Alert: The Replay integration must be added to your [sentryClient:sentry.client.config.js] file. Adding it into [sentryServer:sentry.server.config.js] or [sentryEdge:sentry.edge.config.js] may break your build.',
-            {sentryClient: <code />, sentryServer: <code />, sentryEdge: <code />}
+            'Alert: The Replay integration must be added to your [sentryClient:sentry.client.config.js] file. Adding it to any server-side configuration files (like [instrumentation:instrumentation.ts]) will break your build because the Replay integration depends on Browser APIs.',
+            {
+              sentryClient: <code />,
+              sentryServer: <code />,
+              sentryEdge: <code />,
```
andreiborza, line 195: m: I believe these can be removed now.

static/app/gettingStartedDocs/javascript/nextjs.tsx
```diff
@@ -231,9 +242,22 @@ const feedbackOnboarding: OnboardingConfig = {
...
-        link: 'https://docs.sentry.io/platforms/javascript/guides/nextjs/user-feedback/#crash-report-modal',
-      }),
+      additionalInfo: (
+        <Fragment>
+          {tct(
+            'Alert: The User Feedback integration must be added to your [sentryClient:sentry.client.config.js] file. Adding it to any server-side configuration files (like [instrumentation:instrumentation.ts]) will break your build because the Replay integration depends on Browser APIs.',
+            {
+              sentryClient: <code />,
+              sentryServer: <code />,
+              sentryEdge: <code />,
```
andreiborza, line 252: m: These can also be removed now right? No mention of these in the string.

static/app/utils/gettingStartedDocs/node.ts
```diff
@@ -124,6 +136,66 @@ export function getDefaulServerlessImports({
...
+    params.isProfilingSelected
+      ? `integrations: [
+    nodeProfilingIntegration(),
+  ],`
+      : ''
+  }${
+    params.isPerformanceSelected
+      ? `
+      // Performance Monitoring
+      tracesSampleRate: 1.0, //  Capture 100% of the transactions`
```
Lms24, line 187: l/feel free to disregard for now: I think we should remove all references to "transactions" in SDK-specific documentation:

```suggestion
      tracesSampleRate: 1.0, //  Capture 100% of spans`
```

Now that being said, I'm realizing that capturing 100% of spans is also misleading because the sample rate is only applied to root/segment spans 😢 

It's not too important though and we should also iterate to get this first set of changes in today. So we can just follow up on this later :)

(leaving this comment here for context)

static/app/utils/gettingStartedDocs/node.ts
```diff
@@ -124,6 +136,66 @@ export function getDefaulServerlessImports({
...
+node --require ./instrument.js app.js
+
+# If you are using ECMAScript Modules (ESM) - Note that this is only available from Node 18.19.0 onwards.
+node --import ./instrument.mjs app.mjs`;
+}
+
+export function getImportInstrumentSnippet(): string {
+  return `// IMPORTANT: Make sure to import \`instrument.js'\` at the top of your file.
+  // If you're using ECMAScript Modules (ESM), use \`import './instrument.js';\`
+  require('./instrument.js');`;
```
Lms24, line 150: l-m: I'm afraid this conflicts a bit with users' interpretation of when they're using ESM to write vs to run. I have a couple of ideas but feel free to come up with your own or table this for now and merge anyway:

```suggestion
  // If you're using ECMAScript Modules (ESM) syntax, use \`import './instrument.js';\`:
  require('./instrument.js');`;
```


or

```suggestion
  // If you're `import` syntax, replace this with \`import './instrument.js';\`:
  require('./instrument.js');`;
```

static/app/utils/gettingStartedDocs/node.ts
```diff
@@ -124,6 +136,66 @@ export function getDefaulServerlessImports({
   return imports;
 }
 
+export function getNodeRunCommandSnippet(): string {
+  return `# If you are using CommonJS (CJS)
+node --require ./instrument.js app.js
```
andreiborza, line 141: m: I thought we got rid of this and only suggest `--import` in case of ESM.

andreiborza (changes requested):

Lms24 (approved):
LGTM! I have one concern around our ESM wording and another low prio one RE transactions but other than that I think this is great! 

I think we should get this in today, so all my remarks can be adressed in follow-up PRs. Nothing blocking :)

static/app/utils/gettingStartedDocs/node.ts
s1gr1d: This is an unused function right now. I can delete it.

static/app/utils/gettingStartedDocs/node.ts
s1gr1d: Good suggestion! I can make it more clear that this is about syntax.

andreiborza (approved):
LGTM

PR #70231, opened 2024-05-03
fix(replays): Remove project-id as an aggregation key

cmanallen:
This fixes a bug where the details page would show errors for project's associated with the replay but not other projects (e.g. back-end errors).  This pull does not fix the index page which suffers from a similar problem.


JoshFerge (approved):

JoshFerge:
curious if this will affect perf at all

cmanallen:
@JoshFerge replay_id is indexed so I think we'll probably be fine.

PR #70235, opened 2024-05-03
ref(starfish): create a shared useSeries function

DominikB2014:
1. Create shared `useSeries` which support `useSpanMetricsSeries` and `useMetricsSeries` where SpanMetrics and Metrics are the dataset names.
2. Rename `MetricsResponse` to `SpanMetricsResponse`, `MetricsProperty` to `SpanMetricsProperty`, etc so that they match the datasetname
3. Create `MetricsResponse`, `MetricsProperty` etc for fields in the Metrics dataset


static/app/views/performance/database/getIntervalForMetricFunction.tsx
```diff
@@ -56,4 +56,4 @@ function registerGranularities(
 
 registerGranularities(COUNTER_AGGREGATES, COUNTER_GRANULARITIES);
 registerGranularities(DISTRIBUTION_AGGREGATES, DISTRIBUTION_GRANULARITIES);
-registerGranularities(SPAN_FUNCTIONS, COUNTER_GRANULARITIES);
+registerGranularities(METRICS_FUNCTIONS, COUNTER_GRANULARITIES);
```
edwardgou-sentry, line 59: Does this change any functionality that others might be using? Because the `SPAN_FUNCTIONS` array contains a lot more elements than `METRICS_FUNCTIONS`. Maybe we should register both arrays?

edwardgou-sentry (approved):
Probably not for this pr, but would be nice to eventually get out of the starfish directory

static/app/views/performance/database/getIntervalForMetricFunction.tsx
DominikB2014: Ohh, this was an unintended change, let me change it back! 👍

static/app/views/starfish/queries/useMetricsSeries.ts
gggritso, line 1: Should this be called `useSeries.ts`?

static/app/views/starfish/queries/useSpanMetrics.tsx
```diff
@@ -50,14 +50,15 @@ function getEventView(
   search: MutableSearch | undefined,
   fields: string[] = [],
   sorts: Sort[] = [],
-  pageFilters: PageFilters
+  pageFilters: PageFilters,
+  dataset: DiscoverDatasets = DiscoverDatasets.SPANS_METRICS
```
gggritso, line 54: 🤔 why do this, if this is the `useSpanMetrics` hook? It should always be set to use the `SPANS_METRICS` dataset, no?

gggritso (commented):
👍🏻

static/app/views/starfish/queries/useSpanMetrics.tsx
DominikB2014: Yeah you're right, this wasn't supposed to be here tbh, i was starting to do this for the non series hooks but that will be in another PR, i'll revert it.

static/app/views/starfish/queries/useMetricsSeries.ts
DominikB2014: yes 👍

gggritso (approved):

PR #70238, opened 2024-05-03
feat(seer-grouping): Add factory method to `SeerSimilarIssueData` class

lobsterkatie:
This adds a factory method, `from_raw`, to the `SeerSimilarIssueData` class, which converts `RawSeerSimilarIssueData` instances into instances of `SeerSimilarIssueData`. During conversion, it also a) validates that at least one of parent hash and parent group id has been provided, and b) looks up the parent group id if only a parent hash is present.

### DIFF
--- a/src/sentry/seer/utils.py
+++ b/src/sentry/seer/utils.py
@@ -17,6 +17,14 @@ class SeerException(Exception):
     pass
 
 
+class IncompleteSeerDataError(Exception):
+    pass
+
+
+class SimilarGroupNotFoundError(Exception):
+    pass
+
+
 class BreakpointData(TypedDict):
     project: str
     # For legacy reasons, the group name is always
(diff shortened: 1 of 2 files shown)


src/sentry/seer/utils.py
```diff
@@ -117,6 +126,35 @@ class SeerSimilarIssueData:
     # TODO: See if we end up needing the hash here
     parent_group_hash: str | None = None
 
+    @classmethod
+    def from_raw(cls, project_id: int, raw_similar_issue_data: RawSeerSimilarIssueData) -> Self:
```
jangjodi, line 130: can we add a quick docstring here?

jangjodi (approved):

src/sentry/seer/utils.py
lobsterkatie: Good point. I've gotten so lax now that the linter doesn't yell at me to put a docstring on every function! Will do.

PR #70240, opened 2024-05-03
ref(seer-grouping): Automatically convert `RawSeerSimilarIssueData` into `SeerSimilarIssueData`

lobsterkatie:
This switches `get_similar_issues_embeddings` from returning a list of `RawSeerSimilarIssueData` instances to returning a list of `SeerSimilarIssueData` instances, automatically converting the former to the latter along the way.

The primary difference between the two types is that while the raw version might include the parent group hash rather than the parent group id, the non-raw version is guaranteed to have the parent group id, looked up from the parent group hash if that's all that's sent back from Seer. This means that once this merges, we'll be safe to only send the parent group hash back from Seer, knowing the the parent group id will get filled in in the process of turning the `RawSeerSimilarIssueData` instances into `SeerSimilarIssueData` instances.

Note that this change required switching the code calling `get_similar_issues_embeddings` from using dictionary syntax to using class syntax (because `RawSeerSimilarIssueData` is a typeddict and `SeerSimilarIssueData` is a dataclass). 

To test this change, the "test simple" version of both the similar issues endpoint and `get_similar_issues_embeddings` tests have been replicated so that they now test all three of the possible return options from Seer - only parent group id, only parent group hash, and both - to show that in all cases, the logic which follows gets the data it needs.

### DIFF
--- a/src/sentry/seer/utils.py
+++ b/src/sentry/seer/utils.py
@@ -174,7 +174,7 @@ class SeerSimilarIssueData:
 
 def get_similar_issues_embeddings(
     similar_issues_request: SimilarIssuesEmbeddingsRequest,
-) -> list[RawSeerSimilarIssueData]:
+) -> list[SeerSimilarIssueData]:
     """Call /v0/issues/similar-issues endpoint from seer."""
     response = seer_staging_connection_pool.urlopen(
         "POST",
@@ -199,4 +199,22 @@ def get_similar_issues_embeddings(
         )
         return []
 
-    return response_data.get("responses") or []
+    normalized = []
+
+    for raw_similar_issue_data in response_data.get("responses") or []:
+        try:
+            normalized.append(
+                SeerSimilarIssueData.from_raw(
+                    similar_issues_request["project_id"], raw_similar_issue_data
+                )
+            )
+        except (IncompleteSeerDataError, SimilarGroupNotFoundError) as err:
+            logger.exception(
+                str(err),
+                extra={
+                    "request_params": similar_issues_request,
+                    "raw_similar_issue_data": raw_similar_issue_data,
+                },
+            )
+
+    return normalized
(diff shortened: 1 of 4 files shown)


src/sentry/seer/utils.py
```diff
@@ -183,4 +183,22 @@ def get_similar_issues_embeddings(
         )
         return []
 
-    return response_data.get("responses") or []
+    rv = []
```
jangjodi, line 186: what does rv stand for? can we give it a more descriptive name?

src/sentry/seer/utils.py
```diff
@@ -183,4 +183,22 @@ def get_similar_issues_embeddings(
...
+
+    for raw_similar_issue_data in response_data.get("responses") or []:
+        try:
+            rv.append(
+                SeerSimilarIssueData.from_raw(
+                    similar_issues_request["project_id"], raw_similar_issue_data
+                )
+            )
+        except (IncompleteSeerDataError, SimilarGroupNotFoundError) as err:
+            logger.exception(
```
jangjodi, line 196: should we add some tests for these exceptions and logger?

src/sentry/seer/utils.py
lobsterkatie: Sigh. I've become one of them, LOL. 

It stands for "return value" and you'll see it used in tons of places in our code because it's both Armin and Markus's go-to. But I had to ask, too, way back when, so I'll figure out a better generic name.

src/sentry/seer/utils.py
lobsterkatie: That's the part that codecov was complaining about. I realized there was a bug in my factory method (which I've gone back and fixed) which is why those lines weren't running and codecov was showing them as uncovered. So score one for codecov in this case!

As for testing the logging - I was torn about whether or not that was overkill, but I can.

jangjodi (approved):

PR #70005, opened 2024-04-30
ref(grouping): Small seer similar issues fixes

lobsterkatie:
This is a handful of small fixes pulled out of a few upcoming PRs, in order to make them easier to review.

Changes:

- Switch from using inheritance and `total=False` to using `NotRequired` for optional properties of `SimilarIssuesEmbeddingsRequest`, so everything can be in one class.
- Add exception types thrown by deserialization to the `try-except` block surrounding the parsing of seer response data, and add a log to track if/when things go wrong. Also simplify returning an empty result-set from the `except` block.
- Simplify a few places where we were using `update` to add a single key-value pair to a dictionary.
- Remove an unnecessary check for the existence of entries in the seer result-set. This should have been part of https://github.com/getsentry/sentry/pull/69993, but it got missed.
- Rename `response`/`responses` in  `get_formatted_results` to `similar_issue_data`/`similar_issues_data` in preparation for changing some seer data datatypes.
- Add a TODO to track the fact that we need to at some point let seer know when it's recommending groups which no longer exist.
- Fix the data in a `get_similar_issues_embeddings` test to match the `SimilarIssuesEmbeddingsData` type. (It was given in terms of message and stacktrace similarity rather than distance.)
- Add types to similar issues endpoint helpers.
- Prepend `get_value_if_exists` with an underscore since it's an internal helper.

PR #70244, opened 2024-05-03
ref(seer-grouping): Send group hash to Seer

lobsterkatie:
This adds the group hash to the outgoing payload sent to Seer. Once this is merged, it will be safe to remove the Seer logic handling an incoming group id, since we'll now be able to rely on the group hash being present. 

Taken together, this PR and the ones leading up to it (links below) mean will mean that on the Sentry side we'll be sending in the request and handling in the response both hash and group id (and transforming hashes in the response into group ids if hashes are all that's sent). Once Seer is adjusted to only deal in hashes, we can then remove the sending and handling of group ids from Sentry and the transition will be complete.

Other PRs which are part of the group-id-to-group-hash switch:
- https://github.com/getsentry/sentry/pull/70005, https://github.com/getsentry/sentry/pull/70236, and https://github.com/getsentry/sentry/pull/70237 - various small fixes and tweaks
- https://github.com/getsentry/sentry/pull/70070 and https://github.com/getsentry/sentry/pull/70238 - updates to associated types
- https://github.com/getsentry/sentry/pull/70240 - automatic conversion from hash to group id when handling Seer similar group data

### DIFF
--- a/tests/sentry/seer/test_utils.py
+++ b/tests/sentry/seer/test_utils.py
@@ -80,6 +80,7 @@ def test_simple_similar_issues_embeddings_only_group_id_returned(
 
     params: SimilarIssuesEmbeddingsRequest = {
         "group_id": NonNone(event.group_id),
+        "group_hash": NonNone(event.get_primary_hash()),
         "project_id": default_project.id,
         "stacktrace": "string",
         "message": "message",
@@ -106,6 +107,7 @@ def test_simple_similar_issues_embeddings_only_hash_returned(mock_seer_request, 
 
     params: SimilarIssuesEmbeddingsRequest = {
         "group_id": NonNone(event.group_id),
+        "group_hash": NonNone(event.get_primary_hash()),
         "project_id": default_project.id,
         "stacktrace": "string",
         "message": "message",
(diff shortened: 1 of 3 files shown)


tests/sentry/seer/test_utils.py
```diff
@@ -156,6 +156,7 @@ def test_empty_similar_issues_embeddings(mock_seer_request, default_project):
 
     params: SimilarIssuesEmbeddingsRequest = {
         "group_id": NonNone(event.group_id),
+        "group_hash": NonNone(event.get_primary_hash()),
```
jangjodi, line 159: just to confirm - we're changing this (and all other `group_hash` occurrences) to just `hash` in a follow-up PR and accepting that the similarity endpoint will be a bit broken until that happens?

lobsterkatie: On the name change, yes, that's the next PR to come. As for things being broken... I thought we said nothing was currently relying on `group_hash`? In any case, I'm happy to hold off on merging this until the rename is ready.

jangjodi: yeah nothing's relying on `group_hash` but we're also not expecting it as part of the request right now on the seer side (my PR only will accept `hash).` I can make that change if needed.

lobsterkatie: Will it error out or just ignore it? (I was assuming the latter.) And Seer's not yet requiring the hash, correct?

jangjodi: oh good point, I think you're right. are you able to test it with the current seer implementation (since we're also not accepting `group_hash`)? 

and correct, seer is not requiring the hash yet.

jangjodi: [update]: seer just ignores the unneeded parameter

jangjodi (approved):

PR #70247, opened 2024-05-03
fix(search): fixes short id lookup logic

scefali:
This PR fixes a bug where putting a short id as part of a query like `issue:SENTRY-38VM` actually fails the postgres direct hit search and doesn't redirect the user even though it limits the results to the single result. However, just searching `SENTRY-38VM` works as expected. The problem is the regex for the project slug matches `issue:SENTRY` instead of `SENTRY`.

### DIFF
--- a/src/sentry/models/group.py
+++ b/src/sentry/models/group.py
@@ -67,7 +67,7 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-_short_id_re = re.compile(r"^(?:[^:]+:)?(.*?)(?:[\s_-])([A-Za-z0-9]+)$")
+_short_id_re = re.compile(r"^(?:issue+:)?(.*?)(?:[\s_-])([A-Za-z0-9]+)$")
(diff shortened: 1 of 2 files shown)


src/sentry/models/group.py
```diff
@@ -67,8 +67,7 @@
 
 logger = logging.getLogger(__name__)
 
-_short_id_re = re.compile(r"^(.*?)(?:[\s_-])([A-Za-z0-9]+)$")
-
+_short_id_re = re.compile(r"^(?:[^:]+:)?(.*?)(?:[\s_-])([A-Za-z0-9]+)$")
```
wedamija, line 70: Should this explicitly look for either `<short_id>` or `issue:<short_id>`? Or is the `issue:` already stripped off here?

scefali: @wedamija yea, probably actually. Was thinking maybe people were doing `issue_id` or something as well an needed to capture that but probably not.

wedamija (approved):

PR #70252, opened 2024-05-03
fix(mobile-ui): Frames delay rendering in table

narsaynorath:
Frame delay isn't self-reporting its units as seconds, so handle it in the frontend until we fix this.

Before:
Screenshot 2024-05-03 at 1 00 20 PM

After:
Screenshot 2024-05-03 at 12 59 49 PM


0Calories (approved):
:shipit:

PR #70255, opened 2024-05-03
ref: ensure type of deserialized model is a sentry model

asottile-sentry:
fixes a type error pointed out by upgrading django-stubs:

```
src/sentry/services/hybrid_cloud/import_export/impl.py:203: error: "Model" has no attribute "get_possible_relocation_scopes"  [attr-defined]
src/sentry/services/hybrid_cloud/import_export/impl.py:223: error: "Model" has no attribute "normalize_before_relocation_import"  [attr-defined]
src/sentry/services/hybrid_cloud/import_export/impl.py:233: error: "Model" has no attribute "get_relocation_scope"  [attr-defined]
src/sentry/services/hybrid_cloud/import_export/impl.py:239: error: "Model" has no attribute "write_relocation_import"  [attr-defined]
```

### DIFF
--- a/src/sentry/services/hybrid_cloud/import_export/impl.py
+++ b/src/sentry/services/hybrid_cloud/import_export/impl.py
@@ -26,6 +26,7 @@ from sentry.backup.dependencies import (
 from sentry.backup.findings import InstanceID
 from sentry.backup.helpers import EXCLUDED_APPS, DatetimeSafeDjangoJSONEncoder, Filter, ImportFlags
 from sentry.backup.scopes import ExportScope
+from sentry.db.models.base import BaseModel
 from sentry.models.importchunk import ControlImportChunk, RegionImportChunk
 from sentry.models.user import User
 from sentry.models.userpermission import UserPermission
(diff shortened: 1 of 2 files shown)


src/sentry/services/hybrid_cloud/import_export/impl.py
```diff
@@ -198,15 +199,24 @@ def import_by_model(
                 last_seen_ordinal = min_ordinal - 1
                 for deserialized_object in deserialize("json", json_data, use_natural_keys=False):
                     model_instance = deserialized_object.object
+                    inst_model_name = get_model_name(model_instance)
+
+                    if not isinstance(model_instance, Model):
+                        return RpcImportError(
+                            kind=RpcImportErrorKind.UnexpectedModel,
+                            on=InstanceID(model=str(inst_model_name), ordinal=1),
```
azaslavsky, line 207: nit: imo `ordinal` should be `None` here.

azaslavsky (approved):
Ty for the touch up!

azaslavsky (commented):
Changed approval to comment to block auto-merge.

src/sentry/services/hybrid_cloud/import_export/impl.py
asottile-sentry: should the other one below it also be `None` ?

azaslavsky: Yeah, it makes sense for them too match. Either way is honestly fine.

PR #70260, opened 2024-05-03
ref: remove impossible checks for Event from post_process

asottile-sentry:
after upgrading django-stubs, the type checker helpfully points out that `job['event']` is always `GroupEvent` and so an `isinstance(job['event'], Event)` check will always fail


wedamija (approved):
Good cleanup, thanks!

PR #64825, opened 2024-02-07
Revert "Revert "feat(crons): Allow max_workers to be configured for consumer (#64512)""

wedamija:
This reverts commit a8bbd68b4434632a4869764e1735f82cef99ab6f.

I reverted this yesterday out of an abundance of caution, but it didn't seem to be actually causing issues. This is unchanged from https://github.com/getsentry/sentry/pull/64512

PR #70261, opened 2024-05-03
feat(crons): Add max-workers option

evanpurkhiser:
Follow up to GH-64825


wedamija (approved):
Nice, we should also follow up with ops and reduce our max workers in prod, which should hopefully fix some weird connection spikes we're seeing

Issue #69827, opened 2024-04-26
Issues Stream: Update assignee selector to use new design


malwilley:
After the work in https://github.com/getsentry/sentry/issues/69054 we will have a new assignee selector that looks similar to the old one but with a new dropdown component.

The next step will be to update the trigger to match the new pill design, as pictured below. The component should receive a prop that will either display the trigger with or without text. For the first iteration, we will not display the assignee text on the issue stream.

![CleanShot 2024-04-26 at 15 59 36](https://github.com/getsentry/sentry/assets/10888943/0f310c3a-9549-4086-a50e-3cefa647de06)

MichaelSun48:
Completed in #70991

PR #70269, opened 2024-05-03
feat(tags): Adds border customizability to <Tag/>

MichaelSun48:
This PR adds a `borderStyle` prop to the `<Tag/>` component. It is set to `solid` by default, since that is what was the only option before. 

This will be needed in upcoming changes to the Assignee dropdown trigger on the Issue details page, detailed [here](https://github.com/getsentry/sentry/issues/69827)

### DIFF
--- a/static/app/components/badge/tag.tsx
+++ b/static/app/components/badge/tag.tsx
@@ -18,7 +18,7 @@ import type {Color} from 'sentry/utils/theme';
 import theme from 'sentry/utils/theme';
 
 export interface TagProps extends React.HTMLAttributes<HTMLSpanElement> {
-  borderStyle?: 'solid' | 'dashed' | 'dotted';
+  borderStyle?: 'solid' | 'dashed' | 'dotted' | 'double' | 'none';
   /**
    * Makes the tag clickable. Use for external links.
    * If no icon is passed, it defaults to IconOpen (can be removed by passing icon={null})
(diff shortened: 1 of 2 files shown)


static/app/components/badge/tag.tsx
```diff
@@ -18,6 +18,7 @@ import type {Color} from 'sentry/utils/theme';
 import theme from 'sentry/utils/theme';
 
 export interface TagProps extends React.HTMLAttributes<HTMLSpanElement> {
+  borderStyle?: 'solid' | 'dashed' | 'dotted' | 'double' | 'none';
```
MichaelSun48, line 21: There are [more](https://www.w3schools.com/css/css_border.asp) border styles than this, but I just added the 2d ones that I thought were relevant - feel free to correct me if there's a better way to do this

MichaelSun48:
Is there a good way to write a test for this? I couldn't get `.toHaveStyle('border: hashed')`, for example, to work.

ryan953 (commented):
A note in the story could be useful https://sentry.sentry.io/stories/?name=app/components/badge/tag.stories.tsx

static/app/components/badge/tag.tsx
MichaelSun48: Removed double because it doesn't appear to do anything. 
Removed none because it just becomes text at that point

MichaelSun48:
Updated stories page:
image

ryan953 (approved):

PR #70272, opened 2024-05-03
ref(starfish): create shared useDiscover function

DominikB2014:
This is an extension of https://github.com/getsentry/sentry/pull/70235
This PR does the same thing as that one, except for non-series metrics data.

🤔 I wonder if we should handle series data in the same hook? The typing will probably get even more complicated, so i think having it separate is better. We also have seperate endpoints for series data vs regular data, so this matches that idea as well.

Next steps would be added the indexed span, and indexed transaction stuff here too


static/app/views/starfish/queries/useMetrics.ts
```diff
@@ -18,13 +23,29 @@ interface UseSpanMetricsOptions<Fields> {
...
+  );
+};
+
+export const useMetrics = <Fields extends MetricsProperty[]>(
+  options: UseMetricsOptions<Fields> = {}
+) => {
+  return useMetricsBase<Fields, MetricsResponse>(options, DiscoverDatasets.METRICS);
+};
+
+const useMetricsBase = <T extends Extract<keyof ResponseType, string>[], ResponseType>(
```
DominikB2014, line 40: The naming here is a bit akward, was going for `use<Dataset>`, but i think specifying `base` suffix is a good workaround

gggritso (commented):
There's a weird asymmetry here 🤔 if there's a `useMetricsSeries` that implies that there should be a `useMetrics`, except of course that hook already exists. The series and non-series hooks should match. Maybe `useDiscover` and `useDiscoverSeries`? Or, `useSpanMetrics`, `useTransactionMetrics`, and so on, maybe.

Do you know what I mean? The names should all match and be easy to predict/guess

DominikB2014:
> There's a weird asymmetry here 🤔 if there's a `useMetricsSeries` that implies that there should be a `useMetrics`, except of course that hook already exists. The series and non-series hooks should match. Maybe `useDiscover` and `useDiscoverSeries`? Or, `useSpanMetrics`, `useTransactionMetrics`, and so on, maybe.
> 
> 
> 
> Do you know what I mean? The names should all match and be easy to predict/guess

Yeah I know what u mean! I'm liking useDiscover and useDiscoverSeries, with hooks that build off of these for each dataset. I like it more because those base hooks (useDiscover, useDiscoverSeries) would map better to what the actual api is.

DominikB2014:
@gggritso so i've updated the other hook and file names to be `useDiscoverSeries`

gggritso (approved):
👍🏻 nice! Let's work more on this soon, and resolve the overlap with `useDiscoverQuery` and `useGenericDiscoverQuery`

Issue #69931, opened 2024-04-29
Hover state for checkbox on similar issues tab is weird


mrduncan:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

Go to the similar issues tab on an issue which has similar issues (ex: https://sentry.sentry.io/issues/5225041747/similar/?project=1).

Note for Sentry employees: you can ignore the banner at the top of this page, this checkbox looks the same for external users too.

### Expected Result

Hovering over the checkboxes doesn't have a weird full-height effect.

### Actual Result

![Screenshot 2024-04-29 at 2 41 20 PM](https://github.com/getsentry/sentry/assets/14761/eec8c485-1ef2-4b0f-a9c3-d6543bd0eb6b)

### Product Area

Issues

### Link

https://sentry.sentry.io/issues/5225041747/similar/?project=1

floels:
Hi,

Florian here, aspiring external contributor 😃

I was able to reproduce this locally.

The code that renders these "similar issue details" is this one:

https://github.com/getsentry/sentry/blob/f1ad8977dd6e61f4b57b7e4852f88a19f02c84a2/static/app/views/issueDetails/groupSimilarIssues/similarStackTrace/item.tsx#L130-L147

The `Details` parent component renders a `div` with `display: grid;` and no `align-items` property set. As laid out in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_grid_layout/Box_alignment_in_grid_layout#using_align-items), this means the child components will stretch to fill the parent's height, hence the layout bug on the checkbox child element.

The most natural solution seems to set `align-items: start` on the `<Details />` parent component, and in turn set `height: 100%` on the [`<Diff />` child component](https://github.com/getsentry/sentry/blob/f1ad8977dd6e61f4b57b7e4852f88a19f02c84a2/static/app/views/issueDetails/groupSimilarIssues/similarStackTrace/item.tsx#L142-L146), so that the "Diff" button stays centered vertically:

![ezgif-5-d8813bb858](https://github.com/getsentry/sentry/assets/79283128/1d3d20ce-8dbf-4531-a691-f65e9f00eefe)

Let me know if you'd like me to open a PR for that 🙏

roggenkemper:
@floels If you are interesting in putting up a PR to fix this, i would definitely recommend it! Feel free to follow up on this issue once you have one up!

floels:
@roggenkemper thanks, there it is!

[...]

PR #70274, opened 2024-05-03
fix(ui): Fix layout of checkbox in 'similar issues' table item

floels:
# Goal

The goal of this PR is to fix issue #69931, which concerns a small layout issue in the "similar issues" table row item.

# Bug reproduction

You can reproduce the bug by:

- cloning [this fork of the `sentry` repo](https://github.com/floels/sentry),
- checking out the [`repro-issue-69931` branch](https://github.com/floels/sentry/tree/repro-issue-69931),
- launching the development server,
- visiting [http://localhost:8000/organizations/sentry/issues/16/similar/?project=3](http://localhost:8000/organizations/sentry/issues/16/similar/?project=3) (you will need to log in as admin),
- hovering over the area below the checkbox in the "similar issues" table that appears.

You will see the background color stretching across the table row's height:

ezgif-5-8e59006465

# Fix

The reason the checkbox container stretches across the height of its parent is that the parent has `display: grid;` and no `align-items` property set. As laid out in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_grid_layout/Box_alignment_in_grid_layout#using_align-items), this means the child components will stretch to fill the parent's height.

To fix this, we set `align-items: start` on the parent component, and in turn set `height: 100%` on the `<Diff />` sibling component, so that the "Diff" button stays centered vertically.

# Testing the fix

You can test the fix by following the same instructions as above with the [`repro-fix-issue-69931` branch](https://github.com/floels/sentry/tree/repro-fix-issue-69931) instead. This is the same branch as `repro-issue-69931`, expect with the same changes applied as the ones in this PR (see [this comparison of the two branches](https://github.com/floels/sentry/pull/2)).

ezgif-5-15ec2d424c

# Legal Boilerplate

### DIFF
--- a/static/app/views/issueDetails/groupSimilarIssues/similarStackTrace/item.tsx
+++ b/static/app/views/issueDetails/groupSimilarIssues/similarStackTrace/item.tsx
@@ -189,6 +189,7 @@ const Details = styled('div')`
   ${p => p.theme.overflowEllipsis};
 
   display: grid;
+  align-items: start;
   gap: ${space(1)};
   grid-template-columns: max-content auto max-content;
   margin-left: ${space(2)};
@@ -224,6 +225,7 @@ const StyledCount = styled(Count)`
 `;
 
 const Diff = styled('div')`
+  height: 100%;
   display: flex;
   align-items: center;
   margin-right: ${space(0.25)};


armenzg (approved):
Thank you, @floels !
I will get it merged. Very grateful for your contribution!! 🚀

armenzg:
It looks better. Thanks!

Issue #69249, opened 2024-04-18
[Replay Details - UI] Reflect pause/resume for mobile playback


romtsn:
Because we pause/resume replay when an app goes foreground/background, we should reflect this somehow on the playback (e.g. gray out the video view and also the replay timeline part of it). We could do that based off of the breadcrumb that the SDKs send.

Currently these breadcrumbs have:
- `category`: `navigation`
- `data["state"]`: `foreground|background`

Open question: is this breadcrumb data sufficient or should we introduce a new category/type or add more data?

michellewzhang:
- `ui.blur` & `ui.focus` exist for web replays but we don't display them. we could reuse these for mobile, they correlate to background & foreground
- would be helpful to show, visually, where there are "gaps" (which likely result from switching from foreground/background)
  - would require a large amount of refactoring wrt the scrubber (e.g. we'd probably want to disable clicking into a gap)

PR #70277, opened 2024-05-03
feat(replay): add foreground/background breadcrumbs

michellewzhang:
SCR-20240506-omtl

relates to https://github.com/getsentry/sentry/issues/69249

### DIFF
--- a/static/app/utils/replays/getFrameDetails.tsx
+++ b/static/app/utils/replays/getFrameDetails.tsx
@@ -224,6 +224,20 @@ const MAPPER_FOR_FRAME: Record<string, (frame) => Details> = {
     title: 'User Focus',
     icon: <IconUser size="xs" />,
   }),
+  'app.foreground': () => ({
+    color: 'blue300',
+    description: 'Video started recording',
+    tabKey: TabKey.BREADCRUMBS,
+    title: 'App in Foreground',
+    icon: <IconUser size="xs" />,
+  }),
+  'app.background': () => ({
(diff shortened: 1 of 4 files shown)


static/app/utils/replays/getFrameDetails.tsx
```diff
@@ -224,6 +224,20 @@ const MAPPER_FOR_FRAME: Record<string, (frame) => Details> = {
     title: 'User Focus',
     icon: <IconUser size="xs" />,
   }),
+  'app.foreground': () => ({
+    color: 'blue300',
+    description: 'Video started recording',
```
billyvg, line 229: While this is true, I would describe this in terms of what the application is doing, and not what we/SDK is doing.

michellewzhang: do you have any ideas of what to put instead?

michellewzhang: @billyvg do you have strong opposing thoughts here? i'm also ok with "application went into the background" and variations

billyvg: I'm ok with "Replay <status>" - I suppose I want to avoid "video" because of web replays and how we want to make sure that people know that replays aren't videos

michellewzhang: okay let's go with the replay terminology!

billyvg (approved):

PR #70278, opened 2024-05-03
ref: RedisBuffer partitions is always 1

asottile-sentry:
type checker helped me find this one \o/

with multiple partitions RedisBuffer will always crash with a TypeError:

```
tests/sentry/buffer/test_redis.py:129: in test_group_cache_updated
    self.buf.incr(
src/sentry/buffer/redis.py:373: in incr
    pending_key = self._make_pending_key_from_key(key)
src/sentry/buffer/redis.py:173: in _make_pending_key_from_key
    return self._make_pending_key(crc32(key) % self.pending_partitions)
src/sentry/utils/compat/__init__.py:23: in crc32
    rt = _crc32(*args)
E   TypeError: a bytes-like object is required, not 'str'
```

the last time this setting was flipped was in 2018

### DIFF
--- a/src/sentry/buffer/base.py
+++ b/src/sentry/buffer/base.py
@@ -69,6 +69,8 @@ class Buffer(Service):
             }
         )
 
+    # TODO: `partition` is unused, remove after a deploy
+
     def process_pending(self, partition: int | None = None) -> None:
         return
 
--- a/src/sentry/buffer/redis.py
+++ b/src/sentry/buffer/redis.py
@@ -15,9 +15,8 @@ from rediscluster import RedisCluster
 
 from sentry.buffer.base import Buffer
 from sentry.db import models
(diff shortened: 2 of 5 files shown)


src/sentry/buffer/base.py
```diff
@@ -69,6 +69,8 @@ def incr(
             }
         )
 
+    # TODO: `partition` is unused, remove after a deploy
```
anonrig, line 72:
```suggestion
    # TODO(@asottile-sentry): `partition` is unused, remove after a deploy
```

anonrig (approved):

src/sentry/buffer/base.py
asottile-sentry: meh, I have the followup branch already ready :D

PR #70279, opened 2024-05-03
ref: improve utils typing

asottile-sentry:


src/sentry/utils/env.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
anonrig, line 1: This should be surrounded by TYPE_CHECKING

src/sentry/web/client_config.py
```diff
@@ -43,7 +43,12 @@
 from sentry.utils.email import is_smtp_enabled
 from sentry.utils.http import is_using_customer_domain
 from sentry.utils.settings import is_self_hosted, should_show_beacon_consent_prompt
-from sentry.utils.support import get_support_mail
+
+
+def _get_support_mail() -> str | None:
+    """Returns the most appropriate support email address"""
+
+    return options.get("system.support-email") or options.get("system.admin-email") or None
```
anonrig, line 51: Maybe? 

```suggestion
    return options.get("system.support-email", None) or options.get("system.admin-email", None)
```

anonrig (approved):

src/sentry/utils/env.py
asottile-sentry: logging is a `stdlib` module so it's not going to cause an import cycle -- trying to avoid the import here also doesn't save anything since the module using this module will have already imported `logging`

src/sentry/web/client_config.py
asottile-sentry: the default for `get` is already `None`

anonrig: So, the last `None` is not required?

asottile-sentry: it's a little bit different -- in that options have a builtin default -- and that can be empty or None -- the `or None` handles it being set but to a falsey value

PR #70286, opened 2024-05-03
fix(perf): Do not render contents of closed panels

gggritso:
Right now, when a `SlideOverPanel` is collapsed, its contents still exist in the DOM. This has a whole bunch of downsides:

1. React has to render stuff inside, which is pointless since it's invisible
2. When writing mocks, we have to assert on elements that might exist in the panel even if its closed

This is very annoying! This PR changes it up to use `AnimatePresence` which lets us use a `!collapsed &` condition and _not_ render the panel's contents in the DOM if it's collapsed.

Also clarified some variables names while I'm at it, and removed some unused CSS.

### DIFF
--- a/static/app/views/starfish/components/detailPanel.spec.tsx
+++ b/static/app/views/starfish/components/detailPanel.spec.tsx
@@ -1,0 +1,15 @@
+import {render, screen} from 'sentry-test/reactTestingLibrary';
+
+import DetailPanel from 'sentry/views/starfish/components/detailPanel';
+
+describe('DetailPanel', function () {
+  it('renders toolbar and inner content', function () {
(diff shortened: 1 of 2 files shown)


gggritso:
@edwardgou-sentry what do you think, here? I poked around the app and everything seemed normal, but wondering if I missed any edge cases, or forgot any considerations

DominikB2014 (approved):

edwardgou-sentry (approved):
Thanks, I think this makes sense! If I remember correctly the only reason we didn't unrender the SlideOverPanel was because it would shortcircuit the exit animation. But it looks like AnimatePresence takes care of that!

PR #70289, opened 2024-05-03
ref(crons): Normalize crons incident issues

evanpurkhiser:
Prior to incidents we created issues for each type of faiure (error,
timeout, missed). This is because only one failed check-in was needed to
create an issue. With incidents you can configure how many failures are
needed, meaning there could be 2 missed, 1 timeout, and 1 error.

This removes the various issue occurrence types and replaces them with a
single MonitorIncidentType

### DIFF
--- a/src/sentry/issues/grouptype.py
+++ b/src/sentry/issues/grouptype.py
@@ -247,10 +247,6 @@ class PerformanceGroupTypeDefaults:
     noise_config = NoiseConfig()
 
 
-class CronGroupTypeDefaults:
-    notification_config = NotificationConfig(context=[])
-
-
 class ReplayGroupTypeDefaults:
     notification_config = NotificationConfig(context=[])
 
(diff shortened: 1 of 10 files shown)


wedamija (approved):

src/sentry/issues/grouptype.py
```diff
@@ -518,36 +515,27 @@ class ProfileFunctionRegressionType(GroupType):
...
-    released = True
-    creation_quota = Quota(3600, 60, 60_000)  # 60,000 per hour, sliding window of 60 seconds
-    default_priority = PriorityLevel.HIGH
 
 
 @dataclass(frozen=True)
-class MonitorCheckInMissed(CronGroupTypeDefaults, GroupType):
+@deprecated("This is only kept around for it's type_id")
+class MonitorCheckInMissedDeprecated(MonitorIncidentType, GroupType):
     type_id = 4003
```
wedamija, line 538: I'd recommend keeping all the details, since they're used to describe the issue that you see

wedamija: Nm missed the inheritance

wedamija (approved):

PR #70292, opened 2024-05-03
fix(webhooks): no data migrations for webhooks in the same category

sentaur-athena:
Before this PR any new webhook we add to issues would required a data migration. This PR is a short term refactory so we don't have to run data migrations following up with a larger one to if possible not store `event` at all in DB.

The problem lies in the fact that right now `sentry_app` model stores the webhook `events` but you don't subscribe to webhooks event by event, you subscribe to webhooks by category. 

For example in the UI when someone creates and integration they check the box below and we store events: `[issue.created, issue.resolved, issue.archived, issue.assigned]` in the model. It makes more sense product wise when we add to these events they just get launched to everyone who already signed up for `issues`.
Screenshot 2024-05-02 at 9 46 41 PM

[Slack conversation](https://sentry.slack.com/archives/CD4NYFD34/p1714690555036559) for more context on product decision

### DIFF
--- a/src/sentry/receivers/sentry_apps.py
+++ b/src/sentry/receivers/sentry_apps.py
@@ -10,6 +10,7 @@ from sentry.models.organization import Organization
 from sentry.models.project import Project
 from sentry.models.team import Team
 from sentry.models.user import User
+from sentry.sentry_apps.apps import consolidate_events
 from sentry.services.hybrid_cloud import coerce_id_from
 from sentry.services.hybrid_cloud.app import RpcSentryAppInstallation, app_service
 from sentry.services.hybrid_cloud.user import RpcUser
@@ -59,7 +60,6 @@ def send_issue_resolved_webhook(organization_id, project, group, user, resolutio
 def send_issue_ignored_webhook(project, user, group_list, **kwargs):
     for issue in group_list:
         send_workflow_webhooks(project.organization, issue, user, "issue.ignored")
-        send_workflow_webhooks(project.organization, issue, user, "issue.archived")
 
 
 @issue_unresolved.connect(weak=False)
@@ -118,7 +118,7 @@ def send_comment_deleted_webhook(project, user, group, data, **kwargs):
 def send_comment_webhooks(organization, issue, user, event, data=None):
     data = data or {}
 
-    for install in installations_to_notify(organization, event):
+    for install in installations_to_notify(organization, "comment"):
         build_comment_webhook.delay(
             installation_id=install.id,
             issue_id=issue.id,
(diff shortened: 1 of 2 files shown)


src/sentry/receivers/sentry_apps.py
```diff
@@ -59,7 +60,6 @@ def send_issue_resolved_webhook(organization_id, project, group, user, resolutio
 def send_issue_ignored_webhook(project, user, group_list, **kwargs):
     for issue in group_list:
         send_workflow_webhooks(project.organization, issue, user, "issue.ignored")
-        send_workflow_webhooks(project.organization, issue, user, "issue.archived")
```
sentaur-athena, line 62: we deprecated `issue.ignored` in favor of `issue.archived` but kept sending both for backwards compatibility. The way this worked was that the old installations were subscribe to `ignored` only and the new ones to `archived` only so no one would get 2 webhooks for one event. Instead of sending 2 webhooks I'm just renaming it to what the installation expects in line 142.

src/sentry/receivers/sentry_apps.py
```diff
@@ -136,17 +136,36 @@ def send_workflow_webhooks(
...
 
-def installations_to_notify(organization, event) -> list[RpcSentryAppInstallation]:
+def installations_to_notify(
+    organization: Organization, resource_type: str
+) -> list[RpcSentryAppInstallation]:
     installations = app_service.get_installed_for_organization(organization_id=organization.id)
-    return [i for i in installations if event in i.sentry_app.events]
+    # All issue webhooks are under one subscription, so if an intallation is subscribed to any issue
+    # events it should get notified for all the issue events
+    # TODO: Refactor sentry_app model so it doesn't store event, instead it stores subscription
```
sentaur-athena, line 158: https://github.com/getsentry/team-core-product-foundations/issues/311

src/sentry/receivers/sentry_apps.py
```diff
@@ -118,7 +118,7 @@ def send_comment_deleted_webhook(project, user, group, data, **kwargs):
 def send_comment_webhooks(organization, issue, user, event, data=None):
     data = data or {}
 
-    for install in installations_to_notify(organization, event):
+    for install in installations_to_notify(organization, "comment"):
```
scefali, line 121: @sentaur-athena so we don't need `event` anymore?

sentaur-athena: yes, that's the idea. To know which installations we need to notify, we need to know which ones subscribed to any event in this category. In this case the category for all events is "comment" so I didn't get it from event. I can replace it with event.split(".")[0]

scefali (approved):

PR #70294, opened 2024-05-03
feat(replay): show screen name in url bar for mobile replays

### LINKED ISSUE #67697: [Replay Details - UI] URL should show screen name
It's likely we don't need to do anything on the UI side as SDK should be sending the correct breadcrumbs.

Today the player relies on navigation breadcrumbs. We should aim to rely on the same breadcrumbs (perhaps with small adjustments like `category`)

michellewzhang:
- closes https://github.com/getsentry/sentry/issues/67697
- mirroring the web replay url functions -- show screen name (1 per navigation crumb) in the url bar
- (the replay below has 3 nav frames in a second so it picks the last one)

https://github.com/getsentry/sentry/assets/56095982/24791fd6-ad4b-455e-a1b0-a7a3c483865b

### DIFF
--- a/static/app/components/replays/replayCurrentScreen.tsx
+++ b/static/app/components/replays/replayCurrentScreen.tsx
@@ -1,0 +1,38 @@
+import {useMemo} from 'react';
+import * as Sentry from '@sentry/react';
+
+import {useReplayContext} from 'sentry/components/replays/replayContext';
+import TextCopyInput from 'sentry/components/textCopyInput';
+import {t} from 'sentry/locale';
+import getCurrentScreenName from 'sentry/utils/replays/getCurrentScreenName';
+
+// Screen name component for video/mobile replays - mirrors replayCurrentUrl.tsx
+function ReplayCurrentScreen() {
+  const {currentTime, replay} = useReplayContext();
+  const frames = replay?.getMobileNavigationFrames();
+
+  const screenName = useMemo(() => {
+    try {
+      return getCurrentScreenName(frames, currentTime);
+    } catch (err) {
+      Sentry.captureException(err);
+      return '';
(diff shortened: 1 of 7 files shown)


static/app/components/replays/replayCurrentScreen.tsx
c298lee, line 1: can we add a test for this?

static/app/utils/replays/getCurrentScreenName.tsx
```diff
@@ -0,0 +1,20 @@
+import type {BreadcrumbFrame} from 'sentry/utils/replays/types';
+
+// Gets the current screen name for video/mobile replays - mirrors getCurrentUrl.tsx
+function getCurrentScreenName(
```
c298lee, line 4: can we add a test for this?

michellewzhang: yup

static/app/utils/replays/types.tsx
```diff
@@ -57,8 +57,8 @@ type ExtraBreadcrumbTypes =
   | {
       category: 'navigation';
       data: {
-        from: string;
         to: string;
+        from?: string;
```
michellewzhang, line 61: marked as optional because the SDK currently isn't sending a field for `from` but we can reuse this crumb type

c298lee (approved):

static/app/utils/replays/types.tsx
bruno-garcia: That could be a good comment on top of the prop

michellewzhang: i'll pose it a a question in the channel

PR #70297, opened 2024-05-04
feat(crons): Add way to persist errors during checkin ingestion

wedamija:
This adds persistence for errors generated during checkin ingestion. We're just going to use Redis for this since this data is ephemeral and we want to avoid the overhead of postgres.

We'll store up to 10 errors per monitor, and for any checkins where we couldn't get the specific monitor, we'll store up to 10 errors per project as well. These sets will only stick around for 1 day, although we could keep them longer if we think it's valuable.

We store a fair bit of data about the checkin - the entire payload, and any details about the error that occurred that we can get. We might want to rework the format of the error payload, but that can happen in a separate pr.

### DIFF
--- a/src/sentry/monitors/processing_errors.py
+++ b/src/sentry/monitors/processing_errors.py
@@ -112,7 +112,7 @@ class CheckinProcessingError:
 
 
 class CheckinProcessErrorsManager:
-    def get_cluster(self):
+    def _get_cluster(self):
         return redis.redis_clusters.get(settings.SENTRY_MONITORS_REDIS_CLUSTER)
 
     def store(self, error: CheckinProcessingError, monitor: Monitor | None):
@@ -153,24 +153,25 @@ class CheckinProcessErrorsManager:
         return self._get_for_entity(self.build_project_identifier(project))
 
     def _get_for_entity(self, identifier: str) -> list[CheckinProcessingError]:
-        redis = self.get_cluster()
+        redis = self._get_cluster()
         error_key = f"monitors.processing_errors.{identifier}"
-        raw_errors = redis.zrange(error_key, 0, MAX_ERRORS_PER_SET)
+        raw_errors = redis.zrange(error_key, 0, MAX_ERRORS_PER_SET, desc=True)
         return [CheckinProcessingError.from_dict(json.loads(raw_error)) for raw_error in raw_errors]
 
 
 def handle_processing_errors(item: CheckinItem, error: CheckinValidationError):
     try:
-        organization = Organization.objects.get(projects__id=item.message["project_id"])
+        organization = Organization.objects.get(project__id=item.message["project_id"])
         if not features.has("organizations:crons-write-user-feedback", organization):
             return
 
-        metric_kwargs = {
-            "source": "consumer",
-            "sdk_platform": item.message["sdk"],
-        }
-
-        metrics.incr("monitors.checkin.handle_processing_error", tags=metric_kwargs)
+        metrics.incr(
+            "monitors.checkin.handle_processing_error",
+            tags={
+                "source": "consumer",
+                "sdk_platform": item.message["sdk"],
+            },
+        )
 
         checkin_processing_error = CheckinProcessingError(error.processing_errors, item)
         manager = CheckinProcessErrorsManager()
--- a/src/sentry/monitors/processing_errors.py
+++ b/src/sentry/monitors/processing_errors.py
@@ -7,6 +7,8 @@ from enum import Enum
 from typing import Any, TypedDict
 
 from django.conf import settings
+from redis.client import StrictRedis
+from rediscluster import RedisCluster
 
 from sentry import features
 from sentry.models.organization import Organization
@@ -112,7 +114,7 @@ class CheckinProcessingError:
 
 
 class CheckinProcessErrorsManager:
-    def _get_cluster(self):
+    def _get_cluster(self) -> RedisCluster[str] | StrictRedis[str]:
         return redis.redis_clusters.get(settings.SENTRY_MONITORS_REDIS_CLUSTER)
 
     def store(self, error: CheckinProcessingError, monitor: Monitor | None):
@@ -126,13 +128,13 @@ class CheckinProcessErrorsManager:
             except Monitor.DoesNotExist:
                 pass
         if monitor:
-            error_identifier = f"monitor:{monitor.id}"
+            error_identifier = self.build_monitor_identifier(monitor)
         else:
-            error_identifier = f'project:{error.checkin.message["project_id"]}'
+            error_identifier = self.build_project_identifier(error.checkin.message["project_id"])
 
         error_key = f"monitors.processing_errors.{error_identifier}"
         serialized_error = json.dumps(error.to_dict())
-        redis_client = redis.redis_clusters.get(settings.SENTRY_MONITORS_REDIS_CLUSTER)
+        redis_client = self._get_cluster()
         pipeline = redis_client.pipeline(transaction=False)
         pipeline.zadd(error_key, {serialized_error: error.checkin.ts.timestamp()})
         # Cap the error list to the `MAX_ERRORS_PER_SET` most recent errors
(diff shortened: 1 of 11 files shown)


src/sentry/monitors/processing_errors.py
```diff
@@ -36,3 +54,127 @@ class ProcessingErrorType(Enum):
...
+    @classmethod
+    def from_dict(cls, data: CheckinProcessingErrorData) -> CheckinProcessingError:
+        return cls(
+            errors=[ProcessingError.from_dict(error) for error in data["errors"]],
+            checkin=CheckinItem.from_dict(data["checkin"]),
+        )
+
+
+class CheckinProcessErrorsManager:
+    def _get_cluster(self):
```
evanpurkhiser, line 115: Can we add a type?

src/sentry/monitors/processing_errors.py
```diff
@@ -36,3 +54,127 @@ class ProcessingErrorType(Enum):
...
+
+    @classmethod
+    def from_dict(cls, data: CheckinProcessingErrorData) -> CheckinProcessingError:
+        return cls(
+            errors=[ProcessingError.from_dict(error) for error in data["errors"]],
+            checkin=CheckinItem.from_dict(data["checkin"]),
+        )
+
+
+class CheckinProcessErrorsManager:
```
evanpurkhiser, line 114: Do we actually need the class? Doesn't seem to have any interesting state, just using it as a namespace?

wedamija: Yeah, mostly seemed useful to keep it as a namespace rather than have functions scattered around. Could be a module instead?

evanpurkhiser: yeah could be a module. I'm indifferent

src/sentry/monitors/processing_errors.py
```diff
@@ -36,3 +54,127 @@ class ProcessingErrorType(Enum):
...
+            except Monitor.DoesNotExist:
+                pass
+        if monitor:
+            error_identifier = f"monitor:{monitor.id}"
+        else:
+            error_identifier = f'project:{error.checkin.message["project_id"]}'
+
+        error_key = f"monitors.processing_errors.{error_identifier}"
+        serialized_error = json.dumps(error.to_dict())
+        redis_client = redis.redis_clusters.get(settings.SENTRY_MONITORS_REDIS_CLUSTER)
```
evanpurkhiser, line 135: Isn't this what `_get_cluster` is for?

src/sentry/monitors/processing_errors.py
```diff
@@ -36,3 +54,127 @@ class ProcessingErrorType(Enum):
...
+        pipeline = redis_client.pipeline(transaction=False)
```
evanpurkhiser, line 136: Yagiz would be proud

src/sentry/monitors/processing_errors.py
```diff
@@ -36,3 +54,127 @@ class ProcessingErrorType(Enum):
...
+            # Attempt to get the monitor from the checkin info if we failed to retrieve it during ingestion
+            try:
+                monitor = Monitor.objects.get(
+                    project_id=error.checkin.message["project_id"],
+                    slug=error.checkin.payload["monitor_slug"],
+                )
+            except Monitor.DoesNotExist:
+                pass
+        if monitor:
+            error_identifier = f"monitor:{monitor.id}"
```
evanpurkhiser, line 129:
```suggestion
            error_identifier = self.build_monitor_identifier(monitor)
```

src/sentry/monitors/processing_errors.py
```diff
@@ -36,3 +54,127 @@ class ProcessingErrorType(Enum):
...
+    def build_monitor_identifier(self, monitor: Monitor) -> str:
+        return f"monitor:{monitor.id}"
+
+    def get_for_monitor(self, monitor: Monitor) -> list[CheckinProcessingError]:
+        return self._get_for_entity(self.build_monitor_identifier(monitor))
+
+    def build_project_identifier(self, project: Project) -> str:
+        return f"project:{project.id}"
+
+    def get_for_project(self, project: Project) -> list[CheckinProcessingError]:
```
evanpurkhiser, line 152:
```suggestion
    def get_for_project(self, project: Project) -> Sequence[CheckinProcessingError]:
```

src/sentry/monitors/processing_errors.py
```diff
@@ -36,3 +54,127 @@ class ProcessingErrorType(Enum):
...
+        return self._get_for_entity(self.build_project_identifier(project))
+
+    def _get_for_entity(self, identifier: str) -> list[CheckinProcessingError]:
```
evanpurkhiser, line 155:
```suggestion
    def _get_for_entity(self, identifier: str) -> Sequence[CheckinProcessingError]:
```

tests/sentry/monitors/test_types.py
```diff
@@ -0,0 +1,13 @@
+from sentry.monitors.testutils import build_checkin_item
+from sentry.monitors.types import CheckinItem
+from sentry.testutils.cases import TestCase
+
+
+class CheckinItemTest(TestCase):
```
evanpurkhiser, line 6: good call. Could be it's own PR if you wanted to pull it out

src/sentry/testutils/fixtures.py
```diff
@@ -460,6 +461,23 @@ def create_notification_settings_provider(self, *args, **kwargs):
     def create_user_option(self, *args, **kwargs):
         return Factories.create_user_option(*args, **kwargs)
 
+    def create_monitor(self, **kwargs):
```
evanpurkhiser, line 464: Can probably be reused in a lot of our tests. We have some monitor base test case that has this too

src/sentry/monitors/testutils.py
```diff
@@ -1,7 +1,15 @@
...
+
+from sentry.monitors.processing_errors import (
+    CheckinProcessingError,
+    ProcessingError,
+    ProcessingErrorType,
+)
+from sentry.monitors.types import CheckinItem, CheckinPayload
 
 
 def make_ref_time(**kwargs):
```
evanpurkhiser, line 15: Unrelated but I need to delete this

evanpurkhiser: https://github.com/getsentry/sentry/pull/70429

src/sentry/monitors/processing_errors.py
```diff
@@ -36,3 +54,127 @@ class ProcessingErrorType(Enum):
...
+        redis = self._get_cluster()
+        error_key = f"monitors.processing_errors.{identifier}"
+        raw_errors = redis.zrange(error_key, 0, MAX_ERRORS_PER_SET, desc=True)
+        return [CheckinProcessingError.from_dict(json.loads(raw_error)) for raw_error in raw_errors]
+
+
+def handle_processing_errors(item: CheckinItem, error: CheckinValidationError):
+    try:
+        organization = Organization.objects.get(project__id=item.message["project_id"])
```
evanpurkhiser, line 164: should we get from cache?

src/sentry/monitors/processing_errors.py
```diff
@@ -36,3 +54,127 @@ class ProcessingErrorType(Enum):
...
+                monitor = Monitor.objects.get(
+                    project_id=error.checkin.message["project_id"],
+                    slug=error.checkin.payload["monitor_slug"],
+                )
+            except Monitor.DoesNotExist:
+                pass
+        if monitor:
+            error_identifier = f"monitor:{monitor.id}"
+        else:
+            error_identifier = f'project:{error.checkin.message["project_id"]}'
```
evanpurkhiser, line 131: wish we could use `build_project_identifier`. Tempted to say the signature could be `project_or_project_id` but feels kinda gross

evanpurkhiser (approved):
Left a few comments but I think it's good to go

src/sentry/monitors/processing_errors.py
wedamija: I added a bunch of convenience methods and forgot to use them 🤦

src/sentry/monitors/processing_errors.py
wedamija: Not that clear on what is preferred here. We know we're returning an actual list, isn't it better to be specific? I see both options in the codebase

src/sentry/testutils/fixtures.py
wedamija: Yeah, ideally we'd just use this one method

src/sentry/monitors/processing_errors.py
wedamija: There's no cache for doing this by project id https://github.com/getsentry/sentry/blob/fa0930966ce70d49060dda79c9f951b334d35929/src/sentry/models/organization.py#L203

I could fetch the project by cache, then fetch the org by cache... maybe that's better

src/sentry/monitors/processing_errors.py
evanpurkhiser: You're right https://typing.readthedocs.io/en/latest/source/best_practices.html#arguments-and-return-types

> For return values, prefer concrete types (list, dict, etc.) for concrete implementations. The return values of protocols and abstract base classes must be judged on a case-by-case basis.

PR #70298, opened 2024-05-04
perf(tagstore): sample if timerange longer than 14 days

JoshFerge:
- allow sampling to run on tagkey queries if the timerange is greater than 14 days on the query. We frequently see errors on the tagstore endpoint with 30 and 90 day queries on certain projects

### DIFF
--- a/src/sentry/tagstore/snuba/backend.py
+++ b/src/sentry/tagstore/snuba/backend.py
@@ -3,7 +3,7 @@ import os
 import re
 from collections import defaultdict
 from collections.abc import Iterable, Sequence
-from datetime import timezone
+from datetime import timedelta, timezone
 from typing import Any
 
 from dateutil.parser import parse as parse_datetime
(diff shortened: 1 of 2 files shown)


src/sentry/tagstore/snuba/backend.py
```diff
@@ -440,7 +440,9 @@ def get_tag_keys_for_projects(
         # the sampling that turbo enables so that we get more accurate results.
         # We only want sampling when we have a large number of projects, so
         # that we don't cause performance issues for Snuba.
-        if len(projects) <= max_unsampled_projects:
+        # We also see issues with long timeranges in large projects,
+        # So only disable sampling if the timerange is short enough.
+        if len(projects) <= max_unsampled_projects and end - start > timedelta(days=14):
             optimize_kwargs["sample"] = 1
```
evanh, line 446: What is the sample rate that this eventually sets in the request?

evanh (approved):

src/sentry/tagstore/snuba/backend.py
JoshFerge: looks like 10%? https://github.com/getsentry/snuba/blob/b88683fae94b6a3835b4db10793c7c6ccc0fb723/snuba/settings/__init__.py#L267

PR #70301, opened 2024-05-05
feat(trace): disable/enable reset zoom

JonasBa:
Enable or disable the button based off the zoom state (previously always enabled)


Abdkhan14 (approved):
lgtm

PR #70304, opened 2024-05-05
fix(ui): Fix incorrect search dynamic in 'Search Support' modal

floels:
# Goal

The goal of this PR is to fix a bug in the "Search Support, Docs and More" search bar. When typing into this search bar, the search query sent to the backend in the search request is the previous input value instead of the current one:

Screenshot 2024-05-05 at 20 54 43

See #70305 for more details, including reproduction steps.

# Approach

The bug is caused by an incorrect use of  `componentDidUpdate` in `static/app/components/helpSearch.tsx`:

https://github.com/getsentry/sentry/blob/3e90887a3632585b048f607668e20927321a39a0/static/app/components/search/sources/helpSource.tsx#L56-L60

The first argument of `componentDidUpdate` is the previous props' value, not the new ones', as documented [here](https://legacy.reactjs.org/docs/react-component.html#componentdidupdate).

The fix is straightforward (compare the current props with the previous props, and if they are different call `doSearch` on `this.props.query`, which indeed has the new query value).

We also add an assertion to the relevant test in `sentry/static/app/components/helpSearch.spec.tsx`. You can check that the test now fails if you remove the fix:

Screenshot 2024-05-05 at 21 33 06

# Legal Boilerplate

### DIFF
--- a/static/app/components/helpSearch.spec.tsx
+++ b/static/app/components/helpSearch.spec.tsx
@@ -1,3 +1,5 @@
+import {SentryGlobalSearch} from '@sentry-internal/global-search';
+
 import {render, screen, userEvent} from 'sentry-test/reactTestingLibrary';
 
 import HelpSearch from 'sentry/components/helpSearch';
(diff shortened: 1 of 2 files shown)


evanpurkhiser:
Will take a look first thing Monday! Thanks for this!

evanpurkhiser (approved):

evanpurkhiser:
Thanks so much for this fix!!

PR #70306, opened 2024-05-05
fix(ts): Fix TypeScript issue in integration configuration view

floels:
# Goal

The goal of this PR is to fix a TypeScript error currently ignored with a `@ts-ignore` statement in `static/app/views/settings/organizationIntegrations/configureIntegration.tsx`:

https://github.com/getsentry/sentry/blob/3e90887a3632585b048f607668e20927321a39a0/static/app/views/settings/organizationIntegrations/configureIntegration.tsx#L253-L262

# Approach

There is indeed a type called `OrganizationIntegration` that has an `externalId` property:

https://github.com/getsentry/sentry/blob/3e90887a3632585b048f607668e20927321a39a0/static/app/types/integrations.tsx#L386-L391

but it needs to extend `Integration` instead of `CommonIntegration`:

https://github.com/getsentry/sentry/blob/3e90887a3632585b048f607668e20927321a39a0/static/app/types/integrations.tsx#L358-L380

Indeed, when we look at the API's `integration` model serializers (`src/sentry/api/serializers/models/integration.py`), the only place where we (may) have an `externalId` property is also one where we have `dynamicDisplayInformation` and `scopes`.

After making this change in `static/app/views/settings/organizationIntegrations/configureIntegration.tsx`, we need to also adapt type declarations in `static/app/views/settings/organizationIntegrations/integrationServerlessFunctions.tsx` and `static/app/views/settings/organizationIntegrations/integrationServerlessRow.tsx` for consistency.

# Legal Boilerplate

### DIFF
--- a/static/app/types/integrations.tsx
+++ b/static/app/types/integrations.tsx
@@ -383,7 +383,7 @@ type ConfigData = {
   installationType?: string;
 };
 
-export interface OrganizationIntegration extends CommonIntegration {
+export interface OrganizationIntegration extends Integration {
   configData: ConfigData | null;
   configOrganization: Field[];
   externalId: string;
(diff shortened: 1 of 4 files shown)


evanpurkhiser (approved):
Thanks for fixing this!

floels:
Thanks @evanpurkhiser for the review! Do you think we could merge this? I don't have write access to the repo.

PR #70307, opened 2024-05-05
fix(ts): Remove non-null assertion in integrationExternalUserMappings.tsx

floels:
# Goal

The goal of this PR is to get rid of a non-null assertion (`!`) following an optional chaining operator (`?.`):

https://github.com/getsentry/sentry/blob/3e90887a3632585b048f607668e20927321a39a0/static/app/views/settings/organizationIntegrations/integrationExternalUserMappings.tsx#L105-L112

Optional chain expressions are designed to return `undefined` if the optional property is nullish. Using a non-null assertion after an optional chain expression doesn't make logical sense and introduces a type safety hole into the code.

# Approach

We get rid of the non-null assertion and simply allow the `id` string to be `undefined` in the return value of the `sentryNamesMapper` prop passed down to the `<IntegrationExternalMappings />` child component.

# Legal Boilerplate

### DIFF
--- a/static/app/views/settings/organizationIntegrations/integrationExternalUserMappings.tsx
+++ b/static/app/views/settings/organizationIntegrations/integrationExternalUserMappings.tsx
@@ -107,7 +107,7 @@ class IntegrationExternalUserMappings extends DeprecatedAsyncComponent<Props, St
       .filter(member => member.user)
       .map(({user, email, name}) => {
         const label = email !== name ? `${name} - ${email}` : `${email}`;
-        return {id: user?.id!, name: label};
+        return {id: user?.id, name: label};
       });
   }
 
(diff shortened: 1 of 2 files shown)


static/app/views/settings/organizationIntegrations/integrationExternalUserMappings.tsx
```diff
@@ -107,7 +107,7 @@ class IntegrationExternalUserMappings extends DeprecatedAsyncComponent<Props, St
       .filter(member => member.user)
       .map(({user, email, name}) => {
         const label = email !== name ? `${name} - ${email}` : `${email}`;
-        return {id: user?.id!, name: label};
+        return {id: user?.id, name: label};
```
evanpurkhiser, line 110: I'm not totally convinced on allowing `undefined` into the `id` here. We're already filtering on the user not existing, so it is guaranteed that `user` exists here. I think this null assertion is OK

PR #70319, opened 2024-05-06
fix(metrics): sample position with 0 series

### LINKED ISSUE #70318: Min shows all 0s and samples are in the wrong location
Image

obostjancic:
- fixes #70318

Before:
image

After:
image

### DIFF
--- a/static/app/views/metrics/chart/useMetricChartSamples.tsx
+++ b/static/app/views/metrics/chart/useMetricChartSamples.tsx
@@ -40,12 +40,19 @@ function getValueRectFromSeries(series: Series[]) {
     s.data.map(entry => entry.value)
   );
 
-  return {
+  const rect = {
     xMin: Math.min(...xValues),
     xMax: Math.max(...xValues),
     yMin: Math.min(0, ...yValues) / scalingFactor,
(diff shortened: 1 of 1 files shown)


static/app/views/metrics/chart/useMetricChartSamples.tsx
```diff
@@ -40,12 +40,19 @@ function getValueRectFromSeries(series: Series[]) {
...
 
-  return {
+  const rect = {
     xMin: Math.min(...xValues),
     xMax: Math.max(...xValues),
     yMin: Math.min(0, ...yValues) / scalingFactor,
     yMax: Math.max(0, ...yValues) / scalingFactor,
   };
+
+  // happens when refenceSeries has all 0 values, commonlt seen when using min() aggregation
```
shellmayr, line 50:
```suggestion
  // happens when refenceSeries has all 0 values, commonly seen when using min() aggregation
```

matejminar (approved):
sweet!
so this fixes the samples - what about the line? is it expected/correct that we have all 0s for min?

obostjancic:
> sweet! so this fixes the samples - what about the line? is it expected/correct that we have all 0s for min?

yeah, the line is ok, we really get back all zeroes

PR #70331, opened 2024-05-06
fix(spans): Search on span op and status

Zylphrex:
A few more bugs, searching on span op should support INs and status should map to the corresponding integer.


wmak (approved):
ty lgtm

PR #70334, opened 2024-05-06
chore(perf): Add instrumentation around performance issue detection

gggritso:
Wrap every detector in a span, so we can see which specific detectors are slow on any given transaction.

### DIFF
--- a/src/sentry/utils/performance_issues/performance_detection.py
+++ b/src/sentry/utils/performance_issues/performance_detection.py
@@ -343,7 +343,7 @@ def _detect_performance_problems(
 
     for detector in detectors:
         with sentry_sdk.start_span(op="function", description="run_detector_on_data") as span:
-            span.set_data("detector.type.value", detector.type.value)
+            span.set_tag("detector.type.value", detector.type.value)
             run_detector_on_data(detector, data)
 
     # Metrics reporting only for detection, not created issues.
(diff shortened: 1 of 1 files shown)


shruthilayaj (approved):

src/sentry/utils/performance_issues/performance_detection.py
```diff
@@ -342,7 +342,9 @@ def _detect_performance_problems(
     ]
 
     for detector in detectors:
-        run_detector_on_data(detector, data)
+        with sentry_sdk.start_span(op="function", description="run_detector_on_data") as span:
```
shruthilayaj, line 345: can we name the detector on the span description instead

shruthilayaj: (we can pull percentiles per detector span that way)

gggritso: @shruthilayaj In Metrics? In that case should I use a tag, so it can segment properly?

shruthilayaj: nah, just with indexed data (at least we can check span summary that way and have a row per detector) - also just lets us read the span waterfall easier
if we want to pull metrics, we can have the span op be per detector - either works!

gggritso: 👍🏻 ok

Issue #58410, opened 2023-10-18
Crons: Add locking mechanisms for crons clock driven fan-out tasks


evanpurkhiser:
There exists a problem with the tasks that are triggered due to clocks. Here is a diagram of what each clock tick does:

![image](https://github.com/getsentry/sentry/assets/1421724/43090f7a-90e7-45cb-aee0-2191c5c5967c)

Each tick triggers the `check_missing` and `check_timeout` tasks. These tasks both fan out tasks to mark monitors as missed and check-ins as timed out.

Typically each tick happens about a minute apart. However, these ticks are driven by the check-in kafka topic. So a backlog in this topic may cause ticks to be produced **in rapid succession** (always in order). In this scenario we may publish multiple of these tasks into the celery-beat queue.

How does this affect each task?

### `check_missing`

See the task code for context:

https://github.com/getsentry/sentry/blob/a3264a8cd0366f1df21c4a12023256bd31416428/src/sentry/monitors/tasks.py#L224-L248

When both tasks are put into celery, if the second tick (12:02) happens to run at the same time as the first tick (12:01):

1. We may mark monitors as missed for the SECOND minute and not the first. This is because we use the `next_checkout_latest` column on the table to determine which monitors are missed. If we mark monitors in the second minute as missed before the first the first minute will never get marked as missed since `next_checkout_latest` will have already moved forward.

2. When we skip minutes **this will affect threshold calculations for creating monitor incidents**.

### `check_timeout`

This task appears to be unaffected. Since this task simply looks for check-ins that need to be marked as timed out, it's mostly idempotent.

If one tick marks earlier tick check-ins as timed out out of order it does not matter since 

## Proposed solution

Instead of having the ticks dispatch celery tasks we can introduce a kafka topic for the clock ticks and a consumer of those clock ticks.

The topic should be a single partition topic with a single threaded consumer.

The consumer will consume ticks in order. Each consumer will then run a [`ProcessPoolExecutor`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor) to execute the tasks along with their fan-out tasks. Because the work done is mostly database queries and is not CPU bound we can use a large number of processes / threads.

We will no longer use celery to execute these tasks. instead the consumer will simply execute them all. Once it has finished all work it can continue to consume the next tick.

fpacifici:
> The clock tick consumer will be responsible for doing the following work for each clock tick. This work MUST be completed before consuming the next tick.

[...]

PR #70338, opened 2024-05-06
feat(crons): Backfill missed clock-tick dispatches

evanpurkhiser:
When we detect a clock tick it is possible that we may have skipped a tick when a monitor ingest partition is very slow and does not contain a message for a entire minute (and other partitions have already moved multiple minutes forward).

Previously we would log this and avoid producing a clock tick for these skipped minute(s) as we were using celery to dispatch the check_missing and check_timeout tasks. Since the celery tasks would be produced back-to-back it wasn't unlikely they would be processed out of order

Since the completion of GH-58410 we are now guaranteed that clock tick tasks are processed in order.

### DIFF
--- a/src/sentry/monitors/clock_dispatch.py
+++ b/src/sentry/monitors/clock_dispatch.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import logging
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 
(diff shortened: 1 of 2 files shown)


wedamija (approved):

evanpurkhiser:
Won't merge until we've enabled the consumer in US/DE

wedamija (approved):

PR #70343, opened 2024-05-06
ref(crons): Produce missed check-ins for muted monitors

evanpurkhiser:
With the completion of GH-58410 we should be able to support increasing
the task load for marking muted monitors as missing.

### DIFF
--- a/src/sentry/monitors/clock_tasks/check_missed.py
+++ b/src/sentry/monitors/clock_tasks/check_missed.py
@@ -32,35 +32,18 @@ MONITOR_LIMIT = 10_000
 
 # re-use the monitor exclusion query node across dispatch_check_missing and
 # mark_environment_missing.
-IGNORE_MONITORS = (
-    ~Q(
-        status__in=[
-            MonitorStatus.DISABLED,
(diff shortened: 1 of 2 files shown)


gaprl (approved):

src/sentry/monitors/clock_tasks/check_missed.py
```diff
@@ -32,35 +32,18 @@
...
-        # now that we're using a kafka consumer to process these
-        monitor__is_muted=True,
-    )
-    & ~Q(
-        # TODO(epurkhiser): This is a hold-over from when we were producing
-        # these tasks into celery. Removing this conditional will create many
-        # more monitors for us to mark as missed. We should be able to remove
-        # now that we're using a kafka consumer to process these
-        is_muted=True,
-    )
```
wedamija, line 63: I'm confused about why we'd remove these conditions though... do we actually want to include misses for muted monitors?

evanpurkhiser: Yeah I think it makes sense. Mute isn't "stop checking" it's "stop alerting"

PR #70344, opened 2024-05-06
perf: move assemble status to redis

anonrig:
Dual writes to memcache and redis while transitioning assemble status to redis

Ref: https://github.com/getsentry/sentry/issues/68598

### DIFF
--- a/src/sentry/tasks/assemble.py
+++ b/src/sentry/tasks/assemble.py
@@ -6,9 +6,11 @@ import uuid
 from abc import ABC, abstractmethod
 from datetime import datetime
 from os import path
-from typing import IO, Generic, NamedTuple, Protocol, TypeVar
+from typing import IO, TYPE_CHECKING, Generic, NamedTuple, Protocol, TypeVar
 
+import orjson
 import sentry_sdk
+from django.conf import settings
 from django.db import IntegrityError, router
 from django.db.models import Q
 from django.utils import timezone
@@ -39,12 +41,15 @@ from sentry.models.release import Release
 from sentry.models.releasefile import ReleaseArchive, ReleaseFile, update_artifact_index
 from sentry.silo.base import SiloMode
 from sentry.tasks.base import instrumented_task
-from sentry.utils import metrics
+from sentry.utils import metrics, redis
 from sentry.utils.db import atomic_transaction
 from sentry.utils.files import get_max_file_size
 from sentry.utils.sdk import bind_organization_context, configure_scope
 
 logger = logging.getLogger(__name__)
+
+if TYPE_CHECKING:
+    from rediscluster import RedisCluster
 
 
 class ChunkFileState:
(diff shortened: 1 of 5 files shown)


src/sentry/tasks/assemble.py
```diff
@@ -179,26 +189,38 @@ def get_assemble_status(task, scope, checksum):
     notice or error message.
     """
     cache_key = _get_cache_key(task, scope, checksum)
-    rv = default_cache.get(cache_key)
+
+    if options.get("assemble.read_from_redis"):
+        client = _get_redis_cluster_for_assemble()
+        rv = client.get(cache_key)
```
markstory, line 195: Will this deserialize the tuple that was string cast?

src/sentry/tasks/assemble.py
```diff
@@ -179,26 +189,38 @@ def get_assemble_status(task, scope, checksum):
...
 
+@sentry_sdk.tracing.trace
 def set_assemble_status(task, scope, checksum, state, detail=None):
     """
     Updates the status of an assembling task. It is cached for 10 minutes.
     """
     cache_key = _get_cache_key(task, scope, checksum)
     default_cache.set(cache_key, (state, detail), 600)
+    redis_client = _get_redis_cluster_for_assemble()
+    redis_client.set(name=cache_key, value=str((state, detail)), ex=600)
```
markstory, line 212: Should we be json encoding a list, so that we can de-serialize it on read?

anonrig, line 212: @asottile-sentry Is the `str()` correct for passing tuples to Redis?

asottile-sentry: yeah this probably needs to be explicitly json encoded and decoded

src/sentry/tasks/assemble.py
```diff
@@ -164,12 +168,18 @@ def _get_cache_key(task, scope, checksum):
...
-                str(task).encode("utf-8"),
+                str(task).encode(),
             )
         ).hexdigest()
     )
 
 
+def _get_redis_cluster_for_assemble() -> RedisCluster:
+    cluster_key = settings.SENTRY_ASSEMBLE_CLUSTER
+    return redis.redis_clusters.get(cluster_key)  # type: ignore[return-value]
```
asottile-sentry, line 179: what's the ignore here? this should be typesafe I imagine (maybe it's expecting `RedisCluster[str]` ? or maybe we need a narrowing assertion that this isn't a non-Cluster?)

loewenheim (approved):

src/sentry/tasks/assemble.py
asottile-sentry: wouldn't the normal usage of the return value break self hosted as well? should this preserve the union?

anonrig: Right now, it doesn't break because Cluster and Redis client has the same exact API surface. It will break this and many other places when we upgrade redis and remove py-cluster and rb. 

I don't know which scenario is better.

asottile-sentry: right -- then I think this should `-> RedisCluster | rb.Cluster` instead of `# type: ignore`

anonrig: But technically this can be `StrictRedis` as well :-)


And on `utils/redis.py` we only support `RedisCluster` and `StrictRedis`.

```
self._clusters_bytes: dict[str, RedisCluster[bytes] | StrictRedis[bytes]] = {}
```


Which of them is correct?

image

src/sentry/tasks/assemble.py
anonrig: Yeah, fixed it. Thanks!

markstory (commented):
Logic to read/write values looks good to me once the type issues are sorted.

asottile-sentry (approved):

PR #70352, opened 2024-05-06
ref: improve typing for sentry.utils.performance_issues

asottile-sentry:

### DIFF
--- a/src/sentry/utils/performance_issues/detectors/consecutive_db_detector.py
+++ b/src/sentry/utils/performance_issues/detectors/consecutive_db_detector.py
@@ -83,7 +83,7 @@ class ConsecutiveDBSpanDetector(PerformanceDetector):
     def _add_problem_span(self, span: Span) -> None:
         self.consecutive_db_spans.append(span)
 
-    def _validate_and_store_performance_problem(self):
+    def _validate_and_store_performance_problem(self) -> None:
         self._set_independent_spans(self.consecutive_db_spans)
         if not len(self.independent_db_spans):
             return
@@ -188,7 +188,7 @@ class ConsecutiveDBSpanDetector(PerformanceDetector):
 
         return self.consecutive_db_spans[0].get("description", "")
 
-    def _set_independent_spans(self, spans: list[Span]):
+    def _set_independent_spans(self, spans: list[Span]) -> None:
         """
         Given a list of spans, checks if there is at least a single span that is independent of the rest.
         To start, we are just checking for a span in a list of consecutive span without a WHERE clause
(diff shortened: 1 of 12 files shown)


src/sentry/utils/performance_issues/detectors/consecutive_db_detector.py
```diff
@@ -261,7 +261,7 @@ def is_creation_allowed_for_project(self, project: Project) -> bool:
         return self.settings["detection_enabled"]
 
     @classmethod
-    def is_event_eligible(cls, event, project: Project | None = None) -> bool:
+    def is_event_eligible(cls, event: dict[str, Any], project: Project | None = None) -> bool:
```
gggritso, line 264: What's the current meta for Python typing? `dict[str, Any]` or `Mapping[str, Any]`?

src/sentry/utils/performance_issues/detectors/io_main_thread_detector.py
```diff
@@ -224,10 +224,10 @@ def __init__(self, settings: dict[DetectorType, Any], event: dict[str, Any]) ->
         self.mapper = None
         self.parent_to_blocked_span = defaultdict(list)
 
-    def _fingerprint(self, span_list) -> str:
+    def _fingerprint(self, span_list: list[Span]) -> str:
```
gggritso, line 227: Same question about meta, `list[Span]` vs. `Iterable[Span]`

src/sentry/utils/performance_issues/detectors/io_main_thread_detector.py
```diff
@@ -224,10 +224,10 @@ def __init__(self, settings: dict[DetectorType, Any], event: dict[str, Any]) ->
...
         description_strings = []
         for span in span_list:
-            description_strings.append(span.get("description"))
+            description_strings.append(span.get("description", ""))
```
gggritso, line 230: Note to self: `"-".join([None])` fails, so this makes sense. I don't think any SDK actually sends a span that has well-formed main thread data for an IO span but doesn't have a description. A little funny that the code doesn't actually check for this

gggritso (approved):
Looks good, thanks! One question about typing, just for my personal curiosity

src/sentry/utils/performance_issues/detectors/consecutive_db_detector.py
asottile-sentry: https://typing.readthedocs.io/en/latest/source/best_practices.html#arguments-and-return-types

though generally these are easy enough to be concrete types since we are not dealing with variance and we are not providing a public api

in theory we should be serializing event dicts somewhere so they're not just a big spooky bag of unknown attributes -- but that's probably a bigger project

Issue #70346, opened 2024-05-06
[User Feedback] Enable 'Spam' ui inbox & buttons


ryan953:
It's this one: https://flagr.getsentry.net/#/flags/539

PR #70353, opened 2024-05-06
chore(feedback): show mark as spam buttons & spam mailbox for everyone

michellewzhang:
relates to https://github.com/getsentry/sentry/issues/70346

shows the manual spam feature in these spots:

mailbox:
SCR-20240506-kuku

bulk action:
SCR-20240506-kumo

single feedback action:
SCR-20240506-kuom


ryan953:
To start we should just flip the feature switch ( https://flagr.getsentry.net/#/flags/539) then we can remove the code afterwards

michellewzhang:
> To start we should just flip the feature switch ( https://flagr.getsentry.net/#/flags/539) then we can remove the code afterwards

just added all orgs to the flag in flagr

JoshFerge (approved):

ryan953 (approved):

PR #70356, opened 2024-05-06
ref: fix typing by converting RunTaskWithMultiprocessing to a plain function

asottile-sentry:

### DIFF
--- a/src/sentry/utils/arroyo.py
+++ b/src/sentry/utils/arroyo.py
@@ -5,6 +5,7 @@ from collections.abc import Callable, Mapping
 from functools import partial
 from typing import Any
 
+from arroyo.processing.strategies.run_task import RunTask
 from arroyo.processing.strategies.run_task_with_multiprocessing import (
     MultiprocessingPool as ArroyoMultiprocessingPool,
 )
@@ -12,7 +13,7 @@ from arroyo.processing.strategies.run_task_with_multiprocessing import (
     RunTaskWithMultiprocessing as ArroyoRunTaskWithMultiprocessing,
 )
 from arroyo.processing.strategies.run_task_with_multiprocessing import TResult
-from arroyo.types import TStrategyPayload
+from arroyo.types import Message, TStrategyPayload
 from arroyo.utils.metrics import Metrics
 from django.conf import settings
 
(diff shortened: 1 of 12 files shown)


src/sentry/utils/arroyo.py
```diff
@@ -157,43 +158,36 @@ def close(self) -> None:
             self.__pool.close()
 
 
-class RunTaskWithMultiprocessing(ArroyoRunTaskWithMultiprocessing[TStrategyPayload, TResult]):
+def run_task_with_multiprocessing(
+    *,
+    pool: MultiprocessingPool,
+    function: Callable[[Message[TStrategyPayload]], TResult],
```
asottile-sentry, line 164: splitting this parameter out is important so that the generics can be resolved correctly

src/sentry/utils/arroyo.py
```diff
@@ -157,43 +158,36 @@ def close(self) -> None:
...
+    **kwargs: Any,
```
asottile-sentry, line 165: technically these parameters are unchecked -- ideally they'd be forwarded along to the underlying callers but 🤷

asottile-sentry (commented):
best viewed with [?w=1](https://github.com/getsentry/sentry/pull/70356/files?w=1)

evanpurkhiser:
Wanted to get @lynnagara and @untitaker's eyes on this

untitaker (commented):
i don't quite understand why these changes are necessary, i will try it a bit later to understand what's going on with the type. the class is supposed to be a drop-in replacement for another arroyo class, so I fear that by making it a function it cannot be used as a replacement 100% of the time (even though the return type does look compatible to me, the function itself is of the wrong type)

was this to fix type issues in usage or the internal uses of type:ignore?

asottile-sentry:
> i don't quite understand why these changes are necessary, i will try it a bit later to understand what's going on with the type. the class is supposed to be a drop-in replacement for another arroyo class, so I fear that by making it a function it cannot be used as a replacement 100% of the time (even though the return type does look compatible to me, the function itself is of the wrong type)
> 
> was this to fix type issues in usage or the internal uses of type:ignore?

both -- the implementation on the primary branch is unsound because `__new__` returns instances of unrelated types (and with their generics always as `Any` (!))

the code here is entirely equivalent -- but restores soundness and enables `TResult` to be used safely

asottile-sentry:
~technically the only difference with it being a function is it cannot be used with `isinstance` / `issubclass` -- but that was already the case because `__new__` was unsound (and it's not being used for `isinstance` anywhere)

untitaker (approved):
right okay, i realize now that this is inside `__new__`, that was a dumb hack anyway. i think in terms of API contract RunTaskWithMultiprocessing should be equal to any other ProcessingStrategy and only differ in `__init__`, but it this was always a little bit sketchy and not enforced

src/sentry/utils/arroyo.py
untitaker: ah i think that's why i did it this way actually, with it being a subclass the user of this strategy/class/function has typed args

untitaker: BTW, don't let this convo be a blocker from merging, I don't really care about retaining this property

PR #70360, opened 2024-05-06
feat(cache): add item size column

DominikB2014:
Adds cache.item_size column to the cache module
image


static/app/views/performance/cache/tables/transactionsTable.tsx
```diff
@@ -49,6 +60,11 @@ const COLUMN_ORDER: Column[] = [
     name: t('Project'),
     width: COL_WIDTH_UNDEFINED,
   },
+  {
+    key: `avg(${CACHE_ITEM_SIZE})`,
+    name: t('Avg Value Size'),
```
0Calories, line 65: should this be called `Avg Item Size` instead?

0Calories (approved):

static/app/views/performance/cache/tables/transactionsTable.tsx
DominikB2014: Yeah i see the confusion here, It should be value size, cause that's the term you typically use for caching (key/value). The SDK sends it as `item_size` to we have to mix the terms unfortunately.

PR #70362, opened 2024-05-06
fix(staff): Remove incorrrect u2f error message

schew2381:
Hopefully this fixes the random u2f error msg popping up when it shouldn't be

### DIFF
--- a/static/app/components/superuserStaffAccessForm.tsx
+++ b/static/app/components/superuserStaffAccessForm.tsx
@@ -63,7 +63,7 @@ class SuperuserStaffAccessForm extends Component<Props, State> {
     this.setState({authenticators: authenticators});
 
     // Set the error state if there are no authenticators and U2F is on
-    if (!this.state.authenticators.length && !disableU2FForSUForm) {
+    if (!authenticators.length && !disableU2FForSUForm) {
       this.handleError(ErrorCodes.NO_AUTHENTICATOR);
     }
     this.setState({isLoading: false});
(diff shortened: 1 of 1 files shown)


static/app/components/superuserStaffAccessForm.tsx
```diff
@@ -59,7 +59,9 @@ class SuperuserStaffAccessForm extends Component<Props, State> {
       return;
     }
 
-    await this.getAuthenticators();
+    const authenticators = await this.getAuthenticators();
+    this.setState({authenticators: authenticators});
+
     // Set the error state if there are no authenticators and U2F is on
     if (!this.state.authenticators.length && !disableU2FForSUForm) {
```
dashed, line 66: `componentDidMount()` will only run once on mount, so this part needs to be updated as well:

```suggestion
    if (!authenticators && !disableU2FForSUForm) {
```

schew2381: wait do I need to keep the `.length` part?

dashed: @schew2381 ah yea that too.

schew2381: alright done

dashed (approved):

static/app/components/superuserStaffAccessForm.tsx
```diff
@@ -59,9 +59,11 @@ class SuperuserStaffAccessForm extends Component<Props, State> {
       return;
     }
 
-    await this.getAuthenticators();
+    const authenticators = await this.getAuthenticators();
+    this.setState({authenticators: authenticators});
```
sentaur-athena, line 63: I don't see this state being modified anywhere else. Does it even need to be state?

schew2381: Would removing it look like this? https://github.com/getsentry/sentry/pull/70365/

PR #70364, opened 2024-05-06
fix(highlights): Fix crashing on null tag keys

leeandher:
Resolves https://sentry.sentry.io/issues/5243747663/?project=11276

Not sure why tags are coming in with `null` or `undefined` keys, but we can't display those so they will be omitted, letting the rest of the component render as per usual.

Also adding the team as a codeowner to help assignment for this stuff.

### DIFF
--- a/static/app/components/events/eventTags/eventTagsTree.tsx
+++ b/static/app/components/events/eventTags/eventTagsTree.tsx
@@ -7,6 +7,7 @@ import EventTagsTreeRow, {
 import {useIssueDetailsColumnCount} from 'sentry/components/events/eventTags/util';
 import {space} from 'sentry/styles/space';
 import type {Event, EventTag} from 'sentry/types/event';
+import {defined} from 'sentry/utils';
 
 const MAX_TREE_DEPTH = 4;
 const INVALID_BRANCH_REGEX = /\.{2,}/;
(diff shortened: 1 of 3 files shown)


roggenkemper (approved):

static/app/components/events/eventTags/eventTagsTree.tsx
```diff
@@ -43,6 +44,10 @@ function addToTagTree(
   originalTag: EventTag
 ): TagTree {
   const BRANCH_MATCHES_REGEX = /\./g;
+  if (!defined(tag.key)) {
+    return tree;
+  }
```
asottile-sentry, line 49: shouldn't the backend api which is producing garbage be fixed instead?  this just hides the bug

asottile-sentry (dismissed):

static/app/components/events/eventTags/eventTagsTree.tsx
leeandher: There might be a valid reason for these `null` keys even if I wasn't able to replicate it or test with it myself, it would be hasty to make changes like that since this UI is behind a flag and the previous UI had no issues -- It has been receiving `null` keys for some time with no issues, it'd seem.

This error was triggered from an EA group accessing a new interface I built for the frontend, so this fix provides the best compatibility with users who don't have access to the flags, without changing their experience of the legacy UI.

I will look into where the null keys come from to make sure we can account for it in the new UI, but for now I find this to be the best approach.

asottile-sentry: have you looked into what's causing the `null`?  I really think we should be striving for quality here instead of hacking around our own bugs in the frontend

PR #70365, opened 2024-05-06
nit: Remove authenticators from state

schew2381:
Remove `authenticators` from state because we only set it once

### DIFF
--- a/static/app/components/superuserStaffAccessForm.tsx
+++ b/static/app/components/superuserStaffAccessForm.tsx
@@ -38,7 +38,6 @@ class SuperuserStaffAccessForm extends Component<Props, State> {
   constructor(props) {
     super(props);
     this.authUrl = this.props.hasStaff ? '/staff-auth/' : '/auth/';
-    this.authenticators = [];
     this.state = {
       error: false,
       errorType: '',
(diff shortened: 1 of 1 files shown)


static/app/components/superuserStaffAccessForm.tsx
```diff
@@ -60,16 +59,17 @@ class SuperuserStaffAccessForm extends Component<Props, State> {
     }
 
     const authenticators = await this.getAuthenticators();
-    this.setState({authenticators: authenticators});
+    this.authenticators = authenticators;
```
nhsiehgit, line 62: :squint:

Ok so - we _shouldn't_ be utilizing non lifecycle state variables like this i don't think 🤔

nhsiehgit (commented):
mmmm i'm not quite sure about this change 🤔

static/app/components/superuserStaffAccessForm.tsx
```diff
@@ -39,8 +38,8 @@ class SuperuserStaffAccessForm extends Component<Props, State> {
   constructor(props) {
     super(props);
     this.authUrl = this.props.hasStaff ? '/staff-auth/' : '/auth/';
+    this.authenticators = [];
```
saponifi3d, line 41:
```suggestion
```
I'd recommend removing this and define it inline where you have the types , like this:
```
authenticators: Authenticator[] = [];
```
Which will do the same as what you have here.

saponifi3d (approved):
:shipit: just had the nit.

cathteng:
shoot i reverted this by accident, i can put it back up

PR #70368, opened 2024-05-06
fix(replay): Remove View Sample Replay button

c298lee:
Removes view sample replay button from replay onboarding because we're not supposed to rely on Arcade inside Sentry.

Relates to https://github.com/getsentry/team-replay/issues/433 and https://github.com/getsentry/getsentry/pull/13875


michellewzhang (approved):
🚮

PR #70372, opened 2024-05-06
ref(hybrid-cloud): Remove success logs on integration proxying

leeandher:
Inspired by @JoshFerge, removing some more logs that are too large in volume to be useful. The useful data is already tracked in metrics so we can just retain logging for failure states and debugging instead.


mrduncan (approved):
Nice!

PR #70373, opened 2024-05-06
feat(perf): Update the Spans tab in Transaction Summary

0Calories:
Upgrades the `Spans` tab on the transaction summary to query from the Span metrics dataset and changes the table to have columns more relevant to span metrics.

### Before
image

### After
image

**TODO in followup PRs**
- [ ] Searchbar needs to be implemented (and not auto-suggest irrelevant transaction tags that don't work)


static/app/views/performance/transactionSummary/transactionSpans/spanMetricsTable.tsx
```diff
@@ -0,0 +1,197 @@
...
+export default function SpanMetricsTable(props: Props) {
+  const {project, transactionName} = props;
+  const organization = useOrganization();
+
+  const location = useLocation();
+  const spansCursor = decodeScalar(location.query?.[QueryParameterNames.SPANS_CURSOR]);
+  const spanOp = decodeScalar(location.query?.spanOp);
+
+  const filters: SpanMetricsQueryFilters = {
+    transaction: transactionName,
```
shruthilayaj, line 99: should this also include transaction op and http op filters?

shruthilayaj (approved):

PR #70382, opened 2024-05-06
feat(cache): add cache widget

DominikB2014:
Add cache widget
image

### DIFF
--- a/static/app/views/performance/landing/widgets/widgets/lineChartListWidget.tsx
+++ b/static/app/views/performance/landing/widgets/widgets/lineChartListWidget.tsx
@@ -1,6 +1,7 @@
 import {Fragment, useMemo, useState} from 'react';
 import styled from '@emotion/styled';
 import pick from 'lodash/pick';
+import * as qs from 'query-string';
 
 import Accordion from 'sentry/components/accordion/accordion';
 import {LinkButton} from 'sentry/components/button';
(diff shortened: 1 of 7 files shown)


static/app/views/performance/landing/widgets/widgets/lineChartListWidget.tsx
```diff
@@ -624,6 +684,30 @@ export function LineChartListWidget(props: PerformanceWidgetProps) {
                   </Fragment>
                 </RoutingContextProvider>
               );
+            case PerformanceWidgetSetting.HIGHEST_CACHE_MISS_RATE_TRANSACTIONS:
+              const cacheMissRate = listItem[fieldString];
+              const target = normalizeUrl(
+                `${CACHE_BASE_URL}/?${qs.stringify({transaction: transaction})}`
+              );
```
narsaynorath, line 691:
```suggestion
              const target = normalizeUrl({
                pathname: CACHE_BASE_URL,
                query: {transaction: transaction},
              });
```

I think this also works and avoids the extra import to force this into a string if you want to use it

narsaynorath (approved):
Looks good to me. I found some things during testing like:
1. Could use a "View All" button in the top right that links to the main cache module page
2. Found a weird bug where the data for the samples panel wasn't really getting anything
3. Clicking "Try Different Samples" seemed to always return the same ones, although this might be from sampling and not a caching issue like I had thought

None of these are blockers for this PR

PR #70390, opened 2024-05-06
chore(actor) Remove ActorTuple and use RpcActor instead

markstory:
Replace our usage of `ActorTuple` with `RpcActor`. These classes are very similar but `RpcActor` has far wider usage both in sentry and getsentry. I plan to rename `RpcActor` to `Actor` once the `Actor` model has been removed (soon).


evanpurkhiser (approved):

src/sentry/api/serializers/models/alert_rule.py
```diff
@@ -197,8 +197,8 @@ def get_attrs(
             )
             result[item]["activations"] = serialize(activations, **kwargs)
 
-            actor = ActorTuple.from_id(user_id=item.user_id, team_id=item.team_id)
-            if actor:
+            if item.user_id or item.team_id:
+                actor = RpcActor.from_id(user_id=item.user_id, team_id=item.team_id)
```
wedamija, line 201: Wondering if the interface for `from_id` should just return `None` if both ids are `None`?

src/sentry/api/serializers/rest_framework/mentions.py
```diff
@@ -43,9 +42,9 @@ def extract_user_ids_from_mentions(organization_id, mentions):
     }
 
 
-def separate_actors(actors):
-    users = [actor for actor in actors if actor.type is User]
-    teams = [actor for actor in actors if actor.type is Team]
+def separate_actors(actors: Sequence[RpcActor]):
+    users = [actor for actor in actors if actor.actor_type == ActorType.USER]
+    teams = [actor for actor in actors if actor.actor_type is ActorType.TEAM]
```
wedamija, line 47: Why `is` vs `==`?

wedamija (approved):

src/sentry/api/serializers/rest_framework/mentions.py
markstory: That's a mistake, I'll align on `==`.

PR #70398, opened 2024-05-06
cogs(access-logs): dont log internal methods

JoshFerge:
The purpose of access logs is to monitor how our API is used externally. We shouldn't log access info for internal API calls. volume wise, this is mainly RPC methods taking up the bulk of the logs. https://cloudlogging.app.goo.gl/qzdQHVAyBtPtAPk49

### DIFF
--- a/src/sentry/middleware/access_log.py
+++ b/src/sentry/middleware/access_log.py
@@ -19,6 +19,8 @@ from . import is_frontend_request
 
 api_access_logger = logging.getLogger("sentry.access.api")
 
+EXCLUSION_PATHS = settings.ACCESS_LOGS_EXCLUDE_PATHS + settings.ANONYMOUS_STATIC_PREFIXES
+
 
 @dataclass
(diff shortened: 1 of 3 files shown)


src/sentry/middleware/access_log.py
```diff
@@ -125,8 +127,9 @@ def middleware(request: Request) -> Response:
         if not settings.LOG_API_ACCESS:
             return get_response(request)
 
-        if request.path_info.startswith(settings.ANONYMOUS_STATIC_PREFIXES):
+        if request.path_info.startswith(EXCLUSION_PATHS):
```
vartec, line 130: TIL that `startswith` can takes tuple of prefixes.

vartec (approved):

src/sentry/middleware/access_log.py
JoshFerge: 😅 same

PR #70399, opened 2024-05-06
OPS-5392 add datadog event support

nirajdewani:
## Description

I'd like to add support for [Datadog Events](https://docs.datadoghq.com/service_management/events/guides/dogstatsd/).
Pasting context from [OPS-5392](https://getsentry.atlassian.net/browse/OPS-5392):
> Currently back pressure sends metrics to Datadog ([https://app.datadoghq.com/dashboard/qnx-z63-3di/backpressure-monitoring](https://app.datadoghq.com/dashboard/qnx-z63-3di/backpressure-monitoring?fromUser=false&refresh_mode=paused&view=spans&from_ts=1709132829007&to_ts=1709137896990&live=false)).
We should send datadog events instead (or as well): https://docs.datadoghq.com/service_management/events/ 
This would enable us to do things like:
>- Trigger DD monitors for alerting purposes (eg. notifying slack / PagerDuty)
>- overlay events on top of dashboard widgets
>- add tags to events to support analysis / grouping

## (Potential) To Do
Added a unit test; please let me know if/how I can test this better.

[OPS-5392]: https://getsentry.atlassian.net/browse/OPS-5392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/sentry/metrics/base.py
```diff
@@ -78,3 +78,18 @@ def distribution(
...
+        alert_type: str | None = None,
+        aggregation_key: str | None = None,
+        source_type_name: str | None = None,
+        date_happened: int | None = None,
+        priority: str | None = None,
+        instance: str | None = None,
+        tags: Tags | None = None,
+        stacklevel: int = 0,
+    ) -> None:
+        raise NotImplementedError
```
iambriccardo, line 95: This will backfire when we call `.event()` and the backend configuration uses a `MetricsBackend` implementation that doesn't support this method.

For example, we are running in production the `CompositeExperimentalBackend` so it will crash Sentry immediately if we deploy this PR and call `.event()`.

I think we might have two ways of proceeding:
1. We add `pass` in the parent method.
2. We implement the `event` method in each subclass with `pass` in the body.

Considering the complexity of the change I would opt for 1:
```suggestion
        pass
```

src/sentry/metrics/datadog.py
```diff
@@ -111,3 +111,36 @@ def distribution(
     ) -> None:
         # We keep the same implementation for Datadog.
         self.timing(key, value, instance, tags, sample_rate)
+
+    def event(
```
iambriccardo, line 115: Do we want to add this also for the `dogstatsd` backend?

src/sentry/metrics/base.py
```diff
@@ -78,3 +78,18 @@ def distribution(
         stacklevel: int = 0,
     ) -> None:
         raise NotImplementedError
+
+    def event(
```
anonrig, line 82: No-one is calling this particular function right now. We should do it in this PR to eliminate the chance of having dead-code in the project

fpacifici (changes requested):
Please see comments inline

src/sentry/metrics/base.py
nirajdewani: Reverted to raising error now that the method is implemented for every derived class.
Lemme know if we should still `pass`.

src/sentry/metrics/base.py
nirajdewani: We are planning to use it in the [ops](https://github.com/getsentry/sentry/blob/699ce0cd1281cc3c7275d0a474a595375c769ae8/src/sentry/processing/backpressure/health.py) repo actually.

src/sentry/metrics/base.py
```diff
@@ -78,3 +78,18 @@ def distribution(
...
+        self,
+        key: str,
+        value: str,
+        alert_type: str | None = None,
+        aggregation_key: str | None = None,
+        source_type_name: str | None = None,
+        date_happened: int | None = None,
```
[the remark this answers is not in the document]
nirajdewani: Verified that `date_happened` will default to system time in case it is not passed. [Docs](https://datadogpy.readthedocs.io/en/latest/#datadog.api.Event).
So removed it.

src/sentry/metrics/base.py
fpacifici: We are not planning to use it in the ops repo. The ops repo does not access getsentry.
We are planning to use it in the backpressure system in sentry: https://github.com/getsentry/sentry/blob/master/src/sentry/processing/backpressure/health.py

@anonrig , that will happen in the PR @nirajdewani will write right after this one. So no issue.

src/sentry/metrics/base.py
fpacifici: Please stick to NotImplementedError here. This is an abstract class (even if it is regrettably not defined as such), so it must not be instantiated directly. As a matter of fact all the other methods in this class are raising the same exception.

The right solution is to provde a valid implementation in all subclasses, like Niraj did.

fpacifici: @nirajdewani  Have you checked if there is any subclass of this one in the `getsentry` repo?
If there is some there you will have to either add the method implementation first there and then make the change here, or actually make it temporarily `pass` here, then implement the method there and then revert this one to NotImplementedError

src/sentry/metrics/composite_experimental.py
```diff
@@ -130,3 +130,17 @@ def distribution(
...
+        message: str,
+        alert_type: str | None = None,
+        aggregation_key: str | None = None,
+        source_type_name: str | None = None,
+        priority: str | None = None,
+        instance: str | None = None,
+        tags: Tags | None = None,
+        stacklevel: int = 0,
+    ) -> None:
+        pass
```
fpacifici, line 146: If this is what we use in getsentry production, this will not work as you will not be producing any event to datadog in prod. 
You'll have to forward the call to `self._primary_banckend`

fpacifici (dismissed):
Please see comments in line

src/sentry/metrics/base.py
nirajdewani: Thanks, I checked; `InternalMetrics` in `src/sentry/utils/metrics.py` was the only subclass that was missed.
So should be okay to keep `raise NotImplementedError`.

iambriccardo (approved):
LGTM

mwarkentin (approved):

PR #70402, opened 2024-05-07
fix(middleware): Integration Fetching Logic for Gitlab Parser

iamrajjoshi:
The logic inside  `get_integration_from_request` is flawed. 

The check for "organization_slug" and "integration_id" will never return True because the function is only called when `get_response_from_gitlab_webhook` is called.

`get_response_from_gitlab_webhook` is [called](https://github.com/getsentry/sentry/blob/master/src/sentry/middleware/integrations/parsers/gitlab.py#L104-L107) only when the Endpoint is `GitlabWebhookEndpoint`. This endpoint is a webhook endpoint which will never have the above path params.

I have put this change behind an option so we can toggle it quickly. We will monitor the metric and if it is safe to do so, we will remove the option and the logic.


schew2381 (approved):
lgtm

PR #69759, opened 2024-04-26
ref(list-box): Remove context usage

ArthurKnaus:
Remove compact select specific context usage from list box to make it easier to re-use.

*I recommend hiding whitespace changes when reviewing this PR*

PR #70406, opened 2024-05-07
Revert "ref(list-box): Remove context usage"

scttcper:
Reverts getsentry/sentry#69759

dropdowns have stopped hiding things when filtering in some cases and performance of searchable compact select seems poor


scttcper:
hmm maybe there's just something missing

scttcper:
we can do this instead https://github.com/getsentry/sentry/pull/70408

PR #70415, opened 2024-05-07
feat(ddm): Add support for querying multiple use case ids in the meta endpoints

iambriccardo:
This PR adds support for querying multiple use case IDs while fetching metrics. The need for this arises from the fact that we want to allow querying all public use cases in certain situations (internal admin).

For the metrics product itself, we will always query the endpoint with a single use case per request but it's nice to have the option to load multiple ones.

I have implemented the simple version that just makes 4 queries for each use case id, to avoid using the `IN` operator + group by use case but in case we see a substantial performance problem we can switch the implementation to using the `IN`.


src/sentry/snuba/metrics_layer/query.py
```diff
@@ -549,24 +559,26 @@ def fetch_metric_tag_keys(
...
     Helper function for querying the meta table. This will query across all four metric types, and resolve all the resulting
     values. If an MRI is provided, it is assumed that this function should find unique tag keys for that MRI.
     """
+    if len(use_case_ids) == 0:
+        raise InvalidParams("You must supply at least one use case id")
 
     if mri:
         column_name = "tag_key"
-        metric_id = resolve_weak(use_case_id, org_id, mri)
+        metric_id = resolve_weak(use_case_ids[0], org_id, mri)
```
evanh, line 581: How does this work if the MRI has a different use case from `use_case_ids[0]`?

src/sentry/snuba/dataset.py
```diff
@@ -73,6 +73,10 @@ class EntityKey(Enum):
     GenericMetricsSets = "generic_metrics_sets"
     GenericMetricsCounters = "generic_metrics_counters"
     GenericMetricsGauges = "generic_metrics_gauges"
+    GenericMetricsDistributionsMeta = "generic_metrics_distributions_meta"
```
evanh, line 76: Why did you add these? I don't think the meta tables should be queryable outside of the metrics layer.

src/sentry/snuba/metrics_layer/query.py
```diff
@@ -532,14 +532,24 @@ def convert_snuba_result(
     return snuba_result
 
 
+def fetch_metric_mris_for_use_cases(
```
evanh, line 535: I think you should make this `_fetch_metric_mris_for_use_cases`, to drive home the point that this is not the accepted way to query this data.

evanh (commented):
I'm not sure I understand the utility of this change. If the need is just for an internal admin product, why not call the function multiple times for each use case? Why create an opportunity for misuse?

src/sentry/snuba/metrics_layer/query.py
iambriccardo: It works in exactly the same way as your implementation, it will fail resolution.

src/sentry/snuba/metrics_layer/query.py
iambriccardo: This is a good point.

src/sentry/snuba/dataset.py
iambriccardo: The thing is that this entity key can be queried nonetheless so I would prefer to centralized its definition. The same was done for generic org counters, they are not needed outside of dynamic sampling but they are defined here for clarity.

iambriccardo:
> I'm not sure I understand the utility of this change. If the need is just for an internal admin product, why not call the function multiple times for each use case? Why create an opportunity for misuse?

The rationale of the change is that I do not want to sequentially call the function, I would like to parallelize as much as possible to not slow down admin too much.

evanh:
> I would like to parallelize as much as possible to not slow down admin too much.

Which admin tool is this? Why does it matter if it's slower?

src/sentry/snuba/dataset.py
evanh: The only place these entities are queried is the metrics layer. Adding them here implies that querying them outside the metrics layer is an allowed practice, and I don't think it should be.

src/sentry/snuba/metrics_layer/query.py
evanh: I see. This function has been overloaded and the correct call parameters are no longer as clear. I think there should be a check here that if an `mri` is provided, then exactly one `use_case_id` should be provided. Then it is clear what the intended use of this function is.

evanh (changes requested):

iambriccardo:
Thanks for the comments, will change update the PR.

iambriccardo:
Decided offline to not merge this change and see how sequential queries perform.

PR #70416, opened 2024-05-07
ref(ddm): Use EntityKey instead of hardcoding strings in code

iambriccardo:
This PR slightly refactors meta queries by using the `EntityKey` enum instead of building the entity key string. This enables other people in Sentry to get to know the entities that are available in Snuba.


shellmayr (approved):

evanh (commented):
I mentioned this on your other PR, but I am not convinced these entities should be exposed in a general sense.

The meta tables were built with a specific pattern in mind. If they are used outside of that pattern, there is no guarantee of their performance or utility. I'm not sure how to communicate that through code though.

iambriccardo:
We decided to keep the entities used in meta queries concealed.

PR #70420, opened 2024-05-07
feat(metrics): use new meta tables for metrics meta queries in metrics API

shellmayr:
- Create new functionality in sentry_metrics/querying/metadata that uses new meta tables in Snuba to query metadata about metrics. This should increase query speed considerably.
- Replace calls to sentry.snuba.metrics with calls to sentry_metrics.querying.metadata in the Organization Metrics endpoints to make use of these faster queries in the respective UIs.

### DIFF
--- a/src/sentry/api/endpoints/organization_metrics.py
+++ b/src/sentry/api/endpoints/organization_metrics.py
@@ -37,6 +37,7 @@ from sentry.sentry_metrics.querying.errors import (
     MetricsQueryExecutionError,
 )
 from sentry.sentry_metrics.querying.metadata import MetricCodeLocations, get_metric_code_locations
+from sentry.sentry_metrics.querying.metadata.metrics import get_metrics_meta
 from sentry.sentry_metrics.querying.metadata.tags import get_tag_values
 from sentry.sentry_metrics.querying.metadata.utils import convert_metric_names_to_mris
 from sentry.sentry_metrics.querying.samples_list import get_sample_list_executor_cls
@@ -47,13 +48,7 @@ from sentry.sentry_metrics.use_case_id_registry import (
     get_use_case_id_api_access,
 )
 from sentry.sentry_metrics.utils import string_to_use_case_id
-from sentry.snuba.metrics import (
-    QueryDefinition,
-    get_all_tags,
-    get_metrics_meta,
-    get_series,
-    get_single_metric_info,
-)
+from sentry.snuba.metrics import QueryDefinition, get_all_tags, get_series, get_single_metric_info
 from sentry.snuba.metrics.naming_layer.mri import is_mri
 from sentry.snuba.metrics.utils import DerivedMetricException, DerivedMetricParseException
 from sentry.snuba.referrer import Referrer
@@ -156,10 +151,8 @@ class OrganizationMetricsDetailsEndpoint(OrganizationEndpoint):
         if not projects:
             raise InvalidParams("You must supply at least one project to see its metrics")
 
-        start, end = get_date_range_from_params(request.GET)
-
         metrics = get_metrics_meta(
-            projects=projects, use_case_ids=get_use_case_ids(request), start=start, end=end
+            organization=organization, projects=projects, use_case_ids=get_use_case_ids(request)
         )
 
         return Response(metrics, status=200)
--- a/src/sentry/api/endpoints/organization_metrics.py
+++ b/src/sentry/api/endpoints/organization_metrics.py
@@ -36,10 +36,13 @@ from sentry.sentry_metrics.querying.errors import (
     LatestReleaseNotFoundError,
     MetricsQueryExecutionError,
 )
-from sentry.sentry_metrics.querying.metadata import MetricCodeLocations, get_metric_code_locations
-from sentry.sentry_metrics.querying.metadata.metrics import get_metrics_meta
-from sentry.sentry_metrics.querying.metadata.tags import get_tag_values
-from sentry.sentry_metrics.querying.metadata.utils import convert_metric_names_to_mris
+from sentry.sentry_metrics.querying.metadata import (
+    MetricCodeLocations,
+    convert_metric_names_to_mris,
+    get_metric_code_locations,
+    get_metrics_meta,
+    get_tag_values,
+)
 from sentry.sentry_metrics.querying.samples_list import get_sample_list_executor_cls
 from sentry.sentry_metrics.querying.types import QueryOrder, QueryType
 from sentry.sentry_metrics.use_case_id_registry import (
--- a/src/sentry/api/endpoints/organization_metrics.py
+++ b/src/sentry/api/endpoints/organization_metrics.py
@@ -37,7 +37,7 @@ from sentry.sentry_metrics.querying.errors import (
     MetricsQueryExecutionError,
 )
 from sentry.sentry_metrics.querying.metadata import MetricCodeLocations, get_metric_code_locations
-from sentry.sentry_metrics.querying.metadata.tags import get_tag_keys, get_tag_values
+from sentry.sentry_metrics.querying.metadata.tags import get_tag_values
 from sentry.sentry_metrics.querying.metadata.utils import convert_metric_names_to_mris
 from sentry.sentry_metrics.querying.samples_list import get_sample_list_executor_cls
 from sentry.sentry_metrics.querying.types import QueryOrder, QueryType
@@ -49,6 +49,7 @@ from sentry.sentry_metrics.use_case_id_registry import (
 from sentry.sentry_metrics.utils import string_to_use_case_id
 from sentry.snuba.metrics import (
     QueryDefinition,
+    get_all_tags,
     get_metrics_meta,
     get_series,
     get_single_metric_info,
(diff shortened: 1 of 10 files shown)


src/sentry/api/endpoints/organization_metrics.py
```diff
@@ -250,21 +247,25 @@ def get(self, request: Request, organization, tag_name) -> Response:
...
         try:
-            tag_values = get_tag_values(
-                projects=projects,
-                tag_name=tag_name,
-                metric_names=metric_names,
-                use_case_id=get_use_case_id(request),
-                start=start,
-                end=end,
-            )
+            mris = convert_metric_names_to_mris(metric_names)
```
iambriccardo, line 251: I would drop the support for multiple metric names and restrict the API to a single one. If we make sequential queries for each metric name, we might timeout the request since it takes too long.

src/sentry/sentry_metrics/querying/metadata/metrics.py
```diff
@@ -0,0 +1,103 @@
...
+)
+from sentry.snuba.metrics.utils import BlockedMetric, MetricMeta
+from sentry.snuba.metrics_layer.query import fetch_metric_mris
+
+
+def get_metrics_meta(
+    organization: Organization,
+    projects: Sequence[Project],
+    use_case_ids: Sequence[UseCaseID],
+) -> Sequence[MetricMeta]:
```
iambriccardo, line 21: Can we define our own type for `MetricMeta` which is isolated in this module? I would tie to this the proper serialization. We could do this in a follow up PR, not strictly needed rn, but I would like to remove any dependency we have on `datasource` and `metrics.utils`.

src/sentry/sentry_metrics/querying/metadata/metrics.py
```diff
@@ -0,0 +1,103 @@
...
+    list of project_ids in which the MRI is available as values.
+    """
+    project_ids = [project.id for project in projects]
+    project_id_to_mris = fetch_metric_mris(organization.id, project_ids, use_case_id)
+    mris_to_project_ids = _reverse_mapping(project_id_to_mris)
+
+    return mris_to_project_ids
+
+
+def _reverse_mapping(project_id_to_mris: dict[int, list[str]]):
```
iambriccardo, line 90: nit: maybe make the name less ambiguous since reverse mapping is a term we commonly use for reverse resolving ids to strings in the indexer.

```suggestion
def _convert_to_mri_to_project_ids(project_id_to_mris: dict[int, list[str]]):
```

src/sentry/sentry_metrics/querying/metadata/metrics.py
```diff
@@ -0,0 +1,103 @@
...
+    mris_to_project_ids: dict[str, list[int]] = defaultdict(list)
+
+    mris = set(_flatten([mri for mri in project_id_to_mris.values()]))
+    for mri in mris:
+        for project_id in project_id_to_mris.keys():
+            if mri in project_id_to_mris[project_id]:
+                mris_to_project_ids[mri].append(project_id)
```
iambriccardo, line 97: I am not following the logic here, isn't this simpler:
```suggestion
    mris_to_project_ids = {}        
    for project_id, mris in project_id_to_mris.items():
        for mri in mris:
      	    mris_to_projects_ids.set_default(mri, []).append(project_id)
```

src/sentry/sentry_metrics/querying/metadata/tags.py
```diff
@@ -0,0 +1,60 @@
...
+from sentry.snuba.metrics_layer.query import fetch_metric_tag_keys, fetch_metric_tag_values
+
+
+@dataclass
+class TagValue:
+    key: str
+    value: str
+
+    def __hash__(self):
+        return hash((self.key, self.value))
```
iambriccardo, line 16: Making the dataclass frozen would implement an hash method for you, I would generally implement your own hashing function if the object is mutable or your require equality to be defined on a subset of the fields:
```suggestion
@dataclass(frozen=True)
class TagValue:
    key: str
    value: str
```

src/sentry/sentry_metrics/querying/metadata/tags.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    all_tag_keys: set[str] = set()
+    for mri in mris:
+        for use_case_id in use_case_ids:
+            project_ids = [project.id for project in projects]
+            tag_keys_per_project = fetch_metric_tag_keys(
+                organization.id, project_ids, use_case_id, mri
+            )
+            for tag_keys in tag_keys_per_project.values():
+                all_tag_keys = all_tag_keys.union(tag_keys)
```
iambriccardo, line 37: nit: we can save some allocations and iterations this way:
```suggestion
    all_tag_keys: set[str] = set()
    project_ids = [project.id for project in projects]
    for mri in mris:
        for use_case_id in use_case_ids:
            tag_keys_per_project = fetch_metric_tag_keys(
                organization.id, project_ids, use_case_id, mri
            )
            for tag_keys in tag_keys_per_project.values():
                all_tag_keys = all_tag_keys.union(tag_keys)
```

src/sentry/sentry_metrics/querying/metadata/tags.py
```diff
@@ -0,0 +1,60 @@
...
+    projects: Sequence[Project],
+    use_case_ids: Sequence[UseCaseID],
+    mri: str,
+    tag_key: str,
+) -> list[str]:
+    """
+    Get all available tag values for an MRI and tag key from metrics.
+    """
+    tag_values: set[str] = set()
+    for project in projects:
```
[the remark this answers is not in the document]
iambriccardo: @evanh is there any performance implications for which we have only a single project supported in the tag values query but for the other meta queries not?

src/sentry/sentry_metrics/querying/metadata/tags.py
```diff
@@ -0,0 +1,60 @@
...
+@dataclass
+class TagValue:
+    key: str
+    value: str
+
+    def __hash__(self):
+        return hash((self.key, self.value))
+
+
+def get_tag_keys(
```
iambriccardo, line 19: Are you using this function in this PR?

iambriccardo (commented):
Thanks for the work, left some comments which we should address before rolling it out.

src/sentry/sentry_metrics/querying/metadata/metrics.py
shellmayr: Agreed, we can and should do that, but for now I'd like to keep compliance with the previous format and do that work in a separate PR.

src/sentry/sentry_metrics/querying/metadata/metrics.py
shellmayr: You're absolutely right, thanks for catching this.

src/sentry/sentry_metrics/querying/metadata/metrics.py
shellmayr: Changed

src/sentry/sentry_metrics/querying/metadata/tags.py
shellmayr: We should get rid of this type sooner than later, but at the moment all we need is for this type to be hashable to enable set logic. Adding the frozen keyword adds a level of meaning that would not be intended in my opinion.

src/sentry/sentry_metrics/querying/metadata/tags.py
shellmayr: Absolutely correct, thanks for spotting!

src/sentry/api/endpoints/organization_metrics.py
shellmayr: Added a check and raised an exception if there is more than one metric name in the list - keeping the logic name the same so we can go back if we need it shortly; otherwise I'd then simplify once we know that there are no adverse effects from this change.

src/sentry/sentry_metrics/querying/metadata/tags.py
shellmayr: This is an initial implementation of the function get_tag_keys as a basis for a future PR that will then use this to replace the get_all_tags from snuba/metrics/datasource.py - that logic has proven very hard to detangle, therefore I'd like to move it to a PR of its own.

src/sentry/sentry_metrics/querying/metadata/tags.py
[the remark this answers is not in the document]
shellmayr: Query mode has been changed from single-project to multi-project in this PR: https://github.com/getsentry/sentry/commit/54b747341b9a859adbd945847ca04dec496be462

src/sentry/api/endpoints/organization_metrics.py
```diff
@@ -246,25 +243,45 @@ class OrganizationMetricsTagDetailsEndpoint(OrganizationEndpoint):
 
     def get(self, request: Request, organization, tag_name) -> Response:
         metric_names = request.GET.getlist("metric") or []
+        if len(metric_names) > 1:
+            raise ParseError(
+                "Please supply only a single metric name. Specifying multiple metric names is not supported for this endpoint."
+            )
```
obostjancic, line 249: we should aim to move towards a `metrics/meta/<mri>/` for single metrics info so this is a good middle ground for now!

iambriccardo (approved):
LGTM

obostjancic (approved):
Looks great! 🚢

PR #70427, opened 2024-05-07
fix(issue_stream): Fallback to title if customTitle is null

armenzg:
When an event has a `title` inside of metadata set as `null` the UI would try to use it. In such a case, the UI would not let developers view an Issue from the Issue Stream as it would not be hyperlinked.

This is what it would look like (not clickable issues)
image

In this change, we fall back to the standard title value rather than `metadata.title` which can be `null`.

This happened because of a Relay deployment (https://github.com/getsentry/sentry-options-automator/pull/1367).

### DIFF
--- a/static/app/utils/events.tsx
+++ b/static/app/utils/events.tsx
@@ -126,7 +126,7 @@ export function getTitle(
   grouping = false
 ) {
   const {metadata, type, culprit, title} = event;
-  const customTitle = metadata?.title;
+  const customTitle = metadata?.title || title;
 
   switch (type) {
     case EventOrGroupType.ERROR: {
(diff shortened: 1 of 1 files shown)


static/app/utils/events.tsx
```diff
@@ -178,7 +178,7 @@ export function getTitle(
       };
     case EventOrGroupType.DEFAULT:
       return {
-        title: customTitle ?? metadata.title ?? '',
+        title: customTitle ?? title,
```
armenzg, line 181: As a reference, this is what `customTitle` is set:
https://github.com/getsentry/sentry/blob/c9b49d8c538c58a2a010e714debcb008c4ff1043/static/app/utils/events.tsx#L129

`''` also yields the same problem, thus, we fallback to `title` instead of `''`.

iker-barriocanal:
[Feature flag](https://github.com/getsentry/sentry-options-automator/pull/1367) of normalization in relay.

leeandher (approved):

PR #70430, opened 2024-05-07
fix(pr-comment): don't make an empty comment

cathteng:
The suspect PR comment is only triggered from the suspect commit flow. However, if for some reason we don't find any issue associated with the suspect PR, we shouldn't be commenting.


iamrajjoshi (approved):
🚢

PR #70434, opened 2024-05-07
feat(related_issues): Include project=-1 when opening links

armenzg:
This will ensure that a trace containing issues from multiple projects will be shown to the developer.


scttcper (approved):
you may need to keep in mind that not all orgs support project=-1 depending on billing plan

armenzg:
> you may need to keep in mind that not all orgs support project=-1 depending on billing plan

That's good to know. For now, this is an improvement and not a highly trafficked tab. Thanks, Scott!

PR #70438, opened 2024-05-07
chore(hybrid-cloud): Globally enables multi-region selector

GabeVillalobos:
Globally enables the multi-region selector in SaaS when there are 1 or more regions registered. Updates the backing option to be True by default. Next PR will delete the option entirely, pending removal from options automator.


markstory (approved):
Looks good. There is one reference to this feature flag in getsentry as well in web/identity.py.

GabeVillalobos:
> Looks good. There is one reference to this feature flag in getsentry as well in web/identity.py.

@markstory This should have been addressed by this PR: https://github.com/getsentry/getsentry/pull/13894

Unless there's another reference I missed?

PR #70439, opened 2024-05-07
chore(actor) Remove Actor model from django state

markstory:
Remove the Actor model from django state.

Refs HC-1183


azaslavsky (commented):
Awesome!

I think we need to regenerate a few snapshots (test_releases.py -> test_at_head is on example that immediately comes to mind), and may need to fiddle with the backups.py `create_exhaustive...` methods, but otherwise this looks great relocation-wise.

src/sentry/models/actor.py
```diff
@@ -1,90 +0,0 @@
...
-    )
-    team = FlexibleForeignKey(
-        "sentry.Team",
-        related_name="actor_from_team",
-        db_constraint=True,
-        db_index=True,
-        unique=True,
-        null=True,
-        on_delete=models.CASCADE,
-    )
```
wedamija, line 43: I think we need a separate pr first that drops the constraint, otherwise we can hit errors when attempting to delete teams

wedamija: Or actually, it'd be fine to drop the constraint in this migration as well

markstory: I removed the foreign key between Team -> Actor foreign key in #70090, I'll remove this one now so that both sides are gone.

wedamija (approved):

GabeVillalobos (approved):

markstory:
Both the migration and backup tests are giving me a hard time as Django's ORM state is getting into a bad spot. Because of how Actor is woven into migration history removing it is proving to be more challenging than other models. These changes now only remove the foreign keys and related indexes.

Once I can land these changes, I'll figure out how to remove the rest of the model state and storage.

wedamija:
> Both the migration and backup tests are giving me a hard time as Django's ORM state is getting into a bad spot. Because of how Actor is woven into migration history removing it is proving to be more challenging than other models. These changes now only remove the foreign keys and related indexes.
> 
> Once I can land these changes, I'll figure out how to remove the rest of the model state and storage.

You can feel free to just disable any migration tests that are causing you problems

azaslavsky:
Happy to help debug any backup test errors as well.

vbro (approved):

PR #70444, opened 2024-05-07
ref(feedback): take platform from issue project

michellewzhang:
prevents this bug from happening (showing replay setup for a react native project)

SCR-20240507-ndnd

### DIFF
--- a/static/app/components/feedback/useFetchFeedbackData.tsx
+++ b/static/app/components/feedback/useFetchFeedbackData.tsx
@@ -1,6 +1,6 @@
 import {useEffect, useMemo} from 'react';
 
-import hydrateEventTags from 'sentry/components/feedback/hydrateEventTags';
+import hydrateFeedbackTags from 'sentry/components/feedback/hydrateFeedbackTags';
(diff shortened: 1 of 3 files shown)


static/app/components/feedback/useFetchFeedbackData.tsx
```diff
@@ -1,6 +1,6 @@
 import {useEffect, useMemo} from 'react';
 
-import hydrateEventTags from 'sentry/components/feedback/hydrateEventTags';
+import hydrateFeedbackTags from 'sentry/components/feedback/hydrateFeedbackTags';
```
bruno-garcia, line 3: Seems like an unrelated change? was this intentional?

bruno-garcia (approved):
LGTM other than 1 file seems unrelated

static/app/components/feedback/useFetchFeedbackData.tsx
michellewzhang: renamed the file! bc it's not only event tags

PR #70445, opened 2024-05-07
feat(cache): add average transaction duration to sample sidebar

DominikB2014:
New metric readout for avg transaction duration in sample sidebar
image


static/app/views/performance/cache/samplePanel/samplePanel.tsx
```diff
@@ -76,6 +80,15 @@ export function CacheSamplePanel() {
       referrer: Referrer.SAMPLES_CACHE_METRICS_RIBBON,
     });
 
+  const {data: transactionDurationData, isLoading: isTransactionDurationLoading} =
+    useMetrics({
+      search: MutableSearch.fromQueryObject({
+        transaction: query.transaction,
+      } satisfies MetricsQueryFilters),
+      fields: [`avg(${TRANSACTION_DURATION})`],
```
gggritso, line 88: Nit, but I think it's better to do
```suggestion
      fields: [`avg(${MetricsFields.TRANSACTION_DURATION})`],
```

here. That way it's clear where the constant is actually coming from, and what it means. This shows up all over the place though, so it doesn't have anything to do with this PR specifically

gggritso (approved):

PR #70453, opened 2024-05-07
perf: use orjson on bin, fixtures and scripts

anonrig:
The goal is to get rid of all `sentry.utils.json` usages, and this is a step towards that goal.

### DIFF
--- a/fixtures/apidocs_test_case.py
+++ b/fixtures/apidocs_test_case.py
@@ -1,6 +1,7 @@
 import functools
 import os
 
+import orjson
 from django.conf import settings
 from openapi_core.contrib.django import DjangoOpenAPIRequest, DjangoOpenAPIResponse
 from openapi_core.spec import Spec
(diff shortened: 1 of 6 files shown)


fixtures/apidocs_test_case.py
```diff
@@ -9,16 +10,16 @@
...
 
 
 @requires_snuba
 class APIDocsTestCase(APITestCase):
     @functools.cached_property
     def cached_schema(self):
         path = os.path.join(os.path.dirname(__file__), "../tests/apidocs/openapi-derefed.json")
-        with open(path) as json_file:
-            data = json.load(json_file)
+        # opens the file as binary
```
asottile-sentry, line 20: can probably nuke the comment since `'rb` is fairly explanatory

asottile-sentry (commented):
same for the other `open` comments

fixtures/apidocs_test_case.py
anonrig: Removed them all

asottile-sentry (approved):

PR #70455, opened 2024-05-07
feat(snuba-search): adds logic to handle first and last seen filters for snuba only search

scefali:
This PR adds logic to support filters on the first seen and the last seen of on an issue for the snuba only search.

### DIFF
--- a/src/sentry/search/snuba/executors.py
+++ b/src/sentry/search/snuba/executors.py
@@ -1130,6 +1130,24 @@ class InvalidQueryForExecutor(Exception):
 
 
 class GroupAttributesPostgresSnubaQueryExecutor(PostgresSnubaQueryExecutor):
+    def get_last_seen_filter(self, search_filter: SearchFilter, joined_entity: Entity) -> Condition:
+        # get the max timestamp of the error/search_issue event
+        return Condition(
+            Column("timestamp", joined_entity),
+            Op(search_filter.operator),
+            search_filter.value.raw_value,
+        )
+
+    def get_first_seen_filter(
+        self, search_filter: SearchFilter, joined_entity: Entity
+    ) -> Condition:
+        # use the group first seen from the group dataset
+        return Condition(
+            Column("group_first_seen", self.entities["attrs"]),
+            Op(search_filter.operator),
+            search_filter.value.raw_value,
+        )
+
     def get_basic_group_snuba_condition(
         self, search_filter: SearchFilter, joined_entity: Entity
     ) -> Condition:
@@ -1431,6 +1449,8 @@ class GroupAttributesPostgresSnubaQueryExecutor(PostgresSnubaQueryExecutor):
         "assigned_or_suggested": get_assigned_or_suggested,
         "assigned_to": get_assigned,
         "message": get_message_condition,
+        "first_seen": get_first_seen_filter,
+        "last_seen": get_last_seen_filter,
     }
 
     first_seen = Column("group_first_seen", entities["attrs"])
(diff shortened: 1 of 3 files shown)


src/sentry/search/snuba/executors.py
```diff
@@ -1130,6 +1130,24 @@ class InvalidQueryForExecutor(Exception):
 
 
 class GroupAttributesPostgresSnubaQueryExecutor(PostgresSnubaQueryExecutor):
+    def get_last_seen_filter(self, search_filter: SearchFilter, joined_entity: Entity) -> Condition:
+        # get the max timestamp of the error/search_issue event
```
wedamija, line 1134: This doesn't really get the max does it? Just filters to timestamps before that period.

Does this work correctly for last seen? Last seen should be the aggregated max of each group, not just the timestamp of events.

tests/sentry/issues/endpoints/test_organization_group_index.py
```diff
@@ -3076,6 +3076,45 @@ def test_find_error_by_message_with_snuba_only_search(self):
...
+        )
+        assert len(response.data) == 0, "No issues should be first seen over 24 hours ago"
+        response = self.get_success_response(query="firstSeen:-24h", useGroupSnubaDataset=1)
+        assert len(response.data) == 4, "All issues should be seen in the last 24 hours"
+
+        # Test lastSeen filter
+        response = self.get_success_response(query="lastSeen:-6h", useGroupSnubaDataset=1)
+        assert (
+            len(response.data) == 1
+        ), "Only the most recent issue should be seen in the last 6 hours"
```
wedamija, line 3113: We should have a test here with an absolute date, like `lastSeen:<2024-05-07T00:00:00-07:00`

src/sentry/search/snuba/executors.py
scefali: @wedamija hmm...good point. Need to think about how to restructure this because we can't use an aggregation in a where clause.

wedamija: Yeah, it can be in having, so we'd need some switch logic to enable that. We use that aggregate logic for sorting... at the very least, we filter on it during pagination https://github.com/getsentry/sentry/blob/aa29dbf1bb5afec939cb73e138228ce5f59efb5b/src/sentry/search/snuba/executors.py#L348-L349

Not sure if there are actual user accessible queries that use the aggregates

src/sentry/search/snuba/executors.py
```diff
@@ -1538,11 +1558,15 @@ def query(
...
             ]
+            having = []
             for search_filter in search_filters or ():
                 # use the stored function if it exists in our mapping, otherwise use the basic lookup
                 fn = self.group_conditions_lookup.get(search_filter.key.name)
                 if fn:
                     where_conditions.append(fn(self, search_filter, joined_entity))
+                elif search_filter.key.name == "last_seen":
+                    # last seen modifies the having clause because its an aggregation
+                    having.append(self.get_last_seen_filter(search_filter, joined_entity))
```
wedamija, line 1569: I think we could probably store metadata against the filters so that we know which clause they go into.

like
```
        "first_seen": SnubaFilter(get_first_seen_filter, FilterType.ROW),
        "last_seen": SnubaFilter(get_last_seen_filter, FilterType.AGGREGATE),
```
idk about the names but something like that

wedamija (approved):

src/sentry/search/snuba/executors.py
scefali: Yea good call, let me think about how to do this

wedamija (approved):

Issue #68903, opened 2024-04-15
Replace existing json usages with `orjson`


anonrig:
We currently use `rapidjson` and default `json` library to parse/stringify JSON. The alternative library `orjson` is 6 times faster than `json` and 2-3 times faster than `rapidjson`.

We (me and @fpacifici) validated that simply replacing `rapidjson` with `orjson` results in 25% improvement (from 12ms to 9ms) on a particular task.

For example, here's graph to visualize the difference between `orjson` and `rapidjson`:

![Screenshot 2024-04-15 at 9 14 58 AM](https://github.com/getsentry/sentry/assets/1935246/b1e9d7d3-b734-4c99-a861-a81c418b6a1e)

## The drill

The drill is always this:
1. Pick up a consumer or a component or a class
2. Change the json implementation - surround it with a feature flag
3. Test - enabled feature flags
4. Rollout - usage without feature flags
5. Cleanup - removal of feature flags
6. Done

## Todos

- [ ] Removing all json and rapidjson usages
- [ ] Refactor [`sentry/utils/json.py`](https://github.com/getsentry/sentry/blob/5b88c7e9387903330a15ac5392c61f0e9c512ce9/src/sentry/utils/json.py#L4)

## Task list

| Component | Status | Author | Pull Request | Notes |
| --------- | ------ | ------ | ------------ | ----- |
| [Relay](https://github.com/getsentry/relay/blob/1d06dfbedee08528f0615f6264c01028f3c7c978/py/sentry_relay/processing.py#L206) | 4 | @anonrig | https://github.com/getsentry/sentry/pull/69613 | - |
| Indexer consumer | 6 | @anonrig | https://github.com/getsentry/sentry/pull/68802 | 20-25% overall reduction |
| Ingest processor | 6 | @anonrig | https://github.com/getsentry/sentry/pull/68811 | - |
| Snuba | 6 | @anonrig | https://github.com/getsentry/sentry/pull/68938 | - |
| Slack integration| 4 | @loewenheim | https://github.com/getsentry/sentry/pull/69409 | - |
| reconstruct messages | 4 | @ayirr7 | https://github.com/getsentry/sentry/pull/69378 | 50% reduction in JSON serialization and 25% reduction in average in all indexer payloads |
| outcomes_consumer | 2 | @vbro | https://github.com/getsentry/getsentry/pull/13779 | - |
| event_manager | 4 | @loewenheim | https://github.com/getsentry/sentry/pull/69653 | - |
[...]

PR #70456, opened 2024-05-07
perf: use orjson in all middlewares

anonrig:

### DIFF
--- a/src/sentry/middleware/health.py
+++ b/src/sentry/middleware/health.py
@@ -1,5 +1,6 @@
 import itertools
 
+import orjson
 from django.http import HttpResponse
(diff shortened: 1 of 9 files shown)


asottile-sentry (approved):

fpacifici (approved):

fpacifici:
None of the use case is in the ingestion pipeline or on a critical workflow.

asottile-sentry:
^ that's a duplicate of https://sentry.sentry.io/issues/5090124407/?project=1&query=JSONDecodeError&referrer=issue-stream&statsPeriod=14d&stream_index=2 -- just the type changed

PR #70457, opened 2024-05-07
perf: use orjson in most integrations

anonrig:


asottile-sentry (approved):

asottile-sentry (approved):

loewenheim (approved):
There are a bunch of places where I would have expected a `decode` after `orjson.dumps`, but they're all within test code, and if the tests pass, who am I to argue :)

PR #70467, opened 2024-05-07
feat(performance): Adds a useCheckMessagingMetricExists hook to check if receive latency metric is available before querying

edwardgou-sentry:
The Queues module queries on the `messaging.message.receive.latency` metric, which is an optional metric and may not exist. Currently, these queries will just fail if this metric does not exist. We still want our query to run because we have other metrics that we need.
This PR works around this issue by adding a `useCheckMessagingMetricExists` hook that queries to see if `messaging.message.receive.latency` exists as a metric. Queues queries are updated to use this hook to determine whether or not to include `messaging.message.receive.latency` in their request.


edwardgou-sentry:
no longer needed because this is handled in discover

PR #70472, opened 2024-05-07
ref(rules): Clear Redis buffer after processing

ceorourke:
Clear the related data from the Redis buffer after processing the rules for a project.

### DIFF
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -1,5 +1,4 @@
 import logging
-import math
 import uuid
 from collections import defaultdict
 from datetime import datetime, timedelta, timezone
@@ -240,10 +239,11 @@ def get_group_to_groupevent(
 def process_delayed_alert_conditions(buffer: RedisBuffer) -> None:
     with metrics.timer("delayed_processing.process_all_conditions.duration"):
         fetch_time = datetime.now(tz=timezone.utc)
-        end_time = int(math.ceil(fetch_time.timestamp()))
-        project_ids = buffer.get_sorted_set(PROJECT_ID_BUFFER_LIST_KEY, 0, end_time)
-        for project_id, date_added in project_ids:
-            apply_delayed.delay(project_id, date_added)
+        project_ids = buffer.get_sorted_set(
+            PROJECT_ID_BUFFER_LIST_KEY, min=0, max=fetch_time.timestamp()
+        )
+        for project_id, _ in project_ids:
+            apply_delayed.delay(project_id)
 
         buffer.delete_key(PROJECT_ID_BUFFER_LIST_KEY, min=0, max=fetch_time.timestamp())
 
@@ -257,7 +257,7 @@ def process_delayed_alert_conditions(buffer: RedisBuffer) -> None:
     time_limit=60,  # 1 minute
     silo_mode=SiloMode.REGION,
 )
-def apply_delayed(project_id: int, date_added: float, *args: Any, **kwargs: Any) -> None:
+def apply_delayed(project_id: int, *args: Any, **kwargs: Any) -> None:
     """
     Grab rules, groups, and events from the Redis buffer, evaluate the "slow" conditions in a bulk snuba query, and fire them if they pass
     """
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -323,10 +323,11 @@ def apply_delayed(project_id: int, date_added: float, *args: Any, **kwargs: Any)
                 safe_execute(callback, groupevent, futures, _with_transaction=False)
 
     # Step 8: Clean up Redis buffer data
-    for rule, groups in rules_to_groups.items():
-        for group in groups:
-            buffer.delete_hash(
-                model=Project,
-                filters={"project_id": project_id},
-                field=f"{rule}:{group}",
-            )
+    hashes_to_delete = [
+        f"{rule}:{group}" for rule, groups in rules_to_groups.items() for group in groups
+    ]
+    buffer.delete_hash(
+        model=Project,
+        filters={"project_id": project_id},
+        fields=hashes_to_delete,
+    )
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -1,4 +1,5 @@
 import logging
+import math
 import uuid
 from collections import defaultdict
 from datetime import datetime, timedelta, timezone
@@ -239,7 +240,8 @@ def get_group_to_groupevent(
 def process_delayed_alert_conditions(buffer: RedisBuffer) -> None:
     with metrics.timer("delayed_processing.process_all_conditions.duration"):
         fetch_time = datetime.now(tz=timezone.utc)
-        project_ids = buffer.get_sorted_set(PROJECT_ID_BUFFER_LIST_KEY)
+        end_time = int(math.ceil(fetch_time.timestamp()))
+        project_ids = buffer.get_sorted_set(PROJECT_ID_BUFFER_LIST_KEY, 0, end_time)
         for project_id, date_added in project_ids:
             apply_delayed.delay(project_id, date_added)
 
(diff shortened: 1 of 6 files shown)


src/sentry/rules/processing/delayed_processing.py
```diff
@@ -241,9 +239,9 @@ def get_group_to_groupevent(
 def process_delayed_alert_conditions(buffer: RedisBuffer) -> None:
     with metrics.timer("delayed_processing.process_all_conditions.duration"):
         project_ids = buffer.get_set(PROJECT_ID_BUFFER_LIST_KEY)
-        for project_id in project_ids:
+        for project_id, date_added in project_ids:
```
ceorourke, line 242: Redis responds with a set of tuples of the rule id and the timestamp of when it was added, this was incorrect before.

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -255,10 +253,11 @@ def process_delayed_alert_conditions(buffer: RedisBuffer) -> None:
     time_limit=60,  # 1 minute
     silo_mode=SiloMode.REGION,
 )
-def apply_delayed(project_id: int) -> None:
+def apply_delayed(project_id: int, date_added: float) -> None:
```
wedamija, line 256: Let's add `*args, **kwargs` to the definition as well. This allows us to add/remove fields here without breaking things during deploy

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -320,3 +319,17 @@ def apply_delayed(project_id: int) -> None:
...
                 safe_execute(callback, groupevent, futures, _with_transaction=False)
+
+    # Step 8: Clean up Redis buffer data
+    for rule, groups in rules_to_groups.items():
+        for group in groups:
+            buffer.delete_hash(
+                model=Project,
+                filters={"project_id": project_id},
+                field=f"{rule}:{group}",
+            )
```
wedamija, line 330: As a potential follow up, we might want to implement a batch method for this that builds up a pipeline and sends these in one request.

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -320,3 +319,17 @@ def apply_delayed(project_id: int) -> None:
...
+
+    # if the date the project was added to the buffer is earlier than the date we fetched it, remove it
+    date_added_converted = datetime.fromtimestamp(date_added).replace(tzinfo=timezone.utc)
+    if date_added_converted < time_fetched:
+        buffer.delete_key(PROJECT_ID_BUFFER_LIST_KEY, min=date_added, max=date_added)
```
wedamija, line 335: I think maybe this should be done at the end of `process_delayed_alert_conditions` instead. We don't use `PROJECT_ID_BUFFER_LIST_KEY` in the individual project tasks at all, so I think it makes sense to do
```
fetch_time = now()
project_ids = buffer.get_set(PROJECT_ID_BUFFER_LIST_KEY)
# Schedule all tasks
buffer.delete_key(PROJECT_ID_BUFFER_LIST_KEY, min=0, max=fetch_time)
```

(We use `0` here so that we just delete everything before that time, since we know we've scheduled everything already)

src/sentry/rules/processing/delayed_processing.py
ceorourke: Hmm maybe I can just modify `delete_hash` to be like `add_hash_to_delete_pipeline` (idk maybe keep the name the same) but basically just not do `pipe.execute()` and have another function that only executes?

ceorourke: Doesn't seem like that much work, I can just add that change into this PR

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -240,10 +238,13 @@ def get_group_to_groupevent(
 @redis_buffer_registry.add_handler(BufferHookEvent.FLUSH)
 def process_delayed_alert_conditions(buffer: RedisBuffer) -> None:
     with metrics.timer("delayed_processing.process_all_conditions.duration"):
+        fetch_time = datetime.now(tz=timezone.utc)
         project_ids = buffer.get_set(PROJECT_ID_BUFFER_LIST_KEY)
```
wedamija, line 242: I just realised I've been confused by this name the whole time. I thought it means `get then set`, but it means`get the values from a set`.

It might make sense to rename it as part of the changes here, `get_range_from_set` or something?

ceorourke: Hahaha oh no 😅 yeah I think I had like `get_list` before and then it changed to a set ... what about `get_sorted_set` since we also have `get_hash` and `push_to_sorted_set`?

src/sentry/rules/processing/delayed_processing.py
wedamija: Hmm, I don't know if we want to store the `pipeline` locally in the class, there might be concurrency problems there if we have multiple threads.

It might be easier to just have `buffer.delete_hash_bulk(model=Project, filters={"project_id": project_id}, fields=[<list of all rule and group strings>]`

src/sentry/rules/processing/delayed_processing.py
wedamija: Yeah that seems fine, let's go with that

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -240,10 +238,13 @@ def get_group_to_groupevent(
 @redis_buffer_registry.add_handler(BufferHookEvent.FLUSH)
 def process_delayed_alert_conditions(buffer: RedisBuffer) -> None:
     with metrics.timer("delayed_processing.process_all_conditions.duration"):
-        project_ids = buffer.get_set(PROJECT_ID_BUFFER_LIST_KEY)
-        for project_id in project_ids:
+        fetch_time = datetime.now(tz=timezone.utc)
+        project_ids = buffer.get_sorted_set(PROJECT_ID_BUFFER_LIST_KEY)
+        for project_id, date_added in project_ids:
```
anonrig, line 243: This will call metrics for every project. I think this should be reversed

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -240,10 +238,12 @@ def get_group_to_groupevent(
 @redis_buffer_registry.add_handler(BufferHookEvent.FLUSH)
 def process_delayed_alert_conditions(buffer: RedisBuffer) -> None:
     with metrics.timer("delayed_processing.process_all_conditions.duration"):
-        project_ids = buffer.get_set(PROJECT_ID_BUFFER_LIST_KEY)
-        for project_id in project_ids:
-            with metrics.timer("delayed_processing.process_project.duration"):
-                apply_delayed.delay(project_id=project_id)
+        fetch_time = datetime.now(tz=timezone.utc)
+        project_ids = buffer.get_sorted_set(PROJECT_ID_BUFFER_LIST_KEY)
```
wedamija, line 242: `get_sorted_set` needs to incorporate the same range as `delete_key`. So we know that we're only fetching data between `0` and `fetch_time`.

Maybe a bit pedantic, but it helps us avoid a situation where we get the timestamp, then an extra project id is added to the set before we fetch, which we would end up not deleting.

ceorourke: Updated but the behavior is kinda strange, `zrange` only accepts integers for `start` and `end` versus how `zremrangebyscore` (the Redis docs say `zrange` replaces `zrangebyscore`) accepts floats (aka what timestamps are) for `min` and `max`.

wedamija: Ahh, we need to use https://redis.io/docs/latest/commands/zrangebyscore/

wedamija: Or actually, https://redis.io/docs/latest/commands/zrange/ and use the `byscore` option if that's available for us

wedamija: Nm, I just saw you already did this, ignore me

wedamija (approved):

PR #70477, opened 2024-05-07
fix(chartcuterie): Added Visual Map Field for Endpoint Regression

iamrajjoshi:
There is mismatch in the way we build the EChart Options object in our FE code and how Chartcuterie handles it.

In our FE [code,](https://github.com/getsentry/sentry/blob/master/static/app/components/events/eventStatisticalDetector/breakpointChartOptions.tsx#L89-L104), we wrap the visualMap object in an extra option, which allows us to maintain the hierarchy for styling. However, Chartcuterie cannot handle the wrapped object, so when we pass the service the options, we unwrap it.

I also created a modifier option to modify chart options specifically for slack and removed the icon from the legend icon from there.

example2

### DIFF
--- a/static/app/chartcuterie/performance.tsx
+++ b/static/app/chartcuterie/performance.tsx
@@ -21,6 +21,7 @@ performanceCharts.push({
       backgroundColor: theme.background,
       series: transformedSeries,
       grid: slackChartDefaults.grid,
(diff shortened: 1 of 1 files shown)


sentaur-athena (approved):
Makes sense. Thanks for looking into this.

PR #70482, opened 2024-05-07
chore(icons): Add more options for iconLock

roggenkemper:
this pr updates the options for `iconLock` so that it is never solid and the only choice is whether it is locked. old prop is included since it is being used in `getsentry`, will remove in future pr

Screenshot 2024-05-07 at 3 26 56 PM
Screenshot 2024-05-07 at 3 26 22 PM


vuluongj20:
For simplicity, let's maintain only the non-solid options. We can have a `locked` prop to toggle between these two states:
image

vuluongj20 (approved):

Issue #70411, opened 2024-05-07
Commits do not match the count


anapier70:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

On the releases page the count on the commits tab does not match the number of commits actually shown on the commits tab. The same applied to the files tab, count in the tab header does not match how many files get shown.

This is counting using all available pages so we have went to the next page, but for example on our latest release we have a count of 61 files and only 27 listed out individually

### Expected Result

number of commits should match the number of commits shown

### Actual Result

![image](https://github.com/getsentry/sentry/assets/164612005/31487c99-9adb-4dc2-a90f-81fb4335f39d)

### Product Area

Releases

### Link

https://innovid-xp.sentry.io/releases/jobqueue%40measrelease.20240506/commits/?environment=prod&environment=produsa&project=4506575222800384

### DSN

https://innovid-xp.sentry.io/projects/jobqueue/?project=4506575222800384

schew2381:
Hi thanks for the bug report, I'm looking into this and it appears the problem lies with how we display the commits rather than the commit count.

I'll post future updates on this ticket as I keep exploring

schew2381:
@anapier70 Hi there, the issue should be fixed now. Please feel free to ping me on this issue again if it pops up again!

PR #63860, opened 2024-01-25
feat(releases): Convert release commits to FC

scttcper:

PR #70485, opened 2024-05-07
fix(ui): Revert to using project release commmit API

schew2381:
In https://github.com/getsentry/sentry/pull/63860 we switched from a class component to a FC and also switched from hitting `ProjectReleaseCommitsEndpoint` to `OrganizationReleaseCommitsEndpoint`.

The latter doesn't respect repo name or id in the query param, which makes it so selecting a repo in the dropdown will show you commits from other repos (and also messes up pagination as well, not showing you all commits).

Related to https://github.com/getsentry/sentry/issues/70411


cathteng (approved):
nice catch

PR #70486, opened 2024-05-07
update query_extra logic

nhsiehgit:


src/sentry/incidents/models/alert_rule.py
```diff
@@ -162,14 +163,21 @@ def conditionally_subscribe_project_to_alert_rules(
                             "origin": origin,
                             "query_extra": query_extra,
                             "condition": activation_condition,
+                            "restrict_to_window": restrict_to_window,
                         },
                     )
-                    # attempt to subscribe the alert rule
+                    now = timezone.now()
+                    window_delta = timedelta(seconds=alert_rule.snuba_query.time_window)
```
nhsiehgit, line 170: Snuba requires both a start `timestamp > ...` and an end `timestamp < ...` in the query

In order to calculate the end timestamp, we need the `time_window_ from the `snuba_query`
So i need to construct the query_extra here

src/sentry/incidents/models/alert_rule.py
```diff
@@ -164,14 +164,14 @@ def conditionally_subscribe_project_to_alert_rules(
...
                     )
-                    # attempt to subscribe the alert rule
                     created_subscriptions.extend(
                         alert_rule.subscribe_projects(
                             projects=[project],
                             monitor_type=AlertRuleMonitorType.ACTIVATED,
                             query_extra=query_extra,
                             activation_condition=activation_condition,
                             activator=activator,
+                            timebox=True,
```
nhsiehgit, line 174: activated alerts are timeboxed from `QuerySubscription` creation till end of the `SnubaQuery.time_window`

src/sentry/snuba/models.py
```diff
@@ -117,6 +117,9 @@ class Status(Enum):
     query_extra = models.TextField(
         null=True
     )  # additional query filters to attach to the query created in Snuba such as datetime filters, or release/deploy tags
+    # timebox_start is optional timebox restriction indicator for the query
+    # requires an 'end' on snql query construction derived from snuba_query time_window
+    timebox_start = models.DateTimeField(null=True)
```
saponifi3d, line 122: don't forget to remove once the other pr is merged :)

src/sentry/incidents/models/alert_rule.py
saponifi3d: Rather than creating a boolean and passing it around, could we set a property on the model which checks to see if there's a time_window + time_start associated with it?

src/sentry/snuba/subscriptions.py
```diff
@@ -156,14 +167,20 @@ def create_snuba_subscription(
     :param snuba_query: A `SnubaQuery` instance to subscribe the project to.
     :return: The QuerySubscription representing the subscription
     """
+    timebox_start = None
+    if timebox:
+        timebox_start = timezone.now()
```
saponifi3d, line 172: Rather than passing in `timebox` could we just have the user pass in `timebox_start`? That would make this a bit more flexible allowing us to set the start to whatever we want to, not just `timezone.now()`

src/sentry/snuba/tasks.py
```diff
@@ -218,16 +218,26 @@ def _create_in_snuba(subscription: QuerySubscription) -> str:
...
+            "project_id": [subscription.project_id],
+        }
+        skip_time_conditions = True
+        if subscription.timebox_start:
+            start = subscription.timebox_start.isoformat()
+            window = snuba_query.time_window
+            end = (subscription.timebox_start + window).isoformat()
+            skip_time_conditions = False
+            params[start] = start
+            params[end] = end
```
saponifi3d: Should we just be setting the data as start / end times instead of start / duration? that way we only calculate on write vs read.

nhsiehgit: So - this is 'read' of the QuerySubscription - but it's 'write' of the snql query

The 'end' will always be a derivative of the SnubaQuery.time_window - so there's not really any reason to duplicate that value.
Part of my reason to derive the end here rather than set it on the QuerySubscription is _also_ to ensure that we never introduce a query with too large of a window.
This is the reason why the time conditions was being skipped originally as well

src/sentry/snuba/subscriptions.py
nhsiehgit: Yea - either way could work.
figured it would make the most sense to grab the timestamp closest to creating our model instance though since we _don't_ actually want to make this more flexible.

I'm essentially restricting us from passing in any timestamps we want - and _only_ grabbing the current now timestamp

wedamija (commented):
I think it would be more generic to have both `timebox_start` and `timebox_end` cols. Then if we have some other logic we want to use to determine the length of the timebox we're not tied to the logic from your project

src/sentry/snuba/tasks.py
```diff
@@ -218,16 +218,26 @@ def _create_in_snuba(subscription: QuerySubscription) -> str:
...
             subscription.project.organization_id,
         )
-        # TODO: determine whether concatenating query_extra is proper
+        params = {
+            "organization_id": subscription.project.organization_id,
+            "project_id": [subscription.project_id],
+        }
+        skip_time_conditions = True
+        if subscription.timebox_start:
+            start = subscription.timebox_start.isoformat()
```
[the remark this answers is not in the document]
nhsiehgit: Yea hmmm

So Activated Alert QuerySubscriptions will self delete themselves after their window has elapsed - so we _should_ never need to update the windows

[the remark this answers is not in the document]
wedamija: Wondering if now that this timebox is explicitly part of subscriptions, if we should auto clean up subscriptions with a timebox too, rather than waiting for another system to do it?

src/sentry/snuba/tasks.py
nhsiehgit: hmmm....
I guess i'll actually just ave it to the model as well :/

src/sentry/snuba/subscriptions.py
```diff
@@ -156,12 +168,19 @@ def create_snuba_subscription(
     :param snuba_query: A `SnubaQuery` instance to subscribe the project to.
     :return: The QuerySubscription representing the subscription
     """
+    timebox_start = None
+    timebox_end = None
+    if timebox:
+        timebox_start = timezone.now()
+        timebox_end = timebox_start + timedelta(seconds=snuba_query.time_window)
```
wedamija, line 175: This logic shouldn't live here, we should pass these fields into the function and have your alert type handle the logic

wedamija (commented):
lgtm, just needs some tests

src/sentry/snuba/tasks.py
```diff
@@ -218,16 +221,23 @@ def _create_in_snuba(subscription: QuerySubscription) -> str:
...
             environment=snuba_query.environment,
-            params={
-                "organization_id": subscription.project.organization_id,
-                "project_id": [subscription.project_id],
-            },
+            params=params,
+            skip_time_conditions=skip_time_conditions,
         ).get_snql_query()
 
         return _create_snql_in_snuba(subscription, snuba_query, snql_query, entity_subscription)
```
nhsiehgit, line 243: The request to Snuba is failing in the tests due to 
`Cannot have existing conditions on time field timestamp`

I've found the error message is coming from
[here](https://github.com/getsentry/snuba/blob/eca97d1c332adabb91d0ba99db33cbbf8f53a3e9/snuba/query/validation/validators.py#L197-L228)
but i'm not entirely sure how it gets here or what we would need to add to resolve this 🤔

nhsiehgit: @wmak Would you have an idea?

cc @ayirr7 @enochtangg ?

PR #70490, opened 2024-05-08
perf: replace memcache with redis on assemble

anonrig:
Once https://github.com/getsentry/sentry-options-automator/pull/1358 lands, we can land this pull-request since we're right now officially writing to both Memcache and Redis for assemble status.


markstory (approved):
Looks good to me.

anonrig:
I've removed the flag from the previous test, and force pushed. This PR is ready to land.

PR #70493, opened 2024-05-08
ref(replay): Rage click clicked element name

c298lee:
Since we now use react component names in the selector path, we should modify clicked element to provide more specific info such as class, role etc.
This prevents giving the react component name twice and would give more info for debugging since react component names aren't very specific

Before: 
image


tests/sentry/replays/unit/test_issue_creation.py
```diff
@@ -23,36 +23,48 @@ def test_report_rage_click_issue_with_replay_event(mock_new_issue_occurrence, de
...
+        "selector": "div.xyz > SmartSearchBar",
         "component_name": "SmartSearchBar",
     }
 
     assert (
         issue_occurence_call["evidence_display"][0].to_dict()
-        == IssueEvidence(name="Clicked Element", value="a", important=False).to_dict()
+        == IssueEvidence(
+            name="Clicked Element",
+            value="a[class=class1 class2][role=button][aria-label=test]",
```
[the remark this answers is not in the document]
c298lee: great catch!!

michellewzhang (approved):
looks good!

src/sentry/replays/usecases/ingest/issue_creation.py
```diff
@@ -110,3 +110,12 @@ def _make_tags(replay_id, url, replay_event):
         tags.update(replay_event["tags"])
 
     return tags
+
+
+def _make_clicked_element(node):
+    element = node.get("tagName", "")
+    if "attributes" in node:
+        for key, value in node["attributes"].items():
+            element += "[" + key + '="' + value + '"]'
```
cmanallen, line 119: We should represent ids with a `#` and classes with a `.`.  E.g. `a#id.class.class2[attribute="a"]`

cmanallen (changes requested):

src/sentry/replays/usecases/ingest/issue_creation.py
```diff
@@ -110,3 +110,12 @@ def _make_tags(replay_id, url, replay_event):
         tags.update(replay_event["tags"])
 
     return tags
+
+
+def _make_clicked_element(node):
+    element = node.get("tagName", "")
+    if "attributes" in node:
+        for key, value in node["attributes"].items():
```
aliu3ntry, line 118: Nit:
```suggestion
        str_builder = [element]
        for key, value in node["attributes"].items():
            str_builder.append(f'[{key}="{value}"]')
        return ''.join(str_builder)
```
Pretty nit but this is the preferred/most efficient way to build a string in a loop. Avoids O(n^2) str concatenation 🤓

src/sentry/replays/usecases/ingest/issue_creation.py
```diff
@@ -110,3 +110,27 @@ def _make_tags(replay_id, url, replay_event):
...
+            if key == "alt":
+                element += f'[alt="{value}"]'
+            if key == "data-test-id" or key == "data-testid":
+                element += f'[data-test-id="{value}"]'
+            if key == "aria-label":
+                element += f'[aria="{value}"]'
+            if key == "title":
+                element += f'[title="{value}"]'
+            if key == "data-sentry-component":
+                element += f'[data-sentry-component="{value}"]'
```
cmanallen, line 134: Replace `if`s with `elif`s and we're set.

cmanallen (approved):

PR #70500, opened 2024-05-08
fix(replay): check for undefined property in init crumb

michellewzhang:
fixes JAVASCRIPT-2SZ9
fixes JAVASCRIPT-2T0N

check for undefined properties before we attempt to access them. the replay init breadcrumb doesn't have `data.to`:

https://github.com/getsentry/sentry/blob/257a223719efcfc4d77e4440e3f3e30e88fbbf88/static/app/utils/replays/hydrateBreadcrumbs.tsx#L59-L70


ryan953 (approved):
wanna update the tests with an init breadcrumb?

PR #70501, opened 2024-05-08
fix(ui): Fix squished layout for queues

KevinL10:
Fixes squished layout issues on queue onboarding page.

### **Before**
image

### **After**
image


narsaynorath (approved):
💯

PR #70503, opened 2024-05-08
chore(actor) Remove Actor from django state

markstory:
Remove the Actor model from django state. This required reworking a few migration operations in the squash to avoid creating constraints on a table that doesn't exist anymore. It seems that when a model is removed from state django doesn't create the table and any constraints that were defined earlier in migration history need to be updated.


evanpurkhiser (approved):

src/sentry/migrations/0001_squashed_0484_break_org_member_user_fk.py
```diff
@@ -6576,11 +6576,6 @@ class Migration(CheckedMigration):
                     reverse_sql="DROP INDEX CONCURRENTLY IF EXISTS sentry_team_actor_idx;",
                     hints={"tables": ["sentry_team"]},
                 ),
-                migrations.RunSQL(
-                    sql='ALTER TABLE sentry_team ADD CONSTRAINT "sentry_team_actor_idx_fk_sentry_actor_id" FOREIGN KEY ("actor_id") REFERENCES "sentry_actor" ("id") DEFERRABLE INITIALLY DEFERRED;',
-                    reverse_sql="ALTER TABLE sentry_team DROP CONSTRAINT IF EXISTS sentry_team_actor_idx_fk_sentry_actor_id;",
-                    hints={"tables": ["sentry_team"]},
-                ),
```
wedamija, line 6583: I guess this is the main problem, since django now doesn't properly know about these constraints,

wedamija (approved):
It's nbd to update the state, but we shouldn't need to update earlier migrations to delete a model. I suspect this is either some problem with migration tests, or with migration order dependencies

src/sentry/migrations/0001_squashed_0484_break_org_member_user_fk.py
markstory: With these constraints in place `pytest --migrations` would fail because constraints couldn't be created as `sentry_actor` was not a table. When this first migration is run `sentry_actor`should exist as it is created on line 6553 of this migration. However, with the new migration that remove Actor from state the table is 'missing' when these constraints are created :cry:.

PR #70504, opened 2024-05-08
feat(related_issues): Decouple endpoint

armenzg:
The API tries to find issues for both approaches sequentially (slower), making it harder to iterate on each approach independently.

This change allows using a `type` parameter with the endpoint to execute one approach or the other.

The UI will be changed to make two independent requests, thus, improving load time.

This is required to empower trace issues from the issues details page.

### DIFF
--- a/src/sentry/api/endpoints/issues/related_issues.py
+++ b/src/sentry/api/endpoints/issues/related_issues.py
@@ -6,7 +6,7 @@ from sentry.api.api_publish_status import ApiPublishStatus
 from sentry.api.base import region_silo_endpoint
 from sentry.api.bases.group import GroupEndpoint
 from sentry.issues.related import find_related_issues  # To be deprecated
-from sentry.issues.related import same_root_cause_analysis, trace_connected_analysis
+from sentry.issues.related import RELATED_ISSUES_ALGORITHMS
 from sentry.models.group import Group
 from sentry.types.ratelimit import RateLimit, RateLimitCategory
 
@@ -40,8 +40,8 @@ class RelatedIssuesEndpoint(GroupEndpoint):
 
         if related_type in RELATED_ISSUES_ALGORITHMS:
             data, meta = RELATED_ISSUES_ALGORITHMS[related_type](group)
-            related_issues = [{"type": related_type, "data": data, "meta": meta}]
+            return Response({"type": related_type, "data": data, "meta": meta})
         else:
             # XXX: We will be deprecating this approach soon
             related_issues = find_related_issues(group)
-        return Response({"data": [related_set for related_set in related_issues]})
+            return Response({"data": [related_set for related_set in related_issues]})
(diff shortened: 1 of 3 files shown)


src/sentry/api/endpoints/issues/related_issues.py
```diff
@@ -24,13 +25,26 @@ class RelatedIssuesEndpoint(GroupEndpoint):
...
+        # The type of related issues to retrieve. Can be either `same_root_cause` or `trace_connected`.
+        related_type = request.query_params.get("type")
+        related_issues: list[dict[str, str | list[int] | dict[str, str]]] = []
+
+        if related_type == "same_root_cause":
+            data, meta = same_root_cause_analysis(group)
+            related_issues = [{"type": related_type, "data": data, "meta": meta}]
+        elif related_type == "trace_connected":
+            data, meta = trace_connected_analysis(group)
+            related_issues = [{"type": related_type, "data": data, "meta": meta}]
```
mrduncan, line 46:
```suggestion
        if related_type in RELATED_ISSUES_ALGORITHMS:
            data, meta = RELATED_ISSUES_ALGORITHMS[related_type](group)
            related_issues = [{"type": related_type, "data": data, "meta": meta}]
```

nit: could potentially simplify this by using `sentry.issues.related.RELATED_ISSUES_ALGORITHMS`

tests/sentry/api/endpoints/issues/test_related_issues.py
```diff
@@ -40,15 +40,12 @@ def test_same_root_related_issues(self) -> None:
...
         # https://us.sentry.io/api/0/organizations/sentry/issues-stats/?groups=4741828952&groups=4489703641&statsPeriod=24h
         assert response.json() == {
-            "data": [
-                {"type": "same_root_cause", "data": [5], "meta": {}},
-                {"type": "trace_connected", "data": [], "meta": {}},
-            ],
+            "data": [{"type": "same_root_cause", "data": [5], "meta": {}}],
         }
 
     def test_trace_connected_errors(self) -> None:
```
mrduncan, line 51: Is it worth also having a temporary test for the deprecated case which returns both types?

src/sentry/api/endpoints/issues/related_issues.py
```diff
@@ -24,13 +25,26 @@ class RelatedIssuesEndpoint(GroupEndpoint):
...
+        else:
+            # XXX: We will be deprecating this approach soon
+            related_issues = find_related_issues(group)
         return Response({"data": [related_set for related_set in related_issues]})
```
mrduncan, line 50: This return type ends up feeling a little weird once the deprecated case is removed when the result will only ever contain type. Thoughts on removing the outer array for these new responses?

ex: `{"type": "same_root_cause", "data": [5], "meta": {}}` instead of the current `[{"type": "same_root_cause", "data": [5], "meta": {}}]`

mrduncan (approved):

src/sentry/api/endpoints/issues/related_issues.py
armenzg: You got a good eye Mr. Duncan!
When dealing with the follow up UI changes I could already see how non-sensical it is to keep the old shape.

Issue #64924, opened 2024-02-08
Track and display "viewed by" avatars for replays


IanWoodard:
### Problem Statement

When looking through replays, I often find myself clicking on the same one multiple times thinking I haven't seen it when I already have. Currently there is no indication if you (or any of your teammates) have already seen a given replay.

### Solution Brainstorm

One solution would be to take a similar approach to issues and bolden the titles of the replays you haven't seen before, and possibly show a list of viewers for each replay.

### Product Area

Replays

ryan953:
cc @cmanallen for onboarding

cmanallen:
@aliu3ntry This would be a cool feature to add and I think well within your capabilities.

1. We need a database model capable of tracking the user_id and replay_id.
    - Will require a migration.
3. We need an API endpoint which can retrieve all of the users who have viewed a replay.
4. We need an API endpoint which can create a new replay_id <> user_id row.
5. The Javascript web app needs to be able to fetch and write to these endpoints.
    - Create a new resource blueprint in replays/blueprints/api.md
6. Schedule a meeting with relevent members of the replay engineering team or bring it up at our weekly engineering sync on Wednesday.

Start with the blueprint.  Think about the use cases, how you want to store the data, how you want to present it to clients.

cmanallen:
@aliu3ntry

- [ ] Create a collection endpoint in `sentry/replays/endpoints`.
    - It should return response data matching the blueprint.
    - https://github.com/getsentry/sentry/pull/67140/files#diff-0e4d43542249b0417d788646d1263990affac44f3c80c8740d23849db1d8e42cR612
- [ ] Create a details endpoint in `sentry/replays/endpoints`.
    - Add one method for the `GET` operation and one method for the `POST` operation.
    - It should return response data matching the blueprint.
    - https://github.com/getsentry/sentry/pull/67140/files#diff-0e4d43542249b0417d788646d1263990affac44f3c80c8740d23849db1d8e42cR612
- [ ] Register the endpoints in `sentry/api/urls.py`
- [ ] Define a replay-viewed model in `sentry/replays/models
    - Inherit from the `Model` base class.
    - Prior art: https://github.com/getsentry/sentry/blob/b4a1e5f349864dcc3d6026ba08b24f9ea0557e0c/src/sentry/models/groupseen.py#L10-L29
    - Replay viewed models should be unique per replay <-> user pairing.
[...]

PR #70505, opened 2024-05-08
feat(replay): viewed_by_me search alias

aliu39:
Relates to https://github.com/getsentry/sentry/issues/64924

Frontend change coming soon
I'm ok with removing the `viewed_by_id` parameter from the typeahead. Thoughts?


src/sentry/replays/usecases/query/__init__.py
```diff
@@ -163,6 +165,17 @@ def query_using_optimized_search(
             SearchFilter(SearchKey("environment"), "IN", SearchValue(environments)),
         ]
 
+    for i in range(len(search_filters)):
```
cmanallen, line 168: `for search_filter in search_filters:`

src/sentry/replays/usecases/query/__init__.py
```diff
@@ -163,6 +165,17 @@ def query_using_optimized_search(
...
+        sf = search_filters[i]
+        if sf.key.name in VIEWED_BY_ME_KEY_ALIASES:
+            # "viewed_by_me" is not a valid query field.
+            # It's a convenience alias for users without the admin access to lookup ids.
+            if sf.operator not in ["=", "!="]:
+                raise ParseError("Invalid operator specified for viewed_by_me")
+            search_filters[i] = SearchFilter(
+                SearchKey("viewed_by_id"), sf.operator, SearchValue(request_user_id)
+            )
```
cmanallen, line 177: How is this handled `viewed_by_me:false`?

src/sentry/replays/usecases/query/__init__.py
```diff
@@ -163,6 +165,17 @@ def query_using_optimized_search(
...
+
```
cmanallen, line 178: Needs coverage.  Both fields, true and false values, `=` and `!=` operators.

src/sentry/replays/usecases/query/__init__.py
```diff
@@ -163,6 +165,17 @@ def query_using_optimized_search(
             SearchFilter(SearchKey("environment"), "IN", SearchValue(environments)),
         ]
 
+    for i in range(len(search_filters)):
+        sf = search_filters[i]
+        if sf.key.name in VIEWED_BY_ME_KEY_ALIASES:
+            # "viewed_by_me" is not a valid query field.
+            # It's a convenience alias for users without the admin access to lookup ids.
+            if sf.operator not in ["=", "!="]:
+                raise ParseError("Invalid operator specified for viewed_by_me")
```
cmanallen, line 174: How do you know the user queried by `viewed_by_me`?  Could have queried by `seen_by_me`.

src/sentry/replays/usecases/query/__init__.py
```diff
@@ -163,6 +165,17 @@ def query_using_optimized_search(
...
+            search_filters[i] = SearchFilter(
+                SearchKey("viewed_by_id"), sf.operator, SearchValue(request_user_id)
```
cmanallen, line 176: Even though it sholuldn't be possible we need to handle the case where the user_id is null.  Unless you want to enforce non-nullability of the user in the call site (which would also be a fine outcome).

cmanallen (changes requested):

src/sentry/replays/usecases/query/__init__.py
aliu3ntry: Oh, good point!

src/sentry/replays/usecases/query/__init__.py
```diff
@@ -163,6 +165,18 @@ def query_using_optimized_search(
...
         ]
 
+    for i in range(len(search_filters)):
+        sf = search_filters[i]
+        if sf.key.name in VIEWED_BY_ME_KEY_ALIASES:
+            # "viewed_by_me" is not a valid query field.
+            # It's a convenience alias for users without the admin access to lookup ids.
+            if sf.operator not in ["=", "!="]:
+                raise ParseError("Invalid operator specified for viewed_by_me")
+            is_me = not ((sf.operator == "=") ^ sf.value.value)
```
aliu3ntry, line 175: Got kinda fancy with it. Lmk if this is unreadable lol

src/sentry/replays/usecases/query/__init__.py
```diff
@@ -163,6 +165,20 @@ def query_using_optimized_search(
...
+    for i in range(len(search_filters)):
+        sf = search_filters[i]
+        if sf.key.name in VIEWED_BY_ME_KEY_ALIASES:
+            # "viewed_by_me" is not a valid query field.
+            # It's a convenience alias for users without the admin access to lookup ids.
+            if sf.operator not in ["=", "!="]:
+                raise ParseError(f"Invalid operator specified for {sf.key.name}")
+            is_me = not (
+                (sf.operator == "=") ^ sf.value.value
+            )  # True if both expressions are the same, else False.
```
cmanallen, line 177: Very fancy!  Let's change this though.  Only the `=` operator should be allowed.  I think `!=` doesn't make sense in this context.  Now we only need to care about the value and can drop the xor.

src/sentry/replays/usecases/query/__init__.py
```diff
@@ -163,6 +165,20 @@ def query_using_optimized_search(
...
+            search_filters[i] = SearchFilter(
+                SearchKey("viewed_by_id"), "=" if is_me else "!=", SearchValue(request_user_id)
```
cmanallen, line 179: Let's assign this inline `if` to a variable `operator`.  A little more verbose but it will lower the density of this code.

src/sentry/replays/usecases/query/__init__.py
```diff
@@ -163,6 +165,20 @@ def query_using_optimized_search(
             SearchFilter(SearchKey("environment"), "IN", SearchValue(environments)),
         ]
 
+    for i in range(len(search_filters)):
+        sf = search_filters[i]
+        if sf.key.name in VIEWED_BY_ME_KEY_ALIASES:
+            # "viewed_by_me" is not a valid query field.
+            # It's a convenience alias for users without the admin access to lookup ids.
+            if sf.operator not in ["=", "!="]:
+                raise ParseError(f"Invalid operator specified for {sf.key.name}")
```
cmanallen, line 174: Also need to assert the value is a bool.

cmanallen (changes requested):

src/sentry/replays/usecases/query/__init__.py
aliu3ntry: The problem is the frontend search bar: users don't type in `key=val` or `key!=val`, they type `key:val` and `!key:val`. The latter translates into the != operator. So unless we want to drop searches for `!viewed_by_me:x`, we have to handle !=...

aliu3ntry: I guess if users know its a boolean, it'd be pretty weird to do `!viewed_by_me:true`. So we can just raise ParseError to simplify this..

aliu3ntry:
Sorry all... my rebase git config keeps resetting 🤦‍♂️

MichaelSun48:
Happens to the best of us... 

`git push origin <branch> --force` after a local rebase worked as expected for me before

aliu3ntry:
> Happens to the best of us...
> 
> `git push origin <branch> --force` after a local rebase worked as expected for me before

Thanks..
Nah, I've gone over to the merge side. No rebasing 🙅‍♂️

PR #70506, opened 2024-05-08
ref(crons): Don't error on missed check-ins

evanpurkhiser:
Fixes SENTRY-38Z6

### DIFF
--- a/src/sentry/monitors/clock_tasks/check_timeout.py
+++ b/src/sentry/monitors/clock_tasks/check_timeout.py
@@ -56,11 +56,17 @@ def dispatch_check_timeout(ts: datetime):
 def mark_checkin_timeout(checkin_id: int, ts: datetime):
     logger.info("checkin_timeout", extra={"checkin_id": checkin_id})
 
-    checkin = (
-        MonitorCheckIn.objects.select_related("monitor_environment")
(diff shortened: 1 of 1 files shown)


davidenwang (approved):

src/sentry/monitors/clock_tasks/check_timeout.py
```diff
@@ -56,11 +56,18 @@ def dispatch_check_timeout(ts: datetime):
...
+        checkin = (
+            MonitorCheckIn.objects.select_related("monitor_environment")
+            .select_related("monitor_environment__monitor")
+            .get(id=checkin_id)
+        )
+    except MonitorCheckIn.DoesNotExist:
+        # The check-in may have reached it's retention period
+        #
+        # XXX(epurkhiser): Are there other reasons a check-in may have been
+        # removed between timeout detection and now?
```
wedamija, line 69: I think we should be logging this or at least keeping a metric. The main reason I could see it happening is that they deleted the monitor? Retention period seems pretty unlikely tbh

evanpurkhiser: Ahh yeah. that makes more sense 👍

PR #70507, opened 2024-05-08
ref: fix typing of slack test_requests

asottile-sentry:

### DIFF
--- a/tests/sentry/integrations/slack/test_requests.py
+++ b/tests/sentry/integrations/slack/test_requests.py
@@ -74,36 +74,26 @@ class SlackRequestTest(TestCase):
             },
         }
 
-    def test_validate_existence_of_data(self):
-        type(self.request).DATA = mock.PropertyMock(side_effect=ValueError())
-
-        with pytest.raises(SlackRequestError):
-            self.slack_request.validate()
-
+    @pytest.mark.xfail(strict=True, reason="crashes in _log_request before validation can occur")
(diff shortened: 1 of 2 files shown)


tests/sentry/integrations/slack/test_requests.py
```diff
@@ -74,36 +74,26 @@ def test_disregards_None_logging_values(self):
...
-    def test_validate_existence_of_data(self):
-        type(self.request).DATA = mock.PropertyMock(side_effect=ValueError())
-
-        with pytest.raises(SlackRequestError):
-            self.slack_request.validate()
-
+    @pytest.mark.xfail(strict=True, reason="crashes in _log_request before validation can occur")
     def test_returns_400_on_invalid_data(self):
-        type(self.request).DATA = mock.PropertyMock(side_effect=ValueError())
+        type(self.request).data = mock.PropertyMock(side_effect=ValueError())
```
evanpurkhiser, line 79: Was this even working before lol

evanpurkhiser (approved):

tests/sentry/integrations/slack/test_requests.py
asottile-sentry, line 79: typo on `DATA` so the test wasn't actually checking anything (the SlackRequestError was coming from the integration not being set up!)

even still, once this is fixed it crashes before the validation while logging the test so I disabled this

PR #70512, opened 2024-05-08
fix(metrics): Add missing 'return' statement in <WidgetDetails />

floels:
# Goal

The goal of this PR is to add a missing `return` statement in the rendering logic of the `<WidgetDetails />` component defined in `static/app/views/metrics`:

https://github.com/getsentry/sentry/blob/6b530ffab46d0b50512f0f4c9df8d084c1feb831/static/app/views/metrics/widgetDetails.tsx#L54-L56

This `return` was most likely forgotten when writing this piece of code. This probably doesn't result in any bug: by incorrectly continuing the rendering function, we end up also rendering a `<MetricDetails />` component with `undefined` `mri`, `op` and `query`.

The code makes more sense with the `return` statement though.

# Legal Boilerplate

### DIFF
--- a/static/app/views/metrics/widgetDetails.tsx
+++ b/static/app/views/metrics/widgetDetails.tsx
@@ -52,7 +52,7 @@ export function WidgetDetails() {
 
   // TODO(aknaus): better fallback
   if (selectedWidget?.type === MetricExpressionType.EQUATION) {
-    <MetricDetails onRowHover={handleSampleRowHover} focusArea={focusArea} />;
+    return <MetricDetails onRowHover={handleSampleRowHover} focusArea={focusArea} />;
   }
 
   const {mri, op, query, focusedSeries} = selectedWidget as MetricsQueryWidget;


anonrig (approved):

evanpurkhiser (commented):
Looks important @edwardgou-sentry can you take a look?

edwardgou-sentry (approved):
Code looks fine, but I don't think this actually changes anything because we return `MetricDetails` on L60 anyways, and those props are optional.

cc: @obostjancic might be more familiar with the metrics page

floels:
Thanks @evanpurkhiser and @edwardgou-sentry for the review! Do you think we can get this merged?

ArthurKnaus (approved):
@edwardgou-sentry is right this will not change any behaviour. Though it is cleaner.
thx for fixing @floels 🙌 
We should use the proper type guards here. Will do in a follow-up.

PR #70513, opened 2024-05-08
chore(actor) Make actor type logic simpler

markstory:
Add properties to RpcActor so type logic is easier to build.


wedamija (approved):
Much nicer, thanks!

PR #70514, opened 2024-05-08
feat(traces): More improvements, use sdk names now

k-fish:
### Summary
This uses sdk names as well as project to differentiate when it comes to multi-service single project traces. This also adds some small UI tweaks.

### DIFF
--- a/static/app/views/performance/traces/content.tsx
+++ b/static/app/views/performance/traces/content.tsx
@@ -33,6 +33,7 @@ import {
   ProjectRenderer,
   SpanBreakdownSliceRenderer,
   SpanIdRenderer,
+  SpanTimeRenderer,
   TraceBreakdownContainer,
   TraceBreakdownRenderer,
   TraceIdRenderer,
@@ -141,9 +142,11 @@ export function Content() {
             {t('Duration')}
           </StyledPanelHeader>
           <StyledPanelHeader align="right" lightText>
+            {t('Timestamp')}
+          </StyledPanelHeader>
+          <StyledPanelHeader align="right" lightText>
             {t('Issues')}
           </StyledPanelHeader>
-          <StyledPanelHeader align="right" lightText style={{padding: '5px'}} />
           {isLoading && (
             <StyledPanelItem span={7} overflow>
               <LoadingIndicator />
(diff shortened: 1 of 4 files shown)


static/app/views/performance/traces/content.tsx
```diff
@@ -399,6 +402,10 @@ function useTraces<F extends string>({
       sort,
       per_page: limit,
       maxSpansPerTrace: 10,
+      // breakdownCategory: ['db'],
+      breakdownCategory: [],
+      minBreakdownDuration: 2,
+      minBreakdownPercentage: 1 / 40,
```
Zylphrex, line 408: Did we want to keep the min breakdown duration and percentage now that we're not using the categories?

static/app/views/performance/traces/utils.tsx
```diff
@@ -9,3 +10,28 @@ export function normalizeTraces(traces: TraceResult<string>[] | undefined) {
     (t1, t2) => (t1.name ? '0' : '1').localeCompare(t2.name ? '0' : '1')
   );
 }
+
+export function getStylingSliceName(
+  sliceName: string | null,
+  sliceSecondaryName: string | null
+) {
+  if (sliceSecondaryName) {
+    return sliceSecondaryName.slice(-2) + (sliceName ? sliceName.slice(0, 3) : '');
```
Zylphrex, line 19: Could we leave a comment explaining why this is done this way? Without context, this is very cryptic.

static/app/views/performance/traces/utils.tsx
```diff
@@ -9,3 +10,28 @@ export function normalizeTraces(traces: TraceResult<string>[] | undefined) {
...
+  }
+
+  return sliceName;
+}
+
+const PRIORITIZED_BREAKDOWN_OPS = ['db'];
```
Zylphrex, line 25: Are we keeping the db stuff?

Zylphrex (approved):

static/app/views/performance/traces/content.tsx
k-fish: Nah I'll remove em

static/app/views/performance/traces/utils.tsx
k-fish: nope

PR #70519, opened 2024-05-08
feat(metrics): Open metrics explorer in traces explorer

Zylphrex:
This adds an open in trace explorer button in the metrics page. There's still a few bugs where the traces explorer errors in some cases but this button is behind a feature flag to test.

### DIFF
--- a/static/app/views/performance/traces/content.tsx
+++ b/static/app/views/performance/traces/content.tsx
@@ -3,6 +3,7 @@ import {useTheme} from '@emotion/react';
 import styled from '@emotion/styled';
 import debounce from 'lodash/debounce';
 
+import {Alert} from 'sentry/components/alert';
 import {Button} from 'sentry/components/button';
 import Count from 'sentry/components/count';
 import EmptyStateWarning from 'sentry/components/emptyStateWarning';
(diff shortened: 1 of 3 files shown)


static/app/views/performance/traces/content.tsx
```diff
@@ -119,6 +128,13 @@ export function Content() {
         <EnvironmentPageFilter />
         <DatePageFilter defaultPeriod="2h" />
       </PageFilterBar>
+      {hasMetric && (
+        <StyledAlert type="info" showIcon>
+          {tct('The metric query [metricQuery] is filtering the results below.', {
+            metricQuery: <strong>{`${metricsOp}(${mri}){${metricsQuery || ''}}`}</strong>,
```
matejminar, line 134: There's this helper that you might find useful:
https://github.com/getsentry/sentry/blob/d23ed4797b4a122b40cfc6df3370c80b354a1979/static/app/utils/metrics/index.tsx#L355

static/app/views/performance/traces/content.tsx
```diff
@@ -119,6 +128,13 @@ export function Content() {
         <EnvironmentPageFilter />
         <DatePageFilter defaultPeriod="2h" />
       </PageFilterBar>
+      {hasMetric && (
+        <StyledAlert type="info" showIcon>
+          {tct('The metric query [metricQuery] is filtering the results below.', {
```
matejminar, line 133: Should we also mention the min+max value here?

matejminar (approved):

static/app/views/performance/traces/content.tsx
Zylphrex: I haven't implemented the min/max part yet but will follow up on that

PR #70522, opened 2024-05-08
feat(performance): Update queues module to use span.duration and fixes some minor chart issues

edwardgou-sentry:
- We want to display `span.duration` instead of `span.self_time` for queue spans, because we're interested in the duration of any potential child spans during a consumer/producer task. Updates all instances of `span.self_time` to `span.duration`
 - Also includes some minor fixes for charts, such as correcting avg mark line and tooltip + axis formatting.


gggritso (approved):
Very nice!

PR #70523, opened 2024-05-08
feat(slack): Add support for external issue creation

ykamo001:
Add support for an external issue created activity event notification for Slack threads. To do this, we added a few classes that will parse the specific pieces of information we need, and added logging and metrics to track "bad state" activities, as we assume some data to always be there as a form of contract. Also added test cases to help validate logic and keep it consistent from breaking changes.

With the changes, we should now be able to create a slack notification that looks like the following:
```
<Author> created a (<Provider> issue <Ticket Number>)[Link]
```

Special consideration had to be done for Asana for now, and we might be able to remove that special consideration if we can update the activity data object when an asana issue is created, but the changes would need to be verified that it does not affect other assumptions/contracts.

### DIFF
--- a/src/sentry/integrations/slack/threads/activity_notifications.py
+++ b/src/sentry/integrations/slack/threads/activity_notifications.py
@@ -1,9 +1,16 @@
+import logging
 from collections.abc import Mapping
 from typing import Any
 
+from sentry.models.activity import Activity
 from sentry.notifications.notifications.activity.assigned import (
     AssignedActivityNotification as BaseAssignedActivityNotification,
 )
+from sentry.notifications.notifications.activity.base import GroupActivityNotification
+from sentry.types.activity import ActivityType
+from sentry.utils import metrics
+
+_default_logger = logging.getLogger(__name__)
 
 
 class AssignedActivityNotification(BaseAssignedActivityNotification):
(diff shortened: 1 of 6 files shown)


src/sentry/integrations/slack/threads/activity_notifications.py
```diff
@@ -15,3 +22,185 @@ class AssignedActivityNotification(BaseAssignedActivityNotification):
...
+    for that change are not yet known.
+    """
+
+    _DEFAULT_ASANA_LABEL_VALUE = "Asana Issue"
+
+    def get_ticket_number(self) -> str:
+        # Try to use the base logic if it works as a just in-case
+        stored_value = super().get_ticket_number()
+        if stored_value != "" and stored_value != self._DEFAULT_ASANA_LABEL_VALUE:
+            return stored_value
```
cathteng, line 145: should we be tracking how often this happens?

src/sentry/integrations/slack/threads/activity_notifications.py
```diff
@@ -15,3 +22,185 @@ class AssignedActivityNotification(BaseAssignedActivityNotification):
...
+class ExternalIssueCreatedActivityNotification(GroupActivityNotification):
+    metrics_key = "create_issue"
+    title = "External Issue Created"
+
+    def get_description(self) -> tuple[str, str | None, Mapping[str, Any]]:
+        external_issue = _external_issue_activity_factory(activity=self.activity)
+
+        provider = external_issue.get_provider()
+        # Use proper grammar, so use "an" if it's "external provider" and "a" if it's a regular name
+        if provider == external_issue.DEFAULT_PROVIDER_FALLBACK_TEXT:
```
cathteng, line 186: Asana should use "an" too?

src/sentry/integrations/slack/threads/activity_notifications.py
```diff
@@ -15,3 +22,185 @@ class AssignedActivityNotification(BaseAssignedActivityNotification):
...
+            base_template += " {ticket}"
+
+        link = external_issue.get_link()
+        if link:
+            base_template = "<{link}|" + base_template + ">"
+
+        # Template should look something like "{author} created <{link}| a/an {provider} issue {ticket}>"
+        base_template = "{author} created " + base_template
+
+        return base_template, None, {"provider": provider, "ticket": ticket_number, "link": link}
```
cathteng, line 206: i see this is used in `get_context` so i figured out what it's doing but it's still a lot 😭

cathteng (approved):
nice abstraction. it seems weird to me that the external issue might not have a link but it doesn't hurt to have that check

we also have an entire `ExternalIssue` model but it's not linked to `Activity` at all except for a few things in the data column in `Activity` :/ (but the `ExternalIssue` id isn't even added)

src/sentry/integrations/slack/threads/activity_notifications.py
ykamo001: I guess so, could check for first letter being a vowel

src/sentry/integrations/slack/threads/activity_notifications.py
ykamo001: It would probably just be easier to do a data check through ReDash and check the values. Tracking/metrics on this wouldn't help us prevent anything anyways or have us actively take action

src/sentry/integrations/slack/threads/activity_notifications.py
ykamo001: > but it's still a lot 😭

What's a lot?

cathteng: it's trying to do 3 things in one function

cathteng: by returning 3 things

ykamo001: yeah, I am not sure why this convention was used, definitely mixed responsibilities

ykamo001:
> nice abstraction. it seems weird to me that the external issue might not have a link but it doesn't hurt to have that check
> 
> we also have an entire `ExternalIssue` model but it's not linked to `Activity` at all except for a few things in the data column in `Activity` :/ (but the `ExternalIssue` id isn't even added)

Good find! I'll check out that model and see if there's a way to get the relation and maybe we could just use the source of truth

src/sentry/integrations/slack/threads/activity_notifications.py
ykamo001: I'll pull that into the PR where I'll try to use the externalIssue model

src/sentry/integrations/slack/threads/activity_notifications.py
```diff
@@ -15,3 +22,185 @@ class AssignedActivityNotification(BaseAssignedActivityNotification):
...
+                extra={
+                    "activity_id": self._activity.id,
+                },
+            )
+            ticket_number = ""
+
+        return ticket_number
+
+
+class _AsanaExternalIssueCreatedActivity(_ExternalIssueCreatedActivity):
```
iamrajjoshi, line 132: @ykamo001  why is asana under `/slack/` lol

sentaur-athena: 🤣

PR #70527, opened 2024-05-08
feat(snub-heavy-search): adds exceptions to pre-fetch groupids from postgres under certain conditions

scefali:
There are certain filters such as `bookmarked_by`, `linked`, `subscribed_by` that require us to do joins in Postgres to other tables. These tables do not exist in Snuba and likely never will because these queries are somewhat uncommon. As a result, this PR adds pre-fetching group_ids when users are trying to filter with one of these conditions with the new `GroupAttributesPostgresSnubaQueryExecutor`. The long term goal is to remove `UNSUPPORTED_SNUBA_FILTERS` so we can have all queries use `GroupAttributesPostgresSnubaQueryExecutor` instead of `PostgresSnubaQueryExecutor`.

### DIFF
--- a/src/sentry/search/snuba/executors.py
+++ b/src/sentry/search/snuba/executors.py
@@ -93,6 +93,11 @@ class Clauses(Enum):
     WHERE = auto()
 
 
+# we cannot use snuba for these fields because they require a join with tables that don't exist there
+# if we ever see these fields, we will use postgres to get the group_ids before sending back to ClickHouse
+POSTGRES_ONLY_SEARCH_FIELDS = ["bookmarked_by", "linked", "subscribed_by"]
+
+
 @dataclass
 class TrendsParams:
     # (event or issue age_hours) / (event or issue halflife hours)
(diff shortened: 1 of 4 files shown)


src/sentry/search/snuba/executors.py
```diff
@@ -93,6 +93,11 @@ class Clauses(Enum):
     WHERE = auto()
 
 
+# we cannot use snuba for these fields because they require a join with tables that don't exist there
+# if we ever see these fields, we will use postgres to get the group_ids before sending back to ClickHouse
+POSTGRES_ONLY_SEARCH_FIELDS = ["bookmarked_by", "linked", "subscribed_by"]
```
wedamija, line 98: Seems ok, these tables are usually small enough that this will be fine. I could imagine that possibly `linked` might end up with a lot of rows due to an automation, although in the past when I checked it was still pretty small

src/sentry/search/snuba/executors.py
```diff
@@ -1564,6 +1580,16 @@ def query(
...
             having = []
+            # if we need to prefetch from postgres, we add filter by the group ids
+            if group_ids_to_pass_to_snuba is not None:
+                # will not find any matches, we can return early
+                if len(group_ids_to_pass_to_snuba) == 0:
+                    return self.empty_result
+
+                where_conditions.append(
+                    Condition(Column("group_id", attr_entity), Op.IN, group_ids_to_pass_to_snuba)
+                )
```
wedamija, line 1591: I wonder if we should filter on both the `event` and `attr` group id? I guess the join is doing that already, so maybe it doesn't matter...

wedamija (approved):

src/sentry/search/snuba/executors.py
scefali: @wedamija not a bad idea, worth adding

PR #70529, opened 2024-05-08
fix(issue-priority): Add Chevron-up state for when group priority dropdown is open

MichaelSun48:
Currently, the chevron for the issue priority tag does not animate or flip up when the dropdown is toggled. This PR fixes that.

Before: 
Mov to Gif conversion (1)

After:
Mov to Gif conversion

### DIFF
--- a/static/app/components/badge/groupPriority.tsx
+++ b/static/app/components/badge/groupPriority.tsx
@@ -1,4 +1,4 @@
-import {Fragment, useMemo, useRef} from 'react';
+import {Fragment, useMemo, useRef, useState} from 'react';
 import type {Theme} from '@emotion/react';
 import styled from '@emotion/styled';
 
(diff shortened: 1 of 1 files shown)


static/app/components/badge/groupPriority.tsx
```diff
@@ -211,6 +211,7 @@ export function GroupPriorityDropdown({
     () => makeGroupPriorityDropdownOptions({onChange}),
     [onChange]
   );
+  const [isDropdownOpen, setIsDropdownOpen] = useState(false);
```
vuluongj20, line 214: `DropdownMenu`'s `trigger` prop supplies the open state as the second argument in the render function for cases like this:

image

vuluongj20 (approved):
🚀

PR #70535, opened 2024-05-08
feat(replay): viewed_by_me search alias

aliu39:
Follow up from https://github.com/getsentry/sentry/issues/64924

On the backend, we query by viewed_by_id. Previously, we expose this in the search bar typeahead. But this doesn't really make sense since you need to be an admin to lookup user ids. So we'll use this for convenience - "have **I** seen/not seen this?"


src/sentry/replays/usecases/query/__init__.py
```diff
@@ -163,6 +165,24 @@ def query_using_optimized_search(
...
         ]
 
+    for i in range(len(search_filters)):
+        search_filter = search_filters[i]
+        if (
+            isinstance(search_filter, SearchFilter)
+            and search_filter.key.name in VIEWED_BY_ME_KEY_ALIASES
+        ):
+            # "viewed_by_me" is not a valid query field.
+            # It's a convenience alias for users without the admin access to lookup ids.
```
JoshFerge, line 175: >            # It's a convenience alias for users without the admin access to lookup ids.

isn't this all users? should we just remove this comment as no sentry users have admin access?

aliu3ntry: Yeah, maybe
"It's a convenience alias since only Sentry employees can lookup ids"
?

JoshFerge: why do we need to clarify this?

JoshFerge: (it seems to be a given)

aliu3ntry: Other ppl on our team weren't clear about this when we built it, which is why we exposed viewed_by_id

JoshFerge: right, but the comment implies that us going and looking up ids is a normal use case, but it is not -- i think thats where my confusion lies

aliu3ntry: Oh, I see. I guess that can be misleading, I can remove it then

tests/sentry/replays/test_organization_replay_index.py
```diff
@@ -777,6 +783,13 @@ def test_get_replays_user_filters_invalid_operator(self):
             response = self.client.get(self.url + "?field=id&query=transaction.duration:>0")
             assert response.status_code == 400
 
+    def test_get_replays_user_filters_invalid_value(self):
+        self.create_project(teams=[self.team])
+
+        with self.feature(REPLAYS_FEATURES):
+            response = self.client.get(self.url + "?field=id&query=viewed_by_me:potato")
```
cmanallen, line 790: Needs an invalid operator test too.

src/sentry/replays/usecases/query/__init__.py
```diff
@@ -43,6 +43,46 @@
...
+        # since the value is boolean, negations (!) are not allowed
+        if search_filter.operator != "=":
+            raise ParseError(f"Invalid operator specified for `{search_filter.key.name}`")
+
+        if not isinstance(
+            search_filter.value.value, str
+        ) or search_filter.value.value.lower() not in ["true", "false"]:
+            raise ParseError(f"Could not parse value for `{search_filter.key.name}`")
+
+        operator = "=" if search_filter.value.value.lower() == "true" else "!="
```
cmanallen, line 75: Thinking on this more.  Let's always use the `=` operator.  The value `true` or `false` will change based on user input.

src/sentry/replays/usecases/query/__init__.py
```diff
@@ -43,6 +43,46 @@
...
+            new_filters.append(search_filter)
+            continue
+
+        # since the value is boolean, negations (!) are not allowed
+        if search_filter.operator != "=":
+            raise ParseError(f"Invalid operator specified for `{search_filter.key.name}`")
+
+        if not isinstance(
+            search_filter.value.value, str
+        ) or search_filter.value.value.lower() not in ["true", "false"]:
```
cmanallen, line 72: `value = bool(search_filter.value.value)`.  If an exception is raised return the parse error.

cmanallen (changes requested):

src/sentry/replays/usecases/query/__init__.py
cmanallen: Sorry ignore this.  This is wrong.  I wasn't thinking clearly.

src/sentry/replays/usecases/query/__init__.py
aliu3ntry: python is weird.... i tried in an interpreter and bool("false") = True 🫨

aliu3ntry: any non-empty string is converted to true i think

cmanallen: Oof yeah.  That was bad advice on my part!

JoshFerge (commented):
worth trying to get to 100% code coverage on this one, can see which lines are missing it

aliu3ntry:
> worth trying to get to 100% code coverage on this one, can see which lines are missing it
@JoshFerge 
Looks like it's the request_user_id=None case, which happens when request.user is None -- not sure when this would happen in prod and how to write a test for it. I guess when you view the page without logging into Sentry? Is such a thing possible?

JoshFerge:
> > worth trying to get to 100% code coverage on this one, can see which lines are missing it
> > @JoshFerge
> > Looks like it's the request_user_id=None case, which happens when request.user is None -- not sure when this would happen in prod and how to write a test for it. I guess when you view the page without logging into Sentry? Is such a thing possible?

if its impossible, do we need to handle that case?

JoshFerge:
> > > > worth trying to get to 100% code coverage on this one, can see which lines are missing it
> > > > @JoshFerge
> > > > Looks like it's the request_user_id=None case, which happens when request.user is None -- not sure when this would happen in prod and how to write a test for it. I guess when you view the page without logging into Sentry? Is such a thing possible?
> > 
> > 
> > if its impossible, do we need to handle that case?

I'm not sure if it's possible. The type of request.user is an optional, but let me dig into django docs for this. i don't think it should ever be possible for a non logged in user to make it through auth for this, so at the top of the endpoint you could just say if `request.User is None` then return Unauthorized, or something like that

aliu3ntry:
> > > > > worth trying to get to 100% code coverage on this one, can see which lines are missing it
> > > > > @JoshFerge
> > > > > Looks like it's the request_user_id=None case, which happens when request.user is None -- not sure when this would happen in prod and how to write a test for it. I guess when you view the page without logging into Sentry? Is such a thing possible?
> > > 
> > > 
> > > if its impossible, do we need to handle that case?
> 
> I'm not sure if it's possible. The type of request.user is an optional, but let me dig into django docs for this. i don't think it should ever be possible for a non logged in user to make it through auth for this, so at the top of the endpoint you could just say if `request.User is None` then return Unauthorized, or something like that

Sounds good!

cmanallen (changes requested):

src/sentry/replays/usecases/query/__init__.py
```diff
@@ -43,6 +43,47 @@
...
+    for search_filter in search_filters:
+        if (
+            not isinstance(search_filter, SearchFilter)
+            or search_filter.key.name not in VIEWED_BY_ME_KEY_ALIASES
+        ):
+            new_filters.append(search_filter)
+            continue
+
+        if request_user_id is None:
+            raise NotAuthenticated("No request user provided for viewed_by_me filter")
```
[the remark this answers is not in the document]
aliu3ntry: Ok, got it. My thought process was this might only apply to this filter, but I guess it's worth validating for the whole get method!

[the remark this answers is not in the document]
aliu3ntry: This breaks type-checking though, since calls to the query fxs with request_user=None (e.g. from the delete_replays script) pass through handle_viewed_by_me. How should we handle this?

PR #70538, opened 2024-05-08
feat(analytics): adds analytics for issue search on the backend

scefali:
We want to have a better understanding what queries are being used on the backend for issue search. We have analytics on the FE for issue stream loading but it doesn't cover a lot of places that make searches such as the releases tab.


JoshFerge:
i'm not clear on the use case here -- why do we need to know what individual users did what queries?

scefali:
@JoshFerge we don't necessarily care about the individual user (though that could have some interesting analysis too) but we do care about which searches are used for the purposes of guiding development (especially with the snuba-heavy search). For example, I want to know how often `release.build` is used to gauge the impact of adding a `releases` dataset to ClickHouse.

JoshFerge (approved):

PR #70539, opened 2024-05-08
fix(replays): Fix 'clearQueryCache' method in 'useReplayData' hook

floels:
# Goal

The goal of this PR is to fix a bug currently causing the `clearQueryCache` method defined in `static/app/utils/replays/hooks/useReplayData.tsx` to have no effect (extra arrow function declaration).

# Approach

Remove the duplicated arrow function declaration syntax (`() => { ... }`).

We also add a unit test to cover this behavior of the custom hook. You can see that the test fails if you revert the fix in `useReplayData.tsx`.

# Legal Boilerplate

### DIFF
--- a/static/app/utils/replays/hooks/useReplayData.spec.tsx
+++ b/static/app/utils/replays/hooks/useReplayData.spec.tsx
@@ -32,10 +32,14 @@ jest.mocked(useProjects).mockReturnValue({
   placeholders: [],
 });
 
+const mockInvalidateQueries = jest.fn();
+
 function wrapper({children}: {children?: ReactNode}) {
-  return (
-    <QueryClientProvider client={makeTestQueryClient()}>{children}</QueryClientProvider>
-  );
+  const queryClient = makeTestQueryClient();
+
+  queryClient.invalidateQueries = mockInvalidateQueries;
+
+  return <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>;
 }
 
 function getMockReplayRecord(replayRecord?: Partial<ReplayRecord>) {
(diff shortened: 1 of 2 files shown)


static/app/utils/replays/hooks/useReplayData.tsx
```diff
@@ -214,21 +214,19 @@ function useReplayData({
...
-          `/projects/${orgSlug}/${projectSlug}/replays/${replayId}/recording-segments/`,
-        ],
-      });
-      // The next one isn't optimized
-      // This statement will invalidate the cache of fetched error events for all replayIds
-      queryClient.invalidateQueries({
-        queryKey: [`/organizations/${orgSlug}/replays-events-meta/`],
-      });
-    };
+    queryClient.invalidateQueries({
```
floels, line 217: This is only removing the extraneous `() => {` on line `217` and the corresponding `}` on line `231`.

static/app/utils/replays/hooks/useReplayData.spec.tsx
```diff
@@ -32,10 +32,14 @@ jest.mocked(useProjects).mockReturnValue({
   placeholders: [],
 });
 
+const mockInvalidateQueries = jest.fn();
```
ryan953, line 35: Can we reset this before or after each test? This way if another test gets added the state from a previous test won't affect the new one.

ryan953 (commented):
Looks awesome, thanks for identifying that problem @floels and writing tests too!

static/app/utils/replays/hooks/useReplayData.spec.tsx
floels: Ah yes good point, added in the existing `beforeEach`.

ryan953 (approved):

PR #70543, opened 2024-05-08
ref(rules): Fetch events and occurrences in batches

ceorourke:
Fetch events and issue occurrences in bulk in the delayed rule processor. 

Closes https://github.com/getsentry/team-core-product-foundations/issues/306 and https://getsentry.atlassian.net/browse/ALRT-5

### DIFF
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -4,7 +4,9 @@ from collections import defaultdict
 from datetime import datetime, timedelta, timezone
 from typing import Any, DefaultDict, NamedTuple
 
-from sentry import eventstore
+from google.api_core.exceptions import DeadlineExceeded, ServiceUnavailable
+
+from sentry import nodestore
 from sentry.buffer.redis import BufferHookEvent, RedisBuffer, redis_buffer_registry
 from sentry.eventstore.models import Event, GroupEvent
 from sentry.issues.issue_occurrence import IssueOccurrence
@@ -200,6 +202,31 @@ def parse_rulegroup_to_event_data(
     return parsed_rulegroup_to_event_data
 
 
+def bulk_fetch_events(event_ids: list[str], project_id: int) -> set[Event]:
+    node_id_to_event_id = {
+        Event.generate_node_id(project_id, event_id=event_id): event_id for event_id in event_ids
+    }
+
+    try:
+        bulk_data = nodestore.get_multi(list(node_id_to_event_id.keys()))
+    except (ServiceUnavailable, DeadlineExceeded):
+        try:
+            bulk_data = nodestore.get_multi(list(node_id_to_event_id.keys()))
+        except (ServiceUnavailable, DeadlineExceeded):
+            bulk_data = nodestore.get_multi(list(node_id_to_event_id.keys()))
+
+    bulk_events = set()
+    for node_id, data in bulk_data.items():
+        if node_id in node_id_to_event_id:
+            event_id = node_id_to_event_id[node_id]
+            if data is not None:
+                event = Event(event_id=event_id, project_id=project_id)
+                event.data = data
+            bulk_events.add(event)
+
+    return bulk_events
+
+
 def get_group_to_groupevent(
     parsed_rulegroup_to_event_data: dict[tuple[str, str], dict[str, str]],
     project_id: int,
@@ -208,30 +235,43 @@ def get_group_to_groupevent(
     group_to_groupevent: dict[Group, GroupEvent] = {}
     groups = Group.objects.filter(id__in=group_ids)
     group_id_to_group = {group.id: group for group in groups}
+    event_ids_to_group_ids: dict[str, int] = {}
+    occurrence_ids: list[str] = []
+
     for rule_group, instance_data in parsed_rulegroup_to_event_data.items():
         event_id = instance_data.get("event_id")
-        occurrence_id = instance_data.get("occurrence_id")
         group_id = rule_group[1]
-        group = group_id_to_group.get(int(group_id))
-        if group and event_id:
-            # TODO: fetch events and occurrences in batches
-            event = Event(
-                event_id=event_id,
-                project_id=project_id,
-                snuba_data={
-                    "event_id": event_id,
-                    "group_id": group.id,
-                    "project_id": project_id,
-                },
-            )
-            eventstore.backend.bind_nodes([event])
-            group_event = event.for_group(group)
-            if occurrence_id:
-                occurrence = IssueOccurrence.fetch(occurrence_id, project_id=project_id)
-                if occurrence:
-                    group_event.occurrence = occurrence
+        if event_id:
+            event_ids_to_group_ids[event_id] = int(group_id)
+        occurrence_id = instance_data.get("occurrence_id")
+        if occurrence_id:
+            occurrence_ids.append(occurrence_id)
 
-            group_to_groupevent[group] = group_event
+    bulk_events = bulk_fetch_events(list(event_ids_to_group_ids.keys()), project_id)
+    bulk_occurrences = IssueOccurrence.fetch_multi(occurrence_ids, project_id=project_id)
+
+    for event_id, group_id in event_ids_to_group_ids.items():
+        event = None
+        group_event = None
+        occurrence = None
+
+        for bulk_event in bulk_events:
+            if bulk_event.event_id == event_id:
+                event = bulk_event
+        group = group_id_to_group.get(group_id)
+        if event and group:
+            group_event = event.for_group(group)
+
+        if group_event:
+            for bulk_occurrence in bulk_occurrences:
+                if bulk_occurrence:
+                    if occurrence_id == bulk_occurrence.id:
+                        occurrence = bulk_occurrence
+            if occurrence:
+                group_event.occurrence = occurrence
+            if group:
+                group_to_groupevent[group] = group_event
+
     return group_to_groupevent
 
 
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -35,6 +35,7 @@ from sentry.utils.retries import ConditionalRetryPolicy, exponential_delay
 from sentry.utils.safe import safe_execute
 
 logger = logging.getLogger("sentry.rules.delayed_processing")
+EVENT_LIMIT = 100
 
 
 class UniqueCondition(NamedTuple):
@@ -204,7 +205,6 @@ def parse_rulegroup_to_event_data(
 
 
 def bulk_fetch_events(event_ids: list[str], project_id: int) -> dict[str, Event]:
-    event_limit = 100
     node_id_to_event_id: dict[str, str] = {
         Event.generate_node_id(project_id, event_id=event_id): event_id for event_id in event_ids
     }
@@ -212,7 +212,7 @@ def bulk_fetch_events(event_ids: list[str], project_id: int) -> dict[str, Event]
     fetch_retry_policy = ConditionalRetryPolicy(should_retry_fetch, exponential_delay(1.00))
 
     bulk_data = {}
-    for node_id_chunk in chunked(node_ids, event_limit):
+    for node_id_chunk in chunked(node_ids, EVENT_LIMIT):
         bulk_results = fetch_retry_policy(lambda: nodestore.backend.get_multi(node_id_chunk))
         bulk_data.update(bulk_results)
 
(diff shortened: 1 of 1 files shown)


src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,6 +202,31 @@ def parse_rulegroup_to_event_data(
     return parsed_rulegroup_to_event_data
 
 
+def bulk_fetch_events(event_ids: list[str], project_id: int) -> set[Event]:
```
ceorourke, line 205: This was largely copied from https://github.com/getsentry/getsentry/blob/b2689f4e98cb5d641ecdda4605e9cf84932b2a67/bin/backfill_perf_issue_events_issue_platform.py#L506-L530 - but I removed the `time.sleep`s. It seems like this could be useful somewhere else, I was surprised we didn't already have this function, though I'm not sure where would be a good place for it to live.

saponifi3d: 🤔 If I were looking for this code I'd probably check in `sentry.eventstore.models` where the `Event` is imported from. I'm thinking basically i know what model i want, and then i can either get 1 or many of them from there.

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,6 +202,31 @@ def parse_rulegroup_to_event_data(
...
+        Event.generate_node_id(project_id, event_id=event_id): event_id for event_id in event_ids
+    }
+
+    try:
+        bulk_data = nodestore.get_multi(list(node_id_to_event_id.keys()))
+    except (ServiceUnavailable, DeadlineExceeded):
+        try:
+            bulk_data = nodestore.get_multi(list(node_id_to_event_id.keys()))
+        except (ServiceUnavailable, DeadlineExceeded):
+            bulk_data = nodestore.get_multi(list(node_id_to_event_id.keys()))
```
saponifi3d, line 216: It'd be cool if this just called bulk_fetch_events again instead of repeating the code.

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -208,30 +235,43 @@ def get_group_to_groupevent(
...
-            group_to_groupevent[group] = group_event
+        if group_event:
+            for bulk_occurrence in bulk_occurrences:
+                if bulk_occurrence:
+                    if occurrence_id == bulk_occurrence.id:
+                        occurrence = bulk_occurrence
+            if occurrence:
+                group_event.occurrence = occurrence
+            if group:
+                group_to_groupevent[group] = group_event
```
saponifi3d, line 273: might be nice to decompose this into a new function to reduce the cyclomatic complexity of `get_group_to_groupevent` - it's also hard to tell at a glance what this chunk of code is doing, so pulling it into a function and giving it a name will help improve the readability.

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,6 +202,31 @@ def parse_rulegroup_to_event_data(
...
+            bulk_data = nodestore.get_multi(list(node_id_to_event_id.keys()))
+
+    bulk_events = set()
+    for node_id, data in bulk_data.items():
+        if node_id in node_id_to_event_id:
+            event_id = node_id_to_event_id[node_id]
+            if data is not None:
+                event = Event(event_id=event_id, project_id=project_id)
+                event.data = data
+            bulk_events.add(event)
```
saponifi3d, line 225: Rather than nesting the if statements we can pull them to the top level.
```suggestion
        event_id = node_id_event_id.get(node_id, None)
        if data is not None and event_id is not None:
            event = Event(event_id=event_id, project_id=project_id)
            event.data = data
            bulk_events.add(event)
```

This refactor shows an interesting side effect of your current code, that you could have an event_id but not data - at least in the logic. It's okay because the set will ensure uniqueness and not add the event multiple times, but it could be confusing in the future. It'd be better to only add the event to bulk events if the event is created though.

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -208,30 +235,43 @@ def get_group_to_groupevent(
...
-            )
-            eventstore.backend.bind_nodes([event])
+        if event_id:
+            event_ids_to_group_ids[event_id] = int(group_id)
+        occurrence_id = instance_data.get("occurrence_id")
+        if occurrence_id:
+            occurrence_ids.append(occurrence_id)
+
+    bulk_events = bulk_fetch_events(list(event_ids_to_group_ids.keys()), project_id)
+    bulk_occurrences = IssueOccurrence.fetch_multi(occurrence_ids, project_id=project_id)
```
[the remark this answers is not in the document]
ceorourke: Ok I updated the bulk events, not sure the best way yet to do it for the occurrences since I don't have control over the output of `fetch_multi`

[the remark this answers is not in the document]
wedamija: You can just do `{occurrence.id: occurrence for occurrence in ...}` (might not be the exact property names, but that idea)

src/sentry/rules/processing/delayed_processing.py
ceorourke: I think for _right now_ I wanna keep it where it is, then when I have more confidence it's working as expected it can be moved for others to use.

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,38 +203,90 @@ def parse_rulegroup_to_event_data(
...
+    node_id_to_event_id: dict[str, str] = {}
+
+    for event_id_chunk in chunked(event_ids, event_limit):
+        for event_id in event_id_chunk:
+            node_id = Event.generate_node_id(project_id, event_id=event_id)
+            node_id_to_event_id[node_id] = event_id
+
+    node_ids = list(node_id_to_event_id.keys())
+    fetch_retry_policy = ConditionalRetryPolicy(should_retry_fetch, exponential_delay(1.00))
+    bulk_data = fetch_retry_policy(lambda: nodestore.backend.get_multi(node_ids))
```
wedamija, line 217: This is the part that needs to be batched, right now you're still fetching all the node ids at the same time

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,38 +203,90 @@ def parse_rulegroup_to_event_data(
...
-                    "project_id": project_id,
-                },
-            )
-            eventstore.backend.bind_nodes([event])
-            group_event = event.for_group(group)
-            if occurrence_id:
-                occurrence = IssueOccurrence.fetch(occurrence_id, project_id=project_id)
-                if occurrence:
-                    group_event.occurrence = occurrence
+        group_event = None
```
wedamija, line 241: This is unused

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,38 +203,90 @@ def parse_rulegroup_to_event_data(
...
+    group_id_to_group: dict[int, Group],
 ) -> dict[Group, GroupEvent]:
     group_to_groupevent: dict[Group, GroupEvent] = {}
-    groups = Group.objects.filter(id__in=group_ids)
-    group_id_to_group = {group.id: group for group in groups}
+
     for rule_group, instance_data in parsed_rulegroup_to_event_data.items():
         event_id = instance_data.get("event_id")
         occurrence_id = instance_data.get("occurrence_id")
         group_id = rule_group[1]
```
wedamija, line 240: We may as well just use this directly since it's only used in one place

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,38 +203,90 @@ def parse_rulegroup_to_event_data(
...
+        if event_id:
+            event = bulk_event_id_to_events.get(event_id)
+        group = group_id_to_group.get(int(group_id))
+        if not group or not event:
+            continue
+
+        group_event = event.for_group(group)
+        if occurrence_id:
+            occurrence = bulk_occurrence_id_to_occurrence.get(occurrence_id)
+        if occurrence and group_event:
```
wedamija, line 253: I think we can remove this if, we're guaranteed to have a `group_event`, and if `occurrence` is None then it doesn't hurt to assign none to the occurrence property

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,38 +203,90 @@ def parse_rulegroup_to_event_data(
...
+            group_event.occurrence = occurrence
+        if group_event:
```
wedamija, line 255: Can remove this if as well

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,38 +203,90 @@ def parse_rulegroup_to_event_data(
...
+    parsed_rulegroup_to_event_data: dict[tuple[str, str], dict[str, str]],
+    project_id: int,
+    group_ids: set[int],
+) -> dict[Group, GroupEvent]:
+    groups = Group.objects.filter(id__in=group_ids)
+    group_id_to_group = {group.id: group for group in groups}
+    event_ids: set[str] = set()
+    occurrence_ids: list[str] = []
+
+    for _, instance_data in parsed_rulegroup_to_event_data.items():
```
wedamija, line 271: Nit: Can use `.values()` here and skip the `_,`

src/sentry/rules/processing/delayed_processing.py
ceorourke: I'm not sure what you mean here 🤔

src/sentry/rules/processing/delayed_processing.py
ceorourke: Oh oops lol, hopefully batched the right thing now

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,38 +203,85 @@ def parse_rulegroup_to_event_data(
...
+def bulk_fetch_events(event_ids: list[str], project_id: int) -> dict[str, Event]:
+    event_limit = 100
+    node_id_to_event_id: dict[str, str] = {
+        Event.generate_node_id(project_id, event_id=event_id): event_id for event_id in event_ids
+    }
+    node_ids = list(node_id_to_event_id.keys())
+    fetch_retry_policy = ConditionalRetryPolicy(should_retry_fetch, exponential_delay(1.00))
+
+    for node_id_chunk in chunked(node_ids, event_limit):
+        bulk_data = fetch_retry_policy(lambda: nodestore.backend.get_multi(node_id_chunk))
```
wedamija, line 215: You need to define the dict, and do `bulk_data.update(...)` here, otherwise you're just overwriting results

src/sentry/rules/processing/delayed_processing.py
saponifi3d: I believe Dan's suggesting to move `rule_group[1]` to replace `group_id` in: `group = group_id_to_group.get(int(group_id))` so you can remove this variable, because we're only using `group_id` once.

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,38 +203,87 @@ def parse_rulegroup_to_event_data(
     return parsed_rulegroup_to_event_data
 
 
+def bulk_fetch_events(event_ids: list[str], project_id: int) -> dict[str, Event]:
+    event_limit = 100
```
saponifi3d, line 207: nit: i'd recommend pulling this out as a constant, `EVENT_LIMIT`

src/sentry/rules/processing/delayed_processing.py
wedamija: Just a nit, not that important but
`group = group_id_to_group.get(int(group_id))`
could just be
`group = group_id_to_group.get(int(rule_group[1]))`
Either way is fine

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,38 +203,87 @@ def parse_rulegroup_to_event_data(
...
+    node_id_to_event_id: dict[str, str] = {
+        Event.generate_node_id(project_id, event_id=event_id): event_id for event_id in event_ids
+    }
+    node_ids = list(node_id_to_event_id.keys())
+    fetch_retry_policy = ConditionalRetryPolicy(should_retry_fetch, exponential_delay(1.00))
```
saponifi3d, line 212: woah, nice!

saponifi3d (approved):
lgtm!

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,38 +203,87 @@ def parse_rulegroup_to_event_data(
...
+        event_id = instance_data.get("event_id")
+        occurrence_id = instance_data.get("occurrence_id")
+        group_id = rule_group[1]
+        occurrence = None
+
+        if event_id:
+            event = bulk_event_id_to_events.get(event_id)
+        group = group_id_to_group.get(int(group_id))
+        if not group or not event:
+            continue
```
wedamija, line 247: I think we should probably log something here if one of these isn't present - seems like an error, or a way to have a bunch of silent failures

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -200,38 +203,87 @@ def parse_rulegroup_to_event_data(
...
-    group_to_groupevent: dict[Group, GroupEvent] = {}
     groups = Group.objects.filter(id__in=group_ids)
     group_id_to_group = {group.id: group for group in groups}
-    for rule_group, instance_data in parsed_rulegroup_to_event_data.items():
+    event_ids: set[str] = set()
+    occurrence_ids: list[str] = []
+
+    for instance_data in parsed_rulegroup_to_event_data.values():
         event_id = instance_data.get("event_id")
+        if event_id:
```
wedamija, line 270: We should probably log if this isn't present, since I'd say this is considered a required field

wedamija (approved):
lgtm

PR #70546, opened 2024-05-08
ref(seer grouping): Add constants for Seer model version and URL

lobsterkatie:
This adds two constants, `SEER_SIMILARITY_MODEL_VERSION` and `SEER_SIMILAR_ISSUES_URL`, and uses them in places where we've previously been hardcoding values. it also adds model version to the `SeerSimilarIssueData` dataclass, defaulted to the constant's value.

### DIFF
--- a/src/sentry/conf/server.py
+++ b/src/sentry/conf/server.py
@@ -3950,6 +3950,9 @@ COGS_EVENT_STORE_LABEL = "bigtable_nodestore"
 # Disable DDM entirely
 SENTRY_DDM_DISABLE = os.getenv("SENTRY_DDM_DISABLE", "0") in ("1", "true", "True")
 
+SEER_SIMILARITY_MODEL_VERSION = "v0"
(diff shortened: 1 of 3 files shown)


src/sentry/conf/server.py
```diff
@@ -3950,6 +3950,9 @@ def custom_parameter_sort(parameter: dict) -> tuple[str, int]:
 # Disable DDM entirely
 SENTRY_DDM_DISABLE = os.getenv("SENTRY_DDM_DISABLE", "0") in ("1", "true", "True")
 
+SEER_SIMILARITY_MODEL_VERSION = "0"
```
jangjodi, line 3953: would this not need to be `v0`?

lobsterkatie: Yes. Damn - this is what I get for trying to juggle so many branches at once. I did fix that... just apparently not in this branch. 🤦🏻‍♀️

Good catch.

jangjodi (approved):

PR #69873, opened 2024-04-29
chore(actor) Remove AlertRule.owner from state

markstory:
- Remove AlertRule.owner from django state.
- Remove all usage of owner, and tighten the type checks for owners

Refs HC-1184

PR #69784, opened 2024-04-26
chore(actor) Remove Rule.owner from django state

markstory:
- Remove `Rule.owner` from django state.
- Remove remaining reads to `Rule.owner`
- Remove writes to `Rule.owner`
- Add two new indexes for the new owner columns.

Refs HC-1177

PR #70549, opened 2024-05-08
chore(actor) Remove old actor columns

markstory:
Remove now unused actor columns from rule, alertrule, team, grouphistory. These columns were removed from django state in #69784, #69873, #69976, #70090

Refs HC-1178


wedamija (approved):
🔥

PR #70552, opened 2024-05-08
fix(slack): prevent falling back to token verification if slack headers are not set

scefali:
This PR adds ensures that if `slack.signing-secret` is set, we always using secret verification instead of token verification which is less safe. As it is currently implemented, if you have both `slack.signing-secret` and `slack.verification-token` set, it would fall back to token verification if the request didn't have slack HTTP headers.


scefali:
This PR was wrong; the behavior was already correct before

PR #70553, opened 2024-05-08
feat(utils): Add dictionary filtering helper

lobsterkatie:
While working on an upcoming PR, I found myself wanting to be able to filter a dictionary by key. `filter` seemed promising, but, no, that only works on lists*. Set intersection? Nope, that'll only give you back the intersecting keys, not the full intersecting key-value pairs. And yeah, I could do it manually, but I figured if I were going to do that, I might as well wrap it in a function. 

So this adds an `apply_key_filter` helper, which takes a dictionary and either a list of keys to keep or a function to apply to each key, in order to decide whether to include the key and its corresponding value. (It's implemented both ways because I couldn't figure out which I liked better and ended up trying both.) 

(Maybe at some point [this guy](https://discuss.python.org/t/dictionaries-should-have-key-based-set-operations/28408) will get his way, and we won't have to do this anymore.)

While I was in the JSON utils test file, I also did a little cleanup work there (renamed some things and added missing tests for the already-existing and related `prune_empty_keys` helper).

*Yes, and tuples and sets and the like. But still, it's fundamentally about 1-dimensional objects, and I wanted it to work on a 2D(ish) one.

### DIFF
--- a/src/sentry/utils/json.py
+++ b/src/sentry/utils/json.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import datetime
 import decimal
 import uuid
-from collections.abc import Generator, Mapping
+from collections.abc import Callable, Collection, Generator, Mapping
 from enum import Enum
 from typing import IO, TYPE_CHECKING, Any, NoReturn, TypeVar, overload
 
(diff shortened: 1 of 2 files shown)


src/sentry/utils/json.py
```diff
@@ -177,6 +177,37 @@ def prune_empty_keys(obj: Mapping[TKey, TValue | None] | None) -> dict[TKey, TVa
...
+    If both are given, `keep_keys` takes precedence. If neither is given, an unfiltered shallow copy
+    of the original is returned.
+    """
+
+    if keep_keys:
+        key_filter = lambda key: key in keep_keys
+    elif not keep_keys and not key_filter:
+        key_filter = lambda _key: True
+
+    # `key_filter` can't be None by now, but mypy still thinks it might
```
JoshFerge, line 205: generally re-assignment confuses mypy, so if you didn't want to do this, you could re-write above to not re-assign, but not a big deal

JoshFerge (approved):
this looks good -- i would usually just inline my dict comprehension filters, i think this is also fine

lobsterkatie:
> i would usually just inline my dict comprehension filters, i think this is also fine

Yeah, it's certainly doable. IDK - maybe it's a little more readable with the helper? I'll grant the difference isn't huge if so. 

```python
some_dict = {"some": "dict", "to": "filter"}   
good_keys = ["some"] 
no_short_keys = lambda key: len(key) > 2

{key: some_dict[key] for key in some_dict if key in good_keys}
{key: some_dict[key] for key in some_dict if no_short_keys(key)}

# vs

apply_key_filter(some_dict, keep_keys=good_keys)
apply_key_filter(some_dict, key_filter=no_short_keys)
```

Mostly I'm lazy and somehow writing a helper, along with tests and a PR description, seemed like less work than writing out the comprehension? Sounds a little silly when I say it out loud...

PR #70558, opened 2024-05-08
chore(chartcuterie): Revert Style Changes

iamrajjoshi:
This reverts 1354478967d580699107222d42a95373b5a4dcaa and 173b69012db4737336284ebb4c75adbf08704e70

### DIFF
--- a/static/app/chartcuterie/performance.tsx
+++ b/static/app/chartcuterie/performance.tsx
@@ -1,4 +1,3 @@
-import type {LineChartProps} from 'sentry/components/charts/lineChart';
 import {transformToLineSeries} from 'sentry/components/charts/lineChart';
 import getBreakpointChartOptionsFromData, {
   type EventBreakpointChartData,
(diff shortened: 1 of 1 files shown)


ellisonmarks (approved):

sentaur-athena:
Do we know what caused the failure? Or just reverting because the timing matches

iamrajjoshi:
> Do we know what caused the failure? Or just reverting because the timing matches

We are trying to see if this helps

ellisonmarks:
> Do we know what caused the failure? Or just reverting because the timing matches

I believe the chartcuterie deployment is crashing because it's encountering `URLSearchParams` in https://sentry.io/_chartcuterie-config.js. Presumably that wasn't there before this change?

sentaur-athena:
@ellisonmarks oh I thought this PR fixed it. https://github.com/getsentry/chartcuterie/pull/149
Just checked the logs though and seems like the issue is still going on.

PR #70561, opened 2024-05-08
feat(assignee-selector): Create storybook component for new assignee selector trigger

MichaelSun48:
This PR creates a new component `<AssigneeBadge/>` which will eventually replace  the current assignee selector trigger in the issue stream. This PR does **not** make any changes to the selector, it merely creates the component and a corresponding storybook entry so it can be iterated on independent of any changes made to the issue stream. 

More details of the project can be found [here](https://github.com/getsentry/sentry/issues/69827) (#69827)

At a glance (as of 5/9):
image

Note that unlike the previous assignee selector, there is no longer a "Suggested Assignee" state for this new assignee selector, there is only an Assigned and Unassigned state. 

TODO:
- [x] Figure out what the loading state looks like
- [x] Chevron weight and style match border
- [x] Add tests

### DIFF
--- a/static/app/components/assigneeBadge.tsx
+++ b/static/app/components/assigneeBadge.tsx
@@ -1,5 +1,6 @@
 // import {Fragment} from 'react';
 import {Fragment} from 'react';
+import {css} from '@emotion/react';
 import styled from '@emotion/styled';
 
 import ActorAvatar from 'sentry/components/avatar/actorAvatar';
@@ -86,7 +87,14 @@ export function AssigneeBadge({
 
   return assignedTo ? (
     <Tooltip title={makeAssignedTooltipText(assignedTo)}>
-      <StyledTag icon={makeAssignedIcon(assignedTo)} />
+      <StyledTag
+        data-is-team={assignedTo.type === 'team'}
+        css={css`
+          --avatar-spacing: ${space(0.5)} ${space(0.25)} ${space(0.5)}
+            ${assignedTo.type === 'team' ? space(0.75) : space(0.5)};
+        `}
+        icon={makeAssignedIcon(assignedTo)}
+      />
     </Tooltip>
   ) : (
     <Tooltip title={makeUnAssignedTooltipText()}>
@@ -107,8 +115,9 @@ const StyledTag = styled(Tag)`
   }
   & > div {
     height: 24px;
-    padding: ${space(0.5)};
+    padding: var(--avatar-spacing, ${space(0.5)});
   }
+  cursor: pointer;
 `;
 
 const TooltipSubtext = styled('div')`
--- a/static/app/components/assigneeBadge.tsx
+++ b/static/app/components/assigneeBadge.tsx
@@ -89,6 +89,8 @@ export function AssigneeBadge({
     <Tooltip title={makeAssignedTooltipText(assignedTo)}>
       <StyledTag
         data-is-team={assignedTo.type === 'team'}
+        // Team avatars need extra left padding since the square avatar
+        // is being fit into a rounded edge
         css={css`
           --avatar-spacing: ${space(0.5)} ${space(0.25)} ${space(0.5)}
             ${assignedTo.type === 'team' ? space(0.75) : space(0.5)};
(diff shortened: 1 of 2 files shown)


static/app/components/assigneeBadge.tsx
```diff
@@ -0,0 +1,136 @@
...
+  return assignedTo ? (
+    <Tooltip title={makeAssignedTooltipText(assignedTo)}>
+      <StyledTag
+        data-is-team={assignedTo.type === 'team'}
+        // Team avatars need extra left padding since the square avatar
+        // is being fit into a rounded edge
+        css={css`
+          --avatar-spacing: ${space(0.5)} ${space(0.25)} ${space(0.5)}
+            ${assignedTo.type === 'team' ? space(0.75) : space(0.5)};
+        `}
```
MichaelSun48, line 97: image

This is what it looks like without the extra left padding. imho it looks very cramped and needs the extra spacing, but lmk if you don't think so

roggenkemper (commented):
does this still work for the suspect commit suggested assignee?

MichaelSun48:
> does this still work for the suspect commit suggested assignee?

@roggenkemper 
If the issue has a suggested assignee because of a suspect commit, that suggestion will still be surfaced in the dropdown itself. **However** since there is no longer a suggested assignee state for this new trigger, we will no longer surface that information on hover like in this example from the current UI: 

image

This is not 100% set in stone though, so we can discuss this further if you have strong opinions

roggenkemper:
maybe this is just me but i feel like the padding between the avatar and the name could be increased a bit

static/app/components/assigneeBadge.tsx
scttcper, line 97: i think instead of this, we could put the padding directly on the icon element in makeAssignedIcon

static/app/components/assigneeBadge.tsx
```diff
@@ -0,0 +1,136 @@
...
+        <TooltipSubtext>
+          {tct('You can auto-assign issues by adding [issueOwners:Issue Owner rules].', {
+            issueOwners: (
+              <TooltipSubExternalLink href="https://docs.sentry.io/product/error-monitoring/issue-owners/" />
+            ),
+          })}
+        </TooltipSubtext>
+      </TooltipWrapper>
+    );
+  };
```
scttcper, line 86: i think you can just use constants instead of functions in places which saves a little bit of overhead of creating new functions every render. If they're small enough and not reused, i'd just inline them where they're used
```suggestion
  const makeUnAssignedTooltipText =  (
      <TooltipWrapper>
        <div>{t('Unassigned')}</div>
        <TooltipSubtext>
          {tct('You can auto-assign issues by adding [issueOwners:Issue Owner rules].', {
            issueOwners: (
              <TooltipSubExternalLink href="https://docs.sentry.io/product/error-monitoring/issue-owners/" />
            ),
          })}
        </TooltipSubtext>
      </TooltipWrapper>
    );
```

static/app/components/assigneeBadge.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  const makeAssignedTooltipText = (actor: Actor) => {
+    // Cant use StyledTag's tooltipText prop because
+    // it screws with the nested div style ()
+    return (
+      <TooltipWrapper>
+        {tct('Assigned to [name]', {
+          name: actor.type === 'team' ? `#${actor.name}` : actor.name,
+        })}
```
scttcper, line 67: prefer t over tct if you're not using it to insert components
```suggestion
        {t('Assigned to %s', actor.type === 'team' ? `#${actor.name}` : actor.name)}
```

could even just do regular jsx
```suggestion
        {t('Assigned to')} {actor.type === 'team' ? `#${actor.name}` : actor.name}
```

MichaelSun48:
Updated designs: 
image

static/app/components/assigneeBadge.tsx
```diff
@@ -0,0 +1,127 @@
...
+const StyledTag = styled(Tag)`
+  span {
+    display: flex;
+    align-items: center;
+    gap: ${space(0.5)};
+  }
+  & > div {
+    height: 24px;
+    padding: ${space(0.5)};
+    padding-right: ${space(0.25)};
```
MichaelSun48, line 111: Override the right padding here since there's slightly too much space between the chevron and the end of the border:

Less padding (current):
image

Same padding: 
image

Idk maybe my eyes be crazy

scttcper (approved):

PR #70566, opened 2024-05-09
feat(crons): Add hook for crons landing panel

davidenwang:
Related to: https://github.com/getsentry/getsentry/pull/13908

### DIFF
--- a/static/app/types/hooks.tsx
+++ b/static/app/types/hooks.tsx
@@ -166,6 +166,8 @@ type GuideUpdateCallback = (nextGuide: Guide | null, opts: {dismissed?: boolean}
 
 type MonitorCreatedCallback = (organization: Organization) => void;
 
+type CronsOnboardingPanelProps = {children: React.ReactNode};
+
 type SentryLogoProps = SVGIconProps & {
   pride?: boolean;
 };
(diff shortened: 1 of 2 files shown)


static/app/types/hooks.tsx
```diff
@@ -185,6 +188,7 @@ export type ComponentHooks = {
   'component:codecov-integration-settings-link': () => React.ComponentType<CodecovLinkProps>;
   'component:confirm-account-close': () => React.ComponentType<AttemptCloseAttemptProps>;
   'component:crons-list-page-header': () => React.ComponentType<CronsBillingBannerProps>;
+  'component:crons-onboarding-panel': () => React.ComponentType<CronsOnboardingPanelProps>;
```
evanpurkhiser, line 191: Looks like we're using the world `monitor` down below. Oops 💀

static/app/types/hooks.tsx
```diff
@@ -1,3 +1,4 @@
+import type React from 'react';
```
evanpurkhiser, line 1: Do we need this for something?

evanpurkhiser (approved):

static/app/types/hooks.tsx
davidenwang: dangg ittt, i guess anytime we refer to actual 'monitors' we call them as such, but the product pages we end up calling 'crons'

PR #70081, opened 2024-05-01
feat(api-idorslug): Rename Path paramaters to `organization_id_or_slug`

iamrajjoshi:
I am working on making our APIs work with slug and id in our path parameters. Over the path month, we have updated the endpoints so they can support both parameters. As a final step, we are renaming the parameters from {resource}.slug to {resource}.id_or_slug.

A Breakdown of the changes:
https://github.com/getsentry/sentry/pull/70378
https://github.com/getsentry/sentry/pull/70384
https://github.com/getsentry/sentry/pull/70389

For more information, refer to: https://www.notion.so/sentry/Make-APIs-work-with-Slug-ID-cbaf68b883b843d499191e9cc4c116d3 and https://github.com/getsentry/sentry/pull/69716#issuecomment-2080446567

PR #70567, opened 2024-05-09
feat(api-idorslug): Updated Subset of Orgmember and Integration Endpoints to use `organization_id_or_slug`

iamrajjoshi:
A subset of changes from https://github.com/getsentry/sentry/pull/70081!


pyproject.toml
```diff
@@ -120,14 +120,12 @@ ignore_missing_imports = true
 module = [
     "sentry.api.base",
```
iamrajjoshi, line 121: File updated because I improved typing

src/sentry/api/bases/organization.py
```diff
@@ -242,30 +242,42 @@ class ControlSiloOrganizationEndpoint(Endpoint):
     def convert_args(
         self,
         request: Request,
-        organization_slug: str | int | None = None,
         *args: Any,
         **kwargs: Any,
     ) -> tuple[tuple[Any, ...], dict[str, Any]]:
+        organization_id_or_slug: int | str | None = None
```
iamrajjoshi, line 248: This is a temporary jerry rig so that we maintain backwards compatibility with getsentry endpoints and other endpoints which i haven't merged yet until I update them. They expect convert_args to accept organization_slug, so I can't keep it named, I have to keep it unnamed and extract it. I don't like it, but couldn't think of a better way.

src/sentry/api/bases/organization.py
```diff
@@ -242,30 +242,42 @@ class ControlSiloOrganizationEndpoint(Endpoint):
...
+        if args and args[0] is not None:
```
ykamo001, line 249: but you're allowing `organization_id_or_slug` to be None right, so does it make sense still to check for not none?

iamrajjoshi: hmm, i don't think that check does anything. if no function call will pass none to the method because the serializer will complain if nothing is passed as the `org_id_or_slug` to the endpoint.

ykamo001: > hmm, i don't think that check does anything. if no function call will pass none to the method because the serializer will complain if nothing is passed as the `org_id_or_slug` to the endpoint.

I don't think I follow sorry, could you rephrase or explain with example?

iamrajjoshi: What i meant is args[0] will never be None, so really that check doesn't do anything, Ill delete it.

src/sentry/api/bases/organization.py
```diff
@@ -242,30 +242,42 @@ class ControlSiloOrganizationEndpoint(Endpoint):
...
+            args = args[1:]
+        else:
+            organization_id_or_slug = kwargs.pop("organization_id_or_slug", None) or kwargs.pop(
+                "organization_slug", None
+            )
+
         if not subdomain_is_region(request):
             subdomain = getattr(request, "subdomain", None)
-            if subdomain is not None and subdomain != organization_slug:
+            if subdomain is not None and subdomain != organization_id_or_slug:
```
ykamo001, line 261: `organization_id_or_slug` can be `None, so are we okay with this evaluating into:
```
subdomain != None:
```
which will always be true because of the first check at `subdomain is not None` 😆

iamrajjoshi: yeah the types are a mess here, but if `organization_id_or_slug` is None, we will raise ResourceDoesNotExist because we do this check right under here. So this logic here is generally only for when its not None. I should probably move the conditional above this.

src/sentry/api/bases/organization.py
```diff
@@ -536,32 +548,42 @@ def get_filter_params(
     def convert_args(
         self,
         request: Request,
-        organization_slug: str | int | None = None,
         *args: Any,
         **kwargs: Any,
     ) -> tuple[tuple[Any, ...], dict[str, Any]]:
         """
-        We temporarily allow the organization_slug to be an integer as it actually can be both slug or id
+        We temporarily allow the organization_id_or_slug to be an integer as it actually can be both slug or id
```
sentaur-athena, line 555: nit: I don't think this comment is valid anymore.

src/sentry/api/endpoints/codeowners/external_actor/user_details.py
```diff
@@ -27,7 +27,7 @@ class ExternalUserDetailsEndpoint(OrganizationEndpoint, ExternalActorEndpointMix
     }
     owner = ApiOwner.ENTERPRISE
 
-    def convert_args(  # type: ignore[override]
```
sentaur-athena, line 30: I don't see any type fixes here :/ how come you could remove the comment? 😅

sentaur-athena (approved):

src/sentry/api/endpoints/codeowners/external_actor/user_details.py
iamrajjoshi: The fix was upstream. Before, we use to two a mismatched function signature with the superclass

ykamo001 (approved):

PR #70572, opened 2024-05-09
perf: use orjson in metrics

anonrig:


asottile-sentry (approved):

nikhars (approved):

iambriccardo (approved):
Changes on our code LGTM, thanks for doing this!

anonrig:
> Changes on our code LGTM, thanks for doing this!

Does any of these paths generate a JSON object with integer as the key?

loewenheim (approved):

PR #70573, opened 2024-05-09
fix(ui): Add missing 'return' statement in <SourceMapsDebuggerModal />

floels:
# Goal

The goal of this PR is to add a missing early `return` statement in `static/app/components/ events/interfaces/sourceMapsDebuggerModal.tsx`.

# Approach

We add the missing `return` statement in the component's file, and also add a test file (`sourceMapsDebuggerModal.spec.tsx`) and a first test to cover this behavior of the component.

You can see that this test fails if you remove the fix in `sourceMapsDebuggerModal.tsx`.

According to Istanbul, this increases the statement coverage of the `sourceMapsDebuggerModal.tsx` file from `13%`:

Screenshot 2024-05-09 at 15 15 51

to `39%`:

Screenshot 2024-05-09 at 15 20 08

which shows we need to add more test cases to this test file.

# Legal Boilerplate


lforst (approved):

lforst:
Nice thanks!

floels:
Thanks! I think there are some checks that I need to be able to merge and that I cannot trigger myself as an external contributor. @lforst is that correct? If so, could you trigger them on your end? 🙏

lforst:
Sorry, I was at react conf last week so this kinda slipped. I'll make sure this gets merged. Thanks for the bump!

PR #70574, opened 2024-05-09
perf: use orjson in endpoints

anonrig:

### DIFF
--- a/src/sentry/api/endpoints/event_ai_suggested_fix.py
+++ b/src/sentry/api/endpoints/event_ai_suggested_fix.py
@@ -4,6 +4,7 @@ import logging
 import random
 from typing import Any
 
+import orjson
 from django.conf import settings
 from django.dispatch import Signal
 from django.http import HttpResponse, StreamingHttpResponse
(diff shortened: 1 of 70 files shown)


src/sentry/api/endpoints/event_ai_suggested_fix.py
```diff
@@ -275,10 +275,7 @@ def suggest_fix(event_data, model=settings.SENTRY_AI_SUGGESTED_FIX_MODEL, stream
         temperature=0.7,
         messages=[
             {"role": "system", "content": prompt},
-            {
-                "role": "user",
-                "content": json.dumps(event_info),
-            },
+            dict(role="user", content=orjson.dumps(event_info).decode()),
```
asottile-sentry, line 278: technically the dict literal is faster and should be preferred

asottile-sentry (approved):

src/sentry/api/endpoints/event_ai_suggested_fix.py
anonrig: Yes, you're right. Jetbrains got me...

PR #70576, opened 2024-05-09
ref(replay): update details tab title to be display name

### LINKED ISSUE #70275: Change <title> for individual Replays to not be the Replay ID, but rather what is displayed on Replay Details (email, IP, etc)
Change this from Replay ID to what is being used as the replay identifier in product on the Index/Details pages.

Image

michellewzhang:
fixes https://github.com/getsentry/sentry/issues/70275

replaces the tab title (previously the replay ID) to be the display name instead

https://github.com/getsentry/sentry/assets/56095982/92edbe3e-fdd2-4d73-b6c6-d2ddeb7e561b

### DIFF
--- a/static/app/views/replays/detail/page.tsx
+++ b/static/app/views/replays/detail/page.tsx
@@ -40,7 +40,7 @@ export default function Page({
   isVideoReplay,
 }: Props) {
   const title = replayRecord
-    ? `${replayRecord.user.display_name} — Session Replay — ${orgSlug}`
+    ? `${replayRecord.user.display_name ?? t('Anonymous User')} — Session Replay — ${orgSlug}`
     : `Session Replay — ${orgSlug}`;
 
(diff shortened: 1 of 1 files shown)


static/app/views/replays/detail/page.tsx
```diff
@@ -40,7 +40,7 @@ export default function Page({
   isVideoReplay,
 }: Props) {
   const title = replayRecord
-    ? `${replayRecord.id} — Session Replay — ${orgSlug}`
+    ? `${replayRecord.user.display_name} — Session Replay — ${orgSlug}`
```
c298lee, line 43: just wanted to confirm that user.display_name always exists?

c298lee (approved):

static/app/views/replays/detail/page.tsx
michellewzhang: good point - i added a fallback to Anonymous User

PR #70581, opened 2024-05-09
chore(actor) Improve model interactions with Actor

markstory:
Consolidate the team/user field management logic into the Model classes as much as possible. There is a leak to this abstraction when Model.update() is used. The leak stems from how `update()` works under the hood as it uses QuerySet.update() to only updated the provided attributes instead of all dirty attributes.


src/sentry/types/actor.py
```diff
@@ -259,6 +259,18 @@ def is_user(self) -> bool:
         return self.actor_type == ActorType.USER
 
 
+class ActorOwned(Protocol):
```
wedamija, line 262: How do we apply this protocol to the classes?

wedamija (commented):
This looks much nicer, thanks!

src/sentry/types/actor.py
markstory: Protocols don't have to be applied to classes. They can be used as type annotation in methods/returns and match with [structural subtyping](https://typing.readthedocs.io/en/latest/spec/protocol.html#protocols). There aren't any references to the protocol yet, but I included it in case we need it, and to document the interactions that `Actor` can have with ORM models.

wedamija (approved):

PR #70582, opened 2024-05-09
ref(ui): Consistently use Layout.{Body,Main}

evanpurkhiser:
Without these the feedback views padding was slightly different from all
the other top-level views which consistelty use Body and Main

### DIFF
--- a/static/app/views/feedback/feedbackListPage.tsx
+++ b/static/app/views/feedback/feedbackListPage.tsx
@@ -72,31 +72,35 @@ export default function FeedbackListPage({}: Props) {
             </Layout.HeaderContent>
           </Layout.Header>
           <PageFiltersContainer>
-            <ErrorBoundary>
-              <LayoutGrid data-banner={showWhatsNewBanner}>
-                {showWidgetBanner ? (
-                  <FeedbackWidgetBanner style={{gridArea: 'banner'}} />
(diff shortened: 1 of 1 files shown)


static/app/views/feedback/feedbackListPage.tsx
```diff
@@ -105,9 +109,6 @@ export default function FeedbackListPage({}: Props) {
 }
 
 const LayoutGrid = styled('div')`
-  background: ${p => p.theme.background};
-  overflow: hidden;
```
evanpurkhiser, line 109: afaict this is not needed and causes the focus rings to be cut off on the page filters

ryan953: focus rings arn't cutoff anywhere on /feedback

/replay does have that problem though, hard to fix because of some layers that the Trace tab needs :(

evanpurkhiser:
Your screenshot didn't finish uploading @Jesse-Box

Jesse-Box:
> Your screenshot didn't finish uploading @Jesse-Box

Updated!

evanpurkhiser:
Oh, yeah I'm aware of these problems. That’s why I didn’t merge yet haha

evanpurkhiser:
Will revisit this later

PR #70584, opened 2024-05-09
perf: use orjson in dynamic sampling

anonrig:
Replaces `json` calls with `orjson` and includes a small optimization to avoid sorting of keys.

### DIFF
--- a/src/sentry/dynamic_sampling/rules/utils.py
+++ b/src/sentry/dynamic_sampling/rules/utils.py
@@ -1,11 +1,12 @@
 from enum import Enum
-from typing import Any, Literal, NotRequired, TypedDict, Union
+from typing import Literal, NotRequired, TypedDict, Union
 
+import orjson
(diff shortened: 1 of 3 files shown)


src/sentry/dynamic_sampling/rules/utils.py
```diff
@@ -192,15 +193,18 @@ def get_rule_type(rule: Rule) -> RuleType | None:
...
                 "id": rule["id"],
                 "type": rule["type"],
                 "condition": rule["condition"],
-            }
+            },
+            option=orjson.OPT_SORT_KEYS,
         )
-    ).__hash__()
+        .decode()
+        .__hash__()
```
asottile-sentry, line 206: this should be replaced with a call to `hash(...)` -- calling the double-under directly is usually not recommended

asottile-sentry (approved):
fixnship

PR #70585, opened 2024-05-09
fix(crons): Fix connection spikes during partition assignment

wedamija:
`on_partitions_assigned` is called whenever partitions are assigned to a consumer. This calls `_create_strategy`, which calls `create_with_partitions`.

Since we create the `ThreadPoolExecutor` in `create_parallel_worker`, this means we create a new threadpool whenever we partition assignment changes. The reason this causes spikes and isn't sustained is that presumably the previous `ProcessingStrategy` stops being used and ends up garbage collected, which results in the `ThreadPoolExecutor` being garbage collected too.

### DIFF
--- a/src/sentry/monitors/consumers/monitor_consumer.py
+++ b/src/sentry/monitors/consumers/monitor_consumer.py
@@ -1028,6 +1028,7 @@ class StoreMonitorCheckInStrategyFactory(ProcessingStrategyFactory[KafkaPayload]
     ) -> None:
         if mode == "parallel":
             self.parallel = True
+            self.parallel_executor = ThreadPoolExecutor(max_workers=self.max_workers)
 
         if max_batch_size is not None:
             self.max_batch_size = max_batch_size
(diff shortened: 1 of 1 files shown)


src/sentry/monitors/consumers/monitor_consumer.py
```diff
@@ -1041,8 +1042,6 @@ def shutdown(self) -> None:
             self.parallel_executor.shutdown()
 
     def create_parallel_worker(self, commit: Commit) -> ProcessingStrategy[KafkaPayload]:
-        self.parallel_executor = ThreadPoolExecutor(max_workers=self.max_workers)
```
evanpurkhiser, line 1044: This breaks typing, we need to define `self.parallel_executor` in the class as 

```py
parallel_executor: Executor | None
```

and here we need to do

```py
assert self.parallel_executor
```

evanpurkhiser (approved):

PR #70594, opened 2024-05-09
fix(metrics): Remove byte format within JSON

nikhars:
We need the bytes of metric type only in the kafka headers. Not in the actual payload. Made the change so that the JSON payload gets a string while the kafka header gets a byte.

### DIFF
--- a/tests/sentry/sentry_metrics/test_gen_metrics_multiprocess_steps.py
+++ b/tests/sentry/sentry_metrics/test_gen_metrics_multiprocess_steps.py
@@ -66,6 +66,7 @@ def compare_messages_ignoring_mapping_metadata(actual: Message, expected: Messag
     actual_deserialized = json.loads(actual_payload.value)
     expected_deserialized = json.loads(expected_payload.value)
     del actual_deserialized["mapping_meta"]
+    actual_deserialized.pop("aggregation_option", None)
     assert actual_deserialized == expected_deserialized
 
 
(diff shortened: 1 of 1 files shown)


tests/sentry/sentry_metrics/test_gen_metrics_multiprocess_steps.py
```diff
@@ -66,6 +66,7 @@ def compare_messages_ignoring_mapping_metadata(actual: Message, expected: Messag
     actual_deserialized = json.loads(actual_payload.value)
     expected_deserialized = json.loads(expected_payload.value)
     del actual_deserialized["mapping_meta"]
+    actual_deserialized.pop("aggregation_option", None)
```
nikhars, line 69: In local testing, the metrics payload adds the aggregation option to the transformed payload. Others don't. Since the tests are generic over different payload types, removed the checking of field which is specific to a payload.

anonrig: Can you add a comment to here?

tests/sentry/sentry_metrics/test_gen_metrics_multiprocess_steps.py
```diff
@@ -358,7 +359,7 @@ def test_process_messages() -> None:
                         None,
                         json.dumps(__translated_payload(message_payloads[i])).encode("utf-8"),
                         [
-                            ("metric_type", message_payloads[i]["type"]),
+                            ("metric_type", str.encode(message_payloads[i]["type"])),
```
anonrig, line 362:
```suggestion
                            ("metric_type", message_payloads[i]["type"].encode()),
```

anonrig (approved):

PR #70596, opened 2024-05-09
Add limit to API for Alert Rules

### LINKED ISSUE #67195: Alert rules quantity limits monitoring
### Problem Statement

As I've got information from support - Sentry has limit for 1000 alert rules for 1 organization. 

Due to this limit our team would like to know the way of monitoring quantity of alerts for 1 organization slug. Performing such monitoring makes possible to know exact time when number of alerts reaching limit.

### Solution Brainstorm

Add alerts quantity to any API endpoint that will allow parse exact field and create alert rules quantity monitoring. A little similar example described [here](https://docs.sentry.io/api/alerts/create-an-issue-alert-rule-for-a-project/), but there are no alert rules quantity at all.

### Product Area

Alerts

saponifi3d:
## Description
Fixes: https://github.com/getsentry/sentry/issues/67195

Adds the `x-sentry-rule-limit` header to the list views for alert rules (both the combined-rules and metric alert api).

I tried adding an indicator to the UI, but it was pretty difficult to tell which page we're on in the UI, since we're handling it on the backend with a `Link`. Rather than blocking the API fix on trying to figure out the UI, i figure it's better to get this out and we can add to the UI later.

Screenshot 2024-05-09 at 11 39 43 AM

### DIFF
--- a/src/sentry/api/helpers/constants.py
+++ b/src/sentry/api/helpers/constants.py
@@ -1,2 +1,2 @@
-MAX_QUERY_SUBSCTIPTIONS_HEADER = "X-Sentry-Rule-Limit"
+MAX_QUERY_SUBSCRIPTIONS_HEADER = "X-Sentry-Alert-Rule-Limit"
 ALERT_RULES_COUNT_HEADER = "X-Sentry-Alert-Rule-Hits"
--- a/src/sentry/api/helpers/constants.py
+++ b/src/sentry/api/helpers/constants.py
@@ -1,0 +1,2 @@
+MAX_QUERY_SUBSCTIPTIONS_HEADER = "X-Sentry-Rule-Limit"
+ALERT_RULES_COUNT_HEADER = "X-Sentry-Alert-Rule-Hits"
--- a/src/sentry/incidents/endpoints/organization_alert_rule_index.py
+++ b/src/sentry/incidents/endpoints/organization_alert_rule_index.py
@@ -1,6 +1,7 @@
 from copy import deepcopy
 from datetime import UTC, datetime
 
+from django.conf import settings
 from django.db.models import DateTimeField, IntegerField, OuterRef, Q, Subquery, Value
 from django.db.models.functions import Coalesce
 from drf_spectacular.utils import extend_schema, extend_schema_serializer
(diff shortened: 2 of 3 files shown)


src/sentry/api/helpers/constants.py
```diff
@@ -0,0 +1,2 @@
+MAX_QUERY_SUBSCTIPTIONS_HEADER = "X-Sentry-Rule-Limit"
```
ceorourke, line 1: `SUBSCTIPTIONS` typo 😛 also should this be `X-Sentry-Alert-Rule-Limit`?

src/sentry/api/helpers/constants.py
```diff
@@ -0,0 +1,2 @@
...
+ALERT_RULES_COUNT_HEADER = "X-Sentry-Alert-Rule-Hits"
```
ceorourke, line 2: Maybe this should be "remaining" instead of "hits" to match the API rate limits: https://docs.sentry.io/api/ratelimits/#headers

ceorourke: Eh I could see it either way, up to you

src/sentry/api/helpers/constants.py
saponifi3d: at least i was consistent in my typo. 🤣 - ty - updating!

for `X-Sentry-Alert-Rule-Limit` do you know if the snuba subscription limit only impacts metric alerts? (i believe so, in which case agree that this is a better name)

src/sentry/api/helpers/constants.py
saponifi3d: In this case, I don't think we should rename this because it already existed on the `combined-rules` api. i just also wanted to move this so it's consistent between alert-rules and combined-rules.

src/sentry/api/helpers/constants.py
ceorourke: Yeah, if you look [here](https://github.com/getsentry/sentry/blob/master/src/sentry/incidents/serializers/alert_rule.py#L470-L482) it's checking the `QuerySubscription`s which are only used for metric alerts.

ceorourke (approved):

PR #70600, opened 2024-05-09
feat(new-trace): Fixing routing to discover from trace view bug.

Abdkhan14:
When we have a timestamp in the url, we use `start: timestamp - 1.5days` and `end: timestamp + 1.5days` instead of statsPeriod, when routing to discover to review all events associated with trace id.


armenzg (approved):
Beautiful!

PR #70605, opened 2024-05-09
feat(new-trace): Using codesnippet and stackminitrace from tracedrawer.

Abdkhan14:
We already have all the data needed by `CodeSnippet `and `StackTraceMiniFrame` at the trace level to render DB query description, so using them seems wiser than higher level components like `DBQueryDescription` that re-fetch the data. 
Screenshot 2024-05-09 at 3 31 02 PM


gggritso (approved):
Rad!

PR #70607, opened 2024-05-09
feat(snuba-search): adds handling for additional filters for useGroupSnubaDataset=1

scefali:
This PR adds the capability to filter on the following new attributes:
* regressed_in_release
* release.build
* release.package
* release.dist

All of them require Postgres queries but handle it differently. For `regressed_in_release`, use the query to generate a list of IDs where we have a group history for regression that matches the versions we check for. For the other three, we get the release versions from Postgres and use that in a snuba query checking events of particular release versions.

Note I am leveraging `convert_search_filter_to_snuba_query` to generate the filters which requires some name mapping that I've added. And it adds logic to skip checking `search_issues` for columns that don't appear in it.

Also added `for_review` which is used in the standard way of generating group_ids. And also added support for `times_seen`.

### DIFF
--- a/src/sentry/issues/endpoints/organization_group_index.py
+++ b/src/sentry/issues/endpoints/organization_group_index.py
@@ -51,12 +51,8 @@ allowed_inbox_search_terms = frozenset(["date", "status", "for_review", "assigne
 # these filters are currently not supported in the snuba only search
 # and will use PostgresSnubaQueryExecutor instead of GroupAttributesPostgresSnubaQueryExecutor
 UNSUPPORTED_SNUBA_FILTERS = [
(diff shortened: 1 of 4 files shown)


wedamija (commented):
Does this support more complicated operations on these fields? I feel like we support wildcard searching etc on them too

scefali:
@wedamija It does, i can add tests for it if you think it's worth it

wedamija:
> @wedamija It does, i can add tests for it if you think it's worth it

As long as that functionality is already tested I think it's ok. If the tests are part of the old backend, and we plan to remove that backend, then we should port them over

wedamija (approved):

PR #70608, opened 2024-05-09
perf: remove jsondata usages

anonrig:
By default JSONData is defined as Any. With `orjson`, we don't need `utils/json.py` and can slowly remove it.

Reduces cyclic dependencies.

### DIFF
--- a/src/sentry/api/bases/doc_integrations.py
+++ b/src/sentry/api/bases/doc_integrations.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from typing import Any
+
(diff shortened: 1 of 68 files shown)


asottile-sentry (approved):
~technically would be better as `object` and require a schema to unpack it to a known type -- but for now this is an improvement over the constant which makes it look like it's typed but actually is completely unchecked

anonrig:
> ~technically would be better as object and require a schema to unpack it to a known typ

@asottile-sentry That was my intention as well, but unfortunately it requires **a lot** of work.

PR #70609, opened 2024-05-09
fix: accept non-str keys on autofix

anonrig:
Fixes https://sentry.sentry.io/issues/5326989067/?project=1&referrer=github-pr-bot


anonrig:
@armenzg It seems you reproduced this problem. According to JSON specification, we shouldn't accept key's that are not string, but for the sake of fixing this issue, I added the fix.

armenzg:
Bull's eye 🎯 😄

asottile-sentry (approved):
what's weird is it looks like all the keys _are_ strings -- unless there's something funky in `serialized_event` that I'm not seeing in the sentry error

jennmueng (approved):
yeah second @asottile-sentry, looks like all the immediate keys are strings, probably something with the serialized event?

PR #70616, opened 2024-05-09
perf(assignees): Reduce render time of assignee dropdown

MichaelSun48:
This PR gets rid of a usage of dynamic props in styled components in <baseBadge/> since this is [bad for performance](https://develop.sentry.dev/frontend/using-styled-components/#use-style--css-attributes). It also gets rid of any usages of replaces any usage of idBadge in <assigneeSelectorDropdown/> with userBadge or teamBadge, since idBadge might re-fetch the avatar when we already have it. 

I'm seeing ~~20-30~~ 40-50% faster initial render speeds on local dev.

### DIFF
--- a/static/app/components/assigneeSelectorDropdown.tsx
+++ b/static/app/components/assigneeSelectorDropdown.tsx
@@ -492,6 +492,7 @@ export default function AssigneeSelectorDropdown({
       <CompactSelect
         searchable
         clearable
+        menuWidth={275}
         disallowEmptySelection={false}
         onClick={e => e.stopPropagation()}
         value={
--- a/static/app/components/assigneeSelectorDropdown.tsx
+++ b/static/app/components/assigneeSelectorDropdown.tsx
@@ -314,9 +314,7 @@ export default function AssigneeSelectorDropdown({
           data-test-id="assignee-option"
           displayName={`${userDisplay}${isCurrentUser ? ' (You)' : ''}`}
           hideEmail
-          user={{
-            ...user,
-          }}
+          user={user}
         />
       ),
       // Jank way to pass assignee type (team or user) into each row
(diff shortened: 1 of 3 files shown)


static/app/components/idBadge/baseBadge.tsx
```diff
@@ -43,7 +44,12 @@ export const BaseBadge = memo(
     const wrapperGap: ValidSize = avatarSize <= 14 ? 0.5 : avatarSize <= 20 ? 0.75 : 1;
 
     return (
-      <Wrapper className={className} gap={wrapperGap}>
+      <Wrapper
+        className={className}
+        css={css`
+          --wrapper-gap: ${space(wrapperGap)};
+        `}
```
scttcper, line 51: this does the same thing

scttcper: by that i mean it will still insert a new style into the document, not really saving much

scttcper: inline styles would though? style={{gap: space(wrapperGap)}} since those are not via emotion

static/app/components/assigneeSelectorDropdown.tsx
```diff
@@ -423,10 +417,7 @@ export default function AssigneeSelectorDropdown({
     const memberOptions = {
       value: '_members',
       label: t('Members'),
-      options:
-        memList?.map(member =>
-          makeMemberOption(member.id, member.name || member.email)
-        ) ?? [],
+      options: memList?.map(member => makeMemberOption(member)) ?? [],
```
scttcper, line 420: memList should always be defined
```suggestion
      options: memList.map(member => makeMemberOption(member)),
```

static/app/components/assigneeSelectorDropdown.tsx
```diff
@@ -393,9 +389,7 @@ export default function AssigneeSelectorDropdown({
       } else {
         assignedUser = memList?.find(user => user.id === group.assignedTo?.id);
```
scttcper, line 390:
```suggestion
        assignedUser = memList.find(user => user.id === group.assignedTo?.id);
```

static/app/components/idBadge/baseBadge.tsx
MichaelSun48: That makes sense, will change to inline styles.

I was going off [this section](https://develop.sentry.dev/frontend/using-styled-components/#use-style--css-attributes) in the docs. We should maybe update it to say that styles are better than css insertions if that is the case.

scttcper (approved):

PR #70620, opened 2024-05-09
feat(slack): Add flags to toggle features on organization integration

### LINKED ISSUE #70316: Slack alerts in threads: add option to enable or disable it
### Problem Statement

### Solution Brainstorm

### Product Area

Settings - Integrations

ykamo001:
Features that we add to integrations should have the ability to be turned off by users. Use the Slack Threads feature as a stepping stone for this pattern to add feature toggle to organization integrations, allowing organizations to turn features on or off.

The features are turned on by default as we are about to GA the feature.

### DIFF
--- a/src/sentry/integrations/slack/integration.py
+++ b/src/sentry/integrations/slack/integration.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
+import logging
 from collections import namedtuple
-from collections.abc import Mapping, Sequence
+from collections.abc import Mapping, MutableMapping, Sequence
 from typing import Any
 
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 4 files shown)


src/sentry/integrations/slack/integration.py
```diff
@@ -81,6 +92,61 @@ def get_config_data(self) -> Mapping[str, str]:
...
         return {"installationType": metadata_.get("installation_type", default_installation)}
 
+    def get_organization_config(self) -> Sequence[tuple[str, bool]]:
+        """
+        Not sure why the base class is restricted to a sequence type, doesn't really make sense, however, it is
+        sufficient to our needs for now, and we can utilize it to return toggleable flags/feature on the integration
+        """
+
+        # Specifically using the parent method because the overwritten method on current class is hacked for another
+        # purpose at the integration/provider wide level, which is wrong/incorrect
```
sentaur-athena, line 102: The current class is doing something at installation level not provider wide level. For example for one org and slack integration you can have one workspace installation with `classic_bot` and one with `workspace_app`. No? The value cannot change per installation like your toggle does but it seems installation level.

src/sentry/integrations/slack/integration.py
```diff
@@ -81,6 +92,61 @@ def get_config_data(self) -> Mapping[str, str]:
...
+        base_data = super().get_config_data()
```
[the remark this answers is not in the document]
ykamo001: but there is no update config data flow, so it's technically already the wrong pattern. I think we're both on the same page here, the comments just spanned out over different lines haha

I'll schedule a quick sync

src/sentry/integrations/slack/integration.py
ykamo001: followed the pattern in above comment ^

src/sentry/integrations/slack/integration.py
```diff
@@ -67,18 +68,77 @@
...
+        # Add installationType key to config data
         metadata_ = self.model.metadata
         # Classic bots had a user_access_token in the metadata.
         default_installation = (
             "classic_bot" if "user_access_token" in metadata_ else "workspace_app"
         )
-        return {"installationType": metadata_.get("installation_type", default_installation)}
+        base_data["installationType"] = metadata_.get("installation_type", default_installation)
+
+        # Add missing toggleable feature flags
```
cathteng, line 97: is this so we don't have to do a migration every time we add a flag?

ykamo001: yup! Realistically we should pull the features into it's own table and just link it to the integration installation. This way we you add a new flag the default is already there

cathteng (approved):

PR #70624, opened 2024-05-09
update: remove timebox from query_extra

nhsiehgit:
PR to remove the time restrictions from the activated snuba subscription queries

See: https://github.com/getsentry/sentry/pull/70486

Initially we had wanted to introduce a timebox to our subscriptions to ensure we do not receive any events outside of our timebox.
snuba subscriptions do _not_ support timestamp restrictions though as this could cause zombied queries that just eat memory.

Since we're already restricting on the Release versions and we do not have any use cases currently where this would become an issue, we're rolling forwards with simply removing the timeboxed restrictions.

IF other activation triggers emerge that might be problematic with this set up, then we can reach out to the SNS team to update their apis to enable timestamps for subscriptions

### DIFF
--- a/src/sentry/snuba/models.py
+++ b/src/sentry/snuba/models.py
@@ -117,6 +117,7 @@ class QuerySubscription(Model):
     query_extra = models.TextField(
         null=True
     )  # additional query filters to attach to the query created in Snuba such as datetime filters, or release/deploy tags
+    # TODO: timebox is not utilized. Subscription queries do not support timestamp restrictions
     # timebox_start/end is optional timebox restrictions to apply to the snuba query
     timebox_start = models.DateTimeField(null=True)
     timebox_end = models.DateTimeField(null=True)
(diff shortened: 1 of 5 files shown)


src/sentry/snuba/models.py
```diff
@@ -117,6 +117,7 @@ class Status(Enum):
     query_extra = models.TextField(
         null=True
     )  # additional query filters to attach to the query created in Snuba such as datetime filters, or release/deploy tags
+    # TODO: timebox is not utilized. Subscription queries do not support timestamp restrictions
     # timebox_start/end is optional timebox restrictions to apply to the snuba query
     timebox_start = models.DateTimeField(null=True)
     timebox_end = models.DateTimeField(null=True)
```
wedamija, line 123: Is the plan to delete these?

wedamija (approved):

saponifi3d (approved):

PR #70627, opened 2024-05-09
fix(platform): Using the correct name

sentaur-athena:
sdk developers recommended to use `OS` in the name.

### DIFF
--- a/static/app/components/events/contextSummary/contextIcon.tsx
+++ b/static/app/components/events/contextSummary/contextIcon.tsx
@@ -61,7 +61,7 @@ const LOGO_MAPPING = {
   'net-core': logoNetcore,
   'net-framework': logoNetframework,
   'qq-browser': logoQq,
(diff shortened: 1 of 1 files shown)


iamrajjoshi (approved):

cathteng (approved):

static/app/components/events/contextSummary/contextIcon.tsx
```diff
@@ -61,7 +61,7 @@ const LOGO_MAPPING = {
   'net-core': logoNetcore,
   'net-framework': logoNetframework,
   'qq-browser': logoQq,
-  nintendo: logoNintendo,
+  'nintendo-os': logoNintendo,
```
sentaur-athena, line 64: The way this works is that if sdk passes `Nintendo OS` we convert that to `nintendo-os` and then look for an image that is named that.

PR #70629, opened 2024-05-09
ref(rules): Remove inner lock and add logging on exception

ceorourke:
Part of debugging the delayed rule processor lead here, we don't need the inner lock (it's locked [here](https://github.com/getsentry/sentry/blob/370e78ca10dddc2b018d1e9b6e41e501386cc0d2/src/sentry/buffer/redis.py#L316-L318) before `process_batch` is called).

### DIFF
--- a/src/sentry/buffer/redis.py
+++ b/src/sentry/buffer/redis.py
@@ -315,7 +315,7 @@ class RedisBuffer(Buffer):
         try:
             redis_buffer_registry.callback(BufferHookEvent.FLUSH, self)
         except Exception as e:
-            logger.info("process_batch.error", extra={"exception": e})
+            logger.exception("process_batch.error", extra={"exception": e})
 
     def incr(
         self,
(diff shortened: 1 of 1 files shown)


src/sentry/buffer/redis.py
```diff
@@ -312,15 +312,10 @@ def get_hash(
...
-        lock_key = self._lock_key(client, self.pending_key, ex=10)
-        if not lock_key:
-            return
-
         try:
             redis_buffer_registry.callback(BufferHookEvent.FLUSH, self)
-        finally:
-            client.delete(lock_key)
+        except Exception as e:
+            logger.info("process_batch.error", extra={"exception": e})
```
wedamija, line 318: Tbh I'd probably log an error here if this is failing

wedamija (approved):

src/sentry/buffer/redis.py
```diff
@@ -312,15 +312,10 @@ def get_hash(
...
-        lock_key = self._lock_key(client, self.pending_key, ex=10)
-        if not lock_key:
-            return
-
         try:
             redis_buffer_registry.callback(BufferHookEvent.FLUSH, self)
-        finally:
-            client.delete(lock_key)
+        except Exception as e:
+            logger.exception("process_batch.error", extra={"exception": e})
```
wedamija, line 318: Nit: You don't need the `extra={"exception": e}` here, `.exception` captures all the info about the exception for you

PR #70630, opened 2024-05-09
update: Querysubscription timebox delete column db operation

nhsiehgit:
requires https://github.com/getsentry/sentry/pull/70628


saponifi3d (approved):
:shipit:

PR #70634, opened 2024-05-10
fix(slack): correct suspect commit

cathteng:
The suspect commit is stored via the GroupOwner model. We have been fetching the suspect commit using the old logic (the commit responsible for the first line in the stacktrace) but this is incorrect.

NOTE: This probably won't be the final state, as we are not guaranteed the suspect commit will exist at the time the Slack message is sent because suspect commits processing is async from the alert rule firing.

### DIFF
--- a/src/sentry/integrations/slack/message_builder/issues.py
+++ b/src/sentry/integrations/slack/message_builder/issues.py
@@ -328,8 +328,6 @@ def get_suspect_commit_text(group: Group) -> str | None:
     pull_request = PullRequest.objects.filter(
         merge_commit_sha=commit.key, organization_id=group.project.organization_id
     ).first()
-    if not pull_request:
-        return None
 
     author = commit.author
     commit_id = commit.key
--- a/src/sentry/integrations/slack/message_builder/issues.py
+++ b/src/sentry/integrations/slack/message_builder/issues.py
@@ -43,11 +43,12 @@ from sentry.models.commit import Commit
 from sentry.models.group import Group, GroupStatus
 from sentry.models.project import Project
 from sentry.models.projectownership import ProjectOwnership
+from sentry.models.pullrequest import PullRequest
 from sentry.models.release import Release
+from sentry.models.repository import Repository
 from sentry.models.rule import Rule
 from sentry.models.team import Team
 from sentry.notifications.notifications.base import ProjectNotification
-from sentry.notifications.utils import get_commits
 from sentry.notifications.utils.actions import MessageAction
 from sentry.notifications.utils.participants import (
     dedupe_suggested_assignees,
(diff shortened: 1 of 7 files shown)


src/sentry/integrations/slack/message_builder/issues.py
```diff
@@ -667,10 +663,9 @@ def build(self, notification_uuid: str | None = None) -> SlackBlock:
             blocks.append(self.get_suggested_assignees_block(suggested_assignees))
 
         # add suspect commit info
-        if event_for_tags:
```
isabellaenriquez, line 670: is there a reason this was taken out?

isabellaenriquez (commented):
would it be worth using the same information you grab here for suggested assignees? we use some method there to get suspect commit users but i'm wondering if this could cause discrepancies between the suspect commit listed in the notification and the suggested assignees listed in that same notification

we should also probably add a comment to the old method was (`get_commits`?)  to indicate that it's not how we currently derive suspect commits

src/sentry/integrations/slack/message_builder/issues.py
cathteng: we only need to fetch suspect commit from the group which is already passed in

src/sentry/integrations/slack/message_builder/notifications/issues.py
```diff
@@ -31,5 +31,4 @@ def build(self) -> SlackBlock:
             issue_details=True,
             notification=self.notification,
             recipient=self.recipient,
-            commits=self.context.get("commits", None),
```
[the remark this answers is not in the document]
cathteng: we don't need to pass it at all if we're not going to use it

src/sentry/integrations/slack/message_builder/issues.py
```diff
@@ -316,31 +317,28 @@ def get_suggested_assignees(
     return []
 
 
-def get_suspect_commit_text(
-    project: Project, event: GroupEvent, commits: Sequence[Mapping[str, Any]] | None = None
-) -> SlackBlock:
+def get_suspect_commit_text(group: Group) -> str | None:
```
cathteng, line 320: i would like to see how often this is None, but i don't think there's an easy way to determine if it's None because there's no suspect commit found or because the suspect commit is found after the notification has been sent

cathteng:
@isabellaenriquez suggested assignees will not be guaranteed to have suspect committers because of the same issue affecting the suspect commit (it's an async task, notifications are sent synchronously so it likely won't exist when the notification is firing). so right now it's probably not sending any suggested assignees including suspect committers

src/sentry/integrations/slack/message_builder/issues.py
```diff
@@ -316,31 +317,28 @@ def get_suggested_assignees(
...
         return None
 
-    commit = commits[0]  # get the most recent commit
     suspect_commit_text = "Suspect Commit: "
-    pull_request = commit.get("pull_request")
-    author = commit.get("author")
-    commit_id = commit.get("id")
+    pull_request = PullRequest.objects.filter(
+        merge_commit_sha=commit.key, organization_id=group.project.organization_id
+    ).first()
```
ykamo001, line 330: if you move this query to after the next if statement, we can save on a DB query since it's not used/needed before the early exit

src/sentry/integrations/slack/message_builder/issues.py
```diff
@@ -316,31 +317,28 @@ def get_suggested_assignees(
...
     if not (author and commit_id):  # we need both the author and commit id to continue
         return None
 
-    author_display = author.get("name") if author.get("name") is not None else author.get("email")
+    author_display = author.name if author.name else author.email
     if pull_request:
-        repo = pull_request.get("repository", {})
-        repo_base = repo.get("url")
-        provider = repo.get("provider", {}).get("id")
+        repo = Repository.objects.get(id=pull_request.repository_id)
```
ykamo001, line 339: should we do the `get_related_object(Repository)` on the original query to help save on doing another db query here?

src/sentry/models/group.py
```diff
@@ -803,6 +804,23 @@ def get_recommended_event_for_environments(
             else self.get_latest_event_for_environments([env.name for env in environments])
         )
 
+    def get_suspect_commit(self) -> Commit | None:
```
ykamo001, line 807: Nice addition, thank you! Wish there was a higher level we could keep this at, but a good start

src/sentry/models/group.py
```diff
@@ -43,6 +43,7 @@
     PriorityChangeReason,
     get_priority_for_ongoing_group,
 )
+from sentry.models.commit import Commit
```
ykamo001, line 46: I think this should be imported in the `get_suspect_commit` method only probably

ykamo001 (approved):

src/sentry/models/group.py
```diff
@@ -803,6 +804,23 @@ def get_recommended_event_for_environments(
             else self.get_latest_event_for_environments([env.name for env in environments])
         )
 
+    def get_suspect_commit(self) -> Commit | None:
+        from sentry.models.groupowner import GroupOwner, GroupOwnerType
+
+        suspect_commit_owner = GroupOwner.objects.filter(
```
malwilley, line 810: Looking [here](https://github.com/getsentry/sentry/blob/a297c789d9c2dbf99d74fdb1872eeb078281b268/src/sentry/utils/committers.py#L305) we order by date_added and filter out groupowner rows without context, may want to do that here as well just to ensure we're getting the same one.

src/sentry/models/group.py
malwilley, line 807: It looks like this only works for suspect commits with git blame, but without integrations we still show suspect commits on the issue details page (see https://docs.sentry.io/product/issues/suspect-commits/#missing-suspect-commits and [this comment](https://github.com/getsentry/sentry/blob/a297c789d9c2dbf99d74fdb1872eeb078281b268/src/sentry/utils/committers.py#L339-L343)). We may want to have the same logic here? Also note that you can have multiple suspect commits in this case

cathteng: which suspect commit is shown on the issue page in the case of multiple suspect commits?

malwilley: You can see all by clicking "show more", the first one is the one that is shown most prominently

PR #70637, opened 2024-05-10
feat(performance): Update queues module samples panel to support displaying producer samples.

edwardgou-sentry:
Updates samples panel to support both consumer and producer samples and fixes some minor ui bugs in samples panel.


static/app/views/performance/queues/messageSamplesPanel.tsx
```diff
@@ -23,22 +24,26 @@ import {MetricReadout} from 'sentry/views/performance/metricReadout';
...
-
-export function MessageConsumerSamplesPanel() {
+export function MessageSamplesPanel() {
   const router = useRouter();
   const query = useLocationQuery({
     fields: {
       project: decodeScalar,
       destination: decodeScalar,
       transaction: decodeScalar,
+      op: decodeScalar,
```
gggritso, line 46: `op` is perfectly fine, but `span.op` is a little nicer (even though it's annoying to type) in case people need to search for it, and it's more consistent with how things are filtered everywhere else

static/app/views/performance/queues/messageSamplesPanel.tsx
```diff
@@ -139,16 +152,23 @@ export function MessageConsumerSamplesPanel() {
...
+              {project ? (
                 <SpanSummaryProjectAvatar
                   project={project}
                   direction="left"
                   size={40}
                   hasTooltip
                   tooltip={project.slug}
                 />
+              ) : (
+                <div />
```
gggritso, line 164: `null` also fine here

static/app/views/performance/queues/messageSamplesPanel.tsx
```diff
@@ -168,36 +188,13 @@ export function MessageConsumerSamplesPanel() {
...
-                align="left"
-                title={t('Processed')}
-                value={transactionMetrics?.[0]?.['count()']}
-                unit={'count'}
+            <MetricsRibbonContainer>
+              <MetricsRibbon
+                type={messageActorType}
+                metrics={transactionMetrics}
                 isLoading={aretransactionMetricsFetching}
               />
```
gggritso, line 196: IMO would be just as good, if not better, to do the conditional ribbon here, rather than making a new `MetricsRibbon` component!

```
{messageActorType === MessageActorType.PRODUCER ? <ProducerMetricsRibb...
```

Otherwise, it's ribbons 3 levels down, which is a lot of levels

static/app/views/performance/queues/messageSamplesPanel.tsx
```diff
@@ -257,6 +256,90 @@ export function MessageConsumerSamplesPanel() {
...
+  isLoading: boolean;
+  metrics: Partial<SpanMetricsResponse>[];
+  type: MessageActorType;
+}) {
+  switch (type) {
+    case MessageActorType.PRODUCER:
+      return <MetricsRibbonProducer metrics={metrics} isLoading={isLoading} />;
+    case MessageActorType.CONSUMER:
+    default:
+      return <MetricsRibbonConsumer metrics={metrics} isLoading={isLoading} />;
```
gggritso, line 273: I think `MetricsRibbonProducer` should be `ProducerMetricsRibbon`, otherwise it sounds like this component is a producer of something rather than a ribbon of something

gggritso (approved):
Makes sense! A few nits

static/app/views/performance/queues/messageSamplesPanel.tsx
edwardgou-sentry: I think we actually need an element here because the parent container is a `grid` and the formatting gets messed up without an element:
image

static/app/views/performance/queues/messageSamplesPanel.tsx
edwardgou-sentry: Updated!

static/app/views/performance/queues/messageSamplesPanel.tsx
edwardgou-sentry: Simplified!

static/app/views/performance/queues/messageSamplesPanel.tsx
edwardgou-sentry: lol good point, updated the name!

static/app/views/performance/queues/messageSamplesPanel.tsx
gggritso: Oof, okay!

PR #70639, opened 2024-05-10
feat(crons): Add deletion apis for processing errors

wedamija:
This adds apis at both the monitor and project level for deleting checkin processing errors.


wedamija:
If we want, we could just make this api per project as well, instead of having two separate ones

src/sentry/monitors/endpoints/project_processing_errors_details.py
```diff
@@ -0,0 +1,57 @@
...
+            403: RESPONSE_FORBIDDEN,
+            404: RESPONSE_NOT_FOUND,
+        },
+    )
+    def delete(self, request: Request, project: Project, uuid: str) -> Response:
+        try:
+            parsed_uuid = UUID(uuid)
+        except ValueError:
+            raise ValidationError("Invalid UUID")
+        CheckinProcessErrorsManager().delete_for_project(project, parsed_uuid)
```
evanpurkhiser, line 56: I am starting to feel like maybe this should just be in it's own module lol nbd

evanpurkhiser:
> If we want, we could just make this api per project as well, instead of having two separate ones

could add a query parameter to filter to monitor slug?

src/sentry/monitors/endpoints/project_processing_errors_details.py
wedamija: Yeah... Feels weird to instantiate it. What if I just make all the methods class/static?

evanpurkhiser: sgtm. I am good w/ a module too

wedamija:
> > If we want, we could just make this api per project as well, instead of having two separate ones
> 
> could add a query parameter to filter to monitor slug?

I can delete it without the monitor slug anyway. We only really have to delete the actually error, we don't have to delete from the index since it'll just drop any errors that don't have values. But I think I can also just get the monitor info from the stores processing error. Maybe it's better to just do that?

evanpurkhiser (approved):

PR #70653, opened 2024-05-10
ref(js): Remove usage of browserHistory.getCurrentLocation

evanpurkhiser:
Newer versions of the history library do not support this function [0].
Use `useLocation` in this case instead.

[0]: https://github.com/remix-run/history/blob/485ebc177c1f3f8eef93b0d654fffd1d321c2ecd/packages/history/index.ts#L188

### DIFF
--- a/static/app/utils/replays/hooks/useActiveReplayTab.spec.tsx
+++ b/static/app/utils/replays/hooks/useActiveReplayTab.spec.tsx
@@ -1,26 +1,16 @@
-import type {Location} from 'history';
+import * as qs from 'query-string';
 
 import {renderHook} from 'sentry-test/reactTestingLibrary';
 
 import {browserHistory} from 'sentry/utils/browserHistory';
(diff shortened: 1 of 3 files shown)


ryan953 (dismissed):

scttcper (dismissed):

evanpurkhiser:
This does appear to break things in a weird way.

evanpurkhiser:
Okay so this is only broken in places where we're making synchronous calls to multiple uses of this hook

Specifically here

https://github.com/getsentry/sentry/blob/30419bf9deb18517328dfab0ef0e70532a2dc3b3/static/app/views/replays/detail/network/useSortNetwork.tsx#L53-L64

We can't do this when using `useLocation` because each subsequent call will not have the new value of the location object, so it will remove values updated in the query string by the previous calls.

Not sure what the right fix is yet.

scttcper (approved):

PR #70660, opened 2024-05-10
ref: remove sentry.utils.functional

asottile-sentry:
many of the functions were only used once so I moved them to their consuming modules


src/sentry/api/endpoints/integrations/sentry_apps/installation/external_issue/actions.py
```diff
@@ -42,7 +56,7 @@ def post(self, request: Request, installation) -> Response:
         uri = data.pop("uri")
 
         try:
-            user = extract_lazy_object(request.user)
+            user = _extract_lazy_object(request.user)
             if isinstance(user, User):
                 user = serialize_rpc_user(user)
```
asottile-sentry, line 61: I'm not even sure this is needed since lazy objects support `isinstance` directly in _most_ cases -- but no tests hit this codepath with `User` so I left it alone

armenzg (approved):

IanWoodard (approved):

PR #70664, opened 2024-05-10
fix(replay): if a video replay has no gestures, hide the rrweb canvas

michellewzhang:
This should probably get removed eventually once all mobile replays have gestures

Before:
SCR-20240510-lgxp

After:
SCR-20240510-lgvh

### DIFF
--- a/static/app/components/replays/replayContext.tsx
+++ b/static/app/components/replays/replayContext.tsx
@@ -15,7 +15,6 @@ import type {
   ReplayPrefs,
 } from 'sentry/components/replays/preferences/replayPreferences';
 import useReplayHighlighting from 'sentry/components/replays/useReplayHighlighting';
-import {VideoReplayer} from 'sentry/components/replays/videoReplayer';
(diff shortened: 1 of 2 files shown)


static/app/components/replays/replayContext.tsx
```diff
@@ -479,37 +480,64 @@ function ProviderNonMemo({
         return null;
       }
 
+      const hasGestures = events?.some(e => e.type === 3);
```
billyvg, line 483: What do you think about having this logic inside of `VideoReplayerWithInteractions` instead?

michellewzhang: i thought about that but it seems worse because we'd have to conditionally instantiate the rrweb `Replayer`, and then in all the `play()`, `pause()` etc functions we'd also have to conditionally call the `Replayer` functions depending on if the rrweb inst exists - feels like way more things to change

billyvg (approved):

PR #70668, opened 2024-05-10
ref(seer grouping): `SeerSimilarIssueData.from_raw` improvements

lobsterkatie:
This makes a number of improvements to the `from_raw1 factory method in the `SeerSimilarIssueData` dataclass, all related to the fact that the data being fed to it comes straight from the JSON returned by Seer and so we can't actually guarantee what's in it.

Included changes: 

- Loosen the typing of the `raw_similar_issue_data` parameter, from `RawSeerSimilarIssueData` (implying we know exactly what's in it) to `Mapping[str, Any]` (acknowledging that we don't).
- Add check to ensure all required keys are present.
- Filter out any data with unknown keys, to prevent the `SeerSimilarIssueData` constructor from crashing with an 'unknown keyword argument' error in the case of unexpected data.

### DIFF
--- a/src/sentry/seer/utils.py
+++ b/src/sentry/seer/utils.py
@@ -1,6 +1,7 @@
 import logging
+from collections.abc import Mapping
 from dataclasses import dataclass
-from typing import NotRequired, Self, TypedDict
+from typing import Any, NotRequired, Self, TypedDict
 
 import sentry_sdk
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


src/sentry/seer/utils.py
```diff
@@ -142,7 +143,22 @@ def from_raw(cls, project_id: int, raw_similar_issue_data: RawSeerSimilarIssueDa
         point to an existing group.
 
         """
-        similar_issue_data = raw_similar_issue_data
+        # Unfortunately, we have to hardcode this separately from the `RawSeerSimilarIssueData` type
+        # definition because Python has no way to derive it from the type (nor vice-versa)
+        required_keys = {"stacktrace_distance", "message_distance", "should_group"}
+        optional_keys = {"parent_hash", "parent_group_id"}
+        expected_keys = {*required_keys, *optional_keys}
```
JoshFerge, line 150: nit: since these are constants, you could pull them out of the function to be on the class / a constant at top of file.

JoshFerge (approved):

src/sentry/seer/utils.py
lobsterkatie: Good idea. Done.

PR #70669, opened 2024-05-10
ref(router): Remove ability to pass `to` as a function

evanpurkhiser:
We don't use this and it's not supported in react-router 6


gggritso (approved):
✂️

scttcper (approved):

PR #70672, opened 2024-05-10
cogs(replays): remove some logs and sample more on dom index logs

JoshFerge:
we log a lot in this file. 

https://cloudlogging.app.goo.gl/J6jw8U7i1LjGNQnCA

lets:
* add some sampling to some logs (canvas size)
* increase sampling for some logs (large mutation)
* remove slow click / multi click logs, as I don't think we need the data gathering from this anymore

### DIFF
--- a/src/sentry/replays/usecases/ingest/dom_index.py
+++ b/src/sentry/replays/usecases/ingest/dom_index.py
@@ -128,7 +128,11 @@ def log_canvas_size(
     events: list[dict[str, Any]],
 ) -> None:
     for event in events:
-        if event.get("type") == 3 and event.get("data", {}).get("source") == 9:
+        if (
+            event.get("type") == 3
+            and event.get("data", {}).get("source") == 9
+            and random.randint(0, 499) < 1
+        ):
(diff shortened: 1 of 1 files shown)


src/sentry/replays/usecases/ingest/dom_index.py
```diff
@@ -128,7 +128,11 @@ def log_canvas_size(
     events: list[dict[str, Any]],
 ) -> None:
     for event in events:
-        if event.get("type") == 3 and event.get("data", {}).get("source") == 9:
+        if (
+            event.get("type") == 3
+            and event.get("data", {}).get("source") == 9
+            and random.randint(0, 499) < 1
```
michellewzhang, line 134: any specific reason this one is `499` and the other is `500`?

michellewzhang (approved):

src/sentry/replays/usecases/ingest/dom_index.py
JoshFerge: ah, no reason 😆

PR #70676, opened 2024-05-10
feat(feedback): flag for spam detection actions

JoshFerge:
would like to start taking action on spam detection for LA and soon EA, so make this a feature instead of an option so we can LA it.

### DIFF
--- a/src/sentry/feedback/usecases/create_feedback.py
+++ b/src/sentry/feedback/usecases/create_feedback.py
@@ -8,7 +8,7 @@ from uuid import uuid4
 
 import jsonschema
 
-from sentry import features, options
+from sentry import features
 from sentry.constants import DataCategory
 from sentry.eventstore.models import Event, GroupEvent
 from sentry.feedback.usecases.spam_detection import is_spam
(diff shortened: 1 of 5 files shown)


src/sentry/feedback/usecases/create_feedback.py
```diff
@@ -253,7 +253,7 @@ def create_feedback_issue(event, project_id: int, source: FeedbackCreationSource
         payload_type=PayloadType.OCCURRENCE, occurrence=occurrence, event_data=event_fixed
     )
     if is_message_spam:
-        auto_ignore_spam_feedbacks(project_id, issue_fingerprint)
+        auto_ignore_spam_feedbacks(project, issue_fingerprint)
```
michellewzhang, line 256: is it safe to interchange `project` with `project_id`?

JoshFerge: i changed function signature below!

michellewzhang: ahh i missed that

michellewzhang (approved):

PR #70681, opened 2024-05-10
fix(trace-explorer): Merge parallel spans

Zylphrex:
Parallel spans can create a deeply nested timeline. This tries to merge them together for a cleaner flatten visualization. We're explicitly not merging root spans as that can create strange visualizations.


shruthilayaj (approved):
lgtm

PR #70685, opened 2024-05-10
feat(performance): Adds trace_status_rate function to span metrics

edwardgou-sentry:
Adds trace_status_rate function to span metrics

### DIFF
--- a/src/sentry/search/events/datasets/spans_metrics.py
+++ b/src/sentry/search/events/datasets/spans_metrics.py
@@ -566,6 +566,30 @@ class SpansMetricsDatasetConfig(DatasetConfig):
                     snql_distribution=self._resolve_count_op,
                     default_result_type="integer",
                 ),
+                fields.MetricsFunction(
+                    "trace_status_rate",
+                    required_args=[
+                        SnQLStringArg("status"),
+                    ],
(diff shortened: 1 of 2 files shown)


src/sentry/search/events/datasets/spans_metrics.py
```diff
@@ -566,6 +566,30 @@ def function_converter(self) -> Mapping[str, fields.MetricsFunction]:
...
+                                Column("value"),
+                                Function(
+                                    "equals",
+                                    [
+                                        Column("metric_id"),
+                                        self.resolve_metric("span.self_time"),
+                                    ],
+                                ),
+                            ],
+                        ),
```
DominikB2014, line 588: To make this more robust, should we only count if there is a trace.status?

DominikB2014 (approved):

src/sentry/search/events/datasets/spans_metrics.py
edwardgou-sentry: Hmm, good point. I thought about this for a bit and I _think_ there might scenarios where we want to include spans that don't have a `trace.status`. For cases where we only want to count spans that have a `trace.status`, we should be able to just add something like `has:trace.status` to the discover query to achieve that affect, so I think I'm going to leave this as is for the flexibility.

PR #70688, opened 2024-05-10
fix(revenue): Use metric seconds api name

scttcper:


mbauer404 (approved):

vbro (approved):

scttcper:
false alarm it uses the apiName

PR #70689, opened 2024-05-10
ref(delayed rules): Don't use a decorator, call add_handler instead

ceorourke:
It doesn't seem like any of the delayed processing is being called since it's not imported by anything, so this PR changes the registry pattern a bit so that it does.

### DIFF
--- a/src/sentry/tasks/post_process.py
+++ b/src/sentry/tasks/post_process.py
@@ -994,7 +994,7 @@ def process_rules(job: PostProcessJob) -> None:
     from sentry.rules.processing.delayed_processing import process_delayed_alert_conditions
 
     if not redis_buffer_registry.has(BufferHookEvent.FLUSH):
-        redis_buffer_registry.add_callback(BufferHookEvent.FLUSH, process_delayed_alert_conditions)
+        redis_buffer_registry.add_handler(BufferHookEvent.FLUSH, process_delayed_alert_conditions)
 
     if job["is_reprocessed"]:
         return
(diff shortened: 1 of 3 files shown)


src/sentry/tasks/post_process.py
```diff
@@ -990,6 +990,12 @@ def _get_replay_id(event):
 
 
 def process_rules(job: PostProcessJob) -> None:
+    from sentry.buffer.redis import BufferHookEvent, BufferHookRegistry
```
saponifi3d, line 993:
```suggestion
    from sentry.buffer.redis import BufferHookEvent, redis_buffer_registry
```

src/sentry/tasks/post_process.py
```diff
@@ -990,6 +990,12 @@ def _get_replay_id(event):
...
+    from sentry.rules.processing.delayed_processing import process_delayed_alert_conditions
+
+    if not BufferHookRegistry._registry.get(BufferHookEvent.FLUSH):
+        BufferHookRegistry.add_callback(BufferHookEvent.FLUSH, process_delayed_alert_conditions)
```
saponifi3d, line 997: It's probably better to not expose the internal `_registry` here if possible, instead we could create a quick wrapper to validate this for us:
```suggestion
    if not redis_buffer_registry.has(BufferHookEvent.FLUSH):
        redis_buffer_registry.add_callback(BufferHookEvent.FLUSH, process_delayed_alert_conditions)
```

Then can move what you have into the function.
```python
def has(self, evt):
    return self._registry.get(evt) != None
```

--- 

This is also nice to have incase we want to refactor the registry later to allow for multiple event handlers.

saponifi3d (approved):
lgtm 🤞

PR #70693, opened 2024-05-10
ref(delayed rules): Add instrumentation

ceorourke:
Add instrumentation and logging to the delayed rule processor to measure how long the bulkier functions are taking and how many rules and groups we're processing. 

Closes https://getsentry.atlassian.net/browse/ALRT-19 and https://github.com/getsentry/team-core-product-foundations/issues/308 (a dupe) as a follow up to https://github.com/getsentry/sentry/pull/69830#pullrequestreview-2036397916

### DIFF
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -333,6 +333,12 @@ def apply_delayed(project_id: int, *args: Any, **kwargs: Any) -> None:
     )
     # STEP 2: Map each rule to the groups that must be checked for that rule.
     rules_to_groups = get_rules_to_groups(rulegroup_to_event_data)
+    logger.info("delayed_processing.num_rules", extra={"rules": len(rules_to_groups.keys())})
+
+    num_groups = 0
+    for groups in rules_to_groups.values():
+        num_groups += len(groups)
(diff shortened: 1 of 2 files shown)


src/sentry/rules/processing/delayed_processing.py
```diff
@@ -364,39 +365,42 @@ def apply_delayed(project_id: int, *args: Any, **kwargs: Any) -> None:
...
-            parsed_rulegroup_to_event_data, project.id, group_ids
-        )
-        for group, groupevent in group_to_groupevent.items():
-            rule_statuses = bulk_get_rule_status(alert_rules, group, project)
-            status = rule_statuses[rule.id]
-            if status.last_active and status.last_active > freq_offset:
-                logger.info(
-                    "delayed_processing.last_active",
-                    extra={"last_active": status.last_active, "freq_offset": freq_offset},
+    with metrics.timer("delayed_processing.fire_rules.duration"):
```
saponifi3d, line 368: just so i can understand, what's the difference between a timer and a span in these cases? is one better than the other?

saponifi3d (approved):
lgtm

src/sentry/rules/processing/delayed_processing.py
ceorourke: I think it's just sentry vs datadog

PR #70697, opened 2024-05-10
chore(issues): Simple type improvements

mrduncan:
Another quick batch of followups to #69828.

For review, no cross-file changes so each one is independently reviewable (aside from pyproject of course). Also, there are a handful in here which are not comprehensive - this intent of this PR is just to chip away at some easy improvements even if they're not yet enough to fully opt a module in quite yet.

### DIFF
--- a/tests/sentry/issues/test_escalating.py
+++ b/tests/sentry/issues/test_escalating.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from datetime import datetime, timedelta
 from typing import Any
 from unittest import mock
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 from uuid import uuid4
 
 import pytest
@@ -95,7 +95,7 @@ class HistoricGroupCounts(
 
     @with_feature("organizations:escalating-issues-v2")
     @mock.patch("sentry.issues.escalating.logger")
-    def test_query_single_group(self, mock_logger) -> None:
+    def test_query_single_group(self, mock_logger: MagicMock) -> None:
         event = self._create_events_for_group()
         assert query_groups_past_counts(Group.objects.all()) == [
             self._create_hourly_bucket(1, event)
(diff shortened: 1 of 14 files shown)


tests/sentry/issues/test_escalating.py
```diff
@@ -259,7 +259,7 @@ def test_datetime_number_of_days() -> None:
 
 class DailyGroupCountsEscalating(BaseGroupCounts):
     def save_mock_escalating_group_forecast(
-        self, group: Group, forecast_values=list[int], date_added=datetime
+        self, group: Group, forecast_values: list[int], date_added: datetime
```
mrduncan, line 262: This is the only non-typing change, no actual effect since defaults are never used and they clearly would have been broken if they were.

tests/sentry/issues/test_run.py
```diff
@@ -20,13 +22,15 @@
 
 
 # need to shut down the connections in the thread for tests to pass
-def process_occurrence_group_with_shutdown(*args, **kwargs):
-    process_occurrence_group(*args, **kwargs)
+def process_occurrence_group_with_shutdown(items: list[Mapping[str, Any]]) -> None:
+    process_occurrence_group(items)
```
MichaelSun48, line 26: Feels like only `**kwargs` can be replaced by the `list[Mapping[str, Any]]` type and `*args?` cannot since there's  no keywords? Or is there some Python/mypy magic happening behind the scenes that makes this okay?

MichaelSun48 (approved):
lgtm! assuming there's some magic going on in `test_run.py` that I'm not aware of which is probably the case

If only mypy could automatically assume a function's return type was None if its not explicit 😢

tests/sentry/issues/test_run.py
asottile-sentry: `**kwargs` could have only been passed as `items=...` before as the underlying signature of `process_occurrence_group` is `(list[Mapping[str, Any]]) -> None` -- so this just makes the signature match explicitly

with `*args, **kwargs` mypy infers that as "this is a callable that could take any number of positional or named arguments" -- now it's explicitly "one possibly-named argument possibly positional argument named `items` of type ..."

asottile-sentry (approved):

PR #70698, opened 2024-05-10
feat(autofix): Log github pr analytics events

jennmueng:
Log analytics events for PR events (opened, closed, merged) on PRs created by the autofix bot

Also logs metrics for each of them.

I know we are no longer calling it "AI Autofix" but prefixing the events with `ai.` should allow us to group this and other ai/ml features together in the future.

#### Next Steps
- Should log Autofix run starts too


src/sentry/analytics/events/ai_autofix_pr_opened.py
```diff
@@ -0,0 +1,16 @@
+from sentry import analytics
+
+
+class AiAutofixPrOpenedEvent(analytics.Event):
+    type = "ai.autofix.pr.opened"
+
+    attributes = (
```
ceorourke, line 7: Instead of having this 3 times you can follow this pattern: https://github.com/getsentry/sentry/blob/master/src/sentry/analytics/events/comment_webhooks.py

ceorourke (approved):
lgtm, my comment shouldn't be a blocker. might want someone from ecosystem to give it another once over, the team that's tagged for integrations is just the pm and em

src/sentry/autofix/utils.py
```diff
@@ -0,0 +1,27 @@
...
+        data=json.dumps(
+            {
+                "provider": provider,
+                "pr_id": pr_id,
+            }
+        ),
+        headers={"content-type": "application/json;charset=utf-8"},
+    )
+
+    response.raise_for_status()
```
cathteng, line 21: will errors be raised from the seer url? `raise_for_status` doesn't raise errors with a helpful error message, this is the code

```
        if 400 <= self.status_code < 500:
            http_error_msg = (
                f"{self.status_code} Client Error: {reason} for url: {self.url}"
            )

        elif 500 <= self.status_code < 600:
            http_error_msg = (
                f"{self.status_code} Server Error: {reason} for url: {self.url}"
            )

        if http_error_msg:
            raise HTTPError(http_error_msg, response=self)
```

cathteng (approved):
no big blockers, just some suggestions and wondering how you want to handle errors from the seer API

src/sentry/autofix/utils.py
jennmueng: Shouldn't raise any predefined errors, and any undefined error we capture to sentry on the seer side and we can handle it from there

src/sentry/analytics/events/ai_autofix_pr_events.py
```diff
@@ -0,0 +1,30 @@
...
+        analytics.Attribute("organization_id"),
+        analytics.Attribute("project_id"),
+        analytics.Attribute("group_id"),
+        analytics.Attribute("run_id"),
+        analytics.Attribute("integration"),
+    )
+
+
+class AiAutofixPrClosedEvent(AiAutofixPrEvent):
+    type = "ai.autofix.pr.closed"
```
ykamo001, line 17: this is the same event as the parent, was this intended?

src/sentry/analytics/events/ai_autofix_pr_events.py
```diff
@@ -0,0 +1,30 @@
...
+
+class AiAutofixPrEvent(analytics.Event):
+    type = "ai.autofix.pr.closed"
+
+    attributes = (
+        analytics.Attribute("organization_id"),
+        analytics.Attribute("project_id"),
+        analytics.Attribute("group_id"),
+        analytics.Attribute("run_id"),
+        analytics.Attribute("integration"),
```
ykamo001, line 12: is this the integration type or the id?

src/sentry/analytics/events/ai_autofix_pr_events.py
```diff
@@ -0,0 +1,30 @@
...
+    )
+
+
+class AiAutofixPrClosedEvent(AiAutofixPrEvent):
+    type = "ai.autofix.pr.closed"
+
+
+class AiAutofixPrMergedEvent(analytics.Event):
```
ykamo001, line 20: should this be using the parent class?

src/sentry/autofix/utils.py
```diff
@@ -0,0 +1,27 @@
...
+    result = response.json()
+
+    if result and result["state"]:
+        return result["state"]
+
+    return None
```
ykamo001, line 27: Could you simplify this? Maybe something like:
```suggestion

    if not result:
        return None

    return result.get("state", None)
```

src/sentry/autofix/webhooks.py
```diff
@@ -0,0 +1,38 @@
+from typing import Any
+
+from sentry import analytics
+from sentry.autofix.utils import get_autofix_state_from_pr_id
+from sentry.models.organization import Organization
+from sentry.utils import metrics
+
+AUTOFIX_GITHUB_APP_USER_ID = 157164994
```
ykamo001, line 8: Should this value come in through settings and not be hard coded?

ykamo001: Or some way to make it variable so that users can control the value, while still being locked down for our use case

src/sentry/autofix/webhooks.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def get_webhook_analytics_fields(autofix_state: dict[str, Any]) -> dict[str, Any]:
+    autofix_request = autofix_state.get("request", {})
+    return {
+        "project_id": autofix_request.get("project_id"),
+        "group_id": autofix_request.get("issue", {}).get("id"),
+        "run_id": autofix_state.get("run_id"),
+    }
```
ykamo001, line 17: Could you break this down so it's easier to test and more modular? Something like:
```suggestion
    webhook_analytics_fields = {
        "project_id": autofix_request.get("project_id", None),
    }
    
    issue = autofix_request.get("issue", None)
    webhook_analytics_fields["group_id"] = issue.get("id", None) if issue else None
    
    ...
    return webhook_analytics_fields
```

src/sentry/autofix/webhooks.py
```diff
@@ -0,0 +1,38 @@
...
+
+from sentry import analytics
+from sentry.autofix.utils import get_autofix_state_from_pr_id
+from sentry.models.organization import Organization
+from sentry.utils import metrics
+
+AUTOFIX_GITHUB_APP_USER_ID = 157164994
+
+
+def get_webhook_analytics_fields(autofix_state: dict[str, Any]) -> dict[str, Any]:
```
ykamo001, line 11: Can we have more strict typing for this?

src/sentry/autofix/webhooks.py
```diff
@@ -0,0 +1,38 @@
...
+        "group_id": autofix_request.get("issue", {}).get("id"),
+        "run_id": autofix_state.get("run_id"),
+    }
+
+
+def handle_github_pr_webhook_for_autofix(
+    org: Organization, action: str, pull_request: dict[str, Any], github_user: dict[str, Any]
+) -> None:
+    if github_user["id"] != AUTOFIX_GITHUB_APP_USER_ID:
+        return
```
ykamo001, line 24:
```suggestion
        return None
```

src/sentry/autofix/webhooks.py
```diff
@@ -0,0 +1,38 @@
...
+
+    if action in ["opened", "closed"]:
```
ykamo001, line 26: can we make this early return to make to remove nesting and make it easier to read?
```suggestion
    if action not in ["opened", "closed"]:
        return None
```

src/sentry/analytics/events/ai_autofix_pr_events.py
jennmueng: type

src/sentry/analytics/events/ai_autofix_pr_events.py
jennmueng: whoops

src/sentry/analytics/events/__init__.py
```diff
@@ -1,4 +1,5 @@
 from .advanced_search_feature_gated import *  # noqa: F401,F403
+from .ai_autofix_pr_events import *  # noqa: F401,F403
```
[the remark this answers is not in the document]
jennmueng: I don't know, I thought we needed to import the events here?

src/sentry/analytics/events/ai_autofix_pr_events.py
jennmueng: whoops again

src/sentry/autofix/webhooks.py
jennmueng: let's move it to settings but we can follow up on configurability later as we haven't set up autofix to run locally yet

PR #70706, opened 2024-05-10
ref(assignee-selector): Get rid of old component for assignee selector in issue details 

### LINKED ISSUE #69829: Issue Details: Use new assignee selector component
Update issue details to use the new component in https://github.com/getsentry/sentry/issues/69054

This should be released under a short-lived feature flag.

MichaelSun48:
This PR swaps the old assignee selector dropdown in the issue details page for the new one that was recently introduced in the issue stream. The design of the trigger has not been changed with the exception of the chevron, which has been changed to use `<Chevron>` instead of `<IconChevron>`.

Closes #69829

Old: 
image

New:
image

TODOs:
- [x] Fix old tests

### DIFF
--- a/static/app/components/assigneeSelectorDropdown.tsx
+++ b/static/app/components/assigneeSelectorDropdown.tsx
@@ -66,6 +66,10 @@ export interface AssigneeSelectorDropdownProps {
    */
   loading: boolean;
   /**
+   * Additional styles to apply to the dropdown
+   */
+  className?: string;
+  /**
    * Optional list of members to populate the dropdown with.
    */
   memberList?: User[];
@@ -87,10 +91,6 @@ export interface AssigneeSelectorDropdownProps {
    * Optional list of suggested owners of the group
    */
   owners?: Omit<SuggestedAssignee, 'assignee'>[];
-  /**
-   * Additional styles to apply to the dropdown
-   */
-  style?: React.CSSProperties;
   /**
    * Optional trigger for the assignee selector. If nothing passed in,
    * the default trigger will be used
@@ -196,6 +196,7 @@ export function AssigneeAvatar({
 }
 
 export default function AssigneeSelectorDropdown({
+  className,
   group,
   loading,
   memberList,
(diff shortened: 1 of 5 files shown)


static/app/components/assigneeSelectorDropdown.tsx
```diff
@@ -87,6 +87,10 @@ export interface AssigneeSelectorDropdownProps {
    * Optional list of suggested owners of the group
    */
   owners?: Omit<SuggestedAssignee, 'assignee'>[];
+  /**
+   * Additional styles to apply to the dropdown
+   */
+  style?: React.CSSProperties;
```
malwilley, line 93: I don't think you want to add this prop - `style` should only be used for dynamic styles (they override all CSS which causes issues for styling down the road). Can you add a `className` instead?

static/app/components/group/assignedTo.tsx
```diff
@@ -184,6 +192,40 @@ function AssignedTo({
...
+      }
+
+      await clearAssignment(group.id, organization.slug, 'assignee_selector');
+      return Promise.resolve(null);
+    },
+    onSuccess: (newAssignee: AssignableEntity | null) => {
+      if (onAssign && newAssignee) {
+        onAssign(newAssignee.type, newAssignee.assignee, newAssignee.suggestedAssignee);
+      }
+      setAssigneeLoading(false);
```
malwilley, line 221: No need for a separate loading state in this component, `useMuatation` returns `isLoading` for this purpose

malwilley:
Looks like one of the acceptance tests is failing - you'll probably need to change how the test is selecting the assignee dropdown

static/app/components/group/assignedTo.tsx
MichaelSun48: So I was going to say that I tried this when I replaced the assignee dropdown in issues stream and it refused to work... but then I went to record an instance of it not working in issue details, and it works now??

malwilley (approved):
Looks good!

PR #70707, opened 2024-05-10
fix(files): close mmap in abstract file _prefetch

JoshFerge:
we seem to have a lot of warnings about this function and unclosed sockets. reading python docs, it seems like we should close this mmap. I'm not quite sure if this is the source of the log.

https://cloudlogging.app.goo.gl/Yu321395kTNxZFiWA

### DIFF
--- a/src/sentry/models/files/abstractfile.py
+++ b/src/sentry/models/files/abstractfile.py
@@ -107,6 +107,7 @@ class ChunkedFileBlobIndexWrapper:
                 exe.submit(fetch_file, idx.offset, idx.blob.getfile)
 
         mem.flush()
+        mem.close()
         self._curfile = f
 
     def close(self):


JoshFerge:
looking at this some more, the sockets are network based, not file based, so i don't think this PR is correct. looking further

JoshFerge:
gonna guess its coming from https://github.com/getsentry/getsentry/blob/056e9cc337b98984b37ffd3b40a79adea07c32e5/getsentry/filestore.py#L141

src/sentry/models/files/abstractfile.py
```diff
@@ -107,6 +107,7 @@ def fetch_file(offset, getfile):
                 exe.submit(fetch_file, idx.offset, idx.blob.getfile)
 
         mem.flush()
+        mem.close()
```
asottile-sentry, line 110: mmap supports the context manager protocol so I would recommend that instead

mmap also doesn't trigger a `ResourceWarning` when left unclosed (though arguably it should) so your warnings are definitely coming from somewhere else -- do you have some sample lines?

JoshFerge: yeah, i think its coming from the getfile function that is passed in (https://github.com/getsentry/getsentry/blob/056e9cc337b98984b37ffd3b40a79adea07c32e5/getsentry/filestore.py#L141 in prod)
and example warning is: 
```
/usr/src/sentry/src/sentry/models/files/abstractfile.py:38: ResourceWarning: unclosed <socket.socket fd=179, family=2, type=1, proto=6, laddr=('10.4.161.44', 43014), raddr=('10.8.0.230', 11211)>
  self._prefetch(prefetch_to, delete)
```

JoshFerge:
https://github.com/getsentry/getsentry/pull/13948

asottile-sentry:
this should probably happen anyway -- right now that file descriptor is torn down at an undefined time

PR #70709, opened 2024-05-11
feat(crons): Add stronger typing to error responses

wedamija:
wip


src/sentry/api/urls.py
```diff
@@ -651,6 +651,9 @@
 from ..monitors.endpoints.organization_monitor_processing_errors_index import (
     OrganizationMonitorProcessingErrorsIndexEndpoint,
 )
+from ..monitors.endpoints.project_processing_errors_details import (
```
evanpurkhiser, line 654: Heads up @wedamija looks like sometimes your editor is wanting to do parent relative imports

wedamija: It just copies other patterns it sees, so since it sees relative imports it does dumb stuff like this

PR #70712, opened 2024-05-11
feat(Activated Alerts): implement deploy activator

nhsiehgit:
Deploy reference a specific release being deployed to a specific environment

To activate a monitor for a deploy means to start monitoring per release AND environment for a specific Project.

This PR subscribes the monitor subscription to the creation of a `ReleaseProjectEnvironment` which for all intents and purposes is synonymous with "deploy"

### DIFF
--- a/src/sentry/models/releaseprojectenvironment.py
+++ b/src/sentry/models/releaseprojectenvironment.py
@@ -34,7 +34,7 @@ class ReleaseStages(str, Enum):
 class ReleaseProjectEnvironmentManager(BaseManager["ReleaseProjectEnvironment"]):
     @staticmethod
     def subscribe_project_to_alert_rule(
-        project: Project, release: Release, environment: Environment, activator: str, trigger: str
+        project: Project, release: Release, environment: Environment, trigger: str
     ) -> list[QuerySubscription]:
         """
         TODO: potentially enable custom query_extra to be passed on ReleaseProject creation (on release/deploy)
@@ -44,6 +44,8 @@ class ReleaseProjectEnvironmentManager(BaseManager["ReleaseProjectEnvironment"])
         from sentry.incidents.models.alert_rule import AlertRule
 
         query_extra = f"release:{release.version} and environment:{environment.name}"
+        # TODO: parse activator on the client to derive release version / environment name
+        activator = f"release:{release.version} and environment:{environment.name}"
         return AlertRule.objects.conditionally_subscribe_project_to_alert_rules(
             project=project,
             activation_condition=AlertRuleActivationConditionType.DEPLOY_CREATION,
--- a/src/sentry/models/releaseprojectenvironment.py
+++ b/src/sentry/models/releaseprojectenvironment.py
@@ -59,15 +59,14 @@ class ReleaseProjectEnvironmentManager(BaseManager["ReleaseProjectEnvironment"])
     def post_save(self, instance, created, **kwargs):
         if created:
             release = instance.release
-            projects = instance.project
+            project = instance.project
             environment = instance.environment
-            for project in projects:
-                self.subscribe_project_to_alert_rule(
-                    project=project,
-                    release=release,
-                    environment=environment,
-                    trigger="releaseprojectenvironment.post_save",
-                )
+            self.subscribe_project_to_alert_rule(
+                project=project,
+                release=release,
+                environment=environment,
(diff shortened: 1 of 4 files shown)


src/sentry/models/releaseprojectenvironment.py
```diff
@@ -12,16 +15,60 @@
...
+        """
+        TODO: potentially enable custom query_extra to be passed on ReleaseProject creation (on release/deploy)
+
+        NOTE: import AlertRule model here to avoid circular dependency
+        """
+        from sentry.incidents.models.alert_rule import AlertRule
+
+        query_extra = f"release:{release.version} and environment:{environment.name}"
+        # TODO: parse activator on the client to derive release version / environment name
+        activator = f"release:{release.version} and environment:{environment.name}"
```
saponifi3d, line 50: nit: rather than storing as text blobs, thoughts on building this string in the UI and storing release / env in json? (unclear how large of a refactor this ask is though).

Also, curious why we need to replicate this data? Should the activator only store activation reasons, and we remove activation info from query_extra? My understanding is that query_extra is meant to be meta data about the query itself stored in text. The activator is meant to be the release / deploy / environment information that triggered the alert on activated alerts.

Do we need to do anything to anticipate deploy version info here?

saponifi3d (commented):
just the question about query_extra vs activator - code seems good, please make sure to add some tests around the manager (now will be the easiest time to add them)

nhsiehgit:
Yup looking at tests now

src/sentry/models/releaseprojectenvironment.py
nhsiehgit: Yea - query_extra is actually being tacked on to the query.

`activator` is simply something we can provide to the client to help better understand what triggered the monitor.
For now i mirrored it with the query_extra, btu we could make this any string 🤷 

Deploy version isn't valuable and we do not surface it anywhere else in the UI either :( 
'deploy's are simply releases on certain environments, and that's how we surface it elsewhere so stuck wtih that

nhsiehgit: This trigger isn't being generated from any UI, so we cannot build anything in the UI / have nothing to store in JSON

saponifi3d: I was more thinking that the format could be store as json rather than a string - incase we want to parse it later or share it via an API. it's just a nice to have that might make the future a little easier. Basically, just doing: `json.dumps({ release: release.version, environment: environent.name })` then we can parse it with `json.loads()` on fetch. Just thinking it might be nicer to interact with this as a dictionary than as a string incase we need the release / env name later, but isn't a blocker.

The thing i was thinking with deploys is that in the future we might want to have an activator be a specific deploy, but def don't need to get hung-up on it for the mvp either - so :punt:

tests/sentry/models/test_releaseprojectenvironment.py
```diff
@@ -83,3 +92,68 @@ def test_no_update_too_close(self):
...
+        ReleaseProjectEnvironment.get_or_create(
+            project=self.project,
+            release=self.release,
+            environment=self.environment,
+            datetime=self.datetime_now,
+        )
+
+        assert mock_subscribe_project_to_alert_rule.call_count == 1
+
+    @patch(
```
saponifi3d, line 110: question about sentry testing philosophy... i've seen mocking sparingly and mostly checks for the db to have the conditions completed afterwards. is there a general recommendation?

(I personally like mocking out external logic like this because it improves test performance so much.. but the trade-off is testing implementation a bit more explicitly over the logic of the function)

saponifi3d (approved):

tests/sentry/models/test_releaseprojectenvironment.py
nhsiehgit: mmmm 🤔 

I know we tend to like full e2e tests in our unit tests....
Personally i'm also in favor of testing the individual units and mocking the relations... IMO makes it easier to more clearly understand whats going on and whats expected?

that being said - i have the non patched tests below as well 😂

nhsiehgit:
hmmmm yea- i considered the JSON as well

it would definitely make it easier to parse for the client.

but i'm a little sus of sticking JSON into the db... for now it's just a short charfield, so if we want to enable JSON we'd need to update the column with a bigger size 😬

PR #70736, opened 2024-05-13
ref(ddm): Clean up unused options and feature flags

iambriccardo:
This PR removes some unused feature flags and options that were needed for DDM.


shellmayr:
The flag 'custom-metrics-experimental' seems to be used in quite some parts of the front-end through hasMetricsExperimentalFeature - should this be cleaned up together?

iambriccardo:
> The flag 'custom-metrics-experimental' seems to be used in quite some parts of the front-end through hasMetricsExperimentalFeature - should this be cleaned up together?

Good catch, I forgot frontend uses options with implicit `organizations:` prefix.

shellmayr (approved):

PR #70743, opened 2024-05-13
ref: move integration docs / sdk registry building to sentry.build

asottile-sentry:
I'm planning to kill the setuptools-based build which is blocking our upgrade to python 3.12 -- this is the first step

### DIFF
--- a/src/sentry/build/_integration_docs.py
+++ b/src/sentry/build/_integration_docs.py
@@ -1,0 +1,117 @@
+from __future__ import annotations
+
+import argparse
+import concurrent.futures
+import json
+import multiprocessing
+import os.path
+
+from sentry.build._download import urlopen_with_retries
+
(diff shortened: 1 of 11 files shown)


src/sentry/build/_integration_docs.py
evanpurkhiser, line 1: In the future I think we would like to just completely drop this, the docs website no longer actively derives this and it’s just a hardcoded json file. 

Unfortunately there’s still a good bit of usage of this in the backend I think. Cc @mjq-sentry

evanpurkhiser (approved):

src/sentry/build/_integration_docs.py
asottile-sentry: yeah it certainly is a little spooky that the docs repo can affect the production build, sorta silently too

PR #70751, opened 2024-05-13
feat(metrics): remove percentiles from Metrics API

shellmayr:
As it has been agreed upon that percentiles shall be removed from the product for now, this PR removes the percentile functions from the available operations for distributions.


shellmayr:
Closing PR: discovered additional complexities and dependencies between product verticals that make the "elaborate" solution much more elaborate than useful. Opting for a simpler approach. Percentiles don't need to be removed from *everything*, but only the UI-facing parts of the Metrics product itself.

PR #70752, opened 2024-05-13
ref: replace legacy keys with canonical keys in samples

asottile-sentry:
the first large chunk of `CanoncalKeyDict` errors are related to sample data still having the legacy keys -- this fixes those

### DIFF
--- a/src/sentry/data/samples/android-transaction.json
+++ b/src/sentry/data/samples/android-transaction.json
@@ -4,7 +4,7 @@
   "release": "io.sentry.samples.android@1.1.0+2",
   "dist": "2",
   "platform": "java",
-  "message": "",
+  "logentry": {},
   "tags": [
     ["device", "sdk_gphone64_arm64"],
     ["device.family", "sdk_gphone64_arm64"],
--- a/src/sentry/data/samples/aspnetcore.json
+++ b/src/sentry/data/samples/aspnetcore.json
@@ -3,7 +3,6 @@
   "dist": null,
   "platform": "csharp",
   "culprit": "Samples.AspNetCore.Mvc.Controllers.HomeController in PostIndex",
-  "message": "Param is null!",
   "time_spent": null,
   "tags": [
     ["server-runtime", ".NET Core 2.1.5"],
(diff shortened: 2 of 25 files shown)


jan-auer (commented):
(I didn't check usages in the code, yet) Are these samples used in tests? I'm asking as IIRC at least `message` is still used by SDKs and is documented as a valid key. We could fix such tests by running the payload through librelay's normalize function, if that's the case.

asottile-sentry:
> (I didn't check usages in the code, yet) Are these samples used in tests? I'm asking as IIRC at least `message` is still used by SDKs and is documented as a valid key. We could fix such tests by running the payload through librelay's normalize function, if that's the case.

they're used in tests and I believe public samples -- they are injected directly into eventstore in tests which I believe is why they're triggering canonical rewriting there -- (the canonical rewriting just makes the `message` show up in `logentry`)

src/sentry/data/samples/aspnetcore.json
```diff
@@ -3,7 +3,6 @@
   "dist": null,
   "platform": "csharp",
   "culprit": "Samples.AspNetCore.Mvc.Controllers.HomeController in PostIndex",
-  "message": "Param is null!",
```
asottile-sentry, line 6: this was duplicated below as `logentry` already

src/sentry/data/samples/android-transaction.json
```diff
@@ -4,7 +4,7 @@
   "release": "io.sentry.samples.android@1.1.0+2",
   "dist": "2",
   "platform": "java",
-  "message": "",
+  "logentry": {},
```
jan-auer, line 7: Shouldn't the empty ones be removed?

src/sentry/data/samples/csp.json
```diff
@@ -1,7 +1,7 @@
 {
   "culprit": "script-src 'self' 'unsafe-eval' 'unsafe-inline' www.example.com",
-  "message": "Blocked 'script' from 'example.org'",
-  "sentry.interfaces.Csp": {
+  "logentry": {"formatted": "Blocked 'script' from 'example.org'"},
```
jan-auer, line 3: _nit_: 

```suggestion
  "logentry": {
    "formatted": "Blocked 'script' from 'example.org'"
  },
```

jan-auer (approved):

src/sentry/data/samples/android-transaction.json
asottile-sentry: I think these are intentional to avoid the code which injects a default message here so I left them

PR #70767, opened 2024-05-13
feat(insights): add cache miss rate chart to sidebar

DominikB2014:
You might want to investigate a drop in miss rate scoped to a transaction, so to cover that case we're adding a transaction miss rate chart in the sidebar.

image

### DIFF
--- a/static/app/views/performance/cache/referrers.ts
+++ b/static/app/views/performance/cache/referrers.ts
@@ -9,4 +9,5 @@ export enum Referrer {
   SAMPLES_CACHE_TRANSACTION_DURATION = 'api.performance.cache.samples-cache-transaction-duration',
   SAMPLES_CACHE_SPAN_SAMPLES = 'api.performance.cache.samples-cache-span-samples',
   SAMPLES_CACHE_SPAN_SAMPLES_TRANSACTION_DURATION = 'api.performance.cache.samples-cache-span-samples',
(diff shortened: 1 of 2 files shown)


static/app/views/performance/cache/referrers.ts
```diff
@@ -9,4 +9,5 @@ export enum Referrer {
   SAMPLES_CACHE_TRANSACTION_DURATION = 'api.performance.cache.samples-cache-transaction-duration',
   SAMPLES_CACHE_SPAN_SAMPLES = 'api.performance.cache.samples-cache-span-samples',
   SAMPLES_CACHE_SPAN_SAMPLES_TRANSACTION_DURATION = 'api.performance.cache.samples-cache-span-samples',
+  SAMPLES_CACHE_HIT_MISS_CHART = 'api.performance.cache.samples-cache-hit-miss-chart',
```
narsaynorath, line 12: You might have to add these referrers to the `referrer.py` file and also register them with the events/event stats endpoints to get them to really work

narsaynorath (approved):

static/app/views/performance/cache/referrers.ts
narsaynorath: Yeah no worries, I've often put up backend PRs and linked them to where I'm using them in the frontend in the PR and merged them sequentially. Nowadays I try to think of the referrers ahead of time and add them in bulk way early. It was just a reminder because I noticed it wasn't registered yet

PR #70769, opened 2024-05-13
feat(insights): add analytics to database sample panel

KevinL10:
Adds analytics to the span sample panel to track:
- How often the panels are opened
- How often spans are clicked
- How often “Try different samples” is clicked
- How often filters are interacted with

This PR adds analytics to database only as a first step. I'll be adding analytics to the remaining modules in the next PR.

### DIFF
--- a/static/app/views/starfish/components/samplesTable/spanSamplesTable.tsx
+++ b/static/app/views/starfish/components/samplesTable/spanSamplesTable.tsx
@@ -7,6 +7,7 @@ import Link from 'sentry/components/links/link';
 import {Tooltip} from 'sentry/components/tooltip';
 import {IconProfiling} from 'sentry/icons/iconProfiling';
 import {t} from 'sentry/locale';
+import {trackAnalytics} from 'sentry/utils/analytics';
 import {generateLinkToEventInTraceView} from 'sentry/utils/discover/urls';
 import {useLocation} from 'sentry/utils/useLocation';
 import useOrganization from 'sentry/utils/useOrganization';
@@ -67,6 +68,7 @@ type Props = {
   avg: number;
   data: SpanTableRow[];
   isLoading: boolean;
+  moduleName: string;
   columnOrder?: SamplesTableColumnHeader[];
   highlightedSpanId?: string;
   onMouseLeaveSample?: () => void;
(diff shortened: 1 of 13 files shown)


DominikB2014 (commented):
Changes are looking good, but i'm curious about something!

Will we be able to gather stats relative to the amount of people that are on the page as well? Knowing the total amount of clicks on samples are helpful so we can compare between modules. But also knowing what proportion of people that enter a module arrive to a sample is also an interesting question to answer. 

Answering this question directly might not be too feasible, but perhaps some other metric could help us with this!

KevinL10:
Agree it would be good to have proportion of module visits that arrive to at least 1 sample. For now, we have the total number of people who visit the module itself, so we have a rough estimate of the average # of sample panels opened per page visit. I'll see if there are any other similar metrics we can gather.

static/app/views/starfish/components/samplesTable/spanSamplesTable.tsx
```diff
@@ -67,6 +68,7 @@ type Props = {
   avg: number;
   data: SpanTableRow[];
   isLoading: boolean;
+  moduleName: string;
```
DominikB2014, line 71: We have a moduleName type here, that we should probably use instead of `string`
https://github.com/getsentry/sentry/blob/master/static/app/views/starfish/types.tsx#L13
Although its missing a bunch of modules, so we could extend that.

DominikB2014 (approved):
looking good overall! Feel free to address my comments as you see fit.

Personally i'm not a fan of adding a moduleName prop everywhere, it doesn't really dictate any difference in component behaviour and its used purely for analytics. We could consider determining the name from the URL. I'm cool if we think about this later too!

static/app/views/starfish/components/samplesTable/spanSamplesTable.tsx
gggritso: +1 to this! If you add the missing modules to `ModuleName` and use `ModuleName` instead of strings that'd be great 👍🏻

gggritso (commented):
Makes sense 👍🏻 just needs to rely on `ModuleName` enum rather than plain strings

narsaynorath (approved):
Nice, lgtm!

KevinL10:
thanks all! changed to ModuleName type and added the other modules. May consider inferring module from URL later on.

PR #70771, opened 2024-05-13
ref(insights): Wrap mobile insights module in `ModulePageProviders`

gggritso:
Preparation for URL and breadcrumb changes. The Mobile modules don't wrap pages in `ModulePageProviders`. I need the wrapper so that I can wholesale change the base URL for all modules at once without doing awkward search-and-replace.

Best enjoyed with whitespace off.

## Changes

- Fix typing in `ModulePageProviders`
    
    Delegate to `Feature` for anything passed through.

- Wrap mobile modules in `ModulePageProviders`
    
    `ModulePageProviders` includes `Feature`, `SentryDocumentTitle`, and
    `PageFiltersContainer`, so it's not necessary to wrap in those
    components manually.


gggritso:
@narsaynorath could you take a look at this? Would love a second set of eyes to make sure the feature flag names and page titles are still correct

static/app/views/performance/mobile/ui/screenSummary/index.tsx
```diff
@@ -72,72 +71,82 @@ function ScreenSummary() {
...
-export default ScreenSummary;
+function PageWithProviders() {
+  const location = useLocation<Query>();
+
+  const {transaction} = location.query;
+
+  return (
+    <ModulePageProviders
+      title={[transaction, t('Screen Loads')].join(' — ')}
+      baseURL="/performance/browser/app-startup"
```
narsaynorath, line 141:
```suggestion
      baseURL="/performance/mobile/ui/spans"
```

static/app/views/performance/mobile/appStarts/screenSummary/index.tsx
```diff
@@ -109,120 +109,130 @@ function ScreenSummary() {
...
+}
+
+function PageWithProviders() {
+  const location = useLocation<Query>();
+
+  const {transaction} = location.query;
+
+  return (
+    <ModulePageProviders
+      title={[transaction, ROUTE_NAMES['app-startup']].join(' — ')}
```
narsaynorath, line 226: It looks like this `ROUTE_NAMES` constant is only really used by mobile modules. If you wanted to remove the constant and just define the title inline (i.e. `t(App Starts)`) then that removes another file from the old starfish path

static/app/views/performance/mobile/appStarts/screenSummary/index.tsx
```diff
@@ -109,120 +109,130 @@ function ScreenSummary() {
...
+      baseURL="/performance/browser/app-startup"
```
narsaynorath, line 227:
```suggestion
      baseURL="/performance/mobile/app-startup/spans"
```

I think this is the appropriate path for this file. This component renders the drill down into a screen which uses the `/performance/mobile/app-startup/spans` URL. Unless this component uses `baseURL` as the module list page then the URL should end with `/spans`

static/app/views/performance/mobile/appStarts/index.tsx
```diff
@@ -1,25 +1,34 @@
...
-          docsUrl:
-            'https://docs.sentry.io/product/performance/mobile-vitals/app-starts/#minimum-sdk-requirements',
-        }}
-        content={<AppStartup chartHeight={200} />}
-        title={ROUTE_NAMES['app-startup']}
-      />
-    </Feature>
+    <ModulePageProviders
+      title={ROUTE_NAMES['app-startup']}
+      baseURL="/performance/browser/app-startup"
```
narsaynorath, line 26:
```suggestion
      baseURL="/performance/mobile/app-startup"
```

narsaynorath (commented):
Just a few spots where the URL is still referencing `browser` instead of `mobile`. Also gave a suggestion to remove another old starfish file. Feel free to re-request a review and I can do a quick click through of the modules to make sure everything's as expected. Thanks @gggritso!

static/app/views/performance/mobile/appStarts/screenSummary/index.tsx
gggritso: Ah good call, I can take a look separately

gggritso:
@narsaynorath thanks, good catches! Copy-pasting too fast on my side. The `baseURL` isn't in use yet, so I'll probably have to tweak those a bit once I actually use them for routing, too 👍🏻

narsaynorath (approved):

PR #70775, opened 2024-05-13
fix(assignee-badge): Fix storybook issues 

MichaelSun48:
Fixes JAVASCRIPT-2T1X and an issue where the title for the AssigneeBadge's storybook showed up as `<n/>` instead of `<AssigneeBadge/>`

### DIFF
--- a/static/app/components/assigneeBadge.stories.tsx
+++ b/static/app/components/assigneeBadge.stories.tsx
@@ -1,12 +1,13 @@
 import {Fragment, useState} from 'react';
+import {uuid4} from '@sentry/utils';
 
 import {AssigneeBadge} from 'sentry/components/assigneeBadge';
 import storyBook from 'sentry/stories/storyBook';
-import type {Actor} from 'sentry/types';
+import type {Actor, Team} from 'sentry/types';
 import {useUser} from 'sentry/utils/useUser';
 import {useUserTeams} from 'sentry/utils/useUserTeams';
 
-export default storyBook(AssigneeBadge, story => {
+export default storyBook('AssigneeBadge', story => {
   story('User Assignee', () => {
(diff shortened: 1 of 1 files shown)


static/app/components/assigneeBadge.stories.tsx
```diff
@@ -39,10 +40,30 @@ export default storyBook(AssigneeBadge, story => {
     const [chevron1Toggle, setChevron1Toggle] = useState<'up' | 'down'>('down');
     const [chevron2Toggle, setChevron2Toggle] = useState<'up' | 'down'>('down');
 
+    const team: Team =
+      teams.length > 0
```
michellewzhang, line 44:
```suggestion
      teams.length > 0
```
! NIT ! can just do `teams.length` here!

static/app/components/assigneeBadge.stories.tsx
```diff
@@ -39,10 +40,30 @@ export default storyBook(AssigneeBadge, story => {
...
+            teamRole: null,
+            isMember: true,
+            memberCount: 0,
+            avatar: {avatarType: 'letter_avatar', avatarUuid: uuid4()},
+            flags: {
+              'idp:provisioned': false,
+            },
+            externalTeams: [],
+            hasAccess: false,
+            isPending: false,
```
michellewzhang, line 60: could probably directly use `TeamFixture` here instead of hardcoding a fake team but this also works

michellewzhang (approved):

static/app/components/assigneeBadge.stories.tsx
MichaelSun48: I tried that but for some reason the import wasn't resolving correctly so I just copy pasted the teamfixture object

PR #70776, opened 2024-05-13
ref(insights): Wrap AI module in `ModulePageProviders`

gggritso:
Preparation for URL and breadcrumb changes. The IA module doesn't wrap pages in `ModulePageProviders`. I need the wrapper so that I can wholesale change the base URL for all modules at once without doing awkward search-and-replace.

Best enjoyed with whitespace off.

## Changes

- Wrap IA module in `ModulePageProviders`
    
    `ModulePageProviders` includes `Feature`, `SentryDocumentTitle`, and
    `PageFiltersContainer`, so it's not necessary to wrap in those
    components manually.

- Pass pipeline name through URL
    
    Makes sure the name is available at page load, and is consistent with
    what the user clicked on _and_ saves a loaded column.

### DIFF
--- a/static/app/views/aiMonitoring/aiMonitoringDetailsPage.tsx
+++ b/static/app/views/aiMonitoring/aiMonitoringDetailsPage.tsx
@@ -1,16 +1,12 @@
 import styled from '@emotion/styled';
 
-import Feature from 'sentry/components/acl/feature';
-import {Alert} from 'sentry/components/alert';
 import {Breadcrumbs} from 'sentry/components/breadcrumbs';
 import * as Layout from 'sentry/components/layouts/thirds';
 import NoProjectMessage from 'sentry/components/noProjectMessage';
 import {DatePageFilter} from 'sentry/components/organizations/datePageFilter';
 import {EnvironmentPageFilter} from 'sentry/components/organizations/environmentPageFilter';
 import PageFilterBar from 'sentry/components/organizations/pageFilterBar';
-import PageFiltersContainer from 'sentry/components/organizations/pageFilters/container';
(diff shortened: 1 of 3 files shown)


static/app/views/aiMonitoring/aiMonitoringDetailsPage.tsx
```diff
@@ -25,30 +23,32 @@ import {
...
+  'span.description'?: string;
+};
+
+export function AiMonitoringPage({params}: Props) {
+  const location = useLocation<Query>();
+
   const organization = useOrganization();
   const {groupId} = params;
 
+  const spanDescription = decodeScalar(location.query?.['span.description']);
```
colin-sentry, line 50: smells like XSS but I'll let it pass.

colin-sentry (approved):

static/app/views/aiMonitoring/aiMonitoringDetailsPage.tsx
gggritso: This is a good comment, so I don't want to leave it unanswered before merging. The "smell" here is that I'm rendering URL content into the DOM. This does _smell_ like a reflected XSS vulnerability, but it's safe.

The attack vector is crafting a URL like `https://<org>.sentry.io/ai-monitoring/pipeline-type/<HASH>/?span.description=<script>alert("you win")</script>`. The reason this is _not_ an XSS vulnerability is that React safely escapes all rendered content:

Screenshot 2024-05-14 at 10 07 08 AM

In contrast, using `dangerouslySetInnerHTML` _would_ create a reflected XSS attack here.

PR #70778, opened 2024-05-13
fix(feedback): check latest event for attachment

michellewzhang:
we send screenshots through the feedback event, so instead of checking the issue for attachments, we should check the latest event.

this should fix the issue where the `IconImage` isn't showing up when a feedback has a screenshot attached.

SCR-20240513-jcau

### DIFF
--- a/src/sentry/api/endpoints/group_details.py
+++ b/src/sentry/api/endpoints/group_details.py
@@ -236,7 +236,7 @@ class GroupDetailsEndpoint(GroupEndpoint, EnvironmentMixin):
                 )
                 data.update({"sentryAppIssues": sentry_app_issues})
 
-            if "hasAttachments" in expand:
+            if "latestEventHasAttachments" in expand:
                 if not features.has(
                     "organizations:event-attachments",
                     group.project.organization,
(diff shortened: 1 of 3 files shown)


src/sentry/api/endpoints/group_details.py
```diff
@@ -236,16 +236,19 @@ def get(self, request: Request, group) -> Response:
                 )
                 data.update({"sentryAppIssues": sentry_app_issues})
 
-            if "hasAttachments" in expand:
+            if "latestEventHasAttachments" in expand:
```
JoshFerge, line 239: does this break any behavior on the issue details page?

michellewzhang: ex here https://github.com/getsentry/sentry/blob/79871b806059d17c3014ae0e9c235c544dac4da3/static/app/components/feedback/useFeedbackListQueryKey.tsx#L97

JoshFerge: got it 👍🏼

JoshFerge (approved):

PR #70781, opened 2024-05-13
feat(crons): Remove culprit from crons incident issues

evanpurkhiser:
This is a bit confusing, we don't need a culprit


davidenwang (approved):

JoshFerge (approved):
wonder if in the future we could get the module name or something for the culprit? but not a big deal

evanpurkhiser:
>wonder if in the future we could get the module name or something for the culprit? but not a big deal

Unforutnately highly dependent on how configuration of crons is done

PR #70782, opened 2024-05-13
ref(insights): Use consistent `ModulePageProviders` wrapping

gggritso:
- always lift `ModulePageProviders` out of the component
- always call it `PageWithProviders`

This is just for consistency, so I can find these easier when I move things around later this week.


DominikB2014 (approved):
🧹

PR #70784, opened 2024-05-13
feat(api-idorslug): Update Subset of Monitor & Integration Endpoints and to use `organization_id_or_slug`

iamrajjoshi:
A subset of changes from https://github.com/getsentry/sentry/pull/70081!


src/sentry/monitors/endpoints/monitor_ingest_checkin_attachment.py
isabellaenriquez, line 1: we should update the comments here too to be accurate to what is happening (ie. all the references to slug)

sentaur-athena (approved):
Looks good. Please address the nits before merging.

PR #70785, opened 2024-05-13
feat(crons): Improve occurance evidence Failure reason

evanpurkhiser:
The failure reason of "incident" isn't super helpful

We can instead say something like

> 3 missed check-ins detected
> 2 missed check-ins, 1 timeout check-in and 1 error check-in were detected
> A failed check-in was detected

This implements that

### DIFF
--- a/src/sentry/monitors/logic/mark_failed.py
+++ b/src/sentry/monitors/logic/mark_failed.py
@@ -3,8 +3,9 @@ from __future__ import annotations
 import logging
 import uuid
 from collections import Counter
+from collections.abc import Mapping, Sequence
 from datetime import datetime, timezone
-from typing import TYPE_CHECKING, Mapping, Sequence, TypedDict, cast
+from typing import TYPE_CHECKING, TypedDict, cast
 
 from django.db.models import Q
 from django.utils.text import get_text_list
--- a/src/sentry/monitors/logic/mark_failed.py
+++ b/src/sentry/monitors/logic/mark_failed.py
@@ -2,9 +2,13 @@ from __future__ import annotations
 
 import logging
 import uuid
+from collections import Counter
 from datetime import datetime, timezone
+from typing import TYPE_CHECKING, Mapping, Sequence, TypedDict, cast
 
 from django.db.models import Q
+from django.utils.text import get_text_list
+from django.utils.translation import gettext_lazy as _
 
 from sentry import features
 from sentry.issues.grouptype import MonitorIncidentType
@@ -18,6 +22,9 @@ from sentry.monitors.models import (
 )
 
 logger = logging.getLogger(__name__)
+
+if TYPE_CHECKING:
+    from django.utils.functional import _StrPromise
 
 
 def mark_failed(
@@ -93,8 +100,16 @@ def mark_failed(
         return mark_failed_no_threshold(failed_checkin)
 
 
+class SimpleCheckIn(TypedDict):
+    id: int
+    date_added: datetime
+    status: int
+
+
 def mark_failed_threshold(
-    failed_checkin: MonitorCheckIn, failure_issue_threshold: int, received: datetime | None
+    failed_checkin: MonitorCheckIn,
+    failure_issue_threshold: int,
+    received: datetime | None,
 ):
     from sentry.signals import monitor_environment_failed
 
@@ -103,7 +118,7 @@ def mark_failed_threshold(
     # check to see if we need to update the status
     if monitor_env.status in [MonitorStatus.OK, MonitorStatus.ACTIVE]:
         if failure_issue_threshold == 1:
-            previous_checkins = [
+            previous_checkins: list[SimpleCheckIn] = [
                 {
                     "id": failed_checkin.id,
                     "date_added": failed_checkin.date_added,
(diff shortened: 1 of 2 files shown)


src/sentry/monitors/logic/mark_failed.py
```diff
@@ -304,6 +326,42 @@ def create_issue_platform_occurrence(
...
+    )
+
+    status_counts = Counter(
+        human_names[cast(int, checkin["status"])]
+        for checkin in failed_checkins
+        if checkin["status"] in human_names.keys()
+    )
+
+    if sum(status_counts.values()) == 1:
+        return _("A %(status)s check-in was detected") % {"status": list(status_counts.keys())[0]}
```
gaprl, line 352: Can we say "**An** error check-in was detected"?

gaprl (changes requested):

src/sentry/monitors/logic/mark_failed.py
evanpurkhiser: 👍

evanpurkhiser: Done

src/sentry/monitors/logic/mark_failed.py
```diff
@@ -304,6 +326,42 @@ def create_issue_platform_occurrence(
...
+    human_names: Mapping[int, str] = dict(
+        [
+            (CheckInStatus.ERROR, "error"),
+            (CheckInStatus.MISSED, "missed"),
+            (CheckInStatus.TIMEOUT, "timeout"),
+        ]
+    )
+
+    status_counts = Counter(
+        human_names[cast(int, checkin["status"])]
```
JoshFerge, line 346: why do we need to cast this?

src/sentry/monitors/logic/mark_failed.py
```diff
@@ -304,6 +326,42 @@ def create_issue_platform_occurrence(
...
 
+def get_failure_reason(failed_checkins: Sequence[SimpleCheckIn]):
+    """
+    Builds a humam readible strong from a list of failed check-ins.
+
+    "3 missed check-ins detected"
+    "2 missed check-ins, 1 timeout check-in and 1 error check-in were detected"
+    "A failed check-in was detected"
+    """
+    human_names: Mapping[int, str] = dict(
```
JoshFerge, line 337: why declare it like this instead of the more common style?

```
human_names: Mapping[int, str] = {
    CheckInStatus.ERROR: "error",
    CheckInStatus.MISSED: "missed",
    CheckInStatus.TIMEOUT: "timeout",
}
```

JoshFerge, line 337: also nit, but do we have to declare it per function? or is it constant?

evanpurkhiser: lol doy, some refactoring got me here

evanpurkhiser: can be constant 👍

src/sentry/monitors/logic/mark_failed.py
evanpurkhiser:
```
"CheckInStatus" is incompatible with "int"
```

It's because

```
    human_names: Mapping[int, str] = dict(
        [
            (CheckInStatus.ERROR, "error"),
            (CheckInStatus.MISSED, "missed"),
            (CheckInStatus.TIMEOUT, "timeout"),
        ]
    )
```

I would type this as 

```
    human_names: Mapping[CheckInStatus, str] = dict(
        [
            (CheckInStatus.ERROR, "error"),
            (CheckInStatus.MISSED, "missed"),
            (CheckInStatus.TIMEOUT, "timeout"),
        ]
    )
```

But that produces

```
 Type parameter "_KT@Mapping" is invariant, but "Literal[2, 4, 5]" is not the same as "CheckInStatus"
```

src/sentry/monitors/logic/mark_failed.py
```diff
@@ -304,6 +326,41 @@ def create_issue_platform_occurrence(
     )
 
 
+HUMAN_FAILURE_STATUS_MAP: Mapping[int, str] = {
```
JoshFerge, line 329: i don't think this type is correct? `CheckInStatus` is an enum not an `IntEnum`?

evanpurkhiser: It's not an enum

evanpurkhiser: https://github.com/getsentry/sentry/blob/ac80048be29f718d4caf7000034be7ac19f0e8f3/src/sentry/monitors/models.py#L125

src/sentry/monitors/logic/mark_failed.py
```diff
@@ -304,6 +326,41 @@ def create_issue_platform_occurrence(
...
+    "A failed check-in was detected"
+    """
+    status_counts = Counter(
+        HUMAN_FAILURE_STATUS_MAP[cast(int, checkin["status"])]
+        for checkin in failed_checkins
+        if checkin["status"] in HUMAN_FAILURE_STATUS_MAP.keys()
+    )
+
+    if sum(status_counts.values()) == 1:
+        return _("An %(status)s check-in was detected") % {"status": list(status_counts.keys())[0]}
```
JoshFerge, line 351: not familiar but will the string logic here correctly do `A` vs 'An`?

evanpurkhiser: good point. Let me see what django has for a e i o and u and sometimes y

JoshFerge: worth adding a test for that

JoshFerge (approved):
would try and remove the casts, but other than that lgtm besides the A vs An bit

evanpurkhiser:
Managed to get rid of most of the casts, the problem was the `SimpleCheckIn` was not typing it's status correctly. Indeed status is just a `int`

gaprl (approved):

PR #70789, opened 2024-05-13
feat(related_issues): Create separate sections for each relation type

armenzg:
In https://github.com/getsentry/sentry/pull/70504 I added support for the type parameter which allows getting the data for same-root issues vs trace-connected issues separately.

This UI change allows each related issues section to fetch and render the data independently, thus, making the page load faster.

These changes are required to support fetching trace-connected issues from the Issues Details page without also fetching same-root-connected data.


static/app/views/issueDetails/groupRelatedIssues/index.tsx
```diff
@@ -19,137 +19,136 @@ type RouteParams = {
...
 
+  const traceMeta = relationType === 'trace_connected' ? relatedIssues?.meta : undefined;
+  const issues = relatedIssues?.data ?? [];
+  const query = `issue.id:[${issues}]`;
   // project=-1 allows ensuring that the query will show issues from any projects for the org
   // This is important for traces since issues can be for any project in the org
   const baseUrl = `/organizations/${orgSlug}/issues/?project=-1`;
+  let title;
+  let linkToTrace;
+  let openIssuesButton;
```
scttcper, line 84: defining these as null and giving them a type would prevent them being used before defined and make their type not "any"
```suggestion
  let title: React.ReactNode = null;
  let linkToTrace = null;
  let openIssuesButton = null;
```

static/app/views/issueDetails/groupRelatedIssues/index.tsx
```diff
@@ -19,137 +19,136 @@ type RouteParams = {
...
-                  />
-                </div>
-              ) : (
-                <small>{t('No trace-connected related issues were found.')}</small>
-              )}
-            </HeaderWrapper>
-          </div>
+          <TextButtonWrapper>
+            {linkToTrace ?? null}
+            {openIssuesButton ?? null}
```
scttcper, line 136: could clean up these too then
```suggestion
            {linkToTrace}
            {openIssuesButton}
```

scttcper (approved):

PR #70794, opened 2024-05-13
snuba(devexp): Migrate meta queries to use storage direct queries

enochtangg:
### Overview
Now that snuba supports querying a storage directly (as opposed to querying an entity), meta queries are a great candidate to take advantage of this. All meta queries are executed through the metrics layer, therefore these changes will apply to any product (right now just DDM) running meta queries. The main change is switching out `Entity` for `Storage` in the `match` clause of the query. 

### Testing
This change is guarded behind the sentry option: `sentry-metrics.metrics-layer.use-storage-direct-meta-queries` and will be slowly rolled out + monitored.


src/sentry/snuba/metrics_layer/query.py
```diff
@@ -584,19 +586,34 @@ def _query_meta_table(
...
-        Query(Entity("generic_metrics_counters_meta"))
-        .set_select([Column("project_id"), Column(column_name)])
-        .set_groupby([Column("project_id"), Column(column_name)])
-        .set_where(conditions)
-        .set_orderby(
-            [
-                OrderBy(Column("project_id"), Direction.ASC),
-                OrderBy(Column(column_name), Direction.ASC),
-            ]
+    if in_random_rollout("sentry-metrics.metrics-layer.use-storage-direct-meta-queries"):
```
volokluev, line 589: where you able to run the CI tests using the storage-direct-meta-queries on?

volokluev (approved):
Looks good. If we can run CI with this flag on that would be even better

src/sentry/snuba/metrics_layer/query.py
enochtangg: Tested on CI with flag enabled 👍

PR #70795, opened 2024-05-13
test(ui): Use built-in test providers

scttcper:
removes custom providers that are included in our test render function

```tsx
render(<Component />, {organization, context: routerContext})
```


static/app/views/releases/thresholdsList/thresholdGroupRows.spec.tsx
```diff
@@ -34,35 +33,21 @@ describe('ThresholdGroupRows', () => {
...
-  };
-
-  type RenderProps = ThresholdGroupRowsProps & {org: Organization};
-  const DEFAULT_PROPS: RenderProps = {
+  const DEFAULT_PROPS: ThresholdGroupRowsProps = {
     allEnvironmentNames: ['test'],
     project: ProjectFixture(),
     refetch: () => {},
     setTempError: () => {},
-    org: organization,
```
scttcper, line 53: just moving org to the 2nd argument since its not actually a prop on ThresholdGroupRows but is being passed in as DEFAULT_PROPS

evanpurkhiser (approved):

PR #70797, opened 2024-05-13
ref(rules): Handle improperly formatted rule

ceorourke:
One customer has a couple improperly formatted alert rules that are causing this error: https://sentry.sentry.io/issues/5245475486/. I've already edited the serializer so it's not possible to get into this state again (we don't save data with a value of "") and had support reach out to the customer to address this, but this issue keeps happening so we can just handle the empty string.


saponifi3d (approved):

saponifi3d:
nit: it'd be nice to add a test case for this in the future as well (since we've seen it come up in prod already)

ceorourke:
> nit: it'd be nice to add a test case for this in the future as well (since we've seen it come up in prod already)

I can add one, but there is no future possibility of this happening for new rules, it's just this one org's 2 rules.

PR #70798, opened 2024-05-13
fix(new-trace): Loading trace from sample txn.

Abdkhan14:
We walked away from spreading `location.query` to the trace route and pass every param we need as destructured props to `generateLinkToEventInTraceView`. We forgot to add `demo` to the list.


narsaynorath (approved):
lgtm, ran it locally and I can see the trace for a demo transaction

PR #70803, opened 2024-05-13
ref(screenload): fix display of platformSelector and feedback button

michellewzhang:
Before:
SCR-20240513-lksj

After:
SCR-20240513-loqh


narsaynorath (approved):
thanks! 🙏

PR #70805, opened 2024-05-13
docs(configurations): Add API and protocol blueprints

cmanallen:


src/sentry/configurations/docs/api.md
```diff
@@ -0,0 +1,72 @@
...
+| sample_rate        | number | A number value between 0 and 1.                       |
+| traces_sample_rate | number | A number value between 0 and 1.                       |
+| user_config        | any    | Arbitrary user supplied JSON.                         |
+
+- Response 200
+
+  ```json
+  {
+    "data": {
+      "id": 1,
```
ryan953, line 29: `id`?

cmanallen: The ID of the project whose configuration you're reading.  Should have been more clear in the table above.  We can remove it.

ryan953: Eh. Looks like it's just passed along. I'm unbothered.

src/sentry/configurations/docs/protocol.md
```diff
@@ -0,0 +1,87 @@
+# Remote Configuration Protocol
+
+Host: https://dsn_url.com # TODO!
```
billyvg, line 3: Have we figured out if this will be a new URL, or will we send as a new envelope type?

cmanallen: Same as the envelope.  I just haven't looked up its structure yet.

ryan953 (approved):

src/sentry/configurations/docs/api.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Response 200
+
+  ```json
+  {
+    "data": {
+      "id": "99aabf0dad1c48ad8e47e2a43969f312",
+      "sample_rate": 1.0,
+      "traces_sample_rate": 0.5,
+      "user_config": {
```
Jesse-Box, line 34: Probably another dumb question,

I'm using the term “rule” to describe the action when a user defines configuration in our UI.

Using `"sample_rate": 1.0` as an example, when a user wants to define a custom config rule, will each rule expect a single key and a single value? Or could one rule have multiple keys and values pointing to variables inside the users' application?

I'm trying to understand the mental model here

src/sentry/configurations/docs/api.md
```diff
@@ -0,0 +1,78 @@
...
+Retrieve the DSN's configuration.
+
+**Attributes**
+
+| Column             | Type   | Description                     |
+| ------------------ | ------ | ------------------------------- |
+| id                 | string | Client key.                     |
+| sample_rate        | number | A number value between 0 and 1. |
+| traces_sample_rate | number | A number value between 0 and 1. |
+| user_config        | any    | Arbitrary user supplied JSON.   |
```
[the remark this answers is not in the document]
cmanallen: Since this is a JSON document, the `any` type in this context means "any valid JSON".

src/sentry/configurations/docs/api.md
cmanallen: All of the keys are hard-coded.  There will be a form with these three hard-coded inputs.  If we want more inputs we would hard-code more form elements and update the protocol.

User-defined key, value pairs are not scoped yet and will come after the initial remote-configuration work.

Jesse-Box: So, am I hearing that if a user wants to define any custom config for their application, it will be defined by a single rule?

cmanallen: For the initial internal release that is true.  For the private alpha we will allow the user to define multiple rules.

PR #70806, opened 2024-05-13
chore(hc): Enable stronger typed hybrid cloud tests

mrduncan:
Enables stronger typing for hybrid cloud tests since they were only missing return type annotations in almost all cases which are trivial to fix here.


tests/sentry/hybridcloud/test_organization.py
```diff
@@ -60,7 +60,9 @@ def basic_filled_out_org() -> tuple[Organization, list[User]]:
     return org, [owner, other_user]
 
 
-def parameterize_with_orgs(f: Callable):
+def parameterize_with_orgs(
+    f: Callable[[Callable[[], tuple[Organization, list[User]]]], None]
+) -> Callable[[Callable[[], tuple[Organization, list[User]]]], None]:
```
markstory, line 65:
```suggestion
WithOrgsCallback = Callable[[Callable[[], tuple[Organization, list[User]]]], None]

def parameterize_with_orgs(f: WithOrgsCallback) -> WithOrgsCallback:
```

The duplicate type could be extracted 🤷

markstory (approved):

tests/sentry/hybridcloud/test_caching.py
```diff
@@ -136,15 +137,15 @@ def reader():
                 yield from CacheBackend.set_cache(shared_key, copied_local_value, version)
                 last_length = len(copied_local_value)
 
-    def writer():
+    def writer() -> Iterator[None]:
```
asottile-sentry, line 140: I would be more specific and use `Generator[None, None, None]` though it doesn't matter here I don't think

tests/sentry/hybridcloud/test_caching.py
```diff
@@ -136,15 +137,15 @@ def reader():
...
         nonlocal true_value
         while True:
             for i in range(5):
                 yield
             true_value += "a"
             yield from CacheBackend.delete_cache(shared_key, SiloMode.REGION)
 
-    def cache_death_event():
+    def cache_death_event() -> Iterator[None]:
```
asottile-sentry, line 148: ""

tests/sentry/hybridcloud/test_organization.py
```diff
@@ -60,7 +60,10 @@ def basic_filled_out_org() -> tuple[Organization, list[User]]:
     return org, [owner, other_user]
 
 
-def parameterize_with_orgs(f: Callable):
+WithOrgsCallback = Callable[[Callable[[], tuple[Organization, list[User]]]], None]
+
+
+def parameterize_with_orgs(f: WithOrgsCallback) -> WithOrgsCallback:
     return pytest.mark.parametrize("org_factory", [pytest.param(basic_filled_out_org)])(f)
```
asottile-sentry, line 67: even easier than this:

```python
parameterize_with_orgs = pytest.mark.parametrize("org_factory", [pytest.param(basic_filled_out_org)])
```

if you want to stick with the function (I wouldn't recommend it) there's two easier ways to do this:

1. preserve with `ParamSpec`
```python
P = ParamSpec("P")
R = TypeVar("R")
def parametrize_with_orgs(f: Callable[P, R]) -> Callable[P, R]:
```
2. preserve type with a Callble TypeVar

```python
CallableT = TypeVar("CallableT", bound=Callable[..., object])
def parametrize_with_orgs(f: CallableT) -> CallableT:
```

tests/sentry/hybridcloud/test_organization.py
```diff
@@ -76,23 +79,23 @@ def order_things_by_id(a: list[Any], b: list[Any]) -> None:
     a.sort(key=lambda i: find_ordering(b_ids, i.id))
 
 
-def assert_for_list(a: list[Any], b: list[Any], assertion: Callable) -> None:
+def assert_for_list(a: list[Any], b: list[Any], assertion: Callable[[Any, Any], None]) -> None:
     assert len(a) == len(b)
     order_things_by_id(a, b)
```
asottile-sentry, line 84: probably not your issue but I would maybe rewrite this as `a: list[HasId], b: list[HasId]` and instead of `order_things_by_id` and `find_ordering` I would do `.sort(lambda item: item.id)` on both and compare that way.

would probably also want

```python
class HasId(Protocol):
    @property
    def id(self) -> int: ...
```

asottile-sentry (approved):
fixnship and/or considernship

tests/sentry/hybridcloud/test_organization.py
mrduncan: `parameterize_with_orgs = pytest.mark.parametrize("org_factory", [pytest.param(basic_filled_out_org)])` makes this so much cleaner - thank you!

tests/sentry/hybridcloud/test_organization.py
mrduncan: This one turned out be more of a can of worms than expected, going to leave it as-is for now since it's just a test assertion

PR #70807, opened 2024-05-13
fix(feedback): move call for get_latest_event

michellewzhang:
`get_latest_event` should be called for every `item` in `item_list`


aliu39 (approved):

aliu3ntry:
Quick question: the size of a feedback's group is always 1 right?

PR #70810, opened 2024-05-13
metrics(feedback): Track source emitted by SDK fxs

aliu39:
This is a useful property we're not tracking in metrics -- right now we assume all NEW_FEEDBACK_ENVELOPE's are from the widget, but they could be sent by the public api too.

Code refs:
[source types](https://github.com/getsentry/sentry-javascript/blob/d651f1319c95867be03c55f95c5878771c1e4c26/packages/feedback/src/constants/index.ts#L26)
[widget onSubmit](https://github.com/getsentry/sentry-javascript/blob/d651f1319c95867be03c55f95c5878771c1e4c26/packages/feedback/src/modal/components/Form.tsx#L124)
[sendFeedback (api)](https://github.com/getsentry/sentry-javascript/blob/d651f1319c95867be03c55f95c5878771c1e4c26/packages/feedback/src/core/sendFeedback.ts#L28)


aliu3ntry:
Unsure about naming - could be client_type, client_source_type..

src/sentry/feedback/usecases/create_feedback.py
```diff
@@ -257,7 +257,10 @@ def create_feedback_issue(event, project_id: int, source: FeedbackCreationSource
         auto_ignore_spam_feedbacks(project, issue_fingerprint)
     metrics.incr(
         "feedback.create_feedback_issue.produced_occurrence",
-        tags={"referrer": source.value},
+        tags={
+            "referrer": source.value,
+            "client_source": event["contexts"]["feedback"].get("source"),
```
michellewzhang, line 262: is `event["contexts"]["feedback"].get("source")` different from `source.value`?

aliu3ntry: Yeah -- the feedback context is SDK-defined and passed from there. Think source is required, but using .get in case of typescript 

The `source` in this function is a FeedbackCreationSource enum, which is chosen by Sentry backend before calling `create_feedback_issue`

michellewzhang: ahhh gotcha, makes sense

michellewzhang (approved):

PR #70811, opened 2024-05-13
Sync Mode drift detection 

kneeyo1:
In Sync mode, we do not want options that are not called in configoptions but `set_on_disk` to show up as drift.

Some options have the `prioritize_disk` flag and have values on the db and the disk. 
configoptions should only update these options if they are not set_on_disk. Right now, if it is set_on_disk but also set in the db, we write it as drift. This PR adds a check for that.


src/sentry/runner/commands/configoptions.py
```diff
@@ -273,11 +273,16 @@ def sync(ctx: click.Context) -> None:
                                 )
                                 presenter_delegator.flush()
                                 raise
-
                         presenter_delegator.unset(opt.name)
                     else:
-                        presenter_delegator.drift(opt.name, "")
-                        drift_found = True
+                        if (
```
fpacifici, line 278: Please add tests if you change the logic 
https://github.com/getsentry/sentry/blob/master/tests/sentry/runner/commands/test_configoptions.py

fpacifici (commented):
I think I am missing something here. 
Are you saying that we have options that are set on disk and also on the DB to a value that is not the default value for the option? 

This seems something that should not happen as we do not allow an option to be updated on DB if it is set on disk. 
This could only have happened if the option was already set to a value on the DB (different than the default) and then set on disk to something different again. 

But in that case it means the option cannot be updated with any channel, we should not hide this scenario which is broken.

kneeyo1:
> I think I am missing something here. Are you saying that we have options that are set on disk and also on the DB to a value that is not the default value for the option?
> 
> This seems something that should not happen as we do not allow an option to be updated on DB if it is set on disk. This could only have happened if the option was already set to a value on the DB (different than the default) and then set on disk to something different again.
> 
> But in that case it means the option cannot be updated with any channel, we should not hide this scenario which is broken.

We have options that are set on disk. In the db, they are set to their default value. 
This means that they exist on the DB, but do not read from it, they are still reading the disk value.

src/sentry/runner/commands/configoptions.py
```diff
@@ -273,10 +273,9 @@ def sync(ctx: click.Context) -> None:
                                 )
                                 presenter_delegator.flush()
```
kneeyo1, line 274: above on line 244, I believe we properly check if an option is already set on disk before applying 

https://github.com/getsentry/sentry/blob/fix/options-on-disk/src/sentry/options/manager.py#L188-L200

src/sentry/runner/commands/configoptions.py
```diff
@@ -273,11 +273,18 @@ def sync(ctx: click.Context) -> None:
...
                                 raise
-
                         presenter_delegator.unset(opt.name)
                     else:
-                        presenter_delegator.drift(opt.name, "")
-                        drift_found = True
+                        if (
+                            options.can_update(
```
kneeyo1, line 279: if an option is defined in the automator file, and it is set on disk, we check for it here https://github.com/getsentry/sentry/blob/fix/options-on-disk/src/sentry/options/manager.py#L188-L200

otherwise, if it is set on disk but not defined in the automator, since options.get() and options.isset() will still return true,(even if the options DNE in the DB) we again check to see if the option is on disk. if it is, we continue.

fpacifici (approved):

PR #70812, opened 2024-05-13
feat(snuba-search): snuba group dataset search handle additional columns p2

scefali:
This PR handles additional filters for the snuba-heavy search:

* `issue:`
* `project:`
* `date:`
* `first-release`
* `has:x`
* Generic tags like `server:`

I'm attempting to use as many pieces from existing search code as possible which is why I'm leveraging the discover query builder as well as `format_search_filter`. It's kind of a Frankenstein monster of composed parts, however.

### DIFF
--- a/src/sentry/search/events/builder/discover.py
+++ b/src/sentry/search/events/builder/discover.py
@@ -1322,6 +1322,9 @@ class BaseQueryBuilder:
                 1,
             )
         else:
+            # pull out the aliased expression if it exists
+            if isinstance(lhs, AliasedExpression):
+                lhs = lhs.exp
             condition = Condition(lhs, Op(search_filter.operator), value)
 
         if is_null_condition:
--- a/src/sentry/search/snuba/executors.py
+++ b/src/sentry/search/snuba/executors.py
@@ -1232,6 +1232,9 @@ class GroupAttributesPostgresSnubaQueryExecutor(PostgresSnubaQueryExecutor):
                 raw_column = map_field_name_from_format_search_filter(lhs[1][0])
                 rhs = [query_builder.resolve_column(raw_column)]
                 if len(lhs[1]) > 1:
+                    # example item here: [['ifNull', ['date', "''"]], '>=', 1715707188000]
+                    # which has lhs ['ifNull', ['date', "''"]]
+                    # we need this to become Function('ifNull', [Column('date', entity), ''])
                     rhs.append(lhs[1][1])
                 lhs = Function(lhs[0], rhs)
 
--- a/src/sentry/search/snuba/executors.py
+++ b/src/sentry/search/snuba/executors.py
@@ -57,7 +57,6 @@ from sentry.models.organization import Organization
 from sentry.models.project import Project
 from sentry.models.team import Team
 from sentry.models.user import User
-from sentry.rules.conditions.event_attribute import ATTR_CHOICES
 from sentry.search.events.builder.discover import UnresolvedQuery
 from sentry.search.events.datasets.discover import DiscoverDatasetConfig
 from sentry.search.events.filter import convert_search_filter_to_snuba_query, format_search_filter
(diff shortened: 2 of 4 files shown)


src/sentry/search/events/builder/discover.py
```diff
@@ -1322,6 +1322,9 @@ def default_filter_converter(
                 1,
             )
         else:
+            # pull out the aliased expression if it exists
+            if isinstance(lhs, AliasedExpression):
```
JoshFerge, line 1326: why do we have to do this? will this affect the other discover search uses?

src/sentry/search/snuba/executors.py
```diff
@@ -1215,19 +1213,39 @@ def get_basic_event_snuba_condition(
...
+                # need to convert dates to timestamps
+                if lhs[1][0] == "date":
+                    conver_value_to_date = True
+                # right now we are assuming lhs looks like ['isNull', ['user']]
+                # if there are more complex expressions we will need to handle them
+                raw_column = map_field_name_from_format_search_filter(lhs[1][0])
+                rhs = [query_builder.resolve_column(raw_column)]
+                if len(lhs[1]) > 1:
+                    rhs.append(lhs[1][1])
+                lhs = Function(lhs[0], rhs)
```
JoshFerge, line 1236: why do we have to do this? looks weird

scefali: @JoshFerge we have to deal with nested arrays

src/sentry/search/events/builder/discover.py
scefali: @JoshFerge it would just error out. I am pretty certain it doesn't impact discover or else we'd see the error about `AliasedExpression`

src/sentry/search/snuba/executors.py
JoshFerge: got it, mind adding an example in the comment above?

src/sentry/search/snuba/executors.py
```diff
@@ -1215,19 +1213,39 @@ def get_basic_event_snuba_condition(
...
-                raw_column = ATTR_CHOICES.get(column_name)
-                column_name = raw_column.value.event_name
-
-            column = Column(column_name, joined_entity)
+            lhs = item[0]
+            if isinstance(lhs, str):
+                raw_column = map_field_name_from_format_search_filter(lhs)
+                lhs = query_builder.resolve_column(raw_column)
+            else:
+                # need to convert dates to timestamps
```
JoshFerge, line 1227: why do we need to do this timestamp business?

scefali: @JoshFerge because we need to convert from a UNIX timestamp to datetime

src/sentry/search/snuba/executors.py
scefali: @JoshFerge added

src/sentry/search/snuba/executors.py
JoshFerge: why? do we do this in our other query executors?

scefali: @JoshFerge I realize now that we already handle the date filter with `calculate_start_end` so I can remove that special logic, we just skip that filter

src/sentry/search/events/builder/discover.py
JoshFerge: why does discover not need this previously?

src/sentry/search/snuba/executors.py
```diff
@@ -1681,11 +1697,11 @@ def query(
                     else:
                         raise InvalidQueryForExecutor(f"Invalid clause {clause}")
                 else:
-                    where_conditions.append(
-                        self.get_basic_event_snuba_condition(
-                            search_filter, joined_entity, organization.id, project_ids, environments
-                        )
+                    conditon = self.get_basic_event_snuba_condition(
```
JoshFerge, line 1700: nit: `conditon` -> `condition`

src/sentry/search/events/builder/discover.py
scefali: @JoshFerge It only fails on the `project_id` input that's used as a filter. It seems like we only use this function `default_filter_converter` on small subset of filter types currently even though it's written to accommodate a lot more. I don't really know the history here and the code is convoluted. My guess is this function used to be used for more things but at some point only got used for a subset of possible filters.

JoshFerge: got it. mind just adding more info to that comment then?

JoshFerge (approved):

PR #70816, opened 2024-05-13
feat(performance): queues module destination search

edwardgou-sentry:
Updates the search bar in the Queues module landing page to fully function:
- Search bar placeholder text has been updated
- Search bar applies a substring search on the `messaging.destination.name` tag/column
- Destinations table was updated to respond to search.
- Charts were updated to no longer directly get fetch destination from the query params. Instead this is now passed as a prop. This was needed because the charts on the landing page should not respond to the search bar (same behaviour as other modules).

There's some "sort" and "time spent" placeholders in this pr that doesn't need to be included (sorry). I have a subsequent pr to complete those, but I can pull out of this pr if needed.


static/app/views/performance/queues/queuesLandingPage.tsx
```diff
@@ -21,12 +26,45 @@ import {ModulePageProviders} from 'sentry/views/performance/modulePageProviders'
...
+  const query = useLocationQuery({
+    fields: {
+      destination: decodeScalar,
+      [QueryParameterNames.DOMAINS_SORT]: decodeScalar,
+    },
+  });
+
+  const sort =
+    decodeSorts(query[QueryParameterNames.DOMAINS_SORT]).filter(isAValidSort).at(0) ??
+    DEFAULT_SORT;
```
edwardgou-sentry, line 52: This is placeholder for now as I'm working on completing the table sorting. (`DOMAINS_SORT` will be replaced to something more appropriate).

Can remove from this pr if needed.

gggritso (approved):

static/app/views/performance/queues/queuesTable.tsx
```diff
@@ -73,17 +74,34 @@ const COLUMN_ORDER: Column[] = [
...
+  'messaging.destination.name',
+  'time_spent_percentage()',
+] as const;
+
+type ValidSort = Sort & {
+  field: (typeof SORTABLE_FIELDS)[number];
+};
+
+export function isAValidSort(sort: Sort): sort is ValidSort {
+  return (SORTABLE_FIELDS as unknown as string[]).includes(sort.field);
```
narsaynorath, line 87: Coercing it to unknown seems a little weird to me. Did a quick search and it seems like we can do this

```suggestion
  return (SORTABLE_FIELDS as ReadonlyArray<string>).includes(sort.field);
```

which gives us a touch more type safety by checking if SORTABLE_FIELDS is compatible with the assertion

static/app/views/performance/queues/queuesLandingPage.tsx
```diff
@@ -21,12 +26,45 @@ import {ModulePageProviders} from 'sentry/views/performance/modulePageProviders'
...
+        ...location.query,
+        destination: newDestination === '' ? undefined : newDestination,
+        [QueryParameterNames.DESTINATIONS_CURSOR]: undefined,
+      },
+    });
+  };
+
+  const wildCardDestinationFilter = query.destination
+    ? `*${escapeFilterValue(query.destination)}*`
+    : undefined;
```
narsaynorath, line 67: Why does this need a double ended wildcard? Maybe we should explain that a bit more in a comment?

narsaynorath (approved):
Just two small comments, looks good to me

static/app/views/performance/queues/queuesTable.tsx
edwardgou-sentry: Thanks! Updated!

static/app/views/performance/queues/queuesLandingPage.tsx
edwardgou-sentry: Added a comment! It's to turn the users input into a substring search. `wildCardDestinationFilter` is passed to the `QueuesTable` component which searches based on this value

PR #70818, opened 2024-05-13
chore(hybridcloud) Enable stricter types on more modules

markstory:
Move a few more modules into the stricter allow list.


tests/sentry/types/test_region.py
```diff
@@ -68,11 +69,11 @@ class RegionDirectoryTest(TestCase):
 
     @staticmethod
     @contextmanager
-    def _in_global_state(directory: RegionDirectory):
+    def _in_global_state(directory: RegionDirectory) -> Any:
```
asottile-sentry, line 72: Generator[None, None, None]

asottile-sentry (approved):

PR #70820, opened 2024-05-13
[ALRT-39] fix comparison query timestamp column

nhsiehgit:
Addresses https://sentry.sentry.io/issues/4980183622/?project=1

Fixes ALRT-39

### DIFF
--- a/src/sentry/incidents/subscription_processor.py
+++ b/src/sentry/incidents/subscription_processor.py
@@ -11,7 +11,7 @@ from django.conf import settings
 from django.db import router, transaction
 from django.utils import timezone
 from sentry_redis_tools.retrying_cluster import RetryingRedisCluster
-from snuba_sdk import Column, Condition, Limit, Op
+from snuba_sdk import Column, Condition, Entity, Limit, Op
 
 from sentry import features
 from sentry.constants import CRASH_RATE_ALERT_AGGREGATE_ALIAS, CRASH_RATE_ALERT_SESSION_COUNT_ALIAS
@@ -219,8 +219,12 @@ class SubscriptionProcessor:
             time_col = ENTITY_TIME_COLUMNS[get_entity_key_from_query_builder(query_builder)]
             query_builder.add_conditions(
                 [
-                    Condition(Column(time_col), Op.GTE, start),
-                    Condition(Column(time_col), Op.LT, end),
+                    Condition(
+                        Column(time_col, entity=Entity("events", alias="events")), Op.GTE, start
+                    ),
+                    Condition(
+                        Column(time_col, entity=Entity("events", alias="events")), Op.LT, end
+                    ),
                 ]
             )
             query_builder.limit = Limit(1)
--- a/src/sentry/incidents/subscription_processor.py
+++ b/src/sentry/incidents/subscription_processor.py
@@ -11,7 +11,7 @@ from django.conf import settings
 from django.db import router, transaction
 from django.utils import timezone
 from sentry_redis_tools.retrying_cluster import RetryingRedisCluster
-from snuba_sdk import Column, Condition, Entity, Limit, Op
+from snuba_sdk import Column, Condition, Limit, Op
 
 from sentry import features
 from sentry.constants import CRASH_RATE_ALERT_AGGREGATE_ALIAS, CRASH_RATE_ALERT_SESSION_COUNT_ALIAS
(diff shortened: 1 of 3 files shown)


ceorourke:
Nice, would it be too much to ask to add a test for `get_comparison_aggregation_value`?

src/sentry/incidents/subscription_processor.py
```diff
@@ -215,8 +215,12 @@ def get_comparison_aggregation_value(
             time_col = ENTITY_TIME_COLUMNS[get_entity_key_from_query_builder(query_builder)]
             query_builder.add_conditions(
                 [
-                    Condition(Column(time_col), Op.GTE, start),
-                    Condition(Column(time_col), Op.LT, end),
+                    Condition(
+                        Column(time_col, entity=Entity("events", alias="events")), Op.GTE, start
+                    ),
+                    Condition(
+                        Column(time_col, entity=Entity("events", alias="events")), Op.LT, end
```
kylemumma, line 222: are you sure this is the right entity? is this function strictly for the events? I see other functions like "get entity key from..." which make me think it might not always be the same

nhsiehgit: @ceorourke would you know?

volokluev:
This needs a test. If you can't reproduce the issue in a test then how will you know that it was fixed?

nhsiehgit:
working on the tests now

src/sentry/incidents/subscription_processor.py
ceorourke: I do not, the only time I used this I knew which entity I was using for sure.

src/sentry/incidents/subscription_processor.py
```diff
@@ -212,11 +212,20 @@ def get_comparison_aggregation_value(
                     "end": end,
                 },
             )
-            time_col = ENTITY_TIME_COLUMNS[get_entity_key_from_query_builder(query_builder)]
+            entity_key = get_entity_key_from_query_builder(query_builder)
```
ceorourke, line 215: I think you want to have an [Entity](https://github.com/getsentry/snuba-sdk/blob/65e490a481af00c81693dc590632bd0d9db4697f/snuba_sdk/entity.py#L16), not an [EntityKey](https://github.com/getsentry/sentry/blob/master/src/sentry/snuba/dataset.py#L63).

src/sentry/incidents/subscription_processor.py
```diff
@@ -212,11 +212,20 @@ def get_comparison_aggregation_value(
...
+            time_col = ENTITY_TIME_COLUMNS[entity_key]
             query_builder.add_conditions(
                 [
-                    Condition(Column(time_col), Op.GTE, start),
-                    Condition(Column(time_col), Op.LT, end),
+                    Condition(
+                        Column(time_col, entity=Entity(entity_key.value, alias="entity")),
```
ceorourke, line 220: Is the alias required here? I don't think you need to pass it, it's used later on when parsing the results (I think?)

src/sentry/incidents/subscription_processor.py
nhsiehgit: Yea - the `EntityKey.value` gets translated into the `Entity` below

src/sentry/incidents/subscription_processor.py
nhsiehgit: I asked @kylemumma that also. 🤔

ceorourke: Resolving because it is required

src/sentry/search/events/builder/errors.py
```diff
@@ -126,6 +127,16 @@ def get_snql_query(self) -> Request:
             tenant_ids=self.tenant_ids,
         )
 
+    def add_conditions(self, conditions: list[Condition]) -> None:
```
nhsiehgit, line 130: Oh interesting

So - we don't need to do anything in the subscription processor?
Was entity only necessary for the errors??

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -2045,6 +2046,26 @@ def test_comparison_alert_below(self):
             incident, [self.action], [(50.0, IncidentStatus.CLOSED, mock.ANY)]
         )
 
+    @with_feature("organizations:metric-alert-ignore-archived")
+    def test_is_unresolved_query(self):
```
wedamija, line 2050: Maybe `test_is_unresolved_comparison_query`?

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -2045,6 +2046,26 @@ def test_comparison_alert_below(self):
...
+        rule = self.comparison_rule_above
+        update_alert_rule(rule, query="(event.type:error) AND (is:unresolved)")
```
nhsiehgit, line 2052: Can we add a note here to call out how this explicit query affects the expected results?

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -2045,6 +2046,26 @@ def test_comparison_alert_below(self):
...
+        self.assert_trigger_counts(processor, trigger, 0, 0)
+        self.assert_no_active_incident(rule)
+        self.assert_trigger_does_not_exist(trigger)
+        self.assert_action_handler_called_with_actions(None, [])
+        self.metrics.incr.assert_has_calls(
+            [
+                call("incidents.alert_rules.skipping_update_comparison_value_invalid"),
+                call("incidents.alert_rules.skipping_update_invalid_aggregation_value"),
+            ]
+        )
```
wedamija, line 2067: Is it worth also including data in the comparison period like the other comparison tests do, to verify that it works ok?

tests/sentry/incidents/test_subscription_processor.py
nhsiehgit: eg. utilizes the `ErrorsQueryBuilder` and requires an entity or something

nhsiehgit (commented):
+1

wedamija (approved):

PR #70822, opened 2024-05-13
feat(insights): add analytics to insight sample panels

KevinL10:
Continuation of https://github.com/getsentry/sentry/pull/70769. Adds analytics to the span sample panel to track:
- How often the panels are opened
- How often spans are clicked
- How often “Try different samples” is clicked
- How often specific filters are interacted with

Modules to add analytics to:
- [x] HTTP
- [x] Database
- [x] Cache
- [x] Queues
- [x] Screen Loads
- [x] App Starts
- [x] Resources

Note: Web Vitals does not have a sample panel


static/app/views/starfish/components/tableCells/spanIdCell.tsx
```diff
@@ -14,6 +17,7 @@ interface Props {
 }
 
 export function SpanIdCell({
+  moduleName,
```
KevinL10, line 20: instead of adding a separate moduleName prop, consider inferring the module from path (`location.pathname`). helper function that maps paths to ModuleName enum

static/app/views/starfish/views/spans/selectors/actionSelector.tsx
```diff
@@ -105,8 +105,8 @@ const LABEL_FOR_MODULE_NAME: {[key in ModuleName]: ReactNode} = {
   cache: t('Action'),
   vital: t('Action'),
   queue: t('Action'),
-  screen: t('Action'),
-  startup: t('Action'),
+  screen_load: t('Action'),
+  app_start: t('Action'),
```
narsaynorath, line 109:
```suggestion
  screenLoad: t('Action'),
  appStart: t('Action'),
```
Also, maybe I'm missing context but what are these used for? 🤔

static/app/views/starfish/views/spans/selectors/domainSelector.tsx
```diff
@@ -173,8 +173,8 @@ const LABEL_FOR_MODULE_NAME: {[key in ModuleName]: ReactNode} = {
   cache: t('Domain'),
   vital: t('Domain'),
   queue: t('Domain'),
-  screen: t('Domain'),
-  startup: t('Domain'),
+  screen_load: t('Domain'),
+  app_start: t('Domain'),
```
narsaynorath, line 177:
```suggestion
  screenLoad: t('Domain'),
  appStart: t('Domain'),
```

Similar to my other comment, I'm not exactly sure what's being mapped here to use the `Domain` string

narsaynorath (approved):
Just some small nits and questions because I'm a little unfamiliar with some of the constants you modified

static/app/views/starfish/views/spans/selectors/domainSelector.tsx
KevinL10: These values are used as labels for some specific dropdowns in http and db modules (see "SQL Command" and "Table in screeshot). They aren't used for the remaining modules, but are required in this constant due to the `{[key in ModuleName]: ReactNode}` typing. I included default Action and Domain values for these remaining modules.

image

static/app/views/performance/mobile/appStarts/screenSummary/spanOperationTable.tsx
```diff
@@ -159,7 +160,15 @@ export function SpanOperationTable({
...
       return (
-        <Link to={`${pathname}?${qs.stringify(query)}`}>
+        <Link
+          onClick={() =>
+            trackAnalytics('performance_views.sample_spans.opened', {
+              organization,
+              source: ModuleName.APP_START,
+            })
+          }
+          to={`${pathname}?${qs.stringify(query)}`}
```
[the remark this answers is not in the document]
KevinL10: Yes, I think we can for all of the modules

static/app/views/starfish/views/spans/selectors/domainSelector.tsx
narsaynorath: I see, that's kind of annoying that you had to make this change. I say we can fix the typing here later so we don't have to update this list every time we add a new module but I'm fine with keeping it as-is for now

KevinL10: Agreed, will see if I can fix that in the follow up when I move instrumentation to the panel itself

PR #70824, opened 2024-05-13
feat(generic-metrics): Add dropped percentiles to aggregation options

ayirr7:
Refactoring the way we used an older option for enabling 10s granularity so its purpose is clearer.

Enable org-level and use case-level of disabling percentiles using the newly created Sentry options.


src/sentry/sentry_metrics/aggregation_option_registry.py
```diff
@@ -22,17 +23,41 @@ class TimeWindow(Enum):
...
+            AggregationOption.TEN_SECOND: TimeWindow.SEVEN_DAYS
+        }
+
+    for use_case in options.get("sentry-metrics.drop-percentiles.per-use-case"):
+        USE_CASE_AGG_OPTION[UseCaseID(use_case)] = {
+            AggregationOption.NO_PERCENTILE: TimeWindow.NINETY_DAYS
+        }
+
+
+def get_aggregation_options(mri: str, org_id: int) -> dict[AggregationOption, TimeWindow] | None:
```
nikhars, line 45: Since we are only able to write 1 aggregation option at the moment, what if we added a test case which validated that assumption. I am not exactly sure whether that constraint is enforced somewhere in the code or not. But it would be good to have the constrain in place to avoid potential problems in the near future.

ayirr7: Good point. Added some logic to our existing unit tests where there are a variety of aggregation options set up, which cover all the use cases etc. The unit test asserts that only 1 exists per metric bucket payload, though. 

Once we have multiple aggregation option support, we can remove and refactor these tests to communicate that change in expectations.

src/sentry/sentry_metrics/aggregation_option_registry.py
```diff
@@ -22,17 +23,41 @@ class TimeWindow(Enum):
...
+        USE_CASE_AGG_OPTION[UseCaseID(use_case)] = {
+            AggregationOption.DISABLE_PERCENTILES: TimeWindow.NINETY_DAYS
+        }
+
+
+def get_aggregation_options(mri: str, org_id: int) -> dict[AggregationOption, TimeWindow] | None:
+
+    # Set various aggregation options that
+    # are use case-wide aggregations
+    set_use_case_aggregation_options()
```
jan-auer, line 49: This function internally modifies a global, but this is done on every call of `get_aggregation_options`. The global isn't really needed therefore, it could be just a local variable. This would further allow you to simplify and inline the checks into a single `if`-chain below that would also more clearly show precedence of the options.

src/sentry/sentry_metrics/aggregation_option_registry.py
```diff
@@ -22,17 +23,41 @@ class TimeWindow(Enum):
...
+
     use_case_id: UseCaseID = extract_use_case_id(mri)
 
-    # We check first if the particular metric ID has a specified aggregation
-    if mri in METRIC_ID_AGG_OPTION:
-        return METRIC_ID_AGG_OPTION.get(mri)
+    # We check first if the org ID has disabled percentiles
+    if org_id in options.get("sentry-metrics.drop-percentiles.per-org"):
+        return {AggregationOption.DISABLE_PERCENTILES: TimeWindow.NINETY_DAYS}
```
jan-auer, line 55: There's still an issue in the logic. We need to be able to:
- Disable percentiles for an entire use case
- Override (re-enable) percentiles for a specific org in a specific use case. 

This option will allow us to disable percentiles for specific orgs, but not the inverse due to the fall-through. Ideally, we'd allow for both ways, but if you prefer to keep things simple we can also stick to opt-in for percentiles. What this means is we need two changes:

1. Allow to configure percentiles per usecase on org-level
2. Ensure that the org-level setting _always_ overrides the global one.

Initially, I was considering organization feature flags for this, but given that we better not call out to flagr in the ingest path, sentry options are the better choice.

ayirr7:
Will add more unit tests to check that the logic in `get_aggregation_options` is correct

src/sentry/sentry_metrics/aggregation_option_registry.py
ayirr7: Updated

src/sentry/sentry_metrics/aggregation_option_registry.py
ayirr7: The indexer already assumes a `dict` output and basically chooses the first element. So in any case, we are always guaranteed to have 1 aggregation option per payload. 

I can change the typing and indexer behavior in a subsequent PR, but for now I think I'd like to not have this PR touch too many files at once.

src/sentry/options/defaults.py
```diff
@@ -1260,6 +1260,15 @@
     flags=FLAG_AUTOMATOR_MODIFIABLE,
 )
 
+# Option to remove support for percentiles on a per-use case basis.
+# Add the use case to list to disable percentiles.
```
jan-auer, line 1264: _nit_: This isn't a list, the use case must be an entry with a list value (otherwise causing a crash). Should we make this more clear?

ayirr7: Good callout, thanks

jan-auer (approved):

jan-auer (approved):

PR #70827, opened 2024-05-13
api: add auth test endpoint

mdtro:
@JoshFerge discovered that we were seeing measurable amounts of TruffleHog requests hitting our project index endpoint. It's suspected this would cause a non-trivial amount of load on the backend datastores. At one point, we had 220k requests logged in a single minute from the TruffleHog user agent... The vast majority were rate-limited however.

This PR creates a really basic endpoint that returns a 200 if you are successfully authenticated with a user auth token, org auth token, or a session. If you're not authenticated, we'll return a 403. The endpoint specifically avoids pulling data about the User and returns no body.

The rate limit is quite high, but I did use a bit of math to find the median/average of requests per minute per IP using the TruffleHog user agent over the past 24 hours. It peaked around 10.5k requests (with about 8k of those rate rate limited). The average and median is right around 1,100-1,300 requests/min for those that weren't rate limited (including rate limited requests, it's closer to 4.7k/min).

The hope with this endpoint is to move these scanning tools that attempt to perform verification of the secret. This gives them an option to use something less impactful and to be good stewards of the service. 🙂 We can also encourage them to honor our rate limiting headers and backoff on the requests.


JoshFerge (commented):
is there a missing file?

JoshFerge (approved):
❤️

mdtro:
> is there a missing file?

Yep there sure was. I misnamed some files. Fixed now. :)

PR #70828, opened 2024-05-13
fix(query-builder): Improve the stability of token keys

malwilley:
Previously, the cursor would jump around when entering a free text portion due to the component remounting because the key changed. This makes some changes to avoid keys changing as much as possible:

- Updates token keys from `{type}:{cursor_location}` to `{type}:{type_index}`
- Renames space tokens to free_text since they are treated the same and it keeps the keys stable

### DIFF
--- a/static/app/components/searchQueryBuilder/index.spec.tsx
+++ b/static/app/components/searchQueryBuilder/index.spec.tsx
@@ -26,6 +26,10 @@ const MOCK_SUPPORTED_KEYS: TagCollection = {
 };
 
 describe('SearchQueryBuilder', function () {
+  afterEach(function () {
+    jest.restoreAllMocks();
+  });
+
   const defaultProps: ComponentProps<typeof SearchQueryBuilder> = {
     getTagValues: jest.fn(),
     initialQuery: '',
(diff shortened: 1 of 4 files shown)


static/app/components/searchQueryBuilder/index.spec.tsx
```diff
@@ -178,7 +182,13 @@ describe('SearchQueryBuilder', function () {
         screen.getByRole('combobox'),
         'some free text brow{ArrowDown}'
       );
+
+      // XXX(malwilley): SearchQueryBuilderInput updates state in the render
+      // function which causes an act warning despite using userEvent.click.
+      // Cannot find a way to avoid this warning.
+      jest.spyOn(console, 'error').mockImplementation(jest.fn());
```
malwilley, line 189: Updating the state in a useEffect makes this warning go away, but does cause other issues with having a stale value for a render cycle so I don't want to do that. I tried wrapping everything I could in act to no avail 😞 

Hoping I can eventually get to the bottom of this, but throwing in the towel for now

armenzg (approved):

scttcper (approved):
that act warning will haunt us forever

PR #70829, opened 2024-05-13
feat(api-idorslug): Updated Rest of Endpoints to use `organization_id_or_slug`

iamrajjoshi:
Final PR to update all the endpoints in `url.py` to use `organization_id_or_slug`.

Entire PR is literally updating url.py and then fix all the tests, specifically args of the tests.


isabellaenriquez (approved):
✨

PR #70831, opened 2024-05-13
ref(rules): Fix task name for delayed rule processing

ceorourke:
The task for processing delayed rules wasn't imported into `CELERY_IMPORTS` and the name was incorrect - this fixes both and adds a log message if the buffer hook registry isn't working as expected.

### DIFF
--- a/src/sentry/buffer/redis.py
+++ b/src/sentry/buffer/redis.py
@@ -73,6 +73,7 @@ class BufferHookRegistry:
         try:
             callback = self._registry[buffer_hook_event]
         except KeyError:
+            logger.info("buffer_hook_event.missing", extra={"key_name": buffer_hook_event.value})
             return False
 
         return callback(data)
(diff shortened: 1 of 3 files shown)


schew2381 (approved):

src/sentry/buffer/redis.py
```diff
@@ -73,6 +73,7 @@ def callback(self, buffer_hook_event: BufferHookEvent, data: RedisBuffer) -> boo
         try:
             callback = self._registry[buffer_hook_event]
         except KeyError:
+            logger.info("buffer_hook_event.missing", extra={"key_name": buffer_hook_event.value})
```
saponifi3d, line 76: nit: might be nice to make this an error log instead of an info log, and pass the exception.

saponifi3d (approved):

src/sentry/buffer/redis.py
ceorourke: 👍 I'll change it in a separate PR, really wanna get this out to do testing

PR #70832, opened 2024-05-14
perf: move more implementations to use orjson

anonrig:
Moves more implementations to use `orjson`

### DIFF
--- a/src/sentry/apidocs/build.py
+++ b/src/sentry/apidocs/build.py
@@ -1,12 +1,12 @@
 from typing import Any
 
-from sentry.utils import json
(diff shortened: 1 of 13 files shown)


iamrajjoshi (approved):
🚀 for my knowledge: what benefits do we get by switching?

anonrig:
> 🚀 for my knowledge: what benefits do we get by switching?

@iamrajjoshi I forgot the reference for the original issue: https://github.com/getsentry/sentry/issues/68903. To summarize, performance and JSON specification compliance :-)

PR #70833, opened 2024-05-14
perf: remove trace from sentry.utils.json

anonrig:
Starting a trace has a cost, and for micro-optimizations we should avoid it.

cc @ayirr7

### DIFF
--- a/src/sentry/db/models/fields/gzippeddict.py
+++ b/src/sentry/db/models/fields/gzippeddict.py
@@ -32,7 +32,7 @@ class GzippedDictField(TextField):
         try:
             if not value:
                 return {}
-            return json.loads(value, skip_trace=True)
+            return json.loads(value)
         except (ValueError, TypeError):
             if isinstance(value, str) and value:
                 try:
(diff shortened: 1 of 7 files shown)


src/sentry/db/models/fields/gzippeddict.py
```diff
@@ -32,7 +32,7 @@ def to_python(self, value):
         try:
             if not value:
                 return {}
-            return json.loads(value, skip_trace=True)
```
ayirr7, line 35: What does `skip_trace = True` do?

ayirr7, line 35: Shouldn't this stay `True` so that we skip creating spans?

anonrig: I removed from `loads()` function directly.

https://github.com/getsentry/sentry/pull/70833/files#diff-5c1440eab19b478087e2a08ff5238d26b56c916cc7baa1b481a9645cab9200abR131

ayirr7 (approved):

PR #67563, opened 2024-03-22
Alert rule project backfill migration

nhsiehgit:
Backfill migration to create `AlertRuleProject`s for all AlertRules

Update `create_alert_rule` method to create AlertRuleProject for all new alert rules

Add TODO's for removing projects from query subscription

migration:
Iterate through all QuerySubscriptions and create a new AlertRuleProject given the subscription.projects

PR #70835, opened 2024-05-14
(fix): Sentry-16D3 - Issue accessing the project on a query subscription

saponifi3d:
## Description
Error report: https://sentry.sentry.io/issues/4527458007/?project=1

Looks like this was missed after [we migrated](https://github.com/getsentry/sentry/pull/67563/files) from projects being stored on the alert_rule. This can now be removed because all projects should exist at the alert_rule level.

### DIFF
--- a/src/sentry/incidents/endpoints/organization_alert_rule_details.py
+++ b/src/sentry/incidents/endpoints/organization_alert_rule_details.py
@@ -319,11 +319,6 @@ class OrganizationAlertRuleDetailsEndpoint(OrganizationAlertRuleEndpoint):
             except Exception:
                 pass
 
-            # TODO - Cleanup Subscription Project Mapping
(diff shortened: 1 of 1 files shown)


nhsiehgit (approved):

schew2381 (approved):

saponifi3d:
> What does it mean if there is no project? The sentry issue existed because that happened, is it ok to just pass when it's not found?

If there's not a project associated with an alert, then the check for if the user has access to `None` project will fail, and the user will be shown an error message related to that.

Previously, the failure was happening because the alert rule's snuba query was trying to access a subscription that didn't exist on an activated alert. This happens with activated alerts because they delay creating a snuba subscription until the activation event occurs. By removing the change to lookup the project id from the subscription and directly looking at the alert, we can avoid looking at subscriptions that may not exist yet.

Issue #70152, opened 2024-05-02
Public Launch Flows


matejminar:
```[tasklist]
### Tasks
- [x] AM2 beta banner announcing that we will start asking for money on DATE
- [x] AM1 beta banner asking to update plan until DATE otherwise you'll lose access
- [x] AM1 metrics page experience (replace CTA on empty state with update plan)
```

matejminar:
Raw notes:
```
- AM2 Beta banner announcing that we will start asking for money on DATE
    - owner
    - member
      - HEADS UP design email to buy metric hours (or these members have access to billing)
- AM2 banner quota exceeded
  - owner
  - member
    - CRITICAL design email to buy metric hours (or these members have access to billing)

- AM1 Beta banner asking to upgrade plan until DATE otherwise you'll loose access
  - owner
  - member
    - email to update plan (or these members have access to billing)
- AM1 Beta email from Ale

- GENERIC AM1 metrics page experience (no access)
  - modal
    - owner, upgrade inline or go to settings?
    - member, send email (or these members have access to billing)

Q
- manage subscription - where should it go so that I can buy metric hours
- how to get over quota information so that we can show banner
  - 75%?
- update plan - where to go/what modal to show

addEventsCTA / gsBanner
```

matejminar:
Took these two items out of this task, they'll be tackled separately:
https://github.com/getsentry/sentry/issues/71094
and
https://github.com/getsentry/sentry/issues/71095

PR #70839, opened 2024-05-14
feat(metrics): public launch flow

obostjancic:
- Related to #70152


static/app/views/metrics/billing.tsx
```diff
@@ -0,0 +1,46 @@
...
+          return null;
+        }
+        return (
+          <Alert system type={subscriptionCTAProps.type}>
+            {subscriptionCTAProps.description}
+          </Alert>
+        );
+      }}
+    </MetricsSubscriptionCTAProvider>
+  );
```
matejminar, line 27: nit, we usually just call the hook and let getsentry do all the business logic of rendering or not rendering components

static/app/views/metrics/billing.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export function OnboardingPanelPrimaryActionButtons({children}) {
```
matejminar, line 30: same here, this should ideally be a hook that without getsentry defaults to whatever we want (in this case set up metrics) and getsentry overrides it if needed to update plan

matejminar (approved):

PR #70845, opened 2024-05-14
feat: Add a helper script for Celery troubleshooting

tonyo:
This is a helper script to aid in troubleshooting Celery-related activities.

Mostly for internal use right now, but if proved useful, we can make it a "sentry" subcommand.

The idea is that the script is started in interactive mode (either by calling "./bin/celery-helper.py" or via "python -i ./bin/celery-helper.py"), and then one can run commands in the interactive shell such as:
1. Listing all active tasks
2. Listing all task IDs for the given name
3. Revoking and terminating all tasks by name
...and others.

Additional docs around some basic operations: https://docs.celeryq.dev/en/latest/userguide/workers.html#inspecting-workers

### DIFF
--- a/bin/celery-helper.py
+++ b/bin/celery-helper.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python -i
+# A helper script for Celery that can be used to inspect/revoke specific tasks.
 #
 # NOTE: when started with "python", this script is meant to be used in "interactive" mode:
 #   python -i SCRIPT
@@ -6,7 +7,6 @@
 import enum
 import inspect as insp
 import sys
-from collections.abc import Generator
 
 control = None
 inspect = None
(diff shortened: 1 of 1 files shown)


bin/celery-helper.py
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env python -i
```
beezz, line 1: Any reason for not having this as a runner command to reuse the sentry initiation? e.g. https://github.com/getsentry/sentry/blob/master/src/sentry/runner/commands/queues.py

tonyo: This is very experimental for now, so still early to promote this to a runner command

tonyo:
Converted to draft; this will need more work since we also need to support a federated, non-clustered setup with multiple RMQ brokers.

PR #70846, opened 2024-05-14
ref: stronger type models.eventattachment

asottile-sentry:

### DIFF
--- a/src/sentry/models/eventattachment.py
+++ b/src/sentry/models/eventattachment.py
@@ -1,8 +1,10 @@
+from __future__ import annotations
+
 import mimetypes
 from dataclasses import dataclass
 from hashlib import sha1
 from io import BytesIO
-from typing import IO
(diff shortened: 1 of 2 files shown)


src/sentry/models/eventattachment.py
```diff
@@ -92,7 +96,7 @@ class Meta:
 
     __repr__ = sane_repr("event_id", "name")
 
-    def delete(self, *args, **kwargs):
+    def delete(self, *args: Any, **kwargs: Any) -> tuple[int, dict[str, int]]:
```
anonrig, line 99: isn't object better for this?

asottile-sentry: for parameters `object` wouldn't work because the underlying function it calls _is_ typed and so `object` wouldn't fit in a `bool` for instance

anonrig (approved):

PR #70853, opened 2024-05-14
ref: skip event pre-normalize through CanonicalKeyDict

asottile-sentry:
keys get normalized through the relay normalization

this gets the rest of the tests passing when changing CanonicalKeyDict to raise when legacy keys are passed -- the testsuite passes lots of `{"message": ...}` similar to current sdks!


armenzg (approved):

jan-auer (approved):
🤞 🍪

PR #70854, opened 2024-05-14
feat(traces-explorer): Return slice indexes instead of rounded timest…

Zylphrex:
…amps

Instead of returning the rounded timestamps, we're switching to returning the slice index that the frontend can use directly for rendering.


src/sentry/api/endpoints/organization_traces.py
```diff
@@ -968,32 +981,35 @@ def get_suggested_spans_query(
...
+    span_start: float,
+    span_end: float,
+    trace_range: TraceRange,
+) -> tuple[tuple[int, int], tuple[int, int]]:
     trace_start = trace_range["start"]
     trace_end = trace_range["end"]
 
-    bin_size = trace_range["min"]
+    num_slices = trace_range["slices"]
+    slice_size = trace_range["slice_size"]
```
shruthilayaj, line 993: nit: seems like it's redefined in line 996?

src/sentry/api/endpoints/organization_traces.py
```diff
@@ -968,32 +981,35 @@ def get_suggested_spans_query(
...
+        slice_size = round((trace_end - trace_start) / num_slices)
 
-        # ensure minimum of 1 width
-        if rounded_start == rounded_end:
-            rounded_end += bin_size
-    else:
-        rounded_start = span_start
-        rounded_end = span_end
+        slice_start = round((span_start - trace_start) / slice_size)
+        slice_start = clip(slice_start, 0, num_slices)
```
shruthilayaj, line 999: just for my benefit: 
1. `round((span_start - trace_start) / slice_size)` gives us the slice index this span would start at
2. `clip(slice_start, 0, num_slices)` just makes sure it's within the bounds of the number of allowed slices?

shruthilayaj (approved):

PR #70858, opened 2024-05-14
fix(trace-explorer): Skip floating spans in traces

Zylphrex:
There are some floating spans in a trace like INP spans. These spans are not associated with a transaction and can make the trace look longer. They also don't work in the trace view. So let's skip them for now until we have a clearer story around how it ties into the tracing experience.


edwardgou-sentry (approved):
Thanks!

PR #70861, opened 2024-05-14
ref(insights): Declare module base URLs as constants

gggritso:
More preparation for moving all insights to the `/insights/` URL namespace.

Now that every module is wrapped in `ModulePageProviders` I can set the correct `baseURL` for each one. `BASE_URL` is the _base URL of the whole module_ rather than of the page. It's also not the _full_ base URL, only the module itself, relevant to whatever comes before it (usually "/performance"), which is why they don't have starting slashes (except for AI).

Most importantly, almost nothing in the app actually _uses_ `baseURL` to construct URLs, so this PR has almost no impact. The only affected links:

- resource table
- queries table
- cache samples table


static/app/views/aiMonitoring/aiMonitoringDetailsPage.tsx
```diff
@@ -178,7 +179,7 @@ function PageWithProviders({params}: Props) {
   return (
     <ModulePageProviders
       title={[spanDescription ?? t('(no name)'), t('Pipeline Details')].join(' — ')}
-      baseURL="/ai-monitoring/"
+      baseURL={BASE_URL}
```
anonrig, line 182: This also includes sentry locale package which you don't need in this file. This could easily come to a point where it effects build time and makes it harder to eliminate dead code. I'm against this pattern.

anonrig (changes requested):

static/app/views/aiMonitoring/aiMonitoringDetailsPage.tsx
gggritso: 🤔 I see your point, but:

1. In this PR, every file that imports `BASE_URL` already imports `t`, so I'm not introducing a new import
2. Colocating regular constants and translated strings is a common pattern

If you feel strongly that this pattern is bad, I would love to see proof that it has a meaningful impact on build time, an argument that it's worth sacrificing colocation, and some suggestions on what to do instead

I think the FE TSC or the frontend Slack channel are a more appropriate venue for this, since it needs some discussion. If we all agree, we can introduce this to our developer guidelines and work on improving the situation

In the meantime, is this worth blocking this PR for, or is it a nit? If it's the latter, blocking the PR is probably not appropriate, unless I'm misinterpreting our [PR reviewer guidelines](https://develop.sentry.dev/code-review/#guidelines-forreviewers)

edwardgou-sentry (approved):

anonrig (approved):

Issue #69691, opened 2024-04-25
Issues Stream: Update event stats chart to include issue status


malwilley:
- Current component: `groupChart.tsx`
- Changes
    - Change position of mark line label to be on the right
    - Add status (ongoing, escalating, etc) to bottom of chart
    - Remove status from left portion of issue
- Challenges
    - This loads after the rest of the issue response. Should have a loading state that displays the issue status event while the event stats are still fetching.
- Rollout
    -  New feature flag for internal and EA rollouts

![CleanShot 2024-04-24 at 09 58 03](https://github.com/getsentry/sentry/assets/10888943/5424f755-d7fa-45e8-932d-a95a2961155b)

PR #70866, opened 2024-05-14
feat(events-graph): Add storybook for new events graph component

MichaelSun48:
Related ticket: #69691

This PR creates the storybook and initial design for the new events stats graph component, `<GroupStatusChart/>`. **This PR does not change the existing events graph in the issue stream**. 

The only publicly used file that was modified is miniBarChart.tsx. The `<MiniBarChart/>` component is the underlying component of the existing events graph component, `<GroupChart/>`, and it needs to expose more parameters of the graph as props to allow for the modifications in the new `<GroupStatusChart/>` component to avoid creating an entirely new underlying component\. Prop defaults were set for all new props introduced to ensure that the existing event stats chart is visually unchanged. 

The new component in its current form: 
image

Remaining work to be done (probably in future PRs): 
- [ ] Add test file
- [ ] Add loading state
- [ ] **Swap old events graph with this one**

Open design questions:
- [ ] ~~What should the loading state look like?~~ Gray box (with rounded edges)

### DIFF
--- a/static/app/components/charts/groupStatusChart.stories.tsx
+++ b/static/app/components/charts/groupStatusChart.stories.tsx
@@ -1,0 +1,51 @@
+import GroupStatusChart from 'sentry/components/charts/groupStatusChart';
+import storyBook from 'sentry/stories/storyBook';
+import type {Group, TimeseriesValue} from 'sentry/types';
+
+const stats: ReadonlyArray<TimeseriesValue> = [
+  // [1715554800, 126],
+  [1715558400, 112],
+  // [1715562000, 126],
+  [1715565600, 113],
+  // [1715569200, 118],
+  [1715572800, 87],
+  // [1715576400, 88],
+  [1715580000, 59],
+  // [1715583600, 54],
+  [1715587200, 55],
+  // [1715590800, 52],
+  [1715594400, 48],
+  // [1715598000, 62],
+  [1715601600, 86],
+  // [1715605200, 100],
+  [1715608800, 121],
+  // [1715612400, 124],
+  [1715616000, 129],
+  // [1715619600, 149],
+  [1715623200, 141],
+  // [1715626800, 132],
+  [1715630400, 133],
+  // [1715634000, 127],
+  [1715637600, 82],
+];
+
+const fakeGroup = {
+  stats: {
+    '24h': stats,
+  },
+} as unknown as Group;
+
+export default storyBook(GroupStatusChart, story => {
+  story('Default', () => {
+    return (
+      <div style={{display: 'flex', justifyContent: 'center', marginBottom: '20px'}}>
+        <GroupStatusChart
+          showMarkLine
+          data={fakeGroup}
+          groupStatus="Escalating"
+          statsPeriod="24h"
+        />
+      </div>
+    );
+  });
+});
--- a/static/app/components/charts/groupStatusChart.stories.tsx
+++ b/static/app/components/charts/groupStatusChart.stories.tsx
@@ -1,3 +1,5 @@
+import styled from '@emotion/styled';
+
 import GroupStatusChart from 'sentry/components/charts/groupStatusChart';
 import storyBook from 'sentry/stories/storyBook';
 import type {Group, TimeseriesValue} from 'sentry/types';
@@ -38,14 +40,20 @@ const fakeGroup = {
 export default storyBook(GroupStatusChart, story => {
   story('Default', () => {
     return (
-      <div style={{display: 'flex', justifyContent: 'center', marginBottom: '20px'}}>
+      <GraphContainer>
         <GroupStatusChart
           showMarkLine
           data={fakeGroup}
           groupStatus="Escalating"
           statsPeriod="24h"
         />
-      </div>
+      </GraphContainer>
     );
   });
 });
+
+const GraphContainer = styled('div')`
+  display: flex;
+  justify-content: center;
+  margin-bottom: 20px;
+`;
--- a/static/app/components/charts/groupStatusChart.stories.tsx
+++ b/static/app/components/charts/groupStatusChart.stories.tsx
@@ -2,7 +2,7 @@ import styled from '@emotion/styled';
 
 import GroupStatusChart from 'sentry/components/charts/groupStatusChart';
 import storyBook from 'sentry/stories/storyBook';
-import type {Group, TimeseriesValue} from 'sentry/types';
+import type {TimeseriesValue} from 'sentry/types';
 
 const stats: ReadonlyArray<TimeseriesValue> = [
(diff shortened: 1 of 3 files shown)


static/app/components/charts/groupStatusChart.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+  // [1715605200, 100],
+  [1715608800, 121],
+  // [1715612400, 124],
+  [1715616000, 129],
+  // [1715619600, 149],
+  [1715623200, 141],
+  // [1715626800, 132],
+  [1715630400, 133],
+  // [1715634000, 127],
+  [1715637600, 82],
```
MichaelSun48, line 31: Half of the 24 entries in this series are commented out to align the storybook more with the existing mockups, which currently only 12 bars. This is what it looks like with all 24 entries enabled:
image

static/app/components/charts/groupStatusChart.tsx
```diff
@@ -0,0 +1,150 @@
...
+type Props = {
+  data: Group;
+  statsPeriod: string;
+  groupStatus?: string;
+  height?: number;
+  showMarkLine?: boolean;
+  showSecondaryPoints?: boolean;
+};
+
+function GroupStatusChart({
```
MichaelSun48, line 31: This new component is largely the same as the pre-existing `<GroupChart/>` component, but I would still prefer to have work contained in this new file to avoid any unintended changes to the existing component.

static/app/components/charts/groupStatusChart.tsx
```diff
@@ -0,0 +1,150 @@
...
+    Series[],
+    [string] | undefined,
+    [string] | undefined,
+  ] = useMemo(() => {
+    if (!stats || !stats.length) {
+      return [[], undefined, undefined];
+    }
+
+    let max = 0;
+    const marklinePoints: number[] = [];
```
malwilley, line 58: nit: could be simplified a bit:

```ts
const marklinePoints = stats.map(point => point[1]);
const max = Math.max(...marklinePoints);
```

It does iterate through the list one more time, but usually simpler code is the better tradeoff since it rarely has an actual effect on the user. I try to avoid using `let` whenever possible.

static/app/components/charts/groupStatusChart.tsx
```diff
@@ -0,0 +1,150 @@
...
+      chartSeries.push({
+        seriesName: t('Matching Events'),
+        data: stats.map(asChartPoint),
+      });
+    } else {
+      // Colors are custom to preserve historical appearance where the single series is
+      // considerably darker than the two series results.
+      chartColors = [theme.gray300];
+      chartEmphasisColors = [theme.purple300];
+      secondaryStats;
```
malwilley, line 88: 😕

static/app/components/charts/groupStatusChart.tsx
```diff
@@ -0,0 +1,150 @@
...
+    : EMPTY_STATS;
+
+  const secondaryStats: TimeseriesValue[] | null =
+    statsPeriod && data.filtered ? data.stats[statsPeriod] : null;
+
+  const [series, colors, emphasisColors]: [
+    Series[],
+    [string] | undefined,
+    [string] | undefined,
+  ] = useMemo(() => {
```
scttcper, line 52: this might be easier to read if you give the tuple values names or just return an object and do destructuring 

named tuples look like
```ts
type x = [start: number, end: number]
```

scttcper, line 52: also put the type in the generic or the return type so it gets typed checked earlier
```suggestion
  const [series, colors, emphasisColors] = useMemo<[
    Series[],
    [string] | undefined,
    [string] | undefined,
  ]>(() => {
```

static/app/components/charts/groupStatusChart.tsx
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    const formattedMarkLine = formatAbbreviatedNumber(max);
+
+    const chartSeries: Series[] = [];
+    let chartColors: [string] | undefined = undefined;
+    let chartEmphasisColors: [string] | undefined = undefined;
+
+    if (showSecondaryPoints && secondaryStats && secondaryStats.length) {
+      chartSeries.push({
```
malwilley, line 75: Instead of the `let` above and pushing, you could just return the value in each of the if branches:

```ts
const chartSeries = [...]
return [chartSeries, chartColors, chartEmphasisColors]
```

static/app/components/charts/groupStatusChart.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const stats: ReadonlyArray<TimeseriesValue> = statsPeriod
+    ? data.filtered
+      ? data.filtered.stats?.[statsPeriod]
+      : data.stats?.[statsPeriod]
+    : EMPTY_STATS;
+
+  const secondaryStats: TimeseriesValue[] | null =
+    statsPeriod && data.filtered ? data.stats[statsPeriod] : null;
+
+  const [series, colors, emphasisColors]: [
```
malwilley, line 48: nit: returning an object instead of an array here would prevent bugs from creeping in if someone messed up the ordering of the return value

static/app/components/charts/groupStatusChart.tsx
```diff
@@ -0,0 +1,150 @@
...
+  );
+}
+
+export default GroupStatusChart;
+
+const ChartWrapper = styled('div')`
+  display: flex;
+  flex-direction: column;
+`;
+// TODO: Call this something better
```
malwilley, line 147: Eh, it's fine 😄

static/app/components/charts/groupStatusChart.tsx
```diff
@@ -0,0 +1,150 @@
...
+  return {
+    name: point[0] * 1000,
+    value: point[1],
+  };
+}
+
+const EMPTY_STATS: ReadonlyArray<TimeseriesValue> = [];
+
+type Props = {
+  data: Group;
```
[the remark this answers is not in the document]
MichaelSun48: I completely agree and would take it another step further - the chart component shouldn't even have the responsibility of deciding whether or not to use the filtered or unfiltered stats, especially since it only uses one. The parent component should just decide which one to use and pass the correct stats down to this component.

MichaelSun48:
@malwilley @scttcper thank you for the feedback!! This is infinitely more readable now 

I'll also put up another PR making the same changes to the existing `<GroupChart/>` component since most of the changes were applied to code that I copy pasted from there.

scttcper (approved):

MichaelSun48:
^ This commit will also round the corners of each bar in the bar chart in prod: 

before: 
image


after:
image

PR #70868, opened 2024-05-14
feat(new-trace): add cache metrics to transaction trace details

DominikB2014:
Adds cache metrics section to a transaction node, if the transaction has cache metrics
image


static/app/views/starfish/views/spans/types.tsx
```diff
@@ -49,7 +49,7 @@ export const DataTitles: Record<DataKey, string> = {
   'avg(cache.item_size)': t('Avg Value Size'),
   unsuccessfulHTTPCodes: t('Response Codes (3XX, 4XX, 5XX)'),
   httpCodeBreakdown: t('Response Code Breakdown'),
-  cacheMissRate: t('Miss Rate'),
+  'cache_miss_rate()': t('Miss Rate'),
```
DominikB2014, line 52: Threw in an extra refactor here, sorry about that

static/app/views/performance/newTraceDetails/referrers.ts
```diff
@@ -0,0 +1,3 @@
+export enum Referrer {
```
JonasBa, line 1: This is only used in a single place, can we just inline it there?

static/app/views/performance/newTraceDetails/traceDrawer/details/transaction/index.tsx
```diff
@@ -102,7 +107,17 @@ export function TransactionNodeDetails({
...
+    {
+      search: MutableSearch.fromQueryObject({
+        transaction: node.value.transaction,
+      } satisfies SpanMetricsQueryFilters),
+      fields: ['avg(cache.item_size)', 'cache_miss_rate()'],
+    },
+    Referrer.TRACE_DRAWER_TRANSACTION_CACHE_METRICS
+  );
+
+  if (isLoading || isCacheMetricsLoading) {
```
JonasBa, line 120: Can this be independent from the first request? I'm not a fan of showing loading indicators while multiple requests load, we should instead aim to flush things to the screen asap

static/app/views/performance/newTraceDetails/traceDrawer/details/transaction/index.tsx
```diff
@@ -134,6 +149,7 @@ export function TransactionNodeDetails({
         />
         <AdditionalData event={event} />
         <Measurements event={event} location={location} organization={organization} />
+        {cacheMetrics[0] && <CacheMetrics cacheMetrics={cacheMetrics[0]} />}
```
JonasBa, line 152: Why only cacheMetrics[0]? Do we always expect it to be a single metric or should we instead map and display all of these metric here?

static/app/views/performance/newTraceDetails/referrers.ts
DominikB2014: So this is mainly for consistency between modules, where we declare all referrers in one file. For example
https://github.com/getsentry/sentry/blob/master/static/app/views/performance/cache/referrers.ts

Yes rn its only in a single place, but ideally overtime we extend this to include all referrers used in the trace view. 
Does that work for you? Otherwise I can make it inline.

static/app/views/performance/newTraceDetails/traceDrawer/details/transaction/index.tsx
DominikB2014: Yeah we always expect a single metric. The Discover endpoint just always returns an array of results by design

static/app/views/performance/newTraceDetails/traceDrawer/details/transaction/index.tsx
DominikB2014: We could, but I wanted to avoid the case where the cacheMetrics take longer to load, and suddenly pop into view causing a layout shift. If we're okay with that shift then I can make this independent. 

Alternatively we could do some sort of skeleton for this data, but cacheMetrics won't always be applicable to a transaction, so in that case we probably shouldn't show them.

static/app/views/performance/newTraceDetails/traceDrawer/details/transaction/index.tsx
DominikB2014: For clarification, `cacheMetrics` is of type ` Pick<SpanMetricsResponse, 'avg(cache.item_size)' | 'cache_miss_rate()'>[]`.
We only expect one element in this array (as we're querying for a single transaction), containing the `avg(cache.item_size)` and `cache_miss_rate()` metrics for that transaction.

The only case where we have multiple results is if there are multiple transactions with the same name, but that shouldn't be possible.

static/app/views/performance/newTraceDetails/referrers.ts
JonasBa: Imo, code like this is trivial to refactor and my tendency is to start simple (just less moving parts and files around which hold little value). Eod not a big deal and keeping this consistent across modules is a win.

static/app/views/performance/newTraceDetails/traceDrawer/details/transaction/index.tsx
JonasBa: I think it's ok to show it into the screen later, layout shift is a better problem to have than slow load times. A skeleton sounds like a good idea, and can always be added later. Would you mind sending me a screencap of how this loads when we do have cached metrics vs when we dont? I'm going to show them to the designer on this so we can figure out how what kind of experience we can show to progressively enhance or degrade this

DominikB2014: Sent over slack!

static/app/views/performance/newTraceDetails/traceDrawer/details/transaction/index.tsx
DominikB2014: That's fair, this pattern is used extensively in modules, so it was more of an oversight. I'm okay with rendering all of them too, i like how it has that visual explanation if something weird is happening as you mentioned.
Will update the PR!

JonasBa: Oh yeah, we do this a ton all over sentry and it's just something I want us to be more mindful about.

DominikB2014:
@JonasBa I've addressed you're comments about allowing cache metrics to load independently and mapping all the cache metrics instead of the first element. Lmk what u think!

JonasBa (approved):

PR #70872, opened 2024-05-14
ref(crons): Never allow the clock tick consumer to skip ticks

evanpurkhiser:
We SHOULD process every tick. If there is a problem a backlog is preferred.

We saw this in INC-760, where a query took too long. Luckily the query timeout is longer than the arroyo configured `max.poll.interval.ms`. Meaning the consumer was kicked from the consumer group before we could fail and try to commit


JoshFerge (approved):
yup makes sense

davidenwang (approved):

PR #70878, opened 2024-05-14
feat(related_issues): Support passing an event_id to the endpoint

armenzg:
This enables showing trace connected issues from the issue details page for the event we're currently viewing.

### DIFF
--- a/src/sentry/api/endpoints/issues/related_issues.py
+++ b/src/sentry/api/endpoints/issues/related_issues.py
@@ -39,5 +39,8 @@ class RelatedIssuesEndpoint(GroupEndpoint):
             "event_id": request.query_params.get("event_id"),
             "project_id": request.query_params.get("project_id"),
         }
-        data, meta = RELATED_ISSUES_ALGORITHMS[related_type](group, extra_args)
-        return Response({"type": related_type, "data": data, "meta": meta})
+        try:
+            data, meta = RELATED_ISSUES_ALGORITHMS[related_type](group, extra_args)
+            return Response({"type": related_type, "data": data, "meta": meta})
+        except AssertionError:
+            return Response({}, status=400)
--- a/src/sentry/api/endpoints/issues/related_issues.py
+++ b/src/sentry/api/endpoints/issues/related_issues.py
@@ -35,5 +35,9 @@ class RelatedIssuesEndpoint(GroupEndpoint):
         """
         # The type of related issues to retrieve. Can be either `same_root_cause` or `trace_connected`.
         related_type = request.query_params["type"]
-        data, meta = RELATED_ISSUES_ALGORITHMS[related_type](group)
+        extra_args = {
+            "event_id": request.query_params.get("event_id"),
+            "project_id": request.query_params.get("project_id"),
+        }
+        data, meta = RELATED_ISSUES_ALGORITHMS[related_type](group, extra_args)
         return Response({"type": related_type, "data": data, "meta": meta})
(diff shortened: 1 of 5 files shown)


tests/sentry/api/endpoints/issues/test_related_issues.py
```diff
@@ -47,20 +45,43 @@ def test_same_root_related_issues(self) -> None:
         assert response.json() == {"type": "same_root_cause", "data": [5], "meta": {}}
 
     def test_trace_connected_errors(self) -> None:
-        error_event, _, another_proj_event = self.load_errors(self.project, uuid4().hex[:16])
+        error_event, _, another_proj_event = self.load_errors(self.project)
```
armenzg, line 48: No need to pass the `span_id` anymore.

tests/sentry/api/endpoints/issues/test_related_issues.py
```diff
@@ -47,20 +45,43 @@ def test_same_root_related_issues(self) -> None:
...
-        self.group_id = error_event.group_id  # type: ignore[assignment]
         recommended_event = group.get_recommended_event_for_environments()  # type: ignore[union-attr]
         assert recommended_event is not None  # It helps with typing
 
         assert error_event.group_id != another_proj_event.group_id
         assert error_event.project.id != another_proj_event.project.id
         assert error_event.trace_id == another_proj_event.trace_id
 
+        # This sets the group_id to the one we want to query about
+        self.group_id = error_event.group_id  # type: ignore[assignment]
```
armenzg, line 61: Moving this line closer to the call to the API since it is related.

tests/sentry/api/endpoints/issues/test_related_issues.py
```diff
@@ -47,20 +45,43 @@ def test_same_root_related_issues(self) -> None:
...
+    def test_trace_connected_errors_specific_event(self) -> None:
+        error_event, _, another_proj_event = self.load_errors(self.project)
+
+        # This sets the group_id to the one we want to query about
+        self.group_id = another_proj_event.group_id  # type: ignore[assignment]
+        response = self.get_success_response(
+            qs_params={
+                "type": "trace_connected",
+                "event_id": another_proj_event.event_id,
+                "project_id": another_proj_event.project_id,
```
armenzg, line 82: We test in here that we pass `event_id` and `project_id` unlike the previous test.

src/sentry/api/endpoints/issues/related_issues.py
```diff
@@ -35,5 +35,9 @@ def get(self, request: Request, group: Group) -> Response:
         """
         # The type of related issues to retrieve. Can be either `same_root_cause` or `trace_connected`.
         related_type = request.query_params["type"]
-        data, meta = RELATED_ISSUES_ALGORITHMS[related_type](group)
+        extra_args = {
+            "event_id": request.query_params.get("event_id"),
+            "project_id": request.query_params.get("project_id"),
```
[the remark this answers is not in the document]
vartec: if it's required, shouldn't this validate that it is not none?

src/sentry/issues/related/trace_connected.py
```diff
@@ -11,12 +13,47 @@
...
+    event_id = extra_args.get("event_id")
+    project_id = extra_args.get("project_id")
+    if event_id:
+        # If we are passing an specific event_id, we need to get the project_id
+        assert project_id is not None
+        event = eventstore.backend.get_event_by_id(project_id, event_id, group_id=group.id)
+        # If we are requesting an specific event, we want to be notified with an error
+        assert event is not None
+        # This ensures that the event is actually part of the group and we are notified
+        assert event.group_id == group.id
```
vartec, line 35: AssertionError produced here would bubble up as generic `500 Internal Server Error`, but it should be caught in the front end and return `400 Bad Request`

vartec (approved):

PR #70880, opened 2024-05-14
ref(rules): Raise exception on keyerror

ceorourke:
Follow up from https://github.com/getsentry/sentry/pull/70831#discussion_r1599274482 to make this raise an exception rather than just log an error.

### DIFF
--- a/src/sentry/buffer/redis.py
+++ b/src/sentry/buffer/redis.py
@@ -73,8 +73,7 @@ class BufferHookRegistry:
         try:
             callback = self._registry[buffer_hook_event]
         except KeyError:
-            logger.info("buffer_hook_event.missing", extra={"key_name": buffer_hook_event.value})
-            return False
+            logger.exception("buffer_hook_event.missing")
 
         return callback(data)
 


saponifi3d (approved):
TIL that this will also automatically use the exception. 🔥

gmisiolek-sbm:
After this change there are errors:

```
worker-1                                        | Traceback (most recent call last):
worker-1                                        |   File "/usr/local/lib/python3.11/site-packages/sentry/buffer/redis.py", line 74, in callback
worker-1                                        |     callback = self._registry[buffer_hook_event]
worker-1                                        |                ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
worker-1                                        | KeyError: <BufferHookEvent.FLUSH: 'flush'>
worker-1                                        | 22:11:00 [ERROR] sentry.buffer.redis: buffer_hook_event.missing
worker-1                                        | Traceback (most recent call last):
worker-1                                        |   File "/usr/local/lib/python3.11/site-packages/sentry/buffer/redis.py", line 313, in process_batch
worker-1                                        |     redis_buffer_registry.callback(BufferHookEvent.FLUSH, self)
worker-1                                        |   File "/usr/local/lib/python3.11/site-packages/sentry/buffer/redis.py", line 78, in callback
worker-1                                        |     return callback(data)
worker-1                                        |            ^^^^^^^^
worker-1                                        | UnboundLocalError: cannot access local variable 'callback' where it is not associated with a value
```

why `return` has been removed? In that case it should try except else...

PR #70882, opened 2024-05-14
feat(integrations): Toggle features at organization integration level

### LINKED ISSUE #70316: Slack alerts in threads: add option to enable or disable it

ykamo001:
We need to allow users to be able to toggle features for integrations at their organization integration level. If users do not like or want the features, they should have the power to turn them on or off at their disposal. 

The previous FE implementation was doing this at the organization model level, but we can leverage the organization integration model/object to hold and control these settings. Added a custom component that will handle doing this toggling across the integration, and made it integration agnostic to be used across all the other integrations as well.

We can slowly migrate the github piece to this functionality. There is more work to be done on the FE side to help improve this code area, but it would require a lot of effort and changes that we can follow up with after better planning.

https://github.com/getsentry/sentry/assets/5581484/1213ca25-19d7-417e-89f2-166aada6b7ec

Requires: https://github.com/getsentry/sentry/pull/70620

### DIFF
--- a/src/sentry/integrations/slack/integration.py
+++ b/src/sentry/integrations/slack/integration.py
@@ -92,6 +92,51 @@ class SlackIntegration(SlackNotifyBasicMixin, IntegrationInstallation):
         default_installation = (
             "classic_bot" if "user_access_token" in metadata_ else "workspace_app"
         )
+
+    def get_organization_config(self) -> Sequence[tuple[str, bool]]:
+        """
+        Not sure why the base class is restricted to a sequence type, doesn't really make sense, however, it is
+        sufficient to our needs for now, and we can utilize it to return toggleable flags/feature on the integration
+        """
+
+        # Specifically using the parent method because the overwritten method on current class is hacked for another
+        # purpose at the integration/provider wide level, which is wrong/incorrect
+        base_data = super().get_config_data()
+
+        stored_flag_data = base_data.get(self._FLAGS_KEY, {})
+        flag_statuses = []
+        for flag_name, default_flag_value in self._SUPPORTED_FLAGS_WITH_DEFAULTS.items():
+            flag_value = stored_flag_data.get(flag_name, default_flag_value)
+            flag_statuses.append((flag_name, flag_value))
+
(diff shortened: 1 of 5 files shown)


src/sentry/integrations/slack/integration.py
ykamo001, line 1: merge conflict problem, removing this

cathteng:
should we be using Form and JsonForm? it would keep the styling that's already present in the Github features page
https://github.com/getsentry/sentry/blob/1a913b72556f24a9def277b85d5c672eec772c96/static/app/views/settings/organizationIntegrations/integrationDetailedView.tsx#L388-L405

ykamo001:
> should we be using Form and JsonForm? it would keep the styling that's already present in the Github features page
> 
> https://github.com/getsentry/sentry/blob/1a913b72556f24a9def277b85d5c672eec772c96/static/app/views/settings/organizationIntegrations/integrationDetailedView.tsx#L388-L405

We can't use that because of the way we're calling the API to update all configs as a global state

static/app/views/settings/organizationIntegrations/featureToggle.tsx
```diff
@@ -0,0 +1,119 @@
...
+        toggleableFlags: featureToggles,
+      };
+
+      try {
+        api.request(integrationConfigEndpoint, {
+          method: 'POST',
+          data: updatedConfigData,
+        });
+      } catch (error) {
+        // Nothing to do here
```
sentaur-athena, line 57: What kind of error can happen here? If the toggle is not successful we should reflect that somehow in the UI. Throw a notice maybe?

ykamo001: If the call fails, we'll have to hold onto all the errors and show that correctly. I think at this point, there is no error handling on this component, so we may have to introduce that. Maybe a banner per error, or just a singular error?

ykamo001:
Closing this in favor of following existing convention to keep UI the same

PR #70883, opened 2024-05-14
feat(performance): Adds error rate to queues module

edwardgou-sentry:
Updates the Queues module to now fully display error rate in the:
- landing page destinations table
- destination summary metrics readout
- destination summary transactions table
- samples panels

Also includes a minor fix to producer/consumer sample panel readouts not showing accurate span counts.


static/app/views/performance/queues/destinationSummary/destinationSummaryPage.tsx
```diff
@@ -41,6 +41,7 @@ function DestinationSummaryPage() {
   const destination = decodeScalar(query.destination);
 
   const {data} = useQueuesMetricsQuery({destination});
+  const errorRate = 1 - (data[0]?.['trace_status_rate(ok)'] ?? 0);
```
gggritso, line 44: 😅 there's no `trace_status_rate(error)` for you?

gggritso (approved):
All the `1 - ` are a huge bummer 😢 someone's going to trip over that for sure. If there's any chance of implementing that Discover-side on the backend, it's so worth it! Otherwise LGTM

static/app/views/performance/queues/destinationSummary/destinationSummaryPage.tsx
edwardgou-sentry: Unfortunately the trace context status has a bunch of different error states:
https://develop.sentry.dev/sdk/event-payloads/contexts/#trace-context

ie `cancelled`, `unknown`, `invalid_argument` are all possible values and are all considered errors.

I can update this in discover as a follow up!

gggritso: Makes sense! Definitely worth it to define a helper for this in Discover regardless, if possible

PR #70885, opened 2024-05-14
ci: reduce time running codecov

giovanni-guidini:
`plugin: noop` prevents default plugins from running. These create a new version of the coverage file and that is unnecessary work being done.

Example run showing that it handle airtifact in 5s: https://github.com/giovanni-guidini/sentry/actions/runs/9084832844/job/24966726208


nicholas-codecov (approved):
LGTM

PR #70890, opened 2024-05-14
ref(insights): abstract ScreenLoadSpanSamples side panel for mobile modules

KevinL10:
Currently, the app starts module depends on `ScreenLoadSpanSamples` in `mobile/screenload` to display the side panel. However, we therefore cannot identify which module the panel comes from, which we will need for analytics (e.g. # of times a sample span is clicked per module) - https://github.com/getsentry/sentry/pull/70822.

This PR moves the component under `mobile/components` and adds a moduleName prop to identify the originating module.


gggritso (approved):
👍🏻

narsaynorath (approved):

PR #70891, opened 2024-05-14
ref(rules): Add logger when adding buffer hook event

ceorourke:
Add a logger to see if this is ever getting added in prod, since we see it happening locally and aren't sure what is going on.

### DIFF
--- a/src/sentry/tasks/post_process.py
+++ b/src/sentry/tasks/post_process.py
@@ -994,6 +994,10 @@ def process_rules(job: PostProcessJob) -> None:
     from sentry.rules.processing.delayed_processing import process_delayed_alert_conditions
 
     if not redis_buffer_registry.has(BufferHookEvent.FLUSH):
+        logger.info(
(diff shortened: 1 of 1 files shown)


src/sentry/tasks/post_process.py
```diff
@@ -994,6 +994,10 @@ def process_rules(job: PostProcessJob) -> None:
     from sentry.rules.processing.delayed_processing import process_delayed_alert_conditions
 
     if not redis_buffer_registry.has(BufferHookEvent.FLUSH):
+        logger.info(
+            "post_process.process_rules.adding_bufferhookevent",
+            extra={"key_name": BufferHookEvent.FLUSH.value},
```
saponifi3d, line 999: unclear if it's helpful... but maybe we should also serialize the function here too? i'm mostly thinking that i'd be nice to see as much information as possible.

saponifi3d (approved):
:shipit:

PR #70895, opened 2024-05-14
chore(integrations): better IntegrationFormError error message

cathteng:
Initialize `super()` with `field_errors` so when we call `str(e)` it actually outputs a helpful error message.

### DIFF
--- a/src/sentry/shared_integrations/exceptions/__init__.py
+++ b/src/sentry/shared_integrations/exceptions/__init__.py
@@ -171,8 +171,11 @@ class DuplicateDisplayNameError(IntegrationError):
 
 class IntegrationFormError(IntegrationError):
     def __init__(self, field_errors: Mapping[str, Any]) -> None:
-        super().__init__("Invalid integration action")
-        self.field_errors = field_errors
+        error = "Invalid integration action"
+        if field_errors:
+            error = str(field_errors)
+
+        super().__init__(error)
 
 
 class ClientError(RequestException):


src/sentry/shared_integrations/exceptions/__init__.py
```diff
@@ -171,8 +171,11 @@ class DuplicateDisplayNameError(IntegrationError):
 
 class IntegrationFormError(IntegrationError):
     def __init__(self, field_errors: Mapping[str, Any]) -> None:
-        super().__init__("Invalid integration action")
-        self.field_errors = field_errors
```
schew2381, line 175: Do we use this instance variable for anything?

cathteng: no

cathteng: it's not being helpful atm

sentaur-athena: This is literally the only place it shows up in the code base.

src/sentry/shared_integrations/exceptions/__init__.py
```diff
@@ -171,8 +171,11 @@ class DuplicateDisplayNameError(IntegrationError):
...
+        error = "Invalid integration action"
+        if field_errors:
+            error = str(field_errors)
```
sentaur-athena, line 176: Does map to str always come in an ok format?

sentaur-athena (approved):

src/sentry/shared_integrations/exceptions/__init__.py
cathteng: this should be fine, i just tested it in a python playground

schew2381 (approved):

PR #70896, opened 2024-05-14
featchartcuterie): Update Chart Styling

iamrajjoshi:
The chart styling generated locally looks slightly different from in the container. Namely the font is different. I investigated and it seems to be because we don't pass the default font for the axis labels and the default font in the container is different. Issues team also had this issue when they generated these charts. I really can't test this locally because we need to look at the behavior of the prod container, but for what its worth, the chart still renders correctly locally. This should fix at least the styling of the axis labels.

Local:
image

Container:
image

### DIFF
--- a/static/app/chartcuterie/performance.tsx
+++ b/static/app/chartcuterie/performance.tsx
@@ -8,7 +8,7 @@ import {transformStatsResponse} from 'sentry/utils/profiling/hooks/utils';
 import {lightTheme as theme} from 'sentry/utils/theme';
 import type {NormalizedTrendsTransaction} from 'sentry/views/performance/trends/types';
 
-import {slackChartDefaults, slackChartSize} from './slack';
+import {DEFAULT_FONT_FAMILY, slackChartDefaults, slackChartSize} from './slack';
 import type {RenderDescriptor} from './types';
 import {ChartType} from './types';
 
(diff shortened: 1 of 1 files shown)


sentaur-athena (approved):

static/app/chartcuterie/performance.tsx
```diff
@@ -23,6 +23,15 @@ function modifyOptionsForSlack(options: Omit<LineChartProps, 'series'>) {
...
   options.legend.left = '25';
   options.legend.top = '20';
+  options.grid = slackChartDefaults.grid;
+
+  options.yAxis = options.yAxis || {};
+  options.yAxis.axisLabel = options.yAxis.axisLabel || {};
+  options.yAxis.axisLabel.fontFamily = DEFAULT_FONT_FAMILY;
+
+  options.xAxis = options.xAxis || {};
+  options.xAxis.axisLabel = options.xAxis.axisLabel || {};
```
sentaur-athena, line 33: Are you doing these just to not get lint errors?

You can also do it with `&&` like
```
if (options.xAxis && options.xAxis.axisLabel) ...

iamrajjoshi: hmm, i should actually do it that way next time. ill update this when i modify the style in my next pr.

PR #70897, opened 2024-05-14
chore(pr-comments): metric for possible suspect issue comments on open PRs

cathteng:
https://docs.github.com/en/rest/commits/commits#list-pull-requests-associated-with-a-commit

> Returns the merged pull request that introduced the commit to the repository. If the commit is not present in the default branch, will only return open pull requests associated with the commit.

Adding a metric so we can see if we could be making more suspect issue comments on open PRs

### DIFF
--- a/src/sentry/tasks/commit_context.py
+++ b/src/sentry/tasks/commit_context.py
@@ -88,14 +88,6 @@ def queue_comment_task_if_needed(
             "github_pr_comment.queue_comment_check.open_pr",
             sample_rate=1.0,
         )
-        logger.info(
-            "github.pr_comment.queue_comment_check.open_pr",
-            extra={
-                "organization_id": commit.organization_id,
-                "repository_id": repo.id,
-                "commit_sha": commit.key,
-            },
-        )
         return
 
     merge_commit_sha = response[0]["merge_commit_sha"]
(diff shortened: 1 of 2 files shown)


src/sentry/tasks/commit_context.py
```diff
@@ -83,6 +83,21 @@ def queue_comment_task_if_needed(
             )
         return
```
cathteng, line 84: you could have multiple open PRs with a single commit, but i think it'd be quite noisy if we commented on all of them, so not tracking the metric here

src/sentry/tasks/commit_context.py
```diff
@@ -83,6 +83,21 @@ def queue_comment_task_if_needed(
...
 
+    if response[0]["state"] == "open":
+        metrics.incr(
+            "github_pr_comment.queue_comment_check.open_pr",
+            sample_rate=1.0,
+        )
+        logger.info(
```
JoshFerge, line 91: we're adding a log too, just wanna be sure as the PR says only metric. will want to watch out if this is super noisy

JoshFerge (approved):

src/sentry/tasks/commit_context.py
cathteng: i'll remove the log

PR #70904, opened 2024-05-14
feat(similar-grouping): dry run option for backfill

JoshFerge:
calls the endpoint that will delete the embeddings records in seer, which was added here: https://github.com/getsentry/seer/pull/713/

note that this will delete the records at the _beginning_ of the backfill, so we have a moment to analyze the state of the seer database between runs.

### DIFF
--- a/src/sentry/conf/server.py
+++ b/src/sentry/conf/server.py
@@ -3958,6 +3958,9 @@ SEER_MAX_SIMILARITY_DISTANCE = 0.15  # Not yet in use - Seer doesn't obey this r
 SEER_GROUPING_RECORDS_URL = (
     f"/{SEER_SIMILARITY_MODEL_VERSION}/issues/similar-issues/grouping-record"
 )
+SEER_GROUPING_RECORDS_DELETE_URL = (
+    f"/{SEER_SIMILARITY_MODEL_VERSION}/issues/similar-issues/grouping-record/delete/"
+)
 
 # Devserver configuration overrides.
 ngrok_host = os.environ.get("SENTRY_DEVSERVER_NGROK")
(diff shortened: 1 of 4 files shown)


src/sentry/seer/utils.py
```diff
@@ -290,3 +291,29 @@ def post_bulk_grouping_records(
...
         return {"success": False}
+
+
+def delete_grouping_records(
+    project_id: int,
+) -> bool:
+    try:
+        response = seer_grouping_connection_pool.urlopen(
+            "GET",
+            SEER_GROUPING_RECORDS_DELETE_URL + f"/{project_id}",
```
vartec, line 302: Trailing slash to avoid redirect, and whole thing as f-string to avoid concatenation
```suggestion
           f"{SEER_GROUPING_RECORDS_DELETE_URL}/ {project_id}/",
```

src/sentry/conf/server.py
```diff
@@ -3958,6 +3958,9 @@ def custom_parameter_sort(parameter: dict) -> tuple[str, int]:
 SEER_GROUPING_RECORDS_URL = (
     f"/{SEER_SIMILARITY_MODEL_VERSION}/issues/similar-issues/grouping-record"
 )
+SEER_GROUPING_RECORDS_DELETE_URL = (
+    f"/{SEER_SIMILARITY_MODEL_VERSION}/issues/similar-issues/grouping-record/delete/"
```
vartec, line 3962: you're concatenating that with `"/{project_id}"` so no need for a trailing slash here

```suggestion
    f"/{SEER_SIMILARITY_MODEL_VERSION}/issues/similar-issues/grouping-record/delete"
```

side note: why isn't this just `SEER_GROUPING_RECORDS_URL` with a verb `DELETE`?

vartec (approved):
some minor nitpicks, nothing that will stop it from working

src/sentry/seer/utils.py
JoshFerge: i don't think seer has trailing slashes, at least we don't add them on our other strings

src/sentry/conf/server.py
JoshFerge: the seer records are done in a per-batch fashion per N groups, while this method is simply taking a single project_id and performing all the deletions.

PR #70905, opened 2024-05-14
feat(slack): Only use threads if the feature is enabled

### LINKED ISSUE #70316: Slack alerts in threads: add option to enable or disable it

ykamo001:
Use or do not use threads based on the feature being enabled or disabled via the integration.

Requires: https://github.com/getsentry/sentry/pull/70620

### DIFF
--- a/src/sentry/integrations/slack/integration.py
+++ b/src/sentry/integrations/slack/integration.py
@@ -129,53 +129,6 @@ class SlackIntegration(SlackNotifyBasicMixin, IntegrationInstallation):
                         },
                     )
                     cleaned_flags_data[flag_name] = default_flag_value
-        base_data["installationType"] = metadata_.get("installation_type", default_installation)
-
-        # Add missing toggleable feature flags
-        stored_flag_data = base_data.get(self._FLAGS_KEY, {})
-        for flag_name, default_flag_value in self._SUPPORTED_FLAGS_WITH_DEFAULTS.items():
-            if flag_name not in stored_flag_data:
-                stored_flag_data[flag_name] = default_flag_value
-
-        base_data[self._FLAGS_KEY] = stored_flag_data
-        return base_data
-
-    def _update_and_clean_flags_in_organization_config(
-        self, data: MutableMapping[str, Any]
-    ) -> None:
-        """
-        Checks the new provided data for the flags key.
-        If the key does not exist, uses the default set values.
-        """
-
-        cleaned_flags_data = data.get(self._FLAGS_KEY, {})
-        # ensure we add the default supported flags if they don't already exist
-        for flag_name, default_flag_value in self._SUPPORTED_FLAGS_WITH_DEFAULTS.items():
-            flag_value = cleaned_flags_data.get(flag_name, None)
-            if flag_value is None:
-                cleaned_flags_data[flag_name] = default_flag_value
-            else:
-                # if the type for the flag is not the same as the default, use the default value as an override
-                if type(flag_value) is not type(default_flag_value):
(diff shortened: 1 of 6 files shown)


src/sentry/integrations/slack/integration.py
```diff
@@ -140,6 +140,21 @@ def update_organization_config(self, data: MutableMapping[str, Any]) -> None:
...
+        return data[self._FLAGS_KEY]
+
+    def has_feature(self, feature_name: str) -> bool:
+        """
+        Helper method to check if a particular feature has a flag
+        Default returns to False if the flag does not exist
+        """
+        features = self.get_features()
+        return features.get(feature_name, False)
+
```
sentaur-athena, line 157: Should these as well as toggleableFlags be in the parent class?

src/sentry/integrations/slack/utils/notifications.py
```diff
@@ -88,7 +91,13 @@ def send_incident_alert_notification(
     )
 
     # If a parent notification exists for this rule and action, then we can reply in a thread
-    if parent_notification_message is not None:
+    # However, if the feature is turned off, do not use threads
+    if (
+        parent_notification_message is not None
+        and organization_integration_has_metric_alerts_flag_enabled(
```
sentaur-athena, line 97: `organization_integration_has_metric_alerts_flag_enabled` is basically a one line function and I think `organization_integration_has_feature_enabled(
        integration=integration,
        organization_id=organization_id,
        feature=SlackIntegration.ISSUE_ALERTS_THREAD_FLAG,
    )` is as clear. If it helps with circular imports calling it directly could be a solution.

src/sentry/integrations/slack/utils/feature.py
```diff
@@ -0,0 +1,60 @@
...
+        return False
+
+    if not isinstance(installation, SlackIntegration):
+        _default_logger.info(
+            "installation object is not a slack integration",
+            extra=base_logs,
+        )
+        return False
+
+    return installation.has_feature(feature_name=feature)
```
sentaur-athena, line 36: Didn't spend a lot of time on it but I think this method can be more generic and used in parent integration class instead of being specific to slack. You need one wrapper around intallation.has_feature to make sure installation exists (super.organization_integration_has_feature_enabled) and then the integration specific method check to make sure the integration is slack.

Like instead of utils make these class methods of IntegrationFeatures and then have SlackFeatures inheriting it.

sentaur-athena (commented):
Code looks good. These are mainly just suggestions.

src/sentry/integrations/slack/integration.py
ykamo001: I was thinking about it, but I think I might have a better way, let me see if I can do it in this PR

src/sentry/integrations/slack/utils/feature.py
ykamo001: Yeah, was thinking the same, let me see what I can do

src/sentry/integrations/slack/integration.py
ykamo001: I think I'll tackle those changes in another pr to not block the GA

src/sentry/integrations/slack/utils/feature.py
ykamo001: I'll add that in a follow up PR as to not block GA

ykamo001:
Closing this in favor of following existing convention

PR #70908, opened 2024-05-14
ref: move Alert Rule endpoint serializers directory

nhsiehgit:
Moves the incident alert rules (metric alert rule) endpoint serializers underneath the `incidents` folder


evanpurkhiser:
Incidents are different from Alert runes. Do we want to do this?

evanpurkhiser:
Or are these the incident serializers? The naming is so confusing lol

src/sentry/incidents/charts.py
```diff
@@ -9,11 +9,11 @@
 from sentry.api import client
 from sentry.api.base import logger
 from sentry.api.serializers import serialize
-from sentry.api.serializers.models.alert_rule import AlertRuleSerializer
-from sentry.api.serializers.models.incident import DetailedIncidentSerializer
 from sentry.api.utils import get_datetime_from_stats_period
 from sentry.charts import backend as charts
 from sentry.charts.types import ChartSize, ChartType
+from sentry.incidents.endpoints.incident import DetailedIncidentSerializer
```
saponifi3d, line 15: it looks like we're changing adding `incidents` (👍), but also changing `models` to `endpoints`. Could you explain that a little? just curious if it's meant to represent an inbound/outbound filter or if it's an existing convention or? (Thanks!)

nhsiehgit: Ah!

wait thismight be an oops.
one moment

nhsiehgit:
Hey @evanpurkhiser 

Ya - unfortunately naming is terrible
All the AlertRule logic stuff is shuttled under `incidents` :( 

can make it a follow up task to rename the folder to AlertRules instead

nhsiehgit:
Probably need to break out the actual `incident` api/model stuff with `alert_rule` api/model stuff 🤔

saponifi3d (approved):
lgtm, thanks for doing this :)

PR #70909, opened 2024-05-14
chore(api): Created Test for Path Params

iamrajjoshi:
Here, I introduce a test to make sure endpoint in `sentry` always have `_id_or_slug` in their path params.

The test is based of the convert_args tests I have written before.

I find all the urls and if the url class isn't allowlisted (if its web or auth), I iterate through all the params to make sure it doesn't end with _slug.

If the check fails, we throw an error like:
image


iamrajjoshi:
CI is failing bc a endpoint was recently merged that didn't follow the new convention!

schew2381:
Nice test, I see why we need this cuz the `ProjectBackfillSimilarIssuesEmbeddingsRecords` was just added with just `project_slug`

schew2381 (approved):
lgtm, have to fix the `ProjectBackfillSimilarIssuesEmbeddingsRecords` endpoint first tho before merging this

PR #70913, opened 2024-05-14
feat(api-idorslug): Removed FF Wrapper Function Call

iamrajjoshi:
Here I am removing the wrapper function call as the option has been GA'ed close to a month now. I'll delete the actual function in another pr once I remove calls to the function from `getsentry`


iamrajjoshi:
Tests are failing because of a typo in one of our endpoints, fix here: https://github.com/getsentry/sentry/pull/70918

src/sentry/api/endpoints/organization_region.py
```diff
@@ -65,12 +64,7 @@ def convert_args(
...
         try:
             # We don't use the lookup since OrganizationMapping uses a BigIntField for organization_id instead of a ForeignKey
-            if (
-                id_or_slug_path_params_enabled(
-                    self.convert_args.__qualname__, str(organization_id_or_slug)
-                )
-                and str(organization_id_or_slug).isdecimal()
-            ):
+            if str(organization_id_or_slug).isdecimal():
                 org_mapping = OrganizationMapping.objects.get(
```
schew2381, line 68: Could you remind me again why we don't use the lookup for `OrganizationMapping`

schew2381 (approved):

src/sentry/api/endpoints/organization_region.py
iamrajjoshi: OrganizationMapping is a little different where it doesn't identify an organization by id, rather `organization_id`. It is a 1-off scenario so instead of making another lookup, we just added the logic in the endpoint.

PR #70916, opened 2024-05-15
fix(billing): Fix plural of profileDuration, remove default

scttcper:
Change constant, add comment explaining when plural vs DataCategory would be used


static/app/constants/index.tsx
```diff
@@ -329,7 +329,7 @@ export const DATA_CATEGORY_INFO = {
   [DataCategoryExact.PROFILE_DURATION]: {
     name: DataCategoryExact.PROFILE_DURATION,
     apiName: 'profileSeconds',
-    plural: 'profileSeconds',
+    plural: 'profileDuration',
     displayName: 'profile seconds',
     titleName: t('Profile Seconds'),
```
vbro, line 334: The apiName is `profileDuration` and we display `profile hours`.
See https://github.com/getsentry/getsentry/blob/master/getsentry/billing/utils/data_category.py#L209-L218

static/app/types/core.tsx
```diff
@@ -67,12 +67,11 @@ export type Choice = [
...
 /**
- * This legacy type used plurals which will cause compatibility issues when categories
- * become more complex, e.g. processed transactions, session replays. Instead, access these values
- * with `DATA_CATEGORY_INFO[category].plural`, where category is the `DataCategoryExact` enum value.
+ * These are very similar to the plural types of DATA_CATEGORY_INFO.
+ * DATA_CATEGORY_INFO and DataCategoryExact have additional categories
+ * that are used in stats but not other places like billing.
  */
 export enum DataCategory {
-  DEFAULT = 'default',
```
vbro, line 75: What is the default for?

scttcper: it seems unused, but we can keep it for now

markstory: We have a DataCategory called DEFAULT. It gets used for error events that don't have stack traces like from `capture_message`. Default events consume error quota.

vbro: Perhaps we should investigate further and remove `DataCategory.DEFAULT` in a separate PR if necessary.

vbro (approved):

PR #70917, opened 2024-05-15
fix(similarity): allow single_org sentry users to not be superuser to call backfill

JoshFerge:
still enforces org / project permissions

### DIFF
--- a/src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
+++ b/src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from rest_framework.request import Request
 from rest_framework.response import Response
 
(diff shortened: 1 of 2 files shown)


src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
```diff
@@ -18,9 +19,9 @@ class ProjectBackfillSimilarIssuesEmbeddingsRecords(ProjectEndpoint):
     }
 
     def post(self, request: Request, project) -> Response:
-        if not features.has(
-            "projects:similarity-embeddings-backfill", project
-        ) or not is_active_superuser(request):
+        if not features.has("projects:similarity-embeddings-backfill", project) or (
+            not is_active_superuser(request) and not settings.SENTRY_SINGLE_ORGANIZATION
+        ):
```
vartec, line 24: As I understand it, you're allowed to go through if any of these conditions are true. 
This would make it more obvious. 

```suggestion
        if not (
           features.has("projects:similarity-embeddings-backfill", project)
           or is_active_superuser(request)
           or settings.SENTRY_SINGLE_ORGANIZATION
        ):
```

vartec (approved):

src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
JoshFerge: this is not quite right, but the conditional is confusing.

you need to have the flag, and either sentry_single_organization, or is_active_superuser. i can follow up with a rewrite of this conditional.

JoshFerge: let me make this less confusing

JoshFerge: kk, updated

PR #70941, opened 2024-05-15
fix(insights): Improve span sample trace data fetching

gggritso:
Colin spotted an issue where some span sample links were broken, the trace ID is missing from the URL. This happens because of our fetching strategy, which is:

1. Fetch a list of span samples
2. Find transactions for each of the span IDs in the sample list
3. Merge transaction information into the span information
4. Render the table
5. Construct URL from trace information that's in the transaction information

This doesn't work nicely because the call to fetch transactions for spans might fail or return incomplete information. Since the trace ID and transaction ID are in the spans dataset, too, fetch them from there instead.

I also snuck in a change to how `additionalFields` are declared, because otherwise it was _very_ awkward to mash the extra code in there.

P.S. I didn't see any actual use of the transaction data in any table, so we can probably completely rip that out later.

### DIFF
--- a/static/app/views/performance/browser/resources/resourceSummaryPage/index.tsx
+++ b/static/app/views/performance/browser/resources/resourceSummaryPage/index.tsx
@@ -148,7 +148,6 @@ function ResourceSummary() {
             groupId={groupId}
             moduleName={ModuleName.RESOURCE}
             transactionName={transaction as string}
-            additionalFields={[HTTP_RESPONSE_CONTENT_LENGTH]}
           />
         </Layout.Main>
       </Layout.Body>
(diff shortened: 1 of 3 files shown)


DominikB2014 (approved):
This makes a lot more sense to me

static/app/views/performance/browser/resources/resourceSummaryPage/index.tsx
```diff
@@ -148,7 +148,6 @@ function ResourceSummary() {
             groupId={groupId}
             moduleName={ModuleName.RESOURCE}
             transactionName={transaction as string}
-            additionalFields={[HTTP_RESPONSE_CONTENT_LENGTH]}
```
DominikB2014, line 151: Should we also determine the transaction route prop from the module name within the component to be consistent?

DominikB2014 (approved):

static/app/views/performance/browser/resources/resourceSummaryPage/index.tsx
gggritso: That makes sense! I think this whole sample panel could use an overhaul though

PR #70951, opened 2024-05-15
ref(insights): rename resource module to asset module

DominikB2014:
Renaming the `resources` module to `assets` resonates better with FE devs. This does all the renaming in sentry itself, and there's another PR up for docs https://github.com/getsentry/sentry-docs/pull/10047


static/app/views/performance/browser/resources/settings.ts
```diff
@@ -1,7 +1,11 @@
 import {t} from 'sentry/locale';
 
-export const MODULE_TITLE = t('Resources');
+export const MODULE_TITLE = t('Assets');
 export const BASE_URL = 'browser/resources';
+export const DATA_TYPE = t('Asset'); // Name of the data shown (singular)
```
DominikB2014, line 5: Had trouble finding a decent name for this, but its a variable that represents the singular version of the type of data on the page. It maps to the span, i.e were showing `Asset` spans in this module

static/app/routes.tsx
```diff
@@ -1505,6 +1508,22 @@ function buildRoutes() {
...
+        <Route
+          path="spans/span/:groupId/"
+          component={make(
+            () =>
+              import(
+                'sentry/views/performance/browser/resources/resourceSummaryPage/index'
+              )
+          )}
+        />
+      </Route>
```
DominikB2014, line 1526: `/browser/assets` and `/browser/resources` will just map to the same component so we don't have to rely on the organization to setup routes

static/app/views/performance/browser/resources/settings.ts
```diff
@@ -1,9 +1,16 @@
...
 export const MODULE_DESCRIPTION = t(
   'Find large and slow-to-load resources used by your application and understand their impact on page performance.'
 );
+
+export const PERFORMANCE_MODULE_TITLE = t('Resources');
+export const PERFORMANCE_BASE_URL = 'browser/resources';
+export const PERFORMANCE_DATA_TYPE = t('Resource');
+export const PERFORMANCE_MODULE_DESCRIPTION = t(
+  'Find large and slow-to-load resources used by your application and understand their impact on page performance.'
+);
```
DominikB2014, line 14: Prefixed with `Performance` so that once we move to insights, we just delete these and it all makes sense still

static/app/views/performance/utils/useModuleURL.tsx
```diff
@@ -65,6 +73,14 @@ export function useModuleURLBuilder(bare: boolean = false): URLBuilder {
...
     }
 
+    if (!isInsightsEnabled && PERFORMANCE_MODULE_BASE_URLS[moduleName]) {
+      return bare
+        ? `${insightsURL}/${PERFORMANCE_MODULE_BASE_URLS[moduleName]}`
+        : normalizeUrl(
+            `/organizations/${slug}/${insightsURL}/${PERFORMANCE_MODULE_BASE_URLS[moduleName]}`
+          );
+    }
+
```
gggritso, line 83: This is pretty confusing. Can this be expressed the same way as the AI logic above? e.g., "if the module is Resources, set the correct base URL either `/assets` or `/resources`, and then concatenate". This stuff with checking `PERFORMANCE_MODULE_BASE_URLS` is fine but it's so much more confusing than just having special logic for Resources, which is what's happening here

static/app/components/sidebar/index.tsx
```diff
@@ -355,10 +360,10 @@ function Sidebar() {
   );
 
   const resources = hasOrganization && (
-    <Feature key="resource" features="spans-first-ui">
+    <Feature key="assets" features="spans-first-ui">
```
gggritso, line 363:
```suggestion
    <Feature key="resource" features="spans-first-ui">
```

You can leave it as-is, so it matches with the definition of `ModuleName` enum

gggritso (commented):
Makes sense! Two nits on confusing conditionals, otherwise LGTM!

static/app/views/performance/utils/useModuleTitle.tsx
```diff
@@ -25,9 +29,20 @@ export const MODULE_TITLES: Record<ModuleName, string> = {
...
 type TitleableModuleNames = Exclude<ModuleNameStrings, '' | 'other'>;
 
 export function useModuleTitle(moduleName: TitleableModuleNames) {
+  const organization = useOrganization({allowNull: true});
+  const isInsightsEnabled = organization?.features?.includes('performance-insights');
+
+  if (!isInsightsEnabled && PERFORMANCE_MODULE_TITLES[moduleName]) {
+    return PERFORMANCE_MODULE_TITLES[moduleName];
+  }
+
```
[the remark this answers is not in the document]
DominikB2014: done!

gggritso (approved):

Issue #70953, opened 2024-05-15
Evolved Issues Search: Add basic support for paren characters in query


malwilley:
This has two parts:

1. Modify the search parser to parse parens separately and not error out when there are mismatched parens
2. Implement rendering of parens in the query builder component

[Internal Figma](https://www.figma.com/design/qLPE7uSdRZMOJloVumv8rA/Q1-Smart-Search?node-id=629-1329&t=CMywyt4xjEATCQgg-0)

![CleanShot 2024-05-15 at 09 36 15](https://github.com/getsentry/sentry/assets/10888943/ed28972e-11c0-45da-8a02-1c94408561eb)

PR #70954, opened 2024-05-15
feat(search-syntax): Add option to flatten paren groups

malwilley:
Ref https://github.com/getsentry/sentry/issues/70953

For the new search component, parens will need to be parsed as separate tokens and the parser cannot error out when faced with mismatched parens. Therefore the parser should be updated to allow for this.

I've added a new token type, `PAREN`, which may be returned as part of the parse result. When there are mismatched parentheses, this is now matched which allows the parser to return a successful result. A new config option has been added, `flattenParenGroups` (name could be workshopped), which causes the `PAREN` token to take precedence over `LOGIC_GROUP` and thus return parens separately from the contents when enabled. When disabled (the default), logic groups are still returned as you would expect.

Previous behavior:

- Unmatched parens will throw an error and return nothing from the parser
- Matched parens are parsed as a `LOGIC_GROUP` with boolean operators, free text, and filters inside the `token.inner` array.

New behavior (with the default `flattenParenGroups=false`):

 - Unmatched parens will not throw an error. The parser will return a normal result, with the unmatched parens parsed as a `PAREN` token.
 - Matched parens will return a `LOGIC_GROUP` as normal.

New behavior (with the new `flattenParenGroups=true`):

-  Unmatched parens will not throw an error. The parser will return a normal result, with the unmatched parens parsed as a `PAREN` token.
- Matched parens will _not_ return a `LOGIC_GROUP`

You can see now that syntax highlighting is no longer broken when adding a mismatched paren:

CleanShot 2024-05-15 at 09 47 04
CleanShot 2024-05-15 at 09 47 18


static/app/components/searchSyntax/parser.tsx
```diff
@@ -415,6 +416,14 @@ export class TokenConverter {
     };
   };
 
+  tokenParen = (value: '(' | ')') => ({
+    ...this.defaultTokenFields,
+    value,
+    left: value === '(',
+    right: value === ')',
+    type: Token.PAREN as const,
```
evanpurkhiser, line 424: Small nit, usually the type comes right after `defaultTokenFields`

```suggestion
    ...this.defaultTokenFields,
    type: Token.PAREN as const,
    value,
    left: value === '(',
    right: value === ')',
```

evanpurkhiser (approved):

static/app/components/searchSyntax/parser.tsx
```diff
@@ -1173,6 +1191,10 @@ export type SearchConfig = {
    * Text filter keys we allow to have operators
    */
   textOperatorKeys: Set<string>;
+  /**
+   * When true, the parser will not parse paren groups and will return individual paren tokens
+   */
+  flattenParenGroups?: boolean;
```
JonasBa, line 1197: Out of those scope here, but something to think of for future work is that we could benefit from having the parsing logic not try to generate an AST. Right now the grammar does both lexing and parsing, which makes error recovery (as seen here) tricky to handle. 

We actually want the tokenization pass to succeed regardless if the syntax is valid, as it makes recovering from errors much easier (validation and AST generation is trivial once we have the list of tokens).

static/app/components/searchSyntax/parser.tsx
JonasBa, line 424: I would just use Token.LPAREN | Token.RPAREN
```suggestion
    type: Token.LPAREN | Token.RPAREN
```
The left/right fields will end up in code like Token.type === Token.PAREN && token.left which isnt terrible, but makes it a bit less ergonomic if you were to use the Token.Paren in something like a switch statement.

static/app/components/searchSyntax/parser.tsx
malwilley: Agreed. After this change there are actually very few things that can cause the parser to fail completely, but I agree separating the two would be simpler conceptually and makes things a bit easier on ourselves.

static/app/components/searchSyntax/parser.tsx
malwilley: Ah yeah, I do like that better. Changed

malwilley:
@evanpurkhiser I found a fixed a couple other edge cases I found if you want to take another look!

static/app/components/searchSyntax/parser.tsx
```diff
@@ -415,6 +417,12 @@ export class TokenConverter {
     };
   };
 
+  tokenParen = (value: '(' | ')') => ({
+    ...this.defaultTokenFields,
+    type: value === '(' ? (Token.L_PAREN as const) : (Token.R_PAREN as const),
+    value,
+  });
```
evanpurkhiser, line 424: IMO have two token functions here, makes it a bit easier to understand

```suggestion
  lTokenParen = (value: '(') => ({
    ...this.defaultTokenFields,
    type: Token.L_PAREN as const
    value,
  });
  
  rTokenParen = (value: ')') => ({
    ...this.defaultTokenFields,
    type: Token.R_PAREN as const
    value,
  });
```

PR #70958, opened 2024-05-15
feat(flagpole): Adds bulk option lookup with flag filter

GabeVillalobos:
Adds a new `bulk_retrieve_options` method to OptionsManager to allow flagpole to bulk query matching feature options.

This isn't super optimized as it will invoke the `OptionsManager.get()` method for every registry key match, meaning this could result in multiple DB round trips depending on what lives in the options manager and store caches. If this becomes problematic, we can update the bulk retrieval method to bulk query the store and caches accordingly, but would require us to reimplement this complex get logic.

### DIFF
--- a/src/sentry/options/manager.py
+++ b/src/sentry/options/manager.py
@@ -514,7 +514,7 @@ class OptionsManager:
 
         return None
 
-    def bulk_retrieve_options(self, keys: list[str], flag: int | None = None) -> dict[str, any]:
+    def bulk_retrieve_options(self, keys: list[str], flag: int | None = None) -> dict[str, Any]:
         key_set = set(keys)
         keys = []
 
--- a/src/sentry/options/manager.py
+++ b/src/sentry/options/manager.py
@@ -514,7 +514,7 @@ class OptionsManager:
 
         return None
 
-    def bulk_retrieve_options(self, keys: list[str], flag: int | None = None) -> dict[str, Any]:
+    def bulk_retrieve_options(self, keys: list[str], flag: int | None = None):
         key_set = set(keys)
         keys = []
 
@@ -529,5 +529,5 @@ class OptionsManager:
 
         return self.bulk_retrieve_options_by_key(keys)
 
-    def bulk_retrieve_options_by_key(self, keys) -> dict[str, any]:
+    def bulk_retrieve_options_by_key(self, keys):
         return {key.name: self.get(key.name) for key in keys}
--- a/src/sentry/options/manager.py
+++ b/src/sentry/options/manager.py
@@ -108,6 +108,8 @@ FLAG_BOOL = 1 << 10
 FLAG_AUTOMATOR_MODIFIABLE = 1 << 11
 # Values that are scalar numeric integer values
 FLAG_SCALAR = 1 << 12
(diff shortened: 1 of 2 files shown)


src/sentry/options/manager.py
```diff
@@ -108,6 +108,8 @@ class UnknownOption(KeyError):
 FLAG_AUTOMATOR_MODIFIABLE = 1 << 11
 # Values that are scalar numeric integer values
 FLAG_SCALAR = 1 << 12
+# Values that are meant to be used by the flagpole feature handler
+FLAG_FLAGPOLE_FEATURE = 1 << 13
```
markstory, line 112: Would we use this flag for options that are implicitly registered when features are defined?

src/sentry/options/manager.py
```diff
@@ -511,3 +513,21 @@ def can_update(self, key: str, value, channel: UpdateChannel) -> NotWritableReas
...
+    def bulk_retrieve_options(self, keys: list[str], flag: int | None = None):
+        key_set = set(keys)
+        keys = []
+
+        for key, value in self.registry.items():
+            flag_matches = flag is None or value.flags & flag
+            key_in_set = key in key_set
+
+            if not key_in_set or not flag_matches:
+                continue
```
markstory, line 526: Should we raise an error here? Silently not fetching an option because the flags are wrong could result in KeyErrors in the caller as they expect to get the options they requested.

GabeVillalobos: I don't think we should. In flagpole we want to skip any flags that aren't defined, so in a bulk flag lookup scenario we'll end up raising an exception and evaluating all flags to None which doesn't seem correct. Alternatively, we could just make flagpole individually query options since we're doing that here anyway, avoiding this bulk behavior entirely.

src/sentry/options/manager.py
GabeVillalobos: Yep, that's the idea. I'll be baking this functionality into the feature registration process in getsentry.

src/sentry/options/manager.py
markstory: >  Alternatively, we could just make flagpole individually query options since we're doing that here anyway,

I think that might be the simpler solution. If we prefix the options that flagpole uses we don't need to expand the option flags which are already complicated and have some poorly defined intersections already.

kneeyo1: `configoptions` uses `filter`, which does something very similar (return a list of options filtered on a flag). 
Should we combine the two?

src/sentry/options/manager.py
```diff
@@ -511,3 +513,21 @@ def can_update(self, key: str, value, channel: UpdateChannel) -> NotWritableReas
             return NotWritableReason.DRIFTED
 
         return None
+
+    def bulk_retrieve_options(self, keys: list[str], flag: int | None = None):
```
kneeyo1, line 517: `configoptions` uses `filter`, which does something very similar (return a list of options filtered on a flag). 
Should we combine the two?

src/sentry/options/manager.py
GabeVillalobos: Agreed, closing this change.

PR #70960, opened 2024-05-15
fix(similarity): pass dry run on 

JoshFerge:
backfill should pass dry run on to the next job. also adds some more logs


roggenkemper (approved):

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -58,7 +58,6 @@ class GroupStacktraceData(TypedDict):
     soft_time_limit=60 * 15,
     time_limit=60 * 15 + 5,
 )
-@metrics.wraps(f"{BACKFILL_NAME}.task")
```
JoshFerge, line 61: removing this as the decorator above should supply us with a metric

PR #70962, opened 2024-05-15
chore(metrics-layer): Clean up meta queries to always use storage direct queries

enochtangg:
After rolling out meta queries to directly query a storage in snuba, we can remove the sentry option and the logic that samples these queries.


kylemumma (approved):

volokluev (approved):
🎉 🎉 🎉 🎉

PR #70966, opened 2024-05-15
ref(groupChart): Refactor GroupChart for readability, typing, and scoping

MichaelSun48:
This PR refactors the `<GroupChart>` component to have better typing, and to be more readable. It also removes the `group` prop, since the scope of the entire `group` object is beyond what the component needs. Specific details are included the comments. 

**No design or visual changes should be included in this PR. If you browse the preview and see any visual changes to the events graph, please comment about it and I will rectify.**

High level changes:

- Updates all instances of `groupChart` to reflect refactor
- Replaces `group` and `statsPeriod` props in favor of stats and secondary stats data itself
- Removes all uses of `let` in groupChart
- Better typing

### DIFF
--- a/static/app/components/stream/group.tsx
+++ b/static/app/components/stream/group.tsx
@@ -38,6 +38,7 @@ import type {
   NewQuery,
   Organization,
   PriorityLevel,
+  TimeseriesValue,
   User,
 } from 'sentry/types';
 import {IssueCategory} from 'sentry/types/group';
@@ -438,6 +439,14 @@ function BaseGroupRow({
     <GuideAnchor target="issue_stream" />
   );
 
+  const groupStats: ReadonlyArray<TimeseriesValue> = group.filtered
+    ? group.filtered.stats?.[statsPeriod]
+    : group.stats?.[statsPeriod];
+
+  const groupSecondaryStats: ReadonlyArray<TimeseriesValue> = group.filtered
+    ? group.stats?.[statsPeriod]
+    : [];
+
   return (
     <Wrapper
       data-test-id="group"
@@ -472,8 +481,8 @@ function BaseGroupRow({
       {withChart && !displayReprocessingLayout && issueTypeConfig.stats.enabled && (
         <ChartWrapper narrowGroups={narrowGroups}>
           <GroupChart
-            statsPeriod={statsPeriod!}
-            data={group}
+            stats={groupStats}
+            secondaryStats={groupSecondaryStats}
             showSecondaryPoints={showSecondaryPoints}
             showMarkLine
           />
--- a/static/app/components/stream/groupChart.tsx
+++ b/static/app/components/stream/groupChart.tsx
@@ -4,7 +4,7 @@ import MarkLine from 'sentry/components/charts/components/markLine';
 import MiniBarChart from 'sentry/components/charts/miniBarChart';
 import {LazyRender} from 'sentry/components/lazyRender';
 import {t} from 'sentry/locale';
-import type {Group, TimeseriesValue} from 'sentry/types';
+import type {TimeseriesValue} from 'sentry/types';
 import type {Series} from 'sentry/types/echarts';
 import {formatAbbreviatedNumber} from 'sentry/utils/formatters';
 import theme from 'sentry/utils/theme';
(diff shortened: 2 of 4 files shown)


static/app/components/stream/group.tsx
```diff
@@ -472,8 +481,8 @@ function BaseGroupRow({
       {withChart && !displayReprocessingLayout && issueTypeConfig.stats.enabled && (
         <ChartWrapper narrowGroups={narrowGroups}>
           <GroupChart
-            statsPeriod={statsPeriod!}
-            data={group}
+            stats={groupStats}
+            secondaryStats={groupSecondaryStats}
```
MichaelSun48, line 485: asdf

static/app/components/stream/groupChart.tsx
```diff
@@ -19,70 +19,49 @@ function asChartPoint(point: [number, number]): {name: number | string; value: n
 const EMPTY_STATS: ReadonlyArray<TimeseriesValue> = [];
 
 type Props = {
-  data: Group;
-  statsPeriod: string;
+  stats: ReadonlyArray<TimeseriesValue>;
   height?: number;
+  secondaryStats?: ReadonlyArray<TimeseriesValue>;
   showMarkLine?: boolean;
   showSecondaryPoints?: boolean;
```
MichaelSun48, line 26: The `statsPeriod` props was only being used to extract the correct series from the `group` prop. The GroupChart component shouldn't care about the group itself or which time period is being displayed, it should only care about and receive the actual series data itself, hence why the `stats` and `secondaryStats` props replace this. 

This delegates the logic of extracting the series from the group itself to the parent component of groupChart

static/app/components/stream/groupChart.tsx
```diff
@@ -19,70 +19,49 @@ function asChartPoint(point: [number, number]): {name: number | string; value: n
...
   showMarkLine = false,
 }: Props) {
-  const stats: ReadonlyArray<TimeseriesValue> = statsPeriod
-    ? data.filtered
-      ? data.filtered.stats?.[statsPeriod]
-      : data.stats?.[statsPeriod]
-    : EMPTY_STATS;
-
-  const secondaryStats: TimeseriesValue[] | null =
-    statsPeriod && data.filtered ? data.stats[statsPeriod] : null;
```
MichaelSun48, line 43: This was previously the logic to extract the stats from the group. As mentioned in the above comment, this is now delegated to the parent to limit the scope of this component.

static/app/views/discover/eventDetails/linkedIssue.tsx
```diff
@@ -12,6 +12,7 @@ import ShortId from 'sentry/components/shortId';
 import GroupChart from 'sentry/components/stream/groupChart';
```
MichaelSun48, line 12: I don't believe this component is actually being used anywhere, but out of an abundance of caution, I have still refactored this instance of groupChart

static/app/components/stream/groupChart.tsx
```diff
@@ -19,70 +19,49 @@ function asChartPoint(point: [number, number]): {name: number | string; value: n
...
-    let max = 0;
-    const marklinePoints: number[] = [];
-
-    for (let i = 0; i < stats.length; i++) {
-      const point = stats[i];
-      if (point[1] > max) {
-        max = point[1];
-      }
-      marklinePoints.push(point[1]);
-    }
```
MichaelSun48, line 63: Not sure what the purpose of this marklinePoints array was, but it was not being used anywhere.

static/app/components/stream/group.tsx
```diff
@@ -438,6 +439,14 @@ function BaseGroupRow({
     <GuideAnchor target="issue_stream" />
   );
 
+  const groupStats: ReadonlyArray<TimeseriesValue> = group.filtered
+    ? group.filtered.stats?.[statsPeriod]
+    : group.stats?.[statsPeriod];
+
+  const groupSecondaryStats: ReadonlyArray<TimeseriesValue> = group.filtered
+    ? group.stats?.[statsPeriod]
+    : [];
```
MichaelSun48, line 448: This logic mirrors lines 36-43 in the old `groupChart.tsx`, and is one example of extracting the stats data in the parent component rather than in the groupChart component

static/app/views/discover/eventDetails/linkedIssue.tsx
malwilley: Hmm it looks like there must have been a recent change that redirects from the discover event details to the issue details page, but I don't know where that redirect is happening or why it was added. Probably why you can't find this anywhere.

malwilley (approved):
Looks good! Thanks for going back and refactoring

PR #70967, opened 2024-05-15
feat(replay): viewed_by_me filter

aliu39:
Follow up to https://github.com/getsentry/sentry/issues/64924. We'll support this search filter so users can filter by replays they have or haven't viewed. Previously this was hard because: the viewed_by_id column in Snuba is populated by Sentry user ids (from Postgres), which cannot be looked up from the web app unless you're a superuser.

(Definitely my first time making this PR.)


src/sentry/replays/usecases/query/__init__.py
```diff
@@ -43,6 +43,52 @@
...
+        # since the value is boolean, negations (!) are not supported
+        if search_filter.operator != "=":
+            raise ParseError(f"Invalid operator specified for `{search_filter.key.name}`")
+
+        value = search_filter.value.value
+        if not isinstance(value, str) or value.lower() not in ["true", "false"]:
+            raise ParseError(f"Could not parse value for `{search_filter.key.name}`")
+        value = value.lower() == "true"
+
+        if request_user_id is None:
```
aliu3ntry, line 72: Decided with @cmanallen not going to test this block. It's a very rare edge case that can only be triggered running the `delete_replays` script with viewed_by_me as a tag filter, which we don't expect to happen.

aliu3ntry: The only call site for this function is in `query_using_optimized_search`, and `request_user` is passed down by either 1) the replay index endpoint or 2) delete_replays script. We expect that future callers are exclusively Django endpoints, which automatically return 401 unauthorized for anonymous/not-logged-in users. So _other than the delete script, `request_user_id` will always be non-None._

aliu3ntry: Could add a comment for this?

src/sentry/replays/usecases/query/__init__.py
```diff
@@ -43,6 +43,52 @@
...
+            if value:  # use an always-false filter so query returns nothing
+                return [
+                    SearchFilter(SearchKey("id"), "=", SearchValue("a" * 32)),
+                    "AND",
+                    SearchFilter(SearchKey("id"), "!=", SearchValue("a" * 32)),
+                ]
+            return []
```
cmanallen, line 79: Let's remove everything inside this condition and just raise an Exception.  This error should only be reached in the event of a programmer error.  So lets emit an error to sentry and fix it if it ever comes up.  `raise ValueError("Expected user_id")` or something.

cmanallen (approved):

cmanallen (approved):

PR #70971, opened 2024-05-15
feat(dashboards): Add remaining span metrics to metric samples endpoint

narsaynorath:
To display span metrics properly in dashboards we also need to expose the samples on the widget builder. This involves updating the span sample executor to accept the remaining span MRIs.


tests/sentry/api/endpoints/test_organization_metrics.py
```diff
@@ -288,6 +289,7 @@ def test_span_measurement_samples(self):
 
         for i, mri in enumerate(
             [
+                "g:spans/mobile.slow_frames@none",
```
narsaynorath, line 292: I'm not sure if it's meaningful for me to update the tests further considering the other MRIs would use the same kind of mechanism for this test

obostjancic (approved):
lgtm

PR #70973, opened 2024-05-15
ref(insights): Dynamically generate module URLs everywhere

gggritso:
## tl;dr

Right now: Hardcoded URLs like `/organization/${slug}/performance/browser/resources/spans/`
With this PR: `${moduleURL}/spans/`

## Context

`useModuleURL` and its more specific siblings like `useResourcesModuleURL` check some config variables (the current base URL of Insights, and the current base URL of the relevant module) and generate the prefix automatically.

This is great because:

1. Easier to change URLs. e.g., if we rename a module (happens pretty often), the config variable change is the only thing needed
2. Easier to feature-flag URLs. e.g., when we move insights to `/insights/` I can generate the base URL based on a feature flag

I went through the app and tried to find every single instance of a hard-coded URL and replaced it with this module URL helper.

## Addendum on `baseURL`

You might be wondering "what about `baseURL`"? I'm getting rid of it. As of this PR, `baseURL` is no longer respected by anything. The only modules that respected it were Resources and Queries, and only in one spot

`baseURL` made sense when a single module had to be routed in _multiple contexts at the same time_ (e.g., when `/starfish/database` and `/performance/database` _both_ had to work) but that's no longer true. I hate `baseURL` because of its spooky action at a distance. You misconfigure `baseURL` in one spot, and URLs break somewhere far away. That's no way to live.

I'll delete `RoutingContext` very soon.


static/app/views/aiMonitoring/PipelinesTable.tsx
```diff
@@ -237,9 +240,7 @@ function renderBodyCell(
 
     return (
       <Link
-        to={normalizeUrl(
-          `/organizations/${organization.slug}/ai-monitoring/pipeline-type/${row['span.group']}?${qs.stringify(queryString)}`
-        )}
+        to={`${moduleURL}/pipeline-type/${row['span.group']}?${qs.stringify(queryString)}`}
```
DominikB2014, line 243: 🤔 I wonder if we can expand `useModuleUrl` even more to return us the sub paths too, because now we have the same problem with renaming with `pipeline-type` (and anything after).

Maybe 
```
{getModuleUrl} = useModuleUrl();
const url = getModuleUrl(ModuleName.AI, 'pipeline-type')
```
The second param could be optional

DominikB2014 (approved):
👍

static/app/views/aiMonitoring/PipelinesTable.tsx
gggritso: Yeah, I thought about that! Definitely possible. It'll look a little awkward because these helpers are hooks, so they can't be used conditionally, and the return value has to be passed through helper functions. For just module URLs it's not _too_ bad, but even that is awkward. I wonder if later versions of React Router have built-it helpers for this, that'd be nice

static/app/views/performance/utils/useModuleURL.tsx
```diff
@@ -0,0 +1,55 @@
...
+    // AI Doesn't live under "/performance"
+    return normalizeUrl(`/organizations/${slug}${AI_BASE_URL}`);
+  }
+
+  return normalizeUrl(
+    `/organizations/${slug}${INSIGHTS_BASE_URL}/${MODULE_BASE_URLS[moduleName]}`
+  );
+};
+
+export const useDatabaseModuleURL = partial(useModuleURL, ModuleName.DB);
```
JonasBa, line 46: From discoverability POV, not everyone will know the names of these modules which will make it less likely for them to reach for their usage as the different variations will be fragmented around the codebase and make it more likely that they will hardcode the URLs like they are used to today (or add new module definitions)

Have you thought about changing the hook signature to accept the module name as a string? I think that would help with eliminating the need to import the name enum while preserving type safety and also improving the typeahead experience when using the hook.

From a DX pov, the use of enums can sometimes be overly restrictive, which I suspect is part of why these partial types even exist here in the first place.

gggritso: @JonasBa :eyes:

> From discoverability POV, not everyone will know the names of these modules which will make it less likely for them to reach for their usage as the different variations will be fragmented around the codebase and make it more likely that they will hardcode the URLs like they are used to today (or add new module definitions)

I don't follow! What do you mean by "the different variations will be fragmented"? Also, how does not knowing the module names result in more people hardcoding the URLs or adding new module definitions?

Do you mean that people might add a new module but not add the partial URL function? That's fair if yes. Or that people will use the partials sometimes, and `useModuleURL` other times? Is that bad?

Also, I'd _love_ to fully prevent hard-coded URLs, but I don't know of a lightweight way of doing that!

> Have you thought about changing the hook signature to accept the module name as a string? I think that would help with eliminating the need to import the name enum while preserving type safety and also improving the typeahead experience when using the hook.

That sounds reasonable! I think changing `ModuleName` to a string union type instead of an enum is probably a good idea overall. I'm happy to do that in a follow-up.

> From a DX pov, the use of enums can sometimes be overly restrictive, which I suspect is part of why these partial types even exist here in the first place.

It is not. The partial functions exist because for one thing, that's the API I prefer _and_ because there was already an existing partial hook (`useCacheUrl`) so I set up all partials for consistency. I don't particularly love the partials, but they seem better than always importing `useModuleURL`. I'm fine with either though, if you really don't like the partial!

PR #70975, opened 2024-05-15
Fix: options manager can_update bug

kneeyo1:
Previously if the option value itself is falsy, the options manager module incorrectly assumes that the option isn't set on the disk. 

Examples: `0`, `""`, `false`
This fixes that behavior.


src/sentry/options/manager.py
```diff
@@ -476,7 +476,7 @@ def can_update(self, key: str, value, channel: UpdateChannel) -> NotWritableReas
         opt = self.lookup_key(key)
         if opt.has_any_flag({FLAG_NOSTORE, FLAG_IMMUTABLE}):
             return NotWritableReason.READONLY
-        if opt.has_any_flag({FLAG_PRIORITIZE_DISK}) and settings.SENTRY_OPTIONS.get(key):
+        if opt.has_any_flag({FLAG_PRIORITIZE_DISK}) and key in settings.SENTRY_OPTIONS:
```
kneeyo1, line 479: I also tried `opt.default() != settings.SENTRY_OPTIONS.get(key)` but if the key doesn't exist it will return None. 
None != most default option values (like `0`, or `""`) 
resulting in an incorrect evaluation that the option is set on disk.

tests/sentry/options/test_manager.py
```diff
@@ -258,6 +259,22 @@ def test_flag_prioritize_disk(self):
...
+        )
+        assert self.manager.get("prioritize_disk_falsy") == 1
+        assert self.manager.can_update("prioritize_disk_falsy", 0, UpdateChannel.AUTOMATOR) is None
+
+        with self.settings(SENTRY_OPTIONS={"prioritize_disk_falsy": 0}):
+            assert self.manager.get("prioritize_disk_falsy") == 0
+            assert (
+                self.manager.can_update("prioritize_disk_falsy", 0, UpdateChannel.AUTOMATOR)
+                == NotWritableReason.OPTION_ON_DISK
+            )
```
markstory, line 276: This will mean that options that could previously be modified at runtime will cease to be mutable. Perhaps that is a benefit though as we shouldn't be changing 'disk' options at runtime.

kneeyo1: yes, this is intended, since it has `FLAG_PRIORITIZE_DISK`.

kneeyo1: As long as there is a value on disk, the option will be immutable.

fpacifici (approved):

markstory (approved):

PR #70976, opened 2024-05-15
ref(js): Listen to location to close modals

evanpurkhiser:
Trying to get rid of `browserHistory.listen` since this will not be available in react router 6


static/app/components/globalModal/index.tsx
```diff
@@ -178,7 +179,8 @@ function GlobalModal({onClose}: Props) {
   }, [portal, handleEscapeClose, visible]);
 
   // Close the modal when the browser history changes
-  useEffect(() => browserHistory.listen(() => actionCloseModal()), []);
+  //
```
scttcper, line 182: ?

scttcper (approved):

static/app/components/globalModal/index.tsx
evanpurkhiser: lol oops

PR #70978, opened 2024-05-15
fix(similarity): more backfill fixes and a few more logs

JoshFerge:
- will delete the redis key after no more groups are found
- adds a few more logs in other cases:
  - when we've already done embeddings for a group, we'll log the count of that
  - if snuba returns no results for whatever reason, we'll log that
  - we'll log the number of updated groups

### DIFF
--- a/src/sentry/tasks/backfill_seer_grouping_records.py
+++ b/src/sentry/tasks/backfill_seer_grouping_records.py
@@ -83,12 +83,14 @@ def backfill_seer_grouping_records(
     redis_client = redis.redis_clusters.get(settings.SENTRY_MONITORS_REDIS_CLUSTER)
     if last_processed_id is None:
         last_processed_id = int(redis_client.get(make_backfill_redis_key(project_id)) or 0)
-        if last_processed_id == 0 and dry_run:
-            logger.info(
-                "backfill_seer_grouping_records.delete_all_seer_records",
-                extra={"project_id": project.id},
-            )
-            delete_grouping_records(project_id)
+
+    if last_processed_id == 0 and dry_run:
+        logger.info(
+            "backfill_seer_grouping_records.delete_all_seer_records",
+            extra={"project_id": project.id},
+        )
+        delete_grouping_records(project_id)
+        redis_client.delete(make_backfill_redis_key(project_id))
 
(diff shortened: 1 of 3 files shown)


src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -186,14 +198,18 @@ def backfill_seer_grouping_records(
...
+                    extra={"project_id": project.id, "num_updated": num_updated},
+                )
 
         last_processed_id = group_id_message_data_batch[len(group_id_message_data_batch) - 1][0]
         redis_client.set(
             f"{make_backfill_redis_key(project_id)}",
             last_processed_id if last_processed_id is not None else 0,
             ex=60 * 60 * 24 * 7,
-        )  # needed for typing
+        )
```
JoshFerge, line 212: deleted this comment, not sure what it was referring to

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -108,13 +114,19 @@ def backfill_seer_grouping_records(
         )
         return
 
-    group_id_message_batch = {
+    group_id_message_batch_filtered = {
         group_id: message
         for (group_id, message, data) in group_id_message_data_batch
-        if not get_path(data, "metadata", "embeddings_info", "nn_model_version")
+        if get_path(data, "metadata", "embeddings_info", "nn_model_version") is not None
```
JoshFerge, line 120: since `nn_model_version` this will always be Falsey. changed the condition to `is not None`

src/sentry/tasks/backfill_seer_grouping_records.py
jangjodi: it was referring to this: `last_processed_id if last_processed_id is not None else 0`. cause we can pass in None, but we always will set it to 0 if it is None. but typing was getting mad. not a big deal to delete it though

jangjodi (approved):

src/sentry/tasks/backfill_seer_grouping_records.py
JoshFerge: ah got it.

PR #70981, opened 2024-05-15
feat(flagpole): Merges Operator class into Conditions for brevity

GabeVillalobos:
Removes the Flagpole `OperatorBase` class and its subclasses, replacing them with subclassed Conditions with similar responsibilities. This flattens our features configurations, giving us a more concise structure overall.


src/flagpole/conditions.py
```diff
@@ -1,25 +1,202 @@
...
+            condition_property=condition_property, segment_name=segment_name
+        )
+
+
+ContainsOperatorValueTypes = StrictInt | StrictStr | StrictFloat
+
+
+class ContainsCondition(ConditionBase):
+    operator: Literal[ConditionOperatorKind.CONTAINS] = ConditionOperatorKind.CONTAINS
+    value: StrictInt | StrictStr | StrictFloat
```
markstory, line 134:
```suggestion
    value: ContainsOperatorValueTypes
```

This should be equivalent.

markstory (approved):

PR #70982, opened 2024-05-15
ref(insights): move sample panel open tracking to panel onOpen

KevinL10:
Track `performance_views.sample_spans.opened` on `<DetailPanel onOpen={...}>` instead of tracking the table clicks that trigger the sample span popup.


narsaynorath (approved):
I noticed there's still a `starfish.panel.open` event in two places that I think we should remove (i.e. the call sites and even the event definition) since the analytics you're adding are better. Let me know if you want to do that, or I could throw up a PR after this is merged.

KevinL10:
I'll take out the outdated starfish analytics

PR #70990, opened 2024-05-15
feat(issues): adds backend for "resolve in upcoming release"

roggenkemper:
this pr adds the backend needed to support "resolve in upcoming release" to the ProjectGroupIndex endpoint. this differs from "resolve in next release" because it no longer does the check for if there is an immediate next release to the current release of the issue. It will still create the `GroupResolution` object which will eventually be cleared in `clear_expired_resolutions` but this will now be the only way for it to be resolved, only when a new release is created.


src/sentry/api/helpers/group_index/update.py
```diff
@@ -300,6 +300,34 @@ def update_groups(
...
+                    {"detail": "Cannot set resolved in next release for multiple projects."},
+                    status=400,
+                )
+            release = (
+                status_details.get("inUpcomingRelease")
+                or Release.objects.filter(
+                    projects=projects[0], organization_id=projects[0].organization_id
+                )
+                .extra(select={"sort": "COALESCE(date_released, date_added)"})
+                .order_by("-sort")[0]
```
JoshFerge, line 315: what does this mean semantically, and is this still the correct behavior for "resolve in upcoming release"? (the case where there is no release passed in.

src/sentry/api/helpers/group_index/validators/status_details.py
```diff
@@ -54,3 +56,19 @@ def validate_inNextRelease(self, value: bool) -> "Release":
             raise serializers.ValidationError(
                 "No release data present in the system to form a basis for 'Next Release'"
             )
+
+    def validate_inUpcomingRelease(self, value: bool) -> "Release":
```
JoshFerge, line 60: please use snake case defining methods (serializer definitions / fields are okay)

src/sentry/api/helpers/group_index/validators/status_details.py
```diff
@@ -54,3 +56,19 @@ def validate_inNextRelease(self, value: bool) -> "Release":
...
+        project = self.context["project"]
+
+        if not features.has("organizations:resolve-in-upcoming-release", project.organization):
```
JoshFerge, line 63: could we add some unit tests for these validator error cases?

src/sentry/api/helpers/group_index/update.py
```diff
@@ -300,6 +300,34 @@ def update_groups(
             res_type = GroupResolution.Type.in_next_release
             res_type_str = "in_next_release"
             res_status = GroupResolution.Status.pending
+        elif status_details.get("inUpcomingRelease"):
+            if len(projects) > 1:
```
JoshFerge, line 304: mind adding a quick test case for this branch?

JoshFerge (commented):
overall looks good, one Q about sort ordering, and a few more tests would be useful. great work!

src/sentry/api/helpers/group_index/update.py
roggenkemper: im actually not sure you're able to hit this branch, the endpoint restricts you to only 1 project, so i think through this endpoint you won't be able to have more than 1 project

JoshFerge: got it -- should we remove this code then? or what's the purpose. additionally, you can add a test for this in the corresponding test file (or create it if it doesn't exist) for the serializer instead of the endpoint test for unit testing.

roggenkemper: i think it's good to keep because this function gets called from a bunch of different endpoints. ill add a test to a different endpoint that also calls this method, one where multiple projects can be passed in.

JoshFerge: got it. and feel free to simply add a unit test here instead  / in addition. https://github.com/getsentry/sentry/blob/44093132f41d3b12fbb257942376f6aa0d599bce/tests/sentry/api/helpers/test_group_index.py#L1

src/sentry/api/helpers/group_index/update.py
roggenkemper: this query is used in a couple places to find the most recent release for a project, which we still need for this because in `clear_expired_resolutions` we compare the newly created release and check if there are any `GroupResolution` objects that have a release for a matching project that is older. if there is, then we know a new release for that project is created, so we can clear the pending GroupResolution and say the Group is resolved.

src/sentry/api/helpers/group_index/update.py
```diff
@@ -300,6 +299,34 @@ def update_groups(
...
+                return Response(
+                    {"detail": "Cannot set resolved in next release for multiple projects."},
```
JoshFerge, line 305: shoudl this response detail text be changed to upcoming release?

src/sentry/api/helpers/group_index/update.py
JoshFerge: got it

JoshFerge (approved):
one final q about error detail text but lgtm

src/sentry/api/helpers/group_index/update.py
roggenkemper: good catch , will update

PR #70991, opened 2024-05-15
feat(assignee-dropdown): Swap out old assignee dropdown trigger for <AssigneeBadge/>

MichaelSun48:
This PR swaps the current assignee dropdown trigger out for a new component, `<AssigneeBadge/>` under the feature flag,`organizations:issue-stream-new-assignee-dropdown-trigger`. This feature flag is set to release only in LA currently (although if you're reading this, I have not merged my PR to options automator just yet. I will do that before merging this.)

Before:
image

After:
image

Notice how the height of the issue priority dropdown trigger also increases a bit to match the height of the assignee badge. If the organization does not have the feature enabled, this height change will not be applied. 

This PR will also remove the appearance of the suggested assignees in the issue stream, aka this state:
image

If an issue does have suggested assignees, this new trigger will now just show up as Unassigned, but the suggested assignees and reasons will still be surfaced in the assignee dropdown menu.

### DIFF
--- a/static/app/components/badge/groupPriority.tsx
+++ b/static/app/components/badge/groupPriority.tsx
@@ -1,5 +1,6 @@
 import {Fragment, useMemo, useRef} from 'react';
 import type {Theme} from '@emotion/react';
+import {css} from '@emotion/react';
 import styled from '@emotion/styled';
 
 import bannerStar from 'sentry-images/spot/banner-star.svg';
@@ -102,9 +103,13 @@ export function GroupPriorityBadge({priority, children}: GroupPriorityBadgeProps
   return (
     <StyledTag
       type={getTagTypeForPriority(priority)}
-      bigtag={organization.features.includes(
-        'organizations:issue-stream-new-assignee-dropdown-trigger'
-      )}
+      css={css`
+        --tagHeight: ${organization.features.includes(
+          'organizations:issue-stream-new-assignee-dropdown-trigger'
+        )
+          ? '24px'
+          : '18px'};
+      `}
     >
       {PRIORITY_KEY_TO_LABEL[priority] ?? t('Unknown')}
       {children}
@@ -272,7 +277,7 @@ const DropdownButton = styled(Button)`
   box-shadow: none;
 `;
 
-const StyledTag = styled(Tag)<{bigtag: boolean}>`
+const StyledTag = styled(Tag)`
   span {
     display: flex;
     align-items: center;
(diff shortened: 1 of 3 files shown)


static/app/components/badge/groupPriority.tsx
```diff
@@ -266,12 +272,16 @@ const DropdownButton = styled(Button)`
   box-shadow: none;
 `;
 
-const StyledTag = styled(Tag)`
+const StyledTag = styled(Tag)<{bigTag: boolean}>`
```
MichaelSun48, line 275: I swear I tried for so long to not use dynamic props here but I could not figure out what the syntax was since its also combined with a nested selector 🫠 

If you know how to not use dynamic props here pls lmk

static/app/components/badge/groupPriority.tsx
```diff
@@ -266,12 +272,16 @@ const DropdownButton = styled(Button)`
   box-shadow: none;
 `;
 
-const StyledTag = styled(Tag)`
+const StyledTag = styled(Tag)<{bigtag: boolean}>`
```
MichaelSun48, line 275: I swear I tried for so long to not use dynamic props here but I could not figure out what the syntax was since its also combined with a nested selector 🫠 

If you know how to not use dynamic props here pls lmk

MichaelSun48, line 275: I swear I tried for so long to not use dynamic props here but I could not figure out what the syntax was since its also combined with a nested selector 🫠 

If you know how to not use dynamic props here pls lmk

scttcper (approved):

malwilley (approved):

static/app/components/badge/groupPriority.tsx
```diff
@@ -98,8 +99,14 @@ export function makeGroupPriorityDropdownOptions({
 }
 
 export function GroupPriorityBadge({priority, children}: GroupPriorityBadgeProps) {
+  const organization = useOrganization();
   return (
-    <StyledTag type={getTagTypeForPriority(priority)}>
+    <StyledTag
+      type={getTagTypeForPriority(priority)}
+      isBigtag={organization.features.includes(
```
malwilley, line 106: Was this added to make this match the size of the assignee dropdown? We could probably just update this to be the larger size regardless of the assignee dropdown being release or not

MichaelSun48: That was the intent, yes. Definitely a very small detail, but I just wanted to limit the scope of changes in this PR to be completely under the feature flag - plus I feel like it would be a little weird if users saw a fatter priority button with no other UI changes

PR #70992, opened 2024-05-15
ref(similarity-embedding): Make backfill script smarter

jangjodi:
Sort groups to backfill by times_seen
Remove groups with times_seen = 1, and add metadata for these groups
Add metadata for groups where a nearest neighbor is found instead of inserting a new record


src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -81,26 +91,44 @@ def backfill_seer_grouping_records(
...
+                group.data["metadata"] = {"seer_similarity": seer_similarity_seen_once}
+
+            if not dry_run:
+                Group.objects.bulk_update(groups_seen_once, ["data"])
+
+    group_id_message_data_batch = (
+        Group.objects.filter(project_id=project.id, type=ErrorGroupType.type_id, times_seen__gt=1)
         .values_list("id", "message", "data")
-        .order_by("id")[:BATCH_SIZE]
+        .order_by("times_seen")
```
vartec, line 121: it any of these are still active groups, this order can potentially change between runs.

vartec (approved):

src/sentry/tasks/backfill_seer_grouping_records.py
JoshFerge: yeah, we are accepting this and will monitor if this causes problems during LA/EA

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -171,40 +199,57 @@ def backfill_seer_grouping_records(
...
+                        seer_similarity["results"] = [
+                            asdict(
+                                SeerSimilarIssueData.from_raw(
+                                    project_id, groups_with_neighbor[str(group.id)]
+                                )
+                            )
+                        ]
+                    except (IncompleteSeerDataError, SimilarGroupNotFoundError) as e:
+                        logger.exception(
+                            str(e),
```
JoshFerge, line 221: you don't need to include the stringified error with logger.exception, it's included automatically

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -171,40 +199,57 @@ def backfill_seer_grouping_records(
...
+                                "hash": group_hashes_dict[group.id],
+                            },
+                        )
+
                 if group.data.get("metadata"):
-                    group.data["metadata"]["embeddings_info"] = {
-                        "nn_model_version": 0,
-                        "group_hash": json.dumps([group_hashes_dict[group.id]]),
-                    }
+                    group.data["metadata"]["seer_similarity"] = seer_similarity
```
JoshFerge, line 230: if there is an exception in the above, won't this fail as seer_similarity won't be defined?

JoshFerge (changes requested):
only thing i see is the bulk update should be out of the for loop if we want to do the single event groups at the start of the backfill. other than that looks great! thanks for the quick turn around on this.

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -81,26 +91,44 @@ def backfill_seer_grouping_records(
...
             logger.info(
                 "backfill_seer_grouping_records.delete_all_seer_records",
                 extra={"project_id": project.id},
             )
             delete_grouping_records(project_id)
 
-    group_id_message_data_batch = (
-        Group.objects.filter(
-            project_id=project.id, id__gt=last_processed_id, type=ErrorGroupType.type_id
+        # Set the metadata of groups where times_seen = 1
```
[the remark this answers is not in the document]
jangjodi: this was just for debugging purposes if we wanted to know why something was skipped (ie. by the time we check it, we have more than 1 event and we want to know if we purposely skipped it or our backfill didn't work). I'm not set on keeping it, so up to you.

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -81,26 +91,44 @@ def backfill_seer_grouping_records(
...
         )
+        for group in groups_seen_once:
+            seer_similarity_seen_once = {"times_seen_once": True}
+            if group.data.get("metadata"):
+                group.data["metadata"]["seer_similarity"] = seer_similarity_seen_once
+            else:
+                group.data["metadata"] = {"seer_similarity": seer_similarity_seen_once}
+
+            if not dry_run:
+                Group.objects.bulk_update(groups_seen_once, ["data"])
```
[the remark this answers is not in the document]
jangjodi: this was purely for updating the metadata in bulk for groups with times_seen = 1. if we choose to remove this, I can take it out

src/sentry/tasks/backfill_seer_grouping_records.py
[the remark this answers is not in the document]
JoshFerge: can keep it for now and re-assess as we run in S4S

JoshFerge (approved):

src/sentry/tasks/backfill_seer_grouping_records.py
jangjodi: it won't fail since it's been defined [here](https://github.com/getsentry/sentry/pull/70992/files#diff-073ef410c8c0c0aa12ee71c34018b70edd325d37bf684273293b3db96397b293R216-R218), but it won't be updated to show that a record wasn't created for it. this logic is a bit broken until we start deleting records. I think the correct logic here would be to not update the metadata at all so we can try to backfill again later once we support record deletion

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -174,6 +202,10 @@ def backfill_seer_grouping_records(
             project, rows, group_id_message_batch_filtered, group_hashes_dict
         )
 
+        # If nodestore is down, we should stop
+        if data["data"] == [] and data["stacktrace_list"] == []:
+            return
```
jangjodi, line 207: We already try multiple fallbacks and sleeps in between exceptions for getting data from nodestore - if we reach this point, we should stop trying (the exceptions are logged in the functions where they happen)

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -182,46 +214,76 @@ def backfill_seer_grouping_records(
...
-            f"{make_backfill_redis_key(project_id)}",
-            last_processed_id if last_processed_id is not None else 0,
-            ex=60 * 60 * 24 * 7,
-        )
+            last_processed_index = batch_end_index
+            redis_client.set(
+                f"{make_backfill_redis_key(project_id)}",
+                last_processed_index if last_processed_index is not None else 0,
+                ex=60 * 60 * 24 * 7,
+            )
```
jangjodi, line 267: we should not set the redis key if seer failed

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -182,46 +214,76 @@ def backfill_seer_grouping_records(
...
+                )
+                backfill_seer_grouping_records.apply_async(
+                    args=[project.id, last_processed_index, dry_run],
+                )
+        else:
+            # If seer is down, we should stop
+            logger.info(
+                "backfill_seer_bulk_insert_returned_invald_result",
+                extra={"project_id": project.id},
+            )
```
jangjodi, line 286: if seer returns an invalid response, we should stop trying

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -182,46 +214,76 @@ def backfill_seer_grouping_records(
...
                     extra={"project_id": project.id, "num_updated": num_updated},
                 )
 
-        last_processed_id = group_id_message_data_batch[len(group_id_message_data_batch) - 1][0]
-        redis_client.set(
-            f"{make_backfill_redis_key(project_id)}",
-            last_processed_id if last_processed_id is not None else 0,
-            ex=60 * 60 * 24 * 7,
-        )
+            last_processed_index = batch_end_index
```
jangjodi, line 262: the previous redis key calculation was wrong

JoshFerge (approved):
👍🏼

PR #70997, opened 2024-05-16
fix(assignees): Fix issue with suggested assignees not hoisting to top

MichaelSun48:
This PR fixes an uncommon bug where suggested assignees were not being hoisted to the top of the dropdown menu. This was caused by a combination of two things:

1. A recent change where we limit the number of assignees shown in the menu to 150
2. Suggested assignees were being shown both in the top "suggested" section, and later in the "members" section 

If a suggested assignee was below the 150th option and was removed, it would also be removed from the top of the list as well. This is because both options had the same id (because they were for the same user), and CompactSelect hides the options whose ids appear below the 150th option. 

The solution is to filter out the duplicate from the members list and only show it in the top "suggested" section, thus preventing it from being removed if it's duplicate would have been below the 150th element

### DIFF
--- a/static/app/components/assigneeSelectorDropdown.tsx
+++ b/static/app/components/assigneeSelectorDropdown.tsx
@@ -92,6 +92,10 @@ export interface AssigneeSelectorDropdownProps {
    */
   owners?: Omit<SuggestedAssignee, 'assignee'>[];
   /**
+   * Maximum number of teams/users to display in the dropdown
+   */
+  sizeLimit?: number;
+  /**
    * Optional trigger for the assignee selector. If nothing passed in,
    * the default trigger will be used
    */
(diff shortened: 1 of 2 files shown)


static/app/components/assigneeSelectorDropdown.tsx
```diff
@@ -438,6 +438,18 @@ export default function AssigneeSelectorDropdown({
       }
     }
 
+    // Remove suggested assignees from the member list and team list to avoid duplicates
+    for (let i = 0; i < suggestedAssignees.length; i++) {
```
malwilley, line 442: nit: might be a bit simpler to pull up the `suggestedUsers` and `suggestedTeams` and do it this way:

```ts
const suggestedUsers = suggestedAssignees.filter(
  assignee => assignee.type === 'user'
);
const suggestedTeams = suggestedAssignees.filter(
  assignee => assignee.type === 'team'
);

memList = memList.filter(user => !suggestedUsers.find(suggested => suggested.id === user.id))
assignableTeamList = assignableTeamList.filter(team => !suggestedTeams.find(suggested => suggested.id === team.id))
```

malwilley (commented):
Could you add a test that confirms the new logic works? Helpful to prevent this from regressing in the future

malwilley (approved):

PR #70999, opened 2024-05-16
feat(seer grouping): Add Seer-related ingest helpers

lobsterkatie:
This adds two helpers, `should_call_seer_for_grouping` and `get_seer_similar_issues`, to be used when we (maybe) call Seer as part of event ingestion. 

`should_call_seer_for_grouping` does exactly what you'd think given the name, right now only basing the decision on feature flags and whether or not the event has a usable title and/or stacktrace. In the future we'll also include rate limit and killswitch checks, and any other criteria which it makes sense to add.

`get_seer_similar_issues` is a wrapper around `get_similarity_data_from_seer` (which is what actually makes the API call to Seer). It extracts request data from the given event, makes the request, pulls together metadata about the results, and if a matching group is found and the flag is on, pulls the `Group` record out of the database. (I chose to put the feature flag check there rather than in the code where the the the grouping actually happens so that we can save the trip to the database if we're not going to end up using the results for grouping.)

Code to actually use these helpers is added in https://github.com/getsentry/sentry/pull/71026.

### DIFF
--- a/src/sentry/grouping/ingest/seer.py
+++ b/src/sentry/grouping/ingest/seer.py
@@ -1,0 +1,27 @@
+from sentry import features
+from sentry.constants import PLACEHOLDER_EVENT_TITLES
+from sentry.eventstore.models import Event
+from sentry.models.project import Project
+from sentry.utils.safe import get_path
+
+
+def should_call_seer_for_grouping(event: Event, project: Project) -> bool:
+    """
+    Use event content, feature flags, rate limits, killswitches, seer health, etc. to determine
+    whether a call to Seer should be made.
+    """
+    # TODO: Implement rate limits, kill switches, other flags, etc
+    # TODO: Return False if the event has a custom fingerprint (check for both client- and server-side fingerprints)
+
+    # If an event has no stacktrace, and only one of our placeholder titles ("<untitled>",
+    # "<unknown>", etc.), there's no data for Seer to analyze, so no point in making the API call.
+    if (
+        event.title in PLACEHOLDER_EVENT_TITLES
+        and not get_path(event.data, "exception", "values", -1, "stacktrace", "frames")
+        and not get_path(event.data, "threads", "values", -1, "stacktrace", "frames")
+    ):
+        return False
+
+    return features.has("projects:similarity-embeddings-metadata", project) or features.has(
+        "projects:similarity-embeddings-grouping", project
+    )
--- a/src/sentry/grouping/ingest/seer.py
+++ b/src/sentry/grouping/ingest/seer.py
@@ -1,8 +1,22 @@
+import logging
+from dataclasses import asdict
+
 from sentry import features
+from sentry.api.endpoints.group_similar_issues_embeddings import get_stacktrace_string
 from sentry.constants import PLACEHOLDER_EVENT_TITLES
 from sentry.eventstore.models import Event
+from sentry.grouping.grouping_info import get_grouping_info_from_variants
+from sentry.grouping.result import CalculatedHashes
+from sentry.models.group import Group
 from sentry.models.project import Project
+from sentry.seer.utils import (
+    SeerSimilarIssuesMetadata,
+    SimilarIssuesEmbeddingsRequest,
+    get_similarity_data_from_seer,
+)
 from sentry.utils.safe import get_path
+
+logger = logging.getLogger("sentry.events.grouping")
 
 
 def should_call_seer_for_grouping(event: Event, project: Project) -> bool:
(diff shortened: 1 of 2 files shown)


src/sentry/grouping/ingest/seer.py
```diff
@@ -0,0 +1,95 @@
...
+    if (
+        event.title in PLACEHOLDER_EVENT_TITLES
+        and not get_path(event.data, "exception", "values", -1, "stacktrace", "frames")
+        and not get_path(event.data, "threads", "values", -1, "stacktrace", "frames")
+    ):
+        return False
+
+    return features.has("projects:similarity-embeddings-metadata", project) or features.has(
+        "projects:similarity-embeddings-grouping", project
+    )
```
JoshFerge, line 39: why do we have two flags for this?

src/sentry/grouping/ingest/seer.py
```diff
@@ -0,0 +1,95 @@
...
+    or None if no neighbor was near enough.
+
+    Will also return `None` for the neighboring group if the `projects:similarity-embeddings-grouping`
+    feature flag is off.
+    """
+
+    # TODO: Once we get rid of hierarchical hashing, we'll be able to make `variants` required in
+    # `CalculatedHashes`, meaning we can remove this check. (See note in `CalculatedHashes` class definition.)
+    if primary_hashes.variants is None:
+        raise Exception("Primary hashes missing variants data")
```
JoshFerge, line 64: can we raise a more specific exception? can also just define your own exception that inherits from `Exception`

src/sentry/grouping/ingest/seer.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+
+
+def get_seer_similar_issues(
+    event: Event,
+    primary_hashes: CalculatedHashes,
+    num_neighbors: int = 1,
+) -> tuple[
+    dict[
+        str, str | list[dict[str, float | bool | int | str]]
```
JoshFerge, line 48: worth making a `TypedDict` for this type?

src/sentry/grouping/ingest/seer.py
lobsterkatie: Because we want to be able to turn on the metadata gathering without affecting grouping. There was a convo in one of the grouping sync meetings... IDK, two weeks ago, maybe? where we agreed that it was better to separate the flags.

JoshFerge (approved):

src/sentry/grouping/ingest/seer.py
lobsterkatie: The thing is, this never happens. There are other scenarios where `variants` can be `None`, but this isn't one of them. We need this bit mostly to make mypy happy, since those other scenarios force an `| None` into the type. I can make this clearer in the comment.

As the comment already says, though, all of that is only true (and therefore this check only exists) until we kill the hierarchical grouping code, which is why I didn't go out of my way to make it a super useful error.

src/sentry/grouping/ingest/seer.py
lobsterkatie: _sigh_ This is why I say Python is the IHOP of typing. I agree this isn't great, but there's no way that I've been able to figure out (and googling suggests it's because no way exists) to create a typeddict out of a dataclass. 

So I can, but it'll just mean hard coding a second copy of all of the attributes from both the `SeerSimilarIssuesMetadata` and `SeerSimilarIssueData` dataclasses, which feels worse to me (I think because a change to either one of those classes would be guaranteed to make the hypothetical typeddict need updating whereas there's a good chance they could be changed without this type needing to be any different). 

WDYT?

JoshFerge: yeah totally. its good as is!

vartec: Googling is so 2010s :-P Just ask Copilot ;-) 

```python
def dataclass_to_typeddict(cls):
    return TypedDict(cls.__name__ + 'Dict', {k: v for k, v in get_type_hints(cls).items()})

SeerSimilarIssuesMetadataDict = dataclass_to_typeddict(SeerSimilarIssuesMetadata)
SeerSimilarIssueDataDict = dataclass_to_typeddict(SeerSimilarIssueData)
```

src/sentry/grouping/ingest/seer.py
```diff
@@ -0,0 +1,98 @@
...
+    whether a call to Seer should be made.
+    """
+    # TODO: Implement rate limits, kill switches, other flags, etc
+    # TODO: Return False if the event has a custom fingerprint (check for both client- and server-side fingerprints)
+
+    if (
+        event.title in PLACEHOLDER_EVENT_TITLES
+        and not get_path(event.data, "exception", "values", -1, "stacktrace", "frames")
+        and not get_path(event.data, "threads", "values", -1, "stacktrace", "frames")
+    ):
```
vartec, line 34: Can you add a comment explaining this condition, please?

src/sentry/grouping/ingest/seer.py
lobsterkatie: If only, @vartec. TIL about `get_type_hints`, but my very first attempt to wrestle with the dataclass/typeddict divide was to do something like this:

```python
fields = {"a": int, "b": str}
DerivedTypedDict = TypedDict("DerivedTypedDict", fields)
```

and then construct the dataclass from the same `fields` value. Alas, mypy will have none of it: `error: TypedDict() expects a dictionary literal as the second argument`.

Here's an open issue about it in the mypy repo: https://github.com/python/mypy/issues/4128.

src/sentry/grouping/ingest/seer.py
lobsterkatie: Done.

src/sentry/grouping/ingest/seer.py
vartec: mypy is the worst...

src/sentry/grouping/ingest/seer.py
vartec:
```suggestion
    if not (
        get_path(event.data, "exception", "values", -1, "stacktrace", "frames")
        or get_path(event.data, "threads", "values", -1, "stacktrace", "frames")
        or event.title not in PLACEHOLDER_EVENT_TITLES
    ):
```

PR #71014, opened 2024-05-16
ref(trace_timeline): Make EventItem a reusable component

armenzg:
This is a no-op change which allows EventItem to be used in future work for related issues.

### DIFF
--- a/static/app/views/issueDetails/traceTimeline/traceTimelineTooltip.tsx
+++ b/static/app/views/issueDetails/traceTimeline/traceTimelineTooltip.tsx
@@ -173,7 +173,7 @@ const EventTitle = styled('div')`
 
 const EventDescription = styled('div')`
   ${p => p.theme.overflowEllipsis};
-  margin-left: 1em;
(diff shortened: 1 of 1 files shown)


static/app/views/issueDetails/traceTimeline/traceTimelineTooltip.tsx
```diff
@@ -20,13 +20,6 @@ interface TraceTimelineTooltipProps {
 
 export function TraceTimelineTooltip({event, timelineEvents}: TraceTimelineTooltipProps) {
   const organization = useOrganization();
-  const location = useLocation();
-  const {projects} = useProjects({
-    slugs: [
-      ...timelineEvents.reduce((acc, cur) => acc.add(cur.project), new Set<string>()),
-    ],
-    orgId: organization.slug,
-  });
```
armenzg, line 29: This and the following removed lines are moved into the new `EventItem` component.

scttcper (approved):

PR #71020, opened 2024-05-16
ref(insights): remove outdated starfish.panel.open in favor of new analytics

KevinL10:
Removes `starfish.panel.open` which we no longer use in favor of `performance_views.sample_spans.opened`.


narsaynorath (approved):
:fire: thanks for cleaning this up

PR #71023, opened 2024-05-16
feat(crons): Add processing error types to frontend

davidenwang:
Adds types which match the ones in 'src/sentry/monitors/processing_errors/errors.py'

### DIFF
--- a/static/app/views/monitors/types.tsx
+++ b/static/app/views/monitors/types.tsx
@@ -196,3 +196,106 @@ export interface StatusNotice {
   icon: React.ReactNode;
   label?: React.ReactNode;
 }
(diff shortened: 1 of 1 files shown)


static/app/views/monitors/types.tsx
```diff
@@ -196,3 +196,84 @@ export interface StatusNotice {
   icon: React.ReactNode;
   label?: React.ReactNode;
 }
+
+export enum ProcessingErrorType {
```
evanpurkhiser, line 200: Let's add a comment about where to find this in the backend

evanpurkhiser (approved):

PR #71024, opened 2024-05-16
fix(integrations): set self.field_errors in IntegrationFormError

cathteng:
Fixes SENTRY-398A


iamrajjoshi (approved):

schew2381 (approved):
so this was used somewhere lol

Issue #70454, opened 2024-05-07
[grouping] Consider storing metadata on `GroupHash` records


lobsterkatie:
It's come up a couple times that it'd be great to have more information about how a particular `GroupHash` record came to be and/or came to be in its current state. 

  - Grouping config version (would have to decide if it's first version to calculate that hash value or latest version)
  - If Seer is involved, the fact that the hash is represented in the Seer DB, Seer model version, parent hash (if any) - useful for display in the Grouping Info section of the issue details page, and for debugging the grouping of a given event.
  - If the issue's been merged, maybe the id the `Activity` record of the merge so we can get more data from that if desired
  - Possibly the grouping enhancements currently stored on each event (so that they don't have to be), maybe along with a dump of the stacktrace which was enhanced, and/or the full grouping info `variants` data (so we can know which hashes come from message vs stacktrace vs whatever)
  - Other things?

lobsterkatie:
This is running in prod now. Closing this.

PR #71026, opened 2024-05-16
feat(seer grouping): Call Seer before creating a new group

lobsterkatie:
This uses the helpers added in https://github.com/getsentry/sentry/pull/70999 to - depending on the state of the `projects:similarity-embeddings-metadata` and `projects:similarity-embeddings-grouping` flags - decide whether we should call Seer before creating a new group, make the API call if so, and then store the results and/or use them to actually prevent new group creation in favor of using an existing similar issue. The behavior is as follows:

```
| metadata  | grouping | call  | metadata in | metadata in | use Seer-matched |
|   flag    |  flag    | Seer? |   event?    |   group?    |  group, if any?  |
|-----------|----------|-------|-------------|-------------|------------------|
| off       | off      | no    | -           | -           | -                |
| on        | off      | yes   | yes *       | yes         | no               |
| on or off | on       | yes   | yes *       | only if new | yes              |

* For now, the only event with the data will be the event which triggers the Seer 
call, not subsequent events with that hash. In the long run we will probably need
to store the data on the `GroupHash` record itself. 
See https://github.com/getsentry/sentry/issues/70454.
```

This should be enough for us to run a POC on S4S and measure the effect on grouping.

### DIFF
--- a/src/sentry/event_manager.py
+++ b/src/sentry/event_manager.py
@@ -68,6 +68,7 @@ from sentry.grouping.ingest.metrics import (
     record_hash_calculation_metrics,
     record_new_group_metrics,
 )
+from sentry.grouping.ingest.seer import get_seer_similar_issues, should_call_seer_for_grouping
 from sentry.grouping.ingest.utils import (
     add_group_id_to_grouphashes,
     check_for_category_mismatch,
@@ -136,6 +137,7 @@ from sentry.utils.performance_issues.performance_problem import PerformanceProbl
 from sentry.utils.safe import get_path, safe_execute, setdefault_path, trim
 from sentry.utils.sdk import set_measurement
 from sentry.utils.tag_normalization import normalized_sdk_tag_from_event
+from sentry.utils.types import NonNone
 
 if TYPE_CHECKING:
     from sentry.eventstore.models import BaseEvent, Event
@@ -1477,7 +1479,31 @@ def _save_aggregate(
             # If we still haven't found a matching grouphash, we're now safe to go ahead and create
             # the group.
             if existing_grouphash is None:
-                group = _create_group(project, event, **group_creation_kwargs)
+                seer_matched_group = None
+
+                if should_call_seer_for_grouping(event, project):
+                    try:
+                        # If the `projects:similarity-embeddings-grouping` feature is disabled,
+                        # we'll still get back result metadata, but `seer_matched_group` will be None
+                        seer_response_data, seer_matched_group = get_seer_similar_issues(
+                            event, primary_hashes
+                        )
+                        event.data["seer_similarity"] = seer_response_data
+
+                        # We only want to add this data to new groups, while we're testing
+                        # TODO: Remove this once we're out of the testing phase
+                        if not seer_matched_group:
+                            group_creation_kwargs["data"]["metadata"][
+                                "seer_similarity"
+                            ] = seer_response_data
+
+                    # Insurance - in theory we shouldn't ever land here
+                    except Exception as e:
+                        sentry_sdk.capture_exception(
+                            e, tags={"event": event.event_id, "project": project.id}
+                        )
+
+                group = seer_matched_group or _create_group(project, event, **group_creation_kwargs)
 
                 if root_hierarchical_grouphash is not None:
                     new_hashes = [root_hierarchical_grouphash]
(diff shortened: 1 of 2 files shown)


src/sentry/event_manager.py
```diff
@@ -1488,41 +1526,64 @@ def _save_aggregate(
                     state=GroupHash.State.LOCKED_IN_MIGRATION
                 ).update(group=group)
 
-                is_new = True
-                is_regression = False
+                is_new = False if seer_matched_group else True
```
vartec, line 1529:
```suggestion
                is_new = bool(seer_matched_group)
```

src/sentry/event_manager.py
```diff
@@ -1474,10 +1489,33 @@ def _save_aggregate(
...
+                    try:
+                        # If the `projects:similarity-embeddings-grouping` feature is disabled,
+                        # we'll still get back result metadata, but `seer_matched_group` will be None
+                        seer_response_data, seer_matched_group = get_seer_similar_issues(
+                            event, primary_hashes
+                        )
+                        event.data["seer_similarity"] = seer_response_data
+
+                        # We only want to add this data to new groups, while we're testing
+                        # TODO: Remove this once we're out of the testing phase
```
vartec, line 1507: Wouldn't we want to keep that for debugging even after testing phase?

src/sentry/event_manager.py
lobsterkatie: It would have to be `is_new = not bool(seer_matched_group)`. Is that clearer, do you think?

src/sentry/event_manager.py
```diff
@@ -1488,41 +1526,64 @@ def _save_aggregate(
...
+                    if is_new
+                    else _process_existing_aggregate(
+                        # If `seer_matched_group` were `None`, `is_new` would be true and we
+                        # wouldn't be here
+                        group=NonNone(seer_matched_group),
+                        event=event,
+                        incoming_group_values=group_creation_kwargs,
+                        release=release,
+                    )
+                )
```
vartec, line 1541: purely opinion based, so entirely up to you, but I find just regular ifs more readable than using conditional expressions multiple times. 


```suggestion
                if is_new := not seer_matched_group:
                    is_regression = False
                    outcome = "new_group"
                else:
                     is_regression =  _process_existing_aggregate(
                        # If `seer_matched_group` were `None`, `is_new` would be true and we
                        # wouldn't be here
                        group=NonNone(seer_matched_group),
                        event=event,
                        incoming_group_values=group_creation_kwargs,
                        release=release,
                    )
                    outcome = "seer_match"
```

and below... 
```python
                span.set_tag("outcome", outcome)
                metric_tags["outcome"] = outcome
```

vartec (approved):

src/sentry/event_manager.py
lobsterkatie: It's true, I do love a good ternary. 🙂  I do sometimes get the sense that they're a little more idiomatic in JS/TS (which is - for better or worse - very much my first love, programming-language-wise) than in Python. Also for whatever reason it actually does make more sense to my brain. But you're right that there's a time and a place, and I do probably lean into it a little hard sometimes.

I'm going to leave it as is for now. It works, and there's lots more to do, but I do take the overall point.

(FWIW, the structure of `_save_aggregate_new` is different enough that when I port this there, I think I am going to have to structure it more with `if`s than with ternaries.)

src/sentry/event_manager.py
lobsterkatie: We might end up keeping it, but the current plan is to take it out eventually. For debugging purposes, we have the same information ~on the event~ hopefully added to the corresponding `GroupHash` record. (See https://github.com/getsentry/sentry/issues/70454.)

src/sentry/event_manager.py
```diff
@@ -1474,10 +1489,33 @@ def _save_aggregate(
...
+                        event.data["seer_similarity"] = seer_response_data
+
+                        # We only want to add this data to new groups, while we're testing
+                        # TODO: Remove this once we're out of the testing phase
+                        if not seer_matched_group:
+                            group_creation_kwargs["data"]["metadata"][
+                                "seer_similarity"
+                            ] = seer_response_data
+
+                    except Exception as e:
```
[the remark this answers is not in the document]
lobsterkatie: Much more the latter. Any reasonably-anticipatable errors in that logic are caught before they bubble up this high, but the last thing I want to do is break ingestion, so it seemed wise to try-catch it just in case.

PR #71029, opened 2024-05-16
ref(<Tag>): Remove emotion var and replace with prop

MichaelSun48:
This PR removes the usage of a css property to set the `borderstyle` for the `<Tag/>` component, and replaces it with a emotion prop.

### DIFF
--- a/static/app/components/badge/tag.tsx
+++ b/static/app/components/badge/tag.tsx
@@ -1,5 +1,5 @@
 import {useCallback} from 'react';
-import {css, type Theme} from '@emotion/react';
+import type {Theme} from '@emotion/react';
 import styled from '@emotion/styled';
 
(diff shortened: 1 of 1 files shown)


static/app/components/badge/tag.tsx
```diff
@@ -156,13 +150,16 @@ export default Tag;
 
 const TAG_HEIGHT = '20px';
 
-export const Background = styled('div')<{type: keyof Theme['tag']}>`
+export const Background = styled('div')<{
+  type: keyof Theme['tag'];
+  borderStyle?: TagProps['borderStyle'];
```
MichaelSun48, line 155: Made this optional since other components were complaining that this was required, not sure if this has any unintentional consequences though.

scttcper (approved):

PR #71035, opened 2024-05-16
ref(tsc): convert adminQueue.tsx to FC

michellewzhang:

### DIFF
--- a/static/app/views/admin/adminQueue.tsx
+++ b/static/app/views/admin/adminQueue.tsx
@@ -1,43 +1,54 @@
+import {useState} from 'react';
 import styled from '@emotion/styled';
 
 import {Button} from 'sentry/components/button';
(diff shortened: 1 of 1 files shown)


static/app/views/admin/adminQueue.tsx
```diff
@@ -1,43 +1,54 @@
...
 
 type TimeWindow = (typeof TIME_WINDOWS)[number];
 
-type State = DeprecatedAsyncView['state'] & {
+type State = {
   activeTask: string;
   resolution: string;
   since: number;
-  taskList: string[];
-  taskName: string;
```
michellewzhang, line 21: `taskName` wasn't used; `taskList` was fetched from the endpoint

scttcper (approved):

PR #71036, opened 2024-05-16
fix: ONLY fetch apps for the ORGANIZATION

nhsiehgit:
App installations were fetching ALL installations for an app.

This was causing requests to timeout via HC

### DIFF
--- a/src/sentry/incidents/endpoints/organization_alert_rule_details.py
+++ b/src/sentry/incidents/endpoints/organization_alert_rule_details.py
@@ -58,10 +58,10 @@ def fetch_alert_rule(request: Request, organization, alert_rule):
                     action.get("_sentry_app_installation", {}).get("sentry_app_id", None)
                 )
     if sentry_app_ids:
-        sentry_app_map = {
-            install.sentry_app.id: install.sentry_app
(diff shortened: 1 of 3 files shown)


src/sentry/incidents/endpoints/organization_alert_rule_details.py
```diff
@@ -58,10 +58,10 @@ def fetch_alert_rule(request: Request, organization, alert_rule):
                     action.get("_sentry_app_installation", {}).get("sentry_app_id", None)
                 )
     if sentry_app_ids:
-        sentry_app_map = {
-            install.sentry_app.id: install.sentry_app
-            for install in app_service.get_many(filter=dict(app_ids=sentry_app_ids))
-        }
+        fetched_apps = app_service.get_many(
+            filter=dict(app_ids=sentry_app_ids, organization_id=organization.id)
```
GabeVillalobos, line 62: This seems reasonable. Do we need any additional tests to verify that this behavior is correct?

nhsiehgit: Yup working on that now

markstory (approved):
Makes sense to me.

PR #71039, opened 2024-05-16
chore(pr-comments): generalization for triggering merged PR comments workflow

cathteng:
If we want to be able to have merged PR comments for SCMs other than GitHub, we'll need to generalize some of the flows. In this PR I generalize the flow to trigger the merged PR comments workflow.

1. Check if the repo for the suspect commit has a provider that supports merged PR comments
2. Move the custom logic needed to fetch the merge commit sha from the commit sha into the GitHub client, this can be repeated for other SCMs. We use this to find the PR associated with the merge commit sha in our db.

In a follow up PR, I will generalize the workflow itself.

### DIFF
--- a/src/sentry/integrations/github/client.py
+++ b/src/sentry/integrations/github/client.py
@@ -224,6 +224,25 @@ class GitHubClientMixin(GithubProxyClient):
         """
         return self.get_cached(f"/repos/{repo}/commits/{sha}")
 
+    def get_merge_commit_sha_from_commit(self, repo: str, sha: str) -> str | None:
+        """
+        Get the merge commit sha from a commit sha.
+        """
+        response = self.get_pullrequest_from_commit(repo, sha)
+        if not isinstance(response, list) or len(response) != 1:
+            # the response should return a single merged PR, return if multiple
+            if len(response) > 1:
+                return None
+
+        if response[0]["state"] == "open":
+            metrics.incr(
+                "github_pr_comment.queue_comment_check.open_pr",
+                sample_rate=1.0,
+            )
+            return None
+
+        return response[0].get("merge_commit_sha")
+
     def get_pullrequest_from_commit(self, repo: str, sha: str) -> Any:
         """
         https://docs.github.com/en/rest/commits/commits#list-pull-requests-associated-with-a-commit
--- a/src/sentry/integrations/github/client.py
+++ b/src/sentry/integrations/github/client.py
@@ -233,7 +233,7 @@ class GitHubClientMixin(GithubProxyClient):
             # the response should return a single merged PR, return if multiple
             return None
 
-        pull_request = response[0]
+        (pull_request,) = response
         if pull_request["state"] == "open":
             metrics.incr(
                 "github_pr_comment.queue_comment_check.open_pr",
--- a/src/sentry/integrations/github/client.py
+++ b/src/sentry/integrations/github/client.py
@@ -229,19 +229,19 @@ class GitHubClientMixin(GithubProxyClient):
         Get the merge commit sha from a commit sha.
         """
         response = self.get_pullrequest_from_commit(repo, sha)
-        if not isinstance(response, list) or len(response) != 1:
+        if not response or isinstance(response, list) or len(response) != 1:
             # the response should return a single merged PR, return if multiple
-            if len(response) > 1:
-                return None
+            return None
 
-        if response[0]["state"] == "open":
+        pull_request = response[0]
+        if pull_request["state"] == "open":
             metrics.incr(
                 "github_pr_comment.queue_comment_check.open_pr",
                 sample_rate=1.0,
             )
             return None
 
-        return response[0].get("merge_commit_sha")
+        return pull_request.get("merge_commit_sha")
 
     def get_pullrequest_from_commit(self, repo: str, sha: str) -> Any:
         """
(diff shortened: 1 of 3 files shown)


src/sentry/integrations/github/client.py
```diff
@@ -224,6 +224,25 @@ def get_commit(self, repo: str, sha: str) -> Any:
         """
         return self.get_cached(f"/repos/{repo}/commits/{sha}")
 
+    def get_merge_commit_sha_from_commit(self, repo: str, sha: str) -> str | None:
+        """
+        Get the merge commit sha from a commit sha.
+        """
+        response = self.get_pullrequest_from_commit(repo, sha)
+        if not isinstance(response, list) or len(response) != 1:
+            # the response should return a single merged PR, return if multiple
```
iamrajjoshi, line 233: qq - when does this return multiple prs for the same commit?

cathteng: if there are multiple open PRs that contain the commit, like if somebody cherry picked the commit into different branches

src/sentry/integrations/github/client.py
```diff
@@ -224,6 +224,25 @@ def get_commit(self, repo: str, sha: str) -> Any:
         """
         return self.get_cached(f"/repos/{repo}/commits/{sha}")
 
+    def get_merge_commit_sha_from_commit(self, repo: str, sha: str) -> str | None:
+        """
+        Get the merge commit sha from a commit sha.
+        """
+        response = self.get_pullrequest_from_commit(repo, sha)
```
ykamo001, line 231: Can we add a `None` check since the function can return `Any`?

src/sentry/integrations/github/client.py
```diff
@@ -224,6 +224,25 @@ def get_commit(self, repo: str, sha: str) -> Any:
...
+        if not isinstance(response, list) or len(response) != 1:
+            # the response should return a single merged PR, return if multiple
+            if len(response) > 1:
+                return None
```
ykamo001, line 235: why do we need this check if the parent if already does this check?

src/sentry/integrations/github/client.py
```diff
@@ -224,6 +224,25 @@ def get_commit(self, repo: str, sha: str) -> Any:
...
+
+        if response[0]["state"] == "open":
+            metrics.incr(
+                "github_pr_comment.queue_comment_check.open_pr",
+                sample_rate=1.0,
+            )
+            return None
+
+        return response[0].get("merge_commit_sha")
```
ykamo001, line 244: Can you please pull the first object out into a variable?

src/sentry/integrations/github/client.py
cathteng: i make sure the response is a list in the check below this line, do you mean another check specifically for `None`?

src/sentry/integrations/github/client.py
cathteng: what do you mean by parent? i isolated the logic for getting the github pull request from the commit in this function

cathteng: oh you mean the if above, you're right

src/sentry/integrations/github/client.py
ykamo001: Yeah exactly! I believe `isinstance` is fairly expensive, might help save some performance: https://stackoverflow.com/questions/42378726/why-is-checking-isinstancesomething-mapping-so-slow

src/sentry/tasks/commit_context.py
```diff
@@ -48,6 +48,8 @@
 PR_COMMENT_TASK_TTL = timedelta(minutes=5).total_seconds()
 PR_COMMENT_WINDOW = 14  # days
 
+PR_COMMENT_SUPPORTED_PROVIDERS = {"integrations:github"}
```
cathteng, line 51: when all SCMs have implemented the `get_merge_commit_sha_from_commit` function, we can remove this

src/sentry/integrations/github/client.py
```diff
@@ -224,6 +224,25 @@ def get_commit(self, repo: str, sha: str) -> Any:
...
         return self.get_cached(f"/repos/{repo}/commits/{sha}")
 
+    def get_merge_commit_sha_from_commit(self, repo: str, sha: str) -> str | None:
+        """
+        Get the merge commit sha from a commit sha.
+        """
+        response = self.get_pullrequest_from_commit(repo, sha)
+        if not response or (isinstance(response, list) and len(response) != 1):
+            # the response should return a single merged PR, return if multiple
+            return None
```
RyanSkonnord, line 234: What does it mean if we return on multiple? Should we be logging it?

src/sentry/integrations/github/client.py
```diff
@@ -224,6 +224,25 @@ def get_commit(self, repo: str, sha: str) -> Any:
...
+
+        pull_request = response[0]
```
RyanSkonnord, line 236: Nit: I like the idiom `(pull_request,) = response` because it communicates the expectation that the length is 1 and raises otherwise.

RyanSkonnord (approved):

src/sentry/integrations/github/client.py
cathteng: for GitHub, this means there are multiple open PRs with the commit https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-pull-requests-associated-with-a-commit
Screenshot 2024-07-19 at 09 53 28

PR #71040, opened 2024-05-16
ref(rules): Log after zadd

ceorourke:
Add a log to confirm we're writing project ids to Redis with the data we expect.


nhsiehgit (approved):

saponifi3d (approved):
:shipit:

PR #71045, opened 2024-05-16
feat(search): simplify the sunba heavy search to use convert_search_filter_to_condition

scefali:
This PR does 2 things:
* Leverage `convert_search_filter_to_condition` for the default search conditions for generic event fields. I had more complicated logic before but I realize I didn't need it.
* Adds custom logic for handling `error.unhandled` and `error.handled`. Previously, we used Snuba functions on these but they don't exist on the `search_issues` dataset. Instead, this maps to some logic based on `exception_stacks.mechanism_handled`. The code was ripped from Snuba here: https://github.com/getsentry/snuba/blob/83f3d5a50d2ca59cf13b9a56075a2ed4e3923b3d/snuba/query/processors/logical/handled_functions.py#L89-L116


src/sentry/search/snuba/executors.py
```diff
@@ -1190,62 +1192,16 @@ def get_basic_event_snuba_condition(
...
         """
-        # note this might hit postgres to do queries on releases
-        raw_conditions, projects_to_filter, group_ids = format_search_filter(
-            search_filter,
-            params={
-                "organization_id": organization_id,
-                "project_id": project_ids,
-                "environment": environments,
-            },
+        dataset = Dataset.Events if joined_entity.alias == "e" else Dataset.IssuePlatform
```
wedamija, line 1200: Wondering if this should filter on `name` rather than `alias`? Either way, it might be better to use constants here

src/sentry/search/snuba/executors.py
```diff
@@ -1513,6 +1451,54 @@ def get_last_seen_aggregation(self, joined_entity: Entity) -> Function:
             alias="score",
         )
 
+    def get_handled_condition(
+        self, search_filter: SearchFilter, joined_entity: Entity
+    ) -> Condition:
+        """
+        Returns the handled and unhandled lookup for a search filter.
+        Does not use the functions that Snbua has, instead replicates that logic
```
wedamija, line 1459: Nit: Snuba

wedamija (approved):

PR #71046, opened 2024-05-16
feat(performance): Reorganizes frontend Queues module referrer usage

edwardgou-sentry:
Refactors queues module referrers in the frontend to be a bit more organized


static/app/views/performance/queues/referrers.ts
```diff
@@ -0,0 +1,15 @@
+export enum Referrer {
+  QUEUES_DEFAULT_REFERRER = 'api.performance.queues',
```
gggritso, line 2: This is only used in a few places, right? If yes, would be even better to _never_ use a default referrer, and update the hook arguments to require a referrer at all times

gggritso (approved):
👍🏻

static/app/views/performance/queues/referrers.ts
edwardgou-sentry: 👍 updated!

PR #71049, opened 2024-05-16
feat(related_issues): Show trace-related issues in the Issue Details page

armenzg:
**NOTE**: This UI change will have a follow-up PR creating a new polished component.

If the trace for an issue has 2 different issues involved, show links to the related issues and do not show the trace timeline.

To be clear, this only impacts the case when we have 2 issues regardless of the number of events.

You can load [this event](https://dev.getsentry.net:7999/issues/4208426984/events/abef1576630343a496798c9250fe69db/?project=1) locally with `yarn dev-ui` if you want to compare it with [the production event](https://sentry.sentry.io/issues/4208426984/events/abef1576630343a496798c9250fe69db/).

The following screenshot shows the before and after.

Before: Timeline showing 2 events for 2 different issues (not clear):
image

After: We replaced the timeline with info about the issues involved in the trace:
image

<hr />
This also works for other types of issues besides errors:
image


static/app/views/issueDetails/traceTimeline/traceTimeline.tsx
```diff
@@ -49,28 +62,56 @@ export function TraceTimeline({event}: TraceTimelineProps) {
...
-            size="sm"
-            title={t(
-              'This is a trace timeline showing all related events happening upstream and downstream of this event'
-            )}
-            position="bottom"
-          />
-        </QuestionTooltipWrapper>
-      </TimelineWrapper>
+      {timelineStatus === 'shown' ? (
+        <TimelineWrapper>
```
armenzg, line 66: This `TimelineWrapper` block stays the same.

Using the Split view without whitespaces helps to see it more clearly:
image

static/app/views/issueDetails/traceTimeline/traceTimeline.tsx
```diff
@@ -49,28 +62,56 @@ export function TraceTimeline({event}: TraceTimelineProps) {
...
+              )}
+              position="bottom"
+            />
+          </QuestionTooltipWrapper>
+        </TimelineWrapper>
+      ) : (
+        <Feature features="related-issues-issue-details-page">
+          {issues.length === twoIssues && (
+            <div>
+              {traceEvents.map((traceEvent, index) => (
```
armenzg, line 88: This is incorrect since if there are more than 2 events for 2 issues we would be showing all of them.
I will be fixing this next week when I come back.

static/app/views/issueDetails/traceTimeline/traceTimelineTooltip.tsx
```diff
@@ -173,11 +173,13 @@ const EventTitle = styled('div')`
 
 const EventDescription = styled('div')`
   ${p => p.theme.overflowEllipsis};
-  direction: rtl;
+  margin-left: 1em;
```
armenzg, line 176: Without this change, it would look like this:
image

These two screenshots show the changes in the tooltip which I think they're acceptable. Please let me know.

Before:
image
After:
image

static/app/views/issueDetails/traceTimeline/traceTimeline.tsx
```diff
@@ -1,31 +1,44 @@
...
 import useRouteAnalyticsParams from 'sentry/utils/routeAnalytics/useRouteAnalyticsParams';
 import {useDimensions} from 'sentry/utils/useDimensions';
+import useOrganization from 'sentry/utils/useOrganization';
 
 import {TraceTimelineEvents} from './traceTimelineEvents';
-import {useTraceTimelineEvents} from './useTraceTimelineEvents';
+import {EventItem} from './traceTimelineTooltip';
+import {type TimelineEvent, useTraceTimelineEvents} from './useTraceTimelineEvents';
+
+const twoIssues = 2;
```
malwilley, line 18: nit: could use a more descriptive name, and uppercase naming because it's a constant.

static/app/views/issueDetails/traceTimeline/traceTimelineTooltip.tsx
malwilley: For the tooltip we want to show the end of the endpoint name since that is more likely to be unique, so that should stay the same. 

Instead of using the same component, we should probably make a new one here. With so much extra room, we will probably want to render these items differently than the tooltip does.

static/app/views/issueDetails/traceTimeline/traceTimeline.tsx
malwilley, line 1: We should add a test for the new behavior in `traceTimeline.spec.tsx`

malwilley (commented):
I do have one issue with the general feature - once the trace loads, this will push the rest of the content down ([CLS](https://web.dev/articles/cls)). Since this is at the top, we want to avoid changing content size as much as possible

static/app/views/issueDetails/traceTimeline/traceTimelineTooltip.tsx
armenzg: For the sake of expediency, I will not modify the look of this component and use it as-is.
I will finish this PR (so we can have a minimal working solution for internal testing) and then work on the nicer component on a follow-up PR.

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -152,4 +152,71 @@ describe('TraceTimeline', () => {
     render(<TraceTimeline event={event} />, {organization});
     expect(await screen.findByLabelText('Current Event')).toBeInTheDocument();
   });
+
+  it('skips the timeline and shows related issues (2 issues)', async () => {
```
armenzg, line 156: I need to add a test when there are more than 2 issues.

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -152,4 +152,71 @@ describe('TraceTimeline', () => {
...
+      match: [MockApiClient.matchQuery({dataset: 'issuePlatform'})],
+    });
+    MockApiClient.addMockResponse({
+      url: `/organizations/${organization.slug}/events/`,
+      body: discoverBody,
+      match: [MockApiClient.matchQuery({dataset: 'discover'})],
+    });
+    // I believe the call to projects is to determine what projects a user belongs to
+    MockApiClient.addMockResponse({
+      url: `/organizations/${organization.slug}/projects/`,
```
armenzg, line 169: This mock is not necessary for my feature but when I use the `EventItem` it seems that it gets called.

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -152,4 +152,71 @@ describe('TraceTimeline', () => {
...
+
+    // Enable related issues feature flag
+    organization.features = ['related-issues-issue-details-page'];
+    render(<TraceTimeline event={event} />, {organization});
+
+    // Instead of a timeline, we should see related issues
+    expect(await screen.findByText('Slow DB Query')).toBeInTheDocument();
+    expect(
+      await screen.findByText('AttributeError: Something Failed')
+    ).toBeInTheDocument();
```
armenzg, line 181: These lines test that both issues are displayed.

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -152,4 +152,71 @@ describe('TraceTimeline', () => {
...
+    expect(screen.queryByLabelText('Current Event')).not.toBeInTheDocument();
+    expect(useRouteAnalyticsParams).toHaveBeenCalledWith({
+      trace_timeline_status: 'empty',
+    });
+  });
+
+  it('skips the timeline and shows NO related issues (only 1 issue)', async () => {
+    MockApiClient.addMockResponse({
+      url: `/organizations/${organization.slug}/events/`,
+      body: emptyBody,
```
armenzg, line 191: We have `emptyBody` here in contrast with the previous test.

static/app/views/issueDetails/traceTimeline/traceTimeline.tsx
```diff
@@ -1,31 +1,46 @@
...
+  let timelineSkipped = false;
+  let issuesCount = 0;
   if (hasTraceId && !isLoading) {
-    timelineStatus = traceEvents.length > 1 ? 'shown' : 'empty';
+    if (!organization.features.includes('related-issues-issue-details-page')) {
+      timelineStatus = traceEvents.length > 1 ? 'shown' : 'empty';
+    } else {
+      issuesCount = getIssuesCountFromEvents(traceEvents);
+      // When we have more than 2 issues regardless of the number of events we skip the timeline
+      timelineSkipped = issuesCount === ISSUES_TO_SKIP_TIMELINE;
```
armenzg, line 41: For now, we will only replace the timeline when there are two issues. I have to make some changes here tomorrow.

static/app/views/issueDetails/traceTimeline/traceTimelineTooltip.tsx
```diff
@@ -181,3 +181,5 @@ const TraceItem = styled('div')`
   border-radius: ${p => p.theme.borderRadius};
   border-top: 1px solid ${p => p.theme.innerBorder};
 `;
+
+export {EventItem};
```
armenzg, line 185: I'm using the component as-is without any modifications.

[A local example](https://dev.getsentry.net:7999/issues/4208426984/events/abef1576630343a496798c9250fe69db/?project=1):
image

[Another local example](https://dev.getsentry.net:7999/issues/4912039318/events/7fef9536ef1947a9a17e9ef73c914b2b/?project=1):
image

static/app/views/issueDetails/traceTimeline/traceTimeline.tsx
```diff
@@ -49,28 +64,49 @@ export function TraceTimeline({event}: TraceTimelineProps) {
...
+                'This is a trace timeline showing all related events happening upstream and downstream of this event'
+              )}
+              position="bottom"
+            />
+          </QuestionTooltipWrapper>
+        </TimelineWrapper>
+      ) : (
+        <Feature features="related-issues-issue-details-page">
+          {timelineSkipped && (
+            <div style={{width: '400px'}}>
```
malwilley, line 89: This should be a styled component (although I know this is temporary so probably fine for now)

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -152,4 +152,71 @@ describe('TraceTimeline', () => {
...
+      match: [MockApiClient.matchQuery({dataset: 'discover'})],
+    });
+    // I believe the call to projects is to determine what projects a user belongs to
+    MockApiClient.addMockResponse({
+      url: `/organizations/${organization.slug}/projects/`,
+      body: [],
+    });
+
+    // Enable related issues feature flag
+    organization.features = ['related-issues-issue-details-page'];
```
malwilley, line 174: We want to avoid mutating in tests, since it will make the order of tests matter. Better to add the flag like this:

```ts
render(..., {organization: OrganizationFixture({features: [...]})})
```

static/app/views/issueDetails/traceTimeline/traceTimeline.tsx
armenzg: I will handle this in the next PR.

static/app/views/issueDetails/traceTimeline/traceTimeline.tsx
```diff
@@ -49,28 +64,58 @@ export function TraceTimeline({event}: TraceTimelineProps) {
...
+          {timelineSkipped && (
+            // XXX: Temporary. This will need to be replaced with a styled component
+            <div style={{width: '400px'}}>
+              {traceEvents.map((traceEvent, index) => (
+                <div key={index} style={{display: 'flex', alignItems: 'center'}}>
+                  <div
+                    style={{whiteSpace: 'nowrap', minWidth: '75px'}}
+                    data-test-id={`this-event-${traceEvent.id}`}
+                  >
+                    {event.id === traceEvent.id && <span>This event</span>}
```
armenzg, line 97: @malwilley from yesterday, I added the following feature of showing "This event". Would you test more than the following?

```javascript
    let element = await screen.getByTestId('this-event-1');
    expect(element).toHaveTextContent('This event');
    element = await screen.getByTestId('this-event-abc');
    expect(element).toHaveTextContent('');
```

First screenshot:
<hr />
image

Loading the second event will take us to the associated event and do the opposite ordering:
<hr />
image

malwilley: That looks like a good test to me!

malwilley (approved):

PR #71053, opened 2024-05-16
fix(backup): Avoid flushing outboxes

azaslavsky:
This seems to cause mysterious races for large imports in prod, and is not strictly necessary to do in-band anyway.

### DIFF
--- a/src/sentry/services/hybrid_cloud/import_export/impl.py
+++ b/src/sentry/services/hybrid_cloud/import_export/impl.py
@@ -29,6 +29,7 @@ from sentry.backup.scopes import ExportScope
 from sentry.db.models.base import BaseModel
 from sentry.db.postgres.transactions import in_test_hide_transaction_boundary
 from sentry.models.importchunk import ControlImportChunk, RegionImportChunk
+from sentry.models.outbox import outbox_context
 from sentry.models.user import User
 from sentry.models.userpermission import UserPermission
 from sentry.models.userrole import UserRoleUser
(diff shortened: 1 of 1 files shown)


corps (approved):
This is totally safe, although I am curious what sorts of weird races you saw?

src/sentry/services/hybrid_cloud/import_export/impl.py
```diff
@@ -189,8 +190,12 @@ def import_by_model(
                 logger.info("import_by_model.already_imported", extra=extra)
                 return existing_import_chunk
 
+            # We don't need the control and region silo synced into the correct `*Replica` tables
+            # immediately. The locally silo-ed versions of the models are written by the scripts
+            # themselves, and the remote versions will be synced a few minutes later, well before
+            # any users are likely ot need to get ahold of them to view actual data in the UI.
             using = router.db_for_write(model)
-            with transaction.atomic(using=using):
+            with outbox_context(transaction.atomic(using=using), flush=False):
```
GabeVillalobos, line 198: I don't think you can supply a transaction and the flush option together. If you still want this in a transaction, you'll need to nest the outbox context within the `transaction.atomic` context

azaslavsky:
> This is totally safe, although I am curious what sorts of weird races you saw?

https://sentry.sentry.io/issues/2501462103/events/62572c1746b94fe4ac3aa69ca1d33987/ in particular, and the traces associated with it. Only happens for `organizationteammember` weirdly enough.

src/sentry/services/hybrid_cloud/import_export/impl.py
azaslavsky: I do still want to retain the transaction.

I copied this code from `outboxes.py#L96`, which uses the same pattern. Is that a subtly different usage?

azaslavsky: One reason why I opted to use that, rather than nested context managers: it made for a much smaller diff, since I wouldn't have to push all ~150 lines of nested code to the right as well. But if it is incorrect, I am happy to do so!

GabeVillalobos: Oh whoops, I was misremembering the assertion we make here: https://github.com/getsentry/sentry/blob/08a47c9fa107d4e7018a99060500d39a73b64e1f/src/sentry/models/outbox.py#L854

I thought the options were mutually exclusive for some reason 🤦🏼‍♂️

azaslavsky: So this formulation is ok?

GabeVillalobos: Yep you're good!

GabeVillalobos (approved):

PR #71055, opened 2024-05-16
perf: move interfaces and integrations to orjson

anonrig:
Part of `orjson` move to get rid of `sentry.utils.json`

### DIFF
--- a/src/sentry/testutils/cases.py
+++ b/src/sentry/testutils/cases.py
@@ -17,6 +17,7 @@ from urllib.parse import urlencode
 from uuid import uuid4
 from zlib import compress
 
+import orjson
 import pytest
 import requests
 import responses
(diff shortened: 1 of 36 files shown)


src/sentry/interfaces/http.py
```diff
@@ -64,7 +64,7 @@ def fix_broken_encoding(value):
 
 
 def jsonify(value):
-    return to_unicode(value) if isinstance(value, str) else json.dumps(value)
+    return to_unicode(value) if isinstance(value, str) else orjson.dumps(value).decode()
```
asottile-sentry, line 67: I think this is leftover from python 2 -- can remove the `to_unicode` part of this at least since it does nothing

src/sentry/testutils/cases.py
```diff
@@ -1345,23 +1345,26 @@ def store_outcome(self, group):
...
             ).status_code
             == 200
         )
 
     def store_span(self, span):
         assert (
             requests.post(
-                settings.SENTRY_SNUBA + "/tests/entities/spans/insert", data=json.dumps([span])
+                settings.SENTRY_SNUBA + "/tests/entities/spans/insert",
+                data=orjson.dumps([span]).decode(),
```
asottile-sentry, line 1358: iirc `requests` can take `bytes` for `data`

iambriccardo (approved):
LGTM

asottile-sentry (approved):

PR #71059, opened 2024-05-16
test(backup): Add release test at 24.5.0

azaslavsky:
The test for 24.3.0 has been removed, as it is not outside the support window.


tests/sentry/backup/test_releases.py
```diff
@@ -129,9 +129,9 @@ def test_at_24_4_0(self):
             with open(tmp_path, "rb") as f:
                 import_in_global_scope(f, printer=NOOP_PRINTER)
 
-    def test_at_24_3_0(self):
```
hubertdeng123, line 132: should we delete the test_at_24_3_0.pysnap?

azaslavsky: It is - it's just that the diff here is odd and offset by 1. Look at the new file - only 24_4_N, and 24_5_0.

hubertdeng123: ah misread that, thanks for clarifying

hubertdeng123 (approved):

PR #71066, opened 2024-05-16
chore: update node.js to v20.13.1

anonrig:
Updates Node.js to latest version


evanpurkhiser (approved):
Let's make sure to do it in getsentry as well 👍

PR #71068, opened 2024-05-16
feat(crons): Display service incidents on timeline

evanpurkhiser:

### DIFF
--- a/static/app/views/monitors/components/timeline/gridLines.tsx
+++ b/static/app/views/monitors/components/timeline/gridLines.tsx
@@ -8,6 +8,7 @@ import {DateTime} from 'sentry/components/dateTime';
 import {space} from 'sentry/styles/space';
 import useRouter from 'sentry/utils/useRouter';
 
+import {CronServiceIncidents} from './serviceIncidents';
 import {useTimelineCursor} from './timelineCursor';
 import {useTimelineZoom} from './timelineZoom';
 import type {TimeWindowConfig} from './types';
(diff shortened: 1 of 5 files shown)


static/app/views/monitors/components/timeline/gridLines.tsx
```diff
@@ -23,6 +24,10 @@ interface Props {
    * Enable the timeline cursor
    */
   showCursor?: boolean;
+  /**
+   * Render sentry service incidents as an overlay
+   */
+  showIncidents?: boolean;
```
davidenwang, line 30: hmm are there instances where we actually don't want to show the incidents besides `isLoading`? I wonder if we should just have a `isLoading` here which disables/enables both `showCursor` and `showIncidents`

static/app/views/monitors/components/timeline/serviceIncidents.tsx
```diff
@@ -0,0 +1,170 @@
...
+import type {TimeWindowConfig} from './types';
+
+interface CronServiceIncidentsProps {
+  timeWindowConfig: TimeWindowConfig;
+}
+
+interface GetTimeWindowConfigOptions {
+  inc: StatuspageIncident;
+  /**
+   * Clamo the incident start and end times to be the timeWindowConfig start
```
davidenwang, line 24: nit: Clamp*

davidenwang (approved):
this is sweet

static/app/views/monitors/components/timeline/gridLines.tsx
evanpurkhiser: yeah we don't show it on the visualization either

davidenwang: ahh i see

PR #71075, opened 2024-05-16
fix(trace): adjust timeline for ns precision

JonasBa:
Spans have ns precision while transactions only include ms precision. This means that it is possible for spans to render outside of their "container" if their duration is sub ms and the txn duration is rounded down.


Abdkhan14 (approved):
With this change span children with ns precision will be exceeding the parent txn with ms precision, but at least they will show up within the bounds of the trace. This needs to be followed up with ns precision on txns fromt the trace endpoints.

PR #71076, opened 2024-05-16
feat(flagpole): Adds new feature handler strategy for flagpole

GabeVillalobos:
Adds a new feature handler strategy specifically for flagpole. The current implementation reuses the `OPTIONS` strategy, which is resulting in flags being evaluated by the wrong feature handler, so splitting the two into distinct use cases seems like the better option.

`OPTIONS` will be deprecated in the future once Flagpole lands, unless there's a compelling reason to keep it around.


markstory (approved):
Looks good. A default of `{}` will result in errors being trapped should a flag be used before it has a definition in the option which is what we want.

PR #71079, opened 2024-05-16
fix(filestore): add rate limit to event attachments

volokluev:
Currently we do not rate limit how many attachments a project can save, this causes INC-766 and INC-754

The numbers are pulled out of my butt, we'll see if CI passes


src/sentry/event_manager.py
```diff
@@ -2772,36 +2772,57 @@ def save_attachment(
...
-        size=file.size,
-        sha1=file.sha1,
-        # storage:
-        file_id=file.file_id,
-        blob_path=file.blob_path,
+    is_limited, num_requests, reset_time = ratelimiter.backend.is_limited_with_value(
+        key="event_attachment.save",
+        limit=options.get("sentry.save-event-attachments.project-per-5-minute-limit"),
+        project=project,
+        window=60,
```
vartec, line 2781: description says 5 minutes, but the window is 60 seconds

vartec (approved):
other than the window length detail, LGTM

src/sentry/event_manager.py
volokluev: Thank you

src/sentry/options/defaults.py
```diff
@@ -2529,6 +2529,14 @@
     "api.organization-activity.brownout-duration", default="PT1M", flags=FLAG_AUTOMATOR_MODIFIABLE
 )
 
+
+register(
+    "sentry.save-event-attachments.project-per-5-minute-limit",
+    type=Int,
+    default=20000,
```
JoshFerge, line 2536: i think we should make the default much much higher, or default infinity, and then ratchet it down from there

volokluev: Done

PR #71086, opened 2024-05-16
generating backpressure dd events

nirajdewani:
### Description
Now that sentry [supports](https://github.com/getsentry/sentry/pull/70399) Datadog events, \
updating backpressure code to generate events as well (in addition to metrics).

### Next steps
- Send events using dogshell
- Update event body if the need be


lynnagara:
What is the motivation for this? Does it tell us anything that the metric doesn't already?

mwarkentin:
Hey @lynnagara there was some more detail in the original PR / ticket: https://getsentry.atlassian.net/browse/OPS-5392

This would enable us to do things like:
* overlay events on top of dashboard widgets
* add tags to events to support analysis / grouping
* Trigger DD monitors for alerting purposes (eg. notifying slack / PagerDuty)

mwarkentin:
Here's an example time period where we experienced backpressure on May 6:

image

And split:

image

Some things that aren't clear:

* Why is everything a multiple of 10? 
* When exactly did backpressure trigger? 
* When did it resolve?

Here is one ingest consumer lag graph of the same time:

image

With events we can have discrete events that could include information from the backpressure system for more context. These can be overlaid on top of graphs:

image

src/sentry/processing/backpressure/health.py
```diff
@@ -68,6 +68,15 @@ def is_consumer_healthy(consumer_name: str = "default") -> bool:
                     "reason": reason,
                 },
             )
+            metrics.event(
+                f"{consumer_name}: backpressure.consumer.unhealthy",
```
mwarkentin, line 72: we might not want to include the consumer name in the event, we've got it as a tag and that might make it harder to combine the events for analysis, overlays, etc. Might be worth experimenting with manually to validate what we want name, tags, etc. set as based on the data that we have when we generate the event.

iambriccardo (approved):
LGTM

PR #71088, opened 2024-05-16
add loggers to comparison aggregations

nhsiehgit:


saponifi3d (approved):
not sure if log volume is an issue here or not -- this could get a bit noisy in the future so we'll want to make sure to remove after investigation.

also, is there an existing pattern to filter logs to a single rule / user?

nhsiehgit:
I'll add a TODO and note that this is for investigation and should be removed

src/sentry/incidents/subscription_processor.py
```diff
@@ -231,6 +235,19 @@ def get_comparison_aggregation_value(
...
+        # TODO: logger for investigation. should be removed
+        logger.info(
+            "get_comparison_aggregation_value",
+            extra={
+                "alert_rule_id": self.alert_rule.id,
+                "subscription_id": subscription_update.subscription_id,
+                "organization_id": self.alert_rule.organization_id,
+                "comparison_aggregate": comparison_aggregate,
+                "aggregation_value": aggregation_value,
+                "result_data": results["data"],
```
ceorourke, line 247: maybe do `.get()` here to be safe?

PR #71089, opened 2024-05-17
Fix for configoptions, remove check for isset()

kneeyo1:
We only need check the `last_updated_by`, 

if the option is already unset (removed from db), last_updated_by will be `None`. All other channels we can write as drift.


src/sentry/options/manager.py
```diff
@@ -270,6 +270,12 @@ def isset(self, key: str) -> bool:
 
         return self.is_set_on_disk(key)
 
+    def set_on_db(self, key: str) -> bool:
+        """
+        Check if a key has been set on the database.
```
markstory, line 275:
```suggestion
    def set_in_db(self, key: str) -> bool:
        """
        Check if a key has been set in the database.
```

Values get set *in* databases, and *on* disk.

src/sentry/runner/commands/configoptions.py
```diff
@@ -279,6 +279,10 @@ def sync(ctx: click.Context) -> None:
                         # we can safely continue.
                         if options.is_set_on_disk(opt.name):
                             continue
+                        # isset() will pull from local cache, which gets updated with every get() call.
+                        # configoptions is only concerned with removing the option from the db.
+                        elif not options.set_on_db(opt.name):
```
markstory, line 284: We could also change `isset()` so that it behaves correctly. Currently the docstring for `isset` claims

>  Check if a key has been set to a value and not inheriting from its default.

Which doesn't match its current behavior. We could have isset compare the value from the store to the option default. If the store value and default are the same, the option is not 'set'.

kneeyo1: I originally wanted to do that but serveral other files and methods use `issset()`, worried I could break their functionality

kneeyo1: The other issue is that if this is what we use to compare options, `configoptions` would never actually remove options from the DB if the DB value is just the default value.

markstory: > I originally wanted to do that but serveral other files and methods use issset(), worried I could break their functionality

We could review those callsites and see if their behaviour would change.

> The other issue is that if this is what we use to compare options, configoptions would never actually remove options from the DB if the DB value is just the default value.

If the value is set in the store and equal to the default, wouldn't we still be able to check the option's update channel?

tests/sentry/runner/commands/test_configoptions.py
```diff
@@ -208,6 +208,48 @@ def test_bad_sync(self):
...
 
+    def test_sync_unset_options(self):
+
+        # test options set on disk with and without prioritize disk, tracked
+        # and not tracked
+        # test options set on db, verify that untracked options are properly # deleted
+
+        rv = self.invoke(
+            "-f",
+            "tests/sentry/runner/commands/unsetsync.yaml",
```
markstory, line 219: You could put this yaml file with other test data in `fixtures`. It could be helpful to have an `options` directory for this and other test fixture files.

tests/sentry/runner/commands/test_configoptions.py
```diff
@@ -208,6 +208,48 @@ def test_bad_sync(self):
...
+                ConsolePresenter.SET_MSG % ("map_option", {"a": 1, "b": 2}),
+                ConsolePresenter.SET_MSG % ("list_option", [1, 2]),
+                ConsolePresenter.UNSET_MSG % "str_option",
+                ConsolePresenter.UNSET_MSG % "to_unset_option",
+            ]
+        )
+
+        assert expected_output in rv.output
+
+        assert options.get("int_option") == 20
```
markstory, line 237: If `int_option` wasn't in the `unsetsync.yml` file shouldn't it have been removed?

kneeyo1: it would be removed, and the corresponding `get` would return its default value (which is 20).

markstory (approved):
Looks good to me. Options that were automated and have been removed from the options data will now be removed if the option's last update was through automator. other states that aren't default/disk values will result in drift :+1:

PR #71092, opened 2024-05-17
ref(tsc): convert dataDownload to FC

michellewzhang:

### DIFF
--- a/static/app/views/dataExport/dataDownload.tsx
+++ b/static/app/views/dataExport/dataDownload.tsx
@@ -216,7 +216,7 @@ function DataDownload({params: {orgId, dataExportId}}: Props) {
           {renderOpenInDiscover()}
           <p>
             <small>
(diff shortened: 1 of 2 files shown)


static/app/views/dataExport/dataDownload.tsx
```diff
@@ -206,12 +211,12 @@ class DataDownload extends DeprecatedAsyncView<Props, State> {
...
             <br />
-            {this.renderDate(dateExpired)}
+            {renderDate(dateExpired)}
           </p>
-          {this.renderOpenInDiscover()}
+          {renderOpenInDiscover()}
           <p>
             <small>
-              <strong>SHA1:{checksum}</strong>
+              <strong>{tct('SHA1: ', {checksum})}</strong>
```
JonasBa, line 219: I dont think this will ever translate to anything since SHA1 is just the name of the hashing algorithm

JonasBa (approved):

PR #71098, opened 2024-05-17
ref: rework build to be reusable outside of setup.py

asottile-sentry:
need to get off setuptools so we can get to python 3.12 -- and so we can conditionally do the js build only for self-hosted (since we do it twice for getsentry for no reason!)

### DIFF
--- a/src/sentry/build/_static_assets.py
+++ b/src/sentry/build/_static_assets.py
@@ -1,0 +1,40 @@
+from __future__ import annotations
+
+import argparse
+import os
+import subprocess
+
+
+def _build_static_assets() -> None:
+    node_options = os.environ.get("NODE_OPTIONS", "")
+    env = {
+        **os.environ,
+        # By setting NODE_ENV=production, a few things happen
+        #   * React optimizes out certain code paths
+        #   * Webpack will add version strings to built/referenced assets
+        "NODE_ENV": "production",
+        # TODO: Our JS builds should not require 4GB heap space
+        "NODE_OPTIONS": f"{node_options} --max-old-space-size=4096".lstrip(),
+    }
+
+    def _cmd(*cmd: str) -> None:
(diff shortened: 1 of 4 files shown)


src/sentry/build/_static_assets.py
```diff
@@ -0,0 +1,40 @@
...
+
+    _cmd("yarn", "install", "--production", "--frozen-lockfile", "--quiet")
+    _cmd("yarn", "tsc", "-p", "config/tsconfig.build.json")
+    _cmd("yarn", "build-production", "--bail")
+    _cmd("yarn", "build-chartcuterie-config", "--bail")
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser()
+    parser.parse_args()
```
IanWoodard, line 33: Are we planning to make use of this in the future?

IanWoodard (approved):

webpack.config.ts
```diff
@@ -113,7 +113,7 @@ const ENABLE_CODECOV_BA = env.CODECOV_ENABLE_BA === 'true' ?? false;
 // this is the path to the django "sentry" app, we output the webpack build here to `dist`
 // so that `django collectstatic` and so that we can serve the post-webpack bundles
 const sentryDjangoAppPath = path.join(__dirname, 'src/sentry/static/sentry');
-const distPath = env.SENTRY_STATIC_DIST_PATH || path.join(sentryDjangoAppPath, 'dist');
```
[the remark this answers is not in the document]
asottile-sentry: it was only ever set to exactly the same value as this

src/sentry/build/_static_assets.py
asottile-sentry: it's mostly so `thing --help` doesn't make it just do it -- even though there's no options

PR #71100, opened 2024-05-17
chore(staff): Convert sudo modal to functional

schew2381:
Change this component to functional in order to try using `useContext` in a future PR.

I removed `busy` because it wasn't used anywhere 🤔

Props to @michellewzhang for carrying most of the work on this 😎

### DIFF
--- a/static/app/components/modals/sudoModal.tsx
+++ b/static/app/components/modals/sudoModal.tsx
@@ -28,6 +28,15 @@ type DefaultProps = {
   closeButton?: boolean;
 };
 
+type State = {
+  authenticators: Array<Authenticator>;
+  error: boolean;
+  errorType: string;
+  showAccessForms: boolean;
+  superuserAccessCategory: string;
+  superuserReason: string;
+};
+
 type Props = WithRouterProps &
   DefaultProps &
   Pick<ModalRenderProps, 'Body' | 'Header'> & {
@@ -44,10 +53,20 @@ type Props = WithRouterProps &
     retryRequest?: () => Promise<any>;
   };
 
-function SudoModal(props: Props) {
-  const [state, setState] = useState({
+function SudoModal({
+  closeModal,
+  isSuperuser,
+  location,
+  needsReload,
+  router,
+  retryRequest,
+  api,
+  Header,
+  Body,
+  closeButton,
+}: Props) {
+  const [state, setState] = useState<State>({
     authenticators: [] as Authenticator[],
-    busy: false,
     error: false,
     errorType: '',
     showAccessForms: true,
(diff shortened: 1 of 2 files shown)


static/app/views/organizationContext.tsx
```diff
@@ -24,7 +24,7 @@ export const OrganizationContext = createContext<Organization | null>(null);
 /**
  * Holds a function to load the organization.
  */
-const OrganizationLoaderContext = createContext<null | (() => void)>(null);
+export const OrganizationLoaderContext = createContext<null | (() => void)>(null);
```
michellewzhang, line 27: why did this get exported

static/app/components/modals/sudoModal.tsx
```diff
@@ -44,76 +53,90 @@ type Props = WithRouterProps &
...
+      } catch {
+        // ignore errors
+      }
+    };
 
-  componentDidMount() {
-    this.getAuthenticators();
-  }
+    getAuthenticators();
+  }, [api]);
```
schew2381, line 100: Is this `api` dependency necessary? I tried removing it, but the the test `can delete an org with sudo flow` fails with:
```
Error: No mocked response found for request: GET /authenticators/
```
even though it's being mocked in the `beforeEach` and the actual test passes

saponifi3d (approved):
looks generally good to me, feel free to treat comments as nits.

static/app/components/modals/sudoModal.tsx
```diff
@@ -44,76 +53,90 @@ type Props = WithRouterProps &
...
+function SudoModal({
+  closeModal,
+  isSuperuser,
+  location,
+  needsReload,
+  router,
+  retryRequest,
+  api,
+  Header,
+  Body,
```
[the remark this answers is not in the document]
cathteng: useRouter is deprecated, what should be used instead? we can replace router, api, and location with hooks

static/app/components/modals/sudoModal.tsx
cathteng: yes because we are using it in the useEffect

static/app/components/modals/sudoModal.tsx
```diff
@@ -127,95 +150,77 @@ class SudoModal extends Component<Props, State> {
...
-  handleLogout = async () => {
-    const {api} = this.props;
+  const handleLogout = async () => {
     try {
       await logout(api);
     } catch {
       // ignore errors
     }
-    window.location.assign(this.getAuthLoginPath());
+    window.location.assign(getAuthLoginPath());
```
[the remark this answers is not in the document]
schew2381: We need one more request to validate ⛳️

PR #71105, opened 2024-05-17
fix(slack): Escape assignee dropdown input

isabellaenriquez:
Fixes SENTRY-2MBQ


tests/sentry/integrations/slack/webhooks/options_load/test_dynamic_assignment_dropdown.py
```diff
@@ -103,6 +103,26 @@ def test_simple(self):
...
+        self.original_message["blocks"][0]["block_id"] = orjson.dumps(
+            {"issue": self.group.id}
+        ).decode()
+
+        self.user1 = self.create_user(email="aaa@testing.com", name="Alice")
+        self.create_member(organization=self.organization, user=self.user1, teams=[self.team1])
+        self.store_event(data=self.event_data, project_id=self.project.id)
+
+        # shouldn't fail
+        resp = self.post_webhook(substring="Al[", original_message=self.original_message)
```
sentaur-athena, line 121: why [? 🤔

sentaur-athena: I guess my question is what was the problem for it not working before? Trying to learn 😅

isabellaenriquez: on the user's side nothing seems wrong! but on our side when a user inputs special regex characters like `\` or unterminated `[` we get an [error in sentry](https://sentry.sentry.io/issues/4973916773/events/038116026f204c4fb7cc98e0edae5a50/) because the input is not escaped properly. this pr escapes the input so that when we process the input we don't interpret those characters as special regex and instead just interpret them as is. so `Al[` would become `Al\\[`

ykamo001 (approved):

tests/sentry/integrations/slack/webhooks/options_load/test_dynamic_assignment_dropdown.py
ykamo001: I think this SO post describes what Isabella is trying to fix: https://stackoverflow.com/questions/54135606/python-re-error-unterminated-character-set-at-position

However, it looks like for `[` and `{` we only need to escape the opening, and not the closing. But for `(` we also need to do it for ending `)`

ykamo001: The docs mention this situation with `re.match` method: https://docs.python.org/3/library/re.html#raw-string-notation

iamrajjoshi (approved):

PR #71119, opened 2024-05-17
chore(hybridcloud) Expand OrgMapping update payload

markstory:
Our schema generation and validation tooling doesn't work well with tuple values. Replace this tuple with an 'Optional' that retains the ability to preserve 'undefined' vs 'defined and none'.

Refs HC-1190


src/sentry/services/hybrid_cloud/organization_mapping/model.py
```diff
@@ -32,7 +36,8 @@ class RpcOrganizationMappingUpdate(RpcModel):
     # When not set, no change to customer id performed,
     # when set with a tuple, the customer_id set to either None or the string
     # that is the first element.
-    customer_id: tuple[str | None] | None = None
+    # Using a tuple is deprecated and will be removed.
+    customer_id: tuple[str | None] | CustomerId | None = None
```
GabeVillalobos, line 40: What's the benefit of using a class here vs just passing an int? Will we be adding more attributes to it in the future?

GabeVillalobos (approved):
Changes look good, just had a minor question about the param types

src/sentry/services/hybrid_cloud/organization_mapping/model.py
markstory: The current implementation uses a single element tuple of `str | None` to indicate a value/explicit None vs an implicit None because the key is undefined. The class lets us wrap None and maintain the same semantics as the tuple.

PR #71122, opened 2024-05-17
feat(new-trace): Sorting trace drawer card content by subject.

Abdkhan14:


static/app/views/metrics/customMetricsEventData.tsx
```diff
@@ -228,7 +228,11 @@ export function CustomMetricsEventData({
     });
 
     return (
-      <TraceDrawerComponents.SectionCard title={t('Emitted Metrics')} items={items} />
+      <TraceDrawerComponents.SectionCard
+        title={t('Emitted Metrics')}
+        items={items}
+        enableSorting
```
JonasBa, line 234:
```suggestion
        sortAlphabetically
```

JonasBa (commented):
I'm going to wait for others to review as we dont own this code. Besides the prop name change, the changes look fine to me

static/app/components/events/contexts/contextCard.tsx
```diff
@@ -113,9 +113,13 @@ export default function ContextCard({
     project,
   });
 
-  const content = contextItems.map((item, i) => (
-    <ContextCardContent key={`context-card-${i}`} meta={meta} item={item} />
-  ));
+  const content = contextItems
+    .sort((a, b) => {
+      return a.subject.localeCompare(b.subject);
+    })
```
leeandher, line 119: This will alter the [order for context](https://github.com/getsentry/sentry/blob/master/static/app/components/events/contexts/index.tsx#L40-L45) on issues details page as well. I preserved that order because it was what was there before, but I don't see a reason why we can't change it. Alphabetically seems good to me 👍

Abdkhan14: Heyo @leeandher, it only changes the order of the items inside the card. The context ordering is one level above this, it won't change.

Abdkhan14: Noted @leeandher, lets give sorting alphabetically a go.

Abdkhan14: > @Abdkhan14 .sort mutates the original array, so this might in fact change the order. It would be safer to create a copy of contextItems before sorting


Screenshot 2024-05-21 at 12 17 24 PM

leeandher (approved):
Code looks good, but are there any instances that don't get receive `sortAlphabetically = true`? If not, you can probably skip the prop and just apply the sort for all renders of `SectionCard`

Abdkhan14:
> Code looks good, but are there any instances that don't get receive sortAlphabetically = true? If not, you can probably skip the prop and just apply the sort for all renders of SectionCard

Yeah we don't sort the general Info sections.

PR #71130, opened 2024-05-17
feat(events-graph): Update Event stats graph on issue stream

### LINKED ISSUE #69691: Issues Stream: Update event stats chart to include issue status

MichaelSun48:
This PR updates the event stats graph on the issue stream to a newer design, and it changes the issue sub-status (ongoing, escalating, etc.) from a color-themed pill on the left side of the row, to a single-color label under the new events graph. 

This change is released under a feature flag, which is currently set to 100% LA to Sentry orgs only. 

New UI at a glance: (Preview will be better for this tho, please use that)
image

Full list of changes:
* The issue sub-status is now one color and is a label under the graph
* The markline showing the peak events in a given period has higher opacity and is moved to the right side of the graph
* The bars of the graph now animate and rise from the x-axis upon first render (this feature may need to be turned off if we observe any performance issues) 
* A placeholder has been added for the loading state of the graph (previously there was no loading state at all) 
* The `<Placeholder/>` component has been given a slightly rounded edge


static/app/components/charts/groupStatusChart.stories.tsx
```diff
@@ -5,29 +5,29 @@ import storyBook from 'sentry/stories/storyBook';
 import type {TimeseriesValue} from 'sentry/types';
 
 const stats: ReadonlyArray<TimeseriesValue> = [
```
MichaelSun48, line 7: Just testing purposes, ignore

static/app/components/stream/group.tsx
```diff
@@ -496,7 +514,7 @@ function BaseGroupRow({
             <EventCountsWrapper>{groupCount}</EventCountsWrapper>
           )}
           {withColumns.includes('users') && issueTypeConfig.stats.enabled && (
-            <EventCountsWrapper>{groupUsersCount}</EventCountsWrapper>
+            <EventUsersCountsWrapper>{groupUsersCount}</EventUsersCountsWrapper>
```
MichaelSun48, line 517: The EventCountsWrapper has been given slightly less left margin to better align the event stats graph with the column header

static/app/components/stream/group.tsx
```diff
@@ -480,12 +495,16 @@ function BaseGroupRow({
...
       {withChart && !displayReprocessingLayout && issueTypeConfig.stats.enabled && (
         <ChartWrapper narrowGroups={narrowGroups}>
-          <GroupChart
-            stats={groupStats}
-            secondaryStats={groupSecondaryStats}
-            showSecondaryPoints={showSecondaryPoints}
-            showMarkLine
-          />
+          {organization.features.includes('issue-stream-new-events-graph') ? (
+            groupChart
```
scttcper, line 499: i'd define move the ternary here to avoid GroupStatusChart from being called just to be unused

static/app/components/charts/miniBarChart.tsx
```diff
@@ -265,22 +272,12 @@ function MiniBarChart({
         updated.stack = 'stack1';
       }
       set(updated, 'itemStyle.color', colorList[i]);
-      set(updated, 'itemStyle.opacity', barOpacity); // Opacity of each bar
       set(updated, 'itemStyle.borderRadius', [1, 1, 0, 0]); // Rounded corners on top of the bar
-      set(updated, 'emphasis.itemStyle.opacity', 1.0);
```
MichaelSun48, line 270: These settings were for the entire graph. We now set them manually per bar because we want to conditionally make the bar invisible (set opacity=0 and emphasis opacity=0) if the value of the bar is 0. This logic is reflected in `barChart.tsx`

static/app/components/charts/barChart.tsx
```diff
@@ -12,25 +12,46 @@ export interface BarChartSeries
...
+          value: [name, value],
+          itemStyle:
+            value === 0 && hideZeros
+              ? {...itemStyle, opacity: 0} // Don't show bars when value = 0 (when hideZeros is enabled)
+              : {...itemStyle, opacity: barOpacity},
+          emphasis:
+            value === 0 && hideZeros
+              ? {itemStyle: {opacity: 0}}
+              : {itemStyle: {opacity: 1}},
+        };
```
MichaelSun48, line 54: New logic that makes the bar invisible when the value of the bar is 0. 

The existing events graph in the issue stream shows a bar with minimum height even if the value is 0. This is because the existing graph does not show the x axis, so if we hid the bar with 0 values, the graph would look very disjointed.

malwilley (approved):

PR #71136, opened 2024-05-17
feat(insights): add status and retry filtering to queue sample panel

KevinL10:
Adds status and retry filtering to the queue sample panel. Retry filters are only available for consumers.

image

### DIFF
--- a/static/app/views/performance/queues/destinationSummary/messageSpanSamplesPanel.tsx
+++ b/static/app/views/performance/queues/destinationSummary/messageSpanSamplesPanel.tsx
@@ -120,15 +120,6 @@ export function MessageSpanSamplesPanel() {
   search.addFilterValue('transaction', query.transaction);
   search.addFilterValue('messaging.destination.name', query.destination);
 
-  if (query.status.length > 0) {
-    search.addFilterValue('trace.status', query.status);
-  }
-
-  if (query.retries.length > 0) {
-    // TODO: kevin - index retry.count
-    search.addFilterValue('measurements.messaging.message.retry.count', '1');
-  }
-
   const {data: transactionMetrics, isFetching: aretransactionMetricsFetching} =
     useQueuesMetricsQuery({
       destination: query.destination,
@@ -154,6 +145,24 @@ export function MessageSpanSamplesPanel() {
 
   const durationAxisMax = computeAxisMax([durationData?.[`avg(span.duration)`]]);
 
+  if (query.status.length > 0) {
+    search.addFilterValue('trace.status', query.status);
+  }
+
+  // Note: only consumers should show the retry count filter
+  if (messageActorType === MessageActorType.CONSUMER) {
+    if (query.retries === '0') {
+      search.addFilterValue('measurements.messaging.message.retry.count', '0');
+    } else if (query.retries === '1-3') {
+      search.addFilterValues('measurements.messaging.message.retry.count', [
+        '>=1',
+        '<=3',
+      ]);
+    } else if (query.retries === '4+') {
+      search.addFilterValue('measurements.messaging.message.retry.count', '>=4');
+    }
+  }
+
   const {
     data: durationSamplesData,
     isFetching: isDurationSamplesDataFetching,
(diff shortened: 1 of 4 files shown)


static/app/views/performance/queues/destinationSummary/messageSpanSamplesPanel.tsx
```diff
@@ -106,6 +143,24 @@ export function MessageSpanSamplesPanel() {
 
   const durationAxisMax = computeAxisMax([durationData?.[`avg(span.duration)`]]);
 
+  if (query.status.length > 0) {
```
JonasBa, line 146: I would add comments as to why we are injecting these values here and on L150 so the next person reading this knows why it's there.

static/app/views/performance/queues/destinationSummary/messageSpanSamplesPanel.tsx
```diff
@@ -67,6 +72,38 @@ export function MessageSpanSamplesPanel() {
     ? [query.destination, query.transaction].filter(Boolean).join(':')
     : undefined;
 
+  const handleStatusChange = newStatus => {
```
JonasBa, line 75: While we do allow implicit any in our codebase, I would rather you add a type so that we know what this function expects as an argument.

static/app/views/performance/queues/destinationSummary/messageSpanSamplesPanel.tsx
```diff
@@ -67,6 +72,38 @@ export function MessageSpanSamplesPanel() {
...
+    router.replace({
+      pathname: location.pathname,
+      query: {
+        ...location.query,
+        status: newStatus.value,
+      },
+    });
+  };
+
+  const handleRetriesChange = newRetries => {
```
JonasBa, line 91: Same implicit any comment here

static/app/views/performance/queues/destinationSummary/messageSpanSamplesPanel.tsx
KevinL10: thanks, added a comment and the function types

gggritso: And/or would be even better to name the variables appropriately. e.g.,

```jsx
const timeseriesFilters ...
const sampleFilters ...
```

then people are less likely to update the wrong variable

static/app/views/performance/queues/destinationSummary/messageSpanSamplesPanel.tsx
```diff
@@ -356,6 +438,57 @@ function ConsumerMetricsRibbon({
...
+    'out_of_range',
+    'unimplemented',
+    'internal_error',
+    'unavailable',
+    'data_loss',
+    'unauthenticated',
+  ].map(status => {
+    return {
+      value: status,
+      label: t('%s', status),
```
gggritso, line 468: 🤔 This is interesting. I agree with the intuition to translate this, but is this translated in the trace view, too? If not, this translated value won't match what people see there, and it'd be confusing. Might be better not to translate

static/app/views/performance/queues/destinationSummary/messageSpanSamplesPanel.tsx
```diff
@@ -43,11 +45,14 @@ import {useSampleScatterPlotSeries} from 'sentry/views/starfish/views/spanSummar
 
 export function MessageSpanSamplesPanel() {
   const router = useRouter();
+  const location = useLocation();
   const query = useLocationQuery({
     fields: {
       project: decodeScalar,
       destination: decodeScalar,
       transaction: decodeScalar,
+      retries: decodeScalar,
```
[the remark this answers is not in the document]
gggritso: I think a custom decoder would be better here. If `retries` was a simple integer, then no problem, but it's not, it's string values. It needs validation, a fallback, etc. `decodeScalar` probably won't do. Check the `responseCodeClass` decode for an example, if you like!

gggritso (commented):
Makes sense! A few stray comments, but nothing major

gggritso (approved):
👍🏻 :shipit:

PR #71142, opened 2024-05-17
fix(u2f): Have authenticators GET wait for preloading to finish

schew2381:
In order to have the U2F state be correctly set in the client browser session cookie, we need to wait for preloading requests to finish first to avoid a race condition. This was done by:

1. Converting the `sudo modal` to a functional component so we can use `useContext` in it https://github.com/getsentry/sentry/pull/71100
2. Keeping track of when the preloading finishes with a hook in the `organizationLoadingContext` https://github.com/getsentry/sentry/pull/71112
3. Finally this PR, which finally has the U2F utilize the hook to wait for preloading to finish before making an authenticators GET request

Note that the PRs mentioned above must be merged first

### Local Example

https://github.com/getsentry/sentry/assets/67301797/fef1a475-4e07-40ca-9741-03f6b463d5f7


static/app/components/modals/sudoModal.tsx
```diff
@@ -127,95 +155,77 @@ class SudoModal extends Component<Props, State> {
...
-  handleLogout = async () => {
-    const {api} = this.props;
+  const handleLogout = async () => {
     try {
       await logout(api);
     } catch {
       // ignore errors
     }
-    window.location.assign(this.getAuthLoginPath());
+    window.location.assign(getAuthLoginPath());
```
saponifi3d, line 219: Should this be using `router` to redirect the user or do we need to do a full server request to validate auth cookies?

static/app/components/modals/sudoModal.tsx
```diff
@@ -44,76 +54,94 @@ type Props = WithRouterProps &
...
+function SudoModal({
+  closeModal,
+  isSuperuser,
+  location,
+  needsReload,
+  router,
+  retryRequest,
+  api,
+  Header,
+  Body,
```
saponifi3d, line 66: could any of these props become imports or hooks? for example we could replace `withSentryRouter` with `const router = useRouter()` and define it in the component itself.

static/app/views/organizationContext.tsx
```diff
@@ -82,8 +90,15 @@ export function OrganizationContextProvider({children}: Props) {
       return;
     }
 
+    const logSet = () => {
```
saponifi3d, line 93:
```suggestion
    const updateLoadingState = () => {
```
nit -- can we rename this function or just define it inline for the promise chain (this is already a nested anonymous function, so we won't get any hot process wins by naming it)

static/app/actionCreators/organization.tsx
```diff
@@ -132,49 +132,55 @@ export function fetchOrganizationDetails(
...
-            : typeof err.responseJSON?.detail?.message === 'string'
-              ? err.responseJSON?.detail.message
-              : null;
+        if (err.status === 403 || err.status === 401) {
+          const errMessage =
+            typeof err.responseJSON?.detail === 'string'
+              ? err.responseJSON?.detail
+              : typeof err.responseJSON?.detail?.message === 'string'
+                ? err.responseJSON?.detail.message
+                : null;
```
saponifi3d, line 153: nit: can we destruct / simplify this ternary? Is there a helper to read API error messages already?

static/app/components/modals/sudoModal.tsx
```diff
@@ -44,76 +54,94 @@ type Props = WithRouterProps &
...
+  const {
+    authenticators,
+    error,
+    errorType,
+    showAccessForms,
+    superuserAccessCategory,
+    superuserReason,
+  } = state;
+
+  const isLoading = useContext(OrganizationLoaderContext)?.isLoading;
```
saponifi3d, line 87:
```suggestion
  const  { isLoading } = useContext(OrganizationLoaderContext);
```
nit: Recommend destructing the value (if there's a type error you can do ` || {}` to get around that)

static/app/views/organizationContext.tsx
```diff
@@ -24,18 +33,15 @@ export const OrganizationContext = createContext<Organization | null>(null);
...
+export const OrganizationLoaderContext =
+  createContext<OrganizationLoaderContextProps | null>(null);
 
 /**
  * Ensures that an organization is loaded when the hook is used. This will only
  * be done on first render and if an organization is not already loaded.
  */
 export function useEnsureOrganization() {
-  const loadOrganization = useContext(OrganizationLoaderContext);
+  const loadOrganization = useContext(OrganizationLoaderContext)?.loadOrganization;
```
saponifi3d, line 44:
```suggestion
  const { loadOrganization } = useContext(OrganizationLoaderContext);
```

static/app/views/organizationContext.tsx
```diff
@@ -16,6 +16,15 @@ import useApi from 'sentry/utils/useApi';
 import {useParams} from 'sentry/utils/useParams';
 import {useRoutes} from 'sentry/utils/useRoutes';
 
+interface OrganizationLoaderContextProps {
+  isLoading: boolean;
+  loadOrganization: () => void;
```
saponifi3d, line 21: it might be cool to have this return a `Promise` rather than void, then you wouldn't need to maintain a separate `isLoading` variable, instead you could just continue the promise chain and use the promises state to for the load state.

static/app/views/organizationContext.tsx
```diff
@@ -16,6 +16,15 @@ import useApi from 'sentry/utils/useApi';
...
+}
+
+interface Props {
+  children: React.ReactNode;
```
saponifi3d, line 25: nit: add `type ReactNode` to the react import, it helps illustrate the dependency chain and doesn't utilize global scope then.

saponifi3d (approved):
just nits, generally lgtm.

static/app/views/organizationContext.tsx
schew2381: whoops I think I did this for testing with logs and forgot to switch it back

static/app/actionCreators/organization.tsx
cathteng: i pulled it out into a function

static/app/components/modals/sudoModal.tsx
```diff
@@ -83,25 +84,29 @@ function SudoModal({
...
-    const getAuthenticators = async () => {
+    const getAuthenticators = () => {
+      if (!loadOrganization) return;
       try {
-        const fetchedAuthenticators = await api.requestPromise('/authenticators/');
-        setState(prevState => ({
-          ...prevState,
-          authenticators: fetchedAuthenticators ?? [],
-        }));
+        loadOrganization().then(async () => {
```
saponifi3d, line 96: i don't think we'll need this inside of a `useEffect`, i believe you can just grab the `loadOrganization.then(()` here instead.

This is because the `OrganizationContextLoader` will initialize the promise chain, then here we're referencing the promise directly, so whenever the promise resolves it will call `.then`.

static/app/views/organizationContext.tsx
saponifi3d: 🔥 i'd say this is like 99% there, i left a comment on the `useEffect` but i think you can basically remove it and be good to go.

PR #71146, opened 2024-05-17
feat(crons): Add processing error display to monitor details page

davidenwang:
Shows errors that occurred while processing check-ins specific to a monitor

image

image
image


davidenwang:
Closing this in favor of a view that will be collapsible

PR #71148, opened 2024-05-18
fix(ui): Add missing 'return' statement in <StructuredEventData />

floels:
# Goal

The goal of this PR is to add a missing `return` statement in the declaration of the `<StructuredEventData />` component in `static/app/components/structuredEventData/index.tsx`.

# Approach

We also add a test in the existing test file (`index.spec.tsx`) to cover the corresponding use case (multiline strings). You can check that this test fails if you remove the fix in `index.tsx`.

# Legal Boilerplate


static/app/components/structuredEventData/index.tsx
```diff
@@ -202,16 +202,18 @@ export function StructuredData({
...
-            value={value}
-            meta={meta}
-            withAnnotatedText={withAnnotatedText}
-            withOnlyFormattedText={withOnlyFormattedText}
-          />
-        </ValueMultiLineString>
-      </Wrapper>;
+      return (
+        <Wrapper>
+          <ValueMultiLineString data-test-id="value-multiline-string">
```
floels, line 207: We also add a test id for the purpose of the test added in `index.spec.tsx` above.

HazAT (approved):

PR #71149, opened 2024-05-18
fix(ui): Remove unused expression in 'highlightsDataSection.spec.tsx'

floels:
# Goal

The goal of this PR is to remove an unused expression at the top of file `static/app/components/ events/highlights/highlightsDataSection.spec.tsx`:

https://github.com/getsentry/sentry/blob/3b17f232c2c704a660bfcdaec0e6daa11195228e/static/app/components/events/highlights/highlightsDataSection.spec.tsx#L1-L18

# Legal Boilerplate


leeandher (approved):
Thanks for this!

PR #71152, opened 2024-05-19
feat(trace): search using span. or transaction. entity prefix

JonasBa:
Enables search by narrowing down on the entity types via span. and transaction. prefixes (errors, issues and autogroups still need to be implemented). Previously if two objects would had the same key, the value would be resolved from either of them and both were subject to match. With this change, searching narrows down on the entity by the prefix.

There is a slight annoyance with txns because they contain keys like `transaction.op` as opposed to spans which only contain a key like `op`, which is why we need to perform a direct lookup first, and only then look at the entity.


static/app/components/searchSyntax/evaluator.tsx
```diff
@@ -99,30 +99,29 @@ function processTokenResults(tokens: TokenResult<Token>[]): ProcessedTokenResult
 }
 
 function isBooleanOR(token: ProcessedTokenResult): boolean {
-  return (
-    token && token.type === Token.LOGIC_BOOLEAN && token.value === BooleanOperator.OR
-  );
+  return token?.type === Token.LOGIC_BOOLEAN && token?.value === BooleanOperator.OR;
```
anonrig, line 102: Shouldn't line 101 be `ProcessedTokenResult | undefined`?

static/app/components/searchSyntax/evaluator.tsx
```diff
@@ -99,30 +99,29 @@ function processTokenResults(tokens: TokenResult<Token>[]): ProcessedTokenResult
...
+          // Establishes higher precedence for OR operators.
+          // Whenever the current stack top operator is an OR,
           stack.length > 0 &&
           token.value === BooleanOperator.AND &&
           stack[stack.length - 1].type === Token.LOGIC_BOOLEAN &&
           stack[stack.length - 1].type !== 'L_PAREN' &&
           isBooleanOR(stack[stack.length - 1])
         ) {
-          result.push(stack.pop() as ProcessedTokenResult);
+          result.push(stack.pop()!);
```
anonrig, line 124: We should actually check if `pop()` result exists

static/app/components/searchSyntax/evaluator.tsx
JonasBa: Nah, this narrows down on the type. ProcessedTokenResult has different shapes, so this checks that it has a type and then narrows down to the boolean or as well.

anonrig (approved):

Abdkhan14 (approved):

PR #71155, opened 2024-05-20
perf: replace json parser with orjson

anonrig:
Part of `orjson` move to get rid of `sentry.utils.json`

### DIFF
--- a/src/sentry/api/endpoints/source_map_debug_blue_thunder_edition.py
+++ b/src/sentry/api/endpoints/source_map_debug_blue_thunder_edition.py
@@ -1,5 +1,6 @@
 from typing import Literal, TypedDict
 
+import orjson
 import sentry_sdk
 from django.db.models import QuerySet
 from django.utils.encoding import force_bytes, force_str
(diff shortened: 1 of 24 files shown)


src/sentry/api/endpoints/source_map_debug_blue_thunder_edition.py
```diff
@@ -416,7 +416,7 @@ def _find_source_file_in_artifact_indexes(self):
             self._get_dist_matched_artifact_index_release_file()
         )
         if dist_matched_artifact_index_release_file is not None:
-            raw_data = json.load(dist_matched_artifact_index_release_file.file.getfile())
+            raw_data = orjson.loads(dist_matched_artifact_index_release_file.file.getfile().read())
```
asottile-sentry, line 419: oh interesting, this might be the leaked socket I've been looking for

each call to `getfile` should be wrapped in context manager otherwise it leaks a file descriptor

no need to fix here though

asottile-sentry (approved):

src/sentry/api/endpoints/source_map_debug_blue_thunder_edition.py
anonrig: nice

anonrig:
@asottile-sentry Can you re-review? I had to make some changes due to failing tests.

asottile-sentry (approved):

PR #71162, opened 2024-05-20
fix(autofix): Properly check for existing installation and code mappings

malwilley:
While running through this, I found an issue with how the setup check is looking at the integration, so made a couple changes:

1. Pass `project` in to the integration check so we can find code mappings for that project and error if not there
2. Checks for the integration's installation instead of the organization integration, which I think is more correct
3. Stop checking for whether it is disabled since Autofix only really needs the code mappings

### DIFF
--- a/src/sentry/api/endpoints/group_autofix_setup_check.py
+++ b/src/sentry/api/endpoints/group_autofix_setup_check.py
@@ -12,15 +12,13 @@ from sentry.api.api_owners import ApiOwner
 from sentry.api.api_publish_status import ApiPublishStatus
 from sentry.api.base import region_silo_endpoint
 from sentry.api.bases.group import GroupEndpoint
-from sentry.api.endpoints.event_ai_suggested_fix import get_openai_policy
 from sentry.api.helpers.autofix import (
     AutofixCodebaseIndexingStatus,
     get_project_codebase_indexing_status,
 )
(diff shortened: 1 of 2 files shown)


src/sentry/api/endpoints/group_autofix_setup_check.py
```diff
@@ -119,10 +114,6 @@ def get(self, request: Request, group: Group) -> Response:
 
         return Response(
             {
-                "subprocessorConsent": {
```
malwilley, line 122: I also went ahead and removed this part since we don't actually need it anymore, the `genAIConsent` flag will be off in this instance and the settings page will alert you that configuration is necessary

src/sentry/api/endpoints/group_autofix_setup_check.py
```diff
@@ -30,28 +28,31 @@
...
+    )
+    installation = integration and integration.get_installation(organization_id=organization.id)
 
-    if not integration:
+    if not installation:
         return "integration_missing"
 
-    if integration.status != ObjectStatus.ACTIVE:
-        return "integration_inactive"
+    code_mappings = get_sorted_code_mapping_configs(project)
```
jennmueng, line 53: Oh didn't know this util existed, should be using it for how the other autofix endpoints gets the code mappings too. I'll make a quick refactor

jennmueng (approved):

PR #71167, opened 2024-05-20
fix: allow JSON keys with non-string attributes

anonrig:
It seems we have a missing test coverage for this.

Ref https://github.com/getsentry/sentry/pull/71155#issuecomment-2120783149
Fixes https://sentry-st.sentry.io/issues/5380029738/?project=1513938
Fixes https://sentry.sentry.io/issues/5380113426/?project=1&referrer=github-pr-bot


mitsuhiko (approved):
This matches the older library’s behavior as per docs.

nhsiehgit (approved):

PR #71168, opened 2024-05-20
fix: Attempt to fix flakey backfill test

jangjodi:
Attempt to fix this [flakey test](url) by forcing the task to not call apply_async

### DIFF
--- a/src/sentry/tasks/backfill_seer_grouping_records.py
+++ b/src/sentry/tasks/backfill_seer_grouping_records.py
@@ -266,7 +266,7 @@ def backfill_seer_grouping_records(
                 ex=60 * 60 * 24 * 7,
             )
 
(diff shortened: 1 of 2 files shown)


src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -266,7 +266,7 @@ def backfill_seer_grouping_records(
                 ex=60 * 60 * 24 * 7,
             )
 
-            if last_processed_index <= len(group_id_message_data):
+            if last_processed_index < len(group_id_message_data):
```
jangjodi, line 269: additional small bug fix - this will call the backfill once more than it needs to be called

JoshFerge (approved):

PR #71169, opened 2024-05-20
feat(flagpole): Adds flagpole feature check to organization endpoint

GabeVillalobos:
Adds a temporary feature flag check to the organization serializer for us to start gathering flagpole evaluation metrics.

### DIFF
--- a/src/sentry/api/serializers/models/organization.py
+++ b/src/sentry/api/serializers/models/organization.py
@@ -278,10 +278,14 @@ class OrganizationSerializer(Serializer):
 
         with sentry_sdk.start_span(op="features.check", description="check individual features"):
             # Remaining features should not be checked via the entity handler
+
             for feature_name in org_features:
                 if features.has(feature_name, obj, actor=user, skip_entity=True):
                     # Remove the organization scope prefix
                     feature_list.add(feature_name[len(_ORGANIZATION_SCOPE_PREFIX) :])
+
+        if features.has("organizations:do_a_barrel_roll", obj, actor=user):
+            feature_list.add("do_a_barrel_roll")
 
         # Do not include the onboarding feature if OrganizationOptions exist
         if (


src/sentry/api/serializers/models/organization.py
```diff
@@ -278,11 +278,15 @@ def serialize(
...
         with sentry_sdk.start_span(op="features.check", description="check individual features"):
             # Remaining features should not be checked via the entity handler
+
             for feature_name in org_features:
                 if features.has(feature_name, obj, actor=user, skip_entity=True):
                     # Remove the organization scope prefix
                     feature_list.add(feature_name[len(_ORGANIZATION_SCOPE_PREFIX) :])
 
+        if features.has("organizations:do_a_barrel_roll", obj, actor=user):
+            feature_list.add("do_a_barrel_roll")
```
markstory, line 288: Has this feature been registered yet? I don't see it in `master`. Wouldn't this feature be checked on 256-259 as it is an OrganizationFeature that starts with `organizations`?

GabeVillalobos: Yep! It's been added to GetSentry: https://github.com/getsentry/getsentry/blob/ddadd775868b84c4bc930ef743111e068fca1c1e/getsentry/features.py#L2322

The single feature check above passes `skip_entity=True` to the handler check, which bypasses flagr/flagpole and pulls directly from `settings.SENTRY_FEATURES`:

https://github.com/getsentry/sentry/blob/ea0aa3db9f2be17c4a4d3bd1e8dab411d1c6d5d1/src/sentry/features/manager.py#L251

We do have a batch check before this that I overlooked, but using this would require us to filter the feature names bound for flagr/flagpole and check both. If we're ok with that, I can update the flagpole shim to do this instead.

markstory: In a shell locally

```python
from sentry import features
features.has("organizations:do_a_barrel_roll")
```

raises `FeatureNotRegistered`. Do we need a try/except here for self-hosted/local dev as those environments don't have `getsentry` available.

GabeVillalobos:
Closing this since this has been addressed within getsentry, and is now included as part of the bulk organization checks.

PR #71173, opened 2024-05-20
ref(rules): Temporarily don't remove from Redis

ceorourke:
As part of a continued debugging effort to find out why we aren't finding data in Redis in production (but it works locally), we'll temporarily not clean up the data from Redis so we have a longer period of time in which to attempt to find it.


saponifi3d (approved):
let's try out your other PR first - if that one doesn't help, we can try merging this?

PR #71178, opened 2024-05-20
Revert "feat(slack): Add flags to toggle features on organization integration"

ykamo001:
Reverts getsentry/sentry#70620


iamrajjoshi (approved):

ykamo001:
Since we want to follow existing conventions, the previous changes are unneeded, so reverting code. Revert should be safe as the changes were unused and at the code level

PR #71181, opened 2024-05-20
ref(rules): Import base buffer

ceorourke:
Use the base buffer instead of directly importing the `RedisBuffer` to hopefully fix the problem where this all works locally but in production we get an empty project ID list from Redis perhaps because we're writing to one cluster and reading from another.

### DIFF
--- a/src/sentry/buffer/__init__.py
+++ b/src/sentry/buffer/__init__.py
@@ -1,3 +1,5 @@
+from typing import TYPE_CHECKING
+
 from django.conf import settings
 
 from sentry.utils.services import LazyServiceWrapper
@@ -6,3 +8,18 @@ from .base import Buffer
 
 backend = LazyServiceWrapper(Buffer, settings.SENTRY_BUFFER, settings.SENTRY_BUFFER_OPTIONS)
 backend.expose(locals())
+
+if TYPE_CHECKING:
+    __buffer__ = Buffer()
+    get = __buffer__.get
+    incr = __buffer__.incr
+    process = __buffer__.process
+    process_pending = __buffer__.process_pending
+    process_batch = __buffer__.process_batch
+    validate = __buffer__.validate
+    push_to_sorted_set = __buffer__.push_to_sorted_set
+    push_to_hash = __buffer__.push_to_hash
+    get_sorted_set = __buffer__.get_sorted_set
+    get_hash = __buffer__.get_hash
+    delete_hash = __buffer__.delete_hash
+    delete_key = __buffer__.delete_key
--- a/src/sentry/buffer/__init__.py
+++ b/src/sentry/buffer/__init__.py
@@ -1,5 +1,3 @@
-from typing import TYPE_CHECKING
-
 from django.conf import settings
 
 from sentry.utils.services import LazyServiceWrapper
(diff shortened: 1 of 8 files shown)


src/sentry/rules/processing/delayed_processing.py
```diff
@@ -324,7 +324,6 @@ def apply_delayed(project_id: int, *args: Any, **kwargs: Any) -> None:
     """
     # STEP 1: Fetch the rulegroup_to_event_data mapping for the project from redis
     project = Project.objects.get_from_cache(id=project_id)
-    buffer = RedisBuffer()
```
ceorourke, line 327: In `process_delayed_alert_conditions` should I still be taking `buffer` in or just using the imported buffer?

ceorourke:
Might need to also update a couple tests in `test_post_process.py`

src/sentry/rules/processing/processor.py
```diff
@@ -10,8 +10,7 @@
 from django.core.cache import cache
 from django.utils import timezone
 
-from sentry import analytics, features
-from sentry.buffer.redis import RedisBuffer
```
saponifi3d, line 14: @wedamija is this just moving when we create the connection to redis to application boot time? if so, do we need to do anything additional to make sure we're maintaining a specific redis connection?

src/sentry/rules/processing/delayed_processing.py
saponifi3d: I believe we should be importing buffer here like you've updated rather than initializing a new buffer. it seems like the issue is that calling the constructor is creating a new redis connection which may be to a different instance.

saponifi3d (commented):
generally this makes sense to me as the fix

src/sentry/rules/processing/processor.py
wedamija: The problem is that usually when we create the buffer we do it here: https://github.com/getsentry/sentry/blob/94c896a4a3663abbd31775957f1aa5448fde5491/src/sentry/buffer/__init__.py#L7

When we do this, we also pass options along with it. In prod we pass these: https://github.com/getsentry/getsentry/blob/d7c44c232d49584b17e7ef747b77b145969d3ec6/getsentry/conf/settings/prod.py#L1084-L1085

```
SENTRY_BUFFER = "sentry.buffer.redis.RedisBuffer"
SENTRY_BUFFER_OPTIONS = {"cluster": "buffer"}
```

So it points us to a specific cluster. When we manually instantiate, we don't pass those options. In general, it's better to use `buffer` so that we respect our customer configuration

src/sentry/buffer/base.py
```diff
@@ -44,6 +57,29 @@ def get_hash(
     def get_sorted_set(self, key: str, min: float, max: float) -> list[tuple[int, datetime]]:
         return []
 
+    def push_to_sorted_set(self, key: str, value: list[int] | int) -> None:
+        return None
+
+    def push_to_hash(
```
wedamija, line 63: Btw I'm not sure if we use `RedisBuffer` in self hosted, it's worth checking. In dev we use this, so we should make sure that alerts and so on continue working when we're using `Buffer`. Maybe that just means keeping "organizations:process-slow-alerts" around permanently, and have it disabled in dev by default, but enabled elsewhere?

Alternatively, we could write this to just process in realtime when we enqueue.

wedamija (commented):
You might need to do something like https://github.com/getsentry/sentry/blob/a53126d538024575d9260c9caaf3c033429156a2/src/sentry/processing/realtime_metrics/__init__.py#L16-L27 to fix the typing issues

src/sentry/buffer/__init__.py
```diff
@@ -6,3 +8,18 @@
...
+    process = __buffer__.process
+    process_pending = __buffer__.process_pending
+    process_batch = __buffer__.process_batch
+    validate = __buffer__.validate
+    push_to_sorted_set = __buffer__.push_to_sorted_set
+    push_to_hash = __buffer__.push_to_hash
+    get_sorted_set = __buffer__.get_sorted_set
+    get_hash = __buffer__.get_hash
+    delete_hash = __buffer__.delete_hash
+    delete_key = __buffer__.delete_key
```
asottile-sentry, line 25: this shouldn't be done like this -- the code should use the `backend` instance and access attributes off of that

the eventual plan is to eliminate `expose(...)` entirely

asottile-sentry (dismissed):

src/sentry/buffer/__init__.py
wedamija: Ahh, I wasn't aware of this. So the plan going forward is to use `sentry.buffer.backend.<attrs>`? I'm ok with that, but it might be worth letting people know

asottile-sentry: 99% of the usages have been moved over and the type checker tends to enforce it in new modules already

wedamija (approved):
lgtm, assuming all the typing passes

PR #71182, opened 2024-05-20
fix: use `orjson.OPT_UTC_Z` for timestamps

anonrig:
For timestamps, we should use `OPT_UTC_Z`


anonrig:
Closing in favor of https://github.com/getsentry/sentry/pull/71185

PR #71187, opened 2024-05-20
fix: use date parser to parse timestamps

anonrig:
Recently merged https://github.com/getsentry/sentry/commit/68f2dfc6b73efc374b293c84137d11db8d314d62 can be improved by using the `dateutil.parser`

### DIFF
--- a/src/sentry/utils/dates.py
+++ b/src/sentry/utils/dates.py
@@ -1,20 +1,24 @@
+import logging
 import re
 import zoneinfo
 from collections.abc import Mapping
 from datetime import UTC, date, datetime, timedelta
 from typing import Any, overload
 
-from dateutil.parser import parse
+from dateutil.parser import ParserError, parse
 from django.http.request import HttpRequest
 from django.utils.timezone import is_aware, make_aware
 
-from sentry import quotas
+from sentry import options, quotas
 from sentry.constants import MAX_ROLLUP_POINTS
+from sentry.utils import metrics
 
 epoch = datetime(1970, 1, 1, tzinfo=UTC)
 
 # Factory is an obscure GMT alias
 AVAILABLE_TIMEZONES = frozenset(zoneinfo.available_timezones() - {"Factory"})
+
+logger = logging.getLogger("sentry.utils.dates")
 
 
 def ensure_aware(value: datetime) -> datetime:
@@ -76,12 +80,7 @@ def parse_date(datestr: str, timestr: str) -> datetime | None:
             return None
 
 
-def parse_timestamp(value: Any) -> datetime | None:
-    # TODO(mitsuhiko): merge this code with coreapis date parser
-    if isinstance(value, datetime):
-        return value
-    elif isinstance(value, (int, float)):
-        return datetime.fromtimestamp(value, UTC)
+def _parse_timestamp_old(value: Any) -> datetime | None:
     value = (value or "").rstrip("Z").encode("ascii", "replace").split(b".", 1)
     if not value:
         return None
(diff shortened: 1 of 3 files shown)


src/sentry/utils/dates.py
```diff
@@ -76,31 +76,27 @@ def parse_date(datestr: str, timestr: str) -> datetime | None:
...
-        except Exception:
-            return None
-    except Exception:
-        return None
-    if len(value) == 2:
-        try:
-            rv = rv.replace(microsecond=int(value[1].ljust(6, b"0")[:6]))
-        except ValueError:
+            rv = isoparse(value)
+        except (ParserError, ValueError):
```
wedamija, line 97: I don't know if this is really safe or if we're relying on this in some way. How about we do something like
```
def parse_timestamp(value: datetime | int | float | str | bytes | None) -> datetime | None:
    ... # existing logic
    rv = _parse_timestamp_old(value)
    if options.get("some_flag"):
        new_value = parse(value, ignoretz=True)
        match = "equal"
        if new_value != rv:
            # Maybe log?
            match = "not_equal"
        metrics.incr("parse_date", tags={"result": match})
        rv = value
```

anonrig: Makes sense. I've updated the implementation. Thank you for the review Dan

Swatinem (approved):

loewenheim (approved):

src/sentry/utils/dates.py
```diff
@@ -104,6 +100,33 @@ def parse_timestamp(value: Any) -> datetime | None:
...
+
+    # TODO(@anonrig): Remove this once we make sure both results return the same.
+    if options.get("system.use-date-util-timestamps"):
+        if isinstance(value, bytes):
+            value = value.decode()
+
+        try:
+            new_value = parse(value, ignoretz=True).replace(tzinfo=UTC)
+        except (ParserError, ValueError):
+            new_value = None
```
wedamija, line 122: Log the exception here?

src/sentry/utils/dates.py
```diff
@@ -104,6 +100,33 @@ def parse_timestamp(value: Any) -> datetime | None:
...
+        return datetime.fromtimestamp(value, UTC)
+
+    expected_value = _parse_timestamp_old(value)
+
+    # TODO(@anonrig): Remove this once we make sure both results return the same.
+    if options.get("system.use-date-util-timestamps"):
+        if isinstance(value, bytes):
+            value = value.decode()
+
+        try:
```
wedamija, line 119: Just to be extra paranoid, let's wrap the entire `if` in the try/except?

src/sentry/utils/dates.py
```diff
@@ -104,6 +100,33 @@ def parse_timestamp(value: Any) -> datetime | None:
...
+            new_value = parse(value, ignoretz=True).replace(tzinfo=UTC)
+        except (ParserError, ValueError):
+            new_value = None
+
+        result = "not-equal" if expected_value != new_value else "equal"
```
wedamija, line 124: It could be worth logging the two values to `logging.info` when not equal - but up to you

wedamija (approved):

src/sentry/utils/dates.py
```diff
@@ -104,6 +103,37 @@ def parse_timestamp(value: Any) -> datetime | None:
...
+    expected_value = _parse_timestamp_old(value)
+
+    # TODO(@anonrig): Remove this once we make sure both results return the same.
+    if options.get("system.use-date-util-timestamps"):
+        try:
+            if isinstance(value, bytes):
+                value = value.decode()
+            new_value = parse(value, ignoretz=True).replace(tzinfo=UTC)
+        except (ParserError, ValueError) as err:
+            logger.exception("parse_timestamp", extra={"error": str(err)})
```
wedamija, line 124: Nit: `exception` will already capture all the exception details, so no need for the extra here

wedamija (approved):

src/sentry/utils/dates.py
anonrig: nice

PR #71190, opened 2024-05-20
chore(apis): Update ownership from issues to alerts & notifications

sentaur-athena:
After re-orgs some endpoints are better owned by alerts & notifications vs. issues. I'm not sure if the ownership I'm assigning in this PR is correct so will get help from both teams.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -174,7 +174,19 @@ yarn.lock                                                @getsentry/owners-js-de
 /tests/snuba/incidents/                                   @getsentry/alerts-notifications
 /tests/sentry/snuba/test_query_subscription_consumer.py   @getsentry/alerts-notifications
 /tests/sentry/snuba/test_subscriptions.py                 @getsentry/alerts-notifications
+<<<<<<< HEAD
 /tests/sentry/snuba/test_tasks.py                         @getsentry/alerts-notifications
+=======
+/src/sentry/api/endpoints/notification_defaults.py                          @getsentry/alerts-notifications
+/src/sentry/api/endpoints/broadcast_details.py                              @getsentry/alerts-notifications
+/src/sentry/api/endpoints/broadcast_index.py                                @getsentry/alerts-notifications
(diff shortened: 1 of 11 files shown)


.github/CODEOWNERS
```diff
@@ -174,6 +174,15 @@ yarn.lock                                                @getsentry/owners-js-de
 /tests/snuba/incidents/                                   @getsentry/alerts-notifications
 /tests/sentry/snuba/test_query_subscription_consumer.py   @getsentry/alerts-notifications
 /tests/sentry/snuba/test_subscriptions.py                 @getsentry/alerts-notifications
+/src/sentry/api/endpoints/notification_defaults.py                          @getsentry/alerts-notifications
+/src/sentry/api/endpoints/broadcast_details.py                              @getsentry/alerts-notifications
+/src/sentry/api/endpoints/broadcast_index.py                                @getsentry/alerts-notifications
```
saponifi3d, line 179: sounds like we don't own either of these

src/sentry/api/endpoints/user_subscriptions.py
```diff
@@ -26,7 +26,7 @@ class NewsletterValidator(serializers.Serializer):
 
 @control_silo_endpoint
 class UserSubscriptionsEndpoint(UserEndpoint):
-    owner = ApiOwner.ISSUES
+    owner = ApiOwner.ALERTS_NOTIFICATIONS
```
saponifi3d, line 29: not sure if this should be us or billing? i'm a little overloaded on the term `subscriptions` 😅

mrduncan (approved):
Issues specific changes LGTM - thank you for the cleanup!

vartec (approved):
TY!

src/sentry/api/endpoints/user_subscriptions.py
sentaur-athena: yeah no, it's the emails people get. It's more like marketing but there's no eng team supporting it. I'll add it to unowned.

PR #71145, opened 2024-05-17
feat(org-stats): Show spans + profile hours, hide txns

scttcper:
Shows spans and profile hours on the org stats page and hide transactions

PR #71194, opened 2024-05-20
feat(billing): Add `spans-usage-tracking` org feature

scttcper:
This flag will be attached to the am3 billing plan since they'll measure only spans and not transactions.

part of https://getsentry.atlassian.net/browse/RV-1648
blocks https://github.com/getsentry/sentry/pull/71145

### DIFF
--- a/src/sentry/conf/server.py
+++ b/src/sentry/conf/server.py
@@ -1877,8 +1877,8 @@ SENTRY_FEATURES: dict[str, bool | None] = {
     "organizations:standalone-span-ingestion": False,
     # A single flag for all the new performance UI that relies on span ingestion
     "organizations:spans-first-ui": False,
-    # Hide transactions and display spans
-    "organizations:spans-only": False,
+    # Measure usage by spans instead of transactions
+    "organizations:spans-usage-tracking": False,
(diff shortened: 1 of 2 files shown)


KevinL10 (approved):

src/sentry/conf/server.py
```diff
@@ -1877,6 +1877,8 @@ def custom_parameter_sort(parameter: dict) -> tuple[str, int]:
     "organizations:standalone-span-ingestion": False,
     # A single flag for all the new performance UI that relies on span ingestion
     "organizations:spans-first-ui": False,
+    # Hide transactions and display spans
```
gggritso, line 1880: In checkout flows, or _everywhere_? If it's the former, could you make the flag more specific? Otherwise it'll make a lot of upcoming Performance flags more confusing

scttcper: this is currently going to be used on the org stats page, checkout flows have access to the subscription data categories. but i could see it being used anywhere that we need to know if the org has access to spans (outside of billing)

scttcper: maybe spans-usage-tracking

mbauer404 (approved):

PR #71197, opened 2024-05-20
fix(highlights): Set column count after ref defined

scttcper:
Fixes the number of columns from changing when going between the issues stream and an issue.

can kinda see it going from 1 to 2 columns when the resize observer callback is finally hit

https://github.com/getsentry/sentry/assets/1400464/fc022952-9407-427f-a493-528684bdcbbf


leeandher (approved):
Thanks @scttcper!

PR #71199, opened 2024-05-20
fix(issue-stream): Show event sub status before event chart, fix placeholders

MichaelSun48:
This PR makes a couple follow up improvements to the new `<EventStatsChart/>` component, as well as the Issue stream in which it appears. These improvements include:

1. **Reduces the height of the "Last Seen" placeholder.** Previously, the height of this placeholder was larger than the value itself. This caused all rows in the issue stream to suddenly contract upon initial load. Reducing the height of the placeholder solves this problem. 
2. **Reduces the width of the "Count" and "Users" placeholders.** The extra width made it seem like the number "jumped" into its position, which looked choppy. Now that the placeholder's width is closer to the width of the value itself, the "jump" effect seems reduced (to my eye at least) 
3. **Renders the group sub-status (escalating, ongoing, etc.) before the chart.** Previously, the group sub status wouldn't be loaded until the graph itself was loaded. 

At a glance:
image

### DIFF
--- a/static/app/components/charts/groupStatusChart.tsx
+++ b/static/app/components/charts/groupStatusChart.tsx
@@ -4,6 +4,7 @@ import styled from '@emotion/styled';
 import MarkLine from 'sentry/components/charts/components/markLine';
 import MiniBarChart from 'sentry/components/charts/miniBarChart';
 import {LazyRender} from 'sentry/components/lazyRender';
+import Placeholder from 'sentry/components/placeholder';
 import {t} from 'sentry/locale';
 import type {TimeseriesValue} from 'sentry/types';
 import type {Series} from 'sentry/types/echarts';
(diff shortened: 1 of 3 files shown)


static/app/components/charts/groupStatusChart.tsx
```diff
@@ -104,21 +107,25 @@ function GroupStatusChart({
...
-          isGroupedByDate
-          showTimeInTooltip
-          series={graphOptions.series}
-          colors={graphOptions.colors}
-          emphasisColors={graphOptions.emphasisColors}
-          hideDelay={50}
-          showMarkLineLabel={showMarkLine}
-        />
+        {loading ? (
+          <Placeholder height={`36px`} />
```
malwilley, line 111: nit: backticks not needed, can do height="36px"

malwilley (approved):
Looks really nice!

PR #71201, opened 2024-05-20
feat(onboarding): generic banner for unsupported features

sentaur-athena:
Adding a generic banner for when a feature is not available for a platform. Replay already has one which is where I got the design from.

The code most probably change so feel free to review but I'm still working on the profiling test. Need to mock the filter.

Screenshot 2024-05-17 at 3 49 13 PM
Screenshot 2024-05-17 at 11 15 25 AM

### DIFF
--- a/static/app/views/performance/onboarding.tsx
+++ b/static/app/views/performance/onboarding.tsx
@@ -1,4 +1,4 @@
-import {useEffect} from 'react';
+import {Fragment, useEffect} from 'react';
 import styled from '@emotion/styled';
 import * as Sentry from '@sentry/react';
 
@@ -13,6 +13,7 @@ import {
   addLoadingMessage,
   clearIndicators,
 } from 'sentry/actionCreators/indicator';
+import UnsupportedAlert from 'sentry/components/alerts/unsupportedAlert';
 import {Button} from 'sentry/components/button';
 import ButtonBar from 'sentry/components/buttonBar';
 import type {TourStep} from 'sentry/components/modals/featureTourModal';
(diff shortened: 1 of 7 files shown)


static/app/views/profiling/content.tsx
```diff
@@ -197,40 +221,45 @@ function ProfilingContent({location}: ProfilingContentProps) {
...
               {shouldShowProfilingOnboardingPanel ? (
-                // If user is on m2, show default
-                <ProfilingOnboardingPanel
-                  content={
-                    <ProfilingAM1OrMMXUpgrade
+                <React.Fragment>
+                  {withoutProfilingSupport && (
+                    <UnsupportedAlert featureName="Profiling" />
+                  )}
+                  // If user is on m2, show default
```
sentaur-athena, line 228: Nothing actually changed after this line. Just the tabs show everything new 🙄

static/app/views/performance/onboarding.tsx
```diff
@@ -167,76 +173,81 @@ function Onboarding({organization, project}: Props) {
   }
 
   return (
-    <OnboardingPanel image={<PerfImage src={emptyStateImg} />}>
```
sentaur-athena, line 170: Nothing here is changed except 176-178, the whole thing looks changed here just because of an extra tab 😢

static/app/views/profiling/content.tsx
```diff
@@ -137,6 +139,28 @@ function ProfilingContent({location}: ProfilingContentProps) {
     );
   }, [selection.projects, projects]);
 
+  const withoutProfilingSupport = useMemo((): boolean => {
+    const projectsWithProfilingSupport = new Set(
+      projects
+        .filter(project => !project.platform || profiling.includes(project.platform))
+        .map(project => project.id)
+    );
```
scttcper, line 147: i would maybe put all this inside UnsupportedAlert? Doesn't seem useful at the top level, UnsupportedAlert can choose when to display

scttcper: or maybe not UnsupportedAlert but another small component

scttcper (approved):

PR #71203, opened 2024-05-20
ref: Put animation, pulsing, and text styles into the theme object

ryan953:
Moving animation and text helpers into the theme object, in an effort to make them more discoverable and reusable. 

Also, this will save an import everywhere, which some people care about. 

Relates to https://github.com/getsentry/frontend-tsc/issues/27

See also https://github.com/getsentry/getsentry/pull/14036
See also https://github.com/getsentry/sentry/pull/71208


static/app/styles/text.tsx
```diff
@@ -1,5 +1,10 @@
 import {css} from '@emotion/react';
 
+/**
+ * Use `p.theme.textStyles` instead of importing this.
+ *
+ * @deprecated
+ */
```
ryan953, line 7: This is leftover because there's a callsite in getsentry to remove first.

ryan953: See https://github.com/getsentry/getsentry/pull/14036

static/app/utils/theme.tsx
```diff
@@ -648,6 +649,64 @@ const generateUtils = (colors: BaseColors, aliases: Aliases) => ({
...
+    pre,
+    .auto-select-text,
+    .section,
+    [class^='highlight-'] {
+      margin-bottom: 20px;
+
+      &:last-child {
+        margin-bottom: 0;
+      }
+    }
```
scttcper, line 708: why are people using this? should this just be global? like what is .auto-select-text and why would people want it on their h4

ryan953: This is probably good for markdown generated sections, where we couldn't use an `const H4 = styled('h4')`. Not sure about the callsites.

ryan953:
I'm going to go the other way here... what if we had more things in /styles? 
- Fewer `export Foo = styled('div')...` 
- Keep things out of theme if they don't change in light/dark mode... putting them in there hard-codes them into the bundle.

PR #71238, opened 2024-05-21
chore(ui): Update Codecov bundler plugin

nicholas-codecov:
This PR updates the Codecov bundler plugin version to the latest version `beta-8`, in this latest version we've enhanced the debugging of commit sha's so we can better investigate a suspect issue with merge commits in GH and how we're determining what sha to choose.

Also I've modified the build env vars slightly to grab the head commit sha from the pull request event payload, and pass along through to the webpack config as an upload override to Codecov.

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -137,6 +137,7 @@ jobs:
         if: ${{ always() && needs.files-changed.outputs.backend_all == 'true' }}
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+          commit_sha: ${{ github.event.pull_request.head.sha }}
 
   acceptance-required-checks:
     # this is a required check so we need this job to always run and report a status.
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -65,6 +65,8 @@ jobs:
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         name: Checkout sentry
+        with:
+          fetch-depth: 2
 
       - uses: getsentry/action-setup-volta@e4939d337b83760d13a9d7030a6f68c9d0ee7581 # v2.0.0
 
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -65,8 +65,6 @@ jobs:
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         name: Checkout sentry
-        with:
-          fetch-depth: 2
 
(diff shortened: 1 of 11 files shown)


JonasBa (approved):

.github/workflows/acceptance.yml
```diff
@@ -65,6 +65,8 @@ jobs:
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         name: Checkout sentry
+        with:
+          fetch-depth: 2
```
asottile-sentry, line 69: it's very unclear why this is necessary -- it's been cargo culted since the beginning of time in this repo and I see it copy pasted without explanation (the original commit doesn't really give a hint) -- hoping someone from codecov can actually describe why this is necessary since there's definitely better ways to find the merge commit from the github context

nicholas-codecov: Sorry, I haven't had a chance to update the PR comment explaining this small tweak. We're trying to debug as to why we can't find the merge commit as we have done in previous iterations, this is one thing that stood out to us as what could be part of the problem. I'm all ears to any suggestions as to finding the correct SHA from a merge commit as the `GITHUB_SHA` env var does not point to the correct commit SHA, if you have any suggestions or documentation you can point me to let me know.

nicholas-codecov: Small update, adding this in does allow us to actually grab the correct SHA from the CI workflow. However if you do have any optimizations around grabbing this from another area I'm more than happy to look into it

```
[codecov] Merge commit message: fea4348846c074f9510a4703617dfdf9461233f6 1ac632d4c9e095e3ae8317eb35263aaf4fe4b193
[codecov] Split commit message: fea4348846c074f9510a4703617dfdf9461233f6,1ac632d4c9e095e3ae8317eb35263aaf4fe4b193
[codecov] Using commit: 1ac632d4c9e095e3ae8317eb35263aaf4fe4b193
```

asottile-sentry: there should be these values that you can reference from the event:
- `pull_request.base.sha`
- `pull_request.sha`
- `pull_request.merge_commit_sha`

that should give the full picture without needing to request it from `git` and slowing down the clone operation

nicholas-codecov: thanks for pointing that out, I've updated the changes in this PR to reflect that, let me see if I did it right and we get the correct value.

asottile-sentry: sg2m

asottile-sentry: can you fix the rest of `sentry`? this `depth: 2` is in a bunch of other places as well

asottile-sentry (approved):
lovely, thanks

PR #71247, opened 2024-05-21
feat(insights): support cache.get op in cache module

DominikB2014:
SDKs recently started sending `cache.get` span ops

### DIFF
--- a/static/app/views/performance/cache/settings.ts
+++ b/static/app/views/performance/cache/settings.ts
@@ -19,7 +19,7 @@ export const CHART_HEIGHT = 160;
 export const CACHE_BASE_URL = `/performance/${BASE_URL}`;
 
 export const BASE_FILTERS: SpanMetricsQueryFilters = {
-  'span.op': 'cache.get_item', //  TODO - add more span ops as they become available, we can't use span.module because db.redis is also `cache`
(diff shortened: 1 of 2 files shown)


static/app/views/performance/cache/settings.ts
```diff
@@ -19,7 +19,7 @@ export const CHART_HEIGHT = 160;
 export const CACHE_BASE_URL = `/performance/${BASE_URL}`;
 
 export const BASE_FILTERS: SpanMetricsQueryFilters = {
-  'span.op': 'cache.get_item', //  TODO - add more span ops as they become available, we can't use span.module because db.redis is also `cache`
+  'span.op': '[cache.get_item,cache.get]', //  TODO - add more span ops as they become available, we can't use span.module because db.redis is also `cache`
```
gggritso, line 22: 😭 more motivation to work on a `MutableQuery` alternative that handles array conditions properly. Can you add a second `TODO` about how it's awkward to have to construct an array string here, and why that is?

gggritso (approved):

static/app/views/performance/cache/settings.ts
DominikB2014: 👍 good idea, i'll add another todo.

PR #71250, opened 2024-05-21
fix(insights): fix infinite spinner cache module landing

DominikB2014:
The cache landing page table would contain an infinite spinner if no data is available.
This is because the table requires two requests to populate, if the first request has no data, the second requests would still be in the `loading` state.

~~Also, if anybody remembers a more elegant way to do this, lmk! I feel like this is came up a few times~~

### DIFF
--- a/static/app/views/performance/cache/cacheLandingPage.tsx
+++ b/static/app/views/performance/cache/cacheLandingPage.tsx
@@ -103,6 +103,9 @@ export function CacheLandingPage() {
     Referrer.LANDING_CACHE_TRANSACTION_LIST
   );
 
+  const enableTransactionDuration =
+    !isTransactionsListLoading && transactionsList.length > 0;
+
   const {
     data: transactionDurationData,
     error: transactionDurationError,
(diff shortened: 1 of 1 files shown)


static/app/views/performance/cache/cacheLandingPage.tsx
```diff
@@ -194,7 +197,10 @@ export function CacheLandingPage() {
               <ModuleLayout.Full>
                 <TransactionsTable
                   data={transactionsListWithDuration}
-                  isLoading={isTransactionsListLoading || isTransactionDurationLoading}
+                  isLoading={
+                    isTransactionsListLoading ||
+                    (enableTransactionDuration && isTransactionDurationLoading)
```
gggritso, line 202: I usually do this by using `isFetching` instead of `isLoading`, which works well in this case (though it's also awkward)

```jsx

const { data, isFetching } = useBlahBlah
```

gggritso (commented):
Try `isFetching`, that should work okay!

static/app/views/performance/cache/cacheLandingPage.tsx
DominikB2014: Yup, is fetching is the way to go! thanks!!!

gggritso (approved):

PR #71251, opened 2024-05-21
chore(hybridcloud) Add non-tuple return method for group subscriptions

markstory:
Deprecate an RPC method that has a tuple return and add a new method that returns objects instead.

Refs HC-1190

### DIFF
--- a/src/sentry/services/hybrid_cloud/notifications/impl.py
+++ b/src/sentry/services/hybrid_cloud/notifications/impl.py
@@ -14,6 +14,7 @@ from sentry.notifications.types import (
     NotificationSettingsOptionEnum,
 )
 from sentry.services.hybrid_cloud.notifications import NotificationsService
+from sentry.services.hybrid_cloud.notifications.model import RpcSubscriptionStatus
 from sentry.services.hybrid_cloud.user.service import user_service
 from sentry.types.actor import Actor, ActorType
 from sentry.types.integrations import EXTERNAL_PROVIDERS, ExternalProviderEnum, ExternalProviders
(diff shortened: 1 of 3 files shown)


src/sentry/services/hybrid_cloud/notifications/impl.py
```diff
@@ -120,8 +121,23 @@ def get_subscriptions_for_projects(
...
+            for project_id, sub in result.items()
+        }
+
+    def subscriptions_for_projects(
+        self,
+        *,
+        user_id: int,
+        project_ids: list[int],
+        type: NotificationSettingEnum,
+    ) -> Mapping[int, RpcSubscriptionStatus]:
```
GabeVillalobos, line 140: Something I've been wondering lately: would it be helpful to make these more complex return types into exportable types in the `service.py` file that other code can import just to cut down on the type repetition?

GabeVillalobos, line 140: Should we add some tests for this new RPC method?

GabeVillalobos (approved):

src/sentry/services/hybrid_cloud/notifications/impl.py
markstory: I didn't add any explicit tests as this method is covered by tests for the group serializer, and the old method is calling the new one. Any new tests wouldn't increase coverage.

markstory: > would it be helpful to make these more complex return types into exportable types in the service.py file that other code can import just to cut down on the type repetition?

I think it would be helpful to reduce noise and boilerplate types.

PR #71256, opened 2024-05-21
fix(perf): Adjust transaction name search period

k-fish:
### Summary
This will set a minimum stats period so we don't have cases where you can find an intermittent transaction at low values (eg. 1h). This should eventually be replaced by a materialized view for unique transaction names across all times for faster lookup / completedness, but this is just a quick fix.

### DIFF
--- a/static/app/components/performance/searchBar.tsx
+++ b/static/app/components/performance/searchBar.tsx
@@ -9,6 +9,7 @@ import {t} from 'sentry/locale';
 import type {Organization} from 'sentry/types/organization';
 import {trackAnalytics} from 'sentry/utils/analytics';
 import {browserHistory} from 'sentry/utils/browserHistory';
+import {parsePeriodToHours} from 'sentry/utils/dates';
 import type EventView from 'sentry/utils/discover/eventView';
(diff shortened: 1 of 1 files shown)


static/app/components/performance/searchBar.tsx
```diff
@@ -148,6 +151,15 @@ function SearchBar(props: SearchBarProps) {
...
           }
+          const parsedPeriodHours = eventView.statsPeriod
+            ? parsePeriodToHours(eventView.statsPeriod)
+            : 0;
+          const parsedDefaultHours = parsePeriodToHours(TRANSACTION_SEARCH_PERIOD);
+
+          const statsPeriod =
+            parsedDefaultHours > parsedPeriodHours
+              ? TRANSACTION_SEARCH_PERIOD
+              : eventView.statsPeriod;
```
gggritso, line 162: This wishes it was `Math.max`!

gggritso (approved):

static/app/components/performance/searchBar.tsx
k-fish: I'm not using the hours I'm using the string, otherwise, yeah

PR #71257, opened 2024-05-21
chore(hybridcloud) Add non-tuple returns to notification methods

markstory:
Add new RPC methods that don't return tuples. Methods with tuple returns are not compatible with our RPC schema validation and need to be phased out.

Refs HC-1190

### DIFF
--- a/src/sentry/services/hybrid_cloud/integration/impl.py
+++ b/src/sentry/services/hybrid_cloud/integration/impl.py
@@ -4,6 +4,7 @@ import logging
 from collections.abc import Iterable
 from typing import TYPE_CHECKING, Any
 
+import sentry_sdk
 from django.utils import timezone
 
 from sentry import analytics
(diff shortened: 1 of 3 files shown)


src/sentry/services/hybrid_cloud/integration/impl.py
```diff
@@ -210,6 +212,27 @@ def get_organization_integrations(
...
+    ) -> RpcOrganizationContext:
+        context = self.organization_contexts(
+            organization_id=organization_id,
+            integration_id=integration_id,
+            provider=provider,
+            external_id=external_id,
+        )
+        install = None
+        if context.organization_integrations:
+            install = context.organization_integrations[0]
```
GabeVillalobos, line 231: Do we need any validation here that we only retrieve a single organization integration?

markstory: We could return the install only if the organizationid matches.

GabeVillalobos (approved):

PR #71258, opened 2024-05-21
feat(profiling): enable profiling measurement charts if measurements are sent

### LINKED ISSUE #59989: Show UI frames, CPU and Memory for RN/iOS Mixed Profiles
Since https://github.com/getsentry/sentry-react-native/pull/3349 RN SDK sends profiles with UI frames, CPU and Memory.

https://sentry-sdks.sentry.io/profiling/profile/sentry-react-native/4f0be51bd1534274b5b0fc1874520fd8/flamegraph/

Screenshot 2023-11-15 at 16 10 24

The current UI doesn't display the measurements. 

Note not all RN profiles include the measurements.

JonasBa:
React native uses javascript as the platform value which is used both by browser sdk which doesnt report cpu and memory measures. This means we cant know ahead of time if we expect the measurements and simply enable the charts (this toggles loading and the various no measurement states). 

In order to not enable these charts for browser SDKs and confuse users, we take a different approach and enable the charts on demand, once the data has loaded. We miss out on some of the loading states, but this all happens pretty fast as the profile loads, so I dont think it's a big deal in terms of UX.

Fixes https://github.com/getsentry/sentry/issues/59989

### DIFF
--- a/static/app/components/profiling/flamegraph/flamegraph.tsx
+++ b/static/app/components/profiling/flamegraph/flamegraph.tsx
@@ -240,18 +240,28 @@ function Flamegraph(): ReactElement {
 
   const hasBatteryChart = useMemo(() => {
     const platform = profileGroup.metadata.platform;
-    return platform === 'cocoa';
-  }, [profileGroup.metadata.platform]);
+    return platform === 'cocoa' || profileGroup.measurements?.cpu_energy_usage;
+  }, [profileGroup.metadata.platform, profileGroup.measurements]);
 
   const hasCPUChart = useMemo(() => {
     const platform = profileGroup.metadata.platform;
-    return platform === 'cocoa' || platform === 'android' || platform === 'node';
-  }, [profileGroup.metadata.platform]);
+    return (
+      platform === 'cocoa' ||
+      platform === 'android' ||
+      platform === 'node' ||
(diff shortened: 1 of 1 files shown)


viglia (commented):
@JonasBa in `vroom` in order to differentiate between `browser` _js_ and `react-native` _js_  I check whether the _runtime_ is set to `hermes`.

I think you could make that check as well so that you don't need a different approach and we won't miss out on some of the loading states as you mentioned.

Snippet here below is an extract from the payload returned to the front-end:

```json
    "runtime": {
        "name": "hermes",
        "version": ""
    },
```

JonasBa:
> @JonasBa in `vroom` in order to differentiate between `browser` _js_ and `react-native` _js_ I check whether the _runtime_ is set to `hermes`.
> 
> I think you could make that check as well so that you don't need a different approach and we won't miss out on some of the loading states as you mentioned.
> 
> Snippet here below is an extract from the payload returned to the front-end:
> 
> ```json
>     "runtime": {
>         "name": "hermes",
>         "version": ""
>     },
> ```

The issue is that we need to know this ahead of time in order to show the skeleton and loading UI. If it comes in with the profile response, then it's already too late to show

Zylphrex (approved):

Issue #70942, opened 2024-05-15
[Epic] Remote Configuration - Internal Alpha


cmanallen:
```[tasklist]
### Web UI
- [x] [Add API blueprint](https://github.com/getsentry/sentry/pull/70805/files)
- [x] [Add API endpoints](https://github.com/getsentry/sentry/pull/70799)
- [x] Add Web UI
```

```[tasklist]
### SDK
- [x] [Add protocol blueprint](https://github.com/getsentry/sentry/pull/70805/files)
- [x] [Add Javascript SDK support](https://github.com/getsentry/sentry-javascript/pull/12090)
- [x] ~~[Add Python SDK support](https://github.com/getsentry/sentry-python/pull/3041)~~ (Moved to post alpha)
- [x] Add configuration proxy endpoint
```

```[tasklist]
### Documentation
- [x] [Explore Solution Space](https://www.notion.so/sentry/Remote-Config-Tech-braindump-0567916007bd453daa0e35fb95bd6761)
- [x] [Write Delivery Mechanism DACI and Receive Approval](https://www.notion.so/sentry/DACI-Remote-Configuration-Delivery-Mechanism-530e40d7ecc943f4a15c69d427760f9c)
- [x] Add internal infrastructure documentation
```

PR #71259, opened 2024-05-21
feat(remote-config): Add API endpoints

cmanallen:
Adds API endpoints for managing SDK and user configuration.

Depends on: https://github.com/getsentry/sentry/pull/70805


cmanallen:
Can someone explain to me how this change increases the bundle size by 12 bytes?

ryan953 (approved):

src/sentry/remote_config/storage.py
```diff
@@ -0,0 +1,115 @@
...
+            self.storage.delete(self.key)
+        except Exception:
+            return None
+
+    def _make_storage_options(self) -> dict | None:
+        backend = options.get("configurations.storage.backend")
+        if backend:
+            return {
+                "backend": backend,
+                "options": options.get("configurations.storage.options"),
```
aliu3ntry, line 108: Does get_storage handle the case where options=None?

src/sentry/remote_config/storage.py
```diff
@@ -0,0 +1,115 @@
...
+    def set(self, value: StorageFormat) -> None:
+        self.storage.save(self.key, BytesIO(json.dumps(value).encode()))
+
+    def pop(self) -> None:
+        try:
+            self.storage.delete(self.key)
+        except Exception:
+            return None
+
+    def _make_storage_options(self) -> dict | None:
```
aliu3ntry, line 103:
```suggestion
    def _make_storage_config(self) -> dict | None:
```
To match the name of the parameter for `get_storage`

src/sentry/remote_config/endpoints.py
```diff
@@ -0,0 +1,73 @@
...
+            key = ProjectKey.objects.for_request(request).get(project=project, public_key=key_id)
+        except ProjectKey.DoesNotExist:
+            raise ResourceDoesNotExist
+
+        kwargs["key"] = key
+        return args, kwargs
+
+    def get(self, request: Request, project: Project, key: ProjectKey) -> Response:
+        """Get remote configuration from project options."""
+        remote_config = make_storage(key).get()
```
aliu3ntry, line 51: Is it worth caching the StorageBackends with @lru_cache or something?

src/sentry/remote_config/endpoints.py
```diff
@@ -0,0 +1,73 @@
...
+            return Response("Not found.", status=404)
+        return Response({"data": remote_config}, status=200)
+
+    def post(self, request: Request, project: Project, key: ProjectKey) -> Response:
+        """Set remote configuration in project options."""
+        validator = ConfigurationContainerValidator(data=request.data)
+        if not validator.is_valid():
+            return self.respond(validator.errors, status=400)
+
+        result = validator.validated_data["data"]
```
aliu3ntry, line 62: How is validated_data different from request.data? If it's the same, do we need to return the non-id fields?

src/sentry/remote_config/endpoints.py
```diff
@@ -0,0 +1,73 @@
...
+        "POST": ApiPublishStatus.EXPERIMENTAL,
+        "DELETE": ApiPublishStatus.EXPERIMENTAL,
+    }
+
+    def convert_args(self, request: Request, key_id: str, *args, **kwargs):
+        args, kwargs = super().convert_args(request, *args, **kwargs)
+        project = kwargs["project"]
+
+        try:
+            key = ProjectKey.objects.for_request(request).get(project=project, public_key=key_id)
```
aliu3ntry, line 42: If key_id = project's public key, where's the name `key_id` come from?

src/sentry/remote_config/storage.py
cmanallen: Yes.  None is a valid value for the options.

src/sentry/remote_config/endpoints.py
cmanallen: Nah, we're only initializing a class.  The memory use from caching would be more damaging than the CPU cost of creating a new instance.

src/sentry/remote_config/endpoints.py
cmanallen: Conventional behavior taken from other endpoints which share the route.  Public key could be more contextual but I'm not too worried about it.  Open to change if you feel its important.

src/sentry/remote_config/endpoints.py
aliu3ntry: Cool cool, `get_storage` in the driver is the same thing right, just an import?

aliu3ntry: Why do we have the `make_storage` function instead of directly using the `StorageBackend` constructor?

cmanallen: Its indirection solely for the sake of abstracting the storage definition.  I had another storage driver previously and may add another one in the future.  This makes it easier for me to change things internally without worrying about the implementation in the test suite and in the endpoints.

It can be removed once things have stabilized.

src/sentry/remote_config/endpoints.py
cmanallen: We're just letting the caller know the object that was persisted.  For now its unchanged but in the future these values might differ so we want to return what was committed.

aliu39 (approved):

src/sentry/api/api_owners.py
```diff
@@ -28,3 +28,4 @@ class ApiOwner(Enum):
     TELEMETRY_EXPERIENCE = "telemetry-experience"
     UNOWNED = "unowned"
     WEB_FRONTEND_SDKS = "team-web-sdk-frontend"
+    REMOTE_CONFIG = "replay-backend"
```
sentaur-athena, line 31: replay-backend is already here in the enum

src/sentry/api/urls.py
```diff
@@ -2364,6 +2365,11 @@ def create_group_urls(name_prefix: str) -> list[URLPattern | URLResolver]:
         r"^(?P<organization_id_or_slug>[^\/]+)/(?P<project_id_or_slug>[^\/]+)/keys/(?P<key_id>[^\/]+)/stats/$",
         ProjectKeyStatsEndpoint.as_view(),
     ),
+    re_path(
+        r"^(?P<organization_id_or_slug>[^\/]+)/(?P<project_id_or_slug>[^\/]+)/keys/(?P<key_id>[^\/]+)/configuration/$",
```
sentaur-athena, line 2369: Is the hierarchy here correct? Does configuration belong to a key? If not, key should probably be a query param and url be like ` r"^(?P<organization_id_or_slug>[^\/]+)/(?P<project_id_or_slug>[^\/]+)/configurations/$"`  which would list all configurations

src/sentry/remote_config/endpoints.py
```diff
@@ -0,0 +1,89 @@
...
+    traces_sample_rate = serializers.FloatField(max_value=1.0, min_value=0, required=True)
+    user_config = serializers.JSONField(required=True, allow_null=True)
+
+
+class ConfigurationContainerValidator(Serializer):
+    data = ConfigurationValidator()  # type: ignore[assignment]
+
+
+@region_silo_endpoint
+class ProjectConfigurationEndpoint(ProjectEndpoint):
```
sentaur-athena, line 29: If a configuration belongs to a key the way the url represents it, the endpoint is better be called `ProjectKeyConfigurationEndpoint`

src/sentry/api/api_owners.py
cmanallen: It will be replaced when the remote-config team is created.  This shouldn't affect anything.

src/sentry/api/urls.py
cmanallen: Hierarchy is correct.  We've not spec'd a list endpoint for configurations.  That will come in some later unit of work.

sentaur-athena: Is there a one to one relation between key and configuration?

asottile-sentry:
@cmanallen your new test is flaky -- can you disable it and fix it in a followup?

asottile-sentry:
I don't have anywhere near enough context to approve that and it has non-test changes as well

cmanallen:
@asottile-sentry Yeah fair enough.  I'll get it approved by a team member and merged soon.

PR #71263, opened 2024-05-21
fix(autofix): Add error message when codebase indexing fails

malwilley:


static/app/components/events/autofix/useAutofixCodebaseIndexing.tsx
```diff
@@ -62,24 +64,32 @@ export function useAutofixCodebaseIndexing({
...
+    try {
+      await api.requestPromise(
+        makeCodebaseIndexCreateUrl(organization.slug, projectSlug),
+        {
+          method: 'POST',
+        }
+      );
+    } catch {
+      addErrorMessage(
+        t('Something went wrong while Autofix was indexing your codebase.')
```
jennmueng, line 90: Can we allow a more specific error message to be surfaced from seer? So there can be two ways codebase indexing errors – a 500 from this endpoint would be from initializing the task, stuff like incorrect git setup etc like we just saw, or errors during the indexing itself which can be surfaced from the status endpoint that's polled

malwilley: Added more detailed messaged for the create and status endpoints, this is what the status looks like when errored:
 
CleanShot 2024-05-21 at 10 39 46

jennmueng (approved):

PR #71268, opened 2024-05-21
ref(similarity-embeddings): Update backfill script

jangjodi:
Update the seer grouping record backfill script in the following ways:
- Add only_delete flag that makes the backfill script only delete seer grouping records and the seer similarity data from the group metadata
- Remove dry_run seer grouping record deletion
- Add logging for reaching the end of the group id list in the backfill
- Add order_by id in the group filter


vartec (approved):

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -95,13 +98,19 @@ def backfill_seer_grouping_records(
...
 
-    if last_processed_index == 0 and dry_run:
-        logger.info(
-            "backfill_seer_grouping_records.delete_all_seer_records",
-            extra={"project_id": project.id},
-        )
-        delete_grouping_records(project_id)
-        redis_client.delete(make_backfill_redis_key(project_id))
+    if only_delete:
+        delete_seer_grouping_records(project.id, last_processed_index, True, redis_client)
```
JoshFerge, line 102: should use RangeSetQueryWrapper -- chunk size 100

JoshFerge (approved):

Issue #70859, opened 2024-05-14
CORS error when Assigning and Resolving issues.


Fwang36:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

Some users have reported receiving a CORS error when clicking the resolve or assign buttons of an issue.  I am unable to reproduce, but the users consistently reproduce the issue.

Information shared from the user will be attached to linked [JIRA ticket](https://getsentry.atlassian.net/browse/FEEDBACK-2175).

### Product Area

Issues

markstory:
Closing this as I've not heard of more problems related to this in a long time.

PR #71272, opened 2024-05-21
chore(api) Use capture_message for cors problem tracking

markstory:
I'm not having much luck in getting results from logs. This could be because the scenario isn't coming up, or that our logging can't query effectively over large timespans. Capturing an event will let me be more confident that this scenario isn't happening.

Refs #70859


wedamija (approved):
fwiw the logging should be working fine, but this is ok too

wedamija (commented):
Could also log a metric if you want to be extra sure

PR #71274, opened 2024-05-21
chore(webhooks): Creating a failing test for INC779

sentaur-athena:
Creating a test that would have failed before in case of json mismatch.

### DIFF
--- a/tests/sentry/api/serializers/test_app_platform_event.py
+++ b/tests/sentry/api/serializers/test_app_platform_event.py
@@ -1,3 +1,5 @@
+import datetime
+
 import orjson
 
 from sentry.api.serializers import AppPlatformEvent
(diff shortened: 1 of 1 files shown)


tests/sentry/api/serializers/test_app_platform_event.py
```diff
@@ -2,6 +2,7 @@
 
 from sentry.api.serializers import AppPlatformEvent
 from sentry.testutils.cases import TestCase
+from sentry.utils import json
```
anonrig, line 5: Can you move this to line 22, and can you use `import json` instead of this file, and add a comment to why we have this?

tests/sentry/api/serializers/test_app_platform_event.py
```diff
@@ -18,16 +19,13 @@ def test_no_actor(self):
...
-                    "installation": {"uuid": self.install.uuid},
-                    "data": {},
-                    "actor": {"type": "application", "id": "sentry", "name": "Sentry"},
-                }
-            ).decode()
+        assert result.body == json.dumps(
+            {
+                "action": "triggered",
+                "installation": {"uuid": self.install.uuid},
+                "data": {},
```
anonrig, line 26: Can you add a timestamp to this data?

anonrig (approved):
Perfect! Thank you.

PR #71275, opened 2024-05-21
feat(replay-new-trace): Removing trace meta query.

Abdkhan14:
We have access to the number of events from the `tree: TraceTree`


Abdkhan14:
This won't work since we the `tree:TraceTree` only represents the fetch count with a limit of 10k events.

PR #71281, opened 2024-05-21
feat(performance): Adds tooltip description and links to various insights modules

edwardgou-sentry:
Adds tooltips and doc links to the following insights modules:
- resources
- queries
- requests
- webvitals
- app start
- screen loads
- queues
- caches


gggritso (approved):
B e a utiful

PR #71283, opened 2024-05-21
fix(performance): Fix queues module empty state always displaying when an environment is selected

edwardgou-sentry:
Updates the `useHasData` to also query for `transaction` in order to force discover to use the non light `exclusive_time` metric. `exclusive_time_light` doesn't have an `environment` tag on the queues module, so this is a work around.

### DIFF
--- a/static/app/views/performance/onboarding/useHasData.tsx
+++ b/static/app/views/performance/onboarding/useHasData.tsx
@@ -17,7 +17,7 @@ export function useHasData(mutableSearch: MutableSearch, referrer: string) {
   const {data, isLoading, error} = useSpanMetrics(
     {
       search: mutableSearch,
-      fields: ['count()'],
+      fields: ['count()', 'transaction'],
       pageFilters: ninetyDayPageFilters,
     },
     referrer


DominikB2014 (approved):
Nice, The light vs non light is super uninutitive imo, and we run into this situation a good amount.

We can probably add a comment here why we need transaction for the next person. 

Also, I guess every metric you need is already tagged with environment then? That's why this issue is just showing up now?

edwardgou-sentry:
> Nice, The light vs non light is super uninutitive imo, and we run into this situation a good amount.
> 
> We can probably add a comment here why we need transaction for the next person.
> 
> Also, I guess every metric you need is already tagged with environment then? That's why this issue is just showing up now?

yeah it's pretty easy to trip over this :( 

I just went ahead and made the change in relay since it looks like it's also affecting other page elements:
https://github.com/getsentry/relay/pull/3632

So gunna hold off on merging this for now

edwardgou-sentry:
closing this since i dont think we need this anymore after adding environment as a tag to exclusive_time_light

PR #71286, opened 2024-05-21
chore(api): Publish `OrganizationReleaseDetailsEndpoint`

iamrajjoshi:
This PR removes previous JSON documentation for OrganizationReleaseDetailsEndpoint and adds new documentation to publish the endpoint.

image

image

image

For: https://github.com/getsentry/team-core-product-foundations/issues/325


src/sentry/apidocs/docs/organization_release_details_schema.py
```diff
@@ -0,0 +1,173 @@
+ORGANIZATION_RELEASE_DETAILS_SCHEMA = {
```
iamrajjoshi, line 1: i made this file because there is no serializer for the get response & the actual schema is very large, so defining it using OpenAPI was the easiest.

sentaur-athena:
It's good to have some eyes from the docs team on this. I'll review the code.

src/sentry/apidocs/docs/organization_release_details_schema.py
sentaur-athena: One goal of these changes are that we have docs that change dynamically when code changes happen. I think if someone change the return type of the endpoint the docs won't get updated this way. Can we have a typedict instead which we can use for typing in the code as well as docs?

src/sentry/api/endpoints/organization_release_details.py
```diff
@@ -269,29 +290,47 @@ def get_first_and_last_releases(self, org, environment, project_id, sort):
...
 
 
+@extend_schema(tags=["Releases"])
 @region_silo_endpoint
 class OrganizationReleaseDetailsEndpoint(
     OrganizationReleasesBaseEndpoint,
     ReleaseAnalyticsMixin,
     OrganizationReleaseDetailsPaginationMixin,
 ):
+    owner = ApiOwner.ENTERPRISE
```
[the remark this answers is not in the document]
iamrajjoshi: i asked on our product foundations channel and folks say we are supposed to own releases, but if we actually don't own this, i can change it

src/sentry/api/endpoints/organization_release_details.py
```diff
@@ -33,11 +46,19 @@ class InvalidSortException(Exception):
     pass
 
 
+@extend_schema_serializer(exclude_fields=["headCommits", "status"])
 class OrganizationReleaseSerializer(ReleaseSerializer):
     headCommits = ListField(
-        child=ReleaseHeadCommitSerializerDeprecated(), required=False, allow_null=False
+        child=ReleaseHeadCommitSerializerDeprecated(),
```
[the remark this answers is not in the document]
iamrajjoshi: we should be using `refs` instead. based on the code it seems like headCommits is a legacy param that uses older param names like `previousId` and `currentId`, and maps it to the current params we use in Release model to store refs.

https://github.com/getsentry/sentry/blob/2b1118681ba4ec610933f0a894c314f52b37a619/src/sentry/models/release.py#L578-L591

src/sentry/api/serializers/types.py
```diff
@@ -33,3 +33,146 @@ class OrganizationSerializerResponse(TypedDict):
     features: list[str]
     links: _Links
     hasAuthProvider: bool
+
+
+class ReleaseOptional(TypedDict, total=False):
```
sentaur-athena, line 38: Can you move these to their own file? Asking because right now you have `api.serializer.types.project` which sounds very generic and not specific to release. It can be used by someone by mistake instead of project serializer. Same as `version` it's not clear it's a release version being in the high level module.

src/sentry/api/endpoints/organization_release_details.py
```diff
@@ -33,11 +45,19 @@ class InvalidSortException(Exception):
...
+        required=False,
+        allow_null=False,
+    )
+    refs = ListField(
+        child=ReleaseHeadCommitSerializer(),
+        required=False,
+        allow_null=False,
+        help_text="""An optional way to indicate the start and end commits for each repository included in a release. Head commits must include parameters ``repository`` and ``commit`` (the HEAD sha). They can optionally include ``previousCommit`` (the sha of the HEAD of the previous release), which should be specified if this is the first time you've sent commit data.""",
```
vivianyentran, line 59:
```suggestion
        help_text="""An optional way to indicate the start and end commits for each repository included in a release. Head commits must include parameters ``repository`` and ``commit`` (the HEAD SHA). They can optionally include ``previousCommit`` (the SHA of the HEAD of the previous release), which should be specified if this is the first time you've sent commit data.""",
```

src/sentry/api/serializers/rest_framework/release.py
```diff
@@ -52,11 +52,29 @@ def validate_commit(self, value):
...
+        required=False,
+        allow_null=True,
+        allow_blank=True,
+        help_text="An optional commit reference. This is useful if a tagged version has been provided.",
+    )
+    url = serializers.URLField(
+        required=False,
+        allow_null=True,
+        allow_blank=True,
+        help_text="A URL that points to the release. This can be the path to an online interface to the sourcecode for instance.",
```
vivianyentran, line 65:
```suggestion
        help_text="A URL that points to the release. For instance, this can be the path to an online interface to the source code, such as a GitHub URL.",
```

src/sentry/api/serializers/rest_framework/release.py
```diff
@@ -52,11 +52,29 @@ def validate_commit(self, value):
...
+    )
+    dateReleased = serializers.DateTimeField(
+        required=False,
+        allow_null=True,
+        help_text="An optional date that indicates when the release went live.  If not provided the current time is assumed.",
```
vivianyentran, line 70:
```suggestion
        help_text="An optional date that indicates when the release went live.  If not provided, the current time is used.",
```

src/sentry/apidocs/parameters.py
```diff
@@ -158,6 +160,69 @@ class OrganizationParams:
...
+        required=False,
+        type=str,
+        description="The project id to filter by.",
+    )
+    HEALTH = OpenApiParameter(
+        name="health",
+        location="query",
+        required=False,
+        type=bool,
+        description="Include health data with the release.",
```
vivianyentran, line 183:
```suggestion
        description="Whether or not to include health data with the release. By default, this is ___",
```
Since it's not required, what's the default behavior here?

src/sentry/apidocs/parameters.py
```diff
@@ -158,6 +160,69 @@ class OrganizationParams:
...
+    )
+    ADOPTION_STAGES = OpenApiParameter(
+        name="adoptionStages",
+        location="query",
+        required=False,
+        type=bool,
+        description="Include adoption stages with the release.",
```
vivianyentran, line 190:
```suggestion
        description="Whether or not to include adoption stages with the release. By default, this is ____.",
```

src/sentry/apidocs/parameters.py
```diff
@@ -158,6 +160,69 @@ class OrganizationParams:
...
+    )
+    SUMMARY_STATS_PERIOD = OpenApiParameter(
+        name="summaryStatsPeriod",
+        location="query",
+        required=False,
+        type=str,
+        description="The period of time for the query, default is 14d.",
```
vivianyentran, line 197:
```suggestion
        description="The period of time used to query summary stats for the release. By default, this is 14d.",
```
Might also be helpful to mention where these stats are shown?

src/sentry/apidocs/parameters.py
```diff
@@ -158,6 +160,69 @@ class OrganizationParams:
...
+        enum=STATS_PERIODS.keys(),
+    )
+    HEALTH_STATS_PERIOD = OpenApiParameter(
+        name="healthStatsPeriod",
+        location="query",
+        required=False,
+        type=str,
+        description="The period of time for the query, default is 24h if health is enabled.",
```
vivianyentran, line 205:
```suggestion
        description="The period of time used to query health stats for the release. By default, this is 24h if health is enabled.",
```

src/sentry/apidocs/parameters.py
```diff
@@ -158,6 +160,69 @@ class OrganizationParams:
...
+        enum=STATS_PERIODS.keys(),
+    )
+    SORT = OpenApiParameter(
+        name="sort",
+        location="query",
+        required=False,
+        type=str,
+        description="The field to sort results by, default is `date`.",
```
vivianyentran, line 213:
```suggestion
        description="The field used to sort results by. By default, this is `date`.",
```

src/sentry/apidocs/parameters.py
```diff
@@ -158,6 +160,69 @@ class OrganizationParams:
...
+        enum=["date", "sessions", "users", "crash_free_users", "crash_free_sessions"],
+    )
+    STATUS_FILTER = OpenApiParameter(
+        name="status",
+        location="query",
+        required=False,
+        type=str,
+        description="Filter by release status.",
```
vivianyentran, line 221:
```suggestion
        description="Release statuses that you can filter by.",
```
Original description sounds like this could be a boolean

vivianyentran (approved):
Only looked at the help text & added a few copy suggestions

schew2381 (approved):

PR #71287, opened 2024-05-21
test(ui): Remove mock useOrganization

scttcper:
`render(<Component />, {organization})` will add an organization context for you, no need to mock our hooks which can hide other issues, like hooks being used outside components/hooks.


malwilley (approved):
Much better

ryan953 (approved):

PR #71294, opened 2024-05-21
feat(integrations): add state param to oauth redirect flow

### LINKED ISSUE #55794: Gaps in Integrations Platform external install flow
### Problem Statement

We're trying to use the external install flow described here: https://docs.sentry.io/product/integrations/integration-platform/public-integration/

However, while the flow looks like a standard OAuth2 flow, it does not appear to support passing a `state` query param (and consequently receiving the param back on redirect): https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1

Support for this param has the security benefits outlined in the spec, but most relevant is that it would allow us to explicitly control where external installs happen. In our case, we'd only want them to happen from within our dashboard.

Typically with other oauth flows, we pass in a `state` query param to the authorization URL that is actually a signed JWT that contains relevant context (account and other of our resource IDs that this install operating on). When the flow is redirected back to us, we can then validate that the flow started from us, and also extract the important context to continue with the next steps.

Without this param, we're not only unable to validate where the install started, but we'd be unable to continue since we'd be missing any context about what to do next, e.g. we wouldn't know which of our accounts wants to install our Sentry integration, and on what resources.

### Solution Brainstorm

_No response_

### Product Area

Settings - Integrations

JoshFerge:
Fixes https://github.com/getsentry/sentry/issues/55794

- [x] allows the query param `state` to be passed to the `sentry-apps/<integration>/external-install/` frontend webpage, allowing for integrations initiated externally to utilize this.
- [x] we'll pass this state to the integration oauth redirect, if provided
- [x] test added
- [x] tested locally, looks good

https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1

- Note that this does not add the state parameter for integration setups initiated on the sentry side.

### DIFF
--- a/static/app/views/sentryAppExternalInstallation/index.spec.tsx
+++ b/static/app/views/sentryAppExternalInstallation/index.spec.tsx
@@ -156,6 +156,53 @@ describe('SentryAppExternalInstallation', () => {
 
       (window.location.assign as jest.Mock).mockClear();
     });
+
+    it('installs and redirects with state', async () => {
+      const installUrl = `/organizations/${org1.slug}/sentry-app-installations/`;
+      const install = {
+        uuid: 'fake-id',
+        code: 'some-code',
+      };
+      const installMock = MockApiClient.addMockResponse({
+        url: installUrl,
+        method: 'POST',
+        body: install,
+      });
+
+      const state = 'some-state';
+      const location = {
+        ...RouteComponentPropsFixture(),
+        location: {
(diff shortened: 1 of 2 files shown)


nhsiehgit (approved):
🙏 🙌

saponifi3d (approved):
:shipit: lgtm, thanks for linking to the rfc spec as well.

PR #71295, opened 2024-05-21
fix(feedback): Adjust focus ring color and alignment of feedback list

### LINKED ISSUE #60365: User Feedback focus color for the infinite scrolling list needs fixing in dark-mode
This probably also applies to virtual lists around the app (like replay). We should check that too

But the problem is when you're using your keyboard to navigate the page, the focus color for that list isn't great. We should update the color

What it looks like now:

Image

The code that applies the color is actually coming from browser defaults. See `.ReactVirtualized__Grid .ReactVirtualized__List`  gets `.focus-visible` and `data-focus-visible-added` added, which sets UA styles: `outline: -webkit-focus-ring-color auto 1px;`

The list container could also use some rounded corners at the bottom to match the container/border.

Image

ryan953:
- focus color is purple now
- is-unseen circle is vertically centered
- focus for the Checkbox is not cutoff anymore

**Before**
SCR-20240521-mxvm

**After**
SCR-20240521-mxrg

Fixes https://github.com/getsentry/sentry/issues/60365


aliu39 (approved):

static/app/styles/global.tsx
```diff
@@ -193,6 +193,11 @@ const styles = (theme: Theme, isDark: boolean) => css`
     }
   }
 
+  .ReactVirtualized__Grid:focus-visible,
```
aliu3ntry, line 196: what do these 2 lines do?

billyvg: Is this going to affect our other virtualized grids in unwanted ways?

ryan953: @billyvg It'll also update the focus color of the other ones yeah. We use them a lot for the replay tabs, but `outline` is not a layout changing property, so it's only the colors that will change.

PR #71296, opened 2024-05-21
ref(insights): Dynamically generate the breadcrumbs

gggritso:
1. Re-uses a bit of code (poorly. The right solution would be a some kind of page nesting, where the URL determines the breadcrumb, but one thing at a time)
2. Makes it easier to generate the crumb based on a feature flag, so the root crumb can become "Insights" when we move

Using a hook for this is admittedly a little awk, but it has to check a feature flag (soon, anyway). I can change this to a regular function after we release, if that's necessary.

Also, as a nice benefit, makes all the breadcrumbs consistent

- Always translated
- Always linked


static/app/views/performance/utils/useModuleTitle.tsx
```diff
@@ -0,0 +1,33 @@
...
+  [ModuleName.MOBILE_UI]: MOBILE_UI_MODULE_TITLE,
+  [ModuleName.OTHER]: '',
+  [ModuleName.ALL]: '',
+};
+
+type ModuleNameStrings = `${ModuleName}`;
+type TitleableModuleNames = Exclude<ModuleNameStrings, '' | 'other'>;
+
+export function useModuleTitle(moduleName: TitleableModuleNames) {
+  return MODULE_TITLES[moduleName];
```
AbhiPrasad, line 32: Does this need to be a hook?

AbhiPrasad (approved):

static/app/views/performance/utils/useModuleTitle.tsx
gggritso: It does not! I just liked it for symmetry and because I _think_ I'll be using this hook to also generate page titles for the actual `<title>` tag, where it _does_ need to be a hook.

Good call though! My medium-term plan is to extract something like `InsightsPage` which will set the breadcrumb and title and so on in one place, and the hooks won't be needed anymore

PR #69690, opened 2024-04-25
fix(api-idorslug): Update `monitor_slug` convert_args to use `guid`

iamrajjoshi:
Because crons canonically uses guid instead of the numeric id, I created a `is_uuid` method to determine if the input to the API is a uuid or a slug. I use the method in the convert args. I updated the the path param tests as well.

Requires https://github.com/getsentry/getsentry/pull/13780

PR #71300, opened 2024-05-21
fix(crons): Prioritize lookup by slug over guid

evanpurkhiser:
In GH-69690 we unintentionally introduced a bug where monitors with slugs that are valid UUIDs (that differ from the monitor guid (id)) could not be loaded via APIs.

This is because we prioritized fetching by GUID when the passed slug_or_id when the value was a UUID.

### DIFF
--- a/src/sentry/monitors/endpoints/base.py
+++ b/src/sentry/monitors/endpoints/base.py
@@ -19,14 +19,6 @@ from sentry.utils.sdk import bind_organization_context, configure_scope
 DEPRECATED_INGEST_API_MESSAGE = "We have removed this deprecated API. Please migrate to using DSN instead: https://docs.sentry.io/product/crons/legacy-endpoint-migration/#am-i-using-legacy-endpoints"
 
 
-def is_uuid(monitor_id_or_slug: str | int) -> bool:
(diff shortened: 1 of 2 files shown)


src/sentry/monitors/endpoints/base.py
```diff
@@ -122,24 +114,35 @@ class ProjectMonitorEndpoint(ProjectEndpoint):
...
+            UUID(monitor_id_or_slug, version=4)
+            kwargs["monitor"] = Monitor.objects.get(
+                project_id=kwargs["project"].id, guid=monitor_id_or_slug
+            )
+            return args, kwargs
+        except ValueError:
+            # monitor_id_or_slug does not look like a GUID
+            pass
+        except Monitor.DoesNotExist:
+            pass
```
wedamija, line 143: Just combine both?

sentaur-athena (approved):

wedamija (approved):

src/sentry/monitors/endpoints/base.py
evanpurkhiser: will follow up

PR #71301, opened 2024-05-21
ref: Change grouping record backfill batch size

jangjodi:
Change record backfill batch size from 20 to 10
Fix flaky tests by setting event timestamp in the past


JoshFerge (approved):
thanks!

PR #71306, opened 2024-05-21
ref(tests): Add event_type to store_event

ceorourke:
Unless you add stacktrace information to `data`, events are created as the "default" type rather than error. This trips me up all the time when I expect them to be "error" type by default, so hopefully adding this should help others not have the same pitfall.


src/sentry/testutils/factories.py
```diff
@@ -899,10 +900,19 @@ def inject_performance_problems(jobs, _):
...
+        event_type: str = "default",
+        sent_at: datetime | None = None,
     ) -> Event:
-        # Like `create_event`, but closer to how events are actually
-        # ingested. Prefer to use this method over `create_event`
+        """
+        Like `create_event`, but closer to how events are actually
+        ingested. Prefer to use this method over `create_event`
+        """
+        if event_type == "error":
```
ceorourke, line 913: I'm open to other suggestions here - we have `create_performance_issue` and `store_functions` (which calls `store_event` and passes transactions) already, and you don't _typically_ need tests to have `error` type events, so I was also thinking of making a `create_error_issue` method.

tests/sentry/tasks/test_daily_summary.py
```diff
@@ -72,9 +69,7 @@ def store_event_and_outcomes(
                     data["release"] = release
 
                 event = self.store_event(
-                    data=data,
-                    project_id=project_id,
-                    assert_no_errors=False,
+                    data=data, project_id=project_id, assert_no_errors=False, event_type="error"
```
ceorourke, line 72: I can go through and update other tests that use `DEFAULT_EVENT_DATA` to create error events but for now I wanted to try this on something innocuous.

tests/sentry/api/endpoints/test_event_committers.py
```diff
@@ -28,9 +25,9 @@ def test_simple(self):
                 "fingerprint": ["group1"],
                 "timestamp": min_ago,
                 "release": release.version,
-                "stacktrace": copy.deepcopy(DEFAULT_EVENT_DATA["stacktrace"]),
             },
             project_id=project.id,
+            event_type="error",
```
saponifi3d, line 30: I'd recommend creating this as an enum - that'd mostly just be to help future us though. If it's an enum it would allow us to change the string from `error` to `Error` in 1 place (not a great example, but ya know...)

saponifi3d (approved):
just the enum nit, nbd tho.

saponifi3d (approved):
ty for doing that nit! :shipit:

PR #71307, opened 2024-05-21
ref(crons): Make check timeout qs variable name more explicit

davidenwang:
Got confused looking at this thinking it was checking for missed, changed to be more clear that its for timed out


evanpurkhiser (approved):
good catch

PR #71309, opened 2024-05-22
feat(org-stats): GA `project-stats` feature flag

scttcper:
This flag has been in GA for some time. Remove check in the frontend

https://github.com/getsentry/sentry/blob/master/src/sentry/conf/server.py#L1794


leeandher (approved):
Thanks @scttcper!

PR #71315, opened 2024-05-22
feat(crons): Pass request to quota APIs

evanpurkhiser:
This will allow getsentry to generate audit log entries with information
about the originating user causing seat allocation

This is part of https://github.com/getsentry/getsentry/issues/14009 (but it looks like we're not going to do this)

Proably still worth merging this for now if we do decide to add some kind of audit logs later.


wedamija (approved):
We already modified the getsentry code to be able to accept this right?

wedamija (changes requested):
Actually... if we're not doing this, why pass it down? Seems weird to pass the request anyway

evanpurkhiser:
yeah maybe I should just revert the other PRs lol

PR #71320, opened 2024-05-22
fix(list-item-details): Hidden on Safari

### LINKED ISSUE #69849: Info panel not working on Safari

ArthurKnaus:
### Problem
On Safari the details overlay is not visible once the menu list item is inside a scrollable menu that has a z-index set.
Basically, affecting all out current use cases where absolutely positioned menus float above the page content.

There is an open bug report for this on the webkit bug reporting platform:
https://bugs.webkit.org/show_bug.cgi?id=160953

### Solution
Use `createPortal` to mount the overlay directly in the body and adjust event handling for the combobox to keep interactive elements inside interactive. 

Closes https://github.com/getsentry/sentry/issues/69849

### DIFF
--- a/static/app/components/comboBox/index.tsx
+++ b/static/app/components/comboBox/index.tsx
@@ -123,12 +123,8 @@ function ComboBox<Value extends string>({
     isOpen: state.isOpen,
     position: 'bottom-start',
     offset: [0, 8],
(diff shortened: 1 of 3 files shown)


obostjancic (approved):
🚀

PR #71324, opened 2024-05-22
fix(cardinality-limit): Display default value

ArthurKnaus:
Display the default cardinality limit if none is set.


static/app/views/settings/projectMetrics/cardinalityLimit.tsx
```diff
@@ -3,66 +3,66 @@ import styled from '@emotion/styled';
...
       initialData={{
-        relayCustomMetricCardinalityLimit: project.relayCustomMetricCardinalityLimit,
+        relayCustomMetricCardinalityLimit:
+          project.relayCustomMetricCardinalityLimit ?? DEFAULT_LIMIT,
       }}
     >
       <Panel>
         <PanelHeader>{t('Limits')}</PanelHeader>
         <PanelBody>
-          {isLoading ? (
```
ArthurKnaus, line 40: The loading state was actually not needed here.

matejminar (approved):

PR #71326, opened 2024-05-22
feat(metrics): Expose span.duration

narsaynorath:
Exposes span.duration in the metrics query builder with a very limited set of available tags and replaces the `self_time` and `exclusive_time` metrics. We've decided that we do not want to
show these anymore.

People who have built widgets using this dashboard will still work, but they won't be able to select the field in new widgets, or re-select it in modified widgets.


narsaynorath:
If we want to move forward with this, should this be feature flagged? 🤔

phacops:
> If we want to move forward with this, should this be feature flagged? 🤔

I don't think so, we do not want `exclusive_time` exposed for the release.

obostjancic (approved):

PR #71330, opened 2024-05-22
chore(seer grouping): Add/fix various types

lobsterkatie:
This fixes and improves types in various places in the Seer similar issues code.

- Add types to tests.

- Add type to the `group` parameter in `GroupSimilarIssuesEmbeddingsEndpoint.get`. This in turn gave a type to `latest_event`, which necessitated two other changes:
  - A cast from `GroupEvent` to `Event`. The differences between the two, which have to do with issue platform `occurrence`s and the way that the `group` and `group_id` attributes are stored under the hood, aren't relevant here, and everything downstream expects an `Event`.
  - A non-null assertion on the hash. (Someday we should figure out when, if ever, the hash actually can be `None`, and clean up the typing here.)

- Handle the potential `None`-ness of various event attributes in tests better, by eliminating the `None` option in the `save_new_event` helper rather than being forced to do so on every instance of a value it returns. This works by having it return a subclass of `Event` in which the relevant attributes are guaranteed to have real values.

### DIFF
--- a/src/sentry/api/endpoints/group_similar_issues_embeddings.py
+++ b/src/sentry/api/endpoints/group_similar_issues_embeddings.py
@@ -1,6 +1,6 @@
 import logging
 from collections.abc import Mapping, Sequence
-from typing import Any, TypedDict
+from typing import Any, TypedDict, cast
 
 from django.contrib.auth.models import AnonymousUser
 from rest_framework.request import Request
@@ -12,6 +12,7 @@ from sentry.api.api_publish_status import ApiPublishStatus
 from sentry.api.base import region_silo_endpoint
 from sentry.api.bases.group import GroupEndpoint
 from sentry.api.serializers import serialize
+from sentry.eventstore.models import Event
 from sentry.grouping.grouping_info import get_grouping_info
 from sentry.models.group import Group
 from sentry.models.user import User
@@ -21,6 +22,7 @@ from sentry.seer.utils import (
     get_similarity_data_from_seer,
 )
 from sentry.utils.safe import get_path
+from sentry.utils.types import NonNone
 
 logger = logging.getLogger(__name__)
 MAX_FRAME_COUNT = 50
@@ -133,11 +135,11 @@ class GroupSimilarIssuesEmbeddingsEndpoint(GroupEndpoint):
 
         return [(serialized_groups[group_id], group_data[group_id]) for group_id in group_data]
 
-    def get(self, request: Request, group) -> Response:
+    def get(self, request: Request, group: Group) -> Response:
         if not features.has("projects:similarity-embeddings", group.project):
             return Response(status=404)
 
-        latest_event = group.get_latest_event()
+        latest_event = cast(Event, group.get_latest_event())
         stacktrace_string = ""
         if latest_event.data.get("exception"):
             grouping_info = get_grouping_info(None, project=group.project, event=latest_event)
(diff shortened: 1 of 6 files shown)


src/sentry/api/endpoints/group_similar_issues_embeddings.py
```diff
@@ -133,11 +135,11 @@ def get_formatted_results(
 
         return [(serialized_groups[group_id], group_data[group_id]) for group_id in group_data]
 
-    def get(self, request: Request, group) -> Response:
+    def get(self, request: Request, group: Group) -> Response:
         if not features.has("projects:similarity-embeddings", group.project):
             return Response(status=404)
 
-        latest_event = group.get_latest_event()
+        latest_event = cast(Event, group.get_latest_event())
```
asottile-sentry, line 142: cast should only be used for bugs in the type checker -- there's converters for `GroupEvent` from `Event` if that's what you're looking for -- or if this is an actual `GroupEvent` an `isinstance` check should be done instead

src/sentry/api/endpoints/group_similar_issues_embeddings.py
```diff
@@ -147,7 +149,7 @@ def get(self, request: Request, group) -> Response:
             return Response([])  # No exception, stacktrace or in-app frames
 
         similar_issues_params: SimilarIssuesEmbeddingsRequest = {
-            "hash": latest_event.get_primary_hash(),
+            "hash": NonNone(latest_event.get_primary_hash()),
```
asottile-sentry, line 152: please remove this helper -- this is going to cause a lot of pain to undo.  please don't introduce things like this without a review from the typing owners

asottile-sentry (changes requested):
I stopped when reaching something unacceptable -- and it looks like it has spread across the codebase already -- can you undo the usage of NonNone -- it is not sound and it is making the typing issues _worse_ and not better.  nullability should not be casted away it should be dealt with and handled as an error

asottile-sentry:
> I stopped when reaching something unacceptable -- and it looks like it has spread across the codebase already -- can you undo the usage of NonNone -- it is not sound and it is making the typing issues _worse_ and not better. nullability should not be casted away it should be dealt with and handled as an error

to elaborate on this, even in typescript it is strongly discouraged and we should not allow that here either: https://typescript-eslint.io/rules/no-non-null-assertion/

asottile-sentry:
given this it seems like the type annotation is wrong and you can probably remove the `str | None` entirely -- you may want to try and land a PR like this first (perhaps with a metric to give confidence that it's actually unreachable): https://github.com/getsentry/sentry/pull/71377

lobsterkatie:
I agree that non-null assertion (and to a lesser extent casts) should be used sparingly, but I do think the "I know what this will be in this one circumstance, and I don't have time to go investigating how this thing is used everywhere else in order to see if it's possible to change the underlying type" is a valid use case (and a safer one, frankly - I don't have the context the original author of whatever helper I'm using did, and it doesn't feel wise to do the typing equivalent of over-driving my headlights if I can help it).

And look, I get it. At heart I'm actually a total OCD purist whose natural instinct is _absolutely_ to do the metaphorical DFS to understand everything and fix everything the "right" way. But that's also the thing I have most consistently gotten dinged for (since that's rarely what I'm _supposed_ to be doing), so I've worked hard to train myself to find a balance, while still (sometimes) indulging my strong impulse to leave things better than I found them (which is how I come to fixes like this in the first place).

Anyway, for now I've pulled the test changes out into a [separate PR](https://github.com/getsentry/sentry/pull/71389), added a log for missing hash [here](https://github.com/getsentry/sentry/pull/71397), and will close this and circle back to the endpoint type change.

PR #71339, opened 2024-05-22
chore: remove unused pylibmc

anonrig:
Removes unused pylibmc dependency


asottile-sentry (approved):
I take it the pylibmc experiment didn't pan out?

PR #71342, opened 2024-05-22
feat(profiling): add Dataset and QueryBuilder for profile functions metrics

viglia:
Add Dataset and QueryBuilder utilities for profile functions metrics.

This will allow us to query the metrics data that we have started to ingest into the generic metrics platform.


pyproject.toml
```diff
@@ -406,6 +406,7 @@ module = [
     "sentry.search.events.datasets.metrics",
     "sentry.search.events.datasets.metrics_layer",
     "sentry.search.events.datasets.profile_functions",
+    "sentry.search.events.datasets.profile_functions_metrics",
```
asottile-sentry, line 409: new things should not be added to the typing blocklist -- please fix your annotations and get the type checker passing on this file

asottile-sentry (dismissed):

src/sentry/search/events/builder/profile_functions_metrics.py
```diff
@@ -0,0 +1,50 @@
...
+    column_remapping = {
+        # We want to remap `message` to `name` for the free
+        # text search use case so that it searches the `name`
+        # (function name) when the user performs a free text search
+        "message": "name",
+    }
+
+    @property
+    def use_default_tags(self) -> bool:
+        return False
```
wmak, line 24: I think you want to use default tags here (I'm going to remove this as a config later) and add that constants list of the tags that exist for the profiling metrics

src/sentry/search/events/builder/profile_functions_metrics.py
```diff
@@ -0,0 +1,50 @@
...
+            self.has_transaction = True  # if always true can we skip this?
+        return super().resolve_select(selected_columns, equations)
+
+
+class TimeseriesProfileFunctionsMetricsQueryBuilder(
+    ProfileFunctionsMetricsQueryBuilder, TimeseriesMetricQueryBuilder
+):
+    def resolve_split_granularity(self) -> tuple[list[Condition], Granularity | None]:
+        """Don't do this for timeseries"""
+        return [], self.granularity
```
wmak, line 44:
```suggestion
    pass
```

wmak (commented):
Mostly lgtm, only blocking thing really is the `use_default_tags` comment, and I'd feel a bit better about this if we had some integration tests

src/sentry/search/events/datasets/base.py
```diff
@@ -13,7 +13,7 @@
 class DatasetConfig(abc.ABC):
     custom_threshold_columns: set[str] = set()
     non_nullable_keys: set[str] = set()
-    missing_function_error = InvalidSearchQuery
+    missing_function_error: Any = InvalidSearchQuery
```
asottile-sentry, line 16: this is not a fix, this is a hack -- this should probably be `ClassVar[type[Exception]]` I believe?

src/sentry/search/events/datasets/profile_functions_metrics.py
```diff
@@ -101,10 +102,16 @@ def _resolve_cpm_cond(
         alias: str | None,
         cond: str,
     ) -> SelectType:
+        timestmp = args["timestamp"]
         if cond == "greater":
-            interval = (self.builder.params.end - args["timestamp"]).total_seconds()
+            assert isinstance(self.builder.params.end, datetime) and isinstance(timestmp, datetime)
```
asottile-sentry, line 107: the assertion should include the values otherwise it'll end up being a useless messageless `AssertionError:` on failure

this also seems like this is maybe unsound -- does something else require these to be datetime elsewhere?

src/sentry/search/events/datasets/profile_functions_metrics.py
```diff
@@ -66,6 +66,7 @@ def _resolve_cpm(
         alias: str | None,
         extra_condition: Function | None = None,
     ) -> SelectType:
+        assert self.builder.params.end is not None and self.builder.params.start is not None
```
asottile-sentry, line 69: if these are required to be non-None -- consider updating the class that they come from to enforce that instead of requiring assertions here

src/sentry/search/events/builder/__init__.py
```diff
@@ -21,6 +21,11 @@
     ProfileFunctionsTimeseriesQueryBuilder,
     ProfileTopFunctionsTimeseriesQueryBuilder,
 )
+from .profile_functions_metrics import (  # NOQA
+    ProfileFunctionsMetricsQueryBuilder,
+    TimeseriesProfileFunctionsMetricsQueryBuilder,
+    TopProfileFunctionsMetricsQueryBuilder,
+)
```
asottile-sentry, line 28: we've been trying to reduce barrel files across the codebase as this impacts startup time, test time, makes it very easy to introduce import cycles, and makes it harder to find where something is defined

I'd recommend using an absolute import path and not making the problem worse by expanding this `__init__.py` file

wmak (approved):
lgtm

src/sentry/search/events/builder/__init__.py
asottile-sentry: @viglia please at least acknowledge this comment before merging

PR #71345, opened 2024-05-22
feat(insights): add span search bar to queries/resource sample panel

KevinL10:
Adds a sample span search bar to the queries and resource sample panels, similar to Trace Explorer:
image

This features is gated under `performance-sample-panel-search`.


static/app/views/starfish/utils/useSpanFieldSupportedTags.tsx
```diff
@@ -0,0 +1,61 @@
...
+      dynamicTagQuery.data.map(entry => [entry.key, entry])
+    );
+    return {
+      ...dynamicTags,
+      ...staticTags,
+    };
+  }
+
+  return staticTags;
+}
```
0Calories, line 61: Good call on extracting this out into a hook, just one minor thing to consider; I suggest we put this file in `/performance/utils` instead of `starfish` because this hook is also being used by Traces, and we're moving on from starfish based nomenclature, so we should avoid adding new files there

0Calories (approved):
Nice!

static/app/views/starfish/views/spanSummaryPage/sampleList/index.tsx
```diff
@@ -97,9 +117,12 @@ export function SampleList({
     })}`
   );
 
-  let extraQuery: string[] | undefined = undefined;
+  // set additional query filters from the span search bar and the `query` param
+  let extraQuery: string[] | undefined = spanSearchQuery?.split(' ');
```
narsaynorath, line 121: We might need to be wary about how we're applying this `.split` because I can imagine that some span attributes could very well have spaces in them? Not 100% sure but something worth checking out

KevinL10: Ah actually you're right, you can include spaces like `browser.name:"firefox 120"`. I believe trace explorer sends the query as a single string which the backend parses based on the grammar. I'll update the span-samples route to support this in the next PR.

narsaynorath: Yeah, let's look at the options but because this is feature flagged it's all good to merge and figure it out later. In the frontend we try to use this class called `MutableSearch` which takes a series of conditions and then there's a `.formatString()` method that spits out an escaped string, but we've noticed some issues with its implementation

If the backend also can handle it, that could be a good option as well

narsaynorath (approved):

PR #71346, opened 2024-05-22
ref: fix typing for testutils task_runner

asottile-sentry:
this does two things:

- fix the typing issues for BurstTaskRunner
- fix the mock so it does not leak in weird ways losing enqueued tasks
- clean up useless `burst_task_runner` fixture

I had to fix a bunch of tests which were either leaking enqueued tasks or running in a mixed-mock state.  one test required temporarily undoing the mock which complicated the whole thing


azaslavsky (approved):
Thanks for fixing the spelling error.

PR #71359, opened 2024-05-22
ref(insights): Use documentation links from constants

gggritso:
Consistency, easier to change, etc. You get it.


edwardgou-sentry (approved):
Thanks!

PR #71368, opened 2024-05-22
feat(llm-monitoring): Make LLM monitoring look like other insight modules

colin-sentry:
Use starfish charts and add an empty onboarding state.


gggritso (approved):
Clean!

PR #71372, opened 2024-05-22
ref(OrganizationsStore): Return stable state from getState, strict types

scttcper:
getState will now always return the same object if state has not changed. Removes mutations to state.

The plan is to use `StrictStoreDefinition` for all stores and I'm just showing that a few require basically no changes, OrganizationsStore was mutating and not returning a stable reference from getState.


static/app/stores/organizationsStore.tsx
```diff
@@ -23,27 +21,21 @@ interface OrganizationsStoreDefinition extends State, CommonStoreDefinition<Stat
 }
 
 const storeConfig: OrganizationsStoreDefinition = {
-  organizations: [],
-  loaded: false,
```
scttcper, line 27: the types here were `any`

evanpurkhiser (approved):

PR #71373, opened 2024-05-22
ref(insights): Create sidebar links using `useModuleURL`

gggritso:
Doesn't make sense to have a `useModuleURL` helper if I can't use it to generate links to the modules in the sidebar. This adds a _slightly lower level_ helper, so that the same function makes _all_ links to module. Which, you guessed it, will make it easier to change all the URLs behind a feature flag.

## Changes

- **Allow creating bare module URLs**
- **Split `useModuleURL` into a builder and a wrapper**
- **Remove leading slash from insights base URL**
- **Link sidebar items using `useModuleURLBuilder`**

### DIFF
--- a/static/app/views/performance/utils/useModuleURL.tsx
+++ b/static/app/views/performance/utils/useModuleURL.tsx
@@ -31,15 +31,20 @@ export const MODULE_BASE_URLS: Record<ModuleName, string> = {
 type ModuleNameStrings = `${ModuleName}`;
 type RoutableModuleNames = Exclude<ModuleNameStrings, '' | 'other'>;
 
-export const useModuleURL = (moduleName: RoutableModuleNames): string => {
+export const useModuleURL = (
+  moduleName: RoutableModuleNames,
+  bare: boolean = false
+): string => {
   const {slug} = useOrganization();
 
   if (moduleName === ModuleName.AI) {
     // AI Doesn't live under "/performance"
-    return normalizeUrl(`/organizations/${slug}/${AI_BASE_URL}`);
+    return bare
+      ? `${slug}${AI_BASE_URL}`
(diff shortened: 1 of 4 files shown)


static/app/views/performance/utils/useModuleURL.tsx
```diff
@@ -31,15 +31,31 @@ export const MODULE_BASE_URLS: Record<ModuleName, string> = {
...
+type URLBuilder = (moduleName: RoutableModuleNames, bare?: boolean) => string;
+
+export function useModuleURLBuilder(): URLBuilder {
   const {slug} = useOrganization();
 
-  if (moduleName === ModuleName.AI) {
-    // AI Doesn't live under "/performance"
-    return normalizeUrl(`/organizations/${slug}/${AI_BASE_URL}`);
-  }
+  return function (moduleName: RoutableModuleNames, bare: boolean = false) {
```
AbhiPrasad, line 54: Why is `bare = false`, I feel like `true` makes more sense as a default.

Also, I think `useModuleURLBuilder` should take in `bare`, so then the `builder` just inherits those settings and you can enforce a common set of module routes have the same path.

gggritso: `bare` is `false` by default because most usage of `useModuleURL` expect it to return the full URL, the sidebar is the exception. I can set it to `true` by default though, I don't have a strong opinion.

As for `useModuleURLBuilder` taking in `bare` as a parameter, that works for me, I can make the change

AbhiPrasad: > the sidebar is the exception

Yes you're totally right, I just forgot about that, let's keep it as `false` then!

gggritso: Done!

AbhiPrasad (approved):

PR #71379, opened 2024-05-22
feat(discover): Add dataset to Discover Saved Queries

shruthilayaj:
This is meant to allow us to track whether a Discover Saved Query queries the Errors or Transactions in CH. We want to do this so that eventually any queries going to Transactions can be routed to the Indexed spans dataset and Errors still go to Errors.


src/sentry/migrations/0522_migrate_discover_savedquery_worldmaps.py
```diff
@@ -4,13 +4,13 @@
...
 from django.db.migrations.state import StateApps
 
-from sentry.discover.models import DiscoverSavedQuery
 from sentry.new_migrations.migrations import CheckedMigration
 
 
 def migrate_savedquery_worldmap_display_to_totalPeriod(
     apps: StateApps, schema_editor: BaseDatabaseSchemaEditor
 ) -> None:
+    DiscoverSavedQuery = apps.get_model("sentry", "DiscoverSavedQuery")
```
shruthilayaj, line 13: migration tests were failing because of a direct import here. is it safe to retro fix a migration like this?

src/sentry/discover/models.py
```diff
@@ -8,15 +8,33 @@
...
 MAX_KEY_TRANSACTIONS = 10
 MAX_TEAM_KEY_TRANSACTIONS = 100
 
 
+class DiscoverSavedQueryTypes(TypesClass):
+    ERROR_EVENTS = 0
+    """
+     Error side of the split from Discover.
+    """
+    TRANSACTION_LIKE = 1
```
shruthilayaj, line 26: I'm just using the same names we do on Dashboards

shruthilayaj: Or maybe it's better to name them same as actual datasets? I don't have any strong opinions

src/sentry/migrations/0522_migrate_discover_savedquery_worldmaps.py
wedamija: Yeah that's ok, if we don't fix this then this migration will fail

src/sentry/discover/models.py
wedamija: Feels like it should be the real datasets? Not sure exactly how this is used though

src/sentry/discover/models.py
```diff
@@ -49,6 +67,7 @@ class DiscoverSavedQuery(Model):
     visits = BoundedBigIntegerField(null=True, default=1)
     last_visited = models.DateTimeField(null=True, default=timezone.now)
     is_homepage = models.BooleanField(null=True, blank=True)
+    dataset = BoundedPositiveIntegerField(choices=DiscoverSavedQueryTypes.as_choices(), null=True)
```
wedamija, line 70: Do we want to have a default here, or attempt to backfill? If all existing rows are of one type it could make sense

src/sentry/discover/models.py
k-fish: The problem with using actual datasets, is that in some cases (eg. dashboards) the `Transactions` side of `Errors & Transaction` is not the actual dataset / indexed events, but a series of checks and attempts against using metrics that punch past DS. 

Hence "transactions-like".

src/sentry/discover/models.py
k-fish: When I did this for dashboards, I wasn't sure and did `null`, but we can probably default to `Discover` (eg. `Both`) if we have it as a type.

shruthilayaj: We'll want to backfill, no point in setting a default - it would just be 'discover' or something, it's pretty meaningless

shruthilayaj: There are no other dataset types for discover queries, so I'm like... meh, but I don't have a strong preference

src/sentry/discover/models.py
shruthilayaj: yeah, agreed kevan. i thought about it a bit more, and it needs to be be the "dataset" that the frontend thinks we're using. we'll add trace search to the dataset selector probably, and both transactions and trace search will be hitting the spans dataset.

src/sentry/discover/models.py
markstory: If all the existing records have the same value, a default would let you backfill all existing rows at once. In the future will `null` be a valid state? If not, adding a default and disallowing null would be better.

shruthilayaj: Ah, good point Mark, I'll add a default

shruthilayaj: Added default dataset + a separate db and state migration like our docs suggest https://github.com/getsentry/sentry/pull/71379/commits/3f02bdb7edb8de0c8c074bc3484e241254dd9009

markstory (approved):
Schema changes look good to me.

wedamija (approved):

PR #71382, opened 2024-05-22
ref(context): Generalize components for issue details and trace view

leeandher:
This removes the duplicate code from the trace drawer and consolidates it into one component in `KeyValueData.Card` and `KeyValueData.Content`.

I don't have a lot of familiarity with traces or setting them/their views up locally, so I don't have any screenshots from that side working, I'd appreciate if someone from that side could pull this branch and test it locally, or provide me with instructions.

**Issue Details**

image

**Trace View**

image

**Trace when Subject is null**

333238648-1bda26d5-1925-4cbd-9387-404ce83a65c1


Abdkhan14 (commented):
Hey @leeandher, 
Here's a quick dev-ui link to a rich trace. 
https://sentry.dev.getsentry.net:7999/performance/trace/b3e70f1145aa452ab256b5949c751ceb/?fov=0%2C23663.000106811523&node=trace-root&statsPeriod=1h&timestamp=1716405923

Hey I added being able to pass subject as null for cases where we only have a value, like just like loading indicator or the no tags component. There are more cases. Otherwise, all the other functionality persists:
Screenshot 2024-05-22 at 5 24 12 PM

Abdkhan14 (approved):

PR #71386, opened 2024-05-22
fix(release-alert): Update Query to use Org & Project Filtering for Release Health

iamrajjoshi:
Helps mitigate https://getsentry.atlassian.net/browse/ISSUE-1702

I create an option to query switch between the old query and the new query.

The new query is the same as old query p much, but with few differences. we sort by `org_id` and `project_id` and use the last one as the filter for the next query.

### DIFF
--- a/src/sentry/release_health/release_monitor/metrics.py
+++ b/src/sentry/release_health/release_monitor/metrics.py
@@ -181,7 +181,109 @@ class MetricReleaseMonitorBackend(BaseReleaseMonitorBackend):
             return self.fetch_projects_with_recent_sessions_with_filter()
         return self.fetch_projects_with_recent_sessions_with_offset()
 
-    def fetch_project_release_health_totals(
+    def fetch_project_release_health_totals_with_org_project_filter(
+        self, org_id: int, project_ids: Sequence[int]
+    ) -> Totals:
+        start_time = time.time()
+        prev_org_id = 0
+        prev_project_id = 0
+        totals: Totals = defaultdict(dict)
+        with metrics.timer("release_monitor.fetch_project_release_health_totals.loop"):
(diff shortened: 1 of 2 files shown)


src/sentry/release_health/release_monitor/metrics.py
```diff
@@ -94,7 +95,109 @@ def fetch_projects_with_recent_sessions(self) -> Mapping[int, Sequence[int]]:
...
+                        Condition(Column("org_id"), Op.EQ, org_id),
+                        Condition(Column("project_id"), Op.IN, project_ids),
+                        Condition(
+                            Column("metric_id"),
+                            Op.EQ,
+                            indexer.resolve(
+                                UseCaseID.SESSIONS, org_id, SessionMRI.RAW_SESSION.value
+                            ),
+                        ),
+                        Condition(Column("org_id"), Op.GTE, prev_org_id),
```
iamrajjoshi, line 140: We might not have processed all the projects for an org.

src/sentry/release_health/release_monitor/metrics.py
```diff
@@ -94,7 +95,109 @@ def fetch_projects_with_recent_sessions(self) -> Mapping[int, Sequence[int]]:
...
+                    if not more_results:
+                        break
+            else:
+                logger.error(
+                    "fetch_project_release_health_totals.loop_timeout",
+                    extra={"org_id": org_id, "project_ids": project_ids},
+                )
+        return totals
+
+    def fetch_project_release_health_totals_with_offset(
```
iamrajjoshi, line 200: old query style

iamrajjoshi:
@wedamija if we don't want/need to do this, i can close this PR

PR #71389, opened 2024-05-22
chore(seer grouping): Add/clean up types in tests

lobsterkatie:
This fixes and improves types in various places in the Seer similar issues tests.

- Add types to test parameters.

- Handle the potential `None`-ness of various event attributes in tests better, by eliminating the `None` option in the `save_new_event` helper rather than being forced to do so on every instance of a value it returns. This works by having it return a subclass of `Event` in which the relevant attributes are guaranteed to have real values.

### DIFF
--- a/src/sentry/testutils/helpers/eventprocessing.py
+++ b/src/sentry/testutils/helpers/eventprocessing.py
@@ -1,10 +1,12 @@
-from typing import Any
+from typing import Any, cast
 
 from sentry.event_manager import EventManager
-from sentry.eventstore.models import Event
+from sentry.eventstore.models import Event as _Event
 from sentry.eventstore.processing import event_processing_store
+from sentry.models.group import Group
 from sentry.models.project import Project
 from sentry.testutils.helpers.task_runner import TaskRunner
+from sentry.utils.types import NonNone
 
 
 def write_event_to_cache(event):
@@ -12,6 +14,18 @@ def write_event_to_cache(event):
     cache_data["event_id"] = event.event_id
     cache_data["project"] = event.project_id
     return event_processing_store.store(cache_data)
+
+
+# Wrap the real `Event` type in order to let mypy know that certain attributes are guaranteed
+# to be defined
+class Event(_Event):
+    group: Group
+    group_id: int
+
+    # Note: This function never gets called , but writing it out this way is necessary to change the
+    # return type in mypy's eyes
+    def get_primary_hash(self) -> str:
+        return NonNone(super().get_primary_hash())
 
 
 def save_new_event(event_data: dict[str, Any], project: Project) -> Event:
@@ -25,4 +39,4 @@ def save_new_event(event_data: dict[str, Any], project: Project) -> Event:
     with TaskRunner():
         event = EventManager(event_data).save(project.id)
 
-    return event
+    return cast(Event, event)
--- a/tests/sentry/api/endpoints/test_group_similar_issues_embeddings.py
+++ b/tests/sentry/api/endpoints/test_group_similar_issues_embeddings.py
@@ -18,7 +18,6 @@ from sentry.seer.utils import SeerSimilarIssueData, SimilarIssuesEmbeddingsRespo
 from sentry.testutils.cases import APITestCase
 from sentry.testutils.helpers.eventprocessing import save_new_event
 from sentry.testutils.helpers.features import with_feature
-from sentry.utils.types import NonNone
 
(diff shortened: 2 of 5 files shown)


src/sentry/testutils/helpers/eventprocessing.py
```diff
@@ -25,4 +39,4 @@ def save_new_event(event_data: dict[str, Any], project: Project) -> Event:
     with TaskRunner():
         event = EventManager(event_data).save(project.id)
 
-    return event
+    return cast(Event, event)
```
JoshFerge, line 42: are these changes meant to be in this PR?

lobsterkatie: Yes. That's the second bullet point in the PR description, moving the handling of the `None` possibility into the test helper itself.

asottile-sentry (changes requested):
hold on, this is the same PR without my concerns addressed in https://github.com/getsentry/sentry/pull/71330

lobsterkatie:
> hold on, this is the same PR without my concerns addressed in #71330

No. I responded to your comments, and took out the part where you'd requested changes.

asottile-sentry:
> > hold on, this is the same PR without my concerns addressed in #71330
> 
> No. I [responded](https://github.com/getsentry/sentry/pull/71330#issuecomment-2125715829) to your comments, and took out the part where you'd requested changes.
> 
> > Anyway, for now I've pulled the test changes out into a [separate PR](https://github.com/getsentry/sentry/pull/71389), added a log for missing hash [here](https://github.com/getsentry/sentry/pull/71397), and will close this and circle back to the endpoint type change.

I still see a `cast`, a fake type that undermines the type system, and a `NonNone` call

lobsterkatie:
Yes, there is a `cast`. Some way or another I have to tell `mypy` that these attributes are defined in my tests, and this was my way to do it which _removed_ all the `NonNone`s which were in there.

src/sentry/testutils/helpers/eventprocessing.py
dbennett-sentry: FWIW, this is a bit strange. We already have an `Event` defined. Why do we assume your `Event` has those properties when the _other_ `Event` does not? What if the object you're casting here doesn't have those properties? 

As a developer, if I were looking at `foo(e: Event) -> Rainbows` how would I know which `Event` we're even talking about?

I _think_ that's what @asottile-sentry was referring to regarding typing confusion.

src/sentry/testutils/helpers/eventprocessing.py
```diff
@@ -14,6 +16,18 @@ def write_event_to_cache(event):
...
+# Wrap the real `Event` type in order to let mypy know that certain attributes are guaranteed
+# to be defined
+class Event(_Event):
+    group: Group
+    group_id: int
+
+    # Note: This function never gets called , but writing it out this way is necessary to change the
+    # return type in mypy's eyes
+    def get_primary_hash(self) -> str:
+        return NonNone(super().get_primary_hash())
```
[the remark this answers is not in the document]
lobsterkatie: Yes, you're right, there's still one in there. And I can split it out into `primary_hash = ...; assert primary_hash; return primary_hash` if I must but that's exactly (and all) the `NonNone` is doing. I guess I really don't see how a helper whose only job it to assert something exists is so problematic.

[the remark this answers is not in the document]
asottile-sentry: sometimes when you set up types you're going to have to fix types in other modules -- that's ok!  often individuals of the past annotated things ~slightly wrong because our typing system wasn't set up properly.  sometimes you're going to have to put in a tiny bit of extra work to make everything better for everyone.

this is one of those examples!  the function you've been hacking around is probably typed incorrectly and with a little bit of investigation (a metric if it's `None`, probably never fired) you should be able to fix that and then make everything better for everyone.

my concern with this NonNone thing is that it makes it extremely easy to unsafely cast away nullability.  where the correct fix is usually to handle it as an error *or* fix the underlying nullability at its source. that is why this "helper" should have never existed in the first place and why I'm so strongly trying to discourage its use now

I understand that this probably wasn't obvious but that's why I'm here now. you can always add me or the typing group to your typing PRs if you want earlier feedback.

also, here are some  [guidelines](https://www.notion.so/sentry/sentry-typing-guidelines-best-practices-471d5aacf262476188982ce70f909e00) we have created and a reminder about [cast](https://www.notion.so/sentry/sentry-typing-guidelines-best-practices-471d5aacf262476188982ce70f909e00?pvs=4#4c5f66d5e5f2460991cf1074111fdd85) as well as `# type: ignore` / etc from our best practices document

PR #71391, opened 2024-05-22
fix(u2f): Have authenticators GET wait for preloading to finish

cathteng:
Same as #71142 without `await loadingOrganization` and uses `finally`


saponifi3d (approved):
re: +1. curious about the root cause for the await difference.. 🤔

PR #71393, opened 2024-05-22
fix(release-alert): Update Query to use Org & Project Filtering for `fetch_projects_with_recent_sessions`

iamrajjoshi:
Helps mitigate https://getsentry.atlassian.net/browse/ISSUE-1702, by updating the query in `fetch_projects_with_recent_sessions`.

I create an option to query switch between the old query and the new query.

The new query is the same as old query p much, but with few differences. we sort by org_id and project_id and use the last one as the filter for the next query.


wedamija (approved):
Looks good, thanks!

PR #71397, opened 2024-05-22
chore(grouping): Log when an event has no primary hash

lobsterkatie:
In theory (at least according to our code), sometimes events have no primary hash. It's not clear exactly when that'd happen, though. This adds a log to track it, as a first step towards figuring that out.


jangjodi:
do we have any idea how often this would happen? just want to make sure this log doesn't fire too often

lobsterkatie:
The idea is, hopefully never, at which point we can fix the types.

jangjodi (approved):

PR #71400, opened 2024-05-22
ref(ui): Specialize accordion to replays

evanpurkhiser:
Similar to GH-71380

This component was too specialized for the two use cases it was serving. Instead of trying to generalize it more, for now it's easier to just move it closer to the two places it's being used and give it the necessary interface for each

The main difference is that I've removed all of the logic that was specific to performance widgets. The styling was quite different


static/app/components/accordion/accordion.stories.tsx
```diff
@@ -1,99 +0,0 @@
...
-                  buttonOnLeft={buttoOnLeft}
-                />
-              </SizingWindow>
-            </div>
-          ))}
-        </SideBySide>
-      </Fragment>
-    );
-  });
-});
```
evanpurkhiser, line 99: I've removed the story since this should really be more reserved for components we want to reuse

scttcper (approved):

michellewzhang (approved):

PR #71404, opened 2024-05-22
fix(trace): rerun effect when coming from cached request

JonasBa:
We need to ensure qs path is initialized sync and before the children components render


Abdkhan14 (approved):
lgtm

PR #71409, opened 2024-05-22
feat(insights): add search bar to mobile sample panels

KevinL10:
Adds search bar to mobile sample panels.
image

We have an independent search bar for each of the releases, so that users can narrow down on spans in one release without affecting the sample table of the other release.

### DIFF
--- a/static/app/views/performance/mobile/components/spanSamplesPanel.tsx
+++ b/static/app/views/performance/mobile/components/spanSamplesPanel.tsx
@@ -32,6 +32,9 @@ type Props = {
   transactionRoute?: string;
 };
 
+const PRIMARY_SPAN_QUERY_KEY = 'primarySpanSearchQuery';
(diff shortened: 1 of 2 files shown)


static/app/views/performance/mobile/components/spanSamplesPanel.tsx
```diff
@@ -82,6 +86,26 @@ export function SpanSamplesPanel({
...
+
+  const handleSecondarySearch = (newSpanSearchQuery: string) => {
+    router.replace({
+      pathname: location.pathname,
+      query: {
+        ...query,
+        secondarySpanSearchQuery: newSpanSearchQuery,
+      },
+    });
+  };
```
narsaynorath, line 107: Can we define this as a generic method in the samples container and pass down the key to set in the URL?

narsaynorath (approved):
Just one more comment but this looks good to go!

Issue #71396, opened 2024-05-22
[Hydration Errors] Add setting to "create hydration errors from replays"


ryan953:
similar to `sentry:replay_rage_click_issues`

When the user goes to a page like https://sentry.sentry.io/settings/projects/javascript/replays/ they should see a new checkbox that says something like "Toggles whether or not to create Session Replay Hydration Issues"

![Image](https://github.com/getsentry/sentry/assets/187460/a5750ccf-5c99-4870-841a-519568d5dcef)

aliu3ntry:
Need to add to UI

PR #71411, opened 2024-05-22
feat(replay): project option (toggle in settings) for hydration issues

aliu39:
https://github.com/getsentry/sentry/issues/71396


cmanallen (changes requested):
Missing test coverage.  Write tests showing the option is returned and the option can be written to.

tests/sentry/api/serializers/test_project.py
```diff
@@ -748,6 +748,15 @@ def test_replay_rage_click_flag(self):
...
         assert result["options"]["sentry:replay_rage_click_issues"] is False
 
+    def test_replay_hydration_error_flag(self):
+        result = serialize(self.project, self.user, DetailedProjectSerializer())
+        # default should be true
+        assert result["options"]["sentry:replay_hydration_error_issues"] is True
+
+        self.project.update_option("sentry:replay_hydration_error_issues", False)
+        result = serialize(self.project, self.user, DetailedProjectSerializer())
+        assert result["options"]["sentry:replay_hydration_error_issues"] is False
```
cmanallen, line 758: This is good and should be kept.  But we also need to call the API endpoint that returns these options so we know its being returned properly.

cmanallen (approved):
Approved under the assumption more coverage was added.

PR #71413, opened 2024-05-22
ref(router): useNavigate learns to use LocationDescriptor

evanpurkhiser:
This hook needs to more closely resemble the browserHistory interface to
support a migration to using useNavigate.

### DIFF
--- a/static/app/utils/useNavigate.tsx
+++ b/static/app/utils/useNavigate.tsx
@@ -1,12 +1,19 @@
 import {useCallback, useEffect, useRef} from 'react';
+import type {LocationDescriptor} from 'history';
 
-import {useRouteContext} from 'sentry/utils/useRouteContext';
 import {normalizeUrl} from 'sentry/utils/withDomainRequired';
+
+import useRouter from './useRouter';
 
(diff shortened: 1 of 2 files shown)


davidenwang (approved):

static/app/utils/useNavigate.tsx
```diff
@@ -1,50 +1,65 @@
...
  * Returns an imperative method for changing the location. Used by `<Link>`s, but
  * may also be used by other elements to change the location.
  *
  * @see https://reactrouter.com/hooks/use-navigate
  */
 export function useNavigate() {
-  const route = useRouteContext();
-
-  const navigator = route.router;
+  const router = useRouter();
```
davidenwang, line 25: just checking but the idea here is for all browserHistory/useRouter usages be switched to useNavigate and then once we switch to react router 6 we move off this last usage of `useRouter`?

evanpurkhiser: basically yes. There will be a lot of shim code for react-router 6. We'll then have to migrate off of it.

PR #71415, opened 2024-05-22
feat(insights): add search bar to cache and queue sample panels

KevinL10:
Adds a span search bar to cache and queue sample panels.

Cache:
image

Queues:
image

### DIFF
--- a/static/app/views/performance/cache/samplePanel/samplePanel.tsx
+++ b/static/app/views/performance/cache/samplePanel/samplePanel.tsx
@@ -3,20 +3,24 @@ import styled from '@emotion/styled';
 import keyBy from 'lodash/keyBy';
 import * as qs from 'query-string';
 
+import Feature from 'sentry/components/acl/feature';
 import ProjectAvatar from 'sentry/components/avatar/projectAvatar';
 import {Button} from 'sentry/components/button';
 import {CompactSelect} from 'sentry/components/compactSelect';
+import SearchBar from 'sentry/components/events/searchBar';
 import Link from 'sentry/components/links/link';
 import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
 import {trackAnalytics} from 'sentry/utils/analytics';
 import {DurationUnit, RateUnit, SizeUnit} from 'sentry/utils/discover/fields';
+import {DiscoverDatasets} from 'sentry/utils/discover/types';
 import {PageAlertProvider} from 'sentry/utils/performance/contexts/pageAlert';
 import {decodeScalar} from 'sentry/utils/queryString';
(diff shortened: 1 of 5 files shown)


narsaynorath (approved):
Looks good to me, tested with the Queues module. It looks like there are some tests that are missing a mock for the tags request that `useSpanFieldSupportedTags` requires, but that should be an easy fix

static/app/views/performance/cache/samplePanel/samplePanel.tsx
```diff
@@ -362,6 +384,22 @@ export function CacheSamplePanel() {
             />
           </ModuleLayout.Half>
 
+          <ModuleLayout.Full>
+            <Feature features="performance-sample-panel-search">
```
gggritso, line 388: Would be slightly better to lift `<Feature>` above `ModuleLayout.Full` because an empty `ModuleLayout.Full` would still cause a gap in the layout

static/app/views/performance/queues/destinationSummary/messageSpanSamplesPanel.tsx
```diff
@@ -128,6 +137,13 @@ export function MessageSpanSamplesPanel() {
   sampleFilters.addFilterValue('transaction', query.transaction);
   sampleFilters.addFilterValue('messaging.destination.name', query.destination);
 
+  // filter by key-value filters specified in the search bar query
+  Object.entries(new MutableSearch(query.spanSearchQuery).filters).forEach(
+    ([key, value]) => {
+      sampleFilters.addFilterValues(key, value);
+    }
+  );
+
```
gggritso, line 146: 😭 understandable if there's no way around this, but yuck! More motivation to improve `MutableSearch`

static/app/views/performance/queues/destinationSummary/messageSpanSamplesPanel.tsx
```diff
@@ -333,6 +359,22 @@ export function MessageSpanSamplesPanel() {
             />
           </ModuleLayout.Full>
 
+          <ModuleLayout.Full>
+            <Feature features="performance-sample-panel-search">
```
gggritso, line 363: Same note about nesting `ModuleLayout.Full` inside `Feature` rather than the other way around

static/app/views/performance/queues/destinationSummary/messageSpanSamplesPanel.tsx
KevinL10: yea.. I was looking at changing `MutableSearch.addStringFilter` to support strings containing multiple subfilters, but we would have to add some gnarly parsing logic that seems quite error-prone. I'll look into improving this after I add in the rest of the search bars.

PR #71420, opened 2024-05-23
perf(crons): Attempt to improve performance of `organization_monitor_index_stats` api.

wedamija:
At the moment we're filtering on `monitor_id` and `monitor_environment_id`. There's no specific index on both of these, and so postgres ends up scanning multiple indexes, rather than either `monitor_id, date_added` or `monitor_environment_id, date_added`.

I also dropped the filter on status, which is unnecessary since we're filtering to all statuses.

This should improve things, but if not we can see if a `UNION` helps at all.

### DIFF
--- a/src/sentry/monitors/endpoints/organization_monitor_index_stats.py
+++ b/src/sentry/monitors/endpoints/organization_monitor_index_stats.py
@@ -17,6 +17,14 @@ from sentry.api.helpers.environments import get_environments
 from sentry.models.environment import Environment
 from sentry.monitors.models import CheckInStatus, Monitor, MonitorCheckIn, MonitorEnvironment
 
+TRACKED_STATUSES = [
+    CheckInStatus.IN_PROGRESS,
+    CheckInStatus.OK,
+    CheckInStatus.ERROR,
+    CheckInStatus.MISSED,
+    CheckInStatus.TIMEOUT,
+]
+
 
 def normalize_to_epoch(timestamp: datetime, seconds: int):
     """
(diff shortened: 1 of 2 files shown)


tests/sentry/monitors/endpoints/test_organization_monitor_index_stats.py
```diff
@@ -62,8 +63,8 @@ def setUp(self):
             status=CheckInStatus.TIMEOUT,
         )
 
-        self.add_checkin(self.monitor2, offset={"minutes": 1})
-        self.add_checkin(self.monitor2, offset={"minutes": 2})
+        self.add_checkin(self.monitor2, env=self.env_prod_2, offset={"minutes": 1})
+        self.add_checkin(self.monitor2, env=self.env_prod_2, offset={"minutes": 2})
```
wedamija, line 67: This was incorrectly assigning a monitor environment for another monitor to this monitor, which caused my changes to fail

src/sentry/monitors/endpoints/organization_monitor_index_stats.py
```diff
@@ -112,14 +112,14 @@ def get(self, request: Request, organization) -> Response:
...
-            monitor_id__in=monitor_map.keys(),
-            status__in=tracked_statuses,
             date_added__gte=args["start"],
             date_added__lt=args["end"],
         )
 
         if monitor_environment_map:
             check_ins = check_ins.filter(monitor_environment_id__in=monitor_environment_map.keys())
+        else:
+            check_ins = check_ins.filter(monitor_id__in=monitor_map.keys())
```
wedamija, line 122: Tbh I'm not sure we ever don't have a monitor env anymore?

evanpurkhiser (approved):
looks good

PR #71423, opened 2024-05-23
ref(metrics): move metrics api endpoints to separate files

shellmayr:
We want to have full visibility into the performance of the individual endpoints. Our current instrumentation only works on a per-module / per-file basis, therefore we need to move the endpoints into individual files for there to be endpoint-specific monitoring.


iamrajjoshi (approved):
lgtm

obostjancic (approved):

PR #71425, opened 2024-05-23
ref(seer grouping): Check rate limits before seer ingestion grouping call

lobsterkatie:
This adds both a similarity-service-wide rate limit check and a per-project rate limit check before calling the Seer similarity service during ingestion. The rate limits are based on two new options, `seer.similarity.global-rate-limit` and `seer.similarity.per-project-rate-limit`. 

Note to reviewers: Lacking good intuition about what would make for a reasonable limit, I copied the default values from the Seer severity rate limits. Happy to adjust them, though.

### DIFF
--- a/src/sentry/grouping/ingest/seer.py
+++ b/src/sentry/grouping/ingest/seer.py
@@ -2,6 +2,7 @@ import logging
 from dataclasses import asdict
 
 from sentry import features, options
+from sentry import ratelimits as ratelimiter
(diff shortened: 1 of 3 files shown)


jangjodi (commented):
lgtm, and the rate limits make sense to me. let's wait for Josh to review before merging though

JoshFerge (approved):
makes sense, and we can bump the rate limit numbers up once we roll out if deemed safe.

PR #71432, opened 2024-05-23
fix(metrics): Wrap onboarding panel buttons

ArthurKnaus:
Make the buttons wrap on smaller screens.

Screenshot 2024-05-23 at 16 21 32


matejminar (approved):
Thanks for fixing this!

PR #71433, opened 2024-05-23
chore(hybridcloud) Add service methods to integration service

markstory:
In #71257 I forgot to include the service method stubs. This means the new methods can't be called in a non-local silo mode.

Refs HC-1190


src/sentry/services/hybrid_cloud/integration/service.py
```diff
@@ -147,19 +149,32 @@ def get_organization_context(
 
     @rpc_method
     @abstractmethod
-    def get_organization_contexts(
```
mrduncan, line 150: I might be reading this wrong, is this safe to delete? I'm still seeing usages of this service method

markstory: No, this is wrong. I meant to add the new signatures only.

mrduncan (approved):

PR #71434, opened 2024-05-23
feat(insights): Move Insights modules to `/insights/` URL

gggritso:
If the `performance-insights` flag is present, direct all links to Insights modules to `/insights/`, and update the page titles and breadcrumbs accordingly.

**e.g.,**
Screenshot 2024-05-23 at 10 36 05 AM

The complicated logic creates a _somewhat_ uncomfortable intermediate state, so I'll need to remove this shortly after rollout.

### DIFF
--- a/static/app/views/performance/utils/useInsightsURL.tsx
+++ b/static/app/views/performance/utils/useInsightsURL.tsx
@@ -1,0 +1,25 @@
+import useOrganization from 'sentry/utils/useOrganization';
+import {ModuleName} from 'sentry/views/starfish/types';
+
+type ModuleNameStrings = `${ModuleName}`;
+type RoutableModuleNames = Exclude<ModuleNameStrings, '' | 'other'>;
+
+export function useInsightsURL(moduleName: RoutableModuleName) {
+  const builder = useInsightsURLBuilder();
+  return builder(moduleName);
+}
+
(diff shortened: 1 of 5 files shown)


static/app/views/performance/utils/useInsightsURL.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+type URLBuilder = (moduleName: RoutableModuleNames) => string;
+
+export function useInsightsURLBuilder(): URLBuilder {
+  const organization = useOrganization({allowNull: true}); // Some parts of the app, like the main sidebar, render even if the organization isn't available (during loading, or at all).
+
+  if (!organization) {
+    // If there isn't an organization, items that link to modules won't be visible, so this is a fallback just-in-case, and isn't trying too hard to be useful
+    return () => '';
+  }
```
AbhiPrasad, line 20: Do we need this conditional? Isn't the fallback behaviour on lines 26-28 correct even if organization is not defined?

AbhiPrasad (approved):

static/app/views/performance/utils/useInsightsURL.tsx
gggritso: True! I copy-pasted this from `useModuleURL`! Removed

PR #71437, opened 2024-05-23
build(ui): Speed up hot reloading file resolution in getsentry

scttcper:
enable [resolve.cache](https://webpack.js.org/configuration/resolve/#resolvecache) and resolve.preferAbsolute in webpack and arrange the extensions so we try .js and .tsx first. This seems to mostly affect getsentry, possibly because it has two node_modules folders to look in.

When updating a file there's this huge "watch" process that seems to iterate all of the files and try to resolve them again (couldn't figure out how to get it to not do that).

before watch is 12s
image

after watch is a normal 170ms
image

preferAbsolute seems to reduce the files that are attempted during resolution and the cache does the rest of the work

without preferAbsolute
Screenshot 2024-05-22 at 10 22 15 PM

with preferAbsolute and the new file extension ordering
Screenshot 2024-05-22 at 10 18 30 PM

Measured using this command and waiting for the page to fully load, then changing some text in a file.

```
SENTRY_UI_DEV_ONLY=1 SENTRY_WEBPACK_PROXY_PORT=7999 node --inspect-brk ./node_modules/.bin/webpack serve
```

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -463,8 +463,12 @@ const appConfig: Configuration = {
       'process/browser': require.resolve('process/browser'),
     },
 
+    // Might be an issue if changing file extensions during development
+    cache: true,
+    // Prefers local modules over node_modules
+    preferAbsolute: true,
     modules: ['node_modules'],
-    extensions: ['.js', '.json', '.ts', '.tsx', '.less'],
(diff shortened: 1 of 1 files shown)


billyvg (approved):
v nice

PR #71438, opened 2024-05-23
feat(api): Added `include_feature_flags` Query Param for Organization Details

iamrajjoshi:
In this pr I introduce a new query param `include_feature_flags`, which by default will be False. Eventually, api calls will have to pass `include_feature_flags=1` in order to get all the features of the org. 

There will be 2 follow-up prs, one to update the frontend to use this query param, and one to actually implement the logic to using the query param.

Additionally, the `DELETE` method will be updated so it never returns flags.

For: https://github.com/getsentry/team-core-product-foundations/issues/326

### DIFF
--- a/src/sentry/api/endpoints/organization_details.py
+++ b/src/sentry/api/endpoints/organization_details.py
@@ -689,9 +689,6 @@ class OrganizationDetailsEndpoint(OrganizationEndpoint):
         """
         This method exists as a way for getsentry to override this endpoint with less duplication.
         """
-        # This param will be used to determine if we should include feature flags in the response
-        include_feature_flags = request.GET.get("include_feature_flags", "0") != "0"
-        (include_feature_flags)
 
         if not request.user.is_authenticated:
             return self.respond({"detail": ERR_NO_USER}, status=401)
--- a/src/sentry/api/endpoints/organization_details.py
+++ b/src/sentry/api/endpoints/organization_details.py
@@ -576,6 +576,10 @@ class OrganizationDetailsEndpoint(OrganizationEndpoint):
         :param string detailed: Specify '0' to retrieve details without projects and teams.
         :auth: required
         """
+        # This param will be used to determine if we should include feature flags in the response
+        include_feature_flags = request.GET.get("include_feature_flags", "0") != "0"
+        (include_feature_flags)
+
         serializer = org_serializers.OrganizationSerializer
 
         if request.access.has_scope("org:read") or is_active_staff(request):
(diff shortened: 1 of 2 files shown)


src/sentry/api/endpoints/organization_details.py
```diff
@@ -576,6 +576,10 @@ def get(self, request: Request, organization) -> Response:
         :param string detailed: Specify '0' to retrieve details without projects and teams.
         :auth: required
         """
+        # This param will be used to determine if we should include feature flags in the response
+        include_feature_flags = request.GET.get("include_feature_flags", "0") != "0"
```
iamrajjoshi, line 580: we will only include ff if `include_feature_flags=1`

src/sentry/api/endpoints/organization_details.py
```diff
@@ -606,6 +610,10 @@ def put(self, request: Request, organization) -> Response:
         """
         from sentry import features
 
+        # This param will be used to determine if we should include feature flags in the response
+        include_feature_flags = request.GET.get("include_feature_flags", "0") != "0"
+        (include_feature_flags)
```
ykamo001, line 615: is this a typo?

src/sentry/api/endpoints/organization_details.py
ykamo001: I don't know if that's the right attribute to use here. From reading that, it would make sense to use that for a adjacent domain object, not additional data for the current domain object.

src/sentry/api/endpoints/organization_details.py
iamrajjoshi: i currently don't use the flag because i need to first update calls to the endpoint after this. to prevent linting errors, i added that temporarily, probably should put a comment about it.

ykamo001: ahhh got it! I think you could also just do ` _ = some_call()`

src/sentry/api/endpoints/organization_details.py
iamrajjoshi: hmm, that also makes sense, what do you think @sentaur-athena?

sentaur-athena: yes @ykamo001 is right. It's a weird case with feature flags because they're not real objects that we're serializing. Please ignore my comment.

src/sentry/api/endpoints/organization_details.py
sentaur-athena: I would add a TODO here to avoid temporary confusion

sentaur-athena (approved):

ykamo001 (approved):

src/sentry/api/endpoints/organization_details.py
ykamo001: +1

PR #71447, opened 2024-05-23
ref: upgrade pycountry

asottile-sentry:
this removes one of the last things import pkg_resources


joshuarli (approved):

asottile-sentry:
this breaks getsentry so I'm looking to replace it instead

PR #71452, opened 2024-05-23
fix(backup): Manually drain region org outboxes

azaslavsky:


src/sentry/backup/imports.py
```diff
@@ -435,6 +440,47 @@ def resolve_org_slugs_from_pk_map(pk_map: PrimaryKeyMap):
...
+            # massive backlog of `OrganizationMemberTeam` outboxes that need to clear before this
+            # region can respond. In the worst case scenario, this will result in a timeout of the
+            # original, outer `bulk_create...` call.
+            #
+            # So, the solution we take here is to manually clear all `ORGANIZATION_SCOPE` outboxes
+            # for each imported organization on this side first, so that when this region responds
+            # to the call triggered from the server-side of `bulk_create...`, the
+            # `ORGANIZATION`-scoped outbox queue is empty and is free to only serve requests
+            # specifically related to slug provisioning.
+            for id in slug_mapping.keys():
```
GabeVillalobos, line 458: Is there potential for the slug mapping to contain more than 1 organization ID? If so, wouldn't the break below result in us skipping some organizations?

azaslavsky: If we import multiple orgs at a time (unlikely, though users are allowed to do this on ex: self-hosted), we'll have one entry per org imported. But the `break` should only apply to the inner loop, right?

GabeVillalobos: oh whoops, yes. My bad 😅

GabeVillalobos (approved):

PR #71471, opened 2024-05-23
feat(performance): Updates webvitals to aggregate by avg instead of p75 controlled by feature flag

edwardgou-sentry:
Updates the Web Vitals UI to aggregate web vital values by `avg` instead of `p75` controlled by feature flag.


edwardgou-sentry:
no longer needed

PR #71474, opened 2024-05-23
fix(tags): Allow filtering by both IPv4 and IPv6 tags on EventUsers

### LINKED ISSUE #71461: Unable to load Tag Details for some Issues
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

Select 'Tags' for a particular Issue. This affects some but not all.

### Expected Result

Tag Details should appear.

### Actual Result

Tags fail to load:
image

Seems Inconsistent Handling of IPv6 Addresses blocks users from viewing Tags in Issue Stream, per initial triage.

### Product Area

Issues

### Link

_No response_

### DSN

_No response_

### Version

_No response_

leeandher:
Resolves https://github.com/getsentry/sentry/issues/71461

Prior, when filtering by IP, regardless of which type the address was, we'd apply an IPv4 and IPv6 filter (`toIPv4` and `toIPv6`). For v4s this works fine since the `toIPv4` resolves and `toIPv6` allows for passing v4s. When passing in v6s though, `toIPv4` fails loudly and returns a 500 from the API.

Now, we'll just create two conditions (v4 and v6) and add in all the valid IPs for each form of address after validating them as one or the other. The code was also a bit confusing so it got a heavy refactor, and a few test changes.

### DIFF
--- a/src/sentry/utils/eventuser.py
+++ b/src/sentry/utils/eventuser.py
@@ -39,7 +39,7 @@ REFERRER = "sentry.utils.eventuser"
 
 SNUBA_KEYWORD_SET = {"id", "username", "email", "ip"}
 # Keyword 'ip' is a special case since we need to handle IPv4/IPv6 differently
-SNUBA_KEYWORD_COLUMN_MAP: Mapping[str, str | None] = {
+SNUBA_KEYWORD_COLUMN_MAP = {
     "id": "user_id",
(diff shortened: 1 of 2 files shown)


JoshFerge:
this has annoyed me in the past :D great stuff

wedamija (approved):

PR #71483, opened 2024-05-23
ref(router): Remove search params from discover table links

evanpurkhiser:
React router 6 doesn't let you put query parameters into the pathname


ryan953 (approved):
d'oh

PR #71486, opened 2024-05-24
chore(options): Remove unused options

jan-auer:
These options are no longer used in code and have all been removed from options
automator. To make the list of available options more manageable, we remove the
option from code.


tests/sentry/sentry_metrics/test_gen_metrics_multiprocess_steps.py
```diff
@@ -408,12 +408,7 @@ def test_process_messages_default_card_rollout(set_sentry_option: Callable[...,
 
     outer_message = Message(Value(message_batch, last.committable))
 
-    with set_sentry_option(
-        "sentry-metrics.cardinality-limiter.orgs-rollout-rate",
-        1.0,
```
jan-auer, line 413: 1.0 was the default, but it was also not used.

Dav1dde (approved):

tests/sentry/sentry_metrics/test_gen_metrics_multiprocess_steps.py
ayirr7: IIRC there was a test that used this option? Should we modify that test accordingly?

jan-auer: This is the only test and it set this option to the default. I suppose this change is all we need?

src/sentry/options/defaults.py
```diff
@@ -1292,27 +1280,11 @@
...
-)
-
 # Option to enable orjson for JSON parsing in reconstruct_messages function
 register(
     "sentry-metrics.indexer.reconstruct.enable-orjson", default=0.0, flags=FLAG_AUTOMATOR_MODIFIABLE
 )
 
-# Option to enable direct storage queries for meta queries in the metrics layer
-register(
-    "sentry-metrics.metrics-layer.use-storage-direct-meta-queries",
```
evanh, line 1311: @enochtangg can this be removed?

enochtangg: 👍

enochtangg (approved):

PR #71494, opened 2024-05-24
Store small `EventAttachment` contents inline

Swatinem:
Here, "inline" means storing contents in the `blob_path` field.

ASCII-only contents up to 192 bytes are stored there, prefixed by `:` as marker.

### DIFF
--- a/src/sentry/models/eventattachment.py
+++ b/src/sentry/models/eventattachment.py
@@ -17,6 +17,7 @@ from sentry.attachments.base import CachedAttachment
 from sentry.backup.scopes import RelocationScope
 from sentry.db.models import BoundedBigIntegerField, Model, region_silo_model, sane_repr
 from sentry.db.models.fields.bounded import BoundedIntegerField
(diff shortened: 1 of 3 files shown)


loewenheim (approved):

src/sentry/models/eventattachment.py
```diff
@@ -152,22 +170,27 @@ def putfile(cls, project_id: int, attachment: CachedAttachment) -> PutfileResult
...
         store_blobs = True
 
         if store_blobs:
             size, checksum = get_size_and_checksum(blob)
-            blob_path = "eventattachments/v1/" + FileBlob.generate_unique_path()
 
-            storage = get_storage()
-            compressed_blob = BytesIO(zstandard.compress(attachment.data))
-            storage.save(blob_path, compressed_blob)
+            if can_store_inline(data) and in_random_rollout("eventattachments.store-small-inline"):
```
anonrig, line 186: I think the order should be changed for improving performance

PR #71502, opened 2024-05-24
ref(insights): Add platform selector hook to screenloads

narsaynorath:
There's a lot of code around that's specific for determining if a single project is selected and what platform that project is. I'm adding the platform selector to app starts and before doing that I wanted to extract some of that code to simplify the components more. This is in preparation to add the platform selector to App Starts.

This PR:
- Adds a `usePlatformSelector` hook that returns the project if there's a single selection, the platform, whether it's cross platform, and the platform selector component
- Removes a lot of duplicated code and avoids prop drilling project down through the components until it's necessary to do the query.

I also removed any cases of `organization.features.includes('spans-first-ui');` because these modules should already be gated by that feature flag.


static/app/views/performance/mobile/usePlatformSelector.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const {selection} = usePageFilters();
+  const {projects} = useProjects();
+  const location = useLocation();
+
+  const project = useMemo(() => {
+    if (selection.projects.length !== 1) {
+      return null;
+    }
+
+    return projects.find(p => p.id === String(selection.projects));
```
edwardgou-sentry, line 24:
```suggestion
    return projects.find(p => p.id === String(selection.projects)) ?? null;
```

nit: unless we handle `undefined` vs `null` project differently where `usePlatformSelector` is called, I feel like it might be cleaner if we only return either `undefined` or `null`.

static/app/views/performance/mobile/usePlatformSelector.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const selectedPlatform =
+    decodeScalar(location.query[PLATFORM_QUERY_PARAM]) ??
+    localStorage.getItem(PLATFORM_LOCAL_STORAGE_KEY) ??
+    DEFAULT_PLATFORM;
+
+  return {
+    project,
+    selectedPlatform,
+    isProjectCrossPlatform,
+    PlatformSelector,
```
edwardgou-sentry, line 41: question: how come we return `PlatformSelector` here? Doesn't seem like we make any changes to the component here, so I think we can make this hook lighter by expecting users to import this component on their own.

static/app/views/performance/mobile/usePlatformSelector.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const project = useMemo(() => {
+    if (selection.projects.length !== 1) {
+      return null;
+    }
+
+    return projects.find(p => p.id === String(selection.projects));
+  }, [projects, selection.projects]);
+
+  const isProjectCrossPlatform = useMemo(
+    () => (project ? isCrossPlatform(project) : false),
```
edwardgou-sentry, line 28:
```suggestion
    () => !!(project && isCrossPlatform(project)),
```
nit: similar feedback, if we don't have expect logic on `undefined` values, maybe we can just force this too a boolean

edwardgou-sentry (approved):
looks reasonable to me! just left some nits and a question

static/app/views/performance/mobile/usePlatformSelector.tsx
narsaynorath: Yeah, I thought this was a bit odd but I figured it was okay since this hook was also returning the selected platform, so it made sense to return the component that modified the selection. I looked it up and it's supposed to be pretty anti-pattern-y because it doesn't allow for flexibility so I took it out and renamed the hook to `useCrossPlatformProject`. It's not the best name but I'm unsure of what to call it for now

PR #71503, opened 2024-05-24
feat(related_issues): Use new component instead of EventItem

armenzg:
This change adds a new UI component to render related issues.

Changes:
- Do not show the two issues involved but only one (the one to be visited)
- Move `View Full Trace` link within the `groupEventHeader` component
- Display the issue's title vs the event's title
  - A follow-up PR is needed to speed up fetching the title and subtitle.

We temporarily used `EventItem` to deploy something quickly for feedback (see #71049). Visiting [this local event](https://dev.getsentry.net:7999/issues/4208426984/events/abef1576630343a496798c9250fe69db/?project=1&referrer=issues_trace_issue) via `yarn dev-ui` allows reproducing the screenshots below.

This changes the UI from this:
image
to this:
image

Special attention has been paid to the following:

* Make sure the elements clip nicely and line up with containers around it
* Made the title and subtitle copy matches what an issue would show in the issue details page

This screenshot shows the values that an issue displays:
image

The expected design:
image


static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -0,0 +1,127 @@
...
+  &:hover {
+    background-color: ${p => p.theme.surface200};
+    color: ${p => p.theme.textColor};
+  }
+`;
+
+// The padding-right prevents overflowing the container
+// This can be changed with more CSS expertise
+const TraceIssueDetailsContainer = styled('div')`
+  padding-right: ${space(4)};
```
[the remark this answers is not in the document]
malwilley: If you add `overflow: hidden` here you'll get the truncation you want

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -0,0 +1,127 @@
...
+`;
+
+const NoOverflowDiv = styled('div')`
+  white-space: nowrap;
```
malwilley, line 115: We have a util you can replace these styles with: `${p => p.theme.overflowEllipsis};`

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -161,34 +162,38 @@ describe('TraceTimeline', () => {
...
       match: [MockApiClient.matchQuery({dataset: 'discover'})],
     });
     // I believe the call to projects is to determine what projects a user belongs to
     MockApiClient.addMockResponse({
       url: `/organizations/${organization.slug}/projects/`,
       body: [],
     });
+    // We need the metadata from the issue in order to render the related issue
+    MockApiClient.addMockResponse({
+      url: `/organizations/${organization.slug}/issues/1000/`,
```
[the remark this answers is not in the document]
malwilley: I would love to avoid an extra request here if possible to avoid adding more latency. You might be able to add the `message` field to the discover `/events/` call? If so you should have all the necessary information to render this new design.

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+import type {TimelineEvent} from './useTraceTimelineEvents';
+
+interface TraceIssueEventProps {
+  event: TimelineEvent;
+}
+
+export function TraceIssueEvent({event}: TraceIssueEventProps) {
+  const organization = useOrganization();
+  const {projects} = useProjects({
```
malwilley, line 23: There is a `useProjectFromSlug` hook you could use here instead, but it doesn't matter much since it does the same thing

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -0,0 +1,127 @@
...
+            onClick={() => {
+              trackAnalytics('issue_details.issue_tab.trace_issue_clicked', {
+                organization,
+                event_id: event.id,
+                group_id: issueId,
+              });
+            }}
+          >
+            {project && (
+              <ProjectBadge
```
malwilley, line 68: When `project` does not exist, can you render a placeholder in its place of the same size (`<Placeholder />`)? That way it won't pop in if it needs to be fetched

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -0,0 +1,127 @@
...
+        groupData && (
+          <TraceIssueLinkContainer
+            to={{
+              pathname: `/organizations/${organization.slug}/issues/${issueId}/events/${event.id}/`,
+              query: {
+                referrer: 'issues_trace_issue',
+              },
+            }}
+            onClick={() => {
+              trackAnalytics('issue_details.issue_tab.trace_issue_clicked', {
```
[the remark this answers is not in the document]
malwilley: Yeah I like issue_details.trace_issue_clicked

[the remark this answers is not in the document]
malwilley: Although, as I'm thinking about this more, you may want to use the same key as the current analytic event (the tooltip click event) and add a property that differentiates it. Whether you use the same event or separate events depends on how you want to analyze the data.

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
armenzg: Added locally.

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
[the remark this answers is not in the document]
armenzg: Maybe I can include `groupMetadata` per event. That way we can even fix events in the trace timeline and the new trace view.

static/app/views/issueDetails/groupEventHeader.tsx
```diff
@@ -18,10 +22,27 @@ type GroupEventHeaderProps = {
...
+  // This is also called within the TraceTimeline component but caching will save a second call
+  const {isError, isLoading, oneOtherIssueEvent} = useTraceTimelineEvents({
+    event,
+  });
+  const readyToShow = !isLoading && !isError;
 
   return (
     <StyledDataSection>
       <GroupEventCarousel group={group} event={event} projectSlug={project.slug} />
+      {isRelatedIssuesEnabled && readyToShow && oneOtherIssueEvent === undefined && (
```
armenzg, line 37: When `oneOtherIssueEvent === undefined` is true, we have the normal case where we want to just display the tracelink and potentially the timeline.

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -0,0 +1,136 @@
...
+  event: TimelineEvent;
+}
+
+export function TraceIssueEvent({event}: TraceIssueEventProps) {
+  const organization = useOrganization();
+  const project = useProjectFromSlug({organization, projectSlug: event['project.name']});
+  const issueId = event['issue.id'];
+  // We are view issue X and loading data for issue Y. This means that this call will not be cached,
+  // thus, it is a little bit slow to render the component.
+  // XXX: Create new endpoint that only fetches the metadata for the group
```
armenzg, line 27: I may be able to include the data within the `events` endpoint and save this call altogether.

malwilley (approved):

PR #71527, opened 2024-05-27
fix(metrics-extraction): Fix release selection with on-demand

k-fish:
### Summary
Releases are high cardinality and aren't currently emitted into the tags in on-demand. For now if we detect release selection via conditions we'll opt the widget out. Release conditions should be a provider on the page similar to PageFilters in the future, to disambiguate between conditions saved into the query and conditions coming from controls.


wmak (approved):
slap

PR #71531, opened 2024-05-27
feat(replay-new-trace): Rendering new Trace View under the replays .

Abdkhan14:
- The feature exists under the feature flag: [organizations:replay-trace-view-v1](https://flagr.getsentry.net/#/flags/615)
- Most the functionality aligns with the new trace view in performance, the UI differences include:
- - We adjust trace bounds using replay duration.
- - Replay trace drawer is bottom only.
- Added new analytics and preferences for replay traceview.


static/app/views/performance/newTraceDetails/index.tsx
```diff
@@ -1,5 +1,7 @@
 import type React from 'react';
 import {
+  type Dispatch,
```
evanpurkhiser, line 3: nit: No need to import react types, you can just use the global `React.Dispatch`

jest.config.ts
```diff
@@ -203,7 +203,7 @@ if (
  * node_modules, but some packages which use ES6 syntax only NEED to be
  * transformed.
  */
-const ESM_NODE_MODULES = [];
+const ESM_NODE_MODULES = ['screenfull'];
```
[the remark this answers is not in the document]
scttcper: maybe just mock this module since we're unlikely to be able to test that the functionality does anything

static/app/views/performance/newTraceDetails/traceApi/useReplayTraceMeta.tsx
```diff
@@ -0,0 +1,91 @@
...
+    return {
+      data: meta.data,
+      isLoading: eventsIsLoading || meta.isLoading,
+      isRefetching: meta.isRefetching || eventsIsRefetching,
+      refetch: () => {
+        meta.refetch();
+        eventsRefetch();
+      },
+    };
+  }, [meta, eventsIsLoading, eventsIsRefetching, eventsRefetch]);
```
scttcper, line 88: you can reduce this to one useMemo since only the metaResults are returned

static/app/views/performance/newTraceDetails/traceApi/useTraceMeta.tsx
```diff
@@ -57,18 +86,54 @@ export function useTraceMeta(
...
+      getTraceMetaQueryKey(traceSlug, organization.slug, queryParams)
+    );
+  }, [traceSlugs, organization, queryParams]);
+
+  const results = useApiQueries<TraceMeta>(queryKeys, {
+    enabled: traceSlugs.length > 0 && !!organization.slug && mode !== 'demo',
+    staleTime: Infinity,
+  });
+
+  const {data, isLoading, isRefetching, refetch} = useMemo(() => {
```
scttcper, line 101: would avoid destructing just to make a new object again in the return

static/app/views/performance/newTraceDetails/replayTraceView.tsx
```diff
@@ -0,0 +1,73 @@
...
+      />
+    </Wrapper>
+  );
+}
+
+const Wrapper = styled('div')`
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  padding: 1px;
```
ryan953, line 72: This 1px padding is going to cause alignment issues on the page (there's another alignment issue on the page already too, that affects all tabs, fixed soon)

I notice in here there is a box-shadow, if it were a border then things seem like they would render better, in this page anyway: https://github.com/getsentry/sentry/blob/fb68f763b2cb40107b8a43111d71a0dc17bcc721/static/app/views/performance/newTraceDetails/index.tsx#L995-L996

jest.config.ts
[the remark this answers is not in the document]
Abdkhan14: Yeah, I will just mock the module here.

static/app/views/performance/newTraceDetails/traceApi/useReplayTraceMeta.tsx
scttcper: oh i see its useTraceMeta

static/app/views/performance/newTraceDetails/replayTraceView.tsx
Abdkhan14: Agree, I'll make the change thanks @ryan953

static/app/views/performance/newTraceDetails/traceApi/useReplayTraceMeta.tsx
Abdkhan14: Heyo @scttcper, I don't think we can reduce this to one `useMemo`, since we have the fetching happening in `useTraceMeta`.

static/app/views/performance/newTraceDetails/index.tsx
```diff
@@ -192,7 +201,7 @@ export function TraceView() {
   }, [queryParams, traceSlug]);
 
   const trace = useTrace();
-  const meta = useTraceMeta();
+  const meta = useTraceMeta([traceSlug]);
```
JonasBa, line 204: Why do we need an array if there is only one trace slug?

static/app/views/performance/newTraceDetails/index.tsx
```diff
@@ -306,12 +318,19 @@ function TraceViewContent(props: TraceViewContentProps) {
       return loadingTrace;
     }
 
-    if (props.trace) {
-      return TraceTree.FromTrace(props.trace);
+    if (props.trace && props.status === 'success') {
```
JonasBa, line 321: Why do we need to check for status here?

static/app/views/replays/detail/trace/trace.tsx
```diff
@@ -147,4 +172,18 @@ const OverflowScrollBorderedSection = styled(BorderedSection)`
   }
 `;
 
+function getTraceStatus(traceState: ExternalState) {
```
JonasBa, line 175: Only used in a single place it seems, I would inline it.

JonasBa (changes requested):
My main issue with changes like these is that long term, this will lead to unmaintainable code. It would not be unreasonable to think that we will be adding the trace view to other parts of the product in the future. With an approach like this, we are starting to have nested conditions deep inside our components that are becoming interconnected (example trace meta merging). This will inevitably result in brittle code that will be harder and harder to modify as more changes are added. Instead, I would suggest a different path here of making this UI configurable outside of having to define a `source`. Since we are early in the project development lifecycle, this is easier to do now than later on and I doubt that it will require us to make many changes (we can just continue on this PR). 

Long story short, my opinion is that the source should not exist. If we need to, lets hoist the data fetching and configuration initialization which will decouple the data model from rendering (we already to this in the form of hooks, we just need to lift it a level above), and then supply the view components with the right data and the configuration to behave properly no matter where in the product they are rendering

static/app/views/performance/newTraceDetails/index.tsx
Abdkhan14: A replay has multiple traceIds, so I made useTraceMeta able to handle multiple traceIds.

static/app/views/performance/newTraceDetails/index.tsx
Abdkhan14: I'll remove this, it was part of a debugging effort.

static/app/views/replays/detail/trace/trace.tsx
Abdkhan14: Agree.

JonasBa (approved):

PR #71532, opened 2024-05-27
fix(spans): Free text search on spans should use globs

Zylphrex:
Free text search was previously generating exact match conditions. This changes the behaviour to generate glob conditions so we get better matches for these free text searches. One caveat here is that we should expect free text search to be slow on the spans dataset.

### DIFF
--- a/src/sentry/search/events/builder/spans_indexed.py
+++ b/src/sentry/search/events/builder/spans_indexed.py
@@ -8,7 +8,6 @@ from sentry.search.events.types import SelectType
 
 class SpansIndexedQueryBuilder(QueryBuilder):
     requires_organization_condition = False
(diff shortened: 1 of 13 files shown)


src/sentry/search/events/builder/spans_indexed.py
```diff
@@ -8,7 +8,6 @@
 
 class SpansIndexedQueryBuilder(QueryBuilder):
     requires_organization_condition = False
-    free_text_key = "span.description"
```
Zylphrex, line 11: Not too sure if i should remove this from the `spans_metrics` dataset or not as well.

wmak (approved):

wmak (approved):
lgtm, ty for making this work with metrics too

PR #71538, opened 2024-05-27
ref(insights): create useSpansIndexed

DominikB2014:
1. Create `useSpansIndexed` in `useDiscover.ts` (for type safety and consistency)
2. Remove `useIndexedSpans` and replace with `useSpansIndexed`
3. Move tests for `useIndexedSpans.spec.ts` to `useDiscover.spec.ts`

I think we should rename `SpanIndexedField` -> `SpansIndexedField`, `SpanIndexedResponse` -> `SpansIndexedResponse`, etc (to directly match the datasets, but that can be another PR.


static/app/views/starfish/components/spanDescription.tsx
```diff
@@ -37,17 +37,18 @@ export function DatabaseSpanDescription({
   groupId,
   preliminaryDescription,
 }: Omit<Props, 'op'>) {
-  const {data: indexedSpans, isFetching: areIndexedSpansLoading} = useIndexedSpans({
-    search: MutableSearch.fromQueryObject({'span.group': groupId}),
-    sorts: [INDEXED_SPAN_SORT],
```
DominikB2014, line 42: @gggritso this is a weird one, seems like this sort was never being applied before? Without removing this sort, the test would fail because the test expected no sort in the query object

https://github.com/getsentry/sentry/blob/40ee1fe571af1e543a3216cacc420b4d42372aa1/static/app/views/performance/database/databaseSpanSummaryPage.spec.tsx#L179

edwardgou-sentry (approved):

static/app/views/starfish/components/spanDescription.tsx
edwardgou-sentry: odd, we can probably add the sort in a separate pr.

PR #71539, opened 2024-05-27
fix(insights): Fix horizontal alignment of sidebar feature badge against the power icon

gggritso:
**Before (Jank):**
Screenshot 2024-05-27 at 4 50 45 PM
Screenshot 2024-05-27 at 4 52 42 PM

**After (Class):**
Screenshot 2024-05-27 at 4 50 55 PM
Screenshot 2024-05-27 at 4 52 32 PM

I'll fix vertical alignment separately, that has to happen in `getsentry`


billyvg (approved):
Don't love that the beta tags are misaligned when the business icon is present, but that's a separate issue

PR #71540, opened 2024-05-27
feat(discover): Expose dataset in discover saved query api

shruthilayaj:
Expose dataset field to discover saved query API.
All saved queries currently query the `discover` dataset,
so while the dataset selector feature is still under development,
default dataset to `discover`


wmak (approved):
lgtm

Issue #57518, opened 2023-10-04
MN+1 Issues are not included in the detector threshold settings


nanoburd:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

There is an issue with N+1 query threshold and detectors: a customer set the threshold to 500ms, but this event was raised in his issue page. The total duration of the transaction is 202.27ms, how can the db spans be more than 500ms?
[ISSUE](https://dashfuel.sentry.io/issues/4126710681/events/latest/?environment=production&project=6511968&referrer=latest-event)
[TRANSACTION](https://dashfuel.sentry.io/performance/python-django:c1fcec5babbb45b58a88bc6a6541b035/)
[SETTINGS](https://dashfuel.sentry.io/settings/projects/python-django/performance/)

### Expected Result

The total span duration for MN+1 to adhere to the N+1 Duration Threshold.

### Actual Result

Right now the MN+1 is hardcoded to 100 ms. So its unaffected by the N+1 Threshold Setting

Decision
1) Unify both N+1 and MN+1 total span duration threshold
2) Change total duration to look at all span ops, not just db span ops.
3) for MN+1, ensure at least 10% of total span duration comes from db op spans

### Product Area

Performance

nanoburd:
@brentc FYI

nanoburd:
@Abdkhan14 can you take a look at this?

bcoe:
Closing as stale. We're in the process of overhauling our alerting system, so we'll have new and exciting problems.

PR #71544, opened 2024-05-28
feat(performance-detetector-thresholds): Updated mn plus one detection criteria. 

Abdkhan14:
The decision to make the chages were made in this gh issue: https://github.com/getsentry/sentry/issues/57518

Updates: 
- Use N+1 DB total duration threshold for MN+1 DB issue.
- Consider all spans and not just db spans when calculating total offending spans duration.
- Ensure that db spans make up at least 10% of the total duration of offending spans.


mjq-sentry (approved):
LGTM :+1: Thanks!

PR #71546, opened 2024-05-28
ref(billing): Adds options for release dates for new categories

corps:
Adds options to support https://github.com/getsentry/getsentry/pull/14089 for release dates backed data categories.


corps:
I'm breaking this PR out into separate pieces moving forward, this is for reference.

PR #71551, opened 2024-05-28
fix(security): hide integrations code from members, attempt 2

oioki:
Another attempt of https://github.com/getsentry/sentry/pull/70128
This also adds the missing access parameter to the serializer call in `post()` method.


mdtro (approved):

sentaur-athena:
Thank you @oioki for fixing the security issue.

sentaur-athena (approved):

Issue #71509, opened 2024-05-27
Clean up metric hours in Stats


obostjancic:
We want to have toggle in Stats that will switch between metric hours and metric volume.
Only volume can show negative outcomes.
```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/sentry/pull/71552
```

PR #71552, opened 2024-05-28
feat(stats): metric hours

obostjancic:
- Part of: #71509 
- adds a way to query metric hours from stats endpoint
- moves metric queries from `outcomes.py` to `metrics/stats.py`


src/sentry/sentry_metrics/querying/stats.py
```diff
@@ -0,0 +1,137 @@
...
+    environments: Sequence[Environment],
+):
+
+    start, end = get_date_range_from_params(query)
+    group_by = query.getlist("groupBy", [])
+    interval = parse_stats_period(query.get("interval", "1h"))
+
+    category = query.get("category")
+    # TODO(metrics): remove metrics and metricsSeconds after FE is updated
+    if category == "metrics" or category == "metricSecond" or category == "metricsOutcomes":
```
shellmayr, line 117: metricsOutcomes is not called in the endpoint - is this on purpose?

shellmayr (approved):

src/sentry/sentry_metrics/querying/stats.py
obostjancic: good catch! thx!

PR #71559, opened 2024-05-28
fix(trace): hide trace multiple root warning for js traces with multiple roots

JonasBa:
Removes multiple roots warning for JS traces with multiple roots. A backend change to add SDK to the response needs to be added and the detection needs to be extended to include all other browser sdks


Lms24 (approved):
Thanks for making the change!

PR #71563, opened 2024-05-28
fix(feedback/issues): check if latest_event is None before filtering attachments

michellewzhang:
fixes SENTRY-39PV


JoshFerge (dismissed):
could we add a test for this?

michellewzhang:
> could we add a test for this?

added a test @JoshFerge

cmanallen (approved):

PR #71571, opened 2024-05-28
feat(similarity): Add delete record by hash endpoint

jangjodi:
Add function to call [seer delete grouping records by hash endpoint](https://github.com/getsentry/seer/pull/735)

Cleanup items:
- Rename `backfill.py` to `grouping_records.py` to better express its purpose
- Rename existing delete grouping records functions/variables to include the word `project` to distinguish between deletion of the entire project and deletion by hash


src/sentry/api/helpers/group_index/delete.py
```diff
@@ -45,13 +46,18 @@ def delete_group_list(
     eventstream_state = eventstream.backend.start_delete_groups(project.id, group_ids)
     transaction_id = uuid4().hex
 
+    all_group_hashes = GroupHash.objects.filter(project_id=project.id, group__id__in=group_ids)
+
+    # Tell seer to delete grouping records with these group hashes
+    if features.has("projects:similarity-embeddings-delete-by-hash", project):
+        # TODO (jangjodi): once we store seer grouping info in GroupHash, we should filter by that here
+        delete_seer_grouping_records_by_hash.apply_async(
+            args=[project.id, [group_hash.hash for group_hash in all_group_hashes], 0]
```
jangjodi, line 55: TODO: add countdown

jangjodi:
Another concern is how often this will run - this [query](https://redash.getsentry.net/queries/6828/source?p_today=d_now) suggests the function that calls the task can be run over 150 times/second with up to [1000](https://github.com/getsentry/sentry/blob/b9c389ee2fc90127fa2fb7e0d4931d4045d3d859/src/sentry/api/helpers/group_index/__init__.py#L11) groups in the list. The task will batch by 20, but it looks like this could happen a lot

trillville:
> Another concern is how often this will run - this [query](https://redash.getsentry.net/queries/6828/source?p_today=d_now) suggests 150 groups/second are deleted, where the max is [1000](https://github.com/getsentry/sentry/blob/b9c389ee2fc90127fa2fb7e0d4931d4045d3d859/src/sentry/api/helpers/group_index/__init__.py#L11) groups in the list. The task will batch by 20, but it looks like this could happen a lot

FWIW unless I am missing something, it looks like there are on avg 0.5 deletes/second overall (above query is sorted by count so its showing the spikes). Based on this I would not be too worried, especially if we are batching

src/sentry/api/helpers/group_index/delete.py
JoshFerge: i don't think so, this is fine. we will be bound by our number of celery workers

src/sentry/api/helpers/group_index/delete.py
```diff
@@ -45,13 +46,18 @@ def delete_group_list(
...
+
     # We do not want to delete split hashes as they are necessary for keeping groups... split.
-    GroupHash.objects.filter(
-        project_id=project.id, group__id__in=group_ids, state=GroupHash.State.SPLIT
-    ).update(group=None)
-    GroupHash.objects.filter(project_id=project.id, group__id__in=group_ids).exclude(
-        state=GroupHash.State.SPLIT
-    ).delete()
+    all_group_hashes.filter(state=GroupHash.State.SPLIT).update(group=None)
+    all_group_hashes.exclude(state=GroupHash.State.SPLIT).delete()
```
JoshFerge, line 60: I'm not sure if these querysets weren't intentionally not reused for a reason, so i'd keep these the way they were, and define your own queryset within the feature flag.

JoshFerge (changes requested):
could we split up these changes? let's start with a PR only for the celery task

jangjodi:
> could we split up these changes? let's start with a PR only for the celery task

Good suggestion, thanks! I've modified this PR to only have the seer endpoint change, this [PR](https://github.com/getsentry/sentry/pull/72767) to have the task, and this [PR](https://github.com/getsentry/sentry/pull/72769) to have the actual call to the task

src/sentry/seer/similarity/grouping_records.py
```diff
@@ -75,31 +79,61 @@ def post_bulk_grouping_records(
...
     else:
         logger.error(
-            "seer.delete_grouping_records.failure",
+            "seer.delete_grouping_records.project.failure",
         )
         return False
+
+
+def delete_grouping_records_by_hash(project_id: int, hashes: list[str]) -> bool:
+    extra = {"project_id": project_id, "hashes": json.dumps(hashes)}
```
JoshFerge, line 113: why are we calling json.dumps here? arent the hashes just a list of strings? why do we need to json stringify it?

JoshFerge (approved):

JoshFerge (approved):
lgtm

PR #71576, opened 2024-05-28
fix(backup): Keep `OrganizationMember`s with external inviters

azaslavsky:
There was previously a bug with how we exported `OrganizationMember`s: we assumed that all `invited_id` would also be in the organization, and therefore included in the filtering `pk_map`, but this is not actually true. This is because it is possible for an instance admin who is not a member of this (or really, any) org to invite the user. This creates an exception in the general assumption that "all org scope models reference user IDs already in the org".

To get around this, we simply wrap the exporting function for `OrganizationMember` in such a way that we no longer filter by `inviter_id`; we simply ignore the filed and take all values. The code already has an allowance that enables it to ignore unknown `inviter_id` values at import time, so doing this is okay.

To enable this change, a number of tests have been robustified with different kinds of `OrganizationMember` instances: pending invites, members invited by users in the org, members invited by users outside of the org, admins, superadmins, etc. We should have much better coverage of all of the edge cases associated with this model going forward.

### DIFF
--- a/src/sentry/db/postgres/base.py
+++ b/src/sentry/db/postgres/base.py
@@ -93,7 +93,7 @@ class CursorWrapper:
 
 class DatabaseWrapper(DjangoDatabaseWrapper):
     SchemaEditorClass = DatabaseSchemaEditorProxy
-    queries_limit = 10000
+    queries_limit = 12000
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/fixtures/backup/fresh-install.json
+++ b/fixtures/backup/fresh-install.json
@@ -56,6 +56,14 @@
   }
 },
 {
+  "model": "sentry.email",
+  "pk": 3,
+  "fields": {
+    "email": "superadmin@example.com",
+    "date_added": "2023-06-22T22:59:57.531Z"
+  }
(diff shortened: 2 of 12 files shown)


src/sentry/db/postgres/base.py
```diff
@@ -93,7 +93,7 @@ def executemany(self, sql, paramlist=()):
 
 class DatabaseWrapper(DjangoDatabaseWrapper):
     SchemaEditorClass = DatabaseSchemaEditorProxy
-    queries_limit = 10000
```
hubertdeng123, line 96: Is this limit something we're hitting? Is this intentional?

hubertdeng123 (commented):
I'm a bit confused by this edge case scenario. By instance admin, are you referring to Sentry admins? It seems weird to me that an admin that belongs to a different org is able to invite a user to another org that they don't belong to

src/sentry/db/postgres/base.py
azaslavsky: It just limited how much output the DB can report. For very large tests, we exceed that output (due to adding more users, etc), so I bumped it a bit.

hubertdeng123 (approved):

PR #71577, opened 2024-05-28
ref(feedback): hide pagination for screenshots if only 1 screenshot

### LINKED ISSUE #71573: User Feedback: Screenshot visualization shows < > buttons but we only capture 1 screenshot
image

These feel redundant, since clicking on them does nothing other than animate the button.

Can we hide the buttons if there's only 1 screenshot (which is all we support OOTB today)?

michellewzhang:
if there is 1 screenshot only (which is all we have the ability to attach right now), hide the pagination bar

before:
SCR-20240528-jtvk

after:
SCR-20240528-jtse

closes https://github.com/getsentry/sentry/issues/71573#issuecomment-2135751353

### DIFF
--- a/static/app/components/feedback/feedbackItem/screenshotsModal.tsx
+++ b/static/app/components/feedback/feedbackItem/screenshotsModal.tsx
@@ -34,22 +34,24 @@ export default function ScreenshotsModal({
   return (
     <Fragment>
       <Header closeButton>
-        <PaginationWrapper lightText>
-          <StyledScreenshotPagination
-            nextDisabled={screenshots.length > 1}
-            onNext={() => {
-              setSelectedIndex(prev => prev + 1);
-            }}
-            onPrevious={() => {
-              setSelectedIndex(prev => prev - 1);
-            }}
-            previousDisabled={screenshots.length > 1}
-            headerText={tct('[currentScreenshotIndex] of [totalScreenshotCount]', {
-              currentScreenshotIndex: currentIndex + 1,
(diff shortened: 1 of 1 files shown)


ryan953 (approved):

static/app/components/feedback/feedbackItem/screenshotsModal.tsx
```diff
@@ -34,22 +34,24 @@ export default function ScreenshotsModal({
...
-            headerText={tct('[currentScreenshotIndex] of [totalScreenshotCount]', {
-              currentScreenshotIndex: currentIndex + 1,
-              totalScreenshotCount: screenshots.length,
-            })}
-          />
-        </PaginationWrapper>
+        {screenshots.length > 1 && (
+          <PaginationWrapper lightText>
+            <StyledScreenshotPagination
+              nextDisabled={false}
```
michellewzhang, line 40: @ryan953 the previous logic of `nextDisabled={screenshots.length > 1}` should have been `nextDisabled={screenshots.length <= 1}` actually oops! but since this pagination bar now only shows when there's more than 1 screenshot, this condition doesn't really make sense anymore, so i made the prop always be `false`. not sure if there's a better way to check for disabling the buttons since it looks like the current index can be negative/infinitely increasing

ryan953: if the index is a wild number, positive or negative, then that's ok. keep the buttons enabled!

PR #71581, opened 2024-05-28
chore(api): Remove `id_or_slug` Options

iamrajjoshi:
Here, I am removing the wrapper function that uses the options and the options themselves. The options are related to allowing our apis work with id_or_slug, but this change has been GAed for a month and its a good time to cleanup.

### DIFF
--- a/src/sentry/options/defaults.py
+++ b/src/sentry/options/defaults.py
@@ -296,20 +296,6 @@ register(
     default=False,
     flags=FLAG_MODIFIABLE_BOOL | FLAG_AUTOMATOR_MODIFIABLE,
 )
-# Enable EA endpoints to work with id or slug as path parameters
-register(
-    "api.id-or-slug-enabled-ea-endpoints",
-    type=Sequence,
-    default=[],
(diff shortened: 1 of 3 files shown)


JoshFerge (approved):

src/sentry/options/defaults.py
```diff
@@ -296,20 +296,6 @@
     default=False,
     flags=FLAG_MODIFIABLE_BOOL | FLAG_AUTOMATOR_MODIFIABLE,
 )
-# Enable EA endpoints to work with id or slug as path parameters
-register(
```
schew2381, line 300: Do you need to remove the options from the automator first?

iamrajjoshi: i have that pr ready to be merged, i just wanted to make this pr to figure out if any tests fail because i missed something.

iamrajjoshi: https://github.com/getsentry/sentry-options-automator/pull/1493

iamrajjoshi: I merged the above PR, will merge this tomorrow if we are good.

PR #71583, opened 2024-05-28
fix(flagpole): Adds generic type to context builder, adds AnonymousUser support

GabeVillalobos:
- Adds support for `AnonymousUser` objects being passed into a feature context.
- Updates the context builder to have stricter typing using generics.


corps (approved):
Noice

src/flagpole/sentry_flagpole_context.py
```diff
@@ -1,18 +1,27 @@
...
 
 class InvalidContextDataException(Exception):
     pass
 
 
-def organization_context_transformer(data: dict[str, Any]) -> EvaluationContextDict:
+class SentryContextData(TypedDict):
+    actor: User | RpcUser | AnonymousUser | None
+    organization: Organization | RpcOrganization | None
+    project: Project | RpcProject | None
```
asottile-sentry, line 21: are these supposed to be nullable (`| None`) or sometimes present `NotRequired[...]`? (or maybe both?)

asottile-sentry (commented):
I notie there's a lot of calls to `dict(...)` in the code -- this is slower than a dict literal and generally should be avoided (we probably should add linting for this?)

src/flagpole/sentry_flagpole_context.py
markstory: They need to be nullable because features can be organization or project scoped. When a feature is org scoped it doesn't have projects, and project scoped features often don't get organizations passed to them. There are also feature flag checks that have no `actor` either 🫠

asottile-sentry: oh another question as well -- can this be a real type instead of a TypedDict? dicts are too easy to mess with / mess up and I'd feel more comfortable with a NamedTuple or some other concrete type

markstory: We could have a more formal type for the context builder input data. I don't think having a type for the context data would work well because of the getsentry/sentry boundaries that we have to cross to be compatible with feature flag usage.

markstory (approved):
Looks like a step in the right direction to me. We should continue to improve types for context data though.

GabeVillalobos:
Decided to commit to using `NamedTuple` generics for the implementation. Only challenge here is that GetSentry testing will break temporarily so I've pushed getsentry/getsentry#14101 to alleviate this while we wait for these changes to land.

src/flagpole/sentry_flagpole_context.py
GabeVillalobos: I've converted everything to use dataclasses instead. GetSentry will have its own subclass of `SentryContextData`  with additional GetSentry only data. Let me know if you have any other suggestions for this implementation.

tests/flagpole/test_feature.py
```diff
@@ -1,11 +1,16 @@
 from datetime import datetime, timezone
+from typing import NamedTuple
 
 import pytest
 
 from flagpole import ContextBuilder, EvaluationContext, Feature, InvalidFeatureFlagConfiguration
 from flagpole.conditions import ConditionOperatorKind
 
 
+class ContextData(NamedTuple):
```
markstory, line 10: Should this be a dataclass as well?

markstory (approved):

PR #71585, opened 2024-05-28
feat(remote-config): Update protocol and api blueprints to enable multi-feature

cmanallen:
Removes hard coded `user_config` object and allows multi-key feature definition.  Note: this does not increase the number of API endpoints.  Though we may chose to do so later.


ryan953 (approved):

src/sentry/remote_config/docs/api.md
```diff
@@ -44,16 +68,30 @@ Retrieve the DSN's configuration.
 
 ### Set Configuration [POST]
 
-Set the DSN's configuration.
+Set the DSN's configuration. A hash value of `null` is provided if the object does not exist. If the API finds an existing configuration object an error is returned. If the hash value is provided then it will be compared against the hash of the remote value. A new hash is always returned on successful write. If the hashes do not match an error is returned.
```
billyvg, line 71: Just to clarify, is the hash here used to handle concurrent writes? e.g.

* users A and B have configuration `1`
* user B updates the configuration first, remote now has config w/ hash `2`
* user A attempts to update but fails because their client hash is `1` !== server hash `2`

billyvg (approved):

src/sentry/remote_config/docs/api.md
bruno-garcia, line 71: Since DSNs can be disabled and the app replaced with a new DSN, for the same project. Should we tie this to "project config" instead?

```suggestion
Set the project's configuration. A hash value of `null` is provided if the object does not exist. If the API finds an existing configuration object an error is returned. If the hash value is provided then it will be compared against the hash of the remote value. A new hash is always returned on successful write. If the hashes do not match an error is returned.
```

Here is an example of 2 DSNs for a single project:

image

cmanallen: The point is to split the configurations by DSN.  DSNs allow people to split configurations up.  I.e. the next.js case where client and server share a project.

If a DSN needs to be rotated we can expose UI to copy the configuration data over.

cmanallen: Not concurrent writes just stale writes.  I'm not sure if I want to include it still.  I'm trying to avoid a situation where a stale session overwrites lots of changes.  I'm not sure its worth it atm.

bruno-garcia: I'm not sure this is ideal. Specially now that we're going to have 1 DSN for muliple apps. So a new field for to split config furthers will be used (like `application`).

What do we gain by scoping this to DSN instead of just the project?

cmanallen: What I mentioned above is what we gain.  If Sentry wants to introduce a new scoping concept then we can follow that migration.  I'm tying to the DSN because it exists today.

cmanallen: It's fine if we want to scope to the project.  I'm just stating the reasoning for scoping to the DSN.  If that's not wanted we can drop it.

ryan953: i think it's fair to drop it for now. We only need to build what's minimum to get things working end to end to start and this is extra on that front.

Also, if we look at the previous api design which could be our end-state design... that is very different, so there's time to decide how we would solve for stale and/or concurrent writes at that point. We don't need to have a solution for it with this version of the api

ryan953: I'd even say... we shouldn't need to write down the `features` field yet if it'll be unimplemented by the UI, BE, and/or SDK.

cmanallen: It should be implemented in the UI --mulitple key, value pairs.  The point of the previous API design was to allow the Web UI to move fast.  It was never the final design.  I think the new API is simple enough that should be quick for you to iterate on while also providing the full feature set of the final product.

cmanallen: Billy will be implementing the `features` field on the SDK this week and I've already pushed a branch handling it on the API side.

ryan953: ah, ok
i'll put it into the ui asap too

cmanallen: Decided offline.  Will scope to the project for now.  DSN is imperfect for scoping.

ryan953 (approved):

PR #71597, opened 2024-05-28
ref(crons): Load timeline stats one-by-one

evanpurkhiser:
This improves perceived performance


wedamija:
@evanpurkhiser should we merge this?

PR #71600, opened 2024-05-28
fix(typing): Add types for AuditLogEntry model

snigdhas:


pyproject.toml
```diff
@@ -354,7 +354,6 @@ module = [
     "sentry.middleware.auth",
     "sentry.middleware.ratelimit",
     "sentry.middleware.superuser",
-    "sentry.models.auditlogentry",
```
evanpurkhiser, line 357: I'm confused, it looks like a ton of types are missing in this module, why doesn't mypy complain more?

JoshFerge (approved):

pyproject.toml
asottile-sentry: the top list is things that are _failing_ the type checker -- the bottom list requires all things to be typed

PR #71601, opened 2024-05-28
fix(typing): Add types for codeowners

snigdhas:


JoshFerge (approved):

src/sentry/tasks/codeowners/update_code_owners_schema.py
```diff
@@ -31,9 +31,9 @@ def update_code_owners_schema(
     )
     from sentry.models.projectcodeowners import ProjectCodeOwners
 
-    organization = load_model_from_db(Organization, organization)
+    org: Organization = load_model_from_db(Organization, organization)
```
asottile-sentry, line 34: instead, `load_model_from_db` should be typed -- probably as:

```python
ModelT = TypeVar("ModelT", bound=Model)
def load_model_from_db(tp: type[ModelT], inst: ModelT | int) -> ModelT: ...
```

src/sentry/tasks/base.py
```diff
@@ -65,12 +68,12 @@ def track_memory_usage(metric, **kwargs):
         metrics.distribution(metric, get_rss_usage() - before, unit="byte", **kwargs)
 
 
-def load_model_from_db(cls, instance_or_id, allow_cache=True):
+def load_model_from_db(tp: type[ModelT], instance_or_id: ModelT | int, allow_cache=True) -> ModelT:
```
asottile-sentry, line 71: `allow_cache` can also be annotated `: bool`

asottile-sentry (approved):

PR #71603, opened 2024-05-28
fix(typings): Fix types for test_post_process

snigdhas:


JoshFerge (approved):

tests/sentry/tasks/test_post_process.py
```diff
@@ -2004,7 +2015,7 @@ def test_0_sample_rate_replays(self, incr, kafka_producer, kafka_publisher):
         )
         assert kafka_producer.return_value.publish.call_count == 0
         for args, _ in incr.call_args_list:
-            self.assertNotEqual(args, ("post_process.process_replay_link.id_sampled"))
+            assert args != "post_process.process_replay_link.id_sampled"
```
asottile-sentry, line 2018: I don't think this assertion will ever succeed -- `args` is a `tuple` and it's never going to equal a string

PR #71605, opened 2024-05-28
fix(typing): Fix typing for organizationaccessrequest

snigdhas:


src/sentry/models/organizationaccessrequest.py
```diff
@@ -52,6 +52,8 @@ def send_request_email(self):
 
         if self.requester_id:
             requester = user_service.get_user(user_id=self.requester_id)
+            if not requester:
+                return
```
asottile-sentry, line 56: this feels... not quite right -- what scenario would this user id just suddenly be missing?  this feels like an error case rather than one to just silently ignore (I guess same feedback for below) -- maybe someone from hybrid-cloud could better represent what _should_ happen here in the None case

mrduncan: I'm not sure it's possible to realistically occur here but generally these are most common when the user is deleted. In this case returning without doing anything feels like the right thing to do, although we might want to also log these just for sanity?

armenzg: If we get to such a state I would want to know.

What about this?
```diff
diff --git a/pyproject.toml b/pyproject.toml
index f91a02085f..702139511d 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -350,7 +350,6 @@ module = [
     "sentry.middleware.superuser",
     "sentry.models.integrations.external_issue",
     "sentry.models.integrations.sentry_app_installation",
-    "sentry.models.organizationaccessrequest",
     "sentry.monitors.consumers.monitor_consumer",
     "sentry.monitors.endpoints.base",
     "sentry.monitors.endpoints.monitor_ingest_checkin_attachment",
diff --git a/src/sentry/models/organizationaccessrequest.py b/src/sentry/models/organizationaccessrequest.py
index dc21f94a42..90754e9d4c 100644
--- a/src/sentry/models/organizationaccessrequest.py
+++ b/src/sentry/models/organizationaccessrequest.py
@@ -52,6 +52,7 @@ class OrganizationAccessRequest(Model):

         if self.requester_id:
             requester = user_service.get_user(user_id=self.requester_id)
+            assert requester is not None
             context.update({"requester": requester.get_display_name()})

         msg = MessageBuilder(
@@ -79,6 +80,7 @@ class OrganizationAccessRequest(Model):
         from sentry.utils.email import MessageBuilder

         user = user_service.get_user(user_id=self.member.user_id)
+        assert user is not None
         email = user.email
         organization = self.team.organization
```

mrduncan: I don't have strong opinions so I'll defer to @snigdhas since it's her change but asserting this doesn't feel quite right to me - users get deleted all the time and an assertion here is going to raise errors for cases which we expect to occur at least occasionally

snigdhas:
gonna let this close as stale since I haven't had a chance to look into it further - I'll reopen at a later time. thanks folks!

PR #71606, opened 2024-05-28
fix(typing): Update release health types

snigdhas:


src/sentry/release_health/tasks.py
```diff
@@ -95,7 +96,7 @@ def adopt_releases(org_id: int, totals: Totals) -> Sequence[int]:
                                 environment__organization_id=org_id,
                             )
 
-                            updates = {}
+                            updates: dict[str, Any] = {}
```
asottile-sentry, line 99: this is fine, but could probably be more specific with a `TypedDict`:

```python
class PDEUpdates(TypedDict, total=False):
    adopted: datetime.datetime
    unadopted: None
```

asottile-sentry (approved):

armenzg (approved):

PR #71611, opened 2024-05-28
fix(ui): Avoid extra descender space

evanpurkhiser:
Because the buttons are `inline-block` we're getting some extra
descender spacing here that we don't want. Using display: flex removes
that descender space.

I'm not 100% confident this is the right solution, but it's definitely
going to help

image
image

Then with `display: flex` we can observe the extra descender spacing is not there:

image


ryan953 (approved):
flexbox all the things!

anonrig (approved):

scttcper (approved):

PR #71613, opened 2024-05-28
fix: remove unused killswitch functions

anonrig:
Removes unused functions in killswitch. I couldn't find any usages in getsentry as well.


asottile-sentry (approved):

markstory (approved):
I didn't see any mention of these functions in the [runbook for killswitches either](https://www.notion.so/sentry/Dropping-events-per-project-event-type-organization-killswitches-a32e23c82fd24e18b3591d31a70e4ee4)

PR #71616, opened 2024-05-28
feat(fe): Updated OrganizationDetailsEndpoint Urls to use Query Paramaters

iamrajjoshi:
In this pr, I update the urls for OrganizationDetailsEndpoint to use the new query parameter, `include_feature_flags`, which we will use to determine if the features data get returned in the API response object. 

We mentioned two places where we will be updating the urls in our research [here](https://www.notion.so/sentry/Remove-feature-flags-from-API-response-95ed20b0c6b74b7fad4cbc4495cd4c9b). The third place I modify is `preload-data.html` was because it was also preloading data and the preloaded org should also have all the features since they are used throughout our frontend.

Not sure if there are other gotchas for this.

### DIFF
--- a/src/sentry/api/endpoints/project_details.py
+++ b/src/sentry/api/endpoints/project_details.py
@@ -479,10 +479,6 @@ class ProjectDetailsEndpoint(ProjectEndpoint):
         """
         Return details on an individual project.
         """
-        # This param will be used to determine if we should include feature flags in the response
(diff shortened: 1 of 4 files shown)


ykamo001 (approved):
Thanks for making these changes

ykamo001:
> we will use to determine if we are going to pass all the features
Would rephrase this to "we will use to determine if the features data get returned in the API response object

saponifi3d (approved):

PR #71620, opened 2024-05-28
fix(issues): Use payload_data[fingerprint] for occurrence and status change partition keys

aliu39:
Fixes a problem where status change msgs are processed before occurrences, resulting in [dropped_group_not_found](https://app.datadoghq.com/metric/explorer?fromUser=false&start=1716844861680&end=1716931261680&paused=false#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVBGN8qVoD6gnNtUZCKiOq279VKY6epbINiAiGOrKQdpYAMwGIFCeGE4a+NoAxkrIIOoIulmWgWZxyhp4qshaIoVYAHQ4WVl42kYiOU6i8BoIDX2eeBpOWYwYIvANUNo48opQTjCzePJOIjgILvgiUMAAVDwNGCNZOwgAFACUIHygkdF5sVgALEkpCGkZbTnKBUUlEDPCpVGpwOpxJotNodLo9cwDT4IYajcaTOANXwaDC9A5HE6jc7XW4AXSornceEwoXCFNUVIwMTKCVuFFAdIZTJ0r1JVD6WDQuVA8gwgoQCFyyRwMFG1I0ECySTQojgTjkigqOGVUCVKqc9CYyg27lFrOSEHsmCwaoUkuVIiUPBJPD4IEt4gAwlJhDAUCIqWgeEA). Previously status change had no key, and there are no order guarantees across partitions

### DIFF
--- a/src/sentry/issues/producer.py
+++ b/src/sentry/issues/producer.py
@@ -64,9 +64,8 @@ def produce_occurrence_to_kafka(
     if payload_data is None:
         return
 
-    partition_key = None
-    if occurrence and occurrence.fingerprint:
-        partition_key = occurrence.fingerprint[0].encode()
+    fingerprint = payload_data["fingerprint"]
+    partition_key = fingerprint[0].encode()
     payload = KafkaPayload(partition_key, json.dumps(payload_data).encode("utf-8"), [])
     if settings.SENTRY_EVENTSTREAM != "sentry.eventstream.kafka.KafkaEventStream":
         # If we're not running Kafka then we're just in dev.
--- a/src/sentry/issues/producer.py
+++ b/src/sentry/issues/producer.py
@@ -64,7 +64,9 @@ def produce_occurrence_to_kafka(
     if payload_data is None:
         return
 
-    fingerprint = payload_data["fingerprint"]
+    fingerprint = payload_data[
+        "fingerprint"
+    ]  # guaranteed non-empty by occurrence and status_change __post_init__
     partition_key = fingerprint[0].encode()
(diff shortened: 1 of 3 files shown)


src/sentry/issues/producer.py
```diff
@@ -64,9 +64,8 @@ def produce_occurrence_to_kafka(
     if payload_data is None:
         return
 
-    partition_key = None
-    if occurrence and occurrence.fingerprint:
-        partition_key = occurrence.fingerprint[0].encode()
+    fingerprint = payload_data["fingerprint"]
+    partition_key = fingerprint[0].encode()
```
scefali, line 68: @aliu3ntry should we check if fingerprints is at least 1 item?

aliu3ntry: should i emit some kind of log/metric if empty?

scefali: @aliu3ntry that and prevent it from blowing up :)

aliu3ntry: Tried a different approach that should throw an error in this case before producing.. let me know how it looks

bruno-garcia (commented):
Can you please add a test that verifies this works? The case of 0 fingerprint Steve raised is a good test case too

aliu3ntry:
Closing and reopening at https://github.com/getsentry/sentry/pull/71628. After looking at the tests I wanted to preserve the behavior of partition key = None when we don't have a fingerprint/occurrence

PR #71622, opened 2024-05-28
fix(insights): change web vitals to darker shade of yellow

KevinL10:
before:
image

after:
image
This change only affects the components in WebVitals


edwardgou-sentry (approved):

edwardgou-sentry:
@KevinL10 Seems like there's a similar PR here?
https://github.com/getsentry/sentry/pull/71619

PR #71623, opened 2024-05-28
feat(flagpole): Add date_parsed property to features

GabeVillalobos:
Adds `date_parsed` to features, which can be used for caching TTL checks when evaluating a feature flag. The value of this will always be `datetime.now` when instantiating a new Feature.


src/flagpole/__init__.py
```diff
@@ -100,16 +101,18 @@ def dump_schema_to_file(self, file_path: str) -> None:
     @classmethod
     def from_feature_dictionary(cls, name: str, config_dict: dict[str, Any]) -> Feature:
         try:
+            # Disallow date_parsed from being set in the feature config
+            if "date_parsed" in config_dict:
+                config_dict.pop("date_parsed")
+
```
GabeVillalobos, line 107: Is there a better way to exclude this from our pydantic parsing? I tried to make this a private member variable with getter method, but pydantic seems to get a bit fussy about this.

markstory (approved):

GabeVillalobos:
Closing this out for now as I don't think we need this at the moment, and would probably put this in a different data structure for caching if we do end up wanting it.

PR #71627, opened 2024-05-28
feat(utils): Add `with_circuit_breaker` helper

lobsterkatie:
Currently, our circuit breaker can check if it's been tripped (by too many errors happening within too short a time), but it leaves it to the calling code to actually track those errors. This PR introduces a wrapper function, `with_circuit_breaker`, to combine both tasks into one.

I was deeply conflicted about whether to write this as a wrapper with a callback or as a context manager. I ended up with the wrapper approach because my fear was the that natural tendency with a context manager would be to use it like this:

```python
with circuit_breaker("some_key", config):
	try:
		do_something()
	except Exception:
		handle_exception()
```

The problem with that is that the circuit breaker has to be _inside_ the `try` block - otherwise, the errors it's supposed to be tracking will already have been handled before bubbling up where it could see them. So instead I went with a function which could  be wrapped around `do_something`, because that way - short of pulling the whole `try-except` into `do_something`'s innards - there's no way to make the above mistake. In the end it looks like this:

```python
try:
	with_circuit_breaker("some_key", do_something, config)
except CircuitBreakerTripped:
	logger.log("`do_something` actually did nothing, because the circuit breaker's been tripped.")
except Exception:
	handle_exception()
```

I still don't love it, though - IMHO it's less readable, and you have to stick `do_something` into a lambda if you want to pass it any arguments, which makes it even less readable and is also just kind of a pain. Maybe there's some Python magic I don't know about which would make the above context manager example work? Very open to suggestions on that score, but am going with this for now to keep things moving.

Note: The `_should_call_callback` helper here draws heavy inspiration from the existing `circuit_breaker_activated` function (h/t @snigdhas), the key differences being where the config data comes from and the fact that they return opposite answers (returning True when the other returns False, and vice versa). I left `circuit_breaker_activated` alone rather than refactoring and then wrapping it so as not to have to change everywhere it's used. If those spots choose to switch over to using `with_circuit_breaker`, that can happen in follow-up PRs.

### DIFF
--- a/src/sentry/utils/circuit_breaker.py
+++ b/src/sentry/utils/circuit_breaker.py
@@ -1,9 +1,13 @@
-from typing import TypedDict
+from collections.abc import Callable
+from typing import ParamSpec, TypedDict, TypeVar
 
 from django.core.cache import cache
 
 from sentry import ratelimits as ratelimiter
 from sentry.utils import metrics
+
+# TODO: Right now this circuit breaker is based on count of consecutive errors. We should consider
+# whether basing it on percentage of failed requests would be better.
 
 DEFAULT_ERROR_LIMIT = 30
 ERROR_COUNT_CACHE_KEY = lambda key: f"circuit_breaker:{key}-error-count"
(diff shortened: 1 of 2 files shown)


src/sentry/utils/circuit_breaker.py
```diff
@@ -15,6 +16,42 @@ class CircuitBreakerPassthrough(TypedDict, total=True):
...
+    # calls to bypass the circuit breaker
+    passthrough_interval: int
+
+
+CIRCUIT_BREAKER_DEFAULTS = CircuitBreakerConfig(
+    error_limit=DEFAULT_ERROR_LIMIT,
+    error_limit_window=3600,  # 1 hour
+    allow_passthrough=False,
+    # TODO: No idea what a reasonable value here is
+    passthrough_interval=15,  # in sec
```
snigdhas, line 44: seems reasonable to me, the severity limit is 1 per 10 seconds 

```suggestion
    passthrough_interval=15,  # 15 seconds
```

src/sentry/utils/circuit_breaker.py
```diff
@@ -42,3 +78,100 @@ def circuit_breaker_activated(
...
+    error_count_key: str,
+    error_limit_window: int,
+    reset: bool = False,
+) -> None:
+    """
+    Increment the count at the given key, unless `reset` is True, in which case, reset the count to 0.
+    """
+    current_count = _get_or_set_error_count(error_count_key, error_limit_window)
+    new_count = 0 if reset else current_count + 1
+    cache.set(error_count_key, new_count, error_limit_window)
```
snigdhas, line 171: _technically_ this does an extra cache.get_or_set which can be skipped if `reset=True`

this isn't as clean but 🤷‍♀️ 
```suggestion
    new_count = 0 
    if not reset:
      current_count = _get_or_set_error_count(error_count_key, error_limit_window)
      new_count = current_count + 1
    cache.set(error_count_key, new_count, error_limit_window)
```

snigdhas (approved):
LGTM, thank you for the refac!

src/sentry/utils/circuit_breaker.py
lobsterkatie: Yeah, that actually just occurred to me as I was looking it over again. It's a tiny bit less readable, but I suppose I could just do `new_count = 0 if reset else _get_or_set_error_count(error_count_key, error_limit_window) + 1`.

lobsterkatie: Or I could split up the ternary:

```python
if reset:
    new_count = 0
else:
    new_count = _get_or_set_error_count(error_count_key, error_limit_window) + 1
```

Verbose but very readable. I'll go with that.

PR #71628, opened 2024-05-28
fix(issues): Add partition key for status change kafka msg

aliu39:
Reopened from https://github.com/getsentry/sentry/pull/71620
Previously status change had no key. There are no order guarantees across partitions, resulting in status changes being processed before the first occurrence/group was created.

### DIFF
--- a/tests/sentry/issues/test_producer.py
+++ b/tests/sentry/issues/test_producer.py
@@ -376,3 +376,81 @@ class TestProduceOccurrenceForStatusChange(TestCase, OccurrenceTestMixin):
         )
         assert status_change_1.id
         assert status_change_1.id != status_change_2.id
+
+    @patch(
+        "sentry.issues.producer._prepare_status_change_message", return_value={"mock_data": "great"}
+    )
+    @patch("sentry.issues.producer._occurrence_producer.produce")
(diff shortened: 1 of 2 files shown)


JoshFerge (approved):

tests/sentry/issues/test_producer.py
```diff
@@ -376,3 +376,81 @@ def test_generate_status_changes_id(self) -> None:
         )
         assert status_change_1.id
         assert status_change_1.id != status_change_2.id
+
+    @patch(
```
aliu3ntry, line 380: These are copied from the occurrence tests

wedamija (approved):
This works fine, any reason we can't use the value from the payload though?

aliu3ntry:
> This works fine, any reason we can't use the value from the payload though?

There's no guarantee from the `_prepare` functions that `payload_data` contains a fingerprint, or a non-empty fingerprint. We could enforce this, but we'd have to rewrite the tests that still produce when the fingerprint is missing. 
I'm not sure if this (no fingerprint -> fail) is the behavior issues wants

aliu3ntry:
CI was failing in s4s, want to be around when this goes out so will try again tmrw

PR #71631, opened 2024-05-28
ref(feedback): remove unused creation source

aliu39:
NEW_FEEDBACK_DJANGO_ENDPOINT was used for some experiments but unused now. update_user_reports isn't used anywhere and isn't needed after https://github.com/getsentry/sentry/pull/70200


JoshFerge:
i think update_user_reports is still potentially needed? in cases where post_process fails, etc. if you look at the logs, you can sometimes see that it still finds feedbacks that fall through the cracks, last i checked.

aliu3ntry:
> i think update_user_reports is still potentially needed? in cases where post_process fails, etc. if you look at the logs, you can sometimes see that it still finds feedbacks that fall through the cracks, last i checked.

ok - lets keep it!

michellewzhang (approved):

PR #70632, opened 2024-05-09
feat(rr6): Add shims for React Router 3 → 6

evanpurkhiser:
This is tracked by https://github.com/getsentry/frontend-tsc/issues/23

PR #71634, opened 2024-05-29
ref(charts): Move STARFISH_FIELDS out of charts.ts

evanpurkhiser:
This module imports a TON of stuff, STARFISH_FIELDS is used by
discover/fields. By moving this out we can reduce the import graph
generated for chartcuterie by a magnitude of a few MBs.

This is also needed for GH-70632


static/app/views/starfish/components/chart.tsx
```diff
@@ -49,37 +49,9 @@ import type {AggregationOutputType, RateUnit} from 'sentry/utils/discover/fields
 import {aggregateOutputType} from 'sentry/utils/discover/fields';
 import usePageFilters from 'sentry/utils/usePageFilters';
 import useRouter from 'sentry/utils/useRouter';
-import {SpanIndexedField, SpanMetricsField} from 'sentry/views/starfish/types';
```
evanpurkhiser, line 52: Expand thsi file and just look at how much this was importing 🤯

scttcper (approved):

PR #71635, opened 2024-05-29
ref(seer grouping): Use circuit breaker for Seer call during grouping

lobsterkatie:
This adds a circuit breaker to the Seer call we make before creating a new group, backed up by a new option, `seer.similarity.circuit-breaker-config`. This means that if requests to Seer in that spot error out too many times in a row, over too short a timespan, we'll temporarity block the majority of those calls. (Periodically we'll let one through, to see if whatever was broken has fixed itself, at which point we can resume calling it like normal.)

Right now what "too many errors, too fast" means is more than 30 in an hour, which is the circuit breaker default. If we want to adjust that in the future, we can do so by modifying the new option, which right now only overrides the circuit breaker defaults to turn on that periodic "see if things are okay now" bypass.

### DIFF
--- a/src/sentry/event_manager.py
+++ b/src/sentry/event_manager.py
@@ -125,7 +125,9 @@ from sentry.utils.canonical import CanonicalKeyDict
 from sentry.utils.circuit_breaker import (
     ERROR_COUNT_CACHE_KEY,
     CircuitBreakerPassthrough,
+    CircuitBreakerTripped,
     circuit_breaker_activated,
+    with_circuit_breaker,
 )
 from sentry.utils.dates import to_datetime
 from sentry.utils.event import has_event_minified_stack_trace, has_stacktrace, is_handled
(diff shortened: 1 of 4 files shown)


src/sentry/event_manager.py
```diff
@@ -1493,8 +1494,10 @@ def _save_aggregate(
                     try:
                         # If the `projects:similarity-embeddings-grouping` feature is disabled,
                         # we'll still get back result metadata, but `seer_matched_group` will be None
-                        seer_response_data, seer_matched_group = get_seer_similar_issues(
-                            event, primary_hashes
+                        seer_response_data, seer_matched_group = with_circuit_breaker(
+                            "event_manager.get_seer_similar_issues",
+                            lambda: get_seer_similar_issues(event, primary_hashes),
+                            options.get("seer.similarity.circuit-breaker-config"),
```
vartec, line 1500: where does this handle CircuitBreakerTripped?

lobsterkatie: Yup, just realized this as I was going to post it in the channel. In the middle of fixing it now.

lobsterkatie: Okay, fixed. And actually, we don't want to do anything in that case, because the circuit breaker is really just another check à la those in `should_call_seer_for_grouping`. But it's true that we don't want to capture it as a legit error, so you're right that it needed to be a separate `except` branch from the generic one.

vartec: BTW, I'm a bit late with that, because you've already merged the CircuitBreaker PR, but I think `with_circuit_breaker` might work better as a context manager, checking if the breaker is open, and returning `CircuitBreakerTripped` if it is, otherwise yielding from within try/except block. Let me check if that's something that I can try quickly.

lobsterkatie: Before you do that, please read over the description in https://github.com/getsentry/sentry/pull/71627. I did consider making it a context manager, but decided people were more likely to use it right if I did it as a function.

Not vetoing switching it, but I'd like to address that concern before we do.

JoshFerge (approved):

PR #71646, opened 2024-05-29
Expose a new Relay option to send attachment contents inline

Swatinem:
This will be used in relay starting with this PR: https://github.com/getsentry/relay/pull/3654


loewenheim (commented):
This only registers the option?

Swatinem:
Yes, it is propagated to Relay however for use here: https://github.com/getsentry/relay/pull/3654

loewenheim (approved):

anonrig (approved):

Dav1dde (approved):

PR #71649, opened 2024-05-29
perf: move sentry/monitors to orjson

anonrig:

This pull-request only changes `json` usages in `src/sentry/monitors` folder.

### DIFF
--- a/src/sentry/monitors/processing_errors/manager.py
+++ b/src/sentry/monitors/processing_errors/manager.py
@@ -5,6 +5,7 @@ import uuid
 from datetime import timedelta
 from itertools import chain
 
+import orjson
 from django.conf import settings
 from redis.client import StrictRedis
 from rediscluster import RedisCluster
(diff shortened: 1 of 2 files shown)


wedamija (approved):
Looks good to me. I'd recommend not merging this Friday just in case we have issues, so either today or Monday

src/sentry/monitors/processing_errors/manager.py
```diff
@@ -98,7 +99,9 @@ def store_error(error: CheckinProcessingError, monitor: Monitor | None):
     entity_identifier = _get_entity_identifier_from_error(error, monitor)
     error_set_key = build_set_identifier(entity_identifier)
     error_key = build_error_identifier(error.id)
-    serialized_error = json.dumps(error.to_dict())
+    serialized_error = orjson.dumps(
+        error.to_dict(), option=orjson.OPT_UTC_Z | orjson.OPT_NON_STR_KEYS
```
wedamija, line 103: I'm not sure if there are efficiency improvements here, but I think it's safe to not have the `OPT_NON_STR_KEYS` option for this code.

anonrig: I've removed it. I mostly added it as a `defensive` approach.

anonrig: Thanks for the pull-request review @wedamija!

anonrig:
cc @wedamija this requires an additional review, since another `json.dumps` has been landed before I get to land this pull-request.

wedamija (approved):

PR #71651, opened 2024-05-29
perf: move sentry/interfaces to orjson

anonrig:

This pull-request only changes `json` usages in `src/sentry/interfaces` folder.

### DIFF
--- a/src/sentry/interfaces/security.py
+++ b/src/sentry/interfaces/security.py
@@ -1,8 +1,8 @@
+import orjson
 from django.utils.functional import cached_property
 
 from sentry.interfaces.base import Interface
 from sentry.security import csp
-from sentry.utils import json
 from sentry.web.helpers import render_to_string
 
 __all__ = ("Csp", "Hpkp", "ExpectCT", "ExpectStaple")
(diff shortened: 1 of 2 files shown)


src/sentry/interfaces/security.py
```diff
@@ -173,7 +173,10 @@ def to_python(cls, data, **kwargs):
         return super().to_python(data, **kwargs)
 
     def to_string(self, is_public=False, **kwargs):
-        return json.dumps({"csp-report": self.get_api_context()})
+        return orjson.dumps(
+            {"csp-report": self.get_api_context()},
+            option=orjson.OPT_UTC_Z | orjson.OPT_NON_STR_KEYS,
```
jjbayer, line 178: Why do we have these options here, but not in `stringify` above?

anonrig: Because on `message.py` it gets executed only for int, float and bool, and there is no timestamp or non-string key to serialize. Hence not needed there.

jjbayer: I did a quick search for `stringify` and it doesn't seem to be used at all, so maybe we can remove it? Though I might be missing some Python magic.

jjbayer (approved):

src/sentry/interfaces/security.py
anonrig: I'll follow up on a different PR to remove this. Thanks for the review!

PR #71652, opened 2024-05-29
perf: move sentry/backup to orjson

anonrig:

This pull-request only changes `json` usages in `src/sentry/backup` folder.

### DIFF
--- a/src/sentry/backup/crypto.py
+++ b/src/sentry/backup/crypto.py
@@ -14,7 +14,6 @@ from cryptography.hazmat.primitives.asymmetric import padding
 from google.cloud.kms import KeyManagementServiceClient as KeyManagementServiceClient
 from google_crc32c import value as crc32c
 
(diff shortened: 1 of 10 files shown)


azaslavsky (approved):
LGTM! Note that we still use the default `json` library, aliased as`builtin_json`, in `backup/exports.py`, because we rely on pretty printing that output. If we're confident that `orjson` can pretty print the same way, we can swap that usage as well.

anonrig:
> LGTM! Note that we still use the default `json` library, aliased as`builtin_json`, in `backup/exports.py`, because we rely on pretty printing that output. If we're confident that `orjson` can pretty print the same way, we can swap that usage as well.

`orjson` supports `OPT_INDENT_2` option - https://github.com/ijl/orjson?tab=readme-ov-file#opt_indent_2. Does that produce the same result?

PR #71653, opened 2024-05-29
ref(mqb): Default to not use the metrics indexer

wmak:
Most of the datasets don't use the metrics indexer, so default to it being not used instead of defaulting to using it


mjq (approved):
I don't know about the specifics of the indexer, but the change looks fine.

PR #71667, opened 2024-05-29
feat(releases): Support IN and NOT IN for release filter

narsaynorath:
Adds support for IN and NOT IN filters for the release alias. This addition supports passing along arbitrary releases for release health ordering in dashboards. I'm adding this IN filtering to fix a bug in a release health widget for dashboards where a release is selected, but the widget doesn't display the relevant data.

The bug: to sort release health by date, we make a call to the releases endpoint to fetch <=50 latest releases by date. Those are used to conduct a metrics request and the result is ordered in the frontend. If the release isn't in that first 50 set, then it gets lost. See the relevant code [here](https://github.com/getsentry/sentry/blob/master/static/app/views/dashboards/datasetConfig/releases.tsx#L439)

### DIFF
--- a/src/sentry/api/endpoints/organization_releases.py
+++ b/src/sentry/api/endpoints/organization_releases.py
@@ -87,8 +87,8 @@ def _filter_releases_by_query(queryset, organization, query, filter_params):
 
         if search_filter.key.name == RELEASE_ALIAS:
             query_q = Q()
+            raw_value = search_filter.value.raw_value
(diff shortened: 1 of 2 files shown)


narsaynorath:
I'm not sure who owns this endpoint/has opinions on adding this filter but I see @wedamija you've been active in this file. Would you have time to take a look? Thanks!

wedamija (approved):

PR #71668, opened 2024-05-29
feat(api): Logic for `include_feature_flag` Query Parameter

iamrajjoshi:
In this final PR, I implement the logic to use the `include_feature_flag` logic and selective remove the "features" key from the response.

Closes: https://github.com/getsentry/team-core-product-foundations/issues/326

### DIFF
--- a/src/sentry/api/endpoints/organization_details.py
+++ b/src/sentry/api/endpoints/organization_details.py
@@ -590,6 +590,7 @@ class OrganizationDetailsEndpoint(OrganizationEndpoint):
         :pparam string organization_id_or_slug: the id or slug of the organization the
                                           team should be created for.
         :param string detailed: Specify '0' to retrieve details without projects and teams.
+         :qparam string include_feature_flags: whether or not to include feature flags in the response
         :auth: required
         """
         # This param will be used to determine if we should include feature flags in the response
--- a/src/sentry/api/endpoints/organization_details.py
+++ b/src/sentry/api/endpoints/organization_details.py
@@ -624,6 +624,7 @@ class OrganizationDetailsEndpoint(OrganizationEndpoint):
         :param string name: an optional new name for the organization.
         :param string slug: an optional new slug for the organization.  Needs
                             to be available and unique.
+        :qparam string include_feature_flags: whether or not to include feature flags in the response
         :auth: required
         """
         from sentry import features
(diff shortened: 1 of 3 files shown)


src/sentry/api/endpoints/organization_details.py
```diff
@@ -590,11 +590,11 @@ def get(self, request: Request, organization) -> Response:
         :pparam string organization_id_or_slug: the id or slug of the organization the
                                           team should be created for.
         :param string detailed: Specify '0' to retrieve details without projects and teams.
+         :qparam string include_feature_flags: whether or not to include feature flags in the response
```
sentaur-athena, line 593: nit: indent

sentaur-athena (approved):

src/sentry/api/endpoints/organization_details.py
```diff
@@ -698,6 +701,9 @@ def put(self, request: Request, organization) -> Response:
                 access=request.access,
             )
 
+            if not include_feature_flags:
+                context.pop("features", None)
```
ykamo001, line 705: Is there a way to prevent getting the data in the first place? This helps reduce the amount of bytes transferred on the network/wire and reduced the payload, but we are still taking a hit on the data lookup. Is that something you plan to do on a follow up?

src/sentry/api/endpoints/organization_details.py
```diff
@@ -607,6 +607,9 @@ def get(self, request: Request, organization) -> Response:
 
         context = serialize(organization, request.user, serializer(), access=request.access)
 
+        if not include_feature_flags:
+            context.pop("features", None)
```
ykamo001, line 611: same comment as below for this

src/sentry/api/endpoints/organization_details.py
iamrajjoshi: yeah, i am planning to modify the OrganizationSerializer so that I pass the qparam there and generate the list if the param is true.

ykamo001: Okay, awesome! Is that going to be in this PR or another? Just so I know how to review

iamrajjoshi: ill follow up with another pr.

ykamo001 (approved):

PR #71669, opened 2024-05-29
(CLOSED) fix(issues): Add partition key for status change kafka msg (wrong copy)

aliu39:
(Accidentally reopened from wrong branch)
~~Copied from #71628, which I reverted due to CI problems after merging yday afternoon. Will be around to monitor this one fully~~

~~Previously status change had no key. There are no order guarantees across partitions, resulting in status changes being processed before the first occurrence/group was created.~~


aliu3ntry:
opened from wrong branch

PR #71672, opened 2024-05-29
feat(custom-views): Create new sentry_issueviews table 

### LINKED ISSUE #71592: Custom Views: Create the new table for custom views
We will need to create a new table to store custom views for org members. 

Name: TBD

Schema: 
| Name | Type | Unique Constraints |
|--------|--------|--------|
| `id` | int | Table |
| `name` | string |  |
| `query` | string |  |
| `sort` | string |  |
| `position` | int | org_member_id | 
| `date_added` | datetime | | 
| `org_member_id` | int | |

MichaelSun48:
Implements #71592

This PR creates the ~`sentry_issueviews`~ `sentry_groupsearchview` table for the custom views project. 

**Schema:**
Column Name | Data Type | Unique Constraints | Description
-- | -- | -- | --
id | integer | Table | The default identifier for this table
user_id | integer |   | Id of the user within the org (fk from sentry.user)
organization | integer |  |  Id of the view's org (fk from sentry.organization)
name | string |   | The name of the tab
query | string |   | The search query of the tab
sort | string |   | The sort used for the query
position | integer | user_id, organization | The position of the tab in the tab order
date_added | datetime |   | The date & time the tab was created
date_updated | datetime |   | The date & time the tab was last updated

A full tech spec can be found [here](https://www.notion.so/sentry/Tech-Spec-Backend-Changes-for-New-Saved-Searches-Custom-Tabs-4b4d24b16be04ff5a7addc5f8f32906a) (Internal only)

(I will not merge this PR until the tech spec is completely out of draft stage)


src/sentry/models/issueuserviews.py
```diff
@@ -0,0 +1,40 @@
...
+from sentry.models.savedsearch import SortOptions
+
+
+@region_silo_model
+class IssueUserViews(DefaultFieldsModel):
+    """
+    A model for a user's view of the issue stream
+    """
+
+    __relocation_scope__ = RelocationScope.Organization
```
MichaelSun48, line 17: I copied this over from savedsearches, but I'm not exactly sure what `RelocationScope` means.

src/sentry/models/issueuserviews.py
```diff
@@ -0,0 +1,40 @@
...
+    query = models.TextField()
+    query_sort = models.CharField(
+        max_length=16, default=SortOptions.DATE, choices=SortOptions.as_choices(), null=True
+    )
+    position = models.PositiveSmallIntegerField(null=False)
+    org_member_id = FlexibleForeignKey("sentry.OrganizationMember", null=True)
+
+    class Meta:
+        app_label = "sentry"
+        db_table = "sentry_issueuserviews"
```
MichaelSun48, line 28: I am indifferent to this table name, feel free to suggest alternatives!

src/sentry/models/issueviews.py
```diff
@@ -0,0 +1,40 @@
...
+from sentry.models.savedsearch import SortOptions
+
+
+@region_silo_model
+class IssueViews(DefaultFieldsModel):
+    """
+    A model for a user's view of the issue stream
+    """
+
+    __relocation_scope__ = RelocationScope.Organization
```
MichaelSun48, line 17: I'm not 100% confident this is correct, but this is what was in savedsearches and is what I thought made sense given the description in `scopes.py`. Please correct me if this is wrong!

src/sentry/migrations/0725_create_sentry_issue_views_table.py
```diff
@@ -0,0 +1,67 @@
...
+                ("query", models.TextField()),
+                ("query_sort", models.CharField(default="date", max_length=16, null=True)),
+                ("position", models.PositiveSmallIntegerField()),
+                (
+                    "org_member_id",
+                    sentry.db.models.fields.foreignkey.FlexibleForeignKey(
+                        null=True,
+                        on_delete=django.db.models.deletion.CASCADE,
+                        to="sentry.organizationmember",
+                    ),
```
markstory, line 53: I'll echo the feedback on the spec. Having `organization_id` will be something you'll want in the future. Having a cascade on this column will mean that if an org member is deleted, all their saved views will go with them. This could limit the utility of shared views in the future, as removing a member could delete views that other members are using.

src/sentry/models/issueviews.py
markstory: This looks correct to me. What this does is include the model in self-hosted -> saas relocations, and in region -> region relocations when that is built.

src/sentry/migrations/0725_create_sentry_issue_views_table.py
MichaelSun48: Should we ever support organization level views, I totally agree that it will be necessary to add an `organization_id` and change the cascade behavior depending on the scope of the view. However, for our initial release of this feature, deleting all views upon member removal is the intended behavior since views are scoped to the org member.

wedamija: The main downside of the two column approach is that you don't get that automatic cascade, so if you go that route make sure that you handle it as part of deleting an organization member

src/sentry/models/issueviews.py
```diff
@@ -0,0 +1,40 @@
...
+    name = models.TextField(max_length=128)
+    query = models.TextField()
+    query_sort = models.CharField(
+        max_length=16, default=SortOptions.DATE, choices=SortOptions.as_choices(), null=True
+    )
+    position = models.PositiveSmallIntegerField(null=False)
```
wedamija, line 23: Nit: `null=False` is the default so you can remove it

src/sentry/models/issueviews.py
```diff
@@ -0,0 +1,40 @@
...
+    org_member_id = FlexibleForeignKey("sentry.OrganizationMember", null=True)
```
wedamija, line 24: Nit: Just want to mention this even though you're removing it - this should be `org_member` since you're using a fk here.

src/sentry/migrations/0725_create_sentry_issue_views_table.py
MichaelSun48: Added this as a TODO in the epic for this project, thanks!

src/sentry/models/groupsearchview.py
```diff
@@ -0,0 +1,43 @@
...
+    query = models.TextField()
+    query_sort = models.CharField(
+        max_length=16, default=SortOptions.DATE, choices=SortOptions.as_choices(), null=True
+    )
+    position = models.PositiveSmallIntegerField()
+
+    class Meta:
+        app_label = "sentry"
+        db_table = "sentry_groupsearchview"
+        unique_together = ()
```
wedamija, line 32: Nit: No need to specify this

src/sentry/models/groupsearchview.py
```diff
@@ -0,0 +1,43 @@
...
+        # Two views cannot occupy the same position in an organization user's list of views
+        constraints = [
+            UniqueConstraint(
+                fields=["user_id", "organization_id", "position"],
+                name="sentry_issueviews_unique_view_position_per_org_user",
+            )
+        ]
```
wedamija, line 39: You might need an additional unique here that's basically `organization_id, position where user_id is null`

src/sentry/models/groupsearchview.py
```diff
@@ -0,0 +1,43 @@
...
+@region_silo_model
+class GroupSearchView(DefaultFieldsModel):
+    """
+    A model for a user's view of the issue stream
+    """
+
+    __relocation_scope__ = RelocationScope.Organization
+    name = models.TextField(max_length=128)
+    user_id = HybridCloudForeignKey("sentry.User", on_delete="CASCADE", null=True)
+    organization = FlexibleForeignKey("sentry.Organization", null=True)
```
wedamija, line 21: Can this ever actually be null? It seems like either there will be a view for a `user/org`, or just an org

MichaelSun48: This is actually a mistake, neither `user_id` nor `organization_id` can be null, since we need to know which user created the view, and in which org they created it in. Apologies if this misled you into believing that `user_id` could be null too - there should never be a case (at least in the current product implementation) where a view's `user_id` can be null.

wedamija: Right, I thought we just had a potential future where there are org level default views as well? I guess the not null can be removed when/if that happens

src/sentry/models/groupsearchview.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+
+    __relocation_scope__ = RelocationScope.Organization
+    name = models.TextField(max_length=128)
+    user_id = HybridCloudForeignKey("sentry.User", on_delete="CASCADE")
+    organization = FlexibleForeignKey("sentry.Organization")
+
+    query = models.TextField()
+    query_sort = models.CharField(
+        max_length=16, default=SortOptions.DATE, choices=SortOptions.as_choices(), null=True
```
wedamija, line 25: Will this ever be null?

MichaelSun48: Yes - an empty search query for the view is a valid query

wedamija: Since you have a default here, when will this sort ever be null though?

src/sentry/models/groupsearchview.py
MichaelSun48: Ah I see what you were thinking now - I think if we implement that feature, we would have to add a `scope="user"|"org"` column, since we still want to know which user created a default org level view.

src/sentry/models/groupsearchview.py
MichaelSun48: 🤦 My bad I read the code wrong and thought the null was applying to the query, not the query_sort. You're right - since there is a default sort, `query_sort` should never be null. Fixing that now.

wedamija (approved):

src/sentry/models/groupsearchview.py
```diff
@@ -0,0 +1,43 @@
...
+    position = models.PositiveSmallIntegerField()
+
+    class Meta:
+        app_label = "sentry"
+        db_table = "sentry_groupsearchview"
+        unique_together = ()
+        # Two views cannot occupy the same position in an organization user's list of views
+        constraints = [
+            UniqueConstraint(
+                fields=["user_id", "organization_id", "position"],
```
wedamija, line 35: Btw just a little gotcha with this:

Just say you have rows like
```
id, user_id, org_id, position
1, 100, 1000, 0
2, 100, 1000, 1
```
Now let's say you want to switch the position of these rows. When you attempt the update, you'll run into the constraint blocking you, since you're do `update ... where id = 1 set position = 1`, `update ... where id = 2 set position = 0`, which violates the constraint.

You can mark the constraint deferrable to avoid this https://docs.djangoproject.com/en/5.0/ref/models/constraints/#deferrable

MichaelSun48: This just saved me a huge headache down the line thank you! 🙏  Adding that in now

src/sentry/migrations/0725_create_sentry_groupsearchview_table.py
```diff
@@ -0,0 +1,86 @@
...
+                    ),
+                ),
+            ],
+            options={
+                "db_table": "sentry_groupsearchview",
+            },
+        ),
+        migrations.SeparateDatabaseAndState(
+            database_operations=[
+                migrations.RunSQL(
```
wedamija, line 68: You need to add `hints={"tables": ["sentry_groupsearchview"]},` as a param to `RunSQL`

wedamija (approved):

src/sentry/models/groupsearchview.py
markstory: You'll need to do any position swaps in a transaction or postgres will be sad.

markstory (approved):

malwilley (approved):

PR #71677, opened 2024-05-29
chore(hybridcloud) Reduce outbox generation for orgauthtoken

markstory:
We currently emit two outbox messages for every request that uses an OrgAuthToken. Often the usage patterns for org auth tokens are many requests close together in time. By only emitting one outbox per unique project set per minute we should be able to lower the number of redundant outboxes that are created without greatly impacting the accuracy of the last used time and project ids.

Refs HC-1191

### DIFF
--- a/src/sentry/models/orgauthtoken.py
+++ b/src/sentry/models/orgauthtoken.py
@@ -152,13 +152,13 @@ def get_org_auth_token_id_from_auth(auth: object) -> int | None:
 def update_org_auth_token_last_used(auth: object, project_ids: list[int]):
     org_auth_token_id = get_org_auth_token_id_from_auth(auth)
     organization_id = getattr(auth, "organization_id", None)
-    if not org_auth_token_id and not organization_id:
+    if org_auth_token_id is None or organization_id is None:
         return
 
     # Debounce updates, as we often get bursts of requests when customer
     # run CI or deploys and we don't need second level precision here.
     # We vary on the project ids so that unique requests still make updates
-    project_segment = ",".join([str(i) for i in project_ids])
+    project_segment = ",".join(str(i) for i in project_ids)
     recent_key = f"orgauthtoken:{org_auth_token_id}:last_update:{project_segment}"
     if cache.get(recent_key):
         return
--- a/src/sentry/models/orgauthtoken.py
+++ b/src/sentry/models/orgauthtoken.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import ClassVar, Self
 
+from django.core.cache import cache
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils import timezone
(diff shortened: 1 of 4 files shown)


src/sentry/models/orgauthtoken.py
```diff
@@ -151,10 +152,21 @@ def get_org_auth_token_id_from_auth(auth: object) -> int | None:
...
     org_auth_token_id = get_org_auth_token_id_from_auth(auth)
     organization_id = getattr(auth, "organization_id", None)
-    if org_auth_token_id is not None and organization_id is not None:
-        orgauthtoken_service.update_orgauthtoken(
-            organization_id=organization_id,
-            org_auth_token_id=org_auth_token_id,
-            date_last_used=timezone.now(),
-            project_last_used_id=project_ids[0] if len(project_ids) > 0 else None,
-        )
+    if not org_auth_token_id and not organization_id:
```
szokeasaurusrex, line 155: Why are we no longer checking these values explicitly against `None`, like we are in the previous version? The behavior here could be different if either value is false-y but not `None`; is this behavior change intentional?

src/sentry/models/orgauthtoken.py
```diff
@@ -151,10 +152,21 @@ def get_org_auth_token_id_from_auth(auth: object) -> int | None:
...
+        return
+
+    # Debounce updates, as we often get bursts of requests when customer
+    # run CI or deploys and we don't need second level precision here.
+    # We vary on the project ids so that unique requests still make updates
+    project_segment = ",".join([str(i) for i in project_ids])
```
szokeasaurusrex, line 161:
```suggestion
    project_segment = ",".join(str(i) for i in project_ids)
```

`join` takes an iteratable, so we don't need to build an array here

tests/sentry/api/endpoints/test_organization_releases.py
```diff
@@ -1782,7 +1782,7 @@ def test_org_auth_token(self):
             good_token_str = generate_token(org.slug, "")
             OrgAuthToken.objects.create(
                 organization_id=org.id,
-                name="token 1",
+                name="token 2",
                 token_hashed=hash_token(good_token_str),
                 token_last_characters="ABCD",
                 scope_list=["org:ci"],
```
szokeasaurusrex, line 1788: Why are these changes needed? And, should we perhaps add a test for the caching functionality?

szokeasaurusrex (commented):
Requested some minor changes and left some questions, but overall, I think limiting the updates to once per minute seems like a good idea.

tests/sentry/api/endpoints/test_organization_releases.py
markstory: There were two tokens with a name of 'token 1' for the same org which made debugging harder.

src/sentry/models/orgauthtoken.py
markstory: Projects and orgs cannot have id values of 0. I'll put the none checks back though.

szokeasaurusrex (approved):
Looks good to me!

PR #71683, opened 2024-05-29
remove flag and keep enabled logic

ykamo001:
Feature has been GA'd, so we can remove FF code and keep the enabled logic


tests/sentry/integrations/slack/webhooks/actions/test_status.py
ykamo001, line 1: Removed the tests that were testing the disabled logic which we no longer have

sentaur-athena (approved):
It's best practice to let the feature flag hang around a little longer after launch to turn it off if needed. I'll let you make the decision though since you have better context.

ykamo001:
> It's best practice to let the feature flag hang around a little longer after launch to turn it off if needed. I'll let you make the decision though since you have better context.

Yup, totally fair! Just want to keep this ready to go

cathteng (approved):

PR #71684, opened 2024-05-29
ref(guideStore): Remove state mutation from guide store

scttcper:
- Makes state readonly and removes mutation of state 
- Adds types for previously untyped functions.


malwilley (approved):
Love this

PR #71688, opened 2024-05-29
feat(data-consent): Backend changes for data consent on org creation

roggenkemper:
this pr adds in the backend changes needed to support consenting to the aggregated data consent option on org creation. it also adds in analytics so we can track how many orgs are created with this consent.


malwilley (commented):
Org creation flow is pretty important, so probably good to add a test to cover those lines?

malwilley (approved):

PR #71689, opened 2024-05-29
feat(data-consent): Add data consent check to organization creation

roggenkemper:
this pr adds in a check for aggregate data consent to the organization creation. right now, it is hidden behind a query parameter so we can test it before GA'ing it. Merge after https://github.com/getsentry/sentry/pull/71688


armenzg (approved):
Any tests needed for this?

roggenkemper:
> Any tests needed for this?

I added tests in the backend pr, where the actual option changing is going to happen

static/app/views/organizationCreate/index.tsx
```diff
@@ -119,18 +125,35 @@ function OrganizationCreate() {
...
+                inline={false}
+                stacked
+                required
+              />
+            </TermsWrapper>
+          )}
+          {hasDataConsent && !isSelfHosted && (
             <CheckboxField
-              name="agreeTerms"
+              name="aggregatedDataConsent"
```
malwilley, line 146: It works to use the `isSelfHosted` flag, but would it make more sense for this to be in getsentry?

malwilley (approved):
Otherwise looks good

static/app/views/organizationCreate/index.tsx
```diff
@@ -147,3 +161,7 @@ function OrganizationCreate() {
 }
 
 export default OrganizationCreate;
+
+const TermsWrapper = styled('div')<{hasDataConsent?: boolean}>`
+  margin-bottom: ${p => (p.hasDataConsent ? '0' : '16px')};
```
malwilley, line 166: Instead of this conditional margin, could you do a flex wrapper around both with `gap: ${space(2)};`

roggenkemper: because of how the form is set up, each checkbox by default has 16px bottom margin - my goal was to get rid of that for the terms one so that there isn't a gap if the data consent check is there. the flex wrapper i think would do the opposite of what i'm intending to (add the gap if the data consent check is there)

malwilley: Ah yeah you're right. One other thing you could try is a css selector, but that might be just as gnarly

PR #71691, opened 2024-05-29
fix(typing): Remove test in test_post_process

snigdhas:
https://github.com/getsentry/sentry/pull/71603#discussion_r1619264183

### DIFF
--- a/tests/sentry/tasks/test_post_process.py
+++ b/tests/sentry/tasks/test_post_process.py
@@ -2001,22 +2001,6 @@ class ReplayLinkageTestMixin(BasePostProgressGroupMixin):
         assert kafka_producer.return_value.publish.call_count == 0
         incr.assert_any_call("post_process.process_replay_link.id_sampled")
 
-    def test_0_sample_rate_replays(self, incr, kafka_producer, kafka_publisher):
-        event = self.create_event(
-            data={"message": "testing"},
-            project_id=self.project.id,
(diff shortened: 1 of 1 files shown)


tests/sentry/tasks/test_post_process.py
```diff
@@ -2015,7 +2015,7 @@ def test_0_sample_rate_replays(self, incr, kafka_producer, kafka_publisher):
         )
         assert kafka_producer.return_value.publish.call_count == 0
         for args, _ in incr.call_args_list:
-            assert args != "post_process.process_replay_link.id_sampled"
+            assert args != ("post_process.process_replay_link.id_sampled")
```
JoshFerge, line 2018: do you need to add a comma as this is a single element tuple? https://www.w3schools.com/python/gloss_python_tuple_one_item.asp#:~:text=To%20create%20a%20tuple%20with,the%20variable%20as%20a%20tuple.

asottile-sentry: indeed -- the parens here aren't doing anything

snigdhas: HM this test was broken before the typing changes then

removing this test since it used to test feature-flag dependent behavior for a flag that has since been removed. The test covers the same case as the test right before it.

asottile-sentry (approved):

PR #71695, opened 2024-05-29
feat(context): Add icon to new context cards, remove summary code

leeandher:
Context cards on the issues details page will now display icons if available:

image

I built this with identical logic to the context summary code which is now unusued and removed. I was able to remove the entire `contextSummary/` directory, and put all the utils into `context/` instead.

Follow up to https://github.com/getsentry/sentry/pull/71403


static/app/components/keyValueData/card.tsx
```diff
@@ -193,10 +193,9 @@ const CardPanel = styled(Panel)`
   font-size: ${p => p.theme.fontSizeSmall};
 `;
 
-const Title = styled('p')`
```
scttcper, line 196: yes

scttcper (approved):

PR #71697, opened 2024-05-29
ref(mqb): Remove the flag for defaulting tags

wmak:
- This flag has been GA'ed for a while now so lets just default it now since we should always do this


mjq (approved):
👍

src/sentry/search/events/builder/metrics.py
```diff
@@ -693,13 +666,17 @@ def resolve_tag_key(self, value: str) -> int | str | None:
...
-            else:
-                raise IncompatibleMetricsQuery(f"{value} is not a tag in the metrics dataset")
-        else:
+        # Attempt to resolve tag keys, for the metrics preformance datasets we use an allowlist so ondemand keys don't
+        # overlap. If you don't want this functionality for your dataset set valid_tags to an empty sequence
+        # Ondemand should always resolve the metric_index
+        if (
+            (len(self.valid_tags) > 0 and value in self.valid_tags)
+            or len(self.valid_tags) == 0
+            or self.use_on_demand
```
Zylphrex, line 675: Nit: check the override cases first for better legibility and remove the redundant len > 0 check

```suggestion
            len(self.valid_tags) == 0
            or self.use_on_demand
            or value in self.valid_tags
```

Zylphrex (approved):

PR #71702, opened 2024-05-29
feat(crons): Allow for deleting processing error by type

davidenwang:
Add functionality into the `processing_errors/manager.py` to bulk delete processing errors by type. This will allow a user to bulk dismiss a set of processing errors such as this set of 10 that are all the same:
image

This is mostly straightforward, but problems arise when a single check-in processing error has multiple errors inside of it:
```
CheckinProcessingError {
  errors: [ProcessingError(type: 1), ProcessingError(type: 2), etc...]
  id: ...
  checkin: {...}
}
```
If we wanted to just remove all errors with `type: 1`, we go into the processing error, remove the matching type, and then update the key/value in redis. This has the unfortunate side effect of resetting the key's TTL. 

If this is a problem, we can execute a command to fetch the TTL for the existing key and keep it the same. 
There is also an option `keepttl` which seems to exist in more updated versions of redis. Both solutions I figured might not be worth it, so I kept it as is for now.

### DIFF
--- a/src/sentry/monitors/processing_errors/manager.py
+++ b/src/sentry/monitors/processing_errors/manager.py
@@ -16,7 +16,7 @@ from sentry.monitors.models import Monitor
 from sentry.monitors.types import CheckinItem
 from sentry.utils import json, metrics, redis
 
-from .errors import CheckinProcessingError, ProcessingErrorsException
+from .errors import CheckinProcessingError, ProcessingErrorsException, ProcessingErrorType
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


src/sentry/monitors/processing_errors/manager.py
```diff
@@ -94,6 +94,30 @@ def _delete_for_entity(entity_identifier: str, uuid: uuid.UUID) -> None:
     pipeline.execute()
 
 
+def _delete_for_entity_by_type(entity_identifier: str, type: ProcessingErrorType) -> None:
+    checkin_errors = _get_for_entities([entity_identifier])
+    redis = _get_cluster()
+    pipeline = redis.pipeline()
+    for checkin_error in checkin_errors:
+        errors = checkin_error.errors
+        if not any([error["type"] == type for error in errors]):
```
evanpurkhiser, line 103:
```suggestion
        if not any(error["type"] == type for error in errors):
```

nit, you don't need the list here, can just pass the generator directly to `any`

davidenwang: oh very nice

evanpurkhiser (commented):
Looks good to me, will let dan have the final say.

wedamija (approved):
Looks good to me. Makes me think maybe we should just be storing this stuff differently, since you're now really displaying grouped data rather than data related to a specific checkin failing

PR #71704, opened 2024-05-29
feat(perf): Add span searchbar to Spans tab 

0Calories:
Updates the Search bar in the Spans tab of Transaction Summary. This PR leverages the new suggested tags provided by the Trace Explorer, so that mainly relevant attributes are provided.

The suggested tag set for this page may need to be refined, as there are many tags that cannot be used here as the table is using the metrics dataset. For example, we cannot filter by `span.duration` because the table is making use of aggregates.

image

### DIFF
--- a/static/app/views/performance/transactionSummary/transactionSpans/content.tsx
+++ b/static/app/views/performance/transactionSummary/transactionSpans/content.tsx
@@ -11,6 +11,7 @@ import {EnvironmentPageFilter} from 'sentry/components/organizations/environment
 import PageFilterBar from 'sentry/components/organizations/pageFilterBar';
 import {normalizeDateTimeParams} from 'sentry/components/organizations/pageFilters/parse';
 import Pagination from 'sentry/components/pagination';
+import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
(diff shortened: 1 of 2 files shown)


static/app/views/performance/transactionSummary/transactionSpans/content.tsx
```diff
@@ -233,8 +245,11 @@ function SpansContentV2(props: Props) {
         <StyledSearchBar
           organization={organization}
           projectIds={eventView.project}
-          query={''}
+          query={query.query}
           fields={eventView.fields}
+          placeholder={t('Search for span attributes')}
+          supportedTags={supportedTags}
+          dataset={DiscoverDatasets.SPANS_INDEXED}
```
KevinL10, line 252: maybe add a comment here to explain why we're using SPANS_INDEXED even though the search is querying SPANS_METRICS

KevinL10 (approved):

static/app/views/performance/transactionSummary/transactionSpans/content.tsx
0Calories: Good point, I guess the reason isn't immediately obvious

PR #71707, opened 2024-05-29
fix(traces): Change query timeout message

k-fish:
### Summary
Spans / traces have different constraints and timeout for different reasons than transactions.


mjq (approved):
Typo in error message I think, otherwise LGTM

PR #71710, opened 2024-05-29
ref(querybuilder): Clean up the way we load configs

wmak:
- Instead of a giant if statement for every dataset in the base query builder, set the config in a load config function in each builder


src/sentry/search/events/builder/profile_functions.py
```diff
@@ -19,6 +20,8 @@
 
 
 class ProfileFunctionsQueryBuilderProtocol(Protocol):
+    config_class: type[DatasetConfig] = ProfileFunctionsDatasetConfig
```
Zylphrex, line 23: Does this do anything?

src/sentry/search/events/builder/discover.py
```diff
@@ -344,54 +332,16 @@ def resolve_query(
...
+        self.field_alias_converter = self.config.field_alias_converter
+        self.function_converter = self.config.function_converter
+        self.search_filter_converter = self.config.search_filter_converter
+        self.orderby_converter = self.config.orderby_converter
+        self.config_parsed = True
 
-        return field_alias_converter, function_converter, search_filter_converter, orderby_converter
+    def load_config(self) -> None:
+        self.config = self.config_class(self)
+        self.parse_config()
```
Zylphrex, line 344: This seems like an odd pattern to have the child class call `self.parse_config()` every time it implements `load_config`.

Can we try something like

```python
def load_config(self) -> DatasetConfig:
  return self.config_class(self)
  
def __init__(self, ...):
  # other stuff
  self.config = self.load_config()
  self.parse_config()
```

That way child classes can just return an instance of `DatasetConfig`?

src/sentry/search/events/builder/profile_functions.py
wmak: no, forgot to remove it thx

src/sentry/search/events/builder/discover.py
wmak: Agreed

Zylphrex (approved):

vartec (approved):

PR #71711, opened 2024-05-29
ref(pinned-searches): Save new/updated default searches to new `groupsearchview` table

### LINKED ISSUE #71590: Custom Views: Augment the `organizations/.../pinned-searches/` endpoint to save to/update new custom views database
We will need to update the `organizations/.../pinned-searches/` endpoint to also save and update entries in the new database for custom views. This is because pinned searches are being replaced by custom views, so before we migrate existing pinned searches to the custom views database, we will need to update the endpoint that currently saves pinned searches to double save to the new database as well.

MichaelSun48:
This PR augments the `.../pinned-searches` endpoint. This endpoint previously saved a new pinned (default) search to only the `sentry_savedsearches` table with `visibility="owner_pinned"`. This PR adds new logic to also save a new or updated default to the upcoming custom views table (`sentry_issueuserviews`). This is because we will want to persist user's default searches as a new tab, so we will need to start actively saving default searches to this new table.

### DIFF
--- a/src/sentry/api/endpoints/organization_pinned_searches.py
+++ b/src/sentry/api/endpoints/organization_pinned_searches.py
@@ -7,6 +7,7 @@ from sentry.api.api_publish_status import ApiPublishStatus
 from sentry.api.base import region_silo_endpoint
 from sentry.api.bases.organization import OrganizationEndpoint, OrganizationPinnedSearchPermission
 from sentry.api.serializers import serialize
+from sentry.models.groupsearchview import GroupSearchView
 from sentry.models.savedsearch import SavedSearch, SortOptions, Visibility
 from sentry.models.search_common import SearchType
 
@@ -53,7 +54,7 @@ class OrganizationPinnedSearchEndpoint(OrganizationEndpoint):
             values={"query": result["query"], "sort": result["sort"]},
         )
 
-        GroupSearchView.objects.create_or_update(  # type: ignore
+        GroupSearchView.objects.create_or_update(
             organizaiton=organization,
             user_id=request.user.id,
             values={
--- a/src/sentry/api/endpoints/organization_pinned_searches.py
+++ b/src/sentry/api/endpoints/organization_pinned_searches.py
@@ -55,13 +55,13 @@ class OrganizationPinnedSearchEndpoint(OrganizationEndpoint):
         )
 
         GroupSearchView.objects.create_or_update(
-            organizaiton=organization,
+            organization=organization,
             user_id=request.user.id,
+            position=0,
             values={
                 "name": "Default Search",
                 "query": result["query"],
                 "query_sort": result["sort"],
-                "position": 0,  # 0 position indicates default view
             },
         )
         pinned_search = SavedSearch.objects.get(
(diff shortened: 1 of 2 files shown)


src/sentry/api/endpoints/organization_pinned_searches.py
```diff
@@ -52,6 +53,17 @@ def put(self, request: Request, organization) -> Response:
             visibility=Visibility.OWNER_PINNED,
             values={"query": result["query"], "sort": result["sort"]},
         )
+
+        GroupSearchView.objects.create_or_update(
+            organizaiton=organization,
```
wedamija, line 58: `organization`

src/sentry/api/endpoints/organization_pinned_searches.py
```diff
@@ -52,6 +53,17 @@ def put(self, request: Request, organization) -> Response:
...
+            user_id=request.user.id,
+            values={
+                "name": "Default Search",
+                "query": result["query"],
+                "query_sort": result["sort"],
+                "position": 0,  # 0 position indicates default view
```
wedamija, line 64: I think the position should be part of the select criteria, since the unique is `org, user, position`

src/sentry/api/endpoints/organization_pinned_searches.py
mrduncan: Let's make sure there is a test in place which would catch this, if there isn't already

MichaelSun48: 🤦 
Thanks for catching this! Fixing now

malwilley (commented):
We also want to delete the entry when deleting the pinned search, will that be done here or in a followup PR?

MichaelSun48:
> We also want to delete the entry when deleting the pinned search, will that be done here or in a followup PR?

Good point, I think the scope of this PR is fairly small so I'll add it in here.

tests/sentry/api/endpoints/test_organization_pinned_searches.py
```diff
@@ -35,6 +36,16 @@ def test(self):
             visibility=Visibility.OWNER_PINNED,
         ).exists()
 
+        # TODO(msun): Remove this when custom views are GA'd
```
malwilley, line 39: I think we can remove this entire endpoint when custom views are GA

malwilley (approved):

src/sentry/api/endpoints/organization_pinned_searches.py
```diff
@@ -52,6 +53,18 @@ def put(self, request: Request, organization) -> Response:
             visibility=Visibility.OWNER_PINNED,
             values={"query": result["query"], "sort": result["sort"]},
         )
+
+        # This entire endpoint will be removed once custom views are GA'd
+        GroupSearchView.objects.create_or_update(
```
markstory, line 58: Will we need similar logic for other saved issue searches?

MichaelSun48: No, the intention is to only port over pinned saved searches to custom views - remaining saved searches will be persisted elsewhere

PR #71713, opened 2024-05-29
ref(similarity-embedding): Exclude old groups from backfill

jangjodi:
Exclude groups that are over 90 days old from the record backfill


JoshFerge (approved):

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -127,7 +127,12 @@ def backfill_seer_grouping_records(
             Group.objects.bulk_update(groups_seen_once, ["data"])
 
     group_id_message_data = (
-        Group.objects.filter(project_id=project.id, type=ErrorGroupType.type_id, times_seen__gt=1)
+        Group.objects.filter(
+            project_id=project.id,
+            type=ErrorGroupType.type_id,
+            times_seen__gt=1,
+            last_seen__gt=(datetime.now(UTC) - timedelta(days=90)),
```
mrduncan, line 134: Shouldn't be a problem in the initial cohort but should this account for customers who have custom retention periods?

jangjodi: we've seen a delay even in s4s projects, where the groups are hanging around for a bit longer but have no events

JoshFerge: yeah, some have 30d retention periods

PR #71715, opened 2024-05-29
chore(severity): Remove IssueSeverityFilter

snigdhas:
Removing the IssueSeverityFilter and the corresponding `projects:first-event-severity-alerting` flag. 

Removing the handler in https://github.com/getsentry/getsentry/pull/14118


src/sentry/rules/filters/issue_severity.py
```diff
@@ -1,76 +0,0 @@
...
-}
-CATEGORY_CHOICES = OrderedDict([(f"{gc.value}", str(gc.name).title()) for gc in GroupCategory])
-
-
-class IssueSeverityForm(forms.Form):
-    value = forms.NumberInput()
-
-
-class IssueSeverityFilter(EventFilter):
-    id = "sentry.rules.filters.issue_severity.IssueSeverityFilter"
```
mrduncan, line 28: Are there any existing rules using this filter which need to be cleaned up?

snigdhas: nope, I cleaned up these - there were only two, both for the `sentry` project

mrduncan (approved):

PR #71717, opened 2024-05-29
ref: remove unnecessary build_query_builder helper

nhsiehgit:
the `build_query_builder` helper is just taking the passed `entity_subscription` and invoking it with the remaining args.

This is unnecessary and confusing.

### DIFF
--- a/src/sentry/snuba/tasks.py
+++ b/src/sentry/snuba/tasks.py
@@ -2,18 +2,14 @@ from __future__ import annotations
 
 import logging
 from datetime import timedelta
-from typing import TYPE_CHECKING
 
 import orjson
 import sentry_sdk
 from django.utils import timezone
 
 from sentry import features
-from sentry.models.environment import Environment
-from sentry.search.events.types import ParamsType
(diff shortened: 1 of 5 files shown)


src/sentry/snuba/tasks.py
```diff
@@ -195,16 +188,6 @@ def delete_subscription_from_snuba(query_subscription_id, **kwargs):
...
 
 
-def build_query_builder(
-    entity_subscription: BaseEntitySubscription,
-    query: str,
-    project_ids: list[int],
-    environment: Environment | None,
-    params: ParamsType | None = None,
-) -> QueryBuilder:
-    return entity_subscription.build_query_builder(query, project_ids, environment, params)
```
nhsiehgit, line 205: literally all this helper method is doing is invoking `build_query_builder` on the passed `entity_subscription` 🤔 

This helper is unnecessary and can be removed

saponifi3d: 👍 makes sense to remove, especially with the entity_subscription being an argument 😅

saponifi3d (approved):
:shipit:

PR #71719, opened 2024-05-29
feat(hybridcloud) Add caching decorator for many records

markstory:
The user_service.get_many() call is the highest volume RPC call we have, and it could also benefit from caching. Currently the get_user() method has a hit rate ~70%, and I think get_many can have a high hit rate as well.

These changes only modify the existing single record cache read and add the plumbing for multiple record cache reads.

Refs HC-1185


src/sentry/services/hybrid_cloud/user/service.py
```diff
@@ -193,7 +192,6 @@ def get_user(self, user_id: int) -> RpcUser | None:
 
         :param user_id: The user to fetch
         """
-        metrics.incr("user_service.get_user.call")
```
markstory, line 196: I don't need these metrics anymore. They have proven the theory I had on cache hit rates.

corps (commented):
I have some, skepticism.
is there any chance that many of these get_many calls could just be built differently or use caches on the logic side instead of the service side?

markstory:
> is there any chance that many of these get_many calls could just be built differently or use caches on the logic side instead of the service side?

Of course we could do that as well. It would require roughly the same blend cache and RPC reads to be done though.

My current plan was to introduce a new method - `user_service.get_many_by_id(user_ids: list[int]): list[RpcUser]` - that only accepts a list of userids. I thought this avoids the problematic scenarios we've talked about in the past around complex queries being uncachable (they still are) but will let us skip a ton of RPC calls as most `user_service.get_many` calls are using `user_ids` as the only filter option.

src/sentry/hybridcloud/rpc/services/caching/service.py
```diff
@@ -91,23 +98,137 @@ def resolve_from(
...
+        self.base_key = base_key
+        self.silo_mode = silo_mode
+        self.cb = cb
+        self.type_ = t
+        self.timeout = timeout
+
+    def __call__(self, ids: list[int]) -> list[_R]:
+        if (
+            SiloMode.get_current_mode() != self.silo_mode
+            and SiloMode.get_current_mode() != SiloMode.MONOLITH
```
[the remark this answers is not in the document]
markstory: Yes, going to the DB and bypassing the cache should be fine in a monolith mode. The DB reads aren't the problematic part of our RPC calls it is the network overhead that we want to reduce.

src/sentry/hybridcloud/rpc/services/caching/service.py
```diff
@@ -91,23 +98,137 @@ def resolve_from(
...
+
+        # This result could have different order than missing_object_ids, or have gaps
+        cb_result = self.cb(list(missing.keys()))
+        for record in cb_result:
+            # TODO(hybridcloud) The types/interfaces don't make reading this attribute safe.
+            # We rely on a convention of records having `id` for now. In the future
+            # this could become a decorator parameter intead.
+            record_id = getattr(record, "id")
+            if record_id is None:
+                continue
```
GabeVillalobos, line 181: Couldn't this behavior vary per RPC service? I'd imagine some services care about all queried records existing, while others do not.

tests/sentry/hybridcloud/test_caching.py
```diff
@@ -158,3 +161,118 @@ def cache_death_event() -> Generator[None, None, None]:
...
+    assert len(all_users) == len(users)
+    for i, user in enumerate(all_users):
+        assert user_ids[i] == user.id, "Results should be ordered based on id list"
+
+
+@django_db_all(transaction=True)
+def test_caching_many_missing_ids() -> None:
+    cache.clear()
+
+    @back_with_silo_cache_many(base_key="get_users", silo_mode=SiloMode.REGION, t=RpcUser)
```
GabeVillalobos, line 229: Would it be helpful if we assert on the contents of the cache itself within testing?

tests/sentry/hybridcloud/test_caching.py
```diff
@@ -158,3 +161,118 @@ def cache_death_event() -> Generator[None, None, None]:
...
+            user.update(username=user.username + "moo")
+
+    # Does not include updates made to db
+    after_update_wrapped = get_users(user_ids)
+    for u in after_update_wrapped:
+        assert not u.username.endswith("moo")
+        # Clear cache simulating outbox logic
+        region_caching_service.clear_key(
+            region_name=get_local_region().name, key=get_users.key_from(u.id)
+        )
```
GabeVillalobos, line 194: Do we have any integration style tests that assert this behavior e2e with actual outbox/replication logic? I saw the testing in getsentry/sentry#71783 that validates we're calling the cache clear logic, but it's probably valuable to have realistic testing scenarios for this cache behavior if we don't already.

markstory: I could add that integration testing alongside the first usage of these methods. Without that real usage it is hard to have a end to end scenario.

GabeVillalobos: Agreed, that sounds good to me!

src/sentry/hybridcloud/rpc/services/caching/service.py
markstory: Yeah that could happen. We could set `found[record_id] = None` and let the caller figure out what is going on. 🤷 I chose to go with the skip behavior, as the behavior of `user_service.get_many()` is to return nothing for ids that don't exist. We have similar behavior in other get_many methods.

GabeVillalobos (approved):

PR #65941, opened 2024-02-27
feat: hash support for user auth token middleware

mdtro:
- Update the `UserAuthTokenAuthentication` middleware to support hashed token values. As tokens are used, it will store the appropriate hash value.
- Introduce a temporary option, `apitoken.save-hash-on-create`. This will be used in the model logic and in pre and post backfill migration tests in the future.

PR #71728, opened 2024-05-29
migration: backfill apitoken hashed values

mdtro:
Supports https://github.com/getsentry/rfcs/pull/32

We've been hashing tokens as they are used to authenticate (https://github.com/getsentry/sentry/pull/65941), but it's started to level out. This is a backfill migration to fill in all of the hashed values for the remaining tokens.

Huge thank you to @markstory @wedamija and @GabeVillalobos for helping with the migration test! 🙏


src/sentry/migrations/0726_apitoken_backfill_hashes.py
```diff
@@ -0,0 +1,83 @@
...
+logger = logging.getLogger(__name__)
+
+
+def backfill_hash_values(apps: StateApps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    ApiToken = apps.get_model("sentry", "ApiToken")
+    ControlOutbox = apps.get_model("sentry", "ControlOutbox")
+    try:
+        from sentry.models.outbox import OutboxCategory, OutboxScope
+        from sentry.silo.safety import unguarded_write
+        from sentry.types.region import find_regions_for_user
```
wedamija, line 22: This function uses database models directly and so will fail to run if those models are modified by a future migration.

It's a simple set of functions, so I'd recommend copying them into this migration file and passing the models from `apps.get_model` to them so that this won't break.

src/sentry/migrations/0726_apitoken_backfill_hashes.py
```diff
@@ -0,0 +1,83 @@
...
+from sentry.utils.query import RangeQuerySetWrapperWithProgressBar
+
+logger = logging.getLogger(__name__)
+
+
+def backfill_hash_values(apps: StateApps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    ApiToken = apps.get_model("sentry", "ApiToken")
+    ControlOutbox = apps.get_model("sentry", "ControlOutbox")
+    try:
+        from sentry.models.outbox import OutboxCategory, OutboxScope
```
wedamija, line 20: Generally we prefer to import as little as possible here, so I'd also recommend copy/pasting these into this migration

wedamija (commented):
This is looking good to me, just a couple of small comments to help make sure this migration doesn't break in the future

src/sentry/migrations/0726_apitoken_backfill_hashes.py
mdtro: Hmm, so when I attempt to import `OutboxCategory` via `apps.get_models(...)` it's unable to import it. 

```
E   LookupError: App 'sentry' doesn't have a 'OutboxCategory' model.
```

When I look at the output of `apps.get_models()` I don't see anything from `sentry.models.outbox`. Is there a proper way to do this?

```
...
 <class 'sentry.models.notificationmessage.NotificationMessage'>,
 <class 'sentry.models.notificationsettingoption.NotificationSettingOption'>,
 <class 'sentry.models.notificationsettingprovider.NotificationSettingProvider'>,
 <class 'sentry.models.organizationaccessrequest.OrganizationAccessRequest'>,
 <class 'sentry.models.organizationmemberteam.OrganizationMemberTeam'>,
 <class 'sentry.models.organizationmemberteamreplica.OrganizationMemberTeamReplica'>,
 <class 'sentry.models.organizationonboardingtask.OrganizationOnboardingTask'>,
 <class 'sentry.models.orgauthtoken.OrgAuthToken'>,
 <class 'sentry.models.platformexternalissue.PlatformExternalIssue'>,
 <class 'sentry.models.processingissue.ProcessingIssue'>,
 <class 'sentry.models.processingissue.EventProcessingIssue'>,
 <class 'sentry.models.projectbookmark.ProjectBookmark'>,
...
```

mdtro: Scratch that, these are not in alphabetical order like I had assumed. There are some available from `sentry.models.outbox`. 

```
<class 'sentry.models.outbox.RegionOutbox'>,
 <class 'sentry.models.outbox.ControlOutbox'>,
```

So I looked deeper and realized `OutboxCategory` and `OutboxScope` aren't Django models, but just regular Python Enums. 🤦 

I assume those are safe to import directly?

src/sentry/migrations/0726_apitoken_backfill_hashes.py
mdtro: Done!

src/sentry/migrations/0726_apitoken_backfill_hashes.py
wedamija: For these I actually literally meant to just copy/paste the code inside the migration. It just prevents them breaking if people change code around later

wedamija (approved):

PR #71730, opened 2024-05-29
chore(api): Make `UserDetailsEndpoint` Private

iamrajjoshi:
This pr marks UserDetailsEndpoint as private. 

The endpoint gives a user's information, including all the orgs they are part of and the authenticators associated with their account. Currently it can only be hit by the user themself & staff. The endpoint requires a user to know their id, which isn't something we ever surface.

For: https://github.com/getsentry/team-core-product-foundations/issues/325


schew2381:
You can actually find your user ID if you go to your settings and look at the API request. Still don't know how useful that would be or who would actually do that plus use this API so seems fine to make everything private

schew2381 (approved):

PR #71731, opened 2024-05-29
migration(custom-views): Backfill groupsearchviews with pinned/default views from savedsearches

### LINKED ISSUE #71593: Custom Views: Backfill custom views with previously pinned tabs as defaults
We will need to migrate existing pinned searches from the `sentry_savedsearches` table (where `visibility="owner_pinned"`) as default views (`position=0`) in the new custom views table. 

This means that for every entry in `sentry_savedsearches` where `visibility="owner_pinned"`, we will need to create a new row in the custom views table where `position=0`.

MichaelSun48:
Implements #71593 

This PR migrates any users' default searches to the new `groupsearchview` table. This is necessary because users' default searches will eventually become the default view in the custom views project. 

The logic is as follows: 
1. Get all saved searches where `visibility="owner_pinned"`. These are the default searches for each user 
2. Save or update the query and sort of the default saved search as a row in the `groupsearchview` table. Users may already have a default custom view if they saved/updated a default search between when #71711 and this PR is merged
3. Also add the prioritized view since this will be the new default view

⚠️ This PR is dependent on #71711 being merged. That PR adds logic to save newly created/updated pinned searches to the `groupsearchview` table. Merging this PR first would result in a window of time where new pinned searches are not saved to the `groupsearchview` table.


src/sentry/migrations/0726_backfill_groupsearchviews_with_pinned_searches.py
```diff
@@ -0,0 +1,56 @@
...
+from sentry.utils.query import RangeQuerySetWrapperWithProgressBar
+
+
+def backfill_groupsearchviews_with_pinned_searches(
+    apps: StateApps, schema_editor: BaseDatabaseSchemaEditor
+) -> None:
+    GroupSearchView = apps.get_model("sentry", "GroupSearchView")
+    SavedSearch = apps.get_model("sentry", "SavedSearch")
+
+    pinned_searches = SavedSearch.objects.filter(visibility="owner_pinned")
```
markstory, line 17: This table is pretty small in saas, so the table scan here should be ok. If it isn't you can remove the condition from the query and implement it as a condition in python code.

src/sentry/migrations/0726_backfill_groupsearchviews_with_pinned_searches.py
```diff
@@ -0,0 +1,56 @@
...
+    # of a table.
+    # Here are some things that make sense to mark as post deployment:
+    # - Large data migrations. Typically we want these to be run manually so that they can be
+    #   monitored and not block the deploy for a long period of time while they run.
+    # - Adding indexes to large tables. Since this can take a long time, we'd generally prefer to
+    #   run this outside deployments so that we don't block them. Note that while adding an index
+    #   is a schema change, it's completely safe to run the operation after the code has deployed.
+    # Once deployed, run these manually via: https://develop.sentry.dev/database-migrations/#migration-deployment
+
+    is_post_deployment = False
```
markstory, line 44:
```suggestion
    is_post_deployment = True
```

Databackfills should be run post-deploy. We only have 5s of migration runtime during deploys.

MichaelSun48: Makes sense, thanks!

src/sentry/migrations/0726_backfill_groupsearchviews_with_pinned_searches.py
MichaelSun48: Gotcha. Just out of curiosity, what is the concern with having the condition within the query instead of just in code? Is it just faster for python to loop through all of the records with an if statement instead of having Postgres query  for it?

MichaelSun48:
Currently waiting on some product decisions before moving forward with this backfill - there is a chance we will give users with a previous saved search more custom views than just their old saved search. I will delete this comment once I get more clarity on that

wedamija (commented):
Looks good to me, but we should mark as `is_post_deployment = True`

src/sentry/migrations/0726_backfill_groupsearchviews_with_pinned_searches.py
wedamija: As well as this, sometimes even if you have an index, when you use `RangeQuerySetWrapper` it adds `where id > x order by id` to all queries. So unless the index also includes `id`as the last column you can also run into problems. Iterating over the whole table takes longer as a whole, but each individual query is pretty cheap

src/sentry/migrations/0728_backfill_groupsearchviews_with_pinned_searches.py
```diff
@@ -0,0 +1,68 @@
...
+            organization=pinned_search.organization,
+            user_id=pinned_search.owner_id,
+            position=0,
+            defaults={
+                "name": "Default Search",
+                "query": pinned_search.query,
+                "query_sort": pinned_search.sort,
+            },
+        )
+        GroupSearchView.objects.create(
```
wedamija, line 31: Should this also be `update_or_create`? Just in case this needs to be re-run, then at least it'd idempotent

MichaelSun48: @wedamija It doesn't need to be - this database is currently only being populated with any new/updated pinned searches, which have position = 0, so creating a new object with position = 1 should not create any conflicts. That being said, changing it to `update_or_create` wouldn't cause any harm either, and if it also has the added benefit for being idempotent in the case of a re-run, then I'm happy to change it just for that benefit. What do you think 


Also just for context since this was literally decided on this morning - The idea behind this addition is that we don't want a user's old pinned search to be the _only_ view they see when the new experience is released, so we also want to persist the current Prioritized view as a second view.

src/sentry/migrations/0729_backfill_groupsearchviews_with_pinned_searches.py
```diff
@@ -0,0 +1,68 @@
+# Generated by Django 5.0.6 on 2024-06-03 17:27
+
+from django.db import migrations
+from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+from django.db.migrations.state import StateApps
+
+from sentry.models.savedsearch import SortOptions
```
wedamija, line 7: I recommend just copy/pasting this into the migration. We usually prefer to avoid importing external code if we can avoid it, so that the migration remains self contained.

src/sentry/migrations/0729_backfill_groupsearchviews_with_pinned_searches.py
```diff
@@ -0,0 +1,68 @@
...
+            },
+        )
+        GroupSearchView.objects.create(
+            organization=pinned_search.organization,
+            user_id=pinned_search.owner_id,
+            position=1,
+            defaults={
+                "Name": "Prioritized",
+                "query": PRIORITIZED_QUERY,
+                "query_sort": SortOptions.DATE.value,
```
wedamija, line 38: `SortOptions` isn't actually an enum, so you don't need `.value` here

tests/sentry/migrations/test_0729_backfill_groupsearchviews_with_pinned_searches.py
```diff
@@ -0,0 +1,31 @@
...
+    def test(self):
+        custom_view = GroupSearchView.objects.get(organization=self.org, user_id=self.user.id)
+
+        assert custom_view
+        assert custom_view.organization == self.org
+        assert custom_view.user_id == self.user.id
+        assert custom_view.position == 0
+        assert custom_view.name == "Default Search"
+        assert custom_view.query == "assigned:me"
+        assert custom_view.query_sort == "date"
```
wedamija, line 31: The test should also check the `postition=1` is created too

wedamija (commented):
If users don't have a pinned search should we be creating a `GroupSearchView` row for them too?

MichaelSun48:
> If users don't have a pinned search should we be creating a GroupSearchView row for them too?

No, the logic for which default rows to return in this case will be handled by the `GET` endpoint in this model. This is partially so we avoid having to create dead rows for inactive users, and so we can more easily change which defaults to give users since it'll just be a PR rather than a migration.

wedamija (approved):

PR #71754, opened 2024-05-30
fix(similarity-embedding): Fix backfill filtering and other logic

jangjodi:
Remove adding metadata for groups where times_seen = 1
When nodestore has an exception, log then raise the exception instead of continuing
When nodestore returns no data, log then continue with the backfill instead of stopping
Fix the ordering by times_seen and id

### DIFF
--- a/src/sentry/tasks/backfill_seer_grouping_records.py
+++ b/src/sentry/tasks/backfill_seer_grouping_records.py
@@ -110,22 +110,6 @@ def backfill_seer_grouping_records(
         )
         return
 
-    if last_processed_index == 0:
-        # Set the metadata of groups where times_seen = 1
-        # Do not set the version number, so we can consider it for future backfills later
-        groups_seen_once = Group.objects.filter(
-            project_id=project_id, type=ErrorGroupType.type_id, times_seen=1
-        )
-        for group in groups_seen_once:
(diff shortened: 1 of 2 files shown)


src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -419,6 +388,7 @@ def lookup_group_data_stacktrace_bulk(
                     "tasks.backfill_seer_grouping_records.bulk_event_lookup_exception",
                     extra=extra,
                 )
+                raise ()
```
JoshFerge, line 391: i don't think this is valid python? i think you should just have `raise` here?

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -492,6 +461,7 @@ def lookup_group_data_stacktrace_single(
                 logger.exception(
                     "tasks.backfill_seer_grouping_records.event_lookup_exception", extra=extra
                 )
+                raise ()
```
JoshFerge, line 464: same

JoshFerge (commented):
looks good pending the `raise ()` comments

JoshFerge (approved):
awesome, thanks!

PR #71758, opened 2024-05-30
fix(issues): Move lazy loading wrappers out of render

scttcper:
fixes an issue when unmounting these components, we should always use lazy() outside of the render path.

fixes https://getsentry.atlassian.net/browse/ISSUE-1703


malwilley (approved):
The greatest bug fix of all time

PR #71762, opened 2024-05-30
feat(priority-alerts): Change the default alert condition to NewHighPriorityIssueCondition

snigdhas:
The default alert condition for a new issue alert will now include both the `NewHighPriorityIssueCondition` and `ExistingHighPriorityIssueCondition` conditions for organizations where the feature is enabled. 

image

### DIFF
--- a/static/app/types/alerts.tsx
+++ b/static/app/types/alerts.tsx
@@ -39,7 +39,7 @@ export const enum IssueAlertConditionType {
   EVENT_FREQUENCY = 'sentry.rules.conditions.event_frequency.EventFrequencyCondition',
   EVENT_UNIQUE_USER_FREQUENCY = 'sentry.rules.conditions.event_frequency.EventUniqueUserFrequencyCondition',
   EVENT_FREQUENCY_PERCENT = 'sentry.rules.conditions.event_frequency.EventFrequencyPercentCondition',
-  NEW_HIGH_PRIORITY_ISSUE = 'sentry.rules.conditions.high_priority_issue.NewIssueCondition',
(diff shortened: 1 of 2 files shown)


static/app/types/alerts.tsx
```diff
@@ -39,7 +39,7 @@ export const enum IssueAlertConditionType {
   EVENT_FREQUENCY = 'sentry.rules.conditions.event_frequency.EventFrequencyCondition',
   EVENT_UNIQUE_USER_FREQUENCY = 'sentry.rules.conditions.event_frequency.EventUniqueUserFrequencyCondition',
   EVENT_FREQUENCY_PERCENT = 'sentry.rules.conditions.event_frequency.EventFrequencyPercentCondition',
-  NEW_HIGH_PRIORITY_ISSUE = 'sentry.rules.conditions.high_priority_issue.NewIssueCondition',
```
scefali, line 42: @snigdhas do we have these in the wild already? Won't such a change break things?

snigdhas: This was a typo, but the constant isn't used anywhere so it should be okay

schew2381:
I believe the string may be incorrect [here](https://github.com/getsentry/sentry/blob/4a7adbbac855e04b89ea59eb733c8e5c3234c192/src/sentry/constants.py#L269), it has `new_high_priority_issue` vs `high_priority_issue` as specified in the [condition class](https://github.com/getsentry/sentry/blob/4ba0cf2a048f1b47941d793a88ebe0dc0ef47601/src/sentry/rules/conditions/new_high_priority_issue.py#L18)

snigdhas:
Great catch, ty @schew2381!

scttcper (approved):

PR #71764, opened 2024-05-30
feat(insights): Wrap sidebar navigation items in `sidebar:navigation-item` hook

gggritso:
Declares a new hook called `sidebar:navigation-item`, and wraps all navigation items in it. Here's the scoop on the hook:

- the hook is a _render prop hook_. It accept a render function child, and passes it some computed properties
- the properties are:
    - `disabled`, which indicates if the child should be "active" or not, whatever that means
    - `AdditionalContent` which supplies some React content to render somewhere alongside the child
    - `Wrapper` which supplies a React wrapper to wrap the child in

The `SidebarItem` component uses this hook by checking the `disabled` property, and if present, removes the `to` on the link (thereby disabling it), rendering the `AdditionalContent` next to the item label, and wrapping _the whole thing_ in `Wrapper`.

The default fallthrough behaviour is that the link is enabled, there's no wrapper, and there's no additional content. i.e., no changes to current behaviour.

Why do all this? We want to be able to have _more control_ over the sidebar items from `getsentry`. We want to:

1. Be able to turn off the entire link item (i.e., disable the link)
2. Wrap the whole link in an upsell dialog
3. Pass it some extra content to render (i.e., the Business icon)

The existing `sidebar:item-label` hook doesn't help us, because it doesn't provide a mechanism for wrapping, or turning off the link! See https://github.com/getsentry/getsentry/pull/14127 for the corresponding `getsentry` hook behaviour

### DIFF
--- a/static/app/components/sidebar/sidebarItem.tsx
+++ b/static/app/components/sidebar/sidebarItem.tsx
@@ -216,7 +216,7 @@ function SidebarItem({
       position={placement}
     >
       <SidebarNavigationItemHook id={id}>
-        {({disabled, AdditionalContent, Wrapper}) => (
+        {({disabled, additionalContent, Wrapper}) => (
           <Wrapper>
             <StyledSidebarItem
               {...props}
@@ -239,7 +239,7 @@ function SidebarItem({
                     <LabelHook id={id}>
                       <TruncatedLabel>{label}</TruncatedLabel>
                       {badges}
-                      {AdditionalContent}
+                      {additionalContent}
                     </LabelHook>
                   </SidebarItemLabel>
                 )}
@@ -313,7 +313,7 @@ const SidebarNavigationItemHook = HookOrDefault({
   defaultComponent: ({children}) =>
     children({
       disabled: false,
-      AdditionalContent: <Fragment />,
+      additionalContent: <Fragment />,
       Wrapper: Fragment,
     }),
 });
--- a/static/app/components/sidebar/sidebarItem.tsx
+++ b/static/app/components/sidebar/sidebarItem.tsx
@@ -313,7 +313,7 @@ const SidebarNavigationItemHook = HookOrDefault({
   defaultComponent: ({children}) =>
     children({
       disabled: false,
-      additionalContent: <Fragment />,
+      additionalContent: null,
       Wrapper: Fragment,
     }),
 });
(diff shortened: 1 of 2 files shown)


static/app/types/hooks.tsx
```diff
@@ -646,6 +647,20 @@ type InviteButtonCustomizationHook = () => React.ComponentType<{
...
 }>;
 
+/**
+ * Sidebar navigation item customization allows passing render props to disable
+ * the link, wrap it in an upsell modal, and give it some additional content
+ * (e.g., a Business Icon) to render.
+ */
+type SidebarNavigationItemHook = () => React.ComponentType<{
+  children: (opts: {
+    AdditionalContent: React.ReactElement | null;
```
evanpurkhiser, line 657: Why is this capitalized

gggritso: My bad, it was originally also a renderable, like `Wrapper`! I changed it, but didn't lowercase. I'll update 👍🏻

static/app/components/sidebar/sidebarItem.tsx
```diff
@@ -299,6 +308,16 @@ export function isItemActive(
   );
 }
 
+const SidebarNavigationItemHook = HookOrDefault({
+  hookName: 'sidebar:navigation-item',
+  defaultComponent: ({children}) =>
+    children({
+      disabled: false,
+      AdditionalContent: <Fragment />,
```
evanpurkhiser, line 316: Can just be `null`?

static/app/types/hooks.tsx
```diff
@@ -646,6 +647,20 @@ type InviteButtonCustomizationHook = () => React.ComponentType<{
   organization: Organization;
 }>;
 
+/**
+ * Sidebar navigation item customization allows passing render props to disable
+ * the link, wrap it in an upsell modal, and give it some additional content
+ * (e.g., a Business Icon) to render.
+ */
```
evanpurkhiser, line 654: maybe let's have a TODO above this that sates how we could replace the label hook with this one potentially

evanpurkhiser (approved):

static/app/components/sidebar/sidebarItem.tsx
gggritso: It can! Will change

PR #71765, opened 2024-05-30
feat(autofix): Modify link and styles for autofix github app onboarding

malwilley:
- Removes repo links to draw attention to the GH app link
- Modify link to point to `/installations/new` to save users a click

CleanShot 2024-05-30 at 11 30 10

### DIFF
--- a/static/app/components/modals/autofixSetupModal.tsx
+++ b/static/app/components/modals/autofixSetupModal.tsx
@@ -125,12 +125,10 @@ function GitRepoLink({repo}: {repo: AutofixSetupRepoDefinition}) {
     return (
       <RepoLinkItem>
         <GithubLink>
-          <ExternalLink href={`https://github.com/${repo.owner}/${repo.name}`}>
-            <IconGithub color="linkColor" size="sm" />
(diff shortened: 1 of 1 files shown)


static/app/components/modals/autofixSetupModal.tsx
```diff
@@ -190,7 +188,7 @@ function AutofixGithubIntegrationStep({
           'Install the [link:Sentry Autofix Github App] on your Github organization or each individual repository with write permissions to enable Autofix.',
```
jennmueng, line 188:
```suggestion
          '[link:Install the Sentry Autofix Github App] on your Github organization or each individual repository with write permissions to enable Autofix.',
```
Given the new link is now a link to the install page,i think we should make the link encompass "install" too

jennmueng (approved):

static/app/components/modals/autofixSetupModal.tsx
malwilley: Good call. Will make this update in a followup PR since I'll be modifying that text to make it optional

PR #71766, opened 2024-05-30
fix(llm-monitoring): Fix alerts, remove hacky ai_total_tokens_used function

colin-sentry:
Make alerts go to the spansMetrics dataset, and remove the old hack where `sum` didn't work on counters. Also add an alias to avoid hardcoding MRIs.

### DIFF
--- a/static/app/views/alerts/rules/metric/utils/getMetricDatasetQueryExtras.tsx
+++ b/static/app/views/alerts/rules/metric/utils/getMetricDatasetQueryExtras.tsx
@@ -26,10 +26,14 @@ export function getMetricDatasetQueryExtras({
   const disableMetricDataset =
     decodeScalar(location?.query?.disableMetricDataset) === 'true';
 
-  const queryExtras: Record<string, string> =
-    hasMetricDataset && !disableMetricDataset
-      ? {dataset: getMEPAlertsDataset(dataset, newAlertOrQuery)}
-      : {};
+  const queryExtras: Record<string, string> = {};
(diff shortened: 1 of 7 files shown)


static/app/views/alerts/rules/metric/utils/getMetricDatasetQueryExtras.tsx
```diff
@@ -26,10 +26,14 @@ export function getMetricDatasetQueryExtras({
   const disableMetricDataset =
     decodeScalar(location?.query?.disableMetricDataset) === 'true';
 
-  const queryExtras: Record<string, string> =
-    hasMetricDataset && !disableMetricDataset
-      ? {dataset: getMEPAlertsDataset(dataset, newAlertOrQuery)}
-      : {};
+  const queryExtras: Record<string, string> = {};
```
wmak, line 29: should split FE and BE prs in two

src/sentry/search/events/datasets/spans_metrics.py
```diff
@@ -31,7 +30,6 @@ def search_filter_converter(
         self,
     ) -> Mapping[str, Callable[[SearchFilter], WhereType | None]]:
         return {
-            "message": self._message_filter_converter,
```
wmak, line 34: was this intentionally removed?

colin-sentry: i'm not sure what's going on - some git bug

gggritso (approved):
LGTM except isn't it supposed to go to indexed spans from the PR description?

PR #71767, opened 2024-05-30
fix: When SCIM is disabled, clear Team.idp_provisioned flags

### LINKED ISSUE #71218: Unable to manage teams created via SCIM after SCIM is disabled.
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

1. Configure SSO and SCIM in an organisation
2. Use the SCIM integration to create teams in Sentry
3. Create manual teams in Sentry
4. Disable SCIM provisioning

### Expected Result

After disabling SCIM provisioning, the teams created with SCIM can be manually edited by a user with correct permissions (add/remove members) users can also join and leave the team (if open membership is enabled).

### Actual Result

Teams created with SCIM will remain only managed via the IdP. 

Because this and SCIM being disabled, users are unable to manage team membership via Sentry. A workaround is to recreate the team manually. 

More information is available in the linked Jira ticket due to PII. 

### Product Area

Settings - Auth

### Link

_No response_

### DSN

_No response_

### Version

_No response_

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2178) by [Unito](https://www.unito.io)

RyanSkonnord:
Fix a bug where, after SCIM is disabled, teams are left in a state where team membership can no longer be managed.

Should resolve https://github.com/getsentry/sentry/issues/71218.

### DIFF
--- a/src/sentry/services/hybrid_cloud/organization/impl.py
+++ b/src/sentry/services/hybrid_cloud/organization/impl.py
@@ -555,6 +555,9 @@ class DatabaseBackedOrganizationService(OrganizationService):
                 .bitand(~OrganizationMember.flags["idp:role-restricted"])
             )
 
+        with unguarded_write(using=router.db_for_write(Team)):
+            Team.objects.filter(organization_id=organization_id).update(idp_provisioned=False)
+
     def update_region_user(self, *, user: RpcRegionUser, region_name: str) -> None:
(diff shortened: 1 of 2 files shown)


markstory (approved):
Looks good to me. The unguarded_write should be ok as `idp_provisioned` is not a replicated field.

PR #71773, opened 2024-05-30
feat(remote-config): Add proxy endpoint for configurations

cmanallen:
Temporarily adds an endpoint to proxy configuration requests from Relay.  This is to help us test the proof of concept.

### DIFF
--- a/src/sentry/remote_config/docs/api.md
+++ b/src/sentry/remote_config/docs/api.md
@@ -118,7 +118,7 @@ Delete the project's configuration.
 
 - Response 204
 
-## Configuration Proxy [/projects/<project_id>/configuration/]
+## Configuration Proxy [/remote-config/projects/<project_id>/]
 
 Temporary configuration proxy resource.
 
--- a/src/sentry/remote_config/docs/api.md
+++ b/src/sentry/remote_config/docs/api.md
@@ -118,7 +118,7 @@ Delete the project's configuration.
 
 - Response 204
 
-## Configuration Proxy [/relays/<project_id>/configuration/]
+## Configuration Proxy [/project/<project_id>/configuration/]
 
 Temporary configuration proxy resource.
 
--- a/src/sentry/remote_config/docs/api.md
+++ b/src/sentry/remote_config/docs/api.md
@@ -118,7 +118,7 @@ Delete the project's configuration.
 
 - Response 204
 
-## Configuration Proxy [/project/<project_id>/configuration/]
+## Configuration Proxy [/proxy/projects/<project_id>/configuration/]
 
 Temporary configuration proxy resource.
 
--- a/src/sentry/remote_config/docs/api.md
+++ b/src/sentry/remote_config/docs/api.md
@@ -118,7 +118,7 @@ Delete the project's configuration.
 
 - Response 204
 
-## Configuration Proxy [/proxy/projects/<project_id>/configuration/]
+## Configuration Proxy [/projects/<project_id>/configuration/]
 
 Temporary configuration proxy resource.
 
--- a/src/sentry/remote_config/docs/api.md
+++ b/src/sentry/remote_config/docs/api.md
@@ -10,7 +10,7 @@ Host: https://sentry.io/api/0
 
 ### Get Configuration [GET]
 
-Retrieve the DSN's configuration.
+Retrieve the project's configuration.
 
 **Attributes**
 
@@ -64,7 +64,7 @@ Retrieve the DSN's configuration.
 
 ### Set Configuration [POST]
 
-Set the DSN's configuration.
+Set the project's configuration.
 
 - Request
 
@@ -114,7 +114,7 @@ Set the DSN's configuration.
 
 ### Delete Configuration [DELETE]
 
-Delete the DSN's configuration.
+Delete the project's configuration.
 
 - Response 204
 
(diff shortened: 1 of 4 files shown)


src/sentry/remote_config/docs/api.md
```diff
@@ -117,3 +117,41 @@ Set the DSN's configuration.
 Delete the DSN's configuration.
 
 - Response 204
+
+## Configuration Proxy [/relays/<project_id>/configuration/]
```
Dav1dde, line 121: As outlined in Slack, this route needs to match the route which is exposed from Relay.

Dav1dde (commented):
To make these changes we should first finish the DACI.

cmanallen:
@Dav1dde We don't need to wait for the DACI.  This is temporary.  We're doing it for a proof of concept test.  This will never make it to customers nor will it be deployed to any SDK running in Sentry.

src/sentry/remote_config/docs/api.md
cmanallen: Can you expand on this?  What should the route look like?  What's being exposed in Relay?

Dav1dde: You exposed the route `/api/:project_id/configuration/` from Relay, Sentry should implement the same interface as Relay. You're adding it here as a Relay specific route.

cmanallen: I thought we didn't need to expose a route in Relay and it would forward automatically?

Dav1dde: I don't know how else to put it, this is not a Relay specific API, so it should not be in the Relay specific API routes.

cmanallen: I understand now.  Its a categorization problem not a functionality problem.  Yes, I will re-scope the route to match a more normal pattern.

src/sentry/remote_config/docs/api.md
```diff
@@ -114,6 +114,44 @@ Set the DSN's configuration.
 
 ### Delete Configuration [DELETE]
 
-Delete the DSN's configuration.
+Delete the project's configuration.
 
 - Response 204
+
+## Configuration Proxy [/proxy/projects/<project_id>/configuration/]
```
cmanallen, line 121: @Dav1dde Is this URL name spaced better?  The SDK will call `https://<ingest_host_url>/api/0/proxy/projects/<project_id>/configuration/`.

Dav1dde: I wouldn't encode any information into the url that this is being proxied. I'd treat the API as if it was directly called from an SDK (which it is, just through a layer of proxies).

cmanallen: I'm having trouble name spacing it in a more traditional way because of the lack of org slug.  I could just drop the proxy prefix and make it the singular `project` but it didn't "feel" right.  That said, I can totally do that.  Its just a feeling after all :P

Dav1dde: My initial instinct (not much thought) would have been something like `/api/:project_id/remote_config/`

cmanallen: Got it.  I can rename configuration to remote_config.  The prefix to this url is already `host/api/0` regardless what i define so i think `project/<id>/remote_config/` will have to be the final state of it.

cmanallen: I don't know that it "needs to be" I guess.  But I know I haven't defined any urls in Relay matching what crons has done.  They seem be doing some processing Relay.  I'm using the generic forward route.  But if we're exposing an API on sentry we should follow its convention right?

cmanallen: https://github.com/getsentry/relay/blob/23fa1f3f358ff9966753dd005ebe94919fa654e7/relay-server/src/endpoints/mod.rs#L108-L109

Dav1dde: The fallback only cares about that routes need to start with `/api/`.

Dav1dde: I don't know about Sentry routes, but for me this seems inconsistent, setting the path which is usually referred to as the project_id to 0 just to later still scope it down to a project.

cmanallen: Sorry I think I don't have enough context for this.  Do you want to meet up today/tomorrow and teach me more about how Relay handles routes?

The `/0/` is the api version not a project id.

Dav1dde: > The /0/ is the api version not a project id.

Not for all of [these routes](https://github.com/getsentry/relay/blob/23fa1f3f358ff9966753dd005ebe94919fa654e7/relay-server/src/endpoints/mod.rs#L78-L97). But I am by no means an expert on how Sentry defines its API, I just go off of what I see in Relay ¯\_(ツ)_/¯.

> Sorry I think I don't have enough context for this. Do you want to meet up today/tomorrow and teach me more about how Relay handles routes?

It's relatively simple, it has some handlers defined for a fixed set of routes, everything that doesn't match a specific handler and starts with `/api/` is forwarded to the upstream, everything else is 404'd.

cmanallen: Right, I get that but its going to forward to the `sentry.io` host which expects `/api/<version>/`.  So I have to structure my URL in such a way that Relay will forward it with the correct version.  If I do `/api/<project-id>` Sentry will not know how to handle that route.

It works locally with the current route structure so I know its not a functionality problem.  Its more a taste thing how we structure the route.

aliu39 (approved):

PR #71777, opened 2024-05-30
fix(similarity): fix call logic for seer

JoshFerge:
right now we are running rate limits for _all_ projects regardless if they have the flag enabled or not. this is not great -- see [here](https://console.cloud.google.com/logs/query;query=labels.name%3D~%22sentry.events.grouping%22;cursorTimestamp=2024-05-29T12:04:38.763032760Z;aroundTime=2024-05-29T12:00:10.334Z;duration=PT1H?project=internal-sentry)

we should check feature flags first before looking at rate limits.


mrduncan (approved):
We should get a test on this as a followup

JoshFerge:
> We should get a test on this as a followup

there are tests, just not testing the cascading nature of the checks here

PR #71778, opened 2024-05-30
add queue monitoring option

ykamo001:
We want to monitor the time a task stays on a queue, waiting to get executed. This helps us debug tasks that are more time-sensitive. An option is added to control the default behavior.

### DIFF
--- a/src/sentry/tasks/base.py
+++ b/src/sentry/tasks/base.py
@@ -10,7 +10,6 @@ from typing import Any, TypeVar
 from celery import current_task
 from django.db.models import Model
 
-from sentry import options
 from sentry.celery import app
 from sentry.silo.base import SiloLimit, SiloMode
 from sentry.utils import metrics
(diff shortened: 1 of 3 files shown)


src/sentry/tasks/base.py
```diff
@@ -91,6 +92,10 @@ def instrumented_task(name, stat_suffix=None, silo_mode=None, record_timing=Fals
     - disabling of result collection.
     """
 
+    # Use option to control default behavior of queue time monitoring
+    # Value can be dynamically updated, which is why the evaluation happens during function run-time
+    record_timing = record_timing or options.get("sentry-metrics.monitor-queue-time")
```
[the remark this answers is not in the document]
wedamija: I think this actually needs to be inside of `_wrapped` to work correctly. Otherwise, this will be assessed just when the `instrumented_task` decorator is first called, which is on module load.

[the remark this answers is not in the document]
ykamo001: good catch

wedamija (approved):

src/sentry/tasks/base.py
```diff
@@ -94,6 +94,14 @@ def instrumented_task(name, stat_suffix=None, silo_mode=None, record_timing=Fals
     def wrapped(func):
         @wraps(func)
         def _wrapped(*args, **kwargs):
+            from sentry import options
+
+            # Use option to control default behavior of queue time monitoring
+            # Value can be dynamically updated, which is why the evaluation happens during function run-time
+            record_queue_wait_time = record_timing or options.get(
+                "sentry-metrics.monitor-queue-time"
+            )
```
ykamo001, line 103: Hmm, should I wrap this around a try/catch to contain the blast radius? If any exception doesn't get caught when thrown from the stack all the celery tasks could get effected

wedamija: Yeah might be worth doing so just to be extra cautious

ykamo001: done

ykamo001:
> ## Suspect Issues
> 
> * ‼️ **InvalidSearchQuery: Invalid value '['FRONTEND-Q7N']' for 'issue:' filter** `sentry.snuba.tasks.create_subscription_in_snuba` [View Issue](https://sentry.sentry.io/issues/5430434567/?referrer=github-pr-bot)
> 

This is not related to this PR

PR #71780, opened 2024-05-30
feat(related_issues): Add analytics

armenzg:
We want to know if a related issue was clicked after rendering it.


static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -187,9 +189,29 @@ describe('TraceTimeline', () => {
       await screen.findByText('SELECT "sentry_monitorcheckin"."monitor_id"')
     ).toBeInTheDocument();
     expect(screen.queryByLabelText('Current Event')).not.toBeInTheDocument();
+
+    // Test analytics
     expect(useRouteAnalyticsParams).toHaveBeenCalledWith({
       trace_timeline_status: 'empty',
     });
+    // XXX: The component is being rendered twice, thus, two calls to trackAnalytics
+    expect(trackAnalytics).toHaveBeenCalledTimes(2);
```
armenzg, line 198: The parent component is not rendering twice - only the TraceIssueEvent one. I'm still investigating.

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -21,18 +22,27 @@ export function TraceIssueEvent({event}: TraceIssueEventProps) {
   const {title, subtitle} = getTitleSubtitle(event);
   const avatarSize = parseInt(space(4), 10);
 
-  // If any of data fails to load, we don't want to render the component
-  // Only "One other issue appears in the same trace. View Full Trace (X issues)" would show up
+  const referrer = 'issue_details.related_trace_issue';
+
+  trackAnalytics(`${referrer}.viewed`, {
```
malwilley, line 27: This is a side effect, so it should be inside a `useEffect` callback. But it could be better to add this to the `issue_details.viewed` event anyway

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -21,18 +23,24 @@ export function TraceIssueEvent({event}: TraceIssueEventProps) {
   const {title, subtitle} = getTitleSubtitle(event);
   const avatarSize = parseInt(space(4), 10);
 
-  // If any of data fails to load, we don't want to render the component
-  // Only "One other issue appears in the same trace. View Full Trace (X issues)" would show up
+  const referrer = 'issue_details.related_trace_issue';
+
+  useRouteAnalyticsParams({has_related_trace_issue: true});
```
malwilley, line 28: Could you add `has_related_trace_issue: false` to the top-level call? That will ensure that amplitude records `false` instead of `(none)` when this component isn't rendered 

https://github.com/getsentry/sentry/blob/fb10e1f4cfabf57725ab1ae97ebf3d0a03034064/static/app/views/issueDetails/groupDetails.tsx#L589

malwilley (approved):

PR #71783, opened 2024-05-30
feat(hybridcloud) Add method to fetch sentryapp installs with caching

markstory:
We frequently see spikes of traffic going to this RPC call. These should benefit from caching as orgs occasionally run hot with errors and caching would help reduce the call volume.

I've included cache expiration for when a sentryapp changes, and when an install changes as waiting 1 hour for updates is too long. Once this lands, I'll start using the new cached method behind an option to gradually warm the cache and observe changes in throughput.


src/sentry/receivers/outbox/control.py
```diff
@@ -37,13 +41,33 @@ def process_integration_updates(object_identifier: int, region_name: str, **kwds
...
+
+    # When a sentry app's definition changes purge cache for all the installations.
+    # This could get slow for large applications, but generally big applications don't change often.
+    install_query = SentryAppInstallation.objects.filter(sentry_app=sentry_app).values(
+        "id", "organization_id"
+    )
+    install_map: dict[int, int] = {}
+    for row in install_query:
+        install_map[row["organization_id"]] = row["id"]
+
```
GabeVillalobos, line 60: I don't see an index in the model preventing us from having multiple sentry app installations for a single organization/sentry app pair 😬 Do we need to handle this possible case?

src/sentry/services/hybrid_cloud/app/service.py
```diff
@@ -82,6 +83,14 @@ def get_sentry_app_by_slug(self, *, slug: str) -> RpcSentryApp | None:
     def get_installation_by_id(self, *, id: int) -> RpcSentryAppInstallation | None:
         pass
 
+    def installation_by_id(self, *, id: int) -> RpcSentryAppInstallation | None:
+        """
+        Get a sentryapp install by id
+
+        This method is cached wrapper around get_installation_by_id()
```
GabeVillalobos, line 90:
```suggestion
        This method is a cached wrapper around get_installation_by_id()
```

GabeVillalobos: Would it be worth adding `cached` in the method name to make this behavior bit more obvious?

GabeVillalobos (approved):
Changes look good, just had a couple of questions.

src/sentry/receivers/outbox/control.py
markstory: Good catch, I thought there was a constraint, but there isn't. And even worse, there *are* duplicates in prod.

src/sentry/services/hybrid_cloud/app/service.py
markstory: I'm not sure. We don't have `cached` in the other cached methods.

GabeVillalobos (approved):

PR #71787, opened 2024-05-30
feat(self-hosted): Errors only self-hosted sidebar

hubertdeng123:
Screenshot 2024-05-30 at 12 32 26 PM


azaslavsky:
The front and back end changes should be separate PRs?

hubertdeng123:
@azaslavsky There are only frontend changes in the PR right now

static/app/components/sidebar/index.tsx
```diff
@@ -653,7 +654,7 @@ function Sidebar() {
                   {projects}
                 </SidebarSection>
 
-                {hasNewSidebarHierarchy && (
+                {hasNewSidebarHierarchy && !isSelfHostedErrorsOnly && (
```
azaslavsky, line 657: Can we add a test, at least a jest test but maybe a snapshot too, that this renders the way we want to?

azaslavsky (approved):

static/app/components/sidebar/index.tsx
hubertdeng123: I like the idea of having testing for the sidebar here, but I'm not sure exactly how the new sidebar should look like and how often that is going to change in the upcoming few months so this seems out of scope for the PR. As for snapshot testing, I don't think that is generally used in the frontend of sentry, and it was removed at some point. https://github.com/getsentry/sentry/pull/56037

azaslavsky: > I like the idea of having testing for the sidebar here, but I'm not sure exactly how the new sidebar should look like and how often that is going to change in the upcoming few months so this seems out of scope for the PR.

That's exactly why I think it's important: no one on our team knows everything about sidebar design, and no one on the sidebar-owning team knows everything (or probably, at this stage, anything) about errors only self-hosted. This allows us both to prevent regressions and communicate our intentions clearly.

Otherwise, I think a very possible outcome is that the sidebar folks will change the design in the upcoming few months as you say, end up overlooking the errors only case, and break our next release (which we won't know about either since we don't have a regression test). Whereas if we do have the test, it will break when they redesign the sidebar, in which case they'll either: a. replace it with something that still reasonably handles the errors only case, or b. ask us how to proceed to accomplish that. Either one of those is a much better outcome.

Imo, we should just codify exactly how it looks now in the jest test. Future editors who now more about the sidebar will break the test, then update it as necessary. I imagine there is already such a test for the existing sidebar, so we can just add an "errors only" variant?

hubertdeng123: > Imo, we should just codify exactly how it looks now in the jest test. Future editors who now more about the sidebar will break the test, then update it as necessary. I imagine there is already such a test for the existing sidebar, so we can just add an "errors only" variant?

There actually isn't. For something that is so important in our UI, it appears to be woefully undertested. I've added a test that will break if self-hosted Sentry renders with a sidebar with different items. It asserts on the id of each sidebar item.

hubertdeng123: @azaslavsky  Added some basic tests for how the sidebar should look with some basic feature flags for both the old hierarchy and new hierarchy. Ideally, there should be more thorough testing of feature flags for the sidebar, but that seems out of scope for now.

azaslavsky: Thanks for adding these!

azaslavsky (approved):

hubertdeng123:
wow I think someone merged in a change for the sidebar right as a updated this branch to master, breaking tests

https://github.com/getsentry/sentry/pull/71821/files#diff-bc2278bb86eb9f14ffeeefeb2727758abf56c7086e66fde575be0bf0cb2ea6b1

PR #71791, opened 2024-05-30
ref(api): Refactor Org Serializers to Selectively Include feature flags

iamrajjoshi:
Followup on https://github.com/getsentry/sentry/pull/71668

I refactored the serializers here so they accept a new kwarg - `include_feature_flag`. if the kwarg is passed, it will only determine the `features` field if the flag is `True`.

This should reduce the amount of computation that we do if the response doesn't require feature flags.

### DIFF
--- a/src/sentry/api/serializers/models/organization.py
+++ b/src/sentry/api/serializers/models/organization.py
@@ -239,22 +239,11 @@ class OrganizationSerializer(Serializer):
             for o in item_list
         }
 
-    def serialize(
+    def get_feature_set(
         self, obj: Organization, attrs: Mapping[str, Any], user: User, **kwargs: Any
-    ) -> OrganizationSerializerResponse:
+    ) -> set[str]:
         from sentry import features
         from sentry.features.base import OrganizationFeature
-
-        if attrs.get("avatar"):
-            avatar = {
-                "avatarType": attrs["avatar"].get_avatar_type_display(),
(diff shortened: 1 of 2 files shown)


src/sentry/api/serializers/models/organization.py
```diff
@@ -446,17 +457,20 @@ def get_attrs(
...
     def serialize(  # type: ignore[explicit-override, override]
-        self, obj: Organization, attrs: Mapping[str, Any], user: User, access: Access
+        self, obj: Organization, attrs: Mapping[str, Any], user: User, access: Access, **kwargs: Any
     ) -> DetailedOrganizationSerializerResponse:
         # TODO: rectify access argument overriding parent if we want to remove above type ignore
 
         from sentry import experiments
 
         experiment_assignments = experiments.all(org=obj, actor=user)
+        include_feature_flags = kwargs.get("include_feature_flags", True)
```
iamrajjoshi, line 467: we should include the flag if the kwarg isn't passed in.

ykamo001 (commented):
Love the work here @iamrajjoshi! I think we can clean this up a little and make it more modular and easier to extend for the future, going to put some time on the calendar

ykamo001 (approved):

ykamo001:
> Love the work here @iamrajjoshi! I think we can clean this up a little and make it more modular and easier to extend for the future, going to put some time on the calendar

We can do this in a follow up, so no rush!

PR #71794, opened 2024-05-30
test(ui): Use built in router context in trace component tests

scttcper:
Use the built in router context in the render wrapper instead of a custom setup


static/app/views/performance/newTraceDetails/trace.spec.tsx
```diff
@@ -299,7 +285,7 @@ async function pageloadTestSetup() {
   mockTraceEventDetails();
   mockMetricsResponse();
 
-  const value = render(<TraceViewWithProviders traceSlug="trace-id" />);
+  const value = render(<TraceView />);
```
scttcper, line 288: whoops

michellewzhang (approved):

PR #71795, opened 2024-05-30
feat(feedback): Make the url on the Feedback Details page be clickable

### LINKED ISSUE #71664: [User Feedback] Feedback Url should be clickable
Currently it's a textbox with the auto-copy functionality enabled.

It'd be nice if it was just a link. But it could be spoofed, so if it's a link it should also have a warning popup that people are navigating off-site, like what issues has (title: "Heads up" body: "You're leaving Sentry and will be redirected to the following external website:")

ryan953:
**Before:**
before

**After:**
after

Fixes https://github.com/getsentry/sentry/issues/71664

### DIFF
--- a/static/app/components/feedback/feedbackItem/feedbackItem.tsx
+++ b/static/app/components/feedback/feedbackItem/feedbackItem.tsx
@@ -1,6 +1,7 @@
 import {Fragment, useEffect, useRef} from 'react';
 import styled from '@emotion/styled';
 
+import {openNavigateToExternalLinkModal} from 'sentry/actionCreators/modal';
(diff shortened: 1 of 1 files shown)


aliu3ntry:
Can you still copy the link w/a button?

michellewzhang (approved):

ryan953:
> Can you still copy the link w/a button?

@aliu3ntry there's no button anymore, but people can still click+drag to select and copy it. I think the utility is about the same either way.

aliu3ntry:
> Edit
> Hide
> Delete

Got it, it would be nice to have both, if not nw

PR #71796, opened 2024-05-30
feat(replay-new-trace): Adding replay timestamp indicators.

Abdkhan14:
https://github.com/getsentry/sentry/assets/60121741/d1476a9d-04f2-43cd-8a50-225731e52bad

### DIFF
--- a/static/app/views/performance/newTraceDetails/trace.tsx
+++ b/static/app/views/performance/newTraceDetails/trace.tsx
@@ -445,6 +445,9 @@ export function Trace({
     render: render,
   });
 
+  const traceNode = trace.root.children[0];
+  const traceStartTimestamp = traceNode?.space?.[0];
+
   return (
     <TraceStylingWrapper
       ref={manager.registerContainerRef}
@@ -502,7 +505,12 @@ export function Trace({
             </div>
           );
         })}
-        <VerticalTimestampIndicators viewmanager={manager} tree={treeRef.current} />
+        {traceNode && traceStartTimestamp ? (
+          <VerticalTimestampIndicators
+            viewmanager={manager}
+            traceStartTimestamp={traceStartTimestamp}
+          />
+        ) : null}
       </div>
       <div
         ref={setScrollContainer}
@@ -1632,14 +1640,11 @@ function AutogroupedTraceBar(props: AutogroupedTraceBarProps) {
 
 function VerticalTimestampIndicators({
   viewmanager,
-  tree,
+  traceStartTimestamp,
 }: {
-  tree: TraceTree;
+  traceStartTimestamp: number;
   viewmanager: VirtualizedViewManager;
 }) {
-  const traceNode = tree.root.children[0];
-  const traceStartTimestamp = traceNode.space?.[0];
-
   useEffect(() => {
     function replayTimestampListener({
       currentTime,
@@ -1648,10 +1653,6 @@ function VerticalTimestampIndicators({
       currentHoverTime: number | undefined;
       currentTime: number;
     }) {
-      if (!traceStartTimestamp || !traceNode) {
-        return;
-      }
-
       if (viewmanager.vertical_indicators['replay_timestamp.current']) {
         viewmanager.vertical_indicators['replay_timestamp.current'].timestamp =
           traceStartTimestamp + currentTime;
@@ -1676,7 +1677,7 @@ function VerticalTimestampIndicators({
         replayTimestampListener
       );
     };
-  }, [traceStartTimestamp, traceNode, viewmanager]);
+  }, [traceStartTimestamp, viewmanager]);
 
   const registerReplayCurrentTimestampRef = useCallback(
     (ref: HTMLDivElement | null) => {
@@ -1697,10 +1698,6 @@ function VerticalTimestampIndicators({
     },
     [viewmanager]
   );
-
-  if (!traceNode || !traceStartTimestamp) {
-    return null;
-  }
 
   return (
     <Fragment>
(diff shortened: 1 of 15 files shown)


static/app/views/replays/detail/trace/trace.tsx
```diff
@@ -215,4 +215,4 @@ const TraceViewWaterfallWrapper = styled('div')`
   height: 100%;
 `;
 
-export default Trace;
+export default React.memo(Trace);
```
JonasBa, line 218: This is not/should not be needed

static/app/views/replays/detail/trace/index.tsx
```diff
@@ -18,10 +20,43 @@ function PerfDisabled() {
   );
 }
 
+type EventHandler = (...args: any[]) => void;
+
+class EventEmitter {
```
JonasBa, line 25:
```suggestion
class ReplayEventEmitter {
```

static/app/views/replays/detail/trace/index.tsx
```diff
@@ -18,10 +20,43 @@ function PerfDisabled() {
...
+  private events: {[key: string]: EventHandler[]} = {};
+
+  on(event: string, handler: EventHandler): void {
```
JonasBa, line 28:
```suggestion
  on(event:'replay timestamp change', handler: EventHandler): void {
```

JonasBa: I would also prefer if we had stronger types for this. You can copy the dispatching reducer event emitter declarations or the profiling flamegraph event emitter to see how to do that

static/app/views/replays/detail/trace/index.tsx
```diff
@@ -18,10 +20,43 @@ function PerfDisabled() {
...
+    this.events[event] = this.events[event] || [];
+    this.events[event].push(handler);
```
JonasBa, line 30: With a design like this, it is possible to push the same emitter multiple times and it will result in multiple invocations. Is that expected?

static/app/views/replays/detail/trace/index.tsx
```diff
@@ -18,10 +20,43 @@ function PerfDisabled() {
...
+  }
+
+  emit(event: string, ...args: any[]): void {
+    const handlers = this.events[event] || [];
+    handlers.forEach(handler => handler(...args));
+  }
+
+  removeListener(event: string, handler: EventHandler): void {
```
JonasBa, line 38: In contrast to the on method, this would usually be named or alised to `off`

static/app/views/performance/newTraceDetails/traceRenderers/virtualizedViewManager.tsx
```diff
@@ -127,6 +127,20 @@ export class VirtualizedViewManager {
...
+      timestamp: number | undefined;
+    };
+    hover: {
+      ref: HTMLElement | null;
+      timestamp: number | undefined;
+    };
+  } = {
+    current: {ref: null, timestamp: undefined},
+    hover: {ref: null, timestamp: undefined},
+  };
```
JonasBa, line 142:
```suggestion
  type VerticalIndicator = {
    timestamp: number | undefined,
    ref: HTMLElement | null
  }
  replay_indicators: {
    current: VerticalIndicator,
    hover: VerticalIndicator
    } = {
    current: {ref: null, timestamp: undefined},
    hover: {ref: null, timestamp: undefined},
  };
```

static/app/views/performance/newTraceDetails/trace.tsx
```diff
@@ -1620,6 +1630,63 @@ function AutogroupedTraceBar(props: AutogroupedTraceBarProps) {
   );
 }
 
+function ReplayTimeStampIndicators({
```
JonasBa, line 1633:
```suggestion
function TimelineIndicators({
```

static/app/views/performance/newTraceDetails/traceRenderers/virtualizedViewManager.tsx
JonasBa: A more generic API leaves you more flexibility to add different indicators in the future

static/app/views/performance/newTraceDetails/traceRenderers/virtualizedViewManager.tsx
```diff
@@ -1473,6 +1499,26 @@ export class VirtualizedViewManager {
     }
   }
 
+  drawReplayTimestamps() {
+    const {current, hover} = this.replay_indicators;
+    this.drawReplayTimestamp(current.ref, current.timestamp);
+    this.drawReplayTimestamp(hover.ref, hover.timestamp);
+  }
+
+  drawReplayTimestamp(ref: HTMLElement | null, timestamp: number | undefined) {
```
JonasBa, line 1534: All of this should just be drawIndicators where it does something like

```
for indicator of this.indicators
  draw(indicator)
```

static/app/views/replays/detail/trace/index.tsx
ryan953: There's also https://github.com/getsentry/sentry/blob/35bd66eba1c13177353f457cb442a175515aaed5/static/app/utils/eventDispatcher.tsx#L29

static/app/views/replays/detail/trace/trace.tsx
Abdkhan14: Removed it. Found a way to pass down the `replay`, without having to use `useReplayContext` in parent component.

JonasBa (approved):

PR #71687, opened 2024-05-29
feat(feedback): Iterate on feedback list item highlight style

ryan953:
Here's the after, with something selected, unselected, and hovered:
SCR-20240529-kaal

Fixes https://github.com/getsentry/sentry/issues/71640

PR #71797, opened 2024-05-30
fix(feedback): Fix padding on Feedback List items

ryan953:
Fixes alignment of the header after https://github.com/getsentry/sentry/pull/71687. With this PR the width of the list-items got bigger, because we removed extra padding/whitespace around each item.

Notice that in the 'before' screenshot the "all items" checkbox is not vertically aligned above the checkbox inside the list-item. Now the header is wider, so the checkboxes align vertically.

**Before:**
before

**After:**
after

### DIFF
--- a/static/app/components/feedback/list/feedbackListHeader.tsx
+++ b/static/app/components/feedback/list/feedbackListHeader.tsx
@@ -96,7 +96,7 @@ const HeaderPanel = styled('div')`
 `;
 
 const HeaderPanelItem = styled('div')`
-  padding: ${space(1)} ${space(2)} ${space(1)} ${space(2)};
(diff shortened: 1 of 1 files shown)


aliu39 (approved):

static/app/components/feedback/list/feedbackListHeader.tsx
```diff
@@ -96,7 +96,7 @@ const HeaderPanel = styled('div')`
 `;
 
 const HeaderPanelItem = styled('div')`
-  padding: ${space(1)} ${space(2)} ${space(1)} ${space(2)};
+  padding: ${space(1)} ${space(1.5)} ${space(1)} ${space(1.5)};
```
ryan953, line 99:
```suggestion
  padding: ${space(1)} ${space(1.5)};
```

PR #71817, opened 2024-05-31
ref(pageFilters): Return stable reference from getState

scttcper:
- removes unused hasInitialState
- mashes everything into a readonly state object that is updated on state change


static/app/stores/pageFiltersStore.spec.tsx
```diff
@@ -36,6 +36,12 @@ describe('PageFiltersStore', function () {
     });
   });
 
+  it('returns a stable reference with getState', () => {
+    PageFiltersStore.updateProjects([1], []);
+    const state = PageFiltersStore.getState();
+    expect(Object.is(state, PageFiltersStore.getState())).toBe(true);
```
ryan953, line 42: This assertion could be `.toBe()` which uses `Object.is` internally, and will give better messages if the assertion fails

https://jestjs.io/docs/expect#tobevalue

static/app/stores/pageFiltersStore.tsx
```diff
@@ -117,45 +101,46 @@ interface PageFiltersStoreDefinition
 }
 
 const storeConfig: PageFiltersStoreDefinition = {
-  selection: getDefaultSelection(),
-  pinnedFilters: new Set(),
-  desyncedFilters: new Set(),
-  shouldPersist: true,
-  hasInitialState: false,
+  state: {
```
ryan953, line 104: could be using the `private` or `protected` keyword so callers know to continue to use `getState()`

```suggestion
  private state: {
```

ryan953: internally it seems like there's a mix of calling `this.state` and `this.getState()`. but nbd.

ryan953 (approved):

static/app/stores/pageFiltersStore.tsx
scttcper: ideally, but i think those only exist in classes

scttcper: it is readonly now at least https://github.com/getsentry/sentry/blob/master/static/app/stores/types.tsx#L25

static/app/stores/pageFiltersStore.spec.tsx
scttcper: yeah i'm trying to be explicit on what is being tested. fear people would see it fail and change it to toEqual

PR #71821, opened 2024-05-31
feat(insights): Granular frontend feature flags

gggritso:
The `spans-first-ui` flag is going away, in favour of three flags:

- `insights-entry-points` which enables entry points to Insight UIs if those entry points have upsells (e.g., sidebar links)
- `insights-initial-modules` enables access to UIs that connect to the first set of Insights modules (Queries, Requests, Resources, etc.)
- `insights-addon-modules` enables access to UI that show the newer set of Insights modules (Caches, LLMs, etc.)

This PR gets rid of `spans-first-ui` checks on the frontend, replacing them with the right `insights-` flag.


DominikB2014 (approved):
👍

PR #71822, opened 2024-05-31
ref(metrics): Add metrics for dev script execution times

elramen:
Add metrics for execution times of the scripts in lib.sh.

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -36,24 +36,14 @@ require() {
 
 configure-sentry-cli() {
     if [ -f "${venv_name}/bin/sentry-cli" ]; then
-        path="${venv_name}/bin/sentry-cli"
-    elif command -v "sentry-cli" >/dev/null 2>&1; then
-        path="sentry-cli"
+        return 0
     elif [ -f "${venv_name}/bin/pip" ]; then
+        echo 'installing sentry-cli'
         pip-install sentry-cli
-        path="${venv_name}/bin/sentry-cli"
     else
-        curl -sL https://sentry.io/get-cli/ | SENTRY_CLI_VERSION="2.32.0" bash
-        path="sentry-cli"
-    fi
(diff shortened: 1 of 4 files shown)


scripts/lib.sh
```diff
@@ -34,6 +34,32 @@ require() {
...
+configure-sentry-cli() {
+    if [ -f "${venv_name}/bin/sentry-cli" ]; then
+        path="${venv_name}/bin/sentry-cli"
+    elif command -v "sentry-cli" >/dev/null 2>&1; then
+        path="sentry-cli"
+    elif [ -f "${venv_name}/bin/pip" ]; then
+        pip-install sentry-cli
+        path="${venv_name}/bin/sentry-cli"
+    else
+        curl -sL https://sentry.io/get-cli/ | SENTRY_CLI_VERSION="2.32.0" bash
```
joshuarli, line 46: i don't think we should do this, it would be better to just simplify and prompt for `devenv sync` if `sentry-cli` isn't in the venv

joshuarli: can also just remove all the following code for version checking, i think it's reasonable that devenv sync is run semi-regularly 

otherwise we're effectively invoking sentry-cli two times in do.sh

scripts/lib.sh
```diff
@@ -34,6 +34,22 @@ require() {
...
+configure-sentry-cli() {
+    if [ -f "${venv_name}/bin/sentry-cli" ]; then
+        return 0
+    elif [ -f "${venv_name}/bin/pip" ]; then
+        echo 'installing sentry-cli'
+        pip-install sentry-cli
+    else
+        cat <<EOF
+${red}${bold}
+ERROR: sentry-cli could not be installed. Please run "devenv sync" in sentry and getsentry.
```
joshuarli, line 46: Just say "please run devenv sync".

joshuarli (approved):

PR #71823, opened 2024-05-31
feat(insights): Granular backend feature flags

gggritso:
Sibling to https://github.com/getsentry/sentry/pull/71821 but on the backend. Basically, disable the Aggregate Span Waterfall and Span Duration Comparison features if the organization doesn't have access to Insights modules (AKA span metrics!)


wmak (approved):
lgtm

PR #71826, opened 2024-05-31
feat(hybridcloud) Add cached implementation of user.get_many_by_id

markstory:
We make a ton of RPC calls for lists of users. Many of those calls are simple ID lists. We can cache these quite effectively as the queries are simple and cache keys can be expired via outboxes.

The rollout of using cache is controlled by an option, and I've added metrics to observe the effective hit rate of cached/rpc fetches.

Caching decorators were added in #71719

### DIFF
--- a/tests/sentry/models/test_user.py
+++ b/tests/sentry/models/test_user.py
@@ -1,3 +1,6 @@
+from unittest.mock import patch
+
+import sentry.hybridcloud.rpc.services.caching as caching_module
 from sentry.models.authenticator import Authenticator
 from sentry.models.organizationmember import OrganizationMember
 from sentry.models.savedsearch import SavedSearch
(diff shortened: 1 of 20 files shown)


GabeVillalobos (approved):

tests/sentry/models/test_user.py
```diff
@@ -122,6 +125,23 @@ def test_cascades_to_regions_even_if_user_ownership_revoked(self):
...
+            user.username = "bob2"
+            user.save()
+            mock_caching_service.clear_key.assert_any_call(
+                key=f"user_service.get_many_by_id:{user.id}",
+                region_name=_TEST_REGIONS[0].name,
+            )
+            mock_caching_service.clear_key.assert_any_call(
+                key=f"user_service.get_user:{user.id}",
+                region_name=_TEST_REGIONS[0].name,
+            )
```
markstory, line 143: Is this along the lines of what you were looking for @GabeVillalobos (from a [previous comment](https://github.com/getsentry/sentry/pull/71719/files#r1624797976)) ?

PR #71828, opened 2024-05-31
feat(features) Allow defining default value for features during registration

markstory:
Having to remember to update server.py when adding a new feature is no fun. It contributes bloat to server.py and adds friction to adding/removing features. Consolidating the default value with feature registration will improve ergonomics of defining features.

### DIFF
--- a/src/sentry/features/manager.py
+++ b/src/sentry/features/manager.py
@@ -150,7 +150,7 @@ class FeatureManager(RegisteredFeatureManager):
         name: str,
         cls: type[Feature] = Feature,
         entity_feature_strategy: bool | FeatureHandlerStrategy = False,
-        default: bool | None = None,
+        default: bool = False,
     ) -> None:
         """
         Register a feature.
@@ -181,7 +181,7 @@ class FeatureManager(RegisteredFeatureManager):
             feature_option_name = f"{FLAGPOLE_OPTION_PREFIX}.{name}"
             register(feature_option_name, type=Dict, default={}, flags=FLAG_AUTOMATOR_MODIFIABLE)
 
-        if default is not None:
+        if default is not None and name not in settings.SENTRY_FEATURES:
             settings.SENTRY_FEATURES[name] = default
 
         self._feature_registry[name] = cls
--- a/src/sentry/features/manager.py
+++ b/src/sentry/features/manager.py
@@ -181,7 +181,7 @@ class FeatureManager(RegisteredFeatureManager):
             feature_option_name = f"{FLAGPOLE_OPTION_PREFIX}.{name}"
             register(feature_option_name, type=Dict, default={}, flags=FLAG_AUTOMATOR_MODIFIABLE)
 
-        if default is not None and name not in settings.SENTRY_FEATURES:
+        if name not in settings.SENTRY_FEATURES:
             settings.SENTRY_FEATURES[name] = default
 
         self._feature_registry[name] = cls
--- a/src/sentry/features/manager.py
+++ b/src/sentry/features/manager.py
@@ -150,6 +150,7 @@ class FeatureManager(RegisteredFeatureManager):
         name: str,
         cls: type[Feature] = Feature,
         entity_feature_strategy: bool | FeatureHandlerStrategy = False,
+        default: bool | None = None,
     ) -> None:
         """
         Register a feature.
@@ -158,6 +159,8 @@ class FeatureManager(RegisteredFeatureManager):
         to encapsulate the context associated with a feature.
 
         >>> FeatureManager.has('my:feature', actor=request.user)
+
+        Features that use flagpole will have an option automatically registered.
         """
         entity_feature_strategy = self._shim_feature_strategy(entity_feature_strategy)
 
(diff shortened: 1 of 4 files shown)


src/sentry/features/temporary.py
```diff
@@ -29,15 +29,22 @@ def register_temporary_features(manager: FeatureManager):
...
-    # Organization scoped features that are in development or in customer trials.
-    manager.add("organizations:activated-alert-rules", OrganizationFeature, FeatureHandlerStrategy.REMOTE)
-    manager.add("organizations:ai-analytics", OrganizationFeature, FeatureHandlerStrategy.REMOTE)
+    # Organization scoped features that are in development or in customer trials. #
+    ###############################################################################
+
+    # Enables activated alert rules
+    manager.add("organizations:activated-alert-rules", OrganizationFeature, FeatureHandlerStrategy.REMOTE, default=False)
+    # Enable AI analytics pages (sentry for AI teams)
+    manager.add("organizations:ai-analytics", OrganizationFeature, FeatureHandlerStrategy.REMOTE, default=False)
```
markstory, line 47: Example of what our feature registration would look like if we fold the contents of server.py into the feature registration lists. If folks are happy with this approach I can apply this to all features we currently have.

src/sentry/features/manager.py
```diff
@@ -178,6 +181,9 @@ def add(
             feature_option_name = f"{FLAGPOLE_OPTION_PREFIX}.{name}"
             register(feature_option_name, type=Dict, default={}, flags=FLAG_AUTOMATOR_MODIFIABLE)
 
+        if default is not None:
+            settings.SENTRY_FEATURES[name] = default
```
evanpurkhiser, line 185: How did it get this value before?

wedamija: It was set directly in `SENTRY_FEATURES`

wedamija, line 185: I think this should be `if default is not None and name not in settings.SENTRY_FEATURES`. This would allow customers to still override feature flag settings via settings, which I think they do today?

src/sentry/features/manager.py
```diff
@@ -150,6 +150,7 @@ def add(
         name: str,
         cls: type[Feature] = Feature,
         entity_feature_strategy: bool | FeatureHandlerStrategy = False,
+        default: bool | None = None,
```
wedamija, line 153: This should probably be required?

src/sentry/features/manager.py
markstory: We would have to manually update `server.py` with the default. If no default was provided the default value would be `False` based on logic in `FeatureManager` https://github.com/getsentry/sentry/blob/88a8cbc633096d26da539157ff3edadc9c47e792/src/sentry/features/manager.py#L261

evanpurkhiser: oh duh, I read this backwards lol

src/sentry/features/manager.py
markstory: Agreed, I was playing it safe with this change. Combined with your other suggestion I think making this required and having the default = False is a good solution.

src/sentry/features/manager.py
```diff
@@ -178,6 +181,9 @@ def add(
             feature_option_name = f"{FLAGPOLE_OPTION_PREFIX}.{name}"
             register(feature_option_name, type=Dict, default={}, flags=FLAG_AUTOMATOR_MODIFIABLE)
 
+        if default is not None and name not in settings.SENTRY_FEATURES:
```
wedamija, line 184: Nit: We can now remove `default is not None`

wedamija (commented):
Looks good to me, thanks for making this easier to work with

GabeVillalobos (approved):

wedamija (approved):

src/sentry/features/manager.py
evanpurkhiser: One nice thing about having this in the server file is that it was somewhat documenting for on-prem users. But I'm sure not many people actually looked there.

src/sentry/features/temporary.py
```diff
@@ -29,14 +29,21 @@ def register_temporary_features(manager: FeatureManager):
...
+
+    # Enables user registration.
+    manager.add("auth:register", SystemFeature, FeatureHandlerStrategy.INTERNAL, default=True)
     manager.add("organizations:create", SystemFeature, FeatureHandlerStrategy.INTERNAL)
     manager.add("organizations:multi-region-selector", SystemFeature, FeatureHandlerStrategy.INTERNAL)
     manager.add("relocation:enabled", SystemFeature, FeatureHandlerStrategy.INTERNAL)
 
-    # Organization scoped features that are in development or in customer trials.
+    # Organization scoped features that are in development or in customer trials. #
+    ###############################################################################
```
evanpurkhiser, line 42: Should we add the comment flag for the project scoped ones too

src/sentry/features/temporary.py
```diff
@@ -29,14 +29,21 @@ def register_temporary_features(manager: FeatureManager):
...
+
+    # Enables activated alert rules
```
evanpurkhiser, line 44: Were you thinking of porting over all the comments.

At this point maybe features themselves should just have a `description` hah

evanpurkhiser: Oops sorry missed your comment here https://github.com/getsentry/sentry/pull/71828#discussion_r1622684629

markstory: Making the comment into a feature description isn't a bad idea 🤔

PR #71829, opened 2024-05-31
fix(metrics): Ensure MQB uses transaction for free text search

Zylphrex:
The MQB is exclusively used to search transaction metrics so map `message` to the `transaction` column. But we should be wary of subclasses that are used for other metrics.


wmak (approved):
lgtm

PR #71832, opened 2024-05-31
ref: Move lastOfArray so it can be found and reused more often

ryan953:
Related to "Audit/move functions out of global utils.tsx" https://github.com/getsentry/frontend-tsc/issues/13

### DIFF
--- a/static/app/utils/array/lastOfArray.tsx
+++ b/static/app/utils/array/lastOfArray.tsx
@@ -1,0 +1,5 @@
+export function lastOfArray<T extends Array<unknown> | ReadonlyArray<unknown>>(
+  t: T
+): T[number] {
+  return t[t.length - 1];
+}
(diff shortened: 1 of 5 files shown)


evanpurkhiser (approved):

static/app/utils/array/lastOfArray.tsx
```diff
@@ -0,0 +1,5 @@
+export function lastOfArray<T extends Array<unknown> | ReadonlyArray<unknown>>(
+  t: T
+): T[number] {
+  return t[t.length - 1];
```
ryan953, line 4: this could probably be `t.at(-1)` and not even be a helper function, just do that inline. In both cases, the return value is correct with the current code, or with that change.

The benefit if any is that this is something people can grep for and import, instead of knowing about `.at()` ... or instead of typing the variable name twice.

PR #71833, opened 2024-05-31
fix(query-builder): Fix selected values and click behavior for filter values

malwilley:
- When clicking anywhere in the filter value area while editing, forwards that click event to the input instead
- Fixes the items calculation, before it wasn't filtering out selected values correctly in all cases

### DIFF
--- a/static/app/components/searchQueryBuilder/combobox.tsx
+++ b/static/app/components/searchQueryBuilder/combobox.tsx
@@ -1,4 +1,11 @@
-import {type Key, type MouseEventHandler, useCallback, useMemo, useRef} from 'react';
+import {
+  forwardRef,
+  type Key,
+  type MouseEventHandler,
+  useCallback,
+  useMemo,
+  useRef,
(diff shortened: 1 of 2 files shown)


static/app/components/searchQueryBuilder/combobox.tsx
```diff
@@ -37,158 +44,163 @@ type SearchQueryBuilderComboboxProps = {
...
-  const state = useComboBoxState<SelectOptionWithKey<string>>({
-    children,
-    items,
-    autoFocus,
-    inputValue: filterValue,
-    onSelectionChange,
-    disabledKeys,
-  });
-  const {inputProps, listBoxProps} = useComboBox<SelectOptionWithKey<string>>(
+export const SearchQueryBuilderCombobox = forwardRef(
```
malwilley, line 47: Simply adding `forwardRef` here, no other changes

scttcper (approved):

PR #71834, opened 2024-05-31
ref: Remove trim from utils.tsx

ryan953:
https://github.com/getsentry/frontend-tsc/issues/13


michellewzhang (approved):
missed a spot in `utils.tsx` on line 244

PR #71840, opened 2024-05-31
ref(record-backfill): Query snuba per group

jangjodi:
Change snuba query in backfill script to query events one group at a time, and limit to one
This is necessary since groups with many events were causing the query to timeout (see [here](https://sentry-st.sentry.io/issues/5430094720/?project=1513938))

### DIFF
--- a/src/sentry/tasks/backfill_seer_grouping_records.py
+++ b/src/sentry/tasks/backfill_seer_grouping_records.py
@@ -155,13 +155,13 @@ def backfill_seer_grouping_records(
         )
 
     group_id_batch = list(group_id_message_batch_filtered.keys())
-    time_now = datetime.now()
     events_entity = Entity("events", alias="events")
     redis_client = redis.redis_clusters.get(settings.SENTRY_MONITORS_REDIS_CLUSTER)
 
-    # TODO(jangjodi): Only query per group if it has over 1 million events
-    snuba_event_data, group_id_batch_all = [], copy.deepcopy(group_id_batch)
-    for group_id in group_id_batch_all:
+    # TODO(jangjodi): Only query per group if it has over 1 million events, or batch queries with new where condition
+    snuba_requests = []
+    for group_id in group_id_batch:
+        group = Group.objects.get(id=group_id)
         query = Query(
             match=events_entity,
             select=[
@@ -172,9 +172,15 @@ def backfill_seer_grouping_records(
                 Condition(Column("project_id"), Op.EQ, project.id),
                 Condition(Column("group_id"), Op.EQ, group_id),
                 Condition(
-                    Column("timestamp", entity=events_entity), Op.GTE, time_now - timedelta(days=90)
+                    Column("timestamp", entity=events_entity),
+                    Op.GTE,
+                    group.last_seen - timedelta(days=10),
                 ),
-                Condition(Column("timestamp", entity=events_entity), Op.LT, time_now),
+                Condition(
+                    Column("timestamp", entity=events_entity),
+                    Op.LT,
+                    group.last_seen + timedelta(minutes=5),
+                ),
             ],
             limit=Limit(1),
         )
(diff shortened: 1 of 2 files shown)


src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -159,64 +159,63 @@ def backfill_seer_grouping_records(
...
+                Column("group_id"),
+                Column("event_id"),
+            ],
+            where=[
+                Condition(Column("project_id"), Op.EQ, project.id),
+                Condition(Column("group_id"), Op.EQ, group_id),
+                Condition(
+                    Column("timestamp", entity=events_entity), Op.GTE, time_now - timedelta(days=90)
+                ),
+                Condition(Column("timestamp", entity=events_entity), Op.LT, time_now),
```
wedamija, line 177: You can make this query much more efficient by doing `group.last_seen - 5 mins <= timestamp < group.last_seen + 5 mins`

Potentially, you could also make the query a lot faster for batched groups by doing something like
```
(group_id = 123 and timestamp < <group_123_last_seen> + 5 mins and timestamp >= <group_123_last_seen> - 5 mins)
or (group_id = 456 and timestamp < <group_456_last_seen> + 5 mins and timestamp >= <group_456_last_seen> - 5 mins)
or ...

```

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -159,64 +159,63 @@ def backfill_seer_grouping_records(
...
-                            "organization_id": project.organization.id,
-                            "project_id": project_id,
-                            "group_id": group_id,
-                        },
-                    )
-                    group_id_batch.remove(group_id)
-                    del group_id_message_batch_filtered[group_id]
+        with metrics.timer(f"{BACKFILL_NAME}.bulk_snuba_queries", sample_rate=1.0):
+            result = bulk_snuba_queries(
+                [request], referrer=Referrer.GROUPING_RECORDS_BACKFILL_REFERRER.value
```
wedamija, line 194: You could potentially batch several of these together here and save yourself some time

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -159,64 +159,63 @@ def backfill_seer_grouping_records(
...
-    with metrics.timer(f"{BACKFILL_NAME}.bulk_snuba_queries", sample_rate=1.0):
-        result = bulk_snuba_queries(
-            [request], referrer=Referrer.GROUPING_RECORDS_BACKFILL_REFERRER.value
+        request = Request(
+            dataset=Dataset.Events.value,
+            app_id=Referrer.GROUPING_RECORDS_BACKFILL_REFERRER.value,
+            query=query,
+            tenant_ids={
+                "referrer": Referrer.GROUPING_RECORDS_BACKFILL_REFERRER.value,
+                "cross_org_query": 1,
```
wedamija, line 188: I don't think this is really a cross org query right? Since we're just filtering to a single project and group

src/sentry/tasks/backfill_seer_grouping_records.py
jangjodi: Thank you! I've modified the query to limit it to the 10 minutes around `group.last_seen` and I'll follow up with another PR for trying to do it for batched groups

src/sentry/tasks/backfill_seer_grouping_records.py
jangjodi: It's a "cross_org_query" because we don't want it counting towards an org's regular snuba rate limit and affect anything. This way, just this referrer will count towards its own rate limit

wedamija: Oh, I didn't know that's what that means, good to know!

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -155,45 +155,55 @@ def backfill_seer_grouping_records(
...
-            Condition(
-                Column("timestamp", entity=events_entity), Op.GTE, time_now - timedelta(days=90)
-            ),
-            Condition(Column("timestamp", entity=events_entity), Op.LT, time_now),
-        ],
-        orderby=[OrderBy(Column("group_id"), Direction.ASC)],
-    )
+    # TODO(jangjodi): Only query per group if it has over 1 million events, or batch queries with new where condition
+    snuba_requests = []
+    for group_id in group_id_batch:
```
wedamija, line 163: How big is a batch btw? We might want to be careful that we're not slamming snuba with too many queries at once, or we could hit a rate limit.

wedamija (approved):

src/sentry/tasks/backfill_seer_grouping_records.py
jangjodi: the batch size is 20 and I've taken your suggestion to combine the call into a bulk query (with up to 20 requests)

wedamija: Right, that batch thing just makes 20 queries in a thread pool, so it's still a bunch of queries at once. The thread pool has 10 workers https://github.com/getsentry/sentry/blob/cebec75409806ecb190b1a6db68091b53d42d0c7/src/sentry/utils/snuba.py#L469

So as long as that doesn't put us over the limits for this referrer we should be ok

jangjodi: I see, makes sense. Based on this [comment](https://github.com/getsentry/snuba/pull/5774#issuecomment-2057160235) we have 4 concurrent queries and two threads, but I'll ask about non-current but very close together queries

wedamija: The batch query running will run them concurrently using the thread pool. I'd probably recommend breaking into batches of 4 instead of 20

jangjodi: I've decided to change the batch size to 4 for now, and then follow up to do the queries in batches of 4. Unfortunately I can't do the `(group.id = 1 and group.last_seen - 5 mins <= timestamp < group.last_seen + 5 mins) or ...` since the query always needs an an absolute time range provided

wedamija: That's a good point... Actually, I guess rather than the weird or statement, it'd be better to just do:

`group_id in (<group_ids>) and min(groups, key=lambda g: g.last_seen) - 5 mins <= timestamp and timestamp <= max(groups, key=lambda g: g.last_seen) + 5 mins`

PR #71846, opened 2024-05-31
add extra metadata

ykamo001:
For `apply_async` we weren't passing in `start_time` so the tasks weren't getting the queue time

### DIFF
--- a/src/sentry/celery.py
+++ b/src/sentry/celery.py
@@ -87,10 +87,12 @@ class SentryTask(Task):
     Request = "sentry.celery:SentryRequest"
 
     @classmethod
-    def _add_metadata(cls, kwargs: dict[str, Any]) -> None:
+    def _add_metadata(cls, kwargs: dict[str, Any] | None) -> None:
         """
         Helper method that adds relevant metadata
         """
+        if kwargs is None:
+            return None
         # Add the start time when the task was kicked off for async processing by the calling code
         kwargs["__start_time"] = datetime.now().timestamp()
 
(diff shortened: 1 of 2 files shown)


src/sentry/testutils/helpers/task_runner.py
```diff
@@ -45,9 +45,18 @@ def _apply_async(
         kwargs: dict[str, Any] | None = None,
         countdown: float | None = None,
         queue: str | None = None,
+        **options,
```
ykamo001, line 48: Had to add this to appease testing framework

src/sentry/celery.py
```diff
@@ -85,11 +86,20 @@ def celery_prefork_freeze_gc(**kwargs: object) -> None:
 class SentryTask(Task):
     Request = "sentry.celery:SentryRequest"
 
-    def delay(self, *args, **kwargs):
+    @classmethod
+    def _add_metadata(cls, kwargs: dict[str, Any]) -> None:
+        """
```
ykamo001, line 91: hmm, should I do a `None` check here and also update the type hinting?

markstory: Kwargs should always be a dict based on how this method is being called. A none check wouldn't hurt though.

markstory (approved):

src/sentry/celery.py
ykamo001: done, can't be too safe

wedamija (approved):

PR #71853, opened 2024-05-31
fix(query-builder): Update value and trigger onSearch correctly

malwilley:
- Updates the query state when `initialQuery` value changes (which enables the query to change when switching tabs)
- Stops input blurring from triggering `onSearch`, now only pressing enter will do that


static/app/components/searchQueryBuilder/combobox.tsx
```diff
@@ -24,7 +24,8 @@ type SearchQueryBuilderComboboxProps = {
   inputLabel: string;
   inputValue: string;
   items: SelectOptionWithKey<string>[];
-  onCustomValueSelected: (value: string) => void;
+  onCustomValueBlurred: (value: string) => void;
+  onCustomValueCommitted: (value: string) => void;
```
malwilley, line 28: Instead of a single `onCustomValueSelected`, I broke it into two callbacks. One which fires when the value was blurred, and the other when it was specifically committed with enter.

armenzg (approved):

PR #71854, opened 2024-05-31
ref: Move semverCompare into utils/versions/

ryan953:
Related to https://github.com/getsentry/sentry/pull/71857


ryan953:
I'm also thinking about if this should be in `utils/releases/`.
I think versions is more specific to the datatypes involved, while releases feels more about the product we have. So I feel like it's going into a more stable spot, instead of a flexible product specific spot.

I'll just put it into versions/ to get it somewhere... and with #71857 it won't be alone in there.

michellewzhang (approved):
will we follow up with cleaning up the references in `getsentry` as well?

PR #71856, opened 2024-05-31
fix(related_issues): Fix on click events

armenzg:
We were passing the `org.slug` instead of the `organization`, thus, failing to report the event.

It's unclear why it was letting me do the wrong thing.


static/app/utils/analytics/issueAnalyticsEvents.tsx
```diff
@@ -109,6 +110,7 @@ export type IssueEventParameters = {
   'issue_details.performance.hidden_spans_expanded': {};
   'issue_details.related_trace_issue.trace_issue_clicked': {
     group_id: number;
+    organization: Organization;
```
scttcper, line 113: this shouldn't be necessary and should be required on all analytics, i can look into whats happening

scttcper (approved):
looks like we overloaded org to allow string in some cases. does seem like this can be confusing

https://github.com/getsentry/sentry/blob/master/static/app/utils/analytics/makeAnalyticsFunction.tsx#L8

roggenkemper (approved):

PR #71857, opened 2024-05-31
ref: Move isSemverRelease & formatVersion into utils/versions/

ryan953:
Related to https://github.com/getsentry/frontend-tsc/issues/13


roggenkemper (approved):

ryan953:
I'm also thinking about if this should be in `utils/releases/`.
I think 'versions/' is more specific to the datatypes involved, while releases feels more about the product we have. So I feel like it's going into a more stable spot, instead of a flexible product specific spot.

ryan953:
related to https://github.com/getsentry/sentry/pull/71854

PR #71862, opened 2024-05-31
chore(issues): Add KeyValueData story and small UI fixes

leeandher:
Mainly adding the story for the KeyValueData component  (`stories/?name=app/components/keyValueData/card.stories.tsx`) but also discovered a few small UI fixes:

- Context error icon now appears inline
- Icon lines up with the one from tags
- Tag value will be used even if `null` so long as its set (rather than defaulting to `--`)

image
image


malwilley (approved):
Looks great!

PR #71868, opened 2024-05-31
ref(rules): Make an enum for trigger action methods

ceorourke:
During the TTL week investigations I found this area that would benefit from an enum for "fire" and "resolve".

### DIFF
--- a/src/sentry/incidents/subscription_processor.py
+++ b/src/sentry/incidents/subscription_processor.py
@@ -27,6 +27,7 @@ from sentry.incidents.models.alert_rule import (
     AlertRuleMonitorType,
     AlertRuleThresholdType,
     AlertRuleTrigger,
+    AlertRuleTriggerActionMethod,
     invoke_alert_subscription_callback,
 )
 from sentry.incidents.models.alert_rule_activations import AlertRuleActivations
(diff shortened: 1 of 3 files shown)


src/sentry/incidents/subscription_processor.py
```diff
@@ -733,7 +734,11 @@ def handle_trigger_actions(
         # Grab the first trigger to get incident id (they are all the same)
         # All triggers should either be firing or resolving, so doesn't matter which we grab.
         incident_trigger = incident_triggers[0]
-        method = "fire" if incident_trigger.status == TriggerStatus.ACTIVE.value else "resolve"
+        method = (
+            AlertRuleTriggerActionMethod.FIRE.value
+            if incident_trigger.status == TriggerStatus.ACTIVE.value
+            else AlertRuleTriggerActionMethod.RESOLVE.value
+        )
```
ceorourke, line 741: I get pickle errors if I try to do the enum rather than the value

src/sentry/incidents/tasks.py
```diff
@@ -182,7 +182,7 @@ def handle_trigger_action(
     action_id: int,
     incident_id: int,
     project_id: int,
-    method: str,
+    method: Literal["fire", "resolve"],
```
ceorourke, line 185: I couldn't find a better way to do this, my goal was to make it obvious what the values are but since I can't pass the enum through an async task for pickle reasons this was the best I could come up with.

saponifi3d: Idk if it helps, but i was able to get an enum working the way you're trying to like this: https://github.com/getsentry/sentry/blob/master/src/sentry/incidents/models/alert_rule.py#L50

saponifi3d (approved):
nit; it'd be nice if we didn't use `.value` but saw your comments so don't want to block on that :)

PR #71872, opened 2024-05-31
ref: remove safe_execute in send_confirmation_notification

asottile-sentry:
errors are already handled gracefully in this method and it is creating an unnecessary transaction

I am trying to remove the transaction side-effect of safe_execute since almost all of the callers do not use it

I checked a week of logs and this has never produces a safe execute log line for this function


ceorourke (approved):

saponifi3d (approved):
:shipit:

PR #71873, opened 2024-05-31
ref(custom-views): Delete member's views upon member removal

### LINKED ISSUE #71802: Custom Views: Augment member removal endpoint to delete user's views
Since we have changed the database schema of a view to use `organization_id` and `user_id` instead of `org_member_id`, removing a member of an organization will not cascade delete their views, so we will need to do it manually. Eventually we may need to update this logic to only delete non-organization level views

MichaelSun48:
This PR implements logic to delete an organization member's custom views upon removal from the organization. The table used to store custom views, `sentry_groupsearchview`, does not use `org_member_id` as a foreign key, so deletion of the member will **not** cascade delete their views, hence why we must implement manual logic to do so.

### DIFF
--- a/src/sentry/api/endpoints/organization_member/details.py
+++ b/src/sentry/api/endpoints/organization_member/details.py
@@ -29,6 +29,7 @@ from sentry.apidocs.constants import (
 from sentry.apidocs.examples.organization_examples import OrganizationExamples
 from sentry.apidocs.parameters import GlobalParams
 from sentry.auth.superuser import is_active_superuser
+from sentry.models.groupsearchview import GroupSearchView
 from sentry.models.organization import Organization
 from sentry.models.organizationmember import InviteStatus, OrganizationMember
 from sentry.models.organizationmemberteam import OrganizationMemberTeam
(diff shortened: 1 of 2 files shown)


wedamija (commented):
I think it would be better to handle this via the `deletions/` module, so that it an org member is deleted via some other method this is still removed.

Here's an example: https://github.com/getsentry/sentry/blob/26fa50bb22ea1d491c56e4501dd1a6a6f3a7d126/src/sentry/deletions/defaults/alert_rule_trigger_action.py#L4

MichaelSun48:
> I think it would be better to handle this via the deletions/ module, so that it an org member is deleted via some other method this is still removed.

Ah I assumed that this endpoint was the only entry point for member removal, but if that's not the case then totally agree that the deletions module is better. I'll take down this PR and create a new one that does that, thanks!

Just out of curiosity - how else can a member be removed from an organization besides this endpoint? Or is this just for edge cases like when the entire organization is removed?

wedamija:
> > I think it would be better to handle this via the deletions/ module, so that it an org member is deleted via some other method this is still removed.
> 
> Ah I assumed that this endpoint was the only entry point for member removal, but if that's not the case then totally agree that the deletions module is better. I'll take down this PR and create a new one that does that, thanks!
> 
> Just out of curiosity - how else can a member be removed from an organization besides this endpoint? Or is this just for edge cases like when the entire organization is removed?

It's hard to say - I could imagine a lot of edge cases where it might happen, or that someone might add some other way to delete org members in the future. Generally I prefer to avoid putting logic like this in the endpoint, just because I don't know if it will end up missed

PR #71875, opened 2024-05-31
ref(replays): update videoReplayer code to match new rrweb logic & bump rrweb version

michellewzhang:
see https://github.com/getsentry/rrweb/pull/190

we don't need this snippet anymore because we'll create pointers lazily.


static/app/components/replays/videoReplayerWithInteractions.tsx
```diff
@@ -118,7 +118,9 @@ export class VideoReplayerWithInteractions {
     if (!hasGestures) {
       // If the replay has no gestures, we should hide the mouse
       // @ts-expect-error private
-      this.replayer.mouse.classList.remove('replayer-mouse');
+      for (const [, {pointerEl}] of Object.entries(this.replayer.pointers)) {
+        pointerEl.classList.remove('replayer-mouse');
+      }
```
billyvg, line 123: is this still needed with the new version? don't we only create pointers if it exists in event data?

michellewzhang: oh you're right, this should be removed

billyvg (approved):
can we also update the rrweb library and make sure everything works as expected?

yarn.lock
```diff
@@ -3105,6 +3124,20 @@
     fflate "^0.4.4"
     mitt "^3.0.0"
 
+"@sentry-internal/rrweb@2.17.0":
```
scttcper, line 3127: looking like duplicate packages here

you may need to run `npx yarn-deduplicate && yarn install`

michellewzhang: just ran - does it look correct now @scttcper ?

scttcper: still looks like duplciates, it might be because you haven't updated the other two packages

```    
"@sentry-internal/rrweb-player": "2.12.0",
"@sentry-internal/rrweb-snapshot": "2.12.0",
```

michellewzhang: just updated

scttcper (approved):

PR #71876, opened 2024-05-31
ref: don't hold a django transaction while doing a bunch of redis operations

asottile-sentry:
`_with_transaction` defaults to true and opens an atomic django transaction (for no database operations!)

I'm trying to kill this surprising behaviour -- but first setting each callsite to `_with_transaction=False` before doing so


iker-barriocanal (approved):
Good catch!

PR #71881, opened 2024-05-31
feat(slack): add slack-sdk

cathteng:
Added slack-sdk to pypi here https://github.com/getsentry/pypi/pull/839

See https://www.notion.so/sentry/Task-list-for-Messaging-Vertical-ca4f431d43b84e72b8a91643fe880b66
- We are adding the `slack-sdk` package to be able to use the `WebClient` in place of our custom `SlackClient` implementation

Following https://develop.sentry.dev/python-dependencies/

### DIFF
--- a/requirements-base.txt
+++ b/requirements-base.txt
@@ -67,6 +67,7 @@ sentry-ophio==0.2.7
 sentry-redis-tools>=0.1.7
 sentry-relay>=0.8.67
 sentry-sdk>=2.2.1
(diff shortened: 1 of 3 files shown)


ykamo001 (approved):
One small step for ecosystem, one giant step for all integrations

asottile-sentry (commented):
would prefer this is added when it's actually used -- otherwise this looks like an unused dependency

PR #71882, opened 2024-05-31
feat(profileHours): Add profile hours to org stats

scttcper:
More data category confusion, in this case we were using only part of the data category to match it to the plural which had some overlap between profile and profile_duration.

TODO: put profile hours behind a feature flag


mbauer404 (approved):

scttcper:
@getsentry/profiling Is there an existing feature flag that makes sense for this?

PR #71886, opened 2024-05-31
nit(slack): updated github name in slack threads

iamrajjoshi:
We didn't capitalize github correctly in our slack threads, it should be `GitHub`

### DIFF
--- a/src/sentry/integrations/slack/threads/activity_notifications.py
+++ b/src/sentry/integrations/slack/threads/activity_notifications.py
@@ -128,6 +128,12 @@ class _ExternalIssueCreatedActivity:
 
         return ticket_number
 
+    def get_formatted_provider_name(self, provider: str) -> str:
+        # Make sure to make the proper noun have correct capitalization
+        # I.e. github -> GitHub, jira -> Jira
+        # Special cases like github -> GitHub are implemented in their overriden classes
+        return provider.capitalize()
+
 
 class _AsanaExternalIssueCreatedActivity(_ExternalIssueCreatedActivity):
     """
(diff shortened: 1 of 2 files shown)


ykamo001 (approved):

src/sentry/integrations/slack/threads/activity_notifications.py
```diff
@@ -178,6 +178,13 @@ class ExternalIssueCreatedActivityNotification(GroupActivityNotification):
     metrics_key = "create_issue"
     title = "External Issue Created"
 
+    def format_provider_name(provider: str) -> str:
+        # Make sure to make the proper noun have correct capitalization
+        # Take in account special cases
+        # I.e. github -> GitHub, jira -> Jira
+        special_cases = {"github": "GitHub"}
+        return special_cases.get(provider.lower(), provider.capitalize())
```
ykamo001, line 186: instead of doign this, can we just update `_external_issue_activity_factory` and update it to return a special Github class?

src/sentry/integrations/slack/threads/activity_notifications.py
```diff
@@ -128,6 +128,12 @@ def get_ticket_number(self) -> str:
 
         return ticket_number
 
+    def get_formatted_provider_name(self, provider: str) -> str:
+        # Make sure to make the proper noun have correct capitalization
+        # I.e. github -> GitHub, jira -> Jira
+        # Special cases like github -> GitHub are implemented in their overriden classes
+        return provider.capitalize()
```
ykamo001, line 135: should not accept a parameter, especially `provider` since this is part of the existing class. We can just do `self.get_provider().capitalize()`

src/sentry/integrations/slack/threads/activity_notifications.py
```diff
@@ -161,17 +167,30 @@ def get_ticket_number(self) -> str:
         return last_part
 
 
+class _GithubExternalIssueCreatedActivity(_ExternalIssueCreatedActivity):
+    """
+    Override class for Github, as the provider name that we want to display should be GitHub, not "Github"
+    """
+
+    def get_formatted_provider_name(self, _: str) -> str:
```
ykamo001, line 175: once base class is updated, you can update this

src/sentry/integrations/slack/threads/activity_notifications.py
```diff
@@ -161,17 +167,30 @@ def get_ticket_number(self) -> str:
...
+
 def _external_issue_activity_factory(activity: Activity) -> _ExternalIssueCreatedActivity:
     """
     Returns the correct ExternalIssueCreatedActivity class based on the provider.
     All classes have the same interface, the method for one is simply modified for its use case.
     """
+    activity_classes = {
+        "asana": _AsanaExternalIssueCreatedActivity,
+        "github": _GithubExternalIssueCreatedActivity,
+    }
```
ykamo001, line 187: feel free to move this out into a private global variable so it only initializes once on module load and saves some cpu/memory

ykamo001 (changes requested):

src/sentry/integrations/slack/threads/activity_notifications.py
```diff
@@ -187,9 +208,7 @@ def get_description(self) -> tuple[str, str | None, Mapping[str, Any]]:
             base_template = "an "
         else:
             base_template = "a "
-            # Make sure to make the proper noun have correct capitalization
-            # I.e. github -> Github, jira -> Jira
-            provider = provider.capitalize()
+            provider = external_issue.get_formatted_provider_name()
```
ykamo001, line 211: you could clean up this code if you wanted too, but not a blocker
```
provider = external_issue.get_formated_ptovider()
if external_issue.get_provider() == external_issue.DEFAULT:
   ...
```
and this way you don't need to override the provider value

ykamo001 (approved):

PR #71904, opened 2024-06-03
feat(insights): rename Processing Time column to Span Duration

KevinL10:
Rename Processing Time column in queues module to Span Duration.


mjq (approved):
Makes sense. 👍

PR #71910, opened 2024-06-03
feat(sidebar): expand nested items if parent has no link

DominikB2014:
1. If the sidebar is in the desktop mode and expanded, clicking on the parent item of a sidebar accordion that does not have a link, toggles expansion of the accordion.
image

2. If the sidebar is in mobile mobile or its in desktop mode and collapsed, we disable the hover on the parent item. (Can't click insights in this screenshot)
image


gggritso (approved):
Nice!

PR #71913, opened 2024-06-03
ref(crons): Ignore in progress check-in coming in after OK, ERROR

davidenwang:
Don't generate a processing error for the case where an in-progress comes after an OK check-in. This is to remove noise for fast-running jobs (order of milliseconds) where it is common to see something like:

1. OK/ERROR
2. IN-PROGRESS

### DIFF
--- a/tests/sentry/monitors/consumers/test_monitor_consumer.py
+++ b/tests/sentry/monitors/consumers/test_monitor_consumer.py
@@ -36,6 +36,10 @@ from sentry.utils import json
 from sentry.utils.outcomes import Outcome
 
 
+class ExpectNoProcessingError:
+    pass
+
+
 class MonitorConsumerTest(TestCase):
     def setUp(self):
         super().setUp()
@@ -69,7 +73,7 @@ class MonitorConsumerTest(TestCase):
         guid: str | None = None,
         ts: datetime | None = None,
         consumer: ProcessingStrategy | None = None,
-        expected_error: ProcessingErrorsException | None = None,
+        expected_error: ProcessingErrorsException | ExpectNoProcessingError | None = None,
         expected_monitor_slug: str | None = None,
         **overrides: Any,
     ) -> None:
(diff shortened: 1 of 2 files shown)


tests/sentry/monitors/consumers/test_monitor_consumer.py
```diff
@@ -396,6 +396,27 @@ def test_check_in_update_terminal(self):
...
+            monitor.slug,
+            guid=self.guid,
+            status="ok",
+            expected_error=ProcessingErrorsException(
+                [{"type": ProcessingErrorType.CHECKIN_FINISHED}], monitor
+            ),
+        )
+
+        checkin = MonitorCheckIn.objects.get(guid=self.guid)
+        assert checkin.duration == int(20.0 * 1000)
```
evanpurkhiser, line 418: Does this actually validate that we didn't create a processing error?

davidenwang: nope, unfortunately we don't have validation for when we don't create a processing error. Right now I've just opted to test that we actually do create an error for the case where it should

evanpurkhiser: Can we have it check that no errors were created if `expected_error` is some sentinel value?

davidenwang: Yea right now expected_error is default None, but that just ignores checking for any sort of expected_error. I'm planning on changing the None to validate that no processing_errors were created, but that also involves adding an actual expected_error to many of the tests in this file so I was planning on doing in a follow-up.

Though actually, maybe I can work around this for now 1 sec

tests/sentry/monitors/consumers/test_monitor_consumer.py
```diff
@@ -129,6 +129,10 @@ def check_processing_errors(
             yield
 
             args_list = handle_processing_errors.call_args_list
+            if expected_error.processing_errors == []:
+                assert len(args_list) == 0
+                return
```
evanpurkhiser, line 134: I don't love this, I'm thinking maybe you could define like

```py
class ExpectNoProcessingError:
  pass
```

and then have 

```py
excpected_error: ProcessingErrorsException | ExpectNoProcessingError | None,
```

and pass like

```py
expected_error=ExpectNoProcessingError(),
```

davidenwang: ahhh i see what you're getting at now, tyty just changed

evanpurkhiser (approved):

PR #71614, opened 2024-05-28
feat(insights): ignore duplicate filter fields in search bar

KevinL10:
Removes the filter field as a searchable span field in the sample panel search bars.

PR #71916, opened 2024-06-03
feat(insights): disable filters when overriden by search query

KevinL10:
Disables span sample filters on queues/caches modules when the search query overrides the filter key. This is an alternative to removing the filter key from the search bar entirely, and replaces https://github.com/getsentry/sentry/pull/71614.

image
image

### DIFF
--- a/static/app/views/performance/cache/samplePanel/samplePanel.tsx
+++ b/static/app/views/performance/cache/samplePanel/samplePanel.tsx
@@ -190,9 +190,8 @@ export function CacheSamplePanel() {
   let cacheMissSamplesLimit = SPAN_SAMPLE_LIMIT / 2;
 
   if (isCacheHitFilterOverridden) {
-    // if the search query overrides the `cache.hit` filter, ignore the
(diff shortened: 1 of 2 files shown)


DominikB2014:
> This is an alternative to removing the filter key from the search bar entirely, and replaces #71614.
@KevinL10 Just to clarify, you have two open PRs, but only one of them should be merged?

DominikB2014:
This is largely a design decision,
but my opinion is that we shouldn't allow `cache.hit` in the search bar at all, this matches the existing behaviour of the issues page (there is a dropdown for environment, but you can't search for environment. Same thing for project). To make this less confusing, I also think we should move the dropdown closer to the search bar, it seems odd that they aren't together, but control the same search.

KevinL10:
I discussed with Vasudha and we had decided to gray out the status bar, but if the existing behaviour is to ignore the field from the search bar then I think we should go with that. Re: dropdown, I think that was mostly a design decision to make the layout smoother.

I'll merge https://github.com/getsentry/sentry/pull/71614 and close this pr.

PR #71921, opened 2024-06-03
fix(insights): Use module URL helper in special widgets

gggritso:
This must be the last place where we manually constructed URLs that linked to `/performance/`. Instead, use the `useModuleURL` helper, which will direct people to `/insights/` vs. `/performance/` when necessary.


static/app/views/performance/landing/widgets/widgets/performanceScoreWidget.tsx
```diff
@@ -44,17 +45,16 @@ export function PerformanceScoreWidget(props: PerformanceWidgetProps) {
       }
     : undefined;
 
+  const moduleURL = useModuleURL('vital');
```
DominikB2014, line 48: Ignore, this would be way to much to change rn lol, but I feel like this module should be called `web-vitals` in code.

DominikB2014 (commented):
👍 Nice, I think i missed this during my refactor!

DominikB2014 (approved):

static/app/views/performance/landing/widgets/widgets/performanceScoreWidget.tsx
gggritso: Oh yeah, definitely. Also some of these use underscores vs. dashes and so on 😭

PR #71922, opened 2024-06-03
chore(features) Remove redundant feature defaults

markstory:
Now that features.add() sets a default value, we can remove the now redundant feature default definitions and combine the comments with the feature declarations.

I've only modified features with a default of False that were in `temporary.py` other features will be updated in a separate pull request.

### DIFF
--- a/src/sentry/conf/server.py
+++ b/src/sentry/conf/server.py
@@ -1452,42 +1452,19 @@ SENTRY_EARLY_FEATURES = {
 SENTRY_FEATURES: dict[str, bool | None] = {
     # Enable advanced search features, like negation and wildcard matching.
     "organizations:advanced-search": True,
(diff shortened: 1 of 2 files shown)


evanpurkhiser (approved):

dashed (approved):

vartec:
Wouldn't it make sense to have explicit `default=False` on these, rather than rely on implicit? 
Doesn't really change any functionality, but for documentation sake might make sense.

vartec (approved):

markstory:
> Wouldn't it make sense to have explicit default=False on these, rather than rely on implicit?
Doesn't really change any functionality, but for documentation sake might make sense.

I can make that change, I had received feedback before from @wedamija to skip including the default when it is false.

PR #71926, opened 2024-06-03
fix: Improve "Add Member" dropdown alignment on Team Members settings page

ryan953:
"Add Member" Dropdown in the table header, right side, alignment is fixed:

**Before:**
SCR-20240603-kkce

**After:**
SCR-20240603-kkcz

### DIFF
--- a/static/app/views/settings/organizationTeams/teamMembers.tsx
+++ b/static/app/views/settings/organizationTeams/teamMembers.tsx
@@ -22,6 +22,7 @@ import LoadingIndicator from 'sentry/components/loadingIndicator';
 import Pagination from 'sentry/components/pagination';
 import Panel from 'sentry/components/panels/panel';
 import PanelHeader from 'sentry/components/panels/panelHeader';
+import {Flex} from 'sentry/components/profiling/flex';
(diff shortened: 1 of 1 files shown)


static/app/views/settings/organizationTeams/teamMembers.tsx
```diff
@@ -22,6 +22,7 @@ import LoadingIndicator from 'sentry/components/loadingIndicator';
 import Pagination from 'sentry/components/pagination';
 import Panel from 'sentry/components/panels/panel';
 import PanelHeader from 'sentry/components/panels/panelHeader';
+import {Flex} from 'sentry/components/profiling/flex';
```
evanpurkhiser, line 25: not sure how I feel about importing something from `profiling` into here

michellewzhang: we should move `Flex` out of the profiling folder, we use it everywhere in feedback too!

ryan953: I'll move it in another PR. A few files will get touched in profiling, replay and feedback.

michellewzhang (approved):

PR #71928, opened 2024-06-03
fix: Only format less files, no eslint

ryan953:
A bug was discovered in the CI config with PR https://github.com/getsentry/sentry/pull/71674

Basically we shouldn't be trying to lint a `*.less` file with eslint, we instead need to pass less files into prettier (new config for this) and stylelint (added to existing config for this)

The matching change is applied to getsentry here: https://github.com/getsentry/getsentry/pull/14167


gggritso (approved):
👍🏻

PR #71929, opened 2024-06-03
feat(insights): show is new badge insights/explore

DominikB2014:
Show new badge for insights and explore menus, these badges get removed as soon as the accordion gets expanded.
image

If the parent item does not have a link, we won't show the new badge when the accordion is floating (otherwise the new badge would be permanent here as the item is non-clickable)
I.e Insights in this state would never show `New`
image

### DIFF
--- a/static/app/components/sidebar/index.tsx
+++ b/static/app/components/sidebar/index.tsx
@@ -622,7 +622,6 @@ function Sidebar() {
       icon={<IconSearch />}
       label={<GuideAnchor target="explore">{t('Explore')}</GuideAnchor>}
       id="explore"
-      isNew
       exact={!shouldAccordionFloat}
     >
       {traces}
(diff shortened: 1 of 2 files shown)


static/app/components/sidebar/index.tsx
```diff
@@ -621,6 +622,7 @@ function Sidebar() {
       icon={<IconSearch />}
       label={<GuideAnchor target="explore">{t('Explore')}</GuideAnchor>}
       id="explore"
+      isNew
```
gggritso, line 625:
```suggestion
```

confusingly, the "Explore" accordion is "not" new!

gggritso (commented):
Right on! Watch out for "Explore" though

static/app/components/sidebar/index.tsx
DominikB2014: Good catch, i thought it was, because it's tied to the same `performance-insights` flag

gggritso (approved):

PR #71931, opened 2024-06-03
ref(custom-views): Update tasks to delete custom views upon org or member deletion

### LINKED ISSUE #71802: Custom Views: Augment member removal endpoint to delete user's views

MichaelSun48:
This PR updates the Organization deletion task to make custom views a child of organization, thus deleting any custom views in the org upon org deletion. It also adds a new task for Organization members, and makes custom views a child of org members as well.

### DIFF
--- a/src/sentry/deletions/defaults/organizationmember.py
+++ b/src/sentry/deletions/defaults/organizationmember.py
@@ -1,0 +1,9 @@
+from sentry.deletions.base import ModelDeletionTask, ModelRelation
+from sentry.models.groupsearchview import GroupSearchView
+
+
+class OrganizationMemberDeletionTask(ModelDeletionTask):
+    def get_child_relations(self, instance):
+        relations = [ModelRelation(GroupSearchView, {"user_id": instance.user_id})]
+
+        return relations
--- a/src/sentry/deletions/defaults/organizationmember.py
+++ b/src/sentry/deletions/defaults/organizationmember.py
@@ -8,7 +8,7 @@ class OrganizationMemberDeletionTask(ModelDeletionTask):
         relations = [
             ModelRelation(
                 GroupSearchView,
-                {"user_id": instance.user_id, "organization": instance.organization},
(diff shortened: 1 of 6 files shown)


src/sentry/deletions/defaults/organizationmember.py
```diff
@@ -0,0 +1,10 @@
+from sentry.models.groupsearchview import GroupSearchView
+
+from ..base import ModelDeletionTask, ModelRelation
+
+
+class OrganizationMemberDeletionTask(ModelDeletionTask):
+    def get_child_relations(self, instance):
+        relations = [ModelRelation(GroupSearchView, {"user_id": instance.user_id})]
```
malwilley, line 8: Do we need to send the `organization_id` too? This seems like it would delete all of the user's views, but maybe I'm missing something

MichaelSun48: Yup I think you're right about that, fixing that now. I'll add a check in the tests to double check as well

malwilley (approved):

src/sentry/deletions/defaults/organizationmember.py
```diff
@@ -0,0 +1,15 @@
...
+
+from ..base import ModelDeletionTask, ModelRelation
+
+
+class OrganizationMemberDeletionTask(ModelDeletionTask):
+    def get_child_relations(self, instance):
+        relations = [
+            ModelRelation(
+                GroupSearchView,
+                {"user_id": instance.user_id, "organization": instance.organization},
```
mrduncan, line 11:
```suggestion
                {"user_id": instance.user_id, "organization_id": instance.organization_id},
```
You should be able to skip loading the Organization here since only the ID is needed

PR #71935, opened 2024-06-03
ref(messaging): Moved `types/integrations.py` to `integrations/*.py` 

iamrajjoshi:
https://www.notion.so/sentry/Task-list-for-Messaging-Vertical-ca4f431d43b84e72b8a91643fe880b66#410553168fb143ef857badd06e50ddbd
tldr: Ecosystem Team is working on refactoring our integration logic to be more stable. 

This PR aims to move all the code under `types/integration.py` to the `integration` folder. I split up the code into `integrations/types.py`, `integration/errors.py`, and `integrations/utils/provider.py`.

To make sure we can switch over to these new files, first I will make a PR to to update the imports in `getsentry`, then `sentry`

I will follow up with another PR to remove the file once I update the import statements in `sentry`.

### DIFF
--- a/src/sentry/integrations/errors.py
+++ b/src/sentry/integrations/errors.py
@@ -1,0 +1,6 @@
+class InvalidProviderException(Exception):
+    """
+    Provider that is passed does not exist.
+    """
+
+    pass
--- a/src/sentry/integrations/types.py
+++ b/src/sentry/integrations/types.py
@@ -1,0 +1,67 @@
+from enum import Enum
+
+from sentry.services.hybrid_cloud import ValueEqualityEnum
+
+
+class ExternalProviders(ValueEqualityEnum):
+    UNUSED_GH = 0
+    UNUSED_GL = 1
+
+    EMAIL = 100
+    SLACK = 110
+    MSTEAMS = 120
+    PAGERDUTY = 130
+    DISCORD = 140
+    OPSGENIE = 150
+    GITHUB = 200
+    GITHUB_ENTERPRISE = 201
+    GITLAB = 210
+
+    # TODO: do migration to delete this from database
+    CUSTOM = 700
+
+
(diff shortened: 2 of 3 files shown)


src/sentry/integrations/errors.py
```diff
@@ -0,0 +1,6 @@
+class InvalidProviderException(Exception):
```
iamrajjoshi, line 1: this is a new file I created during refactoring

src/sentry/integrations/types.py
```diff
@@ -0,0 +1,71 @@
+from enum import Enum
```
iamrajjoshi, line 1: this is a new file I created during refactoring

src/sentry/integrations/utils/providers.py
```diff
@@ -0,0 +1,39 @@
+from collections.abc import Sequence
```
iamrajjoshi, line 1: this is a new file I created during refactoring

src/sentry/tasks/integrations/__init__.py
```diff
@@ -54,7 +54,6 @@ def should_comment_sync(
```
iamrajjoshi, line 54: gonna delete this file in a followup

src/sentry/models/integrations/__init__.py
```diff
@@ -16,7 +16,7 @@
 )
```
iamrajjoshi, line 16: gonna delete this file in a followup, if i can

markstory (approved):
> To make sure we can switch over to these new files, first I will make a PR to to update the imports in getsentry, then sentry

Another solution to moving symbols around without breaking getsentry is to move symbols to their new location, and in the old location import from the new location and re-export with `__all__`

cathteng (approved):

PR #71939, opened 2024-06-03
ref: remove expected_errors= from safe_execute

asottile-sentry:
this prevents it from being made typesafe (followup PR) -- fixed the one user of this functionality to just use plain try/except

### DIFF
--- a/src/sentry/tasks/post_process.py
+++ b/src/sentry/tasks/post_process.py
@@ -1280,13 +1280,16 @@ def plugin_post_process_group(plugin_slug, event, **kwargs):
     from sentry.plugins.base import plugins
 
     plugin = plugins.get(plugin_slug)
-    safe_execute(
-        plugin.post_process,
-        event=event,
-        group=event.group,
-        expected_errors=(PluginError,),
-        **kwargs,
-    )
+    try:
+        plugin.post_process(
+            event=event,
+            group=event.group,
+            **kwargs,
+        )
+    except PluginError as e:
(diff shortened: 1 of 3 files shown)


joshuarli (approved):

src/sentry/tasks/post_process.py
```diff
@@ -1280,13 +1280,16 @@ def plugin_post_process_group(plugin_slug, event, **kwargs):
...
-        **kwargs,
-    )
+    try:
+        plugin.post_process(
+            event=event,
+            group=event.group,
+            **kwargs,
+        )
+    except PluginError as e:
+        logger.info("post_process.process_error_ignored", extra={"exception": e})
```
vartec, line 1290: why `extra={"exception": e}` and not just `exc_info=True`?

src/sentry/tasks/post_process.py
```diff
@@ -1280,13 +1280,16 @@ def plugin_post_process_group(plugin_slug, event, **kwargs):
...
+    except Exception as e:
+        logger.exception("post_process.process_error", extra={"exception": e})
```
vartec, line 1292: `extra={"exception": e}` in an `exception()` is kinda superfluous.

vartec (approved):

src/sentry/tasks/post_process.py
asottile-sentry: I am simply matching the (bad) behaviour that `safe_execute` has

PR #71940, opened 2024-06-03
feat(discover): Add top level transactions dataset

shruthilayaj:
Add a top level API to create transaction events
and timeseries queries.
Tests are _very_ similar to the `discover` tests,
except we load transactions data


src/sentry/search/events/datasets/discover.py
```diff
@@ -2002,3 +2005,15 @@ def _error_handled_filter_converter(
...
     def _key_transaction_filter_converter(self, search_filter: SearchFilter) -> WhereType | None:
         return filter_aliases.team_key_transaction_filter(self.builder, search_filter)
+
+    def _event_type_filter_converter(self, search_filter: SearchFilter) -> WhereType | None:
+        if self.builder.dataset == Dataset.Transactions:
+            if search_filter.operator in ["=", "IN"] and search_filter.value.value in [
+                "transaction",
+                ["transaction"],
+            ]:
+                return Condition(Function("equals", [1, 1]), Op.EQ, 1)
```
shruthilayaj, line 2015: No-op on event type filter with the transactions dataset since the field doesn't exist on the transactions table. right now a condition like `event.type:transaction` get converted into a tag condition

mjq (approved):
The 5x repetition of each giant arg list in `src/sentry/snuba/transactions.py` and `src/sentry/snuba/discover.py` hurts me a bit, but this also seems to be the pattern we follow everywhere with these snuba dataset modules, so if we were really going to address it we should probably do it all at once - which is a bigger task than this PR should be forced to address.

TL;DR: LGTM, but I hope we can loop back around to these snuba dataset modules later 😅

src/sentry/search/events/datasets/discover.py
wmak: 🤔 does returning None not work here? (legit forget, but i thought we just ignore the condition in that case)

wmak (approved):
lgtm, before merging please try doing None in the reply, if it doesn't work should be fine

src/sentry/search/events/datasets/discover.py
shruthilayaj: oh nice, that works

shruthilayaj: hmm actually, for an OR condition (for example, `project:foo OR event.type:transaction`), we'll want to no-op instead of setting to None because setting to None will evaluate to `project:foo` which is a more restrictive condition than `project:foo OR event.type:transaction`)

shruthilayaj: my example _is_ as really weird condition though, and would probably evaluate as an ambiguous/"discover" dataset when we try to determine the dataset, so I think I'll return None here for now

PR #71942, opened 2024-06-03
feat(custom-views): add custom views get endpoint

### LINKED ISSUE #71580: Custom Views: Create GET Endpoint
Create `GET` `<ORG_ID_OR_SLUG>/user-views/` endpoint 

This endpoint will be responsible for retrieving an org member's custom tabs from the database. This endpoint will also need to be responsible for returning the default views if an org member has no custom views. It will also need to be gated behind a feature flag until GA. 

Logic: 
* Query database for an org member's views. 
* Return found rows, or default rows if none found

MichaelSun48:
This endpoint creates the `GET` `organizations/<org_id_or_slug>/group-search-views/` Endpoint along with some tests. This endpoint will be responsible for fetching a user's custom views within an organization. 

⚠️ This PR is dependent on #71731  being ~~merged~~ run in production, and this branch will need to be rebased upon that happening

### DIFF
--- a/src/sentry/api/serializers/models/groupsearchview.py
+++ b/src/sentry/api/serializers/models/groupsearchview.py
@@ -1,10 +1,23 @@
+from typing import TypedDict
+
 from sentry.api.serializers import Serializer, register
 from sentry.models.groupsearchview import GroupSearchView
+from sentry.models.savedsearch import SORT_LITERALS
+
+
+class View(TypedDict):
+    id: str
+    name: str
+    query: str
+    querySort: SORT_LITERALS
+    position: int
+    dateCreated: str | None
+    dateUpdated: str | None
 
 
 @register(GroupSearchView)
 class GroupSearchViewSerializer(Serializer):
-    def serialize(self, obj, attrs, user):
+    def serialize(self, obj, attrs, user) -> View:
         return {
             "id": str(obj.id),
             "name": obj.name,
--- a/src/sentry/api/serializers/models/groupsearchview.py
+++ b/src/sentry/api/serializers/models/groupsearchview.py
@@ -1,0 +1,16 @@
+from sentry.api.serializers import Serializer, register
+from sentry.models.groupsearchview import GroupSearchView
+
+
+@register(GroupSearchView)
+class GroupSearchViewSerializer(Serializer):
+    def serialize(self, obj, attrs, user):
+        return {
+            "id": str(obj.id),
+            "name": obj.name,
+            "query": obj.query,
+            "querySort": obj.query_sort,
+            "position": obj.position,
+            "dateCreated": obj.date_added,
+            "dateUpdated": obj.date_updated,
+        }
--- a/src/sentry/api/serializers/models/groupsearchview.py
+++ b/src/sentry/api/serializers/models/groupsearchview.py
@@ -5,7 +5,7 @@ from sentry.models.groupsearchview import GroupSearchView
 from sentry.models.savedsearch import SORT_LITERALS
 
 
-class View(TypedDict):
+class GroupSearchViewSerializerResponse(TypedDict):
     id: str
     name: str
     query: str
@@ -17,7 +17,7 @@ class View(TypedDict):
 
 @register(GroupSearchView)
 class GroupSearchViewSerializer(Serializer):
-    def serialize(self, obj, attrs, user) -> View:
+    def serialize(self, obj, attrs, user, **kwargs) -> GroupSearchViewSerializerResponse:
         return {
             "id": str(obj.id),
             "name": obj.name,
--- a/src/sentry/issues/endpoints/organization_group_search_views.py
+++ b/src/sentry/issues/endpoints/organization_group_search_views.py
@@ -9,12 +9,15 @@ from sentry.api.base import region_silo_endpoint
 from sentry.api.bases.organization import OrganizationEndpoint, OrganizationPermission
 from sentry.api.paginator import SequencePaginator
 from sentry.api.serializers import serialize
-from sentry.api.serializers.models.groupsearchview import GroupSearchViewSerializer, View
+from sentry.api.serializers.models.groupsearchview import (
+    GroupSearchViewSerializer,
+    GroupSearchViewSerializerResponse,
+)
 from sentry.models.groupsearchview import GroupSearchView
 from sentry.models.organization import Organization
 from sentry.models.savedsearch import SortOptions
 
-DEFAULT_VIEWS: list[View] = [
+DEFAULT_VIEWS: list[GroupSearchViewSerializerResponse] = [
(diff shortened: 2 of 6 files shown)


src/sentry/issues/endpoints/organization_group_search_views.py
```diff
@@ -0,0 +1,43 @@
...
+    }
+    owner = ApiOwner.ISSUES
+    permission_classes = (MemberPermission,)
+
+    def get(self, request: Request, organization: Organization) -> Response:
+        """
+        List a organization member's custom views
+        `````````````````````````````````````````
+
+        Retrieve a list of custom views for a given organization member.
```
mrduncan, line 33:
```suggestion
        Retrieve a list of custom views for the current organization member.
```

src/sentry/issues/endpoints/organization_group_search_views.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        List the current organization member's custom views
+        `````````````````````````````````````````
+
+        Retrieve a list of custom views for the current organization member.
+        """
+        if not features.has("organizations:issue-stream-custom-views", organization):
+            return Response(status=status.HTTP_405_METHOD_NOT_ALLOWED)
+
+        query = GroupSearchView.objects.filter(organization=organization, user_id=request.user.id)
```
MichaelSun48, line 40: I will add some logic here in the near future to send defaults tabs if the user has no custom views

src/sentry/issues/endpoints/organization_group_search_views.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def get(self, request: Request, organization: Organization) -> Response:
+        """
+        List the current organization member's custom views
+        `````````````````````````````````````````
+
+        Retrieve a list of custom views for the current organization member.
+        """
+        if not features.has("organizations:issue-stream-custom-views", organization):
+            return Response(status=status.HTTP_405_METHOD_NOT_ALLOWED)
```
malwilley, line 49: Not that big of a deal, but use usually return a 404 for this kind of thing. 405 is specific to the method, we just want to behave like the endpoint doesn't exist.

src/sentry/issues/endpoints/organization_group_search_views.py
```diff
@@ -0,0 +1,68 @@
...
+
+        query = GroupSearchView.objects.filter(organization=organization, user_id=request.user.id)
+
+        # Return only the prioritized view if user has no custom views yet
```
roggenkemper, line 53: do we expect there to be multiple default views eventually?

MichaelSun48: We'll analyze the data after release to make that decision - if lots of people are immediately re-adding the old default tabs, then maybe. If everyone just sticks to the Prioritized tab or adds new tabs that aren't part of the old defaults, then probably not.

src/sentry/issues/endpoints/organization_group_search_views.py
```diff
@@ -0,0 +1,68 @@
...
+from sentry.api.api_publish_status import ApiPublishStatus
+from sentry.api.base import region_silo_endpoint
+from sentry.api.bases.organization import OrganizationEndpoint, OrganizationPermission
+from sentry.api.paginator import SequencePaginator
+from sentry.api.serializers import serialize
+from sentry.api.serializers.models.groupsearchview import GroupSearchViewSerializer
+from sentry.models.groupsearchview import GroupSearchView
+from sentry.models.organization import Organization
+
+DEFAULT_VIEWS = [
```
malwilley, line 16: Can we define this in such a way (with types, or by using the serializer) so that we can guarantee that the keys here are correct? I'm worried that we might add a column and forget to modify this

src/sentry/api/serializers/models/groupsearchview.py
```diff
@@ -0,0 +1,16 @@
+from sentry.api.serializers import Serializer, register
+from sentry.models.groupsearchview import GroupSearchView
+
+
+@register(GroupSearchView)
+class GroupSearchViewSerializer(Serializer):
+    def serialize(self, obj, attrs, user):
```
malwilley, line 7: Can you add a response type here?

src/sentry/issues/endpoints/organization_group_search_views.py
```diff
@@ -0,0 +1,68 @@
...
+    {
+        "id": "",
+        "name": "Prioritized",
+        "query": "is:unresolved issue.priority:[high, medium]",
+        "querySort": "date",
```
malwilley, line 21: A bit better if we can use the enum directly here

src/sentry/api/serializers/models/groupsearchview.py
MichaelSun48: image

Looks like mypy does not like that

malwilley: Are you typing it as a TypedDict? I see other serializers that have done that: https://github.com/getsentry/sentry/blob/8d255a2b22c6d0f7e53ed7a8fe94fb080a2060d0/src/sentry/monitors/serializers.py#L36

MichaelSun48: Ah yes I was, I'll look at that one and see if I can make it work

src/sentry/api/serializers/models/groupsearchview.py
```diff
@@ -0,0 +1,29 @@
...
+from sentry.api.serializers import Serializer, register
+from sentry.models.groupsearchview import GroupSearchView
+from sentry.models.savedsearch import SORT_LITERALS
+
+
+class View(TypedDict):
+    id: str
+    name: str
+    query: str
+    querySort: SORT_LITERALS
```
malwilley, line 12: Can this not be typed as `SortOptions`?

src/sentry/api/serializers/models/groupsearchview.py
```diff
@@ -0,0 +1,29 @@
+from typing import TypedDict
+
+from sentry.api.serializers import Serializer, register
+from sentry.models.groupsearchview import GroupSearchView
+from sentry.models.savedsearch import SORT_LITERALS
+
+
+class View(TypedDict):
```
malwilley, line 8: nit: type name could be more descriptive

src/sentry/api/serializers/models/groupsearchview.py
MichaelSun48: I tried for like an hour to get this to work before just giving up. For some reason, mypy refuses to enforce the type if I set it to `SortOptions` - I set`querySort` of a `view` instance to `"awioef"` and mypy thought it was fine.

malwilley (approved):

PR #71945, opened 2024-06-03
feat(trace-explorer): Return slice index and width for timeline break…

Zylphrex:
…down

Instead of using the start/end to render the timeline, we're moving to using a slice index + offset.


k-fish (approved):
🙏

PR #71952, opened 2024-06-03
feat(crons): Instrument processing error storage with analytics

davidenwang:
Records an analytics event to track when a processing error is stored, will be useful for verifying processing errors as we roll out this feature to new orgs

depends on https://github.com/getsentry/sentry/pull/71951

### DIFF
--- a/src/sentry/monitors/processing_errors/manager.py
+++ b/src/sentry/monitors/processing_errors/manager.py
@@ -9,7 +9,7 @@ from django.conf import settings
 from redis.client import StrictRedis
 from rediscluster import RedisCluster
 
-from sentry import features
+from sentry import analytics, features
 from sentry.models.organization import Organization
 from sentry.models.project import Project
 from sentry.monitors.models import Monitor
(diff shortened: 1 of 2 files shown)


src/sentry/monitors/processing_errors/manager.py
```diff
@@ -178,6 +178,13 @@ def handle_processing_errors(item: CheckinItem, error: ProcessingErrorsException
                 "sdk_platform": item.message["sdk"],
             },
         )
+        analytics.record(
```
evanpurkhiser, line 181: how much latency will this add to the system I wonder

davidenwang: yea thats a fair point. I guess the main motivation for doing something like this is so that as we roll out the feature to a number of orgs we can check in on them and make sure the errors they are receiving make sense. Would a better pattern to take this out of the critical ingest code path be to use signals?

davidenwang: nvm i'm not sure if these signals work async the way i'm thinking about

davidenwang: follow up here, after talking with the team, it seems like adding one call to analytics shouldn't be a huge problem

wedamija (approved):

PR #71954, opened 2024-06-03
feat(relocation): Don't email manually claimed users

azaslavsky:
Sometimes, it is useful for an admin to merge user accounts before "unpausing" an active relocation. In such cases, we do not want to email users that have been manually merged.

### DIFF
--- a/tests/sentry/tasks/test_relocation.py
+++ b/tests/sentry/tasks/test_relocation.py
@@ -1935,14 +1935,11 @@ class NotifyingUsersTest(RelocationTaskTestCase):
             ).inserted_identifiers.values()
         )
         assert len(self.imported_orgs) == 1
-        assert (
-            len(
-                ControlImportChunkReplica.objects.get(
-                    import_uuid=self.uuid, model="sentry.user"
-                ).inserted_map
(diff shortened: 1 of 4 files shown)


tests/sentry/tasks/test_relocation.py
```diff
@@ -1971,6 +1968,38 @@ def test_success(
...
+            notifying_users(self.uuid)
+
+            # Called once for each user imported that has not been manually claimed. Since we
+            # imported 2 users in `fresh-install.json`, but then manually claimed one at the top of
+            # this test, only one user remains.
+            assert mock_relocation_email.call_count == 1
+            email_targets = [
+                mock_relocation_email.call_args_list[0][0][0].username,
+            ]
+            assert sorted(mock_relocation_email.call_args_list[0][0][2]) == self.imported_orgs
```
markstory, line 1993: Mocks also have a few [assertion methods](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_with), you could likely use `mock_relocation_email.assert_called_with()`

markstory (approved):

PR #71955, opened 2024-06-03
ref(messaging): Cleanup `integration/__init__.py`

iamrajjoshi:
Cleaning up `integration/__init__.py` so we don't load everything in memory as well as prevent circular import issues when we refactor things later down the road.

Removed some imports from the file, so had to update import statements to add the complete path.


src/sentry/integrations/__init__.py
```diff
@@ -32,16 +32,11 @@
 )
 
 from .analytics import register_analytics
```
iamrajjoshi, line 34: this file contains the main changes, rest are modifications of imports b/c of this change.

ykamo001 (approved):

PR #71984, opened 2024-06-04
feat(metrics): add tagValuePrefix option to metrics tag details endpoint

### LINKED ISSUE #71227: Add prefix to request for filter suggestions to use new meta endpoint's capabilities & reduce DB load
The new implementation of fetch_metric_tag_values in https://github.com/getsentry/sentry/pull/69803/files makes it possible to specify a tag_key prefix that the results are filtered by. This could potentially reduce query times & load on the database when implemented in the endpoints for metrics metadata.

shellmayr:
Closes https://github.com/getsentry/sentry/issues/71227

### DIFF
--- a/src/sentry/api/endpoints/organization_metrics_tag_details.py
+++ b/src/sentry/api/endpoints/organization_metrics_tag_details.py
@@ -56,6 +56,8 @@ class OrganizationMetricsTagDetailsEndpoint(OrganizationEndpoint):
 
         if len(tag_values_formatted) > 0:
             return Response(tag_values_formatted, status=200)
+        elif len(tag_values_formatted) == 0 and len(tag_value_prefix) > 0:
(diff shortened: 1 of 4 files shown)


src/sentry/api/endpoints/organization_metrics_tag_details.py
```diff
@@ -23,6 +23,7 @@ class OrganizationMetricsTagDetailsEndpoint(OrganizationEndpoint):
 
     def get(self, request: Request, organization, tag_name) -> Response:
         metric_names = request.GET.getlist("metric") or []
+        tag_value_prefix = request.GET.get("tagValuePrefix") or ""
```
obostjancic, line 26: nit: since the url of this endpoint scopes the response to tags, i would name this param `prefix` or `valuePrefix`

obostjancic (approved):
🚀

PR #71988, opened 2024-06-04
feat(discover): Save Discover query split decisions

shruthilayaj:
Save split query decision and return the decision as a part of events meta


src/sentry/api/endpoints/organization_events.py
```diff
@@ -353,91 +358,182 @@ def _data_fn(scopedDataset, offset, limit, query) -> dict[str, Any]:
...
+                if does_widget_have_split and not has_override_feature:
+                    # This is essentially cached behaviour and we skip the check
+                    split_query = scoped_query
+                    if widget.discover_widget_split == DashboardWidgetTypes.ERROR_EVENTS:
+                        split_dataset = discover
+                        split_query = (
+                            f"({scoped_query}) AND !event.type:transaction"
+                            if scoped_query
+                            else "!event.type:transaction"
+                        )
```
shruthilayaj, line 380: made this change because I noticed a bug where if scoped query is an empty string, it evaluates to:
`() AND !event.type:transaction`
and the `()` get added as a free text condition

src/sentry/api/bases/organization_events.py
```diff
@@ -260,6 +261,25 @@ def save_split_decision(self, widget, has_errors, has_transactions_data):
...
+            decision = DiscoverSavedQueryTypes.ERROR_EVENTS
+        elif not has_errors and has_transactions_data:
+            decision = DiscoverSavedQueryTypes.TRANSACTION_LIKE
+        else:
+            # In the case that neither or both datasets return data,
+            # we don't split this yet and can make multiple queries to check each time.
+            # This will help newly created widgets or infrequent count
+            # widgets that shouldn't be prematurely assigned a side.
+            decision = DiscoverSavedQueryTypes.DISCOVER
+        sentry_sdk.set_tag("discover_split_decision", decision)
```
wmak, line 276: nit, can we call this
```suggestion
        sentry_sdk.set_tag("discover.split_decision", decision)
```
so the ui does that thing where we organize this tag with the other discover tags

src/sentry/api/endpoints/organization_events.py
```diff
@@ -353,91 +364,180 @@ def _data_fn(scopedDataset, offset, limit, query) -> dict[str, Any]:
                 on_demand_metrics_type=on_demand_metrics_type,
             )
 
-        def data_fn_factory(scopedDataset):
-            """
-            This factory closes over query and dataset in order to make an additional request to the errors dataset
-            in the case that this request is from a dashboard widget and we're trying to split their discover dataset.
+        def _dashboards_data_fn(scopedDataset, offset, limit, scoped_query, dashboard_widget_id):
```
wmak, line 367: Can we add the sentry_sdk.tracing decorator for this function so we get a span saying how long it took?

wmak (approved):
mostly instrumentation comments

This is behind a flag so not a blocker, but i'm sorta concerned about the perf of it all, I wonder if we could run our data functions in parallel so that there isn't a large increase in runtime?

shruthilayaj:
Good idea, I'll add parallelization in a follow up

PR #71989, opened 2024-06-04
chore: update biome to v1.8.0

anonrig:
Ready to land.

This pull-request updates Biome to latest version and fixes JSON formatter being disabled. This bug was recently resolved in v1.8.0 release. All relevant changes are due to the JSON formatter being disabled until now.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -22,7 +22,8 @@
         "useHookAtTopLevel": "error",
         "useIsNan": "error",
         "noUnusedPrivateClassMembers": "error",
-        "noInvalidUseBeforeDeclaration": "error"
+        "noInvalidUseBeforeDeclaration": "error",
+        "noNodejsModules": "error"
       },
       "complexity": {
         "useFlatMap": "error",
@@ -36,12 +37,11 @@
       },
       "nursery": {
         "noDuplicateJsonKeys": "error",
-        "noNodejsModules": "error",
         "noRestrictedImports": {
           "level": "warn",
           "options": {
             "paths": {
-              "sentry/types":  "Please import directly. For example: import type {Organization} from 'sentry/types/organization'"
+              "sentry/types": "Please import directly. For example: import type {Organization} from 'sentry/types/organization'"
             }
           }
         }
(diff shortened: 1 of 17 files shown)


asottile-sentry (approved):

biome.json
```diff
@@ -125,18 +125,13 @@
   "json": {
     "formatter": {
       "enabled": true
+    },
+    "parser": {
+      "allowComments": true,
+      "allowTrailingCommas": true
     }
```
ryan953, line 132: Looks like this is allowing json files to be created that will not be parsable

Re-reading the docs the hunk below wasn't needed, see: https://biomejs.dev/guides/how-biome-works/#json-like-files, but this seems to be not helpful for folks either. The linter should complain about comments and auto-fix trailing commas before tests start to fail.

Here's a PR and the CI run to demonstrate: 
- https://github.com/getsentry/sentry/pull/72014
- https://github.com/getsentry/sentry/actions/runs/9370576639/job/25797888214?pr=72014

Would also need to be fixed in https://github.com/getsentry/getsentry/pull/14185

anonrig: @ryan953 I personally think this should be handled by the build process, but I'm ok with reverting this change as well. Just FYI, any additions to `overrides` impact the execution time of Biome.

ryan953: How would or could the build process handle this? I'm not able to imagine what you mean

ryan953: Certainly a rules change like this probably shouldn't be hidden inside a PR that purports to update a dependency. So if the purpose was to remove the overrides and improve perf that's probably something to be done independently and to measure. I would assume that the perf difference will amount to a fraction of a percent of CI time

anonrig: > How would or could the build process handle this? 

Whoever loads the content of the JSON file and tries to parse it should and will throw an error if they don't support it. We can keep an explicit list of what is supported and what is not, but most libraries do support JSON files with comment. 

> Certainly a rules change like this probably shouldn't be hidden inside a PR that purports to update a dependency. So if the purpose was to remove the overrides and improve perf that's probably something to be done independently and to measure. I would assume that the perf difference will amount to a fraction of a percent of CI time

Thanks for the feedback 👍

PR #71995, opened 2024-06-04
feat: enable biome css formatter and linter

anonrig:
Probably too early to land this, since there might be some hidden bugs with the CSS formatter and linter, but it's nice to finally remove `stylelint` from Sentry.

cc @getsentry/app-frontend @getsentry/dev-infra


scttcper:
`postcss-styled-syntax` is a related package

this could be removed 
https://github.com/getsentry/sentry/blob/master/config/tsconfig.base.json#L108-L112

scttcper:
still just seeing stylelint and not biome pick these up

image

anonrig:
> `postcss-styled-syntax` is a related package
> 
> this could be removed https://github.com/getsentry/sentry/blob/master/config/tsconfig.base.json#L108-L112

afaict, styled-components are still not yet supported. only CSS/LESS files are supported at the moment.

scttcper:
@anonrig ah i see this is a replacement for linting the css files

PR #71999, opened 2024-06-04
ref(related_issues): Make the code a bit more clear

armenzg:


static/app/views/issueDetails/groupEventHeader.tsx
```diff
@@ -26,23 +26,23 @@ function GroupEventHeader({event, group, project}: GroupEventHeaderProps) {
     'related-issues-issue-details-page'
   );
   // This is also called within the TraceTimeline component but caching will save a second call
-  const {isError, isLoading, oneOtherIssueEvent} = useTraceTimelineEvents({
+  const {oneOtherIssueEvent} = useTraceTimelineEvents({
     event,
   });
-  const readyToShow = !isLoading && !isError;
```
armenzg, line 32: This condition is not really necessary as you will see below.

static/app/views/issueDetails/groupEventHeader.tsx
```diff
@@ -26,23 +26,23 @@ function GroupEventHeader({event, group, project}: GroupEventHeaderProps) {
...
 
   return (
     <StyledDataSection>
       <GroupEventCarousel group={group} event={event} projectSlug={project.slug} />
-      {isRelatedIssuesEnabled && readyToShow && oneOtherIssueEvent === undefined && (
-        <TraceLink event={event} />
-      )}
       {isRelatedIssuesEnabled && oneOtherIssueEvent && (
```
armenzg, line 36: If `oneOtherIssueEvent` is defined, we know that the data was fetched and that there was no error.

static/app/views/issueDetails/groupEventHeader.tsx
```diff
@@ -26,23 +26,23 @@ function GroupEventHeader({event, group, project}: GroupEventHeaderProps) {
...
-  const {isError, isLoading, oneOtherIssueEvent} = useTraceTimelineEvents({
+  const {oneOtherIssueEvent} = useTraceTimelineEvents({
     event,
   });
-  const readyToShow = !isLoading && !isError;
 
   return (
     <StyledDataSection>
       <GroupEventCarousel group={group} event={event} projectSlug={project.slug} />
-      {isRelatedIssuesEnabled && readyToShow && oneOtherIssueEvent === undefined && (
```
armenzg, line 37: Moving this block closer to `<TraceTimeline>` since they would be shown together.

evanpurkhiser (approved):

PR #72002, opened 2024-06-04
Avoid creating new `RawEvent`s and `ProcessingIssue`s

Swatinem:
As part of removing legacy reprocessing, this removes the code which creates new `ProcessingIssue`s and `RawEvent`s from the ingest path.

After a while, these `RawEvent`s will either naturally expire / be resolved, or we will bulk resolve / drop them when continueing with removal of this feature.


loewenheim (approved):

src/sentry/reprocessing.py
```diff
@@ -1,15 +1,11 @@
...
 from typing import Any
 
-import sentry_sdk
-
-from sentry.models.options.project_option import ProjectOption
-
 # Is reprocessing on or off by default?
 REPROCESSING_DEFAULT = False
 REPROCESSING_OPTION = "sentry:reprocessing_active"
+# TODO: make sure to clean these option up at some point:
```
anonrig, line 8: Can we be a little bit more clear about this comment?

Swatinem: since this is a project option, we should bulk-remove these from the DB at some point.
I created an issue for that: https://github.com/getsentry/team-processing/issues/160

PR #72003, opened 2024-06-04
ref: make safe_execute typesafe

asottile-sentry:

### DIFF
--- a/src/sentry/api/serializers/models/group_stream.py
+++ b/src/sentry/api/serializers/models/group_stream.py
@@ -78,7 +78,7 @@ def get_available_issue_plugins(request: Request, group):
         if isinstance(plugin, IssueTrackingPlugin2):
             if is_plugin_deprecated(plugin, project):
                 continue
-            plugin_issues = safe_execute(plugin.plugin_issues, request, group, plugin_issues)
+            safe_execute(plugin.plugin_issues, request, group, plugin_issues)
     return plugin_issues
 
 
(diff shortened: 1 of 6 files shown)


src/sentry/reprocessing2.py
```diff
@@ -240,8 +240,8 @@ def reprocess_event(project_id: int, event_id: str, start_time: float) -> None:
     )
 
 
-def get_original_group_id(event: Event) -> int | None:
-    return get_path(event.data, "contexts", "reprocessing", "original_issue_id")
+def get_original_group_id(event: Event) -> int:
+    return event.data["contexts"]["reprocessing"]["original_issue_id"]
```
asottile-sentry, line 244: this would crash later on when accessing `group_id` as `None` -- so I believe this always has to be present

src/sentry/api/serializers/models/group_stream.py
```diff
@@ -78,7 +78,7 @@ def get_available_issue_plugins(request: Request, group):
         if isinstance(plugin, IssueTrackingPlugin2):
             if is_plugin_deprecated(plugin, project):
                 continue
-            plugin_issues = safe_execute(plugin.plugin_issues, request, group, plugin_issues)
+            safe_execute(plugin.plugin_issues, request, group, plugin_issues)
```
asottile-sentry, line 81: of note, this is in an ignored file but I made sure to check that the blocked errors did not regress

Swatinem (approved):

anonrig (approved):

PR #72015, opened 2024-06-04
feat(insights): show update SDK banner cache module

DominikB2014:
1. If the cache hit miss rate chart aren't showing data, this likely means the customer is on an outdated SDK or they aren't sending `cache.hit`, so we display a message accordingly
image

2. The message is not displayed if it is in an onboarding state

### DIFF
--- a/static/app/utils/performance/contexts/pageAlert.tsx
+++ b/static/app/utils/performance/contexts/pageAlert.tsx
@@ -10,6 +10,7 @@ type PageAlertType = keyof Theme['alert'];
 
 export enum DismissId {
   RESOURCE_SIZE_ALERT = 0,
+  CACHE_SDK_UPDATE_ALERT = 1,
 }
 
 export type PageAlertOptions = {
@@ -81,7 +82,7 @@ export function PageAlertProvider({children}: {children: React.ReactNode}) {
 
 export function PageAlert() {
   const {pageAlert} = useContext(pageErrorContext);
-  const [dismissedAlerts, setDismissedAlerts] = useLocalStorageState<string[]>(
+  const [dismissedAlerts, setDismissedAlerts] = useLocalStorageState<number[]>(
     localStorageKey,
     []
   );
--- a/static/app/views/performance/cache/cacheLandingPage.tsx
+++ b/static/app/views/performance/cache/cacheLandingPage.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useEffect} from 'react';
 import keyBy from 'lodash/keyBy';
 
 import FeatureBadge from 'sentry/components/badge/featureBadge';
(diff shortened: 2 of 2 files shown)


static/app/utils/performance/contexts/pageAlert.tsx
```diff
@@ -81,7 +82,7 @@ export function PageAlertProvider({children}: {children: React.ReactNode}) {
 
 export function PageAlert() {
   const {pageAlert} = useContext(pageErrorContext);
-  const [dismissedAlerts, setDismissedAlerts] = useLocalStorageState<string[]>(
+  const [dismissedAlerts, setDismissedAlerts] = useLocalStorageState<number[]>(
```
DominikB2014, line 85: I have no idea how this didn't give a type error before, dismissIds are numbers

0Calories:
Should we also display an alert banner in the case when there's no data found?

image

static/app/views/performance/cache/cacheLandingPage.tsx
```diff
@@ -45,8 +55,22 @@ import {DataTitles} from 'sentry/views/starfish/views/spans/types';
...
+    {t(
+      `If you're noticing missing cache data, try updating to the latest SDK or ensure spans are manually instrumented with the right attributes. `
+    )}
+    <ExternalLink href={`${MODULE_DOC_LINK}#instrumentation`}>
+      {t('Read the Docs')}
+    </ExternalLink>
+  </Fragment>
+);
+
+const CACHE_ERROR_MESSAGE = 'Column cache.hit was not found in metrics indexer';
```
0Calories, line 69: I'm a little confused because we're referring to the data as `cacheHitRate`, but we're querying on `CACHE_MISS_RATE` and we display the chart as Miss Rate, is this intentional?

0Calories (approved):
LGTM

0Calories:
I'd recommend some test coverage for this too since there's a bit of branching logic and we do slightly finicky things like a conditional on an exact error message

static/app/views/performance/cache/cacheLandingPage.tsx
DominikB2014: Not intentional, we switching from hit rate to miss rate and I didn't update the variable names. I'll clean this up in another PR!

Wrt the line u commented on tho, the error message does make sense for cache_miss_rate, because its discover function uses `cache.hit` under the hood.
 `cache_miss_rate = count(cache.hit == 'false') / (count(cache.hit == 'false') + count(cache.hit == 'true'))

static/app/views/performance/cache/cacheLandingPage.tsx
```diff
@@ -45,8 +55,22 @@ import {DataTitles} from 'sentry/views/starfish/views/spans/types';
 const {CACHE_MISS_RATE} = SpanFunction;
 const {CACHE_ITEM_SIZE} = SpanMetricsField;
 
+const SDK_UPDATE_ALERT = (
+  <Fragment>
+    {t(
+      `If you're noticing missing cache data, try updating to the latest SDK or ensure spans are manually instrumented with the right attributes. `
```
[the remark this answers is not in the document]
DominikB2014: Hmm, i just copied this from the message we used for resource module, but I totally agree

static/app/views/performance/cache/cacheLandingPage.tsx
0Calories: Okay awesome. The error message line made sense to me but I had no way to highlight the cacheHit variables since they weren't part of the changes in this PR

PR #72017, opened 2024-06-04
chore(slack): introduce SlackClient using slack_sdk

cathteng:
Added slack_sdk here https://github.com/getsentry/sentry/pull/71881

This PR adds a `SlackClient` that subclasses `WebClient` from the official Slack SDK. To make the usage the same as the in-house `SlackClient`, the abstraction accepts `integration_id` and finds the access token before initializing an instance of the `WebClient`.

We also want to track the response outcomes for every outgoing API call to Slack, so I added a wrapper that executes right after every single API call to increment a metric.

### DIFF
--- a/src/sentry/integrations/slack/actions/notification.py
+++ b/src/sentry/integrations/slack/actions/notification.py
@@ -22,7 +22,7 @@ from sentry.integrations.slack.message_builder.issues import SlackIssuesMessageB
 from sentry.integrations.slack.message_builder.notifications.rule_save_edit import (
     SlackRuleSaveEditMessageBuilder,
 )
-from sentry.integrations.slack.sdk_client import SlackClient as SlackSdkClient
+from sentry.integrations.slack.sdk_client import SlackSdkClient
 from sentry.integrations.slack.utils import get_channel_id
 from sentry.models.integrations.integration import Integration
 from sentry.models.rule import Rule
@@ -103,8 +103,10 @@ class SlackNotifyServiceAction(IntegrationEventAction):
                 "unfurl_links": False,
                 "unfurl_media": False,
             }
+            json_blocks = None
             if payload_blocks := blocks.get("blocks"):
-                payload["blocks"] = orjson.dumps(payload_blocks).decode()
+                json_blocks = orjson.dumps(payload_blocks).decode()
+                payload["blocks"] = json_blocks
 
             rule = rules[0] if rules else None
             rule_to_use = self.rule if self.rule else rule
@@ -128,6 +130,8 @@ class SlackNotifyServiceAction(IntegrationEventAction):
 
             # Only try to get the parent notification message if the organization is in the FF
             # We need to search by rule action uuid and rule id, so only search if they exist
+            reply_broadcast = False
+            thread_ts = None
             if (
                 features.has(
                     "organizations:slack-thread-issue-alert", event.group.project.organization
@@ -156,26 +160,29 @@ class SlackNotifyServiceAction(IntegrationEventAction):
                     # To reply to a thread, use the specific key in the payload as referenced by the docs
                     # https://api.slack.com/methods/chat.postMessage#arg_thread_ts
                     payload["thread_ts"] = parent_notification_message.message_identifier
+                    thread_ts = parent_notification_message.message_identifier
                     # If this flow is triggered again for the same issue, we want it to be seen in the main channel
                     payload["reply_broadcast"] = True
+                    reply_broadcast = True
 
             organization = event.group.project.organization
 
             if organization.id in options.get("slack.sdk-issue-alert"):
-                client = SlackSdkClient(integration_id=integration.id)
+                sdk_client = SlackSdkClient(integration_id=integration.id)
+                text = str(payload["text"]) if payload["text"] is not None else None
                 try:
-                    response = client.chat_postMessage(
-                        blocks=payload.get("blocks"),
-                        text=payload.get("text"),
-                        channel=payload["channel"],
+                    sdk_response = sdk_client.chat_postMessage(
+                        blocks=json_blocks,
+                        text=text,
+                        channel=channel,
                         unfurl_links=False,
                         unfurl_media=False,
-                        thread_ts=payload.get("thread_ts"),
-                        reply_broadcast=payload.get("reply_broadcast"),
+                        thread_ts=thread_ts,
+                        reply_broadcast=reply_broadcast,
                     )
                 except SlackApiError as e:
                     # Record the error code and details from the exception
-                    new_notification_message_object.error_code = e.status_code
+                    new_notification_message_object.error_code = e.response.status_code
                     new_notification_message_object.error_details = {
                         "msg": str(e),
                         "data": e.response.data,
@@ -196,7 +203,13 @@ class SlackNotifyServiceAction(IntegrationEventAction):
                         extra=log_params,
                     )
                 else:
-                    ts = response.data.get("ts")
+                    ts = None
+                    response_data = sdk_response.data
+                    if isinstance(response_data, dict):
+                        ts = response_data.get("ts")
+
+                    self.logger.info("slack.issue-alert.ts", extra={"ts": ts})
+
                     new_notification_message_object.message_identifier = (
                         str(ts) if ts is not None else None
                     )
--- a/src/sentry/integrations/slack/actions/notification.py
+++ b/src/sentry/integrations/slack/actions/notification.py
@@ -5,8 +5,9 @@ from logging import Logger, getLogger
 from typing import Any
 
 import orjson
+from slack_sdk.errors import SlackApiError
 
-from sentry import features
+from sentry import features, options
 from sentry.api.serializers.rest_framework.rule import ACTION_UUID_KEY
 from sentry.eventstore.models import GroupEvent
 from sentry.integrations.repository import get_default_issue_alert_repository
@@ -21,6 +22,7 @@ from sentry.integrations.slack.message_builder.issues import SlackIssuesMessageB
 from sentry.integrations.slack.message_builder.notifications.rule_save_edit import (
     SlackRuleSaveEditMessageBuilder,
 )
+from sentry.integrations.slack.sdk_client import SlackClient as SlackSdkClient
 from sentry.integrations.slack.utils import get_channel_id
 from sentry.models.integrations.integration import Integration
 from sentry.models.rule import Rule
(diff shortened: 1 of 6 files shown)


src/sentry/integrations/slack/sdk_client.py
```diff
@@ -0,0 +1,24 @@
...
+class SlackClient(WebClient):
+    def __init__(self, integration_id: int, base_url: str | None = None):
+        integration = None
+        if SiloMode.get_current_mode() == SiloMode.REGION:
+            integration = integration_service.get_integration(integration_id=integration_id)
+        else:  # control or monolith (local)
+            integration = Integration.objects.filter(id=integration_id).first()
+
+        if integration is None:
+            raise ValueError(f"Integration with id {integration_id} not found")
```
cathteng, line 17: we don't raise any errors if initialization fails in the existing `SlackClient`. all subsequent calls are no-ops, but i think we should fail loudly so we're aware something is broken

tests/sentry/integrations/slack/test_sdk_client.py
```diff
@@ -0,0 +1,37 @@
...
+
+        with pytest.raises(ValueError):
+            SlackClient(integration_id=self.integration.id)
+
+    def test_authorize(self):
+        client = SlackClient(integration_id=self.integration.id)
+
+        with pytest.raises(SlackApiError):
+            # error raised because it's actually trying to POST
+            client.chat_postMessage(channel="#announcements", text="hello")
```
cathteng, line 37: i was having difficulty mocking responses with the Slack SDK client. it seems like we might have to do something like this https://github.com/slackapi/python-slack-sdk/issues/1188#issuecomment-1577327823

src/sentry/integrations/slack/sdk_client.py
```diff
@@ -0,0 +1,24 @@
+from slack_sdk import WebClient
+
+from sentry.models.integrations import Integration
+from sentry.services.hybrid_cloud.integration import integration_service
+from sentry.silo.base import SiloMode
+
+
+class SlackClient(WebClient):
```
cathteng, line 8: we have a function call `track_response_data` in all of our in-house ApiClients, i'm wondering how we might do the same for this SlackClient... duplicate the method signatures, call super(), and then track the response?

src/sentry/integrations/slack/sdk_client.py
```diff
@@ -0,0 +1,24 @@
+from slack_sdk import WebClient
+
+from sentry.models.integrations import Integration
+from sentry.services.hybrid_cloud.integration import integration_service
+from sentry.silo.base import SiloMode
+
+
+class SlackSdkClient(WebClient):
```
cathteng, line 8: we have a function called track_response_data in all of our in-house ApiClients, i'm wondering how we might do the same for this `SlackClient`... duplicate the method signatures, call super(), and then track the response?

src/sentry/integrations/slack/actions/notification.py
```diff
@@ -154,56 +160,108 @@ def send_notification(event: GroupEvent, futures: Sequence[RuleFuture]) -> None:
...
                     # https://api.slack.com/methods/chat.postMessage#arg_thread_ts
                     payload["thread_ts"] = parent_notification_message.message_identifier
+                    thread_ts = parent_notification_message.message_identifier
                     # If this flow is triggered again for the same issue, we want it to be seen in the main channel
                     payload["reply_broadcast"] = True
+                    reply_broadcast = True
+
+            organization = event.group.project.organization
+
+            if organization.id in options.get("slack.sdk-issue-alert"):
```
ykamo001, line 170: might be good to just pull this out into it's own function, what do you think? Will make it easier to maintain and switch as well when the migration is done

src/sentry/integrations/slack/sdk_client.py
ykamo001: Are we sure we need those abstractions or metrics if we aren't already using them?

src/sentry/integrations/slack/sdk_client.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(self, integration_id: int):
```
ykamo001, line 9: nit
```suggestion
    def __init__(self, integration_id: int) -> None:
```

src/sentry/integrations/slack/sdk_client.py
cathteng: aren't we using them in our integrations health dashboard?

src/sentry/integrations/slack/actions/notification.py
cathteng: good idea

src/sentry/integrations/slack/sdk_client.py
cathteng: `track_response_data` currently tracks all responses. we could also just track errors, if the error rates start climbing above a baseline then we know there's a problem, but it's a little harder to tell without the reference to how many total responses there are

src/sentry/integrations/slack/actions/notification.py
```diff
@@ -101,8 +104,10 @@ def send_notification(event: GroupEvent, futures: Sequence[RuleFuture]) -> None:
                 "unfurl_links": False,
                 "unfurl_media": False,
             }
+            json_blocks = None
             if payload_blocks := blocks.get("blocks"):
-                payload["blocks"] = orjson.dumps(payload_blocks).decode()
+                json_blocks = orjson.dumps(payload_blocks).decode()
```
sentaur-athena, line 109: Is there a test covering this line of code? After #inc-779 I get scared whenever I see `orjson.dumps`

src/sentry/integrations/slack/actions/notification.py
```diff
@@ -154,56 +161,108 @@ def send_notification(event: GroupEvent, futures: Sequence[RuleFuture]) -> None:
...
+                    log_params = {
+                        "error": str(e),
+                        "project_id": event.project_id,
+                        "event_id": event.event_id,
+                        "channel_name": self.get_option("channel"),
+                    }
+                    # temporarily log the payload so we can debug message failures
+                    log_params["payload"] = orjson.dumps(payload).decode()
+
+                    self.logger.info(
```
sentaur-athena, line 202: Why is this just info vs error? Also don't we want this in sentry?

src/sentry/integrations/slack/actions/notification.py
```diff
@@ -154,56 +161,108 @@ def send_notification(event: GroupEvent, futures: Sequence[RuleFuture]) -> None:
...
+                    }
+
+                    log_params = {
+                        "error": str(e),
+                        "project_id": event.project_id,
+                        "event_id": event.event_id,
+                        "channel_name": self.get_option("channel"),
+                    }
+                    # temporarily log the payload so we can debug message failures
+                    log_params["payload"] = orjson.dumps(payload).decode()
```
sentaur-athena, line 200: `orjson` warning! 😅 please make sure there's test coverage

src/sentry/integrations/slack/sdk_client.py
sentaur-athena: Yes, we need total responses to set SLOs. I see it has metrics as well as logger which we really don't need for success responses. Saves us a few buck to only log errors and have metrics only for success.

I also agree with moving as a method here to track response. What's your concern for copying it here?

src/sentry/integrations/slack/sdk_client.py
sentaur-athena: +1

sentaur-athena (commented):
I understand some of my comments are for existing code so I hope they're not too annoying 😅

tests/sentry/integrations/slack/test_sdk_client.py
cathteng: ok i got something working here, please yell loudly if this is not the way to do it

src/sentry/integrations/slack/actions/notification.py
cathteng: i added testing to this area of the code in https://github.com/getsentry/sentry/pull/72066, it should be covered now

src/sentry/integrations/slack/sdk_client.py
cathteng: i think to do it i'm going to have to copy the method signature of every method we want to use from the slack_sdk WebClient, call the WebClient function, and track the response after that

like
```
class SlackSdkClient(WebClient):
    def __init__(self, integration_id: int) -> None:
        ....

    def chat_postMessage(...) -> SlackResponse:
        response = super().chat_postMessage(...)
        track_response_data()
        return response
```

and some of these methods have a lot of possible arguments

ykamo001: I think you might be looking for https://stackoverflow.com/questions/11349183/how-to-wrap-every-method-of-a-class

sentaur-athena: Got it. Thank you for the explanation. So basically each method needs a wrapper for logging purpose only. You can use args and kwargs for the arguments since we're only passing them.

One ugly alternative is also to have just one wrapper for all of them. Like `def call_and_track_response(method_to_call, args)`

cathteng: updated with the metaclass wrapping

src/sentry/integrations/slack/actions/notification.py
cathteng: added a test that goes through this flow no problem :)

src/sentry/integrations/slack/sdk_client.py
```diff
@@ -0,0 +1,91 @@
...
+        sample_rate=1.0,
+        tags={"ok": is_ok, "status": code},
+    )
+
+    extra = {
+        "integration": "slack",
+        "status_string": str(code),
+        "error": error,
+    }
+    logger.info("integration.http_response", extra=extra)
```
ykamo001, line 34: are we going to know which method failed though? Like which API call? Any chance we can capture that as well to help us understand which method is failing?

ykamo001 (approved):
Thanks for taking on this work and listening to the feedback/recommendations!!

src/sentry/integrations/slack/sdk_client.py
cathteng: i think i can pass this in the logger if i pass the method in

cathteng: like method.__name__

cathteng: let me do this in a followup, and move the metaclass stuff to integrations.base. currently the only API call is chat.postMessage

PR #72019, opened 2024-06-04
ref: fix types for sentry.utils.snowflake

asottile-sentry:
a bit of a hack -- I originally tried removing the mixin but there are checks for it in the outbox code -- is there a better way to detect if a model is a snowflake model there rather than `isinstance` of this mixin?


markstory:
> is there a better way to detect if a model is a snowflake model there rather than isinstance of this mixin?

We could check for the method that the mixin provides. I agree with your comments that this logic could be done as a module level function.

markstory (approved):

asottile-sentry:
> > is there a better way to detect if a model is a snowflake model there rather than isinstance of this mixin?
> 
> We could check for the method that the mixin provides. I agree with your comments that this logic could be done as a module level function.

ah well my goal was to kill the mixin entirely since that's the part that can't be typed properly :S

markstory:
> ah well my goal was to kill the mixin entirely since that's the part that can't be typed properly

Making the mixin into a module function is something I can take care of.

asottile-sentry:
> > ah well my goal was to kill the mixin entirely since that's the part that can't be typed properly
> 
> Making the mixin into a module function is something I can take care of.

sweet -- will merge this and let you take care of that -- thanks!

PR #72020, opened 2024-06-04
feat(traces): Add graph for matching spans to traces

k-fish:
### Summary
This includes a graph that displays # of cumulative matching spans according to your search conditions.

Other:
- Small changes to not show 'matching' spans when you have no filters, instead will show 'total' and not have 'of'

#### Screenshots
Screenshot 2024-06-04 at 12 38 27 PM

### DIFF
--- a/static/app/views/performance/traces/content.tsx
+++ b/static/app/views/performance/traces/content.tsx
@@ -31,6 +31,7 @@ import {decodeInteger, decodeList, decodeScalar} from 'sentry/utils/queryString'
 import {useLocation} from 'sentry/utils/useLocation';
 import useOrganization from 'sentry/utils/useOrganization';
 import usePageFilters from 'sentry/utils/usePageFilters';
+import * as ModuleLayout from 'sentry/views/performance/moduleLayout';
 
 import {type Field, FIELDS, SORTS} from './data';
 import {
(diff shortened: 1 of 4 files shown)


static/app/views/performance/traces/content.tsx
```diff
@@ -232,6 +243,10 @@ function TraceRow({
 }) {
   const [expanded, setExpanded] = useState<boolean>(defaultExpanded);
   const [highlightedSliceName, _setHighlightedSliceName] = useState('');
+  const location = useLocation();
+  const queries = useMemo(() => {
+    return decodeList(location.query.query);
+  }, [location.query.query]);
```
0Calories, line 249: you could alternatively use `useLocationQuery` which memoizes the query params from the `location` object for you

static/app/views/performance/traces/content.tsx
```diff
@@ -313,6 +332,11 @@ function SpanTable({
   spans: SpanResult<Field>[];
   trace: TraceResult<Field>;
 }) {
+  const location = useLocation();
+  const queries = useMemo(() => {
+    return decodeList(location.query.query);
+  }, [location.query.query]);
+
```
0Calories, line 339: same suggestion as above

static/app/views/performance/traces/tracesChart.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+interface Props {}
+
+export function TracesChart({}: Props) {
+  const location = useLocation();
+  const pageFilters = usePageFilters();
+
+  const queries = useMemo(() => {
+    return decodeList(location.query.query);
+  }, [location.query.query]);
```
0Calories, line 28: same as above

0Calories (approved):
LGTM :shipit:

static/app/views/performance/traces/content.tsx
k-fish: thanks for the tip Ash! I'll prob do that in a follow up

PR #72021, opened 2024-06-04
feat(similarity): Add read only flag to NN endpoint

jangjodi:
Add read_only flag to get_nearest_neighbor endpoint


JoshFerge (approved):
should we make this not a parameter that's passed in, but just default? any situation where we'd want to run embeddings if the endpoint is hit?

jangjodi:
> should we make this not a parameter that's passed in, but just default? any situation where we'd want to run embeddings if the endpoint is hit?

I thought it'd make sense as a parameter since we only want this for the call from the UI - the main use case is still in ingestion/backfill where we'd want to create the record if no neighbor is found

PR #72022, opened 2024-06-04
fix(similarity): use get_primary_hash in backfill

JoshFerge:
- previously we were getting a random hash associated with the group.
- now, we simply use `event.get_primary_hash()`
- removes passing down these hashes through the various functions as its not needed now
- deletes a few tests where we were going through edge cases where we weren't finding a group_id associated with the hash .

### DIFF
--- a/src/sentry/tasks/backfill_seer_grouping_records.py
+++ b/src/sentry/tasks/backfill_seer_grouping_records.py
@@ -19,7 +19,6 @@ from sentry.grouping.grouping_info import get_grouping_info
 from sentry.issues.grouptype import ErrorGroupType
 from sentry.issues.occurrence_consumer import EventLookupError
 from sentry.models.group import Group, GroupStatus
-from sentry.models.grouphash import GroupHash
 from sentry.models.project import Project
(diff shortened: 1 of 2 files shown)


src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -421,18 +421,23 @@ def lookup_group_data_stacktrace_bulk(
...
+                        invalid_event_ids.add(event_id)
+                        continue
+
                     group_data.append(
                         CreateGroupingRecordData(
                             group_id=group_id,
                             project_id=project_id,
                             message=messages[group_id],
-                            hash=hashes[group_id],
+                            hash=event.get_primary_hash(),
```
jangjodi, line 440: can we just use `primary_hash` instead of re-calling `event.get_primary_hash()`?

jangjodi (approved):

src/sentry/tasks/backfill_seer_grouping_records.py
JoshFerge: fixed!

PR #72023, opened 2024-06-04
ref(similarity): Add read_only to UI NN call

jangjodi:
Add read_only flag to nearest neighbor endpoint call from the UI similar issues tab


JoshFerge (approved):
seee my other PR comment -- when would we ever want to actually have it not be read_only?

jangjodi:
> seee my other PR comment -- when would we ever want to actually have it not be read_only?

ah you're right - I thought ingest/backfill called this endpoint, but it calls the seer endpoint; thanks! I'll fix this

jangjodi:
don't need this since it's sent as default in the endpoint [here](https://github.com/getsentry/sentry/pull/72021)

PR #72028, opened 2024-06-04
ref(test): Remove location from context

evanpurkhiser:
This should have never been in here. Location was never in react routers
context.


scttcper (approved):
no more context

PR #72031, opened 2024-06-04
feat(hybridcloud) Add RPC method for sentry-app components

markstory:
The sentryapp component request flows we have in alert rules could benefit from tailored methods that better match the scenario.

Adding this RPC method will help unblock refactoring to streamline and simplify both issue rule creation and alert rule creation.

Refs HC-1024

### DIFF
--- a/src/sentry/services/hybrid_cloud/app/impl.py
+++ b/src/sentry/services/hybrid_cloud/app/impl.py
@@ -165,7 +165,7 @@ class DatabaseBackedAppService(AppService):
             ).items()
         }
 
-    def get_component_context(
+    def get_component_contexts(
         self, *, filter: SentryAppInstallationFilterArgs, component_type: str
     ) -> list[RpcSentryAppComponentContext]:
         install_query = self._FQ.query_many(filter=filter)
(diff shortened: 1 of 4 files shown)


src/sentry/services/hybrid_cloud/app/impl.py
```diff
@@ -163,6 +165,30 @@ def get_related_sentry_app_components(
             ).items()
         }
 
+    def get_component_context(
```
GabeVillalobos, line 168: nit: Shouldn't this method name be plural if we're returning a list of component contexts?

tests/sentry/hybridcloud/test_app.py
```diff
@@ -42,3 +42,60 @@ def test_find_alertable_services() -> None:
     assert len(services) == 2
     assert services[0].title in [app1.name, app2.name]
     assert services[1].title in [app1.name, app2.name]
+
+
+@django_db_all(transaction=True)
+def test_get_component_context() -> None:
```
GabeVillalobos, line 48: Curious about your thoughts on function db_all tests vs all silo tests when validating RPC behavior. I vaguely remember all silo tests being slightly more exhaustive, while also being slower

GabeVillalobos (approved):

tests/sentry/hybridcloud/test_app.py
markstory: Good point about the all_silos. I'll add that.

PR #72033, opened 2024-06-04
ref(messaging): Removing Integration Namespace Imports

iamrajjoshi:
Importing the entire `integration` module is bad practice and it turns out we only do it to access the `IntegrationManager`. I moved the instantiation of the manager back to the original file and import the manager directly.


src/sentry/models/integrations/utils.py
```diff
@@ -15,7 +15,7 @@
 
 
 def get_provider(instance: Integration | RpcIntegration) -> IntegrationProvider:
-    from sentry import integrations
+    from sentry.integrations.manager import default_manager as integrations
 
     return integrations.get(instance.provider)
```
ykamo001, line 20: Probably a sign that this method doesn't belong here, since the import should be happening the other way around. Or, if anything, move this to a util file under manager

ykamo001 (approved):

PR #72034, opened 2024-06-04
feat(traces): Use new slice quantization from backend

k-fish:
### Summary
This uses the quantization from the backend when provided (trace breakdowns), but still uses time for the spans since they don't have quantization.

|Screenshot 2024-06-04 at 2 47 54 PM| |
|--|--|

### DIFF
--- a/static/app/views/performance/traces/fieldRenderers.tsx
+++ b/static/app/views/performance/traces/fieldRenderers.tsx
@@ -128,6 +128,8 @@ export function TraceBreakdownRenderer({
             sliceEnd={breakdown.end}
             sliceDurationReal={breakdown.duration}
             sliceSecondaryName={breakdown.sdkName}
+            sliceNumberStart={breakdown.sliceStart}
+            sliceNumberWidth={breakdown.sliceWidth}
             trace={trace}
             theme={theme}
             offset={index}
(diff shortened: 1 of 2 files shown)


Zylphrex (approved):

static/app/views/performance/traces/fieldRenderers.tsx
```diff
@@ -170,26 +178,34 @@ export function SpanBreakdownSliceRenderer({
...
+  sliceNumberWidth?: number;
 }) {
-  const traceSliceSize = (trace.end - trace.start) / BREAKDOWN_NUM_SLICES;
-  const traceDuration = BREAKDOWN_NUM_SLICES * traceSliceSize;
+  const traceDuration = trace.end - trace.start;
 
   const sliceDuration = sliceEnd - sliceStart;
 
-  if (sliceDuration <= 0) {
+  if (sliceNumberWidth === 0 && sliceNumberWidth === undefined && sliceDuration <= 0) {
```
mjq, line 188: `sliceNumberWidth === 0 && sliceNumberWidth === undefined`

Isn't that always false?

mjq (commented):
Oh looks like this already got merged, flagging a weird conditional just in case still

static/app/views/performance/traces/fieldRenderers.tsx
k-fish: This should just be entirely removed anyway actually, we have checks on the backend now, don't need this defensive code anymore.

PR #72041, opened 2024-06-04
test(js): Remove all usages of routerContext

evanpurkhiser:
This is replaced with `router`. We already had this and there were
simply two ways of doing things.


scttcper (approved):
router

PR #72043, opened 2024-06-04
bug(slack): add missing stop gap for option

### LINKED ISSUE #71967: Slack alert shows up as thread after disabling the thread option
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

1. have a Slack integration, with enabled Slack threads on Issue Alerts
2. Send in the same error twice, the second will appear as a thread
3. Go to integration settings and disable Slack threads for Issue Alerts
4. Send in the same error again, it will still appear as a thread in Slack

### Expected Result

after disabling the feature, it the notification should not appear as a thread in Slack

### Actual Result

it still appears in a thread

### Product Area

Settings - Integrations

### Link

_No response_

### DSN

_No response_

### Version

_No response_

ykamo001:
Threads should not happen for issues if the option is off

Resolves: https://github.com/getsentry/sentry/issues/71967

### DIFF
--- a/src/sentry/integrations/slack/actions/notification.py
+++ b/src/sentry/integrations/slack/actions/notification.py
@@ -8,6 +8,7 @@ import orjson
 
 from sentry import features
 from sentry.api.serializers.rest_framework.rule import ACTION_UUID_KEY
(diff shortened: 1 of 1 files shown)


cathteng (approved):
how do we write tests for all of these 😭

PR #72045, opened 2024-06-04
fix(contexts): Fix casing for known context

leeandher:
We wanted to preserve casing for unknown context from the sdk in https://github.com/getsentry/sentry/pull/72013, but there are a few non-context context items that weren't accounted for. 

**Before**
image

**After**
image

### DIFF
--- a/static/app/components/events/contexts/utils.tsx
+++ b/static/app/components/events/contexts/utils.tsx
@@ -317,6 +317,16 @@ export function getContextTitle({
       return t('App');
     case 'device':
       return t('Device');
+    case 'browser':
+      return t('Browser');
+    case 'profile':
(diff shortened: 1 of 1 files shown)


static/app/components/events/contexts/utils.tsx
```diff
@@ -328,9 +338,9 @@ export function getContextTitle({
     case 'trace':
       return t('Trace Details');
     case 'otel':
-      return t('OpenTelemetry');
+      return 'OpenTelemetry';
     case 'unity':
-      return t('Unity');
+      return 'Unity';
```
leeandher, line 343: These refer to specific companies and shouldn't be translated

malwilley (approved):

PR #72049, opened 2024-06-04
feat(flagpole): Adds convenience methods for yaml and json conversion

GabeVillalobos:
Adds conversion methods to/from json and yaml in preparation for our feature flag backfill.

The yaml helper methods will also be useful when creating a CLI util to generate feature flag yaml.

Unfortunately, the conversion from a feature to a dictionary is a bit manual, due to the way that we compact our feature yaml.


markstory (approved):
Looks good to me.

PR #72056, opened 2024-06-04
fix(spans): Handle time column for grahing indexed spans

wmak:
- This was defaulted to 1 hour for early prototyping but was never updated after that


k-fish (approved):
TSIQB 👏

PR #72059, opened 2024-06-04
ref(flagpole): Copies sentry flagpole context into sentry src directory

GabeVillalobos:
Copies the Sentry context builder factory util into the Sentry `src` directory, in an attempt to further decouple flagpole from our main sentry codebase.

There will be 2 follow-up PRs to this:
1. Update getsentry import to use the sentry version of the context builder
2. Remove the context builder code from the top-level flagpole directory

After these PRs are completed, the flagpole directory will no longer have _any_ Sentry imports, meaning it can be safely split into a separate module or dependency in the future.


src/sentry/features/flagpole_utils/sentry_flagpole_context.py
```diff
@@ -0,0 +1,102 @@
+from dataclasses import dataclass
```
markstory, line 1: Nit: You could call the module `sentry.features.flagpole`, or even have this in `sentry.features.flagpole_context` :shrug:

markstory (commented):
If you wanted to have more of a 'move' change. You could have the current module import from the new one and re-export the same symbol names with `__all__`.

markstory (approved):

GabeVillalobos:
> If you wanted to have more of a 'move' change. You could have the current module import from the new one and re-export the same symbol names with `__all__`.

Just gave this a try, and it'd still require a GetSentry change to support making this more work unfortunately 😞

markstory:
> Just gave this a try, and it'd still require a GetSentry change to support making this more work unfortunately 😞

Dang, double files it is then.

markstory (approved):

PR #72066, opened 2024-06-04
chore(tests): test SlackNotifyServiceAction

cathteng:
Previously there was no test coverage for SlackNotifyServiceAction (Slack issue alerts). Adding some which is important in light of the new threads feature.


cathteng:
i don't think this is complete at all -- the `send_notification` function in the `after` function is doing a LOT

ykamo001 (approved):
Huge improvements!

PR #72071, opened 2024-06-04
chore(issue-platform): Remove the `organizations:issue-platform` flag from the backend

snigdhas:
Removing the `organizations:issue-platform` feature flag now that it's been released in self-hosted (https://github.com/getsentry/self-hosted/pull/2309).

### DIFF
--- a/src/sentry/monitors/logic/mark_failed.py
+++ b/src/sentry/monitors/logic/mark_failed.py
@@ -11,9 +11,7 @@ from django.db.models import Q
 from django.utils.text import get_text_list
 from django.utils.translation import gettext_lazy as _
 
-from sentry import features
 from sentry.issues.grouptype import MonitorIncidentType
-from sentry.models.organization import Organization
 from sentry.monitors.models import (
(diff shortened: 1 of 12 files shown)


aliu39 (approved):
I think(?) you can also remove the manager.add line from temporary.py, in the same PR

src/sentry/monitors/logic/mark_failed.py
```diff
@@ -196,24 +185,6 @@ def mark_failed_threshold(
...
-
-    monitor_env = failed_checkin.monitor_environment
-
-    monitor_env.update(status=MonitorStatus.ERROR)
-
-    # Do not create event if monitor or monitor environment is muted
-    if monitor_env.monitor.is_muted or monitor_env.is_muted:
-        return True
-
-    create_legacy_event(failed_checkin)
```
wedamija, line 210: Could you also remove this function and any other unused functions that it might call?

wedamija (approved):

wedamija (approved):

PR #72075, opened 2024-06-04
ref(utils): circuit breaker as context manager PoC

vartec:
PoC context manager version of this, which might be nicer to use.

### DIFF
--- a/src/sentry/utils/circuit_breaker.py
+++ b/src/sentry/utils/circuit_breaker.py
@@ -1,4 +1,5 @@
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
+from contextlib import contextmanager
 from typing import ParamSpec, TypedDict, TypeVar
 
 from django.core.cache import cache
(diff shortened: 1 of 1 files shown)


lobsterkatie:
Yup, I can see how that would work, and as I mentioned in [the PR which added `with_circult_breaker`](https://github.com/getsentry/sentry/pull/71627), I definitely considered going the context manager route, but eventually decided against it:

> I was deeply conflicted about whether to write this as a wrapper with a callback or as a context manager. I ended up with the wrapper approach because my fear was the that natural tendency with a context manager would be to use it like this:
> 
> ```python
> with circuit_breaker("some_key", config):
> 	try:
> 		do_something()
> 	except Exception:
> 		handle_exception()
> ```
> 
> The problem with that is that the circuit breaker has to be _inside_ the `try` block - otherwise, the errors it's supposed to be tracking will already have been handled before bubbling up where it could see them. So instead I went with a function which could be wrapped around `do_something`, because that way - short of pulling the whole `try-except` into `do_something`'s innards - there's no way to make the above mistake.

So IDK - is the context manager easier to use? Yeah, maybe a little. But I can't shake the feeling that it's also easier to use _wrong_. (I know, there's a docstring, but why set people up to make a mistake?) 

Actually, one thing which might be neat is if we could do it as a decorator on the helper itself. I can't decide if that would violate separation of concerns, though. (Also, maybe it wouldn't even work because you'd always have to use the same cache key? I'd have to mess around with it.)

Anyway, thoughts?

vartec:
> So IDK - is the context manager easier to use? Yeah, maybe a little. But I can't shake the feeling that it's also easier to use wrong. 

Hmm.. OK, fair enough. I guess by forcing people to use a function it's less likely.

PR #72076, opened 2024-06-04
feat(query-builder): Return focus to input and call onSearch after clicking clear button

malwilley:
Returns focus to input so the user can immediately start typing in the next query

### DIFF
--- a/static/app/components/searchQueryBuilder/index.spec.tsx
+++ b/static/app/components/searchQueryBuilder/index.spec.tsx
@@ -137,22 +137,27 @@ describe('SearchQueryBuilder', function () {
   describe('actions', function () {
     it('can clear the query', async function () {
       const mockOnChange = jest.fn();
+      const mockOnSearch = jest.fn();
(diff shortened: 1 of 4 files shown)


static/app/components/searchQueryBuilder/index.spec.tsx
```diff
@@ -137,22 +137,27 @@ describe('SearchQueryBuilder', function () {
...
           onChange={mockOnChange}
+          onSearch={mockOnSearch}
         />
       );
       userEvent.click(screen.getByRole('button', {name: 'Clear search query'}));
 
       await waitFor(() => {
         expect(mockOnChange).toHaveBeenCalledWith('');
+        expect(mockOnSearch).toHaveBeenCalledWith('');
       });
```
scttcper, line 154: that best practice blog said to have one thing in a waitFor but i don't really see the harm if they're pretty simple

scttcper (approved):

PR #72096, opened 2024-06-05
feat(remote-config): Cache remote-config files in Redis

cmanallen:
Performance and cost optimization.  Removes the need to reach out to GCS (pretty much ever).  This is a temporary caching facade while we build out the real version which will cache in the PoPs.

### DIFF
--- a/src/sentry/remote_config/storage.py
+++ b/src/sentry/remote_config/storage.py
@@ -109,7 +109,9 @@ class ConfigurationBackend:
             return (cache_result, "cache")
 
         storage_result = self.storage.get()
-        self.cache.set(storage_result)
+        if storage_result:
+            self.cache.set(storage_result)
+
         return (storage_result, "store")
 
     def set(self, value: StorageFormat) -> None:
--- a/src/sentry/remote_config/storage.py
+++ b/src/sentry/remote_config/storage.py
@@ -113,8 +113,8 @@ class ConfigurationBackend:
         return storage_result
 
     def set(self, value: StorageFormat) -> None:
-        self.cache.set(value)
         self.storage.set(value)
+        self.cache.set(value)
 
     def pop(self) -> None:
         self.cache.pop()
--- a/src/sentry/remote_config/storage.py
+++ b/src/sentry/remote_config/storage.py
@@ -2,9 +2,10 @@ from io import BytesIO
 from typing import TypedDict
 
 from sentry import options
+from sentry.cache import default_cache
 from sentry.models.files.utils import get_storage
 from sentry.models.project import Project
-from sentry.utils import json
+from sentry.utils import json, metrics
 
 JSONValue = str | int | float | bool | None | list["JSONValue"] | dict[str, "JSONValue"]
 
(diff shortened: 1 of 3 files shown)


anonrig (commented):
Can you add tests that run on actual redis cluster?

cmanallen:
@anonrig How and why?  Presumably the cache facade has that coverage already.

src/sentry/remote_config/storage.py
```diff
@@ -105,5 +95,65 @@ def _make_storage_config(self) -> dict | None:
...
+    def set(self, value: StorageFormat) -> None:
+        self.cache.set(value)
+        self.storage.set(value)
+
+    def pop(self) -> None:
+        self.cache.pop()
+        self.storage.pop()
+
+
+class APIBackendDecorator:
```
aliu3ntry, line 124: Don't think this is technically a decorator but could be confusing / someone might try to use it that way.. can we just call it APIBackend?

src/sentry/remote_config/storage.py
```diff
@@ -30,44 +31,33 @@ class APIFormat(TypedDict):
...
-        return self._deserialize(result)
-
-    def set(self, value: APIFormat) -> None:
-        self.driver.set(self._serialize(value))
+    def get(self) -> StorageFormat | None:
+        cache_result = default_cache.get(self.key)
 
-    def pop(self) -> None:
-        self.driver.pop()
+        if cache_result is None:
```
aliu3ntry, line 41: Could we add a test case for cache miss and pop miss?

aliu39 (changes requested):

src/sentry/remote_config/storage.py
cmanallen: Its a decorator.  See: https://en.wikipedia.org/wiki/Decorator_pattern

aliu3ntry: Ah gotcha, I was thinking about python decorators with the @ syntax.

src/sentry/remote_config/storage.py
```diff
@@ -105,5 +95,65 @@ def _make_storage_config(self) -> dict | None:
             return None
 
 
-def make_storage(key: Project):
-    return StorageBackend(key)
+class ConfigurationBackend:
```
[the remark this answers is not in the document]
cmanallen: Its an abstraction to manage fetching the configs.  I have no idea how long this will live.  Its not intended to be permanent.  We can document it later.

tests/sentry/remote_config/endpoints/test_configuration.py
```diff
@@ -161,12 +184,28 @@ def test_delete_configuration(self):
...
-        assert self.storage.get() is not None
+        assert self.storage.get()[0] is not None
 
         with self.feature(REMOTE_CONFIG_FEATURES):
             response = self.client.delete(self.url)
         assert response.status_code == 204
-        assert self.storage.get() is None
+        assert self.storage.get()[0] is None
+
+    def test_delete_configuration_no_cache(self):
```
aliu3ntry, line 194: Thanks for adding these!

src/sentry/remote_config/storage.py
```diff
@@ -105,5 +95,65 @@ def _make_storage_config(self) -> dict | None:
...
+        return {
+            "features": result["features"],
+            "options": result["options"],
+        }
+
+    def _serialize(self, result: APIFormat) -> StorageFormat:
+        return {
+            "features": result["features"],
+            "options": result["options"],
+            "version": 1,
```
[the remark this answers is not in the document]
aliu3ntry: any thoughts on this?

aliu39 (approved):

src/sentry/remote_config/storage.py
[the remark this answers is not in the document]
cmanallen: If it were used in more than one place I would pull it out.  For now its fine where it is.  I plan to change this substantially in the next few weeks so I'm not too worried about docstrings and constants.

PR #72099, opened 2024-06-05
feat(profiling): add endpoint to fetch a profile given a profiler_id

viglia:
This will take a `profiler_id` and a `time range` as input and will, in turn, fetch the related chunks and merge them into an appropriate profile to be returned.

In this first iteration, if the time-range specified contains too many chunks, we will limit the profile to the latest `N` chunks.
In later iterations, if necessary, we might provide a pagination mechanism so that we'll be able to scroll the whole profile directly from the _flamechart_ page itself.

```bash
curl --location 'http://127.0.0.1:8000/api/0/organizations/sentry/profiling/chunks/?project=<PROJECT_ID>&profiler_id=<PROFILER_ID>&statsPeriod=90d&utc=false'
```

Depends on https://github.com/getsentry/vroom/pull/470


src/sentry/profiles/profile_chunks.py
```diff
@@ -0,0 +1,41 @@
...
+            Column("chunk_id"),
+        ],
+        where=[
+            Condition(Column("start_timestamp"), Op.GTE, params.get("start")),
+            Condition(Column("end_timestamp"), Op.LT, params.get("end")),
+            Condition(Column("project_id"), Op.EQ, project_id),
+            Condition(Column("profiler_id"), Op.EQ, profiler_id),
+        ],
+        orderby=[OrderBy(Column("start_timestamp"), Direction.DESC)],
+    ).set_limit(max_chunks)
```
Zylphrex, line 28: This feels like it should be a paginated API. So for a given project id, profiler id, start/end timestamp, the frontend should be able to request the first page, then load more as needed not restricted to the first 50 results.

viglia: @Zylphrex yes, indeed in the PR description I mentioned this:

> In later iterations, if necessary, we might provide a pagination mechanism so that we'll be able to scroll the whole profile directly from the flamechart page itself.

I had discussed this offline with @phacops and we decided to keep things simple and start with a _fixed size_ solution for the first iteration, but providing pagination is definitely in the plans.

src/sentry/profiles/profile_chunks.py
```diff
@@ -0,0 +1,41 @@
...
+) -> list[dict[str, str]]:
+    max_chunks = options.get("profiling.continuous-profiling.chunks-set.size")
+    query = Query(
+        match=Storage(StorageKey.ProfileChunks.value),
+        select=[
+            Column("chunk_id"),
+        ],
+        where=[
+            Condition(Column("start_timestamp"), Op.GTE, params.get("start")),
+            Condition(Column("end_timestamp"), Op.LT, params.get("end")),
```
Zylphrex, line 23: This might cause some alignment issues when rendering. For example, suppose we have profile chunks for the following time ranges
- 0 to 100
- 100 to 200

And the request asks for 50 to 150. Under in this situation, this query would return no results because
- for 0 to 100, the start timestamp 0 is less than the param start
- for 100 to 200, the end timestamp 200 is greater than the param end

I think the condition you really want here is

```suggestion
            Condition(Column("end_timestamp"), Op.GTE, params.get("start")),
            Condition(Column("start_timestamp"), Op.LT, params.get("end")),
```

viglia: You're correct.

Good catch! thanks!

Zylphrex (approved):

PR #72103, opened 2024-06-05
Set linguist language to Bash for `.envrc`

spenserblack:
Basically, this is a nitpick to use the proper language override instead of a fake shebang.

AFAICT this shebang was added to trick GitHub to apply shell syntax highlighting to the `.envrc` file. Clever! But GitHub uses [linguist](https://github.com/github-linguist/linguist), and using the `linguist-language` attribute is the "official" way to override the detected language and syntax highlighting. I stumbled on this file while searching GitHub for `.envrc` usage examples, and figured that it would be slightly better to not use a fake shebang. Apologies for a PR that is bikeshedding a bit :sweat_smile:

You might also want to look into the `linguist-generated` attribute for your generated files, but Linguist should be smart enough to detect which files are generated on its own. At least for minified files.

### Legal Boilerplate

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,4 +1,3 @@
-#!/not/executable/bash
 # This is the .envrc for sentry, for use with direnv.
 # It's responsible for enforcing a standard dev environment by checking as much state as possible, and either performing
 # initialization (e.g. activating the venv) or giving recommendations on how to reach the desired state.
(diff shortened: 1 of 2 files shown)


spenserblack:
Ah, that makes sense. Depending on which text editors are used, perhaps you'd want to use modelines instead? Sorry for being really nitpicky, I'll stop now :sweat_smile:

PR #72105, opened 2024-06-05
Reapply "ref: upgrade sentry to python 3.12 (#65075)"

asottile-sentry:
This reverts commit 889bf3e4e05a8a0f9346d931375ed4b91e59c7c0.

now with upgraded `pyuwsgi` \o/


joshuarli (approved):

asottile-sentry:
cautiously reverting as there was an uptick in rpc call times afterwards that could be related

PR #72106, opened 2024-06-05
fix(metrics): Duplicate metrics name detection performance

ArthurKnaus:
Detect duplicate metric names in `O(n)` instead of `O(n^2)`.

Part of https://github.com/getsentry/sentry/issues/71979


matejminar (approved):
Thanks for fixing this! 🚀

PR #72113, opened 2024-06-05
chore(features) Move feature defaults for temporary features

markstory:
All feature defaults are now defined when features are registered. Adding and removing a feature no longer requires changes to `server.py`

Refs HC-1152

### DIFF
--- a/src/sentry/conf/server.py
+++ b/src/sentry/conf/server.py
@@ -1452,66 +1452,8 @@ SENTRY_EARLY_FEATURES = {
 # `features.manager.add()`. Defining feature defaults here is deprecated.
 # If you must add a feature here, please maintain alphabetical ordering
 SENTRY_FEATURES: dict[str, bool | None] = {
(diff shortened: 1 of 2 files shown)


src/sentry/conf/server.py
```diff
@@ -1452,66 +1452,8 @@ def custom_parameter_sort(parameter: dict) -> tuple[str, int]:
 # `features.manager.add()`. Defining feature defaults here is deprecated.
 # If you must add a feature here, please maintain alphabetical ordering
 SENTRY_FEATURES: dict[str, bool | None] = {
-    # potentially unused.
-    "organizations:api-keys": False,
```
aliu3ntry, line 1456: Could you add an explicit default=False for this one too?

markstory: Sure, but I don't see this feature flag being used anywhere in sentry or getsentry.

aliu39 (approved):

PR #72118, opened 2024-06-05
feat(releases): Increase max commit author email length

### LINKED ISSUE #2072: Author_email - Ensure this field has no more than 75 characters.
### CLI Version

2.32.1

### Operating System and Architecture

- [X] macOS (arm64)
- [X] macOS (x86_64)
- [X] Linux (i686)
- [X] Linux (x86_64)
- [X] Linux (armv7)
- [X] Linux (aarch64)
- [X] Windows (i686)
- [X] Windows (x86_64)

### Operating System Version

macOS 14.5

### Link to reproduction repository

_No response_

### CLI Command

sentry-cli releases set-commits XXX --local -p YYY --ignore-missing

### Exact Reproduction Steps

1. Create a commit with commit author email longer than 75 chars.
2. Run `sentry-cli releases set-commits XXX --local -p YYY --ignore-missing`

### Expected Results

Remove `author_email` field if it does not require sentry-api expectations (it is [optional](https://github.com/getsentry/sentry/blob/master/src/sentry/api/serializers/rest_framework/commit.py#L28)).

### Actual Results

It doesn't send commits to sentry-api, because we get 400.

I wonder if we could remove `author_email` field if it does not require sentry-api expectations.

If not, can you help me solve the issue on my side? I've changed that email for future commits, but they will be in the history forever. I don't know how to push only newer commits.

### Logs

```
  DEBUG   2024-05-29 08:04:27.462029 +02:00 body: {"commits":{"author_email":["Ensure this field has no more than 75 characters."]}}
error: API request failed
  caused by: sentry reported an error: request failure (http status: 400)
  Object {"commits": Object {"author_email": Array [String("Ensure this field has no more than 75 characters.")]}}
```

### LINKED ISSUE #72114: Increase commit `author_email` character limit
### Problem Statement

We [limit](https://github.com/getsentry/sentry/blob/b48468708281357260a647c1970cb2cd1091d18d/src/sentry/api/serializers/rest_framework/commit.py#L26) commits' `author_email` field to a 75-character maximum. This limit [prevents](https://github.com/getsentry/sentry-cli/issues/2072) users from using the Sentry CLI to create releases containing any commits whose author's email contains more than 75 characters. 

### Solution Brainstorm

Increase the character limit to 254, which is the [maximum length](https://stackoverflow.com/a/574698) for a valid email address.

**Implementing this suggestion would resolve [getsentry/sentry-cli#2072](https://github.com/getsentry/sentry-cli/issues/2072).**

### Product Area

Other

szokeasaurusrex:
Increase the maximum commit author email length to 254 characters, which is the maximum length of an email address per [RFC 3696 (IED 169)](https://www.rfc-editor.org/errata/eid1690).

Fixes #72114
Fixes https://github.com/getsentry/sentry-cli/issues/2072


szokeasaurusrex:
I verified locally that commits with author email addresses of up to 254 characters can now be associated with Sentry releases.

Is any unit testing needed for this change?

asottile-sentry:
I don't think this is that simple -- this is just the serialization layer -- there's several other places where it is truncated to 75 and/or 200 characters (including the model itself) -- I suspect those all need to be updated as well

szokeasaurusrex:
@asottile-sentry, this change, as is, should be sufficient to resolve https://github.com/getsentry/sentry-cli/issues/2072. With the serializer limit at 75, attempting a release with commits authored by an author with an email address longer than 75 characters failed with a 400 status; now, they succeed, but the email address is [silently truncated to 75 characters](https://github.com/getsentry/sentry/blob/85d2b155fd7e28846c743d2f34bf54b5641d9b8e/src/sentry/models/release.py#L738) before being stored.

So, if we are fine with just storing the truncated email, I believe this change is sufficient. Otherwise, I can increase the truncation limit and the [`CommitAuthor` model limit](https://github.com/getsentry/sentry/blob/85d2b155fd7e28846c743d2f34bf54b5641d9b8e/src/sentry/models/commitauthor.py#L31).

asottile-sentry:
up to you -- I have no say in this and I'm not sure who owns this 🤔

szokeasaurusrex:
Yes, it is hard to tell who owns this file @asottile-sentry. Looks like you made the two [most recent changes](https://github.com/getsentry/sentry/commits/master/src/sentry/api/serializers/rest_framework/commit.py), but it is also not a super frequently edited file.

If we were to increase the CommitAuthor email field, this would require a database migration which is owned by @getsentry/owners-migrations – perhaps someone from here can comment on whether we should expand the field in the database, as well?

asottile-sentry:
> Yes, it is hard to tell who owns this file @asottile-sentry. Looks like you made the two [most recent changes](https://github.com/getsentry/sentry/commits/master/src/sentry/api/serializers/rest_framework/commit.py), but it is also not a super frequently edited file.
> 
> If we were to increase the CommitAuthor email field, this would require a database migration which is owned by @getsentry/owners-migrations – perhaps someone from here can comment on whether we should expand the field in the database, as well?

heh well -- could say the same for basically every file in sentry though I just added / fixed some type annotations which I'm trying to do everywhere

markstory:
Accepting longer email addresses only to truncate them before saving commit data seems silly to me. Currently the `CommitAuthor.email` field has a limit of 200 chars. It was lengthened in f24cfb32705f.

Aligning the truncation and serializer to 200 chars to match the schema seems like a good solution. While we could extend the email column, and each other layer to 254 chars, extending the schema has some risk. Because there is an index on `organization_id,email` postgres can invalidate indexes when column length is increased which can result in table scans and higher db load.

szokeasaurusrex:
@markstory I decreased the limit to 200. Let's hope that that is enough for most use cases

markstory (approved):
Looks good to me. Aligning the serializer with the schema length will be safe.

szokeasaurusrex:
@markstory The [`openapi-diff` check](https://github.com/getsentry/sentry/actions/runs/10182244636/job/28164296219?pr=72118) is failing, and I am unsure how to fix it. 

Do we have any developer docs on this, or could you provide some pointers?

PR #72119, opened 2024-06-05
feat(remote-config): Update protocol to include rule and rollout definitions

cmanallen:
This is a proposal.  Subject to change/rejection.


src/sentry/remote_config/docs/protocol.md
```diff
@@ -22,10 +22,58 @@ Retrieve a project's configuration.
...
+
+When evaluating a set of rules and varaints each rule is a logical `AND` while each variant is a logical `OR`. Failure to evaluate one variant does not prevent another variant from successfully evaluating. Failure to evaluate one rule does invalidate the remaining rules and the evaluation proceedure can eagerly terminate.
+
+**Rollout Object**
+
+| Field      | Type   | Description                                                 |
+| ---------- | ------ | ----------------------------------------------------------- |
+| percentage | number | The percentage of requests which should evaluate to true.   |
+| seed       | string | A SHA1 hash used to seed a random number generator.         |
+| target     | string | The name of a property contained within the context object. |
```
ryan953, line 49: I'm thinking it might make more sense to put `seed` and `target` higher up in the schema. If each variant had it's own pair then it might get weird math wise. Like, if you have no conditions on any variant... and all percentages add up to 100% then you still might have users who get none of the variants because of hashing.

src/sentry/remote_config/docs/protocol.md
```diff
@@ -79,11 +127,34 @@ If the server's ETag does not match the request's a 200 response is returned.
       "features": [
         {
           "key": "hello",
-          "value": "world"
+          "value": "world",
```
ryan953, line 130: I'm assuming that `"world"` would be returned if either:
A) the conditions below do not match. ie: request contains `?context={region: Asia, now: 0}`
B) the user_id hashes to the wrong side of the rollout, so 50% of the time europeans will see "world"

src/sentry/remote_config/docs/protocol.md
```diff
@@ -22,10 +22,58 @@ Retrieve a project's configuration.
...
+| -------- | ------------------------------------------- | ----------------------------------------------------------------------- |
+| operator | Operator                                    | A conditional operator that evaluates the value against the target.     |
+| property | string                                      | A context object property name which contains the value of the operand. |
+| value    | union[string, int, bool, null, array[self]] | The value being compared against.                                       |
+
+Rules contain evaluation and extraction instructions. The context object provided must contain the key defined on the `property` attribute. If the SDK can not find the `property` within the context object the evaluation fails and the variant is skipped. The `operator` defines how the context value is compared against the literal provided by the configuration on the `value` attribute. If the value provided in the context object can not be compared against the value provided in the configuration the evaluation fails and the variant is skipped.
```
ryan953, line 61: IMO we should also write down that we're comparing against json values, so `null` is acceptable, `None` (python) and `undefined` (js) are not. 
Similarly, when comparing values we should be specific about equality. Falsey values and type coercion come to mind. Depending on type coercion, a missing properly might match against an expected falsey value and pass the condition, or it might never match and therefore fail the condition. 
But things like `?context={age: false}` with `condition=['age', '<', 21]` need to be consistent in each SDK. Certainly if open feature handles it then we're done here and I'll go read about it!

src/sentry/remote_config/docs/protocol.md
cmanallen: Rollout is not a distribution mechanism so there's no concern with the percentages adding up to 100%.  Percentages can sum up to any values greater than 0.

src/sentry/remote_config/docs/protocol.md
cmanallen: Yes.

src/sentry/remote_config/docs/protocol.md
ryan953: So order is 1. rollout first, 2. evaluate conditions... if both true return `"value"` in the conditional, else return the top-level `"value"`?

Or another way, it's something like "percent of the time we should dive in and evaluate if the conditions are true or not"
sounds good.

src/sentry/remote_config/docs/protocol.md
```diff
@@ -22,17 +22,66 @@ Retrieve a project's configuration.
...
+
+**Variant Object**
+
+| Field   | Type        | Description                                                                               |
+| ------- | ----------- | ----------------------------------------------------------------------------------------- |
+| rollout | Rollout     | An object which contains instructions for rolling out a variant.                          |
+| rules   | array[Rule] | An array of rules which all must evaluate to true in order to return the variant's value. |
+| value   | any         | The value of the feature if the variant is successfully evaluated.                        |
+
+When evaluating a set of rules and varaints each rule is a logical `AND` while each variant is a logical `OR`. Failure to evaluate one variant does not prevent another variant from successfully evaluating. Failure to evaluate one rule does invalidate the remaining rules and the evaluation proceedure can eagerly terminate.
```
ryan953, line 41: variant order is important. sdks need to respect it.

src/sentry/remote_config/docs/protocol.md
```diff
@@ -22,17 +22,66 @@ Retrieve a project's configuration.
...
+**Rollout Object**
+
+| Field         | Type                | Description                                                 |
+| ------------- | ------------------- | ----------------------------------------------------------- |
+| percentage    | number              | The percentage of requests which should evaluate to true.   |
+| sticky        | union[object, null] | An object which controls the stickyness of the rollout.     |
+| sticky.seed   | string              | A SHA1 hash used to seed a random number generator.         |
+| sticky.target | string              | The name of a property contained within the context object. |
+
+The rollout cohort a session is bucketed into is deterministic some long as the `target` value is static. The `target` value controls the bucket a session is placed into. The `seed` value provides randomization _between features_ such that if a session is successfully opted into feature `A` they will not necesssarily be opted into feature `B` even if those features share the same rollout percentage. The `percentage` value controls the number of buckets which evaluate to `true`.
```
ryan953, line 52:
```suggestion
The rollout cohort a session is bucketed into is deterministic so long as the `target` value is static. The `target` value controls the bucket a session is placed into. The `seed` value provides randomization _between features_ such that if a session is successfully opted into feature `A` they will not necesssarily be opted into feature `B` even if those features share the same rollout percentage. The `percentage` value controls the number of buckets which evaluate to `true`.
```

src/sentry/remote_config/docs/protocol.md
cmanallen: Else check the next variant; if no variants match return the top level value.

> "percent of the time we should dive in and evaluate if the conditions are true or not"

Exactly.

src/sentry/remote_config/docs/protocol.md
cmanallen: Yes this needs to be called out explicitly in the blueprint 👍

src/sentry/remote_config/docs/protocol.md
```diff
@@ -22,17 +22,66 @@ Retrieve a project's configuration.
...
 
-| Field | Type   | Description                        |
-| ----- | ------ | ---------------------------------- |
-| key   | string | The name used to lookup a feature. |
-| value | any    | A JSON value.                      |
+| Field    | Type           | Description                                                                     |
+| -------- | -------------- | ------------------------------------------------------------------------------- |
+| key      | string         | The name used to lookup a feature.                                              |
+| value    | any            | A JSON value.                                                                   |
+| variants | array[Variant] | Features can have any number of possible variants which alter the return value. |
```
ryan953, line 29:
```suggestion
| variants | union[array[Variant], null] | Features can have any number of possible variants which alter the return value. |
```

src/sentry/remote_config/docs/protocol.md
```diff
@@ -22,17 +22,66 @@ Retrieve a project's configuration.
...
+
+Features may specify zero or more variants which may alter the evaluation result.
+
+**Variant Object**
+
+| Field   | Type        | Description                                                                               |
+| ------- | ----------- | ----------------------------------------------------------------------------------------- |
+| rollout | Rollout     | An object which contains instructions for rolling out a variant.                          |
+| rules   | array[Rule] | An array of rules which all must evaluate to true in order to return the variant's value. |
```
ryan953, line 38:
```suggestion
| rollout | union[Rollout, null]     | An object which contains instructions for rolling out a variant.                          |
| rules   | union[array[Rule], null] | An array of rules which all must evaluate to true in order to return the variant's value. |
```

src/sentry/remote_config/docs/protocol.md
cmanallen: Let's just return an empty array.  Why bother having a conditional check for null if we can avoid it?

cmanallen: Same concern for rules and rollout.

src/sentry/remote_config/docs/protocol.md
```diff
@@ -22,17 +22,66 @@ Retrieve a project's configuration.
...
+| value   | any         | The value of the feature if the variant is successfully evaluated.                        |
+
+When evaluating a set of rules and varaints each rule is a logical `AND` while each variant is a logical `OR`. Failure to evaluate one variant does not prevent another variant from successfully evaluating. Failure to evaluate one rule does invalidate the remaining rules and the evaluation proceedure can eagerly terminate.
+
+**Rollout Object**
+
+| Field         | Type                | Description                                                 |
+| ------------- | ------------------- | ----------------------------------------------------------- |
+| percentage    | number              | The percentage of requests which should evaluate to true.   |
```
ryan953, line 47: idk, something like this:

```suggestion
| percentage    | number              | The percentage of requests which should evaluate to true. Valid values are `0.0` to `1.0`, increments of `0.001`  |
```

src/sentry/remote_config/docs/protocol.md
```diff
@@ -22,17 +22,68 @@ Retrieve a project's configuration.
...
+| variants | array[Variant] | Features can have any number of possible variants which alter the return value. |
+
+Features may specify zero or more variants which may alter the evaluation result.
+
+**Variant Object**
+
+| Field   | Type                 | Description                                                                               |
+| ------- | -------------------- | ----------------------------------------------------------------------------------------- |
+| rollout | union[Rollout, null] | An object which contains instructions for rolling out a variant.                          |
+| rules   | array[Rule]          | An array of rules which all must evaluate to true in order to return the variant's value. |
```
ryan953, line 38:
```suggestion
| rules   | union[array[Rule], null]          | An array of rules which all must evaluate to true in order to return the variant's value. |
```

src/sentry/remote_config/docs/protocol.md
cmanallen: I'm not sure I'm ready to commit to the range/step just yet.  I was thinking 0-100 as integer values.  Decimal precision seems too granular.  Happy to be convinced otherwise.  lmk.

src/sentry/remote_config/docs/protocol.md
ryan953: fair enough

src/sentry/remote_config/docs/protocol.md
cmanallen: Updated rollout to be nullable 👍

src/sentry/remote_config/docs/protocol.md
ryan953, line 38: was gonna suggest `union[array[Rule], null]` but instead we'll just send down an empty array.

src/sentry/remote_config/docs/protocol.md
```diff
@@ -22,17 +22,68 @@ Retrieve a project's configuration.
...
 **Option Object**
 
-| Field              | Type  | Description        |
-| ------------------ | ----- | ------------------ |
-| sample_rate        | float | Error sample rate. |
-| traces_sample_rate | float | Trace sample rate. |
+| Field              | Type                 | Description        |
+| ------------------ | -------------------- | ------------------ |
+| sample_rate        | OptionVariant<float> | Error sample rate. |
+| traces_sample_rate | OptionVariant<float> | Trace sample rate. |
```
ryan953, line 86: Is `OptionVariant` meant to be our `0.0 to 1.0` type. Could slap that into the top and re-use the type name for `rollout`.

ryan953: oh wait, `Variant` is the name of the existing type above... but in the example it looks like `sample_rate` is using that type.

therefore, we should use the same name here:

```suggestion
| sample_rate        | Variant<float> | Error sample rate. |
| traces_sample_rate | Variant<float> | Trace sample rate. |
```

cmanallen: I need to document it but basically all of the sentry options should have the same feature set as the items in the `features` array.  Except since they're statically typed I'm trying to communicate that its generic over type <T> and the sample_rates are floats.  This part needs work.

cmanallen: Except its missing the `key` field.  So its not exactly the same type.

cmanallen: Added option feature docs.

src/sentry/remote_config/docs/protocol.md
ryan953: I think 1-100 is too narrow, should be a range of 1,000 buckets, or 10,000 buckets imo.

Decimal or integer i don't mind. There'll be a UI that fixes the range and steps for user benefit. SDK just needs to know if it's a float or integer so it can multiple/divide as needed.

src/sentry/remote_config/docs/protocol.md
cmanallen: Turns out its super easy to share code since you're just dropping the key and reusing the generic variants structure.

ryan953: sweet

src/sentry/remote_config/docs/protocol.md
cmanallen: Right I agree we can fix this in UI.  From a product perspective I'm trying to understand the utility of rolling out to 50% versus 50.01%.  Not hard to do from a protocol perspective at all.  We can support it on a "just in case" basis.  But I was just curious of the necessity.

ryan953: bigger apps need more granularity

cmanallen: Yeah I could see "lets start at 0.001% and move up gradually from there".  So maybe the correct approach is 0-100 with arbitrary decimal precision?

cmanallen: Don't forget to refresh and see my previous comment :D

ryan953: I saw. i was typing at the same time and was just like..... ehh i'll post what i have

i think we're on the same page though. arbitrary precision works for me for now! we can revisit that if we have reason to in the future.

src/sentry/remote_config/docs/protocol.md
```diff
@@ -79,16 +144,73 @@ If the server's ETag does not match the request's a 200 response is returned.
...
-        "traces_sample_rate": 0.5
+        "sample_rate": {
+          "value": 0.1,
+          "variants": [
+            {
+              "rollout": null,
+              "rules": [
+                {
+                  "operator": "==",
+                  "property": "device_family",
```
bruno-garcia, line 188: Based on my suggestion above, here's an example where we'd get the value off of the scope, and not (necessarily unless overwritten?) from the user context parameter to the `getConfig` call:

```suggestion
                  "property": "{{sentry.device_family}}",
```

PR #72128, opened 2024-06-05
chore(hybridcloud) Move hacky AlertRule sentryapp logic to serializer

markstory:
These changes echo the changes made in #71956 and #72035 to AlertRule serialization. Getting this logic out of the endpoint and into the serializer will be helpful for reducing complexity here in the future.

Refs HC-1024

### DIFF
--- a/src/sentry/sentry_apps/components.py
+++ b/src/sentry/sentry_apps/components.py
@@ -11,14 +11,15 @@ from django.utils.http import urlencode
 from sentry.mediators.external_requests.select_requester import SelectRequester
 from sentry.models.integrations.sentry_app_component import SentryAppComponent
 from sentry.models.integrations.sentry_app_installation import SentryAppInstallation
+from sentry.services.hybrid_cloud.app.model import RpcSentryAppComponent, RpcSentryAppInstallation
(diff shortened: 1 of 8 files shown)


src/sentry/sentry_apps/components.py
```diff
@@ -11,14 +11,15 @@
...
 from sentry.services.hybrid_cloud.app.serial import serialize_sentry_app_installation
 from sentry.utils import json
 
 
 @dataclasses.dataclass
 class SentryAppComponentPreparer:
-    component: SentryAppComponent
-    install: SentryAppInstallation
+    component: SentryAppComponent | RpcSentryAppComponent
+    install: SentryAppInstallation | RpcSentryAppInstallation
```
GabeVillalobos, line 22: Would it make sense to make `install` an `RpcSentryAppInstallation` only, and add an init method to perform the serialization?

GabeVillalobos (approved):

src/sentry/sentry_apps/components.py
markstory: That could work. I'd like to do that change separately though. There are still a few places where we pass in ORM objects that we could chase down.

PR #72132, opened 2024-06-05
fix(similarity): Use event title in backfill

jangjodi:
Use event.title instead of group.message in record backfill

### DIFF
--- a/src/sentry/tasks/backfill_seer_grouping_records.py
+++ b/src/sentry/tasks/backfill_seer_grouping_records.py
@@ -51,7 +51,6 @@ logger = logging.getLogger(__name__)
 class GroupEventRow(TypedDict):
     event_id: str
     group_id: int
-    message: str
 
 
 class GroupStacktraceData(TypedDict):
@@ -109,7 +108,7 @@ def backfill_seer_grouping_records(
         )
         return
 
-    group_id_message_data = (
+    group_id_data = (
         Group.objects.filter(
             project_id=project.id,
             type=ErrorGroupType.type_id,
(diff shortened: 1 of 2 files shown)


JoshFerge (commented):
sent slack dm as well, but are we sure event.title is equivalent to group.message? and additionally, i think we should change the similarity embeddings endpoint to use the event's title additionally

JoshFerge (approved):

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -436,7 +434,7 @@ def lookup_group_data_stacktrace_bulk(
                         CreateGroupingRecordData(
                             group_id=group_id,
                             project_id=project_id,
-                            message=messages[group_id],
+                            message=event.title,
```
wedamija, line 437: I'm not sure that `event.title` is right here, this is how we get the group title in ingestion https://github.com/getsentry/sentry/blob/251c09dde880f11d052a6120534edd3a7929c36b/src/sentry/event_manager.py#L1308

And event metadata comes from https://github.com/getsentry/sentry/blob/251c09dde880f11d052a6120534edd3a7929c36b/src/sentry/event_manager.py#L858

wedamija: Also, why we we using title in the message? Are you looking for the value from `Group.message`? https://github.com/getsentry/sentry/blob/4c13721d122e9b034b0dcfd99aea9c3a6c91d8e5/src/sentry/models/group.py#L531

JoshFerge: https://github.com/getsentry/sentry/blob/35206b51ef1e019122b98abf6c03043193058b95/src/sentry/grouping/ingest/seer.py#L149

this is how we're getting the title for events when we call seer in ingestion

wedamija (approved):
Discussed title usage on slack, 👍

PR #72134, opened 2024-06-05
feat(charcuterie): Add image to cache

sentaur-athena:
Adding image to cache so the expensive computation of querying events happen only once for all kinds of notifications (slack, email, etc).

[Tech spec](https://www.notion.so/sentry/Chartcuterie-Improvements-Technical-Design-Doc-3dd20eac5c9c494ab0c6bebd0d0a6dad?pvs=4#884676f39698490db944b04bfa39746b)

### DIFF
--- a/src/sentry/integrations/issue_alert_image_builder.py
+++ b/src/sentry/integrations/issue_alert_image_builder.py
@@ -28,9 +28,7 @@ from sentry.utils.services import build_instance_from_options_of_type
 
 logger = logging.getLogger("sentry.chartcuterie")
 locks = LockManager(
-    build_instance_from_options_of_type(
-        LockBackend, settings.SENTRY_POST_PROCESS_LOCKS_BACKEND_OPTIONS
-    )
+    build_instance_from_options_of_type(LockBackend, settings.SENTRY_DEFAULT_LOCKS_BACKEND_OPTIONS)
 )
 
 
@@ -59,6 +57,7 @@ class IssueAlertImageBuilder:
                         image_url = self._get_endpoint_regression_image_url()
                     elif self.group.issue_type == ProfileFunctionRegressionType:
                         image_url = self._get_function_regression_image_url()
+                self.lock.release()
         except UnableToAcquireLock:
             # There is a chance that another thread generated the image
             image_url = cache.get(self.cache_key)
@@ -73,7 +72,9 @@ class IssueAlertImageBuilder:
                 extra={"exception": e, "group_id": self.group.id},
             )
             sentry_sdk.capture_exception()
-        self.lock.release()
+            if self.lock.locked():
+                self.lock.release()
+
         if image_url:
             metrics.incr("chartcuterie.issue_alert.success", tags=self.tags)
             # We don't want to regenerate the image if another type of notification is sending the same one
--- a/src/sentry/integrations/issue_alert_image_builder.py
+++ b/src/sentry/integrations/issue_alert_image_builder.py
@@ -1,6 +1,7 @@
 import logging
 
 import sentry_sdk
+from django.core.cache import cache
 
 from sentry.api import client
 from sentry.charts import backend as charts
@@ -18,7 +19,7 @@ from sentry.models.apikey import ApiKey
 from sentry.models.group import Group
 from sentry.snuba.referrer import Referrer
 from sentry.utils import metrics
-from sentry.utils.performance_issues.detectors.utils import escape_transaction
+from sentry.utils.performance_issues.detectors import utils
 
 logger = logging.getLogger("sentry.chartcuterie")
 
@@ -35,22 +36,41 @@ class IssueAlertImageBuilder:
 
     def get_image_url(self) -> str | None:
         try:
-            image_url = None
             metrics.incr("chartcuterie.issue_alert.attempt", tags=self.tags)
-            if self.group.issue_type == PerformanceP95EndpointRegressionGroupType:
-                image_url = self._get_endpoint_regression_image_url()
-            elif self.group.issue_type == ProfileFunctionRegressionType:
-                image_url = self._get_function_regression_image_url()
-
+            image_url = cache.get(self.cache_key)
             if image_url:
                 metrics.incr("chartcuterie.issue_alert.success", tags=self.tags)
                 return image_url
+
+            lock_key = f"lock_{self.cache_key}"
+            lock_acquired = cache.add(key=lock_key, value="locked", timeout=120)
+            if lock_acquired:
+                # Checking again in case another thread generated the image while
+                # this thread was acquiring the lock
+                image_url = cache.get(self.cache_key)
+                if image_url is None:
+                    if self.group.issue_type == PerformanceP95EndpointRegressionGroupType:
+                        image_url = self._get_endpoint_regression_image_url()
+                    elif self.group.issue_type == ProfileFunctionRegressionType:
+                        image_url = self._get_function_regression_image_url()
+                if image_url:
+                    metrics.incr("chartcuterie.issue_alert.success", tags=self.tags)
+                    cache.delete(key=lock_key)
+                    return image_url
+            else:
+                # There is a chance that another thread generated the image
+                image_url = cache.get(self.cache_key)
+                if image_url:
+                    metrics.incr("chartcuterie.issue_alert.success", tags=self.tags)
+                    return image_url
         except Exception as e:
             logger.exception(
                 "issue_alert_chartcuterie_image.failed",
                 extra={"exception": e, "group_id": self.group.id},
             )
             sentry_sdk.capture_exception()
+            if lock_acquired:
+                cache.delete(key=lock_key)
         return None
 
     def _get_endpoint_regression_image_url(self) -> str | None:
@@ -58,7 +78,7 @@ class IssueAlertImageBuilder:
         event = self.group.get_latest_event_for_environments()
         if event is None or event.transaction is None or event.occurrence is None:
             return None
-        transaction_name = escape_transaction(event.transaction)
+        transaction_name = utils.escape_transaction(event.transaction)
         period = get_relative_time(anchor=get_approx_start_time(self.group), relative_days=14)
         resp = client.get(
             auth=ApiKey(organization_id=organization.id, scope_list=["org:read"]),
@@ -81,6 +101,7 @@ class IssueAlertImageBuilder:
                 "evidenceData": event.occurrence.evidence_data,
                 "percentileData": resp.data["p95(transaction.duration)"]["data"],
             },
+            cache_key=self.cache_key,
         )
 
     def _get_function_regression_image_url(self) -> str | None:
@@ -118,4 +139,5 @@ class IssueAlertImageBuilder:
                 "evidenceData": evidence_data,
                 "rawResponse": resp.data,
             },
+            cache_key=self.cache_key,
         )
--- a/src/sentry/integrations/issue_alert_image_builder.py
+++ b/src/sentry/integrations/issue_alert_image_builder.py
@@ -1,6 +1,7 @@
 import logging
 
 import sentry_sdk
+from django.conf import settings
 from django.core.cache import cache
 
 from sentry.api import client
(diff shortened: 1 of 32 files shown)


tests/acceptance/chartcuterie/test_image_block_builder.py
```diff
@@ -63,13 +65,43 @@ def test_image_block_for_endpoint_regression(self):
...
+        # Start all the threads
+        for thread in threads:
+            thread.start()
+
+        # Wait for all threads to finish
+        for thread in threads:
+            thread.join()
+
+        # make_request()
+        # make_request()
```
sentaur-athena, line 100: If I comment the thread code and instead call back to back request the result gets returned from the cache but then I won't be properly testing the lock logic.

When using thread though the test completely fail with an exception. It seems like each thread is creating the test DB again 🤔  which I don't understand.

sentaur-athena: Pasting the exception here for convenience:
```
________________________ ERROR at teardown of TestSlackImageBlockBuilder.test_caching ________________________
src/sentry/db/postgres/decorators.py:91: in inner
    return func(self, sql, *args, **kwargs)
src/sentry/db/postgres/base.py:85: in execute
    return self.cursor.execute(sql)
E   psycopg2.errors.ObjectInUse: database "test_region" is being accessed by other users
E   DETAIL:  There are 2 other sessions using the database.

During handling of the above exception, another exception occurred:
.venv/lib/python3.11/site-packages/django/db/backends/utils.py:103: in _execute
    return self.cursor.execute(sql)
src/sentry/db/postgres/decorators.py:77: in inner
    raise_the_exception(self.db, e)
src/sentry/db/postgres/decorators.py:75: in inner
    return func(self, *args, **kwargs)
src/sentry/db/postgres/decorators.py:18: in inner
    return func(self, *args, **kwargs)
src/sentry/db/postgres/decorators.py:93: in inner
    raise type(e)(f"{e!r}\nSQL: {sql}").with_traceback(e.__traceback__)
src/sentry/db/postgres/decorators.py:91: in inner
    return func(self, sql, *args, **kwargs)
src/sentry/db/postgres/base.py:85: in execute
    return self.cursor.execute(sql)
E   psycopg2.errors.ObjectInUse: ObjectInUse('database "test_region" is being accessed by other users\nDETAIL:  There are 2 other sessions using the database.\n')
E   SQL: DROP DATABASE "test_region"

The above exception was the direct cause of the following exception:
.venv/lib/python3.11/site-packages/pytest_django/fixtures.py:150: in django_db_setup
    teardown_databases(db_cfg, verbosity=request.config.option.verbose)
.venv/lib/python3.11/site-packages/django/test/utils.py:366: in teardown_databases
    connection.creation.destroy_test_db(old_name, verbosity, keepdb)
.venv/lib/python3.11/site-packages/django/db/backends/base/creation.py:311: in destroy_test_db
    self._destroy_test_db(test_database_name, verbosity)
.venv/lib/python3.11/site-packages/django/db/backends/base/creation.py:327: in _destroy_test_db
    cursor.execute(
.venv/lib/python3.11/site-packages/sentry_sdk/utils.py:1709: in runner
    return original_function(*args, **kwargs)
.venv/lib/python3.11/site-packages/django/db/backends/utils.py:79: in execute
    return self._execute_with_wrappers(
.venv/lib/python3.11/site-packages/django/db/backends/utils.py:92: in _execute_with_wrappers
    return executor(sql, params, many, context)
.venv/lib/python3.11/site-packages/django/db/backends/utils.py:100: in _execute
    with self.db.wrap_database_errors:
.venv/lib/python3.11/site-packages/django/db/utils.py:91: in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
.venv/lib/python3.11/site-packages/django/db/backends/utils.py:103: in _execute
    return self.cursor.execute(sql)
src/sentry/db/postgres/decorators.py:77: in inner
    raise_the_exception(self.db, e)
src/sentry/db/postgres/decorators.py:75: in inner
    return func(self, *args, **kwargs)
src/sentry/db/postgres/decorators.py:18: in inner
    return func(self, *args, **kwargs)
src/sentry/db/postgres/decorators.py:93: in inner
    raise type(e)(f"{e!r}\nSQL: {sql}").with_traceback(e.__traceback__)
src/sentry/db/postgres/decorators.py:91: in inner
    return func(self, sql, *args, **kwargs)
src/sentry/db/postgres/base.py:85: in execute
    return self.cursor.execute(sql)
E   django.db.utils.OperationalError: ObjectInUse('database "test_region" is being accessed by other users\nDETAIL:  There are 2 other sessions using the database.\n')
E   SQL: DROP DATABASE "test_region"

During handling of the above exception, another exception occurred:
.venv/lib/python3.11/site-packages/pytest_django/fixtures.py:152: in django_db_setup
    request.node.warn(
E   pytest.PytestWarning: Error when trying to teardown test databases: OperationalError('ObjectInUse(\'database "test_region" is being accessed by other users\\nDETAIL:  There are 2 other sessions using the database.\\n\')\nSQL: DROP DATABASE "test_region"')
```

src/sentry/integrations/issue_alert_image_builder.py
```diff
@@ -35,30 +36,49 @@ def __init__(self, group: Group, provider: ExternalProviderEnum) -> None:
...
     def get_image_url(self) -> str | None:
         try:
-            image_url = None
             metrics.incr("chartcuterie.issue_alert.attempt", tags=self.tags)
-            if self.group.issue_type == PerformanceP95EndpointRegressionGroupType:
-                image_url = self._get_endpoint_regression_image_url()
-            elif self.group.issue_type == ProfileFunctionRegressionType:
-                image_url = self._get_function_regression_image_url()
-
+            image_url = cache.get(self.cache_key)
```
evanpurkhiser, line 40: Can we move the cache fetching into the `generate_chart` function?

Though I'm guessing you want to do this early before we have to query Snuba again for all the data right?

In that case what I might suggest is adjusting the `generate_chart` so that `data` can be a function that generates the data if needed (ie, the cache lookup is a miss)

Alternatively, let's do all the caching outside the `generate_chart`. Right now we've coupled implementation details of `generate_chart` to implementation details of its caller.

sentaur-athena: Yeah, here I'm trying to query `event-stats` once. I'll take the cache logic out since it's easier 😅

evanpurkhiser: 👍

scttcper:
is there a single size we're generating now? I think in the past email used larger images compared to slack

sentaur-athena:
@scttcper I think we can totally stick to the same size for simplicity. I got feedback that slack images are too small so maybe increase that a bit and use in both.

src/sentry/integrations/issue_alert_image_builder.py
```diff
@@ -35,30 +36,49 @@ def __init__(self, group: Group, provider: ExternalProviderEnum) -> None:
...
             if image_url:
                 metrics.incr("chartcuterie.issue_alert.success", tags=self.tags)
                 return image_url
+
+            lock_key = f"lock_{self.cache_key}"
+            lock_acquired = cache.add(key=lock_key, value="locked", timeout=120)
```
sentaur-athena, line 46: @evanpurkhiser what are your thoughts about timeout and sleep here?
1. Timeout here means the lock will expire after 120s. I think it makes sense for 1 query and 1 service call but we can honestly increase this with low risk. Let's say 240 s
2. At line 61 I need to add a sleep. If lock is not acquired it means someone is creating the image right now so we wait for the image to get to cache. I don't want to put a `while True:` for obvious reason so was thinking to use the same time we use for timeout.

evanpurkhiser: I would recommend using our locking utilities https://github.com/getsentry/sentry/blob/master/src/sentry/utils/locking/lock.py

There is a blocking lock you can use that will do this for you.

sentaur-athena: I didn't know if existed, will definitely reuse and set a timeout of 240s.

evanpurkhiser: 240s feels pretty high no? How long should it take to generate these images?

tests/acceptance/chartcuterie/test_image_block_builder.py
sentaur-athena: Asked @markstory in a separate thread adding response here to close the loop. Django is making new connection in each thread that database is used (which is in every thread here). I'll move to [using existing lock utils](https://github.com/getsentry/sentry/blob/master/src/sentry/utils/locking/lock.py) rather than creating mine from scratch which are better tested and then remove my threads code here.

src/sentry/integrations/issue_alert_image_builder.py
sentaur-athena: Ideally less than 1 mins but I want to set it to worse than the worst case. Here's the trade-off I'm thinking:
Side 1: If I set it at 120s and then it takes one image to get generated 125 the second thread will acquire the lock and start re-generating the image because it's not in the cache yet.
Side 2: If I set it too high, to be sure it's definitely more than any image generation, the 2nd thread will never regenerate. The only downside I see is that lock will be released later if no one releases it.

evanpurkhiser: It would probably make sense to have a timeout configured on the call to chartcuterie and our lock just matches that timeout

sentaur-athena: It's actually reasonably low now that [I look](https://sentry.sentry.io/performance/traces/?project=1&query=span.op%3Acharts.chartcuterie.generate_chart&statsPeriod=14d). both upload and getting response spans are less than 1s.

evanpurkhiser: Yeah I would never expect this to be slow

src/sentry/integrations/issue_alert_image_builder.py
```diff
@@ -32,33 +43,51 @@ def __init__(self, group: Group, provider: ExternalProviderEnum) -> None:
...
+                if image_url is None:
+                    if self.group.issue_type == PerformanceP95EndpointRegressionGroupType:
+                        image_url = self._get_endpoint_regression_image_url()
+                    elif self.group.issue_type == ProfileFunctionRegressionType:
+                        image_url = self._get_function_regression_image_url()
+        except UnableToAcquireLock:
+            # There is a chance that another thread generated the image
+            image_url = cache.get(self.cache_key)
+            if not image_url:
+                logger.warning(
```
sentaur-athena, line 66: I will monitor this and if it happens a lot it means the timeout and acquire numbers should be adjusted. They seem reasonable now. It's a non-breaking thing because worst case the image won't be attached. The alert will still be sent.

src/sentry/integrations/issue_alert_image_builder.py
```diff
@@ -18,9 +20,18 @@
...
+from sentry.utils.locking import UnableToAcquireLock
+from sentry.utils.locking.backends import LockBackend
+from sentry.utils.locking.manager import LockManager
+from sentry.utils.performance_issues.detectors import utils
+from sentry.utils.services import build_instance_from_options_of_type
 
 logger = logging.getLogger("sentry.chartcuterie")
+locks = LockManager(
+    build_instance_from_options_of_type(
+        LockBackend, settings.SENTRY_POST_PROCESS_LOCKS_BACKEND_OPTIONS
```
evanpurkhiser, line 32: Is this the right options to use?

sentaur-athena: This lock util is not widely used in codebase like I'm the only person using the blockacquire which seems a bit scary.

src/sentry/integrations/issue_alert_image_builder.py
```diff
@@ -32,33 +43,51 @@ def __init__(self, group: Group, provider: ExternalProviderEnum) -> None:
...
+                    "issue_alert_chartcuterie_image.lock.failed",
+                    extra={"group_id": self.group.id},
+                )
         except Exception as e:
             logger.exception(
                 "issue_alert_chartcuterie_image.failed",
                 extra={"exception": e, "group_id": self.group.id},
             )
             sentry_sdk.capture_exception()
+        self.lock.release()
```
evanpurkhiser, line 76: Is the lock always locked? Looks like it's behind `if image_url is None` no?

src/sentry/integrations/issue_alert_image_builder.py
evanpurkhiser: I'll let you figure it out, I'm not exactly sure.

src/sentry/integrations/issue_alert_image_builder.py
sentaur-athena: Yes, if the image is already in cache we don't need lock. lock.release does nothing if the lock is not acquired, it fails silently. I'm adding it at the end to avoid writing it in all the ifs and elses.

src/sentry/integrations/issue_alert_image_builder.py
sentaur-athena: Updated to the correct one.

evanpurkhiser (approved):

PR #72135, opened 2024-06-05
feat(crons): Add DELETE method for deleting processing errors by type

davidenwang:
Adds a DELETE method onto the existing `project_monitor_processing_errors_index` endpoint

1. Accepts a query parameter for `errortype`
2. Deletes all processing errors by type for the monitor supplied in path parameters

### DIFF
--- a/src/sentry/monitors/endpoints/project_monitor_processing_errors_index.py
+++ b/src/sentry/monitors/endpoints/project_monitor_processing_errors_index.py
@@ -1,4 +1,5 @@
 from drf_spectacular.utils import extend_schema
+from rest_framework.exceptions import ValidationError
 from rest_framework.response import Response
 
 from sentry.api.api_owners import ApiOwner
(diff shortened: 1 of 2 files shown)


src/sentry/monitors/endpoints/project_monitor_processing_errors_index.py
```diff
@@ -51,3 +63,28 @@ def get(self, request: AuthenticatedHttpRequest, project, monitor) -> Response:
...
+            401: RESPONSE_UNAUTHORIZED,
+            403: RESPONSE_FORBIDDEN,
+            404: RESPONSE_NOT_FOUND,
+        },
+    )
+    def delete(
+        self, request: AuthenticatedHttpRequest, project: Project, monitor: Monitor
+    ) -> Response:
+        try:
+            error_type = ProcessingErrorType(int(request.GET.get("errortype")))
```
evanpurkhiser, line 85: Do we need some error handling here. This will 500 if we don't pass errortype right?

davidenwang: hmmm my understanding is that we would get a ValueError here and then return a ValidationError back to the user, but lemme confirm

evanpurkhiser: maybe a quick test?

davidenwang: yee my bad lemme write a test

wedamija (approved):

PR #72137, opened 2024-06-05
deps(ui): Upgrade react testing library dependencies

scttcper:
upgrades `@testing-library/react` from 14 -> 16.
upgrades `@testing-library/dom` from 9 -> 10

Notes:
- @testing-library/dom is now a peer dependency, have to install
- Support for react 19
- Upgrades aria-query to 5.3.0 https://github.com/testing-library/dom-testing-library/pull/1241
  - changes getByRole('link') `a element now requires a set href attribute to have an implicit link role`


static/app/components/sidebar/index.spec.tsx
```diff
@@ -257,7 +257,7 @@ describe('Sidebar', function () {
       const {unmount} = renderSidebar({organization});
 
       // This will start timer to mark as seen
-      await userEvent.click(await screen.findByRole('link', {name: "What's new"}), {
+      await userEvent.click(await screen.findByTestId('sidebar-broadcasts'), {
```
evanpurkhiser, line 260: Why did it have to change?

scttcper: this one was missing an href

evanpurkhiser (approved):

static/app/components/sidebar/sidebarDropdown/switchOrganization.spec.tsx
```diff
@@ -92,11 +92,11 @@ describe('SwitchOrganization', function () {
 
     expect(screen.getByRole('list')).toBeInTheDocument();
 
-    const org1Link = screen.getByRole('link', {name: 'org1 Organization 1'});
+    const org1Link = screen.getByRole('link', {name: 'O Organization 1'});
```
scttcper, line 95: i think it was able to get the aria label on the svg before.
now it is the literal text of the avatar

PR #72138, opened 2024-06-05
feat(analytics): update search query analytics 

KevinL10:
Adds an `isMultiProject` field to `search.searched` analytics and a new Amplitude event to track individual key/value filters.


malwilley (commented):
Is this meant to handle the cases where the user is searching "My Projects" (when there are no projects selected) or "All Projects" (when the selected project is `-1`)?

KevinL10:
Thanks for the catch. I've passed `projectIds` through the props and now handling `isMultiProject` in the SmartSearchBar itself.

static/app/views/replays/list/replaySearchBar.tsx
```diff
@@ -91,7 +91,7 @@ function ReplaySearchBar(props: Props) {
         orgSlug: organization.slug,
         tagKey: tag.key,
         search: searchQuery,
-        projectIds: projectIdStrings,
+        projectIds: projectIds.map(String),
```
aliu3ntry, line 94: Noticed the old code uses `?`, what happens if projectIds is null here?
Same applies for feedbackSearch

KevinL10: I removed them because projectIds is always `number[]` and I believe defaults to an empty array. I've added them back in case.

aliu39 (approved):

PR #72140, opened 2024-06-05
chore(hybrid-cloud): Adds timeout and retry count options to rpc_method decorator

GabeVillalobos:
# [WIP] RPC Method Options
Adds `timeout_override_seconds` and `retry_count_override` options to the `rpc_method` decorator, in order to allow per-method overrides of these settings.

Organization provisioning flows have different requirements than most of our typical RPCs and may require additional configuration. Being able to set a retry count and timeout should help with transient failure rates.


GabeVillalobos:
Not going with this approach, closing.

PR #72145, opened 2024-06-05
feat(custom-views): Add custom views PUT endpoint

### LINKED ISSUE #71587: Custom Views: Create PUT Endpoint
Create `PUT` `<ORG_ID_OR_SLUG>/user-views/` endpoint 

This endpoint will be responsible for deleting, creating, and bulk updating new views for an org member. This endpoint will need to be gated behind a feature flag until GA.

Parameters:
* `org_member_id`
* `view[] | view = {id, name, query, sort, position}`

Logic:
* Create new row if view has no id 
* Delete row if exists in database but not in request 
* Update row if view has same id

MichaelSun48:
This PR implements the `PUT` `organizations/.../group-search-views` endpoint, which bulk updates a user's custom views within an organization.

### DIFF
--- a/src/sentry/api/serializers/rest_framework/groupsearchview.py
+++ b/src/sentry/api/serializers/rest_framework/groupsearchview.py
@@ -1,0 +1,23 @@
+from rest_framework import serializers
+
+from sentry.models.savedsearch import SortOptions
+
+MAX_VIEWS = 50
+
+
+class ViewSerializer(serializers.Serializer):
+    id = serializers.CharField(required=False)
+    name = serializers.CharField(required=True)
+    query = serializers.CharField(required=True)
+    query_sort = serializers.ChoiceField(
+        choices=SortOptions.as_choices(), default=SortOptions.DATE, required=False
+    )
+
+
+class GroupSearchViewRestSerializer(serializers.Serializer):
+    views = serializers.ListField(child=ViewSerializer(), required=True, max_length=MAX_VIEWS)
+
+    def validate(self, data):
+        if len(data.get("views", [])) == 0:
+            raise serializers.ValidationError("Must provide at least one view")
+        return data
--- a/src/sentry/api/serializers/rest_framework/groupsearchview.py
+++ b/src/sentry/api/serializers/rest_framework/groupsearchview.py
@@ -15,9 +15,9 @@ class ViewValidator(serializers.Serializer):
 
 
 class GroupSearchViewValidator(serializers.Serializer):
-    views = serializers.ListField(child=ViewValidator(), required=True, max_length=MAX_VIEWS)
+    views = serializers.ListField(
+        child=ViewValidator(), required=True, min_length=1, max_length=MAX_VIEWS
+    )
 
     def validate(self, data):
-        if len(data.get("views", [])) == 0:
-            raise serializers.ValidationError("Must provide at least one view")
         return data
--- a/src/sentry/api/serializers/rest_framework/groupsearchview.py
+++ b/src/sentry/api/serializers/rest_framework/groupsearchview.py
@@ -9,7 +9,7 @@ class ViewSerializer(serializers.Serializer):
     id = serializers.CharField(required=False)
     name = serializers.CharField(required=True)
     query = serializers.CharField(required=True)
-    query_sort = serializers.ChoiceField(
+    querySort = serializers.ChoiceField(
         choices=SortOptions.as_choices(), default=SortOptions.DATE, required=False
     )
 
--- a/src/sentry/api/serializers/rest_framework/groupsearchview.py
+++ b/src/sentry/api/serializers/rest_framework/groupsearchview.py
@@ -5,7 +5,7 @@ from sentry.models.savedsearch import SortOptions
 MAX_VIEWS = 50
 
 
-class ViewSerializer(serializers.Serializer):
+class ViewValidator(serializers.Serializer):
     id = serializers.CharField(required=False)
     name = serializers.CharField(required=True)
     query = serializers.CharField(required=True)
@@ -14,8 +14,8 @@ class ViewSerializer(serializers.Serializer):
     )
 
 
-class GroupSearchViewRestSerializer(serializers.Serializer):
-    views = serializers.ListField(child=ViewSerializer(), required=True, max_length=MAX_VIEWS)
+class GroupSearchViewValidator(serializers.Serializer):
+    views = serializers.ListField(child=ViewValidator(), required=True, max_length=MAX_VIEWS)
 
     def validate(self, data):
         if len(data.get("views", [])) == 0:
--- a/src/sentry/issues/endpoints/organization_group_search_views.py
+++ b/src/sentry/issues/endpoints/organization_group_search_views.py
@@ -83,7 +83,7 @@ class OrganizationGroupSearchViewsEndpoint(OrganizationEndpoint):
         serializer = GroupSearchViewRestSerializer(data=request.data)
 
         if not serializer.is_valid():
-            return Response(serializer.errors, status=400)
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
         validated_data = serializer.validated_data
         bulk_update_views(organization, request.user.id, validated_data)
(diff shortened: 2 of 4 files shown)


src/sentry/issues/endpoints/organization_group_search_views.py
```diff
@@ -72,3 +75,64 @@ def get(self, request: Request, organization: Organization) -> Response:
             order_by="position",
             on_results=lambda x: serialize(x, request.user, serializer=GroupSearchViewSerializer()),
         )
+
+    def put(self, request: Request, organization: Organization) -> Response:
+        if not features.has("organizations:issue-stream-custom-views", organization):
+            return Response(status=status.HTTP_405_METHOD_NOT_ALLOWED)
```
roggenkemper, line 81: I think you're using 404 in the GET

MichaelSun48: Good catch - changing now

tests/sentry/issues/endpoints/test_organization_group_search_views.py
```diff
@@ -103,3 +106,198 @@ def test_get_default_views(self):
...
+    def setUp(self):
+        self.login_as(user=self.user)
+        self.base_data = self.create_base_data()
+
+        self.url = reverse(
+            "sentry-api-0-organization-group-search-views",
+            kwargs={"organization_id_or_slug": self.organization.slug},
+        )
+
+    @with_feature({"organizations:issue-stream-custom-views": True})
```
roggenkemper, line 160: might be going crazy but i feel like most of the time i see `with_feature` i just see it as `@with_feature("organizations:issue-stream-custom-views")` . is there a reason why you have it this way?

MichaelSun48: Nope I have no idea why I made it that way either - probably just copy pasted it from somewhere else. Looks like it still works as expected though

malwilley (approved):
This looks great! Very well tested 💯

src/sentry/api/serializers/rest_framework/groupsearchview.py
```diff
@@ -0,0 +1,23 @@
+from rest_framework import serializers
+
+from sentry.models.savedsearch import SortOptions
+
+MAX_VIEWS = 50
+
+
+class ViewSerializer(serializers.Serializer):
+    id = serializers.CharField(required=False)
```
JoshFerge, line 9: what is id? should this be an int?

src/sentry/issues/endpoints/organization_group_search_views.py
```diff
@@ -72,3 +76,70 @@ def get(self, request: Request, organization: Organization) -> Response:
             order_by="position",
             on_results=lambda x: serialize(x, request.user, serializer=GroupSearchViewSerializer()),
         )
+
+    def put(self, request: Request, organization: Organization) -> Response:
```
JoshFerge, line 80: are we updating multiple `GroupView`s in this put?
why PUT instead of POST then?

src/sentry/issues/endpoints/organization_group_search_views.py
```diff
@@ -72,3 +76,70 @@ def get(self, request: Request, organization: Organization) -> Response:
...
+        if "id" not in view:
+            _create_view(org, user_id, view, position=idx)
+        else:
+            _update_existing_view(view, position=idx)
+
+
+def _delete_missing_views(org: Organization, user_id: int, view_ids_to_keep: list[int]):
+    GroupSearchView.objects.filter(organization=org, user_id=user_id).exclude(
+        id__in=view_ids_to_keep
+    ).delete()
```
JoshFerge, line 125: hmm should the deletions be done in a separate endpoint?

JoshFerge (commented):
few Qs

src/sentry/issues/endpoints/organization_group_search_views.py
JoshFerge: talked offline, this is a "Backend for the Frontend" oriented endpoint.

src/sentry/issues/endpoints/organization_group_search_views.py
```diff
@@ -72,3 +76,70 @@ def get(self, request: Request, organization: Organization) -> Response:
...
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+        validated_data = serializer.validated_data
+        try:
+            with transaction.atomic(using=router.db_for_write(GroupSearchView)):
+                bulk_update_views(organization, request.user.id, validated_data)
+
+                query = GroupSearchView.objects.filter(
+                    organization=organization, user_id=request.user.id
+                )
```
JoshFerge, line 96: since query doesn't actually get called until it's evaluated, i don't think it needs to be in the transaction here.

src/sentry/issues/endpoints/organization_group_search_views.py
```diff
@@ -72,3 +76,70 @@ def get(self, request: Request, organization: Organization) -> Response:
...
+        return self.paginate(
+            request=request,
+            queryset=query,
+            order_by="position",
+            on_results=lambda x: serialize(x, request.user, serializer=GroupSearchViewSerializer()),
+        )
+
+
+def bulk_update_views(org: Organization, user_id: int, validated_data):
+    views = validated_data.get("views")
```
JoshFerge, line 109: should views always exist? why do you need to use `.get`?

src/sentry/issues/endpoints/organization_group_search_views.py
MichaelSun48: This PUT endpoint is designed to reduce complexity on the frontend by allowing it to just send over the list of views it sees over to the backend. The backend takes this list of `GroupSearchView`s, and handles the logic for adding, updating, and deleting to the database in bulk. We'll consider adding public facing endpoints that adhere to RESTful conventions (like a POST endpoint that updates one view) in the future once we get closer to launching this feature.

This should also address this comment below:
> hmm should the deletions be done in a separate endpoint?

src/sentry/issues/endpoints/organization_group_search_views.py
```diff
@@ -72,3 +76,70 @@ def get(self, request: Request, organization: Organization) -> Response:
...
+
+        validated_data = serializer.validated_data
+        try:
+            with transaction.atomic(using=router.db_for_write(GroupSearchView)):
+                bulk_update_views(organization, request.user.id, validated_data)
+
+                query = GroupSearchView.objects.filter(
+                    organization=organization, user_id=request.user.id
+                )
+        except IntegrityError:
```
JoshFerge, line 97: should we capture a sentry error here? seems like right now it would silently fail

src/sentry/api/serializers/rest_framework/groupsearchview.py
MichaelSun48: This is the id of the view in the database. Most other endpoints cast ids to string when returning, and it seems like it has become precedent to expect a string in the input as well ([example](https://github.com/getsentry/sentry/blob/13975b06a59de20face9fd2d1ed9a3f61ccecc54/src/sentry/api/serializers/rest_framework/dashboard.py#L140))

JoshFerge: got it 👍🏼

src/sentry/api/serializers/rest_framework/groupsearchview.py
```diff
@@ -0,0 +1,23 @@
...
+    query = serializers.CharField(required=True)
+    querySort = serializers.ChoiceField(
+        choices=SortOptions.as_choices(), default=SortOptions.DATE, required=False
+    )
+
+
+class GroupSearchViewValidator(serializers.Serializer):
+    views = serializers.ListField(child=ViewValidator(), required=True, max_length=MAX_VIEWS)
+
+    def validate(self, data):
```
JoshFerge, line 20: why can't you use min_length above?

JoshFerge (approved):
looks good. i do think using member_id could end up simplifying things but this makes sense as is. also added a comment about the serializer

src/sentry/api/serializers/rest_framework/groupsearchview.py
MichaelSun48: No reason, just missed that when writing it somehow. Changing it now

PR #72148, opened 2024-06-05
ref: improve typing of sentry.tasks.store

asottile-sentry:
Event is a sneaky Any alias so I improved it to MutableMapping / Mapping

once CanonicalKeyDict is gone these can probably become dict directly

### DIFF
--- a/src/sentry/datascrubbing.py
+++ b/src/sentry/datascrubbing.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
 import copy
-from typing import Any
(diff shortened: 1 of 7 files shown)


joshuarli (approved):

Swatinem (approved):

anonrig (approved):

snigdhas (approved):
`Event = Any` 💀

asottile-sentry:
ready for review again -- now that `LazyServiceWrapper` is type checked I no longer need the extra function \o/

PR #72153, opened 2024-06-05
fix(dynamic-sampling): add project-level permission check before adding custom rule

oioki:
If an organization user does not have project-level permission to a project (for example, with Open Membership disabled), they should not be allowed to create or view custom rules (aka _Get Samples_).


matejminar (approved):
Thank you for fixing this! 🙏

PR #72156, opened 2024-06-05
feat(insights): Collapse "Insights" sidebar accordion by default

gggritso:
Request from Design to, on first page load, collapse the "Insights" accordion, so it's not as distracting. After that, works as usual, and stores open/closed state in `localStorage`

**e.g.,**

Screenshot 2024-06-05 at 2 55 29 PM


static/app/components/sidebar/sidebarAccordion.tsx
```diff
@@ -25,9 +25,14 @@ import SidebarItem, {isItemActive} from './sidebarItem';
 
 type SidebarAccordionProps = SidebarItemProps & {
   children?: React.ReactNode;
+  initiallyExpanded?: boolean;
```
evanpurkhiser, line 28: what about `collapsedByDefault`

static/app/components/sidebar/sidebarAccordion.tsx
```diff
@@ -25,9 +25,14 @@ import SidebarItem, {isItemActive} from './sidebarItem';
...
 };
 
-function SidebarAccordion({children, ...itemProps}: SidebarAccordionProps) {
+function SidebarAccordion({
+  children,
+  initiallyExpanded,
```
evanpurkhiser, line 33:
```suggestion
  initiallyExpanded = true,
```

static/app/components/sidebar/sidebarAccordion.tsx
gggritso: @evanpurkhiser you tell me! I went with `initiallyExpanded` to match how it's stored and referenced. Otherwise, you get:

```tsx
  const [expanded, setExpanded] = useLocalStorageState(
    `sidebar-accordion-${id}:expanded`,
    !collapsedByDefault
  );
```

evanpurkhiser (approved):

PR #72157, opened 2024-06-05
feat(query-builder): Add support for boolean operators

malwilley:
Similar to parens, so I pulled out the common logic into a new `DeletableToken` component.

CleanShot 2024-06-05 at 11 56 06

### DIFF
--- a/static/app/components/searchQueryBuilder/deletableToken.tsx
+++ b/static/app/components/searchQueryBuilder/deletableToken.tsx
@@ -1,0 +1,145 @@
+import type React from 'react';
+import {useRef} from 'react';
+import styled from '@emotion/styled';
+import {mergeProps} from '@react-aria/utils';
+import type {ListState} from '@react-stately/list';
+import type {Node} from '@react-types/shared';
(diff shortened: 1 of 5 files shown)


static/app/components/searchQueryBuilder/deletableToken.tsx
malwilley, line 1: This is all taken from `paren.tsx`

static/app/components/searchQueryBuilder/input.tsx
```diff
@@ -215,7 +215,7 @@ function SearchQueryBuilderInputInternal({
   tabIndex,
   state,
 }: SearchQueryBuilderInputInternalProps) {
-  const trimmedTokenValue = token.value.trim();
+  const trimmedTokenValue = token.text.trim();
```
malwilley, line 218: Only difference is that text includes quotes, which I noticed was necessary if you want to type `"foo and bar"` without `and` being tokenized

scttcper (approved):

PR #72160, opened 2024-06-05
update clickhouse version to 23.3.19.33

MeredithAnya:
As of today we have finally upgrade all SaaS clickhouse clusters to `23.3.19.33`, we should now be using that as the default for local development. Self-hosted is actually already defaulting to `23.8`, which will be the next version SaaS/ST will be upgrading to shortly.


nikhars (approved):
🎉

PR #72164, opened 2024-06-05
ref(seer grouping): Don't call Seer when there's a custom fingerprint

lobsterkatie:
This adds a custom fingerprint check to `should_call_seer_for_grouping`, so that if the user has (or we have) forced the event to have a certain fingerprint, the grouping based on that fingerprint is always used, rather than deferring to Seer. (This new check thus prevents us from calling Seer, since we know we won't use whatever it might tell us.)


JoshFerge (approved):
awesome, great work!

PR #72165, opened 2024-06-05
fix(performance): Only display Interaction samples with a target

edwardgou-sentry:
Some interaction samples may not have a target associated. This can happen for a number of different reasons. This change updates the webvitals module to only fetch interactions that have a target.


static/app/views/performance/browser/webVitals/utils/queries/useInpSpanSamplesWebVitalsQuery.tsx
```diff
@@ -30,6 +30,8 @@ export function useInpSpanSamplesWebVitalsQuery({
   const {data, isLoading, ...rest} = useSpansIndexed(
     {
       search: MutableSearch.fromQueryObject({
+        has: 'message',
```
edwardgou-sentry, line 33: For some reason, `has:span.description` doesn't seem to work (probably due to some special case handling in discover). `message` does work though, and it's mapped to the same tag in discover, so we're using that instead.

0Calories (approved):

PR #72166, opened 2024-06-05
ref: add mypy plugin for LazyServiceWrapper

asottile-sentry:
prior to this all accesses on LazyServiceWrapper were Any -- now they are the type of the base service class

### DIFF
--- a/src/sentry/utils/lazy_service_wrapper.py
+++ b/src/sentry/utils/lazy_service_wrapper.py
@@ -1,0 +1,119 @@
+"""split out from sentry.utils.services to handle mypy plugin specialization"""
+from __future__ import annotations
+
+import enum
+import inspect
+import itertools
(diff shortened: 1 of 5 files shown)


src/sentry/utils/lazy_service_wrapper.py
```diff
@@ -0,0 +1,119 @@
+"""split out from sentry.utils.services to handle mypy plugin specialization"""
```
evanpurkhiser, line 1: Feels like this should be a regular comment instead of a docstring no? since this is more implementation detail than module documentation

asottile-sentry: yeah or probably just deleted entirely

evanpurkhiser (approved):
really cool. Nice work on this

PR #72047, opened 2024-06-04
chore(ui): Improve sidebar specs

gggritso:
Follow-up to https://github.com/getsentry/sentry/pull/72012. Updates the tests a bit, moves things around, and expands test coverage.

- Move link set tests together

- Simplify `ConfigStore` init
    
    - init in `beforeEach`
    - remove `act`

- Update spec structure
    
    1. Set up all possible sidebar feature flags
    2. Await `Broadcast` mount
    3. Fetch elements by role, not by ID
    4. Check element by accessible title, not by ID

PR #72167, opened 2024-06-05
chore(ui): Improve sidebar specs II

gggritso:
Re-do of https://github.com/getsentry/sentry/pull/72047 which I had to revert due to CI failures. It intersected badly with the RTL upgrade that happend around the same time.

The secret sauce was adding a proper `useLocation` mock. Without it, the sidebar links didn't have a valid `href=` attribute, and therefore didn't have a `link` role, so `getAllByRole('link')` didn't return anything.

NOTE: Setting `aria-role="link"` does _not_ do it! It doesn't like to be tricked.


scttcper (approved):
wild

Issue #70199, opened 2024-05-02
[Epic] Hydration Errors


ryan953:
Project Details: https://www.notion.so/sentry/FY-25-Q2-Hydration-Errors-3d5673e38a2743f8819cee8fdf94b3c4#2e5548009b7c43ebab4c7889910252c6
KTTR Items:
- https://www.notion.so/sentry/Hydration-Errors-in-Issue-Platform-5949fe74682b461c981f448ceaa2b322?pvs=23
- https://www.notion.so/sentry/Hydration-Errors-diff-tooling-936ba93dceff4fdea6c5cf543857ac86
- https://www.notion.so/sentry/Hydration-Errors-GA-219e258cfbdc43e1b7216c1a3b1b14ff

These new Issues are made visible via this flag: https://flagr.getsentry.net/#/flags/621

Relates to https://github.com/getsentry/sentry/issues/45664

```[tasklist]
### Core LA Tasks
- [ ] https://github.com/getsentry/sentry/issues/71392
- [ ] https://github.com/getsentry/sentry/issues/71395
- [ ] https://github.com/getsentry/sentry/issues/71396
- [ ] https://github.com/getsentry/sentry/issues/72067
- [ ] https://github.com/getsentry/sentry/issues/72068
- [x] https://sentry.sentry.io/issues/5451660180/?project=6520785&referrer=github-pr-bot
- [ ] https://github.com/getsentry/sentry/issues/72150
- [ ] https://github.com/getsentry/sentry/issues/72151
- [ ] https://github.com/getsentry/sentry/issues/72152
- [ ] https://github.com/getsentry/sentry/issues/72266
```

```[tasklist]
### Diff Tooling
- [ ] Add icons to diff-tool tabs, to differentiate text-based tools from image-based ones
- [x] Implement an image-diff tool where two images (replay views) are stacked atop each other, and a slider reveals parts of one or the other. AKA the "image slider" tool.
- [ ] https://github.com/getsentry/sentry/issues/74361
- [x] Replace the inline diff-tools tab bar with a single image-based tool on the Issue Details page, users can still click to open the modal which contains all diff tools.
- [ ] https://github.com/getsentry/sentry/issues/74358
- [ ] https://github.com/getsentry/sentry/issues/74360
- [ ] https://github.com/getsentry/sentry/issues/74458
```

```[tasklist]
### Fit & Finish Tasks - GA
- [ ] https://github.com/getsentry/sentry/issues/72493
[...]

PR #72169, opened 2024-06-05
feat(replay): Add breadcrumb event id as a tag in hydration issues

aliu39:
Ref https://github.com/getsentry/sentry/issues/70199


aliu3ntry:
I didn't understand the problem correctly, discussed on a call with @ryan953 -  no change is needed

PR #72171, opened 2024-06-05
deps(dnd-kit): Upgrade dnd-kit to latest version

malwilley:
We want to use this library for a new feature that will need drag and drop, but it's currently a few major versions behind.

This is only used in dashboards when you are creating a new widget with a timeseries chart. I've verified that it still works:

https://github.com/getsentry/sentry/assets/10888943/99cb5fb8-6f9a-4da3-9547-0badb67634a6


evanpurkhiser (approved):

static/app/views/dashboards/widgetBuilder/buildSteps/groupByStep/groupBySelector.tsx
```diff
@@ -123,7 +123,7 @@ export function GroupBySelector({
           <DndContext
             collisionDetection={closestCenter}
             onDragStart={({active}) => {
-              setActiveId(active.id);
+              setActiveId(active.id.toString());
```
malwilley, line 126: This was the only change necessary, the type for ID changed from `string` to `string | number`

scttcper (approved):
amazing

PR #72172, opened 2024-06-05
ref(ui): Strict group store state

scttcper:
- getstate returns the same object on every call (until state changes)
- saves the merged items as state that is updated in updateItems
- changes a few methods to Readonly otherwise they're mutating the wrong state


evanpurkhiser (approved):

static/app/stores/groupStore.tsx
```diff
@@ -206,8 +237,8 @@ const storeConfig: GroupStoreDefinition = {
     return -1;
   },
 
-  addActivity(id, data, index = -1) {
-    const group = this.get(id);
```
scttcper, line 210: this function is mutating the group, this.get() now returns the wrong item to mutate so i'm switching it to just this.items.find

PR #72176, opened 2024-06-05
feat(self-hosted): Gate features on project onboarding page for errors-only self-hosted

hubertdeng123:
Ensure that the project onboarding page disables all non-errors features.
1. Checkbox for performance, session replay, and profiling are disabled for configure SDK code
2. Button to navigate to performance and session replay pages are disabled

Screenshot 2024-06-05 at 2 44 38 PM


static/app/components/onboarding/productSelection.tsx
```diff
@@ -40,6 +41,7 @@ function getDisabledProducts(organization: Organization): DisabledProducts {
   const hasSessionReplay = organization.features.includes('session-replay');
   const hasPerformance = organization.features.includes('performance-view');
   const hasProfiling = organization.features.includes('profiling-view');
+  const isSelfHostedErrorsOnly = ConfigStore.get('isSelfHostedErrorsOnly');
 
   const reason = t('This feature is not enabled on your Sentry installation.');
```
azaslavsky, line 46: Can we not just remove the buttons entirely? I feel like by the time the user sees this, they will already be too far down the errors-only path (ie, they've selected it during install.sh) to do much about this now.

hubertdeng123: We can remove the buttons entirely, and I considered doing that. The reason I kept the buttons is that as time goes on and we add more features, it would indicate to users that a certain feature that is new is disabled. Maybe then they would make the tradeoff of going to full self-hosted Sentry or SaaS to include this feature.

static/app/components/onboarding/productSelection.tsx
```diff
@@ -48,6 +50,18 @@ function getDisabledProducts(organization: Organization): DisabledProducts {
...
+  if (isSelfHostedErrorsOnly) {
+    const products = Object.values(ProductSolution);
+    products.forEach(product => {
+      if (product !== ProductSolution.ERROR_MONITORING) {
+        disabledProducts[product] = {
+          reason,
+        };
+      }
+    });
+    return disabledProducts;
```
azaslavsky, line 62:
```suggestion
    return Object.values(ProductSolution)
      .filter(product => product !== ProductSolution.ERROR_MONITORING)
      .reduce((acc, prod) => {
        acc[prod] = {reason};
        return acc;
      }, {});
```

static/app/views/projectInstall/platform.spec.tsx
```diff
@@ -136,5 +138,46 @@ describe('ProjectInstallPlatform', function () {
         name: 'Configure Browser JavaScript SDK',
       })
     ).toBeInTheDocument();
+
+    expect(await screen.findByText('Take me to Issues')).toBeInTheDocument();
```
azaslavsky, line 142: Since we expect them all to appear, can we not use getByText for all 3?

hubertdeng123: ah yes, we can. changed!

static/app/components/onboarding/productSelection.tsx
hubertdeng123: done

azaslavsky (approved):

PR #72178, opened 2024-06-05
ref(custom-views): Return default view query if enabled

### LINKED ISSUE #71594: Custom Views: Augment the default search endpoint to use default from custom views table
The default search endpoint is called so that the initial issue stream page can be loaded with issues as fast as possible. This used to return the user's default saved search, if it exists. Since saved searches are being replaced with custom views, we will need to update this endpoint to return the query from the default custom view if the custom view feature flag is enabled.

MichaelSun48:
Closes #71594

This PR branches the issues/ endpoint, specifically the part that returns the query and query sort that the user will see upon loading the page. Before, it would use either the saved search you were currently on, or fallback to your pinned search if you were logging in fresh. 

Now, it will first check if you have custom views enabled as a feature, and if so, it will return the query and sort from either the view you are currently on (`selected_view_id`), or your default view. 

We will need to remember to populate the `groupsearchview_id` property on the frontend when making the request to this endpoint with custom views enabled

### DIFF
--- a/src/sentry/api/helpers/group_index/index.py
+++ b/src/sentry/api/helpers/group_index/index.py
@@ -90,7 +90,7 @@ def build_query_params_from_request(
     query = query.strip()
 
     if request.GET.get("savedSearch") == "0" and request.user and not has_query:
-        if features.has("organizations:issue-stream-custom-views"):
+        if features.has("organizations:issue-stream-custom-views", organization):
             selected_view_id = request.GET.get("groupSearchViewId")
             if selected_view_id:
                 default_view = GroupSearchView.objects.get(id=int(selected_view_id))
--- a/src/sentry/api/helpers/group_index/index.py
+++ b/src/sentry/api/helpers/group_index/index.py
@@ -93,15 +93,15 @@ def build_query_params_from_request(
         if features.has("organizations:issue-stream-custom-views", organization):
             selected_view_id = request.GET.get("searchId")
             if selected_view_id:
-                default_view = GroupSearchView.objects.get(id=int(selected_view_id))
+                default_view = GroupSearchView.objects.filter(id=int(selected_view_id)).first()
             else:
                 default_view = GroupSearchView.objects.filter(
                     organization=organization,
                     user_id=request.user.id,
                     position=0,
-                )
+                ).first()
 
-            if default_view.exists():
+            if default_view:
                 query_kwargs["sort_by"] = default_view.query_sort
                 query = default_view.query
         else:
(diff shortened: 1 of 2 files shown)


src/sentry/api/helpers/group_index/index.py
```diff
@@ -89,31 +90,46 @@ def build_query_params_from_request(
...
-            .extra(order_by=["name"])
-        )
-        selected_search_id = request.GET.get("searchId", None)
-        if selected_search_id:
-            # saved search requested by the id
-            saved_search = saved_searches.filter(id=int(selected_search_id)).first()
-        else:
-            # pinned saved search
-            saved_search = saved_searches.filter(visibility=Visibility.OWNER_PINNED).first()
+        if features.has("organizations:issue-stream-custom-views"):
```
scttcper, line 93: i think this takes the org as the 2nd parameter

MichaelSun48: 🤦 
Fixed, thanks for flagging!

scttcper (approved):
group index route is very important, probably needs a test

src/sentry/api/helpers/group_index/index.py
```diff
@@ -89,31 +90,46 @@ def build_query_params_from_request(
...
-        )
-        selected_search_id = request.GET.get("searchId", None)
-        if selected_search_id:
-            # saved search requested by the id
-            saved_search = saved_searches.filter(id=int(selected_search_id)).first()
-        else:
-            # pinned saved search
-            saved_search = saved_searches.filter(visibility=Visibility.OWNER_PINNED).first()
+        if features.has("organizations:issue-stream-custom-views", organization):
+            selected_view_id = request.GET.get("groupSearchViewId")
```
MichaelSun48, line 94: Make sure to populate the frontend's request with this `groupSearchViewId` parameter

src/sentry/api/helpers/group_index/index.py
```diff
@@ -89,31 +90,46 @@ def build_query_params_from_request(
...
+            if selected_view_id:
+                default_view = GroupSearchView.objects.get(id=int(selected_view_id))
+            else:
+                default_view = GroupSearchView.objects.get(
```
malwilley, line 98: Does this cover the case where the user has no default views (and we should give them the prioritized tab)?

malwilley: oh and I also think `.get()` will error here if it doesn't find anything which we don't want

MichaelSun48: I think the `.get()` is *technically fine since this part of the code should not be reachable if the user has no custom views, but that would have to be enforced by the frontend, so I think, to your point, it's probably still better to double check here as well.

malwilley: Maybe I'm forgetting how this works, but don't users start with no views? I thought that we only created views when they modified them.

MichaelSun48: @malwilley okay so after confirming with Richard I think my previous understanding of how this logic worked was incorrect - feel free to just ignore everything I just said above. I believe the most up to date version addresses your concerns and is the correct implementation now.

malwilley (approved):
Logic looks good!

PR #72179, opened 2024-06-05
feat(replay): Add Hydration Error Detected search title text

ryan953:
Noticed our new hydration error issue type was missing from this enum. The enum help populate help-docs within the search dropdown.

### DIFF
--- a/static/app/types/group.tsx
+++ b/static/app/types/group.tsx
@@ -97,6 +97,7 @@ export enum IssueType {
 
   // Replay
   REPLAY_RAGE_CLICK = 'replay_click_rage',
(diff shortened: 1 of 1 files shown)


ryan953:
For React 19 users this might be a good fit: https://www.notion.so/sentry/React-JS-19-23a7a2c5f88d4ef59a4b9647c84333a3

With that they could disable their inbound filter and start consuming error-quota and get hydration-errors. These would include useful stack-traces. If the React19 user also has replays installed, then they'd get replays attached.

michellewzhang (approved):

PR #72183, opened 2024-06-05
ref(similarity): Run single query for events with large event counts

jangjodi:
Modify record backfill script to only run single queries for groups with over 1 million events
Run bulk queries with modified timestamp for groups with under 1 million events


src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -108,102 +123,84 @@ def backfill_seer_grouping_records(
...
-        )
-        snuba_requests.append(request)
+    # Handle groups with over 1 million events
+    high_event_count_results = []
+    high_event_count_group_ids = [
+        group_id
+        for (group_id, _, times_seen, _) in group_values_batch
+        if times_seen >= HIGH_GROUP_EVENT_COUNT and group_id in group_ids_batch_filtered
+    ]
+    high_event_count_results = single_query_event_per_group(high_event_count_group_ids, project_id)
```
jangjodi, line 180: for groups with over 1 million events, run single snuba query per group

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -108,102 +123,84 @@ def backfill_seer_grouping_records(
...
+        group_id
+        for group_id in group_ids_batch_filtered
+        if group_id not in high_event_count_group_ids
+    ]
+    low_event_count_groups_last_seen = [
+        last_seen
+        for (group_id, _, _, last_seen) in group_values_batch
+        if group_id in low_event_count_group_ids
+    ]
+    low_event_count_results = bulk_query_events(
```
jangjodi, line 193: for groups with less than 1 million events, run the bulk snuba query

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -554,6 +551,114 @@ def delete_seer_grouping_records(
...
 
+def single_query_event_per_group(
+    high_event_count_group_ids: list[int], project_id: int
+) -> list[EventGroupSnubaResult]:
+    """
+    Query events for a group one at a time, running the requests in groups of SNUBA_QUERY_RATELIMIT.
+    """
+    group_event_rows: list[EventGroupSnubaResult] = []
+    events_entity = Entity("events", alias="events")
+    for group_ids_chunk in chunked(high_event_count_group_ids, SNUBA_QUERY_RATELIMIT):
```
jangjodi, line 562: in chunks groups of size SNUBA_QUERY_RATELIMIT (4), run the event query per group.
make a single request of the 4 queries, appending the results to a list

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -554,6 +551,114 @@ def delete_seer_grouping_records(
...
+            Condition(Column("group_id"), Op.IN, low_event_count_group_ids),
+            Condition(
+                Column("timestamp", entity=events_entity),
+                Op.GTE,
+                min(low_event_count_groups_last_seen) - timedelta(minutes=5),
+            ),
+            Condition(
+                Column("timestamp", entity=events_entity),
+                Op.LT,
+                max(low_event_count_groups_last_seen) + timedelta(minutes=5),
```
jangjodi, line 638: the regular bulk snuba event query, but modify the timestamp condition to only look between the lowest and highest `times_seen` of the batch (+- 5 mins) instead of querying all 90 days

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -115,13 +135,23 @@ def backfill_seer_grouping_records(
...
     last_group_id = groups_to_backfill_with_no_embedding[-1]
 
-    snuba_results = get_data_from_snuba(project, groups_to_backfill_with_no_embedding)
+    # Run single snuba query per group for high event counts, otherwise run bulk query
+    snuba_results_high_count = get_data_from_snuba_single_group_query(
+        project, group_id_last_seen_no_embeddings_high_count
+    )
+    snuba_results_low_count = get_data_from_snuba_bulk_groups_query(
+        project, group_id_last_seen_no_embeddings_low_count
+    )
```
JoshFerge, line 144: I think it would be cleaner if you put this logic into one function which this main function would call, and then you'd combine the results in that function. so that way further down in the logic, it doesn't have to care what a high result / low result is, it's all snuba results that are treated the same

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -115,13 +135,23 @@ def backfill_seer_grouping_records(
...
 
     (
         filtered_snuba_results,
         groups_to_backfill_with_no_embedding_has_snuba_row,
-    ) = filter_snuba_results(snuba_results, groups_to_backfill_with_no_embedding, project)
-
+    ) = filter_snuba_results(
+        snuba_results_high_count,
+        snuba_results_low_count,
```
JoshFerge, line 151: so this function's arguments could remain the same if you do the above refactor

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -61,6 +74,11 @@ class GroupStacktraceData(TypedDict):
     stacktrace_list: list[str]
 
 
+class EventGroupSnubaResult(TypedDict):
```
JoshFerge, line 77: this doesn't appear to be used?

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -342,13 +398,70 @@ def get_data_from_snuba(project, groups_to_backfill_with_no_embedding):
...
+    groups_last_seen = [
+        group_id_last_seen_no_embeddings_low_count[group_id]
+        for group_id in group_id_last_seen_no_embeddings_low_count
+    ]
+    group_ids = [group_id for group_id in group_id_last_seen_no_embeddings_low_count]
+    query = Query(
+        match=events_entity,
+        select=[
+            Column("group_id"),
+            Function("max", [Column("event_id")], "event_id"),
```
JoshFerge, line 427: can we change the function to be `any` instead of `max`?

JoshFerge:
and one final request: could we make it so that the bulk queries are controled by an option? e.g. if the option is false, we only run the single event queries.

if the option is true, we'll try to do the bulk as well

i have a feeling we may run into snuba issues even if we filter out high event counts from the bulk query.

PR #72184, opened 2024-06-05
ref(replay): Refactor the replayCompareModal

ryan953:
so we can re-use the diff tools inline on another page, outside of any modal.

Relates to https://github.com/getsentry/sentry/issues/72150

### DIFF
--- a/static/app/components/replays/replayDiff.tsx
+++ b/static/app/components/replays/replayDiff.tsx
@@ -1,0 +1,210 @@
+import {Fragment, useEffect, useState} from 'react';
+import styled from '@emotion/styled';
+import beautify from 'js-beautify';
+
+import Alert from 'sentry/components/alert';
+import {Flex} from 'sentry/components/container/flex';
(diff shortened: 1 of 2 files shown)


static/app/components/replays/replayDiff.tsx
```diff
@@ -0,0 +1,210 @@
+import {Fragment, useEffect, useState} from 'react';
+import styled from '@emotion/styled';
+import beautify from 'js-beautify';
```
scttcper, line 3: wherever we end up using this we'll still want to be lazy loaded since js-beautify and diff are pretty big

scttcper (approved):

static/app/components/replays/replayDiff.tsx
ryan953: ya for sure. It'll be on Issue Details, so I'll tap you again for that PR.

I think the lazy-load boundary will basically include the `useReplayData()` call, and then return this component to render. modulo there will be some wrapper divs/panels to make it look nice, but hopefully that's the extent of the lazy-loaded wrapper for the diff component.

PR #72185, opened 2024-06-05
ref(issue-stream): Remove feature flag for event stats graph and assignee selector

MichaelSun48:
This PR unregisters and removes all instances of the feature flags used to rollout the new event stats graph and assignee selector. It has been EA'd for a week and has been in GA since Monday without any issues. I will merge this and the associated options PR later tomorrow, 06/06.

[Associated options PR](https://github.com/getsentry/sentry-options-automator/pull/1556)

### DIFF
--- a/src/sentry/features/temporary.py
+++ b/src/sentry/features/temporary.py
@@ -154,10 +154,6 @@ def register_temporary_features(manager: FeatureManager):
     manager.add("organizations:issue-stream-custom-views", OrganizationFeature, FeatureHandlerStrategy.OPTIONS)
     # Enable the updated empty state for issues
     manager.add("organizations:issue-stream-empty-state", OrganizationFeature, FeatureHandlerStrategy.REMOTE)
-    # Enable the new assignee dropdown trigger on issue stream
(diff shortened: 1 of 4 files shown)


malwilley (commented):
@MichaelSun48 you actually want to split this up into two PRs - one to remove the references to the flag, and another to remove the flag definition itself.

Because the frontend and backend are deployed separately, we want to first ensure that the frontend deploy isn't referencing the flag before merging the backend PR that removes it completely. And I don't quite remember the order required with options, but I think you want to unregister the option before removing the flag definition from sentry.

MichaelSun48:
>you actually want to split this up into two PRs - one to remove the references to the flag, and another to remove the flag definition itself.

>Because the frontend and backend are deployed separately, we want to first ensure that the frontend deploy isn't referencing the flag before merging the backend PR that removes it completely. And I don't quite remember the order required with options, but I think you want to unregister the option before removing the flag definition from sentry.

That makes sense, thanks for pointing that out! I'll close this PR now and put up two separate ones.

PR #72186, opened 2024-06-06
ref(messaging): Use `ExternalProviderEnum.value`

iamrajjoshi:
This PR is required to fix the tests in https://github.com/getsentry/getsentry/actions/runs/9389683374/job/25875213718?pr=14211

Because we use the actual Enum class in some of our logic, when I attempt to use the new copied enum class that i created under the integration folder, we actually have a mismatch in our code as `getsentry` will use the newer class while `sentry` won't. 

Specifically in this [line](https://github.com/getsentry/sentry/blob/master/src/sentry/notifications/notificationcontroller.py#L602),  because the defaultdict is keyed with the Enum class (and not the value), the same enum in a different module will actually not exist in the dict 😢 

We need to use the actual enum value (which is a str), which would be the same for both classes.

thanks to @ykamo001  for helping out!

### DIFF
--- a/src/sentry/integrations/types.py
+++ b/src/sentry/integrations/types.py
@@ -51,6 +51,8 @@ EXTERNAL_PROVIDERS_REVERSE = {
     ExternalProviderEnum.CUSTOM: ExternalProviders.CUSTOM,
 }
 
+EXTERNAL_PROVIDERS_REVERSE_VALUES = {k.value: v for k, v in EXTERNAL_PROVIDERS_REVERSE.items()}
+
 EXTERNAL_PROVIDERS = {
     ExternalProviders.EMAIL: ExternalProviderEnum.EMAIL.value,
     ExternalProviders.SLACK: ExternalProviderEnum.SLACK.value,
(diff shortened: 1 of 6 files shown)


src/sentry/integrations/types.py
```diff
@@ -51,6 +51,8 @@ class ExternalProviderEnum(Enum):
     ExternalProviderEnum.CUSTOM: ExternalProviders.CUSTOM,
 }
 
+EXTERNAL_PROVIDERS_REVERSE_VALUES = {k.value: v for k, v in EXTERNAL_PROVIDERS_REVERSE.items()}
```
iamrajjoshi, line 54: created this temporarily, will eventually put a wrapper function/remove this once we completely get rid of old imports.

src/sentry/types/integrations.py
```diff
@@ -52,6 +52,10 @@ class ExternalProviderEnum(Enum):
     ExternalProviderEnum.CUSTOM: ExternalProviders.CUSTOM,
 }
 
+
+EXTERNAL_PROVIDERS_REVERSE_VALUES = {k.value: v for k, v in EXTERNAL_PROVIDERS_REVERSE.items()}
```
iamrajjoshi, line 56: i also needed to update the old types b/c currently `sentry` will use this, while the pr i mentioned in the description will use the symbols in the  file under integrations folder.

ykamo001 (approved):

PR #72187, opened 2024-06-06
ref(reports): Assign batch IDs to investigate duplicate sends

RyanSkonnord:
Refactor to assign a batch ID to each run of `schedule_organizations`.

Store batch IDs in Redis and check for duplicate send operations to the same organization and user. We hope to gain information about the cause of the duplicate sends. It may also provide a good way to detect and interrupt them if there is no better solution.


src/sentry/tasks/summaries/weekly_reports.py
```diff
@@ -211,76 +224,123 @@ def prepare_organization_report(
...
+    def _prevent_duplicate_delivery(self) -> None:
+        from sentry.utils import redis  # TODO: Use sentry.buffer.redis instead?
+
+        cluster = redis.clusters.get("default").get_local_client_for_key("weekly_reports")
+        key = f"{self.batch_id}:{self.ctx.organization.id}"
+        count = cluster.incr(key)
+        cluster.expire(key, timedelta(days=1))
+
+        if count > 1:
+            raise Exception(f"Duplicate report delivery detected for batch {self.batch_id}")
```
RyanSkonnord, line 257: Ideally this shouldn't be hit at all. Our hypothesis is that it may be hit regardless. Raising an exception would prevent users from getting spammed with extra messages until we deliver a fix.

However, it incurs some risk that it could incorrectly prevent the *first* delivery. If we want to be more cautious, we could downgrade it to an error log and see if it gets hit at all.

src/sentry/tasks/summaries/weekly_reports.py
```diff
@@ -211,76 +224,127 @@ def prepare_organization_report(
...
+
+    dry_run: bool = False
+    target_user: User | None = None
+    email_override: str | None = None
+
+    def _prevent_duplicate_delivery(self) -> None:
+        from sentry.utils import redis  # TODO: Use sentry.buffer.redis instead?
+
+        cluster = redis.clusters.get("default").get_local_client_for_key("weekly_reports")
+        key = f"{self.batch_id}:{self.ctx.organization.id}"
```
ceorourke, line 255: I can try to dig through the data from 2 weeks ago, but I'm curious if we know if this happens to an org, does every user get duplicate sends? Or should we also store the user id to filter to a potentially smaller list of user ids to send to?

src/sentry/tasks/summaries/weekly_reports.py
```diff
@@ -211,76 +224,127 @@ def prepare_organization_report(
...
+            user_list = list(
+                OrganizationMember.objects.filter(
+                    user_is_active=True,
+                    organization_id=self.ctx.organization.id,
                 )
-    else:
-        user_list = list(
-            OrganizationMember.objects.filter(
-                user_is_active=True,
-                organization_id=ctx.organization.id,
```
ceorourke, line 262: Don't we still need to filter by active users?

src/sentry/tasks/summaries/weekly_reports.py
```diff
@@ -211,76 +224,127 @@ def prepare_organization_report(
...
+            )
+
+            # TODO see if we can use the UUID to track if the email was sent or not
+            logger.info(
+                "weekly_report.send_email",
+                extra={
+                    "batch_id": self.batch_id,
+                    "organization": self.ctx.organization.id,
+                    "uuid": template_ctx["notification_uuid"],
+                },
```
ceorourke, line 343: Can you add the user id here? When I was looking at these I was wishing we had it

src/sentry/tasks/summaries/weekly_reports.py
RyanSkonnord: Yes, we have a few options as to where in the flow we want to call this. We could definitely store a key per user, provided that won't take up too much space in Redis.

ceorourke: I can try to pull how many users it would be but I imagine it's quite a large number 😬 Do you know how to find out the limits of what we can put in Redis?

RyanSkonnord:
With https://github.com/getsentry/sentry/pull/72187/commits/8126c5ad940993e4a9d31bb224422020c7fc9d1a, this should be in approximately the state we want to ship it in:

1. The check is now done per individual user in each org.
2. Detected duplicates will only cause errors to be logged. We won't raise any exceptions until we know more about what's going on.

The new checks still need unit test coverage, which I'll be working on next.

src/sentry/tasks/summaries/weekly_reports.py
```diff
@@ -211,76 +225,179 @@ def prepare_organization_report(
...
+            )
+
+            message.add_users((user_id,))
+            message.send_async()
+
+        delivery_record.record_delivery()
+
+
+@dataclass(frozen=True)
+class _ReportDeliveryRecord:
```
ceorourke, line 344: I really like how this whole class was implemented :)

src/sentry/tasks/summaries/weekly_reports.py
ceorourke: For posterity I pulled this and it's around 3m users, we verified over Slack that Redis will handle this just fine. 

I made a [Redash query](https://redash.getsentry.net/queries/6899/source) to validate that we don't necessarily see the same number of duplicates across an organization (and cross referenced `mail.queued` logs and checked the counts per user for the weekly report email) so I feel good about us using the user id in the Redis key as well.

ceorourke (approved):

PR #72197, opened 2024-06-06
fix(screen-loads): Align TTID/TTFD/count naming

philipphofmann:
Use avg TTID / avg TTFD and total count to algin the naming with the screens landing page, as different namings can confuse users.


gggritso (approved):
Funny enough, in Insights modules we usually use "Avg" rather than "AVG", but some of the UIs auto-capitalize it? Anyway, I leave it to you, but "Avg" is more common!

PR #72198, opened 2024-06-06
fix(screen-loads): Prefix duration with AVG

philipphofmann:
Prefix the duration for the span summary with avg to clearly communicate the duration is an average.


gggritso (approved):
Should be `"Avg"` to match the rest, I think, but fine to fix in a separate PR if you wish

PR #72212, opened 2024-06-06
fix(app-start): Display correct change color

philipphofmann:
The app starts module screen summary page displayed the wrong change color in the metrics readout. This is fixed now.

| Before | After |
|--------|--------|
| CleanShot 2024-06-06 at 14 42 32@2x | CleanShot 2024-06-06 at 14 45 10@2x |


gggritso (approved):
Nice!

PR #72214, opened 2024-06-06
fix(profiling): improve copy source precision by including lineno and colno

JonasBa:


static/app/components/profiling/flamegraph/flamegraphZoomView.tsx
```diff
@@ -55,6 +55,13 @@ function isHighlightingAllOccurrences(
   );
 }
 
+function makeSourceCodeLink(frame: FlamegraphFrame['frame']): string | undefined {
+  const path = frame.path || frame.file;
+  const lineComponents = [frame.line, frame.column]
```
viglia, line 60: nit: if `line` is undefined but `col` is not (or vice versa), how do we know what that number represents? it could be `path (col)` or `path (line)`

Or that case can not happen?

JonasBa: No, you are correct. I think it can happen in theory, but I doubt it does in practice because iirc colno is line relative. I'll add a guard from this just in case though

Zylphrex (approved):

Issue #72219, opened 2024-06-06
Problems with Org Member APIs


Dhrumil-Sentry:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

Question from Linear:

Two issues. We were previously using `/organizations/<org>/users?project=<project>` to list members which appears to be an unsupported api (I can't find documentation for it). This is some old code, so probably deprecated at some point? We are getting errors on ~10% of requests. Errors like this: `invalid json response body at ... reason: Unexpected end of JSON input`. Feels reminiscent of the previous error you had related to invalid signatures, which as I recall was related to some JSON handling (and which also had a roughly 10% error rate?).
Anyway. I tried to migrate us to `organizations/<org>/members` which [appears to be the supported api.](https://docs.sentry.io/api/organizations/list-an-organizations-members/) But I have two issues with it. All members are supposed to have a `user` property, but I'm getting back `null` very regularly and I'm also still getting the same JSON error message.
For now I'll revert to `/organizations/<org>/users?project=<project>`, but would love some help here.

### Expected Result

.

### Actual Result

.

### Product Area

Settings - Members

PR #72222, opened 2024-06-06
fix(api) Correct and expand org members API docs

markstory:
- User is optional as invites don't have users yet
- Add an example including an invite.

Refs #72219


isabellaenriquez (approved):

isabellaenriquez:
after test fixes lol

PR #72242, opened 2024-06-06
fix: use explicit webpack define

anonrig:
We need to be explicit when calling define plugin. The proposed change is more correct than previous one.

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -348,15 +348,13 @@ const appConfig: Configuration = {
      * Defines environment specific flags.
      */
     new webpack.DefinePlugin({
-      'process.env': {
-        NODE_ENV: JSON.stringify(env.NODE_ENV),
-        IS_ACCEPTANCE_TEST: JSON.stringify(IS_ACCEPTANCE_TEST),
-        DEPLOY_PREVIEW_CONFIG: JSON.stringify(DEPLOY_PREVIEW_CONFIG),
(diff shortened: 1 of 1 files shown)


webpack.config.ts
```diff
@@ -348,15 +348,13 @@ const appConfig: Configuration = {
...
-        SENTRY_RELEASE_VERSION: JSON.stringify(SENTRY_RELEASE_VERSION),
-        USE_REACT_QUERY_DEVTOOL: JSON.stringify(USE_REACT_QUERY_DEVTOOL),
-      },
+      'process.env.IS_ACCEPTANCE_TEST': JSON.stringify(IS_ACCEPTANCE_TEST),
+      'process.env.NODE_ENV': JSON.stringify(env.NODE_ENV),
+      'process.env.DEPLOY_PREVIEW_CONFIG': JSON.stringify(DEPLOY_PREVIEW_CONFIG),
+      'process.env.EXPERIMENTAL_SPA': JSON.stringify(SENTRY_EXPERIMENTAL_SPA),
+      'SPA_DSN': JSON.stringify(SENTRY_SPA_DSN),
+      'SENTRY_RELEASE_VERSION': JSON.stringify(SENTRY_RELEASE_VERSION),
+      'USE_REACT_QUERY_DEVTOOL': JSON.stringify(USE_REACT_QUERY_DEVTOOL),
```
evanpurkhiser, line 357: Why don't these get the prefix?

anonrig: Oops. Fixed. Thanks!

scttcper (approved):

PR #72244, opened 2024-06-06
refactor: move ts-fork options to webpack

anonrig:
I didn't understand why we defined this in package.json rather than in webpack.config where it belongs.

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -372,9 +372,9 @@ const appConfig: Configuration = {
               configOverwrite: {
                 compilerOptions: {incremental: true},
               },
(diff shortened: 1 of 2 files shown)


webpack.config.ts
```diff
@@ -372,9 +372,9 @@ const appConfig: Configuration = {
               configOverwrite: {
                 compilerOptions: {incremental: true},
```
scttcper, line 373: fun fact, incremental doesn't actually work because its not setup correctly. i think it needs a path for where to put that file or something

scttcper (approved):

PR #72250, opened 2024-06-06
ref(similarity): Add more logs to backfill

jangjodi:
Add more logs to backfill since it's stopping without any logs


src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -200,7 +200,9 @@ def backfill_seer_grouping_records(
 
     group_id_batch_all = copy.deepcopy(group_id_batch_filtered)
     if snuba_results and snuba_results[0].get("data"):
-        rows: list[GroupEventRow] = [snuba_result["data"][0] for snuba_result in snuba_results]
+        rows: list[GroupEventRow] = [
+            snuba_result["data"][0] for snuba_result in snuba_results if snuba_result["data"]
```
jangjodi, line 204: empty list check that I caught while testing locally

snigdhas (approved):

PR #72252, opened 2024-06-06
feat(uptime): Add do-nothing uptime results consumer

evanpurkhiser:
This reads and decodes results from the uptime-results topic and just logs them

### DIFF
--- a/src/sentry/conf/server.py
+++ b/src/sentry/conf/server.py
@@ -2255,6 +2255,9 @@ SENTRY_USE_REPLAY_ANALYZER_SERVICE = False
 # This flag activates Spotlight Sidecar in the development environment
 SENTRY_USE_SPOTLIGHT = False
 
+# This flag activates uptime checks in the developemnt environment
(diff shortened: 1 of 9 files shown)


src/sentry/conf/server.py
```diff
@@ -2255,6 +2255,9 @@ def custom_parameter_sort(parameter: dict) -> tuple[str, int]:
 # This flag activates Spotlight Sidecar in the development environment
 SENTRY_USE_SPOTLIGHT = False
 
+# This flag activates uptime checks in the developemnt environment
+SENTRY_USE_UPTIME = False
```
evanpurkhiser, line 2259: We don't have one for monitors, I wonder if we should

wedamija: Yeah I'm not sure. It's kind of a pain to have to know to enable a bunch of settings just to get a feature working, but on the other hand nicer to not have to spin up a bunch of unnecessary services

wedamija (approved):

PR #72253, opened 2024-06-06
ref(error_eventype): Easier to read

armenzg:
This changes includes:

* Include the module for stronger typing
* Create more functions to reduce main function complexity

### DIFF
--- a/src/sentry/eventtypes/error.py
+++ b/src/sentry/eventtypes/error.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from collections.abc import MutableMapping
 from typing import Any
 
--- a/src/sentry/eventtypes/error.py
+++ b/src/sentry/eventtypes/error.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 from collections.abc import MutableMapping
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


src/sentry/eventtypes/error.py
```diff
@@ -22,42 +21,44 @@ def get_crash_location(data):
...
-        # If the event data has been marked with a main_exception_id, then we should be able to
-        # find the exception with the matching metadata.exception_id and use that one.
-        # This can be the case for some exception groups.
-
-        # Otherwise, the default behavior is to use the last one in the list.
-
+        # With chained exceptions, the SDK could set the main_exception_id to
+        # indicate which exception to use for title & subtitle
         main_exception_id = get_path(data, "main_exception_id")
-        exception = (
```
armenzg, line 44: Moved this logic into `get_exception`
image

src/sentry/eventtypes/error.py
```diff
@@ -22,42 +21,44 @@ def get_crash_location(data):
...
+    "dart-flutter",
+    "unity",
+    "dotnet-xamarin",
+)
 
 
 class ErrorEvent(BaseEvent):
     key = "error"
 
     def extract_metadata(self, data: MutableMapping[str, Any]) -> Metadata:
```
armenzg, line 50: I'm mostly making this function easier to read.

src/sentry/eventtypes/error.py
```diff
@@ -1,5 +1,3 @@
-from __future__ import annotations
```
vartec, line 1: This hasn't been mainlined, in fact it never will, so it's still needed for deferred annotations.

vartec (approved):

src/sentry/eventtypes/error.py
```diff
@@ -22,42 +23,44 @@ def get_crash_location(data):
 
         func = get_function_name_for_frame(frame, data.get("platform"))
         return frame.get("filename") or frame.get("abs_path"), func
+    return None
+
+
+NATIVE_MOBILE_PLATFORMS = (
```
vartec, line 29: nit: we really should put these types of constants in one well defined place, we keep on redefining them in bunch of places
 https://github.com/getsentry/sentry/blob/88d3b9eedda90f6d710a431ad841220aa0f35f56/src/sentry/stacktraces/platform.py
 
https://github.com/getsentry/relay/blob/a92fc6a6f5538c74a9344f9bd458bcbc93306ccb/relay-event-schema/src/protocol/constants.rs#L2

asottile-sentry (approved):

PR #72255, opened 2024-06-06
chore(insights): Remove `spans-first-ui` flag registration

gggritso:
Not in use. In da bin.


Zylphrex (approved):

aldy505:
Should we also remove the one defined on self-hosted too? https://github.com/getsentry/self-hosted/blob/419d6ccf4e4d0dde2db301e43370d019f9a4c69b/sentry/sentry.conf.example.py#L303

gggritso:
@aldy505 yes, that makes sense! No harm in keeping it around, but it's no in use any more

PR #72256, opened 2024-06-06
ref: fix typing for tagstore

asottile-sentry:
signatures mismatched between baseclass and snuba implementation


src/sentry/api/endpoints/group_tagkey_details.py
```diff
@@ -49,7 +49,7 @@ def get(self, request: Request, group, key) -> Response:
 
         if group_tag_key.count is None:
             group_tag_key.count = tagstore.backend.get_group_tag_value_count(
-                group.project_id, group.id, environment_id, lookup_key, tenant_ids=tenant_ids
+                group, environment_id, lookup_key, tenant_ids=tenant_ids
             )
```
asottile-sentry, line 53: this call couldn't have succeeded as written -- I don't see any errors in prod so perhaps this codepath is dead?

evanh (approved):

PR #72262, opened 2024-06-06
ref(messaging): Updated Integration Types Imports

iamrajjoshi:
Final PR to update the rest of the integration imports in sentry over to the new module that I created. I edit 1 `__init__.py` file, plan is to clean it up completely in a followup pr.

### DIFF
--- a/src/sentry/tasks/integrations/__init__.py
+++ b/src/sentry/tasks/integrations/__init__.py
@@ -54,7 +54,6 @@ settings.CELERY_IMPORTS += (
 
 from .create_comment import create_comment
 from .kick_off_status_syncs_impl import kick_off_status_syncs
-from .migrate_issues import migrate_issues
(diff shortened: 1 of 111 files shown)


src/sentry/tasks/integrations/__init__.py
```diff
@@ -54,7 +54,6 @@ def should_comment_sync(
 
 from .create_comment import create_comment
 from .kick_off_status_syncs_impl import kick_off_status_syncs
-from .migrate_issues import migrate_issues
```
cathteng, line 57: does this need to be removed?

iamrajjoshi: it causes circular dependency, plan is to delete all these imports so other files so specific ones.

ykamo001 (approved):

Issue #71998, opened 2024-06-04
Issue unexpectedly marked as "resolved due to inactivity" after escalating


nbrustein:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

I don't know how to reproduce this issue. Looking at the activity tab for [this issue](https://pedago.sentry.io/issues/5392087181/activity/?alert_rule_id=751720&alert_type=issue&environment=Production+-+Client&notification_uuid=2ad4c6ca-5c14-4938-bf93-4cabbb31fed2&project=1491374&referrer=slack), I see

 * May 23, 1:44pm: Nate Brustein archived this issue until it happens 1 time(s)
 * June 3, 3:27pm: Sentry flagged this issue as escalating because 1 event happened
 * June 3, 3:27pm: Sentry updated the priority value of this issue to be high after it escalated
 * **June 3, 3:30pm: Sentry marked this issue as resolved due to inactivity**

### Expected Result

The issue should not be marked as "resolved due to inactivity minutes after escalating.

### Actual Result

The issue was marked as "resolved due to inactivity minutes after escalating.

https://pedago.sentry.io/issues/5392087181/activity/?alert_rule_id=751720&alert_type=issue&environment=Production+-+Client&notification_uuid=2ad4c6ca-5c14-4938-bf93-4cabbb31fed2&project=1491374&referrer=slack

### Product Area

Issues

### Link

https://pedago.sentry.io/issues/5392087181/activity/?alert_rule_id=751720&alert_type=issue&environment=Production+-+Client&notification_uuid=2ad4c6ca-5c14-4938-bf93-4cabbb31fed2&project=1491374&referrer=slack

snigdhas:
Hey @nbrustein thanks for reporting this! It's definitely a bug in our auto-resolution logic and seems to be hitting a race condition with the issue auto-transitioning to `escalating`. I've brought this to our team's attention and will keep the ticket updated as we work towards a fix.

snigdhas:
Closing this as it should have been fixed in #72269 - please reopen if you notice this happen again!

PR #72269, opened 2024-06-06
fix(auto-resolve0: Check Group.last_seen before performing auto-resolve update

snigdhas:
Issues recently marked as escalating are sometimes auto-resolved sooner than the resolve_age. This is likely because we buffer updating last_seen and the auto_resolve task (runs every 15m) is working off of an old last_seen. 

Adding a second check to make sure we're only updating issues with a last_seen older than the resolve age, in case the value is updated between the when we query the groups to update and perform the actual update.

This is likely only a partial fix for https://github.com/getsentry/sentry/issues/71998.


wedamija (approved):
Yeah this will only catch a small number of edge cases I think, likely fetching the value from buffer will be more effective

PR #72270, opened 2024-06-06
ref(crons): Remove analytics for processing error storage

davidenwang:
These analytics were put in place so that we can verify everything was alright as we rolled out processing error storage for 10% of customers.

After instrumentation it does seem like the volume of analytics would be much too high (already almost at 1 mil events per day) if we kept it in a scenario with a 100% rollout.

Instead lets just stick to using the datadog metric, and maybe add a new tag there to count based on processing error type.


wedamija (approved):
If these are useful you could also sample them and just write 1%, but if not removing is fine

davidenwang:
> If these are useful you could also sample them and just write 1%, but if not removing is fine

@wedamija oh didn't really consider this, is this an option we pass to analytics.record? Trying to find examples but failing haha

wedamija:
> > If these are useful you could also sample them and just write 1%, but if not removing is fine
> 
> @wedamija oh didn't really consider this, is this an option we pass to analytics.record? Trying to find examples but failing haha

I don't think it has an option, but you can just write a simple thing yourself like `if random.random() < 0.01: send`

davidenwang:
> > > If these are useful you could also sample them and just write 1%, but if not removing is fine
> > 
> > 
> > @wedamija oh didn't really consider this, is this an option we pass to analytics.record? Trying to find examples but failing haha
> 
> I don't think it has an option, but you can just write a simple thing yourself like `if random.random() < 0.01: send`

oooh haha i see, yea in that case lemme opt for that, will make a new PR

PR #72272, opened 2024-06-06
ref: fix typing of EventStream

asottile-sentry:
fixes errors when enabling type checking of LazyServiceWrapper

### DIFF
--- a/src/sentry/eventstream/snuba.py
+++ b/src/sentry/eventstream/snuba.py
@@ -89,7 +89,7 @@ class SnubaProtocolEventStream(EventStream):
         is_regression: bool,
         is_new_group_environment: bool,
         primary_hash: str | None,
-        received_timestamp: float,
+        received_timestamp: float | datetime,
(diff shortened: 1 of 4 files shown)


src/sentry/eventstream/snuba.py
```diff
@@ -288,9 +286,9 @@ def end_unmerge(self, state: Mapping[str, Any]) -> None:
             state_copy["project_id"], "end_unmerge", extra_data=(state_copy,), asynchronous=False
         )
 
-    def start_delete_tag(self, project_id: int, tag: str) -> Mapping[str, Any] | None:
+    def start_delete_tag(self, project_id: int, tag: str) -> Mapping[str, Any]:
         if not tag:
-            return None
```
asottile-sentry, line 293: these `None`s (same as above) would immediately crash in the `end_*` methods so I changed this to an error and removed the nullability

joshuarli (approved):

PR #72279, opened 2024-06-06
feat(search-query-builder): Add tabbed empty state for discovering filter keys

### LINKED ISSUE #69723: Evolved Issues Search: Add improved empty state for discovering filter keys

malwilley:
Closes https://github.com/getsentry/sentry/issues/69723

This adds a tabbed version of the combobox menu when the user has not yet started searching for a specific key. This list is configurable with the new `filterKeySections` prop, which replaces the old `supportedKeys` prop.

CleanShot 2024-06-06 at 14 30 16

### DIFF
--- a/static/app/components/searchQueryBuilder/input.tsx
+++ b/static/app/components/searchQueryBuilder/input.tsx
@@ -5,17 +5,16 @@ import {Item, Section} from '@react-stately/collections';
 import type {ListState} from '@react-stately/list';
 import type {Node} from '@react-types/shared';
 
-import type {SelectOptionWithKey} from 'sentry/components/compactSelect/types';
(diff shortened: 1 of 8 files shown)


static/app/components/searchQueryBuilder/input.tsx
```diff
@@ -292,7 +269,7 @@ function SearchQueryBuilderInputInternal({
 
   return (
     <SearchQueryBuilderCombobox
-      items={items}
+      items={sections}
```
malwilley, line 272: We were originally sending in a flat list of items, but now are sending list of sections with child items. Below we are using a callback render function, which is why I needed to make the component props generic.

scttcper (approved):

PR #72280, opened 2024-06-06
feat(priority-alerts): Add analytic event for alert rule row deleted

snigdhas:
Add an analytic event `edit_alert_rule.delete_row` to track when rows are deleted from an alert rule in the edit view. This will help us track how often users remove the new high priority alert conditions.


scttcper (approved):
SEER

PR #72281, opened 2024-06-06
ref(replay): Refactor eventReplay to split onboarding and eligibility checks from preview render

ryan953:
Refactor all the things!

To read this you wanna start at `export default function EventReplay` inside index.tsx, checkout all the if-statements as you go into `function EventReplayContent()` within the same file. Compare that flow to the new `export default function EventReplay()` which flows into `replayClipSection.tsx`. All the conditions should be logically the same.

We've got all the onboarding and guards and stuff inside index now... and the actual meat of the rendering is in the new `replayClipSection.tsx`

After this change I'm going to followup with a new `replayDiffSection` that can rely on the shared guards/onboarding stuff happening first.


scttcper (approved):

static/app/components/events/eventReplay/index.tsx
```diff
@@ -1,63 +1,50 @@
...
+
+  const {hasOrgSentReplays, fetching: fetchingHasSentReplays} =
+    useHasOrganizationSentAnyReplayEvents();
+  const project = useProjectFromSlug({organization, projectSlug});
+
+  const replayId = getReplayIdFromEvent(event);
 
-  if (fetching) {
-    return null;
+  if (!hasReplaysFeature || fetchingHasSentReplays) {
```
michellewzhang, line 33: what happens if we return `false` here? is it the same as returning `null`?

ryan953: Meant to be null. Caught me copy paste again! 

Null would be better.

static/app/components/events/eventReplay/index.tsx
```diff
@@ -1,63 +1,50 @@
...
+    return false;
```
ryan953, line 34:
```suggestion
    return null;
```

PR #72296, opened 2024-06-07
feat(chartcuterie): Add image to email

sentaur-athena:
Depends on : https://github.com/getsentry/sentry/pull/72134

I'm adding a very generic component to all alert emails for chart_image. For now though these images only exist for endpoint regression and function regression. For everything else `get_image_url()` return none.

There's no test endpoint regression issue in dev so I tested with another performance issue and faked the data to get this image.
Screenshot 2024-06-06 at 4 44 05 PM


src/sentry/notifications/notifications/rules.py
```diff
@@ -39,6 +39,7 @@
 from sentry.services.hybrid_cloud.user_option.service import get_option_from_list
 from sentry.types.actor import Actor
 from sentry.types.group import GroupSubStatus
+from sentry.types.integrations import ExternalProviderEnum, ExternalProviders
```
iamrajjoshi, line 42:
```suggestion
from sentry.integrations.types import ExternalProviderEnum, ExternalProviders
```
i moved the imports as part of the new messaging refactor, will delete the old module so folks don't get confused.

iamrajjoshi (approved):

src/sentry/templates/sentry/emails/issue_alert_base.html
```diff
@@ -169,6 +169,9 @@ <h3>
         {% endif %}
       </div>
 
+      {% if chart_image %}
+        <img src="{{chart_image}}" />
```
scttcper, line 173: a quick alt text might be helpful

sentaur-athena:
I merged the parent PR which caused the rebase issues. Will update them later.

PR #72316, opened 2024-06-07
feat(insights): Remove "Beta" badges from Caches and Queues

gggritso:
We're ready for GA very soon, so the "Beta" badges are not needed.


0Calories (approved):
🫡

PR #72317, opened 2024-06-07
ref: fix typing of tsdb get_range

asottile-sentry:
every implementation except redis returns ints (and even redis returned whole-number floats!)

### DIFF
--- a/src/sentry/tsdb/redis.py
+++ b/src/sentry/tsdb/redis.py
@@ -322,7 +322,7 @@ class RedisTSDB(BaseTSDB):
         jitter_value: int | None = None,
         tenant_ids: dict[str, str | int] | None = None,
         referrer_suffix: str | None = None,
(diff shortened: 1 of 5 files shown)


src/sentry/tsdb/redis.py
```diff
@@ -339,7 +339,7 @@ def get_range(
         self.validate_arguments([model], [environment_id])
 
         rollup, series = self.get_optimal_rollup_series(start, end, rollup)
-        _series: list[datetime] = [to_datetime(item) for item in series]
+        _series = [to_datetime(item) for item in series]
```
asottile-sentry, line 342: `series` here is a sequence of `int`s -- so below the `.timestamp()` will return whole number floats that we can safely turn back into ints without truncation

evanh (approved):

PR #72321, opened 2024-06-07
ref(tagStore): Make TagStore state readonly

scttcper:
another strict store for https://github.com/getsentry/frontend-tsc/issues/66

### DIFF
--- a/static/app/components/smartSearchBar/index.spec.tsx
+++ b/static/app/components/smartSearchBar/index.spec.tsx
@@ -22,16 +22,18 @@ describe('SmartSearchBar', function () {
 
   beforeEach(function () {
     TagStore.reset();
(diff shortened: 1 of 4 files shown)


static/app/components/smartSearchBar/index.spec.tsx
```diff
@@ -22,16 +22,18 @@ describe('SmartSearchBar', function () {
 
   beforeEach(function () {
     TagStore.reset();
-    TagStore.loadTagsSuccess(TagsFixture());
+    TagStore.loadTagsSuccess([
```
scttcper, line 25: love that its `loadTagsSuccess`

michellewzhang (approved):

PR #72324, opened 2024-06-07
chore(hybridcloud) Remove SnowflakeIdMixin

markstory:
This mixin complicates typing and can be implemented as a decorator and module functions.

Refs HC-1192

### DIFF
--- a/src/sentry/utils/snowflake.py
+++ b/src/sentry/utils/snowflake.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING
 
 from django.conf import settings
 from django.db import IntegrityError, router, transaction
+from django.db.models import Model
 from redis.client import StrictRedis
 from rest_framework import status
 from rest_framework.exceptions import APIException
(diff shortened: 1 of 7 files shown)


src/sentry/utils/snowflake.py
```diff
@@ -15,44 +16,52 @@
...
+    Check if a model class has been annotated as a snowflake id model
+    """
+    name = f"{model_class.__module__}.{model_class.__name__}"
+    return name in _snowflake_models
+
+
+def save_with_snowflake_id(
+    instance: BaseModel, snowflake_redis_key: str, save_callback: Callable[[], object]
+) -> None:
+    for _ in range(settings.MAX_REDIS_SNOWFLAKE_RETRY_COUNTER):
```
corps, line 48: Should this assert that the model in question `uses_snowflake_id`?

corps (approved):

asottile-sentry (commented):
nice! much better

src/sentry/utils/snowflake.py
```diff
@@ -15,44 +16,56 @@
...
 if TYPE_CHECKING:
-    from sentry.db.models.base import Model
+    from sentry.db.models.base import Model as BaseModel
 
 _TTL = timedelta(minutes=5)
 
+_snowflake_models: set[str] = set()
+
+
+def snowflake_id_model(model_class: type[Model]) -> object:
```
asottile-sentry, line 26: this should be

```python
ModelT = TypeVar("ModelT", bound=Model)

def snowflake_id_model(model_class: type[ModelT]) -> type[ModelT]:
   ...
```

that way the decorator preserves the type

asottile-sentry: github is acting very strangely -- this comment might post twice

dashed (approved):

PR #72325, opened 2024-06-07
feat(tracing): Add experimental span trace endpoint

wmak:
- This endpoint will attempt to return an entire traces spans along with all the spans data
- For now I just took a pile of columns and threw it into the select clause -- will need to update this list later with the precise ones that the frontend need to actually render the trace
- Mostly concerned to see if we hit the limit of a trace often -- added a measurement to check that


src/sentry/api/endpoints/organization_spans_trace.py
```diff
@@ -0,0 +1,100 @@
...
+    publish_status = {
+        "GET": ApiPublishStatus.PRIVATE,
+    }
+
+    def query_trace_data(self, params, trace_id):
+        builder = SpansIndexedQueryBuilder(
+            Dataset.SpansIndexed,
+            params,
+            query=f"trace:{trace_id}",
+            selected_columns=[
```
volokluev, line 26: this query will be quite slow because of the amount of columns scanned. Just as an FYI. I know you said you'll pair it down later

wmak: :+1: yeah I think the endstate is something like 1 query that only grabs the minimum columns (timestamp, parent, name) and secondary queries that pull details for a single transaction.

But I'm curious to see if this "give me everything!" approach is feasible, and this endpoint is behind a feature flag so should have minimal impact

src/sentry/api/urls.py
```diff
@@ -1555,6 +1556,11 @@ def create_group_urls(name_prefix: str) -> list[URLPattern | URLResolver]:
         OrganizationEventsTraceMetaEndpoint.as_view(),
         name="sentry-api-0-organization-events-trace-meta",
     ),
+    re_path(
+        r"^(?P<organization_id_or_slug>[^\/]+)/spans-trace/(?P<trace_id>(?:\d+|[A-Fa-f0-9-]{32,36}))/$",
```
Zylphrex, line 1560: Thoughts on just naming this `/traces/`

src/sentry/api/endpoints/organization_spans_trace.py
Zylphrex: @wmak have you tried this query in the admin tool to see how it performs yet?

wmak: Not yet, I can give that a try

wmak: Huh nice, "only" 2.5s for a trace with 6500 spans (97 txn) but actually going to remove all the extra columns for now so FE can just focus on rendering first

evanh: I would note too this basically isn't using the primary index at all (I think). So the performance will only improve.

Zylphrex (approved):
LGTM

PR #72326, opened 2024-06-07
fix(similarity): Remove invalid group ids from backfill

jangjodi:
Remove invalid group ids in backfill script

### DIFF
--- a/src/sentry/tasks/backfill_seer_grouping_records.py
+++ b/src/sentry/tasks/backfill_seer_grouping_records.py
@@ -198,7 +198,7 @@ def backfill_seer_grouping_records(
             snuba_requests, referrer=Referrer.GROUPING_RECORDS_BACKFILL_REFERRER.value
         )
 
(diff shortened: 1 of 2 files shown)


JoshFerge (approved):

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -458,13 +482,22 @@ def lookup_group_data_stacktrace_bulk(
...
         sample_rate=1.0,
     )
 
+    logger.info(
+        "backfill_seer_grouping_records.invalid_event_ids",
+        extra={
+            "project_id": project_id,
+            "invalid_event_group_ids": json.dumps(invalid_event_group_ids),
+        },
+    )
```
JoshFerge, line 499: this log name is confusing as the name is invalid_event_ids, but here we're logging the group_ids, but can follow up

PR #72328, opened 2024-06-07
ref(rule): Add description column to AlertRuleTriggerAction model

ceorourke:
Add a new column `description` for metric alert rule actions so we can let users enter some text that will be displayed on the resulting notification.

Part of https://getsentry.atlassian.net/browse/ALRT-83


src/sentry/incidents/models/alert_rule.py
```diff
@@ -497,6 +497,7 @@ class AlertRuleTriggerAction(AbstractNotificationAction):
     status = BoundedPositiveIntegerField(
         default=ObjectStatus.ACTIVE, choices=ObjectStatus.as_choices()
     )
+    description = models.CharField(max_length=1000, null=True)
```
schew2381, line 500: Did we decide to do this per metric alert? From the ticket it seemed like we only really needed to support this for slack

ceorourke: It's per action and yes it will be scoped to Slack only (we'll have to limit the UI)

wedamija: Just want to check if this description length makes sense for sending to slack, or will it be too long?

wedamija (approved):
Migration lgtm

ceorourke:
@wedamija the length was the first thing I checked 😅 the message is so short for metric alerts that we have a lot of extra space

ceorourke:
I don't know if I did the snapshots thing right, I just followed the directions in the error message

src/sentry/incidents/models/alert_rule.py
schew2381: If it's only for slack, is there a reason to not follow what we did for issue alerts? I could see a scenario where we want to support a description for another action, but the field is reserved for slack

wedamija: Right, should this be at the actual alert rule level, so it doesn't need to be configured for each action (assuming we add other actions later)?

schew2381: For metric alerts without slack actions, we'd also have this null description field which would potentially be confusing

ceorourke: Issue alerts uses django forms so I was able to just pull the data from the form - metric alerts UI is built with React so I need to store this data somewhere. It won't limit a future for other actions to use the field, it's just only being used for Slack right now. Nothing about this field is Slack specific - I was just thinking we'd only display the textfield for Slack in the UI.

ceorourke: I asked product if they wanted this per alert or per action and I was told per action.

schew2381: I'm not too familiar with metric alert actions (only issue alerts), but could we store this description in `AlertRuleTriggerAction` instead?

ceorourke: That is where it's being stored

schew2381: oh whoops I misunderstood this whole time. I thought it was on the alert rule model 🙈

schew2381: was so confused there

ceorourke: Hahah no problem, there are a lot of models in this file

schew2381: I guess final thing is just what dan mentioned above, the text limit which I have no clue on

ceorourke: Oh yeah I replied outside the thread: https://github.com/getsentry/sentry/pull/72328#issuecomment-2155333814 but I checked that

schew2381 (approved):
lgtm!

tests/sentry/backup/snapshots/ReleaseTests/test_at_head.pysnap.new
azaslavsky, line 1: Why is there a new copy of this snapshot? I think updating the existing one is fine, though tbh I'm not sure why this PR necessitates it.

ceorourke: I don't know I just followed the instructions 😢 [this test](https://github.com/getsentry/sentry/blob/master/tests/sentry/backup/test_releases.py#L65) failed and it said to either run `SENTRY_SNAPSHOTS_WRITEBACK=new pytest /Users/ceorourke/code/sentry/tests/sentry/backup/test_releases.py` and then `make review-python-snapshots` (which I did, but then I didn't know what I was approving so I exited) or `SENTRY_SNAPSHOTS_WRITEBACK=1 pytest /Users/ceorourke/code/sentry/tests/sentry/backup/test_releases.py` which was automatic. What am I supposed to do?

ceorourke:
Closing in favor of https://github.com/getsentry/sentry/pull/72364 after more discussion

PR #72330, opened 2024-06-07
ref(ui): Various global store readonly state

scttcper:
a lot of these were already doing the right thing for https://github.com/getsentry/frontend-tsc/issues/66

### DIFF
--- a/static/app/stores/IssueListCacheStore.tsx
+++ b/static/app/stores/IssueListCacheStore.tsx
@@ -3,7 +3,7 @@ import {createStore} from 'reflux';
 
 import type {Group} from 'sentry/types/group';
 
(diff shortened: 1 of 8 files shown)


static/app/stores/IssueListCacheStore.tsx
```diff
@@ -78,7 +73,7 @@ const storeConfig: IssueListCacheStoreDefinition = {
   },
 
   getState() {
-    return this.state?.cache ?? null;
```
scttcper, line 81: not used at the moment, just changing it to match the rest of the stores

evanpurkhiser (approved):
these are the ones I did 😏

PR #72334, opened 2024-06-07
chore(hybridcloud) Mainline cached user reads and add generic metrics

markstory:
- Make `get_many_by_id` use caching by default as we're getting a good hit rate on this currently.
- Add generalized metrics for rpc caching so it can be plotted.

Refs HC-1193

### DIFF
--- a/tests/sentry/notifications/notifications/test_assigned.py
+++ b/tests/sentry/notifications/notifications/test_assigned.py
@@ -136,17 +136,17 @@ class AssignedNotificationAPITest(APITestCase):
         ).exists()
 
         assert len(mail.outbox) == 3
-        txt_msg = f"{user1.email} assigned {self.group.qualified_short_id} to {user2.email}"
-        html_msg = f"{self.group.qualified_short_id}</a> to {user2.email}"
-        self.validate_email(mail.outbox, 1, user2.email, txt_msg, html_msg)
-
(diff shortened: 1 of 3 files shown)


tests/sentry/notifications/notifications/test_assigned.py
```diff
@@ -209,16 +210,16 @@ def test_sends_reassignment_notification_team(self):
...
 
         msg = f"Issue assigned to the {team2.slug} team by {user1.email}"
-        self.validate_slack_message(msg, group, project, user2.id, index=2)
-        self.validate_slack_message(msg, group, project, user3.id, index=3)
-        self.validate_slack_message(msg, group, project, user4.id, index=4)
-        self.validate_slack_message(msg, group, project, user1.id, index=5)
+        self.validate_slack_message(msg, group, project, user1.id, index=2)
+        self.validate_slack_message(msg, group, project, user2.id, index=3)
+        self.validate_slack_message(msg, group, project, user3.id, index=4)
+        self.validate_slack_message(msg, group, project, user4.id, index=5)
```
markstory, line 225: These tests were relying on the previous behavior of `user_service.get_many()` having a different result order than the input parameters. The new cached method always returns results in the same order as the user_id parameter and these tests had to be updated accordingly.

GabeVillalobos (approved):

PR #72335, opened 2024-06-07
fix: Improve trimming of Swift function names

Steelskin:
Some Swift for Windows function names were trimmed to incorrect tokens in the UI, resulting in hard-to-read stack traces. For instance, some functions were shown as `"throws"` and closures were shown as `"#1"`.

This change fixes the former issue by using the correct function name. Closures are now displayed as `"closure in <function name>"`, and initialization expressions are displayed as `"initializer expression of <value>"`, improving the user experience when browsing stack traces.

### Legal Boilerplate

### DIFF
--- a/tests/sentry/stacktraces/test_functions.py
+++ b/tests/sentry/stacktraces/test_functions.py
@@ -103,6 +103,30 @@ from sentry.stacktraces.functions import (
         ],
         ["pthread_cond_timedwait@@GLIBC_2.3.2", "pthread_cond_timedwait"],
         ["glob64@GLIBC_2.2", "glob64"],
+        [
+            "static Namespace.ThrowingFunction() throws -> Namespace.ExitValue?",
+            "Namespace.ThrowingFunction",
+        ],
+        [
+            "closure #1 @Swift.MainActor () -> () in static Foo.CallFunction(args: [Swift.String]) -> ()",
+            "closure in Foo.CallFunction",
+        ],
+        [
+            "closure #1 () -> () in Bar.PostTask(() -> ()) -> ()",
+            "closure in Bar.PostTask",
+        ],
+        [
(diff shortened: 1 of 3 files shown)


tests/sentry/stacktraces/test_functions.py
```diff
@@ -103,6 +103,18 @@
         ],
         ["pthread_cond_timedwait@@GLIBC_2.3.2", "pthread_cond_timedwait"],
         ["glob64@GLIBC_2.2", "glob64"],
+        [
+            "static WinUI.WindowsXamlManager.getForCurrentThread() throws -> WinUI.WindowsXamlManager?",
```
tristanlabelle, line 107: This example may be confusing to folks who don't expect to see Windows APIs in Swift code, but no biggie.

tristanlabelle (approved):

tests/sentry/stacktraces/test_functions.py
Steelskin: Changed it to something more generic.

JoshFerge:
thanks for the PR! looking at this now. will make sure it gets reviewed.

brustolin (commented):
Everything seems good to me, but I think we should keep the closure number.

A function (specially in SwiftUI) may have many closures.
`closure #2 in MediaSlideshow.toSources` sounds more helpful

armenzg (approved):

Steelskin:
> Everything seems good to me, but I think we should keep the closure number.
> 
> A function (specially in SwiftUI) may have many closures. `closure #2 in MediaSlideshow.toSources` sounds more helpful

The closure code is hard to parse, especially when we are dealing with closures in closures. Right now, on macOS, we just display `closure`. On Windows, we don't display the right closure number if we are in a closure in a closure. I feel like the current PR is an improvement for both cases. And with #72336 fixed, we can have the full prototype on hover.

PR #72338, opened 2024-06-07
feat(self-hosted): Create topics on upgrade command

hubertdeng123:
Pass in an optional bool flag to optionally create kafka topics. This will mainly be applicable to self-hosted, so that we don't have to manually create kafka topics in https://github.com/getsentry/self-hosted/blob/419d6ccf4e4d0dde2db301e43370d019f9a4c69b/install/create-kafka-topics.sh


hubertdeng123:
passes self hosted e2e tests here that uses a docker image that is built by this PR:

https://github.com/getsentry/self-hosted/actions/runs/9423572834/job/25962243419

lynnagara (approved):

PR #72339, opened 2024-06-07
feat(trace-explorer): Add new traces stats endpoint

Zylphrex:
This adds a new /traces-stats/ endpoint to load the number of matching spans given the n span conditions.


k-fish (approved):

wmak (approved):
lgtm

PR #72341, opened 2024-06-07
fix(insights): Add keys to parsed tokens on queries page

0Calories:
Fixes the unique prop key warning that is always present when viewing the Queries page. This was occuring because the SQL parser was outputting the tokens as a list HTML elements without keys.

image

### DIFF
--- a/static/app/views/starfish/components/tableCells/spanDescriptionCell.tsx
+++ b/static/app/views/starfish/components/tableCells/spanDescriptionCell.tsx
@@ -1,4 +1,4 @@
-import {cloneElement, Fragment, useMemo} from 'react';
+import {Fragment, useMemo} from 'react';
 import styled from '@emotion/styled';
 
 import {Hovercard} from 'sentry/components/hovercard';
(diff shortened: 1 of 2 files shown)


JonasBa (approved):

static/app/views/starfish/components/tableCells/spanDescriptionCell.tsx
```diff
@@ -33,7 +33,12 @@ export function SpanDescriptionCell({
       return rawDescription;
     }
 
-    return formatter.toSimpleMarkup(rawDescription);
+    return (
+      formatter
+        .toSimpleMarkup(rawDescription)
+        // Since the order of the tokens will never change, it's safe to assign the index as the key
+        .map((e, i) => cloneElement(e, {key: i}))
+    );
```
gggritso, line 41: 🤔 can this code live in `sqlish/formatters/simpleMarkup.tsx` instead? That was every usage will have correct `key=` props

gggritso (commented):
Makes sense, thanks! One question about code placement

static/app/views/starfish/components/tableCells/spanDescriptionCell.tsx
0Calories: I completely missed this file, thanks! I'll move the key logic here

0Calories: Moved it to `simpleMarkup.tsx`, how's it look now?

gggritso (approved):

static/app/views/starfish/components/tableCells/spanDescriptionCell.tsx
gggritso: Nice 👍🏻

PR #72344, opened 2024-06-07
feat(chartcuterie): email

sentaur-athena:
reverted the [wrong PR](https://github.com/getsentry/sentry/pull/72296). There was nothing wrong with it 🤦‍♀️


iamrajjoshi (approved):

cathteng (approved):

asottile-sentry:
this introduces a typing regression:

```
src/sentry/integrations/issue_alert_image_builder.py:112: error: Argument "size" to "generate_chart" of "ChartRenderer" has incompatible type "dict[str, int]"; expected "ChartSize | None"  [arg-type]
src/sentry/integrations/issue_alert_image_builder.py:150: error: Argument "size" to "generate_chart" of "ChartRenderer" has incompatible type "dict[str, int]"; expected "ChartSize | None"  [arg-type]
```

sentaur-athena:
@asottile-sentry thanks for pointing it out. Curious why doesn't mypy catch it and where did you see the issue?

PR #72345, opened 2024-06-07
feat(query-builder): Numeric filter improvements

malwilley:
- Adds suggestions for numeric filters (based on the number you type in)
- Allows changing the operator to other things like `<` and `>` instead of just `is`
- When creating a numeric filter, prefills a value (I chose an arbitrary `100`) and doesn't allow you to change it to an invalid one. This was necessary because the parser will not see `timesSeen:` or `timesSeen:abc` as a numeric filter because it does not have a valid value. That resulted in some unwanted behavior. We can change the grammar in the future, but this was a good solution for now.  

Before:

CleanShot 2024-06-07 at 10 13 39

After:

CleanShot 2024-06-07 at 10 13 54

CleanShot 2024-06-07 at 13 13 30

### DIFF
--- a/static/app/utils/fields/index.ts
+++ b/static/app/utils/fields/index.ts
@@ -520,7 +520,7 @@ const EVENT_FIELD_DEFINITIONS: Record<AllEventFieldKeys, FieldDefinition> = {
   [FieldKey.AGE]: {
     desc: t('The age of the issue in relative time'),
     kind: FieldKind.FIELD,
-    valueType: FieldValueType.DURATION,
+    valueType: FieldValueType.DATE,
   },
   [FieldKey.ASSIGNED]: {
     desc: t('Assignee of the issue as a user ID'),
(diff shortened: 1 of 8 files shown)


static/app/utils/fields/index.ts
```diff
@@ -520,7 +520,7 @@ const EVENT_FIELD_DEFINITIONS: Record<AllEventFieldKeys, FieldDefinition> = {
   [FieldKey.AGE]: {
     desc: t('The age of the issue in relative time'),
     kind: FieldKind.FIELD,
-    valueType: FieldValueType.DURATION,
+    valueType: FieldValueType.DATE,
```
malwilley, line 523: While testing I noticed that this type was just wrong, it's a date field

scttcper (approved):

mrduncan:
I think we have a duplicate equality filter - the after screenshot has both "is" and "=" options showing.

malwilley:
@mrduncan this is actually something that we may need to think more about, since `is` is different from `=` in that it can support multiple values. we may want to enforce that `is` is only used when for multiple values, but we may want to think about what the UI for that would look like.

PR #72348, opened 2024-06-07
ref(slack): add integration auto-disable logic to SDK client

cathteng:
All integrations currently have integration auto-disable logic and we should have this for the new Slack client too. This is a bit trickier since we are building on top of the existing WebClient.

We can wrap the all functions inside the metaclass's `__call__` method to make sure we have the integration_id, and then use it to in the auto-disable logic.

### DIFF
--- a/src/sentry/integrations/base.py
+++ b/src/sentry/integrations/base.py
@@ -12,8 +12,11 @@ from urllib.request import Request
 
 from rest_framework.exceptions import NotFound
 
-from sentry import audit_log
+from sentry import audit_log, features
+from sentry.constants import ObjectStatus
 from sentry.exceptions import InvalidIdentity
+from sentry.integrations.notify_disable import notify_disable
+from sentry.integrations.request_buffer import IntegrationRequestBuffer
 from sentry.models.identity import Identity
 from sentry.models.integrations.external_actor import ExternalActor
 from sentry.models.integrations.integration import Integration
@@ -40,12 +43,14 @@ from sentry.shared_integrations.exceptions import (
     IntegrationFormError,
     UnsupportedResponseType,
 )
-from sentry.utils.audit import create_audit_entry
+from sentry.utils.audit import create_audit_entry, create_system_audit_entry
 from sentry.utils.sdk import configure_scope
 
 if TYPE_CHECKING:
     from sentry.services.hybrid_cloud.integration import RpcOrganizationIntegration
     from sentry.services.hybrid_cloud.integration.model import RpcIntegration
+
+logger = logging.getLogger(__name__)
 
 FeatureDescription = namedtuple(
     "FeatureDescription",
@@ -463,3 +468,82 @@ class IntegrationInstallation:
 
     def remove_notification_settings(self, actor_id: int, provider: str) -> None:
         pass
+
+
+def is_response_success(resp) -> bool:
+    if resp.status_code:
+        if resp.status_code < 300:
+            return True
+    return False
+
+
+def is_response_error(resp) -> bool:
+    if resp.status_code:
+        if resp.status_code >= 400 and resp.status_code != 429 and resp.status_code < 500:
+            return True
+    return False
+
+
+def disable_integration(
+    buffer: IntegrationRequestBuffer, redis_key: str, integration_id: int
+) -> None:
+    from sentry.services.hybrid_cloud.integration import integration_service
+
+    result = integration_service.organization_contexts(integration_id=integration_id)
+    rpc_integration = result.integration
+    rpc_org_integrations = result.organization_integrations
+    if rpc_integration and rpc_integration.status == ObjectStatus.DISABLED:
+        return
+
+    org = None
+    if len(rpc_org_integrations) > 0:
+        org_context = organization_service.get_organization_by_id(
+            id=rpc_org_integrations[0].organization_id,
+            include_projects=False,
+            include_teams=False,
+        )
+        if org_context:
+            org = org_context.organization
+
+    extra = {
+        "integration_id": integration_id,
+        "buffer_record": buffer._get_all_from_buffer(),
+    }
+    if len(rpc_org_integrations) == 0 and rpc_integration is None:
+        extra["provider"] = "unknown"
+        extra["organization_id"] = "unknown"
+    elif len(rpc_org_integrations) == 0:
+        extra["provider"] = rpc_integration.provider
+        extra["organization_id"] = "unknown"
+    elif rpc_integration is None:
+        extra["provider"] = "unknown"
+        extra["organization_id"] = rpc_org_integrations[0].organization_id
+    else:
+        extra["provider"] = rpc_integration.provider
+        extra["organization_id"] = rpc_org_integrations[0].organization_id
+
+    logger.info(
+        "integration.disabled",
+        extra=extra,
+    )
+
+    if org and (
+        (rpc_integration.provider == "slack" and buffer.is_integration_fatal_broken())
+        or (rpc_integration.provider == "github")
+        or (
+            features.has("organizations:gitlab-disable-on-broken", org)
+            and rpc_integration.provider == "gitlab"
+        )
+    ):
+        integration_service.update_integration(
+            integration_id=rpc_integration.id, status=ObjectStatus.DISABLED
+        )
+        notify_disable(org, rpc_integration.provider, redis_key)
+        buffer.clear()
+        create_system_audit_entry(
+            organization_id=org.id,
+            target_object=org.id,
+            event=audit_log.get_event_id("INTEGRATION_DISABLED"),
+            data={"provider": rpc_integration.provider},
+        )
+    return
--- a/src/sentry/integrations/base.py
+++ b/src/sentry/integrations/base.py
@@ -470,13 +470,13 @@ class IntegrationInstallation:
         pass
 
 
-def is_response_success(resp) -> bool:
+def is_response_success(resp: Any) -> bool:
     if resp.status_code and resp.status_code < 300:
         return True
     return False
 
 
-def is_response_error(resp) -> bool:
+def is_response_error(resp: Any) -> bool:
     if not resp.status_code:
         return False
     return resp.status_code >= 400 and resp.status_code != 429 and resp.status_code < 500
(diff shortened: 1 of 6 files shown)


src/sentry/integrations/slack/sdk_client.py
```diff
@@ -35,16 +39,43 @@ def track_response_data(response: SlackResponse, method: str, error: str | None
...
         try:
             response = method(*args, **kwargs)
-            if isinstance(response, SlackResponse):
-                track_response_data(response=response, method=method.__name__)
+            track_response_data(response=response, method=method.__name__)
         except SlackApiError as e:
             if e.response:
                 track_response_data(response=e.response, error=str(e), method=method.__name__)
+
+                record_response_for_disabling_integration(e.response, integration_id)
```
cathteng, line 78: @ceorourke does this need to record all responses or just error responses? i notice in the existing logic we seem to only be recording success responses. the `record_response_for_disabling_integration` function isn't in any of the except statements https://github.com/getsentry/sentry/blob/e796c94b46e5fa06bf75eb21f201619726aba07b/src/sentry/shared_integrations/client/base.py#L298

src/sentry/integrations/base.py
```diff
@@ -463,3 +468,82 @@ def notify_remove_external_team(self, external_team: ExternalActor, team: Team)
 
     def remove_notification_settings(self, actor_id: int, provider: str) -> None:
         pass
+
+
+def is_response_success(resp) -> bool:
+    if resp.status_code:
+        if resp.status_code < 300:
+            return True
+    return False
```
ykamo001, line 477:
```suggestion
    if resp.status_code and resp.status_code < 300:
        return True
    return False
```

src/sentry/integrations/base.py
```diff
@@ -463,3 +468,82 @@ def notify_remove_external_team(self, external_team: ExternalActor, team: Team)
...
+
+
+def is_response_error(resp) -> bool:
+    if resp.status_code:
+        if resp.status_code >= 400 and resp.status_code != 429 and resp.status_code < 500:
+            return True
+    return False
```
ykamo001, line 484:
```suggestion
    if not resp.status_code:
        return False
     return resp.status_code >= 400 and resp.status_code != 429 and resp.status_code < 500
```

src/sentry/integrations/base.py
```diff
@@ -463,3 +468,82 @@ def notify_remove_external_team(self, external_team: ExternalActor, team: Team)
...
+def disable_integration(
+    buffer: IntegrationRequestBuffer, redis_key: str, integration_id: int
+) -> None:
+    from sentry.services.hybrid_cloud.integration import integration_service
+
+    result = integration_service.organization_contexts(integration_id=integration_id)
+    rpc_integration = result.integration
+    rpc_org_integrations = result.organization_integrations
+    if rpc_integration and rpc_integration.status == ObjectStatus.DISABLED:
+        return
```
ykamo001, line 496:
```suggestion
        return None
```

src/sentry/integrations/base.py
```diff
@@ -463,3 +468,82 @@ def notify_remove_external_team(self, external_team: ExternalActor, team: Team)
...
+
+def disable_integration(
+    buffer: IntegrationRequestBuffer, redis_key: str, integration_id: int
+) -> None:
+    from sentry.services.hybrid_cloud.integration import integration_service
+
+    result = integration_service.organization_contexts(integration_id=integration_id)
+    rpc_integration = result.integration
+    rpc_org_integrations = result.organization_integrations
+    if rpc_integration and rpc_integration.status == ObjectStatus.DISABLED:
```
ykamo001, line 495: you could use the walrus here if you wanted

ykamo001:
```suggestion
    if rpc_integration := result.integration and rpc_integration.status == ObjectStatus.DISABLED:
```

src/sentry/integrations/base.py
```diff
@@ -463,3 +468,82 @@ def notify_remove_external_team(self, external_team: ExternalActor, team: Team)
...
+        extra["organization_id"] = "unknown"
+    elif len(rpc_org_integrations) == 0:
+        extra["provider"] = rpc_integration.provider
+        extra["organization_id"] = "unknown"
+    elif rpc_integration is None:
+        extra["provider"] = "unknown"
+        extra["organization_id"] = rpc_org_integrations[0].organization_id
+    else:
+        extra["provider"] = rpc_integration.provider
+        extra["organization_id"] = rpc_org_integrations[0].organization_id
```
ykamo001, line 523: can this be simplified?
```suggestion
extra["provider"] = "unknown" if rpc_integration is None else rpc_integration.provider
extra["organization_id"] = "unknown" if len(rpc_org_integrations) == 0 else rpc_org_integrations[0].organization_id
```

src/sentry/integrations/base.py
```diff
@@ -463,3 +468,82 @@ def notify_remove_external_team(self, external_team: ExternalActor, team: Team)
...
+            integration_id=rpc_integration.id, status=ObjectStatus.DISABLED
+        )
+        notify_disable(org, rpc_integration.provider, redis_key)
+        buffer.clear()
+        create_system_audit_entry(
+            organization_id=org.id,
+            target_object=org.id,
+            event=audit_log.get_event_id("INTEGRATION_DISABLED"),
+            data={"provider": rpc_integration.provider},
+        )
```
ykamo001, line 548: could we move this to another function? Seems like we're mixing responsibilities here and we could make the function easier to test/maintain

cathteng: isn't this actually disabling the integration and notifying recipients?

src/sentry/integrations/base.py
```diff
@@ -463,3 +468,72 @@ def notify_remove_external_team(self, external_team: ExternalActor, team: Team)
 
     def remove_notification_settings(self, actor_id: int, provider: str) -> None:
         pass
+
+
+def is_response_success(resp) -> bool:
```
ykamo001, line 473: nit: we should at least add `Any` as a type hint

src/sentry/integrations/base.py
```diff
@@ -463,3 +468,72 @@ def notify_remove_external_team(self, external_team: ExternalActor, team: Team)
...
+    if resp.status_code and resp.status_code < 300:
+        return True
+    return False
+
+
+def is_response_error(resp) -> bool:
```
ykamo001, line 479: same comment here

ykamo001: Also, would this function be the opposite of the function above? Or is there a scenario where it's not an error or success?

src/sentry/integrations/base.py
ykamo001: Yeah true, the audit and buffer clearing threw me off. It's conditionally based, so it's not applicable to the entire flow in this function, so thought it would be cleaner to read in a new function

ykamo001 (approved):

src/sentry/integrations/base.py
ceorourke: This looks about the same as the base definition - https://github.com/getsentry/sentry/blob/master/src/sentry/shared_integrations/client/base.py#L122-L126 are you re-defining it since you're redoing the client and it won't inherit from this anymore? I'm a little out of the loop of what you're doing here

src/sentry/integrations/base.py
```diff
@@ -463,3 +468,72 @@ def notify_remove_external_team(self, external_team: ExternalActor, team: Team)
...
+    if not resp.status_code:
+        return False
+    return resp.status_code >= 400 and resp.status_code != 429 and resp.status_code < 500
+
```
ceorourke, line 483: Do you want to retain the functionality of checking for fatal errors? It's pretty specific to avoid false positives https://github.com/getsentry/sentry/blob/master/src/sentry/integrations/slack/client.py#L74-L82

src/sentry/integrations/base.py
cathteng: yes it doesn't inherit from the client anymore and the response class is different

src/sentry/integrations/base.py
cathteng: i don't know, i'm copying what the previous person did. it seems like they don't count 429s or 500s as meaning the integration is broken https://www.notion.so/sentry/Integration-Auto-Disable-Notify-5266cce6e5834b1d8120cfde93050fb8

src/sentry/integrations/base.py
ceorourke: oh nvm I see it below

cathteng: i'm retaining it here https://github.com/getsentry/sentry/pull/72348/files#diff-06439c73605aa1447ba2ca4f2dfe7bb2b579dad459893ace24dd7252f5003355R42-R46, i moved it out of the old SlackClient

src/sentry/integrations/base.py
```diff
@@ -463,3 +468,72 @@ def notify_remove_external_team(self, external_team: ExternalActor, team: Team)
...
+
+def disable_integration(
```
ceorourke, line 485: Is there any way you can use the [shared_integration version of this](https://github.com/getsentry/sentry/blob/master/src/sentry/shared_integrations/client/base.py#L418)? Like could that be moved somewhere everyone can access it? It feels bad to have the same code twice.

cathteng: ? i moved it out of `shared_integation`, everywhere is importing from here now

src/sentry/integrations/base.py
ceorourke: It's been a while but I think we did this to avoid disabling the integration in case there are temporary issues

src/sentry/integrations/slack/sdk_client.py
ceorourke: I think it should just be error responses, I see what you're saying but I dont think it'd be working at all if it's only recording successes 🤔 and we know it is disabling stuff from checking logs

src/sentry/integrations/base.py
ceorourke: oh nvm I can't read

src/sentry/integrations/slack/sdk_client.py
cathteng: okay the spec here says it should be error and successes https://www.notion.so/sentry/Tech-Spec-Notify-on-Disabled-Integration-Spec-e7ea0f86ccd6419cb3e564067cf4a2ef

PR #72349, opened 2024-06-07
ref(groupingStore): Stricter grouping store state

scttcper:
Can't make this store super strict but, this is more than it was. Part of https://github.com/getsentry/frontend-tsc/issues/66


static/app/stores/groupingStore.tsx
```diff
@@ -585,54 +571,46 @@ const storeConfig: GroupingStoreDefinition = {
...
     };
-    this.trigger(state);
-    return state;
+    this.trigger(this.state);
+    return this.state;
   },
 
   triggerUnmergeState() {
-    const state = pick(this, [
+    const state = pick(this.state, [
```
scttcper, line 613: could probably just get rid of this

malwilley (approved):

PR #72357, opened 2024-06-07
ref: fix a few types in sentry.digests.notifications

asottile-sentry:
this fixes an error in an ignored file when LazyServiceWrapper becomes type checked

### DIFF
--- a/src/sentry/digests/__init__.py
+++ b/src/sentry/digests/__init__.py
@@ -1,9 +1,8 @@
 from __future__ import annotations
 
 import datetime as datetime_mod
-from collections import namedtuple
 from collections.abc import MutableMapping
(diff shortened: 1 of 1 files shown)


src/sentry/digests/__init__.py
```diff
@@ -23,13 +22,20 @@
 backend.expose(locals())
 
 
-class Record(namedtuple("Record", "key value timestamp")):
+class Record(NamedTuple):
+    key: str
+    value: Any  # TODO: I think this is `Notification` ?
```
asottile-sentry, line 27: I tried fixing this as well -- but there are too many callers which stuff other things in here -- or build `Notification` objects with incorrect data

will follow up and fix that later!

schew2381 (approved):

PR #72362, opened 2024-06-07
feat(autofix): Add frontend for iterative user feedback

jennmueng:
Introduces a User Response step:
Screenshot 2024-06-05 at 3 02 17 PM

And a user input for autofix that only shows for the last changes card:
Screenshot 2024-06-07 at 10 10 15 AM

The input is gated behind a flag in the autofix state from seer: `autofixState.options.interative_feedback` which means the input won't show in prod yet. The user response card won't be added to the steps either.

### DIFF
--- a/static/app/components/events/autofix/autofixInputField.tsx
+++ b/static/app/components/events/autofix/autofixInputField.tsx
@@ -12,6 +12,7 @@ import LoadingIndicator from 'sentry/components/loadingIndicator';
 import Panel from 'sentry/components/panels/panel';
 import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
+import {isCtrlKeyPressed} from 'sentry/utils/isCtrlKeyPressed';
 import {setApiQueryData, useMutation, useQueryClient} from 'sentry/utils/queryClient';
 import useApi from 'sentry/utils/useApi';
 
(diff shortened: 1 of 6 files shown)


static/app/components/events/autofix/autofixInputField.tsx
```diff
@@ -0,0 +1,128 @@
...
+};
+
+export function AutofixInputField({groupId, runId}: {groupId: string; runId: string}) {
+  const {sendInstruction, instruction, setInstruction, isSubmitting} =
+    useAutofixUserInstruction(groupId, runId);
+
+  return (
+    <Card>
+      <Title>{t("Doesn't look right? Tell Autofix what needs to be changed")}</Title>
+      <FormRow>
```
malwilley, line 81: This should be an actual `form` element so the textarea and button can be connected semantically. Probably want to use the `Form` component?

static/app/components/events/autofix/autofixInputField.tsx
```diff
@@ -0,0 +1,128 @@
...
+        <TextArea
```
malwilley, line 82: nit: might be nice if cmd+enter submitted the form, like github

static/app/components/events/autofix/autofixInputField.tsx
jennmueng: This originally used the `Form` component, but had to add the `FormField` wrapper to get the text area to work with the form, by doing so it added padding/margin that I couldn't remove

malwilley (approved):

PR #72366, opened 2024-06-07
feat(profiling): Format profile durations to hours

scttcper:
I believe they're in ms, we want to display them as hours

example org with profile duration stats https://testorg-am3launch-am3-business.sentry.io/stats/?dataCategory=profileDuration&project=4507392064290816&statsPeriod=14d

### DIFF
--- a/static/app/views/organizationStats/utils.tsx
+++ b/static/app/views/organizationStats/utils.tsx
@@ -47,7 +47,7 @@ export function formatUsageWithUnits(
     dataCategory === DATA_CATEGORY_INFO.profileDuration.plural &&
     Number.isFinite(usageQuantity)
   ) {
-    // Profile duration is in miliseconds and we want hours
+    // Profile duration is in milliseconds, convert to hours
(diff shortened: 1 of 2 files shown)


static/app/views/organizationStats/utils.tsx
```diff
@@ -31,21 +31,31 @@ export function formatUsageWithUnits(
...
+      maximumFractionDigits: 2,
+    });
   }
 
-  const usageGb = usageQuantity / GIGABYTE;
   return options.isAbbreviated
-    ? `${abbreviateUsageNumber(usageGb)} GB`
-    : `${usageGb.toLocaleString(undefined, {maximumFractionDigits: 2})} GB`;
+    ? abbreviateUsageNumber(usageQuantity)
+    : usageQuantity.toLocaleString();
```
scttcper, line 58: rearranging this to have the two special cases first and the default case last

dashed (approved):

vbro (approved):

PR #72367, opened 2024-06-07
ref: fix typing for sentry.digests.notifications

asottile-sentry:
so this ended up being a pretty big change -- even with splitting out most of the smaller digestible parts

the main problem preventing this from being typesafe was the original code stuffed `user_count` / `event_count` attributes into `Group` instances

refactoring that out required breaking the FP structure of digests.notifications (though an argument can be made that it wasn't FP at all since it mutated the inputs (!))

this actually ended up simplifying the digest grouping significantly, and made it a lot easier to follow what was happening

I was able to preserve the tests for the most part as well!

I did have to adjust the digest-text template since it was relying on the attribute stuffing -- I also needed to adjust the debug mail view since the way it was faking organizations was relying on the previous attribute stuffing


src/sentry/digests/notifications.py
```diff
@@ -72,148 +75,117 @@ def event_to_record(
...
-) -> tuple[Digest | None, Sequence[str]]:
+    groups: dict[int, Group],
+    rules: dict[int, Rule],
+    event_counts: dict[int, int],
+    user_counts: dict[int, int],
+) -> Digest:
+    # sans-io implementation details
+    bound_records = _bind_records(records, groups, rules)
+    grouped = _group_records(bound_records, groups, rules)
+    return _sort_digest(grouped, event_counts=event_counts, user_counts=user_counts)
```
nhsiehgit, line 150: 👨‍🍳 💋

nhsiehgit (approved):

PR #72375, opened 2024-06-08
fix(crons): Add option to specify monitors by guid in bulk edit

davidenwang:
Allows for the client to specify a list of monitor guids which correspond to the monitors they are trying to edit. Keeping the old functionality as we transition the frontend to use the new request payload

Related to: https://github.com/getsentry/team-crons/issues/184

### DIFF
--- a/src/sentry/monitors/validators.py
+++ b/src/sentry/monitors/validators.py
@@ -396,7 +396,7 @@ class MonitorCheckInValidator(serializers.Serializer):
         return attrs
 
 
-class MonitorBulkEditValidator(MonitorValidator):
+class MonitorBulkEditValidatorLegacy(MonitorValidator):
     slugs = serializers.ListField(
         child=SentrySerializerSlugField(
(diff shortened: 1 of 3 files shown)


src/sentry/monitors/validators.py
```diff
@@ -410,3 +410,17 @@ def validate_slugs(self, value):
...
+    ids = serializers.ListField(
+        child=serializers.UUIDField(format="hex"),
+        required=True,
+    )
+
+    def validate_ids(self, value):
+        if Monitor.objects.filter(
+            guid__in=value,
+        ).count() != len(value):
+            raise ValidationError("Not all ids are valid for this organization.")
```
evanpurkhiser, line 425: This doesn't look like it's actually doing an organization check like the one in the legacy validator?

davidenwang: ooh good catch, yea we should make sure the guids they specify are all from their own organization

evanpurkhiser (approved):

PR #72383, opened 2024-06-10
chore(metrics): remove metrics blocking feature flag

obostjancic:
Metrics blocking feature should be available to anyone with access to custom metrics


iambriccardo (approved):
LGTM, thanks for doing this!

PR #72384, opened 2024-06-10
chore(metrics): remove use metrics layer in alerts flag

obostjancic:
Replacing `use-metrics-layer-in-alerts` with `custom-metrics`


iambriccardo (approved):

shellmayr (approved):
LGTM - we should make sure the other mentions of the flag also get cleaned up:

https://github.com/search?q=org%3Agetsentry%20use-metrics-layer-in-alerts&type=code

PR #72387, opened 2024-06-10
ref(profiling): fix typing for profile_functions and remove from mypy ignore list

viglia:


src/sentry/search/events/datasets/profile_functions.py
```diff
@@ -580,10 +582,17 @@ def _resolve_cpm_cond(
         alias: str | None,
         cond: str,
     ) -> SelectType:
+        timestmp = args["timestamp"]
```
asottile-sentry, line 585: typo

asottile-sentry (approved):
fixnship

PR #72389, opened 2024-06-10
feat: Update sourcemaps debugger logic for supported SDKs

### LINKED ISSUE #72302: Update Source Maps Debug Modal that NextJs Supports DebugIDs
### Problem Statement

Previous NextJS couldn't support debugIds due to a being tied to 1.x webpack under the hood. With V8 this is no longer the case

The modal misleads the user that they're set up is not supported, when for Next.js they need to just update to >v8

### Solution Brainstorm

Update modal for Next.js projects/issues <8.x that user is directed upgrade to >v8

- others are impacted potenitally example Gatsby

### Product Area

Issues - Source Maps

lforst:
Sveltekit and Next.js now support debug IDs with version 8 and onwards.

Resolves https://github.com/getsentry/sentry/issues/72302

### DIFF
--- a/src/sentry/api/endpoints/source_map_debug_blue_thunder_edition.py
+++ b/src/sentry/api/endpoints/source_map_debug_blue_thunder_edition.py
@@ -41,6 +41,7 @@ from sentry.utils.urls import non_standard_url_join
 MIN_JS_SDK_VERSION_FOR_DEBUG_IDS = "7.56.0"
 MIN_REACT_NATIVE_SDK_VERSION_FOR_DEBUG_IDS = "5.11.1"
 MIN_ELECTRON_SDK_VERSION_FOR_DEBUG_IDS = "4.6.0"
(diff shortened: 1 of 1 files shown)


src/sentry/api/endpoints/source_map_debug_blue_thunder_edition.py
```diff
@@ -643,6 +642,7 @@ def get_sdk_debug_id_support(event_data):
             "sentry.javascript.react",
             "sentry.javascript.react-native",
             "sentry.javascript.remix",
+            "sentry.javascript.solid",
```
andreiborza, line 645: Maybe a silly question but why's this part of the PR?

lforst: Not a silly question. Just a little improvement while we're in this file.

andreiborza (approved):

Lms24 (approved):

PR #72394, opened 2024-06-10
fix(replay-onboarding): Adjust Replay onboarding instructions for Astro

Lms24:
This PR adjusts the in-product Replay onboarding doc by:

- stating that Replay is enabled by default in the Astro SDK (=> configuration becomes optional)
- diffrentiating between which options can be set in the simple `astro.config.mjs` vs which options can only be set by creating a dedicated `sentry.client.config.js` file. 
- Adjusting text as necessary.

Before (taken from https://github.com/getsentry/sentry-javascript/issues/12408): 

image

After:
image

closes https://github.com/getsentry/sentry-javascript/issues/12408

cc @whitep4nth3r

### DIFF
--- a/static/app/gettingStartedDocs/javascript/astro.tsx
+++ b/static/app/gettingStartedDocs/javascript/astro.tsx
@@ -18,10 +18,7 @@ import {
   getFeedbackSDKSetupSnippet,
 } from 'sentry/components/onboarding/gettingStartedDoc/utils/feedbackOnboarding';
 import {getJSMetricsOnboarding} from 'sentry/components/onboarding/gettingStartedDoc/utils/metricsOnboarding';
-import {
-  getReplayConfigureDescription,
-  getReplaySDKSetupSnippet,
-} from 'sentry/components/onboarding/gettingStartedDoc/utils/replayOnboarding';
+import {getReplaySDKSetupSnippet} from 'sentry/components/onboarding/gettingStartedDoc/utils/replayOnboarding';
 import {t, tct} from 'sentry/locale';
 
 type Params = DocsParams;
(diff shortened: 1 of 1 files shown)


mydea (approved):
hmm, should we also document/show how to deactivate replay? If users do not have replay selected in the onboarding, we should show some config that removes it, I guess? (Or is this already happening?)

Lms24:
> Also — why would we include the replay configuration in both files? Surely it makes more sense just to include it in the client config?

We don't. The idea is:
- either you use the simpe `astro.config.mjs` config which has [limited options](https://docs.sentry.io/platforms/javascript/guides/astro/manual-setup/#astro-integration-options). This is the default experience. You can control replay with the replay sample rates. If both are 0, replay is not enabaled, otherwise the integration is added. By default we sample with 0.1 on session and 1.0 on error.
- or you use `sentry.client.config.js` which works like any other `Sentry.init` call. This however, is opt-in. Once you created this file, the configuration in `astro.config.mjs` has no effect anymore (with the exception of source maps upload).

So this updated guide tries to accommodate both of these use cases

> This makes more sense, but still doesn't explain why I see no replays after a week now 😅

Hmm nope 😅 Do you see debug logs indicating if Replay was installed or not? do you have a link to your project repo by chance?

whitep4nth3r:
Conversation was continued in Slack and I got it set up. Essentially we've all concluded that we need some clearer instructions around configuring Astro, given that you can use three files to configure things and some settings might override the others.

static/app/gettingStartedDocs/javascript/astro.tsx
```diff
@@ -206,20 +203,74 @@ const onboarding: OnboardingConfig = {
 };
 
 const replayOnboarding: OnboardingConfig = {
-  install: () => getInstallConfig(),
+  install: () => [
+    {
+      ...getInstallConfig()[0],
+      additionalInfo:
+        'Session Replay is enabled by default when you install the Astro SDK!',
```
michellewzhang, line 210: do we need to state that it's enabled by default twice? it's here and also on the first line of the configure step

static/app/gettingStartedDocs/javascript/astro.tsx
```diff
@@ -206,20 +203,74 @@ const onboarding: OnboardingConfig = {
...
+          code: <code />,
+          code2: <code />,
+          link: (
+            <ExternalLink
+              href={
+                'https://docs.sentry.io/platforms/javascript/guides/astro/session-replay/'
+              }
+            />
+          ),
+        }
```
michellewzhang, line 228:
```suggestion
        {
          link: (
            <ExternalLink
              href={
                'https://docs.sentry.io/platforms/javascript/guides/astro/session-replay/'
              }
            />
          ),
        }
```

static/app/gettingStartedDocs/javascript/astro.tsx
Lms24: Yeah I was thinking better twice than confuse people why there's the configure section but you're right, let's leave it here and remove it from the configure section.

static/app/gettingStartedDocs/javascript/astro.tsx
```diff
@@ -206,20 +203,74 @@ const onboarding: OnboardingConfig = {
...
+            }
+          ),
+        },
         {
           code: [
             {
               label: 'JavaScript',
               value: 'javascript',
               language: 'javascript',
+              filename: 'sentry.client.config.js',
```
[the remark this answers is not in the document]
Lms24: Yup, good suggestion! I added this:
image

Lms24:
Thx for the reviews! I applied the changes. I'm on vacation though for the next two weeks so please feel free to merge this :)

michellewzhang (approved):
lgtm! thanks for making the changes!

ryan953 (approved):

ryan953:
> Thx for the reviews! I applied the changes. I'm on vacation though for the next two weeks so please feel free to merge this once you think this is good to go :)

@michellewzhang do you want to click merge if you've got the bandwidth to support it?

michellewzhang:
> > Thx for the reviews! I applied the changes. I'm on vacation though for the next two weeks so please feel free to merge this once you think this is good to go :)
> 
> 
> 
> @michellewzhang do you want to click merge if you've got the bandwidth to support it? 

yup, i'll merge and watch over

michellewzhang (approved):

PR #72398, opened 2024-06-10
feat(metrics): use new meta tables to query MRI-based metrics, retain old API for non-MRI metrics and track

shellmayr:
We want to use the new meta tables for querying for metrics tags. 
There are multiple cases that need to be handled: 
1. MRIs that are correctly parseable and correspond to one of the metric types c:, s:, d:, g:
2. MRIs that are correctly parseable, but correspond to the derived metric type e:
3. Metric names that are not parseable as MRIs 

Case 1 can use the new meta tables
Case 2 is not contained in the new meta tables, and should therefore still use the old function
Case 3 will not be supported at all anymore - only valid MRIs are allowed


obostjancic (approved):
lfg

PR #72403, opened 2024-06-10
feat(related_issues): Empty state should not show any sections

armenzg:
In the Similar Issues tab, if no related issues are found do not show any empty sections.

This is what it would look like before this change:
image


static/app/views/issueDetails/groupRelatedIssues/index.spec.tsx
```diff
@@ -75,37 +71,6 @@ describe('Related Issues View', function () {
     jest.clearAllMocks();
   });
 
-  it('renders with no data', async function () {
```
armenzg, line 78: This test makes no sense anymore.

scttcper (approved):

PR #72409, opened 2024-06-10
feat(traces): Use traces-stats endpoint

k-fish:
### Summary
This uses the traces-stats endpoint and fixes the referrer.

Traces-stats is required since we need to do a multi-project event stats search on spans, but we don't want to expose this api outside of the specific use case of trace explorer.

### DIFF
--- a/static/app/views/starfish/utils/useSpansQuery.tsx
+++ b/static/app/views/starfish/utils/useSpansQuery.tsx
@@ -66,11 +66,13 @@ export function useWrappedDiscoverTimeseriesQuery<T>({
   initialData,
   referrer,
   cursor,
+  overriddenRoute,
 }: {
   eventView: EventView;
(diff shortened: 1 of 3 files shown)


static/app/views/starfish/utils/useSpansQuery.tsx
```diff
@@ -83,7 +85,7 @@ export function useWrappedDiscoverTimeseriesQuery<T>({
     },
     DiscoverQueryProps
   >({
-    route: 'events-stats',
+    route: overriddenRoute ?? 'events-stats',
```
k-fish, line 88: We want to allow overriding instead of cloning all these hooks since `traces-stats` is a direct sub-set of `event-stats` for now.

Zylphrex (approved):

PR #72410, opened 2024-06-10
fix(similarity): Filter out null from event title

jangjodi:
Filter out null byte from event title since it cannot be saved in records table

fixes [SEER-63](https://sentry.sentry.io/issues/5473319314/)

### DIFF
--- a/src/sentry/grouping/ingest/seer.py
+++ b/src/sentry/grouping/ingest/seer.py
@@ -10,7 +10,11 @@ from sentry.models.group import Group
 from sentry.models.project import Project
 from sentry.seer.similarity.similar_issues import get_similarity_data_from_seer
 from sentry.seer.similarity.types import SeerSimilarIssuesMetadata, SimilarIssuesEmbeddingsRequest
-from sentry.seer.similarity.utils import event_content_is_seer_eligible, get_stacktrace_string
+from sentry.seer.similarity.utils import (
+    event_content_is_seer_eligible,
+    filter_null_from_event_title,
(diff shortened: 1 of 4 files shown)


src/sentry/grouping/ingest/seer.py
```diff
@@ -163,7 +167,7 @@ def get_seer_similar_issues(
         "hash": event_hash,
         "project_id": event.project.id,
         "stacktrace": stacktrace_string,
```
trillville, line 169: is it possible for this to affect the stacktrace as well?

trillville (approved):

src/sentry/grouping/ingest/seer.py
jangjodi: this likely is possible (though it maybe be properly filtered out already), but since we're only saving the vector embedding of the stacktrace in postgres, it's not a problem

trillville: hm do you have the string handy by any chance? maybe we should check what happens when we call `encode` on it

jangjodi: yeah I tried it out earlier - it'll just translate the byte string into raw (ex. `b'Title with null \x00, "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" is null'`)

trillville: kk cool ty for verifying!!

PR #72415, opened 2024-06-10
ref: fix typing for hc CacheVersion model

asottile-sentry:
- `.first(...)` is nullable
- `.filter(...).values(...)[...]` is unnecessary conversion to `dict`

fixes errors in this module when BaseManager becomes typed

### DIFF
--- a/src/sentry/hybridcloud/models/cacheversion.py
+++ b/src/sentry/hybridcloud/models/cacheversion.py
@@ -19,7 +19,7 @@ class CacheVersionBase(Model):
             cls.objects.create_or_update(
                 key=key, defaults=dict(version=1), values=dict(version=F("version") + 1)
             )
-            return cls.objects.filter(key=key).values("version").first()["version"]
+            return cls.objects.get(key=key).version
 
     @classmethod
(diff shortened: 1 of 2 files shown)


src/sentry/hybridcloud/models/cacheversion.py
```diff
@@ -19,7 +19,7 @@ def incr_version(cls, key: str) -> int:
             cls.objects.create_or_update(
                 key=key, defaults=dict(version=1), values=dict(version=F("version") + 1)
             )
-            return cls.objects.filter(key=key).values("version").first()["version"]
+            return cls.objects.get(key=key).version
```
evanpurkhiser, line 22: This is a little different now right, since it's turned into a `select *`

asottile-sentry: I suspect it isn't because the only fields are (key, version)

evanpurkhiser (approved):

src/sentry/hybridcloud/models/cacheversion.py
evanpurkhiser: sgtm

PR #72422, opened 2024-06-10
ref(similarity): backfill refactors

JoshFerge:
makes some changes to the backfill to improve readability. no behavior changes intended.

- splits up logic into different functions, renames variables
- attempts to hoist logic to separate data calling from transformations
- attempts to remove any instances of mutability to make clear what each variable at each step is

i did delete a couple of tests that now felt not as helpful with the refactors.

please follow along at the commit level, as each commit should show a digestable change.


src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -79,6 +90,136 @@ def backfill_seer_grouping_records(
...
+
+    try:
+        project, redis_client, last_processed_index = initialize_backfill(
+            project_id, last_processed_index, dry_run
+        )
+    except FeatureError:
+        logger.info(
+            "backfill_seer_grouping_records.no_feature",
+            extra={"project_id": project_id},
+        )
```
jangjodi, line 102: I think we can remove this log statement since it's also in `initialize_backfill`

JoshFerge: thanks! fixed (removed log in function)

jangjodi (approved):

PR #72427, opened 2024-06-10
ref(replay): Refactor OpenReplayComparisonButton to take more params

ryan953:
Refactoring OpenReplayComparisonButton to take more params, so I can more easily customize it to be part of the Issue Details page.

Related to https://github.com/getsentry/sentry/issues/70199

### DIFF
--- a/static/app/components/replays/breadcrumbs/openReplayComparisonButton.tsx
+++ b/static/app/components/replays/breadcrumbs/openReplayComparisonButton.tsx
@@ -1,10 +1,10 @@
-import {Fragment, lazy, Suspense} from 'react';
+import {Fragment, lazy, type ReactNode, Suspense} from 'react';
 import {css} from '@emotion/react';
 
 import {openModal} from 'sentry/actionCreators/modal';
-import {Button} from 'sentry/components/button';
+import FeatureBadge from 'sentry/components/badge/featureBadge';
+import {Button, type ButtonProps} from 'sentry/components/button';
 import LoadingIndicator from 'sentry/components/loadingIndicator';
-import {t} from 'sentry/locale';
 import type ReplayReader from 'sentry/utils/replays/replayReader';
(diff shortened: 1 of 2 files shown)


static/app/components/replays/breadcrumbs/openReplayComparisonButton.tsx
```diff
@@ -39,7 +43,12 @@ export function OpenReplayComparisonButton({
               fallback={
                 <Fragment>
                   <deps.Header closeButton>
-                    <deps.Header>{t('Hydration Error')}</deps.Header>
+                    <deps.Header>
+                      <h4>
+                        Hydration Error
```
michellewzhang, line 48: does this need to be translated?

michellewzhang (approved):

static/app/components/replays/breadcrumbs/openReplayComparisonButton.tsx
ryan953: I un-translated it because on the server we're creating errors with the english name "Hydration Error".
https://github.com/getsentry/sentry/blob/9f3bfebb247fe6ac67a67831f7d3138545babf64/src/sentry/replays/usecases/ingest/issue_creation.py#L13-L19

Also, help text for hydration errors are not translated:
https://github.com/getsentry/sentry/blob/9f3bfebb247fe6ac67a67831f7d3138545babf64/static/app/utils/issueTypeConfig/errorConfig.tsx#L201-L215

And the modal loading state doesn't have it translated:
https://github.com/getsentry/sentry/blob/9f3bfebb247fe6ac67a67831f7d3138545babf64/static/app/components/replays/breadcrumbs/replayComparisonModal.tsx#L32-L35

ryan953:
Related to https://github.com/getsentry/sentry/issues/70199

PR #72432, opened 2024-06-10
fix(replay): Fix hydration diff modal tabs

ryan953:
Added in the missing `<Tabs>` wrapper which creates context for the `<TabList.Item>` and `<TabPanels.Item>` components below. Things work much better using the defaults like this. 

We really need a story to show all the parts and explain how they're used together (and also explain what should be a child of what)

### DIFF
--- a/static/app/components/replays/replayDiff.tsx
+++ b/static/app/components/replays/replayDiff.tsx
@@ -1,4 +1,4 @@
-import {Fragment, useEffect, useState} from 'react';
+import {useEffect, useState} from 'react';
 import styled from '@emotion/styled';
 import beautify from 'js-beautify';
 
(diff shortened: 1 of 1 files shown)


michellewzhang (approved):

static/app/components/replays/replayDiff.tsx
```diff
@@ -48,7 +52,7 @@ export default function ReplayDiff({leftTimestamp, replay, rightTimestamp}: Prop
   const isSameTimestamp = leftBody && rightBody && leftBody === rightBody;
 
   return (
-    <Fragment>
+    <Tabs<DiffType> defaultValue={defaultTab}>
```
michellewzhang, line 55: whoa, TIL you can do this

static/app/components/replays/replayDiff.tsx
```diff
@@ -58,70 +62,61 @@ export default function ReplayDiff({leftTimestamp, replay, rightTimestamp}: Prop
...
       <Flex gap={space(1)} column>
-        <TabList
-          selectedKey={activeTab}
-          onSelectionChange={tab => setActiveTab(tab as Tab)}
-        >
-          <TabList.Item key={Tab.HTML}>{t('Html Diff')}</TabList.Item>
-          <TabList.Item key={Tab.VISUAL}>{t('Visual Diff')}</TabList.Item>
+        <TabList>
+          <TabList.Item key={DiffType.VISUAL}>{t('Visual Diff')}</TabList.Item>
+          <TabList.Item key={DiffType.HTML}>{t('Html Diff')}</TabList.Item>
```
ryan953, line 67: I flipped the order of these, and the default.

The HTML item has some kind of problem refreshing itself on initial render, so it renders blank. This is a little change to avoid having to debug and fix that right now

PR #72433, opened 2024-06-10
docs(remote-config): Add support for named, versioned configurations and allow features to be disabled

cmanallen:
- Adds support for explicitly naming configurations.
- Adds support for associating configurations to environments.
- Adds support for versioning configurations.
- Adds support for disabling and enabling features.

Open Questions:

- When do we push? Automatically? Manually? How do we stage changes so the user can modify a feature without necessarily publishing it. How do we prevent other users from accidentally publishing staged feature changes?
- How do we handle environment renames?
   - Need to associate to a foreign-key.
   - Push the associated configuration on environment rename?
   - Do SDKs automatically update on environment rename?

### DIFF
--- a/src/sentry/remote_config/docs/api.md
+++ b/src/sentry/remote_config/docs/api.md
@@ -50,6 +50,15 @@ Fetch a project's configuration. Responses should be proxied exactly to the SDK.
 
 ## Configurations [/projects/<organization_id_or_slug>/<project_id_or_slug>/configurations/]
 
+**Parameters**
+
+- offset (optional, number)
+  Default: 0
+- per_page (optional, number)
+  Default: 10
+
+**Attributes**
+
 | Field              | Type             | Description                                                                   |
 | ------------------ | ---------------- | ----------------------------------------------------------------------------- |
 | environment        | optional[string] | The environment the configuration is associated with.                         |
@@ -170,6 +179,23 @@ Delete a configuration.
 - Response 204
 
 ## Features [/projects/<organization_id_or_slug>/<project_id_or_slug>/configuration/<configuration_id>/features/]
+
+**Parameters**
+
+- offset (optional, number)
+  Default: 0
+- per_page (optional, number)
+  Default: 10
+
+**Attributes**
+
+| Field       | Type   | Description                                           |
+| ----------- | ------ | ----------------------------------------------------- |
+| description | string | The environment the configuration is associated with. |
+| id          | string | A server generated unique identifier.                 |
+| is_enabled  | bool   | Disabled features are not visible to clients.         |
+| key         | string | A unique, per-config feature identifier.              |
+| value       | any    | The value associated to the key.                      |
 
 ### Get Features [GET]
 
--- a/src/sentry/remote_config/docs/api.md
+++ b/src/sentry/remote_config/docs/api.md
@@ -6,6 +6,44 @@ Host: https://sentry.io/api/0
 
 @cmanallen
 
+## Configuration Proxy [/remote-config/projects/<project_id>/]
+
+Temporary configuration proxy resource.
+
+### Get Configuration [GET]
+
+Fetch a project's configuration. Responses should be proxied exactly to the SDK.
+
+- Response 200
+
+  - Headers
+
+    Cache-Control: public, max-age=3600
+    Content-Type: application/json
+    ETag: a7966bf58e23583c9a5a4059383ff850
+
+  - Body
+
+    ```json
+    {
+      "features": [
+        {
+          "key": "hello",
+          "value": "world"
+        },
+        {
+          "key": "has_access",
+          "value": true
+        }
+      ],
+      "options": {
+        "sample_rate": 1.0,
+        "traces_sample_rate": 0.5
+      },
+      "version": 1
+    }
+    ```
+
 ## Configuration [/projects/<organization_id_or_slug>/<project_id_or_slug>/configuration/]
 
 ### Get Configuration [GET]
@@ -117,41 +155,3 @@ Set the project's configuration.
 Delete the project's configuration.
 
 - Response 204
-
-## Configuration Proxy [/remote-config/projects/<project_id>/]
-
-Temporary configuration proxy resource.
-
-### Get Configuration [GET]
-
-Fetch a project's configuration. Responses should be proxied exactly to the SDK.
-
-- Response 200
-
-  - Headers
-
-    Cache-Control: public, max-age=3600
-    Content-Type: application/json
-    ETag: a7966bf58e23583c9a5a4059383ff850
-
-  - Body
-
-    ```json
-    {
-      "features": [
-        {
-          "key": "hello",
-          "value": "world"
-        },
-        {
-          "key": "has_access",
-          "value": true
-        }
-      ],
-      "options": {
-        "sample_rate": 1.0,
-        "traces_sample_rate": 0.5
-      },
-      "version": 1
-    }
-    ```
--- a/src/sentry/remote_config/docs/api.md
+++ b/src/sentry/remote_config/docs/api.md
@@ -182,6 +182,7 @@ Retrieve configuration features.
     "data": [
       {
         "description": "A feature description.",
+        "id": "59d105e1872d4280b5206015d68e95a8",
         "is_enabled": true,
         "key": "hello",
         "value": 22.3
@@ -213,6 +214,7 @@ Create a configuration feature.
   {
     "data": {
       "description": "Another key.",
+      "id": "59d105e1872d4280b5206015d68e95a8",
       "is_enabled": true,
       "key": "other",
       "value": "key"
@@ -232,6 +234,7 @@ Retrieve a configuration feature.
   {
     "data": {
       "description": "A feature description.",
+      "id": "59d105e1872d4280b5206015d68e95a8",
       "is_enabled": true,
       "key": "hello",
       "value": 22.3
@@ -259,6 +262,7 @@ Update a configuration feature.
   {
     "data": {
       "description": "A feature description.",
+      "id": "59d105e1872d4280b5206015d68e95a8",
       "is_enabled": false,
       "key": "hello",
       "value": 22.3
--- a/src/sentry/remote_config/docs/api.md
+++ b/src/sentry/remote_config/docs/api.md
@@ -49,7 +49,7 @@ Fetch a project's configuration. Responses should be proxied exactly to the SDK.
 | Field              | Type             | Description                                                                   |
 | ------------------ | ---------------- | ----------------------------------------------------------------------------- |
 | environment        | optional[string] | The environment the configuration is associated with.                         |
-| id                 | string           | A server generted unique identifier.                                          |
+| id                 | string           | A server generated unique identifier.                                          |
 | name               | optional[string] | A custom name distinguishing the configuration from the default project name. |
 | sample_rate        | number           |                                                                               |
 | traces_sample_rate | number           |                                                                               |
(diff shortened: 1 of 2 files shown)


src/sentry/remote_config/docs/api.md
```diff
@@ -53,105 +39,231 @@ Retrieve the project's configuration.
...
+      "version": 1
+    }
+    ```
+
+## Configurations [/projects/<organization_id_or_slug>/<project_id_or_slug>/configurations/]
+
+| Field              | Type             | Description                                                                   |
+| ------------------ | ---------------- | ----------------------------------------------------------------------------- |
+| environment        | optional[string] | The environment the configuration is associated with.                         |
+| id                 | string           | A server generted unique identifier.                                          |
```
billyvg, line 52:
```suggestion
| id                 | string           | A server generated unique identifier.                                          |
```

src/sentry/remote_config/docs/api.md
```diff
@@ -53,105 +39,231 @@ Retrieve the project's configuration.
...
+        "traces_sample_rate": 0.75,
+        "version": 1
       }
+    ]
+  }
+  ```
+
+### Create Configuration [POST]
+
+Create a new configuration. The version field is ignored on create. Version is initialized to 1.
```
ryan953, line 81: is `version` an auto-increment whenever a change is made? or is it like a schema-version?

cmanallen: Its manually incremented by user update.  Its to prevent concurrent writes.  The integer version can be associated with errors/traces/whatever and act as a version identifier but that's not totally its intent.

src/sentry/remote_config/docs/api.md
```diff
@@ -53,105 +39,231 @@ Retrieve the project's configuration.
...
+      },
+      "version": 1
+    }
+    ```
+
+## Configurations [/projects/<organization_id_or_slug>/<project_id_or_slug>/configurations/]
+
+| Field              | Type             | Description                                                                   |
+| ------------------ | ---------------- | ----------------------------------------------------------------------------- |
+| environment        | optional[string] | The environment the configuration is associated with.                         |
```
ryan953, line 51: Does this field need to be unique across all the configs inside a given project? 

That opens a can of worms if not... like flattening configs or something idk... there's no order so flattening here isn't gonna work

cmanallen: There should be a unique constraint on `(project_id, environment, name)`.  Definitely no merging/flattening.

src/sentry/remote_config/docs/api.md
```diff
@@ -53,105 +39,231 @@ Retrieve the project's configuration.
...
+      "traces_sample_rate": 0.5
+    }
+  }
+  ```
+
+- Response 201
+
+  ```json
+  {
+    "data": {
```
ryan953, line 100: the GET is returning `data: [ {...}, {...} ]` (arraY) and the POST returns `data: { ... }` (same object as `GET /projects/<organization_id_or_slug>/<project_id_or_slug>/configurations/<configuration_id>/` it seems)

Just writing for clarity.

src/sentry/remote_config/docs/api.md
```diff
@@ -53,105 +39,231 @@ Retrieve the project's configuration.
...
-### Set Configuration [POST]
+  ```json
+  {
+    "data": {
+      "environment": null,
+      "id": "0b88ac27a7b444a6baeb312c0493aed5",
+      "sample_rate": 0,
+      "name": null,
+      "traces_sample_rate": 0.75,
+      "version": 1
```
ryan953, line 127: Can we alpha-sort this, and all the examples so it's easier to compare fields

cmanallen: How dare you.

cmanallen: (consider it done)

src/sentry/remote_config/docs/api.md
ryan953: `environment` `id` and `name` also seem to be a trio to look at. 

Maybe we should be looking at a different url scheme:
1. `GET /projects/<organization_id_or_slug>/<project_id_or_slug>/configurations/` returns a list of envs basically (still the same config objects, but the unique constraint is built-in)
2. Rename other endpoint from `<configuration_id>` to `<env>`: `GET /projects/<organization_id_or_slug>/<project_id_or_slug>/configurations/<env>`
2. Drop `name` field from inside the body. It's basically the same as concatenating project+env. Could be added back later.
3. Drop `id` field from inside the body. Unless it's really needed by the db (and it could be!) but it's already replaced by the url at this point.

still have to deal with the `null` environment example. But that could be where merging comes back into play. What's better is that it serves as a special default/fallback.

I think on the topic of env names changing... we shouldn't try to be smart about that. People should be able to create a config for an env that doesn't exist yet. And if the SDK stops sending data to an env we shouldn't move the config or do anything with it either. Instead we can show volume of requests against an env to let people delete things that are unused if needed.

src/sentry/remote_config/docs/api.md
```diff
@@ -53,105 +43,231 @@ Retrieve the project's configuration.
...
+- Response 200
+
+  ```json
+  {
+    "data": [
+      {
+        "description": "A feature description.",
+        "is_enabled": true,
+        "key": "hello",
+        "value": 22.3
```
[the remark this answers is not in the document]
cmanallen: > unless we also have a type: 'number' field to tell the SDK and the Settings UI about it

You should have that since the type of the value field would be a number.

> But when the user comes back to edit the form we wouldn't know their preference unless we inspect each value... and that would be annoying if the user wanted to change the value during an edit session...

Ultimately I want the user to have a simple UI experience.  I don't want them to paste JSON or to type the keyword `true` when they want to control if the feature is on or off.  On update if we can inspect the type and provide a simple interface to interact with then great.  On create (and update when changing types) a type-drop-down or something could be used to generate a helpful interface.

> I guess during an edit session the user could just blow away an existing key/value pair then create a new one with a different type and then click save..

I want to avoid this because eventually people will have complicated rule-sets and roll-outs and variants and the rest.  They won't want to lose that data.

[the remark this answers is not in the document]
cmanallen: The SDK knows the type because the user encodes it into their request.  E.g. `sdk.get_boolean_value(key)`

src/sentry/remote_config/docs/api.md
```diff
@@ -6,40 +6,30 @@ Host: https://sentry.io/api/0
...
 
-**Attributes**
+Temporary configuration proxy resource.
 
-| Column   | Type           | Description                                   |
-| -------- | -------------- | --------------------------------------------- |
-| features | array[Feature] | Custom, user-defined configuration container. |
-| options  | Option         | Sentry SDK options container.                 |
+- Parameters
+  - environment (optional, string) - The environment to scope the configuration by.
```
ryan953, line 14: if this is optional, then the `env=null` config becomes important.

if it's not optional, and the env is missing from the database are we falling back to the `null` config or nothing?

src/sentry/remote_config/docs/api.md
```diff
@@ -53,105 +43,231 @@ Retrieve the project's configuration.
...
+
+- Response 201
+
+  ```json
+  {
+    "data": {
+      "description": "Another key.",
+      "is_enabled": true,
+      "key": "other",
+      "value": "key"
```
ryan953, line 218: We should be returning the `id` field with the response here, and with the other feature objects returned.

src/sentry/remote_config/docs/api.md
```diff
@@ -53,105 +43,231 @@ Retrieve the project's configuration.
...
 - Response 204
 
-## Configuration Proxy [/remote-config/projects/<project_id>/]
+## Features [/projects/<organization_id_or_slug>/<project_id_or_slug>/configuration/<configuration_id>/features/]
 
-Temporary configuration proxy resource.
+### Get Features [GET]
 
-### Get Configuration [GET]
+Retrieve configuration features.
```
ryan953, line 176: Should we document the pagination params here?

src/sentry/remote_config/docs/api.md
```diff
@@ -53,105 +43,231 @@ Retrieve the project's configuration.
...
+| environment        | optional[string] | The environment the configuration is associated with.                         |
+| id                 | string           | A server generated unique identifier.                                         |
+| name               | optional[string] | A custom name distinguishing the configuration from the default project name. |
+| sample_rate        | number           |                                                                               |
+| traces_sample_rate | number           |                                                                               |
+| version            | number           | The current configuration version. Initialized to 1.                          |
+
+### Get Configurations [GET]
+
+Retrieve configurations.
```
ryan953, line 64: Should we document the pagination params here?

src/sentry/remote_config/docs/api.md
cmanallen: Its to "help" prevent concurrent writes.  This doesn't actually prevent it without a lock.  I need to think more about how to handle concurrent publishes.

src/sentry/remote_config/docs/api.md
cmanallen: Right now I'm not sure what to do.  Environment is optionally defined so it is optional.  We can have a null config or fallback to a default like "production".  I think falling back to "production" is best but I'm not totally sure.  lmk your thoughts.

src/sentry/remote_config/docs/api.md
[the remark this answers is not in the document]
ryan953: > The SDK knows the type because the user encodes it into their request. E.g. sdk.get_boolean_value(key)
IMO, best case is the user has picked a request method (`get_boolean`) and the value needs to be coerced into that type and returned.

But there are issues with that, i think i mentioned type coersion before in a comment, can't find it just now. But we have to be aware of different deployed app versions with different request methods, and current/cached values that don't conform to those expected values within the app.

For example, if the code says: `sdk.get_number('isEnabled')`
and the payload has: `{ features: [ {key: 'isEnabled', value: true } ] }` then what?
its totally possible for a user to fix the payload, changing it from `true` to `1.0`
it's also possible for there to be two app versions out there (iOS, etc) that call `get_boolean('isEnabled')` and also `get_number('isEnabled')`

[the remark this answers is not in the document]
cmanallen: If you have two different app versions with two different type expectations it seems like you have the opportunity and the motive to add a new feature flag which supports the type you want to use.

IMO, the better behavior is to fail and return the default in the event of a type mismatch.  Because while yes you can coerce some values to some types, in practice I think this leads to a lot of work for SDK engineers, a lot of opportunities for mismatch between SDKs (and SDK versions), and, when its working correctly, a lot of undefined or magic behavior that customer's will guess at rather than know for certain.

If nothing else we should be strict to start and loosen our requirements later when there is a clear demand for it.

> current/cached values that don't conform to those expected values within the app.

I don't think caching is the primary concern here.  This is a migration problem.

If you want to migrate a type the consumer of the type needs to be updated to handle both the new and old case.  Then the producer is updated to the new case.  Finally the consumer drops the old case.  Coercion fixes this in that you don't need to explicitly handle both cases but at the expense of hidden behavior.

IMO, type migration will be uncommon so optimizing for it at the expense of a simplified, direct behavior in the common case is a mistake.

> its totally possible for a user to fix the payload, changing it from true to 1.0
it's also possible for there to be two app versions out there (iOS, etc) that call get_boolean('isEnabled') and also get_number('isEnabled')

If you have a legacy installation using the wrong type, you recognize its the wrong type, and want to change it then the appropriate solution is to either coerce it in your code base explicitly (e.g. `toFloat(get_boolean(key))` or `toBool(get_number(key))`) or migrate to a new key (e.g. `isEnabled` and `isEnabledFixed`).

---

One final thought on coercion.  If coercion is important then we should provide an untyped interface `sdk.get_value(key)` which will return anything.  The user can explicitly or the programming language implicitly can coerce the result however they like depending on the context.  The SDK should remove itself from that situation.

PR #72436, opened 2024-06-10
ref(slack): use sdk client in handle_parent_notification

cathteng:
Adds feature-flagged used of Slack SDK Client (initialized in `notify_all_threads_for_activity`) and adds tests for `notify_all_threads_for_activity` where there were previously none.

### DIFF
--- a/tests/sentry/integrations/slack/service/test_slack_service.py
+++ b/tests/sentry/integrations/slack/service/test_slack_service.py
@@ -49,6 +49,7 @@ class TestGetNotificationMessageToSend(TestCase):
         assert result == "admin@localhost archived BAR-1"
 
 
+@mock.patch("sentry.integrations.slack.service")
 class TestNotifyAllThreadsForActivity(TestCase):
     def setUp(self) -> None:
         self.service = SlackService.default()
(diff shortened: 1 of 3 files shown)


ykamo001 (approved):

tests/sentry/integrations/slack/service/test_slack_service.py
```diff
@@ -40,6 +49,114 @@ def test_simple(self) -> None:
...
+                name="slack",
+                provider="slack",
+                external_id="slack:1",
+                metadata={"access_token": "xoxb-access-token"},
+            )
+
+    def test_none_group(self, mock_logger):
+        self.activity.update(group=None)
+        self.service.notify_all_threads_for_activity(activity=self.activity)
+        assert mock_logger.info.called_with("no group associated on the activity, nothing to do")
```
asottile-sentry, line 100: this assertion is doing nothing -- you either want `mock_logger.info.assert_called_with(...)` or to assert on `call_args_list`

~generally you want to make sure your test fails _before_ the implementation such that you don't end up with tests which aren't actually testing anything

additionally, `caplog` is usually easier to use than mocking a logger -- and is built into pytest!

PR #72440, opened 2024-06-10
chore(hybridcloud) Remove an unused RPC method

markstory:
This RPC method has no more callers as all callsites were removed in #72346


corps (approved):

corps:
May run into backwards incompatibility issues with our api checks.

PR #72441, opened 2024-06-10
feat(traces): Make separate graphs for traces

k-fish:
### Summary
This makes a separate line for each set of span conditions. Doing long period counts for the sum of 'within' each trace condition matches is computationally expensive and likely to be too slow. This way you can at least see each set of conditions.

### Screenshot
Screenshot 2024-06-10 at 3 09 11 PM

### DIFF
--- a/static/app/views/traces/tracesChart.tsx
+++ b/static/app/views/traces/tracesChart.tsx
@@ -65,7 +65,7 @@ export function TracesChart({}: Props) {
   return (
     <ChartContainer>
       <ChartPanel
-        title={areQueriesEmpty(queries) ? t('Total Spans') : t('All Matching Spans')}
+        title={areQueriesEmpty(queries) ? t('All Spans') : t('All Matching Spans')}
       >
         <Chart
           height={CHART_HEIGHT}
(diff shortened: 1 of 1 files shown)


static/app/views/traces/tracesChart.tsx
```diff
@@ -21,34 +22,50 @@ interface Props {}
...
-      ),
-      yAxis: ['count()'],
-      interval: getInterval(pageFilters.selection.datetime, 'metrics'),
-      overriddenRoute: 'traces-stats',
-    },
-    'api.trace-explorer.stats'
-  );
+  const firstCountSeries = useTraceCountSeries(queries[0]);
+  const secondCountSeries = useTraceCountSeries(queries[1]);
+  const thirdCountSeries = useTraceCountSeries(queries[2]);
```
0Calories, line 32: Are we only ever going to allow a max of 3 queries?

0Calories (approved):
LGTM :shipit:

static/app/views/traces/tracesChart.tsx
Zylphrex, line 32: As we discussed, we should be able to disable the 2nd and 3rd queries if they're unused

PR #72445, opened 2024-06-10
ref(smart-search): Define issue type suggestions outside of component

malwilley:
I want to use the friendly names in the new search component as well, so these are better defined as inputs to the component instead of a special case within it. This is done by changing the predefined values to `SearchGroup`s instead of simple text items.
 
Suggestions should work the same as they did previously:

CleanShot 2024-06-10 at 12 24 48@2x


roggenkemper (dismissed):

roggenkemper (commented):
i'm not seeing the suggestion dropdown anymore on the deploy of this branch
Screenshot 2024-06-10 at 2 15 38 PM

malwilley:
@roggenkemper good catch - needed to modify the new component code to render search group items: 

CleanShot 2024-06-11 at 12 48 23

roggenkemper (approved):

PR #72449, opened 2024-06-10
feat(insights): track hasData on insight module views

KevinL10:
Adds "Insights: .. Page Load" analytic events to track insight module page views and whether the user has data in the module. We use `useHasData` to look back 90-days in `SpanMetrics` with a specific filter for each insight module, to check that there is at least 1 data point.

Note: each separate project selection results in a separate analytic event. If the user is on the "Flask" project and switches over to "Android", we will send two of the same `insights.page_loads....` event type, possibly with a different `has_data` param.


static/app/views/llmMonitoring/landing.tsx
```diff
@@ -25,13 +25,20 @@ import * as ModuleLayout from 'sentry/views/performance/moduleLayout';
...
 import {useModuleBreadcrumbs} from 'sentry/views/performance/utils/useModuleBreadcrumbs';
 
 export function LLMMonitoringPage() {
   const organization = useOrganization();
 
   const crumbs = useModuleBreadcrumbs('ai');
 
+  useHasDataTrackAnalytics(
+    new MutableSearch('span.category:"ai"'),
+    'api.performance.ai.landing-llm-monitoring-onboarding',
```
0Calories, line 38: Another nit, up to you whether you want to change this or not:

I noticed that for all these referrers, we're referring to them as `<module>-onboarding`, but not sure how they relate to onboarding? Should they be `<module>-onboarding` instead?

static/app/views/performance/browser/webVitals/webVitalsLandingPage.tsx
```diff
@@ -73,6 +75,14 @@ export function WebVitalsLandingPage() {
   const fidDeprecationTimestampString =
     moment(FID_DEPRECATION_DATE).format('DD MMMM YYYY');
 
+  useHasDataTrackAnalytics(
+    new MutableSearch(
+      'span.op:[ui.interaction.click,ui.interaction.hover,ui.interaction.drag,ui.interaction.press]'
```
0Calories, line 80: A more exhaustive check on this would be to check for all web vitals rather than just INP, since INP is not supported on all browsers yet. You could do this via the `has` check on all these measurement tags, i.e `has:measurements.cls OR has:measurement.fcp OR has:measurements.inp OR has:measurements.fid OR has:measurements.lcp ...`

image

0Calories (approved):
Nice! :shipit:

static/app/views/performance/browser/webVitals/webVitalsLandingPage.tsx
KevinL10: No option to filter by `has:..` and OR conditionals in useSpanMetrics, so deciding to stick with the INP interactions. INP interactions are supported by all Chromium-browsers; if the user receives at least 1 interaction from a Chrome browser, then the metrics search will return a positive count.

PR #72450, opened 2024-06-10
typing(grouping): Stricter typing for GroupingComponent

armenzg:
This helps understand the code a bit better.

### DIFF
--- a/src/sentry/grouping/component.py
+++ b/src/sentry/grouping/component.py
@@ -1,6 +1,9 @@
-from sentry.grouping.utils import hash_from_values
+from __future__ import annotations
 
-DEFAULT_HINTS = {"salt": "a static salt"}
+from collections.abc import Generator, Sequence
+from typing import Any
+
+from sentry.grouping.utils import hash_from_values
 
 # When a component ID appears here it has a human readable name which also
 # makes it a major component.  A major component is described as such for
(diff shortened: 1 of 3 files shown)


src/sentry/grouping/component.py
```diff
@@ -1,6 +1,9 @@
-from sentry.grouping.utils import hash_from_values
+from __future__ import annotations
+
+from collections.abc import Generator, Sequence
+from typing import Any
 
-DEFAULT_HINTS = {"salt": "a static salt"}
```
armenzg, line 3: I have no idea why there was a need for this dictionary.

src/sentry/grouping/component.py
```diff
@@ -37,44 +41,35 @@ class GroupingComponent:
...
-        self.contributes = None
+        self.hint = hint
+        self.contributes = contributes
         self.variant_provider = variant_provider
-        self.values = []
-        self.tree_label = None
-        self.is_prefix_frame = False
-        self.is_sentinel_frame = False
-
-        self.update(
```
armenzg, line 60: I do not know why we would call `self.update` instead of initializing the lines above.

asottile-sentry (approved):

src/sentry/grouping/component.py
```diff
@@ -37,44 +41,35 @@ class GroupingComponent:
...
+        tree_label: str | None = None,
+        is_prefix_frame: bool | None = None,
+        is_sentinel_frame: bool | None = None,
     ):
         self.id = id
 
         # Default values
-        self.hint = DEFAULT_HINTS.get(id)
-        self.contributes = None
+        self.hint = hint
```
vartec, line 56: to keep it exactly as it worked before, it would need to be 
```suggestion
        self.hint = hint or DEFAULT_HINTS.get(id)
```
Not sure if that default value is actually needed though.

vartec (approved):

PR #72452, opened 2024-06-10
fix(grouping): Provide defaults for `CalculatedHashes` fields

lobsterkatie:
This adds default values for all fields in the `CalculatedHashes` class except for `hashes`. The advantages of doing this are twofold:

- There are a bunch of places were we pass in the default values, not because we actually need them, but because otherwise the constructor crashes. This lets us clean those up.
- By doing this, we can force the `variants` field to always have a value, which then saves us needing to check whether or not it does, especially in places where we know ahead of time it will. (This latter reason is what inspired the change, as those spots were starting to proliferate.)


JoshFerge (approved):
nice change. lgtm

PR #72454, opened 2024-06-10
ref(rules): Percent of sessions affected uses floats

ceorourke:
I began rolling out the delayed rule processor and it triggered this issue: https://sentry.sentry.io/issues/5474453009 - it turns out the `EventFrequencyPercentCondition` value is stored as a float, even though the docs and all tests use integers.


schew2381 (approved):
nice catch, also open PR comments is cool

PR #72455, opened 2024-06-10
fix(insights): sample panel overflowing

DominikB2014:
1. Fixed overflow in transaction title in sample sidebar, vertical center icon / text, and fix overlapping

This applies to the following module, as these appeared to have the copy and pasted sidebar title
1. Requests
2. Queries
3. Assets
4. Caches
5. Queues

| Before | After |
|--------|--------|
| image | image |
| image | image |
|image | image |
|image | image |

7. Removes unused `spanDescription` prop from sampleList and corresponding rendered item (made css stuff a bit easier too)

### DIFF
--- a/static/app/views/performance/cache/samplePanel/samplePanel.tsx
+++ b/static/app/views/performance/cache/samplePanel/samplePanel.tsx
@@ -276,22 +276,20 @@ export function CacheSamplePanel() {
                   tooltip={project.slug}
                 />
               )}
-              <TitleContainer>
-                <Title>
-                  <Link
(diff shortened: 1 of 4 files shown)


gggritso (approved):
Wrapping the transaction under the project avatar also looks a little strange, but one thing at a time! Thanks!

DominikB2014:
> Wrapping the transaction under the project avatar also looks a little strange, but one thing at a time! Thanks!

@gggritso  Yeah totally agree, that's cause we have a media query that removes the grid columns when the screen is small. Would be an easy fix outside this PR

PR #72464, opened 2024-06-10
feat(query-builder): Add better relative date suggestions

### LINKED ISSUE #69728: Evolved Issues Search: Better suggestions for date filter values

malwilley:
Closes https://github.com/getsentry/sentry/issues/69728

Similar to what was done for numeric suggestions (https://github.com/getsentry/sentry/pull/72345) where suggestions are filled in as you type, but also includes descriptions so the user knows what `-14d` and `+1w` means.

Suggestions used to be a simple array of strings, but now are objects with `value` and optional `description`

CleanShot 2024-06-10 at 14 03 52@2x

CleanShot 2024-06-10 at 14 04 08@2x

### DIFF
--- a/static/app/components/searchQueryBuilder/combobox.tsx
+++ b/static/app/components/searchQueryBuilder/combobox.tsx
@@ -496,6 +496,7 @@ const StyledPositionWrapper = styled(PositionWrapper, {
 
 const StyledOverlay = styled(Overlay)`
   max-height: 400px;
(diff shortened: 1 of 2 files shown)


scttcper (approved):

roggenkemper (approved):
have we thought about having a "more than" option in the default dropdown so users know that it is something they can do? instead of just having all "last hour/day/week/etc." options

PR #72465, opened 2024-06-10
fix(span-summary): Incorrect data when scoped to just span op

0Calories:
Fixes a bug where the data in the header of the span summary (avg duration, time spent, span counts) was incorrect when the summary is generic and scoped to a specific span op. This was because it was assuming there is only one data point in the API response, but when the scope is expanded to an entire span operation, there are multiple groups returned.

Fix this simply by adding the data for each group, and fetch the average duration separately


static/app/views/performance/transactionSummary/transactionSpans/spanSummary/content.tsx
```diff
@@ -107,33 +110,62 @@ function SpanSummaryContent(props: ContentProps) {
...
     </Fragment>
   );
 }
+
+function parseSpanHeaderData(data: Partial<SpanMetricsResponse>[]) {
+  if (!data) {
+    return undefined;
+  }
+
+  if (data.length <= 1) {
```
mjq, line 151: nit: `!data` above already protects from the case where `data.length` is `0` so it's not a problem, but in isolation this check looks like a bug in waiting (if `data.length` is `0`, `<= 1` is still true but `data[0]` will error).

_Extremely_ minor.

```suggestion
  if (data.length === 1) {
```

static/app/views/performance/transactionSummary/transactionSpans/spanSummary/spanSummaryHeader.tsx
```diff
@@ -10,11 +10,11 @@ import {formatMetricUsingUnit} from 'sentry/utils/metrics/formatters';
...
 type Props = {
-  avgDuration: number;
-  spanCount: number;
-  spanDescription: string;
-  spanOp: string;
-  timeSpent: number;
+  avgDuration: number | undefined;
+  spanCount: number | undefined;
+  spanDescription: string | undefined;
+  spanOp: string | undefined;
```
mjq, line 16: Is there a case where `spanOp` might be `undefined` now? (The other ones are clear from the changes, just not sure about that one.)

mjq (approved):
One q, one nit, looks fine to me either way :+1:

static/app/views/performance/transactionSummary/transactionSpans/spanSummary/spanSummaryHeader.tsx
0Calories: `spanOp` should always be returned AFAIK, but since the header isn't handling any loading state at the moment it could be `undefined` if the query has not finished. Speaking of which though, I should probably make use of the loading state in this case instead

static/app/views/performance/transactionSummary/transactionSpans/spanSummary/content.tsx
0Calories: Good callout, yeah this is pretty wonky. `!data` resolves to `false` though if `data.length` is `0` (thanks JavaScript) so I'll change the first conditional to `if (!data || data.length === 0)` as well as your suggestion here!

PR #72466, opened 2024-06-10
ref(slack): use sdk client to send metric alerts

cathteng:
Adds feature-flagged usage of Slack SDK client to send metric alerts. Also adds tests for the code area.

### DIFF
--- a/src/sentry/integrations/slack/utils/notifications.py
+++ b/src/sentry/integrations/slack/utils/notifications.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
 from collections.abc import Mapping
-from typing import Any
 
 import orjson
 import sentry_sdk
+from slack_sdk.errors import SlackApiError
 
 from sentry import features
 from sentry.constants import METRIC_ALERTS_THREAD_DEFAULT, ObjectStatus
@@ -18,6 +18,7 @@ from sentry.integrations.repository.metric_alert import (
 )
 from sentry.integrations.slack.client import SlackClient
 from sentry.integrations.slack.message_builder.incidents import SlackIncidentsMessageBuilder
+from sentry.integrations.slack.sdk_client import SlackSdkClient
 from sentry.models.integrations.integration import Integration
 from sentry.models.options.organization_option import OrganizationOption
 from sentry.services.hybrid_cloud.integration import integration_service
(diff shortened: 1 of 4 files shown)


ykamo001 (approved):

src/sentry/integrations/slack/utils/notifications.py
```diff
@@ -111,40 +115,78 @@ def send_incident_alert_notification(
...
+                "incident_status": new_status,
+            }
+            logger.info("slack.metric_alert.error", exc_info=True, extra=log_params)
+        else:
+            success = True
+
+            ts = None
+            response_data = sdk_response.data
+            if isinstance(response_data, dict):
+                ts = response_data.get("ts")
```
ykamo001, line 150: is there not an easier way to grab this from the `sdk_response` object?

cathteng: oh yeah i think i can do response.get("ts")

ykamo001: Ahh, I was hoping they had an actual method on the dataclass object due to the SDK! Let me see, it's not a blocker for now though

ykamo001:
I know this was on the other PR, but could we have used the validate function/method instead of the `check_response_success` and `check_response_error` functions? https://github.com/slackapi/python-slack-sdk/blob/main/slack_sdk/web/slack_response.py#L186

src/sentry/integrations/slack/utils/notifications.py
```diff
@@ -111,40 +115,74 @@ def send_incident_alert_notification(
...
+    if features.has("organizations:slack-sdk-metric-alert", organization):
+        try:
+            sdk_client = SlackSdkClient(integration_id=integration.id)
+            sdk_response = sdk_client.chat_postMessage(
+                blocks=json_blocks,
+                text=text,
+                channel=str(channel),
+                thread_ts=thread_ts,
+                reply_broadcast=reply_broadcast,
+            )
```
ykamo001, line 128: we should add the `sdk_response.validate()` call here: https://github.com/slackapi/python-slack-sdk/blob/main/slack_sdk/web/slack_response.py#L186

cathteng:
@ykamo001 validate raises SlackApiError for 429s and 500s, the tech spec seems to not want to count those as errors

ykamo001:
> @ykamo001 validate raises SlackApiError for 429s and 500s, the tech spec seems to not want to count those as errors

Ahh, can we still leverage the call though? This way we get the side-effect benefits of the validation, and as they expand the SDK, the improvements will continue to come.

src/sentry/integrations/slack/utils/notifications.py
cathteng: shouldn't we do this in the client, when we wrap all the functions?

ykamo001: It depends where we want to do error handling, and how we want to do error raising. Let's take a look at where it would fit the best

cathteng: actually, it already calls `validate` in the base client https://github.com/slackapi/python-slack-sdk/blob/main/slack_sdk/web/base_client.py#L311-L319

PR #72468, opened 2024-06-10
fix(assignee-selector): Fix avatar not rendering issue

MichaelSun48:
This PR fixes a bug where the avatar badge would not render any image for the avatar, like in this photo:

image

This was caused by an assigned actor not being present in the member list, and our actorAvatar component not handling this edge case properly. This situation could have occurred if, for example, a user was assigned to an issue, but was then removed from an organization, thus removing them from the org's memberlist.

The new behavior simply adds a fall back to render a letter avatar if the user cannot be found in the member list : 

image

In the future, we may want to consider some special visual treatment in this case, just to make it clear that the assignee is no longer part of the organization

### DIFF
--- a/static/app/components/avatar/actorAvatar.spec.tsx
+++ b/static/app/components/avatar/actorAvatar.spec.tsx
@@ -56,6 +56,20 @@ describe('ActorAvatar', function () {
     expect(screen.getByText('CT')).toBeInTheDocument();
   });
 
+  it('should show an avatar even if the user is not in the memberlist', function () {
+    render(
+      <ActorAvatar
+        actor={{
+          id: '2',
+          name: 'Jane Vloggs',
+          type: 'user',
+        }}
+      />
+    );
+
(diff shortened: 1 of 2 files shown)


scttcper:
Can we add a test for this

MichaelSun48:
> Can we add a test for this

Yup I can throw one in for the assignee selector, but the root issue of this is that user is an optional parameter in `<UserBadge/>`, and the component [silently returns null ](https://github.com/getsentry/sentry/blob/6ee0d9297dcef0a39d7bb5610fde1aecb1d6cbb0/static/app/components/avatar/userAvatar.tsx#L43)when user is not passed in, obfuscating this error. I gave a quick look into how easy it would be to switch it to be required, and it's not a quick refactor.

scttcper:
@MichaelSun48 i'd test static/app/components/avatar/actorAvatar.tsx directly

roggenkemper (approved):

scttcper (approved):

PR #72469, opened 2024-06-10
feat(performance): Update starfish charts to show incomplete data points with dotted line

edwardgou-sentry:
Currently we provide a `preserveIncompletePoints` option to control displaying the last data point in insights charts. This is because the last datapoint often contains incomplete data.

This update removes the `preserveIncompletePoints` option and updates insights and trace view charts to always display incomplete data points with a dotted line. This is the same behaviour as the Metrics chart.

Reuses `createIngestionSeries` and `getIngestionDelayBucketCount` from the Metrics chart.

image
image

### DIFF
--- a/static/app/views/starfish/components/chart.tsx
+++ b/static/app/views/starfish/components/chart.tsx
@@ -1,5 +1,5 @@
 import type {RefObject} from 'react';
-import {useEffect, useRef, useState} from 'react';
+import {useEffect, useMemo, useRef, useState} from 'react';
 import {useTheme} from '@emotion/react';
 import styled from '@emotion/styled';
 import type {LineSeriesOption} from 'echarts';
@@ -13,7 +13,6 @@ import type {
 } from 'echarts/types/dist/shared';
 import max from 'lodash/max';
 import min from 'lodash/min';
-import moment from 'moment';
 
 import type {AreaChartProps} from 'sentry/components/charts/areaChart';
 import {AreaChart} from 'sentry/components/charts/areaChart';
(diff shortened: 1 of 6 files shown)


narsaynorath (approved):

static/app/views/starfish/components/chart.tsx
```diff
@@ -198,6 +201,45 @@ function Chart({
...
+    yAxisIndex: 0,
+    xAxisIndex: 0,
+    id: values.seriesName,
+    color: colors[index],
+  }));
+  let incompleteSeries: Series[] = [];
+
+  const bucketSize =
+    new Date(series[0]?.data[1]?.name).getTime() -
+    new Date(series[0]?.data[0]?.name).getTime();
```
Zylphrex, line 215: Are we guaranteed to have 2 data points here? `new Date(undefined).getTime()` returns `NaN`.

static/app/views/starfish/components/chart.tsx
```diff
@@ -370,6 +407,16 @@ function Chart({
...
+                  ...incompleteSeries.map(({seriesName, data: seriesData, ...options}) =>
+                    LineSeries({
+                      ...options,
+                      name: seriesName,
+                      data: seriesData?.map(({value, name}) => [name, value]),
+                      animation: false,
+                      animationThreshold: 1,
+                      animationDuration: 0,
+                    })
+                  ),
```
Zylphrex, line 419: Should this be placed between `series` and `scatterPlot`? IIRC, echarts stack the charts on top of one another and this can place the incomplete series on top of the scatter plot.

static/app/views/starfish/components/chart.tsx
edwardgou-sentry: Good question. I added a check below to just assume a delay bucket count of `1` if we're not able to get `bucketSize`:

https://github.com/getsentry/sentry/pull/72469/files#diff-c743a95f0d4036d070f8488ec9eda1a53b4c4c8b473d8aff81ad68f0fb9469f1R219

I think it's pretty unlikely this will happen, and we probably short circuit the chart somewhere else in this scenario, but this should work

static/app/views/starfish/components/chart.tsx
edwardgou-sentry: Yeah this will have the side effect of overlapping on top of any scatter plots. But IMO the effect is barely noticeable:
image

I opted to put the series at the end to minimize code change, because we have some charts + hover handlers that always assume the scatter plot comes after the initial line series.

lmk if you have a strong opinion on this though

PR #72481, opened 2024-06-11
fix(hybridcloud) Add cors headers to route error response

markstory:
Our API routes include a catch all which is separate from django's error handlers. This catchall route should include CORS headers so that missing route errors don't also cause CORS errors.

Fixes HC-1006

### DIFF
--- a/src/sentry/api/endpoints/catchall.py
+++ b/src/sentry/api/endpoints/catchall.py
@@ -2,7 +2,7 @@ from django.http import HttpResponse, JsonResponse
 from django.views.decorators.csrf import csrf_exempt
 from rest_framework.request import Request
 
-from sentry.api.base import Endpoint, all_silo_endpoint
+from sentry.api.base import Endpoint, all_silo_endpoint, allow_cors_options
 
 
 @all_silo_endpoint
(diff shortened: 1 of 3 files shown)


src/sentry/api/endpoints/catchall.py
```diff
@@ -2,14 +2,15 @@
...
-from sentry.api.base import Endpoint, all_silo_endpoint
+from sentry.api.base import Endpoint, all_silo_endpoint, allow_cors_options
 
 
 @all_silo_endpoint
 class CatchallEndpoint(Endpoint):
     permission_classes = ()
 
     @csrf_exempt
+    @allow_cors_options
```
asottile-sentry, line 13: I think this "fixed" the mypy errors list because these are supposed to be method decorators via `@method_decorator(...)` and this is not actually working properly 🤔

asottile-sentry: that or `allow_cors_options` is untyped so it hides an error behind `Any`

dashed (approved):

src/sentry/api/endpoints/catchall.py
markstory: allow_cors_options is currently untyped :cry: However, the decorator is doing its job of adding the necessary headers.

PR #72482, opened 2024-06-11
ref(ui): Readonly selected group store state

scttcper:

makes state readonly, returns the same object from getState until state changes.

### DIFF
--- a/static/app/stores/selectedGroupStore.tsx
+++ b/static/app/stores/selectedGroupStore.tsx
@@ -2,7 +2,7 @@ import {createStore} from 'reflux';
 
 import GroupStore from 'sentry/stores/groupStore';
 
-import type {CommonStoreDefinition} from './types';
+import type {StrictStoreDefinition} from './types';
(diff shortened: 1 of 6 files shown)


malwilley (changes requested):
I'm not sure what's wrong here but selecting issues on the stream isn't working with these changes

static/app/stores/selectedGroupStore.tsx
```diff
@@ -48,12 +43,11 @@ const storeConfig: SelectedGroupStoreDefinition = {
...
 
   reset() {
-    this.records = new Map();
-    this.lastSelected = null;
+    this.state = {records: new Map(), lastSelected: null};
   },
 
   getState() {
-    return Object.fromEntries(this.records);
+    return this.state;
```
malwilley, line 50: Oh it's probably this returning different data, the issue stream component using it needs to be changed

malwilley (approved):

PR #72483, opened 2024-06-11
ref(rules): Render Slack issue alert label better

ceorourke:
Currently the text displayed for Slack actions in issue alerts looks pretty bad: 

We always show the tags and notes fields whether or not there's anything entered there. This is partially because of how we're using Django forms for issue alerts, but we can format it to look nicer.

**With no tags or notes**
Screenshot 2024-06-10 at 5 44 52 PM
**With tags**
Screenshot 2024-06-10 at 5 45 07 PM
**With tags and notes**
Screenshot 2024-06-10 at 5 45 27 PM
**With only notes**
Screenshot 2024-06-10 at 5 45 38 PM

Rule edit confirmation notifications look a lot cleaner too - here's an example from before:
image

And here's after
Screenshot 2024-06-10 at 5 45 50 PM
Screenshot 2024-06-10 at 5 45 56 PM
Screenshot 2024-06-10 at 5 46 01 PM

Closes https://github.com/getsentry/team-core-product-foundations/issues/293

### DIFF
--- a/src/sentry/integrations/slack/actions/notification.py
+++ b/src/sentry/integrations/slack/actions/notification.py
@@ -359,14 +359,28 @@ class SlackNotifyServiceAction(IntegrationEventAction):
 
     def render_label(self) -> str:
         tags = self.get_tags_list()
+        channel = self.get_option("channel").lstrip("#")
+        workspace = self.get_integration_name()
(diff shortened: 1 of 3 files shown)


schew2381 (approved):
lgtm!

PR #72489, opened 2024-06-11
ref: fix spans_metrics dataset types and remove from mypy ignore list

viglia:


src/sentry/search/events/fields.py
```diff
@@ -2165,7 +2165,7 @@ class MetricArg(FunctionArg):
     def __init__(
         self,
         name: str,
-        allowed_columns: Sequence[str] | None = None,
+        allowed_columns: Sequence[str] | set[str] | None = None,
```
asottile-sentry, line 2168: I believe this should be something like `Iterable[str]` or `Container[str]` instead of `Sequence | set` (which is usually a mistake)

src/sentry/search/events/datasets/spans_metrics.py
```diff
@@ -731,10 +731,15 @@ def _resolve_total_span_duration(self, alias: str, scope: str, column: str) -> S
     def _resolve_time_spent_percentage(
         self, args: Mapping[str, str | Column | SelectType | int | float], alias: str
     ) -> SelectType:
+        scope = args["scope"]
+        column = args["column"]
```
asottile-sentry, line 735: I suspect that `args` should be a `TypedDict` here and then the arguments can be safely passed and unpacked without needing `isinstance` narrowing

asottile-sentry (approved):

PR #72501, opened 2024-06-11
feat(insights): add search analytics to insight modules

KevinL10:
Adds search analytics for all insight modules.


0Calories (approved):
👌

PR #72503, opened 2024-06-11
chore(replays): Log oversized segments

cmanallen:
Gives visibility into oversized segments.

### DIFF
--- a/src/sentry/replays/usecases/ingest/__init__.py
+++ b/src/sentry/replays/usecases/ingest/__init__.py
@@ -218,7 +218,7 @@ def recording_post_processor(
     if len(segment_bytes) >= 4_000_000:
         logger.info(
             # Logging to the sentry.replays.slow_click namespace because
-            # its the only one configured to use BigQuery at the moment.
+            # it's the only one configured to use BigQuery at the moment.
             #
             # NOTE: Needs an ops request if we want to create a new dataset.
             "sentry.replays.slow_click",
--- a/src/sentry/replays/usecases/ingest/__init__.py
+++ b/src/sentry/replays/usecases/ingest/__init__.py
@@ -215,6 +215,22 @@ def recording_post_processor(
     replay_event_bytes: bytes | None,
     transaction: Span,
 ) -> None:
+    if len(segment_bytes) >= 4_000_000:
+        logger.info(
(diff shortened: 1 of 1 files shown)


src/sentry/replays/usecases/ingest/__init__.py
```diff
@@ -215,6 +215,22 @@ def recording_post_processor(
     replay_event_bytes: bytes | None,
     transaction: Span,
 ) -> None:
+    if len(segment_bytes) >= 4_000_000:
+        logger.info(
+            # Logging to the sentry.replays.slow_click namespace because
+            # its the only one configured to use BigQuery at the moment.
+            #
+            # NOTE: Needs an ops request if we want to create a new dataset.
```
billyvg, line 223: Would it be worth getting this done for "general" logging?

src/sentry/replays/usecases/ingest/__init__.py
```diff
@@ -215,6 +215,22 @@ def recording_post_processor(
     replay_event_bytes: bytes | None,
     transaction: Span,
 ) -> None:
+    if len(segment_bytes) >= 4_000_000:
+        logger.info(
+            # Logging to the sentry.replays.slow_click namespace because
+            # its the only one configured to use BigQuery at the moment.
```
michellewzhang, line 221:
```suggestion
            # it's the only one configured to use BigQuery at the moment.
```

michellewzhang (approved):

src/sentry/replays/usecases/ingest/__init__.py
```diff
@@ -215,6 +215,22 @@ def recording_post_processor(
     replay_event_bytes: bytes | None,
     transaction: Span,
 ) -> None:
+    if len(segment_bytes) >= 4_000_000:
```
billyvg, line 218: Any reasoning for this size?

billyvg (approved):
just some side comments

src/sentry/replays/usecases/ingest/__init__.py
cmanallen: Yeah that's basically what we use this namespace for anyway.  Could extract this out into a global and not have to worry about comments like these anymore.

PR #72505, opened 2024-06-11
feat(new-trace): Updated breadcrumbs for all trace view entry points.

Abdkhan14:
This PR:
- Updates navigation breadcrumbs from all trace view entry points.
Screenshot 2024-06-11 at 12 19 07 PM


static/app/views/performance/newTraceDetails/traceMetadataHeader.tsx
```diff
@@ -20,16 +25,344 @@ interface TraceMetadataHeaderProps {
...
+
+function getInsightsModuleBreadcrumbs(location: Location, organization: Organization) {
+  const crumbs: Crumb[] = [];
+
+  if (organization.features.includes('performance-insights')) {
+    crumbs.push({
+      label: t('Insights'),
+    });
+
+    switch (location.query.referrer) {
```
scttcper, line 162: are you sure you want to use referrer for this? referrer is often stripped by adblockers to prevent tracking and our app tends to use referrer only for click tracking

Abdkhan14: Ahh I see, @scttcper, help me understand this a bit. Would changing `referrer` to something like `entry_point` work?

scttcper: @Abdkhan14 yeah entryPoint would make sense. or tabConfiguration. Some indication that the query parameter is controlling what is on the page. I wouldn't expect referrer to change anything

Abdkhan14: Got it, I'll make the change.

Abdkhan14: I think `source` fits the best here.

static/app/views/performance/newTraceDetails/traceMetadataHeader.tsx
```diff
@@ -20,16 +25,344 @@ interface TraceMetadataHeaderProps {
   traceSlug: string;
 }
 
+export enum TraceViewReferrers {
```
scttcper, line 28: been trying to use more const enums where possible since they compile down better
```suggestion
export const enum TraceViewReferrers {
```

scttcper (approved):

JonasBa (approved):

PR #72512, opened 2024-06-11
ref(slack): use sdk client to notify recipients

cathteng:
From the docstring for the function
> Send an "activity" or "alert rule" notification to a Slack user or team, but NOT to a channel directly.
> Sending Slack notifications to a channel is in integrations/slack/actions/notification.py

### DIFF
--- a/src/sentry/integrations/slack/notifications.py
+++ b/src/sentry/integrations/slack/notifications.py
@@ -8,6 +8,7 @@ from typing import Any
 import orjson
 import sentry_sdk
 
+from sentry import features
 from sentry.integrations.mixins import NotifyBasicMixin
 from sentry.integrations.notifications import get_context, get_integrations_by_channel_by_recipient
 from sentry.integrations.slack.message_builder import SlackBlock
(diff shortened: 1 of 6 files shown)


src/sentry/integrations/slack/notifications.py
```diff
@@ -109,17 +110,26 @@ def _notify_recipient(
         if SiloMode.get_current_mode() == SiloMode.CONTROL:
             post_message_task = post_message_control
 
+        has_sdk_flag = features.has(
+            "organizations:slack-sdk-notify-recipient", notification.organization
+        )
+
```
cathteng, line 116: i have to do the backend equivalent of prop drill with this :(

ykamo001 (approved):

src/sentry/integrations/slack/notifications.py
iamrajjoshi: its gonna be so much nicer once we can remove the ff and delete all of the flags

iamrajjoshi (approved):

PR #72516, opened 2024-06-11
fix(reports): Send weekly report to each user as individual task

RyanSkonnord:
The intent is to prevent duplicate sends, on the theory that errors on individual sends cause the `prepare_organization_report` task to be retried. Now individual sends can be retried in their own Celery task.

TBD: Performance/capacity implementation of increasing number of Celery tasks.


RyanSkonnord:
The correctness of this approach depends on the errors that cause the retries originating in `send_email`, not `prepare_organization_report`. [SENTRY-2RBR](https://sentry.sentry.io/issues/5012013434/) suggests that this is not the case.

PR #53286, opened 2023-07-20
ref(repository): remove repo org_id + name constraint

cathteng:
From Steve on #53024
> I believe this will cause some bad states because we have uniqueness constraints on the repo that cause bugs that occasionally have to be fixed by deleting a repo for another integration: https://github.com/getsentry/sentry/blob/master/src/sentry/models/repository.py#L41
> 
> I think if we changed the uniqueness constraints then we can hide repos but we have to fix that first

PR #72523, opened 2024-06-11
fix(github): update repo name in webhook

cathteng:
The method of updating repo names was removed a while ago because there were too many IntegrityErrors happening. These should occur less frequently because we don't have the constraint on the repo name anymore, it was removed here https://github.com/getsentry/sentry/pull/53286


schew2381 (approved):
lgtm!

PR #72524, opened 2024-06-11
ref(slack): Logging & Metrics in Unlink User Flow

iamrajjoshi:

Addresses `bot-1` task. Added logging and metrics to the flow. 

Will start working on a dashboard for all the bot commands.

### DIFF
--- a/src/sentry/integrations/slack/webhooks/base.py
+++ b/src/sentry/integrations/slack/webhooks/base.py
@@ -19,9 +19,16 @@ UNLINK_USER_MESSAGE = "<{associate_url}|Click here to unlink your identity.>"
 NOT_LINKED_MESSAGE = "You do not have a linked identity to unlink."
 ALREADY_LINKED_MESSAGE = "You are already linked as `{username}`."
 
+from sentry.utils import metrics
+
+from ..utils import logger
+
 
 class SlackDMEndpoint(Endpoint, abc.ABC):
     slack_request_class = SlackDMRequest
+
+    _METRICS_SUCCESS_KEY = "slack.integrations.slack.dm_endpoint.success."
+    _METRIC_FAILURE_KEY = "slack.integrations.slack.dm_endpoint.failure."
 
     def post_dispatcher(self, request: SlackDMRequest) -> Response:
         """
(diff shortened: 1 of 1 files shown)


src/sentry/integrations/slack/webhooks/base.py
```diff
@@ -19,10 +19,17 @@
 NOT_LINKED_MESSAGE = "You do not have a linked identity to unlink."
 ALREADY_LINKED_MESSAGE = "You are already linked as `{username}`."
 
+from sentry.utils import metrics
+
+from ..utils import logger
```
cathteng, line 24: should we do `logger = logging.getLogger(__name__)`

iamrajjoshi: when should we use the common logger + add additional namespace when logging vs make a new one?

src/sentry/integrations/slack/webhooks/base.py
```diff
@@ -19,10 +19,19 @@
...
+from sentry.utils import metrics
+
+logger = logging.getLogger(__name__)
+
 
 class SlackDMEndpoint(Endpoint, abc.ABC):
     slack_request_class = SlackDMRequest
 
+    _METRICS_SUCCESS_KEY = "slack.integrations.slack.dm_endpoint.success."
+    _METRIC_FAILURE_KEY = "slack.integrations.slack.dm_endpoint.failure."
```
cathteng, line 33: going by what we have for discord, i think these should be `sentry.integrations.slack.dm_endpoint.success `and `sentry.integrations.slack.dm_endpoint.failure`

```suggestion
    _METRICS_SUCCESS_KEY = "sentry.integrations.slack.dm_endpoint.success"
    _METRIC_FAILURE_KEY = "sentry.integrations.slack.dm_endpoint.failure"
```

cathteng: for success and failure, instead of emitting a new metric should we group all failures together and differentiate them by tags?

cathteng: e.g. `metrics.incr(self._METRICS_SUCCESS_KEY, tags={"action": "unlink_user"}, sample_rate=1.0)`

cathteng: these are just options, i don't think anything is blocking except the key naming with `sentry` in front

iamrajjoshi: ill update how do do metrics once i figure out which ones are worth keeping after collecting some data.

cathteng: ok

cathteng (approved):

PR #72525, opened 2024-06-11
chore(perf): Adjust flags for suspect spans table

0Calories:
Adjusts some of the feature flag logic for span summary and the suspect spans table:

- Hide the suspect spans table and new span summary for orgs that don't have the `insights-initial-modules` flag, as only these plans will have span metrics ingestion

### DIFF
--- a/static/app/views/performance/transactionSummary/transactionOverview/content.tsx
+++ b/static/app/views/performance/transactionSummary/transactionOverview/content.tsx
@@ -403,7 +403,7 @@ function SummaryContent({
           />
         </PerformanceAtScaleContextProvider>
 
-        {hasNewSpansUIFlag && (
+        {!hasNewSpansUIFlag && (
           <SuspectSpans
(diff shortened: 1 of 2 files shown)


static/app/views/performance/transactionSummary/transactionOverview/content.tsx
```diff
@@ -399,18 +403,21 @@ function SummaryContent({
...
-          eventView={eventView}
-          totals={
-            defined(totalValues?.['count()'])
-              ? {'count()': totalValues!['count()']}
-              : null
-          }
-          projectId={projectId}
-          transactionName={transactionName}
-        />
+        {hasNewSpansUIFlag ? null : (
```
KevinL10, line 406:
```
{hasNewSpansUIFlag && <SuspectSpans />}
```
may be easier to read

KevinL10 (approved):

PR #72533, opened 2024-06-11
feat(flagpole): Adds identity fields to Eval Context for more stable partial rollouts

GabeVillalobos:
Introduces the concept of `identity_fields` to both the evaluation context and the context builder classes in Flagpole.

These fields can be specified when adding a new context transformer, ideally matching the name of one of the properties the transformer provides. These are then provided to the Evaluation context constructor, which uses them to generate a stable object ID that can be used for partial rollout checks.

Because this ID never changes (the context is immutable), I made the ID immutable as well and set it at init time.

This PR also removes the sentry flagpole context file from the flagpole directory as it is no longer used in production, thus cutting all sentry references from Flagpole.


src/flagpole/evaluation_context.py
```diff
@@ -13,40 +13,78 @@
...
-    def id(self) -> int:
+        for field in self.__identity_fields:
+            if field not in self.__data:
+                invalid_identity_fields.append(field)
+
+        if invalid_identity_fields:
+            raise InvalidIdentityFieldException(
+                f"One or more invalid identity fields specified: {invalid_identity_fields}"
+            )
+
```
GabeVillalobos, line 54: This logic probably has to go. Our evaluation context won't always contain project IDs, so I think we just have to allow it to skip over undefined fields and remove them from the identity fields set. Of course, this also means an empty ID set in any way should just use all keys as the ID fields, but this feels like really obscure behavior.

src/flagpole/evaluation_context.py
```diff
@@ -13,40 +13,78 @@
...
+        keys = list(self.__identity_fields)
         vector = []
         for key in sorted(keys):
             vector.append(key)
             vector.append(str(self.__data[key]))
         hashed = hashlib.sha1(":".join(vector).encode("utf8"))
         return int.from_bytes(hashed.digest(), byteorder="big")
 
+    @property
+    def id(self):
```
markstory, line 72:
```suggestion
    def id(self) -> int:
```

You should be able to define the return of a property and have the typings workout.

src/flagpole/evaluation_context.py
```diff
@@ -13,40 +13,78 @@
...
 class EvaluationContext:
     """
     Prepared by the application and passed to flagpole to evaluate
     feature conditions.
     """
 
-    def __init__(self, data: EvaluationContextDict):
+    __data: EvaluationContextDict
+    __identity_fields: set[str]
+    id: int
```
markstory, line 28: I don't think you would need this if you type the return of the `id` property.

src/flagpole/evaluation_context.py
```diff
@@ -61,20 +99,27 @@ class ContextBuilder(Generic[T_CONTEXT_DATA]):
...
 
     def add_context_transformer(
-        self, context_transformer: Callable[[T_CONTEXT_DATA], EvaluationContextDict]
+        self,
+        context_transformer: Callable[[T_CONTEXT_DATA], EvaluationContextDict],
+        identity_fields: list[str] | None = None,
     ) -> ContextBuilder[T_CONTEXT_DATA]:
         self.context_transformers.append(context_transformer)
+        if identity_fields is not None:
+            self.identity_fields.update(identity_fields)
```
markstory, line 121: Does `build()` on `ContextBuilder` need to forward `identity_fields` to the Context?

src/flagpole/evaluation_context.py
markstory: Removing unknown fields from the identity_fields makes sense to me. If we end up with zero identity fields, then the all fields are part of the identity.

src/flagpole/evaluation_context.py
GabeVillalobos: yep! Was planning to write tests around this to ensure we properly generate the same Id for different contexts as well.

src/flagpole/evaluation_context.py
GabeVillalobos: Should be addressed now.

markstory (approved):
Looks good to me. The rollout rates for user scoped features will be wobbly as users update their proflies, but we don't frequently use user scoped features and users infrequently update their profile data.

PR #72542, opened 2024-06-11
ref(seer grouping): Prevent events with hybrid fingerprints from calling Seer

lobsterkatie:
This adds hybrid fingerprints to the check for custom fingerprints we do before calling Seer. (Hybrid fingerprints those which include `{{ default }}` alongside at least one other value, which causes events to group as they normally would, but then further differentiate using the other value(s).)

With this change, events with such fingerprints will stop triggering a Seer call. In the long run, once calling Seer becomes a more integral part of our grouping process, we will likely want to make the Seer call anyway, and somehow incorporate that into the `{{ default }}` part of the fingerprint, but for now this prevents us from overriding grouping which has been specifically set by the user.


JoshFerge (approved):
makes sense!

PR #72548, opened 2024-06-11
ref(tsc): convert adminWarnings to FC

michellewzhang:

### DIFF
--- a/static/app/views/admin/adminWarnings.tsx
+++ b/static/app/views/admin/adminWarnings.tsx
@@ -1,58 +1,57 @@
 import {Fragment} from 'react';
 
+import LoadingIndicator from 'sentry/components/loadingIndicator';
 import {t} from 'sentry/locale';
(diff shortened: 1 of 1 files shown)


evanpurkhiser (approved):

static/app/views/admin/adminWarnings.tsx
```diff
@@ -1,58 +1,57 @@
...
-              {warnings.map((warning, i) => (
-                <li key={i}>{warning}</li>
-              ))}
-            </ul>
-          </Fragment>
-        )}
-      </div>
-    );
+  if (!data || isError) {
+    return null;
```
evanpurkhiser, line 22: Why not RouteError or whatever that component is?

michellewzhang: hmmm... old code just returned `null` so didn't want to change the behavior, but i can update to `LoadingError` if that's better?

evanpurkhiser: ah ok lol

PR #72550, opened 2024-06-11
ref(slack): use sdk client to get team info during integration installation

cathteng:
Replace old slack client with the Slack SDK client when fetching team info during integration installation. There's no `integration_id` available at this point but we do have the access token.

### DIFF
--- a/src/sentry/integrations/slack/integration.py
+++ b/src/sentry/integrations/slack/integration.py
@@ -6,6 +6,8 @@ from typing import Any
 
 from django.utils.translation import gettext_lazy as _
 from django.views import View
+from slack_sdk import WebClient
+from slack_sdk.errors import SlackApiError
 
 from sentry.identity.pipeline import IdentityProviderPipeline
 from sentry.integrations.base import (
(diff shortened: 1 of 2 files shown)


src/sentry/integrations/slack/integration.py
```diff
@@ -133,6 +135,19 @@ def get_pipeline_views(self) -> Sequence[View]:
 
     def _get_team_info(self, access_token: str) -> Any:
         # Manually add authorization since this method is part of slack installation
+
+        # first try with new SDK client (not attached to integration)
+        try:
+            client = WebClient(token=access_token)
```
cathteng, line 141: i went with this because the new `SlackSdkClient` subclasses `WebClient`, and uses `integration_id` to fetch the access token to initialize the parent. we don't have integration_id here because it doesn't exist yet.

if we already have the access token we don't need to use the SlackSdkClient unless we care about tracking the metric via DD. we also don't need to track successes and fails in order to disable the integration here because it doesn't exist

ykamo001 (approved):

PR #72552, opened 2024-06-11
chore(grouping): Remove unused `GroupHashesSplitEndpoint`

lobsterkatie:
In the process of auditing all of the places we access the `GroupHash` table (as prep for making a number of changes to it), I was reminded that we have an endpoint, `GroupHashesSplitEndpoint`, which stopped being used when the last references to it were removed [three years ago](https://github.com/getsentry/sentry/pull/26733). As it's part of the hierarchical grouping system which is slated to be removed anyway, I went ahead and killed it now, so it stops polluting my search results.


iamrajjoshi (approved):

src/sentry/apidocs/api_publish_status_allowlist_dont_modify.py
```diff
@@ -20,7 +20,6 @@
     "/api/0/{var}/{issue_id}/{var}/": {"GET", "POST"},
     "/api/0/{var}/{issue_id}/{var}/{note_id}/": {"DELETE", "PUT"},
     "/api/0/{var}/{issue_id}/hashes/": {"GET", "DELETE"},
-    "/api/0/{var}/{issue_id}/hashes/split/": {"DELETE", "GET", "PUT"},
```
iamrajjoshi, line 23: thank you for updating this 🙏🏽

PR #72561, opened 2024-06-11
feat(replay): Calculate hydration diff timestamps based on related hydration breadcrumbs

ryan953:
Related to https://github.com/getsentry/sentry/issues/70199

### DIFF
--- a/static/app/utils/replays/getDiffTimestamps.tsx
+++ b/static/app/utils/replays/getDiffTimestamps.tsx
@@ -1,0 +1,56 @@
+import type {Event} from 'sentry/types/event';
+import type ReplayReader from 'sentry/utils/replays/replayReader';
+import type {ReplayFrame} from 'sentry/utils/replays/types';
+
+export function getReplayDiffOffsetsFromFrame(
+  replay: ReplayReader | null,
+  frame: ReplayFrame
+) {
+  return {
+    leftOffsetMs: frame.offsetMs,
+    rightOffsetMs:
+      (frame.data.mutations.next?.timestamp ?? 0) -
+      (replay?.getReplay().started_at.getTime() ?? 0),
+  };
+}
+
+export function getReplayDiffOffsetsFromEvent(replay: ReplayReader, event: Event) {
+  const startTimestampMS =
+    'startTimestamp' in event ? event.startTimestamp * 1000 : undefined;
+  const timeOfEvent = event.dateCreated ?? startTimestampMS ?? event.dateReceived;
+  const eventTimestampMs = timeOfEvent ? Math.floor(new Date(timeOfEvent).getTime()) : 0;
+  // `event.dateCreated` is the most common date to use, and it's in seconds not ms
+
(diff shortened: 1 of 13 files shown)


static/app/utils/replays/getDiffTimestamps.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  const frames = replay.getRRWebFrames();
+  const replayStartTimestamp = replay?.getReplay().started_at.getTime() ?? 0;
+
+  const leftReplayFrameIndex = replay
+    .getRRWebFrames()
+    .findIndex(frame => frame.timestamp < eventTimestampMs);
+  const leftFrame = frames.at(Math.max(0, leftReplayFrameIndex));
+  const leftOffsetMs = replayStartTimestamp - (leftFrame?.timestamp ?? 0);
```
michellewzhang, line 45: am i misunderstanding, or is `replayStartTimestamp` smaller than `leftFrame.timestamp`?  so this could be negative?

michellewzhang: (or is it being negative expected)

static/app/utils/replays/getDiffTimestamps.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  const rightReplayFrameIndex = replay
+    .getRRWebFrames()
+    .findLastIndex(frame => frame.timestamp <= eventTimestampMs + 1);
+
+  const rightFrame = frames.at(Math.min(frames.length, rightReplayFrameIndex + 1));
+  const rightOffsetMs =
```
michellewzhang, line 52: why the +1?

static/app/utils/replays/getDiffTimestamps.tsx
ryan953: I set some Math.max() in here, so we'll never get a negative number back.

For the left side, a `0` offset is probably ok to deal with. We could have either an empty html page (unlikely, because we need some html to load the sdk!) or it's just a really early snapshot. 

If the right side is `0` then it's a problem for the diff views... I think in that case those views should deal with that themselves... for example we could show an error to the user. `0` in that case might mean that there's no 2nd snapshot to compare against.

static/app/utils/replays/types.tsx
```diff
@@ -17,6 +17,17 @@ import invariant from 'invariant';
...
+// These stub types should be coming from the sdk, but they're hard-coded until
+// the SDK updates to the latest version... once that happens delete this!
+type StubBreadcrumbTypes = {
+  category: 'replay.hydrate-error';
+  timestamp: number;
+  type: '';
+  data?: {
+    url?: string;
+  };
+};
```
ryan953, line 29: depends on https://github.com/getsentry/sentry-javascript/pull/12521 being released

michellewzhang (approved):

michellewzhang:
nice tests👍

PR #72562, opened 2024-06-11
feat(ui): Add loading indicator to page filters wrapper

scttcper:
Previously showed a blank screen until projects were loaded.

### DIFF
--- a/static/app/components/organizations/pageFilters/container.tsx
+++ b/static/app/components/organizations/pageFilters/container.tsx
@@ -120,7 +120,6 @@ function PageFiltersContainer({
   // This happens when we mount the container.
   useEffect(() => {
     if (!projectsLoaded) {
-      // doInitialization();
       return;
     }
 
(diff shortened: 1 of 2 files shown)


mrduncan (approved):

static/app/components/organizations/pageFilters/container.tsx
```diff
@@ -185,8 +186,14 @@ function PageFiltersContainer({
   }, [location.query]);
 
   // Wait for global selection to be ready before rendering children
+  // TODO: Not waiting for projects to be readly but initalizing the correct page filters
```
malwilley, line 189:
```suggestion
  // TODO: Not waiting for projects to be ready but initializing the correct page filters
```

malwilley (approved):
IMO this is a _very slightly_ worse UX for most orgs since you see a flash of a spinner very briefly, but it's a lot better for orgs where the projects take a long time to load.

PR #72563, opened 2024-06-11
fix(perf): Navigating to Spans Tab propagates incompatible filters

0Calories:
If you navigate to the spans tab from the transaction summary, it propagates any of the filters you had in the transaction summary. However, the filters used on transactions are very likely to be incompatible with the filter on the spans tab.

Take this example flow: 

Click into a transaction from the performance landing page with an HTTP method. It will automatically propagate the searchbar an `http.method` filter:
image

When you click on the Spans tab from here, it will pass on the same filter:
image

Unfortunately, you can't filter spans with the same `http.method` tag. This PR fixes this by checking all filters in the URL and confirming whether they are compatible with spans, and then sanitizing it.

Note: As a byproduct, the searchbar also does not allow you to input any incompatible tags in the search bar, but I think this is a nice thing to have:

image


KevinL10 (approved):
👍 only small UI thing is when the user searches in the Spans tab it'll replace the unsupported filters and the user will lose what they searched for in Overview. Seems fine to me though.

0Calories:
thanks @KevinL10 this helped me realize that I massively over-engineered the solution 😂 

I'm going to close this and open up a PR with a much simpler solution that should work better too.

PR #72570, opened 2024-06-11
fix(issues): occurence_consume, deduping status updates

vartec:
work related to INC-786

### DIFF
--- a/src/sentry/issues/occurrence_consumer.py
+++ b/src/sentry/issues/occurrence_consumer.py
@@ -418,7 +418,7 @@ def process_occurrence_group(items: list[Mapping[str, Any]]) -> None:
     if len(status_changes) > 1:
         items = [
             item for item in items if item.get("payload_type") != PayloadType.STATUS_CHANGE.value
-        ] + status_changes[-1]
+        ] + status_changes[-1:]
 
     for item in items:
         cache_key = f"occurrence_consumer.process_occurrence_group.{item['id']}"
--- a/src/sentry/issues/occurrence_consumer.py
+++ b/src/sentry/issues/occurrence_consumer.py
@@ -428,7 +428,7 @@ def process_occurrence_group(items: list[Mapping[str, Any]]) -> None:
             item for item in items if item.get("payload_type") == PayloadType.STATUS_CHANGE.value
         ]
 
-        if len(status_changes) > 1:
+        if status_changes:
             items = [
                 item
                 for item in items
--- a/src/sentry/issues/occurrence_consumer.py
+++ b/src/sentry/issues/occurrence_consumer.py
@@ -374,6 +374,7 @@ def _process_batch(worker: ThreadPoolExecutor, message: Message[ValuesBatch[Kafk
     By batching we're able to process occurrences in parallel while guaranteeing
     that no occurrences are processed out of order per group.
     """
+
     batch = message.payload
 
     occcurrence_mapping: Mapping[str, list[Mapping[str, Any]]] = defaultdict(list)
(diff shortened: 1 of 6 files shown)


wedamija (commented):
Shouldn't the idea here to be to look at a whole batch of status updates for an occurrence, and remove all but the last once from that batch?

So basically in here: https://github.com/getsentry/sentry/blob/58376fef7ba62923543dfb282470d86b29c26580/src/sentry/issues/occurrence_consumer.py#L413-L417

Just go through items and drop all extra status messages, then continue to just process in order. We could even go further and split them all into `occurrences` and `status_updates`, process all the occurrences first, then just process `if status_updates: _process_message(status_update[-1])`

vartec:
@wedamija OK, here's WIP of what you've suggested.

src/sentry/issues/occurrence_consumer.py
```diff
@@ -392,21 +393,40 @@ def _process_batch(worker: ThreadPoolExecutor, message: Message[ValuesBatch[Kafk
...
             worker.submit(process_occurrence_group, group) for group in occcurrence_mapping.values()
         ]
-        wait(futures)
+        status_change_futures = [
+            worker.submit(process_occurrence_group, group[-1])
+            for group in status_change_mapping.values()
+        ]
+
+        wait(occurrence_futures)
+        wait(status_change_futures)
```
wedamija, line 429: Oh sorry, what I meant was to do this within `process_occurrence_group`. If we split them all up here, we have no guarantee that `status_changes` related to a specific occurrence will be processed after.

I was just thinking that within an occurrence group, maybe it makes sense to split them up? I'm not sure if there's ever a risk of a status update somehow ending up on the topic before an occurrence, but maybe that just doesn't make sense

vartec: Hmm... OK, let me move that logic...

src/sentry/issues/occurrence_consumer.py
```diff
@@ -414,6 +416,16 @@ def process_occurrence_group(items: list[Mapping[str, Any]]) -> None:
...
     """
+
+    status_changes = [
+        item for item in items if item.get("payload_type") == PayloadType.STATUS_CHANGE.value
+    ]
+
+    if len(status_changes) > 1:
+        items = [
+            item for item in items if item.get("payload_type") != PayloadType.STATUS_CHANGE.value
+        ] + status_changes[-1]
```
vartec, line 427: processing the last status change _after_ all the occurrences in the batch;

alternatively we can keep them in order and remove all status changes but the last one.

wedamija, line 427: lgtm, but `+ status_changes[-1:]`

vartec: good catch, TY

wedamija (commented):
Could we also log some metrics about how many we're dropping here? Might be useful to help monitor this incident

vartec:
> Could we also log some metrics about how many we're dropping here? Might be useful to help monitor this incident
@wedamija yeah, I still have to add metrics, an option to roll this out, and some tests.

src/sentry/issues/occurrence_consumer.py
```diff
@@ -414,6 +416,29 @@ def process_occurrence_group(items: list[Mapping[str, Any]]) -> None:
...
+        return
+
+    project = Project.objects.get_from_cache(id=items[0]["project_id"])
+    organization = Organization.objects.get_from_cache(id=project.organization_id)
+    if features.has("organizations:occurence-consumer-prune-status-changes", organization):
+        status_changes = [
+            item for item in items if item.get("payload_type") == PayloadType.STATUS_CHANGE.value
+        ]
+
+        if len(status_changes) > 1:
```
wedamija, line 430: I think we can just remove the `if`. `status_changes[-1:]` will work even if there are 0 elements in the list. Metric could be `amount=max(0, len(status_changes) - 1)`. Although I'm not sure if a value of 0 is going to do anything other than no-op...

vartec: I'm leaving this as is, because although `status_changes[-1:]` would work regardless, when theres one or zero status changes, we can avoid recreating `items` list.

wedamija: Right, but don't we still want to put the 1 item at the end of the list?

vartec: Fair point, I'll update

wedamija (approved):

PR #72574, opened 2024-06-11
feat(uptime): Add an issue platform `GroupType` for uptime failures

wedamija:
This adds a new issue platform type for uptime failures. We may need to bump the creation quota before we fully release

### DIFF
--- a/src/sentry/issues/grouptype.py
+++ b/src/sentry/issues/grouptype.py
@@ -583,10 +583,10 @@ class FeedbackGroup(GroupType):
 
 
 @dataclass(frozen=True)
-class UptimeCheckFailure(GroupType):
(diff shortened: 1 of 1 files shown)


evanpurkhiser (approved):

src/sentry/issues/grouptype.py
```diff
@@ -581,6 +582,18 @@ class FeedbackGroup(GroupType):
     notification_config = NotificationConfig(context=[])
 
 
+@dataclass(frozen=True)
+class UptimeCheckFailure(GroupType):
+    type_id = 7001
+    slug = "uptime"
```
JoshFerge, line 588: nit: `uptime_failure` instead of `uptime`? same w/ description

PR #72576, opened 2024-06-11
[feat] FK subscriptions to Incident model

nhsiehgit:
Currently Alert notifications assume a 1:1 relation between AlertRule's and Incidents.
When an AlertRule is triggered, then we create an incident - and that incident won't be resolved until the resolved threshold is triggered.

Notification's are simple because we do not need to reference which snuba subscription was triggered - since there would only ever be a single subscription per alert rule per incident.

...

with Activated Alert Rules - we now have the possibility of multiple subscriptions per AlertRule.
Notification's become a bit more complicated since they still only reference the Incident - but now the Incident itself doesn't have the relevant data since linking simply to the AlertRule does not tell the end user which subscription itself was triggered.

FK'ing the subscription to the Incident will now give us more clarity into which subscription triggered the AlertRule

TODO:
- UI/UX updates to enable multiple Incident's per AlertRule 🤔


src/sentry/incidents/models/incident.py
```diff
@@ -60,28 +60,39 @@ class Meta:
...
+    SUB_CACHE_KEY = "incidents:active:%s:%s:%s"
 
     def fetch_for_organization(self, organization, projects):
         return self.filter(organization=organization, projects__in=projects).distinct()
 
     @classmethod
-    def _build_active_incident_cache_key(cls, alert_rule_id, project_id):
+    def _build_active_incident_cache_key(cls, alert_rule_id, project_id, subscription_id):
+        if subscription_id:
+            return cls.SUB_CACHE_KEY % (alert_rule_id, project_id, subscription_id)
```
nhsiehgit, line 71: When running a migration to add subscriptions to Incidents - how will this be affected?

we'll no longer find anything in the cache, and will build a new cache key with the subscription.
will there be any issue with the zombie cached items?
will there be any issue with querying for the incident once more?

src/sentry/migrations/0728_add_subscription_fk_to_incident.py
```diff
@@ -0,0 +1,85 @@
...
+        if not alert_rule:
+            logger.info(
+                "Incident found with no alert_rule. skipping",
+                extra={"incident_id": incident.id},
+            )
+            continue
+
+        snuba_query = alert_rule.snuba_query
+        subscriptions = list(snuba_query.subscriptions.all())
+        if len(subscriptions) == 0 or len(subscriptions) > 1:
```
nhsiehgit, line 34: Currently the only alert rules with >1 subscription would be the activated alert rules 🤔 
these subscriptions should be temporary and will expire soon.

should we backfill the incidents to just the first subscription?
This would could cause multiple `Incident`'s to be tied to the same `Subscription`...

OR

We just allow subscriptions to be  nullable on the Incident model and any _new_ Incident's will be created with a subscription

src/sentry/incidents/models/incident.py
markstory: > will there be any issue with the zombie cached items?

There shouldn't be. You're not overridding the default TTL so keys will naturally expire in an hour. Memcache will reclaim the space used by old keys automatically.

src/sentry/incidents/models/incident.py
```diff
@@ -97,28 +108,33 @@ def get_active_incident(self, alert_rule, project):
...
-            )
+            subscription = instance.subscription
+            if subscription:
+                key = cls._build_active_incident_cache_key(
+                    instance.alert_rule_id, project.id, subscription.id
+                )
+            else:
+                key = cls._build_active_incident_cache_key(instance.alert_rule_id, project.id, None)
+            cache.delete(key)
+            assert cache.get(key) is None
```
markstory, line 121: You could get burned here by concurrent requests.

src/sentry/incidents/models/incident.py
```diff
@@ -97,28 +108,33 @@ def get_active_incident(self, alert_rule, project):
...
-                    instance.incident.alert_rule_id, instance.project_id
-                )
+        else:
+            key = cls._build_active_incident_cache_key(
+                instance.incident.alert_rule_id, project_id, None
             )
-            is None
-        )
+        cache.delete(key)
+        assert cache.get(key) is None
```
markstory, line 137: I would be careful with this as it could fail from concurrent requests that re-create the cache key.

src/sentry/migrations/0728_add_subscription_fk_to_incident.py
```diff
@@ -0,0 +1,87 @@
...
+    # of a table.
+    # Here are some things that make sense to mark as post deployment:
+    # - Large data migrations. Typically we want these to be run manually so that they can be
+    #   monitored and not block the deploy for a long period of time while they run.
+    # - Adding indexes to large tables. Since this can take a long time, we'd generally prefer to
+    #   run this outside deployments so that we don't block them. Note that while adding an index
+    #   is a schema change, it's completely safe to run the operation after the code has deployed.
+    # Once deployed, run these manually via: https://develop.sentry.dev/database-migrations/#migration-deployment
+
+    is_post_deployment = False
```
markstory, line 65: We have ~27M rows in the affected tables. The backfill should be a post-deploy migration, which means you'd need to separate it from the schema change.

nhsiehgit: Aha that makes sense.
thanks for the catch!

I separated out the column add to a [separate PR here](https://github.com/getsentry/sentry/pull/72666)

src/sentry/incidents/models/incident.py
wedamija: > will there be any issue with querying for the incident once more?
Querying for the incident will be fine too, it just means slightly higher db load until we populate the cache.

src/sentry/incidents/models/incident.py
wedamija: I'm not totally sure why this get is here, it looks like it's meant to validate that the cache clear worked, but not sure why it's in code vs a test

src/sentry/incidents/models/incident.py
```diff
@@ -97,28 +108,33 @@ def get_active_incident(self, alert_rule, project):
...
-                cache.get(cls._build_active_incident_cache_key(instance.alert_rule_id, project.id))
-                is None
-            )
+            subscription = instance.subscription
+            if subscription:
+                key = cls._build_active_incident_cache_key(
+                    instance.alert_rule_id, project.id, subscription.id
+                )
+            else:
+                key = cls._build_active_incident_cache_key(instance.alert_rule_id, project.id, None)
```
wedamija, line 119: Nit:
`key = cls._build_active_incident_cache_key(
                    instance.alert_rule_id, project.id, subscription.id if subscription else None
                )

src/sentry/incidents/models/incident.py
```diff
@@ -97,28 +108,33 @@ def get_active_incident(self, alert_rule, project):
...
             )
-        )
-        assert (
-            cache.get(
-                cls._build_active_incident_cache_key(
-                    instance.incident.alert_rule_id, instance.project_id
-                )
+        else:
+            key = cls._build_active_incident_cache_key(
+                instance.incident.alert_rule_id, project_id, None
```
wedamija, line 134: Same nit about removing this if/else as above

src/sentry/incidents/subscription_processor.py
```diff
@@ -109,18 +109,33 @@ def __init__(self, subscription: QuerySubscription) -> None:
...
+            incident = Incident.objects.get_active_incident(
+                alert_rule=self.alert_rule,
+                project=self.subscription.project,
+                subscription=self.subscription,
             )
+            if not incident:
+                # TODO: make subscription required
+                incident = Incident.objects.get_active_incident(
+                    alert_rule=self.alert_rule, project=self.subscription.project
+                )
```
wedamija, line 125: No need for this conditional, we always have a subscription here since we require it to get the project

src/sentry/incidents/subscription_processor.py
```diff
@@ -633,21 +652,27 @@ def trigger_alert_threshold(
...
             # Now create (or update if it already exists) the incident trigger so that
             # we have a record of this trigger firing for this incident
-            incident_trigger = self.incident_triggers.get(trigger.id)
+            # NOTE: `incident_triggers` is derived from `self.active_incident`
+            incident_trigger = self.incident_trigger_map.get(trigger.id)
             if incident_trigger:
                 incident_trigger.status = TriggerStatus.ACTIVE.value
+                if not incident_trigger.subscription_id:
+                    # NOTE: until Incidents are per subscription, this will overwrite the subscription_id
+                    incident_trigger.subscription_id = self.subscription.id
```
wedamija, line 665: Why are we setting the subscription id on the trigger? I thought this was at the incident level?

Is this line covered by a test?

src/sentry/incidents/subscription_processor.py
```diff
@@ -716,11 +742,13 @@ def handle_trigger_actions(
             if incident_trigger.status == TriggerStatus.ACTIVE.value
             else AlertRuleTriggerActionMethod.RESOLVE.value
         )
-        try:
+
+        # NOTE: all incident_triggers are derived from self.active_incident
+        incident = self.active_incident
```
wedamija, line 747: Are you sure there's not some other reason we do this? Is there's no active incident at the moment does it cause a problem?

src/sentry/incidents/subscription_processor.py
```diff
@@ -716,11 +742,13 @@ def handle_trigger_actions(
...
+        if not incident:
+            # NOTE: resolve incident would have cleared the active incident cache
+            # So fetch the incident if it has
             incident = Incident.objects.get(id=incident_trigger.incident_id)
```
wedamija, line 751: Why'd we remove the `try/except` here that returns? This will cause an error if we don't find the incident

src/sentry/migrations/0729_add_subscription_fk_to_incident.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # use RangeQuerySetWrapper to avoid loading all Incidents into memory
+    for incident in RangeQuerySetWrapper(Incident.objects.all()):
+        alert_rule = incident.alert_rule
+        if not alert_rule:
+            logger.info(
+                "Incident found with no alert_rule. skipping",
+                extra={"incident_id": incident.id},
+            )
+            continue
```
wedamija, line 28: This isn't possible, this column is not null

src/sentry/migrations/0729_add_subscription_fk_to_incident.py
```diff
@@ -0,0 +1,76 @@
...
+                "Incident found with no or multiple subscriptions. skipping",
+                extra={
+                    "incident_id": incident.id,
+                    "snuba_query_id": snuba_query.id,
+                    "subscription_ids": [
+                        subscription.id for subscription in subscriptions if len(subscriptions) > 0
+                    ],
+                },
+            )
+            continue
```
wedamija, line 43: Do we not want to just always add the subscription to the incident? Or is the concern we don't know which one to add?

src/sentry/migrations/0729_add_subscription_fk_to_incident.py
```diff
@@ -0,0 +1,76 @@
...
+
+        subscription = subscriptions[0]
+        incident.subscription = subscription
+        incident.save()
```
wedamija, line 47: `incident.save(update_fields=["subscription"])`

src/sentry/migrations/0729_add_subscription_fk_to_incident.py
```diff
@@ -0,0 +1,76 @@
...
+logger = logging.getLogger(__name__)
+
+
+def _backfill_incident_subscription(
+    apps: StateApps, schema_editor: BaseDatabaseSchemaEditor
+) -> None:
+    Incident = apps.get_model("sentry", "Incident")
+
+    # use RangeQuerySetWrapper to avoid loading all Incidents into memory
+    for incident in RangeQuerySetWrapper(Incident.objects.all()):
```
wedamija, line 21: Maybe if the `incident` already has a `subscription` then skip?

src/sentry/incidents/models/incident.py
nhsiehgit: oops - thanks 😅

src/sentry/incidents/models/incident.py
nhsiehgit: hmmm yea
this was previously here though 🤔 

maybe i'll leave it for now?

src/sentry/incidents/subscription_processor.py
nhsiehgit: Oh - 
we haven't made subscription required as a fk to the `Incident` model yet though

this is just covering the case where there's an incident without a subscription fk'd

src/sentry/incidents/subscription_processor.py
nhsiehgit: oops - this was left over!

thanks for the catch 🙏

src/sentry/incidents/subscription_processor.py
wedamija: This is in the subscription processor, `self.subscription.project` is already used here

src/sentry/incidents/subscription_processor.py
nhsiehgit: Ah yea - so previously the `trigger_resolve_threshold` would clear teh `active_incident` before we got to this point
This is why we fetched the incident previously.

BUT - the incident fetched via the `incident_trigger` here _is just the active incident_ 
So - rather than making another query to grab the `Incident`, i figured we could just grab the available active incident.

src/sentry/incidents/subscription_processor.py
nhsiehgit: Looking at the dd charts - this metric was never actually fired 
makes sense to me - because in order to have an `IncidentTrigger` - we _must_ have an existing correlated `Incident`

src/sentry/migrations/0729_add_subscription_fk_to_incident.py
nhsiehgit: Yea - don't know which one to add

for the moment i think this should be fine since the only alert rules with multiple subscirptions sould be our test rules

src/sentry/incidents/subscription_processor.py
nhsiehgit: hmmm not sure if i'm understanding you here.

In this conditional i'm making a query to `get_active_incident` first _with_ the `self.subscription` filter - and if that doesn't exist, then i make the query _without_ the subscription filter.

The check isn't for whether subscription exists in the processor (yes, this always exists)
but the check is for whether or not we've backfilled the `subscription` to the `Incident` instance yet or not

wedamija: Lol sorry, I skimmed this and didn't read the conditional properly. I thought it was `if not self.subscription`. Ignore me

src/sentry/incidents/models/incident.py
```diff
@@ -60,28 +60,39 @@ class Meta:
...
-                    )
-                    .exclude(status=IncidentStatus.CLOSED.value)
-                    .order_by("-date_added")[0]
+                incident_query = Incident.objects.filter(
+                    type=IncidentType.ALERT_TRIGGERED.value,
+                    alert_rule=alert_rule,
+                    projects=project,
                 )
+                if subscription:
+                    incident_query = incident_query.filter(subscription=subscription)
```
wedamija, line 92: If we don't have a subscription, should we filter to rows with a null subscription value?

src/sentry/incidents/models/incident.py
wedamija: I think we can leave it. It's not causing a problem today I think?

src/sentry/incidents/subscription_processor.py
wedamija: Yeah that's fair, we have a not null on the incident field on the trigger too

src/sentry/incidents/subscription_processor.py
wedamija: Could we explain this more explicitly in the comments? 

"We attempt to fetch the active incident here to save a query. When this is triggered by a resolve, there will be no incident so we fall back to fetching from the db"

wedamija: I'm also wondering if we explicitly want to reload it from the db here though...

wedamija (approved):

src/sentry/incidents/models/incident.py
nhsiehgit: Ah - that's a good call 🤔

src/sentry/incidents/subscription_processor.py
nhsiehgit: still not sure reloading it is necessary since it's explicitly being set and utilized in the previous steps (`trigger_alert_threshold`)

PR #72578, opened 2024-06-12
WIP: Create new U2F flow that doesn't use the session cookie

schew2381:
This PR will be split into multiple parts:
1. Frontend U2F components - https://github.com/getsentry/sentry/pull/72654
2. Backend U2F activation and validation behind an option - https://github.com/getsentry/sentry/pull/72659
3. There are then 4 backend paths that need to be modified for this change
        - [BaseAuthIndexEndpoint](https://github.com/getsentry/sentry/blob/8e33e6624985ce96f2bb22cec87461f7215ce2f0/src/sentry/api/endpoints/auth_index.py#L35), which used for activating superuser + staff
        - [SudoView](https://github.com/getsentry/sentry/blob/8e33e6624985ce96f2bb22cec87461f7215ce2f0/src/sentry/web/frontend/sudo.py#L14) which is used for when you forget your password
        - [TwoFactorAuthView](https://github.com/getsentry/sentry/blob/de45c9258f2a86fd46feb917ce246fc57e6360fe/src/sentry/web/frontend/twofactor.py#L32) which is used for getting into orgs with U2F - https://github.com/getsentry/sentry/pull/72750


src/sentry/api/endpoints/authenticator_index.py
schew2381, line 1: Just typing fixes on this file

PR #72582, opened 2024-06-12
feat(starfish): Add count_if for span metrics

philipphofmann:
We need count_if similar to avg_if to display the count of spans per release for mobile starfish.


narsaynorath (approved):
:+1: looks good. We could drop the `span.op` from the allowed if columns but not a big deal for now.

In the future, if we expose this dataset (and these functions as API) through other places like Dashboards we'll probably have to modify this to have a more generic footprint, same with `avg_if` too probably

Issue #30164, opened 2021-11-23
Slack alerts do not include error message


adamgsg:
### Environment

SaaS (https://sentry.io/)

### Version

SaaS

### Steps to Reproduce

1.  Create an issue alert that notifies in Slack
2. Trigger an exception with a custom error message

### Expected Result

The Slack notification includes the custom error message

### Actual Result

The Slack notification includes the exception text without the message

getsentry-release:
Routing to @getsentry/ecosystem for [triage](https://develop.sentry.dev/processing-tickets/#3-triage). ⏲️

manuzope:
Hi @adamgsg, what do you mean by custom error message and how is it being generated?

adamgsg:
If I export the issue as JSON, I can see that the field used for Slack Notifications is coming from either:

- .exception.values[0].value
- .metadata.value

But the field we want to see in slack is just the ".message" field.

adamgsg:
I found the place where this is happening:  https://github.com/getsentry/sentry/blob/d81985575e1b0a9eebdb65f240db02c6f52ec2ff/src/sentry/integrations/slack/message_builder/issues.py#L74

So this is definitely using the ".metadata.value" field for slack notifications.  How can we make it use the ".message" field isntead?

adamgsg:
If it helps, this is how we're sending our events via the go sdk:

```
event := sentry.NewEvent()

event.EventID = ...
event.Environment = ...
event.Message = ...
event.User = ...
event.Timestamp = ...
event.ServerName = ...
event.Release = ...
event.Level = ...
event.Logger = ...
event.Tags = ...
event.Platform = ...
event.Contexts  = ...
event.Exception  = ...
event.Fingerprint = ...

hint :=  ...

client.CaptureEvent(event, &hint, nil)
```

adamgsg:
Any idea, @manuzope ?    This may also be related to a similar issue I just reported in #30566

adamgsg:
bump

kamilogorek:
@adamgsg if an event contains an exception (those its an "error" event), the message will always be formed out of internal exception type and value. `message` field is used for exception-less events, so you'd need to skip it if you want provided message to be your "main" value.
However, the builder itself doesn't look for the message value as you already pointed out in the link above, so it'd need to be changed there first.

Dhrumil-Sentry:
[...]

Issue #26006, opened 2021-05-12
Include culprit of event in Slack message


5long:
## Summary

Include culprit of event in Slack message

## Motivation

It helps users identify the issue quicker.

The legacy Slack integration already does this but the modern integration does not. It'd be good to see this feature preserved in the modern integration.

## Additional Context

There is a post on [forum.sentry.io](https://forum.sentry.io/t/is-it-possible-to-we-have-the-issue-message-culprit-transacation-in-slack-integration/8919) asking for the same feature.

getsentry-release:
Routing to @getsentry/ecosystem for [triage](https://develop.sentry.dev/processing-tickets/#3-triage). ⏲️

manuzope:
Thanks for opening the issue @5long, going to add this to our backlog.

For anyone else looking at this, please :+1: this if interested and we will prioritize it higher.

Dhrumil-Sentry:
Hi @5long , Apologies for the delayed response, when you say culprit of the event are you referring to the event's message?

Dhrumil-Sentry:
This issue is duplicate of https://github.com/getsentry/sentry/issues/30164 , Please track this request in this ticket

PR #72587, opened 2024-06-12
feat(slack): Show culprit in notifications

olagjo:
Relates to/part of #30164 (which has absorbed #26006). 

This adds a context-like culprit-block directly under the title, which mirrors how the culprit is presented in the web UI.

### Legal Boilerplate

### DIFF
--- a/src/sentry/integrations/slack/message_builder/issues.py
+++ b/src/sentry/integrations/slack/message_builder/issues.py
@@ -521,6 +521,12 @@ class SlackIssuesMessageBuilder(BlockSlackMessageBuilder):
 
         return self.get_markdown_block(title_text)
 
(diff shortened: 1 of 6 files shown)


olagjo:
I see that this is red, but I believe that it is only because I am missing some permissions as an outsider, and an employee needs to allow some of the checks to run?

iamrajjoshi (commented):
@olagjo thanks for the PR! Could you add a screen shot on how the notification looks after these changes?

olagjo:
Hi! Of course! 
<pre align="center"></pre>
<pre align="center"></pre>

iamrajjoshi:
Thanks for your work here!, I am gonna toss this behind a feature flag and get this reviewed by the team 🙌🏽

olagjo:
Thank you! I will just close this PR then and follow-up discussion can happen on yours :)

PR #72590, opened 2024-06-12
ref: remove sentry.profiles.* from mypy list of ignored modules

viglia:


pyproject.toml
```diff
@@ -572,7 +572,6 @@ module = [
     "sentry.nodestore.django.models",
     "sentry.nodestore.filesystem.backend",
     "sentry.nodestore.models",
-    "sentry.profiles.*",
```
asottile-sentry, line 575: this is the _stronger_ list -- not the ignored list!

viglia: ah, my bad @asottile-sentry !

Thanks!

Issue #64088, opened 2024-01-29
With chained errors, which one counts as the main one?


lobsterkatie:
There have been a number of issues which have come up lately having to do with which one of a chained set of errors we use for the purposes of grouping, titling an issue, calculating `stack.module`, etc. In many cases, what we're doing is the opposite of what the user expects, i.e., we're picking the wrong error to use. Here are a few examples:

- About how we title an issue: https://github.com/getsentry/sentry/issues/64074
- About how we automatically group issues: https://github.com/getsentry/sentry/issues/59679
- About how we use custom fingerprints when grouping issues: https://github.com/getsentry/sentry/issues/63882

To solve this we need to:

- Decide for sure which error in a chain it makes sense to use. It's possible this may differ by platform, but within a platform, we should be consistent everywhere across the product.
- Find all the places where we handle chained errors and have to make that determination. In some instances, the choice will be based on the value of `main_exception_id`, which is sent by the SDK. There may be other places, too, though (if we're hardcoding to always look at the top error or the bottom error, for example), so we should make sure to find those as well.
- Where we're making the wrong choice (according to our agreement in step 1), update the code and/or get the SDK to send a different `main_exception_id` value.

lobsterkatie:
Just wanted to highlight part of @markushi's [comment](https://github.com/getsentry/sentry-java/pull/3184#issuecomment-1935749973) on https://github.com/getsentry/sentry-java/pull/3184, as I think it's an interesting idea to consider vis-à-vis titling. (The whole thread there is relevant and worth reading, but didn't want to lose track of this idea in particular.)

> > Are the error message and stacktrace of the wrapper error always the same and/or do they provide any informational value?
> 
> Whilst the wrapped exception itself and it's stacktrace don't provide any extra information, it indicates one crucial semantic information: An app crash is about to happen.
> 
> Thinking more about future solutions: I'm wondering if in situations where the wrapping exception contains no in-app frames, we simply could decide to just use the inner exception for titleing.

lobsterkatie:
More discussion of the above: https://github.com/getsentry/sentry/issues/62299#issuecomment-1971939190

armenzg:
[An example](https://github.com/getsentry/sentry/pull/72593) of picking the title from the `cause` error.

InterstellarStella:
Another case: https://sentry.zendesk.com/agent/tickets/162443

PR #72593, opened 2024-06-12
feat(issues): Use title from React caused errors

armenzg:
Starting with React 19, errors can include a `cause` error. Instead of using the generic `Error` title, we can use the title from the `cause` error, thus, making each issue unique.

This design shows what the Issue Stream looks like now vs after this change:
image

### DIFF
--- a/src/sentry/grouping/strategies/newstyle.py
+++ b/src/sentry/grouping/strategies/newstyle.py
@@ -1,11 +1,10 @@
 from __future__ import annotations
 
 import itertools
+import logging
 import re
 from collections.abc import Generator
 from typing import Any
-
-import sentry_sdk
 
 from sentry.eventstore.models import Event
 from sentry.grouping.component import GroupingComponent, calculate_tree_label
@@ -24,6 +23,8 @@ from sentry.interfaces.exception import Mechanism, SingleException
 from sentry.interfaces.stacktrace import Frame, Stacktrace
 from sentry.interfaces.threads import Threads
 from sentry.stacktraces.platform import get_behavior_family_for_platform
+
+logger = logging.getLogger(__name__)
 
 _ruby_erb_func = re.compile(r"__\d{4,}_\d{4,}$")
 _basename_re = re.compile(r"[/\\]")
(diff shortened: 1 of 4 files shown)


eps1lon:
Why is this specific to React hydration errors? `Error.prototype.cause` is standardized nowadays so it seems to me it shouldn't be special cased for React.

armenzg:
> Why is this specific to React hydration errors? `Error.prototype.cause` is standardized nowadays so it seems to me it shouldn't be special cased for React.

It would take more time to get the data to determine the extent of the impact of the change. I am currently investigating getting the data.

eps1lon:
Makes sense. React uses `cause` in other places as well e.g. for "There was an error during concurrent rendering but React was able to recover [...]".

src/sentry/grouping/strategies/newstyle.py
```diff
@@ -899,3 +906,29 @@ def threads_variant_processor(
...
     return remove_non_stacktrace_variants(variants)
+
+
+def react_hydration_error_with_cause(exceptions: list[SingleException]) -> int | None:
+    main_exception_id = None
+    # Starting with React 19, hydration errors can contain a cause error which is useful to display
+    if (
+        exceptions[0].type == "Error"
+        and exceptions[0].value
+        == "There was an error while hydrating but React was able to recover by instead client rendering from the nearest Suspense boundary."
```
vartec, line 917: is it guaranteed to always be just this message and nothing else?

armenzg: At the moment, that's what we have noticed.
We may be fine with dropping checking the `value` but I need data to determine what other platforms could be impacted.

I assume it would be fine but I just don't know.

armenzg: @vartec I have added the other test case based on the data @eps1lon has provided.

src/sentry/grouping/strategies/newstyle.py
```diff
@@ -899,3 +906,33 @@ def threads_variant_processor(
...
+    main_exception_id = None
+    # Starting with React 19, errors can also contain a cause error which
+    # is useful to display instead of the default message
+    if (
+        exceptions[0].type == "Error"
+        and exceptions[0].value
+        in [
+            "There was an error during concurrent rendering but React was able to recover by instead synchronously rendering the entire root.",
+            "There was an error while hydrating but React was able to recover by instead client rendering from the nearest Suspense boundary.",
+        ]
```
vartec, line 921: maybe pull these out into a constant?

vartec (approved):

lobsterkatie:
@eps1lon - It's not always clear whether we want to use `cause`. For example, in our codebase we attach generic API errors as the `cause` of various more specific errors, and the latter are what we want to use for titling and grouping.

For lots more on this, see https://github.com/getsentry/sentry/issues/64088 and the various issues it links to.

PR #72599, opened 2024-06-12
Enable negated transaction filtering for dynamic sampling custom rules

shellmayr:
- To date, not filtering for event.type:transactions causes a Bad Request 400 on /dynamic-sampling/custom-rules/, even if something valid like a negated other event type is entered. This PR hotfixes this by adding the logic to the extraction code.
- Contributes to https://github.com/getsentry/sentry/issues/72300


shellmayr:
Closing - fixing on front end only for now.

PR #72603, opened 2024-06-12
ref(similarity): Change backfill snuba query to batch on ratelimit

jangjodi:
Change backfill single group snuba query to chunk on snuba ratelimit instead of relying on batch size = snuba ratelimit
Now the single group snuba query will execute in batches of size `snuba ratelimit`, regardless of batch size


JoshFerge (approved):
nice

jangjodi:
Closed since I redid it with base branch of main: https://github.com/getsentry/sentry/pull/72630

PR #72605, opened 2024-06-12
fix(hybridcloud) Don't use RPC to fetch user.id

markstory:
Within post_process_group we trigger several signal handlers. Many of these handlers fetch an organization, and the owner of that organization. Fetching the user via RPC to only use its `id` is very wasteful. Refactoring this logic to reduce the number of SQL queries (organization reads) and not using RPC should help reduce RPC throughput and make post_process_group marginally faster.


dashed (approved):
👏

PR #72610, opened 2024-06-12
fix(perf): Use separate query param for span search in spans tab

0Calories:
Same fix as https://github.com/getsentry/sentry/pull/72563, but simpler. Prevents the search query from the transaction overview from propagating into the searchbar in the spans tab. This is done because the search params for transactions will not work for spans.


KevinL10 (approved):
👍

PR #72611, opened 2024-06-12
fix(sudo): Selectively log user email on superuser access

RyanSkonnord:
This can leak PII for users from orgs other than the staff org, per https://github.com/getsentry/team-core-product-foundations/issues/315


RyanSkonnord:
Having found no good way to bind this to an actual auth provider, I've kludged in a single-purpose conf value that should do the trick. It should be pretty easy to refactor out again if anyone wants to take another pass later.

sentaur-athena (approved):

PR #72614, opened 2024-06-12
feat(trace-explorer): add analytics

KevinL10:
Analytics for trace explorer, excluding the search_result analytics.

### DIFF
--- a/static/app/views/traces/content.tsx
+++ b/static/app/views/traces/content.tsx
@@ -24,6 +24,7 @@ import {t, tct} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
 import type {PageFilters} from 'sentry/types/core';
 import type {MRI} from 'sentry/types/metrics';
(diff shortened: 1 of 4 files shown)


static/app/views/traces/content.tsx
```diff
@@ -386,10 +412,12 @@ function SpanTable({
 }
 
 function SpanRow({
+  organization,
   span,
   trace,
   setHighlightedSliceName,
 }: {
+  organization: Organization;
```
narsaynorath, line 420: Any particular reason we didn't just put `useOrganization` here instead of passing it down?

narsaynorath (approved):

static/app/views/traces/content.tsx
KevinL10: This is rendered for every row, I moved it to the parent Table instead so we only do it once.

PR #72619, opened 2024-06-12
feat(issues): Add front end for resolve in upcoming release

roggenkemper:
this pr makes the front end changes to support resolve in upcoming release. backend changes in https://github.com/getsentry/sentry/pull/70990. the changes made include adding an option to the Resolve button on issue details (hidden behind a feature flag), adding a subtitle to the resolve  dropdown to clarify behavior, and updating the resolution box to support this form of resolution. the group activities item shouldn't need any updates as it mirrors the behavior as "resolve in next release"


malwilley (approved):

malwilley (commented):
Might be nice to add a test, not sure how well tested this component is

PR #72621, opened 2024-06-12
ref: remove unused CanonicalKeyView

asottile-sentry:
after removing the last usage this should be safe to clean up now \o/

still need to do CanonicalKeyDict after this


joshuarli (approved):
🔥🔥🔥

PR #72622, opened 2024-06-12
ref(issues): don't persist sort param when clicking from issue list to tabs

### LINKED ISSUE #72617: Issue list sort persists on issue replay tab
Ran into this issue trying to look at replays for our top sentry issues. the "sort" param looks to be persisted as "user," which isn't something supported on the replays tab I guess?

To Reproduce:
• Go to issues page
• Sort by users
• Click on an issue
• Click on replay tab -- Screenshot is shown here

[Slack Message](https://sentry.slack.com/archives/C04KZQBNQ2U/p1718211570540469?thread_ts=1718211570.540469&cid=C04KZQBNQ2U)

michellewzhang:
Persisting the `sort` param from the issue list to issue details sometimes causes the replay tab table to break, since some of the issue list sort params like `users` and `trends` don't apply to the replay table. This PR removes the `sort` param from the URL once we click into the replay tab, to prevent this.

Before:

https://github.com/getsentry/sentry/assets/56095982/1f1831ec-57e1-4216-b223-11f4f6863249

After:

https://github.com/getsentry/sentry/assets/56095982/17e3b427-1576-4b64-b30c-34400393ddc9

closes https://github.com/getsentry/sentry/issues/72617

Also made the change to remove the `sort` param from every tab on the issue details (see @aliu39's comment below).

### DIFF
--- a/static/app/views/issueDetails/header.tsx
+++ b/static/app/views/issueDetails/header.tsx
@@ -47,7 +47,6 @@ type Props = {
 interface GroupHeaderTabsProps extends Pick<Props, 'baseUrl' | 'group' | 'project'> {
   disabledTabs: Tab[];
   eventRoute: LocationDescriptor;
-  replayRoute: LocationDescriptor;
 }
 
 function GroupHeaderTabs({
@@ -56,14 +55,15 @@ function GroupHeaderTabs({
   eventRoute,
   group,
   project,
-  replayRoute,
 }: GroupHeaderTabsProps) {
   const organization = useOrganization();
+  const location = useLocation();
 
   const {getReplayCountForIssue} = useReplayCountForIssues({
     statsPeriod: '90d',
   });
   const replaysCount = getReplayCountForIssue(group.id, group.issueCategory);
+  const queryParams = omit(location.query, ['sort']);
 
   const projectFeatures = new Set(project ? project.features : []);
   const organizationFeatures = new Set(organization ? organization.features : []);
@@ -93,7 +93,7 @@ function GroupHeaderTabs({
         key={Tab.ACTIVITY}
         textValue={t('Activity')}
         disabled={disabledTabs.includes(Tab.ACTIVITY)}
-        to={`${baseUrl}activity/${location.search}`}
+        to={{pathname: `${baseUrl}activity/`, query: queryParams}}
       >
         {t('Activity')}
         <IconBadge>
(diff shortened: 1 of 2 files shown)


JoshFerge:
will that remove the sort when the user goes back to issues? maybe replays just shouldn't use the sort field there?

michellewzhang:
> will that remove the sort when the user goes back to issues? maybe replays just shouldn't use the sort field there?

@JoshFerge what do you mean by 'goes back to issues'? do you mean by clicking this backlink here in the header?
SCR-20240612-jvkh

JoshFerge:
> @JoshFerge what do you mean by 'goes back to issues'? do you mean by clicking this backlink here in the header?

ah i guess we don't generally preserve this across clicks. nvmind

static/app/views/issueDetails/header.tsx
```diff
@@ -209,6 +211,14 @@ function GroupHeader({
     };
   }, [location, baseUrl]);
 
+  const replayRoute = useMemo(() => {
+    const searchTermWithoutSort = omit(location.query, ['sort']);
```
aliu39, line 215: Looking at the diff, should it be location.search? What's the difference?

michellewzhang: `location.search` is a string but omit takes in an object

michellewzhang: (`search` is kinda like the stringified version of `query`)

aliu39 (approved):

aliu39:
Actually, I'm thinking we could omit the sort param for _all_ tabs to avoid similar problems in the future..

aliu39:
That change could be beefy though, since it affects every tab. Seems like right now this only affects replays, so I'm cool merging this for now

michellewzhang:
> That change could be beefy though, since it affects every tab. Seems like right now this only affects replays, so I'm cool merging this for now

i just made changes to every tab - let me get some issues team eyes on this, we can always revert back to just the changes to replay

static/app/views/issueDetails/header.tsx
```diff
@@ -57,11 +57,13 @@ function GroupHeaderTabs({
   project,
 }: GroupHeaderTabsProps) {
   const organization = useOrganization();
+  const location = useLocation();
 
   const {getReplayCountForIssue} = useReplayCountForIssues({
     statsPeriod: '90d',
   });
   const replaysCount = getReplayCountForIssue(group.id, group.issueCategory);
+  const queryParams = omit(location.query, ['sort']);
```
malwilley, line 66: I don't see any issues with omitting sort from the query params, but a comment here explaining why this is necessary would be great!

aliu39 (approved):
oh awesome, lgtm!

PR #72626, opened 2024-06-12
feat(crons): Differentiate processing errors from differing projects

davidenwang:
When looking at the cron processing errors we currently only group checkin-errors by type, but we should also differentiate by project. If there are errors from multiple different projects we will show a view that looks like this: (otherwise we will not show project information)

After:
Screenshot 2024-06-12 at 10 50 07 AM

Before:
image

### DIFF
--- a/static/app/views/monitors/components/processingErrors/monitorProcessingErrors.tsx
+++ b/static/app/views/monitors/components/processingErrors/monitorProcessingErrors.tsx
@@ -8,11 +8,13 @@ import {Button} from 'sentry/components/button';
 import {Chevron} from 'sentry/components/chevron';
 import {DateTime} from 'sentry/components/dateTime';
 import {Hovercard} from 'sentry/components/hovercard';
+import ProjectBadge from 'sentry/components/idBadge/projectBadge';
 import List from 'sentry/components/list';
 import ListItem from 'sentry/components/list/listItem';
 import StructuredEventData from 'sentry/components/structuredEventData';
 import {t, tct} from 'sentry/locale';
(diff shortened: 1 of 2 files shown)


static/app/views/monitors/components/processingErrors/monitorProcessingErrors.tsx
```diff
@@ -51,45 +65,70 @@ export default function MonitorProcessingErrors({
...
-          </List>
-        )}
-      </ErrorGroup>
-    );
-  });
+  const showingMultipleProjects = Object.keys(errorsByProjectByType).length > 1;
+  const [expanded, setExpanded] = useState('');
+  const accordionErrors = Object.entries(errorsByProjectByType).map(
+    ([projectId, errorsByType]) => {
+      const project = projects.find(({id}) => id === projectId);
```
davidenwang, line 72: this is definitely not ideal, and feeling a need for a ids->projects type of functionality within useProjects :(

static/app/views/monitors/components/processingErrors/monitorProcessingErrors.tsx
```diff
@@ -102,6 +141,10 @@ const ErrorsList = styled('div')`
   gap: ${space(1)};
 `;
 
+const ProjectGroupsList = styled(ErrorsList)`
+  gap: ${space(1.5)};
```
davidenwang, line 145: want additional spacing if we are showing groups of errors by project

evanpurkhiser (approved):

PR #72629, opened 2024-06-12
feat(spans): Support autocompleting project slugs

Zylphrex:
Project slugs are not in clickhouse so override the behaviour and implement autocompletion logic in sentry directly.


wmak (approved):
lgtm

PR #72630, opened 2024-06-12
ref(similarity): Change backfill snuba query to batch on ratelimit

jangjodi:
Change backfill snuba query to chunk on snuba ratelimit instead of relying on batch size = snuba ratelimit
Now the snuba query will execute in batches of size snuba ratelimit, regardless of batch size

### DIFF
--- a/src/sentry/tasks/backfill_seer_grouping_records.py
+++ b/src/sentry/tasks/backfill_seer_grouping_records.py
@@ -43,6 +43,7 @@ from sentry.utils.snuba import bulk_snuba_queries
 
 BACKFILL_NAME = "backfill_grouping_records"
 BULK_DELETE_METADATA_CHUNK_SIZE = 100
+SNUBA_QUERY_RATELIMIT = 4
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


JoshFerge (dismissed):
nice!

src/sentry/tasks/backfill_seer_grouping_records.py
```diff
@@ -305,47 +306,51 @@ def get_data_from_snuba(project, groups_to_backfill_with_no_embedding):
...
+                app_id=Referrer.GROUPING_RECORDS_BACKFILL_REFERRER.value,
+                query=query,
+                tenant_ids={
+                    "referrer": Referrer.GROUPING_RECORDS_BACKFILL_REFERRER.value,
+                    "cross_org_query": 1,
+                },
+            )
+            snuba_requests.append(request)
+
+        with metrics.timer(f"{BACKFILL_NAME}.bulk_snuba_queries", sample_rate=1.0):
```
JoshFerge, line 348: should this be indented? seems like this is outside the above loop?

JoshFerge: ah no i see, this is correct

JoshFerge (approved):

PR #72633, opened 2024-06-12
fix(meta): Skip tag keys of zero

evanh:
For metrics with no tags, there is a default tag key of 0 populated in the table. Ignore this value
when querying for tag keys for a given metric.


volokluev (approved):
Should we have a test for this?

shellmayr (approved):
LGTM

evanh:
> Should we have a test for this?

So there is already a test for this, I am just refactoring.

PR #72635, opened 2024-06-12
ref: fix some typing in hybrid cloud modules

asottile-sentry:
fixes most errors in these modules when BaseManager becomes typed

### DIFF
--- a/src/sentry/hybridcloud/models/webhookpayload.py
+++ b/src/sentry/hybridcloud/models/webhookpayload.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import datetime
-from typing import Any, Self
+from typing import Any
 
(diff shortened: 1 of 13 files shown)


src/sentry/hybridcloud/models/webhookpayload.py
```diff
@@ -79,7 +79,7 @@ def create_from_request(
         identifier: int | str,
         request: HttpRequest,
         integration_id: int | None = None,
-    ) -> Self:
+    ) -> WebhookPayload:
```
asottile-sentry, line 82: due to an annoying variance issue interacting with a microbug in django-stubs it can't inference `cls.objects` properly here as also being `BaseManager[Self]`

markstory (approved):

PR #72642, opened 2024-06-12
ref: delete unused auth provider send reminders endpoint

asottile-sentry:
last reference removed in b5d2efe29b57a9678e8d1fbfe0eef76f49ab3942

the type checker helpfully pointed out that the attributes accessed in this module were impossible \o/

```
src/sentry/api/endpoints/organization_auth_provider_send_reminders.py:40: error: "AuthProvider" has no attribute "key"  [attr-defined]
```


markstory (approved):
Nice find :clap:

PR #72643, opened 2024-06-12
feat(spans): Support autocompleting span statuses

Zylphrex:
Span statuses are stored as UInt8 in clickhouse so we need to adapt the condition accordingly.


wmak (approved):
lgtm

PR #72649, opened 2024-06-12
fix(insights): Fix webvitals sample query when user filters on more than 1 tag

edwardgou-sentry:
Fix webvitals sample query when user filters on more than 1 tag

### DIFF
--- a/static/app/views/performance/browser/webVitals/utils/queries/storedScoreQueries/useTransactionSamplesWebVitalsScoresQuery.tsx
+++ b/static/app/views/performance/browser/webVitals/utils/queries/storedScoreQueries/useTransactionSamplesWebVitalsScoresQuery.tsx
@@ -50,6 +50,14 @@ export const useTransactionSamplesWebVitalsScoresQuery = ({
     sortableFields: filteredSortableFields as unknown as string[],
   });
 
+  const mutableSearch = new MutableSearch([
+    'transaction.op:pageload',
+    'has:measurements.score.total',
+  ]).addStringFilter(`transaction:"${transaction}"`);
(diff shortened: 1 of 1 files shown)


static/app/views/performance/browser/webVitals/utils/queries/storedScoreQueries/useTransactionSamplesWebVitalsScoresQuery.tsx
```diff
@@ -50,6 +50,14 @@ export const useTransactionSamplesWebVitalsScoresQuery = ({
     sortableFields: filteredSortableFields as unknown as string[],
   });
 
+  const mutableSearch = new MutableSearch([
+    'transaction.op:pageload',
+    'has:measurements.score.total',
+  ]).addStringFilter(`transaction:"${transaction}"`);
```
narsaynorath, line 56: Why can't we just slap this into the original array argument to MutableSearch?

narsaynorath (approved):

static/app/views/performance/browser/webVitals/utils/queries/storedScoreQueries/useTransactionSamplesWebVitalsScoresQuery.tsx
edwardgou-sentry: When we add the transaction filter inside the `MutableSearch` array arg, the `transaction` filter value doesn't get escaped. Using `addStringFilter` escapes the value by default. We need to escape the value because sometimes the transaction name has an `*` character which becomes a wildcard search if not escaped.

PR #72654, opened 2024-06-12
feat(u2f): Frontend fixes for passing u2f state directly in response

schew2381:
We need our U2F components to be able to handle receiving the auth state directly in the response body, and send it back to either the `sudo` and `twofactor` views or `auth_index`/`staff_auth_index` endpoints.

Right now, these changes do nothing because the state is not being sent in the response.

Split from https://github.com/getsentry/sentry/pull/72578


static/app/components/u2f/u2finterface.tsx
```diff
@@ -116,11 +121,18 @@ class U2fInterface extends Component<Props, State> {
...
+              : this.props.stateData;
 
             if (this.state.responseElement) {
               // eslint-disable-next-line react/no-direct-mutation-state
               this.state.responseElement.value = u2fResponse;
             }
+            if (this.state.stateElement && stateData) {
+              // eslint-disable-next-line react/no-direct-mutation-state
+              this.state.stateElement.value = stateData;
+            }
```
schew2381, line 135: If we're using a view, we need to set `stateData` so we can access it in `response.POST`

static/app/components/u2f/u2finterface.tsx
```diff
@@ -116,11 +121,18 @@ class U2fInterface extends Component<Props, State> {
           () => {
             const u2fResponse = this.getU2FResponse(data);
             const challenge = JSON.stringify(this.props.challengeData);
+            const stateData = this.props.stateData
+              ? JSON.stringify(this.props.stateData)
+              : this.props.stateData;
```
schew2381, line 126: `this.props.stateData` is currently undefined, but when we later need to have a django view render an html template which renders this component, it will be set to an empty string

static/app/components/u2f/u2finterface.tsx
```diff
@@ -131,6 +143,7 @@ class U2fInterface extends Component<Props, State> {
               .onTap({
                 response: u2fResponse,
                 challenge,
+                ...(stateData ? {state: stateData} : {}),
```
schew2381, line 146: Only send stateData back using whatever the onTap request is if it's set

static/app/components/u2f/u2fContainer.tsx
```diff
@@ -20,7 +20,12 @@ function U2fContainer({className, authenticators, ...props}: Props) {
     <div className={className}>
       {authenticators.map(auth =>
         auth.id === 'u2f' && auth.challenge ? (
-          <U2fSign key={auth.id} {...props} challengeData={auth.challenge} />
+          <U2fSign
+            key={auth.id}
+            {...props}
```
nhsiehgit, line 25: 🤔 
why are we splatting props in here after the key?
should the prop splat be first? Or last?

static/app/components/u2f/u2finterface.tsx
```diff
@@ -5,12 +5,13 @@ import * as cbor from 'cbor-web';
...
 import {t, tct} from 'sentry/locale';
 import ConfigStore from 'sentry/stores/configStore';
-import type {ChallengeData} from 'sentry/types';
+import type {ChallengeData, StateData} from 'sentry/types';
 
 type TapParams = {
   challenge: string;
   response: string;
   isSuperuserModal?: boolean;
+  state?: string;
```
nhsiehgit, line 14: state is very generic.... also kind of a confusing attr to pass, especially for a react component
can we call this soemthing else?

maybe tapState at least?

nhsiehgit: Ah wait i see...

we're typing the arguments for the onTap helper 🤔 

would still prefer if we did not call this `state` for the onTap unless we're actually passing the full state object in. which it looks like we're not

nhsiehgit (changes requested):
Can we rename everything `state` to something less generic?

`state` is sort of a reserved word in react world 😬

static/app/components/u2f/u2finterface.tsx
schew2381: I can call it `authState` and rename it everywhere

static/app/components/u2f/u2fContainer.tsx
schew2381: I'm not entirely sure, I was just copying how it was done before 😅

saponifi3d: eh, i think this is fine for (for now since it existed perviously).

---

In general, this is good to avoid as it creates a "prop drilling" problem where props are passed through many layers. a "future" solution would be to create an auth context where we could add any of these options.

static/app/types/auth.tsx
```diff
@@ -100,6 +102,11 @@ export type ChallengeData = {
   webAuthnRegisterData: string;
 };
 
+export type AuthStateData = {
+  challenge: string;
+  user_verification: 'required' | 'preferred' | 'discouraged' | null;
```
saponifi3d, line 107: I'd recommend pulling the `required` etc into an enum, then making `user_verification` optional so you don't need the `| null`.

```ts
// is there a better name for this? AuthStateVerification?
export enum UserVerificationState = {
   REQUIERD: 'required',
   PREFERRED: 'preferred',
   DISCOURAGED: 'discouraged',
};
```

```suggestion
  user_verification?: UserVerificationState;
```
This will give a bit better type error information, and be a bit closer to the represented state (rather than null, we don't have it yet)

static/app/components/u2f/u2finterface.tsx
```diff
@@ -21,16 +22,19 @@ type Props = {
...
     superuserReason,
   }: TapParams) => Promise<void>;
   silentIfUnsupported: boolean;
+  authStateData?: AuthStateData;
   children?: React.ReactNode;
   style?: React.CSSProperties;
 };
 
 type State = {
+  authStateElement: HTMLInputElement | null;
```
saponifi3d, line 37:
```suggestion
  authStateElement?: HTMLInputElement;
```
Recommend changing the type as it matches with the intent better.

static/app/components/u2f/u2finterface.tsx
```diff
@@ -49,6 +53,7 @@ class U2fInterface extends Component<Props, State> {
     hasBeenTapped: false,
     deviceFailure: null,
     responseElement: null,
+    authStateElement: null,
```
saponifi3d, line 56:
```suggestion
    authStateElement,
```
i'd recommend against setting it as null by default, and instead utilize undefined. `null` is a better representation for when things are in a null state (vs a state yet to be defined)

static/app/components/u2f/u2finterface.tsx
```diff
@@ -116,11 +121,18 @@ class U2fInterface extends Component<Props, State> {
...
+              ? JSON.stringify(this.props.authStateData)
+              : this.props.authStateData;
 
             if (this.state.responseElement) {
               // eslint-disable-next-line react/no-direct-mutation-state
               this.state.responseElement.value = u2fResponse;
             }
+            if (this.state.authStateElement && authStateData) {
+              // eslint-disable-next-line react/no-direct-mutation-state
+              this.state.authStateElement.value = authStateData;
```
saponifi3d, line 134: i'd really recommend against doing this directly manipulating state like this (it can cause a lot of difficult to detect issues).

instead take advantage of the react render life cycle around state updates, using `setState` to trigger changes checking for specific component conditions.

static/app/components/u2f/u2finterface.tsx
schew2381: This line complains if I'm missing null because `ref` is `HTMLInputElement | null`.
```typescript
  bindAuthStateElement: React.RefCallback<HTMLInputElement> = ref =>
    this.setState({authStateElement: ref});
```
I think best bet is to leave the default null off, and change the type to be undefined or null

saponifi3d (approved):
:shipit:

static/app/components/u2f/u2fContainer.tsx
nhsiehgit: for cleanliness sake - mind deconstructing this manually just so we can be more explicit?
Looks like this should only include `onTap` and `displayMode`?

static/app/components/u2f/u2finterface.tsx
```diff
@@ -116,11 +120,19 @@ class U2fInterface extends Component<Props, State> {
           () => {
             const u2fResponse = this.getU2FResponse(data);
             const challenge = JSON.stringify(this.props.challengeData);
+            const authStateData = this.props.authStateData
```
nhsiehgit, line 123: 🤔 so if it exists, we're stringifying it, otherwise we're keeping it as is? Does this just mean we're keeping it `null`?

Can we change this to specifically be a string? and default it to `''` if it doesn't exist?

nhsiehgit (commented):
few more comments

static/app/components/u2f/u2finterface.tsx
```diff
@@ -116,11 +120,19 @@ class U2fInterface extends Component<Props, State> {
...
+              ? JSON.stringify(this.props.authStateData)
+              : this.props.authStateData;
 
             if (this.state.responseElement) {
               // eslint-disable-next-line react/no-direct-mutation-state
               this.state.responseElement.value = u2fResponse;
             }
+            if (this.state.authStateElement && authStateData) {
+              this.setState({
+                authStateElement: authStateData as unknown as HTMLInputElement,
```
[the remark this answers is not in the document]
schew2381: oh it doesn't let me cast it directly to `HTMLInputElement` b/c there's some error saying there's no overlap btwn the classes. I poked around and it looks like we do this in some other parts of the code.
`authStateData` is a JSON that's been stringified

static/app/components/u2f/u2finterface.tsx
schew2381: It will either get passed in as `''` or undefined. I can set it to be the empty string

static/app/components/u2f/u2finterface.tsx
```diff
@@ -131,6 +143,7 @@ class U2fInterface extends Component<Props, State> {
               .onTap({
                 response: u2fResponse,
                 challenge,
+                ...(authStateData ? {authState: authStateData} : {}),
```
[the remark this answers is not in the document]
schew2381: right now `onTap` is drilled down from whatever flow's component is calling this generic component, so I wanted to split out the changes btwn the specific flows vs this generic part

static/app/components/u2f/u2finterface.tsx
```diff
@@ -39,18 +42,19 @@ type State = {
...
-    hasBeenTapped: false,
     deviceFailure: null,
-    responseElement: null,
-    isSafari: false,
     failCount: 0,
+    formElement: null,
+    hasBeenTapped: false,
+    isSafari: false,
+    isSupported: null,
+    responseElement: null,
```
schew2381, line 57: Sorted these alphabetically

static/app/components/u2f/u2finterface.tsx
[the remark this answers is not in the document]
nhsiehgit: Oof... 😬 🤔 

well... it _does_ seem like this _shouldn't_ be an `HtmlInputElement` right?
`authStateData` is a jsonified string? is this being converted into an element somewhere??

rather than perpetuating this hack - I would prefer we `@ts-ignore` this particular error until we properly set the value?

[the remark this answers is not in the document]
schew2381: I'm not sure how this `HtmlInputElement` stuff works, but I think we need it so that our view which calls a django html template which renders this component can send that data back in the response body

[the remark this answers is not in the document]
schew2381: I can `@ts-ignore` it for now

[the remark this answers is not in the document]
nhsiehgit: Ohhh uggh...

So this is an element being injected into a template
who's results are now being written back to state? 🤔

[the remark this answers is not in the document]
nhsiehgit: Responded in slack

but it looks like here we were attempting to copy the previous ref element example
```
this.state.responseElement.value = u2fResponse;
```

but rather than setting the element value, we're attempting to replace the entire element.

This looks incorrect?

PR #72666, opened 2024-06-12
feat: Incident subscription fk

nhsiehgit:
Just the column add change from https://github.com/getsentry/sentry/pull/72576

Purpose:

Currently Alert notifications assume a 1:1 relation between AlertRule's and Incidents.
When an AlertRule is triggered, then we create an incident - and that incident won't be resolved until the resolved threshold is triggered.

Notification's are simple because we do not need to reference _which_ snuba subscription was triggered - since there would only ever be a single subscription per alert rule per incident.

...

with Activated Alert Rules - we now have the possibility of multiple subscriptions _per_ AlertRule.
Notification's become a bit more complicated since they still only reference the Incident - but now the Incident itself doesn't have the relevant data since linking simply to the AlertRule does not tell the end user _which_ subscription itself was triggered.

FK'ing the subscription to the Incident will now give us more clarity into which subscription triggered the AlertRule


saponifi3d (approved):
Changes LGTM, should we also include the spec `tests/sentry/migrations/test_0728_add_subscription_fk_to_incident.py` from your other PR to test the migration?

nhsiehgit:
No need for the tests since those are backfill migration specific!

src/sentry/incidents/models/incident.py
```diff
@@ -203,6 +203,9 @@ class Incident(Model):
     activation = FlexibleForeignKey(
         "sentry.AlertRuleActivations", on_delete=models.SET_NULL, null=True
     )
+    subscription = FlexibleForeignKey(
+        "sentry.QuerySubscription", null=True, on_delete=models.PROTECT
+    )
```
wedamija, line 208: Are we sure we want protect here? I feel like we can recreate the subscription when we update the rule, which will now be blocked if we protect here. Do we want to set null insted?

wedamija (commented):
I have a feeling that `CREATE INDEX CONCURRENTLY "sentry_incident_subscription_id_ba61b3bb" ON "sentry_incident" ("subscription_id");` is going to fail with a timeout, since this table has 27m rows

Issue #72609, opened 2024-06-12
Metrics selector duplicate metrics issues


matejminar:
![Image](https://github.com/getsentry/sentry/assets/9060071/05b28eb3-bc5e-4ee4-b099-97b0ddac5680)

PR #72671, opened 2024-06-13
fix(metrics): Duplicated custom metrics in dropdown

ArthurKnaus:
Filter duplicated metrics and metrics without projects from metrics meta response.
Meant as a temporary fix until it is solved on the BE.

Relates to https://github.com/getsentry/sentry/issues/72609


priscilawebdev (approved):
lgtm

matejminar (approved):

PR #72676, opened 2024-06-13
ref(onboarding): Simplify next steps logic

ArthurKnaus:
Allow `nextSteps` entries to be `null` to ease the process of making them conditional.


priscilawebdev (approved):
lgtm

PR #72691, opened 2024-06-13
ref: remove EventDict inheritance of CanonicalKeyDict

asottile-sentry:
CanonicalKeyDict is a noop, for now implement EventDict as a simple MutableMapping -- though I may be able to remove EventDict as well in the future

### DIFF
--- a/src/sentry/models/event.py
+++ b/src/sentry/models/event.py
@@ -1,11 +1,17 @@
+from __future__ import annotations
+
+from collections.abc import Iterator, Mapping, MutableMapping
+from typing import Self, TypeVar
+
 import orjson
 from sentry_relay.processing import StoreNormalizer
 
 from sentry.db.models import NodeData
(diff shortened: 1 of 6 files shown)


src/sentry/models/event.py
```diff
@@ -1,11 +1,17 @@
...
+from collections.abc import Iterator, Mapping, MutableMapping
+from typing import Self, TypeVar
+
 import orjson
 from sentry_relay.processing import StoreNormalizer
 
 from sentry.db.models import NodeData
-from sentry.utils.canonical import CanonicalKeyDict
 
+V = TypeVar("V")
```
anonrig, line 11: What does this corresponds to? Can you add a comment?

anonrig (approved):

src/sentry/models/event.py
asottile-sentry: I think every caller uses it as `Any` right now -- but it's the (conventional name for) value type used below

PR #72692, opened 2024-06-13
feat(admin): Impove user account merging

azaslavsky:
We now ensure that user account merges are much more complete. All exportable (and even some non-exportable!) data associated with the user-being-merged is now properly pointed at the newly combined `User` and `OrganizationMember` models, and awkward collisions (what if the users had invited one another to a certain team?) are handled gracefully.

### DIFF
--- a/src/sentry/backup/dependencies.py
+++ b/src/sentry/backup/dependencies.py
@@ -6,12 +6,13 @@ from enum import Enum, auto, unique
 from functools import lru_cache
 from typing import NamedTuple
 
-from django.db import models
-from django.db.models import UniqueConstraint
+from django.db import IntegrityError, models, router, transaction
+from django.db.models import Q, UniqueConstraint
 from django.db.models.fields.related import ForeignKey, OneToOneField
 
 from sentry.backup.helpers import EXCLUDED_APPS
 from sentry.backup.scopes import RelocationScope
+from sentry.db.postgres.transactions import enforce_constraints
 from sentry.silo.base import SiloMode
 from sentry.utils import json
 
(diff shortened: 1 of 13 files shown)


src/sentry/backup/dependencies.py
```diff
@@ -701,3 +701,28 @@ def get_exportable_sentry_models() -> set[type[models.base.Model]]:
...
+    user in question will be pointed at the new user instead.
+    """
+
+    from sentry.models.organization import Organization
+    from sentry.models.user import User
+
+    model_relations = dependencies()[get_model_name(model)]
+    user_refs = {k for k, v in model_relations.foreign_keys.items() if v.model == User}
+    org_refs = {
+        k if k.endswith("_id") else f"{k}_id"
```
hubertdeng123, line 720: Why do we need to check for `_id` in the name for org but not user?

src/sentry/backup/dependencies.py
```diff
@@ -701,3 +701,28 @@ def get_exportable_sentry_models() -> set[type[models.base.Model]]:
...
+        for k, v in model_relations.foreign_keys.items()
+        if v.model == Organization
+    }
+    for_this_org = Q(**{field_name: organization_id for field_name in org_refs})
```
hubertdeng123, line 724: Is there a more descriptive variable name/comment that can be added to describe what's going on here? A bit lost reading through here

src/sentry/backup/dependencies.py
azaslavsky: Because user fields are basically just ints already ("HybridCloudForeignKey"), whereas `FlexibleForeignKey` is more convoluted and makes a `foo` and `foo_id` field for each such relation. Since all orgs are `FlexibleForeignKey`s here (we are in the region silo for this function), we want to make sure we get the `int`.

hubertdeng123 (approved):

PR #72694, opened 2024-06-13
fix(slack): prevent duplicate unfurl in metric alert when using sdk client

cathteng:
Forgot to put unfurl_links=False and unfurl_media=False, I think we're sending a link with the metric alert and it's getting unfurled.


iamrajjoshi (approved):

src/sentry/integrations/slack/utils/notifications.py
```diff
@@ -125,6 +125,8 @@ def send_incident_alert_notification(
                 channel=str(channel),
                 thread_ts=thread_ts,
                 reply_broadcast=reply_broadcast,
+                unfurl_links=False,
+                unfurl_media=False,
```
cathteng, line 129: we're passing this in the old logic

PR #72695, opened 2024-06-13
feat(query-builder): Add undo stack with ctrl-z

malwilley:
Adds a basic undo stack so that ctrl/cmd+z will revert to the last state. Note that this only supports undoing for now, but can be modified to support redoing in the future.

https://github.com/getsentry/sentry/assets/10888943/5c1056b7-86c7-427f-933a-b581f313c33c

### DIFF
--- a/static/app/components/searchQueryBuilder/useQueryBuilderState.tsx
+++ b/static/app/components/searchQueryBuilder/useQueryBuilderState.tsx
@@ -20,7 +20,11 @@ type QueryBuilderState = {
 
 type ClearAction = {type: 'CLEAR'};
 
-type UpdateQueryAction = {query: string; type: 'UPDATE_QUERY'};
+type UpdateQueryAction = {
+  query: string;
+  type: 'UPDATE_QUERY';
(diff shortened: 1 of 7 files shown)


static/app/components/searchQueryBuilder/useQueryBuilderState.tsx
```diff
@@ -78,8 +82,8 @@ export type QueryBuilderActions =
   | DeleteLastMultiSelectFilterValueAction;
 
 function removeQueryToken(query: string, token: TokenResult<Token>): string {
-  return (
-    query.substring(0, token.location.start.offset) +
+  return removeExcessWhitespaceFromParts(
```
malwilley, line 85: Some modifications in this file to trimming whitespace, because I was seeing that the stack would have duplicates of things like `is:unresolved browser.name:Firefox` and `is:unresolved  browser.name:Firefox` (multiple spaces in between). So I modified this to be more aggressive with enforcing a single space between tokens

roggenkemper (approved):

PR #72718, opened 2024-06-13
ref: fix typing for event_id (it is str)

asottile-sentry:
split out from my Job is TypedDict branch

### DIFF
--- a/src/sentry/issues/related/trace_connected.py
+++ b/src/sentry/issues/related/trace_connected.py
@@ -28,7 +28,7 @@ def trace_connected_analysis(
     if event_id:
         # If we are passing an specific event_id, we need to get the project_id
         assert project_id is not None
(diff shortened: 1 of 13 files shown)


src/sentry/issues/related/trace_connected.py
```diff
@@ -28,7 +28,7 @@ def trace_connected_analysis(
     if event_id:
         # If we are passing an specific event_id, we need to get the project_id
         assert project_id is not None
-        event = eventstore.backend.get_event_by_id(project_id, event_id, group_id=group.id)
+        event = eventstore.backend.get_event_by_id(int(project_id), event_id, group_id=group.id)
```
asottile-sentry, line 31: @armenzg is going to handle the api endpoint doing the validation itself instead of here so this `int(...)` call should go away eventually (and the `extra_args` would probably become a `TypedDict` or normal parameters)

armenzg (approved):

PR #72722, opened 2024-06-13
feat(related_issues): Validate parameters received in endpoint

armenzg:
Before, it would allow passing `project_id` as a `str` which is not the case.


src/sentry/api/endpoints/issues/related_issues.py
```diff
@@ -10,6 +11,12 @@
 from sentry.types.ratelimit import RateLimit, RateLimitCategory
 
 
+class RequestSerializer(serializers.Serializer):
+    type = serializers.CharField(required=True)
```
asottile-sentry, line 15: I think you can use `ChoiceField` here as well to constrain it to the particular algorithm names

src/sentry/api/endpoints/issues/related_issues.py
```diff
@@ -33,6 +40,8 @@ def get(self, request: Request, group: Group) -> Response:
         :pparam Request request: the request object
         :pparam Group group: the group object
         """
+        serializer = RequestSerializer(data=request.data)
+        serializer.is_valid()
```
asottile-sentry, line 44: I think you need to use this to produce an error somehow? and then retrieve the data from this and pass it along

src/sentry/api/endpoints/issues/related_issues.py
```diff
@@ -10,6 +11,12 @@
 from sentry.types.ratelimit import RateLimit, RateLimitCategory
 
 
+class RequestSerializer(serializers.Serializer):  # type: ignore[type-arg]
```
armenzg, line 14: Otherwise I get this error:
> error: Missing type parameters for generic type "Serializer"  [type-arg]

src/sentry/api/endpoints/issues/related_issues.py
armenzg: thank you!

src/sentry/api/endpoints/issues/related_issues.py
```diff
@@ -33,6 +40,9 @@ def get(self, request: Request, group: Group) -> Response:
         :pparam Request request: the request object
         :pparam Group group: the group object
         """
+        serializer = RequestSerializer(data=request.query_params)
+        if not serializer.is_valid():
```
armenzg, line 44: @asottile-sentry thank you! I was doing it all wrong. I had thought it raises an error but it doesn't.

src/sentry/api/endpoints/issues/related_issues.py
asottile-sentry, line 14: idk why my comment didn't post here in the first revision I'll try and rewrite it again!

Serializers come in two forms, one where they are registered for a particular type and another where they are just used to accept view parameters -- this is the latter one and so the generic type can just be `[None]`

src/sentry/api/endpoints/issues/related_issues.py
```diff
@@ -33,6 +40,9 @@ def get(self, request: Request, group: Group) -> Response:
...
+            return Response(serializer.errors, status=400)
         # The type of related issues to retrieve. Can be either `same_root_cause` or `trace_connected`.
         related_type = request.query_params["type"]
         extra_args = {
```
asottile-sentry, line 48: the extra_args and related_type should then be pulled out of the serializer and passed along

you'll want to change the type of the dictionary received by the implementations -- probably to take named arguments instead of a dict

you'll also be able to clean up the `int(...)` call I introduced as well

src/sentry/api/endpoints/issues/related_issues.py
```diff
@@ -1,15 +1,23 @@
...
 from rest_framework.request import Request
 from rest_framework.response import Response
 
 from sentry.api.api_owners import ApiOwner
 from sentry.api.api_publish_status import ApiPublishStatus
 from sentry.api.base import region_silo_endpoint
 from sentry.api.bases.group import GroupEndpoint
-from sentry.issues.related import RELATED_ISSUES_ALGORITHMS
+from sentry.issues.related.same_root_cause import same_root_cause_analysis
+from sentry.issues.related.trace_connected import trace_connected_analysis
```
armenzg, line 10: I've removed the indirection. I don't see us adding more than these two cases.

src/sentry/api/endpoints/issues/related_issues.py
```diff
@@ -33,14 +41,19 @@ def get(self, request: Request, group: Group) -> Response:
...
-            data, meta = RELATED_ISSUES_ALGORITHMS[related_type](group, extra_args)
+            data, meta = (
+                same_root_cause_analysis(group)
+                if related_type == "same_root_cause"
+                else trace_connected_analysis(
+                    group, event_id=_data.get("event_id"), project_id=_data.get("project_id")
+                )
+            )
             return Response({"type": related_type, "data": data, "meta": meta})
         except AssertionError:
```
asottile-sentry, line 58: unrelated to this PR but `except AssertionError` is something that should really only be in tests -- assertion errors shouldn't really be used at runtime since they're meant to be "programmer assertions" and not business logic

asottile-sentry (approved):

PR #72723, opened 2024-06-13
chore(insights): Update documentation URLs

gggritso:
The URLs have changed! Redirects exist, but why rely on redirects?


AbhiPrasad (approved):
save the environment (by not paying for server redirects)

PR #72724, opened 2024-06-13
ref(perf): Use `sum(span.duration)` instead of `sum(span.self_time)` in Spans tab & Span Summary Page

0Calories:
Due to technical limitations, self_time will eventually not be available on spans. This PR refactors to use duration instead of self_time in the spans tab and span summary page

### DIFF
--- a/static/app/views/performance/transactionSummary/transactionSpans/spanMetricsTable.spec.tsx
+++ b/static/app/views/performance/transactionSummary/transactionSpans/spanMetricsTable.spec.tsx
@@ -27,7 +27,7 @@ describe('SuspectSpansTable', () => {
             'span.op': 'db',
             'span.description': 'SELECT thing FROM my_cool_db',
             'spm()': 4.448963396488444,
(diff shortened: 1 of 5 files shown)


static/app/views/performance/transactionSummary/transactionSpans/spanMetricsTable.spec.tsx
```diff
@@ -65,7 +65,7 @@ describe('SuspectSpansTable', () => {
     expect(descriptionCell).toHaveTextContent('SELECT thing FROM my_cool_db');
     expect(throughputCell).toHaveTextContent('4.45/s');
     expect(avgDurationCell).toHaveTextContent('30.90s');
-    expect(timeSpentCell).toHaveTextContent('2.04wk');
+    expect(timeSpentCell).toHaveTextContent('2.04wk'); //
```
edwardgou-sentry, line 68:
```suggestion
    expect(timeSpentCell).toHaveTextContent('2.04wk');
```

edwardgou-sentry (approved):

PR #72726, opened 2024-06-13
fix(issues): Prevent error when stream ids don't match group store

scttcper:
There seems to be a race between updating the group ids in the issues stream and the groups being removed from the group store. Ideally we would not use the group store and just pass the group from the stream. For now, avoid throwing errors.

fixes JAVASCRIPT-2TK8

replays show the user resolving or another action that removes the issues from the current stream

### DIFF
--- a/static/app/components/stream/group.tsx
+++ b/static/app/components/stream/group.tsx
@@ -34,17 +34,16 @@ import GroupStore from 'sentry/stores/groupStore';
 import SelectedGroupStore from 'sentry/stores/selectedGroupStore';
 import {useLegacyStore} from 'sentry/stores/useLegacyStore';
 import {space} from 'sentry/styles/space';
+import type {TimeseriesValue} from 'sentry/types/core';
 import type {
   Group,
   GroupReprocessing,
   InboxDetails,
-  NewQuery,
-  Organization,
   PriorityLevel,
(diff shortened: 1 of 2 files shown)


static/app/components/stream/group.tsx
```diff
@@ -245,9 +239,31 @@ function BaseGroupRow({
...
+    if (!group) {
+      return [];
+    }
+
+    return group.filtered ? group.stats?.[statsPeriod] : [];
+  }, [group, statsPeriod]);
+
+  if (!group) {
+    return null;
+  }
```
scttcper, line 265: mostly re-arranging things so i can return early

static/app/components/stream/group.tsx
```diff
@@ -245,9 +239,31 @@ function BaseGroupRow({
...
+    if (!group) {
+      return [];
+    }
+
+    return group.filtered
+      ? group.filtered.stats?.[statsPeriod]
+      : group.stats?.[statsPeriod];
+  }, [group, statsPeriod]);
+
+  const groupSecondaryStats = useMemo<ReadonlyArray<TimeseriesValue>>(() => {
```
scttcper, line 255: hooks needed to be moved up a bit

malwilley (approved):

PR #72730, opened 2024-06-13
fix(reports): Use duplicate report detection to halt delivery

RyanSkonnord:
Upgrade the Redis-based weekly report duplicate delivery detection so that it halts the duplicate delivery instead of merely logging an error.

Refactor the context manager to a small class, because a context manager can't actually interrupt the body of the `with`-block.

### DIFF
--- a/src/sentry/tasks/summaries/weekly_reports.py
+++ b/src/sentry/tasks/summaries/weekly_reports.py
@@ -375,7 +375,7 @@ class _DuplicateDeliveryCheck:
     def record_delivery(self) -> bool:
         """Record in Redis that the delivery was completed successfully."""
         if self.count is None:
-            raise ValueError("This object has not had `check_before` called yet")
+            raise ValueError("This object has not had `check_for_duplicate_delivery` called yet")
         cluster = self._get_redis_cluster()
         count_after = cluster.incr(self._redis_name)
         cluster.expire(self._redis_name, timedelta(days=1))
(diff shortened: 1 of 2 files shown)


src/sentry/tasks/summaries/weekly_reports.py
```diff
@@ -331,41 +333,66 @@ def send_email(self, template_ctx: Mapping[str, Any], user_id: int) -> None:
...
+        """Check whether this delivery has been recorded in Redis already."""
+        if self.count is not None:
+            raise ValueError("This object has already checked a delivery")
+        cluster = self._get_redis_cluster()
+        self.count = int(cluster.get(self._redis_name) or 0)
+
+        is_duplicate_detected = self.count > 0
+        if is_duplicate_detected:
+            logger.error(
+                "weekly_report.delivery_record.duplicate_detected", extra=self._get_log_extras()
```
ceorourke, line 371: Could we ever have more than a count of 1 here? It might be interesting to log the count but if it's always 1 here that's not very interesting data.

ceorourke (approved):
Looks great! Looking forward to seeing how well this works tomorrow

src/sentry/tasks/summaries/weekly_reports.py
RyanSkonnord: Good idea -- we definitely could have a count of more than 1. I'll add it.

PR #72740, opened 2024-06-13
fix(slack): actually send the metric alert in a thread

cathteng:
The tests check that we have the correct relationship between the parent and child message, but we still need to update the values of `thread_ts` and `reply_broadcast` to pass to `client.chat_postMessage`. These are pulled out separately for typing purposes.


ykamo001 (approved):

ykamo001:
Which PR had removed this? I'm on master and it looks the same?

cathteng:
@ykamo001 i never added this, i initialized these two variables to the non-threads values but forgot to set them when they should be going in a thread

ykamo001:
> Which PR had removed this? I'm on master and it looks the same?

My bad, this is adding the values, was looking at another piece of code in my IDE

PR #72743, opened 2024-06-13
feat(perf): Make suggested fields on Spans tab metrics only

0Calories:
The Spans tab queries on span metrics, but the searchbar's autosuggest includes indexed fields. This PR adds a new hook that suggests only metrics-compatible fields for span search

image


KevinL10 (approved):
👍

PR #72750, opened 2024-06-13
feat(u2f): Have 2FA view use u2f state in request + response

schew2381:
Waiting for frontend PR to merge before testing locally

### DIFF
--- a/src/sentry/web/frontend/twofactor.py
+++ b/src/sentry/web/frontend/twofactor.py
@@ -207,22 +207,30 @@ class TwoFactorAuthView(BaseView):
 
         #  If a challenge and response exists, validate
         if challenge:
-            response = request.POST.get("response")
-            if response:
+            if response := request.POST.get("response"):
                 response = json.loads(response)
-                if interface.validate_response(request, challenge, response):
+
+                # TODO(schew2381): Eventually check state in if statement
+                if state := request.POST.get("state"):
+                    state = json.loads(state)
+                else:
+                    # state may be an empty string, so we want to explicitly set
+                    # it to None before passing to validate_response
+                    state = None
+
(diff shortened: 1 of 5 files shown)


src/sentry/web/frontend/twofactor.py
```diff
@@ -207,22 +207,30 @@ def handle(self, request: Request) -> HttpResponse:
...
 
         return render_to_response(
-            ["sentry/twofactor_%s.html" % interface.interface_id, "sentry/twofactor.html"],
-            {
+            template=["sentry/twofactor_%s.html" % interface.interface_id, "sentry/twofactor.html"],
+            context={
                 "form": form,
                 "interface": interface,
                 "other_interfaces": self.get_other_interfaces(interface, interfaces),
                 "activation": activation,
```
schew2381, line 231: We don't need to do any extra work for sending the state back in the GET response because it's nested inside this activation object which is of type [ActivationChallengeResult](https://github.com/getsentry/sentry/pull/72659/files#diff-020539479c15bbc08dfb2b59d410993b51a102ac3d53630131ff5ff70ee3f478R43)

tests/sentry/web/frontend/test_twofactor.py
```diff
@@ -1,107 +1,139 @@
...
-        assert msg.to == [user.email]
+        assert msg.to == [self.user.email]
         assert msg.subject == "[Sentry]Suspicious Activity Detected"
         url = absolute_uri(reverse("sentry-account-settings-security"))
         assert url in msg.body
         assert "IP address: 127.0.0.1" in msg.body
+
+
+@control_silo_test
+class U2FTest(TwoFactorBaseTestCase):
```
schew2381, line 94: There were no existing U2F tests for this flow beforehand 😓

PR #72752, opened 2024-06-13
ref(slack): use sdk client to get channel id

iamrajjoshi:
Adds feature-flagged usage of Slack SDK client to check channel id by schedule-sending a message and then immediately deleting. Also adds tests for the code area.

### DIFF
--- a/src/sentry/integrations/slack/utils/__init__.py
+++ b/src/sentry/integrations/slack/utils/__init__.py
@@ -1,6 +1,5 @@
 __all__ = (
     "get_channel_id",
-    "get_channel_id_with_timeout",
     "get_slack_data_by_user",
     "get_users",
     "is_valid_role",
@@ -20,12 +19,7 @@ import logging
 logger = logging.getLogger("sentry.integrations.slack")
 
 from .auth import is_valid_role, set_signing_secret
-from .channel import (
-    get_channel_id,
-    get_channel_id_with_timeout,
-    strip_channel_name,
-    validate_channel_id,
-)
+from .channel import get_channel_id, strip_channel_name, validate_channel_id
 from .notifications import send_incident_alert_notification, send_slack_response
 from .rule_status import RedisRuleStatus
 from .users import get_slack_data_by_user, get_users
--- a/src/sentry/integrations/slack/utils/channel.py
+++ b/src/sentry/integrations/slack/utils/channel.py
@@ -3,8 +3,11 @@ from __future__ import annotations
 import time
 
 from django.core.exceptions import ValidationError
+from slack_sdk.errors import SlackApiError
 
+from sentry import features
 from sentry.integrations.slack.client import SlackClient
+from sentry.integrations.slack.sdk_client import SlackSdkClient
 from sentry.models.integrations.integration import Integration
 from sentry.models.organization import Organization
 from sentry.services.hybrid_cloud.integration import RpcIntegration
@@ -65,8 +68,12 @@ def get_channel_id(
     # This means some users are unable to create/update alert rules. To avoid this, we attempt
     # to find the channel id asynchronously if it takes longer than a certain amount of time,
     # which I have set as the SLACK_DEFAULT_TIMEOUT - arbitrarily - to 10 seconds.
+    if features.has("organizations:slack-sdk-get-channel-id", organization):
+        has_sdk_flag = True
+    else:
+        has_sdk_flag = False
 
-    return get_channel_id_with_timeout(integration, channel_name, timeout)
+    return get_channel_id_with_timeout(integration, channel_name, timeout, has_sdk_flag)
 
 
 def validate_channel_id(name: str, integration_id: int | None, input_channel_id: str) -> None:
@@ -103,6 +110,7 @@ def get_channel_id_with_timeout(
     integration: Integration | RpcIntegration,
     name: str | None,
     timeout: int,
+    has_sdk_flag: bool,
 ) -> tuple[str, str | None, bool]:
     """
     Fetches the internal slack id of a channel using scheduled message.
@@ -122,14 +130,16 @@ def get_channel_id_with_timeout(
     }
 
     time_to_quit = time.time() + timeout
-
-    client = SlackClient(integration_id=integration.id)
+    if has_sdk_flag:
+        client = SlackSdkClient(integration_id=integration.id)
+    else:
+        client = SlackClient(integration_id=integration.id)
     id_data: tuple[str, str | None, bool] | None = None
     found_duplicate = False
     prefix = ""
     channel_id = None
     try:  # Check for channel
-        channel_id = check_for_channel(client, name)
+        channel_id = check_for_channel(client, name, has_sdk_flag)
         prefix = "#"
     except ApiError as e:
         if str(e) != "channel_not_found":
@@ -188,26 +198,64 @@ def get_channel_id_with_timeout(
     return prefix, channel_id, False
 
 
-def check_for_channel(client: SlackClient, name: str) -> str | None:
-    msg_response = client.post(
-        "/chat.scheduleMessage",
-        data={
-            "channel": name,
-            "text": "Sentry is verifying your channel is accessible for sending you alert rule notifications",
-            "post_at": int(time.time() + 500),
-        },
-    )
+def check_for_channel(
+    client: SlackClient | SlackSdkClient, name: str, has_sdk_flag: bool
+) -> str | None:
+    if has_sdk_flag:
+        try:
+            msg_response = client.chat_scheduleMessage(
+                channel=name,
+                text="Sentry is verifying your channel is accessible for sending you alert rule notifications",
+                post_at=int(time.time() + 500),
+            )
+        except SlackApiError as e:
+            logger_params = {
+                "error": str(e),
+                "integration_id": client.integration_id,
+                "channel": name,
+                "post_at": int(time.time() + 500),
+            }
+            logger.exception("slack.chat_scheduleMessage.error", extra=logger_params)
+            raise
 
-    if "channel" in msg_response:
-        client.post(
-            "/chat.deleteScheduledMessage",
-            params=dict(
-                {
-                    "channel": msg_response["channel"],
+        if "channel" in msg_response:
+            try:
+                client.chat_deleteScheduledMessage(
+                    channel=msg_response["channel"],
+                    scheduled_message_id=msg_response["scheduled_message_id"],
+                )
+                return msg_response["channel"]
+            except SlackApiError as e:
+                logger_params = {
+                    "error": str(e),
+                    "integration_id": client.integration_id,
+                    "channel": name,
+                    "channel_id": msg_response["channel"],
                     "scheduled_message_id": msg_response["scheduled_message_id"],
                 }
-            ),
+                logger.exception("slack.chat_deleteScheduledMessage.error", extra=logger_params)
+                raise
+
+    else:
+        msg_response = client.post(
+            "/chat.scheduleMessage",
+            data={
+                "channel": name,
+                "text": "Sentry is verifying your channel is accessible for sending you alert rule notifications",
+                "post_at": int(time.time() + 500),
+            },
         )
 
-        return msg_response["channel"]
+        if "channel" in msg_response:
+            client.post(
+                "/chat.deleteScheduledMessage",
+                params=dict(
+                    {
+                        "channel": msg_response["channel"],
+                        "scheduled_message_id": msg_response["scheduled_message_id"],
+                    }
+                ),
+            )
+
+            return msg_response["channel"]
     return None
--- a/src/sentry/integrations/slack/utils/channel.py
+++ b/src/sentry/integrations/slack/utils/channel.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
 import time
+from collections.abc import Iterable, Mapping
+from dataclasses import dataclass
+from typing import Any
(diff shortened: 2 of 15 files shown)


src/sentry/integrations/slack/utils/channel.py
```diff
@@ -65,8 +68,12 @@ def get_channel_id(
     # This means some users are unable to create/update alert rules. To avoid this, we attempt
     # to find the channel id asynchronously if it takes longer than a certain amount of time,
     # which I have set as the SLACK_DEFAULT_TIMEOUT - arbitrarily - to 10 seconds.
+    if features.has("organizations:slack-sdk-get-channel-id", organization):
+        has_sdk_flag = True
+    else:
+        has_sdk_flag = False
```
cathteng, line 74:
```suggestion
    has_sdk_flags = features.has("organizations:slack-sdk-get-channel-id", organization):
```

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -65,8 +68,12 @@ def get_channel_id(
...
 
-    return get_channel_id_with_timeout(integration, channel_name, timeout)
+    return get_channel_id_with_timeout(integration, channel_name, timeout, has_sdk_flag)
```
ykamo001, line 76: could we simplify this with:
```
return get_channel_id_with_timeout(integration, channel_name, timeout, features.has("organizations:slack-sdk-get-channel-id", organization))
```

or
```
has_sdk_flag = features.has("organizations:slack-sdk-get-channel-id", organization)
get_channel_id_with_timeout(integration, channel_name, timeout, has_sdk_flag)
```

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -122,14 +130,16 @@ def get_channel_id_with_timeout(
...
+        client = SlackSdkClient(integration_id=integration.id)
+    else:
+        client = SlackClient(integration_id=integration.id)
     id_data: tuple[str, str | None, bool] | None = None
     found_duplicate = False
     prefix = ""
     channel_id = None
     try:  # Check for channel
-        channel_id = check_for_channel(client, name)
+        channel_id = check_for_channel(client, name, has_sdk_flag)
```
cathteng, line 142: do we need to be passing the sdk flag if we can infer which method to use (`client.post` or `client.chat_postMessage`) based on the type of the client?

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -188,26 +198,64 @@ def get_channel_id_with_timeout(
...
+            logger_params = {
+                "error": str(e),
+                "integration_id": client.integration_id,
+                "channel": name,
+                "post_at": int(time.time() + 500),
+            }
+            logger.exception("slack.chat_scheduleMessage.error", extra=logger_params)
+            raise
+
+        if "channel" in msg_response:
```
ykamo001, line 221: what should we do if channel is not in msg_response? This would simply return `None` right?

ykamo001: could we just early return here?
```
if "channel" not in msg_response:
    return None
```

ykamo001 (approved):

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -188,26 +198,64 @@ def get_channel_id_with_timeout(
...
+            )
+        except SlackApiError as e:
+            logger_params = {
+                "error": str(e),
+                "integration_id": client.integration_id,
+                "channel": name,
+                "post_at": int(time.time() + 500),
+            }
+            logger.exception("slack.chat_scheduleMessage.error", extra=logger_params)
+            raise
```
cathteng, line 219: in the parent function that calls this function, we except ApiError and raise only `if str(e) != "channel_not_found"`... i'm not sure why we continue on with the logic though because we don't even use the `channel_id` to paginate through the users?

iamrajjoshi: talked about this on slack, need to refactor.

iamrajjoshi:
had to update some other tests because of the feature flag check 😢

src/sentry/integrations/slack/utils/channel.py
cathteng: add a TODO comment so we don't forget?

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -188,26 +198,68 @@ def get_channel_id_with_timeout(
...
+
+        if "channel" not in msg_response:
+            return None
+
+        try:
+            client.chat_deleteScheduledMessage(
+                channel=msg_response["channel"],
+                scheduled_message_id=msg_response["scheduled_message_id"],
```
cathteng, line 229:
```suggestion
                channel=msg_response.get("channel"),
                scheduled_message_id=msg_response.get("scheduled_message_id"),
```
the `SlackResponse` class has a built in method called `get` that fetches the information from `response.data`

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -188,26 +198,68 @@ def get_channel_id_with_timeout(
...
+            )
+            return msg_response["channel"]
```
cathteng, line 231:
```suggestion
            return msg_response.get("channel")
```

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -188,26 +198,68 @@ def get_channel_id_with_timeout(
...
+                "error": str(e),
+                "integration_id": client.integration_id,
+                "channel": name,
+                "post_at": int(time.time() + 500),
+            }
+            logger.exception("slack.chat_scheduleMessage.error", extra=logger_params)
+            raise
+
+        if "channel" not in msg_response:
+            return None
```
cathteng, line 224:
```suggestion
        if not msg_response.get("channel"):
            return None
```

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -188,26 +198,68 @@ def get_channel_id_with_timeout(
...
+                scheduled_message_id=msg_response["scheduled_message_id"],
+            )
+            return msg_response["channel"]
+        except SlackApiError as e:
+            logger_params = {
+                "error": str(e),
+                "integration_id": client.integration_id,
+                "channel": name,
+                "channel_id": msg_response["channel"],
+                "scheduled_message_id": msg_response["scheduled_message_id"],
```
cathteng, line 238:
```suggestion
                "channel_id": msg_response.get("channel"),
                "scheduled_message_id": msg_response.get("scheduled_message_id"),
```

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -188,26 +198,68 @@ def get_channel_id_with_timeout(
...
+            }
+            logger.exception("slack.chat_deleteScheduledMessage.error", extra=logger_params)
+            raise
```
cathteng, line 241: should we be catching this exception in `get_channel_id_with_timeout`? we continue with logic to try to fetch a user if the error here is `channel_not_found`

tests/sentry/integrations/slack/test_utils.py
```diff
@@ -73,6 +99,27 @@ def test_valid_private_channel_selected(self):
...
+
+        self.run_valid_test("#my-private-channel", CHANNEL_PREFIX, "m-p-c", False)
+
+        mock_metrics.incr.assert_called_with(
+            SLACK_DATADOG_METRIC,
+            sample_rate=1.0,
+            tags={"ok": True, "status": 200},
+        )
+
     def test_valid_member_selected(self):
```
cathteng, line 123: can we test this with the sdk? i don't think the current logic will correctly search for a user if a channel does not exist with the id

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -99,11 +116,145 @@ def validate_channel_id(name: str, integration_id: int | None, input_channel_id:
...
+def get_users_sdk(
+    client: SlackSdkClient, integration: Integration | RpcIntegration
+) -> Iterable[Mapping[str, Any]]:
+    payload = {
+        "exclude_archived": False,
+        "exclude_members": True,
+        "types": "public_channel,private_channel",
+    }
+    try:
+        users_list = client.users_list(limit=SLACK_GET_CHANNEL_ID_PAGE_SIZE, kwargs=payload)
```
iamrajjoshi, line 128: will replace this after cathy's refactor [pr](https://github.com/getsentry/sentry/pull/73085) gets merged

src/sentry/integrations/slack/utils/__init__.py
```diff
@@ -1,6 +1,5 @@
 __all__ = (
```
iamrajjoshi, line 1: I am planning to add a task to clean these files at the end of the refactoring

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -17,11 +23,20 @@
...
 from . import logger
 
+SLACK_GET_CHANNEL_ID_PAGE_SIZE = 200
 SLACK_DEFAULT_TIMEOUT = 10
 MEMBER_PREFIX = "@"
 CHANNEL_PREFIX = "#"
 strip_channel_chars = "".join([MEMBER_PREFIX, CHANNEL_PREFIX])
 
+
+@dataclass
```
cathteng, line 33: should this be frozen?

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -99,11 +116,145 @@ def validate_channel_id(name: str, integration_id: int | None, input_channel_id:
...
+        profile = user.get("profile")
+        if profile and profile.get("display_name") == name:
+            if id_data.channel_id is not None:
+                raise DuplicateDisplayNameError(name)
+            else:
+                id_data = SlackChannelIdData("@", user["id"], False)
+
+        # TODO: This is a problem if we don't go through all the users and eventually run in to someone with duplicate display name
+        if time.time() > time_to_quit:
+            break
```
cathteng, line 200: (i'm wondering what the timeout is for? i don't see this being kicked off as an async task. maybe some API calls have a timeout)

should we return `SlackChannelIdData` with `timed_out=True` for the timeout here?

src/sentry/integrations/slack/utils/channel.py
iamrajjoshi: It can't be because i modify the instance in the function.

src/sentry/integrations/slack/utils/channel.py
iamrajjoshi: There is a timeout in case the member list is extremely long, in which case  the function will return early if it can't find a user that matches in time. I should return early here as you said.

cathteng: i guess i'm wondering what is necessitating the timeout in terms of the call stack, it's probably some upstream caller then

RyanSkonnord (changes requested):

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -99,11 +116,145 @@ def validate_channel_id(name: str, integration_id: int | None, input_channel_id:
...
+    logger_params = {
+        "integration_id": integration.id,
+        "name": name,
+    }
+    try:  # Check for channel
+        id_data.channel_id = check_for_channel(client, name)
+        id_data.prefix = "#"
+    except SlackApiError:
+        logger.exception("rule.slack.channel_check_error", extra=logger_params)
+        return id_data
```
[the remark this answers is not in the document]
iamrajjoshi: i have looked through callers, most of them  complain loudly when the `channel_id` is None. its a good call to document this case.

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -99,11 +116,145 @@ def validate_channel_id(name: str, integration_id: int | None, input_channel_id:
...
+    except SlackApiError as e:
+        logger_params = {
+            "error": str(e),
+            "integration_id": client.integration_id,
+            "channel": name,
+            "post_at": int(time.time() + 500),
+        }
+        logger.exception("slack.chat_scheduleMessage.error", extra=logger_params)
+        if "channel_not_found" in str(e):
+            return None
```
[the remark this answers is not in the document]
iamrajjoshi: Essentially this channel checks for either a channel or a particular user. if we get a channel not found, we attempt to find a user with the particular name, which is why we have to end up checking for this error

RyanSkonnord (approved):

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -103,16 +124,141 @@ def get_channel_id_with_timeout(
...
+
+def check_user_with_timeout(
+    integration: Integration, name: str, time_to_quit: int
+) -> SlackChannelIdData:
+    """
+    If the channel is not found, we check if the name is a user.
+    """
+
+    _channel_id = None
+    _prefix = ""
```
cathteng, line 170: just wondering why you chose to have underscores here? these aren't global variables?

cathteng (approved):

src/sentry/integrations/slack/utils/channel.py
iamrajjoshi: i did that cause intellisense kept suggesting the variables defined in the previous function were in scope here.

cathteng: weird

PR #72754, opened 2024-06-14
If a GroupEvent has no occurance - don't follow through to the else

nhsiehgit:
we already default the title

If a GroupEvent has no occurance - then there's no point in fetching the latest event of the group again


wedamija (approved):
lgtm, might be worth letting @ceorourke have a look first just to be sure

saponifi3d (approved):
👍 lgtm, traced it back down to where the GroupEvent's title and it all seems legit.

PR #72763, opened 2024-06-14
ref(seer grouping): Send error type to Seer

lobsterkatie:
This adds `exception_type` to the data we send to Seer, which may be a string or `None` (if we don't find it at `exception["values"][-1]["type"]`). I chose to make it a required field (even though we might not have data) so that everywhere we use our functions which call Seer, we at least try to retrieve the data. Doing this shouldn't pollute the Seer database (where the column is nullable), because if we send Seer a `None`, it will turn into `null` when it's JSONified, and that will in turn become SQL `NULL` when we add it the database.

As I was updating tests, I discovered that a test I was expecting to have to update didn't in fact even exist, so this PR also adds (and updates) that test (`GetSeerSimilarIssuesTest.test_sends_expected_data_to_seer`).


armenzg (approved):

lobsterkatie:
Ignore my above linking to the fingerprint API issue - I initially pasted the wrong link there, and even fixing it didn't make this backlink disappear.

JoshFerge (approved):
awesome stuff. i'll merge and deploy my seer migration now

PR #72767, opened 2024-06-14
feat(similarity): Add delete record by hash task

jangjodi:
Add the task that calls the seer delete record by hash endpoint
Given a list of hashes, batch by size 20


JoshFerge (commented):
do we need to register the celery task? and could we make batch size an option? i have a feeling we can probably go pretty high with it, into the thousands

jangjodi:
> do we need to register the celery task? and could we make batch size an option? i have a feeling we can probably go pretty high with it, into the thousands

The task is registered [here](https://github.com/getsentry/sentry/blob/583ac6c1ee09c11609ecdf269f4330fe834077b8/src/sentry/conf/server.py#L855). and yes, I've made it an option, good idea - thanks!

JoshFerge (approved):

src/sentry/tasks/delete_seer_grouping_records_by_hash.py
```diff
@@ -0,0 +1,31 @@
...
+) -> None:
+    """
+    Task to delete seer grouping records by hash list.
+    Calls the seer delete by hash endpoint with batches of hashes of size `BATCH_SIZE`.
+    """
+    batch_size = options.get("embeddings-grouping.seer.delete-record-batch-size")
+    len_hashes = len(hashes)
+    end_index = min(last_deleted_index + batch_size, len_hashes)
+    delete_grouping_records_by_hash(project_id, hashes[last_deleted_index:end_index])
+    if end_index < len_hashes:
```
JoshFerge, line 30: any reason to call the task again instead of just calling the `delete_grouping_records_by_hash` function repetitively?

jangjodi: it's based on the logic we used for the backfill - I didn't want this task to timeout, so I'm respawning the task instead

JoshFerge: what is our current timeout set to on the celery task?

jangjodi: haha whoops just realized I didn't even add a timeout, but I was thinking  of using the same timeouts from the backfill `soft_time_limit=60 * 15`, `time_limit=60 * (15 + 5)`

JoshFerge: got it. i'd consider thinking about increasing the batch size to 100, and optionally doing some calls in a loop, so that way we have to spawn less celery tasks, as each one should run pretty quick (less than a second)

jangjodi: nice, I didn't realize they were that quick - thanks!

PR #72768, opened 2024-06-14
feat(metrics): add metrics extraction rule endpoint

shellmayr:


src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -0,0 +1,72 @@
...
+        except Exception as e:
+            return Response(status=400, data={"detail": str(e)})
+
+        return Response(status=200, data=updated_rules)
+
+    def delete(self, request: Request, project: Project) -> Response:
+        if not self.has_feature(project.organization, request):
+            return Response(status=404)
+
+        rules_update = request.data.get("metricsExtractionRules") or ""
```
obostjancic, line 60: nit: I'd expect the deserealization to be in the here in the endpoint and functions `delete_metrics_extraction_rules ` and `update_metrics_extraction_rules` to accept a sequence of rules instead of strings

src/sentry/sentry_metrics/extraction_rules.py
```diff
@@ -0,0 +1,114 @@
...
+class MetricsExtractionRule:
+    span_attribute: str
+    type: str
+    unit: str
+    tags: set[str]
+
+    @classmethod
+    def from_dict(cls, dictionary: Mapping[str, Any]) -> "MetricsExtractionRule":
+        return MetricsExtractionRule(
+            span_attribute=dictionary["span_attribute"],
```
obostjancic, line 25: nit: afaik we try to use camel case in the request/response objects
```suggestion
            span_attribute=dictionary["spanAttribute"],
```

src/sentry/sentry_metrics/extraction_rules.py
```diff
@@ -0,0 +1,114 @@
...
+            type=dictionary["type"],
+            unit=dictionary["unit"],
+            tags=set(dictionary.get("tags") or set()),
+        )
+
+    def to_dict(self) -> Mapping[str, Any]:
+        return self.__dict__
+
+    def generate_mri(self, use_case: str = "custom"):
```
obostjancic, line 34: nit: maybe validate the use case here using `use_case_utils.string_to_use_case_id`

obostjancic: nit: also since `generate_mri` is called in both update and delete, it might be easier to have it as a field

src/sentry/sentry_metrics/extraction_rules.py
```diff
@@ -0,0 +1,114 @@
...
+    def save_to_project(self, project: Project) -> None:
+        metrics_extraction_rules = [rule.to_dict() for rule in self.rules.values()]
+
+        json_payload = json.dumps(metrics_extraction_rules)
+        project.update_option(METRICS_EXTRACTION_RULES_OPTION_KEY, json_payload)
+        # once this is live, we need to invalidate the relay project config here.
+        return
+
+    def get_rules(self) -> Sequence[MetricsExtractionRule]:
+        return list(self.rules.values())
```
obostjancic, line 78:
```suggestion
        return self.rules.values()
```

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -0,0 +1,72 @@
...
+
+        try:
+            persisted_rules = delete_metrics_extraction_rules(project, rules_update)
+            updated_rules = serialize(
+                persisted_rules, request.user, MetricsExtractionRuleSerializer()
+            )
+        except Exception as e:
+            return Response(status=500, data={"detail": str(e)})
+
+        return Response(status=200, data=updated_rules)
```
obostjancic, line 72: according to [`get_success_response` ](https://github.com/getsentry/sentry/blob/8bf93a27544d21d29e3bfc9487542f7533254950/src/sentry/testutils/cases.py#L763)
```suggestion
        return Response(status=204, data=updated_rules)
```

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -0,0 +1,72 @@
...
+from sentry.models.project import Project
+from sentry.sentry_metrics.extraction_rules import (
+    delete_metrics_extraction_rules,
+    update_metrics_extraction_rules,
+)
+
+"""
+Open Questions
+1. Do we need to register the key metricsExtractionRules in the project config somehow?
+2. Do we need to prevent breaking something in the project config somehow?
```
obostjancic, line 20: ideally, we would have validation provided by sentry_relay, similar to https://github.com/getsentry/sentry/blob/8bf93a27544d21d29e3bfc9487542f7533254950/src/sentry/relay/config/metric_extraction.py#L784

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -0,0 +1,72 @@
...
+from sentry.api.serializers.models.metrics_extraction_rules import MetricsExtractionRuleSerializer
+from sentry.models.project import Project
+from sentry.sentry_metrics.extraction_rules import (
+    delete_metrics_extraction_rules,
+    update_metrics_extraction_rules,
+)
+
+"""
+Open Questions
+1. Do we need to register the key metricsExtractionRules in the project config somehow?
```
obostjancic, line 19: From what i can see, yes.

src/sentry/api/endpoints/project_metrics_extraction_rules.py
shellmayr: 204 is No Content - since we are returning the updated rules, I think 200 is the way to go

obostjancic: ah yes, but why do we return the rules in the first place? 😁

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -0,0 +1,138 @@
...
+            persisted_rules = create_metrics_extraction_rules(project, state_update)
+            updated_rules = serialize(
+                persisted_rules, request.user, MetricsExtractionRuleSerializer()
+            )
+        except Exception as e:
+            return Response(status=400, data={"detail": str(e)})
+
+        return Response(status=200, data=updated_rules)
+
+    def put(self, request: Request, project: Project) -> Response:
```
ArthurKnaus, line 98: @shellmayr As of how the rules are identified via the MRI, this only allows updating the tags, right?
Afaik, as soon as something else is updated a new rule will be created?

obostjancic (approved):

PR #72769, opened 2024-06-14
feat(similarity): Add delete record by hash call on group delete

jangjodi:
Add call to delete record by hash task on group deletion


src/sentry/api/endpoints/project_details.py
```diff
@@ -932,6 +933,10 @@ def delete(self, request: Request, project) -> Response:
         if updated:
             scheduled = RegionScheduledDeletion.schedule(project, days=0, actor=request.user)
 
+            # Tell seer to delete all the project's grouping records
+            if features.has("projects:similarity-embeddings-delete-by-hash", project):
+                delete_project_grouping_records(project.id)
```
jangjodi, line 938: Call the project record deletion endpoint

src/sentry/deletions/defaults/group.py
```diff
@@ -128,6 +131,17 @@ def delete_bulk(self, instance_list):
...
+        # Tell seer to delete grouping records with these group hashes
+        if group_ids:
+            group = Group.objects.get(id=group_ids[0])
+            project = group.project if group else None
+        if project and features.has("projects:similarity-embeddings-delete-by-hash", project):
+            # TODO (jangjodi): once we store seer grouping info in GroupHash, we should filter by that here
+            group_hashes = GroupHash.objects.filter(project_id=project.id, group__id__in=group_ids)
+            delete_seer_grouping_records_by_hash.apply_async(
+                args=[project.id, [group_hash.hash for group_hash in group_hashes], 0]
+            )
```
jangjodi, line 143: Call the delete by hash endpoint. This code is called by all group deletion, including projects - still figuring out if there's a way to skip it if it's from project deletion

src/sentry/deletions/defaults/group.py
```diff
@@ -128,6 +131,17 @@ def delete_bulk(self, instance_list):
 
         group_ids = [group.id for group in instance_list]
 
+        # Tell seer to delete grouping records with these group hashes
+        if group_ids:
+            group = Group.objects.get(id=group_ids[0])
+            project = group.project if group else None
+        if project and features.has("projects:similarity-embeddings-delete-by-hash", project):
```
jangjodi, line 138: I feel like there should be a better way to do this

src/sentry/api/endpoints/project_details.py
JoshFerge: does this spawn a celery task or call the deletion directly?

src/sentry/deletions/defaults/group.py
JoshFerge, line 143: wanna put this logic in its own function? `delete_embeddings_grouping_records`?

src/sentry/deletions/defaults/group.py
```diff
@@ -128,6 +131,9 @@ def delete_bulk(self, instance_list):
 
         group_ids = [group.id for group in instance_list]
 
+        # Tell seer to delete grouping records with these group hashes
+        call_delete_seer_grouping_records_by_hash(group_ids)
```
JoshFerge, line 135: should we add an option for this too?

jangjodi: I have the feature check in the function itself since we needed to extract the project and it looked a bit messy

JoshFerge: ah gotcha 👍🏼

JoshFerge (approved):

src/sentry/api/endpoints/project_details.py
jangjodi: I moved it into its own task just in case seer takes a long time

PR #72773, opened 2024-06-14
fix(perf): cannot read property of undefined area chart

DominikB2014:
Fixes JAVASCRIPT-2TK4

### DIFF
--- a/static/app/components/charts/areaChart.tsx
+++ b/static/app/components/charts/areaChart.tsx
@@ -25,7 +25,7 @@ export function transformToAreaSeries({
     AreaSeries({
       stack: stacked ? 'area' : undefined,
       name: seriesName,
-      data: data.map(({name, value}) => [name, value]),
+      data: data?.map(({name, value}) => [name, value]) || [],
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):

static/app/components/charts/areaChart.tsx
```diff
@@ -25,7 +25,7 @@ export function transformToAreaSeries({
     AreaSeries({
       stack: stacked ? 'area' : undefined,
       name: seriesName,
-      data: data.map(({name, value}) => [name, value]),
+      data: data?.map(({name, value}) => [name, value]),
```
AbhiPrasad, line 28: would be nice to have a test

DominikB2014: Agreed, would have to figure a good place for it, there's no tests for this component. So maybe somewhere higher up? 
Any objections to doing this in another PR to unblock?

AbhiPrasad: let's do it in another PR!

PR #72779, opened 2024-06-14
fix(pinned-searches): Delete ALL groupsearchviews upon savedsearch deletion

MichaelSun48:
We need to ensure that users' saved searches are persisted as custom views, AND that those users with a persisted saved saerch also have a second "Priorizted" tab. This PR updates the pinned searches endpoint to maintain this consistency by deleting ALL groupsearchviews upon savedsearch deletion (as opposed to only the groupsearchview associated with the user's saved search), and adding/updating the "Prioritized" view when a saved search is added/updated. 

This needs to be merged before the savedsearch -> groupsearchview [backfill](https://github.com/getsentry/sentry/pull/71731) is run

### DIFF
--- a/src/sentry/api/endpoints/organization_pinned_searches.py
+++ b/src/sentry/api/endpoints/organization_pinned_searches.py
@@ -85,7 +85,5 @@ class OrganizationPinnedSearchEndpoint(OrganizationEndpoint):
             type=search_type.value,
             visibility=Visibility.OWNER_PINNED,
         ).delete()
-        GroupSearchView.objects.filter(
-            organization=organization, user_id=request.user.id, position=0
-        ).delete()
+        GroupSearchView.objects.filter(organization=organization, user_id=request.user.id).delete()
         return Response(status=204)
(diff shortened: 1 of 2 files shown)


malwilley (commented):
Logic looks good! Can you add or modify tests to capture this behavior?

MichaelSun48:
> Logic looks good! Can you add or modify tests to capture this behavior?

On it!

src/sentry/api/endpoints/organization_pinned_searches.py
```diff
@@ -65,6 +65,18 @@ def put(self, request: Request, organization) -> Response:
                 "query_sort": result["sort"],
             },
         )
+
+        GroupSearchView.objects.create_or_update(
```
malwilley, line 69: nit: this isn't really a big deal since this endpoint will be going away after the search views are released, but it's not really clear (to others reading the code) why were are creating these entries here. An explanatory comment would be nice

malwilley (approved):

PR #72783, opened 2024-06-14
chore(insights): Remove special handling of LLM module

gggritso:
LLM module is in Open Beta! Which means:

1. It always and only lives at `/insights/llm-monitoring/`
2. Its breadcrumb always says "Insights -> LLM Monitoring"
3. Its title is always "Insights > LLM Monitoring"

No need for special conditions in the UI. Also adding a permanent redirect from `'/llm-monitoring/'` to `'/insights/llm-monitoring/'` which might not be necessary but seems like a good idea.

### DIFF
--- a/static/app/routes.tsx
+++ b/static/app/routes.tsx
@@ -1437,16 +1437,16 @@ function buildRoutes() {
     </Route>
   );
 
-  const llmMonitoringRoutes = (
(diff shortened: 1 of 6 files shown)


AbhiPrasad (approved):

colin-sentry (approved):
should be fine if it works locally

PR #72786, opened 2024-06-14
fix(query-builder): Fix bug where menu state is closed but displays as open

malwilley:
Followup to https://github.com/getsentry/sentry/pull/72776

Missed one extra part that was missing - the test was checking the state of the menu, but the component was actually ignoring that state erroneously. Added the check for the state and it all works as expected.

### DIFF
--- a/static/app/components/searchQueryBuilder/index.spec.tsx
+++ b/static/app/components/searchQueryBuilder/index.spec.tsx
@@ -326,7 +326,6 @@ describe('SearchQueryBuilder', function () {
         'placeholder',
         '-1d'
       );
-      await userEvent.click(screen.getByRole('combobox', {name: 'Edit filter value'}));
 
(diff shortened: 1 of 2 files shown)


static/app/components/searchQueryBuilder/index.spec.tsx
```diff
@@ -326,7 +326,6 @@ describe('SearchQueryBuilder', function () {
         'placeholder',
         '-1d'
       );
-      await userEvent.click(screen.getByRole('combobox', {name: 'Edit filter value'}));
```
malwilley, line 329: These lines now caused the test to fail (as they should) because clicking the combobox again should toggle it closed. Because we were missing the `state.isOpen` check these tests were passing.

scttcper (approved):

PR #72788, opened 2024-06-14
Incident subscription fk - don't cascade delete

nhsiehgit:
Inciden'ts should not be deleted when subscriptions are deleted

editing the migration should be fine since this migration is applied post_deployment

### DIFF
--- a/tests/sentry/migrations/test_0730_add_subscription_fk_to_incident.py
+++ b/tests/sentry/migrations/test_0730_add_subscription_fk_to_incident.py
@@ -140,6 +140,7 @@ class AlertRuleProjectBackfillTest(TestMigrations):
 
         self.query_subscription.delete()
         self.incident.refresh_from_db()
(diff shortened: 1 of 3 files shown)


tests/sentry/migrations/test_0730_add_subscription_fk_to_incident.py
```diff
@@ -138,6 +138,10 @@ def test(self):
         assert self.incident.subscription is not None
         assert self.incident.subscription == self.query_subscription
 
+        self.query_subscription.delete()
+        self.incident.refresh_from_db()
+        assert self.incident.subscription is None
```
saponifi3d, line 143: should we also assert that the incident still exists or would that throw when trying to access the subscription?

saponifi3d (approved):
d'oh, that makes sense - sorry i didn't catch in other pr. :shipit:

PR #72791, opened 2024-06-14
fix(insights): fix setting markline to undefined

edwardgou-sentry:
Updates the insights chart component to remove `markLine` from the incompleteSeries instead of setting it to `undefined`, which can lead to bugs.


AbhiPrasad (commented):
Can we revert the change in https://github.com/getsentry/sentry/pull/72773/files because of this and also add a test?

edwardgou-sentry:
> Can we revert the change in https://github.com/getsentry/sentry/pull/72773/files because of this and also add a test?

Added a test for this specific scenario, and also included the revert to `areaChart` in this pr. Tested it out on the page we were having issues with, so it should be safe.

AbhiPrasad (approved):

PR #72793, opened 2024-06-14
feat(uptime): Create issue occurrence when we get failures from the uptime checker

wedamija:
This creates an issue occurrence for the `UptimeDomainCheckFailure` issue platform type. Also add some testing framework things to support tests in the future

### DIFF
--- a/src/sentry/uptime/consumers/results_consumer.py
+++ b/src/sentry/uptime/consumers/results_consumer.py
@@ -9,9 +9,10 @@ from arroyo.processing.strategies.commit import CommitOffsets
 from arroyo.processing.strategies.run_task import RunTask
 from arroyo.types import BrokerValue, Commit, FilteredPayload, Message, Partition
 from sentry_kafka_schemas.codecs import Codec
-from sentry_kafka_schemas.schema_types.uptime_results_v1 import CheckResult
+from sentry_kafka_schemas.schema_types.uptime_results_v1 import CHECKSTATUS_FAILURE, CheckResult
 
 from sentry.conf.types.kafka_definition import Topic, get_topic_codec
+from sentry.uptime.issue_platform import create_issue_platform_occurrence
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


src/sentry/uptime/consumers/results_consumer.py
```diff
@@ -24,6 +25,8 @@ def process_result(message: Message[KafkaPayload | FilteredPayload]):
 
     try:
         result: CheckResult = UPTIME_RESULTS_CODEC.decode(message.payload.value)
+        if result["status"] == CHECKSTATUS_FAILURE:
+            create_issue_platform_occurrence(result)
 
         # XXX(epurkhiser): This consumer literally does nothing except log right now
```
evanpurkhiser, line 31: Can prob drop this comment. not critical if nothing else in the PR needs changed

src/sentry/uptime/issue_platform.py
```diff
@@ -0,0 +1,90 @@
...
+            ),
+        )
+
+    return IssueOccurrence(
+        id=uuid.uuid4().hex,
+        resource_id=None,
+        project_id=settings.UPTIME_POC_PROJECT_ID,  # TODO: Get this from the subscription or subscription like thing
+        event_id=uuid.uuid4().hex,
+        fingerprint=[
+            result["subscription_id"]
```
evanpurkhiser, line 62: This is going to be different every time. idk if we should finger print on this yet? otherwise new issue for every failure no?

src/sentry/uptime/issue_platform.py
```diff
@@ -0,0 +1,90 @@
...
+        ],  # TODO: Should be the specific monitor id related to the subscription
+        type=UptimeDomainCheckFailure,
+        issue_title="Uptime Check Failed for https://sentry.io",  # TODO: Get this from the uptime check details
```
evanpurkhiser, line 65: it's actually not checking sentry.io now, it's checking that uptime thing

evanpurkhiser (approved):

src/sentry/uptime/issue_platform.py
wedamija: Yeah, I figured it didn't matter too much - I'll follow up with improvements later

src/sentry/uptime/issue_platform.py
wedamija: Discussed on slack and we'll fix in the rust code instead, so it's easier to change. Also safe to merge now since we won't create issue platform issues for this type yet

PR #72797, opened 2024-06-14
ref: remove OptionMixin

asottile-sentry:
this fixes types for the various models involved

mixins aren't sound in typing unless they have an associated protocol -- and even then there aren't dependent types so this mixin couldn't possibly be typed properly.  it also wasn't being used at runtime either so it was easier to just write the methods directly in the two subclasses


JoshFerge (approved):
👏

markstory (approved):
Looks good to me.

PR #72799, opened 2024-06-14
ref: remove dead code in models.eventattachment

asottile-sentry:
this fixes type checking in this module when BaseManager becomes checked


asottile-sentry:
best viewed with [?w=1](https://github.com/getsentry/sentry/pull/72799/files?w=1)

joshuarli (approved):

PR #72809, opened 2024-06-14
ref: fix typing for sentry.buffer.base

asottile-sentry:

### DIFF
--- a/src/sentry/buffer/base.py
+++ b/src/sentry/buffer/base.py
@@ -150,7 +150,7 @@ class Buffer(Service):
                     # continue
                     pass
                 else:
-                    group.update(**update_kwargs)
+                    group.update(using=None, **update_kwargs)
                 created = False
             else:
                 _, created = model.objects.create_or_update(values=update_kwargs, **filters)


src/sentry/buffer/base.py
```diff
@@ -150,7 +150,7 @@ def process(
                     # continue
                     pass
                 else:
-                    group.update(**update_kwargs)
+                    group.update(using=None, **update_kwargs)
```
asottile-sentry, line 153: this one is a bit annoying, because `using` *could* be passed as a named or positional it *could* collide with arbitrary names in `update_kwargs` so the type checker rejects it.   explicitly passing the default here appeases the checker though

joshuarli: what's the annoyance? seems fine to me as it's just overridden if present in update_kwargs

joshuarli (approved):

src/sentry/buffer/base.py
asottile-sentry: in practice we don't have fields named `using` so it couldn't possibly collide

PR #72813, opened 2024-06-14
ref: fix some types for sentry.issues

asottile-sentry:

### DIFF
--- a/src/sentry/issues/endpoints/organization_activity.py
+++ b/src/sentry/issues/endpoints/organization_activity.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from datetime import datetime
 
 from rest_framework.request import Request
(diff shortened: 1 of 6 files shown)


src/sentry/issues/endpoints/organization_activity.py
```diff
@@ -70,7 +72,7 @@ def get(
             ).values_list("id", flat=True)
         )
 
-        union_qs = Activity.objects.none()
+        union_qs = Activity.objects.none().values_list("id", flat=True)
```
asottile-sentry, line 75: this convinces mypy the type is `QuerySet[Activity, int]`

src/sentry/issues/endpoints/organization_searches.py
```diff
@@ -110,6 +110,7 @@ def post(self, request: Request, organization: Organization) -> Response:
             sort=result["sort"],
             visibility=result["visibility"],
         )
+        assert saved_search.type is not None
```
joshuarli, line 113: wouldn't it be better to check result["type"] in the validator?

joshuarli (approved):

src/sentry/issues/endpoints/organization_searches.py
asottile-sentry: it is checked there -- but the model can do some defaulting on `type` so I have to check it here as well

technically the database schema is wrong (it should be `NOT NULL` for that column) but fixing that is a lot more involved

armenzg (approved):

PR #72817, opened 2024-06-14
perf(trace-explorer): Service Entry spans always have group 0

Zylphrex:
Group is only computed for the spans for the various insights modules. Service entry spans always have a group of 0, so leverage that for optimization.


wmak (approved):

Zylphrex:
Reverting since this assumption seems to be false for http.server transactions.

PR #72821, opened 2024-06-14
ref(rules): Addtl final touches on rule create/edit confirm notifications

ceorourke:
More minor changes following feedback 
Screenshot 2024-06-14 at 2 15 50 PM


rachrwang:
As discussed in slack, we'll add a carriage return between header & text for both create & update messages. Thanks!

schew2381 (approved):

PR #72823, opened 2024-06-14
fix(replay): Fix error state quickly flashing when loading replay

### LINKED ISSUE #70077: Error screen very briefly flashes during loading state of a replay
When you visit replay details, you can see that the error screen very briefly flashes during the loading state inside of the replay area.

https://github.com/getsentry/sentry/assets/79684/7e1f9f4f-b14a-44ed-8932-a2ceb26a82b4

billyvg:
This fixes the error state that quickly flashes when loading a replay.

The `useReplayData` hook makes ~5 requests, but they do not all happen in parallel. The fetching state ends up looking like the following:

* fetch replay, fetch extra errors, fetch platform ("fetching" == true)
* above requests finish ("fetching" == false)
* fetch attachments, fetch extra errors, fetch platform errors ("fetching" == true)

This intermittent fetching state means that we have [components that erroneously render](https://github.com/getsentry/sentry/blob/78334b95bc5ff69816fcaeafff8f67fadf0727c7/static/app/components/replays/replayView.tsx#L43) an error state because it believes API requests have completed and that the data is missing.

To test:
Loading a replay should no longer flicker and should no longer flicker with a red error screen. We use `useReplayReader` in quite a few places besides replay details, but I think all of them should benefit from this change.

Closes https://github.com/getsentry/sentry/issues/70077

### DIFF
--- a/static/app/utils/replays/hooks/useReplayData.tsx
+++ b/static/app/utils/replays/hooks/useReplayData.tsx
@@ -1,4 +1,4 @@
-import {useCallback, useMemo} from 'react';
+import {useCallback, useMemo, useRef} from 'react';
 
 import {ALL_ACCESS_PROJECTS} from 'sentry/constants/pageFilters';
 import useFetchParallelPages from 'sentry/utils/api/useFetchParallelPages';
@@ -76,8 +76,8 @@ function useReplayData({
   errorsPerPage = 50,
   segmentsPerPage = 100,
 }: Options): Result {
+  const hasFetchedAttachments = useRef(false);
   const projects = useProjects();
-
   const queryClient = useQueryClient();
 
   // Fetch every field of the replay. The TS type definition lists every field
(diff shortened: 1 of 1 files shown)


static/app/utils/replays/hooks/useReplayData.tsx
```diff
@@ -230,12 +233,23 @@ function useReplayData({
...
   return useMemo(() => {
+    // This hook can enter a state where `fetching` below is false
+    // before it is entirely ready (i.e. it has not fetched
+    // attachemnts yet). This can cause downstream components to
+    // think it is no longer fetching and will display an error
+    // because there are no attachments. The below will require
+    // that we have attempted to fetch an attachment once (or if it
+    // errors) before we toggle fetching state to false.
+    hasFetchedAttachments.current =
+      hasFetchedAttachments.current || isFetchingAttachments;
```
billyvg, line 244: please double check my logic, but i'm pretty sure this is right and pretty sure we _always_ want to make at least 1 request for attachments or something is wrong.

static/app/utils/replays/hooks/useReplayData.tsx
```diff
@@ -76,8 +76,8 @@ function useReplayData({
   errorsPerPage = 50,
   segmentsPerPage = 100,
 }: Options): Result {
+  const hasFetchedAttachments = useRef(false);
```
billyvg, line 79: Does this need to be a ref?

billyvg: (It does, otherwise value will get lost between re-renders)

static/app/utils/replays/hooks/useReplayData.tsx
```diff
@@ -230,12 +233,25 @@ function useReplayData({
   }, [orgSlug, replayId, projectSlug, queryClient]);
 
   return useMemo(() => {
+    // This hook can enter a state where `fetching` below is false
+    // before it is entirely ready (i.e. it has not fetched
+    // attachemnts yet). This can cause downstream components to
+    // think it is no longer fetching and will display an error
+    // because there are no attachments. The below will require
+    // that we have attempted to fetch an attachment once (or if it
```
michellewzhang, line 241:
```suggestion
    // that we have attempted to fetch an attachment once (or it
```

michellewzhang (approved):

aliu39 (approved):

Issue #68109, opened 2024-04-02
[Replay Details] Address UI jumping when loading in mobile replay details


michellewzhang:
Right now, we're loading the replay details assuming it's a web replay. When we realize it's a video replay, then we hide/change around a bunch of things: namely

- [x]  "right click to inspect" blue dom banner goes away
- [x]  some tabs are hidden
- [x]   "contact us" button -> "give feedback" button in top right corner
- [ ]   disabled playback settings
- [x]   rage/dead clicks in header go away
- [x]   browser icon in url area goes away
- [x]   configure replay button in top right goes away

See video below:

https://github.com/getsentry/sentry/assets/56095982/174142f9-0804-484c-b168-451c54702802

It's tricky because we don't know whether it's a video replay until all the data is loaded, so the "jumping" is unavoidable at the moment. We should have a better experience here without degrading our current web replays experience.

michellewzhang:
playback button greying out doesn't seem like the biggest deal. we'll add playback functionality back soon enough anyway

PR #72825, opened 2024-06-14
fix(replay): fix some ui jumping between mobile/web

michellewzhang:
this PR fixes some of the UI 'jumping' when we realize a mobile replay is a mobile replay (and not a web replay). 

namely, this adds two loading placeholders which hides the following discrepancies while we're waiting for our `isVideoReplay` check to evaluate:
- web "contact us" button vs. mobile "give feedback" button in top right corner
- no rage/dead clicks in header for mobile replays
- configure replay button in top right goes away for mobile replays

also updated the header placeholder style so they're consistent, and less jarring white.

before: 

https://github.com/getsentry/sentry/assets/56095982/094bfea3-0d01-46f3-87d2-c93395c25dbb

after (mobile) - testing on `dev-ui` so the 'give feedback' button doesn't show up but it will on prod:

https://github.com/getsentry/sentry/assets/56095982/1d4e7c70-42f1-4189-bffd-583848d6c43e

after (web):

https://github.com/getsentry/sentry/assets/56095982/50c8184c-fe15-4899-af98-a464e41fb553

it's not perfect and still looks slightly janky (probably because the dimensions aren't exact - they were chosen by me arbitrarily and can change; open to suggestions). also because the 'seen by' avatar bit loads in at a different time 🫠

the logic i'm using to check for the loading state is based on whether the rrweb events have all loaded in, because for normal replays (both mobile/web) there are at least 2 rrweb events - otherwise there was some sort of processing error. 

will follow up with changes to fix the rest of the discrepancies; wanted to get feedback on this new logic/loading state first.

relates to https://github.com/getsentry/sentry/issues/68109

### DIFF
--- a/static/app/views/replays/detail/page.tsx
+++ b/static/app/views/replays/detail/page.tsx
@@ -11,9 +11,7 @@ import Placeholder from 'sentry/components/placeholder';
 import ConfigureReplayCard from 'sentry/components/replays/configureReplayCard';
 import DetailsPageBreadcrumbs from 'sentry/components/replays/header/detailsPageBreadcrumbs';
 import FeedbackButton from 'sentry/components/replays/header/feedbackButton';
-import HeaderPlaceholder from 'sentry/components/replays/header/headerPlaceholder';
 import ReplayMetaData from 'sentry/components/replays/header/replayMetaData';
-import {useReplayContext} from 'sentry/components/replays/replayContext';
 import SentryDocumentTitle from 'sentry/components/sentryDocumentTitle';
 import TimeSince from 'sentry/components/timeSince';
 import {IconCalendar, IconDelete, IconEllipsis, IconUpload} from 'sentry/icons';
@@ -30,6 +28,7 @@ type Props = {
   projectSlug: string | null;
   replayErrors: ReplayError[];
   replayRecord: undefined | ReplayRecord;
+  isLoading?: boolean;
   isVideoReplay?: boolean;
 };
 
@@ -40,6 +39,7 @@ export default function Page({
   projectSlug,
   replayErrors,
   isVideoReplay,
+  isLoading,
 }: Props) {
   const title = replayRecord
     ? `${replayRecord.user.display_name ?? t('Anonymous User')} — Session Replay — ${orgSlug}`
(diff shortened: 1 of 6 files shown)


static/app/views/replays/detail/page.tsx
```diff
@@ -46,6 +48,15 @@ export default function Page({
...
   const onDeleteReplay = useDeleteReplay({replayId: replayRecord?.id, projectSlug});
 
+  const {replay} = useReplayContext();
+  const rrwebFrames = replay?.getRRWebFrames();
+  // The replay data takes a while to load in, which causes our `isVideoReplay`
+  // to return an early `false`, which used to cause UI jumping.
+  // One way to check whether it's finished loading is by checking the length
+  // of the rrweb frames, which should always be > 2 for any given replay.
+  // By default, the 2 frames are replay.start and replay.end
+  const isLoading = !rrwebFrames || (rrwebFrames && rrwebFrames.length <= 2);
```
aliu39, line 58: Overall lgtm, but I don't have the knowledge to confirm this logic

aliu39, line 58:
```suggestion
  const isLoading = !rrwebFrames || (rrwebFrames.length <= 2);
```

aliu39 (approved):

static/app/views/replays/detail/page.tsx
michellewzhang: i think we need the `rrwebFrames` check here in case it's `undefined` or `null`

aliu39, line 58: Could be more specific with `isLoadingFrames`, `isLoadingVideo`?

aliu39: Yeah but the `||` should short circuit if undefined or null right?

michellewzhang: you get a typescript error if you try to access `.length` or something undefined

ryan953, line 58: Can we hoist this variable up into static/app/views/replays/details.tsx and pass it down? It's similar in my mind to `isVideoReplay` and should be set together.

Eventually this logic could be returned from `useReplayData`, so it can be re-used

ryan953 (approved):

PR #72826, opened 2024-06-14
chore(billing): Update docs links for managing quotas

isabellaenriquez:
update the quota docs links for transactions and spans :)

### DIFF
--- a/static/app/views/settings/account/notifications/utils.tsx
+++ b/static/app/views/settings/account/notifications/utils.tsx
@@ -40,9 +40,9 @@ export function getDocsLinkForEventType(
 ) {
   switch (event) {
     case 'transaction':
-      return 'https://docs.sentry.io/product/performance/transaction-summary/#what-is-a-transaction';
+      return 'https://sentry-docs-git-move-quota-to-billing.sentry.dev/pricing/quotas/manage-transaction-quota/';
     case 'span':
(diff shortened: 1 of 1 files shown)


static/app/views/settings/account/notifications/utils.tsx
```diff
@@ -40,9 +40,9 @@ export function getDocsLinkForEventType(
 ) {
   switch (event) {
     case 'transaction':
-      return 'https://docs.sentry.io/product/performance/transaction-summary/#what-is-a-transaction';
+      return 'https://sentry-docs-git-move-quota-to-billing.sentry.dev/pricing/quotas/manage-transaction-quota/';
```
schew2381, line 43: why is this link so strange? I tested it and it works but still

schew2381: weird cuz then going anywhere else on docs also uses that domain

scttcper: this is the domain we're putting am2 docs on for now

scttcper (approved):

static/app/views/settings/account/notifications/utils.tsx
schew2381: ah ok makes sense

isabellaenriquez: i'll double check with the team but this is specifically for billing with transactions (old) vs billing with spans (new)

PR #72848, opened 2024-06-17
feat(starfish): Add count_if for metrics

philipphofmann:
We need count_if also for metrics and not just for span metrics. Follow-up of GH-72582.

### DIFF
--- a/src/sentry/search/events/datasets/metrics.py
+++ b/src/sentry/search/events/datasets/metrics.py
@@ -178,6 +178,10 @@ class MetricsDatasetConfig(DatasetConfig):
                     "count_if",
                     required_args=[
                         fields.MetricArg(
+                            "column",
+                            allowed_columns=constants.METRIC_DURATION_COLUMNS,
+                        ),
+                        fields.MetricArg(
                             "if_col",
                             allowed_columns=["release"],
                         ),
@@ -185,6 +189,7 @@ class MetricsDatasetConfig(DatasetConfig):
                             "if_val", unquote=True, unescape_quotes=True, optional_unquote=True
                         ),
                     ],
+                    calculated_args=[resolve_metric_id],
                     snql_distribution=lambda args, alias: Function(
                         "countIf",
                         [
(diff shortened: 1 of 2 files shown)


src/sentry/search/events/datasets/metrics.py
```diff
@@ -174,6 +174,42 @@ def function_converter(self) -> Mapping[str, fields.MetricsFunction]:
...
                 ),
+                fields.MetricsFunction(
+                    "count_if",
+                    required_args=[
+                        fields.MetricArg(
+                            "if_col",
+                            allowed_columns=["release"],
+                        ),
+                        fields.SnQLStringArg(
+                            "if_val", unquote=True, unescape_quotes=True, optional_unquote=True
```
wmak, line 185: Please update this to match the syntax of the other datasets, and include the condition argument as well?

wmak (commented):
BTW, this wasn't included originally because there were concerns with eventually having the query builders work with the metrics layer. I don't think this is a blocker just a heads up

Also not sure if it doesn't work for some reason, but can you do a query where you group by release, and filter with an IN `release:[1.0.0,2.0.0]`

philipphofmann:
> Also not sure if it doesn't work for some reason, but can you do a query where you group by release, and filter with an IN release:[1.0.0,2.0.0]

I'm not used to Snuba datasets and such yet. How would that be useful, @wmak 🤔?

narsaynorath:
> Also not sure if it doesn't work for some reason, but can you do a query where you group by release, and filter with an IN release:[1.0.0,2.0.0]

@wmak I think this approach works if we're passing the data down to individual components (e.g. like the metrics we show in the headers of the detail pages), but if the intention is to add this to a table (e.g. surface count in the Screens Overview page) then we'll need this function to keep the results in the same row. I'll let @philipphofmann confirm how it's being used because I believe this was for the overview table, but could also be for both use cases depending on what dataset is being used

narsaynorath (approved):
This looks good to me. I'm approving it because I think we can add it and if there are issues we can always remove/modify it. I think there are use cases that make this valid for the UI module.

philipphofmann:
I currently don't plan to use the data in a table. Only for the metrics ribbon. Thanks for explaining this in detail, @narsaynorath. I agree that we can always change it later. @wmak, I'm going to merge this now. If you want it changed, please add a comment, and I will follow up with another PR.

PR #72855, opened 2024-06-17
Remove one minute interval as an option for custom metrics

vgrozdanic:
Part of: [#72843](https://github.com/getsentry/sentry/issues/72843)

Alerts for custom metrics and time interval of 1 minute are not reliable due to a big latency (around 3 minutes) when custom metrics are ingested. Because of that, we will disable users to use 1 minute alert interval for custom metrics alerts. This change is only present on custom metrics alert creation.

### DIFF
--- a/static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
+++ b/static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
@@ -142,6 +142,17 @@ class RuleConditionsForm extends PureComponent<Props, State> {
 
   get timeWindowOptions() {
     let options: Record<string, string> = TIME_WINDOW_MAP;
+    const {alertType} = this.props;
+
+    if (alertType === 'custom_metrics') {
+      // Do not show ONE MINUTE interval as an option for custom_metrics alert
+      options = Object.keys(TIME_WINDOW_MAP)
+        .filter(key => key !== TimeWindow.ONE_MINUTE.toString())
+        .reduce((obj, key) => {
+          obj[key] = options[key];
+          return obj;
+        }, {});
+    }
 
     if (isCrashFreeAlert(this.props.dataset)) {
       options = pick(TIME_WINDOW_MAP, [
--- a/static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
+++ b/static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
@@ -143,15 +143,11 @@ class RuleConditionsForm extends PureComponent<Props, State> {
 
   get timeWindowOptions() {
     let options: Record<string, string> = TIME_WINDOW_MAP;
-    const {alertType, timeWindow, onTimeWindowChange} = this.props;
+    const {alertType} = this.props;
 
     if (alertType === 'custom_metrics') {
       // Do not show ONE MINUTE interval as an option for custom_metrics alert
       options = omit(options, TimeWindow.ONE_MINUTE.toString());
-
-      if (timeWindow === TimeWindow.ONE_MINUTE) {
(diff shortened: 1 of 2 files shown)


static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
```diff
@@ -142,6 +142,17 @@ class RuleConditionsForm extends PureComponent<Props, State> {
 
   get timeWindowOptions() {
     let options: Record<string, string> = TIME_WINDOW_MAP;
+    const {alertType} = this.props;
+
+    if (alertType === 'custom_metrics') {
+      // Do not show ONE MINUTE interval as an option for custom_metrics alert
+      options = Object.keys(TIME_WINDOW_MAP)
```
obostjancic, line 149: we often use lodash/omit for this like:
```suggestion
      options = omit(options, TimeWindow.ONE_MINUTE.toString())
```

vgrozdanic:
@ArthurKnaus fixed - now it is moved to 5 minutes if 1 minute was selected

static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
```diff
@@ -142,6 +143,16 @@ class RuleConditionsForm extends PureComponent<Props, State> {
...
     let options: Record<string, string> = TIME_WINDOW_MAP;
+    const {alertType, timeWindow, onTimeWindowChange} = this.props;
+
+    if (alertType === 'custom_metrics') {
+      // Do not show ONE MINUTE interval as an option for custom_metrics alert
+      options = omit(options, TimeWindow.ONE_MINUTE.toString());
+
+      if (timeWindow === TimeWindow.ONE_MINUTE) {
+        onTimeWindowChange(TimeWindow.FIVE_MINUTES);
+      }
```
ArthurKnaus, line 154: Imho a getter should not trigger a state update.

Other side effects like this are handled here:
https://github.com/getsentry/sentry/blob/ef282f16269785e57ade409b3d8f1f91015920ae/static/app/views/alerts/rules/metric/ruleForm.tsx#L512

static/app/views/alerts/rules/metric/ruleForm.tsx
```diff
@@ -520,6 +521,15 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
...
     }
 
+    if (
+      name === 'dataset' &&
+      value === 'generic_metrics' &&
+      this.state.timeWindow === TimeWindow.ONE_MINUTE
+    ) {
+      // ONE MINUTE interval is not allowed for custom metrics alerts
+      this.setState({timeWindow: TimeWindow.FIVE_MINUTES});
+    }
```
ArthurKnaus, line 531: Couldn't you handle it where the `alert_type` field changes? (the condition above)
```ts
if (name === 'alertType') {
  this.setState(({dataset, timeWindow}) => ({
    alertType: value as MetricAlertType,
    dataset: this.checkOnDemandMetricsDataset(dataset, this.state.query),
    // ONE MINUTE interval is not allowed for custom metrics alerts
    timeWindow: value === 'custom_metrics' && timeWindow === TimeWindow.ONE_MINUTE
      ? TimeWindow.FIVE_MINUTES
      : timeWindow 
  }));
}
```

ArthurKnaus (approved):
Looks good! Just one small thing :)

static/app/views/alerts/rules/metric/ruleForm.tsx
ArthurKnaus: If possible you should reference current state values by calling setState as a function to avoid race conditions.

PR #72858, opened 2024-06-17
fix(dynamic-sampling): Fix unsupported search query FE

### LINKED ISSUE #72857: Fix the front-end of discover so the API is not called with default wrong arguments

priscilawebdev:
closes https://github.com/getsentry/sentry/issues/72857
relates https://github.com/getsentry/sentry/issues/72300

### DIFF
--- a/static/app/components/dynamicSampling/investigationRule.tsx
+++ b/static/app/components/dynamicSampling/investigationRule.tsx
@@ -18,6 +18,8 @@ import {useApiQuery, useMutation, useQueryClient} from 'sentry/utils/queryClient
 import type RequestError from 'sentry/utils/requestError/requestError';
 import useApi from 'sentry/utils/useApi';
 import useOrganization from 'sentry/utils/useOrganization';
+import {Datasource} from 'sentry/views/alerts/rules/metric/types';
+import {getQueryDatasource} from 'sentry/views/alerts/utils';
 
 // Number of samples under which we can trigger an investigation rule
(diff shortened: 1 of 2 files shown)


static/app/components/dynamicSampling/investigationRule.tsx
```diff
@@ -174,34 +178,20 @@ const InvestigationInProgressNotification = styled('span')`
   gap: ${space(0.5)};
 `;
 
-function checkIsTransactionQueryMissing(error: RequestError | null) {
```
priscilawebdev, line 177: since the request won't be fired if `event.type='transaction'` is not present in the query, this won't happen anymore - therefore I am removing this code

static/app/components/dynamicSampling/investigationRule.tsx
```diff
@@ -174,34 +178,20 @@ const InvestigationInProgressNotification = styled('span')`
...
   const organization = props.organization;
   const period = props.eventView.statsPeriod || null;
   const query = props.eventView.getQuery();
+  const isTransactionQueryMissing =
+    getQueryDatasource(query)?.source !== Datasource.TRANSACTION;
   const createInvestigationRule = useCreateInvestigationRuleMutation();
   const {
     data: rule,
-    isLoading,
+    isFetching: isLoading,
```
priscilawebdev, line 191: had to use `isFetching` instead of `isLoading`, because, in this current version of react-query, the `isLoading` state is always `true` if `enabled` equals `false`

ArthurKnaus (approved):

PR #72863, opened 2024-06-17
chore(slack): remove old slack client from notify recipient

cathteng:
Default to using the Slack SDK Client for notifying recipients (sending an "activity" or "alert rule" notification to a Slack user or team, but NOT to a channel directly). The feature has been GA'd.

Also adds metrics for success/failure of chat_postMessage for notifying recipients.

Fixes a LOT of tests that were using the `responses` library to mock Slack responses, but this needed to be replaced because slack_sdk uses urllib. These tests check the call args to `chat_postMessage`.

### DIFF
--- a/src/sentry/notifications/notifications/organization_request/abstract_invite_request.py
+++ b/src/sentry/notifications/notifications/organization_request/abstract_invite_request.py
@@ -78,7 +78,7 @@ class AbstractInviteRequestNotification(OrganizationRequestNotification, abc.ABC
             MessageAction(
                 name="Reject",
                 style="danger",
-                action_id="approve_request",
+                action_id="reject_request",
                 value="reject_member",
                 label="Reject",
             ),
--- a/src/sentry/tasks/integrations/slack/post_message.py
+++ b/src/sentry/tasks/integrations/slack/post_message.py
@@ -23,7 +23,7 @@ def _send_message_to_slack_channel(
     payload: Mapping[str, Any],
     log_error_message: str,
     log_params: Mapping[str, Any],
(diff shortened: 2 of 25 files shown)


src/sentry/notifications/notifications/organization_request/abstract_invite_request.py
```diff
@@ -75,7 +75,7 @@ def get_message_actions(
             MessageAction(
                 name="Reject",
                 style="danger",
-                action_id="approve_request",
```
sentaur-athena, line 78: 😅

src/sentry/tasks/integrations/slack/post_message.py
```diff
@@ -20,29 +23,27 @@ def _send_message_to_slack_channel(
     payload: Mapping[str, Any],
     log_error_message: str,
     log_params: Mapping[str, Any],
-    has_sdk_flag: bool | None = False,
+    has_sdk_flag: bool | None = True,  # TODO: remove all these
```
sentaur-athena, line 26: Just curious. Why can't you remove it in this PR?

cathteng: this is in a task, it's possible if we remove it, we are still spawning tasks from one pod that pass the argument but the pod that's picking it up has updated code that doesn't have the argument in the function

cathteng: i had this issue where i added and was passing the argument to the task in one PR and it errored in Sentry for a couple minutes during deployment

iamrajjoshi (approved):

PR #72864, opened 2024-06-17
feat(query-builder): Use CompactSelect for operator button

malwilley:
Because the operator has a success state, we should be using CompactSelect instead of a DropdownMenu. This also sets us up better for making more complex operator actions.

CleanShot 2024-06-17 at 09 02 52

### DIFF
--- a/static/app/components/compactSelect/control.tsx
+++ b/static/app/components/compactSelect/control.tsx
@@ -444,6 +444,8 @@ export function Control({
       if (e.key === 'ArrowDown' || e.key === 'ArrowUp') {
         e.preventDefault(); // Prevent scroll
         overlayState.open();
+      } else {
+        e.continuePropagation();
       }
     },
   });
--- a/static/app/components/searchQueryBuilder/useQueryBuilderGridItem.tsx
+++ b/static/app/components/searchQueryBuilder/useQueryBuilderGridItem.tsx
@@ -69,7 +69,10 @@ export function useQueryBuilderGridItem(
         return;
       }
 
-      const walker = getFocusableTreeWalker(e.currentTarget, {wrap: false});
+      const walker = getFocusableTreeWalker(e.currentTarget, {
+        wrap: false,
+        accept: node => node.tagName === 'BUTTON',
+      });
       walker.currentNode = e.target as FocusableElement;
 
       // On ArrowRight, we want to focus the next grid cell if there is one.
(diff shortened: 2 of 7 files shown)


static/app/components/compactSelect/control.tsx
```diff
@@ -444,6 +444,8 @@ export function Control({
       if (e.key === 'ArrowDown' || e.key === 'ArrowUp') {
         e.preventDefault(); // Prevent scroll
         overlayState.open();
+      } else {
+        e.continuePropagation();
```
malwilley, line 448: Without continuing propagation, arrow left/right while the op trigger was focused wouldn't do anything. We probably should have had this here already, I don't foresee any issues with adding it.

static/app/components/searchQueryBuilder/useQueryBuilderGridItem.tsx
```diff
@@ -69,7 +69,10 @@ export function useQueryBuilderGridItem(
         return;
       }
 
-      const walker = getFocusableTreeWalker(e.currentTarget, {wrap: false});
+      const walker = getFocusableTreeWalker(e.currentTarget, {
+        wrap: false,
+        accept: node => node.tagName === 'BUTTON',
```
malwilley, line 74: Had to add this to make the tests happy, but I think it's also a good change. All the focusable elements we care about in a filter are buttons.

roggenkemper:
one thing i noticed is that if i click the operator and then click the values dropdown - it used to close the operators dropdown and open the dropdown of options for values but now it just closes compact select which means you have to double click to get to the values dropdown.

malwilley:
@roggenkemper you're right, and it's because of this https://github.com/adobe/react-spectrum/issues/2061

It's the intended behavior actually, which I sort of agree with, but it does make the interactions feel less fluid. Especially since the comboboxes lose focus more readily. I may end up adding some extra logic for items in the searchbar if it ends up becoming something that is an issue.

roggenkemper (approved):

MichaelSun48:
I think this is a bug, but correct me if this is the intended behavior: click inside the filter operator, then click outside. The filter operator is still highlighted despite not being in focus. If this is a bug and the fix isn't within the scope of this PR, lmk and I'll add it to the feedback doc

https://github.com/getsentry/sentry/assets/55160142/c92462e8-de94-479e-bbf1-385d900d1fcb

malwilley:
@MichaelSun48 yeah that's related to the above comment, and is how all of our CompactSelects work (though they have some special logic that executes when you click another dropdown). 

Though now that you both have commented on it, I'll follow up with some work to try and fix this behavior. Will probably require some changes to that component or `useOverlay` so I want to do it in a separate PR.

PR #72867, opened 2024-06-17
fix(crons): Fix bug where processing error data growth is not capped.

wedamija:
This fixes a bug where we don't delete the body of a processing error after evicting it from the set of data for that monitor. This occurred when we started storing the bodies of the errors in a separate key and we forgot to delete it.

### DIFF
--- a/src/sentry/monitors/processing_errors/manager.py
+++ b/src/sentry/monitors/processing_errors/manager.py
@@ -131,10 +131,15 @@ def store_error(error: CheckinProcessingError, monitor: Monitor | None):
     pipeline = redis_client.pipeline(transaction=False)
     pipeline.zadd(error_set_key, {error.id.hex: error.checkin.ts.timestamp()})
     pipeline.set(error_key, serialized_error, ex=MONITOR_ERRORS_LIFETIME)
-    # Cap the error list to the `MAX_ERRORS_PER_SET` most recent errors
-    pipeline.zremrangebyrank(error_set_key, 0, -(MAX_ERRORS_PER_SET + 1))
     pipeline.expire(error_set_key, MONITOR_ERRORS_LIFETIME)
-    pipeline.execute()
+    pipeline.zrange(error_set_key, 0, -(MAX_ERRORS_PER_SET + 1))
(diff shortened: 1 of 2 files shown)


src/sentry/monitors/processing_errors/manager.py
```diff
@@ -131,10 +131,15 @@ def store_error(error: CheckinProcessingError, monitor: Monitor | None):
     pipeline = redis_client.pipeline(transaction=False)
     pipeline.zadd(error_set_key, {error.id.hex: error.checkin.ts.timestamp()})
     pipeline.set(error_key, serialized_error, ex=MONITOR_ERRORS_LIFETIME)
-    # Cap the error list to the `MAX_ERRORS_PER_SET` most recent errors
-    pipeline.zremrangebyrank(error_set_key, 0, -(MAX_ERRORS_PER_SET + 1))
     pipeline.expire(error_set_key, MONITOR_ERRORS_LIFETIME)
-    pipeline.execute()
+    pipeline.zrange(error_set_key, 0, -(MAX_ERRORS_PER_SET + 1))
+    results = pipeline.execute()[3]
```
evanpurkhiser, line 136: What's the 3rd tuple?

wedamija: The pipeline returns results for all commands. The `zrange` is in index 3

evanpurkhiser: I would almost be tempted to destructure and use underscore prefixes for the ones we don't care about, so it's more clear

evanpurkhiser (approved):
Possible to write a test?

wedamija:
> Possible to write a test?

? I updated a test to check for it

davidenwang (approved):

src/sentry/monitors/processing_errors/manager.py
wedamija: I'm not sure if that's clearer... what if we had 10 operations here, should we destructure all of them?

Although I wonder if it's safer to use `-1` and then try to keep this as the last operation...

PR #72869, opened 2024-06-17
ref(replay): update mobile breadcrumb messages

michellewzhang:
Some updates to the mobile breadcrumb messaging for `device.connectivity` and `app.foreground` / `app.background`

SCR-20240617-jhka
SCR-20240617-jgzc
SCR-20240617-jgya

cc @Jesse-Box

### DIFF
--- a/static/app/utils/replays/getFrameDetails.tsx
+++ b/static/app/utils/replays/getFrameDetails.tsx
@@ -53,6 +53,13 @@ interface Details {
   title: ReactNode;
 }
 
+const DEVICE_CONNECTIVITY_MESSAGE: Record<string, string> = {
+  wifi: 'Device connected to wifi',
+  offline: 'Internet connection was lost',
+  cellular: 'Device connected to cellular network',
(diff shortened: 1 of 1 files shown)


static/app/utils/replays/getFrameDetails.tsx
```diff
@@ -220,14 +227,14 @@ const MAPPER_FOR_FRAME: Record<string, (frame) => Details> = {
   }),
   'app.foreground': () => ({
     color: 'blue300',
-    description: 'Replay started',
+    description: 'The user is currently focused on your application',
```
aliu39, line 230: Nit: remove the "currently"?

aliu39 (approved):

ryan953 (approved):
Can we translate the messages. Seems like the standard is for titles to not be.

static/app/utils/replays/getFrameDetails.tsx
michellewzhang: @Jesse-Box suggested these copies

Issue #25291, opened 2021-04-15
Merging origin/master into my feature branch causes almost all JS tests to run in pre-commit


untitaker:
Our `sentry.lint.engine`-based hooks for running JS lints + tests are also triggered upon `git fetch && git merge origin/master` in a feature branch. This stands in contrast to our pre-commit ([the framework](https://pre-commit.com/)) -based Python hooks.

@joshuarli insinuated on Slack that moving them over to pre-commit (the framework) solves this problem. Then we could potentially delete `sentry.lint.engine`

natemoo-re:
Closing this as it appears that a solution was merged in https://github.com/getsentry/sentry/pull/72870, but this issue was never closed.

PR #72870, opened 2024-06-17
devenv: move legacy setup-git to sync

joshuarli:
First step towards cleaning up all the legacy stuff. This moves setup-git to sync. context: https://github.com/getsentry/getsentry/pull/13540/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R28

Also removes the old `make bootstrap` and `make develop` now just runs `devenv sync`.


devenv/sync.py
```diff
@@ -128,17 +128,20 @@ def main(context: dict[str, str]) -> int:
...
-            ),
-        ),
+        (("pre-commit dependencies", ("pre-commit", "install", "--install-hooks", "-f")),),
     ):
         return 1
 
+    git_config = configparser.ConfigParser()
+    git_config.read(f"{reporoot}/.git/config")
+    git_config["blame"] = {"ignoreRevsFile": ".git-blame-ignore-revs"}
+    git_config["branch"] = {"autosetuprebase": "always"}
```
asottile-sentry, line 138: I know people won't want this setting always -- I wonder what we can do about that 🤔 

maybe we move that setting to `bootstrap` and then leave it unmanaged afterwards?

joshuarli: Mmm, good idea

joshuarli: leaning towards putting that in a new `post-bootstrap.py` instead of releasing new devenv just for this change, wdyt?

asottile-sentry: wfm

joshuarli: well, will still need a new devenv release for post-bootstrap but that's a one time deal :P

config/hooks/pre-commit
```diff
@@ -1,58 +0,0 @@
-#!/usr/bin/env python
```
asottile-sentry, line 1: annoyingly I think we still need this thing since we haven't ported the js checks over to `.pre-commit-config.yaml`

devenv/sync.py
joshuarli: https://github.com/getsentry/devenv/pull/126

config/hooks/pre-commit
joshuarli: hmm... i'll look into getting that done before this PR then

joshuarli: lol been wanting to do that since 2021 apparently https://github.com/getsentry/sentry/issues/25291

IanWoodard (approved):

PR #72871, opened 2024-06-17
ref(slack): use sdk client to get users with pagination

cathteng:
Use the Slack SDK client to get users during integration post-install. The `slack_sdk` `SlackResponse` class has a built in method of pagination using an iterator, so we can use that here.

### DIFF
--- a/src/sentry/integrations/slack/utils/users.py
+++ b/src/sentry/integrations/slack/utils/users.py
@@ -3,7 +3,11 @@ from __future__ import annotations
 from collections.abc import Mapping, MutableMapping, Sequence
 from typing import Any
 
+from slack_sdk.errors import SlackApiError
+
+from sentry import features
 from sentry.integrations.slack.client import SlackClient
+from sentry.integrations.slack.sdk_client import SlackSdkClient
 from sentry.models.integrations.integration import Integration
 from sentry.models.organization import Organization
 from sentry.models.user import User
@@ -18,31 +22,58 @@ SLACK_GET_USERS_PAGE_SIZE = 200
 
 
 def get_users(integration: Integration, organization: Organization) -> Sequence[Mapping[str, Any]]:
-    client = SlackClient(integration_id=integration.id)
-
     user_list = []
     next_cursor = None
-    for i in range(SLACK_GET_USERS_PAGE_LIMIT):
+
+    if features.has("organizations:slack-sdk-get-users", organization):
+        sdk_client = SlackSdkClient(integration_id=integration.id)
+
         try:
-            next_users = client.get(
-                "/users.list",
-                params={"limit": SLACK_GET_USERS_PAGE_SIZE, "cursor": next_cursor},
-            )
-        except ApiError as e:
+            users = sdk_client.users_list(cursor=next_cursor, limit=SLACK_GET_USERS_PAGE_SIZE)
+        except SlackApiError as e:
             logger.info(
-                "post_install.fail.slack_users.list",
+                "slack.post_install.get_users.error",
                 extra={
                     "error": str(e),
                     "organization": organization.slug,
                     "integration_id": integration.id,
                 },
             )
-            break
-        user_list += next_users["members"]
+            return []
 
-        next_cursor = next_users["response_metadata"]["next_cursor"]
-        if not next_cursor:
-            break
+        next_users = iter(users)
+        for _ in range(SLACK_GET_USERS_PAGE_LIMIT):
+            try:
+                user_list += next_users.get("members")
+                next_users = next(users)
+            except StopIteration:
+                break
+            except SlackApiError:
+                break
+    else:
+        client = SlackClient(integration_id=integration.id)
+
+        for _ in range(SLACK_GET_USERS_PAGE_LIMIT):
+            try:
+                next_users = client.get(
+                    "/users.list",
+                    params={"limit": SLACK_GET_USERS_PAGE_SIZE, "cursor": next_cursor},
+                )
+            except ApiError as e:
+                logger.info(
+                    "post_install.fail.slack_users.list",
+                    extra={
+                        "error": str(e),
+                        "organization": organization.slug,
+                        "integration_id": integration.id,
+                    },
+                )
+                break
+            user_list += next_users["members"]
+
+            next_cursor = next_users["response_metadata"]["next_cursor"]
+            if not next_cursor:
+                break
 
     return user_list
 
--- a/src/sentry/integrations/slack/utils/users.py
+++ b/src/sentry/integrations/slack/utils/users.py
@@ -45,7 +45,7 @@ def get_users(integration: Integration, organization: Organization) -> Sequence[
         for _ in range(SLACK_GET_USERS_PAGE_LIMIT):
             try:
                 user_list += next_users.get("members")
-                next_users = next(users)
+                next_users = next(next_users)
             except StopIteration:
                 break
             except SlackApiError:
--- a/src/sentry/integrations/slack/utils/users.py
+++ b/src/sentry/integrations/slack/utils/users.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from collections.abc import Generator, Mapping, MutableMapping, Sequence
+from collections.abc import Iterable, Mapping, MutableMapping, Sequence
 from dataclasses import dataclass
 from typing import Any
 
@@ -19,6 +19,13 @@ from ..utils import logger
 
 SLACK_GET_USERS_PAGE_LIMIT = 100
 SLACK_GET_USERS_PAGE_SIZE = 200
+
+
+@dataclass(frozen=True)
+class SlackUserData:
+    email: str
+    team_id: str
+    slack_id: str
 
 
 def get_users(integration: Integration, organization: Organization) -> Sequence[Mapping[str, Any]]:
@@ -54,13 +61,11 @@ def get_users(integration: Integration, organization: Organization) -> Sequence[
 
 def get_slack_info_by_email(
     integration: Integration, organization: Organization
-) -> Mapping[str, Mapping[str, str]]:
+) -> Mapping[str, SlackUserData]:
     return {
-        member["profile"]["email"]: {
-            "email": member["profile"]["email"],
-            "team_id": member["team_id"],
-            "slack_id": member["id"],
-        }
+        member["profile"]["email"]: SlackUserData(
+            email=member["profile"]["email"], team_id=member["team_id"], slack_id=member["id"]
+        )
         for member in get_users(integration, organization)
         if not member["deleted"] and member["profile"].get("email")
     }
@@ -70,9 +75,9 @@ def get_slack_data_by_user(
     integration: Integration | RpcIntegration,
     organization: Organization | RpcOrganization,
     emails_by_user: Mapping[User, Sequence[str]],
-) -> Mapping[User, Mapping[str, str]]:
+) -> Mapping[User, SlackUserData]:
     slack_info_by_email = get_slack_info_by_email(integration, organization)
-    slack_data_by_user: MutableMapping[User, Mapping[str, str]] = {}
+    slack_data_by_user: MutableMapping[User, SlackUserData] = {}
     for user, emails in emails_by_user.items():
         for email in emails:
             if email in slack_info_by_email:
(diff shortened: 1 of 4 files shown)


src/sentry/integrations/slack/utils/users.py
```diff
@@ -18,31 +22,58 @@
...
+                "slack.post_install.get_users.error",
                 extra={
                     "error": str(e),
                     "organization": organization.slug,
                     "integration_id": integration.id,
                 },
             )
-            break
-        user_list += next_users["members"]
+            return []
```
cathteng, line 42: this function is called inside an async task to link slack user identities inside integration post_install. we could error loudly and have retries on the task?

src/sentry/integrations/slack/utils/users.py
```diff
@@ -18,31 +22,58 @@
...
-            break
+        next_users = iter(users)
+        for _ in range(SLACK_GET_USERS_PAGE_LIMIT):
+            try:
+                user_list += next_users.get("members")
+                next_users = next(users)
+            except StopIteration:
+                break
+            except SlackApiError:
+                break
```
cathteng, line 52: similarly we could error loudly here 🤔

src/sentry/integrations/slack/utils/users.py
```diff
@@ -18,31 +22,58 @@
...
+            return []
 
-        next_cursor = next_users["response_metadata"]["next_cursor"]
-        if not next_cursor:
-            break
+        next_users = iter(users)
+        for _ in range(SLACK_GET_USERS_PAGE_LIMIT):
+            try:
+                user_list += next_users.get("members")
+                next_users = next(users)
```
ykamo001, line 48: shouldn't this be `next_users = next(next_users)`?

cathteng: oh oops yeah

src/sentry/integrations/slack/utils/users.py
ykamo001: Hmmm, would it be better to return `None` and update the function contract? Might also be good to just reraise and let the caller decide what it wants to do instead of implicitly forcing it

src/sentry/integrations/slack/utils/users.py
```diff
@@ -18,31 +22,58 @@
...
             )
-            break
-        user_list += next_users["members"]
+            return []
 
-        next_cursor = next_users["response_metadata"]["next_cursor"]
-        if not next_cursor:
-            break
+        next_users = iter(users)
+        for _ in range(SLACK_GET_USERS_PAGE_LIMIT):
```
ykamo001, line 45: once we reach the limit, we just end silently? Is that the best corse of action? I think this flow needs to be updated, why not make this return a generator and yield each batch?

ykamo001: I know you're just trying to keep the functionality the same as before, but this is a limitation we should address

ykamo001 (approved):
LGTM, we should fix the pagination limit flow though

src/sentry/integrations/slack/utils/users.py
cathteng: should we have the generator and iterate through until an error happens or we finish?

cathteng: i added a generator function for this

src/sentry/integrations/slack/utils/users.py
```diff
@@ -74,3 +78,55 @@ def get_slack_data_by_user(
...
+# USING SDK CLIENT
+
+
+def format_slack_data_by_user(emails_by_user, slack_info_by_email):
+    slack_data_by_user: MutableMapping[User, Mapping[str, str]] = {}
+    for user, emails in emails_by_user.items():
+        for email in emails:
+            if email in slack_info_by_email:
+                slack_data_by_user[user] = slack_info_by_email[email]
+                break
```
ykamo001, line 92: This is a good example of a pattern you can look at and know using functions would be better. Instead of a break, using a return is cleaner and easier to read, and easier to extend as well for the future

src/sentry/integrations/slack/utils/users.py
```diff
@@ -74,3 +78,55 @@ def get_slack_data_by_user(
                 slack_data_by_user[user] = slack_info_by_email[email]
                 break
     return slack_data_by_user
+
+
+# USING SDK CLIENT
+
+
+def format_slack_data_by_user(emails_by_user, slack_info_by_email):
```
ykamo001, line 86: nit: type hints

src/sentry/integrations/slack/utils/users.py
```diff
@@ -74,3 +78,55 @@ def get_slack_data_by_user(
...
+    slack_data_by_user: MutableMapping[User, Mapping[str, str]] = {}
+    for user, emails in emails_by_user.items():
+        for email in emails:
+            if email in slack_info_by_email:
+                slack_data_by_user[user] = slack_info_by_email[email]
+                break
+    return slack_data_by_user
+
+
+def format_slack_info_by_email(users: dict[str, Any]):
```
ykamo001, line 96: nit: return type hint

ykamo001, line 96: can we use a dataclass instead of dictionaries please?

src/sentry/integrations/slack/utils/users.py
```diff
@@ -74,3 +78,55 @@ def get_slack_data_by_user(
...
+    sdk_client = SlackSdkClient(integration_id=integration.id)
+    try:
+        users_list = sdk_client.users_list(limit=SLACK_GET_USERS_PAGE_SIZE)
+
+        for page in users_list:
+            users: dict[str, Any] = page.get("members")
+            slack_info_by_email = format_slack_info_by_email(users)
+            slack_data_by_user = format_slack_data_by_user(emails_by_user, slack_info_by_email)
+
+            yield slack_data_by_user
```
ykamo001, line 122: amazing, thank you

src/sentry/integrations/slack/utils/users.py
```diff
@@ -74,3 +78,55 @@ def get_slack_data_by_user(
...
+        for member in users
+        if not member["deleted"] and member["profile"].get("email")
+    }
+
+
+def get_slack_data_by_user_via_sdk(
+    integration: Integration | RpcIntegration,
+    organization: Organization | RpcOrganization,
+    emails_by_user: Mapping[User, Sequence[str]],
+) -> Generator[Any, Any, Any] | None:
```
ykamo001, line 112: can we have a stricter contract for the data object?

src/sentry/integrations/slack/utils/users.py
cathteng: the break is to break out of the inner for loop?

src/sentry/integrations/slack/utils/users.py
cathteng: done

src/sentry/integrations/slack/utils/users.py
cathteng: sure

src/sentry/integrations/slack/utils/users.py
```diff
@@ -74,3 +79,63 @@ def get_slack_data_by_user(
                 slack_data_by_user[user] = slack_info_by_email[email]
                 break
     return slack_data_by_user
+
+
+# USING SDK CLIENT
+
+
+@dataclass
```
RyanSkonnord, line 87: I recommend always writing `@dataclass(frozen=True)` for a new dataclass that doesn't need to be mutated after creation.

src/sentry/integrations/slack/utils/users.py
```diff
@@ -74,3 +79,63 @@ def get_slack_data_by_user(
...
+        logger.info(
+            "slack.post_install.get_users.error",
+            extra={
+                "error": str(e),
+                "organization": organization.slug,
+                "integration_id": integration.id,
+            },
+        )
+
+    return None
```
RyanSkonnord, line 141: I don't think this line is necessary. Reaching the end of the function should just cause the generator to stop yielding. Typically you would use `return` in a generator function if you want to halt iteration early.

src/sentry/integrations/slack/utils/users.py
```diff
@@ -74,3 +79,63 @@ def get_slack_data_by_user(
...
+        for member in users
+        if not member["deleted"] and member["profile"].get("email")
+    }
+
+
+def get_slack_data_by_user_via_sdk(
+    integration: Integration | RpcIntegration,
+    organization: Organization | RpcOrganization,
+    emails_by_user: Mapping[User, Sequence[str]],
+) -> Generator[Mapping[User, Mapping[str, str]], None, None] | None:
```
RyanSkonnord, line 120: The `| None` should be unnecessary unless I'm missing something.

Also, `Generator[Mapping[User, Mapping[str, str]], None, None]` can be simplified to `Iterator[Mapping[User, Mapping[str, str]]]`.

src/sentry/integrations/slack/utils/users.py
```diff
@@ -74,3 +79,63 @@ def get_slack_data_by_user(
...
+    sdk_client = SlackSdkClient(integration_id=integration.id)
+    try:
+        users_list = sdk_client.users_list(limit=SLACK_GET_USERS_PAGE_SIZE)
+
+        for page in users_list:
+            users: dict[str, Any] = page.get("members")
+            slack_info_by_email = format_slack_info_by_email(users)
+            slack_data_by_user = format_slack_data_by_user(emails_by_user, slack_info_by_email)
+
+            yield slack_data_by_user
```
RyanSkonnord, line 130: This seems to be inconsistent with the type hints? `format_slack_data_by_user` returns `Mapping[User, SlackUserData]` but here we're supposed to be yielding `Mapping[User, Mapping[str, str]]`.

src/sentry/integrations/slack/utils/users.py
```diff
@@ -74,3 +79,63 @@ def get_slack_data_by_user(
...
+    except SlackApiError as e:
```
RyanSkonnord, line 131: Can we narrow the scope of the `try` block to just the `sdk_client.users_list` statement?

src/sentry/tasks/integrations/slack/link_slack_user_identities.py
```diff
@@ -66,3 +87,38 @@ def link_slack_user_identities(
...
+            logger.info(
+                "post_install.identity_linked_different_user",
+                extra={
+                    "idp_id": idp.id,
+                    "external_id": slack_id,
+                    "object_id": matched_identity.id,
+                    "user_id": user.id,
+                    "type": idp.type,
+                },
+            )
```
RyanSkonnord, line 124: This is a lot of duplicated code from `update_identities`. It would be better to adapt the existing function to work with both cases.

There are a few ways we could go about that. My first choice would be to rewrite the existing function to act on `SlackUserData` objects, and convert the `Mapping[str, str]` values from old function calls into the new dataclass. If that won't work for some reason, we could also refactor it into a very small abstract class.

Let me know if you'd like to pair on that.

src/sentry/integrations/slack/utils/users.py
cathteng: oh yeah, missed that here, thanks!

src/sentry/integrations/slack/utils/users.py
cathteng: i had a typing issue earlier where it complained i wasn't returning anything in the function, removing the `| None` from the return type solved it

src/sentry/tasks/integrations/slack/link_slack_user_identities.py
cathteng: it's copy-pasted, using the new dataclass for the old logic sounds good to me

src/sentry/integrations/slack/utils/users.py
cathteng: when i'm iterating through the results of `sdk_client.users_list` it will paginate through the API (the `SlackResponse` object has a `__next__` function that does this). i think it's possible that `SlackApiError` is raised while iterating through the pages, am I interpreting that correctly?

https://github.com/slackapi/python-slack-sdk/blob/v3.27.1/slack_sdk/web/slack_response.py#L118

RyanSkonnord (approved):
Nice!

PR #72879, opened 2024-06-17
chore(hybrid-cloud): Adds option-based overrides for RPC retry count and timeout duration

GabeVillalobos:
Adds the ability to override retry count and request timeout using Sentry options. Some of our less idempotent service requests shouldn't be using retries by default, as additional requests will have side-effects.

To combat this, adding the ability to default a specific method of a specific service to use its own values seems reasonable.

This unfortunately means we'll still have to define a default option in the Sentry repository, prior to using it which is less than ideal.

**Retry Count Option:**
Expected to be a dictionary containing keys in the format of `service_name:method_name`, with int values.
`hybridcloud.rpc.method_retry_overrides`

e.g.
```python
{
    "user_service:get_user_by_id": 2
}
```

**Timeout Option:**
Expected to be a dictionary containing keys in the format of `service_name:method_name`, with float values.
`hybridcloud.rpc.method_timeout_overrides`

e.g.
```python
{
    "user_service:get_user_by_id": 15.0
}
```


markstory (approved):
Looks good to me. We'll be able to tune timeouts/retries more easily with options vs trying to wedge these config values into the method definitions.

PR #72880, opened 2024-06-17
chore(trace-explorer): Add disabled project selector

Zylphrex:
Make it clear how to select projects in the trace explorer.

### DIFF
--- a/static/app/components/organizations/projectPageFilter/index.tsx
+++ b/static/app/components/organizations/projectPageFilter/index.tsx
@@ -55,8 +55,15 @@ export interface ProjectPageFilterProps
    */
   footerMessage?: React.ReactNode;
   /**
-   * Override the selected projects that are displayed.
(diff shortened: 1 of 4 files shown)


static/app/components/organizations/projectPageFilter/index.tsx
```diff
@@ -54,6 +54,11 @@ export interface ProjectPageFilterProps
    * Message to show in the menu footer
    */
   footerMessage?: React.ReactNode;
+  /**
+   * Override the selected projects that are displayed.
+   * Does NOT override the projects in the store.
+   */
+  projectOverride?: number[];
```
k-fish, line 61: I think you should give a bit more context in this comment, that this is only meant to show a disabled state for all projects not that we're allowing an override when a user can actually use it

k-fish (approved):

PR #72884, opened 2024-06-17
fix(trace) account for head or tail end of icons

JonasBa:
Span bars sometimes have icons at the head or tail end, which means they can overlap with the duration bars and cause visual issues. This PR computes the span bar space according to the min and max timestamps so that the duration labels no longer overlap.

Before:

https://github.com/getsentry/sentry/assets/9317857/d8ff6efc-7145-49ff-a7e1-4ab6a8c21f0a

After:

https://github.com/getsentry/sentry/assets/9317857/fd23c752-e5b3-4623-885f-e51460074a75

### DIFF
--- a/static/app/views/performance/newTraceDetails/traceRenderers/virtualizedViewManager.tsx
+++ b/static/app/views/performance/newTraceDetails/traceRenderers/virtualizedViewManager.tsx
@@ -1685,8 +1685,8 @@ function getIconTimestamps(
 ) {
   let min_icon_timestamp = span_space[0];
   let max_icon_timestamp = span_space[0] + span_space[1];
-  
-    if (!node.errors.size && !node.performance_issues.size) {
+
+  if (!node.errors.size && !node.performance_issues.size) {
     return [min_icon_timestamp, max_icon_timestamp];
   }
 
--- a/static/app/views/performance/newTraceDetails/traceRenderers/virtualizedViewManager.tsx
+++ b/static/app/views/performance/newTraceDetails/traceRenderers/virtualizedViewManager.tsx
@@ -1173,23 +1173,25 @@ export class VirtualizedViewManager {
     const TEXT_PADDING = 2;
 
     const icon_width_config_space = (18 * this.span_to_px[0]) / 2;
-    const text_left = span_space[0] > this.to_origin + this.trace_space.width * 0.8;
+    const text_anchor_left =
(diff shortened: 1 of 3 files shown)


static/app/views/performance/newTraceDetails/traceRenderers/virtualizedViewManager.tsx
```diff
@@ -1170,57 +1170,28 @@ export class VirtualizedViewManager {
...
-
-    // |   text|
+    const right_outside = this.computeTransformXFromTimestamp(text_right) + TEXT_PADDING;
+    // |---text|
     const right_inside =
-      this.computeTransformXFromTimestamp(span_space[0] + span_space[1]) -
-      width -
-      TEXT_PADDING;
-    // |text   |
+      this.computeTransformXFromTimestamp(span_space[0] + span_space[1]) - TEXT_PADDING;
```
Zylphrex, line 1189: Did this need to subtract `width` as well?

static/app/views/performance/newTraceDetails/traceRenderers/virtualizedViewManager.tsx
```diff
@@ -1704,6 +1675,50 @@ export class VirtualizedViewManager {
...
+  node: TraceTreeNode<any>,
+  span_space: [number, number],
+  icon_width: number
+) {
+  if (!node.errors.size && !node.performance_issues.size) {
+    return [span_space[0], span_space[0] + span_space[1]];
+  }
+
+  let min_icon_timestamp = span_space[0];
+  let max_icon_timestamp = span_space[0] + span_space[1];
```
Zylphrex, line 1691: Nit: Just makes what you're returning a little more obvious

```suggestion
  let min_icon_timestamp = span_space[0];
  let max_icon_timestamp = span_space[0] + span_space[1];
  
    if (!node.errors.size && !node.performance_issues.size) {
    return [min_icon_timestamp, max_icon_timestamp];
  }
```

Zylphrex (approved):
There's a mix of snake case and camel case here that you may want to clean up at some point.

static/app/views/performance/newTraceDetails/traceRenderers/virtualizedViewManager.tsx
JonasBa: Yup, and the padding needs to be negative here too.

PR #72895, opened 2024-06-17
ref: fix types in sentry.rules.processing

asottile-sentry:


saponifi3d (approved):
🙏 thanks for this cleanup!

PR #72897, opened 2024-06-17
ref: fix types for EventUser

asottile-sentry:

### DIFF
--- a/src/sentry/utils/eventuser.py
+++ b/src/sentry/utils/eventuser.py
@@ -9,6 +9,7 @@ from functools import cached_property
 from ipaddress import IPv4Address, IPv6Address, ip_address
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


src/sentry/utils/eventuser.py
```diff
@@ -117,7 +118,7 @@ def get_display_name(self):
     @classmethod
     def for_projects(
         self,
-        projects: list[Project],
+        projects: QuerySet[Project] | list[Project],
```
asottile-sentry, line 121: this is kind of a gross union -- but tests call this with a list so it'd be annoying to fix those (it can't be `Sequence` because a QuerySet isn't one, it can't be `Iterable` because accesses are done using indices!)

IanWoodard (approved):

PR #72902, opened 2024-06-17
chore(hybridcloud) Merge customer-domains feature into multi-region

markstory:
To reduce the number of features flags and the combinations of feature flags, I am merging the `customer-domains` feature into `system:multi-region`.

I've started to cleanup some of the logic around customer-domains behavior as we'll no longer have ways to go in and out of customer domains for specific organizations. There is more to do here and I'll do the remaining cleanup once customer-domains feature can be removed.

These changes also include moving away from `window.__initialData` being used across the application. Instead we can use `ConfigStore` to read application configuration.

Refs #72537


static/app/stores/configStore.spec.tsx
```diff
@@ -19,12 +29,7 @@ describe('ConfigStore', () => {
   });
 
   it('should have customerDomain', () => {
-    const customerDomain = ConfigStore.get('customerDomain');
-    expect(customerDomain).toEqual({
-      subdomain: 'foobar',
-      organizationUrl: 'https://foobar.sentry.io',
-      sentryUrl: 'https://sentry.io',
-    });
+    expect(ConfigStore.get('customerDomain')).toEqual(null);
```
markstory, line 32: This is kind of a low value test as it mostly validates the default data loaded into `ConfigStore`.

static/app/components/sidebar/sidebarDropdown/switchOrganization.spec.tsx
```diff
@@ -35,42 +47,10 @@ describe('SwitchOrganization', function () {
     jest.useRealTimers();
   });
 
-  it('uses organizationUrl when customer domain is enabled', async function () {
```
markstory, line 38: I removed this test because the `multi-region` flag cannot be turned off at the organization level like `customer-domain` could be.

scttcper (approved):

PR #72905, opened 2024-06-17
fix(insights): updated empty-state copy for caches / queues

bcoe:
Better describe the benefits of using the cache / queue insights modules:

Screenshot 2024-06-18 at 2 44 34 PM
Screenshot 2024-06-18 at 2 44 27 PM


DominikB2014 (approved):
Nice! This is a lot more descriptive

static/app/views/performance/onboarding/modulesOnboarding.tsx
```diff
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled';
 
-import emptyStateImg from 'sentry-images/spot/performance-waiting-for-span.svg';
```
DominikB2014, line 3: We should delete the old image if its no longer being used

bcoe: Done.

PR #72906, opened 2024-06-17
fix(ui) Fix broken redirect on org rename

markstory:
When an organization changes their slug we should redirect to the new slug. This isn't working currently because of changes to OrganizationDetails that elide `features` by default. (#71791)

As customer-domains will soon be retried and replaced with a system configuration feature flag, I've chosen to adapt the logic for org details settings to read from `ConfigStore` like it will in the future.

Fixes HC-1096


iamrajjoshi (approved):
thanks for this fix! 🙏🏽

scttcper (approved):

PR #72909, opened 2024-06-17
ref(replay): use placeholder for loading in tabs

michellewzhang:
relates to https://github.com/getsentry/sentry/issues/68109

adds loading placeholders for the tabs focus area, since some tabs go away for mobile replays.

https://github.com/getsentry/sentry/assets/56095982/1387add2-cf98-46d0-9f01-173e4a0471bf

### DIFF
--- a/static/app/views/replays/detail/layout/index.tsx
+++ b/static/app/views/replays/detail/layout/index.tsx
@@ -2,6 +2,7 @@ import {useRef} from 'react';
 import styled from '@emotion/styled';
 
 import ErrorBoundary from 'sentry/components/errorBoundary';
+import Placeholder from 'sentry/components/placeholder';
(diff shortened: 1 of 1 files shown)


ryan953:
lets just block out the whole area.

for example: if the page is loaded with `?t_main=a11y` then we'll see the search bar for accessibility, but that tab is invalid for mobile, so it'll cause jank when the replay loads in. Better for maintenance to have it be a big loading space imo

michellewzhang:
> lets just block out the whole area.
> 
> for example: if the page is loaded with `?t_main=a11y` then we'll see the search bar for accessibility, but that tab is invalid for mobile, so it'll cause jank when the replay loads in. Better for maintenance to have it be a big loading space imo

done!

ryan953 (approved):

PR #72919, opened 2024-06-18
feat(uptime): Introduce `remote_subscriptions` module and model

wedamija:
This introduces the `remote_subscriptions` module. The purpose of this module is to make a generic version of `QuerySusbcription` that can be used with other similar remote systems that send results down on a periodic schedule.

We should be able to re-use a lot of code from query subscription around creating/updating/deleting subscriptions, checking that the subscriptions were created successfully, and so on.

The goal is to integrate this with uptime checkers first, and then attempt to migrate `QuerySubscription` data over if the alerts team agrees with this model.


src/sentry/migrations/0732_remote_subscription.py
```diff
@@ -0,0 +1,49 @@
...
+    # - Adding indexes to large tables. Since this can take a long time, we'd generally prefer to
+    #   run this outside deployments so that we don't block them. Note that while adding an index
+    #   is a schema change, it's completely safe to run the operation after the code has deployed.
+    # Once deployed, run these manually via: https://develop.sentry.dev/database-migrations/#migration-deployment
+
+    is_post_deployment = False
+
+    dependencies = [
+        ("sentry", "0731_add_insight_project_flags"),
+    ]
```
markstory, line 27: If you're making a new module, that module could contain models/migrations so that you don't have to compete for the migration head.

src/sentry/remote_subscriptions/models.py
```diff
@@ -0,0 +1,36 @@
...
+from django.db import models
+
+from sentry.backup.scopes import RelocationScope
+from sentry.db.models import BaseManager, DefaultFieldsModel, region_silo_model
+
+
+@region_silo_model
+class RemoteSubscription(DefaultFieldsModel):
+    # TODO: This should be included in export/import, but right now it has no relation to
+    # any projects/orgs. Will fix this in a later pr
```
markstory, line 14: How will the project/org relation work for this model?

evanpurkhiser (approved):

src/sentry/remote_subscriptions/models.py
wedamija: Via whichever service implements a more specific subscription. So for uptime, someone might make a subscription to monitor http://sentry.io. So they'll create a
`ProjectUptimeSubscription` - This ties an UptimeSubscription to a specific project
`UptimeSubscription` - this is generic, since we want to debounce monitors for the same url. So if multiple projects across orgs make the same monitor, they'll share this
`RemoteSubscription`

We'll need some kind of cleanup process that allows orphaned rows to be properly removed.

PR #72920, opened 2024-06-18
feat(replay): Cross link from inbound-filters to replay settings pages

### LINKED ISSUE #72180: [Hydration Errors] Cross link from inbound-filters to replay-settings
When, or even whether, https://www.notion.so/sentry/React-JS-19-23a7a2c5f88d4ef59a4b9647c84333a3 is implemented we should have links between the inbound-filters settings page, and the replays-settings page. This should explain the impact of enabling each toggle, in terms of quota, what to expect on the issue stream, and what to expect on the issue-details page itself.

For example:
When that notion docs is implemented: users could disable their inbound filter and start consuming error-quota and get hydration-errors. These would include useful stack-traces. If the React19 user also has replays installed, then they'd get replays attached.

ryan953:
New copy with links:

| Page | Copy |
| --- | --- |
| Replay Settings | SCR-20240617-qfvb
| Inbound Filters | SCR-20240617-qfvw

Fixes https://github.com/getsentry/sentry/issues/72180


static/app/views/settings/project/projectFilters/projectFiltersSettings.tsx
```diff
@@ -537,9 +538,22 @@ export function ProjectFiltersSettings({project, params, features}: Props) {
                       type: 'boolean',
                       name: 'filters:react-hydration-errors',
                       label: t('Filter out hydration errors'),
-                      help: t(
-                        'React falls back to do a full re-render on a page and these errors are often not actionable.'
-                      ),
+                      help: organization.features.includes(
```
michellewzhang, line 541: if they're not under the feature flag, should we add a comment explaining that they don't have access or that sort of thing?

michellewzhang (approved):

static/app/views/settings/project/projectFilters/projectFiltersSettings.tsx
ryan953: i think if they're not under the feature flag they don't need to know about the replay-hydration settings page. we're just testing it basically, so if it all went away it could be confusing to folks. they'll get it when it's released!

If they don't have replay at all though.. then that's a good question. We could link to the page and have the whole page be disabled or something like that.

PR #72933, opened 2024-06-18
fix(trace_timeline): Support for trace timeline and related issues (team/free plans)

armenzg:
If we fetch the events endpoint without specifying a project it queries across all projects. Organizations without the `global-views` feature (e.g. free and team plans) would fail to return any events, thus, the trace timeline would not be displayed.

This changes the `useTraceTimelineEvents` to always include the `project` query parameter by either setting it to `-1` (which queries all projects) or to the current project (e.g. the organization is on the free plan).

This will enable the trace timeline and related issues for free plans for events/issues within the same project.

If a trace contains events for other projects, we will only show the link to the trace but not the trace timeline or related issues.

### DIFF
--- a/static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
+++ b/static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
@@ -249,7 +249,7 @@ describe('TraceTimeline', () => {
     });
   });
 
-  it('works for free plans (no global-views feature)', async () => {
+  it('works for plans with no global-views feature', async () => {
     MockApiClient.addMockResponse({
       url: `/organizations/${organization.slug}/events/`,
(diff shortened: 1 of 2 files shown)


static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -244,4 +248,36 @@ describe('TraceTimeline', () => {
       trace_timeline_status: 'empty',
     });
   });
+
+  it('works for free plans (no global-views feature)', async () => {
```
armenzg, line 252:
```suggestion
  it('works for plans with no global-views feature', async () => {
```

scttcper (approved):

PR #72934, opened 2024-06-18
ref(metric-widget): Prevent users leaving route if unsaved changes

### LINKED ISSUE #72846: Dashboard: metric widget edit button behind experimental feature flag
Image

- remove experimental feature flag check
- change icon to full screen
- close modal button should discard changes, but ask if you're sure before that

priscilawebdev:
### Goal

Display a warning message about unsaved changes before users close the metrics widget modal. If they confirm, the changes are discarded. If they stay, they can continue editing.

### Challenges

Initially, this task seemed straightforward since we could have used the `setRouteLeaveHook` function, as we do in the [widget builder](https://github.com/getsentry/sentry/blob/4e071ead0bcf491ffbf8f34fa60d2f2071d5896a/static/app/views/dashboards/widgetBuilder/widgetBuilder.tsx#L357-L366). However, this is a modal, and route updates do not always occur, such as when a new metrics widget is being created. Making this work with route changes would require more time investment.

Another potential solution was to use `browserHistory.listen`, but again, we are using a modal to handle metrics widgets, complicating this approach.

### Solution

For now, we have updated the global modal component's `CloseButton` to allow the use of a custom click function. Now, if a user clicks the X button at the top of a metrics widget and has unsaved changes, a native browser message will be displayed, asking if they are sure they want to leave without saving.

### Notes

- Since it was also possible to close the modal using the **Esc** key, and adding this change specifically for the metrics widget would be a bit hacky, I have updated the code to disable closing the modal using the keyboard.

- Applying this change only to metrics widgets could confuse users, so it has been disabled for all widgets.

- The ideal solution would be to verify route changes, as users can still navigate elsewhere (e.g., clicking the 'Open in Traces' or profiling button) with unsaved changes - even with the changes presented in this PR (Let's discuss this).

### Preview

https://github.com/getsentry/sentry/assets/29228205/67585029-8885-4c21-9280-d90e8f158bef

_____

closes https://github.com/getsentry/sentry/issues/72846


static/app/actionCreators/modal.tsx
```diff
@@ -355,7 +355,7 @@ export async function openWidgetViewerModal({
   const {default: Modal, modalCss} = mod;
 
   openModal(deps => <Modal {...deps} {...options} />, {
-    closeEvents: 'escape-key',
+    closeEvents: 'none',
```
evanpurkhiser, line 358: Does this disable escape to close for all modals?

evanpurkhiser: Never mind I see, just for this modal

static/app/components/modals/metricWidgetViewerModal.tsx
```diff
@@ -224,6 +237,25 @@ function MetricWidgetViewerModal({
...
+        // eslint-disable-next-line no-alert
+        window.confirm(t('You have unsaved changes, are you sure you want to close?'))
+      ) {
+        closeModal();
+      }
+      return;
+    }
+
+    closeModal();
+  }, [userHasModified, closeModal]);
```
evanpurkhiser, line 257:
```suggestion
  const handleClose = useCallback(() => {
    if (
        userHasModified &&
        // eslint-disable-next-line no-alert
        !window.confirm(t('You have unsaved changes, are you sure you want to close?')
    ) {
      return;
    }
    closeModal();
  }, [userHasModified, closeModal]);
```

static/app/views/dashboards/widgetCard/widgetCardContextMenu.tsx
```diff
@@ -92,14 +92,6 @@ function WidgetCardContextMenu({
     }
   };
 
-  const openWidgetViewerIcon =
-    hasMetricsExperimentalFeature(organization) &&
-    widget.widgetType === WidgetType.METRICS ? (
-      <IconEdit />
-    ) : (
-      <IconExpand />
-    );
```
evanpurkhiser, line 101: unrelated changes?

priscilawebdev: oh no, it is part of the [ticket](https://github.com/getsentry/sentry/issues/72846) . It is a very small change though and if you have strong feelings, I can create a new PR for it.

ArthurKnaus (approved):

PR #72940, opened 2024-06-18
feat(query-builder): Add loading state and improved menu styles/positioning

malwilley:
- Add loading state when tag values are being fetched (and there aren't any items to show)
- Decrease menu text size and increase max width to better fit suggestions
- Recalculate popperjs positioning when menu size might change (because of loading state and width changes, it would get out of place at times)

https://github.com/getsentry/sentry/assets/10888943/a2f65098-dd64-4621-9e37-19c8d23817bb

### DIFF
--- a/static/app/components/searchQueryBuilder/combobox.tsx
+++ b/static/app/components/searchQueryBuilder/combobox.tsx
@@ -4,12 +4,11 @@ import {
   type MouseEventHandler,
   type ReactNode,
   useCallback,
+  useLayoutEffect,
   useMemo,
   useRef,
   useState,
 } from 'react';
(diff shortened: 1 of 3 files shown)


MichaelSun48 (approved):
Looks great! I just noticed a bug (?) the menu never shows up for the  `project_slug` after the loading state disappears. Not sure if this is because there are just 0 options or a different bug, but if its the former, then I think an empty state would be great. 

https://github.com/getsentry/sentry/assets/55160142/77b65857-6408-4a49-a77d-583183d6b53c

malwilley:
@MichaelSun48 good callout, an empty state would make a lot of sense for those tags. Just need to figure out how to show it in a non-intrusive way.

PR #72941, opened 2024-06-18
fix(spans): Use span type for traces searches

Zylphrex:
This was using the event type so searches from other parts of the product like discover were being mixed here. Switch to use it's own type.

Depends on #72938


k-fish (approved):
You can prob base the branch off the other one so it doesn't go in accidentally ahead of time and updates when that merges

PR #72947, opened 2024-06-18
ref(slack): Refactor Unlink Team View

iamrajjoshi:

Addresses `bot-4` task.

I created a dataclass for the input params, added logging & metrics, and instead of raising errors, we instead render the error page

### DIFF
--- a/src/sentry/integrations/slack/views/types.py
+++ b/src/sentry/integrations/slack/views/types.py
@@ -21,3 +21,23 @@ class IdentityParams:
         self.slack_id = slack_id
         self.channel_id = channel_id
         self.response_url = response_url
+
+
+@dataclass
+class TeamIdentityParams:
+    organization_id: str
+    integration_id: str
+    channel_id: str
+    channel_name: str
+    slack_id: str
+    response_url: str | None = None
+
+    def __init__(
+        self, organization_id, integration_id, channel_id, channel_name, slack_id, response_url=None
+    ):
+        self.organization_id = organization_id
+        self.integration_id = integration_id
+        self.channel_id = channel_id
+        self.channel_name = channel_name
+        self.slack_id = slack_id
+        self.response_url = response_url
--- a/src/sentry/integrations/slack/views/types.py
+++ b/src/sentry/integrations/slack/views/types.py
@@ -14,30 +14,12 @@ class IdentityParams:
     channel_id: str
     response_url: str | None = None
 
-    def __init__(self, organization, integration, idp, slack_id, channel_id, response_url=None):
-        self.organization = organization
-        self.integration = integration
-        self.idp = idp
-        self.slack_id = slack_id
-        self.channel_id = channel_id
-        self.response_url = response_url
-
 
 @dataclass
-class TeamIdentityParams:
+class TeamIdentityRequest:
     organization_id: str
     integration_id: str
     channel_id: str
     channel_name: str
     slack_id: str
     response_url: str | None = None
-
-    def __init__(
-        self, organization_id, integration_id, channel_id, channel_name, slack_id, response_url=None
-    ):
-        self.organization_id = organization_id
-        self.integration_id = integration_id
-        self.channel_id = channel_id
-        self.channel_name = channel_name
-        self.slack_id = slack_id
-        self.response_url = response_url
--- a/src/sentry/integrations/slack/views/types.py
+++ b/src/sentry/integrations/slack/views/types.py
@@ -5,7 +5,7 @@ from sentry.models.integrations.integration import Integration
 from sentry.services.hybrid_cloud.organization import RpcOrganization
 
 
-@dataclass
+@dataclass(frozen=True)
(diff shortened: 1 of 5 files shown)


src/sentry/integrations/slack/views/types.py
```diff
@@ -21,3 +21,23 @@ def __init__(self, organization, integration, idp, slack_id, channel_id, respons
...
+
+    def __init__(
+        self, organization_id, integration_id, channel_id, channel_name, slack_id, response_url=None
+    ):
+        self.organization_id = organization_id
+        self.integration_id = integration_id
+        self.channel_id = channel_id
+        self.channel_name = channel_name
+        self.slack_id = slack_id
+        self.response_url = response_url
```
ykamo001, line 43: you don't need to specify an init with a dataclass unless you're overriding it somehow

src/sentry/integrations/slack/views/types.py
```diff
@@ -21,3 +21,23 @@ def __init__(self, organization, integration, idp, slack_id, channel_id, respons
         self.slack_id = slack_id
         self.channel_id = channel_id
         self.response_url = response_url
+
+
+@dataclass
+class TeamIdentityParams:
```
ykamo001, line 27: can we name this to `TeamIdentityRequest`?

src/sentry/integrations/slack/views/unlink_team.py
```diff
@@ -51,32 +56,59 @@ class SlackUnlinkTeamView(BaseView):
     Django view for unlinking team from slack channel. Deletes from ExternalActor table.
     """
 
+    _METRICS_SUCCESS_KEY = "sentry.integrations.slack.unlink_team_view.success"
+    _METRICS_FAILURE_KEY = "sentry.integrations.slack.unlink_team_view.failure"
```
cathteng, line 60: fyi i'm collecting all slack metrics in `src/sentry/integrations/slack/metrics.py` if you want to put these metrics in there, it's easier to see all of them at once that way

src/sentry/integrations/slack/views/types.py
```diff
@@ -14,10 +14,12 @@ class IdentityParams:
...
 
-    def __init__(self, organization, integration, idp, slack_id, channel_id, response_url=None):
-        self.organization = organization
-        self.integration = integration
-        self.idp = idp
-        self.slack_id = slack_id
-        self.channel_id = channel_id
-        self.response_url = response_url
+
+@dataclass
```
RyanSkonnord, line 18: Can we make it `@dataclass(frozen=True)`?

src/sentry/integrations/slack/views/unlink_team.py
```diff
@@ -51,32 +56,59 @@ class SlackUnlinkTeamView(BaseView):
...
+            _logger.exception("dispatch.type_error")
+            metrics.incr(self._METRICS_FAILURE_KEY, sample_rate=1.0)
+            return render_error_page(request, status=400, body_text="HTTP 400: Invalid parameters")
+
+        logger_params = {
+            "organization_id": params.organization_id,
+            "integration_id": params.integration_id,
+            "channel_id": params.channel_id,
+            "channel_name": params.channel_name,
+            "slack_id": params.slack_id,
```
cathteng, line 87: you could do something like `asdict(params)` here instead of typing all of the fields in params out!

src/sentry/integrations/slack/views/unlink_team.py
```diff
@@ -86,21 +118,34 @@ def handle(self, request: Request, signed_params: str) -> HttpResponse:
             external_id=channel_id,
         )
         if len(external_teams) == 0:
+            _logger.info(
+                "no-team-found",
+                extra={logger_params},
```
cathteng, line 123: should this be just logger_params? i think it's already a dict

src/sentry/integrations/slack/views/unlink_team.py
```diff
@@ -86,21 +118,34 @@ def handle(self, request: Request, signed_params: str) -> HttpResponse:
...
+            )
+            metrics.incr(self._METRICS_FAILURE_KEY + ".get_team", sample_rate=1.0)
             return render_error_page(request, status=404, body_text="HTTP 404: Team not found")
 
         team = external_teams[0].team
         if team is None:
+            _logger.info(
+                "no-team-found",
+                extra={logger_params},
```
cathteng, line 132: same thing here

src/sentry/integrations/slack/views/unlink_team.py
iamrajjoshi: thanks for the heads up, ill update some of the other metrics i made as well

src/sentry/integrations/slack/views/unlink_team.py
RyanSkonnord: Yeah, I think this would make it a singleton set.

RyanSkonnord (approved):

PR #72954, opened 2024-06-18
chore(performance): Cache events_stats endpoint responses

colin-sentry:
We make a lot of requests for relatively similar data right now, this will let us split up the frontend graph loading into a few cached chunks.

### DIFF
--- a/src/sentry/api/endpoints/organization_events_stats.py
+++ b/src/sentry/api/endpoints/organization_events_stats.py
@@ -1,3 +1,4 @@
+import time
 from collections.abc import Mapping, Sequence
 from datetime import datetime, timedelta
 from typing import Any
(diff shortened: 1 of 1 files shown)


src/sentry/api/endpoints/organization_events_stats.py
```diff
@@ -453,20 +454,30 @@ def fn(
...
+            )
+            resp = Response(
+                response_data,
                 status=200,
             )
+            if (
+                response_data.get("end")
+                and request.GET.get("end")
+                and request.GET.get("project")  # only works if projects are explicitly specified
+                and time.time() - response_data.get("end") > 7200
```
wmak, line 477: Could we explicitly check this isn't a statsPeriod query for a small period?

wmak: oh nvm i misread this, its checking that the end of the query is >2hr ago right?

wmak (commented):
Thoughts on adding a metric to see how often these conditions are met first? I think this change is fine, I'm just worried that we'll have weird debugging edge cases

PR #72958, opened 2024-06-18
nit: Add project_id to delayed processing logs

schew2381:
Add extra logging to sort logs by `project_id` and `group_id`

### DIFF
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -231,6 +231,7 @@ def build_group_to_groupevent(
     bulk_event_id_to_events: dict[str, Event],
     bulk_occurrence_id_to_occurrence: dict[str, IssueOccurrence],
     group_id_to_group: dict[int, Group],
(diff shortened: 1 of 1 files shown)


src/sentry/rules/processing/delayed_processing.py
```diff
@@ -383,7 +397,11 @@ def apply_delayed(project_id: int, *args: Any, **kwargs: Any) -> None:
                 if status.last_active and status.last_active > freq_offset:
                     logger.info(
                         "delayed_processing.last_active",
-                        extra={"last_active": status.last_active, "freq_offset": freq_offset},
+                        extra={
+                            "last_active": status.last_active,
+                            "freq_offset": freq_offset,
+                            project_id: project_id,
```
saponifi3d, line 403:
```suggestion
                            "project_id": project_id,
```

saponifi3d (approved):
🪵- note to self, once we roll this out we should review these logs to see what we want to maintain longer term.

PR #72961, opened 2024-06-18
fix(related_issues): Orgs with open membership should always show related issues

armenzg:
A member may not belong to Project B, however, if their organization has enabled Open Membership they can then navigate any project issues.

The `events` endpoint should allow serving events for any project if Open Membership is enabled since the UI allows the member to view the issue from the projects they do not belong to.


src/sentry/api/bases/organization_events.py
```diff
@@ -118,8 +118,14 @@ def get_snuba_dataclass(
...
-
-                if not has_global_views and len(params.projects) > 1 and not fetching_replay_data:
+                if (
+                    not has_global_views
+                    and len(params.projects) > 1
+                    and not fetching_replay_data
+                    # If a developer can view issues of a project they do not belong to
+                    # via opem membership, we will also allow the endpoint to return events for it
+                    and not organization.flags.allow_joinleave
+                ):
```
vartec, line 128: nit: bit more readable?   

```suggestion
                if not any(
                    has_global_views,
                    len(params.projects) <= 1,
                    fetching_replay_data, 
                    # If a developer can view issues of a project they do not belong to
                    # via opem membership, we will also allow the endpoint to return events for it
                    organization.flags.allow_joinleave, 
                ):
```

vartec (approved):

src/sentry/api/bases/organization_events.py
```diff
@@ -118,8 +118,16 @@ def get_snuba_dataclass(
...
                 fetching_replay_data = request.headers.get("X-Sentry-Replay-Request") == "1"
-
-                if not has_global_views and len(params.projects) > 1 and not fetching_replay_data:
+                if not any(
+                    [
+                        has_global_views,
+                        len(params.projects) <= 1,
+                        fetching_replay_data,
+                        # If a developer can view issues of a project they do not belong to
+                        # via opem membership, we will also allow the endpoint to return events for it
```
armenzg, line 127:
```suggestion
                        # via open membership, we will also allow the endpoint to return events for it
```

PR #72967, opened 2024-06-18
ref: have models.User extend Model instead of BaseModel

asottile-sentry:
most of our db helpers expected .id to be a Field[int, int] and use Model as the type for that

no migration was generated so I believe this to be a fancy noop


markstory (approved):
No migration makes sense to me as there weren't any relevant changes to django model state or the schema.

PR #72981, opened 2024-06-18
chore(issue-priority): Remove issue-priority flag

snigdhas:
The feature has been GA'd for a few weeks - cleaning up the feature flag.


src/sentry/models/activity.py
```diff
@@ -42,17 +42,14 @@ def get_activities_for_group(self, group: Group, num: int) -> Sequence[Activity]
         activity_qs = self.filter(group=group).order_by("-datetime")
         initial_priority = None
```
wedamija, line 43: Nit: We don't ever use this value now without the conditional

wedamija (approved):

PR #72988, opened 2024-06-19
chore(slack): speed up wrapping functions in SDK client

cathteng:
We are getting some [TimeoutErrors](https://sentry.sentry.io/issues/?project=1&query=The+read+operation+timed+out&referrer=issue-list&statsPeriod=7d) from wrapping all the methods in the SDK client. From inspecting the stack trace it looks like we're calling `setattr` and wrapping every single method in every ancestor class of `SlackSdkClient` (`WebClient` and `BaseClient`). We could speed it up a lot by just wrapping the `api_call` function from `BaseClient`, which every single function in `WebClient` calls.

### DIFF
--- a/src/sentry/integrations/slack/sdk_client.py
+++ b/src/sentry/integrations/slack/sdk_client.py
@@ -18,7 +18,7 @@ SLACK_DATADOG_METRIC = "integrations.slack.http_response"
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


ykamo001 (approved):

RyanSkonnord (approved):
I'm fine with this in the short term. I'll follow up with some broader refactoring ideas.

PR #72991, opened 2024-06-19
ref(snql): Add arg count to exception message

philipphofmann:
Add the actual arg count to the raised exception of `validate_argument_count`.


narsaynorath (approved):
Looks good to me, the extra context is helpful

PR #72995, opened 2024-06-19
feat(insights): Add division_if for span metrics

philipphofmann:
Add division_if for span metrics so we can calculate rates such as slow and frozen rates.


tests/snuba/api/endpoints/test_organization_events_span_metrics.py
```diff
@@ -120,6 +120,53 @@ def test_count_if(self):
         assert data[0][fieldRelease2] == 1
         assert meta["dataset"] == "spansMetrics"
 
+    def test_division_if(self):
+        self.store_span_metric(
+            1,
+            internal_metric=constants.SPAN_METRICS_MAP["mobile.slow_frames"],
+            timestamp=self.three_days_ago,
+            tags={"release": "1.0.0"},
+        )
```
narsaynorath, line 129: These look like they'd pass but that's just because of how the data is being mocked and where the mock data is set up. There's another example of slow frames tests here where the first arg is actually passed as an object which is consistent for the gauge type, I'd suggest changing these to match that type to make sure it's working the way you want for these metrics, e.g.

```
        self.store_span_metric(
            {
                "min": 1,
                "max": 1,
                "sum": 1,
                "count": 1,
                "last": 1,
            },
            entity="metrics_gauges",
            metric="mobile.slow_frames",
            timestamp=self.six_min_ago,
            tags={"release": "foo"},
        )
```

Total frames might be the only one that needs to have different values for min/max/sum/count/last when entered

narsaynorath (approved):
Just a comment about the test setup, but I think the code looks good

src/sentry/search/events/datasets/spans_metrics.py
```diff
@@ -173,6 +173,30 @@ def function_converter(self) -> Mapping[str, fields.MetricsFunction]:
...
+                fields.MetricsFunction(
+                    "division_if",
+                    required_args=[
+                        fields.MetricArg(
+                            # needs to be named column, otherwise the query builder won't be able to determine the correct target table
+                            "column",
+                            allow_custom_measurements=False,
+                        ),
+                        fields.MetricArg(
+                            "column1",
```
narsaynorath, line 185: Can we give the `column1` a more descriptive name? Maybe like `divisorColumn` to match its role in the division.

narsaynorath (approved):
Looks good, just left a comment on naming. Thanks for finding that `column` is actually a special arg that we _need_ to define, we won't be able to change that one but we could do the second column

PR #73011, opened 2024-06-19
chore(hybridcloud) Remove the customer-domains feature

markstory:
With all checks no combined with multi-region the customer-domains feature flag is redundant.

Refs HC-1096
Refs getsentry/getsentry#14385

### DIFF
--- a/tests/sentry/web/frontend/test_home.py
+++ b/tests/sentry/web/frontend/test_home.py
@@ -76,7 +76,7 @@ class HomeTest(TestCase):
             assert response.redirect_chain == [
                 (f"http://{org.slug}.testserver/restore/", 302),
             ]
-            assert "activeorg" not in self.client.session
+            assert "activeorg" in self.client.session
 
     def test_customer_domain_org_deletion_in_progress(self):
         org = self.create_organization(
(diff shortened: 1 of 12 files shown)


tests/sentry/web/frontend/test_react_page.py
```diff
@@ -402,7 +395,7 @@ def test_customer_domain_org_deletion_in_progress(self):
             assert response.redirect_chain == [
                 ("http://testserver/organizations/new/", 302),
             ]
-            assert "activeorg" not in self.client.session
+            assert "activeorg" in self.client.session
```
markstory, line 398: This flip and the one on 379 are how the system was always working. However, the results were opposite before because `SENTRY_USE_CUSTOMER_DOMAINS` was `False`. Because of that, the logic that sets `activeorg` would not have run before.

tests/sentry/web/frontend/test_home.py
```diff
@@ -95,4 +95,4 @@ def test_customer_domain_org_deletion_in_progress(self):
             assert response.redirect_chain == [
                 ("http://testserver/organizations/new/", 302),
             ]
-            assert "activeorg" not in self.client.session
+            assert "activeorg" in self.client.session
```
GabeVillalobos, line 98: Is it ok that we're changing the behavior of this test now?

GabeVillalobos (approved):

PR #73013, opened 2024-06-19
chore(hybridcloud) Remove redundant signed link handling

markstory:
We don't need to handle signed link requests in multiple places. Instead we can handle that logic in the authentication adapter. I've combined AuthenticationMiddleware and its mixin as there is no other usage of that mixin in sentry or getsentry.

### DIFF
--- a/src/sentry/middleware/auth.py
+++ b/src/sentry/middleware/auth.py
@@ -16,7 +16,6 @@ from sentry.api.authentication import (
 )
 from sentry.models.userip import UserIP
 from sentry.utils.auth import AuthUserPasswordExpired, logger
-from sentry.utils.linksign import process_signature
 
 
 def get_user(request):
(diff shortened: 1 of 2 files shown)


src/sentry/middleware/auth.py
```diff
@@ -83,25 +88,3 @@ def process_exception(
...
-            return
-
-        # If there is a valid signature on the request we override the
-        # user with the user contained within the signature.
-        user = process_signature(request)
-
-        if user is not None:
-            request.user = user
-            request.user_from_signed_request = True
-            return
```
GabeVillalobos, line 105: We no longer need this signature override behavior?

markstory: Not in the middleware. We have the same logic applied to the two endpoints that need it via an Authentication adapter.

corps:
Was there just one call site?  Was for certain there were 2.

markstory:
> Was there just one call site? Was for certain there were 2.

@corps There is another still. When I fixed notification unsubscribes, I moved the other callsite into an Authentication adapter so that signed requests can work the same as other token based requests.

GabeVillalobos (approved):

PR #73015, opened 2024-06-19
fix(dashboards): Hide add widget button in templates

narsaynorath:
Adding a widget shouldn't be allowed in templates. The current behaviour is if you have a template preview open and you click "Add Widget", you'll go to a new dashboard with just the new widget. Hide it until we decide that this flow should be officially supported.


edwardgou-sentry (approved):
might be nice to have a test for preview mode!

PR #73020, opened 2024-06-19
feat(related_issues): Track trace timeline with two issues

armenzg:
The Related Issue feature replaces the Trace Timeline when only two issues are involved. This will help measure the impact of it.


leeandher (approved):
🚀

PR #73023, opened 2024-06-19
feat(related_issues): Track analytic even when feature is enabled

armenzg:
This is a follow up to #73020 since we are not tracking the metric when `related-issues-issue-details-page` is enabled.

### DIFF
--- a/static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
+++ b/static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
@@ -205,6 +205,13 @@ describe('TraceTimeline', () => {
 
     // Test analytics
     await userEvent.click(await screen.findByText('Slow DB Query'));
+    expect(useRouteAnalyticsParams).toHaveBeenLastCalledWith({
+      has_related_trace_issue: true,
+      trace_timeline_status: 'empty',
+      // Even though the trace timeline has not been rendered, we still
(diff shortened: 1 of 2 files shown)


static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -205,6 +205,13 @@ describe('TraceTimeline', () => {
 
     // Test analytics
     await userEvent.click(await screen.findByText('Slow DB Query'));
+    expect(useRouteAnalyticsParams).toHaveBeenLastCalledWith({
+      has_related_trace_issue: true,
```
armenzg, line 209: This test case is when we test that a related issue has been rendered.

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -205,6 +205,13 @@ describe('TraceTimeline', () => {
...
+      trace_timeline_status: 'empty',
```
armenzg, line 210: When the feature is enabled, the timeline will now be shown.
https://github.com/getsentry/sentry/blob/5c1824ad54e205b8b522daedcc7602789e4b6da6/static/app/views/issueDetails/traceTimeline/traceTimeline.tsx#L33-L35

leeandher (approved):

PR #73024, opened 2024-06-19
feat(trace-explorer): Use separate endpoint to load span samples

Zylphrex:
Once we move to this separate endpoint, we can remove the samples from the main query which can then be better optimized.

### DIFF
--- a/static/app/utils/analytics/tracingEventMap.tsx
+++ b/static/app/utils/analytics/tracingEventMap.tsx
@@ -55,7 +55,6 @@ export type TracingEventParameters = {
   'trace_explorer.search_success': {
     has_data: boolean;
     num_traces: number;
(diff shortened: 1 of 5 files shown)


static/app/utils/analytics/tracingEventMap.tsx
```diff
@@ -55,7 +55,6 @@ export type TracingEventParameters = {
   'trace_explorer.search_success': {
     has_data: boolean;
     num_traces: number;
-    project_platforms: string[];
```
Zylphrex, line 58: @KevinL10 is the project platforms necessary for analytics? It's being delegated to a separate query so it's no longer to possible to get a full list anymore.

k-fish (approved):
Looks good Tony 👍

PR #73030, opened 2024-06-19
feat(insights): add additional trace analytics

KevinL10:
more analytics for trace viewer/explorer.

### DIFF
--- a/static/app/views/performance/newTraceDetails/traceAnalytics.tsx
+++ b/static/app/views/performance/newTraceDetails/traceAnalytics.tsx
@@ -1,19 +1,38 @@
 import * as Sentry from '@sentry/react';
 
 import type {Organization} from 'sentry/types/organization';
+import type {Project} from 'sentry/types/project';
 import {trackAnalytics} from 'sentry/utils/analytics';
 import type {TraceTree} from 'sentry/views/performance/newTraceDetails/traceModels/traceTree';
 import type {TraceType} from 'sentry/views/performance/newTraceDetails/traceType';
 
-const trackTraceMetadata = (tree: TraceTree, organization: Organization) => {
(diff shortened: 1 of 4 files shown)


static/app/views/performance/newTraceDetails/traceAnalytics.tsx
```diff
@@ -1,19 +1,38 @@
...
+  const projectSlugs = [
+    ...new Set(
+      tree.list.map(node => node.metadata.project_slug).filter(slug => slug !== undefined)
+    ),
+  ];
+
+  const projectPlatforms = projectSlugs
+    .map(slug => projects.find(p => p.slug === slug))
+    .filter(project => project !== undefined)
+    .map(project => project?.platform ?? '');
```
narsaynorath, line 27:
```suggestion
  const projectPlatforms = projects
    .filter(({slug}) => projectSlugs.includes(slug))
    .map(project => project?.platform ?? '');
```

Just a small suggestion, instead of nesting the `.find` inside the `.map` we could just filter outright for the projects and I think that eliminates the need to do an undefined check

narsaynorath: If the platform _is_ undefined (I do think there are data issues where this does happen, or somehow was able to happen) should we leave it as the empty string or should we give it a different name like 'unknown'? Whatever works for the intended analytics.

narsaynorath (approved):

static/app/views/performance/newTraceDetails/traceAnalytics.tsx
KevinL10: thanks, updated the find/map. I'm keeping the project platforms empty because that's our standard behaviour on Amplitude for missing entries.

PR #73037, opened 2024-06-20
chore: move sentry.lint.engine to pre-commit

joshuarli:
sentry.lint.engine (very old at this point) is unused. This removes it, and consolidates all the JS linting stuff from the frontend CI workflow into pre-commit. This eliminates some otherwise redundant stuff with automatically committing file fixer changes, and gives pre-commit users full JS linting support.

Only small downside is versions in pre-commit config need to be manually synced. There's no easy way to install only devDependencies if we want to use `language: system`. pre-commit-managed environments are better, anyways, and we can take advantage of existing pre-commit's cache than have another node_modules cache.


scttcper:
@joshuarli to confirm, this would run eslint on the staged files matching the file extension?

.pre-commit-config.yaml
```diff
@@ -92,6 +92,16 @@ repos:
         entry: python3 -m tools.mypy_helpers.check_stronglist
         files: ^pyproject\.toml$
         language: python
+      - id: eslint
+        name: eslint
+        entry: ./node_modules/.bin/eslint
+        language: node
```
asottile-sentry, line 98: for now, these should be `language: system` or `language: script` -- otherwise this'll provision a throwaway js environment for nothing -- ideally these would be managed by pre-commit itself but that's more work than just moving them over

evanpurkhiser: https://github.com/pre-commit/mirrors-eslint

Any reason we shouldn't use this? Or does it make more sense to just use the one yarn brings in?

evanpurkhiser:
```
language: system
```

will use the volta node right?

evanpurkhiser:
How does this affect CI? We already have eslint running in CI right?

evanpurkhiser:
We have this thing

https://github.com/getsentry/sentry/blob/c015319b6020ee2c2130ac0513b680a10db2b371/.github/workflows/frontend.yml#L122-L128

@joshuarli if you want to just have this PR kill the lint engine, that seems fine? I don't think anything is even using it?

.pre-commit-config.yaml
asottile-sentry: yeah it'll use the one in the current user's environment -- I think for now (despite a manage environment typically being better) this is the most apples-to-apples conversion of the current script

joshuarli: pre-commit mirror didn't have the version we have specified in package.json, and also i wasn't sure if there was a pre-commit-supported way to use the version in package.json (is there?)

joshuarli:
> How does this affect CI? We already have eslint running in CI right?

Doesn't affect CI. This simply lets pre-commit users run all the js tools.

joshuarli:
> We have this thing
> 
> https://github.com/getsentry/sentry/blob/c015319b6020ee2c2130ac0513b680a10db2b371/.github/workflows/frontend.yml#L122-L128
> 
> @joshuarli if you want to just have this PR kill the lint engine, that seems fine? I don't think anything is even using it?

Yeah, I figured no one was using it, jsut wanted to double check. So, i'm gonna kill that and make pre-commit users happy.

evanpurkhiser (approved):

joshuarli:
Ok wait, now we have two workflows running js stuff and trying to commit them.

joshuarli:
I suspect if pre-commit succeeds we'd get two commits from both, it's just fast failing.

Maybe it would be better to keep the current paradigm of pre-commit for python only stuff for now?

Alternatively we read versions from package.json and move all the lint parts over from the "typescript and lint" action over to pre-commit. I think this would be nice to consolidate.

joshuarli:
I'm gonna consolidate into pre-commit, will figure out a way to just install devDependencies as I;'d like to avoid pulling in the whole node_modules. prettier isn't in there for some reason.

joshuarli:
hmm this is in CI but can't repro locally even with repo node_modules wiped

```
eslint...................................................................Failed
- hook id: eslint
- exit code: 2

Oops! Something went wrong! :(

ESLint: 8.57.0

ESLint couldn't find the plugin "eslint-plugin-jest".

(The package "eslint-plugin-jest" was not found when loaded as a Node module from the directory "/home/runner/work/sentry/sentry".)

 The plugin "eslint-plugin-jest" was referenced from the config file in ".eslintrc.js » eslint-config-sentry-app/strict » eslint-config-sentry-app » eslint-config-sentry-react".
```

joshuarli:
also pre-commit <-> volta isn't working

```
runner@fv-az692-824:~/.cache/pre-commit/repohqonok8l/node_env-default/bin$ ./node --version
v22.3.0
```

joshuarli:
ok i can repro locally now (not sure why), snuck in a console.log in eslint and things look fine...

```
  NODE_VIRTUAL_ENV: '/Users/josh/.cache/pre-commit/repoxcdoecuw/node_env-20.13.1',
  NPM_CONFIG_PREFIX: '/Users/josh/.cache/pre-commit/repoxcdoecuw/node_env-20.13.1',
  npm_config_prefix: '/Users/josh/.cache/pre-commit/repoxcdoecuw/node_env-20.13.1',
  NODE_PATH: '/Users/josh/.cache/pre-commit/repoxcdoecuw/node_env-20.13.1/lib/node_modules'
}

Oops! Something went wrong! :(

ESLint: 8.57.0

ESLint couldn't find the plugin "eslint-plugin-jest".

(The package "eslint-plugin-jest" was not found when loaded as a Node module from the directory "/Users/josh/dev/sentry".)
```

PR #73044, opened 2024-06-20
feat(insights): Screen count data per release

philipphofmann:
Show screen counts for app starts and screen loads per release.

| Module | Before | After |
|------------|--------|--------|
| App Starts    | CleanShot 2024-06-20 at 11 03 23@2x | CleanShot 2024-06-20 at 11 02 47@2x | 
| Screen Loads    | CleanShot 2024-06-20 at 11 04 33@2x | CleanShot 2024-06-20 at 11 04 50@2x |


static/app/views/insights/mobile/screenload/views/screenLoadSpansPage.tsx
```diff
@@ -67,6 +67,9 @@ function ScreenLoadSpans() {
     spanDescription,
   } = location.query;
 
+  const countIfPrimaryRelease = `count_if(measurements.time_to_initial_display,release,${primaryRelease})`;
+  const countIfSecondaryRelease = `count_if(measurements.time_to_initial_display,release,${secondaryRelease})`;
```
mjq, line 71: Personally I'd probably keep these inlined so a reader doesn't have to jump around but that's a tiny nitpick.

mjq (approved):

static/app/views/insights/mobile/screenload/views/screenLoadSpansPage.tsx
philipphofmann: I had to change it a few times in two places, but either we have variables for all of these or none. I change it back to having it duplicated.

PR #73052, opened 2024-06-20
feat(tracing) add tracing session integration

JonasBa:
Add a session integration so we can see if it would be feasible to connect traces in a similar way as how replay connects them via replay id so that we can provide a better navigation experience


mjq (changes requested):
The current session is read from `sessionStorage`, but I can't see anywhere where we ever write it to `sessionStorage`? 🤔

JonasBa:
Oh yea

> The current session is read from `sessionStorage`, but I can't see anywhere where we ever write it to `sessionStorage`? 🤔

You are correct. I think we'll abandon this concept tbh and I'm going to close the PR. I think there are two different features that we want, one is to show related traces if the current trace has a replay and another feature is to show similar traces. We'll first tackle the issue of showing related traces and see if we can find a more general approach that would work for the rest of the SDK which are not necessarily just frontend.

PR #73056, opened 2024-06-20
fix(feedback): Create a store so we can track feedback integration loading and re-render

ryan953:
This adds a context that we can use to store the loaded Feedback integration, and re-render children when that thing is loaded. It's got to be at the top of the react heirarchy like this because feedback is also loaded high in the tree and rendered everywhere.

See https://github.com/getsentry/getsentry/pull/14390

Related to https://github.com/getsentry/team-replay/issues/438

### DIFF
--- a/static/app/components/feedback/widget/useFeedbackWidget.tsx
+++ b/static/app/components/feedback/widget/useFeedbackWidget.tsx
@@ -5,6 +5,9 @@ import * as Sentry from '@sentry/react';
 import {t} from 'sentry/locale';
 import ConfigStore from 'sentry/stores/configStore';
 import {useLegacyStore} from 'sentry/stores/useLegacyStore';
+import useAsyncSDKIntegrationStore from 'sentry/views/app/asyncSDKIntegrationProvider';
+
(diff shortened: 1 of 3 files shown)


michellewzhang (approved):

static/app/components/feedback/widget/useFeedbackWidget.tsx
```diff
@@ -18,7 +21,12 @@ export default function useFeedbackWidget({
   messagePlaceholder,
 }: Props) {
   const config = useLegacyStore(ConfigStore);
-  const feedback = Sentry.getFeedback();
+  const {state} = useAsyncSDKIntegrationStore();
+
+  // TODO(ryan953): remove the fallback `?? Sentry.getFeedback()` after
+  // getsentry is calling `store.add(feedback);`
+  const feedback =
+    (state.Feedback as FeedbackIntegration | undefined) ?? Sentry.getFeedback();
```
ryan953, line 29: the `state.Feedback` here must sync up with the setState call in getsentry. Very important stuff!

`Feedback` in this case is also the proper name of the integration, so that's consistent too.

PR #73062, opened 2024-06-20
fix(processing): Use correct comparison interval

schew2381:
From Slack:
Comparison conditions are using the incorrect comparison interval (comparison interval is "how far back should we go". interval is "what period of time should we query")

Basically when deciding how far back to go for the past session, we were using 1 instead of 2
Screenshot 2024-06-20 at 10 21 34 AM

Additional PR to fix test bug: https://github.com/getsentry/sentry/pull/73235

Fixes ALRT-100

### DIFF
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -15,6 +15,7 @@ from sentry.models.rule import Rule
 from sentry.models.rulesnooze import RuleSnooze
 from sentry.rules import history, rules
 from sentry.rules.conditions.event_frequency import (
+    DEFAULT_COMPARISON_INTERVAL,
     BaseEventFrequencyCondition,
     ComparisonType,
     EventFrequencyConditionData,
@@ -143,7 +144,7 @@ def get_condition_group_results(
 
         _, duration = condition_inst.intervals[unique_condition.interval]
         comparison_interval = condition_inst.intervals[
-            condition_data.get("comparisonInterval", "5m")
+            condition_data.get("comparisonInterval", DEFAULT_COMPARISON_INTERVAL)
         ][1]
         comparison_type = (
             condition_data.get("comparisonType", ComparisonType.COUNT)
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -48,7 +48,7 @@ class UniqueCondition(NamedTuple):
 
 
 class DataAndGroups(NamedTuple):
-    data: EventFrequencyConditionData | None
+    data: EventFrequencyConditionData
     group_ids: set[int]
 
     def __repr__(self):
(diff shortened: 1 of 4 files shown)


src/sentry/rules/processing/delayed_processing.py
```diff
@@ -48,7 +48,7 @@ def __repr__(self):
 
 
 class DataAndGroups(NamedTuple):
-    data: EventFrequencyConditionData | None
+    data: EventFrequencyConditionData
```
schew2381, line 51: Small typing fix, this is always set

saponifi3d (approved):

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -142,7 +142,9 @@ def get_condition_group_results(
             return None
 
         _, duration = condition_inst.intervals[unique_condition.interval]
-        comparison_interval = condition_inst.intervals[unique_condition.interval][1]
+        comparison_interval = condition_inst.intervals[
+            condition_data.get("comparisonInterval", "5m")
```
schew2381, line 146: The default is not really necessary b/c all percent conditions should have this set, and we only use condition_interval in `get_rate_bulk` if the condition is a percent condition in the first place.

However, it's good to be safe and match what we have for regular processing
https://github.com/getsentry/sentry/blob/598b30da271b9ff83226c0f6d85912e4f3e64af1/src/sentry/rules/conditions/event_frequency.py#L168

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -142,7 +142,9 @@ def get_condition_group_results(
...
+        ][1]
```
saponifi3d, line 147: It'd be nice to include a unit test for this to make sure we don't run into the issue again in the future.

ceorourke: ^ this and/or a test with a rule where the comparison interval and the comparison value are different

src/sentry/rules/processing/delayed_processing.py
saponifi3d: yeah, nice to include the default. i think extra credit would be to have that default set as a constant. then we can reuse it for the API or other areas of code in the future.

wedamija, line 146: So just keep in mind here that `condition_data` comes from the first condition we see for that `unique_condition`. So since we don't currently unique on different `comparisonInterval`, this will still produce incorrect results if we have two rules with conditions like

`% change of events in the last hour compared to 1 day ago`
`% change of events in the last hour compared to week day ago`

Whichever of these conditions is processed first will have  the comparison interval set, and the second will incorrectly compare to the previous conditions `comparisonInterval`.

I assume you're going to fix this bug separately so I'm ok with merging, but just wanted to make sure you're aware.

src/sentry/rules/processing/delayed_processing.py
wedamija: +1 on a test here, we need something that reproduces this bug without the fix

src/sentry/rules/processing/delayed_processing.py
schew2381: oh yes this was just the easier of the 2 bugs to tackle first

src/sentry/rules/processing/delayed_processing.py
schew2381: ok wrote the test and confirmed it fails with the previous code but works now

PR #73067, opened 2024-06-20
fix(autofix): Hide when stack trace does not contain any frames

malwilley:
We used to be looking for the existence of one of the interfaces, but we really should be checking for frames since Autofix will fail without those. This matches the logic in seer.


roggenkemper (approved):
lgtm , do we not care about EntryType.Stacktrace with this new implementation?

malwilley:
> lgtm , do we not care about EntryType.Stacktrace with this new implementation?

We actually aren't checking that in Seer so we need to change that first. Not actually sure which events produce stack trace entry types

Issue #51994, opened 2023-06-30
Sunset the App Store Connect Integration


philipphofmann:
### Problem Statement

Since Xcode 14, you can't download dSYMs anymore from AppStore Connect. See

> Debug symbols can only be downloaded from App Store Connect / TestFlight for existing bitcode submissions and are no longer available for submissions made with Xcode 14. (86118779) - https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes#Deprecations

We don't know why our integration worked for a couple of Xcode 14 versions. 

Anyways, since around June 28th, 2023, users reported problems ([Twitter](https://twitter.com/denrase/status/1673968897257603072?s=20)),  that our App Store Connect integration fails to download the dSYMs. The integration only confuses users and causes problems. Therefore, we have to remove it.

### Tasks

```[tasklist]
### User-facing
- [x] platform docs: we updated the platform SDK docs for Apple, providing clarity around why bitcode can't be used anymore and that dSYMs need to be manually managed
- [x] product docs: not featured (anymore)
- [x] decide on deprecation/grace period
- [x] ~~add deprecation note to ASC integration settings~~
- [x] (optional) email orgs that currently have ASC integration set up
- [x] hide ASC integration for all orgs that don't use it yet https://github.com/getsentry/sentry/pull/72202
- [x] wait for deprecation period to pass (July 10, 2024)
- [x] hide ASC integration settings for all orgs https://github.com/getsentry/sentry/pull/73071
```

```[tasklist]
### Cleanup
- [ ] https://github.com/getsentry/sentry/pull/72203
- [ ] https://github.com/getsentry/sentry/pull/72206
- [ ] https://github.com/getsentry/getsentry/pull/14222
- [ ] https://github.com/getsentry/sentry/pull/71987
- [ ] https://github.com/getsentry/sentry/pull/73952
- [ ] https://github.com/getsentry/sentry/pull/73960
- [ ] https://github.com/getsentry/sentry/pull/74100
```

### Product Area

Integrations

kahest:
Wrt deprecation plans:

- [x] platform docs: we updated the platform SDK docs for Apple, providing clarity around why bitcode can't be used anymore and that dSYMs need to be manually managed
- [x] product docs: not featured (anymore)
- [ ] decide on deprecation/grace period (e.g. 1-2 months)
- [ ] add deprecation note to ASC integration settings
- [ ] (optional) email orgs that currently have ASC integration set up
[...]

PR #73071, opened 2024-06-20
chore: remove appstore-connect frontend code

anonrig:
Removes the frontend code from Sentry for `appstore-connect` integration which will be sunsetted soon.

After merging this in, we also need to remove the backend implementation.

cc @kahest @jernejstrasner 

Ref: https://github.com/getsentry/sentry/issues/51994


armenzg (approved):

static/app/components/projects/appStoreConnectContext.tsx
```diff
@@ -1,144 +0,0 @@
...
-
-    if (project.symbolSources) {
-      setProjectDetails(project);
-      return undefined;
-    }
-
-    let unmounted = false;
-
-    api
-      .requestPromise(`/projects/${organization.slug}/${project.slug}/`)
```
scttcper, line 55: great i was actually just looking at this file, makes an extra project request

Swatinem (approved):

PR #73074, opened 2024-06-20
ref(metrics): Change existing MetricsQuery to DeprecatingMetricsQuery

nikhars:
We have the existing MetricsQuery which we would like to deprecate sometime in the fulture. Also, its name is same as the new MetricsQuery which snuba sdk provides [here.](https://github.com/getsentry/snuba-sdk/blob/main/snuba_sdk/metrics_query.py#L19) Which makes it really hard to differentiate between the 2. Renaming the existing one to make things simpler


iambriccardo (approved):
Looks good to me! Thanks for doing this, it's an important step towards the adoption of the new layer

shellmayr (approved):

vgrozdanic (approved):

wmak (approved):

PR #73075, opened 2024-06-20
feat(similarity): add functionality to backfill a cohort of projects

JoshFerge:
- [x] since we need to daisy chain our celery tasks, we can't simply loop through a list of projects and call the backfill on them, we need to daisy chain the celery call over the projects
- [x] add new parameters, `cohort`, `last_processed_project_index` to the celery task parameters
- [x] in places where we'd simply return before, we'll now try to call the next backfill, and if there are more project_ids in the cohort, we'll start on the next project instead of returning
- [x] we add logic to correctly increment the project_id_index between jobs
- [x] adds tests. there are a few tests that could be written to test the redis key and feature branches, but we can follow on if these scenarios cause issues..


src/sentry/tasks/embeddings_grouping/backfill_seer_grouping_records_for_project.py
```diff
@@ -138,48 +189,106 @@ def backfill_seer_grouping_records_for_project(
...
+                batch_project_id,
+                None,
+                cohort,
+                last_processed_project_index,
+                only_delete,
+            ],
+        )
+
+
+def get_project_for_batch(last_processed_project_index, cohort_list, cohort_name):
```
jangjodi, line 288: nit: would this be better in utils.py?

src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -106,13 +111,28 @@ def initialize_backfill(project_id, last_processed_index):
...
+        last_processed_group_index_ret = int(
+            redis_client.get(make_backfill_grouping_index_redis_key(project_id)) or 0
+        )
+    else:
+        last_processed_group_index_ret = last_processed_group_index
+
+    if last_processed_project_index is None:
+        if cohort and isinstance(cohort, str):
+            last_processed_project_index_ret = int(
+                redis_client.get(make_backfill_project_index_redis_key(cohort, project_id)) or 0
```
jangjodi, line 124: what value would this be getting? I only see us setting it using `last_processed_project_index` instead of `project_id`

JoshFerge: ah great catch! 🙌🏼 thank you

jangjodi (commented):
lgtm!

src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -452,11 +472,16 @@ def lookup_event(project_id: int, event_id: str, group_id: int) -> Event:
     return event
 
 
-def make_backfill_redis_key(project_id: int):
-    redis_key = "grouping_record_backfill.last_processed_index"
+def make_backfill_grouping_index_redis_key(project_id: int):
+    redis_key = "grouping_record_backfill.last_processed_grouping_index"
```
lobsterkatie, line 476: Nit:

Why `grouping_index` here and `group_index` elsewhere? Aren't they all keeping track of which group you're on?

src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -487,3 +512,12 @@ def delete_seer_grouping_records(
...
         Group.objects.bulk_update(groups_with_seer_metadata, ["data"])
+
+
+def get_project_for_batch(last_processed_project_index, cohort_list, cohort_name):
+    next_cohort_index = last_processed_project_index + 1
+    if next_cohort_index >= len(cohort_list):
+        return None, None
+    project_id = cohort_list[next_cohort_index]
+    last_processed_project_index = next_cohort_index
+    return project_id, last_processed_project_index
```
lobsterkatie, line 523: `cohort_name` doesn't seem like it's used.

Naming nits:

```suggestion
def get_project_for_batch(last_processed_project_index, cohort_projects, cohort_name):
    next_project_index = last_processed_project_index + 1
    if next_project_index >= len(cohort_projects):
        return None, None
    project_id = cohort_projects[next_project_index]
    last_processed_project_index = next_project_index
    return project_id, last_processed_project_index
```

lobsterkatie (approved):
Looks good overall! There were some spots that weren't super clear to me (as someone who hasn't been in the backfill code all along) which I pointed out, but feel free to take or leave those suggestions as you see fit.

src/sentry/tasks/embeddings_grouping/backfill_seer_grouping_records_for_project.py
```diff
@@ -36,34 +40,66 @@
...
         )
+        # TODO: let's just delete this branch since feature is on
         return
 
     if only_delete:
         delete_seer_grouping_records(project.id, redis_client)
         logger.info(
             "backfill_seer_grouping_records.deleted_all_records",
-            extra={"project_id": project.id},
+            extra={"current_project_id": project.id},
```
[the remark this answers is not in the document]
JoshFerge: doesn't really matter, i'll standardize 👍🏼

src/sentry/tasks/embeddings_grouping/backfill_seer_grouping_records_for_project.py
```diff
@@ -89,11 +134,13 @@ def backfill_seer_grouping_records_for_project(
```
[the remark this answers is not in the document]
JoshFerge: yes, and a few reasons: 1) the row got expired in clickhouse, and for nodestore 2) it's not so much that there were no nodestore rows, but that the stacktraces _in_ nodestore weren't eligible for seer.

>do some of the ones we're backfilling already have embeddings (and if so, why are we backfilling them
yes, in case the backfill gets ran twice or something. we don't backfill them.

> What is the hashes dict, and what would make it empty?
it is the mapping of group hashes to group_id 

>Possibly out of scope for this PR, since it's not new here, but reading the backfill code for the first time, it's not clear quite what's happening here between lines 130 and 166. Why are we filtering the snuba results? has_snuba_row and has_nodestore_row imply that there might be some without snuba and/or nodestore rows? How/when would that happen? Similarly, with_no_embeddings... do some of the ones we're backfilling already have embeddings (and if so, why are we backfilling them)? What is the hashes dict, and what would make it empty?

> Could we add a few comments explaining what's happening and why?

if we end up needing to put a lot more work into the backfill, yes, but i imagine as we release this featue we'll end up deleting it soon, so could add comments if it becomes opaque to us as we work o it.

src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -92,12 +92,17 @@ def filter_snuba_results(snuba_results, groups_to_backfill_with_no_embedding, pr
 
 
 @sentry_sdk.tracing.trace
-def initialize_backfill(project_id, last_processed_index):
+def initialize_backfill(
+    project_id: int,
+    cohort: str | list[int] | None,
+    last_processed_group_index: int | None,
+    last_processed_project_index: int | None,
+):
```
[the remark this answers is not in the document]
JoshFerge: this gets called at the start of every backfill batch, which is what one celery task is. and can add docstrings / comments if we end up needing to loop more folks into this code.

src/sentry/tasks/embeddings_grouping/utils.py
JoshFerge: no reason, just a small inconsistency. can end up fixing if it causes confusion

src/sentry/tasks/embeddings_grouping/utils.py
JoshFerge: yup. thanks

src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -106,13 +111,28 @@ def initialize_backfill(project_id, last_processed_index):
...
 
-    if last_processed_index is None:
-        last_processed_index = int(redis_client.get(make_backfill_redis_key(project_id)) or 0)
-    return project, redis_client, last_processed_index
+    if last_processed_group_index is None:
+        last_processed_group_index_ret = int(
+            redis_client.get(make_backfill_grouping_index_redis_key(project_id)) or 0
+        )
+    else:
+        last_processed_group_index_ret = last_processed_group_index
```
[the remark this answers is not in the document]
JoshFerge: mypy hates re-assignment :(

PR #73099, opened 2024-06-21
fix(processing): Separate second query of percentage comparison

schew2381:
Fixes count and percent conditions bleeding into each other. The original bug occurred as follows:
1. `UniqueCondition` only accounts for interval, env, and condition class. Percent and count queries could be grouped into the same `UniqueCondition`
2. When querying, we [checked](https://github.com/getsentry/sentry/pull/73099/files#diff-f906e75a0e4419db4870fa45ca5a1608ca79beaa052c8bc50b4805607a665d27L149-L152) what condition type was in `DataAndGroups`.
3. `get_rate_bulk` for `EventFrequencyConditions` decides how to query depending on the condition type we pass in this. However, count and percent queries can share the same `UniqueCondition` and thus the same `DataAndGroups`, so depending on whatever condition type came first when creating the [initial `UniqueCondition`](https://github.com/getsentry/sentry/pull/73099/files#diff-f906e75a0e4419db4870fa45ca5a1608ca79beaa052c8bc50b4805607a665d27L120-L121), we would either be making a percent calculation for count queries or vice versa.

This PR fixes the bug by adding `comparisonInterval` to `UniqueCondition`, allowing us to distinguish btwn the second percent query and count/first percent queries. This way, we can still optimize the first percent query + count queries that are able to share the same scan.

One caveat is we have to move the percent calculation out of `get_rate_bulk` and into `delayed_processing`, but this is a small fix.

Fixes ALRT-101


src/sentry/rules/processing/delayed_processing.py
```diff
@@ -142,26 +168,60 @@ def get_condition_group_results(
...
+            "delayed_processing.missing_query_results", extra={"exception": e, "group_id": group_id}
+        )
+        return False
+
+    calculated_value = query_values[0]
+    # If there's a second query we must have a percent comparison condition.
+    if len(query_values) == 2:
+        calculated_value = percent_increase(calculated_value, query_values[1])
+
+    target_value = float(condition_data["value"])
```
wedamija, line 218: Previously we converted this to a string then float, not sure if that's important though

wedamija (commented):
This makes sense to me, I'll give it a second review once you've finished tests

src/sentry/rules/conditions/event_frequency.py
```diff
@@ -261,15 +261,24 @@ def disable_consistent_snuba_mode(
         return option_override_cm
 
     def get_comparison_start_end(
-        self, interval: timedelta, duration: timedelta
+        self, end: datetime, duration: timedelta, comparison_interval: timedelta | None = None
```
nhsiehgit, line 264: can we call `comparison_interval` `comparison_offset` instead?

The verbage is a bit confusing

src/sentry/rules/conditions/event_frequency.py
```diff
@@ -261,15 +261,24 @@ def disable_consistent_snuba_mode(
...
-        "The issue is seen more than 100 times in 10 minutes"
+        Calculate the start and end times for the query.
+        Take these example conditions that generate queries:
+            '# of issues in 1 day is {}'
+            "# of issues is {}% higher in 1 day compared to 12 hours ago"
+
+        "duration" is the period of time we're querying over, in this case 1 day
+        for both queries.
+        "comparison_interval" only applies for percent comparison queries (the
+        second condition) and is the time diff between the two queries we derive a
```
nhsiehgit, line 275: > diff between the two queries we derive a percentage from
🤔 this is a bit confusing and the example isn't really helping me much here...

I think what you're trying to say is that while the duration is 24hrs, we've offset our comparison by 12hrs. So the query becomes comparing [24hrs duration from now] to [24hrs duration from 12hrs ago]?

src/sentry/rules/conditions/event_frequency.py
```diff
@@ -261,15 +261,24 @@ def disable_consistent_snuba_mode(
...
+        percentage from, in this case 12 hours. The two queries would be:
+            1. now -> 24 hours ago
+            2. 12 hours ago -> 36 hours ago
         """
-        end = timezone.now() - interval
+        if comparison_interval:
+            end = end - comparison_interval
```
nhsiehgit, line 281: If we're already passing in the `end` - why do we pass in the offset?
Can we just calculate the new `end` when invoking this helper?

src/sentry/rules/conditions/event_frequency.py
```diff
@@ -281,14 +290,17 @@ def get_rate(
         environment_id: int,
         comparison_type: str,
     ) -> int:
-        start, end = self.get_comparison_start_end(timedelta(), duration)
+        current_time = timezone.now()
+        start, end = self.get_comparison_start_end(current_time, duration)
```
nhsiehgit, line 294: lets be explicit in args and key them

src/sentry/rules/conditions/event_frequency.py
```diff
@@ -281,14 +290,17 @@ def get_rate(
...
         with self.disable_consistent_snuba_mode(duration):
             result = self.query(event, start, end, environment_id=environment_id)
             if comparison_type == ComparisonType.PERCENT:
                 # TODO: Figure out if there's a way we can do this less frequently. All queries are
                 # automatically cached for 10s. We could consider trying to cache this and the main
                 # query for 20s to reduce the load.
-                start, end = self.get_comparison_start_end(comparison_interval, duration)
+                start, end = self.get_comparison_start_end(
```
nhsiehgit, line 301: RE: comments above

1 - lets make this kwargs
2 - lets pre-calucate the `end` here

src/sentry/rules/conditions/event_frequency.py
```diff
@@ -297,31 +309,32 @@ def get_rate(
...
-        comparison_type: str,
+        current_time: datetime,
+        comparison_interval: timedelta | None,
     ) -> dict[int, int]:
-        start, end = self.get_comparison_start_end(timedelta(), duration)
+        """
+        Make a batch query for multiple groups. The return value is a dictionary
+        of group_id to the result for that group.
+        """
+        # If comparison_interval is not None, we're making the second query in a
```
nhsiehgit, line 321: should these notes go into the docstring?

nhsiehgit (commented):
some pre-lim comments

haven't looked over the tests yet, lookin gnow

tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -598,3 +608,93 @@ def test_apply_delayed_percent_condition_comparison_interval(self):
...
+        """
+        Test that having both count and percent comparison type conditions do
+        not affect each other and the order the different types of conditions
+        are processed does not matter.
+
+        Also ensure the count query and first percent query share the same
+        query, but the second percent query is separate.
+        """
+
+        def ordered_callthrough(descending=False):
```
nhsiehgit, line 622: 🤔 can we call this `mock_get_condition_group`? or toss `side_effect`/`callback` somewhere in teh name of this?

nhsiehgit: Also - does this need to be a separate helper?

Why are we modifying the args being passed into this method? This means we _aren't_ properly capturing how the implementation would behave.

This feels smelly 🤔

nhsiehgit (commented):
some more comments on the tests

src/sentry/rules/conditions/event_frequency.py
schew2381: I can cleanup the explanation

src/sentry/rules/conditions/event_frequency.py
schew2381: We could but I liked having this function handle all the time calculations instead of doing it in `get_rate_bulk` and `get_rate`

tests/sentry/rules/processing/test_delayed_processing.py
schew2381: We're modifying the order of the list of rules being passed into the function. We need to do this to test the behavior when count and percent conditions can share a query, so that ordering doesn't matter.

Originally they come in a random order, so we test for both orderings (count first then percent and vice versa)

src/sentry/rules/processing/delayed_processing.py
schew2381: @ceorourke do you remember why we originally did that?

ceorourke: I think it was just to match this https://github.com/getsentry/sentry/blob/master/src/sentry/rules/conditions/event_frequency.py - what's the data type initially? A float?

schew2381: It's a string but I think we can just convert that directly, like [here](https://github.com/getsentry/sentry/blob/master/src/sentry/rules/conditions/event_frequency.py#L156)

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -39,13 +40,20 @@
...
+    """
+
     cls_id: str
     interval: str
     environment_id: int
+    comparisonInterval: str | None = None
 
-    def __repr__(self):
-        return f"id: {self.cls_id},\ninterval: {self.interval},\nenv id: {self.environment_id}"
+    # def __repr__(self):
```
ceorourke, line 55: Is this supposed to be in here?

schew2381: oh whoops I removed that to make debugging easier, will change it back

tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -681,17 +702,18 @@ def test_apply_delayed_percent_comparison_condition_interval(self):
         the correct comparison interval, e.g. # events is ... compared to 1 hr ago
         """
         percent_condition = self.create_event_frequency_condition(
-            interval="5m",
+            interval="1h",
```
schew2381, line 705: The interval and comparison interval need to be changed for this test in order to steer clear of any snuba rollup issues. The rollup for intervals of <= 1 hr is only 10 seconds, so it's effect is minimal on shifting the query boundaries

src/sentry/rules/conditions/event_frequency.py
saponifi3d: +1 on just subtracting the `comparison_interval` from the `end` here since we are already doing the logic to check for comparisons here.

Perhaps we rename the function to like `get_query_window` or something a little more descriptive to it as well.

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -92,42 +100,75 @@ def get_rules_to_slow_conditions(
...
+            interval=condition_data["interval"],
+            environment_id=environment_id,
+        )
+    ]
+    if condition_data.get("comparisonType") == ComparisonType.PERCENT:
+        # We will later compare the first query results against the second query to calculate
+        # a percentage for percentage comparison conditions.
+        comparison_interval = condition_data.get("comparisonInterval", DEFAULT_COMPARISON_INTERVAL)
+        second_comparison_query = unique_queries[0]._replace(comparisonInterval=comparison_interval)
+        unique_queries.append(second_comparison_query)
```
saponifi3d, line 124: Hmm - something about this feels a little off. it seems like we're creating a unique query, modifying it, and then appending it to the list.

I'm a little nervous about the mutation to the [unique_queries[0]._replace](https://github.com/getsentry/sentry/blob/bc54bac083ab7c585b778e8e1686cc274d8d2de8/src/sentry/rules/processing/delayed_processing.py#L123) since it reads like a mutation of the first element. While `._replace` creates a copy rather than mutating, maybe we help our future selves by making a copy of the query and then updating it?

saponifi3d (approved):
left a few nits. just want to make sure you're confident in this fix and that we won't be introducing any new regressions.

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -39,13 +40,20 @@
...
+    """
+    Represents all the data that uniquely identifies a condition class and its
+    single respective Snuba query that must be made. Multiple instances of the
+    same condition class can share the single query.
+    """
+
     cls_id: str
     interval: str
     environment_id: int
+    comparisonInterval: str | None = None
```
nhsiehgit, line 53: snake case rather than camel case?

src/sentry/rules/processing/delayed_processing.py
nhsiehgit: agreed with this.
Can we simple construct another class instance and splat in the values while overwriting the comparion interval instead?

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -92,42 +100,75 @@ def get_rules_to_slow_conditions(
...
     for rule in alert_rules:
         # We only want a rule's slow conditions because alert_rules are only added
         # to the buffer if we've already checked their fast conditions.
         slow_conditions = get_slow_conditions(rule)
         for condition_data in slow_conditions:
-            if condition_data:
-                unique_condition = UniqueCondition(
-                    condition_data["id"], condition_data["interval"], rule.environment_id
-                )
+            for unique_cond in _generate_unique_queries(condition_data, rule.environment_id):
```
nhsiehgit, line 143: maybe just `condition` rather than abbreviating?

nhsiehgit (commented):
few nit comments

src/sentry/rules/processing/delayed_processing.py
schew2381: oh right, whoops

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -174,23 +247,17 @@ def get_rules_to_fire(
...
+                    condition_group_results, slow_condition, group_id, alert_rule.environment_id
+                ):
+                    if action_match == "any":
+                        rules_to_fire[alert_rule].add(group_id)
+                        break
+                    conditions_matched += 1
+                else:
+                    if action_match == "all":
+                        # We failed to match all conditions for this group, skip
+                        break
```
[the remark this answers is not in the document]
schew2381: We have the break there to stop looping over slow conditions if any/all fail, with the new code I think the looping would continue even tho we know that the condition has already failed

src/sentry/rules/processing/delayed_processing.py
schew2381: I was following the docs [here](https://docs.python.org/dev/library/collections.html#collections.somenamedtuple._replace) but for readability I can explicitly make a new instance

schew2381: ok I can't create a copy and then update it's value cuz it's a tuple :|
I can use unpacking but it's a bit ugly as it doesn't handle overrides.

```Python
first_query_copy = unique_queries[0]._asdict()
first_query_copy["comparison_interval"] = comparison_interval
second_comparison_query = UniqueConditionQuery(**first_query_copy)
```

src/sentry/rules/processing/delayed_processing.py
schew2381: Swapped to `condition_query`

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -120,7 +120,9 @@ def _generate_unique_queries(
         # We will later compare the first query results against the second query to calculate
         # a percentage for percentage comparison conditions.
         comparison_interval = condition_data.get("comparisonInterval", DEFAULT_COMPARISON_INTERVAL)
-        second_comparison_query = unique_queries[0]._replace(comparisonInterval=comparison_interval)
+        first_query_copy = unique_queries[0]._asdict()
+        first_query_copy["comparison_interval"] = comparison_interval
```
saponifi3d, line 124: **nit**: on the names that `first_query_copy` is the same thing as `second_query` and it might be a little easier to grok that we're building the second query rather than naming it a copy.

```python
second_query_data = deepcopy(unique_queries[0])
second_query_data["comparison_interval"] = comparison_interval
unique_queries.append(UniqueConditionQuery(**second_query_data))
```
This would reduce 4 lines to 3, and reads a little easier. we also don't reference the `second_query_comparison` later, so we can just append it to the array.

saponifi3d (approved):
lgtm, thanks for the cleanup! i'd say to save my comment for when you're taking a look at adding more unit tests as it's a noop.

nhsiehgit (approved):
👍

tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -718,3 +717,116 @@ def test_apply_delayed_percent_comparison_condition_interval(self):
...
+        # not be triggered.
+        project_id = _setup_events()
+        with patch(
+            "sentry.rules.processing.delayed_processing.get_condition_query_groups",
+            side_effect=mock_get_condition_group(descending=False),
+        ):
+            apply_delayed(project_id)
+        assert_results(iteration=1)
+
+        # Have a count condition be processed first. It's calculated value is 2,
```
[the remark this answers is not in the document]
schew2381: ok I split up the ordering test into two tests

PR #73100, opened 2024-06-21
feat(uptime): Use subscriptions to process uptime messages.

wedamija:
This abstracts the subscription processor and starts attempting to fetch subscriptions when processing results. To keep our dummy test case going, for now we'll create dummy models if the subscription can't be found.


src/sentry/remote_subscriptions/consumers/result_consumer.py
```diff
@@ -0,0 +1,97 @@
...
+from sentry_kafka_schemas.codecs import Codec
+from sentry_kafka_schemas.schema_types.uptime_results_v1 import CheckResult
+
+from sentry.conf.types.kafka_definition import Topic, get_topic_codec
+from sentry.db.models import FlexibleForeignKey, Model
+from sentry.remote_subscriptions.models import RemoteSubscription
+
+logger = logging.getLogger(__name__)
+
+UPTIME_RESULTS_CODEC: Codec[CheckResult] = get_topic_codec(Topic.UPTIME_RESULTS)
```
evanpurkhiser, line 22: Why does this live here now?

src/sentry/remote_subscriptions/consumers/result_consumer.py
```diff
@@ -0,0 +1,97 @@
...
+    @abc.abstractmethod
+    def get_subscription_id(self, result: T) -> str:
+        pass
+
+    @abc.abstractmethod
+    def handle_result(self, result: T):
+        pass
+
+
+class ResultsStrategyFactory(ProcessingStrategyFactory[KafkaPayload], Generic[T, U]):
```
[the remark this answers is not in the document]
evanpurkhiser: I guess I'm not exactly sure why we keep them separate, but seems like a common pattern for how we've written our consumers. Maybe @lynnagara or @untitaker has better reasoning for why the `function` doesn't just existing the strategy factory?

src/sentry/remote_subscriptions/consumers/result_consumer.py
wedamija: Oh I meant to totally remove this, it's unused

src/sentry/remote_subscriptions/consumers/result_consumer.py
```diff
@@ -0,0 +1,96 @@
...
+    def get_subscription_id(self, result: T) -> str:
+        pass
+
+    @abc.abstractmethod
+    def handle_result(self, subscription: U, result: T):
+        pass
+
+
+class ResultsStrategyFactory(ProcessingStrategyFactory[KafkaPayload], Generic[T, U]):
+    def __init__(self, topic: Topic) -> None:
```
evanpurkhiser, line 80: Where does this get passed from?

evanpurkhiser (approved):

src/sentry/remote_subscriptions/consumers/result_consumer.py
wedamija: Oops, yeah this wasn tested and doesn't work. Adding a test that covers this and fixing

PR #73105, opened 2024-06-21
feat(seer grouping): Handle non-200 responses from similar issues endpoint

lobsterkatie:
This adds handling of non-200 responses to requests to the Seer similar issues endpoint, which is called both during ingestion and when loading the Similar Issues tab of the issue details page. I chose to count a redirect as an error, because it would mean we'd updated the endpoint on the Seer end but not on the Sentry end, which is something which we would want to fix if it ever happened.

Working on this made me realize that we need to change the way we're using the circuit breaker, as neither erroring responses such as those handled here nor other errors already being handled by this function will increment the circuit breaker's error counter. (In order for that to happen, the error would need to be raised here and bubble up to and out of the circuit breaker's callback; by handling the errors internally, this function prevents that from happening.) So that needs to be fixed, but I'll do so in a separate PR.

Fixes https://github.com/getsentry/team-issues/issues/23.

### DIFF
--- a/src/sentry/seer/similarity/similar_issues.py
+++ b/src/sentry/seer/similarity/similar_issues.py
@@ -23,7 +23,6 @@ seer_grouping_connection_pool = connection_from_url(
 )
 
 
-# TODO: Handle non-200 responses
 def get_similarity_data_from_seer(
     similar_issues_request: SimilarIssuesEmbeddingsRequest,
 ) -> list[SeerSimilarIssueData]:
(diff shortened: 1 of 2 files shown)


src/sentry/seer/similarity/similar_issues.py
```diff
@@ -40,6 +39,19 @@ def get_similarity_data_from_seer(
...
 
+    if response.status > 200:
+        redirect = response.get_redirect_location()
+        if redirect:
+            logger.error(
+                f"Encountered redirect when calling Seer endpoint {SEER_SIMILAR_ISSUES_URL}. Please update `SEER_SIMILAR_ISSUES_URL` in `sentry.conf.server` to be '{redirect}'."  # noqa
+            )
+        else:
+            logger.error(
+                f"Received {response.status} when calling Seer endpoint {SEER_SIMILAR_ISSUES_URL}.",  # noqa
```
jangjodi, line 50: lgtm, but just curious why `# noqa` was needed here?

jangjodi (approved):

src/sentry/seer/similarity/similar_issues.py
lobsterkatie: To avoid triggering the no-f-strings [lint rule](https://github.com/adamchainz/flake8-logging#log011-avoid-pre-formatting-log-messages). In principle I get it, but in this case I actually _want_ different statuses to group separately. (It occurs to me as I'm writing this that our parameterization might thwart me on that even so, but still...)

PR #73107, opened 2024-06-21
fix(jira): validate domain of `jira_url` parameter in autocomplete

oioki:
* domain of `jira_url` has to match the domain of the jira instance configured in the plugin
* also added validation for `jira_url` presence and denested `if jira_url:` branch

### DIFF
--- a/src/sentry_plugins/jira/plugin.py
+++ b/src/sentry_plugins/jira/plugin.py
@@ -303,83 +303,101 @@ class JiraPlugin(CorePluginMixin, IssuePlugin2):
                 return Response({field: issues})
 
         jira_url = request.GET.get("jira_url")
-        if jira_url:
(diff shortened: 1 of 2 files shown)


src/sentry_plugins/jira/plugin.py
```diff
@@ -303,83 +303,101 @@ def view_autocomplete(self, request: Request, group, **kwargs):
...
+        instance_url = self.get_option("instance_url", group.project)
+
+        if parsed.netloc != urlsplit(instance_url).netloc:
+            return Response(
+                {
+                    "error_type": "validation",
+                    "errors": [{"jira_url": "domain must match"}],
+                },
+                status=400,
+            )
```
markstory, line 326: It would be good to cover the domain mismatch with a test.

markstory (approved):

PR #73111, opened 2024-06-21
ref: Move getFormattedTimestamp until utils/date/

ryan953:
Related to https://github.com/getsentry/frontend-tsc/issues/13


leeandher (approved):

scttcper:
i don't think i want anyone other than breadcrumbs to use this date function. it doesn't use the user's selected timezone, it doesn't use a <time> element. it doesn't use the user's 24 hour date preference

ryan953:
> i don't think i want anyone other than breadcrumbs to use this date function. it doesn't use the user's selected timezone, it doesn't use a `<time>` element. it doesn't use the user's 24 hour date preference

With all those problems why should breadcrumbs even use it?

scttcper:
i don't know, some of this stuff doesn't even make sense https://github.com/getsentry/sentry/blob/ryan953/ref-getFormattedTimestamp/static/app/utils/date/getFormattedTimestamp.tsx#L15

ryan953:
@scttcper I want to move the file first, if things were wrong before then they'll still be wrong after i move it around. But we'll create a collection of all these date related things and be able to fix them up. 

There's a similar collection of duration related files in utils/duration/ and i'm sure loads of the code in there is either duplicate, or somehow overly specialized and omits things like preferences and locale, just like this. Making the problem visible is the first step to solving it!

PR #73112, opened 2024-06-21
ref(replay): update screen name right away for mobile replay

### LINKED ISSUE #72233: [Replay Details] Populate the screen name right away (might require new data from SDKs)
need to `urls` field to be filled out on the replay data

michellewzhang:
- closes https://github.com/getsentry/sentry/issues/72233
- this pr sets the screen name in the top right immediately, using the data from the `urls` array 

SCR-20240621-jmre

### DIFF
--- a/static/app/utils/replays/getCurrentScreenName.tsx
+++ b/static/app/utils/replays/getCurrentScreenName.tsx
@@ -1,7 +1,9 @@
 import type {BreadcrumbFrame} from 'sentry/utils/replays/types';
+import type {ReplayRecord} from 'sentry/views/replays/types';
 
 // Gets the current screen name for video/mobile replays - mirrors getCurrentUrl.tsx
 function getCurrentScreenName(
+  replayRecord: undefined | ReplayRecord,
   frames: undefined | BreadcrumbFrame[],
   currentOffsetMS: number
 ) {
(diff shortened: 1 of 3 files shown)


static/app/utils/replays/getCurrentScreenName.tsx
```diff
@@ -14,6 +16,10 @@ function getCurrentScreenName(
     return '';
   }
 
+  if ('category' in mostRecentFrame && mostRecentFrame.category === 'replay.init') {
+    return replayRecord?.urls[0] ?? '';
+  }
```
ryan953, line 21: idk if this is needed too? 

```suggestion
  if (repalyRecord && 'category' in mostRecentFrame && mostRecentFrame.category === 'replay.init') {
    return replayRecord?.urls[0] ?? '';
  }
```

ryan953 (approved):

static/app/utils/replays/getCurrentScreenName.tsx
michellewzhang: this is what's in `getCurrentUrl.tsx`, which doesn't have that check:
https://github.com/getsentry/sentry/blob/15997098796a3d3b1297227e80b6dd06ab5a461b/static/app/utils/replays/getCurrentUrl.tsx#L28-L30

ryan953: but that snippet isn't reading from `replayRecord?.urls`

There's probably never a case where `replayRecord` is undefined, and `mostRecentFrame` has some useful value, I was just thinking through the typesystem. nbd!

PR #73119, opened 2024-06-21
feat(trace-explorer): Experiment with sorting by timestamp

Zylphrex:
This is an experiment to try supporting sorting traces by timestamp.

### DIFF
--- a/src/sentry/api/endpoints/organization_traces.py
+++ b/src/sentry/api/endpoints/organization_traces.py
@@ -530,7 +530,7 @@ class TracesExecutor:
         builder, timestamp_column = self.get_traces_matching_span_conditions_query(
             params,
             snuba_params,
-            orderby=[self.sort, "trace"],
+            orderby=[self.sort],
         )
 
         executor = OrderedTracesExecutor(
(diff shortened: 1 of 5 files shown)


src/sentry/api/endpoints/organization_traces.py
```diff
@@ -495,50 +508,65 @@ def get_traces_matching_metric_conditions(
...
 
-        all_queries.append(query)
+        executor = OrderedTracesExecutor(
+            params=params,
+            snuba_params=snuba_params,
+            builder=builder,
+            limit=self.limit,
+            timestamp_column=timestamp_column,
+            max_execution_seconds=30,
+            max_parallel_queries=3,
```
k-fish, line 532: Good call on parallel to speed it up a bit, are we sure this won't spam Snuba and hit rate limits?

k-fish (approved):

src/sentry/api/endpoints/organization_traces.py
Zylphrex: 3 x 8 hour queries should stay within the rate limits of snuba but I'll confirm before merging. And I explicitly moved this to a different referrer to help alleviate this concern slightly.

PR #73120, opened 2024-06-21
feat(related_issues): Post GA clean up

armenzg:
This will require a backend PR to remove the feature flag and remove the flagr flag.

### DIFF
--- a/static/app/views/issueDetails/groupEventHeader.tsx
+++ b/static/app/views/issueDetails/groupEventHeader.tsx
@@ -9,9 +9,7 @@ import type {Project} from 'sentry/types/project';
 import {getConfigForIssueType} from 'sentry/utils/issueTypeConfig';
 import useOrganization from 'sentry/utils/useOrganization';
 import {GroupEventCarousel} from 'sentry/views/issueDetails/groupEventCarousel';
-import {TraceLink} from 'sentry/views/issueDetails/traceTimeline/traceLink';
 import {TraceTimeline} from 'sentry/views/issueDetails/traceTimeline/traceTimeline';
-import {useTraceTimelineEvents} from 'sentry/views/issueDetails/traceTimeline/useTraceTimelineEvents';
 
 type GroupEventHeaderProps = {
   event: Event;
@@ -22,25 +20,11 @@ type GroupEventHeaderProps = {
 function GroupEventHeader({event, group, project}: GroupEventHeaderProps) {
   const organization = useOrganization();
   const issueTypeConfig = getConfigForIssueType(group, group.project);
-  const isRelatedIssuesEnabled = organization.features.includes(
-    'related-issues-issue-details-page'
-  );
-  // This is also called within the TraceTimeline component but caching will save a second call
-  const {isLoading, oneOtherIssueEvent} = useTraceTimelineEvents({
-    event,
-  });
-  const showTraceLink = isRelatedIssuesEnabled && !isLoading;
 
   return (
     <StyledDataSection>
       <GroupEventCarousel group={group} event={event} projectSlug={project.slug} />
-      {/* XXX: Remove TraceLink from groupEventCarousel when we GA. */}
-      {showTraceLink && (
-        <StyledTraceLink>
-          {oneOtherIssueEvent && <span>One other issue appears in the same trace.</span>}
-          <TraceLink event={event} />
-        </StyledTraceLink>
-      )}
+      {/* This renders either the trace timeline or the trace-related issue */}
       {issueTypeConfig.traceTimeline && <TraceTimeline event={event} />}
       <StyledGlobalAppStoreConnectUpdateAlert
         project={project}
@@ -60,13 +44,6 @@ const StyledDataSection = styled(DataSection)`
   @media (min-width: ${p => p.theme.breakpoints.medium}) {
     padding: ${space(1.5)} ${space(4)} 0;
   }
-`;
-
-const StyledTraceLink = styled('div')`
-  display: flex;
-  white-space: nowrap;
-  overflow: hidden;
-  gap: ${space(0.25)};
 `;
 
 export default GroupEventHeader;
--- a/static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
+++ b/static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
@@ -20,10 +20,11 @@ describe('TraceTimeline', () => {
   const organization = OrganizationFixture({
     features: ['global-views'],
   });
+  const firstEventTimestamp = '2024-01-24T09:09:01+00:00';
   // This creates the ApiException event
   const event = EventFixture({
     // This is used to determine the presence of seconds
-    dateCreated: '2024-01-24T09:09:03+00:00',
+    dateCreated: firstEventTimestamp,
     contexts: {
       trace: {
         // This is used to determine if we should attempt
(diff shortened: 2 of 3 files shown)


static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -94,7 +108,6 @@ describe('TraceTimeline', () => {
     expect(useRouteAnalyticsParams).toHaveBeenCalledWith({
       has_related_trace_issue: false,
       trace_timeline_status: 'shown',
-      trace_timeline_two_issues: true,
```
armenzg, line 97: We're not going to report this metric anymore. It was to compare to the previous trace timeline CTR.

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -181,19 +199,17 @@ describe('TraceTimeline', () => {
     });
     MockApiClient.addMockResponse({
       url: `/organizations/${organization.slug}/events/`,
-      body: emptyBody,
+      body: discoverBody,
       match: [MockApiClient.matchQuery({dataset: 'discover', project: -1})],
     });
-    // I believe the call to projects is to determine what projects a user belongs to
```
armenzg, line 187: I was not sure when I originally added the comment.

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -205,21 +221,16 @@ describe('TraceTimeline', () => {
 
     // Test analytics
     await userEvent.click(await screen.findByText('Slow DB Query'));
-    expect(useRouteAnalyticsParams).toHaveBeenLastCalledWith({
+    expect(useRouteAnalyticsParams).toHaveBeenCalledWith({
```
armenzg, line 221: I'm making this more consistent with the other tests.

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -181,19 +199,17 @@ describe('TraceTimeline', () => {
...
-    // I believe the call to projects is to determine what projects a user belongs to
+    // Used to determine the project badge
     MockApiClient.addMockResponse({
       url: `/organizations/${organization.slug}/projects/`,
       body: [],
     });
 
     render(<TraceTimeline event={event} />, {
-      organization: OrganizationFixture({
-        features: ['related-issues-issue-details-page', 'global-views'],
```
armenzg, line 195: We don't need `related-issues-issue-details-page` since it will be the new default.

`'global-views'` is already defined in the `organization` mock, thus, I do not need to pass it in here again.

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -259,29 +268,28 @@ describe('TraceTimeline', () => {
     expect(useRouteAnalyticsParams).toHaveBeenCalledWith({
       has_related_trace_issue: false,
       trace_timeline_status: 'empty',
-      trace_timeline_two_issues: false,
     });
   });
 
   it('works for plans with no global-views feature', async () => {
+    // This test will call the endpoint without the global-views feature, thus,
+    // we will only look at the current project (project: event.projectID) instead of passing -1
```
armenzg, line 273: The two comments below were a bit unclear to me. This is a bit more elaborate and hopefully clearer to future readers.

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -259,39 +265,75 @@ describe('TraceTimeline', () => {
...
           dataset: 'issuePlatform',
-          // Since we don't have global-views, we only look at the current project
           project: event.projectID,
         }),
       ],
     });
     MockApiClient.addMockResponse({
       url: `/organizations/${organization.slug}/events/`,
-      body: emptyBody,
+      body: twoIssuesBody,
```
armenzg, line 286: We need 3 issues for the trace timeline to be rendered.

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -259,39 +265,75 @@ describe('TraceTimeline', () => {
...
     render(<TraceTimeline event={event} />, {
       organization: OrganizationFixture({
-        features: [], // No global-views feature
+        features: [],
       }),
     });
     expect(await screen.findByLabelText('Current Event')).toBeInTheDocument();
   });
+
+  it('trace-related issue works for plans with no global-views feature', async () => {
```
armenzg, line 303: This test is similar to the one above, however, it tests that trace-related works on orgs without `global-views`.

static/app/views/issueDetails/traceTimeline/traceTimeline.spec.tsx
```diff
@@ -259,39 +265,75 @@ describe('TraceTimeline', () => {
...
+      match: [
+        MockApiClient.matchQuery({
+          dataset: 'issuePlatform',
+          project: event.projectID,
+        }),
+      ],
+    });
+    MockApiClient.addMockResponse({
+      url: `/organizations/${organization.slug}/events/`,
+      body: discoverBody,
```
armenzg, line 318: Using `discoverBody` & `issuePlatformBody` will yield two issues, thus, trace-related issues will be rendered instead of the trace-timeline.

static/app/views/issueDetails/traceTimeline/traceTimeline.tsx
```diff
@@ -8,7 +8,6 @@ import {space} from 'sentry/styles/space';
 import type {Event} from 'sentry/types/event';
 import useRouteAnalyticsParams from 'sentry/utils/routeAnalytics/useRouteAnalyticsParams';
 import {useDimensions} from 'sentry/utils/useDimensions';
-import useOrganization from 'sentry/utils/useOrganization';
```
armenzg, line 11: We don't need to check for the `related-issues-issue-details-page` feature, thus, not needed.

static/app/views/issueDetails/groupEventHeader.tsx
```diff
@@ -22,25 +20,11 @@ type GroupEventHeaderProps = {
...
       <GroupEventCarousel group={group} event={event} projectSlug={project.slug} />
-      {/* XXX: Remove TraceLink from groupEventCarousel when we GA. */}
-      {showTraceLink && (
-        <StyledTraceLink>
-          {oneOtherIssueEvent && <span>One other issue appears in the same trace.</span>}
-          <TraceLink event={event} />
-        </StyledTraceLink>
-      )}
+      {/* This renders either the trace timeline or the trace-related issue */}
       {issueTypeConfig.traceTimeline && <TraceTimeline event={event} />}
```
mrduncan, line 28: If the `TraceTimeline` component can render itself conditionally based on whether the event has a trace id I think we can remove this configuration. The only issue type I'm aware of that it's disabled for is rage clicks which I don't think will ever have a trace id (and if they do, arguably it'd be nice to see the related issue/timeline anyway).

mrduncan (approved):

static/app/views/issueDetails/groupEventHeader.tsx
armenzg: Thanks for pointing that out. I will look a bit into it and report back:
```
git grep "issueTypeConfig.traceTimeline"
static/app/views/issueDetails/groupEventCarousel.tsx:        {issueTypeConfig.traceTimeline && <TraceLink event={event} />}
static/app/views/issueDetails/groupEventHeader.tsx:      {issueTypeConfig.traceTimeline && <TraceTimeline event={event} />}
```

PR #73121, opened 2024-06-21
ref: fix passing slug=None to HC slug check

asottile-sentry:
this currently produces an INFO log on every logged out request

```
16:38:59 server  | 16:38:59 [INFO] sentry.services.hybrid_cloud: Organization by slug [None] not found
```

this fixes these two type errors as well:

```
src/sentry/web/frontend/auth_login.py:509: error: Argument "slug" to "check_organization_by_slug" of "OrganizationService" has incompatible type "str | None"; expected "str"  [arg-type]
src/sentry/web/frontend/auth_login.py:568: error: Argument "slug" to "check_organization_by_slug" of "OrganizationService" has incompatible type "str | None"; expected "str"  [arg-type]
```

### DIFF
--- a/src/sentry/web/frontend/auth_login.py
+++ b/src/sentry/web/frontend/auth_login.py
@@ -504,10 +504,11 @@ class AuthLoginView(BaseView):
         """
         Returns True if the organization passed in a request exists.
         """
-        return bool(
+        return request.subdomain is not None and (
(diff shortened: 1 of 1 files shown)


joshuarli (approved):

joshuarli:
man, typing is good

PR #73123, opened 2024-06-21
nit(mes-ref): Moved Metrics to `metrics.py`

iamrajjoshi:
small nit to move the metrics key to where we are aggregating all of them. The period is also not needed at the end, (currently we would have `..`)

### DIFF
--- a/src/sentry/integrations/slack/webhooks/base.py
+++ b/src/sentry/integrations/slack/webhooks/base.py
@@ -7,6 +7,10 @@ from rest_framework.response import Response
 
 from sentry.api.base import Endpoint
 from sentry.integrations.slack.message_builder.help import SlackHelpMessageBuilder
+from sentry.integrations.slack.metrics import (
+    SLACK_WEBHOOK_DM_ENDPOINT_FAILURE_DATADOG_METRIC,
+    SLACK_WEBHOOK_DM_ENDPOINT_SUCCESS_DATADOG_METRIC,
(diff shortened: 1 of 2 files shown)


src/sentry/integrations/slack/webhooks/base.py
```diff
@@ -32,8 +36,8 @@
 class SlackDMEndpoint(Endpoint, abc.ABC):
     slack_request_class = SlackDMRequest
 
-    _METRICS_SUCCESS_KEY = "sentry.integrations.slack.dm_endpoint.success."
-    _METRIC_FAILURE_KEY = "sentry.integrations.slack.dm_endpoint.failure."
+    _METRICS_SUCCESS_KEY = SLACK_WEBHOOK_DM_ENDPOINT_SUCCESS_DATADOG_METRIC
+    _METRIC_FAILURE_KEY = SLACK_WEBHOOK_DM_ENDPOINT_FAILURE_DATADOG_METRIC
```
cathteng, line 40: if we're already importing these, do we need to use another variable for them?

iamrajjoshi: i keep it cause the name is rather long and often makes 1 line stuff multiple lines after we auto format it.

cathteng: ok

cathteng (approved):

PR #73130, opened 2024-06-21
feat(codeowners): Ignore lines with no owners, only comment

### LINKED ISSUE #72480: CODEOWNERs doesn't support excluding subdirectories
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

When parsing CODEOWNER lines, we ignore lines which don't have an owner present.

For example (borrowed from [GitHub](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file)), with the following lines:

```
# In this example, @octocat owns any file in the `/apps`
# directory in the root of your repository except for the `/apps/github`
# subdirectory, as its owners are left empty.
/apps/ @octocat
/apps/github # Skip
```

### Expected Result

The line should exclude ownership of a subdirectory.

### Actual Result

The line is ignored.

### Product Area

Issues

### Link

_No response_

### DSN

_No response_

### Version

_No response_

scttcper:
ignores codeowners lines with a comment instead of an owner
```codeowners
/apps/github # Skip
```

fixes #72480

### DIFF
--- a/src/sentry/ownership/grammar.py
+++ b/src/sentry/ownership/grammar.py
@@ -35,9 +35,12 @@ ownership_grammar = Grammar(
 
 ownership = line*
 
-line = _ (comment / rule / empty) newline?
+line = _ (comment / rule / rule_comment / empty) newline?
 
 rule = _ matcher owners
+
+# A partial rule is a rule without owners
+rule_comment = _ matcher comment?
 
 matcher      = _ matcher_tag any_identifier
 matcher_tag  = (matcher_type sep)?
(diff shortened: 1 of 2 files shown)


src/sentry/ownership/grammar.py
```diff
@@ -35,10 +35,13 @@
 
 ownership = line*
 
-line = _ (comment / rule / empty) newline?
+line = _ (comment / rule / rule_comment / empty) newline?
 
 rule = _ matcher owners
 
+# A partial rule is a rule without owners
+rule_comment = _ matcher comment?
```
mrduncan, line 43: One more edge case:

> Escaping a pattern starting with # using \ so it is treated as a pattern and not a comment

`/apps/\#1 @user` should get parsed as `/apps/#1 @user` and not as though it has a trailing comment

PR #73137, opened 2024-06-21
fix(codeowners): Do not parse inline comments as owners

scttcper:
would prevent a banner from showing when a line only has comments instead of codeowners

related to #72480

### DIFF
--- a/src/sentry/ownership/grammar.py
+++ b/src/sentry/ownership/grammar.py
@@ -345,6 +345,14 @@ def get_codeowners_path_and_owners(rule: str) -> tuple[str, Sequence[str]]:
     # Regex does a negative lookbehind for a backslash. Matches on whitespace without a preceding backslash.
     pattern = re.compile(r"(?<!\\)\s")
     path, *code_owners = (i for i in pattern.split(rule.strip()) if i)
+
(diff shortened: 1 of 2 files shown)


src/sentry/ownership/grammar.py
```diff
@@ -345,6 +345,14 @@ def get_codeowners_path_and_owners(rule: str) -> tuple[str, Sequence[str]]:
     # Regex does a negative lookbehind for a backslash. Matches on whitespace without a preceding backslash.
     pattern = re.compile(r"(?<!\\)\s")
     path, *code_owners = (i for i in pattern.split(rule.strip()) if i)
+
+    # Find index of # in code_owners, assume everything after is a comment
+    try:
+        comment_index = code_owners.index("#")
```
mrduncan, line 351: I think we need to handle escaping `#` as well with a `\` but not entirely sure

> Escaping a pattern starting with # using \ so it is treated as a pattern and not a comment

`/apps/\#1 @user` should get parsed as `/apps/#1 @user` and not as though it has a trailing comment

mrduncan: Actually, this seems fine as-is since that would not match here

mrduncan (approved):

PR #73138, opened 2024-06-22
chore(hybridcloud) Add more logging to debug features in s4s

markstory:
Our internal sentry instance is exhibiting strange behavior with entity_handler. The entity_handler is registered during startup but is then lost when checking features.

Logging the manager instances, and init should let us see if the memory address of the manager changes between a pod starting and handling requests.

### DIFF
--- a/src/sentry/api/serializers/models/organization.py
+++ b/src/sentry/api/serializers/models/organization.py
@@ -260,6 +260,8 @@ class OrganizationSerializer(Serializer):
                 extra={
                     "org_features": org_features,
                     "user.id": user.id if not user.is_anonymous else None,
+                    "manager": features.default_manager,
+                    "entity_handler": type(features.default_manager._entity_handler),
                 },
             )
         with sentry_sdk.start_span(op="features.check", description="check batch features"):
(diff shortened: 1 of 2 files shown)


src/sentry/api/serializers/models/organization.py
```diff
@@ -260,6 +260,8 @@ def get_feature_set(
                 extra={
                     "org_features": org_features,
                     "user.id": user.id if not user.is_anonymous else None,
+                    "manager": features.default_manager,
```
corps, line 263: since we're throwing the kitchen sink for debugging, why not a good ol `id(features.default_manager)` in?

corps (approved):

src/sentry/api/serializers/models/organization.py
corps: Or I guess the default __str__ has a memory identifier?

markstory: Yeah, default str()/repr() includes a hex address/id of the object.

markstory:
I think getsentry/getsentry#14413 might be the solution to our problem, and I'd like to try that instead of more logging.

PR #73139, opened 2024-06-22
feat(uptime): Implement hostname detection and ranking

wedamija:
This implements the first half of automatic hostname detection and monitoring. Here we examine incoming events in `post_process_group` and if they have a valid hostname in the request url we store them in our ranking system.

The ranking system keeps track of how many total events with a valid hostname we've seen on the project, as well as the top hostnames for the project, ranked by times seen.

Details about each are included in comments in the code.

### DIFF
--- a/src/sentry/conf/server.py
+++ b/src/sentry/conf/server.py
@@ -136,6 +136,7 @@ SENTRY_METRIC_META_REDIS_CLUSTER = "default"
 SENTRY_ESCALATION_THRESHOLDS_REDIS_CLUSTER = "default"
 SENTRY_SPAN_BUFFER_CLUSTER = "default"
 SENTRY_ASSEMBLE_CLUSTER = "default"
+SENTRY_UPTIME_DETECTOR_CLUSTER = "default"
 
 # Hosts that are allowed to use system token authentication.
 # http://en.wikipedia.org/wiki/Reserved_IP_addresses
--- a/src/sentry/features/temporary.py
+++ b/src/sentry/features/temporary.py
@@ -458,6 +458,8 @@ def register_temporary_features(manager: FeatureManager):
     manager.add("organizations:transaction-name-normalize", OrganizationFeature, FeatureHandlerStrategy.INTERNAL, default=True)
     # Sanitize transaction names in the ingestion pipeline. # Deprecated
     manager.add("organizations:transaction-name-sanitization", OrganizationFeature, FeatureHandlerStrategy.INTERNAL)
+    # Enables automatic hostname detection in uptime
+    manager.add("organizations:uptime-automatic-hostname-detection", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE)
     manager.add("organizations:use-metrics-layer", OrganizationFeature, FeatureHandlerStrategy.REMOTE)
     # Enable User Feedback v2 ingest
     manager.add("organizations:user-feedback-ingest", OrganizationFeature, FeatureHandlerStrategy.INTERNAL)
--- a/src/sentry/tasks/post_process.py
+++ b/src/sentry/tasks/post_process.py
@@ -26,6 +26,7 @@ from sentry.signals import event_processed, issue_unignored, transaction_process
 from sentry.silo.base import SiloMode
 from sentry.tasks.base import instrumented_task
 from sentry.types.group import GroupSubStatus
+from sentry.uptime.detectors.detector import detect_hostname_for_project
 from sentry.utils import json, metrics
 from sentry.utils.cache import cache
 from sentry.utils.event_frames import get_sdk_name
@@ -1512,6 +1513,11 @@ def detect_new_escalation(job: PostProcessJob):
         return
 
 
+def detect_hostnames_for_uptime(job: PostProcessJob):
+    url = get_path(job.event.data, "request", "url")
+    detect_hostname_for_project(job.event.project, url)
+
+
 GROUP_CATEGORY_POST_PROCESS_PIPELINE = {
     GroupCategory.ERROR: [
         _capture_group_stats,
@@ -1533,6 +1539,7 @@ GROUP_CATEGORY_POST_PROCESS_PIPELINE = {
         sdk_crash_monitoring,
         process_replay_link,
         link_event_to_user_report,
+        detect_hostnames_for_uptime,
     ],
     GroupCategory.FEEDBACK: [
         feedback_filter_decorator(process_snoozes),
--- a/src/sentry/tasks/post_process.py
+++ b/src/sentry/tasks/post_process.py
@@ -26,7 +26,7 @@ from sentry.signals import event_processed, issue_unignored, transaction_process
 from sentry.silo.base import SiloMode
 from sentry.tasks.base import instrumented_task
 from sentry.types.group import GroupSubStatus
-from sentry.uptime.detectors.detector import detect_hostname_for_project
+from sentry.uptime.detectors.detector import detect_base_url_for_project
 from sentry.utils import json, metrics
 from sentry.utils.cache import cache
 from sentry.utils.event_frames import get_sdk_name
@@ -1513,9 +1513,9 @@ def detect_new_escalation(job: PostProcessJob):
         return
 
 
-def detect_hostnames_for_uptime(job: PostProcessJob):
+def detect_base_urls_for_uptime(job: PostProcessJob):
     url = get_path(job["event"].data, "request", "url")
-    detect_hostname_for_project(job["event"].project, url)
+    detect_base_url_for_project(job["event"].project, url)
 
 
 GROUP_CATEGORY_POST_PROCESS_PIPELINE = {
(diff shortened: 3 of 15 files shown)


src/sentry/features/temporary.py
```diff
@@ -458,6 +458,8 @@ def register_temporary_features(manager: FeatureManager):
     manager.add("organizations:transaction-name-normalize", OrganizationFeature, FeatureHandlerStrategy.INTERNAL, default=True)
     # Sanitize transaction names in the ingestion pipeline. # Deprecated
     manager.add("organizations:transaction-name-sanitization", OrganizationFeature, FeatureHandlerStrategy.INTERNAL)
+    # Enables automatic hostname detection in uptime
+    manager.add("organizations:uptime-automatic-hostname-detection", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE)
```
wedamija, line 462: We'll add a separate feature flag for the actual creation, so potentially we can be detecting for all projects but only creating on projects we opt in.

src/sentry/tasks/post_process.py
```diff
@@ -1512,6 +1513,11 @@ def detect_new_escalation(job: PostProcessJob):
         return
 
 
+def detect_hostnames_for_uptime(job: PostProcessJob):
+    url = get_path(job.event.data, "request", "url")
+    detect_hostname_for_project(job.event.project, url)
```
wedamija, line 1518: This is pretty fast so I haven't made it a task, but we could easily split it out into a task since we just need the url and project id so we can fetch the project.

src/sentry/uptime/detectors/ranking.py
```diff
@@ -0,0 +1,76 @@
...
+    project_incr_result = pipeline.execute()[0]
+    if project_incr_result == 1:
+        # We don't want to constantly set expire on these rows to avoid load on redis.
+        # Instead we just set it when we increment the project for the first time, so
+        # that we know it's when we create the zset.
+        pipeline.expire(rank_key, KEY_EXPIRY)
+        # We don't know if this has been expired yet, but doing it once per project
+        # shouldn't be too heavy
+        pipeline.expire(bucket_key, KEY_EXPIRY)
+        pipeline.execute()
```
wedamija, line 67: I'm trying to avoid a lot of commands here since this will be reasonably high volume

src/sentry/uptime/detectors/ranking.py
```diff
@@ -0,0 +1,76 @@
...
+RANKED_TRIM_CHANCE = 0.01
+# How many urls should we trim the ranked list to?
+RANKED_MAX_SIZE = 20
+# Expiry we should set on all keys. Set the expiry to twice the flush frequency
+# so that we definitely have time to process.
+KEY_EXPIRY = PROJECT_FLUSH_FREQUENCY * 2
+
+
+def _get_cluster() -> RedisCluster | StrictRedis[str]:
+    return redis.redis_clusters.get(settings.SENTRY_UPTIME_DETECTOR_CLUSTER)
```
wedamija, line 30: Oops I should define this

wedamija (commented):
Still need to add tests but I think this is code complete

src/sentry/uptime/detectors/detector.py
```diff
@@ -0,0 +1,21 @@
...
+from sentry import features
+from sentry.uptime.detectors.hostname_extraction import extract_hostname_from_url
+from sentry.uptime.detectors.ranking import add_hostname_to_rank
+
+if TYPE_CHECKING:
+    from sentry.models.project import Project
+
+
+def detect_hostname_for_project(project: Project, url: str) -> str | None:
+    if not features.has("organizations:uptime-automatic-hostname-detection", project.organization):
```
evanpurkhiser, line 14: Flagpole is fast for ingestion stuff now yeah?

src/sentry/uptime/detectors/hostname_extraction.py
```diff
@@ -0,0 +1,49 @@
...
+    pass
+
+url_validator = URLValidator()
+
+# Configure TLDExtract to not make external calls and just use a local snapshot of tld data
+extractor = TLDExtract(
+    cache_dir=None,
+    suffix_list_urls=(),
+    fallback_to_snapshot=True,
+)
```
evanpurkhiser, line 20: Should we expand on this comment at all with _why_ we don't want it to make external requests?

src/sentry/uptime/detectors/detector.py
evanpurkhiser: `project.organization` I assume is already cached

src/sentry/uptime/detectors/ranking.py
```diff
@@ -0,0 +1,78 @@
...
+# Expiry we should set on all keys. Set the expiry to twice the flush frequency
+# so that we definitely have time to process.
+KEY_EXPIRY = PROJECT_FLUSH_FREQUENCY * 2
+
+
+def _get_cluster() -> RedisCluster | StrictRedis:
+    return redis.redis_clusters.get(settings.SENTRY_UPTIME_DETECTOR_CLUSTER)
+
+
+def add_hostname_to_rank(project: Project, hostname: str):
```
evanpurkhiser, line 35: What is the reason we rank them?

src/sentry/uptime/detectors/detector.py
wedamija: Yep, it's all backed by options automator

src/sentry/uptime/detectors/ranking.py
wedamija: We want to find the most common hostnames in a project and attempt to monitor the top one.

evanpurkhiser: Ahh right 👍

src/sentry/conf/server.py
```diff
@@ -136,6 +136,7 @@ def env(
 SENTRY_ESCALATION_THRESHOLDS_REDIS_CLUSTER = "default"
 SENTRY_SPAN_BUFFER_CLUSTER = "default"
 SENTRY_ASSEMBLE_CLUSTER = "default"
+SENTRY_UPTIME_DETECTOR_CLUSTER = "default"
```
evanpurkhiser, line 139: This will need to be adjusted in getsentry before we turn this on yeah? Unless we want to bring sentry down

src/sentry/uptime/detectors/url_extraction.py
```diff
@@ -0,0 +1,53 @@
...
+)
+
+
+def extract_base_url(url: str | None) -> str | None:
+    """
+    Examines a url string and returns a url containing just the
+    protocol and fully qualified domain name from it.
+    Filters out invalid TLDs and domains with ip addresses. The
+    goal is to reduce the set of hostnames down so that we only test
+    hostnames that might actually return a valid response.
```
evanpurkhiser, line 32: Might be useful to toss in some examples. I don't fully understand what the extractor does.

Does it canonicalize URLs like `sentry.sentry.io` to just `sentry.io`?

tests/sentry/uptime/detectors/test_url_extraction.py
```diff
@@ -0,0 +1,19 @@
...
+        self.run_test("", None)
+        self.run_test("192.168.0.1", None)
+        self.run_test("https://192.168.0.1", None)
+        self.run_test("https://www.sentry.io/", "https://www.sentry.io")
+        self.run_test("https://www.sentry.io/some/page", "https://www.sentry.io")
+        self.run_test("www.sentry.io", None)
+        self.run_test("www.sentry.io/something", None)
+        self.run_test("https://www.sentry.nope/", None)
+        self.run_test("https://www.sentry.nope/some/page", None)
+        self.run_test("https://www.ato.gov.au/", "https://www.ato.gov.au")
```
evanpurkhiser, line 19: Probably could drop the class based test and use the pytest parameterize thing like here https://github.com/getsentry/sentry/blob/495754e103f72dab80b8474640f38e6f75c12de6/tests/sentry/security/test_csp.py#L6-L20

evanpurkhiser (approved):

src/sentry/uptime/detectors/url_extraction.py
wedamija: It takes a url like `http://sentry.io/hello` and returns the protocol + fqdn, which is `http://sentry.io`. So `http://sentry.sentry.io/issues` would be `http://sentry.sentry.io`

evanpurkhiser: ah ok, so it isn't doing anything with https://publicsuffix.org/ yet

wedamija: It will validate whether the tld is valid as part of it, but we're not doing anything to extract specific hostnames yet.

src/sentry/conf/server.py
wedamija: Yep!

PR #73140, opened 2024-06-22
chore(seer grouping): Remove `seer.similar_issue_request.parent_issue` metric

lobsterkatie:
The original purpose of the `seer.similar_issue_request.parent_issue` metric we collect in `get_similarity_data_from_seer` was to track how often Seer was recommending groups which no longer exist - with the idea that that was a thing which was going to happen sometimes. Once @jangjodi's deletion work is done, though, and Seer records get deleted as soon as their corresponding `GroupHash` records disappear, it shouldn't ever happen - and if it does, we want to be able to track it down, because it will signal a real problem.

In anticipation of that shift, this PR makes the following changes:

- The existing metric is removed. We do still want to keep track of this, but going forward we'll do so as part of a new metric (to be added in an upcoming PR) tracking all of the different errors which can happen in the `get_similarity_data_from_seer`. 
- In the case in which we do run into a missing parent group, we now log the project id and missing hash, the better to be able to debug the situation.

### DIFF
--- a/src/sentry/seer/similarity/similar_issues.py
+++ b/src/sentry/seer/similarity/similar_issues.py
@@ -11,7 +11,7 @@ from sentry.seer.similarity.types import (
     SimilarGroupNotFoundError,
     SimilarIssuesEmbeddingsRequest,
 )
-from sentry.utils import json, metrics
+from sentry.utils import json
 from sentry.utils.json import JSONDecodeError
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


JoshFerge (approved):

src/sentry/seer/similarity/similar_issues.py
```diff
@@ -91,7 +87,14 @@ def get_similarity_data_from_seer(
                 },
             )
         except SimilarGroupNotFoundError:
-            metrics.incr("seer.similar_issue_request.parent_issue", tags={"outcome": "not_found"})
```
corps, line 94: For reference, how often does this occur in practice?

lobsterkatie: > Once @jangjodi's deletion work is done, though, and Seer records get deleted as soon as their corresponding GroupHash records disappear, it shouldn't ever happen 

Very soon, theoretically never.

PR #73141, opened 2024-06-22
ref(seer grouping): Refactor `get_similarity_data_from_seer` and associated tests

lobsterkatie:
This is a bunch of small refactors to `get_similarity_data_from_seer` and its associated tests, pulled out of an upcoming PR in order to make it easier to review.

Changes to `get_similarity_data_from_seer`:

- Dig into `response.data` earlier.
- Handle case of there being no similar groups separately.

Changes to tests:

- Pull common operations into `setUp` and a new `_make_response` helper.
- Add new tests and/or new cases to existing tests to cover the following missing scenarios: 
  - similar but not matching groups being found
  - no similar groups being found
  - `response.data` being `None`
  - results missing required data
  - a non-existent parent hash being returned.

### DIFF
--- a/src/sentry/seer/similarity/similar_issues.py
+++ b/src/sentry/seer/similarity/similar_issues.py
@@ -53,7 +53,7 @@ def get_similarity_data_from_seer(
         return []
 
     try:
-        response_data = json.loads(response.data.decode("utf-8"))
+        response_data = json.loads(response.data.decode("utf-8")).get("responses")
     except (
         AttributeError,  # caused by a response with no data and therefore no `.decode` method
         UnicodeError,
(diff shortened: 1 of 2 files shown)


src/sentry/seer/similarity/similar_issues.py
```diff
@@ -69,15 +69,17 @@ def get_similarity_data_from_seer(
         )
         return []
 
-    normalized = []
+    if not response_data:
+        return []
```
JoshFerge, line 73: should we log if this happens? doesn't seem like it should in theory.

JoshFerge (approved):

src/sentry/seer/similarity/similar_issues.py
lobsterkatie: It'll be included in the metric, but this is a totally normal case - it will happen whenever Seer doesn't find any groups that are matching or even close enough to count as similar.

(You may be being thrown off by the change to dig out the `responses` data earlier, making `response_data` now the value of `response.data["responses"]`.)

PR #73142, opened 2024-06-22
feat(seer grouping): Add `seer.similar_issues_request` metric

lobsterkatie:
This adds a new metric, `seer.similar_issues_request`, to track the results of our calls to the seer similarity endpoint. In addition to counting the calls themselves, the new metric tracks the response status code, the result of the call (matching group, similar groups, no similar groups, error), and in the erroring case, what type of error occurred. 

Notes:

- For now the sample rate is set to 100%, so that we can get some decent signal even while the feature is only rolled out to a few customers. Once we GA, we'll probably want to change it to a lower rate.

- In keeping with our handling of non-200 responses, redirects are considered errors, because they imply that a change has been made on the Seer side and we haven't correctly updated the Sentry code to match. They are however tracked separately from other request errors.

### DIFF
--- a/src/sentry/seer/similarity/similar_issues.py
+++ b/src/sentry/seer/similarity/similar_issues.py
@@ -11,10 +11,15 @@ from sentry.seer.similarity.types import (
     SimilarGroupNotFoundError,
     SimilarIssuesEmbeddingsRequest,
 )
-from sentry.utils import json
+from sentry.utils import json, metrics
 from sentry.utils.json import JSONDecodeError
 
 logger = logging.getLogger(__name__)
+
+# TODO: Keeping this at a 100% sample rate for now in order to get good signal as we're rolling out
+# and calls are still comparatively rare. Once traffic gets heavy enough, we should probably ramp
+# this down.
+SIMILARITY_REQUEST_METRIC_SAMPLE_RATE = 1.0
 
 
 seer_grouping_connection_pool = connection_from_url(
@@ -39,6 +44,8 @@ def get_similarity_data_from_seer(
         ),
     )
 
+    metric_tags: dict[str, str | int] = {"response_status": response.status}
+
     if response.status > 200:
         redirect = response.get_redirect_location()
         if redirect:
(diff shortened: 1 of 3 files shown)


src/sentry/seer/similarity/similar_issues.py
```diff
@@ -67,20 +85,47 @@ def get_similarity_data_from_seer(
                 "response_code": response.status,
             },
         )
+        metrics.incr(
+            "seer.similar_issues_request",
+            sample_rate=SIMILARITY_REQUEST_METRIC_SAMPLE_RATE,
+            tags={**metric_tags, "outcome": "error", "error": type(e).__name__},
```
JoshFerge, line 91: let's not include the error type here as a metric tag, this could be too high cardinality.

JoshFerge (commented):
just the one note about the error type

src/sentry/seer/similarity/similar_issues.py
lobsterkatie: Given that the except block isn't a generic `except Exception`, won't it only ever be one of `AttributeError`, `UnicodeError`, and `JSONDecodeError`? (I used the name attribute just because I didn't want to have to write a separate `except` clause for each of the three error types, only for the purpose of hard-coding the tag value.)

JoshFerge: sure. we capture the sentry error so i also just think this isn't a very important tag to add

JoshFerge (approved):

src/sentry/seer/similarity/similar_issues.py
lobsterkatie: We're adding it for every other thing than can go wrong, though, so we need _some_ value there, to distinguish it from getting a bad response status, for example. If we're worried about cardinality, I could make it a more general `DataError` to cover all three cases with one tag, but in point of fact, once we fix Seer to send error JSON rather than the generic Flask error page HTML (causing the JSON decode error), we shouldn't really ever see any of these errors, so I think we're talking more theory than practice here in any case.

PR #73146, opened 2024-06-24
ref(settings-org-teams): Convert to functional component

priscilawebdev:
### Goal

### Note

- Changes were manually tested locally

### DIFF
--- a/static/app/views/settings/organizationTeams/index.tsx
+++ b/static/app/views/settings/organizationTeams/index.tsx
@@ -1,40 +1,35 @@
+import {useCallback, useEffect, useState} from 'react';
 import type {RouteComponentProps} from 'react-router';
 
 import {loadStats} from 'sentry/actionCreators/projects';
-import type {Client} from 'sentry/api';
+import LoadingError from 'sentry/components/loadingError';
+import LoadingIndicator from 'sentry/components/loadingIndicator';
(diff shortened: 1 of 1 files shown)


static/app/views/settings/organizationTeams/index.tsx
```diff
@@ -1,83 +1,82 @@
...
+    },
+    [requestList]
+  );
 
-  renderBody() {
-    const {organization} = this.props;
+  // can an organization be null here?
+  if (!organization) {
+    return null;
+  }
```
ArthurKnaus, line 60: For this not to break, you would need to pass `{allowNull: true}` to `useOrganization()`.

static/app/views/settings/organizationTeams/index.tsx
```diff
@@ -1,83 +1,82 @@
...
+      if (isApproved && requestToRemove) {
+        const team = requestToRemove.team;
+        TeamStore.onUpdateSuccess(team.slug, {
+          ...team,
+          memberCount: team.memberCount + 1,
+        });
+      }
+    },
+    [requestList]
+  );
```
ArthurKnaus, line 55: You could update the query cache here, instead of a separate `useState()`.

ArthurKnaus (approved):

PR #73147, opened 2024-06-24
ref(metrics): renames operation to aggregation

obostjancic:
- renames `op` to `aggregation` in metric types to distinguish it form `span.op`


static/app/utils/metrics/dashboardImport.tsx
```diff
@@ -315,9 +315,11 @@ export class WidgetParser {
     const appliedFunctionMatch = str.match(/\.(\w+)\(\)/);
     const appliedFunction = appliedFunctionMatch ? appliedFunctionMatch[1] : undefined;
 
-    if (!op) {
-      this.errors.push(`widget.request.query - could not parse op: ${str}, assuming sum`);
-      op = 'sum';
+    if (!aggregation) {
+      this.errors.push(
+        `widget.request.query - could not parse aggregation: ${str}, assuming sum`
+      );
```
priscilawebdev, line 321: are we displaying these errors somewhere to clients? shouldn't we use the translation function here?

priscilawebdev:
since metrics like the average and p90 are categorized under "operations," it makes sense to rename them as "aggregates." 

The code looks good to me, but it is hard to review due to so many needed changes. I will leave it to those more familiar with this part of the code to approve.

static/app/utils/metrics/dashboardImport.tsx
obostjancic: yes, but they are visible only to sentry staff since this is an internal feature

priscilawebdev: ok we just have to remember to update it when the feature is available to clients

vgrozdanic (approved):
Let's 🐑 this

PR #73176, opened 2024-06-24
feat[apis]: Pass frontend to Snuba - phase 1

sentaur-athena:
EDIT: updated boolean is_frontend with query_source based on feedback
---

For better debugability we're passing a new field to snuba while querying: `is_frontend`. I decided the field is a better fit for request header since that's how we pass the `referrer` in the Snuba request and `is_frontend` is kind of a break down of referrer. 

This PR right now only apply the changes to one endpoint, putting it up for discussion before adding it everywhere else.

### DIFF
--- a/src/sentry/snuba/query_sources.py
+++ b/src/sentry/snuba/query_sources.py
@@ -1,0 +1,6 @@
+from enum import Enum
+
+
+class QuerySource(Enum):
+    FRONTEND = "frontend"
+    API = "api"
--- a/src/sentry/utils/snuba.py
+++ b/src/sentry/utils/snuba.py
@@ -35,6 +35,7 @@ from sentry.models.releases.release_project import ReleaseProject
 from sentry.net.http import connection_from_url
 from sentry.snuba.dataset import Dataset
 from sentry.snuba.events import Columns
+from sentry.snuba.query_sources import QuerySource
 from sentry.snuba.referrer import validate_referrer
 from sentry.utils import json, metrics
 from sentry.utils.dates import outside_retention_with_modified_start
@@ -856,8 +857,8 @@ ResultSet = list[Mapping[str, Any]]
 def raw_snql_query(
     request: Request,
     referrer: str | None = None,
-    is_frontend: (
-        bool | None
+    query_source: (
+        QuerySource | None
     ) = None,  # TODO: @athena Make this field required after updated all the callsites
     use_cache: bool = False,
 ) -> Mapping[str, Any]:
@@ -868,15 +869,15 @@ def raw_snql_query(
     # other functions do here. It does not add any automatic conditions, format
     # results, nothing. Use at your own risk.
     return bulk_snuba_queries(
-        requests=[request], referrer=referrer, is_frontend=is_frontend, use_cache=use_cache
+        requests=[request], referrer=referrer, query_source=query_source, use_cache=use_cache
     )[0]
 
 
 def bulk_snuba_queries(
     requests: list[Request],
     referrer: str | None = None,
-    is_frontend: (
-        bool | None
+    query_source: (
+        QuerySource | None
     ) = None,  # TODO: @athena Make this field required after updated all the callsites
     use_cache: bool = False,
 ) -> ResultSet:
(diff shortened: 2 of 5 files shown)


src/sentry/utils/snuba.py
```diff
@@ -923,12 +933,17 @@ def get_cache_key(query: SnubaQuery) -> str:
...
+        bool | None
+    ) = None,  # TODO: @athena Make this field required after updated all the callsites
     use_cache: bool | None = False,
 ) -> ResultSet:
     headers = {}
     validate_referrer(referrer)
     if referrer:
         headers["referer"] = referrer
+    if is_frontend is not None:
+        headers["is_frontend"] = str(is_frontend)
```
evanh, line 946:
```suggestion
        headers["is-frontend"] = str(is_frontend)
```

Small nit.

evanh (commented):
I like this idea, I think it would be helpful.

Is the Sentry rate limiter already using this middleware?

Zylphrex:
Instead of a boolean `is-frontend` header, what if it's called something like `query-source` and can take on values like
- `browser` for ones initiated by the frontend
- `api` for ones initiated by the api
- `job` for ones initiated by celery jobs in sentry
- etc

sentaur-athena:
@evanh which middleware are you talking about? The API rate limiting is not directly related to this PR, this is milestone 1 of the [tech spec](https://www.notion.so/sentry/Ratelimitting-Clarity-Internal-and-External-229b6897639841b683a79e3a6e16a7ea) here for more details.

sentaur-athena:
@Zylphrex shouldn't referrer be different for each of these? My thinking is that if the caller is different the referrer has to be different. I can only think of 1 place where the code is shared -> browser vs. backend API call so referrer will be the same and this flag is needed. I'm not familiar with all snuba calls though so open to feedback.

Zylphrex:
> @Zylphrex shouldn't referrer be different for each of these? My thinking is that if the caller is different the referrer has to be different. I can only think of 1 place where the code is shared -> browser vs. backend API call so referrer will be the same and this flag is needed. I'm not familiar with all snuba calls though so open to feedback.

For most referrers, this is true but there are some cases where this assumption doesn't hold. For example, the `/api/0/organizations/{organization_id_or_slug}/events/` endpoint can be used by the frontend, directly as an API by a user, and I've seen some backend code that uses the endpoint as well. This endpoint uses a default referrer when there isn't a valid one specified so everything gets grouped into one. So it can be helpful to be able to distinguish between the various sources.

sentaur-athena:
@Zylphrex, I think the issue you're talking about is that:
> This endpoint uses a default referrer when there isn't a valid one specified so everything gets grouped into one

I'm trying to go through all APIs that have snuba calls in them and fix the shared referrer as well. I'll look into the one you mentioned. Thanks for the context.

volokluev (commented):
The `is_frontend` should go into the `tenant_ids` of the snuba Request. That's what we use to make capacity allocation decisions

evanh:
> @evanh which middleware are you talking about? The API rate limiting is not directly related to this PR, this is milestone 1 of the [tech spec](https://www.notion.so/sentry/Ratelimitting-Clarity-Internal-and-External-229b6897639841b683a79e3a6e16a7ea) here for more details.

Is the Sentry API rate limiter also going to make use of this `is-frontend` field? So we can have different limits on non-frontend requests?

> I'm trying to go through all APIs that have snuba calls in them and fix the shared referrer as well. I'll look into the one you mentioned. Thanks for the context.

There are also shared modules that get called by many different endpoints. `[tsdb](https://github.com/getsentry/sentry/tree/master/src/sentry/tsdb)` is one example.

sentaur-athena:
@volokluev I'll make that change. It's not needed in the header at all then?

sentaur-athena:
@evanh 
> Is the Sentry API rate limiter also going to make use of this is-frontend field? So we can have different limits on non-frontend requests?

No, this is only a clarity/monitoring attempt so in case of incident we know what's happening. The is_frontend flag is not super accurate and we shouldn't use it to determine logic. Basically no matter how we define it a direct API call can mimic the behavior and pretend to be a frontend call.

src/sentry/snuba/query_sources.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
+
+
+class QuerySource(Enum):
```
sentaur-athena, line 4: will add more as I discover

volokluev (approved):

PR #73178, opened 2024-06-24
feat(metrics): allow span attribute extraction without condition

### LINKED ISSUE #73151: extraction rules without filters are not generated

obostjancic:
- closes: #73151


tests/sentry/relay/config/test_metric_extraction.py
```diff
@@ -2174,5 +2180,12 @@ def test_get_span_attribute_metrics(default_project: Project) -> None:
                     {"field": "span.data.bar", "key": "bar"},
                     {"field": "span.data.foo", "key": "foo"},
                 ],
-            }
+            },
+            {
+                "category": "span",
+                "condition": None,
```
shellmayr, line 2186: Would it make sense to add a test for something like empty-string or \s? A condition that doesn't do anything but is not None may use some different execution paths.

shellmayr (approved):

vgrozdanic (approved):

tests/sentry/relay/config/test_metric_extraction.py
obostjancic: [c81910b](https://github.com/getsentry/sentry/pull/73178/commits/c81910ba0b3e546cb109054f1df9f4adaee05d8d)

PR #73180, opened 2024-06-24
ref(metrics) Move default per page limit to 1000

### LINKED ISSUE #73159: Increase default perPage for /metrics/extraction-rules to global max

vgrozdanic:
Previous default per page was 25 items, now default per page is set to 1000
Closes https://github.com/getsentry/sentry/issues/73159

### DIFF
--- a/src/sentry/api/endpoints/project_metrics_extraction_rules.py
+++ b/src/sentry/api/endpoints/project_metrics_extraction_rules.py
@@ -71,7 +71,9 @@ class ProjectMetricsExtractionRulesEndpoint(ProjectEndpoint):
             on_results=lambda x: serialize(
                 x, user=request.user, serializer=MetricsExtractionRuleSerializer()
             ),
-            default_per_page=25,
+            default_per_page=1000,
+            max_per_page=1000,
+            max_limit=1000,  # overrides default max_limit of 100 when creating paginator object
         )
 
     def post(self, request: Request, project: Project) -> Response:
--- a/tests/sentry/api/endpoints/test_project_metrics_extraction_rules.py
+++ b/tests/sentry/api/endpoints/test_project_metrics_extraction_rules.py
@@ -409,12 +409,12 @@ class ProjectMetricsExtractionEndpointTestCase(APITestCase):
         json_payload = {
             "metricsExtractionRules": [
                 {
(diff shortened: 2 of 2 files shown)


src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -71,7 +71,9 @@ def get(self, request: Request, project: Project) -> Response:
             on_results=lambda x: serialize(
                 x, user=request.user, serializer=MetricsExtractionRuleSerializer()
             ),
-            default_per_page=25,
+            default_per_page=1000,
+            max_per_page=1000,
+            max_limit=1000,  # overrides default max_limit of 100 when creating paginator object
```
obostjancic, line 76: This is ok for now, but I am pretty sure we will need to limit the number of extraction rules per project, as they have a big impact on the size of project configs sent to relay.

tests/sentry/api/endpoints/test_project_metrics_extraction_rules.py
```diff
@@ -409,12 +409,12 @@ def test_get_pagination(self):
...
             "metricsExtractionRules": [
                 {
-                    "spanAttribute": f"count_clicks_{i:02d}",
+                    "spanAttribute": f"count_clicks_{i:04d}",
                     "type": "c",
                     "unit": "none",
                     "tags": ["tag1", "tag2", "tag3"],
                 }
-                for i in range(0, 60)
+                for i in range(0, 2050)
```
obostjancic, line 417: can you check if there is a noticeable performance impact of this change. Our tests are anyways slow so i would like to avoid this getting even slower

vgrozdanic: There is, I measured locally a few times, there is around 0.5s increase in test execution time (from 9.3s to 9.8s), but the main reason for that is not due to this iteration, but due to the fact that we are now retuning 1000 items instead of 25 from the same API endpoint, so the fetching takes more time.

I have tried this loop with the test case when there is default per page limit of 25 and there is no significant difference in execution time.

obostjancic: ok! thanks!

obostjancic (approved):

shellmayr (approved):

PR #73191, opened 2024-06-24
feat(metrics) Reenable percentiles option

### LINKED ISSUE #73157: Re-enable percentiles
They need to get returned from /metrics/meta endpoint and frontend will just work(TM).

vgrozdanic:
Percentile option is reenabled by default.
Closes https://github.com/getsentry/sentry/issues/73157

This PR will be merged few hours after the percentiles are enabled on the storage and relay side


tests/sentry/api/endpoints/test_organization_details.py
```diff
@@ -449,7 +449,7 @@ def test_various_options(self, mock_get_repositories):
             "genAIConsent": True,
             "issueAlertsThreadFlag": False,
             "metricAlertsThreadFlag": False,
-            "metricsActivatePercentiles": True,
+            "metricsActivatePercentiles": False,
```
vgrozdanic, line 452: `metricsActivatePercentiles` are now by default true, so this test checks if it is possible to update them to non-default value (to false) and if the audit log will contain record of that update. Audit log entry check can be found on the line 521 in this file

priscilawebdev (approved):
lets 🛳️ it

PR #73197, opened 2024-06-24
feat(insights): return has_insights_xx flag in project serializers

KevinL10:
Adds `has_insight_xx` flags for each of the insight modules to project serializers.

Continuation of https://github.com/getsentry/sentry/pull/73028, where we are replacing an expensive 90 day look-back with a single flag to track whether data has been sent for each insight module.


iamrajjoshi (commented):
If we are updating the api examples, should we also update the api documentation? Or do we not want to explicitly document these flags? If so, do we really have to update the examples?

KevinL10:
Thanks for the catch. Will update the API documentation.

KevinL10:
Actually, I don't think we need to update anything on the docs side. I'm looking at the other hasXXX flags and they are not explicitly specified.

mjq (approved):
I think it's probably fine to leave these undocumented, as they don't have much/any use outside of Sentry. We can always loop back and add them if this changes.

PR #73203, opened 2024-06-24
chore(insights): Removes deprecated FID code paths from the web vitals module

edwardgou-sentry:
Deletes all instances of FID from webvitals code, since they have been deprecated for a while and those code paths are no longer used.


0Calories (approved):
LGTM

PR #73210, opened 2024-06-24
fix(feedback): Add special cases for apple getting started docs

scttcper:
fixes i think most of JAVASCRIPT-2TM9


michellewzhang (approved):
thanks for fixing!

vbro (approved):

PR #73211, opened 2024-06-24
ref: fix typing stronglist after HC deletions

asottile-sentry:
I'll need to come up with a better way to enforce this for deletions, but for now I'm going to fix the list


pyproject.toml
```diff
@@ -590,14 +590,12 @@ module = [
     "sentry.services.hybrid_cloud.organization_mapping.*",
     "sentry.services.hybrid_cloud.organizationmember_mapping.*",
     "sentry.services.hybrid_cloud.orgauthtoken.*",
-    "sentry.services.hybrid_cloud.pagination",
```
markstory, line 593: Did we want to add the new module locations to the stronglist?

asottile-sentry: ideally it would be `sentry.hybridcloud.*` but it's not passing at the moment

markstory: I'll get the issues there solved and get it added to the stronglist.

markstory (approved):

PR #73216, opened 2024-06-24
ref(slack): Refactored SlackCommandsEndpoint and Link Identity View

iamrajjoshi:

Addresses `bot-6` task.

Added logging and metrics, not much else to clean up here

### DIFF
--- a/src/sentry/integrations/slack/views/link_team.py
+++ b/src/sentry/integrations/slack/views/link_team.py
@@ -234,7 +234,7 @@ class SlackLinkTeamView(BaseView):
             "organizations:team-workflow-notifications", team.organization
         )
         # Turn on notifications for all of a team's projects.
-        # TODO: Remove this once the flag is removed
+        # TODO(jangjodi): Remove this once the flag is removed
         if not has_team_workflow:
             notifications_service.enable_all_settings_for_provider(
                 external_provider=ExternalProviderEnum.SLACK,
--- a/src/sentry/integrations/slack/views/link_team.py
+++ b/src/sentry/integrations/slack/views/link_team.py
@@ -26,7 +26,6 @@ from sentry.notifications.services import notifications_service
 from sentry.notifications.types import NotificationSettingEnum
 from sentry.services.hybrid_cloud.identity import identity_service
 from sentry.services.hybrid_cloud.integration import RpcIntegration, integration_service
-from sentry.services.hybrid_cloud.notifications import notifications_service
 from sentry.utils import metrics
 from sentry.utils.signing import unsign
 from sentry.web.frontend.base import BaseView, region_silo_view
--- a/src/sentry/integrations/slack/views/link_team.py
+++ b/src/sentry/integrations/slack/views/link_team.py
@@ -1,17 +1,22 @@
 from __future__ import annotations
 
+import logging
 from collections.abc import Sequence
+from dataclasses import asdict
 from typing import Any
 
 from django import forms
 from django.core.signing import BadSignature, SignatureExpired
-from django.http import Http404
 from django.http.response import HttpResponseBase
 from django.utils.decorators import method_decorator
 from rest_framework.request import Request
 
 from sentry import analytics, features
-from sentry.integrations.services.integration import RpcIntegration, integration_service
+from sentry.integrations.slack.metrics import (
+    SLACK_BOT_COMMAND_LINK_TEAM_FAILURE_DATADOG_METRIC,
+    SLACK_BOT_COMMAND_LINK_TEAM_SUCCESS_DATADOG_METRIC,
+)
+from sentry.integrations.slack.views.types import TeamLinkRequest
 from sentry.integrations.types import ExternalProviderEnum, ExternalProviders
 from sentry.models.integrations.external_actor import ExternalActor
 from sentry.models.integrations.integration import Integration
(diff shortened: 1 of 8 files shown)


src/sentry/integrations/slack/views/link_team.py
```diff
@@ -69,23 +78,36 @@ class SlackLinkTeamView(BaseView):
...
         if request.method not in ALLOWED_METHODS:
             return render_error_page(request, status=405, body_text="HTTP 405: Method not allowed")
 
         try:
-            params = unsign(signed_params)
-        except (SignatureExpired, BadSignature):
+            converted = unsign(signed_params)
+            params = TeamLinkRequest(**converted)
+        except (SignatureExpired, BadSignature) as e:
+            _logger.warning("handle.signature_error", exc_info=e)
```
cathteng, line 93: should we specify that this is slack handling in the logger

src/sentry/integrations/slack/views/link_team.py
```diff
@@ -69,23 +78,36 @@ class SlackLinkTeamView(BaseView):
...
-        integration = integration_service.get_integration(integration_id=params["integration_id"])
+        logger_params = asdict(params)
+        logger_params["user_id"] = request.user.id
+
+        integration = integration_service.get_integration(integration_id=params.integration_id)
         if integration is None:
-            raise Http404
+            _logger.info("integration.not_found", extra=logger_params)
+            metrics.incr(self._METRICS_FAILURE_KEY + ".get_integration", sample_rate=1.0)
+            raise render_error_page(
```
cathteng, line 108: should this be `return`? do we want to know about Http404s?

cathteng (commented):
are all of the metrics we're adding actionable? if it's user error like the team is already linked, do we need to be tracking that or is a logger sufficient?

src/sentry/integrations/slack/views/link_team.py
```diff
@@ -35,6 +42,8 @@
 SUCCESS_LINKED_TITLE = "Team linked"
 SUCCESS_LINKED_MESSAGE = "The {slug} team will now receive issue alert{workflow_addon} notifications in the {channel_name} channel."
 
+_logger = logging.getLogger(__name__)
```
sentaur-athena, line 45: Why _? Other loggers I see in the codebase are just `logger`

src/sentry/integrations/slack/views/link_team.py
```diff
@@ -69,23 +78,36 @@ class SlackLinkTeamView(BaseView):
...
     @method_decorator(never_cache)
     def handle(self, request: Request, signed_params: str) -> HttpResponseBase:
         if request.method not in ALLOWED_METHODS:
             return render_error_page(request, status=405, body_text="HTTP 405: Method not allowed")
 
         try:
-            params = unsign(signed_params)
-        except (SignatureExpired, BadSignature):
+            converted = unsign(signed_params)
+            params = TeamLinkRequest(**converted)
```
sentaur-athena, line 91: nit: should we rename this to team_link_request now that it's not params

src/sentry/integrations/slack/views/link_team.py
```diff
@@ -173,6 +209,8 @@ def handle(self, request: Request, signed_params: str) -> HttpResponseBase:
 
         if not created:
             message = ALREADY_LINKED_MESSAGE.format(slug=team.slug)
+            _logger.info("team.already_linked", extra=logger_params)
```
sentaur-athena, line 212: Is this a useful log?

src/sentry/integrations/slack/views/link_team.py
```diff
@@ -202,17 +240,23 @@ def handle(self, request: Request, signed_params: str) -> HttpResponseBase:
...
         )
+
         integration_service.send_message(
             integration_id=integration.id,
             organization_id=team.organization_id,
             channel=channel_id,
             message=message,
         )
+
+        _logger.info("team.linked_success", extra=logger_params)
```
sentaur-athena, line 257: Do we need both log as well as metric?

src/sentry/integrations/slack/views/link_team.py
iamrajjoshi: its a standard that Yash is trying to follow the convention of putting the `_` to emphasize that the logger is used within the file and shouldn't be exported. as part of the refactory, we have been removing references to common loggers in favor of this.

src/sentry/integrations/slack/views/link_team.py
iamrajjoshi: we should be able to get based on how we initialized the logger right `__name__`?

cathteng: that's true

src/sentry/integrations/slack/views/link_team.py
iamrajjoshi: 1. yep should be return. 
2. we will still return a 404, but instead erroring and not returning anything to the user and leaving them with no context on what went wrong, we instead show them an error page. the logging and metric should give us info about the 404s anyway.

src/sentry/integrations/slack/views/link_team.py
iamrajjoshi: your right, ill remove it from here and a few other places i added this log.

src/sentry/integrations/slack/views/link_team.py
iamrajjoshi: good call, its unactionable, Ill remove it

RyanSkonnord (approved):

PR #73235, opened 2024-06-24
fix(processing): Correct timestamp in test

schew2381:
The `comparison_interval` time is inclusive so the test needs to btwn 5 and 10 min.

I checked and this test still fails without https://github.com/getsentry/sentry/pull/73062 so all good


saponifi3d (approved):
:shipit:

PR #73236, opened 2024-06-24
feat(issues): Reduce flicker when switching replays

scttcper:
- Replaces a state that was used to reset replay context + player when switching between replays
- Replay table is now used in one place and rerenders avoided by not being called in an area that constantly rerenders


static/app/views/replays/replayTable/index.tsx
```diff
@@ -43,189 +42,186 @@ type Props = {
   showDropdownFilters?: boolean;
 };
 
-// Memoizing this component to avoid unnecessary re-renders
-const ReplayTable = memo(
-  ({
-    fetchError,
-    isFetching,
```
scttcper, line 50: this should just be whitespace changes after taking off memo

static/app/views/issueDetails/groupReplays/replayClipPreviewWrapper.tsx
```diff
@@ -0,0 +1,53 @@
...
+  pageLinks: string | null;
+  replaySlug: string;
+  replays: ReplayListRecord[] | undefined;
+  selectedReplayIndex: number;
+  setSelectedReplayIndex: (index: number) => void;
+  visibleColumns: ReplayColumn[];
+  overlayContent?: React.ReactNode;
+};
+
+export function ReplayClipPreviewWrapper(props: Props) {
```
scttcper, line 20: this is ReplayTableWrapper without the table anymore

static/app/views/issueDetails/groupReplays/groupReplays.tsx
```diff
@@ -221,31 +208,36 @@ function GroupReplaysTable({
...
+      selectedReplayIndex={selectedReplayIndex}
+      overlayContent={overlayContent}
+      organization={organization}
+      group={group}
+      replaySlug={selectedReplay.id}
+      pageLinks={replayListData.pageLinks}
+      replays={replays}
+    >
+      {replayTable}
+    </GroupReplaysTableInner>
```
ryan953, line 237: This is much better already. There were too many layers in here before!

ryan953 (approved):

PR #73239, opened 2024-06-24
feat(<DraggableTabs>): Add new draggable tabs component to storybook

### LINKED ISSUE #73221: Custom Views: MVP for dnd tabs component
Drag and drop tabs must be implemented to support custom views. This ticket tracks the progress of creating the basic drag and drop tab component that will eventually be used to for custom views. 

MVP features include:
- [ ] Unstyled tabs that can be dragged with the cursor 
- [ ] Unstyled tabs that can be dropped between any other tabs in the tab bar 
- [ ] A basic drop indicator indicating where a dragged tab will drop 
- [ ] A storybook entry displaying this component

MichaelSun48:
**Note:** This PR merges into the `msun/draggableTabsComponent`, not `master`. The `msun/draggableTabsComponent` branch contains the `components/draggableTabs` directory with with the files from `components/tabs` (except for the test file). This is so that the diff for this PR is easier to read and more intuitive – it effectively shows all upgrades to the existing tabs components to make the tabs draggable. 

This PR implements an MVP for drag and drop tabs. Visually, these tabs are identical to the current tabs component, but you can drag them and drop them within the TabList, and a vertical line appears to indicate the drop position for a tab that is currently being dragged. You can play around with this component in the vercel deployment's storybook. 

Quick Demo:

https://github.com/getsentry/sentry/assets/55160142/d36b8fd6-4d99-4a1a-9aa9-04e293cfef21

Known issues:
- Tabs seem to not drag unless you focus on them first 
- The drop indicator seems to stutter between the first and second positions sometimes (visible in the screen recording above)


static/app/components/draggableTabs/draggableTab.tsx
```diff
@@ -73,28 +112,80 @@ function BaseTab(
...
+    {
+      target: {type: 'item', key: item.key, dropPosition: 'on'},
+    },
+    dropState,
+    ref
+  );
+
+  function Draggable({children}) {
+    const {dragProps, dragButtonProps, isDragging} = useDrag({
+      getAllowedDropOperations: () => ['move'],
```
scttcper, line 125: you'll need to pull Draggable out of this component since things like calling hooks inside here will violate rules of hooks

static/app/components/draggableTabs/draggableTab.tsx
```diff
@@ -73,28 +112,80 @@ function BaseTab(
...
 
+  const {dropProps, isDropTarget} = useDroppableItem(
+    {
+      target: {type: 'item', key: item.key, dropPosition: 'on'},
+    },
+    dropState,
+    ref
+  );
+
+  function Draggable({children}) {
```
scttcper, line 123: define prop types here to avoid implicit any

static/app/components/draggableTabs/draggableTab.tsx
```diff
@@ -73,28 +112,80 @@ function BaseTab(
...
+      />
+      <TabWrap
+        {...mergeProps(tabProps, dropProps)}
+        hidden={hidden}
+        selected={isSelected}
+        overflowing={overflowing}
+        ref={ref}
+        className={`option ${isFocusVisible ? 'focus-visible' : ''} ${
+          isDropTarget ? 'drop-target' : ''
+        }`}
```
scttcper, line 166: is there a plan for these classnames?

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -223,14 +249,47 @@ function BaseTabList({
...
-   */
+export function DraggableTabList({
+  items,
+  tabs,
+  setTabs,
+  ...props
+}: DraggableTabListProps) {
+  const onInsert = async (e: DroppableCollectionInsertDropEvent) => {
+    const dropItem = e.items[0] as TextDropItem;
+    const eventTab = JSON.parse(await dropItem.getText('tab'));
```
scttcper, line 275: lets add some types here since we know what object is expected

static/app/components/draggableTabs/index.tsx
```diff
@@ -75,6 +78,35 @@ export function Tabs<T extends string | number>({
...
+  key: Key;
+  label: string;
+}
+
+export interface DragAndDropTabBarProps {
+  tabs: Tab[];
+}
+
+export function DraggableTabBar(props: DragAndDropTabBarProps) {
+  const [tabs, setTabs] = useState<Tab[]>(props.tabs);
```
scttcper, line 92: might need a useEffect to update this state if the props.tabs change otherwise it will only use the props.tabs from the first render

scttcper:
dragging an item onto itself seems to move it to the next position

https://github.com/getsentry/sentry/assets/1400464/14998161-d9e2-4757-8337-16df834bb9b5

scttcper:
feels like the draggable element is hijacking the click to switch tabs

https://github.com/getsentry/sentry/assets/1400464/62eb43cc-56ed-4b61-b01a-798badf39881

MichaelSun48:
> feels like the draggable element is hijacking the click to switch tabs

Yup you're totally right about this – moving Draggable outside of the component fixed the issue with clicks being hijacked. Just curious - what was your intuition there? How did you know that it was the fact that Draggable was violating the rule of hooks that was causing the issue with clicks being weird?

static/app/components/draggableTabs/draggableTab.tsx
MichaelSun48: Nope, this was copy pasted over from the react-aria guides. I'll delete them if/until they become useful

MichaelSun48:
> dragging an item onto itself seems to move it to the next position 
Fixed in the latest commits, thanks for catching this!

MichaelSun48:
> feels like the draggable element is hijacking the click to switch tabs

This is still broken - working on a fix

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -223,14 +249,51 @@ function BaseTabList({
...
+        }
+        const targetIdx = updatedTabs.indexOf(targetTab);
+        if (e.target.dropPosition === 'before') {
+          updatedTabs.splice(targetIdx, 0, draggedTab);
+        } else if (e.target.dropPosition === 'after') {
+          updatedTabs.splice(targetIdx + 1, 0, draggedTab);
+        }
+        setTabs(updatedTabs);
+      }
+    }
```
scttcper, line 294: i think we can simplify stuff here by returning early and combining the find and indexOf
```suggestion
    const draggedTab = tabs.find(tab => tab.key === eventTab.key);
    if (!draggedTab || e.target.key === draggedTab.key) {
      return; // Do nothing if the dragged tab is dropped on itself
    }

    const updatedTabs = tabs.filter(tab => tab.key !== draggedTab.key);
    const targetIdx = tabs.findIndex(tab => tab.key === e.target.key);
    if (targetIdx > -1) {
      if (e.target.dropPosition === 'before') {
        updatedTabs.splice(targetIdx, 0, draggedTab);
      } else if (e.target.dropPosition === 'after') {
        updatedTabs.splice(targetIdx + 1, 0, draggedTab);
      }
      setTabs(updatedTabs);
    }
```

scttcper (approved):

malwilley (commented):
Looks like a good start! We can improve on the drag interactions in subsequent PRs, but before merging this one I'd like to try a couple things:

- Add a test file with some basic tests (similar to the original tabs component). Ideally we would want a test that validates the drag and drop is working correctly, but that may be difficult with the jsdom test env. You can try something like this https://testing-library.com/docs/user-event/pointer and see if that works though.
- Ideally we would want to use some of the Tab components directly instead of copy/pasting them. I'd give that a shot and at least uncover what kind of issues you run in to and note them down so that we can tackle the code deduplication later on.

malwilley (approved):

PR #73242, opened 2024-06-24
fix: attach activations to incidents on trigger

nhsiehgit:
nit - check the `monitor_type` to the enum `value` rather than the enum itself

Update tests

### DIFF
--- a/src/sentry/incidents/subscription_processor.py
+++ b/src/sentry/incidents/subscription_processor.py
@@ -630,7 +630,7 @@ class SubscriptionProcessor:
             if not self.active_incident:
                 detected_at = self.calculate_event_date_from_update_date(self.last_update)
                 activation: AlertRuleActivations | None = None
-                if self.alert_rule.monitor_type == AlertRuleMonitorType.ACTIVATED:
+                if self.alert_rule.monitor_type == AlertRuleMonitorType.ACTIVATED.value:
                     activations = list(self.subscription.alertruleactivations_set)
                     if len(activations) != 1:
                         logger.error(
--- a/src/sentry/incidents/subscription_processor.py
+++ b/src/sentry/incidents/subscription_processor.py
@@ -631,7 +631,7 @@ class SubscriptionProcessor:
                 detected_at = self.calculate_event_date_from_update_date(self.last_update)
                 activation: AlertRuleActivations | None = None
                 if self.alert_rule.monitor_type == AlertRuleMonitorType.ACTIVATED.value:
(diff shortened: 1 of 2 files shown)


src/sentry/incidents/subscription_processor.py
```diff
@@ -630,8 +630,8 @@ def trigger_alert_threshold(
             if not self.active_incident:
                 detected_at = self.calculate_event_date_from_update_date(self.last_update)
                 activation: AlertRuleActivations | None = None
-                if self.alert_rule.monitor_type == AlertRuleMonitorType.ACTIVATED:
```
nhsiehgit, line 633: was previously comparing to the enum itself rather than the `.value`

src/sentry/incidents/subscription_processor.py
```diff
@@ -630,8 +630,8 @@ def trigger_alert_threshold(
...
-                    activations = list(self.subscription.alertruleactivations_set)
+                if self.alert_rule.monitor_type == AlertRuleMonitorType.ACTIVATED.value:
+                    activations = list(self.subscription.alertruleactivations_set.all())
```
schew2381, line 634: What is `alertruleactivations_set`?

nhsiehgit, line 634: this is another functional change that was missed

cannot `list`ify the `_set` - but still need to grab `.all()` in order to `list`ify

schew2381: jk saw the other comment

src/sentry/incidents/subscription_processor.py
schew2381: Instead of using value, could we switch `AlertRuleMonitorType` to an [IntEnum](https://docs.python.org/3/library/enum.html#enum.IntEnum)?

nhsiehgit: The Enum isn't quite an `IntEnum` though.
`... but its members are also integers `

the members of the enum are not ints but more descriptors of what the enum is for 🤷

schew2381: by members there I think it means the values that represent those members?

schew2381: I think the example is maybe a bit extra in that the members are ints and can be manipulated freely, but we really just care about the equality part.

nhsiehgit: for now i'll leave it as is.

out of scope for this PR to change the enum type / value type

schew2381 (approved):
lgtm

PR #73244, opened 2024-06-24
fix(similarity): return empty lists from snuba if no results

JoshFerge:
if snuba returns no results, we should turn empty lists, not nothing. 

Fixes SENTRY-3B1X


JoshFerge:
missed this edge case during writing tests. can follow up with one if this case gives us any more problems.

jangjodi (approved):

mrduncan (approved):
Would a return type annotation here help mypy catch this?

JoshFerge:
> Would a return type annotation here help mypy catch this?

indeed. another +1 for having stricter mypy on by default

PR #73250, opened 2024-06-24
chore(<DraggableTabs>): Make new draggableTabs directory, copied from tabs directory

MichaelSun48:
This PR duplicates all files from the `sentry/app/components/tabs` directory to a new directory, `sentry/app/components/draggableTabs`. There are **no public changes** to this PR, and the sole purpose of this PR is to make reviewing the PR that actually implements the drag-and-drop tabs components easier. That PR (currently drafted [here](https://github.com/getsentry/sentry/pull/73239)) is at nearly +1000 lines of code, and the vast majority of the changes simply copied over from `.../tabs`. This PR would make the diff in that PR significantly more readable and intuitive. 

The contents of each file in `.../draggableTabs` have **not** been modified, but the names of any files starting with `tab` has been changes to `draggableTab` to reflect the future behavior of these components. I have also removed the `.spec.tsx` file and the `stories` file since these will be significantly different for drag and drop tabs compared to regular tabs.

### DIFF
--- a/static/app/components/draggableTabs/draggableTab.tsx
+++ b/static/app/components/draggableTabs/draggableTab.tsx
@@ -1,0 +1,236 @@
+import {forwardRef, useCallback} from 'react';
+import type {Theme} from '@emotion/react';
+import styled from '@emotion/styled';
+import type {AriaTabProps} from '@react-aria/tabs';
+import {useTab} from '@react-aria/tabs';
+import {useObjectRef} from '@react-aria/utils';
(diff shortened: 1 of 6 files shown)


malwilley:
@MichaelSun48 another option here would be to create a feature branch and open PRs that merge into that instead of master. You can do that until the feature branch has a basic working version of the component, then merge that branch into master. That way we won't have a bunch of duplicated files hanging around for too long.

PR #73251, opened 2024-06-24
chore(empty-states): Use existing docs for empty state

roggenkemper:
this pr updates the updated issue stream empty state to use existing docs, instead of hardcoded ones, for the existing 4 projects. this change, while not having any impact on what users actually see, will allow for other projects to easily be added while also keeping the instructions as up to date as possible


static/app/components/updatedEmptyState.tsx
```diff
@@ -143,32 +41,83 @@ export default function UpdatedEmptyState({project}: {project?: Project}) {
...
+
+  const currentPlatformKey = project?.platform ?? 'other';
+  const currentPlatform = platforms.find(
+    p => p.id === currentPlatformKey
+  ) as PlatformIntegration;
+
+  const platformPath = getPlatformPath(currentPlatform);
+
+  useEffect(() => {
+    async function getGettingStartedDoc() {
```
malwilley, line 55: I'm seeing this save useEffect in a few places, would it be possible to make this a common util?

static/app/components/updatedEmptyState.tsx
```diff
@@ -143,32 +41,83 @@ export default function UpdatedEmptyState({project}: {project?: Project}) {
...
   ) {
     return null;
   }
 
   const dsn = projectKeys[0].dsn.public;
-  const language = project?.platform === 'node' ? 'javascript' : 'python';
-
+  const {default: docs} = module;
+
+  const docParams: DocsParams<any> = {
```
malwilley, line 83: Does this have to be `any`?

static/app/components/updatedEmptyState.tsx
roggenkemper: will merge a separate PR to do this first, then update this one !

static/app/components/updatedEmptyState.tsx
roggenkemper: it's not great but that's what we have across a bunch of components right now, i can come back and see if it really is necessary after this pr

malwilley (approved):

PR #73252, opened 2024-06-25
fix(issues): Reduce shift when switching replays

scttcper:
Reduces layout shift when switching between replays on the issue replays tab.

Adds a placeholder for some text that appears after load
image

### DIFF
--- a/static/app/views/issueDetails/groupReplays/groupReplays.tsx
+++ b/static/app/views/issueDetails/groupReplays/groupReplays.tsx
@@ -64,6 +64,7 @@ function GroupReplays({group}: Props) {
   }, []);
 
   if (!eventView) {
+    // Shown on load and no replay data available
     return (
       <StyledLayoutPage withPadding>
         <ReplayCountHeader>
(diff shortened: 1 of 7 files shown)


static/app/views/issueDetails/groupReplays/groupReplays.tsx
```diff
@@ -242,11 +252,11 @@ function GroupReplaysTable({
...
       <ReplayCountHeader>
-        <StyledIconUser size="sm" />
+        <IconUser size="sm" />
         {t(
-          'Replay captured %s users experiencing this issue across %s events.',
-          replayCount,
-          group.count
+          'Replay captured %s experiencing this issue across %s.',
+          tn('%s user', '%s users', replayCount ?? 0),
+          tn('%s event', '%s events', group.count)
```
ryan953, line 259: Could we put the whole sentence in to the ternary and tn() calls. drop the t() wrapper part. Should be easier for  translators to have the full sentence there.

scttcper: i was thinking you could have 1 user but 2 events or 2 events and 1 user

ryan953 (approved):
Rest looks good. Data-test-id all the things!

static/app/views/issueDetails/groupReplays/groupReplays.tsx
ryan953: Fair enough. 

Tough to use t18n lib :(

PR #73253, opened 2024-06-25
fix(feedback): enforce case insensitivity for spam detection

### LINKED ISSUE #73066: [User Feedback] spam detection should filter messages uniformly without taking into account capitalization

aliu39:
Fixes https://github.com/getsentry/sentry/issues/73066
Lower-cases all input messages and examples in the prompt. See experiment results below.
Includes some minor refactors


JoshFerge:
how did this look in the analysis notebook? change results at all?

aliu39:
**Before**
Correct Predictions: 92%
Precision: 94.44%
Recall: 73.91%
False Positives: 2
False Negatives: 12
True Positives: 34
True Negatives: 133

**After**
Correct Predictions: 93%
Precision: 92.5%
Recall: 80.43%
False Positives: 3
False Negatives: 9
True Positives: 37
True Negatives: 132

**Before, each character of inference data has a 50% chance of being upper-case**
Correct Predictions: 80%
Precision: 56.58%
Recall: 93.48%
False Positives: 33
False Negatives: 3
True Positives: 43
True Negatives: 102

aliu39:
Did some more experiments, with original dataset but chance of capital = 0.2. What I found:
- lowering the input improves results across the board, no matter what other strategies are tried
- best performing was lower case input, lower case the current examples.
Correct Predictions: 94%
Precision: 92.68%
Recall: 82.61%
False Positives: 3
False Negatives: 8
True Positives: 38
True Negatives: 132

The other strategies I tried were 
a) adding an explicit instruction and/or 
b) duplicating 2 examples and using mixed capitalization (spam + notspam)
Both these resulted in slightly lower accuracy, 91%. However, precision > recall by almost 20%, meaning filters are a little weaker (don't filter spam as much).

aliu39:
Results on original dataset (no random capitalizations):
Correct Predictions: 93%
Precision: 92.31%
Recall: 78.26%
False Positives: 3
False Negatives: 10
True Positives: 36
True Negatives: 132

Fails on similar examples too

JoshFerge (approved):

PR #73255, opened 2024-06-25
Update AlertRuleMonitorType to be an IntEnum

saponifi3d:
# Description
Migrate the `AlertRuleMonitorType` to use an `IntEnum` as the base. This is possible because the `monitor_type` field is an [`IntegerField`](https://github.com/getsentry/sentry/blob/2256c3c7665ffccf02d58af987c0fc117bf6a2d5/src/sentry/incidents/models/alert_rule.py#L270), by doing this we can simplify how we use the enum in code, and improve typing.

### Review Callout
I wasn't sure if i should update the migration or not though (by updating it, we can support potential rollbacks. the risk i'm not sure of is if django or our ci will pick it up as a change and trigger any new migrations).

### DIFF
--- a/src/sentry/incidents/models/alert_rule.py
+++ b/src/sentry/incidents/models/alert_rule.py
@@ -324,14 +324,14 @@ class AlertRule(Model):
             "Subscribing projects to alert rule",
             extra={
                 "alert_rule.monitor_type": self.monitor_type,
-                "conditional_monitor_type": monitor_type.value,
+                "conditional_monitor_type": monitor_type,
                 "query_extra": query_extra,
             },
         )
         # NOTE: AlertRuleMonitorType.ACTIVATED will be conditionally subscribed given activation triggers
         # On activated subscription, additional query parameters will be added to the constructed query in Snuba
         created_subscriptions = []
-        if self.monitor_type == monitor_type.value:
+        if self.monitor_type == monitor_type:
             # NOTE: QuerySubscriptions hold reference to Projects which should match the AlertRule's project reference
             created_subscriptions = bulk_create_snuba_subscriptions(
                 projects,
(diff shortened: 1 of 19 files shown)


src/sentry/incidents/models/alert_rule.py
```diff
@@ -225,7 +225,7 @@ class Meta:
         unique_together = (("alert_rule", "project"),)
 
 
-class AlertRuleMonitorType(Enum):
+class AlertRuleMonitorType(IntEnum):
```
nhsiehgit, line 228: 👍 nice

Can we actually append `Int` to the enum name?

As we transition over to `IntEnum`'s, i think it would make it more clear why some of our enums are being `.value`'d whereas others are being compared directly 😅

nhsiehgit (commented):
awesome quick comment

Thanks for doing this!

src/sentry/incidents/models/alert_rule.py
```diff
@@ -47,11 +47,11 @@ def __call__(self, subscription: QuerySubscription, *args: Any, **kwargs: Any) -
         ...
 
 
-alert_subscription_callback_registry: dict[AlertRuleMonitorType, SubscriptionCallback] = {}
+alert_subscription_callback_registry: dict[AlertRuleMonitorTypeInt, SubscriptionCallback] = {}
```
nhsiehgit, line 50: Oh hmm - the key will be different so how might this registry be affected?

don't think it should matter, but just want to double check

nhsiehgit (approved):

azaslavsky (approved):

src/sentry/incidents/models/alert_rule.py
saponifi3d: 👍 yeah - confirm this is just a name change, the type is still the same (it'll actually be a slightly better type now that it's an IntEnum 😄)

PR #73258, opened 2024-06-25
feat(uptime): Tasks for automatic hostname detection and monitoring

wedamija:
This implements most of the second half of automatic hostname detection and monitoring. It has a few components:
 - `schedule_detections` runs every minute. It looks at the last time it was run, and fires off tasks for each minute bucket.
 - `process_detection_bucket` runs for a specific datetime bucket. It just fetches all projects from this bucket and fires `process_project_url_ranking` for each of them. Note that each project will only belong to a single bucket.
 - `process_project_url_ranking` runs for a specific project. This is where the meat of the logic exists. We get the ranked list of urls collected from this project's event data and run it through various criteria to see whether we should attempt to monitor the url

There are a fair few todos around that I need to finish up, as well as building the logic to send subscriptions to the topic for our rust consumer.

### DIFF
--- a/src/sentry/uptime/detectors/tasks.py
+++ b/src/sentry/uptime/detectors/tasks.py
@@ -15,8 +15,8 @@ from sentry.uptime.detectors.ranking import (
 )
 from sentry.uptime.models import ProjectUptimeSubscription, UptimeSubscription
 from sentry.utils import metrics
+from sentry.utils.hashlib import md5_text
 from sentry.utils.locking import UnableToAcquireLock
-from sentry_plugins.jira.client import md5
 
 LAST_PROCESSED_KEY = "uptime_detector_last_processed"
 SCHEDULER_LOCK_KEY = "uptime_detector_scheduler_lock"
@@ -92,7 +92,7 @@ def process_project_url_ranking(project_id: int, project_url_count: int):
         return
 
     for url, url_count in get_candidate_urls_for_project(project)[:5]:
-        if check_url(project, project_url_count, url, url_count):
+        if process_candidate_url(project, project_url_count, url, url_count):
             # TODO: On success, we want to mark this project as not needing to be checked for a while
             break
     else:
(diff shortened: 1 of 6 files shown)


src/sentry/uptime/detectors/tasks.py
```diff
@@ -0,0 +1,164 @@
...
+            break
+    else:
+        # TODO: If we don't find any urls to monitor, we want to increment a counter in redis and check the value.
+        # After a number of failures, we want to stop checking for this project for a while.
+        pass
+
+    delete_candidate_urls_for_project(project)
+
+
+def check_url(project: Project, project_url_count: int, url: str, url_count: int) -> bool:
```
wedamija, line 101: Tbh I'd prefer to break most of the logic out of `tasks.py` but just haven't gotten around to it yet.

src/sentry/uptime/detectors/tasks.py
```diff
@@ -0,0 +1,164 @@
...
+        SCHEDULER_LOCK_KEY,
+        duration=60,
+        name="uptime.detection.schedule_detections",
+    )
+    try:
+        with lock.acquire():
+            cluster = _get_cluster()
+            last_processed = cluster.get(LAST_PROCESSED_KEY)
+            if last_processed is None:
+                last_processed = datetime.now().replace(second=0, microsecond=0)
```
evanpurkhiser, line 46: Can we use a timezone aware datetime here in UTC?

src/sentry/uptime/detectors/tasks.py
```diff
@@ -0,0 +1,169 @@
...
+            if last_processed is None:
+                last_processed = timezone.now().replace(second=0, microsecond=0)
+            else:
+                last_processed = datetime.datetime.fromtimestamp(
+                    int(last_processed), tz=datetime.UTC
+                )
+
+            minutes_since_last_processed = int(
+                (timezone.now() - last_processed) / timedelta(minutes=1)
+            )
```
evanpurkhiser, line 57: This is probably more clear, but you can use the `//` integer division operator too

```suggestion
            minutes_since_last_processed = (timezone.now() - last_processed) // timedelta(minutes=1)
```

yeah way less clear lol

src/sentry/uptime/detectors/tasks.py
```diff
@@ -0,0 +1,169 @@
...
+            break
+    else:
+        # TODO: If we don't find any urls to monitor, we want to increment a counter in redis and check the value.
+        # After a number of failures, we want to stop checking for this project for a while.
+        pass
+
+    delete_candidate_urls_for_project(project)
+
+
+def check_url(project: Project, project_url_count: int, url: str, url_count: int) -> bool:
```
evanpurkhiser, line 106: `check_url` kind of makes it sound like it's going to make HTTP requests. Maybe a better name could be `url_meets_criteria` or something?

evanpurkhiser: maybe it just needs a doc string

evanpurkhiser (approved):

PR #73264, opened 2024-06-25
feat(performance): UI updates to regression emails

sentaur-athena:
Updating UI for regression emails so the more important information gets more visibility. Also regression emails have no span so removing span section from the email which currently always empty.

Some notes: 
* styles between the two image are different because before is from a real email. I didn't change any styling. 
* Ignore the data mismatch in the image, I just rendered fake data

Before:
Screenshot 2024-06-25 at 11 50 59 AM

After:
Screenshot 2024-06-25 at 11 50 27 AM

### DIFF
--- a/src/sentry/notifications/notifications/rules.py
+++ b/src/sentry/notifications/notifications/rules.py
@@ -12,7 +12,12 @@ from sentry.db.models import Model
 from sentry.eventstore.models import GroupEvent
 from sentry.integrations.issue_alert_image_builder import IssueAlertImageBuilder
 from sentry.integrations.types import ExternalProviderEnum, ExternalProviders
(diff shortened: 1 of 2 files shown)


src/sentry/notifications/notifications/rules.py
```diff
@@ -146,6 +151,14 @@ def get_image_url(self) -> str | None:
             return image_builder.get_image_url()
         return None
 
+    def is_new_design(self) -> bool:
```
sentaur-athena, line 154: I will clean this up after removing feature flags. Right now 3 feature flags are involved which makes the logic complicated.

cathteng (approved):

PR #73269, opened 2024-06-25
ref(settings-debug-files): Convert to functional component

priscilawebdev:
### Goal

**Preview:**

https://github.com/getsentry/sentry/assets/29228205/19fb30c4-5dde-42ac-8cf4-31c42d0498ff

https://github.com/getsentry/sentry/assets/29228205/52f4488f-6ab3-47af-b131-2f88759c6815

### Note

- Changes were manually tested locally


static/app/views/settings/projectDebugFiles/index.tsx
```diff
@@ -27,211 +34,168 @@ type Props = RouteComponentProps<{projectId: string}, {}> & {
...
+        ...location,
+        query: {...location.query, cursor: undefined, query},
+      });
+    },
+    [navigate, location]
+  );
+
+  const query = useMemo(() => {
+    return typeof location.query === 'string' ? location.query : undefined;
+  }, [location.query]);
```
ArthurKnaus, line 84: As the return value of this `useMemo` is atomic (`string | undefined`) it will actually not prevent any re-renders, right?
Also the original code was accessing `location.query.query`
So we could just do:
```ts
const queryParam = location.query.query
const query = queryParam === 'string' ? queryParam : undefined
```

static/app/views/settings/projectDebugFiles/index.tsx
```diff
@@ -27,211 +34,168 @@ type Props = RouteComponentProps<{projectId: string}, {}> & {
...
-              builtinSymbolSourceOptions={builtinSymbolSources ?? []}
-              isLoading={loading}
+    },
+    onMutate: () => {
+      addLoadingMessage('Deleting debug file');
+    },
+    onSuccess: () => {
+      addSuccessMessage('Successfully deleted debug file');
+      refetchDebugFiles();
+      refetchBuiltinSymbolSources();
```
ArthurKnaus, line 101: I am not sure about it, but I think refetch will also fetch a query where `enabled: false` 🤔 (could you check that?)
Normally, I would just invalidate them.

static/app/views/settings/projectDebugFiles/index.tsx
```diff
@@ -27,211 +34,168 @@ type Props = RouteComponentProps<{projectId: string}, {}> & {
...
-    });
-
-    this.api.request(
-      `/projects/${organization.slug}/${params.projectId}/files/dsyms/?id=${id}`,
+  const {
+    data: debugFiles,
+    getResponseHeader: getDebugFilesResponseHeader,
+    isLoading: isLoadingDebugFiles,
+    refetch: refetchDebugFiles,
+  } = useApiQuery<DebugFile[] | null>(
```
ArthurKnaus, line 48: The old implementation would have surfaced loading errors, right?
This is missing in the new one.

static/app/views/settings/projectDebugFiles/index.tsx
priscilawebdev: good catch!!

static/app/views/settings/projectDebugFiles/index.tsx
priscilawebdev: yes you are right in the DeprecatedAsyncView - I have handled it now .. thanks for the good catch!

ArthurKnaus (approved):
🙌

PR #73271, opened 2024-06-25
feat(metrics): Limit query operators

### LINKED ISSUE #73150: Only allow equal and IN filters

ArthurKnaus:
Make sure only `equal` and `in` can be used as operators in the query of a metric extraction rule.

Remove not equal from the list of suggested operators.
Screenshot 2024-06-25 at 12 54 39

Mark filter as invalid when a negation is used.
Screenshot 2024-06-25 at 12 54 46

Closes https://github.com/getsentry/sentry/issues/73150

### DIFF
--- a/static/app/components/searchSyntax/parser.spec.tsx
+++ b/static/app/components/searchSyntax/parser.spec.tsx
@@ -208,6 +208,29 @@ describe('searchSyntax/parser', function () {
     });
   });
 
+  it('applies dissalowNegation', () => {
+    const result = parseSearch('!foo:bar', {
+      disallowNegation: true,
+      invalidMessages: {
+        [InvalidReason.NEGATION_NOT_ALLOWED]: 'Custom message',
+      },
+    });
+
+    // check with error to satisfy type checker
+    if (result === null) {
+      throw new Error('Parsed result as null');
+    }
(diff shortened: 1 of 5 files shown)


static/app/components/searchSyntax/parser.spec.tsx
```diff
@@ -208,6 +208,29 @@ describe('searchSyntax/parser', function () {
     });
   });
 
+  it('applies dissalowNegation', () => {
```
shellmayr, line 211: nit: typo (dissalow -> disallow)

static/app/components/searchSyntax/parser.spec.tsx
```diff
@@ -208,6 +208,29 @@ describe('searchSyntax/parser', function () {
     });
   });
 
+  it('applies dissallowNegation', () => {
```
vgrozdanic, line 211: Nit: typo in `dissallowNegation` , it should be disallow, only one 's'

ArthurKnaus, line 211:
```suggestion
  it('applies disallowNegation', () => {
```

vgrozdanic (approved):

PR #73272, opened 2024-06-25
fix(metrics): Duplicate name warning spanning use cases

### LINKED ISSUE #73149: Duplicate metrics with different units check needs to take namespace into consideration

ArthurKnaus:
Avoid displaying duplication warnings if the affected metrics are from different use cases. 

Closes https://github.com/getsentry/sentry/issues/73149

### DIFF
--- a/static/app/components/metrics/mriSelect/index.tsx
+++ b/static/app/components/metrics/mriSelect/index.tsx
@@ -66,7 +66,7 @@ export function getMetricsWithDuplicateNames(metrics: MetricMeta[]): Set<MRI> {
 
   for (const metric of metrics) {
     const parsedMri = parseMRI(metric.mri);
-    // Include the use case to avaoid warning of conflicts between different use cases
+    // Include the use case to avoid warning of conflicts between different use cases
     const metricName = `${parsedMri?.useCase}_${parsedMri?.name}`;
     if (!metricName) {
       continue;
--- a/static/app/components/metrics/mriSelect/index.tsx
+++ b/static/app/components/metrics/mriSelect/index.tsx
@@ -65,7 +65,9 @@ export function getMetricsWithDuplicateNames(metrics: MetricMeta[]): Set<MRI> {
   const duplicateNames: string[] = [];
 
   for (const metric of metrics) {
-    const metricName = parseMRI(metric.mri)?.name;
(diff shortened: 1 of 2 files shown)


static/app/components/metrics/mriSelect/index.tsx
```diff
@@ -65,7 +65,9 @@ export function getMetricsWithDuplicateNames(metrics: MetricMeta[]): Set<MRI> {
   const duplicateNames: string[] = [];
 
   for (const metric of metrics) {
-    const metricName = parseMRI(metric.mri)?.name;
+    const parsedMri = parseMRI(metric.mri);
+    // Include the use case to avaoid warning of conflicts between different use cases
```
priscilawebdev, line 69:
```suggestion
    // Include the use case to avoid warning of conflicts between different use cases
```

static/app/components/metrics/mriSelect/index.tsx
```diff
@@ -65,7 +65,9 @@ export function getMetricsWithDuplicateNames(metrics: MetricMeta[]): Set<MRI> {
...
+    const metricName = `${parsedMri?.useCase}_${parsedMri?.name}`;
```
priscilawebdev, line 70: `metricName` will be always defined because of `_`

ArthurKnaus: that is actually fine as the we should always valid MRI's from the BE 👍 
The types of `parseMRI` are a bit lacking...

ArthurKnaus: @priscilawebdev  Follow up: https://github.com/getsentry/sentry/pull/73277

priscilawebdev: but the code will never go to the `continue` described bellow . shall we remove it?

ArthurKnaus: that is true, good catch!
I would merge the other PR first and then remove the `continue` + change all `parsedMri?.` to `parsedMri.` in this file.

ArthurKnaus: done!

priscilawebdev (approved):
lgtm

Issue #65869, opened 2024-02-27
Change Hub clones to isolation scope forks


szokeasaurusrex:
Once Sentry Python SDK v2.0 is released, and once we are using 2.0 in Sentry, we should update manual Hub clones to isolation scope forks. The Hub API is deprecated from v2.0; forking an isolation scope is analogous to cloning the Hub in the new API.
```[tasklist]
### Implementing PRs
- [ ] https://github.com/getsentry/sentry/pull/73266
- [ ] https://github.com/getsentry/sentry/pull/73275
- [ ] #73279
- [ ] #73280
```

PR #73275, opened 2024-06-25
ref(utils): Replace Hub with isolation scope in concurrent utils

szokeasaurusrex:
Depends on #73276

ref #65869


antonpirker (approved):
Looks good

PR #73276, opened 2024-06-25
ref(utils): Improve `PriorityTask` typing

szokeasaurusrex:
Type the items in the `PriorityTask` named tuple. This change will allow the change to isolation scope from Hub, which we introduce in #73275, to be more explicitly documented.

### DIFF
--- a/src/sentry/utils/concurrent.py
+++ b/src/sentry/utils/concurrent.py
@@ -9,7 +9,7 @@ from concurrent.futures._base import FINISHED, RUNNING
 from contextlib import contextmanager
 from queue import Full, PriorityQueue
 from time import time
-from typing import TYPE_CHECKING, NamedTuple, TypeVar
+from typing import TYPE_CHECKING, Generic, NamedTuple, TypeVar
 
 from sentry_sdk import Hub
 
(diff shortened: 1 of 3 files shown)


src/sentry/utils/concurrent.py
```diff
@@ -46,7 +45,10 @@ def run():
 
 
 @functools.total_ordering
-class PriorityTask(collections.namedtuple("PriorityTask", "priority item")):
+class PriorityTask(NamedTuple):
+    priority: int
+    item: tuple[Callable, Hub, FutureBase]
```
asottile-sentry, line 50: `Callable` is an incomplete type -- this should have a signature probably?

mypy should probably be running with `--disallow-any-generics`

szokeasaurusrex: @asottile-sentry This was completely untyped previously – so what we have here is already an improvement – and it is not entirely obvious to me what the type here should be. 

If you prefer, I can explicitly type this callable as `Callable[..., Any]`, but I believe adding more specific typing or changing mypy configuration would be out of scope for this PR.

asottile-sentry: ah but if you add this file to the stronglist it will require that setting -- at the bottom of `pyproject.toml`

asottile-sentry: looking at the usage below this appears to be `Callable[[], R]`

szokeasaurusrex: OK, I was able to make some changes to type the callable as `Callable[[], T]`. This file now passes with `--disallow-any-generics`; however, we cannot add it to the stronglist as the file still fails `--disallow-untyped-defs` due to untyped functions. I believe we should fix the `--disallow-untyped-defs` separately, since it is not directly related to the `PriorityTask` typing we are trying to improve here.

asottile-sentry (approved):

asottile-sentry (approved):

PR #73277, opened 2024-06-25
ref(metrics): Make parseMRI type aware

ArthurKnaus:
Make `parseMRI` type aware.
Its return type should be `ParseResult` if it is supplied with an MRI.

```ts
parseMRI(string) -> ParseResult | null;
parseMRI(MRI) -> ParseResult;
```

It will enable us to get rid of some casts in the code.

### DIFF
--- a/static/app/utils/metrics/mri.tsx
+++ b/static/app/utils/metrics/mri.tsx
@@ -18,8 +18,10 @@ export function isMRI(mri?: unknown): mri is MRI {
   }
 }
 
-type ParseResult<T extends MRI | string> = T extends MRI ? ParsedMRI : ParsedMRI | null;
-export function parseMRI<T extends MRI | string>(mri?: T): ParseResult<T> {
+type ParseResult<T extends MRI | string | null> = T extends MRI
+  ? ParsedMRI
+  : ParsedMRI | null;
+export function parseMRI<T extends MRI | string | null>(mri?: T): ParseResult<T> {
   if (!mri) {
     // TODO: How can this be done without casting?
     return null as ParseResult<T>;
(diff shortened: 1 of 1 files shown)


static/app/utils/metrics/mri.tsx
```diff
@@ -7,21 +7,28 @@ export const DEFAULT_MRI: MRI = 'c:custom/sentry_metric@none';
 export const DEFAULT_METRIC_ALERT_FIELD = `sum(${DEFAULT_MRI})`;
 
 export function isMRI(mri?: unknown): mri is MRI {
-  return !!parseMRI(mri);
+  try {
+    _parseMRI(mri as string);
```
priscilawebdev, line 11:
```suggestion
    _parseMRI(String(mri));
```

static/app/utils/metrics/mri.tsx
```diff
@@ -7,21 +7,28 @@ export const DEFAULT_MRI: MRI = 'c:custom/sentry_metric@none';
...
+    return true;
+  } catch (e) {
```
priscilawebdev, line 13:
```suggestion
  } catch  {
```

ArthurKnaus: oh, is there no linting rule for that? 😮

static/app/utils/metrics/mri.tsx
ArthurKnaus: thinking about it, I should simply do a `typeof mri !== string` before calling `_parseMRI`

priscilawebdev: yes that is even better

static/app/utils/metrics/mri.tsx
priscilawebdev: not sure, but it would be a nice to have

priscilawebdev (approved):

vgrozdanic (approved):

PR #73278, opened 2024-06-25
chore(on-demand-metrics): add prefill metric

obostjancic:


iambriccardo (approved):

iambriccardo:
The metric might have too high cardinality but if we keep it temporary, we should be fine.

vgrozdanic (approved):

obostjancic:
> The metric might have too high cardinality but if we keep it temporarily, we should be fine.

According to other metrics, there are only a handful of orgs that should meet this condition, there is a limiter in place, and i will monitor the rollout of this :D

PR #73280, opened 2024-06-25
ref(replays): Use isolation scope instead of Hub in recording consumers

szokeasaurusrex:
The Hub API is deprecated as of Sentry Python SDK 2.0.0, so we should isolation scope forks instead of Hub clones in the recording consumers

ref #65869

### DIFF
--- a/src/sentry/replays/consumers/recording.py
+++ b/src/sentry/replays/consumers/recording.py
@@ -28,7 +28,7 @@ RECORDINGS_CODEC: Codec[ReplayRecording] = get_topic_codec(Topic.INGEST_REPLAYS_
 class MessageContext:
     message: bytes
     transaction: Span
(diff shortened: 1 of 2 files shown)


antonpirker (approved):

szokeasaurusrex:
I would appreciate an additional review from someone on @getsentry/replay-backend before merging this change

cmanallen (approved):

cmanallen:
@szokeasaurusrex Isolation scope is thread safe right?

szokeasaurusrex:
@cmanallen, the isolation scope is a Python contextvar, which makes it local to the current execution "context," e.g., the current thread. Changes in the scope on one thread, therefore, would not be expected to bleed into the other thread's scope

PR #73284, opened 2024-06-25
ref(trace) use event emitter to schedule events

JonasBa:
Split the rendering logic to react to events emitted by the scheduler (very similar to how profiling works). This will allow us to add independent components that can react to view changes and rerender their state however they need to and avoid react lifecycle updates where they are not needed.

By doing this we also fix a race condition where we failed to scroll to the row in the list (because we can await virtualized list initialization on initial render)

This prepares the UI work for minimap as the timeline indicators can subscribe to updates and are decoupled from the main span view

### DIFF
--- a/static/app/views/performance/newTraceDetails/trace.tsx
+++ b/static/app/views/performance/newTraceDetails/trace.tsx
@@ -499,7 +499,7 @@ export function Trace({
             >
               <div className="TraceIndicatorLabel">
                 {indicatorTimestamp > 0
-                  ? formatTraceDuration(manager.trace_view.x + indicatorTimestamp)
+                  ? formatTraceDuration(manager.view.trace_view.x + indicatorTimestamp)
                   : '0s'}
               </div>
               <div className="TraceIndicatorLine" />
--- a/static/app/views/performance/newTraceDetails/trace.tsx
+++ b/static/app/views/performance/newTraceDetails/trace.tsx
@@ -30,6 +30,10 @@ import {clamp} from 'sentry/utils/profiling/colors/utils';
 import useApi from 'sentry/utils/useApi';
 import useOrganization from 'sentry/utils/useOrganization';
 import useProjects from 'sentry/utils/useProjects';
+import type {
+  TraceEvents,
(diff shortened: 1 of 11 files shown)


static/app/views/performance/newTraceDetails/trace.tsx
```diff
@@ -206,6 +212,43 @@ export function Trace({
...
+    const onTraceSpaceChange: TraceEvents['initialize trace space'] = () => {
+      manager.recomputeTimelineIntervals();
+      manager.recomputeSpanToPXMatrix();
+      manager.draw();
+    };
+    const onDividerResize: TraceEvents['divider resize'] = view => {
+      manager.recomputeTimelineIntervals();
+      manager.recomputeSpanToPXMatrix();
+      manager.draw(view);
+    };
```
JonasBa, line 235: This is what we want the span view to react to. When it happens, we recompute the areas affected by timeline change and redraw the space

static/app/views/performance/newTraceDetails/traceRenderers/traceView.tsx
```diff
@@ -1,8 +1,78 @@
 import {mat3} from 'gl-matrix';
 
+import clamp from 'sentry/utils/number/clamp';
+
 // Computes the transformation matrix that is used to render scaled
 // elements to the DOM and draw the view.
 export class TraceView {
```
JonasBa, line 7: This was moved from the manager since the view is a separate concept that is not related to the virtualized list.

static/app/views/performance/newTraceDetails/traceRenderers/virtualizedViewManager.tsx
```diff
@@ -11,7 +11,9 @@ import type {
 } from 'sentry/views/performance/newTraceDetails/traceModels/traceTree';
```
JonasBa, line 11: A lot of the changes from this file is just code that was moved to TraceView and updates to access the view on the stored field

Abdkhan14 (approved):

PR #73286, opened 2024-06-25
fix(trace-explorer): Order traces by last seen

Zylphrex:
We need to sort by the trace last seen, not just the order of the trace ids because it's possible to match on a span in the beginning of 1 trace but at the end of another.


wmak (approved):
lgtm

PR #73287, opened 2024-06-25
feat(settings-metrics): Add extrapolate metrics field

priscilawebdev:
### Goal

- Add a toggle to the metrics settings UI to enable the metrics extrapolation feature.

**Preview:**
image

### Notes

- The feature is opt-in and therefore disabled by default.
- The UI is only visible if the organization feature flag `organizations:metrics-extrapolation` is enabled.
- Currently, the toggle is in the metrics settings of a project, but it will likely be moved elsewhere soon.
- UI tests will be created in a separate PR.
- Feedback is highly appreciated, especially regarding the copies.

Part of https://github.com/getsentry/sentry/issues/73185


static/app/views/settings/projectMetrics/extrapolationField.tsx
```diff
@@ -0,0 +1,76 @@
...
+          extrapolateMetrics: value,
+        },
+      });
+    },
+    onMutate: () => {
+      addLoadingMessage(t('Toggling metrics extrapolation'));
+    },
+    onSuccess: updatedProject => {
+      addSuccessMessage(t('Successfully toggled metrics extrapolation'));
+      ProjectsStore.onUpdateSuccess(updatedProject);
```
priscilawebdev, line 49: It appears that we are not yet using a cache for the project conf request, which is why I couldn't invalidate it here using react query

ArthurKnaus (approved):
🚀

PR #73289, opened 2024-06-25
feat(settings-metrics): Add project/organization option extrapolatedMetrics

priscilawebdev:
### Goal

Implement a flag in both the project and organization configurations to enable or disable metrics extrapolation. This toggle will initially be used in metrics settings via the frontend PR [#73287](https://github.com/getsentry/sentry/pull/73287).

### Notes

- This new functionality will be controlled by the feature flag `organizations:metrics-extrapolation`.
- While the organization setting won't be used in the frontend immediately, it is being added now to ensure Relay has all necessary components in place.


src/sentry/api/serializers/models/project.py
```diff
@@ -1005,6 +1006,16 @@ def serialize(
                 },
             }
         )
+
+        if features.has("organizations:metrics-extrapolation", obj.organization):
```
priscilawebdev, line 1010: @obostjancic @shellmayr @vgrozdanic
Could you please clarify how we are guarding project properties? The current implementation won't work because the organization is a string, which isn't what we need. Could you point me to the correct way to check for an organization feature flag in the project?

obostjancic: So since `obj` is Project, `obj.organization` should be an Organization model, so this should work.

priscilawebdev: i thought the same but when testing it locally I do not get an object 🤔

src/sentry/features/temporary.py
```diff
@@ -86,6 +86,8 @@ def register_temporary_features(manager: FeatureManager):
     manager.add("organizations:custom-metrics-experimental", OrganizationFeature, FeatureHandlerStrategy.REMOTE)
     # Enables custom metrics extraction rule endpoint
     manager.add("organizations:custom-metrics-extraction-rule", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE)
+    # Enables metrics extrapolation feature
+    manager.add("organizations:metrics-extrapolation", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE)
```
priscilawebdev, line 90: This was added to this PR to ensure the project tests pass. When registering the feature, I chose to add it to `features.py` in getsentry based on the guide:

image

tests/sentry/api/endpoints/test_organization_details.py
vgrozdanic, line 1: Can you also add a check if option is updated in a test.
For example:
```python
assert options.get("sentry:extrapolate_metrics") is True
```

tests/sentry/api/endpoints/test_project_details.py
```diff
@@ -1325,6 +1325,12 @@ def test_symbol_sources_deletion(self):
             assert resp.status_code == 200
             assert project.get_option("sentry:symbol_sources", orjson.dumps([source1]).decode())
 
+    @with_feature("organizations:metrics-extrapolation")
+    def test_extrapolate_metrics(self):
+        resp = self.get_success_response(self.org_slug, self.proj_slug, extrapolateMetrics=False)
+        assert self.project.get_option("sentry:extrapolate_metrics") is False
+        assert resp.data["extrapolateMetrics"] is False
```
vgrozdanic, line 1332: Can you please below this, in a same test case, add a request where you also update `extrapolateMetrics` to `True` and checks that everything is updated as expected. Because default value for `extrapolateMetrics` is False and there might be the situation in the future when something will change and update won't work anymore, but the test won't cover it

vgrozdanic (approved):
Looks good

PR #73297, opened 2024-06-25
Let dependabot create PRs for senty-sdk python package

antonpirker:


antonpirker:
we will not do it this way, but with using the bump-version workflow in this repository

PR #73301, opened 2024-06-25
chore(slack): remove block kit flag from get actions list in webhook

cathteng:
Finish removing the block kit flag from the Slack action webhook handler. From #70207

### DIFF
--- a/src/sentry/integrations/slack/webhooks/action.py
+++ b/src/sentry/integrations/slack/webhooks/action.py
@@ -36,7 +36,6 @@ from sentry.models.rule import Rule
 from sentry.notifications.utils.actions import BlockKitMessageAction, MessageAction
 from sentry.services.hybrid_cloud.integration import integration_service
 from sentry.services.hybrid_cloud.notifications import notifications_service
-from sentry.services.hybrid_cloud.organization import organization_service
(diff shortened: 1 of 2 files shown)


src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -715,11 +714,9 @@ def get_action_option(cls, slack_request: SlackActionRequest) -> str | None:
...
 
     @classmethod
-    def get_action_list(
-        cls, slack_request: SlackActionRequest, use_block_kit: bool
-    ) -> list[MessageAction]:
+    def get_action_list(cls, slack_request: SlackActionRequest) -> list[MessageAction]:
         action_data = slack_request.data.get("actions")
-        if use_block_kit and action_data:
+        if action_data:
             # XXX(CEO): this is here for backwards compatibility - if a user performs an action with an "older"
```
ceorourke, line 720: I think you can remove this comment now? Unless I'm reading it wrong

ceorourke (approved):

src/sentry/integrations/slack/webhooks/action.py
cathteng: idk you kept it in your PR, is action data both present in old and new slack notifications?

PR #73311, opened 2024-06-25
fix(slack): fix SlackService early return tests

cathteng:
These weren't actually testing anything, fixed the mocked logger to properly test where we early return in SlackService for notifying Slack threads of activities.

### DIFF
--- a/tests/sentry/integrations/slack/service/test_slack_service.py
+++ b/tests/sentry/integrations/slack/service/test_slack_service.py
@@ -46,10 +46,10 @@ class TestGetNotificationMessageToSend(TestCase):
         assert result == "admin@localhost archived BAR-1"
 
 
-@mock.patch("sentry.integrations.slack.service")
 class TestNotifyAllThreadsForActivity(TestCase):
     def setUp(self) -> None:
         self.service = SlackService.default()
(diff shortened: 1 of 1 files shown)


schew2381 (approved):
lgtm, typing has an issue tho

tests/sentry/integrations/slack/service/test_slack_service.py
```diff
@@ -93,6 +92,7 @@ def setUp(self) -> None:
 
     def test_none_group(self):
         self.activity.update(group=None)
+        self.service._logger = mock.MagicMock()
```
asottile-sentry, line 95: mypy was helping you here, monkeypatching a class like this leaks outside of the scope of the test

this should use a context manager so it is undone after the test finishes:

```python
with mock.patch.object(self.service, '_logger') as self.mock_logger:
    ...
```

cathteng: i get the same mypy error when i add the context manager

asottile-sentry: can you show the mypy error you're referring to?

PR #73312, opened 2024-06-25
chore(metrics): Add observability into different event builders

nikhars:
There are different dataset builders which use metrics. Would like to get some observability into the usage pattern of different builders in order to determine how often each one of them is used. The insights could influence the decision on how to approach query migration strategy when no indexer is involved.


wmak (commented):
I'm about to remove all this code in https://github.com/getsentry/sentry/pull/71710

nikhars:
> I'm about to remove all this code in #71710

Nice. Will close this PR then

PR #73315, opened 2024-06-25
feat(u2f): Await bootstrap promises

scttcper:
attempts to wait until all of the bootstrap promises are finished before making the authenticators request that sets the correct cookie

https://www.notion.so/sentry/Superuser-Staff-Race-Condition-e06af8baea9f4b81b1b4fa219702aa4e

Fixes ALRT-34

### DIFF
--- a/static/app/components/modals/sudoModal.tsx
+++ b/static/app/components/modals/sudoModal.tsx
@@ -113,10 +113,6 @@ function SudoModal({
       }
     };
 
-    if (!organizationPromise) {
-      return;
-    }
-
     getAuthenticators();
   }, [api, organizationPromise]);
 
--- a/static/app/components/modals/sudoModal.tsx
+++ b/static/app/components/modals/sudoModal.tsx
@@ -1,16 +1,15 @@
 import {Fragment, useContext, useEffect, useState} from 'react';
-import type {WithRouterProps} from 'react-router';
 import styled from '@emotion/styled';
 import trimEnd from 'lodash/trimEnd';
 
 import {logout} from 'sentry/actionCreators/account';
 import type {ModalRenderProps} from 'sentry/actionCreators/modal';
-import type {Client} from 'sentry/api';
 import {Alert} from 'sentry/components/alert';
 import {Button} from 'sentry/components/button';
 import SecretField from 'sentry/components/forms/fields/secretField';
 import Form from 'sentry/components/forms/form';
 import Hook from 'sentry/components/hook';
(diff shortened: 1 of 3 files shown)


schew2381:
I used `dev-ui` to make sure this works for an org with a lot of projects, bit hard to tell but the longest 6 second one is projects GET and bottom one is authenticators request

Screenshot 2024-06-25 at 1 55 40 PM

static/app/components/modals/sudoModal.tsx
```diff
@@ -66,13 +61,16 @@ function SudoModal({
   Body,
   closeButton,
 }: Props) {
+  const router = useRouter();
+  const api = useApi();
```
saponifi3d, line 65: appreciate this.

static/app/components/modals/sudoModal.tsx
```diff
@@ -84,29 +82,43 @@ function SudoModal({
...
+        setState(prevState => ({
+          ...prevState,
+          isLoading: false,
+        }));
+      }
     };
 
+    if (!organizationPromise) {
+      return;
+    }
```
schew2381, line 118: I believe we can delete this, we only added it previously b/c we were calling `.finally` on it.

schew2381 (commented):
lgtm but will have one more person take a look at it b/c FE isn't my specialty

static/app/components/modals/sudoModal.tsx
saponifi3d: looks like this is being used to gate when we call `getAuthenticators` is invoked and we need it because the initial value is set to null.

saponifi3d (approved):
lgtm! thanks for the quick fix and moving router / api to hooks in the process 🎉

static/app/components/modals/sudoModal.tsx
```diff
@@ -84,29 +82,43 @@ function SudoModal({
...
-            authenticators: fetchedAuthenticators ?? [],
-          }));
-        });
+        // Await all preload requests
+        await Promise.allSettled([
+          organizationPromise,
+          ...Object.values(window.__sentry_preload),
+        ]);
       } catch {
         // ignore errors
```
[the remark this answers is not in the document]
scttcper: these request always fail with 403 since you can't access the org yet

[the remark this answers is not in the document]
schew2381: oh when you have active staff the org request will actually pass b/c we need it for _admin.

PR #73318, opened 2024-06-25
feat(frontend): Implement a GlobalDrawer

leeandher:
This PR implements a global drawer, similar to current `GlobalModal` set up. It uses render props to allow call sites to determine its contents. The `DrawerPanel` itself lends a lot from [`detailPanel.tsx`](https://github.com/getsentry/sentry/blob/master/static/app/views/insights/common/components/detailPanel.tsx) bu without any of the docking features. The default appearance of drawer is a header with a close button and a body which adds padding.

image

By default, it will close when:
- escape is pressed
- outside the drawer is clicked
- location changes
- closeDrawer is invoked
- default close button is pressed

https://github.com/getsentry/sentry/assets/35509934/d42a9a38-3f26-4a10-8f68-8ad33770b687

I am building this out to accomodate some new designs which rely on a drawer for issue details page. Rather than having every call site also have a rendered empty panel on the page (as is the case for current starfish pages), they will now all be able to mutate the same global drawer.

https://github.com/getsentry/sentry/assets/35509934/7e050b8b-7fd5-4cfa-957e-d9383d1a4e0a


static/app/components/globalDrawer/components.tsx
```diff
@@ -0,0 +1,97 @@
...
+      setIsDrawerOpen(false);
+    }
+  });
+
+  useEffect(() => {
+    if (isOpen && escapeKeyPressed && closeOnEscapeKeypress) {
+      onClose?.();
+      setIsDrawerOpen(false);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
leeandher, line 42: This was carried over from [detailPanel](https://github.com/getsentry/sentry/blob/master/static/app/views/insights/common/components/detailPanel.tsx#L76). I had issues with maximum call stack errors when I added the rest of the deps, so I'm not sure what to do here.

static/app/components/globalDrawer/components.tsx
```diff
@@ -0,0 +1,97 @@
...
+  }, [escapeKeyPressed, closeOnEscapeKeypress]);
+
+  useEffect(() => {
+    setIsDrawerOpen(isOpen);
+  }, [isOpen]);
+
+  return (
+    <div role="complementary" aria-hidden={!isDrawerOpen}>
```
leeandher, line 50: Seemed like a good and likely unused accessibility role  - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/complementary_role

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+    <DrawerContainer data-test-id="drawer-container">
+      <DrawerPanel
+        isOpen={isDrawerOpen}
+        onClose={closeDrawer}
+        onOpen={options?.onOpen}
+        closeOnOutsideClick={closeOnOutsideClick}
+        closeOnEscapeKeypress={closeOnEscapeKeypress}
+      >
+        {renderedChild}
+      </DrawerPanel>
```
leeandher, line 71: Compared to the `GlobalModal`, this component but it controls whether or not it is visible. The contents however, **are removed** when the drawer is closed.

static/app/stores/drawerStore.tsx
```diff
@@ -0,0 +1,53 @@
+import {createStore} from 'reflux';
```
billyvg, line 1: Should we be adding more reflux stores?

evanpurkhiser: We don't need reflux for this. This should be a global context.

I want to move modal to be a global context, but I haven't had time.

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  closeDrawer: () => void;
+}
+
+interface GlobalDrawerProps {
+  onClose?: () => void;
+}
+
+export default function GlobalDrawer({onClose}: GlobalDrawerProps) {
+  const location = useLocation();
+  const {renderer, options} = useLegacyStore(DrawerStore);
```
evanpurkhiser, line 35: This should just use a react contexdt. No need for a reflux store.

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+    closeDrawerAction();
+    // From GlobalDrawer usage, refocus main content
+    onClose?.();
+  }, [onClose, options]);
+
+  // Close the drawer when the browser history changes.
+  //
+  // XXX: We're using useEffectAfterFirstRender primarily to support tests
+  // which render the GlobalDrawer after a drawer has already been registered in
+  // the drawer store, meaning it would be closed immediately.
```
evanpurkhiser, line 51: We don't need `useEffectAfterFirstRender` here. I only added this in the global modal because we already had a ton of test that were rendering the global modal AFTER a modal was already opened within the store, meaning on first mount it would just immediately close the modal.

Since nothing is using this in tests yet (and because this should be converted to a context anyway) we can us a regular useEffect

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  /** Callback for when the drawer opens */
+  onOpen?: () => void;
+}
+
+export interface DrawerRenderProps {
+  /** Body container for the drawer */
+  Body: typeof DrawerBody;
+  /** Close the drawer */
+  closeDrawer: () => void;
+}
```
ryan953, line 27: Please prefer `//` for comments like this. It makes it easier to comment whole chunks of code later without the extra `*/` getting in the way.

Also, vscode (if that's your editor) will insert/remove `//` if you type `CMD+/` on the current row(s)

static/app/views/app/index.tsx
```diff
@@ -228,14 +229,15 @@ function App({children, params}: Props) {
...
+  const focusMainContainer = useCallback(() => mainContainerRef.current?.focus?.(), []);
 
   return (
     <Profiler id="App" onRender={onRenderCallback}>
       <OrganizationContextProvider>
         <AsyncSDKIntegrationContextProvider>
           <MainContainer tabIndex={-1} ref={mainContainerRef}>
-            <GlobalModal onClose={handleModalClose} />
+            <GlobalModal onClose={focusMainContainer} />
+            <GlobalDrawer onClose={focusMainContainer} />
```
evanpurkhiser, line 240: Switching this to a context provider will mean the GlobalDrawer wraps everything.

static/app/actionCreators/drawer.tsx
```diff
@@ -0,0 +1,13 @@
...
+export function openDrawer(
+  renderer: (renderProps: DrawerRenderProps) => React.ReactNode,
+  options?: DrawerOptions
+) {
+  DrawerStore.openDrawer(renderer, options ?? {});
+}
+
+export function closeDrawer() {
+  DrawerStore.closeDrawer();
+}
```
scttcper, line 13: i wouldn't add any new action creators, its the same thing as using the DrawerStore directly

static/app/actionCreators/drawer.tsx
```diff
@@ -0,0 +1,13 @@
+import type {DrawerOptions, DrawerRenderProps} from 'sentry/components/globalDrawer';
+import DrawerStore from 'sentry/stores/drawerStore';
+
+export function openDrawer(
+  renderer: (renderProps: DrawerRenderProps) => React.ReactNode,
+  options?: DrawerOptions
+) {
+  DrawerStore.openDrawer(renderer, options ?? {});
```
scttcper, line 8: openDrawer should handle the optional options argument

tests/js/sentry-test/reactTestingLibrary.tsx
```diff
@@ -125,6 +126,22 @@ function renderGlobalModal(options?: Options) {
...
+function renderGlobalDrawer(options?: Options) {
+  const result = render(<GlobalDrawer />, options);
+
+  /**
+   * Helper that waits for the drawer to be hidden from the DOM. You may need to
+   * wait for the drawer to be removed to avoid any act warnings.
+   */
+  function waitForDrawerToHide() {
+    return rtl.waitFor(() => {
+      expect(rtl.screen.queryByRole('complimentary')).not.toBeInTheDocument();
```
evanpurkhiser, line 138: Let's give it a label to so we can use the `name` option. Since I can imagine there are other complimentary elements.

static/app/components/globalDrawer/components.tsx
```diff
@@ -0,0 +1,96 @@
...
+import {Button} from 'sentry/components/button';
+import type {DrawerOptions} from 'sentry/components/globalDrawer';
+import {IconClose} from 'sentry/icons/iconClose';
+import {t} from 'sentry/locale';
+import {space} from 'sentry/styles/space';
+import useKeyPress from 'sentry/utils/useKeyPress';
+import useOnClickOutside from 'sentry/utils/useOnClickOutside';
+import SlideOverPanel from 'sentry/views/insights/common/components/slideOverPanel';
+
+export interface DrawerPanelProps extends DrawerOptions {
```
scttcper, line 13: props like onClose should always be defined, i think the props should be something like `Required<DrawerOptions>` or something that is a bit closer to how its being used

static/app/components/globalDrawer/components.tsx
```diff
@@ -0,0 +1,96 @@
...
+import styled from '@emotion/styled';
+
+import {Button} from 'sentry/components/button';
+import type {DrawerOptions} from 'sentry/components/globalDrawer';
+import {IconClose} from 'sentry/icons/iconClose';
+import {t} from 'sentry/locale';
+import {space} from 'sentry/styles/space';
+import useKeyPress from 'sentry/utils/useKeyPress';
+import useOnClickOutside from 'sentry/utils/useOnClickOutside';
+import SlideOverPanel from 'sentry/views/insights/common/components/slideOverPanel';
```
malwilley, line 11: We should pull this out into the top level components directory (and perhaps into this folder?) if it's a subcomponent here.

static/app/components/globalDrawer/index.tsx
leeandher: I get what you mean, I'm going to make these multi-line to make it more transparent but these are formatted to provide the IDE with descriptions of each key, not just info for those reading this specific file. Good catch!

static/app/components/globalDrawer/components.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  useOnClickOutside(panelRef, () => {
+    if (isOpen && closeOnOutsideClick) {
+      onClose?.();
+      setIsDrawerOpen(false);
+    }
+  });
+
+  useEffect(() => {
+    if (isOpen && escapeKeyPressed && closeOnEscapeKeypress) {
```
malwilley, line 38: I wouldn't use a `useEffect` for this, it can be handled in the event handler. You can listen to the escape key in a `document.addEventListener('keydown', handler)`

static/app/components/globalDrawer/components.tsx
```diff
@@ -0,0 +1,103 @@
...
+  onOpen,
+  closeOnOutsideClick = true,
+  closeOnEscapeKeypress = true,
+}: DrawerPanelProps) {
+  const [isDrawerOpen, setIsDrawerOpen] = useState(isOpen);
+  const panelRef = useRef<HTMLDivElement>(null);
+
+  useOnClickOutside(panelRef, () => {
+    if (isOpen && closeOnOutsideClick) {
+      onClose?.();
```
scttcper, line 31: nit
```suggestion
      onClose();
```

static/app/components/globalDrawer/context.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  const ctx: DrawerContextProps = {
+    config: drawerConfig,
+    closeDrawer,
+    openDrawer,
+  };
+
+  return (
+    <DrawerContext.Provider value={ctx}>
+      <GlobalDrawer />
```
scttcper, line 61: is there any way to not render the global drawer until its used for the first time. mostly thinking about the rtl tests

leeandher: Yeah a bunch of tests were failing, I added an option (`includeDrawer`) to explcitly render the drawer when using `render`. I'm not sure how obvious it will be to find as someone writing tests, but it fixes all the tests failing from `toBeEmptyDOMElement` assertions.

I saw we were already doing something like this with organization/router context, but if this feels hacky, lmk.

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  Body: typeof DrawerBody;
+  //
+  // Close the drawer
+  //
+  closeDrawer: () => void;
+}
+
+interface GlobalDrawerProps {}
+
+export default function GlobalDrawer(_props: GlobalDrawerProps) {
```
evanpurkhiser, line 40: I think you can collapse the context provider stuff just into this component

tests/js/sentry-test/reactTestingLibrary.tsx
```diff
@@ -20,6 +21,11 @@ import {instrumentUserEvent} from '../instrumentedEnv/userEventIntegration';
 import {initializeOrg} from './initializeOrg';
 
 interface ProviderOptions {
+  /**
+   * Mounts the GlobalDrawer, and its provider.
+   * Must be specified excplitly since .toBeEmptyDOMElement assertions will fail otherwise
+   */
+  includeDrawer?: boolean;
```
scttcper, line 28: i was thinking more like another state that hides `<GlobalDrawer />` might even behave differently in tests since the animation matters a lot less. Like it renders once you open a drawer

```suggestion
  /**
   * Mounts the GlobalDrawer, and its provider.
   * Must be specified excplitly since .toBeEmptyDOMElement assertions will fail otherwise
   */
  includeDrawer?: boolean;
```

static/app/components/globalDrawer/index.tsx
evanpurkhiser: Like, the only component using this component `GlobalDrawer` is `DrawerContextProvider`. So why have this be separate and read the context, when all this components code and just be inside of `DrawerContextProvider`

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const isDrawerOpen = typeof renderer === 'function';
+  const renderedChild =
+    renderer?.({
+      Body: DrawerBody,
+      closeDrawer,
+    }) ?? null;
+
+  return (
+    <DrawerContainer data-test-id="drawer-container">
```
evanpurkhiser, line 64: Do we need to render any of this if it's not open? why?

evanpurkhiser, line 64: Possible to not have a data-test-id here? What are we using this for. Ideally we should avoid using `data-test-id`

See https://testing-library.com/docs/queries/bytestid/

> In the spirit of [the guiding principles](https://testing-library.com/docs/guiding-principles), it is recommended to use this only after the other queries don't work for your use case. Using data-testid attributes do not resemble how your software is used and should be avoided if possible. That said, they are way better than querying based on DOM structure or styling css class names. Learn more about data-testids from the blog post ["Making your UI tests resilient to change"](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change)

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    </DrawerContainer>
+  );
+}
+
+const DrawerContainer = styled('div')`
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
```
evanpurkhiser, line 83:
```suggestion
  inset: 0;
```

https://developer.mozilla.org/en-US/docs/Web/CSS/inset

static/app/components/globalDrawer/index.tsx
leeandher: That makes sense, I've refactored it into one component, thanks!

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+const DrawerContext = createContext<DrawerContextProps>(DEFAULT_DRAWER_CONTEXT);
+
+export function DrawerContextProvider({children}) {
+  const location = useLocation();
+  const [drawerConfig, setDrawerConfig] = useState<DrawerConfig>(
+    DEFAULT_DRAWER_CONTEXT.config
+  );
+  const openDrawer = useCallback(
+    (renderer, options = {}) => setDrawerConfig({renderer, options}),
+    [setDrawerConfig]
```
evanpurkhiser, line 36: I actually think you don't even need to put this in here, the eslint checker knows that `setDrawerConfig` will never change reference I think
```suggestion
    []
```

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+  );
+  const closeDrawer = useCallback(
+    () => setDrawerConfig(DEFAULT_DRAWER_CONTEXT.config),
+    [setDrawerConfig]
```
evanpurkhiser, line 40: Same here

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+        <DrawerPanel
+          isOpen={isDrawerOpen}
+          onClose={handleClose}
+          onOpen={options?.onOpen}
+          closeOnOutsideClick={closeOnOutsideClick}
+          closeOnEscapeKeypress={closeOnEscapeKeypress}
+        >
+          {renderedChild}
+        </DrawerPanel>
+      </DrawerContainer>
```
evanpurkhiser, line 78: I think you only need to render this if `drawerConfig.renderer !== null`

static/app/components/globalDrawer/components.tsx
```diff
@@ -0,0 +1,110 @@
...
+  );
+
+  useEffect(() => {
+    document.addEventListener('keydown', handleEscapePress);
+    return () => document.removeEventListener('keydown', handleEscapePress);
+  }, [handleEscapePress]);
+
+  useEffect(() => {
+    setIsDrawerOpen(isOpen);
+  }, [isOpen]);
```
evanpurkhiser, line 53: Why do you need this?

static/app/components/globalDrawer/components.tsx
```diff
@@ -0,0 +1,110 @@
...
+}: DrawerPanelProps) {
+  const [isDrawerOpen, setIsDrawerOpen] = useState(isOpen);
+  const panelRef = useRef<HTMLDivElement>(null);
+
+  useOnClickOutside(panelRef, () => {
+    if (isOpen && closeOnOutsideClick) {
+      onClose();
+      setIsDrawerOpen(false);
+    }
+  });
```
evanpurkhiser, line 34: The handler function probably should be in a `useCallback` otherwise this effect is going to re-run every render (https://github.com/getsentry/sentry/blob/6e3cddfd060aa6db961af54197718331cc7a531b/static/app/utils/useOnClickOutside.tsx#L37)

static/app/components/globalDrawer/components.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const handleEscapePress = useCallback(
+    (e: KeyboardEvent) => {
+      if (e.key === 'Escape' && isOpen && closeOnEscapeKeypress) {
+        onClose();
+        setIsDrawerOpen(false);
+      }
+    },
+    [isOpen, closeOnEscapeKeypress, onClose]
+  );
```
evanpurkhiser, line 44: We have a `useHotkeys` hook you can use here instead of manually setting up this listener.

The modal probably should use it too if it's not

Like this https://github.com/getsentry/sentry/blob/6e3cddfd060aa6db961af54197718331cc7a531b/static/app/views/settings/components/settingsSearch/index.tsx#L15

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+      closeOnOutsideClick: true,
+    },
+  },
+  openDrawer: () => {},
+  closeDrawer: () => {},
+};
+
+const DrawerContext = createContext<DrawerContextProps>(DEFAULT_DRAWER_CONTEXT);
+
+export function DrawerContextProvider({children}) {
```
evanpurkhiser, line 29: At this point, I probably wouldn't even call it a context provider (that's just an implementation detail), just call it `GlobalDrawer`

static/app/components/slideOverPanel.tsx
evanpurkhiser, line 1: I see this compoennt has the `AnimatePresence` inside of it. That really probably needs to move up into the `globalDrawer/index` component so that it can stay rendered. Then all the animation stuff will just magically work.

You might have to refactor it in a way so that existing usages that expect the `AnimatePresence` to be in this component still work (maybe just export a version w/o the animate presence)

tests/js/sentry-test/reactTestingLibrary.tsx
```diff
@@ -125,6 +136,16 @@ function renderGlobalModal(options?: Options) {
   return {...result, waitForModalToHide};
 }
 
+/**
+ * Helper that waits for the drawer to be hidden from the DOM. You may need to
+ * wait for the drawer to be removed to avoid any act warnings.
+ */
+function waitForDrawerToHide() {
+  return rtl.waitFor(() => {
+    expect(rtl.screen.queryByRole('complimentary')).not.toBeInTheDocument();
```
evanpurkhiser, line 145: Still think we should give the drawer some kind of label

```suggestion
    expect(rtl.screen.queryByRole('complimentary', {name: "Popout drawer"})).not.toBeInTheDocument();
```

static/app/components/globalDrawer/components.tsx
evanpurkhiser: This probably can be in the `GlobalDrawer` component also btw once you get rid of the state in here

static/app/components/globalDrawer/components.tsx
```diff
@@ -0,0 +1,110 @@
...
+        collapsed={!isDrawerOpen}
+        ref={panelRef}
+        onOpen={onOpen}
+      >
+        <DrawerHeader>
+          <CloseButton
+            priority="link"
+            size="zero"
+            borderless
+            data-test-id="close-drawer-control"
```
evanpurkhiser, line 68: Let's drop this teset id, you can already select this based on it's accessibility attributes.

static/app/components/globalDrawer/components.tsx
```diff
@@ -0,0 +1,110 @@
...
+    <div role="complementary" aria-hidden={!isDrawerOpen} data-test-id="drawer-content">
+      <SlideOverPanel
+        slidePosition="right"
+        collapsed={!isDrawerOpen}
+        ref={panelRef}
+        onOpen={onOpen}
+      >
+        <DrawerHeader>
+          <CloseButton
+            priority="link"
```
evanpurkhiser, line 65: Is this needed?

static/app/components/globalDrawer/index.tsx
leeandher: Moved all usage of `data-test-id` to the fake elements added in tests :+1:

tests/js/sentry-test/reactTestingLibrary.tsx
leeandher: missed this the first time, thanks for following up

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+          onOpen={options?.onOpen}
+          closeOnOutsideClick={closeOnOutsideClick}
+          closeOnEscapeKeypress={closeOnEscapeKeypress}
+        >
+          {renderedChild}
+        </DrawerPanel>
+      )}
+      {children}
+    </DrawerContext.Provider>
+  );
```
evanpurkhiser, line 71:
```suggestion
  const isDrawerOpen = renderer !== null;
  const renderedChild =
    renderer?.({
      Body: DrawerBody,
      closeDrawer: handleClose,
    }) ?? null;

  return (
    <DrawerContext.Provider value={drawerContextValue}>
      {isDrawerOpen !== null && (
        <DrawerPanel
          isOpen={isDrawerOpen}
          onClose={handleClose}
          onOpen={options?.onOpen}
          closeOnOutsideClick={closeOnOutsideClick}
          closeOnEscapeKeypress={closeOnEscapeKeypress}
        >
          {renderedChild}
        </DrawerPanel>
      )}
      {children}
    </DrawerContext.Provider>
  );
```

Probably a bit cleaner, but now it's even more questionable why `isOpen` exists lol

static/app/components/globalDrawer/components.tsx
leeandher: It's to remove the background from the button for styling when hovering. This is the appearance without it:
image

And with it:

image

static/app/components/slideOverPanel.tsx
```diff
@@ -59,6 +59,9 @@ function SlideOverPanel(
             stiffness: 500,
             damping: 50,
           }}
+          role="complementary"
+          aria-hidden={collapsed}
+          aria-label="slide-out-drawer"
```
evanpurkhiser, line 64: This should be a human-readable label. Screen readers will announce the label whit the panel appears

static/app/components/globalDrawer/components.tsx
evanpurkhiser: Let it have a background use `size="xsmall` instead of zero. This is the pattern use use for most of our buttons that look like this

static/app/components/globalDrawer/index.tsx
leeandher: It lingered from the components I mimicked from the starfish implementation. Those are always rendered so they control whether or not they're expanded. I've removed it for now but still have to do the animation fix when it unmounts :+1:

static/app/components/globalDrawer/components.tsx
leeandher: Updated! and removed state from the style components.

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const isDrawerOpen = renderer !== null;
+  const renderedChild =
+    renderer?.({
+      Body: DrawerBody,
+      closeDrawer: handleClose,
+    }) ?? null;
+
+  return (
+    <DrawerContext.Provider value={drawerContextValue}>
```
evanpurkhiser, line 77: You probably don't need this interim variable, can get rid of `TDrawerContext` then

```suggestion
    <DrawerContext.Provider value={{openDrawer, closeDrawer}}>
```

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+import {useLocation} from 'sentry/utils/useLocation';
+import useOnClickOutside from 'sentry/utils/useOnClickOutside';
+
+const DEFAULT_DRAWER_CONFIG: DrawerConfig = {
+  renderer: null,
+  options: {
+    closeOnEscapeKeypress: true,
+    closeOnOutsideClick: true,
+  },
+};
```
evanpurkhiser, line 18: Do we even need this default, what if it was just `undefined` when not set and you just had

```
  const [drawerConfig, setDrawerConfig] = useState<DrawerConfig>();
```

Then instead of checking if the render is set you can just check if the config is not undefined. Should clean up a bit more code

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const DrawerContext = createContext<TDrawerContext>(DEFAULT_DRAWER_CONTEXT);
+
+export function GlobalDrawer({children}) {
+  const location = useLocation();
+  const [drawerConfig, setDrawerConfig] = useState<DrawerConfig>(DEFAULT_DRAWER_CONFIG);
+  const openDrawer = useCallback(
+    (renderer, options = {}) => setDrawerConfig({renderer, options}),
+    []
+  );
```
evanpurkhiser, line 33: Might be good to type this correctly. It's `any` right now.

evanpurkhiser:
```suggestion
  const openDrawer = useCallback<DrawerContext['openDrawer']>(
    (renderer, options = {}) => setDrawerConfig({renderer, options}),
    []
  );
```

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const handleClose = useCallback(() => {
+    // Callsite callback when closing the drawer
+    options?.onClose?.();
+    // Actually close the drawer component
+    closeDrawer();
+  }, [options, closeDrawer]);
+
+  // Close the drawer when the browser history changes.
+  useEffect(() => closeDrawer(), [location.pathname, closeDrawer]);
```
evanpurkhiser, line 50: Should be `handleClose`?

static/app/components/globalDrawer/types.tsx
```diff
@@ -0,0 +1,46 @@
...
+   */
+  closeOnOutsideClick?: boolean;
+  /**
+   * Callback for when the drawer closes
+   */
+  onClose?: () => void;
+  /**
+   * Callback for when the drawer opens
+   */
+  onOpen?: () => void;
```
evanpurkhiser, line 19: Do we even need this? Don't they have to call `openDrawer` anyway, can't they just call whatever they would pass to that right after they call `openDrawer`?

static/app/components/globalDrawer/index.spec.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+    await userEvent.click(screen.getByTestId('drawer-test-outside'));
+
+    expect(closeSpy).toHaveBeenCalled();
+    expect(screen.queryByTestId('drawer-test-content')).not.toBeInTheDocument();
+  });
+
+  // TODO(Leander): I'll have to look into this further in a separate PR.
+  // The useHotKeys hook doesn't seem to register the userEvents for this test
+  it.skip('calls onClose handler when escape key is pressed', async function () {
```
leeandher, line 109: @evanpurkhiser Do you know why this may not be working? Manually setting the event listeners had this test passing, but with the `useHotKeys` hook it doesn't seem to be registering the `userEvent.keyboard` signal.

It works in the browser but I couldn't figure out a way to get this test passing again.

static/app/components/deprecatedDropdownMenu.spec.tsx
```diff
@@ -134,10 +134,10 @@ describe('dropdownMenuDeprecated', function () {
     );
 
     // Make sure this is only called when menu is open
-    expect(addSpy).not.toHaveBeenCalled();
+    expect(addSpy).not.toHaveBeenCalledWith('click', expect.anything(), true);
 
     await userEvent.click(screen.getByRole('button'));
-    expect(addSpy).toHaveBeenCalled();
+    expect(addSpy).toHaveBeenCalledWith('click', expect.anything(), true);
```
evanpurkhiser, line 140: Unrelated?

static/app/components/globalDrawer/components.tsx
```diff
@@ -0,0 +1,66 @@
+import {type ForwardedRef, forwardRef} from 'react';
```
evanpurkhiser, line 1: nit: No need to import tye ForwardedRef type, just use `React.ForwardedRef`

static/app/components/globalDrawer/index.spec.tsx
evanpurkhiser: hrm. I'm not sure. Might just have to investigate what `useHotKeys` is doing

static/app/components/globalDrawer/index.spec.tsx
```diff
@@ -0,0 +1,198 @@
...
+          },
+        }}
+      />
+    );
+
+    await userEvent.click(screen.getByTestId('drawer-test-open'));
+
+    const content = screen.getByTestId('drawer-test-content');
+
+    await userEvent.keyboard('{Escape}');
```
evanpurkhiser, line 183: Maybe drop a todo here to fix whatever needs fixed when the useHotKey stuff is fixed

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+        )}
+      </AnimatePresence>
+      {children}
+    </DrawerContext.Provider>
+  );
+}
+
+export default function useDrawer() {
+  return useContext(DrawerContext);
+}
```
evanpurkhiser, line 84: nit: Would be nice to have a small documentation comment above this with an example of how to use it

evanpurkhiser (approved):

evanpurkhiser:
Great work!! Glad we widdled it down to a simpler implementation :)

static/app/components/deprecatedDropdownMenu.spec.tsx
leeandher: This test was failing since the `useHotKeys` hook adds the event listener for all tests. I changed it to only look for the event listener it intends to test.

evanpurkhiser: oh can you just enable the hotkey when it's open?

leeandher: I don't think so, since the hotkey is on the provider now. If i moved it back to the panel maybe, but otherwise we can't call the hook conditionally IIRC.

evanpurkhiser: we could just add a `disabled` option to `useHotKeys` but let's not worry about it for now maybe

static/app/components/slideOverPanel.tsx
```diff
@@ -45,25 +45,24 @@ function SlideOverPanel(
...
+      exit={collapsedStyle}
+      slidePosition={slidePosition}
+      transition={{
+        type: 'spring',
+        stiffness: 500,
+        damping: 50,
+      }}
+      role="complementary"
+      aria-hidden={collapsed}
+      aria-label="slide out drawer"
```
gggritso, line 62: With a role of `"complementary"` this label is redundant. Would be nicer for screen readers to allow `ariaLabel` as a prop, so we can pass descriptions like "Query Samples" and the reader would read out something like "page landmark, complementary, Query Samples" rather than something like "page landmark, complementary, slide out drawer"

gggritso (commented):
Makes sense! A few things though:

- a global drawer makes sense, but if you ever implement docking to the bottom, it might be harder with a global element, since it's lifted out of the DOM context. For Insights, we get correct positioning for free by putting the sidebar in the right place in the DOM, so it doesn't overlap the sidebar. Handy, since the sidebar changes width and repositions itself
- also, if the panel is lifted out of the DOM, will its screen reader behaviour still make sense? You might have an "aside" that starts before the rest of the page, or way after the footer
- if Design's direction is to have a close button on the left with a text label, it makes sense that Insights drawers would look the same, right? Should I update those?
- I think having a global sidebar that exposes a hook that has to be passed a renderer function that acts like a React component is a bit of a code somersault. It's simpler to do this in a React-y way, which is through the DOM by including a `<SidePanel>` component of some kind on the page, and passing it `children` and props. I'm not seeing the benefit of having a global panel, but I trust you!

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const handleClose = useCallback(() => {
+    // Callsite callback when closing the drawer
+    drawerConfig?.options?.onClose?.();
+    // Actually close the drawer component
+    closeDrawer();
+  }, [drawerConfig, closeDrawer]);
+
+  // Close the drawer when the browser history changes.
+  useEffect(() => closeDrawer(), [location?.pathname, closeDrawer]);
```
JonasBa, line 42:
```suggestion
  useLayoutEffect(() => closeDrawer(), [location?.pathname, closeDrawer]);
```
This will trigger the close before the next paint, avoiding the chance of a flicker on the target screen

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const closeDrawer = useCallback<TDrawerContext['closeDrawer']>(
+    () => setDrawerConfig(undefined),
+    []
+  );
+
+  const handleClose = useCallback(() => {
+    // Callsite callback when closing the drawer
+    drawerConfig?.options?.onClose?.();
+    // Actually close the drawer component
+    closeDrawer();
```
JonasBa, line 38: These seem self explanatory. If the order matters, then I would add a comment, but else I think it's fine without
```suggestion
    drawerConfig?.options?.onClose?.();
    closeDrawer();
```

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,109 @@
+import {createContext, useCallback, useContext, useEffect, useRef, useState} from 'react';
+import {AnimatePresence} from 'framer-motion';
+
+import ErrorBoundary from 'sentry/components/errorBoundary';
+import {DrawerBody, DrawerPanel} from 'sentry/components/globalDrawer/components';
+import type {
+  DrawerConfig,
+  DrawerContext as TDrawerContext,
```
JonasBa, line 8: Nit (feel free to dismiss), but I havent seen us use the T conventions for types decl naming

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  openDrawer: () => {},
+  closeDrawer: () => {},
+});
+
+export function GlobalDrawer({children}) {
+  const location = useLocation();
+  const [drawerConfig, setDrawerConfig] = useState<DrawerConfig | undefined>();
+  // If no 'drawerConfig' is set, the global drawer is closed.
+  const isDrawerOpen = drawerConfig !== undefined;
+  const openDrawer = useCallback<TDrawerContext['openDrawer']>(
```
JonasBa, line 25: I assume this means that if a drawer is already open, calling open with a different config will just override and render a different drawer? From a visual experience pov, do we want to allow that or should we require for the current drawer to be closed first?

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }, [drawerConfig, isDrawerOpen, handleClose]);
+  useOnClickOutside(panelRef, handleClickOutside);
+
+  // Close the drawer when escape is pressed and options allow it.
+  const handleEscapePress = useCallback(() => {
+    const allowEscapeKeypress = drawerConfig?.options?.closeOnOutsideClick ?? true;
+    if (isDrawerOpen && allowEscapeKeypress) {
+      handleClose();
+    }
```
JonasBa, line 59: Based off the change in L24, this would just become
```suggestion
    if (drawerConfig?.options?.closeOnOutsideClick ?? true) {
      handleClose();
    }
```

static/app/components/globalDrawer/index.spec.tsx
```diff
@@ -0,0 +1,207 @@
...
+    const content = screen.getByTestId('drawer-test-content');
+    expect(content).toBeInTheDocument();
+
+    // TODO(Leander): Implement the following test case when `useHotKeys` hook allows for userEvent signals
+    // await userEvent.keyboard('{Escape}');
+    // await waitForDrawerToHide();
+
+    // expect(closeSpy).toHaveBeenCalled();
+    // expect(content).not.toBeInTheDocument();
+  });
```
JonasBa, line 139: I dont think this is testing what anything new, you already assert drawer-test-content presence in other tests. Feel free to mark the test as todo or remove it in this case. I think the only thing that isnt tested is the closing on location change, but that is trivial.

static/app/components/globalDrawer/index.spec.tsx
```diff
@@ -0,0 +1,207 @@
+import {
```
JonasBa, line 1: 🥇🥇🥇🥇🥇 For the tests

evanpurkhiser:
@gggritso

> not seeing the benefit of having a global panel

one benefit of a global panel like this is that it can properly animate out when page transitions happen, similar to the modal

static/app/components/slideOverPanel.tsx
evanpurkhiser: 👍 to a label defined by the options

static/app/components/globalDrawer/index.tsx
evanpurkhiser: We should do this in the modal too

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  // If no 'drawerConfig' is set, the global drawer is closed.
+  const isDrawerOpen = drawerConfig !== undefined;
+  const openDrawer = useCallback<TDrawerContext['openDrawer']>(
+    (renderer, options = {}) => setDrawerConfig({renderer, options}),
+    []
+  );
+  const closeDrawer = useCallback<TDrawerContext['closeDrawer']>(
+    () => setDrawerConfig(undefined),
+    []
+  );
```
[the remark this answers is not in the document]
evanpurkhiser: I think the stability of `closeDrawer` is important here no?

static/app/components/globalDrawer/index.tsx
evanpurkhiser: yeah this would just swap the content completely out.

If we want the effect of one drawer closing and another opening we would need to render elements based on some unique id so that both get rendered at once and one animates out while the other animates in

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+const DrawerContext = createContext<TDrawerContext>({
+  openDrawer: () => {},
+  closeDrawer: () => {},
+});
+
+export function GlobalDrawer({children}) {
+  const location = useLocation();
+  const [drawerConfig, setDrawerConfig] = useState<DrawerConfig | undefined>();
+  // If no 'drawerConfig' is set, the global drawer is closed.
+  const isDrawerOpen = drawerConfig !== undefined;
```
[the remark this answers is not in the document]
evanpurkhiser: I suggested the `!== undefined` just to be explicit, kind of stylistic nit I suppose lol

gggritso:
@evanpurkhiser ah, for full route transitions to a completely different page? Yeah makes sense! In Insights, we don't need to worry about that because the panel is always present in the routes where it must animate in/out. I guess for this kind of sidebar maybe that doesn't apply?

static/app/components/globalDrawer/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  useOnClickOutside(panelRef, handleClickOutside);
+
+  // Close the drawer when escape is pressed and options allow it.
+  const handleEscapePress = useCallback(() => {
+    const allowEscapeKeypress = drawerConfig?.options?.closeOnOutsideClick ?? true;
+    if (isDrawerOpen && allowEscapeKeypress) {
+      handleClose();
+    }
+  }, [drawerConfig, isDrawerOpen, handleClose]);
+  useHotkeys([{match: 'Escape', callback: handleEscapePress}], [handleEscapePress]);
```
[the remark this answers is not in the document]
evanpurkhiser: good question, I think that last parameter is for dependencies.

leeandher:
Thanks for the reviews everyone! way better than it started 💪

PR #73321, opened 2024-06-25
feat(query-builder): Add absolute date picker for date filters

malwilley:
Sorry for the large diff, much of this is copying over an existing component from SmartSearchBar (the original will be deleted along with that component when it is no longer needed).

Requires https://github.com/getsentry/sentry/pull/73324 to be merged first

https://github.com/getsentry/sentry/assets/10888943/cc9be863-b83a-4808-b798-ae629e6b2e7e


static/app/components/searchQueryBuilder/specificDatePicker.tsx
malwilley, line 1: This was mostly copied from SearchBarDatePicker in SmartSearchBar, but needed enough changes that it should be its own component. The original will be deleted when SmartSearchBar is anyway.

static/app/components/searchSyntax/grammar.pegjs
```diff
@@ -339,8 +339,8 @@ ms_format   = [0-9] [0-9]? [0-9]? [0-9]? [0-9]? [0-9]?
 tz_format   = [+-] num2 ":" num2
 
 iso_8601_date_format
-  = date_format time_format? ("Z" / tz_format)? &end_value {
-      return tc.tokenValueIso8601Date(text());
+  = date:date_format time:time_format? tz:("Z" / tz_format)? &end_value {
```
malwilley, line 342: Adding names to these parameters so that we can put them into the token data object to be used in the query builder.

malwilley: I'm actually going to pull this out into a separate PR to make this diff a bit more bearable

static/app/components/searchQueryBuilder/specificDatePicker.tsx
```diff
@@ -0,0 +1,319 @@
...
+            </TimeUtcWrapper>
+          </ControlsWrapper>
+          <ButtonsFooter>
+            <ButtonBar gap={1}>
+              <Button
+                size="xs"
+                icon={<IconArrow direction="left" />}
+                onClick={handleBack}
+              >
+                Back
```
scttcper, line 191:
```suggestion
                {t('Back')}
```

static/app/components/searchQueryBuilder/specificDatePicker.tsx
```diff
@@ -0,0 +1,319 @@
...
+              </Button>
+              <Button
+                size="xs"
+                priority="primary"
+                disabled={!dateString}
+                onClick={() => {
+                  handleSave?.(dateString!);
+                }}
+              >
+                Save
```
scttcper, line 201:
```suggestion
                {t('Save'})
```

scttcper (approved):
it hurts to approve popperjs

PR #73323, opened 2024-06-25
feat(discover): Use split query decision to populate dataset slector

shruthilayaj:
Use the split query decision from the events endpoint to populate
the dataset selector


static/app/views/discover/savedQuery/utils.tsx
```diff
@@ -242,3 +246,18 @@ export function displayModeToDisplayType(displayMode: DisplayModes): DisplayType
       return DisplayType.LINE;
   }
 }
+
+export function getDataset(
+  location: Location,
+  savedQuery: SavedQuery | undefined,
+  splitDecision?: string
```
edwardgou-sentry, line 253: Would it be difficult to type `splitDecision` as a `SavedQueryDatasets`? I noticed it's being passed around as a string in the other files, and in this function we eventually force it to `SavedQueryDatasets`.

edwardgou-sentry (approved):
lgtm, left 1 nit, feel free to address or not!

PR #73325, opened 2024-06-25
nit: include query_extra for metric alert notifications

nhsiehgit:
we're constructing `query_extra` in a few spots now.

This consolidates the query_extra string all to utilize a single method to derive the query extra from the existing snuba_query and subscription

NOTE:

There are a handful of areas where we are referencing `snuba_query.query` _without_ taking into account the `query_extra`

This might possibly lead to discrepancies if we are updating the snuba_query `query`, comparing queries, or executing one off snuba queries 🤔 

eg. 
- snuba/tasks.py::update_subscription_in_snuba
- incidents/logic.py::update_alert_rule
- subscription_processor.py::get_comparison_aggregation_value
etc.etc.

TODO: we may want to audit all uses and determine whether we should be constructing the query from the subscription rather than the `snuba_query` 🤔

### DIFF
--- a/src/sentry/incidents/charts.py
+++ b/src/sentry/incidents/charts.py
@@ -22,7 +22,8 @@ from sentry.models.organization import Organization
 from sentry.models.user import User
 from sentry.snuba.dataset import Dataset
 from sentry.snuba.entity_subscription import apply_dataset_query_conditions
-from sentry.snuba.models import SnubaQuery
+from sentry.snuba.models import QuerySubscription, SnubaQuery
+from sentry.snuba.utils import build_query_extra
 
 CRASH_FREE_SESSIONS = "percentage(sessions_crashed, sessions) AS _crash_rate_alert_aggregate"
 CRASH_FREE_USERS = "percentage(users_crashed, users) AS _crash_rate_alert_aggregate"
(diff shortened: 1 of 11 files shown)


src/sentry/incidents/charts.py
```diff
@@ -162,8 +163,7 @@ def build_metric_alert_chart(
     end: str | None = None,
     user: Optional["User"] = None,
     size: ChartSize | None = None,
-    query_extra: str | None = None,
-    project_id: int | None = None,
+    subscription: QuerySubscription | None = None,
```
nhsiehgit, line 166: rather than passing query_extra here and relying on implementations to construct the query_extra, instead we accept the subscription and construct the query_extra within the chart builder.

src/sentry/integrations/metric_alerts.py
```diff
@@ -217,6 +221,8 @@ def metric_alert_attachment_info(
     if latest_incident:
         last_triggered_date = latest_incident.date_started
 
+    activation = selected_incident.activation
+    # TODO: determine whether activated alert data is useful for integration messages
```
nhsiehgit, line 225: preemptively include activated alert data into the metric alert attachment info

static/app/views/alerts/rules/metric/details/metricActivity.tsx
```diff
@@ -77,6 +77,7 @@ function MetricAlertActivity({organization, incident}: MetricAlertActivityProps)
               }/releases/${encodeURIComponent(activation.activator)}/`,
               query: {project: project},
             }}
+            style={{textOverflow: 'ellipsis', overflowX: 'inherit'}}
```
nhsiehgit, line 80: oops. this is an unrelated change to make the alert row slightly nicer.

shouldn't have any effect on the rest of the PR...

I can break this out if anyone feels strongly

src/sentry/snuba/utils.py
```diff
@@ -32,3 +33,13 @@
 
 def get_dataset(dataset_label: str) -> Any | None:
     return DATASET_OPTIONS.get(dataset_label)
+
+
+def build_query_extra(subscription: QuerySubscription | None, snuba_query: SnubaQuery) -> str:
```
saponifi3d, line 38: nit: might be nice to add some tests to this as we break it out.

saponifi3d (approved):
:shipit: thanks for doing this, i feel like this is a really great cleanup pr and does a great job of DRYing up the code.

PR #73326, opened 2024-06-25
feat(frontend): Add timestamp on issue public share page

### LINKED ISSUE #71642: Include Timestamp on Issue Public Share page
### Problem Statement

When I share an event with a public link, there is no timestamp, which makes it a lot less useful. Could you add the event timestamp to the public share page?

### Solution Brainstorm

_No response_

### Product Area

Issues

shayna-ch:
Added timestamp of latest event on the issue public share page. image
The timestamp shows local time, and a tooltip with date occurred and date received in UTC appears on hover. 

resolves #71642

### DIFF
--- a/static/app/views/sharedGroupDetails/sharedGroupHeader.tsx
+++ b/static/app/views/sharedGroupDetails/sharedGroupHeader.tsx
@@ -51,7 +51,7 @@ function SharedGroupHeader({group}: Props) {
                   title={<EventCreatedTooltip event={event} />}
                   overlayStyle={{maxWidth: 300}}
                 >
-                  <DateTime date={date.toLocaleString()} />
+                  <DateTime date={date} />
                 </Tooltip>
               </EventTimeLabel>
             </TimeStamp>
--- a/static/app/views/sharedGroupDetails/sharedGroupHeader.tsx
+++ b/static/app/views/sharedGroupDetails/sharedGroupHeader.tsx
@@ -20,7 +20,7 @@ type Props = {
 
 function SharedGroupHeader({group}: Props) {
   const date = group.latestEvent?.dateReceived ?? group.latestEvent?.dateCreated;
-  const d = new Date(date as string);
+  const date_obj = new Date(date as string);
   const event = group.latestEvent as Event;
   return (
     <Wrapper>
@@ -55,7 +55,7 @@ function SharedGroupHeader({group}: Props) {
                       title={<EventCreatedTooltip event={event} />}
                       overlayStyle={{maxWidth: 300}}
                     >
-                      <DateTime date={d.toLocaleString()} />
+                      <DateTime date={date_obj.toLocaleString()} />
                     </Tooltip>
                   ),
                 })}
--- a/static/app/views/sharedGroupDetails/sharedGroupHeader.tsx
+++ b/static/app/views/sharedGroupDetails/sharedGroupHeader.tsx
@@ -8,10 +8,8 @@ import ShortId from 'sentry/components/shortId';
 import {Tooltip} from 'sentry/components/tooltip';
 import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
-import type {Event} from 'sentry/types/event';
 import type {Group} from 'sentry/types/group';
 import {IssueCategory} from 'sentry/types/group';
-import getDynamicText from 'sentry/utils/getDynamicText';
 import EventCreatedTooltip from 'sentry/views/issueDetails/eventCreatedTooltip';
 
 type Props = {
@@ -19,9 +17,11 @@ type Props = {
 };
 
 function SharedGroupHeader({group}: Props) {
-  const date = group.latestEvent?.dateReceived ?? group.latestEvent?.dateCreated;
-  const date_obj = new Date(date as string);
-  const event = group.latestEvent as Event;
+  const date = new Date(
+    (group.latestEvent?.dateCreated ?? group.latestEvent?.dateReceived) as string
+  );
+  const event = group.latestEvent;
+
   return (
     <Wrapper>
       <Details>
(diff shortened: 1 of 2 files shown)


vartec:
QQ: why UTC and not local TZ?

static/app/views/sharedGroupDetails/sharedGroupHeader.tsx
```diff
@@ -33,6 +41,27 @@ function SharedGroupHeader({group}: Props) {
...
+                {getDynamicText({
+                  fixed: 'Jan 1, 12:00 AM',
+                  value: (
+                    <Tooltip
+                      isHoverable
+                      showUnderline
+                      title={<EventCreatedTooltip event={event} />}
+                      overlayStyle={{maxWidth: 300}}
+                    >
+                      <DateTime date={date_obj.toLocaleString()} />
```
MichaelSun48, line 58: I think we can avoid lines 22 and 23 by passing in the raw date time string into the `<DateTime>` component like this:

```
<DateTime
  date={
    group.latestEvent?.dateReceived ??
    group.latestEvent?.dateCreated
  }
/>
```

Also nit: I saw that we prefer dateCreated over dateReceived in the issue details page ([here](https://github.com/getsentry/sentry/blob/469ad6afb896aa6e67e1bd58a6fe54959e1386b2/static/app/views/issueDetails/groupEventCarousel.tsx#L417)), should we maybe switch the order just to keep things consistent?

static/app/views/sharedGroupDetails/sharedGroupHeader.tsx
```diff
@@ -1,19 +1,27 @@
...
+import EventCreatedTooltip from 'sentry/views/issueDetails/eventCreatedTooltip';
 
 type Props = {
   group: Group;
 };
 
 function SharedGroupHeader({group}: Props) {
+  const date = group.latestEvent?.dateReceived ?? group.latestEvent?.dateCreated;
+  const date_obj = new Date(date as string);
+  const event = group.latestEvent as Event;
```
MichaelSun48, line 24: Good thinking to cast `group.latestEvent` as an Event to solve the type checking error you got, but this could lead to unexpected consequences down the road. This is because `group.latestEvent` has type `Event | undefined`, which means it _could_ be `undefined`. If it is, then your component will error out since it will attempt to get the `dateCreated` attribute of `undefined` on line 47. 

This is a tricky situation: the `group` object this component receives isn't guaranteed to have a `latestEvent`, but we depend on `latestEvent` to get the last seen timestamp. Perhaps we can somehow modify this component to account for this situation?  Let me know if any of that could use some clarification!

MichaelSun48 (commented):
Just a missed edge case and a nit, but otherwise this looks great to me! Huge props for getting this out despite having never worked with React and Javascript before 💪

static/app/views/sharedGroupDetails/sharedGroupHeader.tsx
```diff
@@ -33,6 +41,27 @@ function SharedGroupHeader({group}: Props) {
...
           </ShortIdWrapper>
+
+          <TimeStamp data-test-id="sgh-timestamp">
+            {'Last seen '}
+            {(event.dateCreated ?? event.dateReceived) && (
+              <EventTimeLabel>
+                {getDynamicText({
+                  fixed: 'Jan 1, 12:00 AM',
+                  value: (
+                    <Tooltip
```
scttcper, line 52: we don't need to use `getDynamicText` anymore, just use everything inside the value starting with <Tooltip

static/app/views/sharedGroupDetails/sharedGroupHeader.tsx
```diff
@@ -81,3 +110,14 @@ const Title = styled('h3')`
     font-size: ${p => p.theme.headerFontSize};
   }
 `;
+
+const TimeStamp = styled('div')`
+  color: ${p => p.theme.headingColor};
+  font-size: ${p => p.theme.fontSizeMedium};
+  line-height: ${p => p.theme.text.lineHeightHeading};
+  margin-top: ${space(0)};
```
scttcper, line 118: should be able to omit this
```suggestion
```

static/app/views/sharedGroupDetails/sharedGroupHeader.tsx
```diff
@@ -33,6 +41,27 @@ function SharedGroupHeader({group}: Props) {
               />
             )}
           </ShortIdWrapper>
+
+          <TimeStamp data-test-id="sgh-timestamp">
+            {'Last seen '}
```
scttcper, line 46: add the translation wrapper on text
```suggestion
            {t('Last seen ')}
```

shayna-ch:
> QQ: why UTC and not local TZ?

The tooltip on the main issues page is in UTC, so thought I would do the same for consistency. The timestamp shows local time zone though.

static/app/views/sharedGroupDetails/sharedGroupHeader.tsx
```diff
@@ -81,3 +104,13 @@ const Title = styled('h3')`
...
+
+const TimeStamp = styled('div')`
+  color: ${p => p.theme.headingColor};
+  font-size: ${p => p.theme.fontSizeMedium};
+  line-height: ${p => p.theme.text.lineHeightHeading};
+`;
+
+const EventTimeLabel = styled('span')`
+  color: ${p => p.theme.subText};
+`;
```
MichaelSun48, line 116:
```suggestion
const TimeStamp = styled('div')`
  color: ${p => p.theme.headingColor};
  font-size: ${p => p.theme.fontSizeMedium};
  line-height: ${p => p.theme.text.lineHeightHeading};
  margin-top: ${space(0.25)};
`;

const EventTimeLabel = styled('span')`
  color: ${p => p.theme.subText};
  margin-left: ${space(0.25)};
`;
```

Added  \*slightly\* more margin to the top and left of the timestamp just to improve readability:

Before: 
image

After: 
image

MichaelSun48 (approved):
Small formatting nit but otherwise lgtm!

static/app/views/sharedGroupDetails/sharedGroupHeader.tsx
```diff
@@ -33,6 +41,21 @@ function SharedGroupHeader({group}: Props) {
...
+            <TimeStamp data-test-id="sgh-timestamp">
+              {t('Last seen ')}
+              <EventTimeLabel>
+                <Tooltip
+                  isHoverable
+                  showUnderline
+                  title={<EventCreatedTooltip event={event} />}
+                  overlayStyle={{maxWidth: 300}}
+                >
+                  <DateTime date={date.toLocaleString()} />
```
scttcper, line 54: don't pass locale strings around should be able to pass the date object or toIsoString
```suggestion
                  <DateTime date={date} />
```

keithbrink:
So helpful, thank you!

PR #73327, opened 2024-06-25
ref(slack): use sdk webhook client to respond to archive/resolve modals

cathteng:
Use the Slack SDK webhook client to respond to actions taken in the archive and resolve modals. "Respond" meaning we update the Slack message a user took the action on.

### DIFF
--- a/tests/sentry/integrations/slack/service/test_slack_service.py
+++ b/tests/sentry/integrations/slack/service/test_slack_service.py
@@ -92,59 +92,64 @@ class TestNotifyAllThreadsForActivity(TestCase):
 
     def test_none_group(self):
         self.activity.update(group=None)
-        self.service._logger = mock.MagicMock()
-        self.service.notify_all_threads_for_activity(activity=self.activity)
-        self.service._logger.info.assert_called_with(
-            "no group associated on the activity, nothing to do",
-            extra={"activity_id": self.activity.id},
-        )
+
+        with mock.patch.object(self.service, "_logger") as mock_logger:
+            self.service.notify_all_threads_for_activity(activity=self.activity)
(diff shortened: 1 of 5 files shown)


tests/sentry/integrations/slack/service/test_slack_service.py
```diff
@@ -92,59 +92,64 @@ def setUp(self) -> None:
...
-            "no group associated on the activity, nothing to do",
-            extra={"activity_id": self.activity.id},
-        )
+
+        with mock.patch.object(self.service, "_logger") as mock_logger:
+            self.service.notify_all_threads_for_activity(activity=self.activity)
+            mock_logger.info.assert_called_with(
+                "no group associated on the activity, nothing to do",
+                extra={"activity_id": self.activity.id},
+            )
```
cathteng, line 101: just fixing some tests in this file

tests/sentry/integrations/slack/service/test_slack_service.py
```diff
@@ -92,59 +92,64 @@ def setUp(self) -> None:
...
     def test_none_group(self):
         self.activity.update(group=None)
-        self.service._logger = mock.MagicMock()
-        self.service.notify_all_threads_for_activity(activity=self.activity)
-        self.service._logger.info.assert_called_with(
-            "no group associated on the activity, nothing to do",
-            extra={"activity_id": self.activity.id},
-        )
+
+        with mock.patch.object(self.service, "_logger") as mock_logger:
```
RyanSkonnord, line 96: The fact that we're patching it in tests makes `_logger` effectively part of the module's public contract. IMHO that would justify removing the leading underscore.

RyanSkonnord (approved):

PR #73329, opened 2024-06-25
ref(querybuilder): Move BaseQueryBuilder out

wmak:
- Move `BaseQueryBuilder` into its own file since it should not be tied specifically to Discover
- Nearly functionally identical, any functional changes are contained to 37837ab828015e9a9bb06f18f57a60b398ba6eac


src/sentry/search/events/builder/__init__.py
```diff
@@ -1,5 +1,5 @@
+from .base import BaseQueryBuilder  # NOQA
```
vartec, line 1: Why does it need NOQA? AFAIK we don't warn on relative imports, and this shouldn't warn of unused import, because it's in `__all__`

src/sentry/search/events/datasets/base.py
```diff
@@ -15,6 +18,10 @@ class DatasetConfig(abc.ABC):
     non_nullable_keys: set[str] = set()
     missing_function_error: ClassVar[type[Exception]] = InvalidSearchQuery
 
+    @abc.abstractmethod
+    def __init__(self, builder: "BaseQueryBuilder"):
```
vartec, line 22: You've imported it above:

```suggestion
    def __init__(self, builder: BaseQueryBuilder):
```

vartec (approved):

src/sentry/search/events/datasets/base.py
wmak: That doesn't work, its only imported for type checking

src/sentry/search/events/builder/__init__.py
wmak: 🤷 just following the template, pretty sure this wasn't working before

asottile-sentry, line 1: please use absolute imports instead of reexports in `__init__.py` -- this makes it really easy to end up with import cycles and impacts startup time / test time since it makes it much more likely that any import pulls in the whole app

src/sentry/search/events/datasets/base.py
asottile-sentry: add `from __future__ import annotations` to the file

src/sentry/search/events/builder/__init__.py
wmak: I'll remove this line for now, and the others in a follow up PR

PR #73331, opened 2024-06-25
[CapMan visibility] add CapMan metadata as tags to Sentry spans

xurui-c:
**if your local tests are failing, see https://github.com/getsentry/sentry/pull/73331#issuecomment-2195065684**

https://getsentry.atlassian.net/browse/SNS-2798

Allocation policies are our mechanism for doing traffic management for Snuba queries. Currently, the result of applying allocation policies in the internal API is simply accept/reject/throttle. In a [previous PR](https://github.com/getsentry/snuba/pull/6045), capacity management metadata are sent over to Sentry within the payload. We want to reflect the quota metadata in Sentry spans as well. This PR adds metadata as tags to Sentry spans.

### DIFF
--- a/src/sentry/utils/snuba.py
+++ b/src/sentry/utils/snuba.py
@@ -1025,6 +1025,14 @@ def _bulk_snuba_query(
                 raise SnubaError("Failed to parse snuba error response")
             raise UnexpectedResponseError(f"Could not decode JSON response: {response.data!r}")
 
+        if "quota_allowance" in body and body["quota_allowance"]:
+            quota_allowance_summary = body["quota_allowance"]["summary"]
+            span.set_tag("threads_used", quota_allowance_summary["threads_used"])
+            for k, v in quota_allowance_summary["throttled_by"].items():
+                span.set_tag(k, v)
+            for k, v in quota_allowance_summary["rejected_by"].items():
+                span.set_tag(k, v)
+
         if response.status != 200:
             _log_request_query(snuba_param_list[index][0])
             metrics.incr(


evanh (approved):

src/sentry/utils/snuba.py
```diff
@@ -1025,6 +1025,18 @@ def _bulk_snuba_query(
                 raise SnubaError("Failed to parse snuba error response")
             raise UnexpectedResponseError(f"Could not decode JSON response: {response.data!r}")
 
+        if "quota_allowance" in body and body["quota_allowance"]:
+            with sentry_sdk.start_span(
+                op="snuba_query",
+                description=query_referrer,
+            ) as span:
```
volokluev, line 1032: You don't need this. The span is already defined up top. You can simply attach these things to the existing span
```suggestion
```

src/sentry/utils/snuba.py
```diff
@@ -1025,6 +1025,18 @@ def _bulk_snuba_query(
...
+                quota_allowance_summary = body["quota_allowance"]["summary"]
+                span.set_tag("threads_used", quota_allowance_summary["threads_used"])
+                for k, v in quota_allowance_summary["throttled_by"].items():
+                    span.set_tag(k, v)
+                for k, v in quota_allowance_summary["rejected_by"].items():
+                    span.set_tag(k, v)
+
         if response.status != 200:
```
volokluev, line 1040: are you going to add the details to the error case in this PR or a separate one?

xurui-c: what do you mean by error case? if you're referring to adding the details when `response.status == 429`, then `error["message"]` already contains the summary. On Sentry issues, the summary is currently displayed

Screenshot 2024-06-26 at 3 26 23 PM

volokluev: some of these things should be tags. e.g.

threads_used, rejected_by_policy, rejected_by_quota_used, rejected_by_quota_unit

that makes them searchable in the sentry product

volokluev (approved):

src/sentry/utils/snuba.py
```diff
@@ -1025,6 +1025,14 @@ def _bulk_snuba_query(
                 raise SnubaError("Failed to parse snuba error response")
             raise UnexpectedResponseError(f"Could not decode JSON response: {response.data!r}")
 
+        if "quota_allowance" in body and body["quota_allowance"]:
+            quota_allowance_summary = body["quota_allowance"]["summary"]
```
sentaur-athena, line 1029: Looks like here if "quota_allowance" exists it doesn't necessary mean that the "summary" exists as well. At least not in the test.

`File "/Users/athena1/code/sentry/src/sentry/utils/snuba.py", line 1036, in _bulk_snuba_query
    quota_allowance_summary = body["quota_allowance"]["summary"]
                              ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
KeyError: 'summary'`

sentaur-athena (commented):
@xurui-c there is a test failing in master which looks related to this PR.
`pytest tests/apidocs/endpoints/releases/test_organization_releases.py::OrganizationReleasesDocsTest::test_get`

sentaur-athena:
Hm! I realized this might be an issue with my dev setup. Since I can't repro in CI. Nevermind.

wmak:
If anyone else lands in here cause your local tests are failing, you gotta update your local sentry with something like
```sentry devservices down && sentry devservices up```

PR #73333, opened 2024-06-25
chore(slack): Use Slack SDK for SlackEventEndpoints

iamrajjoshi:
Updated SlackEventEndpoint to use the the SlackSdkClient.


src/sentry/integrations/slack/webhooks/event.py
```diff
@@ -39,17 +47,28 @@ class SlackEventEndpoint(SlackDMEndpoint):
...
     slack_request_class = SlackEventRequest
 
     def reply(self, slack_request: SlackDMRequest, message: str) -> Response:
         payload = {"channel": slack_request.channel_id, "text": message}
-        client = SlackClient(integration_id=slack_request.integration.id)
-        try:
-            client.post("/chat.postMessage", data=payload, json=True)
-        except ApiError as e:
-            logger.error("slack.event.on-message-error", extra={"error": str(e)})
+        if options.get("slack.event-endpoint-sdk"):
```
iamrajjoshi, line 59: I am using an option here which isn't scoped to an org b/c adding an extra 1/2 RPC call will take longer and i am afraid we will start timing out.

src/sentry/integrations/slack/webhooks/event.py
```diff
@@ -109,19 +181,23 @@ def on_link_shared(self, request: Request, slack_request: SlackDMRequest) -> boo
...
+            "user_id": slack_request.user_id,
+            "channel": slack_request.channel_id,
+            **data,
+        }
+
         # An unfurl may have multiple links to unfurl
         for item in data.get("links", []):
             try:
                 url = item["url"]
-                slack_shared_link = parse_link(url)
```
iamrajjoshi, line 116: deleted this b/c it created a log for every unfurl and we don't track this currently.

src/sentry/integrations/slack/webhooks/event.py
cathteng: how do you plan to LA this?

src/sentry/integrations/slack/webhooks/event.py
```diff
@@ -195,13 +273,25 @@ def on_link_shared(self, request: Request, slack_request: SlackDMRequest) -> boo
...
-            logger.exception(
-                "slack.event.unfurl-error", extra={"error": str(e), "payload": payload}
-            )
+        if features.has("organizations:slack-sdk-slack-event-endpoint", organization):
+            client = SlackSdkClient(integration_id=slack_request.integration.id)
+            try:
+                client.chat_unfurl(
+                    channel=data["channel"],
+                    ts=data["message_ts"],
+                    unfurls=orjson.dumps(results).decode(),
```
cathteng, line 282: instead of dumping twice, we could also store this in a variable

tests/sentry/integrations/slack/webhooks/events/test_discover_link_shared.py
```diff
@@ -68,6 +104,60 @@ def share_discover_links(self, mock_match_link, mock_):
...
+    @patch(
+        "sentry.integrations.slack.webhooks.event.link_handlers",
+        {
+            LinkType.DISCOVER: Handler(
+                matcher=[re.compile(r"test")],
+                arg_mapper=make_type_coercer({}),
+                fn=Mock(return_value={"link1": "unfurl", "link2": "unfurl"}),
+            )
+        },
+    )
```
cathteng, line 127: if you have the same patch decorators for most functions in a test class, you can move them to the top of the class and it will apply to every function

RyanSkonnord (approved):

tests/sentry/integrations/slack/webhooks/events/test_discover_link_shared.py
iamrajjoshi: ill do this when i ga and delete the old client tests

src/sentry/integrations/slack/webhooks/event.py
```diff
@@ -39,17 +47,39 @@ class SlackEventEndpoint(SlackDMEndpoint):
...
+
         payload = {"channel": slack_request.channel_id, "text": message}
-        client = SlackClient(integration_id=slack_request.integration.id)
-        try:
-            client.post("/chat.postMessage", data=payload, json=True)
-        except ApiError as e:
-            logger.error("slack.event.on-message-error", extra={"error": str(e)})
+        if options.get("slack.event-endpoint-sdk") and slack_request.integration.id in options.get(
+            "slack.event-endpoint-sdk-integration-ids"
+        ):
```
cathteng, line 70: i think this should be OR
either the feature is GA'd OR the org is part of LA.

can you add a test for just the LA part?

src/sentry/integrations/slack/webhooks/event.py
```diff
@@ -79,11 +110,37 @@ def prompt_link(self, slack_request: SlackDMRequest) -> None:
...
+        if options.get("slack.event-endpoint-sdk") or slack_request.integration.id in options.get(
+            "slack.event-endpoint-sdk-integration-ids"
+        ):
+            client = SlackSdkClient(integration_id=slack_request.integration.id)
+            try:
+                client.chat_postEphemeral(
+                    channel=slack_request.channel_id,
+                    user=slack_request.user_id,
+                    text=payload["text"],
+                    **SlackPromptLinkMessageBuilder(associate_url).as_payload(),
```
cathteng, line 131: nit: this message builder only returns `{"blocks": ...}`, can we just get the blocks out of the payload instead of building them again?

cathteng (approved):

PR #73334, opened 2024-06-25
feat(billing): Introduce Outcomes Dead Letter Queue type

dashed:
Introduce `Topic.OUTCOMES_DLQ` and `Topic.OUTCOMES_BILLING_DLQ` to be used for the dead letter queue on the outcomes consumer.

For https://github.com/getsentry/getsentry/pull/14437


lynnagara (commented):
Lgtm, but all topics need to have their configuration defined in https://github.com/getsentry/sentry-kafka-schemas  before they can be used

lynnagara (approved):

PR #73343, opened 2024-06-25
feat(replay): add playback speed for mobile

### LINKED ISSUE #67842: [Replay Details] Playback speed functionality
- Right now playback speed doesn't work for video replays
- Timer needs to also account for playback speed

michellewzhang:
- closes https://github.com/getsentry/sentry/issues/67842
- i disabled fast forward for now
- tested a bunch of scenarios:

playing all the way through on 8x:

https://github.com/getsentry/sentry/assets/56095982/01d9562d-bb74-4350-bedc-370ad6f281aa

seeking through the replay, pausing, and trying various speeds:

https://github.com/getsentry/sentry/assets/56095982/5496fa7c-3d8e-4a77-822a-99bf8944c684

seeking throughout the video on one speed:

https://github.com/getsentry/sentry/assets/56095982/5db16aea-de91-4a5f-b7c0-30a3910acd0b

playing thru video with various speeds (no pause):

https://github.com/getsentry/sentry/assets/56095982/b1cb5c4c-783e-4678-97a6-24861f0e8b4c

### DIFF
--- a/static/app/utils/replays/timer.tsx
+++ b/static/app/utils/replays/timer.tsx
@@ -34,7 +34,7 @@ export class Timer extends EventTarget {
    */
   start(seconds?: number) {
     this._pausedAt = 0;
-    this._start = window.performance.now() - (seconds ?? 0);
+    this._start = window.performance.now() - (seconds ?? 0) / this._speed;
     this.resume();
     this.step();
   }
--- a/static/app/utils/replays/timer.tsx
+++ b/static/app/utils/replays/timer.tsx
@@ -9,13 +9,14 @@ export class Timer extends EventTarget {
   private _pausedAt: number = 0;
   private _additionalTime: number = 0;
   private _callbacks: Map<number, (() => void)[]> = new Map();
+  private _speed: number = 1;
 
   step = () => {
     if (!this._active) {
       return;
     }
 
-    this._time = window.performance.now() - this._start;
+    this._time = (window.performance.now() - this._start) * this._speed;
     // We don't expect _callbacks to be very large, so we can deal with a
     // linear search
     this._callbacks.forEach((value, key, callbacks) => {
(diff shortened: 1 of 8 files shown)


c298lee (approved):
everything lgtm! 🚀

aliu39 (approved):

static/app/utils/replays/timer.tsx
```diff
@@ -33,7 +34,7 @@ export class Timer extends EventTarget {
    */
   start(seconds?: number) {
     this._pausedAt = 0;
-    this._start = window.performance.now() - (seconds ?? 0);
+    this._start = window.performance.now() - (seconds ?? 0) / this._speed;
```
ryan953, line 37: I don't understand why this is here. It seems like `start(seconds)` should just set the time directly. I don't follow why we'd adjust that requested timestamp, like if the video is paused we'd still want to start at `seconds`. I'm probably just missing something about the math, or about what `_start` does i guess.

michellewzhang: i'll add a comment explaining the math!

ryan953:
```
// we're dividing by the speed here because we mutiply
    // by the same factor up in step(). doing the math,
    // you'll see that this._time turns out to be just `seconds`.
```

awesome, that's what I needed 👍

ryan953 (approved):

bruno-garcia (approved):
I got FOMO so I'll approve this too

static/app/utils/replays/timer.tsx
```diff
@@ -33,7 +34,10 @@ export class Timer extends EventTarget {
    */
   start(seconds?: number) {
     this._pausedAt = 0;
-    this._start = window.performance.now() - (seconds ?? 0);
+    // we're dividing by the speed here because we mutiply
```
bruno-garcia, line 37:
```suggestion
    // we're dividing by the speed here because we multiply
```

michellewzhang: good catch😵

PR #73345, opened 2024-06-25
perf(similarity): in backfill lower time bounds

JoshFerge:
lower time bounds to 1 minute, as we'll scan less data, and i think this should be plenty of range for it to find the last event sent.


roggenkemper (approved):

iadityagupta19:
lower time bounds to 1 minute, as we'll scan less data, and i think this should be plenty of range for it to find the last event sent.

PR #73348, opened 2024-06-26
feat(uptime): Implement uptime subscription system

wedamija:
This implements a system for creating subscriptions to our url checker. At the moment this is heavily copy/pasted from the snuba subscriptions system, and we plan to abstract the logic between these systems better in the future.

This handles creating and deleting subscriptions. We send subscriptions via kafka, and use tasks to make sure that the subscriptions are successfully created.

Note that we explicitly don't allow `UptimeSubscription` rows to be updated. Since these are shared, we can't modify them without risking affecting other users.

### DIFF
--- a/src/sentry/uptime/subscriptions/tasks.py
+++ b/src/sentry/uptime/subscriptions/tasks.py
@@ -1,0 +1,166 @@
+from __future__ import annotations
+
+import logging
+from datetime import timedelta
+from uuid import uuid4
+
+from arroyo import Topic as ArroyoTopic
+from arroyo.backends.kafka.configuration import build_kafka_configuration
+from arroyo.backends.kafka.consumer import KafkaPayload, KafkaProducer
+from sentry_kafka_schemas.codecs import Codec
+from sentry_kafka_schemas.schema_types.uptime_configs_v1 import CheckConfig
+
+from sentry.conf.types.kafka_definition import Topic, get_topic_codec
+from sentry.snuba.models import QuerySubscription
+from sentry.tasks.base import instrumented_task
+from sentry.uptime.models import UptimeSubscription
+from sentry.utils import metrics
+from sentry.utils.hashlib import md5_text
+from sentry.utils.kafka_config import get_kafka_producer_cluster_options, get_topic_definition
+
+logger = logging.getLogger(__name__)
+
+
+SUBSCRIPTION_STATUS_MAX_AGE = timedelta(minutes=10)
+uptime_config_producer: KafkaProducer | None = None
+uptime_config_codec: Codec | None = None
+
+
+def _get_subscription_producer() -> KafkaProducer:
+    global uptime_config_producer
+    if uptime_config_producer is None:
+        cluster_name = get_topic_definition(Topic.UPTIME_CONFIG)["cluster"]
+        producer_config = get_kafka_producer_cluster_options(cluster_name)
+        producer_config.pop("compression.type", None)
+        producer_config.pop("message.max.bytes", None)
+        uptime_config_producer = KafkaProducer(
+            build_kafka_configuration(default_config=producer_config)
+        )
+
+    return uptime_config_producer
+
+
+def _get_config_codec() -> Codec:
+    global uptime_config_codec
+    if uptime_config_codec is None:
+        uptime_config_codec = get_topic_codec(Topic.UPTIME_CONFIG)
+    return uptime_config_codec
+
+
+def get_uptime_config_topic() -> ArroyoTopic:
+    return ArroyoTopic(get_topic_definition(Topic.UPTIME_CONFIG)["real_topic_name"])
+
+
+@instrumented_task(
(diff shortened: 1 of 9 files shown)


src/sentry/uptime/subscriptions/tasks.py
```diff
@@ -0,0 +1,137 @@
...
+    return {
+        "subscription_id": subscription_id,
+        "url": subscription.url,
+        "interval_seconds": subscription.interval_seconds,  # type: ignore[typeddict-item]
+        "timeout_ms": subscription.timeout_ms,
+    }
+
+
+def send_uptime_config_deletion(subscription_id: str) -> None:
+    send_uptime_config_message(subscription_id, b"")
```
evanpurkhiser, line 128: It's actually not empty bytes, it's `None`. The types in arroyo are wrong though

src/sentry/uptime/subscriptions/tasks.py
```diff
@@ -0,0 +1,137 @@
...
+
+def _get_config_codec() -> Codec:
+    global uptime_config_codec
+    if uptime_config_codec is None:
+        uptime_config_codec = get_topic_codec(Topic.UPTIME_CONFIG)
+    return uptime_config_codec
+
+
+def get_uptime_config_topic() -> ArroyoTopic:
+    return ArroyoTopic(get_topic_definition(Topic.UPTIME_CONFIG)["real_topic_name"])
```
evanpurkhiser, line 51: Can all be replaced with the stuff in `uptime.config_producer`

src/sentry/uptime/subscriptions/tasks.py
```diff
@@ -0,0 +1,137 @@
...
+    # Whenever we create/update a config we always want to generate a new subscription id. This allows us to validate
+    # that the config took effect
+    subscription_id = uuid4().hex
+    check_config = uptime_subscription_to_check_config(subscription, subscription_id)
+    send_uptime_config_message(subscription_id, _get_config_codec().encode(check_config))
+    return subscription_id
+
+
+def uptime_subscription_to_check_config(
+    subscription: UptimeSubscription, subscription_id: str
```
evanpurkhiser, line 117: Why do you have to pass `subscription_id` isn't it in the subscription already?

src/sentry/uptime/subscriptions/tasks.py
```diff
@@ -0,0 +1,137 @@
...
+) -> CheckConfig:
+    return {
+        "subscription_id": subscription_id,
+        "url": subscription.url,
+        "interval_seconds": subscription.interval_seconds,  # type: ignore[typeddict-item]
```
evanpurkhiser, line 122: Can cast as well if we want to be more explcit.

src/sentry/uptime/subscriptions/tasks.py
evanpurkhiser: Can be replaced with the stuff in `uptime.config_producer` anyway

src/sentry/uptime/subscriptions/tasks.py
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+
+def send_uptime_config_deletion(subscription_id: str) -> None:
+    send_uptime_config_message(subscription_id, b"")
+
+
+def send_uptime_config_message(subscription_id: str, value: bytes) -> None:
+    producer = _get_subscription_producer()
+    message_key = md5_text(subscription_id).hexdigest().encode()
```
evanpurkhiser, line 133: In what I wrote the key is just the bytes of the subscription_id. But I think we were talking about making it a composite key of something else?

Right now the uptime checker config consumer expects it to be the subscription key and will fail to consume if it's not https://github.com/getsentry/uptime-checker/blob/eee8326ce849eb11b277a561964e0a2d18835c61/src/config_consumer.rs#L50-L53

evanpurkhiser (approved):
Looking good.

IMO we should update the SingletonProducer to return the future so we can wait on it, and then we can use that and not have to instantiate the producer ourselves.

There's some differences in the producer you wrote and the one I wrote, so we'll want to reconsicle that.

FYI my producer does get called in a getsentry script, so if we move stuff around we'll want to update that too (though I do think it makes sense to have the producer live in it's own module)

src/sentry/uptime/subscriptions/tasks.py
wedamija: I don't store it in the subscription until I know I've successfully sent the update to Kafka

src/sentry/uptime/subscriptions/tasks.py
wedamija: We just need a composite key for the results, not the config

src/sentry/uptime/subscriptions/tasks.py
wedamija: I remember this being a pain and wasn't sure how to cast to the literal

PR #73353, opened 2024-06-26
fix(metrics) Prevent negative crash_free_rate value to be returned to the client

### LINKED ISSUE #73172: Negative session.crash_free_rate
When there are more crashes than sessions, the crash_free_rate is negative.

vgrozdanic:
Sometimes due to the problems in other parts of the system, data gets corrupted and `crash_free_rate` is returned as negative value (which should never happen)

This solution guards against corrupted data by returning 0 if the calculated `crash_free_rate` is less than 0

Fixes: https://github.com/getsentry/sentry/issues/73172

### DIFF
--- a/src/sentry/snuba/metrics/fields/snql.py
+++ b/src/sentry/snuba/metrics/fields/snql.py
@@ -655,18 +655,12 @@ def rate_snql_factory(
     alias: str | None = None,
 ) -> Function:
     return Function(
-        "max",
+        "divide",
         [
-            0.0,
-            Function(
-                "divide",
-                [
-                    Function("countIf", [Column("value"), aggregate_filter]),
-                    Function("divide", [numerator, denominator]),
-                ],
-                alias=alias,
-            ),
+            Function("countIf", [Column("value"), aggregate_filter]),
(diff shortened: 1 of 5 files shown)


src/sentry/snuba/metrics/fields/snql.py
```diff
@@ -589,6 +589,13 @@ def miserable_users(org_id: int, metric_ids: Sequence[int], alias: str | None =
     )
 
 
+def non_negative_crash_free_rate(crash_rate, alias: str | None = None) -> Function:
+    # prevents crash_free_rate value to be negative when crash_rate > 1
+    # (this can happen due to a bug in other parts of the system)
+    # for more info see: https://github.com/getsentry/sentry/issues/73172
+    return Function("max2", [0, Function("minus", [1, crash_rate])], alias)
```
vgrozdanic, line 596: `max2` is a Clickhouse function to return maximum of 2 elements: https://clickhouse.com/docs/en/sql-reference/functions/arithmetic-functions#max2

priscilawebdev (approved):
🚀

obostjancic (approved):

PR #73355, opened 2024-06-26
feat(uptime): Hook up subscription creation to the detector

wedamija:
This allows the detector to start creating subscriptions for valid urls. It also adds a mode column to the `ProjectUptimeSubscription` table. We put it on this table and not the `UptimeSubscription` table since we want each project to perform validation itself, and also because we don't necessarily know that the subscription has been validated before - possibly it's a manual subscription that a user set up.

This isn't fully complete but is directional. Should mostly need tests to validate behaviour

### DIFF
--- a/src/sentry/uptime/detectors/tasks.py
+++ b/src/sentry/uptime/detectors/tasks.py
@@ -14,7 +14,14 @@ from sentry.uptime.detectors.ranking import (
     get_candidate_urls_for_project,
     get_project_bucket,
 )
-from sentry.uptime.models import ProjectUptimeSubscription, UptimeSubscription
+from sentry.uptime.models import ProjectUptimeSubscriptionMode
+from sentry.uptime.subscriptions.subscriptions import (
+    create_project_uptime_subscription,
+    create_uptime_subscription,
+    delete_project_uptime_subscription,
+    get_auto_monitored_subscriptions_for_project,
+    is_url_auto_monitored_for_project,
+)
 from sentry.utils import metrics
 from sentry.utils.hashlib import md5_text
 from sentry.utils.locking import UnableToAcquireLock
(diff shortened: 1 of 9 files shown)


src/sentry/uptime/detectors/tasks.py
```diff
@@ -24,6 +31,8 @@
 FAILED_URL_RETRY_FREQ = timedelta(days=7)
 URL_MIN_TIMES_SEEN = 5
 URL_MIN_PERCENT = 0.05
+ONBOARDING_SUBSCRIPTION_INTERVAL_SECONDS = int(timedelta(minutes=60).total_seconds())
+ONBOARDING_SUBSCRIPTION_TIMEOUT_MS = 1000
```
evanpurkhiser, line 35: Feels like some of these could maybe use comments

src/sentry/uptime/detectors/tasks.py
```diff
@@ -141,46 +150,50 @@ def process_candidate_url(
...
     logger.info(
         "uptime.url_autodetected",
         extra={
             "url": url,
             "project": project.id,
         },
     )
+    # If we hit this point, then the url looks worth monitoring. Create an uptime subscription in monitor mode.
+    # Also check if there's already an existing auto-detected monitor for this project. If so, delete it.
+    monitor_url_for_project(project, url)
```
evanpurkhiser, line 177: Should this be behnd a flag for now?

src/sentry/uptime/detectors/tasks.py
```diff
@@ -141,46 +150,50 @@ def process_candidate_url(
...
-    # TODO: Implement subscriptions
     logger.info(
         "uptime.url_autodetected",
         extra={
             "url": url,
             "project": project.id,
         },
     )
+    # If we hit this point, then the url looks worth monitoring. Create an uptime subscription in monitor mode.
+    # Also check if there's already an existing auto-detected monitor for this project. If so, delete it.
```
evanpurkhiser, line 176: This comment feels like it should be in the `monitor_url_for_project` function

evanpurkhiser (approved):

evanpurkhiser (approved):

PR #73367, opened 2024-06-26
ref(metrics) fix incorrect exception being raised

### LINKED ISSUE #71561: Incorrect Exception raised
As the `get_blended_sample_rate` can return either `float | None` raising an Exception on None is misleading and creates confusion. The None case of `get_blended_sample_rate` could be handled using a default value.

https://github.com/getsentry/sentry/blob/41a2e0a3dd8048c41d551d7301709ea0600d1c59/src/sentry/dynamic_sampling/rules/base.py#L51

vgrozdanic:
Closes https://github.com/getsentry/sentry/issues/71561

### DIFF
--- a/src/sentry/dynamic_sampling/rules/base.py
+++ b/src/sentry/dynamic_sampling/rules/base.py
@@ -49,9 +49,8 @@ def get_guarded_blended_sample_rate(organization: Organization, project: Project
     )
 
     # get_blended_sample_rate returns None if the organization doesn't have dynamic sampling
-    # see: https://github.com/getsentry/sentry/blob/5499277c21f2dc997da79343087565b6676a6f50/src/sentry/quotas/base.py#L585
     if sample_rate is None:
-        sample_rate = 1.0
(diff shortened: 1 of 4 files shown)


src/sentry/dynamic_sampling/rules/base.py
```diff
@@ -48,9 +48,10 @@ def get_guarded_blended_sample_rate(organization: Organization, project: Project
         organization_id=organization.id, project=project
     )
 
-    # If the sample rate is None, it means that dynamic sampling rules shouldn't be generated.
+    # get_blended_sample_rate returns None if the organization doesn't have dynamic sampling
+    # see: https://github.com/getsentry/sentry/blob/5499277c21f2dc997da79343087565b6676a6f50/src/sentry/quotas/base.py#L585
```
iambriccardo, line 52: nit: I would say there is no need to link to the function definition itself.

iambriccardo (approved):
LGTM

PR #73375, opened 2024-06-26
fix(insights): removes unneeded empty state logic from Web Vitals module

edwardgou-sentry:
Deletes a redundant empty state from the Web Vitals module when no transactions are found. This is redundant and unneeded because we already have a different and better empty state when no score metrics are found

image


KevinL10 (approved):
Looks good. Do we plan on using `<ModulesOnboarding />` for the empty state in the future?

edwardgou-sentry:
> Looks good. Do we plan on using `<ModulesOnboarding />` for the empty state in the future?

Possibly, there is already an onboarding state for the webvitals (same as the performance page) that is different from `<ModulesOnboarding />`. I think we've only started using `<ModulesOnboarding />` for recent modules, we might need to iterate on the content a little bit to work with webvitals.

PR #73377, opened 2024-06-26
chore(hybridcloud) Add sentry.hybridcloud.* to stronglist

markstory:
Fix the existing issues, as I continue to move modules I'll address typing problems so that we can keep the new home in the stronglist.


GabeVillalobos (approved):

src/sentry/hybridcloud/rpc/filter_query.py
```diff
@@ -132,4 +132,5 @@ def get_many(self, filter: FILTER_ARGS) -> list[RPC_RESPONSE]:
         return [self.serialize_rpc(o) for o in self.query_many(filter=filter)]
 
     def get_many_ids(self, filter: FILTER_ARGS) -> list[int]:
-        return [o.id for o in self.query_many(filter=filter, select_related=False)]
+        # Using getattr to avoid mypy errors that Model does not have attr id
+        return [getattr(o, "id") for o in self.query_many(filter=filter, select_related=False)]
```
asottile-sentry, line 136: this should be `bound=Model` but our base `Model` in `sentry.db.models.base` (it has `id: Field[int, int]`)

src/sentry/hybridcloud/tasks/deliver_webhooks.py
```diff
@@ -71,7 +72,7 @@ class DeliveryFailed(Exception):
     queue="webhook.control",
     silo_mode=SiloMode.CONTROL,
 )
-def schedule_webhook_delivery(**kwargs) -> None:
```
asottile-sentry, line 74: I think `**kwargs: Never` would make the type checker help more here (since we don't expect this to take args)

src/sentry/hybridcloud/rpc/filter_query.py
markstory: I tried that. However, the models in `socialauth` don't inherit from `BaseModel` so type errors show up in different parts of the codebase.

asottile-sentry: I've been fixing those when running into them -- shouldn't be too bad to move them over (yeah I know they are vendored modules, but we've already modified them so much that we're never going to be able to cleanly pull or unfork)

PR #73380, opened 2024-06-26
chore(insights): deletes unused hooks and wrappers from the web vitals module

edwardgou-sentry:
Deletes some unused query hooks and wrappers from the web vitals module.


KevinL10 (approved):
👍

PR #73382, opened 2024-06-26
chore(slack): completely remove block kit flag

cathteng:
What it says ❗

### DIFF
--- a/src/sentry/integrations/slack/webhooks/options_load.py
+++ b/src/sentry/integrations/slack/webhooks/options_load.py
@@ -94,7 +94,7 @@ class SlackOptionsLoadEndpoint(Endpoint):
             logger.exception(
                 "slack.options_load.request-error",
                 extra={
-                    "group_id": group.id if group else None,
+                    "group_id": slack_request.group_id,
                     "organization_id": group.project.organization.id if group else None,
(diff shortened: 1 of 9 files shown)


cathteng:
typing failures is unrelated, will probably wait for a fix

cathteng:
typing failure is unrelated, will probably wait for a fix

src/sentry/integrations/slack/webhooks/options_load.py
```diff
@@ -91,9 +90,7 @@ def post(self, request: Request) -> Response:
             .first()
         )
 
-        if not group or not features.has(
-            "organizations:slack-block-kit", group.project.organization
-        ):
+        if not group:
```
isabellaenriquez, line 93: sorry to intrude 😬 but with this removal, the `group_id` in the extra below can be replaced with `slack_request.group_id` since `group` will always be None when the logger is reached now

(i did this in #71105 and removed the flags there too but i'll just wait til this PR is merged and update my branch after so all the remaining flags can be removed at the same time 🫡 )

cathteng: thanks!

iamrajjoshi (approved):
🚀

PR #73385, opened 2024-06-26
fix(performance): fix issue with mobile widget breaking due to non unique series names

edwardgou-sentry:
We use `formatVersion` to produce series name for the mobile performance widget, but `formatVersion` can produce the same string for different releases if they have the same semver. Updates to use `formatVersion` with the `withPackage` set to `true` to prevent this clashing.


Zylphrex (approved):
Perhaps consider always showing the package name, not just when they overlap

edwardgou-sentry:
> Perhaps consider always showing the package name, not just when they overlap

👍 updated to just always show package name to reduce complexity. I dont see why not 😄

PR #73199, opened 2024-06-24
feat(insights): add project has_insight_xx flags to frontend

KevinL10:
Continuation of https://github.com/getsentry/sentry/pull/73197. Adds has_insight_xx flags to the `Project` type and `useHasFirstSpan` hook to eventually replace `useHasData`.

PR #73386, opened 2024-06-26
feat(insights): replace useHasDataTrackAnalytics with new insight spans 

KevinL10:
Replaces useHasData analytics with the new project-based flag.

Requires https://github.com/getsentry/sentry/pull/73199


gggritso (approved):
This makes a lot of sense, but I think the `has_data` tag might need a new name 🤔 before this change, the hook actually checked whether the user has any recent data, right? _Now_, `has_data` is actually a check for whether the user has _ever_ sent data! So, someone might come back to Sentry after a few months off, and not see any data, but still have `has_data: true` which would be confusing when looking at analytics

Not saying this behaviour is _wrong_, just that it would be a lot clearer if the tag was something like `has_ever_sent_data` or whatever, rather than `has_data`, which implies that they're looking at an empty screen

Speaking of which, I assume Ben knows that this change in meaning is coming, right? I leave it to you two to decide the correct behaviour and interpretation!

PR #73387, opened 2024-06-26
fix(slack): Update How we Iterate to Slack Response in `users.list`

iamrajjoshi:
Addresses https://sentry.sentry.io/issues/5539974714/events/2af1ab068807409ab8e23ffba33b260c/?project=1

Basically the generator itself returns a list of users, we need to iterate through them further to get an individual user. I tested this with my local Slack setup.

I wish we had smoke tests, would have caught this.

### DIFF
--- a/src/sentry/integrations/slack/utils/channel.py
+++ b/src/sentry/integrations/slack/utils/channel.py
@@ -174,9 +174,14 @@ def check_user_with_timeout(
         "exclude_members": True,
         "types": "public_channel,private_channel",
     }
-    users = get_slack_user_data(integration, organization=None, kwargs=payload)
 
-    for user in users:
+    # Get each user from a page from the Slack API
+    page_generator = (
+        user
+        for page in get_slack_user_data(integration, organization=None, kwargs=payload)
+        for user in page
+    )
+    for user in page_generator:
         # The "name" field is unique (this is the username for users)
         # so we return immediately if we find a match.
         # convert to lower case since all names in Slack are lowercase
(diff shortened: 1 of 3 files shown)


cathteng (dismissed):
since this was confusing, it might be helpful to rename to function to something that denotes it returns a list of users?

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -174,9 +174,14 @@ def check_user_with_timeout(
         "exclude_members": True,
         "types": "public_channel,private_channel",
     }
-    users = get_slack_user_data(integration, organization=None, kwargs=payload)
 
-    for user in users:
+    # Get each user from a page from the Slack API
+    page_generator = (
```
cathteng, line 179: nit: should this be `user_generator` since it yields the user?

tests/sentry/integrations/slack/test_utils.py
```diff
@@ -198,3 +225,36 @@ def test_rate_limiting(self):
...
+
+    @patch("slack_sdk.web.client.WebClient._perform_urllib_http_request")
+    @with_feature("organizations:slack-sdk-get-channel-id")
+    def test_user_list_pagination_sdk_client(self, mock_api_call):
+        self.response_json["response_metadata"] = {"next_cursor": "dXNlcjpVMEc5V0ZYTlo"}
+        mock_api_call.return_value = {
+            "body": orjson.dumps(self.response_json).decode(),
+            "headers": {},
+            "status": 200,
+        }
```
cathteng, line 237: can we add a page to the successful test?

cathteng: like test that a user on the second page is parsed correctly too

cathteng (approved):
🚢

PR #73388, opened 2024-06-26
fix(typing) Fix typing errors in master

markstory:
These didn't show up on my pull request, but are causing master to fail.


roggenkemper (approved):

corps (approved):

GabeVillalobos (approved):

src/sentry/hybridcloud/rpc/__init__.py
```diff
@@ -30,13 +30,13 @@
 
 
 class ValueEqualityEnum(Enum):
-    def __eq__(self, other):
+    def __eq__(self, other: Any) -> bool:
```
asottile-sentry, line 33: the signature for `__eq__` should always be `(object) -> bool`

PR #73389, opened 2024-06-26
feat(insights): replace useHasData and ModuleOnboarding with project hasInsightSpan

KevinL10:
Replaces `useHasData` with the project-based flag for each insight module. This avoids the need to look back 90-days to determine whether a module contains data.

Requires https://github.com/getsentry/sentry/pull/73199 and https://github.com/getsentry/sentry/pull/73386


gggritso (approved):
TS error aside, makes sense!

edwardgou-sentry:
lgtm but I think theres one last spot where `useHasData` is still being used in `useHasDataTrackAnalytics.tsx`

PR #73392, opened 2024-06-26
Project Setting Template Options Manager

saponifi3d:
## Description
[Tech Spec](https://www.notion.so/sentry/Service-Settings-Templates-ca6fdb38b15f4a6db61d319bc342b86c?pvs=4#8cddeba5606d43f49c0d80a71bcaaedb)
This is the implementation of a caching layer for ProjectTemplateOptions. This layer will then be used will be used to create the layering effect (described in tech spec) for the `Project` model.

This code is meant to replicate the [ProjectOptionsManager](https://github.com/getsentry/sentry/blob/a2c96d4822961b76b71f40e9cb82f14277104a8f/src/sentry/models/options/project_option.py#L79), but there are a few minor differences. These differences are why I ended up replicating the system rather than trying to create a single manager.

Currently this code is not executable in production.

### DIFF
--- a/tests/sentry/models/test_projecttemplateoption.py
+++ b/tests/sentry/models/test_projecttemplateoption.py
@@ -11,12 +11,40 @@ class ProjectTemplateOptionManagerTest(TestCase):
             name="test_project_template", organization=self.org
         )
 
-    def create_test_option(self, key="key", value="value"):
-        return ProjectTemplateOption.objects.create(
-            project_template=self.project_template,
-            key=key,
-            value=value,
+    def test_set_value(self):
+        ProjectTemplateOption.objects.set_value(self.project_template, "foo", "bar")
+        assert (
+            ProjectTemplateOption.objects.get(
+                project_template=self.project_template, key="foo"
+            ).value
+            == "bar"
         )
+
+    def test_get_value(self):
+        result = ProjectTemplateOption.objects.get_value(self.project_template, "foo")
+        assert result is None
+
+        ProjectTemplateOption.objects.create(
+            project_template=self.project_template, key="foo", value="bar"
+        )
+        result = ProjectTemplateOption.objects.get_value(self.project_template, "foo")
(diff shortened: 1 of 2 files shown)


tests/sentry/models/test_projecttemplateoption.py
```diff
@@ -3,6 +3,113 @@
...
+        assert result == "bar"
+
+        # Test validator false
+        result = ProjectTemplateOption.objects.get_value(
+            self.project_template, "foo", None, lambda x: x != "bar"
+        )
+        assert result is None
+
+    def test_unset_value(self):
+        ProjectTemplateOption.objects.unset_value(self.project_template, "foo")
```
ceorourke, line 55: why do you unset it here?

tests/sentry/models/test_projecttemplateoption.py
```diff
@@ -3,6 +3,113 @@
...
+        ).exists()
+
+    def test_get_value_bulk(self):
+        result = ProjectTemplateOption.objects.get_value_bulk([self.project_template], "foo")
+        assert result == {}
+
+        ProjectTemplateOption.objects.create(
+            project_template=self.project_template, key="foo", value="bar"
+        )
+        result = ProjectTemplateOption.objects.get_value_bulk([self.project_template], "foo")
```
ceorourke, line 71: might be nice to do one w/ multiple

tests/sentry/models/test_projecttemplateoption.py
saponifi3d: ah, whoops! this exposed an issue and i forgot to clean it up, will remove :)

ceorourke (approved):
overall lgtm!

tests/sentry/models/test_projecttemplateoption.py
saponifi3d: 👍 will add a test for that.

PR #73394, opened 2024-06-26
perf(similarity): use mulithreadpool to do nodestore calls concurrently

JoshFerge:
use a threadpool and chunk into batches of 5 for bigtable reads, with up to 5 threads


src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -377,33 +378,19 @@ def update_groups(project, seer_response, group_id_batch_filtered, group_hashes_
...
+            node_keys,
             retries=3,
             delay=2,
         )
     except (ServiceUnavailable, DeadlineExceeded) as e:
         extra = {
             "organization_id": project.organization.id,
             "project_id": project.id,
-            "group_data": json.dumps(rows),
+            "node_keys": json.dumps(node_keys),
```
JoshFerge, line 393: not logging rows here because i don't want to pass them in, node keys should be enough for us to debug

JoshFerge: (i don't believe this error has happened, either

vartec:
is `nodestore.backend` known to be thread-safe?

vartec (approved):

PR #73396, opened 2024-06-26
chore(hybridcloud) Move apigateway middleware

markstory:
Move the middleware class to be with the rest of the apigateway logic.

Refs HC-982


corps (approved):
Curious to see if our new grouping logic correctly groups api gateway errors after the package rename! @JoshFerge

JoshFerge:
> Curious to see if our new grouping logic correctly groups api gateway errors after the package rename! @JoshFerge

bring it on!!

JoshFerge:
also: have we considered marking some of the hybrid cloud files as non-in-app in our stacktrace rules?

markstory:
> also: have we considered marking some of the hybrid cloud files as non-in-app in our stacktrace rules?

They are in-app to *me* though 😄

PR #73397, opened 2024-06-26
feat: Activated Alert UI/UX updates

nhsiehgit:
updates:

Filter table on the alert details page when an alert is selected ✅
Button / link to remove selected Alert query param ✅
Remove start/finish monitoring row from the alert details page✅
Add ‘per new deploy/release’ into the alert details side bar ✅

Jun-26-2024 16-22-21


src/sentry/incidents/endpoints/serializers/alert_rule.py
```diff
@@ -272,6 +273,10 @@ def serialize(
         aggregate = translate_aggregate_field(
             obj.snuba_query.aggregate, reverse=True, allow_mri=allow_mri
         )
+        condition_type = None
+        if obj.activation_condition.exists():
+            activation_condition = obj.activation_condition.first()
+            condition_type = activation_condition.condition_type if activation_condition else None
```
saponifi3d, line 279: might be a little cleaner to use `getattr` here:
```suggestion
            condition_type = getattr(activation_condition, 'condition_type', None)
```

If you wanted to get a bit fancier you can merge lines 278 and 279 like this:
```py
condition_type = obj.activation_condition.values_list('condition_type', flat=True).first()
```

This works by using [`values_list`](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet.values_list) to limit the properties to `condition_type` and return the first one, by default then it'd return `None` from that.

static/app/views/alerts/rules/metric/details/metricChart.tsx
```diff
@@ -536,6 +538,27 @@ class MetricChart extends PureComponent<Props, State> {
...
 
+    let activationFilter = '';
+    if (
+      rule.monitorType === MonitorType.ACTIVATED &&
+      selectedIncident &&
+      selectedIncident.activation
+    ) {
+      const activation = selectedIncident.activation;
+      const activator = activation.activator;
+      const conditionType = activation.conditionType;
```
saponifi3d, line 549: Nit: Recommend using destruct for this, it's a bit simpler as you can DRY the variable names, and then remove the 2nd line creating conditionType (and then easier in the future if there's more to destruct)
```suggestion
      const { activation } = selectedIncident;
      const { activator, conditionType } = activation;
```

static/app/views/alerts/rules/metric/details/metricHistory.tsx
```diff
@@ -20,34 +20,35 @@ type Props = {
...
-      if (activation.isComplete) {
-        activationTriggers.push({
-          type: ActivationTrigger.FINISHED,
-          activator: activation.activator,
-          conditionType: activation.conditionType,
-          dateCreated: activation.finishedAt,
-        });
-      }
-    });
+    // NOTE: disabling start/finish trigger rows for now
```
saponifi3d, line 30: any idea how long this will be disabled for? (just thinking we could pull it out of git history if it's going to be a while, and clean up this file a little more for now)

static/app/views/alerts/rules/metric/details/sidebar.tsx
```diff
@@ -140,6 +141,21 @@ export function MetricDetailsSidebar({
 
   const ownerId = rule.owner?.split(':')[1];
   const teamActor = ownerId && {type: 'team' as Actor['type'], id: ownerId, name: ''};
+  let conditionType;
+  switch (
+    rule.monitorType === MonitorType.ACTIVATED &&
+    typeof rule.activationCondition !== 'undefined' &&
+    rule.activationCondition
```
saponifi3d, line 148: thoughts on pulling this out as a variable? It might read a little more cleanly if you have something like:
```ts
const activationConditionType = rule.monitorType === MonitorType.ACTIVATED &&
    typeof rule.activationCondition !== 'undefined' &&
    rule.activationCondition;
    
switch(activationConditionType) {
  ...
}
```
(or maybe it could just be an if statement? 🤔)

static/app/views/alerts/rules/metric/details/sidebar.tsx
```diff
@@ -205,6 +221,13 @@ export function MetricDetailsSidebar({
             keyName={t('Environment')}
             value={<OverflowTableValue>{rule.environment ?? '-'}</OverflowTableValue>}
           />
+          {rule.monitorType === MonitorType.ACTIVATED &&
+            typeof rule.activationCondition !== 'undefined' && (
```
saponifi3d, line 225: nit: i believe you can just check to see if it's undefined (i'm guessing we can't just do !rule.activationCondition cause there's a `0` case?)
```suggestion
            rule?.activationCondition !== undefined && (
```

saponifi3d (approved):
Just nits / clean-up suggestions from me 🎉 - it might be nice to also add some FE tests for the new component logic (hoping it's not tooooo crazy to add those)

src/sentry/incidents/endpoints/serializers/alert_rule.py
```diff
@@ -89,6 +89,7 @@ class AlertRuleSerializerResponse(AlertRuleSerializerResponseOptional):
     createdBy: dict
     monitorType: int
     activations: list[dict]
+    activationCondition: int | None
```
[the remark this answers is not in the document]
nhsiehgit: Not in the serializer response.
This is typing out what we're returning to the FE

src/sentry/incidents/endpoints/serializers/alert_rule.py
nhsiehgit: Oh nice i like that 🙌

static/app/views/alerts/rules/metric/details/metricChart.tsx
nhsiehgit: ohya duh 👍

static/app/views/alerts/rules/metric/details/metricHistory.tsx
nhsiehgit: I'll leave a note.

Just until we determine whether it's useful or not during EA

static/app/views/alerts/rules/metric/details/sidebar.tsx
nhsiehgit: 🤷 
I'll keep a switch statement since this could potentially become more than just these 2
fine wiht breaking out into a const var

static/app/views/alerts/rules/metric/details/sidebar.tsx
nhsiehgit: yea i think either works

PR #73400, opened 2024-06-26
feat(u2f): Register fallback promises

scttcper:
in https://github.com/getsentry/sentry/blob/master/static/app/actionCreators/organization.tsx#L82-L92 we use `getPreloadedDataPromise` to make a followup request if the first one fails. Register the fallbacks to the __sentry_preload object so we can wait for the fallback promises to finish.

when testing locally they seem to be registered in time 

follow up to #73315

### DIFF
--- a/static/app/utils/getPreloadedData.ts
+++ b/static/app/utils/getPreloadedData.ts
@@ -12,7 +12,9 @@ export async function getPreloadedDataPromise(
   const wrappedFallback = () => {
     const fallbackAttribute = `${name}_fallback`;
     const promise = fallback();
-    data[fallbackAttribute] = promise;
+    if (data) {
+      data[fallbackAttribute] = promise;
+    }
     return promise;
   };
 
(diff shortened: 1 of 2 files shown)


static/app/utils/getPreloadedData.ts
```diff
@@ -14,15 +25,15 @@ export async function getPreloadedDataPromise(
       !data[name] ||
       !data[name].then
     ) {
-      return await fallback();
+      return await wrappedFallback();
     }
-    const result = await data[name].catch(fallback);
+    const result = await data[name].catch(() => null);
```
scttcper, line 30: in this case we were making 3 requests, switching it to just 2

saponifi3d (approved):
lgtm!

static/app/utils/getPreloadedData.ts
```diff
@@ -4,8 +4,21 @@ export async function getPreloadedDataPromise(
   fallback: () => Promise<any>,
   usePreload?: boolean
 ) {
+  const data = (window as any).__sentry_preload;
+  /**
+   * Save the fallback promise to `__sentry_preload` to allow the sudo modal to wait
+   * for the promise to resolve
+   */
+  const wrappedFallback = () => {
```
saponifi3d, line 12: any chance that has file is tested and we could add a unit test to make sure the wrappedFallback is invoked as expected?

PR #73413, opened 2024-06-27
Setup increase the number of issue alerts by org slug

saponifi3d:
## Description
This PR creates a helper function, `get_max_alerts`. This enables us to change the max limit of issue alerts for an organization to 1000.

[Slack discussion for additional context](https://sentry.slack.com/archives/C039ZGT5K/p1719433765759889)


src/sentry/features/temporary.py
```diff
@@ -176,6 +176,7 @@ def register_temporary_features(manager: FeatureManager):
     # Adds the ttid & ttfd vitals to the frontend
     manager.add("organizations:mobile-vitals", OrganizationFeature, FeatureHandlerStrategy.REMOTE)
     # Enables higher limit for alert rules
+    manager.add("organizations:more-fast-alerts", OrganizationFeature, FeatureHandlerStrategy.INTERNAL)
```
saponifi3d, line 179: TODO - investigate how to add this feature flag / target only a single org (kinda like the `more-slow` handler) or if i need to handle it in options automator.

src/sentry/conf/server.py
```diff
@@ -3196,6 +3196,7 @@ def custom_parameter_sort(parameter: dict) -> tuple[str, int]:
 MAX_MORE_SLOW_CONDITION_ISSUE_ALERTS = 300
 MAX_FAST_CONDITION_ISSUE_ALERTS = 500
 MAX_QUERY_SUBSCRIPTIONS_PER_ORG = 1000
+MAX_MORE_FAST_CONDITION_ISSUE_ALERTS = 1000
```
nhsiehgit, line 3199: 🤨 hmmmm dang. wish we had chosen verbage better than 'max more fast' and 'max more slow' 😭

nhsiehgit (approved):
👍

src/sentry/conf/server.py
saponifi3d: hahah it's true.. if you look at the commit history on the pr i thought about changing it... 🤣

PR #73419, opened 2024-06-27
feat(metrics-extraction): Change extraction condition to object

ArthurKnaus:
Change the extraction conditions from strings to objects.

### DIFF
--- a/static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
+++ b/static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
@@ -247,7 +247,7 @@ export function MetricsExtractionRuleForm({isEdit, project, onSubmit, ...props}:
                 <Fragment>
                   <ConditionsWrapper hasDelete={value.length > 1}>
                     {conditions.map((condition, index) => (
-                      <Fragment key={index}>
+                      <Fragment key={condition.id}>
                         <SearchWrapper hasPrefix={index !== 0}>
                           {index !== 0 && <ConditionLetter>{t('or')}</ConditionLetter>}
                           <SearchBar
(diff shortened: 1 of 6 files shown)


static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
```diff
@@ -92,6 +93,16 @@ export function aggregatesToGroups(aggregates: MetricsAggregate[]): AggregateGro
   }
   return groups;
 }
+
+export function createCondition(): MetricsExtractionCondition {
+  return {
+    query: '',
+    // id and mris will be set by the backend after creation
+    id: uniqueId(),
```
ArthurKnaus, line 101: @shellmayr I assign a uniqueId to each condition when it gets created (to use as key in the FE - like we do in dashboards). Can this cause a problem, when creating / updating a extraction rule?
the id is in uuid v4 format

obostjancic (approved):
lgtm

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
shellmayr: When creating, it just gets ignored, a new integer ID is created in the database, and the database ID is returned in the response. The format of the ID created on the frontend is therefore irrelevant. 
When updating, you'll need to pass the database ID to specify which rule should change, if you specify some random uuid in the update, it will not work.

priscilawebdev (approved):
left 1 suggestion but other than that it looks good

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
```diff
@@ -218,20 +229,33 @@ export function MetricsExtractionRuleForm({isEdit, project, onSubmit, ...props}:
             flexibleControlStateSize
           >
             {({onChange, initialData, value}) => {
-              const conditions = (value || initialData) as string[];
+              const conditions = (value ||
+                initialData ||
+                []) as MetricsExtractionCondition[];
+
+              const handleChange = (queryString: string, index: number) => {
```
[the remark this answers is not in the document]
ArthurKnaus: Will not prevent re-renders unless I move the whole condition item into a separate component.
Can do this refactoring in a separate PR, but the performance gains are not noticeable in this case.

priscilawebdev (approved):

vgrozdanic (approved):

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
ArthurKnaus: k, will only send it if it exists 👍

PR #73421, opened 2024-06-27
feat(metrics): persist span attribute extraction rules in django

shellmayr:
Span attribute extraction was previously saved as a ProjectOption. This PR moves this functionality and the persistence to a Django model and updates both the CRUD endpoint and the serialization to project config for relay to use this model. 

This PR is made up of 4 major parts:
- the CRUD API endpoints
- SpanAttributeExtractionConfig is the configuration model available via CRUD to the user
- each config can have 0 or more SpanAttributeExtractionConditions that allow the user to narrow down the extraction volume
- When generating the project config for Relay, MetricsExtractionRules are generated based on the configs&conditions, which then are converted to MetricSpecs understandable by Relay


obostjancic (approved):
LGTM 🚢 🚀 🔥

src/sentry/sentry_metrics/configuration.py
```diff
@@ -172,3 +172,23 @@ def initialize_main_process_state(config: MetricsIngestConfiguration) -> None:
...
+
+
+HARD_CODED_UNITS = {"span.duration": "millisecond"}
+ALLOWED_TYPES = {"c", "d", "s", "g"}
+
+# METRICS_AGGREGATES specifies the aggregates that are available for a metric type - AGGREGATES_TO_METRICS reverses this,
+# and provides a map from the aggregate to the metric type in the form {'count': 'c', 'avg':'g', ...}. This is needed
+# when the UI lets the user select the aggregate, and the backend infers the metric_type from it. It is programmatic
+# and not hard-coded, so that in case of a change, the two mappings are aligned.
+METRICS_AGGREGATES = {
```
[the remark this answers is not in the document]
shellmayr: Good suggestion, will change

src/sentry/snuba/metrics/span_attribute_extraction.py
```diff
@@ -79,7 +79,7 @@ class SpanAttributeMetricSpec(TypedDict):
 def convert_to_metric_spec(extraction_rule: MetricsExtractionRule) -> SpanAttributeMetricSpec:
 
     field = _get_field(extraction_rule)
-    parsed_conditions = _parse_conditions(extraction_rule.conditions)
+    parsed_conditions = _parse_conditions([extraction_rule.condition])
```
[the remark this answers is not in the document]
shellmayr: Added

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -42,15 +40,18 @@ def delete(self, request: Request, project: Project) -> Response:
...
             return Response(status=204)
 
         try:
-            state_update = self._generate_deleted_rule_objects(rules_update)
-            delete_metrics_extraction_rules(project, state_update)
+            with transaction.atomic(router.db_for_write(SpanAttributeExtractionRuleConfig)):
+                for obj in config_update:
+                    SpanAttributeExtractionRuleConfig.objects.filter(
+                        project=project, span_attribute=obj["spanAttribute"]
+                    ).all().delete()
```
vgrozdanic, line 52: You do not need `all()` here, just `filter()` is enough
```suggestion
                    SpanAttributeExtractionRuleConfig.objects.filter(
                        project=project, span_attribute=obj["spanAttribute"]
                    ).delete()
```

src/sentry/sentry_metrics/models/metricsextractionrules.py
```diff
@@ -0,0 +1,108 @@
...
+        config.save()
+
+        for condition in dictionary["conditions"]:
+            condition_obj = SpanAttributeExtractionRuleCondition.objects.create(
+                created_by_id=user_id,
+                project=project,
+                value=condition["value"],
+                config=config,
+            )
+            condition_obj.save()
```
vgrozdanic, line 89: This `condition_obj.save()` is redundant, object is already created in database with previous command, this line will just add another query to the database, and it will do update of the object with the same values with which the object was created few lines above

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -60,16 +61,17 @@ def get(self, request: Request, project: Project) -> Response:
...
-            extraction_rules = get_metrics_extraction_rules(project)
+            configs = SpanAttributeExtractionRuleConfig.objects.filter(project=project)
+
         except Exception as e:
             return Response(status=500, data={"detail": str(e)})
 
         return self.paginate(
             request,
-            queryset=extraction_rules,
+            queryset=list(configs),
```
vgrozdanic, line 71: Nit for future work: We should think of using different pagination in the future, but this is currently out of the scope of this PR.

How we currently doing this, we will fetch all the configs from the database, and then slice the list of the fetched configs and return only part of that list. Instead we should do the pagination during querying the database (relay on OFFSET and LIMIT from Postgres)

src/sentry/sentry_metrics/models/metricsextractionrules.py
shellmayr: Great catch - changed!

src/sentry/api/endpoints/project_metrics_extraction_rules.py
shellmayr: Added a TODO comment for this

src/sentry/sentry_metrics/models/metricsextractionrules.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+@region_silo_model
+class SpanAttributeExtractionRuleCondition(Model):
+    __relocation_scope__ = RelocationScope.Organization
+
+    date_modified = models.DateTimeField(default=timezone.now)
+    date_added = models.DateTimeField(default=timezone.now)
+    created_by_id = HybridCloudForeignKey("sentry.User", on_delete="SET_NULL", null=True)
+    project = FlexibleForeignKey("sentry.Project")
```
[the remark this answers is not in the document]
shellmayr: Changed

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -81,62 +83,90 @@ def post(self, request: Request, project: Project) -> Response:
...
+
+                        # create new conditions
+                        else:
+                            condition_obj = SpanAttributeExtractionRuleCondition.objects.create(
+                                created_by_id=request.user.id,
+                                project=project,
+                                value=condition["value"],
+                                config=config,
+                            )
+                            condition_obj.save()
```
vgrozdanic: More Django way of doing this would be to use `update_or_create` ([docs](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#update-or-create)):

```suggestion
                        condition_id = condition["id'] if 'id' in condition else None
                        SpanAttributeExtractionRuleCondition.objects.update_or_create(id=condition_id, project=project, config=config, defaults={"value":condition["value"], "created_by_id":request.user.id}
```
Django will first check if the object exist and all the values in `default` dict will be updated. If it doesn't exist, then new record will be created with all the values passed in args

src/sentry/relay/config/metric_extraction.py
```diff
@@ -825,11 +825,15 @@ def _generate_span_attribute_specs(project: Project) -> list[HashedMetricSpec]:
     ):
         return []
 
-    extraction_rules_state = MetricsExtractionRuleState.load_from_project(project)
+    extraction_configs = SpanAttributeExtractionRuleConfig.objects.filter(project=project).all()
```
vgrozdanic, line 828: `all()` is not needed, usually, `all()` is used only if you want to select all object without using the filter method.
e.g.:
```
SpanAttributeExtractionRuleConfig.objects,all()
```

src/sentry/sentry_metrics/models/metricsextractionrules.py
```diff
@@ -0,0 +1,105 @@
...
+from django.utils import timezone
+
+from sentry.backup.scopes import RelocationScope
+from sentry.db.models import ArrayField, FlexibleForeignKey, Model, region_silo_model
+from sentry.db.models.fields.hybrid_cloud_foreign_key import HybridCloudForeignKey
+from sentry.models.project import Project
+from sentry.sentry_metrics.configuration import HARD_CODED_UNITS
+from sentry.sentry_metrics.extraction_rules import MetricsExtractionRule
+
+logger = logging.getLogger(__name__)
```
vgrozdanic, line 17: Is this used anywhere in the file?

src/sentry/api/endpoints/project_metrics_extraction_rules.py
shellmayr: Nice suggestion, changed

src/sentry/sentry_metrics/models/metricsextractionrules.py
shellmayr: Nope, removing

vgrozdanic (approved):
Nice work 👏💪

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -81,62 +88,86 @@ def post(self, request: Request, project: Project) -> Response:
...
+            with transaction.atomic(router.db_for_write(SpanAttributeExtractionRuleConfig)):
+                for obj in config_update:
+                    configs.append(
+                        SpanAttributeExtractionRuleConfig.from_dict(obj, request.user.id, project)
+                    )
+                schedule_invalidate_project_config(
+                    project_id=project.id, trigger="span_attribute_extraction_configs"
+                )
+
+        except Exception:
```
markstory, line 109: Do you want any errors captured here or are failures always user error?

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -81,62 +88,86 @@ def post(self, request: Request, project: Project) -> Response:
...
+                                "value": condition["value"],
+                                "created_by_id": request.user.id,
+                            },
+                        )
+                schedule_invalidate_project_config(
+                    project_id=project.id, trigger="span_attribute_extraction_configs"
+                )
+
+        except Exception:
+            return Response(status=400)
```
markstory, line 165: Should we capture errors here as well, or are all errors invalid user input?

src/sentry/api/serializers/models/metrics_extraction_rules.py
```diff
@@ -1,18 +1,39 @@
...
             "unit": attrs.get("unit"),
             "tags": list(attrs.get("tags") or set()),
             "conditions": list(attrs.get("conditions") or set()),
+            "id": attrs.get("id"),
         }
+
+
+class SpanAttributeExtractionRuleConfigSerializer(Serializer):
+    def __init__(self, *args, **kwargs):
+        Serializer.__init__(self, *args, **kwargs)
```
markstory, line 24:
```suggestion
        super().__init__(*args, **kwargs)
```

You should be able to use super() here.

src/sentry/sentry_metrics/models/metricsextractionrules.py
```diff
@@ -0,0 +1,102 @@
...
+from sentry.sentry_metrics.configuration import HARD_CODED_UNITS
+from sentry.sentry_metrics.extraction_rules import MetricsExtractionRule
+
+
+@region_silo_model
+class SpanAttributeExtractionRuleCondition(Model):
+    __relocation_scope__ = RelocationScope.Organization
+
+    date_modified = models.DateTimeField(default=timezone.now)
+    date_added = models.DateTimeField(default=timezone.now)
```
markstory, line 22: You probably want to inherit from `DefaultFieldsModel`. By default django won't manage `date_modified` fields. However, `DefaultFieldsModel` will manage it's `date_updated` field.

src/sentry/sentry_metrics/models/metricsextractionrules.py
```diff
@@ -0,0 +1,102 @@
...
+    def generate_mris(self):
+        mris = []
+        metric_types = MetricsExtractionRule.infer_types(self.config.aggregates)
+        for metric_type in metric_types:
+            mris.append(f"{metric_type}:custom/span_attribute_{self.id}@{self.config.unit}")
+        return mris
+
+
+@region_silo_model
+class SpanAttributeExtractionRuleConfig(Model):
```
markstory, line 45: Same here with inheriting from `DefaultFieldsModel`.

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -60,16 +65,18 @@ def get(self, request: Request, project: Project) -> Response:
             return Response(status=404)
 
         try:
-            extraction_rules = get_metrics_extraction_rules(project)
+            configs = SpanAttributeExtractionRuleConfig.objects.filter(project=project)
```
markstory, line 68: How will existing metric extraction rules be ported into new storage?

src/sentry/api/endpoints/project_metrics_extraction_rules.py
shellmayr: Good point, I added a call to capture_exception()

src/sentry/api/endpoints/project_metrics_extraction_rules.py
shellmayr: Added a call to capture_exception()

tests/sentry/models/test_user.py
```diff
@@ -387,6 +393,8 @@ def test_only_source_user_is_member_of_organization(self, expected_models: list[
         RuleActivity,
         RuleSnooze,
         SavedSearch,
+        SpanAttributeExtractionRuleConfig,
```
azaslavsky, line 396: To pass this test, and the one above, you'll need to add this model to the list at `src/sentry/services/hybrid_cloud/organization/impl.py#559`. If you want to make this sharp edge a little easier to handle for future editors of this code, it'd be nice if you edit the test failure message (which I agree is currently quite uninformative!) to explain what went wrong and (the expected list of merged models did not much the ones that were actually merged).

shellmayr: Thank you, this worked - I'll add a more verbose error message to those two tests indicating how to solve them.

PR #73427, opened 2024-06-27
feat(metrics): extend counter condition

### LINKED ISSUE #73426: Extend metric extraction spec condition for counters
counters that target span attribute `foo` should have their conditions implicitly extended with `has:foo` to ensure we are only counting spans that contain the attribute `foo`

obostjancic:
- closes: #73426


vgrozdanic (approved):

shellmayr (approved):
LGTM

PR #73434, opened 2024-06-27
ref(project-details): Convert class to func component

priscilawebdev:
### Goal

### Note

Although this component was using the `AsyncView`, it did not trigger any initial requests, and therefore did not utilize the load states as expected


static/app/views/projectDetail/projectDetail.tsx
```diff
@@ -48,158 +49,100 @@ type RouteParams = {
...
+        tagKey: key,
+        search,
+        projectIds: location.query.project ? [location.query.project] : undefined,
+        endpointParams: location.query,
+      });
+    },
+    [api, organization.slug, location.query]
+  );
+
+  const syncProjectWithSlug = useCallback(() => {
```
ArthurKnaus, line 110: Nit: you could move this into the `useEffect` as it is only used there. Then it would be clearer from the dependency array, when exactly it gets executed.

static/app/views/projectDetail/projectDetail.tsx
```diff
@@ -48,158 +49,100 @@ type RouteParams = {
...
+  const {query} = location.query;
+  const hasPerformance = organization.features.includes('performance-view');
+  const hasDiscover = organization.features.includes('discover-basic');
+  const hasTransactions = hasPerformance && project?.firstTransactionEvent;
+  const projectId = project?.id;
+  const isProjectStabilized =
+    defined(project?.id) &&
+    project.id === location.query.project &&
+    project.id === String(selection.projects[0]);
+  const visibleCharts = ['chart1'];
```
ArthurKnaus, line 70: You could memoize `visibleCharts` as it is not just used for iterating but also as a prop for `<ProjectCharts />`.

ArthurKnaus (approved):
🚀

vgrozdanic (approved):

PR #73437, opened 2024-06-27
fix(clipped-box): Fix content being clipped when `prefers-reduced-motion` is set

### LINKED ISSUE #73408: Stack trace variables values are clipped when expanding objects/arrays with prefers-reduced-motion
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

1. Set your browser preference for `prefers-reduced-motion: reduce` 
2. Load an issue stack trace with variable data
3. Expand/collapse values

### Expected Result

Variable values should always be fully visible, and the container should expand in size with them.

### Actual Result

With prefers-reduced-motion, the container does not expand correctly, clipping the values.

### Product Area

Issues

### Link

_No response_

### DSN

_No response_

### Version

_No response_

malwilley:
Fixes https://github.com/getsentry/sentry/issues/73408

This was only happening when the content started at a height that was not clipped, and the height was changed by toggling an inner element. The max-height was set but never removed by the `onTransitionEnd` handler.

Adds a util to check for this in JS in the future if necessary.

### DIFF
--- a/static/app/utils/usePrefersReducedMotion.tsx
+++ b/static/app/utils/usePrefersReducedMotion.tsx
@@ -1,0 +1,5 @@
+import useMedia from 'sentry/utils/useMedia';
+
+export function usePrefersReducedMotion() {
+  return !useMedia('(prefers-reduced-motion: no-preference)');
+}
--- a/static/app/utils/usePrefersReducedMotion.tsx
+++ b/static/app/utils/usePrefersReducedMotion.tsx
@@ -1,5 +1,0 @@
-import useMedia from 'sentry/utils/useMedia';
-
-export function usePrefersReducedMotion() {
-  return !useMedia('(prefers-reduced-motion: no-preference)');
-}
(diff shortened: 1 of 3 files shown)


static/app/utils/usePrefersReducedMotion.tsx
```diff
@@ -0,0 +1,5 @@
+import useMedia from 'sentry/utils/useMedia';
+
+export function usePrefersReducedMotion() {
+  return !useMedia('(prefers-reduced-motion: no-preference)');
+}
```
evanpurkhiser, line 5: Should we use https://www.framer.com/motion/guide-accessibility/#usereducedmotion

evanpurkhiser (approved):

static/app/utils/usePrefersReducedMotion.tsx
malwilley: Yup, didn't realize we were already using that

PR #73439, opened 2024-06-27
ref(onboarding): add hook for gettingStartedDocs

roggenkemper:
this replaces a useEffect that was used in a few different places with a custom hook.


static/app/components/onboarding/gettingStartedDoc/utils/useLoadGettingStarted.tsx
```diff
@@ -0,0 +1,57 @@
...
+  feedbackOnboardingPlatforms,
+  replayPlatforms,
+} from 'sentry/data/platformCategories';
+import type {PlatformKey} from 'sentry/types/project';
+
+export default function useLoadGettingStarted({
+  platformId,
+  platformPath,
+  replayPlatformCheck,
+  feedbackPlatformCheck,
```
michellewzhang, line 15: what if instead of having these two props, we did something like

`productType?: 'replay' | 'feedback'` and then do similar casing below

static/app/components/onboarding/gettingStartedDoc/utils/useLoadGettingStarted.tsx
```diff
@@ -0,0 +1,57 @@
...
+  useEffect(() => {
+    async function getGettingStartedDoc() {
+      if (replayPlatformCheck && !replayPlatforms.includes(platformId)) {
+        setModule('none');
+        return;
+      }
+      if (feedbackPlatformCheck && !feedbackOnboardingPlatforms.includes(platformId)) {
+        setModule('none');
+        return;
+      }
```
michellewzhang, line 39:
```suggestion
      if ((productType === 'replay' && !replayPlatforms.includes(platformId)) || (productType === 'feedback' && !feedbackOnboardingPlatforms.includes(platformId))) {
        setModule('none');
        return;
      }
```
something like this could remove a little duplication (sorry for bad formatting)

michellewzhang (approved):

michellewzhang (commented):
this is super nit and also my fault, but i don't like how i named the feedback one `useLoadFeedbackOnboardingDoc` and the replays one just `useLoadOnboardingDoc` ... do you mind renaming these to be either the same or the replay one to be `useLoadReplayOnboardingDoc`?

PR #73440, opened 2024-06-27
feat(new-trace): Incrementally rendering traces in replay details

Abdkhan14:
- Loads the traces associated with a trace incrementally.
- Makes already loaded traces intractable.


static/app/views/performance/newTraceDetails/index.tsx
```diff
@@ -331,12 +334,20 @@ export function TraceViewWaterfall(props: TraceViewWaterfallProps) {
...
+    if (traceData) {
+      return TraceTree.FromTrace(traceData, props.replayRecord, {
+        api,
+        filters,
+        additionalTraceDataRows: props.traceDataRows?.slice(1),
+        organization: props.organization,
+        traceLimit: undefined,
+        urlParams,
+        rerender: forceRerender,
+      });
```
JonasBa, line 346: There is no need to put this in a constructor. We should construct the tree in the same way as we did before with a minimal set of arguments needed to create a valid trace tree and call a separate method later on to fetch this.

JonasBa:
@Abdkhan14 This looks a bit too complicated than it needs to be so I'd like us to pair review this and make some adjustments

static/app/views/performance/newTraceDetails/index.tsx
Abdkhan14: Hey, @JonasBa I pulled out the incremental logic from the constructor and added a function called `incrementallyFetchTraces` that does the job.

static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
```diff
@@ -596,6 +624,124 @@ export class TraceTree {
     return tree.build();
   }
 
+  async fetchTraces(options: IncrementalTraceFetchOptions) {
```
JonasBa, line 627:
```suggestion
  async fetchTraces(options: TraceFetchOptions) {
```

static/app/views/performance/newTraceDetails/index.tsx
```diff
@@ -331,12 +335,32 @@ export function TraceViewWaterfall(props: TraceViewWaterfallProps) {
...
+    if (trace) {
+      return TraceTree.FromTrace(trace, props.replayRecord);
     }
 
     throw new Error('Invalid trace state');
-  }, [props.traceSlug, props.trace, props.status, projects, props.replayRecord]);
+  }, [props.traceSlug, trace, isLoading, error, projects, props.replayRecord]);
+
+  useEffect(() => {
+    if (props.replayTraces) {
```
JonasBa, line 346: You dont want to start fetching before the tree is assembled
```suggestion
    if (!props.replayTraces.length || tree.type !== 'trace') {
```

Abdkhan14: I check for that inside the method

JonasBa: Yes, but that is unnecessarily hidden. Just do it here and only call it if you need to

static/app/views/performance/newTraceDetails/index.tsx
```diff
@@ -338,6 +349,26 @@ export function TraceViewWaterfall(props: TraceViewWaterfallProps) {
...
+        urlParams,
+        rerender: forceRerender,
+      });
+    }
+
+    return () => {
+      tree.stopFetching();
+    };
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
JonasBa, line 369: We should remove this and add all deps. The if condition inside the effect should be the place where we determine if we want to fetch this or not

JonasBa (approved):

PR #73441, opened 2024-06-27
feat(query-builder): Combine filter key with operator button

malwilley:
This solves a couple issues with the current UI:

- is and had filters didn't make sense as `is is` and `has is`
- `is` button had a very small click area without adding too much visual padding
- Filters felt unbalanced with a key that is non-editable (could add key editing, but that adds a lot of complexity and isn't often useful)

`is` and `has` are now special cases where we show custom operator labels. Everything else should look similar, just with a merged button.

CleanShot 2024-06-27 at 09 49 15

CleanShot 2024-06-27 at 09 48 57

CleanShot 2024-06-27 at 09 49 02


roggenkemper (approved):
looks good! definitely an improvement

PR #73442, opened 2024-06-27
[CapMan visibility] emit a warning when a query gets throttled by the capacity management system

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2799

Allocation policies are our mechanism for doing traffic management for Snuba queries. Currently, the result of applying allocation policies in the internal API is simply accept/reject/throttle, and Snuba sends back a payload to Sentry that contains metadata about those policy decisions. In the case of a throttled query, we want to emit a warning to GCP as well as a warning to Sentry Issues

### DIFF
--- a/src/sentry/utils/snuba.py
+++ b/src/sentry/utils/snuba.py
@@ -1058,6 +1058,13 @@ def _bulk_snuba_query(
                     span.set_tag(k, v)
                     sentry_sdk.set_tag(k, v)
 
+                if (
+                    "throttled_by" in quota_allowance_summary
+                    and quota_allowance_summary["throttled_by"]
+                ):
+                    logger.warning("Query is throttled", extra={"response.data": response.data})
+                    sentry_sdk.capture_message("Query is throttled", level="warning")
+
             if response.status != 200:
                 _log_request_query(snuba_param_list[index][0])
                 metrics.incr(


src/sentry/utils/snuba.py
```diff
@@ -1053,6 +1053,9 @@ def _bulk_snuba_query(
             for k, v in quota_allowance_summary["rejected_by"].items():
                 span.set_tag(k, v)
 
+            if "rejected_by" in quota_allowance_summary and quota_allowance_summary["rejected_by"]:
```
volokluev, line 1056: `rejected_by` is rejection not a throttle.

src/sentry/utils/snuba.py
```diff
@@ -1053,6 +1053,9 @@ def _bulk_snuba_query(
...
+                logger.warning("Query is throttled")
```
volokluev, line 1057: You probably want more info here. This is going to create a sentry issue. Think about the product developer looking at it, what are they supposed to do with this information?

xurui-c: I think only rejected queries will create a sentry issue. The response.status of throttled queries should still be 200. Since there is no sentry issue that displays all the helpful metadata for throttled queries, I figured it doesn't hurt to just log the entire response.data, because we don't know what info product developers can need.

volokluev: logging.warning is how you create warnings in sentry (which become sentry issues)

volokluev (approved):

PR #73443, opened 2024-06-27
ref(querybuilder): Change metric summaries base class

wmak:
- Stop inheriting Discover QueryBuilder in favour of the BaseQueryBuilder


wmak:
This is getting annoying going to make 1 big PR

PR #73445, opened 2024-06-27
ref(querybuilder): Change span indexed base class

wmak:
- Change the base class for span table queries to the BaseQueryBuilder


wmak:
ok this is getting annoying, going to make 1 big PR instead

PR #73451, opened 2024-06-27
ref(flag): Move performance-use-metrics to flagpole

k-fish:
### Summary
We need this to be options backed since we're testing with it in STs. Moves the flag to flagpole.


Zylphrex (approved):

k-fish:
STs don't have flagpole setup yet

PR #73460, opened 2024-06-27
perf(query-builder): Use startsWith/endsWith for some glob conditions

Zylphrex:
Additional optimization here for common glob patterns for starts and ends with.


vartec (approved):
LGTM.

nit: lot of that code would benefit from using [match](https://docs.python.org/3/tutorial/controlflow.html#match-statements) instead of chained if-elif

src/sentry/api/event_search.py
```diff
@@ -387,6 +387,56 @@ def is_wildcard(self) -> bool:
...
+                # the pattern also matched on some leading slashes.
+                leading_wildcard = True
+            elif end == len(self.raw_value):
+                # It only needs to match on end because if it matches on
+                # some slashes before the *, that's okay.
+                trailing_wildcard = True
+            else:
+                middle_wildcard = True
+
+        if not middle_wildcard:
```
wmak, line 413: To confirm I understand this, middle wildcard is something like
`foo*bar` which is a classification of wildcard we don't care about because we can't optimize it right?

wmak (approved):
LGTM, minor nit, I think the middle_wildcard becoming `other` might be better explained with a comment (assuming i understood that correctly)

src/sentry/api/event_search.py
Zylphrex: It's a little more generic than that. `middle_wildcard` is there to capture all cases where there's a `*` in the middle of the query like your example `foo*bar` or `f*o*o`. There's no easy way to optimize these types of wildcards, and they're less common so I choose to just not optimize them.

PR #73464, opened 2024-06-27
ref(onboarding): Rename `useLoadOnboardingDoc` for replays

roggenkemper:
this pr updates the name so it better aligns with feedback


michellewzhang (approved):
thanks!

PR #73465, opened 2024-06-27
perf:(similarity) add one more bigtable thread

JoshFerge:
add one more thread to bigtable mutlithread get so for a batch size of 30, we make 5 concurrent calls.


JoshFerge:
on a separate note, i'm not sure what the optimal number of threads is, nor the optimal number of rows to read from bigtable. if we want to tune this we should probably make both of these options and play around with it., but for now this is good enough. I know for sure any bigtable reads with more than 10 rows are much slower.

JoshFerge:
closing, gonna make it an option

PR #73467, opened 2024-06-27
ref(replay): Refactor some replay stepping stuff for re-use with diff tools

ryan953:


static/app/utils/replays/extractDomNodes.tsx
```diff
@@ -14,8 +14,19 @@ export type Extraction = {
...
+   */
   frames: ReplayFrame[] | undefined;
+
+  /**
+   * The rrweb events that constitute the replay
+   */
   rrwebEvents: RecordingFrame[] | undefined;
+
+  /**
+   * The replay startTimestampMs
```
michellewzhang, line 28:
```suggestion
   * The replay start time, in ms
```

static/app/utils/replays/extractDomNodes.tsx
```diff
@@ -14,8 +14,19 @@ export type Extraction = {
 };
 
 type Args = {
+  /**
+   * frames where we should stop and extract html for a given dom node
```
michellewzhang, line 18:
```suggestion
   * Frames where we should stop and extract html for a given dom node
```

michellewzhang (approved):
feel free to ignore the nits

c298lee (approved):

PR #73473, opened 2024-06-27
ref(slack): use sdk webhook client for webhook actions

cathteng:
Adds feature-flagged usage of the Slack SDK WebhookClient to update slack messages when actions are taken on them (archive, assign, resolve, etc).

### DIFF
--- a/src/sentry/integrations/slack/webhooks/action.py
+++ b/src/sentry/integrations/slack/webhooks/action.py
@@ -10,9 +10,9 @@ from django.urls import reverse
 from rest_framework import serializers
 from rest_framework.request import Request
 from rest_framework.response import Response
-from slack_sdk import WebhookClient
 from slack_sdk.errors import SlackApiError
 from slack_sdk.models.views import View
+from slack_sdk.webhook import WebhookClient
 
 from sentry import analytics, features
 from sentry.api import client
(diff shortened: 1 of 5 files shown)


tests/sentry/integrations/slack/webhooks/actions/test_status.py
```diff
@@ -271,30 +271,6 @@ def test_ask_linking(self):
         assert resp.data["response_type"] == "ephemeral"
         assert resp.data["text"] == LINK_IDENTITY_MESSAGE.format(associate_url=associate_url)
 
-    def test_archive_issue(self):
```
cathteng, line 274: this was testing for backwards compatibility which is no longer necessary https://github.com/getsentry/sentry/pull/73248/files#diff-09dc9b9a6030a931823ed51799486b635133b203c044824cafb66debb62fe1c6L290

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -666,15 +666,26 @@ def _handle_group_actions(
...
-        except ApiError as e:
-            logger.error("slack.action.response-error", extra={"error": str(e)})
+
+        response_url = slack_request.data["response_url"]
+        if not features.has("organizations:slack-sdk-webhook-handling", group.project.organization):
+            slack_client = SlackClient(integration_id=slack_request.integration.id)
+            try:
+                slack_client.post(response_url, data=response, json=True)
+            except ApiError as e:
+                logger.error("slack.action.response-error", extra={"error": str(e)})
```
iamrajjoshi, line 680: Do we want a metric here?

iamrajjoshi (approved):

src/sentry/integrations/slack/webhooks/action.py
cathteng: i add metrics during feature flag cleanup, but yes

Issue #71665, opened 2024-05-29
[Replay Details] Implement support for visualizing 'gaps' in the replay timeline


ryan953:
Per figma: https://www.figma.com/design/LC8Uhws6ruiNYUkObfqcjv/Specs%3A-Segment-Gaps-in-Mobile-Replay?node-id=0-1&t=9GoyanWjtcULZMIX-0

![Image](https://github.com/getsentry/sentry/assets/187460/b96fdf12-a512-4d96-b8bf-5ca7048e5e3a)

There might be cases where the 'gap' still has events inside it. Therefore the gap should be rendered behind all other content, including the 'ticks' within the timeline.

Initially the gap should be merged with support for mobile `background` & `foreground` events. But it should be applicable to use web `focus` & `blur` events just as easily, we can decide on that later.
```[tasklist]
### Tasks
- [x] Create gaps in timeline
- [x] Add in  tooltip
- [x] Implement gap in segments visualization
- [ ] Refactor tooltip to look like designs (tooltip with time is on top of the timeline + scrubber & hovering on scrubber has same behaviour as timeline))
- [x] [BUG] zooming timeline breaks in https://sentry-sdks.sentry.io/replays/2298b9c818db427cae385baf7945dc5b/?project=5428559&project=5428557&project=5428561&query=&referrer=%2Freplays%2F&statsPeriod=14d&yAxis=count%28%29
```

bruno-garcia:
Gaps are currently available for Sentry for dogfooding, and it's based on timestamps of frame breadcrumbs (https://github.com/getsentry/sentry/pull/74406).
We'll still look at adding support for gaps when segments are missing.

c298lee:
Timeline gaps are based on video events as of https://github.com/getsentry/sentry/pull/79274. Currently all gaps are visible on the timeline for the SDK team to see how often there's gaps in the video. Before releasing this feature, this would need to be revisited to ensure that the gaps shown don't cause any confusion (probably by implementing a minimum gap time)

bruno-garcia:
Gaps are looking good in general but we'll verify the dev experience in react native before GA'ing

cmanallen:
Consider closing this issue if we're just waiting on GA.

c298lee:
Closing as this has been GA'd

PR #73475, opened 2024-06-27
feat(replay): Gap in replay timeline

c298lee:
Creates gaps in the timeline between `App in Background` and `App in Foreground` frames. If there's no `App in Foreground` frame after `App in Background`, the gap continues to the end of the replay.

Example gap:
image

Example gap to end of replay:
image

### DIFF
--- a/static/app/components/replays/breadcrumbs/timelineGaps.tsx
+++ b/static/app/components/replays/breadcrumbs/timelineGaps.tsx
@@ -27,7 +27,8 @@ export default function TimelineGaps({durationMs, frames, width}: Props) {
   );
 
   // returns all numbers in the range
-  const range = (start, stop) => Array.from({length: stop - start}, (_, i) => start + i);
+  const range = (start, stop) =>
+    Array.from({length: stop - start}, (_, i) => start + i + 1);
 
   const gapCol: number[] = [];
 
@@ -56,11 +57,11 @@ export default function TimelineGaps({durationMs, frames, width}: Props) {
 
     // create gap if we found have start (background frame) and end (foreground frame)
     if (start !== -1 && end !== -1) {
-      gapCol.push(...range(start + 1, end + 1));
+      gapCol.push(...range(start, end));
     }
     // if we have start but no end, that means we have a gap until end of replay
     if (start !== -1 && end === -1) {
-      gapCol.push(...range(start + 1, totalColumns + 1));
+      gapCol.push(...range(start, totalColumns));
     }
   }
 
@@ -74,7 +75,7 @@ export default function TimelineGaps({durationMs, frames, width}: Props) {
 }
 
 const Gap = styled(Timeline.Col)<{column: number}>`
-  grid-column: ${p => Math.floor(p.column)};
+  grid-column: ${p => p.column};
   background: ${p => p.theme.red300};
   text-align: right;
   line-height: 14px;
--- a/static/app/components/replays/breadcrumbs/timelineGaps.tsx
+++ b/static/app/components/replays/breadcrumbs/timelineGaps.tsx
@@ -76,8 +76,7 @@ export default function TimelineGaps({durationMs, frames, width}: Props) {
 
 const Gap = styled(Timeline.Col)<{column: number}>`
   grid-column: ${p => p.column};
-  background: ${p => p.theme.red300};
-  text-align: right;
+  background: ${p => p.theme.gray400};
   line-height: 14px;
-  opacity: 50%;
+  opacity: 16%;
 `;
--- a/static/app/components/replays/breadcrumbs/timelineGaps.tsx
+++ b/static/app/components/replays/breadcrumbs/timelineGaps.tsx
@@ -26,7 +26,7 @@ export default function TimelineGaps({durationMs, frames, width}: Props) {
     totalColumns
   );
 
-  // returns all numbers in the range
+  // returns all numbers in the range, exclusive of start and inclusive of stop
   const range = (start, stop) =>
     Array.from({length: stop - start}, (_, i) => start + i + 1);
 
--- a/static/app/components/replays/breadcrumbs/timelineGaps.tsx
+++ b/static/app/components/replays/breadcrumbs/timelineGaps.tsx
@@ -1,0 +1,70 @@
+import styled from '@emotion/styled';
+
+import * as Timeline from 'sentry/components/replays/breadcrumbs/timeline';
+import {getFramesByColumn} from 'sentry/components/replays/utils';
+import {
+  isBackgroundFrame,
+  isForegroundFrame,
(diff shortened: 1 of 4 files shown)


michellewzhang (commented):
generally it looks good so far! will leave the floor open for others to comment :)

static/app/components/replays/breadcrumbs/timelineGaps.tsx
```diff
@@ -0,0 +1,70 @@
...
+    frames.filter(f => isBackgroundFrame(f)),
+    totalColumns
+  );
+  const framesByColEnd = getFramesByColumn(
+    durationMs,
+    frames.filter(f => isForegroundFrame(f)),
+    totalColumns
+  );
+  const range = (start, stop) =>
+    Array.from({length: (stop - start) / 1}, (_, i) => start + i);
```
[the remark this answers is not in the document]
c298lee: thanks for catching that!

static/app/components/replays/breadcrumbs/timelineGaps.tsx
```diff
@@ -0,0 +1,70 @@
...
+  let end = endKeys.next();
+
+  while (!start.done) {
+    while (start.value > end.value) {
+      console.log(start.value, end.value);
+      end = endKeys.next();
+      if (end.done) {
+        end.value = totalColumns - 1;
+      }
+    }
```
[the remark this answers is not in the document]
c298lee: Updated the logic and added comments, lmk if everything makes sense!

static/app/components/replays/breadcrumbs/timelineGaps.tsx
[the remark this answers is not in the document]
c298lee: ty for looking into this!!

static/app/components/replays/breadcrumbs/timelineGaps.tsx
```diff
@@ -0,0 +1,70 @@
...
+} from 'sentry/utils/replays/types';
+
+interface Props {
+  durationMs: number;
+  frames: ReplayFrame[];
+  width: number;
+}
+
+export default function TimelineGaps({durationMs, frames, width}: Props) {
+  const markerWidth = frames.length < 200 ? 4 : frames.length < 500 ? 6 : 10;
```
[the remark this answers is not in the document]
c298lee: I think it's used to determine the width of each column, not completely sure about why it differs tho, I copied it over from `ReplayTimelineEvents`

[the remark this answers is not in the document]
michellewzhang: ahhh cool! didn't know it was copied over, seems to work so cool w me🥳

static/app/components/replays/breadcrumbs/timelineGaps.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const totalColumns = Math.floor(width / markerWidth);
+  const framesByCol = getFramesByColumn(
+    durationMs,
+    frames.filter(f => isBackgroundFrame(f) || isForegroundFrame(f)),
+    totalColumns
+  );
+
+  // returns all numbers in the range
```
michellewzhang, line 29:
```suggestion
  // returns all numbers in the range, exclusive of start and inclusive of stop
```

michellewzhang (approved):
new logic & comments look good to me - seems like we have all the cases covered. 

- no background/foreground fames --> skip the big while loop, no gaps rendered.
- only background frames for some reason --> we'll only set `start`, and it'll be a gap from `start` to the end.
- only foreground frames for some reason --> `start` never gets set, and we'll reach the end of `gapFrames`. we don't push anything to `gapCol`, so we don't render any gaps.
- mix of background/foreground frames --> iterate through the `gapFrames` matching up `start` and `end` to render the gaps, until we reach one of the cases above

(lmk if any of those sound wrong)

michellewzhang:
would be good to test with a few more replays before merging - you can ping @romtsn or @brustolin or @krystofwoldrich to generate some!

static/app/components/replays/breadcrumbs/timelineGaps.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const range = (start, stop) =>
+    Array.from({length: stop - start}, (_, i) => start + i + 1);
+
+  const gapCol: number[] = [];
+
+  const gapFrames = framesByCol.entries();
+  let currFrame = gapFrames.next();
```
michellewzhang, line 36: do we need to reset `currFrame` after each while loop iteration? looks like it's outside the main loop

i see we're resetting `start` and `end` to be `-1` each time we push something into `gapCol` though, which is good

static/app/components/replays/breadcrumbs/timelineGaps.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+// create gaps in the timeline by finding all columns between a background frame and foreground frame
+// or background frame to end of replay
+export default function TimelineGaps({durationMs, frames, width}: Props) {
+  const markerWidth = frames.length < 200 ? 4 : frames.length < 500 ? 6 : 10;
+
+  const totalColumns = Math.floor(width / markerWidth);
+  const framesByCol = getFramesByColumn(
+    durationMs,
+    frames.filter(f => isBackgroundFrame(f) || isForegroundFrame(f)),
```
ryan953, line 25: This type of filter could be put into ReplayReader and memoized. `replayTimeline.tsx` could call something like `replay.getFooFrames()` to fetch just the list we need here.

static/app/utils/replays/types.tsx
```diff
@@ -193,6 +193,14 @@ export function isHydrationErrorFrame(
   return frame.category === 'replay.hydrate-error';
 }
 
+export function isBackgroundFrame(frame: ReplayFrame): frame is BreadcrumbFrame {
+  return frame && 'category' in frame && frame.category === 'app.background';
+}
+
+export function isForegroundFrame(frame: ReplayFrame): frame is BreadcrumbFrame {
```
ryan953, line 200: It could be useful if we had more specific types instead of `frame is BreadcrumbFrame`

ryan953 (approved):
Accepting to unblock

static/app/utils/replays/types.tsx
c298lee: I get type issues when I try to use `BackgroundFrame` instead :(

ryan953: No. We can iterate on it later

krystofwoldrich:
[Here](https://sentry-sdks.sentry.io/replays/7249042e9fb040cb813f7e4412824615) is an example replay with multiple app suspensions. Looks good.

PR #73476, opened 2024-06-27
feat(query-builder): Organize and sort issue/event filters better

malwilley:
- Removes event properties from the "Issue filters" tab and places them in a new "Event filters" category
- Sort event tags by count within the "Event tags" section so that more useful ones are at the top

CleanShot 2024-06-27 at 15 24 38

CleanShot 2024-06-27 at 15 24 42

CleanShot 2024-06-27 at 15 26 54


scttcper (approved):
wow very organized

PR #73477, opened 2024-06-27
nit(processing): Use frozen time for tests

schew2381:


tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -117,24 +122,25 @@ def setUp(self):
         event_frequency_percent_condition = self.create_event_frequency_condition(
             interval="5m", id="EventFrequencyPercentCondition", value=1.0
         )
-        self.now = datetime.now(UTC)
```
saponifi3d, line 120: FYI for next time - it could be a little easier to just set `self.now` to the FROZEN_TIMESTAMP to reduce the amount of refactoring you need to do.

saponifi3d (approved):

Issue #70687, opened 2024-05-10
[Mobile Replay - Bug] Video player plays wrong clip


michellewzhang:
See comments below for relevant examples

michellewzhang:
Looked into this a bit. Seems to be a weird replay because in `loadSegmentAtTime`, `segmentIndex` is always undefined, which seems to relate to a gap: cc @billyvg 

https://github.com/getsentry/sentry/assets/56095982/7e7bbc84-eea5-4d8c-ad57-c5f0fbf6b7ad

https://github.com/getsentry/sentry/blob/1491155c67bec5eb0c4a9f89de282cbbe2316cf9/static/app/components/replays/videoReplayer.tsx#L555-L560

michellewzhang:
example replay: https://sentry-sdks.sentry.io/replays/e63c8004ce07468a9e461d75d4d5944b/?project=5428557&query=&referrer=%2Freplays%2F&statsPeriod=90d&yAxis=count%28%29

this one & the example above may be due to the fact that the ios sdk is sending `video.duration` in seconds (rather than milliseconds). need to validate when the SDKs are sending the duration in ms

michellewzhang:
additional example with android SDK (which is not related to the seconds/ms problem): https://sentry-sdks.sentry.io/replays/f7797884538f49288eccf9497def2634/?project=5428559&query=&referrer=%2Freplays%2F&statsPeriod=14d&yAxis=count%28%29

if you play/pause around 5, 10, 15, and ~19 seconds, you'll notice some flickering as well. i believe that this is related to this chunk of code: https://github.com/getsentry/sentry/blob/0298eef00139c5bb6838e382064aaacd6177926e/static/app/components/replays/videoReplayer.tsx#L321-L323

c298lee:
> additional example with android SDK (which is not related to the seconds/ms problem): https://sentry-sdks.sentry.io/replays/f7797884538f49288eccf9497def2634/?project=5428559&query=&referrer=%2Freplays%2F&statsPeriod=14d&yAxis=count%28%29
> 
> if you play/pause around 5, 10, 15, and ~19 seconds, you'll notice some flickering as well. i believe that this is related to this chunk of code - maybe something to do with the fact that our segment is ending early and we've accidentally clicked into this gap:
> 
> https://github.com/getsentry/sentry/blob/0298eef00139c5bb6838e382064aaacd6177926e/static/app/components/replays/videoReplayer.tsx#L321-L323

[...]

PR #73481, opened 2024-06-27
fix(replay): fix flickering when play/pause mobile replay

michellewzhang:
- relates to https://github.com/getsentry/sentry/issues/70687
- this fixes some of the bugs where playing/pausing a mobile replay results in two different frames being shown
- this one was happening because we were seeking into a part of the replay where the segment ends, and we didn't put in any sort of buffer. ex. the video duration is sent in as 5000ms but the segment ends before that, so we run out of content to show.

before: notice the flickering

https://github.com/getsentry/sentry/assets/56095982/846f99d5-2145-480c-9081-be228e0c6a9c

after:

https://github.com/getsentry/sentry/assets/56095982/21c26ebe-64c2-405d-8ec6-8b60abe42c2b

another example before:

https://github.com/getsentry/sentry/assets/56095982/98f837d7-ce30-4fdc-b946-8b2bf6346ace

after:

https://github.com/getsentry/sentry/assets/56095982/c50d30cc-d073-4f34-8382-571e96233cfd

### DIFF
--- a/static/app/components/replays/videoReplayer.tsx
+++ b/static/app/components/replays/videoReplayer.tsx
@@ -318,13 +318,15 @@ export class VideoReplayer {
     }
 
     // Final check in case replay was stopped immediately after a video
-    if (!this._isPlaying) {
+    if (!this._isPlaying && nextIndex !== this._currentIndex) {
       // we can get here if we play/pause into a gap at the end of the video segment
       // to avoid showing flickering states when the user pauses,
       // (e.g., showing the next frame too eagerly), we want to stay on the last old
       // frame until it's time to show the next frame.
       const currSeg = this.getSegment(nextIndex - 1);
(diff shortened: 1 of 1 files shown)


static/app/components/replays/videoReplayer.tsx
```diff
@@ -318,7 +318,15 @@ export class VideoReplayer {
     }
 
     // Final check in case replay was stopped immediately after a video
-    if (!this._isPlaying) {
+    if (!this._isPlaying && nextIndex !== this._currentIndex) {
+      // we can get here if we play/pause into a gap at the end of the video segment
+      // to avoid showing flickering states when the user pauses,
+      // (e.g., showing the next frame too eagerly), we want to stay on the last old
```
michellewzhang, line 324:
```suggestion
      // (e.g., showing a different frame too eagerly), we want to stay on the last old
```

PR #73482, opened 2024-06-27
fix(sentry-sdk): Correctly fork the scope for the thread pool

Zylphrex:
I'm not sure I fully understand isolation vs current scopes in the sdk yet but this is adapted from https://github.com/getsentry/sentry-python/blob/b38f9c707e0b0ca42587a20cc1498b575fa4f895/sentry_sdk/integrations/threading.py.

The current implementation causing the validation and run spans to not be attached to the transaction. I believe what's happening is that the transaction is only on the current scope and not the isolation scope (I don't know why). As a result, when starting a span in the thread pool function, it's not attached to the transaction. This mimicks what the threading integration does and passes both the isolation and current scopes and sets them in the thread pool function.

### DIFF
--- a/src/sentry/utils/snuba.py
+++ b/src/sentry/utils/snuba.py
@@ -1004,7 +1004,13 @@ def _bulk_snuba_query(
                 _query_thread_pool.map(
                     _snuba_query,
                     [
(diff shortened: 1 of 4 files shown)


antonpirker (approved):
I think this should work like this!

PR #73483, opened 2024-06-28
ref(feedback): clarify user report notification setting

aliu39:
Displayed in user-feedback project settings. IMO the old title is confusing because both "crash report" and "user report" link to crash report docs. Also the "User Report API" isn't really mentioned in our docs, since it's deprecated. Correct me if I'm wrong about any of this, including the SDK version

### DIFF
--- a/static/app/data/forms/userFeedback.tsx
+++ b/static/app/data/forms/userFeedback.tsx
@@ -28,7 +28,7 @@ const formGroups: JsonFormObject[] = [
         label: t('Enable Crash Report Notifications'),
         help: () =>
           tct(
-            'Get notified on [crashReportModalDocsLink: Crash Report Modal and User Report API submissions]. [feedbackWidgetDocsLink: Feedback widget] notifications are not affected by this setting and are on by default.',
+            'Get notified on [crashReportModalDocsLink: Crash Report Modal] and User Feedback API (SDK v7.X and older) submissions. [feedbackWidgetDocsLink: Feedback widget] notifications are not affected by this setting and are on by default.',
             {
               crashReportModalDocsLink: (
                 <a href="https://docs.sentry.io/platforms/javascript/user-feedback/#crash-report-modal" />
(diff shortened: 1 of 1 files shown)


JoshFerge (approved):

michellewzhang:
you can probably link to the user feedback api docs: https://docs.sentry.io/platforms/javascript/user-feedback/#user-feedback-api

michellewzhang:
also not sure if the v7 comment is needed -- it's JS specific, but the user might be using a different sentry SDK right?

aliu39:
> you can probably link to the user feedback api docs: https://docs.sentry.io/platforms/javascript/user-feedback/#user-feedback-api

I believe this is the updated api, which won't send user reports

aliu39:
> also not sure if the v7 comment is needed -- it's JS specific, but the user might be using a different sentry SDK right?

Cc @bruno-garcia - what are your thoughts on how to word "old" vs "new" here? I don't know much about our SDK versioning

static/app/data/forms/userFeedback.tsx
```diff
@@ -28,7 +28,7 @@ const formGroups: JsonFormObject[] = [
         label: t('Enable Crash Report Notifications'),
         help: () =>
           tct(
-            'Get notified on [crashReportModalDocsLink: Crash Report Modal and User Report API submissions]. [feedbackWidgetDocsLink: Feedback widget] notifications are not affected by this setting and are on by default.',
+            'Get notified on [crashReportModalDocsLink: Crash Report Modal] and User Feedback API (web endpoint and pre-SDK v8) submissions. [feedbackWidgetDocsLink: Feedback widget] notifications are not affected by this setting and are on by default.',
```
bruno-garcia, line 31: Could we add a `webApiEndpointLink` also?

```suggestion
            'Get notified on [crashReportModalDocsLink: Crash Report Modal] and User Feedback API ([webApiEndpointLink:web endpoint] and pre-JS SDK v8) submissions. [feedbackWidgetDocsLink: Feedback widget] notifications are not affected by this setting and are on by default.',
```

bruno-garcia (approved):

bruno-garcia:
Thanks @aliu39

PR #73484, opened 2024-06-28
feat(rate-limiting): leaky bucket rate limiter

vartec:
[Leaky bucket](https://en.wikipedia.org/wiki/Leaky_bucket) rate limiter which allows for defining both max burst rate as well as sustained rate limit. 

- While the bucket is not yet full, the requests are accepted until it's full (bucket size = max burst limit)
- When the bucker is full, new request are accepted only after a drop drips out (drip rate = sustained rate limit)

This is a metered (i.e. counter) implementation, which is far less resource heavy, but only limits the traffic, doesn't attempt to artificially smooth it out (i.e. doesn't implement a queue mechanism). 

## Why it's better than sliding window?

When set up for the same average sustained traffic, when actual traffic that is significantly higher than sustained rate the _sliding_window_ limiter will create short bursts every window_size seconds, while _leaky_bucket_ will allow for initial burst, but after that will only let in a steady stream of traffic.
bucket vs sliding window


src/sentry/ratelimits/leaky_bucket.py
```diff
@@ -0,0 +1,213 @@
...
+
+            @limiter(key, limited_handler=my_limited_handler)
+            def my_function():
+                rv = do_something()
+            ```
+
+        """
+
+        def decorator(func):
+            def wrapper(*args, **kwargs):
```
vgrozdanic, line 201: Is there a reason why we are not using `functools.wraps` decorator here?

vartec: that's a good point, might be useful if we generate docs or such, I'll add it.

src/sentry/ratelimits/leaky_bucket.py
```diff
@@ -0,0 +1,212 @@
...
+
+        return LeakyBucketLimitInfo(
+            self.burst_limit,
+            self.drip_rate,
+            last_drip,
+            current_level,
+            max(0, (current_level - self.burst_limit) / self.drip_rate),
+        )
+
+    def context(self, key: str) -> ContextManager[LeakyBucketLimitInfo]:
```
vgrozdanic, line 101: What is the benefit of having a context manager here, why not just use `is_limited` method for checking?

To me it seems like an added complexity, we need to handle possible exception that is being raised from it, there is no clean up that is being done after the context is exited, and it seems to me that it can all be achieved with just calling `limiter.is_limited(key)`, but maybe I am missing some very obvious reason

src/sentry/ratelimits/leaky_bucket.py
```diff
@@ -0,0 +1,212 @@
...
+        return f"{self.NAMESPACE}:{key}"
+
+    def validate(self) -> None:
+        try:
+            self.client.ping()
+            self.client.connection_pool.disconnect()
+        except Exception as e:
+            raise InvalidConfiguration(str(e))
+
+    def use_and_get_info(self, key: str, timestamp: float | None = None) -> LeakyBucketLimitInfo:
```
armenzg, line 54: nit: Could `key` be part of the `__init__` method? That way you don't need to pass the key when calling the various methods.

src/sentry/ratelimits/leaky_bucket.py
```diff
@@ -0,0 +1,212 @@
...
+            bucket_size, drip_rate, last_drip, current_level, wait_time = leaky_bucket_info(
+                [redis_key],
+                [self.burst_limit, self.drip_rate, timestamp],
+                client=self.client,
+            )
+            last_drip = float(last_drip)
+            current_level = float(current_level)
+            wait_time = float(wait_time)
+            info = LeakyBucketLimitInfo(bucket_size, drip_rate, last_drip, current_level, wait_time)
+            return info
```
armenzg, line 69: You can probably do this in just one line instead of two.

armenzg (approved):

wedamija:
This makes sense, although I'm wondering if it's much different to `RedisSlidingWindowRateLimiter` in practice?

vartec:
@wedamija 
> This makes sense, although I'm wondering if it's much different to `RedisSlidingWindowRateLimiter` in practice?

It allows absorbing bursts (bucket size) while at the same time allows defining significantly lower sustained rate (drip rate), without needing to tie both of them together.  

Though to your point, if you'd set _LeakyBucket_ to burst rate `X` and drip rate to `Y` per second, it's would give you similar result as setting _SlidingWindow_ to quota `X` and window length `X/Y` seconds and granularity of at least `1/Y` seconds.

That being said LeakyBucket is far simpler algorithm and this implementation doesn't have the granularity vs data size trade-off. With this implementation you have microsecond granularity storing just a single hash with two floats (_current_level_, and _last_drip_). 

Also, this is more opinion based, but I believe _burst_ and _sustained rate_ are easier to reason about than sliding window params.

src/sentry/scripts/ratelimits/leaky_bucket.lua
```diff
@@ -0,0 +1,74 @@
...
+local wait_time = tostring((new_level - bucket_size) / drip_rate)
+
+if allowed then
+  redis.call("hset", key, "current_level", new_level)
+  redis.call("hset", key, "last_drip", current_time)
+  redis.call("expire", key, max_tll_seconds)
+  wait_time = 0
+end
+
+-- Redis converts all Lua numbers to ints, so we need to convert flaots to strings, then back to floats in Python
```
wedamija, line 73: Nit: float

wedamija (approved):

src/sentry/ratelimits/leaky_bucket.py
vartec: made it optional, also made decorator use original function's fully qualified name as default key

vartec:
> Though to your point, if you'd set _LeakyBucket_ to burst rate `X` and drip rate to `Y` per second, it's would give you similar result as setting _SlidingWindow_ to quota `X` and window length `X/Y` seconds and granularity of at least `1/Y` seconds.

Actually giving it more thought, this is not true... The overall throughput will be the same, but behavior with sustained traffic will be different.

bucket vs sliding window

src/sentry/ratelimits/leaky_bucket.py
```diff
@@ -0,0 +1,203 @@
...
+
+            @limiter(key, limited_handler=my_limited_handler)
+            def my_function():
+                rv = do_something()
+            ```
+
+        """
+
+        def decorator(func: Callable[[Any], Any]) -> Callable[[Any], Any]:
+            @functools.wraps(func)
```
asottile-sentry, line 185: this should use `Callable[P, R] -> Callable[P, R]` to preserve the signature (`P` being the conventional name for a `ParamSpec` and `R` being a TypeVar for the return value): https://youtu.be/fwZoxWyMGM8

vartec: @asottile-sentry https://github.com/getsentry/sentry/pull/74687

PR #73486, opened 2024-06-28
nit(rollup): Use reverse instead of sorting

schew2381:
We're creating the list of buckets in the series by appending going backwards in time, meaning the largest timestamp will be at the front of the list and the last item will be the smallest.

We can avoid sorting by just reversing the list.

### DIFF
--- a/src/sentry/tsdb/base.py
+++ b/src/sentry/tsdb/base.py
@@ -275,6 +275,7 @@ class BaseTSDB(Service):
         while timestamp >= start:
             series.append(self.normalize_to_epoch(timestamp, rollup))
             timestamp = timestamp - timedelta(seconds=rollup)
+
         return rollup, series[::-1]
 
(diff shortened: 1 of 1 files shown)


src/sentry/tsdb/base.py
```diff
@@ -276,7 +276,7 @@ def get_optimal_rollup_series(
             series.append(self.normalize_to_epoch(timestamp, rollup))
             timestamp = timestamp - timedelta(seconds=rollup)
```
schew2381, line 277: In this loop, `timestamp` is only being set backwards in time so it's guaranteed that we're appending to list in reverse order (so the list is descending)

cathteng (approved):
🧠

saponifi3d (approved):
looks like the existing tests should also cover this change 🎉  - :shipit:

Issue #73263, opened 2024-06-19
Remove Sentry SDK `Hub` 


antonpirker:
We still use the Hub in `sentry`. We must remove it everywhere, and replace the `Hub` usages with `Scope` usages.

```[tasklist]
### Subtasks
- [ ] #65869 
- [ ] https://github.com/getsentry/sentry/pull/73357
- [ ] #73360
- [ ] #73422
- [ ] #73424
- [ ] #73433
- [ ] https://github.com/getsentry/sentry/pull/73488
```

Related issue: #73423

PR #73488, opened 2024-06-28
test(minimetrics): Replace `Hub` with `Scope`

szokeasaurusrex:
Since the `Hub` API is deprecated, replace all uses of the `Hub` API in the minimetrics tests with the equivalent `Scope` API.

ref #73263


antonpirker (approved):
looks cool

Issue #73423, opened 2024-06-27
Migrate `sentry_sdk.push_scope` calls to SDK 2.0 API


szokeasaurusrex:
`sentry_sdk.push_scope` is deprecated as of Sentry Python SDK version 2.0.0. Each of these calls should be replaced with the (roughly) equivalent `sentry_sdk.isolation_scope`.

See also [#73263](https://github.com/getsentry/sentry/issues/73263)
```[tasklist]
### Implementing PRs
- [ ] #73490
- [ ] https://github.com/getsentry/sentry/pull/73491
- [ ] #73492
- [ ] https://github.com/getsentry/sentry/pull/73495
- [ ] https://github.com/getsentry/sentry/pull/73501
- [ ] https://github.com/getsentry/sentry/pull/73502
- [ ] https://github.com/getsentry/sentry/pull/73503
- [ ] https://github.com/getsentry/sentry/pull/73504
```

antonpirker:
Will this change some data available after the with block? (sorry for asking again, I know we have talked this through yesterday, but cant remember the outcome)

szokeasaurusrex:
@antonpirker As far as I can tell, the only difference between `push_scope` and `isolation_scope` is that `push_scope` allows the current scope to bleed outside the block, whereas `isolation_scope` restores the `current_scope` after exiting the block. So, if we put something on the current scope while inside the `push_scope` block, that thing is still there after the `push_scope` block. However, if we replace the `push_scope` with an `isolation_scope`, anything added to the current scope within the `isolation_scope` block is not available on the current scope outside the block.

This change in behavior would not matter for someone who changes all of their push_scopes to isolation_scopes at the time when they upgrade to 2.0, since current scopes are a completely new concept in 2.0 with no analog in 1.0, as we discussed offline yesterday. However, someone who keeps using `push_scope` for awhile on 2.0 and then switches to `isolation_scope` could notice a change in behavior. I would argue, however, that the fact that `push_scope` does not clear the current scope is a bug with `push_scope`, since I believe this means that `push_scope` will not behave similarly in 2.0 as it does in 1.0.

In short, right now I believe that `push_scope` in 1.x is equivalent to `isolation_scope` in 2.x, but `push_scope` in 2.x is slightly different due to the current scope bleed (and the fact that it is different could be considered a bug)

PR #73492, opened 2024-06-28
ref: Replace `push_scope` with `isolation_scope` in backpressure code

szokeasaurusrex:
ref #73423

### DIFF
--- a/src/sentry/processing/backpressure/health.py
+++ b/src/sentry/processing/backpressure/health.py
@@ -77,7 +77,7 @@ def is_consumer_healthy(consumer_name: str = "default") -> bool:
                     "reason": reason,
                 },
             )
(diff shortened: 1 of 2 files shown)


antonpirker:
### Note

Changing:
```python
with sentry_sdk.push_scope() as scope:
    scope.set_tag("tagkey", "tagvalue")
```
to 
```python
with sentry_sdk.isolation_scope() as scope:
    scope.set_tag("tagkey", "tagvalue")
```
will change the data available after the `with` block. 

The data changes made in the `with push_scope()` block where still there after exiting the with block. 
Using `with isolation_scope` will restore the scope after exiting the with block, so all data changes inside the block will be lost after the with block.

antonpirker (approved):

PR #73494, opened 2024-06-28
ref(metrics): rename name param to mri

obostjancic:
`BaseMetricsTestCase.store_metric` first parameter is named `name` even though the method expects a valid MRI to be passed

This PR:
- changes the parameter name from `name` to `mri`
- uses the `mri` to extract `use_case_id` and `type` (metric type)
- removes `use_case_id` and `type` parameters


shellmayr (approved):
Nice, it's good that we refactor these things to make APIs more coherent - LGTM! 🚢

PR #73495, opened 2024-06-28
ref: Replace `push_scope` with `isolation_scope` in `src/sentry/tasks`

szokeasaurusrex:
ref #73423

### DIFF
--- a/src/sentry/tasks/app_store_connect.py
+++ b/src/sentry/tasks/app_store_connect.py
@@ -203,7 +203,7 @@ def inner_refresh_all_builds() -> None:
     options = ProjectOption.objects.filter(key=appconnect.SYMBOL_SOURCES_PROP_NAME)
     count = 0
     for option in options:
-        with sdk.push_scope() as scope:
+        with sdk.isolation_scope() as scope:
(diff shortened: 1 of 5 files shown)


loewenheim (approved):

antonpirker (approved):

antonpirker:
### Note

Changing:
```python
with sentry_sdk.push_scope() as scope:
    scope.set_tag("tagkey", "tagvalue")
```
to 
```python
with sentry_sdk.isolation_scope() as scope:
    scope.set_tag("tagkey", "tagvalue")
```
will change the data available after the `with` block. 

The data changes made in the `with push_scope()` block where still there after exiting the with block. 
Using `with isolation_scope` will restore the scope after exiting the with block, so all data changes inside the block will be lost after the with block.

Swatinem (approved):

szokeasaurusrex:
@antonpirker There is no difference in your example because `push_scope` yields the isolation scope`, which _does_ reset after exiting the `push_scope` block. There is only a difference when you explicitly get the current scope within the `push_scope` block

PR #73496, opened 2024-06-28
ref(onboarding): Consolidate ionic and capacitor docs

priscilawebdev:
### Problem

In our user feedback product, we currently display capacitor getting started documentation even when ionic is selected. This results from the strong integration between ionic and capacitor, which are commonly used together and share the same Sentry SDK configuration.

However, inconsistencies exist in how we present the ionic documentation between user feedback and the main onboarding process:

- The user feedback documentation offers additional features such as selecting a sibling framework (Angular 12+, React, Vue, etc.), whereas the onboarding documentation for ionic does not provide these options.
- The onboarding documentation for capacitor allows users to opt-in or opt-out of performance and replay products, whereas the onboarding documentation for ionic lacks these options.

### Solution

To address these issues, we propose removing the ionic getting started documentation entirely. Instead, when users select ionic as their platform during the main onboarding process, we will integrate the capacitor documentation. This approach will ensure consistency across platforms and provide users with uniform options for integrating performance and replay functionalities.


priscilawebdev:
@lucas-zimerman, I was wondering if we don't need to add new in front of SentrySibling. Does it create an instance without it?

Screenshot 2024-06-28 at 11 18 21

priscilawebdev:
Also, just to be absolutely sure:

We didn't display the information about uploading source maps, and the guide for verifying errors in Ionic was different from that for Capacitor... Is that all resolved now? I assume so, right?

static/app/utils/gettingStartedDocs/getPlatformPath.ts
```diff
@@ -3,6 +3,12 @@ import type {PlatformIntegration} from 'sentry/types/project';
 export function getPlatformPath(platform: PlatformIntegration) {
   // some platforms use a naming convention that combines 'language' and 'id' with a hyphen in between. For example, 'react-native'.
   if (platform.id === platform.language) {
+    // Ionic is a special case as it is closely intertwined with Capacitor and often used together.
+    // Consequently, they can utilize the same SDK documentation, specifically that of Capacitor.
+    if (platform.language === 'ionic') {
+      return 'capacitor/capacitor';
+    }
```
ArthurKnaus, line 10: Nit: I think it would be a bit more explicit to have a `ionix.tsx` file that imports and exports the capacitor docs.
That way other places that might import it will automatically function correctly.

e.g.:
https://github.com/getsentry/sentry/blob/master/static/app/components/replaysOnboarding/useLoadOnboardingDoc.tsx

We do the same for `apple.tsx` already.

static/app/components/onboarding/productSelection.tsx
```diff
@@ -92,6 +92,7 @@ export const platformProductAvailability = {
   'apple-ios': [ProductSolution.PERFORMANCE_MONITORING, ProductSolution.PROFILING],
   'apple-macos': [ProductSolution.PERFORMANCE_MONITORING, ProductSolution.PROFILING],
   bun: [ProductSolution.PERFORMANCE_MONITORING],
+  capacitor: [ProductSolution.PERFORMANCE_MONITORING, ProductSolution.SESSION_REPLAY],
```
priscilawebdev, line 95: this change is just to keep the list in alphabetical order ^^

static/app/gettingStartedDocs/capacitor/capacitor.tsx
```diff
@@ -63,6 +63,50 @@ const platformOptions: Record<PlatformOptionKey, PlatformOption> = {
 type PlatformOptions = typeof platformOptions;
 type Params = DocsParams<PlatformOptions>;
 
+function getIntegrations(params: Params, siblingOption: string) {
```
priscilawebdev, line 66: these changes make the code a bit more readable

ArthurKnaus (commented):
🚀 Code looks good to me, will leave approval up to @lucas-zimerman.

Also with this merged, we can finally `de-duplicate` the `getPlatformPath` logic in
https://github.com/getsentry/sentry/blob/master/static/app/components/feedback/feedbackOnboarding/useLoadFeedbackOnboardingDoc.tsx
and
[static/app/components/replaysOnboarding/useLoadOnboardingDoc.tsx](https://github.com/getsentry/sentry/blob/master/static/app/components/replaysOnboarding/useLoadOnboardingDoc.tsx)

Also we can consolidate those hooks with the logic in 
https://github.com/getsentry/sentry/blob/master/static/app/components/onboarding/gettingStartedDoc/sdkDocumentation.tsx

static/app/utils/gettingStartedDocs/getPlatformPath.ts
```diff
@@ -3,6 +3,12 @@ import type {PlatformIntegration} from 'sentry/types/project';
 export function getPlatformPath(platform: PlatformIntegration) {
   // some platforms use a naming convention that combines 'language' and 'id' with a hyphen in between. For example, 'react-native'.
   if (platform.id === platform.language) {
+    // Ionic is a special case as it is closely intertwined with Capacitor and often used together.
+    // Consequently, they can utilize the same SDK documentation, specifically that of Capacitor.
+    if (platform.language === 'ionic') {
```
priscilawebdev, line 8: If we continue adding special use cases, we may need to reconsider this function.

priscilawebdev:
hey @ArthurKnaus thank you for your review!

Yes, I plan to focus on cleaning up the code in future PRs 🙂

static/app/utils/gettingStartedDocs/getPlatformPath.ts
priscilawebdev: Initially, I did that, but then I thought, "Why keep a .tsx file with just 3 lines of code and a .spec.tsx file to test the same thing we are already testing in Capacitor?" I don't have strong opinions on this, though

ArthurKnaus: I think it is more transparent. As an engineer that is not familiar with this code base I would not find this redirect in `getPlatformPath`.  No hard opinion on the test file 🤷 Most important, in the end it should be consistently handled over all the platforms.

lucas-zimerman:
> @lucas-zimerman, I was wondering if we don't need to add new in front of SentrySibling. Does it create an instance without it? I mean it is a pseudo code but....
> 
> Screenshot 2024-06-28 at 11 27 37

for `browserTracingIntegration` new is not needed, despite being on V7, this specific integration initializes the same way on both V7 and V8.

lucas-zimerman:
> We didn't display the information about uploading source maps for ionic before... I assume that since everything now is a capacitor, it does not matter right?

Looks good to me, the same method for uploading source-maps also applies to ionic

lucas-zimerman (approved):
Looks good to me, LGTM!

ArthurKnaus (approved):
🙌

PR #73497, opened 2024-06-28
feat(apis): Pass query source to snuba - phase 2

sentaur-athena:
Continuing https://github.com/getsentry/sentry/pull/73176 to pass query source to snuba for more debuggability. 

OrganizationSpansSamplesEndpoint
OrganizationsEventsMetaEndpoint
OrganizationsEventsStatsEndpoint

### DIFF
--- a/src/sentry/api/endpoints/organization_events_meta.py
+++ b/src/sentry/api/endpoints/organization_events_meta.py
@@ -14,7 +14,9 @@ from sentry.api.helpers.group_index import build_query_params_from_request
 from sentry.api.serializers import serialize
 from sentry.api.serializers.models.group import GroupSerializer
 from sentry.api.utils import handle_query_errors
+from sentry.middleware import is_frontend_request
 from sentry.snuba import spans_indexed, spans_metrics
+from sentry.snuba.query_sources import QuerySource
 from sentry.snuba.referrer import Referrer
 
 
@@ -39,6 +41,10 @@ class OrganizationEventsMetaEndpoint(OrganizationEventsEndpointBase):
                 params=params,
                 query=request.query_params.get("query"),
                 referrer=Referrer.API_ORGANIZATION_EVENTS_META.value,
+                # TODO: @athena - add query_source when all datasets support it
+                # query_source=(
+                #     QuerySource.FRONTEND if is_frontend_request(request) else QuerySource.API
+                # ),
             )
 
         return Response({"count": result["data"][0]["count"]})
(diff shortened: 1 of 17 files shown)


src/sentry/api/endpoints/organization_events_meta.py
```diff
@@ -39,6 +41,10 @@ def get(self, request: Request, organization) -> Response:
                 params=params,
                 query=request.query_params.get("query"),
                 referrer=Referrer.API_ORGANIZATION_EVENTS_META.value,
+                # TODO: @athena - add query_source when all datasets support it
+                # query_source=(
+                #     QuerySource.FRONTEND if is_frontend_request(request) else QuerySource.API
+                # ),
```
Zylphrex, line 47: Did you mean to leave this commented out?

src/sentry/api/endpoints/organization_events_stats.py
```diff
@@ -201,6 +215,8 @@ def get(self, request: Request, organization: Organization) -> Response:
                 if referrer in ALLOWED_EVENTS_STATS_REFERRERS.union(METRICS_ENHANCED_REFERRERS)
                 else Referrer.API_ORGANIZATION_EVENT_STATS.value
             )
+            if referrer in SENTRY_BACKEND_REFERRERS:
+                query_source = QuerySource.SENTRY_BACKEND
```
Zylphrex, line 219: Can this logic be moved into `is_frontend_request`? Also, is there no other way to verify that the request came from the backend by inspecting the request like the authentication method?

Zylphrex (approved):
Mostly looks good to me, though it would be ideal if the logic to detect if the query source can be collected in 1 place instead of having some endpoints override it with some custom logic.

src/sentry/api/endpoints/organization_events_meta.py
sentaur-athena: Yes, not all datasets currently expect query_source as a param and they're not typed to safely take care of that. So I'll uncomment it after my last PR.

src/sentry/api/endpoints/organization_events_stats.py
sentaur-athena: `is_frontend_request` is using authentication but really no matter how we implement it a customer's direct api call can look like a frontend request so this is just a good estimate.

I don't think it's a good idea to move this to is_frontend_request because that's really independent of Snuba. Let me create a method in `Endpoint` or a util 🤔

PR #73499, opened 2024-06-28
feat(apis): Pass query source to snuba - phase 3

sentaur-athena:
Continuing https://github.com/getsentry/sentry/pull/73497 to pass query source to snuba for more debuggability.

OrganizationEventsTraceBaseEndpoint
OrganizationsEventsNewTrendsStatsEndpoint
OrganizationsEventsEndpoint

### DIFF
--- a/src/sentry/api/endpoints/organization_events.py
+++ b/src/sentry/api/endpoints/organization_events.py
@@ -21,7 +21,6 @@ from sentry.apidocs.parameters import GlobalParams, OrganizationParams, Visibili
 from sentry.apidocs.utils import inline_sentry_response_serializer
 from sentry.discover.models import DiscoverSavedQuery, DiscoverSavedQueryTypes
 from sentry.exceptions import InvalidParams
-from sentry.middleware import is_frontend_request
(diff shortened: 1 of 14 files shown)


src/sentry/api/endpoints/organization_events.py
```diff
@@ -338,6 +340,8 @@ def get(self, request: Request, organization) -> Response:
         elif referrer not in ALLOWED_EVENTS_REFERRERS:
             referrer = Referrer.API_ORGANIZATION_EVENTS.value
 
+        query_source = QuerySource.FRONTEND if is_frontend_request(request) else QuerySource.API
```
Zylphrex, line 343: Just a thought but it'd be nice if this was wrapped into a helper function like

```python
def get_request_source(request: Request) -> QuerySource:
  if is_frontend_request(request):
    return QuerySource.FRONTEND
  return QuerySource.API
```

This also gives you a single place to update in the future if there are additional query sources instead of having to update every caller with additional checks.

Zylphrex (approved):

PR #73508, opened 2024-06-28
ref: remove sentry.lint.engine and get all js linting and formatting into pre-commit

joshuarli:
Cleaned up version of https://github.com/getsentry/sentry/pull/73037.

I was trying very hard to only install what's needed in pre-commit-managed nodeenvs, but wasn't able to get it to work and now I'm opting of just using the entire node_modules (setting up volta + node deps takes 25s on a warm cache).

```
      # We're using the local node_modules to simplify things, otherwise would
      # need an easy way to keep versions in sync. Furthermore some repositories
      # are either not on pre-commit mirrors or are missing some tagged versions.
      # pre-commit-managed nodeenv environments also do not work out of the box because
      # additional_dependencies does not install a flattened dependency tree which is needed by,
      # for example, our eslint setup.
```


.github/workflows/pre-commit.yml
```diff
@@ -53,6 +53,12 @@ jobs:
             all:
               - added|modified: '**/*'
 
+      - uses: getsentry/action-setup-volta@e4939d337b83760d13a9d7030a6f68c9d0ee7581 # v2.0.0
+
+      - name: Install node dependencies
+        run: |
+          yarn install --frozen-lockfile
```
evanpurkhiser, line 60: @scttcper I wonder if there's some way we could say "Just install what's needed for linting"

.pre-commit-config.yaml
```diff
@@ -100,6 +100,57 @@ repos:
...
+      - id: prettier
+        name: prettier (yaml, markdown, tsx, jsx, css)
+        language: system
+        # TODO: Remove tsx and jsx when Biome supports styled CSS formatting.
+        types_or: [yaml, markdown, tsx, jsx, css]
+        entry: ./node_modules/.bin/prettier --log-level=error --write
+        # https://pre-commit.com/#regular-expressions
+        exclude: |
+          (?x)^($^
+            # prettier crashes on some of these, by intent:
```
evanpurkhiser, line 139: crashes or has errors?

.github/workflows/frontend.yml
```diff
@@ -76,65 +68,14 @@ jobs:
           echo "::add-matcher::.github/tsc.json"
           echo "::add-matcher::.github/eslint-stylish.json"
```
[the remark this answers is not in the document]
joshuarli: What do those github matches look like? It's probably fine since the pre-commit action will just autocommit changes.

.github/workflows/pre-commit.yml
joshuarli: Yeah, I already tried very hard to do that with the v1 PR. Couldn't figure it out. 25 seconds from warm cache is fine though, I realized it wasn't worth it to keep going.

.pre-commit-config.yaml
joshuarli: I have no idea, this was from the previous hook.

.github/workflows/frontend.yml
```diff
@@ -47,22 +47,14 @@ jobs:
           filters: .github/file-filters.yml
           list-files: shell
 
-  typescript-and-lint:
+  typescript:
```
billyvg, line 50: will renaming this break anything in getsentry, or some other GHA task?

joshuarli: nope, this is encapsulated in the Frontend required check

gggritso:
Does this change mean that we've never running linters/formatters on the entire codebase, since the "all files" steps are gone? If we're not, is that safe? I assume it is, since `tsc` still runs on everything, but wondering if there are any cases where changing a file might cause a problem in an unchanged file and it won't be linted

joshuarli:
> Does this change mean that we've never running linters/formatters on the entire codebase, since the "all files" steps are gone? If we're not, is that safe? I assume it is, since `tsc` still runs on everything, but wondering if there are any cases where changing a file might cause a problem in an unchanged file and it won't be linted

Yes, it's fine, tsc still runs on everything. The pre-commit action'll commit autofixes to branches, and it's required to merge, so I don't expect tsc to fail on master.

.pre-commit-config.yaml
```diff
@@ -100,6 +100,57 @@ repos:
...
+      # are either not on pre-commit mirrors or are missing some tagged versions.
+      # pre-commit-managed nodeenv environments also do not work out of the box because
+      # additional_dependencies does not install a flattened dependency tree which is needed by,
+      # for example, our eslint setup.
+
+      - id: biome
+        name: biome (javascript, typescript, json)
+        language: system
+        types: [text]
+        files: \.(jsx?|tsx?|css)$
```
gggritso, line 115: Maybe `types_or` would be better here? Not a strong opinion

gggritso (approved):
I like it. I think putting `pre-commit` in charge of figuring out what's changed and linting it makes sense. It will make sure that more bad changes are caught locally _and_ will auto-fix them in PRs rather than just letting them fail linting checks.

The possibility that bad changes will somehow slip through and get into `master` seems like it's more me being paranoid than a real thing that will happen and it's easy to fix, so, IMO, go forth

.pre-commit-config.yaml
joshuarli: oh, will just remove this entirely, `files` there is fine + we have that for most of the other hooks

PR #73514, opened 2024-06-28
chore(similarity): Add metric for no stacktrace frame

jangjodi:
Add metric for when an event has a non-empty stacktrace string that we'd send to seer, but has no stacktrace frames
Limit metric to backend calls

### DIFF
--- a/src/sentry/seer/similarity/utils.py
+++ b/src/sentry/seer/similarity/utils.py
@@ -2,6 +2,7 @@ import logging
 from typing import Any
 
 from sentry.eventstore.models import Event
+from sentry.utils import metrics
 from sentry.utils.safe import get_path
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


src/sentry/seer/similarity/utils.py
```diff
@@ -74,6 +75,15 @@ def get_stacktrace_string(data: dict[str, Any]) -> str:
         if frame_str:
             stacktrace_str += frame_str
 
+    # Increment metric if we send seer a non-empty stacktrace, but there are no frames
+    if stacktrace_str and not has_contributing_frame and referrer == "backend":
+        metrics.incr(
+            "grouping.similarity.no_contributing_stacktrace_frame",
+            # TODO: Consider lowering this (in all the spots this metric is
+            # collected) once we roll Seer grouping out more widely
+            sample_rate=1.0,
```
mrduncan, line 84: Not necessary as a part of this change - we have a handful of these high sample rates, thoughts on having a single ML grouping sample rate setting somewhere that we can lower/change over time?

lobsterkatie: Yeah, I did that [here](https://github.com/getsentry/sentry/blob/7d1d7699e67e2bb8718aeff031dcde250186d711/src/sentry/seer/similarity/similar_issues.py#L24). It's only used for the one metric, but we could make it more general.

@jangjodi, if you want to go ahead and get this merged, I can handle doing that.

jangjodi: Sounds good, thanks @lobsterkatie

PR #73515, opened 2024-06-28
chore(trace-explorer): Put back project selector

Zylphrex:
Reverting this decision, the project selector will be put back and we'll go with an UX solution to explain the trace root coming from a different project.


Zylphrex:
Timing of this PR is bad, will put this up again when it's ready.

PR #73521, opened 2024-06-28
fix: Disable Invite Members button if org requires SSO

### LINKED ISSUE #56889: Disabling `Invite Members` directly from Sentry if SSO is enabled and required
### Problem Statement

A lot of our customers manage their Sentry membership with third-party identity providers, and their requests and approvals workflow for access lives entirely outside of Sentry. Even with SSO and/or SCIM enabled and required, users still have the option to Invite Members and request for access to org Owners and Managers as shown in the screenshot.

This is problematic for 2 reasons:
1. Users will believe that is the only way to get access to Sentry, and
2. Org Owners and Managers get noisy notifications (by default through email) about people requesting access to Sentry even if they're not the ones in charge of the IdP
Screenshot 2023-09-25 at 9 32 20 AM

### Solution Brainstorm

One solution is to make it an option to disable Inviting Members through Sentry if you have SSO enabled. One step further would be to allow Organization access admins to customize the message there and direct people to the correct place to request for Sentry access.

The main result of this would be to disable the noisy alerts that Owners and Managers are currently receiving.

### Product Area

Settings - Members

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-1946) by [Unito](https://www.unito.io)

RyanSkonnord:
Requires https://github.com/getsentry/sentry/pull/73593

Resolves https://github.com/getsentry/sentry/issues/56889


RyanSkonnord:
Up for preliminary review. Still needs some new unit tests, and probably to be split into front-end and back-end PRs.

static/app/types/organization.tsx
```diff
@@ -70,6 +70,7 @@ export interface Organization extends OrganizationSummary {
     projectLimit: number | null;
   };
   relayPiiConfig: string | null;
+  requiresSso: boolean;
```
cathteng, line 73: if we merge BE and FE at the same time, this is not guaranteed to be sent from the backend when the frontend is merged, especially since backend deploys take longer. it might error unless we also allow `null` here

RyanSkonnord:
Back end changes split into https://github.com/getsentry/sentry/pull/73593. Now ready for review.

static/app/types/organization.tsx
RyanSkonnord: Would it be more sensible to do `requiresSso?`. I see no problem treating `null` and `false` interchangeably.

cathteng: i think that works too

RyanSkonnord:
Hmm, I'm getting a type error due to https://github.com/getsentry/getsentry/blob/master/static/getsentry/gsApp/components/inviteMembersButtonCustomization.tsx#L4

So there's a chicken-and-egg problem in syncing the type change between Sentry and Getsentry. AFAICT a simple fix would be not to change the name of `disabled` and to make `disabledBySso` optional. But I prefer not to have `disabled` be ambiguous like that. If there's a good way to sync up the changes otherwise, let me know.

static/app/types/organization.tsx
evanpurkhiser: Let's just deploy the backend first.

static/app/types/hooks.tsx
```diff
@@ -631,10 +631,8 @@ type IntegrationsFeatureGatesHook = () => {
  */
 type InviteButtonCustomizationHook = () => React.ComponentType<{
   children: (opts: {
-    /**
-     * Whether the Invite Members button is active or not
-     */
-    disabled: boolean;
+    disabledByFlag: boolean;
+    disabledBySso: boolean;
```
evanpurkhiser, line 635: Can we add comments? What is "disabled by flag"

evanpurkhiser, line 635:
```suggestion
    disabledByFlag: boolean;
    disabledBySso: boolean;
```

evanpurkhiser, line 635: I think you'll need to keep the `disabled` around for BC otherwise getsentry is going to fail no?

static/app/views/settings/organizationMembers/organizationMembersWrapper.tsx
```diff
@@ -172,6 +175,19 @@ function renderInviteMembersButton({
     >
       {action}
     </Hovercard>
+  ) : disabledBySso ? (
+    <Hovercard
+      body={
+        <Fragment>
+          {t(
+            `Your organization must use its single sign-on provider to invite
+            new members.`
```
evanpurkhiser, line 184: Is the phrasing here right? Are they "inviting" members via SSO, or are they "registering" members?

```suggestion
            `Your organization must use its single sign-on provider to register new members.`
```

static/app/views/settings/organizationMembers/organizationMembersWrapper.tsx
```diff
@@ -172,6 +175,19 @@ function renderInviteMembersButton({
     >
       {action}
     </Hovercard>
+  ) : disabledBySso ? (
+    <Hovercard
```
evanpurkhiser, line 179: Should this be a hovercard? Feels like this should just be a tooltip?

static/app/types/hooks.tsx
RyanSkonnord: Yeah, that's essentially what I was asking about [here](https://github.com/getsentry/sentry/pull/73521#issuecomment-2200739255). I'll revert the name change.

static/app/views/settings/organizationMembers/organizationMembersWrapper.tsx
RyanSkonnord: Sure. I was just imitating the other "disabled" message.

Any further, subjective tweaks to phrasing or presentation are welcome. I was just winging it on the message to the user.

static/app/views/settings/organizationMembers/organizationMembersWrapper.tsx
```diff
@@ -161,17 +164,27 @@ function renderInviteMembersButton({
...
-      }
-    >
-      {action}
-    </Hovercard>
+    isSsoRequired ? (
+      <Tooltip
+        title={t(
+          `Your organization must use its single sign-on provider to register new members.`
+        )}
+      >
```
evanpurkhiser, line 172: I think you can add `skipWrappper` here
```suggestion
      <Tooltip
        skipWrapper
        title={t(
          `Your organization must use its single sign-on provider to register new members.`
        )}
      >
```

cathteng (approved):

PR #73524, opened 2024-06-28
chore(trace-explorer): Start enforcing project selection behind flag

Zylphrex:
We're reverting the decision to allow cross project queries. This starts enforcing it again behind a feature flag.


k-fish (approved):
👍

PR #73527, opened 2024-06-28
fix(similarity): use acks late and provisioned celery queue

JoshFerge:
https://cloudlogging.app.goo.gl/8gf44SpPH74KLP6F9

we observe our celery task getting terminated pre-maturely. let's try using acks_late. also, ops provisioned a queue for us, so let's use it!


JoshFerge:
docs on acks_late: https://docs.celeryq.dev/en/main/userguide/tasks.html

jangjodi (approved):

PR #73531, opened 2024-06-28
fix(stacktrace-link): Fix gitlab integration url being incorrectly generated

MichaelSun48:
Fixes an issue where gitlab urls were not being formatted correctly. Specifically:

* We were not stripping filepaths of their trailing slash
* We did change the url generation logic to parse backslashes correctly after Windows code mappings were supported

### DIFF
--- a/src/sentry/integrations/utils/code_mapping.py
+++ b/src/sentry/integrations/utils/code_mapping.py
@@ -387,8 +387,6 @@ def convert_stacktrace_frame_path_to_source_path(
     """
 
     stack_root = code_mapping.stack_root
-    if "\\" in code_mapping.stack_root:
-        stack_root = code_mapping.stack_root.replace("\\", "/")
 
     # In most cases, code mappings get applied to frame.filename, but some platforms such as Java
     # contain folder info in other parts of the frame (e.g. frame.module="com.example.app.MainActivity"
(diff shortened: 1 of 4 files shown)


malwilley:
@MichaelSun48 could you add a test to verify that backslashes will work with stack trace linking? I'm not sure this is the right place since the request to check if a file exists is in the gitlab client `check_file()` which doesn't look like it calls this function.

I _think_ the right place to do this would be in `convert_stacktrace_frame_path_to_source_path` so that the transformation happens earlier.

MichaelSun48:
@malwilley Thanks for the call out! After digging into it I'm pretty sure you're right – I added a test and modified `convert_stacktrace_frame_path_to_source_path` to support backslash paths and code mappings

src/sentry/integrations/utils/code_mapping.py
```diff
@@ -399,13 +397,21 @@ def convert_stacktrace_frame_path_to_source_path(
...
+            .replace("\\", "/")
+        )
 
     # Some platforms only provide the file's name without folder paths, so we
     # need to use the absolute path instead. If the code mapping has a non-empty
     # stack_root value and it matches the absolute path, we do the mapping on it.
     if frame.abs_path and frame.abs_path.startswith(code_mapping.stack_root):
-        return frame.abs_path.replace(stack_root, code_mapping.source_root, 1)
+        return (
+            frame.abs_path.replace(stack_root, code_mapping.source_root, 1)
```
MichaelSun48, line 411: We do `.replace("\\", "/")` _after_ we replace the stack root with the source root in the abs_path because the stack_root maintains the slash type from the stacktrace path it was derived from. 

This is what acuse the issue before: on lines 390-391 of the previous version, we replaced the backslashes with slashes in the stack root _before_ we applied the code mapping to the abs_path. But since the abs_path does not also have this transformation, it will never apply correctly. 

(There's **a lot** of variables and things to keep track of here, lmk if some more clarification into my thinking would be helpful)

malwilley: I think this makes sense. It _does_ break some existing functionality, if someone had a stack_root with backslashes but had code with front slashes. But that doesn't sound very likely. You could be super safe and run the replacement on everything to avoid that if you wanted to

src/sentry/integrations/utils/code_mapping.py
```diff
@@ -399,13 +397,21 @@ def convert_stacktrace_frame_path_to_source_path(
...
+            .lstrip("/")
```
malwilley, line 412: Might make more sense to `.lstrip("/")` after the replacement in case there is a leading backslash

src/sentry/integrations/utils/code_mapping.py
MichaelSun48: Yeah that definitely is pretty unlikely, and I would argue that if that were the case, then it would make more sense to generate an entirely new code mapping for stacktrace frame with forward slashes – there are some nuances that make it not just a 1 to 1 translation if that makes sense

malwilley (approved):

PR #73535, opened 2024-06-28
fix(logging): updating logging 'throwaways' to Python 3.12 + adding Formatter fields

vartec:
Underlying problem — `flake8-logging` does not fully catch these if they're coming for a dynamic dictionary rather than a literal.  

Changes:
- Updating the `throwaway` list of `extra` fields that class with `LogRecord` fields to Python 3;
- Adding `"message"` and `"asctime"` which are injected into `LogRecord` via `Formatter` and also cause clashes;
- Adding `"taskName"` for forwards compatibility with Python 3.12;

The list can be also be auto-generated via:
```python
_rec = logging.makeLogRecord({})
logging.Formatter("%(asctime)s:%(levelname)s:%(name)s:%(message)s").format(_rec)
throwaways = frozenset(rec.__dict__.keys())
```

### DIFF
--- a/src/sentry/logging/handlers.py
+++ b/src/sentry/logging/handlers.py
@@ -9,26 +9,35 @@ from sentry.utils import json, metrics
 
 # These are values that come default from logging.LogRecord.
 # They are defined here:
-# https://github.com/python/cpython/blob/2.7/Lib/logging/__init__.py#L237-L310
+# https://github.com/python/cpython/blob/e6543daf12051e9c660a5c0437683e8d2706a3c7/Lib/logging/__init__.py#L286-L385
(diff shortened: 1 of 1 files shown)


vartec:
@lobsterkatie FYI, regarding renaming rather than dropping, I'll make that a separate PR

lobsterkatie (approved):
Looks good!

How would you feel about alphabetizing this list? There may never be occasion for someone to be checking against it in the future, but when I was investigating this problem earlier I kept having to scan the list to find if something was in there and wished it were in order.

vartec:
This actually needs fixin' in the SDK, not here.

PR #73537, opened 2024-06-28
feat(replay): support tag LIKE queries for replay typeahead

### LINKED ISSUE #60025: Replay Search: Typeahead doesn't filter down matching values as you type
Searching on Replays (e.g: https://sentry.sentry.io/replays/) 

image

The spinner comes up when you type, but values are not filtered. When compared to for example issues (https://sentry.sentry.io/issues/):

image

aliu39:
Fixes https://github.com/getsentry/sentry/issues/60025. The relevant tag endpoint is OrganizationTagKeyValuesEndpoint.
Note since it's built on SQL's LIKE function, the filter is CASE SENSITIVE. This is the current behavior for issues, too.


src/sentry/replays/query.py
```diff
@@ -184,12 +184,23 @@ def query_replays_dataset_tagkey_values(
...
+        Condition(Column("project_id"), Op.IN, project_ids),
+        Condition(Column("timestamp"), Op.LT, end),
+        Condition(Column("timestamp"), Op.GTE, start),
+        Or(
+            [
+                Condition(Column("is_archived"), Op.EQ, 0),
+                Condition(Column("is_archived"), Op.IS_NULL),
+            ]
+        ),
+    ]
```
cmanallen, line 203: Its fine to move this here but it was in the place it was to keep it local to the execution site.  Less scrolling.  You don't have to change this I'm just giving the reasoning.  I defer to your judgement.

cmanallen (changes requested):
Needs test coverage.

src/sentry/replays/query.py
aliu39: Got it, thanks for explaining. I moved it up pretty arbitrarily when coding, don't mind moving it back

cmanallen (approved):

PR #73543, opened 2024-06-28
feat(custom-views): Add styles to draggable tabs 

MichaelSun48:
The purpose of this PR is to style the draggable tabs component to match the mockups, and to add some basic functionality to make it easier to connect the backend endpoints in the near future. In addition to changes made to the draggable tabs component, the a new prop, `isNewVariant` has been added to the existing `<Tabs>` component, which controls whether to use the new variant's styles or the existing styles. 

This also implements some basic functionality such as:
* Deleting tabs 
* `+ Add View` button
* Temporary tabs 

Visual things that are still incomplete:
* Overflowing tabs is not working 
* Temporary tabs still need some work 
* Tab content wraps if the screen is too narrow, which looks janky


static/app/components/tabs/tab.tsx
```diff
@@ -82,23 +82,15 @@ export const BaseTab = forwardRef(
...
-            orientation={orientation}
-            higherOpacity={isSelected}
-          />
-          <FocusLayer orientation={orientation} />
-          {props.children}
-          <TabSelectionIndicator orientation={orientation} selected={isSelected} />
-        </InnerWrap>
-      </TabWrap>
+        {props.children}
+      </NewTabWrap>
```
MichaelSun48, line 93: For sake of getting feedback sooner rather than later, I have edited the BaseTab component directly, but I will decouple the style changes I made here just for draggable tabs from the current BaseTab styles before this PR merges.

static/app/components/assigneeSelectorDropdown.spec.tsx
```diff
@@ -501,6 +501,29 @@ describe('AssigneeSelectorDropdown', () => {
     expect(screen.getByTestId('assignee-selector')).toHaveTextContent('CD');
   });
 
+  it('filters users based on their email addres', async () => {
```
roggenkemper, line 504:
```suggestion
  it('filters users based on their email address', async () => {
```

static/app/components/tabs/tab.tsx
```diff
@@ -54,13 +54,24 @@ interface BaseTabProps {
    * by <DraggableTab> to pass in props used for drag-and-drop functionality.
    */
   additionalProps?: React.HTMLAttributes<HTMLElement>;
+  isTempTab?: boolean;
```
malwilley, line 57: We shouldn't be adding anything specific to our use case in this Tab component

static/app/components/tabs/tab.tsx
```diff
@@ -54,13 +54,24 @@ interface BaseTabProps {
...
+  newVariant?: boolean;
```
malwilley, line 58: 1. We probably want a better name here
2. Modifying the Tabs component to support he new variant is something that we can break out and make a separate PR to master for

static/app/components/tabs/tab.tsx
```diff
@@ -138,6 +160,44 @@ export const Tab = forwardRef(
   }
 );
 
+const NewTabWrap = styled('li', {shouldForwardProp: tabsShouldForwardProp})<{
+  isTempTab: boolean;
+  overflowing: boolean;
+  selected: boolean;
+}>`
+  ${p =>
+    p.selected
```
malwilley, line 169: Instead of using the prop, it is better to use attributes when applicable since it does not require styles to be generated dynamically:

```css
&[aria-selected="true"] {
  /* selected styles */
}
```

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -134,27 +142,70 @@ function BaseDraggableTabList({
...
+                  <TabDivider />
+                )}
+            </Reorder.Item>
+          ))}
+          <AddViewButton borderless size="zero" onClick={() => setIsTempTabVisible(true)}>
+            <IconAdd size="xs" style={{margin: '2 4 2 2'}} />
+            Add View
+          </AddViewButton>
+          {isTempTabVisible && <TabDivider />}
+          {isTempTabVisible && tempTab && (
```
malwilley, line 193: I think it makes more sense to render the "Add view" button and temp tab _outside_ of this component. They are separate from the list tabs and it will be easier to make this component more generic that way.

I also think if you are able to remove the implementation-specific details here, we should be ready to merge this component into master.

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -134,27 +142,70 @@ function BaseDraggableTabList({
...
+                isChanged
+              />
+              {state.selectedKey !== item.key &&
+                state.collection.getKeyAfter(item.key) !== state.selectedKey && (
+                  <TabDivider />
+                )}
+            </Reorder.Item>
+          ))}
+          <AddViewButton borderless size="zero" onClick={() => setIsTempTabVisible(true)}>
+            <IconAdd size="xs" style={{margin: '2 4 2 2'}} />
```
malwilley, line 189: I this margin necessary? The button styles should have centered it correctly I'd . If they _are_ necessary it should be in a styled component.

PR #73547, opened 2024-06-28
feat(group-attributes): Add priority and first_release to the GroupAttributes snapshot 

snigdhas:
Add the priority and first_release fields to the snapshots when populating GroupAttributes. These fields will be used to power the snuba-first search. 

We should start writing these fields for groups before we run a backfill for the columns.


vartec (approved):

src/sentry/issues/attributes.py
```diff
@@ -192,6 +204,8 @@ def _bulk_retrieve_snapshot_values(
             "group_id": group_value.id,
             "status": group_value.status,
             "substatus": group_value.substatus,
+            "priority": group_value.priority,
+            "first_release": group_value.first_release.id if group_value.first_release else None,
```
wedamija, line 208: I'd recommend just using `group_value.first_release_id`, and just storing the `first_release_id` in the `GroupValues` structure too

wedamija (approved):

src/sentry/issues/attributes.py
```diff
@@ -192,6 +203,8 @@ def _bulk_retrieve_snapshot_values(
             "group_id": group_value.id,
             "status": group_value.status,
             "substatus": group_value.substatus,
+            "priority": group_value.priority,
+            "first_release": group_value.first_release,
```
wedamija, line 207: This should be `first_release_id` so that when we have a `Group` it works correctly,  (i think a group is passed here sometimes?). So I guess we need to rename `GroupValues.first_release`?

if it's a huge pain then feel free to just skip this review though

snigdhas: ahh that makes sense, updated!

PR #73550, opened 2024-06-29
ref(feedback): raise specific exception for vertex errors and don't log success

aliu39:
The current code raises KeyError when a request to VertexAI fails. https://console.cloud.google.com/logs/query;cursorTimestamp=2024-06-28T13:19:08.894346671Z;duration=P1D;query=resource.type%3D%22k8s_container%22%0Alabels.name%3D~%22sentry.feedback.usecases.create%22%0Atimestamp%3D%222024-06-28T13:19:08.894346671Z%22%0AinsertId%3D%22ht87o5dzkxma0imr%22;summaryFields=:true:32:beginning?project=internal-sentry

Also moved the spam_detection counter to create_feedback_issue, so we can see with one metric the T + F + failed (None) decisions.

### DIFF
--- a/src/sentry/feedback/usecases/create_feedback.py
+++ b/src/sentry/feedback/usecases/create_feedback.py
@@ -198,7 +198,11 @@ def create_feedback_issue(event, project_id: int, source: FeedbackCreationSource
         except Exception:
             # until we have LLM error types ironed out, just catch all exceptions
             logger.exception("Error checking if message is spam")
-            metrics.incr("feedback.create_feedback_issue.spam_check_failed")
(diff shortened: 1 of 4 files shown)


JoshFerge:
is there not a sentry error for this? what problems does this cause?

JoshFerge (approved):

aliu39:
> is there not a sentry error for this? what problems does this cause?

Wdym by sentry error?
Not a huge problem but the KeyError happens because we don't raise anything when the vertex request fails, so thought this would make things more clear

PR #73551, opened 2024-06-29
log(feedback): log user reports filtered from the denylist

aliu39:
relates to https://github.com/getsentry/team-replay/issues/436. Helps us see which orgs are filtered and when.

Anything wrong with having "too many"(?) extras, or too many logs? 😄


michellewzhang (approved):

michellewzhang:
looks like the denylist function check was imported elsewhere, might need to bring it back

src/sentry/ingest/userreport.py
```diff
@@ -33,6 +33,16 @@ def save_userreport(
...
         if is_org_in_denylist(project.organization):
+            metrics.incr("user_report.create_user_report.filtered", tags={"reason": "org.denylist"})
+            logger.info(
+                "Filtered an org in the deny list",
+                extra={
+                    "organization_id": project.organization.id,
+                    "organization_slug": project.organization.slug,
+                    "project_id": project.id,
+                    "project_slug": project.slug,
+                },
```
bruno-garcia, line 44: Could we get away with the metric only for now? I wonder if this would cost a ton as we can get millions of dropped data from spammy automated reports (e.g: unreal engine)

PR #73557, opened 2024-07-01
Use ``transaction`` for subtitle instead of deprecated ``culprit``

DanielNoord:
After digging into the code a bit I figured out this is why our Sentry project was no longer showing the correct subtitle after we switched to doing tracing through OpenTelemetry instead of using the Sentry based version.
We use the Python SDK which probably still sets the `culprit` key "correctly" whenever we are tracing. However, since there is no way to set this using the SDK we now have a lot of issues with incorrect subtitles. This makes it much harder to triage the issues.

As can be seen in the [documentation](https://develop.sentry.dev/sdk/event-payloads/types/#event) this key is deprecated and should probably not be used anymore.

### Legal Boilerplate

### DIFF
--- a/static/app/utils/events.tsx
+++ b/static/app/utils/events.tsx
@@ -42,7 +42,7 @@ export function getMessage(
     return event.culprit || '';
   }
 
-  const {metadata, type, culprit} = event;
+  const {metadata, type, transaction} = event;
 
   switch (type) {
     case EventOrGroupType.ERROR:
@@ -57,7 +57,7 @@ export function getMessage(
     case EventOrGroupType.GENERIC:
       return metadata.value;
     default:
-      return culprit || '';
+      return transaction || '';
   }
 }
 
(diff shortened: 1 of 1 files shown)


static/app/utils/events.tsx
```diff
@@ -42,7 +42,7 @@ export function getMessage(
     return event.culprit || '';
```
DanielNoord, line 42: I left because I have no clue what a Tombstone is but it sounds legacy and this key is legacy so I thought I'd leave it.

DanielNoord:
Created https://github.com/getsentry/sentry/issues/73558 as this seems like a bigger issue than I can solve on my own. Would appreciate if somebody would pick this up or help me fix it.

armenzg:
This file is using `culprit` from the `Group` model:
https://github.com/getsentry/sentry/blob/ac83b15608a785ac60ae46231da83c8dee5bfe00/src/sentry/models/group.py#L532-L534

In my PR I am using `culprit` from an event which is a [generated value](https://github.com/getsentry/sentry/blob/23f2b42028b197415b1fc55d1f233c1e5414c89f/src/sentry/culprit.py#L15-L43).

Ideally, we would not use the same field name to prevent confusion.

PR #73559, opened 2024-07-01
Clean up inline-attachments options

Swatinem:
Usage of the option downstream in relay was removed in https://github.com/getsentry/relay/pull/3711

Inlining small attachments in `EventAttachment` was also turned on for quite a while, so lets hardcode it now and remove the option later.


Dav1dde (commented):
The option `relay.inline-attachments.rollout-rate` still exists: https://github.com/getsentry/sentry/blob/0c4fbc1e46f9cc0b7c387301cd3684f642f7cae0/src/sentry/options/defaults.py#L1040-L1041

But I am not sure what the policy is on removing options.

Swatinem:
> The option still exists

Yes, all that is a 3-part process: https://github.com/getsentry/sentry-options-automator#deleting-an-option

loewenheim (approved):

Issue #73561, opened 2024-07-01
limit number of specs emitted by span attribute extraction rules


matejminar:
For now 100 - make it configurable option.

```[tasklist]
- [ ] https://github.com/getsentry/sentry/pull/73566
- [ ] https://github.com/getsentry/sentry/pull/73627
```

PR #73566, opened 2024-07-01
feat(metrics): add span attribute spec limit

obostjancic:
- Part of #73561

### DIFF
--- a/src/sentry/relay/config/metric_extraction.py
+++ b/src/sentry/relay/config/metric_extraction.py
@@ -4,7 +4,7 @@ from collections import defaultdict
 from collections.abc import Sequence
 from dataclasses import dataclass
 from datetime import timedelta
-from typing import Any, NotRequired, TypedDict, cast
+from typing import Any, Literal, NotRequired, TypedDict, cast
 
 import sentry_sdk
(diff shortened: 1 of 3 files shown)


src/sentry/relay/config/metric_extraction.py
```diff
@@ -841,6 +841,9 @@ def _generate_span_attribute_specs(project: Project) -> list[HashedMetricSpec]:
         except ValueError:
             logger.exception("Invalid span attribute metric spec", extra=rule.to_dict())
 
+    max_specs = options.get("metric_extraction.max_span_attribute_specs")
+    (specs, _) = _trim_if_above_limit(specs, max_specs, project, "span_attributes")
```
ArthurKnaus, line 845: Not for this PR, but we should log that somewhere so we notice if the limit was reached.

obostjancic: `_trim_if_above_limit` already does that for us with a [capture_exception](https://github.com/getsentry/sentry/blob/ogi/feat/span-attribute-spec-limit/src/sentry/relay/config/metric_extraction.py#L376)

ArthurKnaus (approved):

PR #73570, opened 2024-07-01
Do not allow crash rate value to be greater than 1

### LINKED ISSUE #73563: Do not allow crash_rate value greater than 1
Connected to https://github.com/getsentry/sentry/issues/73172 we should not allow `crash_rate` value to be greater than one.
This can happen when the data is corrupted

vgrozdanic:
Sometimes due to the problems in other parts of the system, data gets
corrupted and `crash_rate` is returned as a value greater than 1 (which
should never happen)

This solution guards against corrupted data by returning 1 if the
calculated `crash_rate` is greater than 0

closes [#73563](https://github.com/getsentry/sentry/issues/73563)

### DIFF
--- a/src/sentry/api/endpoints/organization_release_health_data.py
+++ b/src/sentry/api/endpoints/organization_release_health_data.py
@@ -10,6 +10,7 @@ from sentry.api.paginator import GenericOffsetPaginator
 from sentry.exceptions import InvalidParams
 from sentry.sentry_metrics.use_case_utils import get_use_case_id
 from sentry.snuba.metrics import DerivedMetricException, QueryDefinition, get_series
+from sentry.snuba.metrics.naming_layer import SessionMetricKey
 from sentry.types.ratelimit import RateLimit, RateLimitCategory
 from sentry.utils import metrics
 from sentry.utils.cursors import Cursor, CursorResult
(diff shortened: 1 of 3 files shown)


src/sentry/api/endpoints/organization_release_health_data.py
```diff
@@ -63,6 +63,10 @@ def data_fn(offset: int, limit: int):
                 # due to possible data corruption crash_free_rate value can be less than 0,
                 # which is not valid behavior, so those values have to be bottom capped at 0
                 metrics.ensure_non_negative_crash_free_rate_value(data, request, organization)
+                # due to possible data corruption crash_rate value can be greater than 1,
+                # which is not valid behavior, so those values have to be capped at 1
+                metrics.ensure_crash_rate_not_greater_than_1(data, request, organization)
```
obostjancic, line 68: can we do something like `metrics.ensure_crash_rate_in_bounds` that performs both checks so that we do not iterate twice over the response?

vgrozdanic: There is no much overhead in iterating through the response few times (usually number of groups in that API endpoint is 1 or 2, rarely over that), and every other iteration is key specific, which means it won't happen unless a metric key for which we do checks is present.

I could combine all of this condition into one function which checks everything, but it would make the code much more complex and I would prefer approach with simpler and cleaner code, but slight overhead of few iterations more.

I have refactored this code to use only one function for all crash rate bound checks, so it is easier to maintain and refactor later.

obostjancic: Yeah, agreed, this is great. Thanks!

obostjancic (approved):

PR #73293, opened 2024-06-25
chore(api): Rewrite Org Details docs

isabellaenriquez:
Updates the OrganizationDetailsEndpoint docs to use the modern format + more detailed parameters + updated examples and schema 

localhost_3000_api_organizations_update-an-organization_ (5)

localhost_3000_api_organizations_update-an-organization_ (4)

PR #73593, opened 2024-07-01
feat(api): In org details, show if SSO is required

RyanSkonnord:
Supports https://github.com/getsentry/sentry/pull/73521

### DIFF
--- a/src/sentry/api/serializers/models/organization.py
+++ b/src/sentry/api/serializers/models/organization.py
@@ -445,6 +445,7 @@ class DetailedOrganizationSerializerResponse(_DetailedOrganizationSerializerResp
     metricsActivatePercentiles: bool
     metricsActivateLastForGauges: bool
     extrapolateMetrics: bool
+    requiresSso: bool
 
 
 class DetailedOrganizationSerializer(OrganizationSerializer):
@@ -598,6 +599,7 @@ class DetailedOrganizationSerializer(OrganizationSerializer):
         if access.role is not None:
             context["role"] = access.role  # Deprecated
             context["orgRole"] = access.role
+        context["requiresSso"] = access.requires_sso
         context["pendingAccessRequests"] = OrganizationAccessRequest.objects.filter(
             team__organization=obj
         ).count()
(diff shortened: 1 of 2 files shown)


cathteng (approved):

iamrajjoshi (approved):
I am not sure if we need to update our API docs for this change.

src/sentry/api/serializers/models/organization.py
```diff
@@ -598,6 +599,7 @@ def serialize(  # type: ignore[explicit-override, override]
         if access.role is not None:
             context["role"] = access.role  # Deprecated
             context["orgRole"] = access.role
+        context["requiresSso"] = access.requires_sso
```
evanpurkhiser, line 602: Is there another way we can make this check without having to update the org serializer? I'm sure we must know somewhere on the frontend if the org requires SSO no?

evanpurkhiser: Chatted with @RyanSkonnord about this. We really don't expose this anywhere to the UI for the most part. I think adding this here is fine (despite it being just another field on an organization).

evanpurkhiser (approved):

src/sentry/api/serializers/models/organization.py
RyanSkonnord: Some notes on my investigation, for the record:

The source of truth for `access.requires_sso` is `query_sso_state` at https://github.com/getsentry/sentry/blob/master/src/sentry/auth/services/access/service.py#L60, which inspects the associated `AuthProvider` object. The flag is generally available through an `Access` object and isn’t associated directly with the org model. Given that the org serializer is already adding few other `Access` attributes to the API output (right above here), there seems to be precedent. However, I did have the same initial intuition as @evanpurkhiser — it's surprising that it wasn’t already visible somehow.

RyanSkonnord:
> I am not sure if we need to update our API docs for this change.

According to @isabellaenriquez an update would be overwritten by https://github.com/getsentry/sentry/pull/73293 anyway, so we can add it afterward.

PR #73600, opened 2024-07-01
docs(useQuery): Write a story that visually shows useQuery loading/error/success states

ryan953:
There's always been some slight confusion, and lack of clear examples, when it comes to using `isFetching` and `isLoading` from react-query. Also we're not super consistent on the names when we create our own loading state variables, which just adds to the confusion...

So I made a story that illustrates a few examples where we "load" in data slowly enough to see the state changes happening. Turns out that checking both `isFetching` and `isLoading` is probably what we should do all the time. `<Suspense>` is also a killer feature to leverage.

SCR-20240701-lspr

### DIFF
--- a/static/app/utils/api/useQuery.stories.tsx
+++ b/static/app/utils/api/useQuery.stories.tsx
@@ -282,11 +282,11 @@ export default storyBook('useQuery', story => {
 
   story('suspense: true, useErrorBoundary: false', () => {
     return (
-      <ToggleMounted queryKeys={[['story-suspense-success'], ['story-suspense-error']]}>
+      <ToggleMounted queryKeys={[['story-suspense1-success'], ['story-suspense1-error']]}>
         <Suspense fallback={<LoadingFallback />}>
           <DataContainer
             options={{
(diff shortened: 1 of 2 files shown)


static/app/utils/api/useQuery.stories.tsx
```diff
@@ -0,0 +1,343 @@
...
+        You should prefer to import from <code>sentry/utils/queryClient</code> as much as
+        possible to benefit from easier and consistent data fetching within sentry.
+      </p>
+    </Fragment>
+  ));
+  story('TL/DR', () => (
+    <Fragment>
+      <p>
+        It seems like you can get really far by checking BOTH{' '}
+        <code>isFetching || isLoading</code> when you are waiting for the first render,
```
malwilley, line 151: I not sure we should encourage checking for `isFetching || isLoading`, since you'll lose the benefits of background refetches if you do so

ryan953: i'll make more examples!

ryan953:
Relates to https://github.com/getsentry/sentry/pull/76407

PR #73601, opened 2024-07-01
ref(slack): retry link slack user identities task

cathteng:
Add retries for the Slack link user identities task, and bubble up any `SlackApiErrors` so we can retry if that happens rather than exit and potentially never link certain users.

This is for the INFRA-1 task for Slack messaging refactor.

### DIFF
--- a/src/sentry/tasks/integrations/slack/link_slack_user_identities.py
+++ b/src/sentry/tasks/integrations/slack/link_slack_user_identities.py
@@ -22,16 +22,20 @@ logger = logging.getLogger("sentry.integrations.slack.tasks")
     name="sentry.integrations.slack.link_users_identities",
     queue="integrations.control",
     silo_mode=SiloMode.CONTROL,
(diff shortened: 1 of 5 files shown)


src/sentry/tasks/integrations/slack/link_slack_user_identities.py
```diff
@@ -22,16 +22,20 @@
...
-    integration_id: int | None = None,
-    organization_id: int | None = None,
+    integration_id: int,
+    organization_id: int,
 ) -> None:
-    if integration_id is not None:
-        integration = integration_service.get_integration(integration_id=integration_id)
-    if organization_id is not None:
-        organization = organization_service.get_organization_by_id(id=organization_id).organization
-    assert organization is not None and integration is not None
```
RyanSkonnord, line 34: Good change; this didn't make sense in the first place.

RyanSkonnord (approved):

PR #73607, opened 2024-07-01
ref(replay): Refactor text extraction inside replay diff toolset

ryan953:
The text extraction stuff was a big janky, it relied on rendering the VISUAL tab so we could peek at the inside of the `<ReplayPlayer>` instances. 

Now we just setup some hidden instances and read from those. In a followup we'll be able to do that on demand instead of having this dependency between the tabs.

### DIFF
--- a/static/app/utils/replays/hooks/useExtractedPageHtml.tsx
+++ b/static/app/utils/replays/hooks/useExtractedPageHtml.tsx
@@ -1,0 +1,21 @@
+import {useQuery} from 'sentry/utils/queryClient';
+import extractPageHtml from 'sentry/utils/replays/extractPageHtml';
+import type ReplayReader from 'sentry/utils/replays/replayReader';
+
+interface Props {
+  offsetMsToStopAt: number[];
+  replay: ReplayReader | null;
+}
+
+export default function useExtractedPageHtml({replay, offsetMsToStopAt}: Props) {
+  return useQuery(
+    ['extactPageHtml', replay],
+    () =>
+      extractPageHtml({
+        offsetMsToStopAt,
+        rrwebEvents: replay?.getRRWebFrames(),
+        startTimestampMs: replay?.getReplay().started_at.getTime() ?? 0,
+      }),
+    {enabled: Boolean(replay), cacheTime: Infinity}
+  );
+}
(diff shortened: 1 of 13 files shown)


static/app/utils/replays/hooks/useExtractedPageHtml.tsx
```diff
@@ -0,0 +1,21 @@
...
+import extractPageHtml from 'sentry/utils/replays/extractPageHtml';
+import type ReplayReader from 'sentry/utils/replays/replayReader';
+
+interface Props {
+  offsetMsToStopAt: number[];
+  replay: ReplayReader | null;
+}
+
+export default function useExtractedPageHtml({replay, offsetMsToStopAt}: Props) {
+  return useQuery(
```
scttcper, line 11: why useQuery instead of useMemo

ryan953: useQuery will await the promise. It's a pattern i used before here: https://github.com/getsentry/sentry/blob/02f53d9b3a8251a0e108fd91104f529816e21a98/static/app/utils/replays/hooks/useExtractedDomNodes.tsx#L5-L17

scttcper: didn't even see the promise, that makes sense

scttcper (approved):

static/app/utils/replays/hooks/useExtractedPageHtml.tsx
```diff
@@ -0,0 +1,21 @@
...
+    ['extactPageHtml', replay],
```
scttcper, line 12: not sure what is hashing the replay to compare it later but it does look like it might access every property on the object? https://tanstack.com/query/latest/docs/framework/react/guides/query-keys#query-keys-are-hashed-deterministically

ryan953: We've implemented `toJSON` to avoid that specific problem! It was indeed crazy without that

https://github.com/getsentry/sentry/blob/d72428194f97d82e2f6d668673175213b0e6e5f2/static/app/utils/replays/replayReader.tsx#L383

PR #73613, opened 2024-07-01
ref(replay): Extract useTimelineScale from useReplayContext and tweak ui styles

ryan953:
The UI tweaks come from this older figma: https://www.figma.com/design/tlGOyIijfu309du7zXTDTM/Library%3A-Replay?node-id=0-1&t=iRFVZSPdf7cuccpS-0

Changes:
- Down Triangle for both sides
- Triangle is a little higher, so it's not cutoff by the scrubber circle when they point to the same spot (see image below)
- Unicode `x`
- font size/weight/color changes for the current/total time labels

**After:**
SCR-20240701-oerr


Jesse-Box:
@ryan953 is it a pain to change the multiplier label to be the same size as the current time/total time label?

ryan953:
> @ryan953 is it a pain to change the multiplier label to be the same size as the current time/total time label?

the `4x` to be smaller? nbd

c298lee (approved):

PR #73618, opened 2024-07-02
Add ability to get options if project has a template

saponifi3d:
# Descriptions
Add a `template_manager` which mimics the `option_manager`, except with templated values. When using `get_value` from a project, this will now check for a template association. if there is a template and there isn't an option set, check to see if the template has that option.

[Tech spec](https://www.notion.so/sentry/Service-Settings-Templates-ca6fdb38b15f4a6db61d319bc342b86c?pvs=4#2f0d1c466f684c9aa735224e854ecf1d)

### DIFF
--- a/src/sentry/models/options/project_option.py
+++ b/src/sentry/models/options/project_option.py
@@ -152,6 +152,9 @@ class ProjectOptionManager(OptionManager["ProjectOption"]):
     def post_delete(self, instance: ProjectOption, **kwargs: Any) -> None:
         self.reload_cache(instance.project_id, "projectoption.post_delete")
 
+    def isset(self, project: Project, key: str) -> bool:
+        return self.get_value(project, key, default=Ellipsis) is not Ellipsis
+
 
 @region_silo_model
 class ProjectOption(Model):
--- a/src/sentry/models/project.py
+++ b/src/sentry/models/project.py
@@ -418,9 +418,9 @@ class Project(Model, PendingDeletionMixin):
         self, key: str, default: Any | None = None, validate: Callable[[object], bool] | None = None
     ) -> Any:
         # don't pass the default, we need to check the template if it doesn't exist
-        option = self.option_manager.get_value(self, key, None, validate)
+        option = self.option_manager.get_value(self, key, default, validate)
 
-        if option is None and self.template is not None:
+        if option == default and self.template is not None:
             return self.template_manager.get_value(self.template, key, default, validate)
 
         # check if the option is set, if not return the default
--- a/src/sentry/models/project.py
+++ b/src/sentry/models/project.py
@@ -50,6 +50,7 @@ from sentry.utils.snowflake import save_with_snowflake_id, snowflake_id_model
 
 if TYPE_CHECKING:
     from sentry.models.options.project_option import ProjectOptionManager
+    from sentry.models.options.project_template_option import ProjectTemplateOptionManager
     from sentry.models.user import User
 
 SENTRY_USE_SNOWFLAKE = getattr(settings, "SENTRY_USE_SNOWFLAKE", False)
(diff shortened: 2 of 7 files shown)


tests/sentry/models/test_project.py
```diff
@@ -407,6 +409,55 @@ def test_remove_team_clears_alerts(self):
         assert alert_rule.user_id is None
 
 
+class ProjectSettings(TestCase):
+    def setUp(self):
```
ceorourke, line 413: Could you add a docstring that just says something to the effect of "`ProjectOption` trumps `ProjecTemplateOption`? It was a little hard to understand at first when reading the tests.

tests/sentry/models/test_project.py
```diff
@@ -407,6 +409,55 @@ def test_remove_team_clears_alerts(self):
...
+        assert self.project.get_option(self.option_key) is None
+        ProjectOption.objects.set_value(self.project, self.option_key, True)
+        assert self.project.get_option(self.option_key) is True
+
+    def test_get_template_option(self):
+        assert self.project.get_option(self.option_key) is None
+        ProjectTemplateOption.objects.set_value(self.project_template, self.option_key, "test")
+        assert self.project.get_option(self.option_key) == "test"
+
+    def test_get_option__overide_template(self):
```
ceorourke, line 437: nittiest nit of all nits but are these test names meant to have double underscores?

ceorourke: Also override with 2 r's

ceorourke (approved):
lgtm but seems it's broken some other tests - not gonna block

src/sentry/models/project.py
```diff
@@ -407,10 +408,23 @@ def option_manager(self) -> ProjectOptionManager:
...
     ) -> Any:
-        return self.option_manager.get_value(self, key, default, validate)
+        # don't pass the default, we need to check the template if it doesn't exist
+        option = self.option_manager.get_value(self, key, None, validate)
+
+        if option is None and self.template is not None:
+            return self.template_manager.get_value(self.template, key, default, validate)
+
+        # check if the option is set, if not return the default
+        return default if option is None else option
```
ceorourke, line 427: Does this need to be `return self.option_manager.get_value(self, key, default, validate) if option is None else option`?

saponifi3d: yeah, that probably would have fixed the test 🤣 i ended up making it a littttle different and just stored the result in `option` then if it's the `default` value, check the template. so logically it works the same in this function, but it will correctly utilize the `default` / `vaildator` stuff (that your recommendation would have also fixed)

tests/sentry/models/test_project.py
saponifi3d: yeah - the double underscores are used to split apart the "name" of the test with the "context" of the test. so in this case i'm "testing" the "get_option" function and the context is that i want to "override the template".

idk - something i've done in the past in python to help organize all the tests, but happy to remove them if it's too confusing.

---

will fix the 2 r's - and curious about your opinion on ^ for organizing tests.

ceorourke: ah I see - that makes sense, I just haven't seen tests named like that before.

src/sentry/models/options/project_option.py
```diff
@@ -152,6 +152,9 @@ def post_save(self, instance: ProjectOption, **kwargs: Any) -> None:
     def post_delete(self, instance: ProjectOption, **kwargs: Any) -> None:
         self.reload_cache(instance.project_id, "projectoption.post_delete")
 
+    def isset(self, project: Project, key: str) -> bool:
```
ceorourke, line 155: could you make this `is_set`? or is isset something else?

ceorourke: also TIL about `Ellipsis`

saponifi3d: for `isset` it seems like there's already a pattern established: https://github.com/search?q=repo:getsentry/sentry%20isset&type=code - specifically around project options too. #duckTypingAndAllThat

ceorourke: weird, I don't like it but ok!

PR #73620, opened 2024-07-02
Emit Sentry warning and GCP log for sampled throttled queries

xurui-c:
Allocation policies are our mechanism for doing traffic management for Snuba queries. Currently, the result of applying multiple allocation policies in the internal API is simply accept/reject/throttle.

In a [previous PR](https://github.com/getsentry/sentry/pull/73442), we emit a Sentry warning and a GCP log for every query throttled by Snuba's capacity management system. However, we quickly realized [a large volume of queries were throttled](https://sentry.sentry.io/issues/5550501231/?project=1&query=&referrer=issue-stream&statsPeriod=24h&stream_index=5). 

To address this, we sample the throttled queries such that we will only emit a Sentry warning and a GCP log for every `SAMPLING_FREQUENCY`th throttled query for a given referrer. For example, if `SAMPLING_FREQUENCY = 100`, then we will emit the warning and log for every 100th throttled query for the same referrer.

### DIFF
--- a/src/sentry/utils/snuba.py
+++ b/src/sentry/utils/snuba.py
@@ -228,6 +228,8 @@ METRICS_COLUMN_MAP = {
     "organization_id": "org_id",
 }
 
+REFERRER_THROTTLED_QUERIES_COUNT_MAP: dict[str, int] = {}
+SAMPLING_FREQUENCY = 100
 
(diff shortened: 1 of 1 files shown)


xurui-c:
close in favor of https://github.com/getsentry/sentry/pull/73709

PR #73624, opened 2024-07-02
feat(metrics): persist span attribute extraction rules in django (v2)

shellmayr:
Continued after revert of https://github.com/getsentry/sentry/pull/73421

Entries in auth_permission with a length of over 50, causing the migration to fail due to the length of the class name. This can be fixed by adding a verbose_name field to the Meta.


obostjancic (approved):

vgrozdanic (approved):
Let's goooo

PR #73626, opened 2024-07-02
feat(metrics-extraction): Introduce virtual metrics

ArthurKnaus:
Add the metrics type `v`
Add a new hook `useVirtualizedMetricsMeta` to fetch metric meta and map span based metrics to virtual ones.
Add a stub implementation of the virtual metrics context.

### Background

Introduce virtual metrics as a concept to hide the internal complexity of span-based metrics from the user.
The additional metric type `v` (virtual) can be used to distinguish them from other metrics. Via a context provider (still TODO) we will be able to inject callbacks to map between virtual metrics and span-based metrics wherever necessary.

Part of https://github.com/getsentry/sentry/issues/73418

### DIFF
--- a/static/app/utils/metrics/useMetricsMeta.tsx
+++ b/static/app/utils/metrics/useMetricsMeta.tsx
@@ -1,7 +1,9 @@
 import {useMemo} from 'react';
 
 import type {PageFilters} from 'sentry/types/core';
+import {isExtractedCustomMetric} from 'sentry/utils/metrics';
 import {formatMRI, getUseCaseFromMRI} from 'sentry/utils/metrics/mri';
+import {useVirtualMetricsContext} from 'sentry/utils/metrics/virtualMetricsContext';
 import type {ApiQueryKey} from 'sentry/utils/queryClient';
 import {useApiQuery} from 'sentry/utils/queryClient';
 import useOrganization from 'sentry/utils/useOrganization';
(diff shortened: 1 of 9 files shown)


static/app/utils/metrics/useMetricsMeta.tsx
```diff
@@ -58,6 +61,65 @@ export function useMetricsMeta(
...
+    const otherMetrics = data.filter(meta => {
+      if (!isExtractedCustomMetric(meta)) {
+        return true;
+      }
+
+      const virtualMRI = getVirtualMRI(meta.mri);
+      // If there is no virtual MRI, we don't want to show this metric
+      if (!virtualMRI) {
+        return false;
+      }
```
ArthurKnaus, line 103: As the stub implementation of `getVirtualMRI` always returns `null`, this will have the effect of hiding internal extracted metrics from the UI wherever it is used.

static/app/utils/metrics/virtualMetricsContext.tsx
```diff
@@ -0,0 +1,21 @@
...
+    throw new Error('Not implemented');
+  },
+  isLoading: false,
+});
+
+export function useVirtualMetricsContext() {
+  return useContext(Context);
+}
+
+// TODO(aknaus): Write a provider
```
ArthurKnaus, line 21: Will happen in a follow up PR.

obostjancic (approved):

PR #73627, opened 2024-07-02
feat(metrics): limit number of specs during creation

obostjancic:
- Part of: #73561
- Checks the number of specs that are going to be generated when creating/updating a rule
- swaps sentry_sdk for a logger


vgrozdanic (approved):

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -104,22 +108,27 @@ def post(self, request: Request, project: Project) -> Response:
...
+            return Response(data=persisted_config, status=200)
 
-        persisted_config = serialize(
-            configs,
-            request.user,
-            SpanAttributeExtractionRuleConfigSerializer(),
-        )
-        return Response(data=persisted_config, status=200)
+        except MetricsExtractionRuleValidationError as e:
+            logger.warning("Failed to update extraction rule", exc_info=True)
```
vgrozdanic, line 126: Is there a reason why are we logging this?
If we want to count how many times it is happening, is there a reason why aren't we using Sentry for this?

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -104,22 +108,27 @@ def post(self, request: Request, project: Project) -> Response:
...
+            return Response(status=400, data={"detail": str(e)})
+
+        except Exception:
+            logger.exception("Failed to update extraction rule")
```
vgrozdanic, line 130: This is nit, but if some general Exception happens that we are not aware of, I would also log the exception (maybe even in Sentry) so that we can act on it if needed

PR #73633, opened 2024-07-02
fix(metrics) do not allow generic metric without operation

### LINKED ISSUE #73364: AssertionError
Sentry Issue: [SENTRY-3AVD](https://sentry.sentry.io/issues/5506836919/?referrer=github_integration)

```
AssertionError: 
(6 additional frame(s) were not displayed)
...
  File "sentry/api/endpoints/organization_release_health_data.py", line 82, in get_result
    data = self.data_fn(offset=offset, limit=limit + 1)
  File "sentry/api/endpoints/organization_release_health_data.py", line 56, in data_fn
    data = get_series(
  File "sentry/snuba/metrics/datasource.py", line 1071, in get_series
    ).get_snuba_queries()
  File "sentry/snuba/metrics/query_builder.py", line 1121, in get_snuba_queries
    metric_field_obj = metric_object_factory(field.op, field.metric_mri)
  File "sentry/snuba/metrics/fields/base.py", line 1939, in metric_object_factory
    assert op is not None
```

vgrozdanic:
Do not allow to query for a generic metric without operation.
For example:
    - querying for `sentry.sessions.session` is not valid because operation is missing
    - querying for `sum(sentry.sessions.session)` is valid because operation `sum` exists

Closes https://github.com/getsentry/sentry/issues/73364

### DIFF
--- a/tests/sentry/api/endpoints/test_organization_release_health_data.py
+++ b/tests/sentry/api/endpoints/test_organization_release_health_data.py
@@ -18,6 +18,7 @@ from sentry.snuba.metrics.naming_layer.public import (
     TransactionStatusTagValue,
     TransactionTagsKey,
 )
+from sentry.snuba.metrics.query import MetricField
 from sentry.testutils.cases import MetricsAPIBaseTestCase
 from sentry.testutils.helpers.datetime import freeze_time
 from sentry.utils.cursors import Cursor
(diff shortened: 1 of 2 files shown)


tests/sentry/api/endpoints/test_organization_release_health_data.py
```diff
@@ -1658,6 +1661,8 @@ def test_derived_metric_incorrectly_defined_as_singular_entity(
         mocked_get_mri_query.return_value = "crash_free_fake"
         mocked_reverse_mri.return_value = "crash_free_fake"
         mocked_parse_mri.return_value = ParsedMRI("e", "sessions", "crash_free_fake", "none")
+        mocked_parse_field.return_value = MetricField(None, "e:sessions/crashed_free_fake@none")
```
vgrozdanic, line 1664: This has to be mocked so that it passes newly added validation, but triggers error in other part of code, which was original intention. This metric is fully mocked and it does not exists, it was only created to test specific use case

obostjancic (approved):

Issue #73562, opened 2024-07-01
Implement pagination on metrics extraction rule table


matejminar:
Tasks:

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/sentry/pull/73648
- [ ] https://github.com/getsentry/sentry/pull/73640
```

PR #73648, opened 2024-07-02
feat(metrics) Add pagination to metrics extraction rule table

vgrozdanic:
Add pagination for metrics extraction rule table

Part of [#73562](https://github.com/getsentry/sentry/issues/73562)


vgrozdanic:
Closing because we won't implement it soon

PR #73657, opened 2024-07-02
feat(profiler_id): add profiler id to response

JonasBa:


mjq (approved):

wmak (changes requested):
Unfortunately this is the code for the non spans-based response which isn't in use anymore

PR #73659, opened 2024-07-02
chore(slack): Use Slack SDK Client in `validate_channel_id`

iamrajjoshi:
Update `validate_channel_id` to use to Slack SDK. Wrote Tests and Tested Locally. Addresses `SDK-2.22`


src/sentry/integrations/slack/utils/channel.py
```diff
@@ -97,14 +99,33 @@ def validate_channel_id(name: str, integration_id: int | None, input_channel_id:
...
-        if e.text == "channel_not_found":
-            raise ValidationError("Channel not found. Invalid ID provided.")
-        logger.info("rule.slack.conversation_info_failed", extra={"error": str(e)})
-        raise IntegrationError("Could not retrieve Slack channel information.")
+
+    logger_params = {
+        "integration_id": integration_id,
+        "channel_name": name,
+        "input_channel_id": input_channel_id,
+    }
```
RyanSkonnord, line 107: Nit: Are we trying to follow a convention of declaring `logger_params` as a local variable? I don't think it's necessary unless the same params are logged in multiple places.

src/sentry/integrations/slack/utils/channel.py
```diff
@@ -97,14 +99,33 @@ def validate_channel_id(name: str, integration_id: int | None, input_channel_id:
...
+            raise IntegrationError("Could not retrieve Slack channel information.")
+    else:
+        client = SlackClient(integration_id=integration_id)
+        try:
+            results = client.get("/conversations.info", params=payload)
+        except ApiError as e:
+            if e.text == "channel_not_found":
+                raise ValidationError("Channel not found. Invalid ID provided.")
+            _logger.info("rule.slack.conversation_info_failed", extra={"error": str(e)})
+            raise IntegrationError("Could not retrieve Slack channel information.")
```
RyanSkonnord, line 128: All four of these `raise` statements should have `from e` at the end.

RyanSkonnord (approved):

PR #73664, opened 2024-07-02
fix(queues): remove option for timing queued tasks

cathteng:
This is already default to `False` and the option has never been set. This also causes problems with `tests/sentry/tasks/test_options.py` because it's getting an option and polluting the calls to the mocked function.


cathteng:
@ykamo001

iamrajjoshi (approved):

nhsiehgit (approved):

src/sentry/tasks/base.py
```diff
@@ -98,16 +98,6 @@ def _wrapped(*args, **kwargs):
 
             # Use a try/catch here to contain the blast radius of an exception being unhandled through the options lib
             # Unhandled exception could cause all tasks to be effected and not work
```
asottile-sentry, line 100: the comment is orphaned now

PR #66847, opened 2024-03-13
ref(rules): Add changes to rule edit confirmation notification

ceorourke:
Follow up to https://github.com/getsentry/sentry/pull/66285 to add details about what changed happened when a rule was edited to the notification.

Screenshot 2024-03-12 at 4 40 24 PM

Note that I've made a couple follow up tickets to address poor rendering of issue type enums and frequency: https://github.com/getsentry/team-core-product-foundations/issues/158 and https://github.com/getsentry/team-core-product-foundations/issues/157

PR #73665, opened 2024-07-02
fix(rules): Fix environment fetch

ceorourke:
Copy/paste bug where the `.get()` was on the wrong variable for the environment and was always returning None so if an environment was present on the rule and then was removed, it wouldn't be added to the confirmation notification. See https://github.com/getsentry/sentry/pull/66847/files#r1662799957

### DIFF
--- a/src/sentry/rules/actions/utils.py
+++ b/src/sentry/rules/actions/utils.py
@@ -126,7 +126,7 @@ def get_changed_data(
     if rule_data_before.get("environment_id") and not rule_data.get("environment_id"):
         environment = None
         try:
-            environment = Environment.objects.get(id=rule_data_before.get("environment_id"))
+            environment = Environment.objects.get(id=rule_data_before["environment_id"])
         except Environment.DoesNotExist:
             pass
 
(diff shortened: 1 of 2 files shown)


src/sentry/rules/actions/utils.py
```diff
@@ -126,7 +126,7 @@ def get_changed_data(
     if rule_data_before.get("environment_id") and not rule_data.get("environment_id"):
         environment = None
         try:
-            environment = Environment.objects.get(id=rule_data.get("environment_id"))
+            environment = Environment.objects.get(id=rule_data_before.get("environment_id"))
```
asottile-sentry, line 129: these should be square brackets (same as above) -- we know it is present and non-None here already (part of why the type checker flagged it for me!)

src/sentry/rules/actions/utils.py
```diff
@@ -126,7 +126,7 @@ def get_changed_data(
     if rule_data_before.get("environment_id") and not rule_data.get("environment_id"):
         environment = None
         try:
-            environment = Environment.objects.get(id=rule_data.get("environment_id"))
+            environment = Environment.objects.get(id=rule_data_before["environment_id"])
```
asottile-sentry, line 129: can you fix the one on 119 too -- will save me a PR :D

asottile-sentry (approved):

saponifi3d (approved):
:shipit:

PR #73672, opened 2024-07-02
fix(rules): Use the right comparison interval mapping

schew2381:
`EventFrequencyPercentCondition` with a percent comparison would fail with a KeyError when trying to get the timedelta for the `comparison_interval`. This is b/c we were using the class variable `intervals`
https://github.com/getsentry/sentry/blob/3b3b250b5268263577ba9189f0b707d299b11d58/src/sentry/rules/processing/delayed_processing.py#L192 instead of the `COMPARISON_INTERVALS` which is used correctly when not bulk querying:
https://github.com/getsentry/sentry/blob/3b3b250b5268263577ba9189f0b707d299b11d58/src/sentry/rules/conditions/event_frequency.py#L176

`EventFrequencyPercentCondition` has different class `intervals` that's missing some keys from `COMPARISON_INTERVALS`, causing the `KeyError` to occur. For example, if `comparison_interval` was `"1d"` or `"1w"`.
https://github.com/getsentry/sentry/blob/3b3b250b5268263577ba9189f0b707d299b11d58/src/sentry/rules/conditions/event_frequency.py#L490-L496

Fixes SENTRY-3BB5
Fixes ALRT-123

### DIFF
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -15,6 +15,7 @@ from sentry.models.rule import Rule
 from sentry.models.rulesnooze import RuleSnooze
 from sentry.rules import history, rules
 from sentry.rules.conditions.event_frequency import (
+    COMPARISON_INTERVALS,
     DEFAULT_COMPARISON_INTERVAL,
     BaseEventFrequencyCondition,
     ComparisonType,
@@ -189,7 +190,7 @@ def get_condition_group_results(
         # The comparison interval is only set for the second query of a percent
         # comparison condition.
         comparison_interval = (
-            condition_inst.intervals[unique_condition.comparison_interval][1]
+            COMPARISON_INTERVALS[unique_condition.comparison_interval][1]
             if unique_condition.comparison_interval
             else None
         )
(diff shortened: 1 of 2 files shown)


nhsiehgit (approved):

tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -733,6 +733,40 @@ def test_apply_delayed_percent_comparison_condition_interval(self):
         assert (percent_comparison_rule.id, group5.id) in rule_fire_histories
         self.assert_buffer_cleared(project_id=self.project.id)
 
+    def test_apply_delayed_event_frequency_percent_comparison_interval(self):
```
schew2381, line 736: I'm not creating any sessions for this test b/c I don't want to trigger this condition, but rather that it's just using the right comparison_interval mapping. It fails with a KeyError without the fix

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -189,7 +190,7 @@ def get_condition_group_results(
         # The comparison interval is only set for the second query of a percent
         # comparison condition.
         comparison_interval = (
-            condition_inst.intervals[unique_condition.comparison_interval][1]
+            COMPARISON_INTERVALS[unique_condition.comparison_interval][1]
```
saponifi3d, line 193: Thoughts on making this a bit more resilient and using `.get` to access the dictionary attribute? that way we could provide a default value  here as well.

saponifi3d (approved):

PR #73673, opened 2024-07-02
ref(seer grouping): Only send 20 frames if stacktrace is fully minified

lobsterkatie:
When creating embeddings, Seer first has to tokenize the stacktrace string it's sent, and the number of tokens it finds is one of the strongest predictors of its performance. 

In non-minified code, there is generally one instruction per line, resulting in a handful of tokens. In minified code, however, not only are multiple instructions squished onto the same line, but all the nice, however-many-characters-long, human-readable function and variable names have been replaced by one- or two-character names, making even more instructions fit into a single context line. (We truncate context lines to roughly 140 characters, which saves us from collecting an entire bundled script as the context line, but that still can let in a fair number of instructions, given they all look like `if x:`, `te+=1`, `m=e(q,d)`, etc.)

Minified code is therefore significantly more token-dense than non-minified code. This - combined with the fact that minified code groups terribly and therefore events consisting of it are more likely to be sent to Seer in the first place, making them a disproportionate share of Seer's workload - means that it becomes much easier for Seer to get overloaded.

To prevent this, this PR lowers the maximum number of frames in a Seer stacktrace string from 50 to 20 in cases where we have a fully minified stacktrace. (An analysis comparing the results before and after this change showed no significant drop in accuracy.)

### DIFF
--- a/src/sentry/seer/similarity/utils.py
+++ b/src/sentry/seer/similarity/utils.py
@@ -52,12 +52,10 @@ def get_stacktrace_string(data: dict[str, Any]) -> str:
                     for frame in exception_value["values"]
                     if frame.get("id") == "frame" and frame.get("contributes")
                 ]
-                num_frames = len(contributing_frames)
-                if frame_count + num_frames > MAX_FRAME_COUNT:
-                    remaining_frame_count = MAX_FRAME_COUNT - frame_count
-                    contributing_frames = contributing_frames[-remaining_frame_count:]
-                    num_frames = remaining_frame_count
-                frame_count += num_frames
+                contributing_frames = _discard_excess_frames(
+                    contributing_frames, MAX_FRAME_COUNT, frame_count
+                )
+                frame_count += len(contributing_frames)
 
                 for frame in contributing_frames:
                     frame_dict = {"filename": "", "function": "", "context-line": ""}
--- a/src/sentry/seer/similarity/utils.py
+++ b/src/sentry/seer/similarity/utils.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any
+from typing import Any, TypeVar
 
 from sentry.eventstore.models import Event
 from sentry.utils.safe import get_path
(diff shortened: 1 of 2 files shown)


src/sentry/seer/similarity/utils.py
```diff
@@ -52,27 +56,42 @@ def get_stacktrace_string(data: dict[str, Any]) -> str:
...
+    # maximum, since minified frames are significantly more token-dense than non-minified ones
+    result_max_frames = (
+        MAX_FRAME_COUNT
+        if found_non_snipped_context_line
+        else FULLY_MINIFIED_STACKTRACE_MAX_FRAME_COUNT
+    )
+    final_frame_count = 0
+
+    for header, frame_strings in result_parts:
+        frame_strings = _discard_excess_frames(frame_strings, result_max_frames, final_frame_count)
```
jangjodi, line 92: would this extra `_discard_excess_frames` call only need to occur if `found_non_snipped_context_line = True`? otherwise the call on line 59 should truncate to `MAX_FRAME_COUNT` already?

lobsterkatie: I put the conditional on the max because it seemed simpler (and `_discard_excess_frames` short-circuits pretty quickly), but you're right, I could move it here. Lemme do that and re-run the tests.

JoshFerge (approved):
looks good!

jangjodi (approved):

JoshFerge (commented):
my only other thought is, should we restrict this only to events w/ javascript platform?

lobsterkatie:
> my only other thought is, should we restrict this only to events w/ javascript platform?

That happens implicitly, since the `{snip}` comes from the JS processor: https://github.com/getsentry/sentry/blob/d077a5bb7e13a5927794b35d9ae667a4f181feb7/src/sentry/lang/javascript/utils.py#L72-L77

I can add a comment to that effect. I'd rather not put a hard restriction on it, because it'll just be another place to forget to update as we add in support for RN, electron, etc.

JoshFerge:
ah got it. good to know, thanks.!

PR #73674, opened 2024-07-02
feat(new-trace): Route to traceView if traceSlug from query is not null.

Abdkhan14:
Initially we were only routing to the trace from when the event_id was associated to a transaction.

### DIFF
--- a/static/app/views/discover/table/tableView.tsx
+++ b/static/app/views/discover/table/tableView.tsx
@@ -190,7 +190,7 @@ function TableView(props: TableViewProps) {
 
       let target;
 
-      if (dataRow.trace === null) {
+      if (dataRow.trace == null) {
         if (dataRow['event.type'] === 'transaction') {
           throw new Error(
             'Transaction event should always have a trace associated with it.'
@@ -198,7 +198,6 @@ function TableView(props: TableViewProps) {
         }
 
         const project = dataRow.project || dataRow['project.name'];
-
         target = {
           // NOTE: This uses a legacy redirect for project event to the issue group event link
           // This only works with dev-server or production.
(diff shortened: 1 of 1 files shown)


static/app/views/discover/table/tableView.tsx
```diff
@@ -189,13 +189,23 @@ function TableView(props: TableViewProps) {
...
           throw new Error(
             'Transaction event should always have a trace associated with it.'
           );
         }
 
+        const project = dataRow.project || dataRow['project.name'];
+
+        target = {
+          // NOTE: This uses a legacy redirect for project event to the issue group event link
+          // This only works with dev-server or production.
```
narsaynorath, line 204: Just tested this out a bit on my `dev-ui` and it seems to work. Is this comment still necessary? I didn't get any weird infinite redirects or anything.

narsaynorath: Also should we wrap the `to={target}` in the styled link with `normalizeUrl()`?

narsaynorath (approved):
Tested it locally on the org that raised the issue

static/app/views/discover/table/tableView.tsx
Abdkhan14: Good point, I wrapped the path from the `trace:null` case with `normalizeUrl()` the other path already uses it.

static/app/views/discover/table/tableView.tsx
```diff
@@ -190,15 +190,14 @@ function TableView(props: TableViewProps) {
 
       let target;
 
-      if (dataRow.trace === null) {
+      if (dataRow.trace == null) {
```
JonasBa, line 193: This is misleading and can be really hard to catch when looking for bugs as it is actually checking datarow.trace === null || datarow.trace === undefined. You should either use !datarow.trace or do an explicit check for both value types

PR #73679, opened 2024-07-02
feat(autofix): Remove codebase indexing setup step if flag enabled

jennmueng:
In the autofix setup modal, if the `organizations:autofix-disable-codebase-indexing` is enabled, does not show the codebase indexing setup step and allows skipping of the write integration step if needed.

Screenshot 2024-07-03 at 2 13 53 AM

### DIFF
--- a/static/app/components/guidedSteps/guidedSteps.tsx
+++ b/static/app/components/guidedSteps/guidedSteps.tsx
@@ -17,7 +17,10 @@ import {space} from 'sentry/styles/space';
 import usePrevious from 'sentry/utils/usePrevious';
 
 type GuidedStepsProps = {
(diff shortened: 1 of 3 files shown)


static/app/components/guidedSteps/guidedSteps.tsx
```diff
@@ -17,7 +17,10 @@ import {space} from 'sentry/styles/space';
 import usePrevious from 'sentry/utils/usePrevious';
 
 type GuidedStepsProps = {
-  children: React.ReactElement<StepProps> | React.ReactElement<StepProps>[];
+  children:
```
malwilley, line 20: I think this can actually be React.ReactNode now - I forgot to modify the types after I changed it to stop using Children.map which required the children to be Steps

malwilley (approved):

PR #73682, opened 2024-07-02
feat(assignee-selector): Make user searchable by email in dropdown

### LINKED ISSUE #73584: Add support for also filtering on email in assignee selector
### Problem Statement

In Sentry organizations which have names in multiple languages it can be difficult to find assignees since the assignee selector only filters on names.

Internal customer case: 495

### Solution Brainstorm

Add support for filtering the assignee selector by both name and email.

### Product Area

Issues

MichaelSun48:
Fixes #73584

Makes users in the assignee dropdown searchable by their email address

### DIFF
--- a/static/app/components/assigneeSelectorDropdown.spec.tsx
+++ b/static/app/components/assigneeSelectorDropdown.spec.tsx
@@ -501,6 +501,29 @@ describe('AssigneeSelectorDropdown', () => {
     expect(screen.getByTestId('assignee-selector')).toHaveTextContent('CD');
   });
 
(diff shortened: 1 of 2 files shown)


roggenkemper (commented):
does this only allow for email searching? it doesn't seem like name searching works anymore

MichaelSun48:
> does this only allow for email searching? it doesn't seem like name searching works anymore

It shouldn't be, how did you reproduce this?

scttcper (approved):

roggenkemper (approved):

PR #73684, opened 2024-07-02
feat(trace-view): Add UI fallback to old view for STs

k-fish:
### Summary
Singletenant (and on-prem etc.) may not be fully rolled out for span extraction. This means the new trace-view which is entirely span based won't work. This causes links to fallback to the old trace view if the organization is one without span-extraction rolled out, then proceeds to check the rollout date so we can offer the new experience where possible right away.

#### Aside
- `OrganizationSummary` didn't need to be narrowed, so switched it out for the full `Organization` provided at the top level by `useOrganization`, can't see a reason why we wouldn't.

### DIFF
--- a/static/app/types/organization.tsx
+++ b/static/app/types/organization.tsx
@@ -48,8 +48,8 @@ export interface Organization extends OrganizationSummary {
   allowJoinRequests: boolean;
   allowSharedIssues: boolean;
   attachmentsRole: string;
+  /** @deprecated use orgRoleList instead. */
   availableRoles: {id: string; name: string}[];
-  // Deprecated, use orgRoleList
   dataScrubber: boolean;
   dataScrubberDefaults: boolean;
   debugFilesRole: string;
--- a/static/app/types/organization.tsx
+++ b/static/app/types/organization.tsx
@@ -48,7 +48,8 @@ export interface Organization extends OrganizationSummary {
   allowJoinRequests: boolean;
   allowSharedIssues: boolean;
   attachmentsRole: string;
-  availableRoles: {id: string; name: string}[]; // Deprecated, use orgRoleList
+  availableRoles: {id: string; name: string}[];
+  // Deprecated, use orgRoleList
   dataScrubber: boolean;
   dataScrubberDefaults: boolean;
   debugFilesRole: string;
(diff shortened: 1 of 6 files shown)


static/app/types/organization.tsx
```diff
@@ -48,7 +48,8 @@ export interface Organization extends OrganizationSummary {
   allowJoinRequests: boolean;
   allowSharedIssues: boolean;
   attachmentsRole: string;
-  availableRoles: {id: string; name: string}[]; // Deprecated, use orgRoleList
+  availableRoles: {id: string; name: string}[];
+  // Deprecated, use orgRoleList
```
mjq, line 52: The new line break here makes the comment's location misleading IMO.

static/app/views/performance/traceDetails/utils.tsx
```diff
@@ -68,12 +79,28 @@ export function getTraceDetailsUrl({
...
+ *
+ * This code can be removed at the time we're sure all STs have rolled out span extraction.
+ */
+export function shouldForceRouteToOldView(
+  organization: Organization,
+  timestamp: string | number | undefined
+) {
+  const usableTimestamp = getTimeStampFromTableDateField(timestamp);
+  if (!usableTimestamp) {
+    // Timestamps must always be provided for the new view, if it doesn't exist, fall back to the old view.
```
Abdkhan14, line 96: @k-fish , it just helps make the query faster. The traceview still loads without it.

static/app/types/organization.tsx
k-fish: Deprecated should actually be annotated, which makes it show up in editors and also moves with the key properly, going to change it to that instead.

Screenshot 2024-07-04 at 2 58 08 PM

Screenshot 2024-07-04 at 2 58 13 PM

static/app/views/performance/traceDetails/utils.tsx
k-fish: I realize it works on the api but the frontend it should be mandatory no? it significantly improves trace performance, I had assumed all callsites were sending it where appropriate for the new view.

k-fish:
@Abdkhan14 are there any other call sites / a better way instead of a redirect to do this?

static/app/views/performance/traceDetails/utils.tsx
Abdkhan14: Yeah we do it from all callsites. However, even without it the new traceview should load faster than the old traceview I believe. But yeah, just wanted to point it out in case you thought that the new view wouldn't work at all without timestamp.

Abdkhan14:
> @Abdkhan14 are there any other call sites / a better way instead of a redirect to do this?

@k-fish, you've covered all entry points. We need the traceSlug from the fetched event so re-routing seems to be a straightforward solution here.

Abdkhan14 (approved):

PR #73688, opened 2024-07-02
feat(relocation): Add relocatin provenance

azaslavsky:
This will allow each relocation to track whether it is self-hosted or Saas->Saas. We'll use this information to differentiate logic slightly where appropriate.


src/sentry/models/relocation.py
```diff
@@ -81,6 +81,15 @@ class EmailKind(Enum):
         def get_choices(cls) -> list[tuple[int, str]]:
             return [(key.value, key.name) for key in cls]
 
+    class Provenance(Enum):
+        SELF_HOSTED = 0
+        SAAS_TO_SAAS = 1
```
hubertdeng123, line 86: nit: technically shouldn't this be just SAAS?

hubertdeng123 (approved):

src/sentry/migrations/0733_relocation_provenance.py
```diff
@@ -0,0 +1,54 @@
...
+    ]
+
+    operations = (
+        [
+            migrations.SeparateDatabaseAndState(
+                database_operations=(
+                    [
+                        migrations.RunSQL(
+                            """
+                        ALTER TABLE "sentry_relocation" ADD COLUMN "provenance" integer NOT NULL DEFAULT 1;
```
markstory, line 34:
```suggestion
                        ALTER TABLE "sentry_relocation" ADD COLUMN "provenance" integer NOT NULL DEFAULT 0;
```

Default values should be aligned.

src/sentry/models/relocation.py
```diff
@@ -81,6 +81,15 @@ class EmailKind(Enum):
         def get_choices(cls) -> list[tuple[int, str]]:
             return [(key.value, key.name) for key in cls]
 
+    class Provenance(Enum):
```
markstory, line 84: If you inherit from IntEnum you won't have to use `.value` when using the enum.

src/sentry/migrations/0733_relocation_provenance.py
azaslavsky: Done.

src/sentry/models/relocation.py
azaslavsky: Done.

src/sentry/models/relocation.py
azaslavsky: Think `SAAS_TO_SAAS` is a bit clearer and more greppable in this context.

markstory (approved):

PR #73689, opened 2024-07-02
feat(relocation): Implement SaaS -> SaaS export

azaslavsky:
To perform a SaaS -> SaaS relocation, we must first successfully export the serialized organization data from the source region to the target region. This PR enables this capability, using a middleman proxy service on the control silo to coordinate this work.


azaslavsky:
@markstory: this is not yet ready for full review (no tests, etc), but I wanted to get it on your radar to make sure that the "control silo proxy + using GCS to persist work" strategy we discussed is reasonably implemented here.

src/sentry/receivers/outbox/control.py
```diff
@@ -25,6 +27,7 @@
 from sentry.organizations.services.organization import RpcOrganizationSignal, organization_service
 from sentry.receivers.outbox import maybe_process_tombstone
 from sentry.sentry_apps.services.app.service import get_by_application_id, get_installation
+from sentry.services.hybrid_cloud.relocation_export import region_relocation_export_service
```
markstory, line 30: The `sentry/services/hybrid_cloud` module was removed in master. Could this be moved to `sentry.backup.services` with the existing import_export service?

src/sentry/services/hybrid_cloud/relocation_export/impl.py
```diff
@@ -0,0 +1,201 @@
...
+from uuid import UUID
+
+from django.db import router
+from sentry_sdk import capture_exception
+
+from sentry.models.files.file import File
+from sentry.models.files.utils import get_relocation_storage
+from sentry.models.outbox import ControlOutbox, OutboxCategory, OutboxScope
+from sentry.models.relocation import Relocation, RelocationFile
+from sentry.services.hybrid_cloud.relocation_export.model import (
```
markstory, line 17: Could we move this to `sentry.backup.services` or `sentry.hybridcloud.services`

src/sentry/services/hybrid_cloud/relocation_export/impl.py
```diff
@@ -0,0 +1,201 @@
...
+            fp = BytesIO(encrypted_contents)
+            file = File.objects.create(name="raw-relocation-data.tar", type=RELOCATION_FILE_TYPE)
+            file.putfile(fp, blob_size=RELOCATION_BLOB_SIZE, logger=logger)
+            logger.info("SaaS -> SaaS relocation underlying File created", extra=logger_data)
+
+            RelocationFile.objects.create(
+                relocation=relocation,
+                file=file,
+                kind=RelocationFile.Kind.RAW_USER_DATA.value,
+            )
```
markstory, line 114: We'll need a way to make this idempotent. Outboxes can be retried multiple times because of timeouts/connection errors.

src/sentry/tasks/relocation.py
```diff
@@ -46,12 +53,17 @@
...
 )
 from sentry.models.user import User
-from sentry.organizations.services.organization import organization_service
+from sentry.services.hybrid_cloud.lost_password_hash import lost_password_hash_service
+from sentry.services.hybrid_cloud.organization import organization_service
+from sentry.services.hybrid_cloud.relocation_export.model import (
+    RelocationExportReplyWithExportParameters,
+)
+from sentry.services.hybrid_cloud.relocation_export.service import control_relocation_export_service
+from sentry.services.hybrid_cloud.user.service import user_service
```
markstory, line 62: All of these have moved around.

src/sentry/tasks/relocation.py
```diff
@@ -131,9 +145,197 @@
...
-# <=10MB, and one for large imports >10MB. Then we should limit the number of daily executions of
-# the latter.
+@instrumented_task(
+    name="sentry.relocation.uploading_start",
+    queue="relocation",
+    autoretry_for=(Exception,),
+    max_retries=MAX_FAST_TASK_RETRIES,
+    retry_backoff=RETRY_BACKOFF,
+    retry_backoff_jitter=True,
+    soft_time_limit=FAST_TIME_LIMIT,
```
markstory, line 155: You likely also want `acks_late=True`

src/sentry/tasks/relocation.py
```diff
@@ -131,9 +145,197 @@
...
+    In the case of a `SAAS_TO_SAAS` relocation, we'll need to export an organization from the
+    exporting region (ER) back to the requesting region (RR - where this method is running). Because
+    region-to-region messaging is forbidden, all of this messaging will need to be proxied via the
+    control silo (CS). Thus, to accomplish this export-and-copy operation, we'll need to use
+    sequenced RPC calls to fault-tolerantly execute code in these three siloed locations.
+
+    Due to of our system architecture, the sequence of remote functions executed can be a bit hard
+    to follow, so it is diagrammed and listed below. Each function executed is given a sequential
+    numerical identifier and is annotated with information about where to find the source and which
+    silo it will be executed in:
```
markstory, line 173: Could the behavior of this be simpler to understand if we structured the code differently?

src/sentry/tasks/relocation.py
```diff
@@ -131,9 +145,197 @@
...
+        |            |            |     09     |            |            |
+        |            |            |     10     |            |            |
+        |            |<----11-----|            |            |            |
+        |     12     |            |            |            |            |
+        |            |            |            |            |            |
+
+
+    01. (RR) .../tasks/relocation.py::uploading_start: This first function grabs this (aka the
+        "requesting" region) region's public key, then sends an RPC call to the control silo,
+        requesting an export of some `org_slug` from the `replying_region_name` in which is lives.
```
markstory, line 197: Is the `replying_region_name` the same as the exporting region?

src/sentry/tasks/relocation.py
```diff
@@ -131,9 +145,197 @@
...
+    10. (CS) .../receiver/outbox/control.py::process_relocation_reply_with_export: Whenever an
+        outbox draining attempt occurs, this code will be called to read the export data from the
+        local relocation-specific GCS bucket, then forward it into the requesting region.
+    11. The `DBBackedExportService::reply_with_export` call is sent over the wire from the control
+        silo back to the requesting region.
+    12. (RR) .../relocation_export/impl.py::DBBackedRelocationExportService::reply_with_export:
+        We've made it all the way back! The export data gets saved to a `RelocationFile` associated
+        with the `Relocation` that originally triggered `uploading_start`, and the next task in the
+        sequence (`uploading_complete`) is scheduled.
+    """
```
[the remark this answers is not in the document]
markstory: Did you considering having the relocation process start in control? A sequence diagram like

```mermaid
sequenceDiagram

participant User
participant Destination
participant Control
participant Source

User ->> Control : move acme from us to de
Control ->> Control : save outbox for export request
Control ->> Destination : get public key for export
Control ->> Source : request export for slug=acme
Source ->> Source : export organization
Source ->> Control : send export blob
Control ->> Control : save blob and outbox
Control ->> Destination : send export blob and start import
Destination ->> Destination : save outbox + blob
Destination ->> Destination : process import
```

This division would have one fewer outbox driven step, but is otherwise mostly the same 🤷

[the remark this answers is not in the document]
azaslavsky: I think we still need an outbox after `export organization` and before `export blob`, since we don't want to have to redo the work of creating the export if the send call fails? I think it would end up with the same number of outboxes in that case.

More generally: I considered this as an option, but determined that it's a bit easier to start things on the "region where the relocation will go side", since that is where regular relocations go as well, and where the UUID-generating `Relocation` model lives. If we start it from `Control`, then we don't have the ID until we hit the requesting region for the first time with the `get public key for export` call.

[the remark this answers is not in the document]
azaslavsky: Also, just my own bias here, but the control flow feels a bit easier to reason about for me with the first diagram: we always move our request left to right (RR -> CS -> ER), generate the response, then move it back right to left (RR <- CS <- ER).

src/sentry/receivers/outbox/control.py
azaslavsky: See below, resolved.

src/sentry/tasks/relocation.py
azaslavsky: Fixed.

src/sentry/services/hybrid_cloud/relocation_export/impl.py
azaslavsky: I've added a small migration, restricting `RelocationFile` to uniqueness per relocation_uuid + relocation_file_kind pair. This means that even if these outbox retries trigger (wasteful, but not otherwise horrible) duplications at other parts of the chain, the successful export response can only be committed at most once in the requesting region. So for example, if each time we manually write an outbox in this chain we end up sending 2 retries, this method would be triggered 8 times for this particular relocation UUID. However, since only one of those files can win the race, the other 7 will result in errors, and no `uploading_complete` being scheduled.

src/sentry/tasks/relocation.py
azaslavsky: I've added this, as well as a bit of timeout handling, since (I think? Celery docs were a bit unclear on this) we can now get ourselves into an infinite loop of the task failing and never acking.

src/sentry/tasks/relocation.py
azaslavsky: Addressed in the other comment re: the architecture.

src/sentry/tasks/relocation.py
azaslavsky: Yes. The only reason I went with `exporting region` here is because `replying region` would have the same acronym. I can change if it is too confusing.

src/sentry/migrations/0738_one_relocation_file_kind_per_relocation.py
```diff
@@ -0,0 +1,32 @@
...
+    is_post_deployment = False
+
+    dependencies = [
+        ("sentry", "0737_add_discover_saved_query_dataset_source"),
+    ]
+
+    operations = [
+        migrations.AlterUniqueTogether(
+            name="relocationfile",
+            unique_together={("relocation", "file"), ("relocation", "kind")},
```
markstory, line 30: I checked the existing tables and there are no rows that would violate these constraints. :+1:

src/sentry/receivers/outbox/control.py
```diff
@@ -127,3 +130,47 @@ def process_issue_email_reply(shard_identifier: int, payload: Any, **kwds):
...
     )
+
+
+# See the comment on /src/sentry/tasks/relocation.py::uploading_start for a detailed description of
+# how this outbox drain handler fits into the entire SAAS->SAAS relocation workflow.
+@receiver(process_control_outbox, sender=OutboxCategory.RELOCATION_EXPORT_REQUEST)
+def process_relocation_request_new_export(payload: Mapping[str, Any], **kwds):
+    encrypt_with_public_key = (
+        payload["encrypt_with_public_key"].encode("utf-8")
+        if isinstance(payload["encrypt_with_public_key"], str)
```
markstory, line 141: Could you get bytes or another type here?

src/sentry/services/hybrid_cloud/relocation_export/impl.py
markstory: As long as the race settles out to the same state and the duplicate attempts aren't counted as errors that sounds good to me.

src/sentry/tasks/relocation.py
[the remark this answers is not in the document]
markstory: > Also, just my own bias here, but the control flow feels a bit easier to reason about for me with the first diagram:

That is fair, the chevron does have a simplicity to it.

hubertdeng123 (approved):
approving to unblock, but don't have much context here

PR #73691, opened 2024-07-02
feat(trace-explorer): Render list of trace related project icons

Zylphrex:
Instead of always rendering a single project icon for the trace root, render a list of project icons for all related projects in the trace. We try to prioritize projects that are selected in order to avoid users thinking it's making a cross project query.

# Screenshot

image


static/app/views/traces/fieldRenderers.tsx
```diff
@@ -50,6 +55,126 @@ export function SpanDescriptionRenderer({span}: {span: SpanResult<Field>}) {
   );
 }
 
+interface ProjectsRendererProps {
+  projectSlugs: string[];
+  maxVisibleProjects?: number;
+}
+
+export function ProjectsRenderer({
```
Zylphrex, line 63: This was largely adapted from `AvatarList` however that component is very much specialized right now and doesn't extend to this use case very well. A good place to clean up in the future.

static/app/views/traces/content.tsx
```diff
@@ -348,9 +389,13 @@ function TraceRow({defaultExpanded, trace}: {defaultExpanded; trace: TraceResult
...
         <Description>
-          {trace.project ? (
-            <ProjectRenderer projectSlug={trace.project} hideName />
-          ) : null}
+          <ProjectBadgeWrapper>
+            {traceProjects.length > 0 ? (
+              <ProjectsRenderer projectSlugs={traceProjects} />
+            ) : trace.project ? (
+              <ProjectsRenderer projectSlugs={[trace.project]} />
+            ) : null}
```
k-fish, line 397: Maybe just arrayify it inline? nit though

```suggestion
            <ProjectsRenderer projectSlugs={traceProjects.length ? traceProjects : [trace.project]} />
```

static/app/views/traces/content.tsx
```diff
@@ -796,13 +841,13 @@ const StyledPanel = styled(Panel)`
...
   display: grid;
-  grid-template-columns: repeat(1, min-content) auto repeat(2, min-content) 85px 112px 66px;
+  grid-template-columns: 116px auto repeat(2, min-content) 85px 112px 66px;
 `;
 
 const SpanPanelContent = styled('div')`
   width: 100%;
   display: grid;
-  grid-template-columns: repeat(1, min-content) auto repeat(1, min-content) 160px 85px;
+  grid-template-columns: 100px auto repeat(1, min-content) 160px 85px;
```
k-fish, line 850: Why? doesn't this cover ids?

static/app/views/traces/fieldRenderers.spec.tsx
```diff
@@ -0,0 +1,315 @@
...
+  ProjectRenderer,
+  ProjectsRenderer,
+  SpanDescriptionRenderer,
+  SpanIdRenderer,
+  TraceIdRenderer,
+  TraceIssuesRenderer,
+  TransactionRenderer,
+} from 'sentry/views/traces/fieldRenderers';
+
+describe('Renderers', function () {
```
k-fish, line 21: Nice 👏  adding this

static/app/views/traces/fieldRenderers.spec.tsx
```diff
@@ -0,0 +1,315 @@
...
+      'failed_precondition',
+      'aborted',
+      'out_of_range',
+      'unimplemented',
+      'internal_error',
+      'unavailable',
+      'data_loss',
+      'unauthenticated',
+    ])('renders span status %s', function (spanStatus) {
+      const span = makeSpan(projects[0], {'span.status': spanStatus});
```
k-fish, line 120: Unless you have a strong reason to suspect other values would not render or other error paths, I'd just do `ok` and maybe one other since this list is just testing the backend sent data. Otherwise we're just running tests for mocks that will always be true.


```suggestion
      'ok',
    ])('renders span status %s', function (spanStatus) {
      const span = makeSpan(projects[0], {'span.status': spanStatus});
```

static/app/views/traces/fieldRenderers.spec.tsx
```diff
@@ -0,0 +1,315 @@
...
+      expect(collapsed).toBeInTheDocument();
+      expect(collapsed).toHaveTextContent('+2');
+    });
+  });
+
+  describe('ProjectRenderer', function () {
+    it('renders project badge without name', function () {
+      render(<ProjectRenderer projectSlug={projects[0].slug} hideName />);
+      expect(screen.getByRole('img')).toBeInTheDocument();
+    });
```
k-fish, line 157: You're not actually asserting the lack of the name here. Generally `not` is a bad idea, you can decorate with a test-id if you really want to test this empty state. If not I'd just delete this test.

static/app/views/traces/fieldRenderers.spec.tsx
```diff
@@ -0,0 +1,315 @@
...
+          transactionId={span['transaction.id']}
+          onClick={onClickHandler}
+        />
+      );
+
+      const link = screen.getByText('11111111');
+      expect(link).toBeInTheDocument();
+      expect(link).toHaveAttribute(
+        'href',
+        `/organizations/${organization.slug}/performance/${projects[0].slug}:00000000000000000000000000000000/?`
```
k-fish, line 187: span.id?

k-fish (approved):
Nothing breaking, a few code quality comments though.

static/app/views/traces/content.tsx
Zylphrex: we dont use a monospace font for the id so depending on the id, the trace/span id could end up with different widths using min-content

k-fish: Ah now you say it I did notice that

PR #73697, opened 2024-07-02
fix(settings): Fix redirect for settings search

### LINKED ISSUE #73692: Incorrect redirect when searching for 'Let Members Create and Edit Alerts' setting.
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

In the settings search, I get an option to go to a link to "Let Members Create and Edit Alerts" that has a description "Allow members to create, edit, and delete alert rules by granting them the "alerts:write" scope, but it links to https://<org_slug>.sentry.io/settings/#alertsMemberWrite which redirects to the settings page generically. 
Screenshot 2024-07-02 at 2 28 56 PM

### Expected Result

Redirected to https://<org_slug>.sentry.io/settings/organization/

### Actual Result

Redirect to https:/<org_slug>.sentry.io/settings/#alertsMemberWrite

### Product Area

Settings - General

### Link

_No response_

### DSN

_No response_

### Version

_No response_

roggenkemper:
this pr fixes [an issue](https://github.com/getsentry/sentry/issues/73692) with redirecting that came up with the search results for anything on the general organization settings page. since the route didn't include "organization", it was sending users to the wrong place

### DIFF
--- a/static/app/data/forms/organizationGeneralSettings.tsx
+++ b/static/app/data/forms/organizationGeneralSettings.tsx
@@ -6,7 +6,7 @@ import type {BaseRole} from 'sentry/types/organization';
 import slugify from 'sentry/utils/slugify';
 
 // Export route to make these forms searchable by label/help
-export const route = '/settings/:orgId/';
+export const route = '/settings/organization/:orgId';
 
 const formGroups: JsonFormObject[] = [
   {


MichaelSun48 (approved):
Don't think its super important, but maybe worth changing `:orgId` to `:orgIdOrSlug` to be verbose since we're supporting that now

static/app/data/forms/organizationGeneralSettings.tsx
```diff
@@ -6,7 +6,7 @@ import type {BaseRole} from 'sentry/types/organization';
 import slugify from 'sentry/utils/slugify';
 
 // Export route to make these forms searchable by label/help
-export const route = '/settings/:orgId/';
+export const route = '/settings/organization/:orgId';
```
scttcper, line 9: i think this breaks self hosted

scttcper (changes requested):
i guess i also tried. confused as to what the right url should be that would get normalized for both domain and without domain

static/app/data/forms/organizationGeneralSettings.tsx
scttcper, line 9:
```suggestion
export const route = '/settings/organization/:orgId/';
```

PR #73698, opened 2024-07-02
test(rules): Create a unit test for get_condition_groups

schew2381:
Closes https://getsentry.atlassian.net/browse/ALRT-130

### DIFF
--- a/tests/sentry/rules/processing/test_delayed_processing.py
+++ b/tests/sentry/rules/processing/test_delayed_processing.py
@@ -15,8 +15,11 @@ from sentry.rules.conditions.event_frequency import (
     EventFrequencyCondition,
     EventFrequencyConditionData,
 )
-from sentry.rules.processing.delayed_processing import (  # build_group_to_groupevent,; bulk_fetch_events,; get_condition_group_results,; get_group_to_groupevent,; get_rules_to_fire,; ; ; parse_rulegroup_to_event_data,
+from sentry.rules.processing.delayed_processing import (  # bulk_fetch_events; get_group_to_groupevent; parse_rulegroup_to_event_data,
+    DataAndGroups,
+    UniqueConditionQuery,
     apply_delayed,
+    get_condition_group_results,
     get_condition_query_groups,
     get_rules_to_groups,
     get_rules_to_slow_conditions,
@@ -54,6 +57,39 @@ def mock_get_condition_group(descending=False):
     return _callthrough_with_order
 
 
+@freeze_time(FROZEN_TIME)
+class CreateEventTestCase(TestCase, APITestCase):
+    def create_event(
+        self,
+        project_id: int,
+        timestamp: datetime,
+        fingerprint: str,
+        environment=None,
+        tags: list[list[str]] | None = None,
+    ) -> GroupEvent:
+        data = {
+            "timestamp": iso_format(timestamp),
+            "environment": environment,
+            "fingerprint": [fingerprint],
+            "level": "error",
+            "user": {"id": uuid4().hex},
+            "exception": {
+                "values": [
+                    {
+                        "type": "IntegrationError",
+                        "value": "Identity not found.",
+                    }
+                ]
+            },
+        }
+        if tags:
+            data["tags"] = tags
+
+        return self.store_event(
+            data=data, project_id=project_id, assert_no_errors=False, event_type=EventType.ERROR
+        )
+
+
 class BuildGroupToGroupEventTest(TestCase):
     def test_build_group_to_groupevent(self):
         pass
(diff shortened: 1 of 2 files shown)


tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -54,6 +57,39 @@ def _callthrough_with_order(*args, **kwargs):
     return _callthrough_with_order
 
 
+@freeze_time(FROZEN_TIME)
+class CreateEventTestCase(TestCase, BaseEventFrequencyPercentTest):
```
schew2381, line 61: This needs to inherit from `BaseEventFrequencyPercentTest` in order to use the correct `create_event` which doesn't store created events btwn test runs

tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -64,9 +100,50 @@ def test_bulk_fetch_events(self):
         pass
 
 
-class GetConditionGroupResultsTest(TestCase):
+class GetConditionGroupResultsTest(CreateEventTestCase):
```
saponifi3d, line 103: It looks like we're only testing the happy path here, let's make unit tests to validate each branch in logic. this is also magnified by the number of parameters as each param could use a test as well.

### Example
Function to test:
https://github.com/getsentry/sentry/blob/1f7741e0847c031157946907e7862ecdf77943ee/src/sentry/rules/processing/delayed_processing.py#L90-L101


tests: https://github.com/getsentry/sentry/blob/1f7741e0847c031157946907e7862ecdf77943ee/tests/sentry/rules/processing/test_delayed_processing.py#L111-L132

In those tests, i'm testing a number of different loop combinations, and results. each time i use a different context there's a `__with_` that then explains the thing i'm isolating in those tests.

by doing this, we can simplify the tests themselves (moving a lot of scaffolding into a `setUp` function). It also allows us to quickly identify issues in the future, if a test name includes the function and context for the logic it's testing, it's much faster to find the issue.

---

The end goal of these tests are to ensure that our code works in a variety of scenarios;

So for this code:
https://github.com/getsentry/sentry/blob/1f7741e0847c031157946907e7862ecdf77943ee/src/sentry/rules/processing/delayed_processing.py#L168-L207

we should have a _lot_ of tests...
- ensure each of the params works in edge cases
 - condition_groups is empty list
 - project is correct for group, project is incorrect for group

Then start to look at the branches in the code itself; where are there `if` statements, where are there loops, etc. For each of those pieces of logic, we should have a test to verify it.

tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -54,6 +57,39 @@ def _callthrough_with_order(*args, **kwargs):
     return _callthrough_with_order
 
 
+@freeze_time(FROZEN_TIME)
+class CreateEventTestCase(TestCase, BaseEventFrequencyPercentTest):
+    def create_event(
```
saponifi3d, line 62: 🎉 thanks for creating a shared base.

saponifi3d (commented):
I think this is a good start, but we're only testing 1 scenario of the function. i'd recommend breaking the that you have up into a `setUp` function that creates data elements for you, then creating a number of different tests to tryout those scenarios.

let me know if it'd help to hop on a call and walk through any of this too!

tests/sentry/rules/processing/test_delayed_processing.py
schew2381: parameterizing does not work with unittest sad 😔

tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -64,9 +122,123 @@ def test_bulk_fetch_events(self):
...
+        mock_logger.warning.assert_called_once()
+
+    def test_group_does_not_belong_to_project(self):
+        condition_data = self.create_event_frequency_condition(interval=self.interval)
+        condition_groups, group_id, unique_queries = self.create_condition_groups([condition_data])
+
+        results = get_condition_group_results(condition_groups, self.create_project())
+        assert results == {
+            unique_queries[0]: {group_id: 2},
+        }
```
schew2381, line 195: This exactly matches the count comparison test below b/c we don't actually use the passed in project to `get_condition_groups`. We use it to instantiate the class shown below:
https://github.com/getsentry/sentry/blob/8fef42a9c3cebd1c1ddef69362b25bc9ebb39bf4/src/sentry/rules/processing/delayed_processing.py#L183
but `self.project` is not used anywhere in **EventFrequencyAction**. It's required for the base abstract class **RuleBase** which is the parent of the abstract **EventCondition**.
It seems like `self.project` is rarely used, and the few examples I could find were all for **EventAction**, such as this use for **SlackNotifyServiceAction**.

https://github.com/getsentry/sentry/blob/8163594293cd777fd9ca79c976113bed3894227a/src/sentry/integrations/slack/actions/notification.py#L88-L90

It's a bit hard to resolve this by refactoring needing `self.project` in the base class without breaking LSP and tracking down everywhere it's used.
We could add a check in the function, but I'm not sure if this is necessary b/c it would mean our ingestion incorrectly mapped an event to a project in `self.enqueue_rule` in `processor.py`

tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -64,9 +122,121 @@ def test_bulk_fetch_events(self):
...
+    def test_count_and_percent_comparison_conditions(self):
+        count_data = self.create_event_frequency_condition(interval=self.interval)
+        percent_data = self.create_event_frequency_condition(
+            interval=self.interval,
+            comparison_type=ComparisonType.PERCENT,
+            comparison_interval=self.comparison_interval,
+        )
+        condition_groups, group_id, unique_queries = self.create_condition_groups(
+            [count_data, percent_data]
+        )
```
saponifi3d, line 229: thoughts on also adding an invalid condition to this list to ensure it processes as expected?

tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -64,9 +122,121 @@ def test_bulk_fetch_events(self):
...
+        assert get_condition_group_results({}, self.project) == {}
+
+    @patch("sentry.rules.processing.delayed_processing.logger")
+    def test_nonexistent_condition_cls(self, mock_logger):
+        first_query = UniqueConditionQuery(cls_id="fake_id", interval="", environment_id=1)
+        fake_data_groups = DataAndGroups(
+            data=self.create_event_frequency_condition(id="fake_id"),
+            group_ids={1},
+        )
+        assert get_condition_group_results({first_query: fake_data_groups}, self.project) is None
```
saponifi3d, line 169: Do we want to return `None` for all conditions or just the specific condition? It looks like in code we will return `None` for all conditions if 1 condition doesn't exist.

My expectation is more that the this condition result would be none, not that _all_ condition results are none.

saponifi3d (approved):
generally lgtm, curious about the invalid condition case though...

tests/sentry/rules/processing/test_delayed_processing.py
schew2381: oh yeah good catch, will fix this

PR #73702, opened 2024-07-02
chore(hybridcloud) Add newly moved services to mypy stronglist

markstory:
Resolve the outstanding typing issues and mistakes that were preventing these modules from being added to the stronglist before.

Refs HC-982


src/social_auth/backends/__init__.py
```diff
@@ -703,7 +703,7 @@ def do_auth(self, access_token, *args, **kwargs):
 
 
 # Cache for discovered backends.
-BACKENDSCACHE: dict[str, type[SocialAuthBackend]] = {}
+BACKENDSCACHE: dict[str, type[BaseAuth]] = {}
```
markstory, line 706: I think this type was wrong before. With this type in-place I was getting

```
src/social_auth/utils.py:39: error: Item "type" of "Any | type[SocialAuthBackend]" has no attribute "AUTH_BACKEND"  [union-attr]
```

Which points to the type being wrong. Furthermore, the `get_backends()` function (in this module) picks properties out of the module's `BACKENDS` dict by name and appears to always get a `BaseAuth` which has `AUTH_BACKEND` as a property that is a `SocialAuthBackend`.

asottile-sentry: looks like I missed the bit below and assumed the `issubclass(..., SocialAuthBackend)` check applied to the values

asottile-sentry (approved):

PR #73703, opened 2024-07-02
fix(grouping): Fix missing system-only stacktraces in similar issues diff view

### LINKED ISSUE #72607: The Diff button inside the Similar Issues tab should show grouping information, not just error message
### Problem Statement

When looking at the differences between groups, I am trying to understand whether they can be grouped together, in other words, why did Sentry group them differently in the first place?

Right now the Diff button, when pressed, shows only error messages.  For us, they're almost always identical.

This is the button I'm talking about:
Screenshot by Dropbox Capture
This is what I see when I click it.  Very unhelpful.
Screenshot by Dropbox Capture

### Solution Brainstorm

I'd instead like to see the reason for the grouping diffed so it actually tells me why they're not the same.

In most cases that will mean the stack trace and exception type, but whatever is in the grouping details should ideally be supported.  I'll take just stack trace and exception type as an incremental improvement, though.

### Product Area

Issues

lobsterkatie:
As reported in https://github.com/getsentry/sentry/issues/72607, sometimes the diff view (used on both the merged and similar issues tabs of the issue details page) shows no stacktraces, even when the events being compared do in fact have them.

The culprit turns out to be the check [here](https://github.com/getsentry/sentry/blob/ff5efb59a42e5936507bcbc2a36888bc966742a8/static/app/components/events/interfaces/crashContent/stackTrace/rawContent.tsx#L176-L179), which filters the frames based on in-app status when the Seer similar issues flag is on. On an individual frame level, this works great, but it can leave the stacktrace overall in a bad state, with no frames to show, which happens whenever the flag is on and the stacktrace consists of nothing but system frames.

This fixes that by filtering the stacktrace as a whole before passing frames to `getFrame`, so we can recover if it turns out said filtering would leave us with no frames. It also refactors the tests a bit to make sure we're covering all cases. 

Fixes https://github.com/getsentry/sentry/issues/72607.

### DIFF
--- a/static/app/components/events/interfaces/crashContent/stackTrace/rawContent.spec.tsx
+++ b/static/app/components/events/interfaces/crashContent/stackTrace/rawContent.spec.tsx
@@ -122,39 +122,76 @@ describe('RawStacktraceContent', function () {
       );
     });
 
-    const data_with_non_in_app: StacktraceType = {
+    const inAppFrame = (fnName, line) =>
+      FrameFixture({
+        function: fnName,
+        module: 'example.application',
+        lineNo: line,
+        filename: 'application',
+        platform: undefined,
+      });
+    const systemFrame = (fnName, line) =>
+      FrameFixture({
+        function: fnName,
+        module: 'example.application',
+        lineNo: line,
+        filename: 'application',
+        platform: undefined,
+        inApp: false,
+      });
+
+    const onlyInAppFrames: StacktraceType = {
       hasSystemFrames: false,
       framesOmitted: null,
       registers: {},
-      frames: [
-        FrameFixture({
-          function: 'main',
-          module: 'example.application',
-          lineNo: 1,
-          filename: 'application',
-          platform: undefined,
-        }),
-        FrameFixture({
-          function: 'doThing',
-          module: 'example.application',
-          lineNo: 2,
-          filename: 'application',
-          platform: undefined,
-          inApp: false,
-        }),
-      ],
+      frames: [inAppFrame('main', 1), inAppFrame('doThing', 2)],
     };
 
-    it('renders non in-app frames example', () => {
-      expect(displayRawContent(data_with_non_in_app, 'python', exception)).toEqual(
-        `Error: an error occurred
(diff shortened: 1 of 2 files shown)


static/app/components/events/interfaces/crashContent/stackTrace/rawContent.tsx
```diff
@@ -170,15 +170,18 @@ export default function displayRawContent(
...
-      frames.push(getFrame(frame, frameIdx, platform));
-    }
+  const hasInAppFrames = rawFrames.some(frame => frame.inApp);
+  const shouldFilterOutSystemFrames = hasSimilarityEmbeddingsFeature && hasInAppFrames;
+
+  const framesToUse = shouldFilterOutSystemFrames
+    ? rawFrames.filter(frame => frame.inApp)
+    : rawFrames;
+
+  framesToUse.forEach((frame, frameIdx) => {
```
malwilley, line 183: nit: could replace this `forEach` and `push` with a `const frames = framesToUse.map(...)`

static/app/components/events/interfaces/crashContent/stackTrace/rawContent.spec.tsx
```diff
@@ -122,39 +122,76 @@ describe('RawStacktraceContent', function () {
       );
     });
 
-    const data_with_non_in_app: StacktraceType = {
```
malwilley, line 125: snakecase begone!

malwilley (approved):

static/app/components/events/interfaces/crashContent/stackTrace/rawContent.tsx
lobsterkatie: Oh, good call. I was just riffing off of what was there, but you're right, that's cleaner.

static/app/components/events/interfaces/crashContent/stackTrace/rawContent.spec.tsx
lobsterkatie: Ha - not mine, but I hadn't even noticed! All this time programming in python has corrupted me!

PR #73704, opened 2024-07-02
ref(metrics): Add edit route

### LINKED ISSUE #73420: Onboarding Updates for Span-based Metrics
**Problem**: The current empty state of the Metrics explorer page is still focused around setting up the "old" custom metrics. We need to update it to reflect the new way of configuring metrics from span attributes.

**Solution**:

**(1) If they’re already using performance and don't have old custom metrics**
- Show the explorer page
- Remove the banner to set up custom metrics
- In the query 3 dot menu - Update “Metric Settings” so it’s not disabled for the out-of-the-box metrics (e.g. span.duration), but leads to the “Configure Metrics” screen with the attribute pre-selected
- Show button "Add New Metric" on top right that leads the customers to the Metrics Settings page.

Image

Image

**(2)  If they’re not using performance and don't have old custom metrics**
- Update Description
  - Track and solve what matters
  - “Query and plot metrics extracted from your span data, to visualise trends and identify anomalies. To get started, you need to enable tracing (https://docs.sentry.io/concepts/key-terms/tracing/).”
- Update CTA button
  - “Set Up Tracing” links to Tracing setup docs (https://docs.sentry.io/product/performance/getting-started/)

**(3) First visit to the Metrics page**
- Remove Modal with beta disclaimer (we would need to update it, it does not feel worth it at this point since there is a warning on the page already about the upcoming API changes)

Image

**(4) If they have old custom metrics**
- Display red banner telling them that this old method is deprecated
   - Description: "We've released a new API to submit metrics. Metrics using with the old API will stop being ingested on XX/YY. Learn more [link to docs or help article will be provided later]."
- CTA on top right is renamed from "Add Custom Metric" to "Add New Metric" and instead of opening the setup instructions panel, it leads the customers to the Metrics Settings page.
- The rest continues to work the same
  - Clicking on Metrics Settings brings them to the Metric Details page as-is

Image

priscilawebdev:
### Goal

- When users are on the metrics page and a span attribute is selected, clicking the three-dot button will redirect them to the metrics settings. This action will open the metrics rule modal, with all information pre-filled.

- Users on the metrics page by clicking on the edit icon, shall see the metric rule modal, with all information pre-filled.

**Preview:**

https://github.com/getsentry/sentry/assets/29228205/e1102409-a45f-480e-b458-9b33fc016bab

closes https://github.com/getsentry/sentry/issues/73420

### DIFF
--- a/static/app/routes.tsx
+++ b/static/app/routes.tsx
@@ -550,6 +550,10 @@ function buildRoutes() {
             () => import('sentry/views/settings/projectMetrics/projectMetricsDetails')
           )}
         />
+        <Route
+          path=":spanAttribute/edit/"
+          component={make(() => import('sentry/views/settings/projectMetrics'))}
+        />
       </Route>
       <Route
         path="replays/"
--- a/static/app/views/metrics/metricQueryContextMenu.tsx
+++ b/static/app/views/metrics/metricQueryContextMenu.tsx
@@ -152,8 +152,14 @@ export function MetricQueryContextMenu({
             organization,
           });
           Sentry.metrics.increment('ddm.widget.settings');
+
+          const {useCase, name} = parseMRI(metricsQuery.mri) ?? {};
+
+          const isSpanBasedMetric = useCase === 'spans';
+
           if (
             !hasCustomMetricsExtractionRules(organization) ||
+            !isSpanBasedMetric ||
             isCustomMetric({mri: metricsQuery.mri})
           ) {
             navigateTo(
@@ -163,7 +169,6 @@ export function MetricQueryContextMenu({
               router
             );
           } else {
-            const {name} = parseMRI(metricsQuery.mri) ?? {};
             navigateTo(`/settings/projects/:projectId/metrics/${name}/edit/`, router);
           }
         },
--- a/static/app/views/metrics/metricQueryContextMenu.tsx
+++ b/static/app/views/metrics/metricQueryContextMenu.tsx
@@ -32,6 +32,7 @@ import {
   hasCustomMetricsExtractionRules,
   hasMetricAlertFeature,
 } from 'sentry/utils/metrics/features';
+import {parseMRI} from 'sentry/utils/metrics/mri';
 import {
   isMetricsQueryWidget,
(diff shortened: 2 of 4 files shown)


static/app/views/metrics/metricQueryContextMenu.tsx
```diff
@@ -162,8 +163,8 @@ export function MetricQueryContextMenu({
               router
             );
           } else {
-            // TODO(telemetry-experience): As soon as the span-based-metrics data has an unique identifier, we should use it here
-            navigateTo(`/settings/projects/:projectId/metrics/`, router);
+            const {name} = parseMRI(metricsQuery.mri) ?? {};
+            navigateTo(`/settings/projects/:projectId/metrics/${name}/edit/`, router);
```
ArthurKnaus, line 167: This logic branch will now be executed for all non-custom metrics. Also those witch are in namespaces, that are not covered by span extraction rules e.g. `transaction.duration`. Is that intentional?

static/app/routes.tsx
```diff
@@ -550,6 +550,10 @@ function buildRoutes() {
             () => import('sentry/views/settings/projectMetrics/projectMetricsDetails')
           )}
         />
+        <Route
+          path=":spanAttribute/edit/"
+          component={make(() => import('sentry/views/settings/projectMetrics'))}
+        />
```
ArthurKnaus, line 556: Just wondering here: Would there be any benefit in using a query param instead? 🤔

ArthurKnaus (approved):

static/app/routes.tsx
priscilawebdev: I was wondering the same thing, but since we have everything else we need on the page, for now we don't need the other params

static/app/views/metrics/metricQueryContextMenu.tsx
priscilawebdev: I thought we would still update the field on the metrics page so that the list is aligned with the table we display in settings. But this is not the intention, is there any function I can use to filter this?

PR #73710, opened 2024-07-03
feat(flagpole): Adds sentry CLI command for generating new flagpole features

GabeVillalobos:
Adds a new CLI command `getsentry createflag` for generating skeleton feature yaml. By default, the CLI will prompt a user to enter a name, owner name, and a default segment, although this can be skipped in a number of ways.

Includes the following options:
`--name`: Provide a name via the CLI invocation. If not provided, the user will be prompted for one.
`--scope`: Provide a scope ('organization' or 'project'). If not provided, the user will be prompted for one.
`--owner`: Provide a feature owner field. If not provided, the user will be prompted for one.
`--blank`: Skip generating any segments and directly print the yaml template for the flag.

The only major things the CLI currently does not handle:
- Autofilling or showing all possible context properties a developer can match against.
- Providing prompts to enter values for each generated condition.

The hope is that this is enough to easily scaffold new feature flag yaml for use in flagpole, without prompting for too many fields or being overly burdensome.

Here's an example CLI invocation for a new feature with a couple of segments:
```
$ sentry createflag
Owner (team name or email address): Hybrid Cloud
Feature name: Super New Feature
Feature scope (organization, project) [organization]: organization
Would you like to create a new segment? [Y/n]:
Name: GA Segment
Rollout percentage [100]:
Would you like to create some conditions? [Y/n]: n
Continue creating segments? [y/N]: y
Name: Early Access Segment
Rollout percentage [100]:
Would you like to create some conditions? [Y/n]: y
Operator type (in, not_in, contains, not_contains, equals, not_equals): equals
Here are some example condition properties available:

project_slug
project_id
project_name
user_email
organization_is-early-adopter
user_id
organization_slug
organization_id

Context property name: organization_is-early-adopter
Continue creating conditions? [y/N]:
Continue creating segments? [y/N]:

=== GENERATED YAML ===

feature.organization:super-new-feature:
  created_at: '2024-07-03T04:42:04.575683'
  enabled: true
  owner: Hybrid Cloud
  segments:
  - conditions: []
    name: GA Segment
    rollout: 100
  - conditions:
    - operator: equals
      property: organization_is-early-adopter
      strict_validation: false
      value: ''
    name: Early Access Segment
    rollout: 100
```

### DIFF
--- a/src/flagpole/__init__.py
+++ b/src/flagpole/__init__.py
@@ -68,7 +68,7 @@ import orjson
 import yaml
 from pydantic import BaseModel, Field, ValidationError, constr
 
-from flagpole.conditions import Segment
+from flagpole.conditions import ConditionBase, Segment
 from flagpole.evaluation_context import ContextBuilder, EvaluationContext
 
 
@@ -156,4 +156,11 @@ class Feature(BaseModel):
         return orjson.dumps(self.to_dict()).decode()
 
 
-__all__ = ["Feature", "InvalidFeatureFlagConfiguration", "ContextBuilder", "EvaluationContext"]
+__all__ = [
+    "Feature",
+    "InvalidFeatureFlagConfiguration",
+    "ContextBuilder",
+    "EvaluationContext",
+    "Segment",
+    "ConditionBase",
+]
--- a/src/sentry/runner/commands/createflag.py
+++ b/src/sentry/runner/commands/createflag.py
@@ -1,0 +1,141 @@
+import click
+
+from flagpole import Feature, Segment
+from flagpole.conditions import (
+    ConditionBase,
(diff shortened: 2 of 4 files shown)


markstory (commented):
Looking good. Should we have a test for a simple use case? [click has some affordances](https://click.palletsprojects.com/en/8.1.x/testing/) for responding to prompts.

GabeVillalobos:
> Looking good. Should we have a test for a simple use case? [click has some affordances](https://click.palletsprojects.com/en/8.1.x/testing/) for responding to prompts.

@markstory yep, can do! I'll add those shortly, but wanted to hold off until I got some feedback on the format and prompt.

markstory (approved):

PR #73711, opened 2024-07-03
feat(replay): Create a replay-compare tool that uses a slider to overlay two videos

ryan953:
This creates a 'slider diff' to compare two replays, or two moments from the same replay in a visual way.

By dragging the purple bar left/right you can reveal more or less of the replay below. Making it possible to see pixel-by-pixel what has changed. In this case we're comparing the same replay, but at different timestamps to show a hydration error:
SCR-20240702-sycx

This diff tool also appears inside the "Open Diff Viewer" modal, as the first/default tab selected. Also by default, the purple line starts in the center; so it's obvious it's there, but not always obvious where the difference is.
SCR-20240702-szvu

I also refactored some code out into new files, because the one file was getting really large.

Relates to https://github.com/getsentry/sentry/pull/73609 (this is seriously needed)


static/app/components/replays/diff/replayDiffChooser.tsx
```diff
@@ -0,0 +1,75 @@
...
+import type ReplayReader from 'sentry/utils/replays/replayReader';
+
+interface Props {
+  leftOffsetMs: number;
+  replay: null | ReplayReader;
+  rightOffsetMs: number;
+  defaultTab?: DiffType;
+}
+
+export enum DiffType {
```
scttcper, line 18: just still recommending const enums since they work like strings instead of a weird object
```suggestion
export const enum DiffType {
```

static/app/components/replays/diff/replaySliderDiff.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+export function ReplaySliderDiff({leftOffsetMs, replay, rightOffsetMs}: Props) {
+  const positionedRef = useRef<HTMLDivElement>(null);
+  const viewDimensions = useDimensions({elementRef: positionedRef});
+
+  const width = toPixels(viewDimensions.width);
+  return (
+    <WithPadding>
+      <Positioned ref={positionedRef}>
+        {viewDimensions.width ? (
```
billyvg, line 28: Is this what we want if width is 0?

billyvg (approved):

static/app/components/replays/diff/replaySliderDiff.tsx
ryan953: i think so? up until this morning it was easy/common to get `0` during the first render, but now we have a `useLayoutEffect` happening https://github.com/getsentry/sentry/pull/73958

I think if the width is 0, then it's not visible and we can skip inserting the children, it should re-render when the size is increased.

PR #73714, opened 2024-07-03
ref(metrics): Update route cond for metrics settings

priscilawebdev:
related to the feedback https://github.com/getsentry/sentry/pull/73704#discussion_r1663564799


ArthurKnaus (approved):
Just one thing:
```
disabled: !isCustomMetric({mri: metricsQuery.mri}),
```
You want to disable this item if it is not a custom metric.

PR #73727, opened 2024-07-03
feat(trace_timeline): Remove the need to check flag

armenzg:
Trace timelines are rendered conditionally, thus, we do not need to check the property.


malwilley (approved):
It makes sense that we probably always want to show this when there's a trace_id in the context, looks good

PR #73734, opened 2024-07-03
fix(snuba): Allow common suffixes when validating referrers

### LINKED ISSUE #71570: referrer is not part of Referrer Enum
### Self-Hosted Version

24.4.2

### CPU Architecture

x86_64

### Docker Version

20.10.24+dfsg1

### Docker Compose Version

2.24.6

### Steps to Reproduce

?

I guess this is internal code (to Sentry) calling with such referrers

### Expected Result

No errors in logs

### Actual Result

```
web-1                                           | Traceback (most recent call last):
web-1                                           |   File "/usr/local/lib/python3.11/site-packages/sentry/snuba/referrer.py", line 824, in validate_referrer
web-1                                           |     raise Exception(error_message)
web-1                                           | Exception: referrer api.performance.http.domain-summary-transactions-list.secondary is not part of Referrer Enum
web-1                                           | 21:20:56 [WARNING] sentry.snuba.referrer: referrer api.performance.http.domain-summary-transactions-list.secondary is not part of Referrer Enum
web-1                                           | Traceback (most recent call last):
web-1                                           |   File "/usr/local/lib/python3.11/site-packages/sentry/snuba/referrer.py", line 824, in validate_referrer
web-1                                           |     raise Exception(error_message)
web-1                                           | Exception: referrer api.performance.http.domain-summary-metrics-ribbon.primary is not part of Referrer Enum
web-1                                           | 21:20:56 [WARNING] sentry.snuba.referrer: referrer api.performance.http.domain-summary-metrics-ribbon.primary is not part of Referrer Enum
web-1                                           | Traceback (most recent call last):
web-1                                           |   File "/usr/local/lib/python3.11/site-packages/sentry/snuba/referrer.py", line 824, in validate_referrer
web-1                                           |     raise Exception(error_message)
web-1                                           | Exception: referrer api.performance.http.domain-summary-transactions-list.secondary is not part of Referrer Enum
```

### Event ID

_No response_

gggritso:
The query builder adds `".primary"` and `".secondary"` to the query referrer string when it constructs compound queries (when fetching from multiple tables). It's a pain to have to declare both a primary _and_ a secondary _and_ a regular referrer for each of those cases, especially because the fields in the query can change for all kinds of reasons. Instead, add a list of common referrer suffixes that are considered valid. Closes https://github.com/getsentry/sentry/issues/71570

P.S. There are already a bunch of referrers defined with those suffixes, which I guess I could have removed, but I didn't bother in this PR. Should I trash 'em?

### DIFF
--- a/src/sentry/snuba/referrer.py
+++ b/src/sentry/snuba/referrer.py
@@ -887,15 +887,25 @@ class Referrer(Enum):
 
 VALUES = {referrer.value for referrer in Referrer}
 
+# These suffixes are automatically added by Query Builder code in certain conditions. Any valid referrer with these suffixes is still a valid referrer.
+VALID_SUFFIXES = ["primary", "secondary"]
+
 
 def validate_referrer(referrer: str | None) -> None:
     if not referrer:
         return
 
+    if referrer in VALUES:
+        return
+
+    for suffix in VALID_SUFFIXES:
+        if referrer.removesuffix(f".{suffix}") in VALUES:
+            return
+
     error_message = f"referrer {referrer} is not part of Referrer Enum"
+
     try:
-        if referrer not in VALUES:
-            raise Exception(error_message)
+        raise Exception(error_message)
     except Exception:
         metrics.incr("snql.sdk.api.new_referrers", tags={"referrer": referrer})
         logger.warning(error_message, exc_info=True)
--- a/tests/snuba/test_referrer.py
+++ b/tests/snuba/test_referrer.py
@@ -31,6 +31,21 @@ class ReferrerTest(TestCase):
         assert warn_log.call_count == 1
 
     @patch("sentry.snuba.referrer.logger.warning")
+    def test_referrer_validate_common_suffixes(self, warn_log):
+        assert warn_log.call_count == 0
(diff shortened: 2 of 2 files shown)


src/sentry/snuba/referrer.py
```diff
@@ -887,15 +887,25 @@ class Referrer(Enum):
...
+    for suffix in VALID_SUFFIXES:
+        if referrer.removesuffix(f".{suffix}") in VALUES:
+            return
+
     error_message = f"referrer {referrer} is not part of Referrer Enum"
+
     try:
-        if referrer not in VALUES:
-            raise Exception(error_message)
+        raise Exception(error_message)
```
wmak, line 908: Hmm i think we can just remove the try/except here and directly call `metrics.incr` and `logger.warning`. I think you'll need to drop `exc_info` but i'm not sure

wmak (approved):
nice

src/sentry/snuba/referrer.py
gggritso: I kept this around from the original PR because apparently this is what creates the Sentry exception, which I think the Snuba team wanted

enochtangg (approved):

PR #73737, opened 2024-07-03
chore(slack): Fix Bug when Slack Rate Limits in `get_channel_id`

iamrajjoshi:
There was a small regression in `get_channel_id`. When Slack Rate Limits, the previous implementation forced the caller handles the exception, in the rewrite, we weren't raising it so I fixed it. 

Previous logic:
https://github.com/getsentry/sentry/blob/master/src/sentry/integrations/slack/utils/channel.py#L323-L328

Surprised we haven't run into this, but this probably cause people aren't abusing alerts creation.

I also made the tests reflect the Rate Limiting logic.

### DIFF
--- a/src/sentry/integrations/slack/utils/channel.py
+++ b/src/sentry/integrations/slack/utils/channel.py
@@ -228,8 +228,10 @@ def check_user_with_timeout(
             # TODO: This is a problem if we don't go through all the users and eventually run in to someone with duplicate display name
             if time.time() > time_to_quit:
                 return SlackChannelIdData(prefix=_prefix, channel_id=None, timed_out=True)
-    except SlackApiError:
+    except SlackApiError as e:
         _logger.exception("rule.slack.user_check_error", extra=logger_params)
+        if "ratelimited" in str(e):
+            raise ApiRateLimitedError("Slack rate limited") from e
 
     return SlackChannelIdData(prefix=_prefix, channel_id=_channel_id, timed_out=False)
 
--- a/tests/sentry/integrations/slack/test_utils.py
+++ b/tests/sentry/integrations/slack/test_utils.py
@@ -3,6 +3,7 @@ from unittest.mock import patch
 import orjson
 import pytest
 import responses
+from slack_sdk.web.slack_response import SlackResponse
 
 from sentry.integrations.slack.sdk_client import SLACK_DATADOG_METRIC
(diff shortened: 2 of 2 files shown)


src/sentry/integrations/slack/utils/channel.py
```diff
@@ -228,8 +228,10 @@ def check_user_with_timeout(
             # TODO: This is a problem if we don't go through all the users and eventually run in to someone with duplicate display name
             if time.time() > time_to_quit:
                 return SlackChannelIdData(prefix=_prefix, channel_id=None, timed_out=True)
-    except SlackApiError:
+    except SlackApiError as e:
         _logger.exception("rule.slack.user_check_error", extra=logger_params)
+        if "ratelimited" in str(e):
+            raise ApiRateLimitedError("Slack rate limited") from e
```
cathteng, line 234: hmm i think this warrants more discussion. since we used to use the in-house client for everything, i think this was the default. but if we want callers to be able to handle the various SlackApiErrors appropriately maybe we want to always convert the SlackApiError into the equivalent ApiError?

RyanSkonnord: Given that `ApiRateLimitedError` extends `ApiError`, is that what this is already doing?

RyanSkonnord (approved):
I'm open to further discussion about the structure of our exception types but this looks good to me for now.

src/sentry/integrations/slack/utils/channel.py
cathteng: i mean should we always catch SlackApiError and raise ApiError from the SlackApiError?

cathteng: similar to https://github.com/getsentry/sentry/pull/73666#discussion_r1664752164, maybe we should split the string by `\n` and check in the smaller string?

cathteng:
we can also raise this from inside here https://github.com/getsentry/sentry/blob/2d7bbe5e0173c081104cae1bed418773661553a6/src/sentry/integrations/slack/sdk_client.py#L74-L86, it seems like there's more than one place where we're catching ApiRateLimitedError

cathteng:
ignore me about the raising ApiRateLimited error inside the wrapper, i realize we're only catching SlackApiError everywhere else

PR #73740, opened 2024-07-03
feat(discover): Show banner for split decision

shruthilayaj:
1. Show banner for split decision
2. Use saved query dataset to decide which actual dataset to query
3. Switching datasets preserves old state and only switches dataset instead of resetting the whole query when already in a saved query - it's a better experience if we guess the wrong dataset, the state remains intact as the switch between datasets
4. Removes references to queryDataset in EventView to keep event dataset selector agnostic


static/app/views/discover/eventDetails/index.spec.tsx
```diff
@@ -212,19 +212,19 @@ describe('Discover > EventDetails', function () {
     expect(await screen.findByText('Firefox')).toBeInTheDocument();
     expect(screen.getByRole('link', {name: 'Firefox'})).toHaveAttribute(
       'href',
-      '/organizations/org-slug/discover/results/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All%20Events&query=browser%3AFirefox%20title%3A%22Oh%20no%20something%20bad%22&queryDataset=discover&sort=-timestamp&statsPeriod=24h&yAxis=count%28%29'
+      '/organizations/org-slug/discover/results/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All%20Events&query=browser%3AFirefox%20title%3A%22Oh%20no%20something%20bad%22&sort=-timestamp&statsPeriod=24h&yAxis=count%28%29'
```
shruthilayaj, line 215: No longer passing the `queryDataset=discover` param if the feature flag isn't enabled

static/app/utils/discover/eventView.tsx
```diff
@@ -38,12 +38,12 @@ import {
   DISPLAY_MODE_FALLBACK_OPTIONS,
```
shruthilayaj, line 38: The changes in this file is almost entirely removing the `queryDataset` param from EventView. the discover component instead handles figuring which `dataset` value to pass to Eventview.

static/app/views/discover/homepage.tsx
```diff
@@ -108,10 +109,18 @@ class HomepageQueryAPI extends DeprecatedAsyncComponent<Props, HomepageQueryStat
 
   renderBody(): React.ReactNode {
     const {savedQuery, loading} = this.state;
+    const {organization} = this.props;
+    let savedQueryWithDataset = savedQuery;
+    if (
+      organization.features.includes('performance-discover-dataset-selector') &&
+      savedQuery
+    ) {
+      savedQueryWithDataset = getSavedQueryWithDataset(savedQuery);
```
shruthilayaj, line 118: infers `dataset` based on saved query dataset. `dataset` is then passed down to event view so the queries are made correctly.

static/app/views/discover/results.spec.tsx
```diff
@@ -648,12 +648,12 @@ describe('Results', function () {
 
       expect(screen.getByRole('link', {name: 'timestamp'})).toHaveAttribute(
         'href',
-        'undefined?field=title&field=event.type&field=project&field=user.display&field=timestamp&id=1&name=new&query=&queryDataset=discover&sort=-timestamp&statsPeriod=24h&topEvents=5'
+        'undefined?field=title&field=event.type&field=project&field=user.display&field=timestamp&id=1&name=new&query=&sort=-timestamp&statsPeriod=24h&topEvents=5'
```
shruthilayaj, line 651: No longer passing the queryDataset=discover param if the feature flag isn't enabled

static/app/views/discover/savedQuery/datasetSelector.spec.tsx
```diff
@@ -1,26 +1,57 @@
...
+      'Transactions',
+      'Unknown',
+    ]);
   });
 
-  it('pushes new event view', async function () {
-    render(<DatasetSelector isHomepage={false} savedQuery={undefined} />, {
-      router,
-    });
+  it('pushes new deafult event view if not a saved query', async function () {
```
narsaynorath, line 47:
```suggestion
  it('pushes new default event view if not a saved query', async function () {
```

static/app/utils/discover/eventView.tsx
```diff
@@ -609,6 +599,13 @@ class EventView {
       delete newQuery.query;
     }
 
+    if (this.dataset) {
+      newQuery.queryDataset = getSavedQueryDatasetFromLocationOrDataset(
+        undefined,
+        this.dataset
+      );
+    }
```
narsaynorath, line 607: I'm a little confused about why we need to pass along the `queryDataset` field to a new query if these changes were meant to remove this, could `EventView` just handle a single `dataset` field? Is this being used?

narsaynorath (approved):
Just a small typo and a comment r.e. `queryDataset` in `EventView` but otherwise looks good

static/app/utils/discover/eventView.tsx
shruthilayaj: I went back and forth on this a bit, `toNewQuery()` is primarily used to create the payload for creating/updating a DiscoverSavedQuery object. So I figured I'd rather implement it here than at every individual discover saved query api call. When we implement this change for dashboards, we could add another helper method to EventView like `toNewWidgetQuery()` and do something similar.

narsaynorath: I see, yeah, I think if this is supposed to produce the most valid instance of a `DiscoverSavedQuery` object then I'd rather us keep the logic together than having to remember to add the `queryDataset` attribute everywhere it's called. Thanks for clarifying why this is here!

PR #73749, opened 2024-07-03
chore(slack): Use Slack SDK Client to Open View for Archive/Resolved

iamrajjoshi:
Using Slack SDK Client to show view when you click Archive/Resolved on a Sentry Issue in Slack. (`SDK-2.16` & `SDK-2.17`)


src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -399,32 +400,49 @@ def open_resolve_dialog(self, slack_request: SlackActionRequest, group: Group) -
...
-                extra={
-                    "error": str(e),
-                    "organization_id": org.id,
-                    "integration_id": slack_request.integration.id,
+        if features.has("organizations:slack-sdk-action-view-open", group.project.organization):
+            slack_client = SlackSdkClient(integration_id=slack_request.integration.id)
+            try:
+                slack_client.views_open(
+                    trigger_id=slack_request.data["trigger_id"],
+                    view=modal_payload,
```
cathteng, line 410: any chance we can make `modal_payload` into a `View` from `slack_sdk`? `views_open` converts the `View` back into a dict but it would be good to verify the payload

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -337,7 +338,51 @@ def build_resolve_modal_payload(self, callback_id):
...
+                        "type": "plain_text",
+                        "text": text,
+                        "emoji": True,
+                    },
+                    "value": value,
+                }
+            )
+
+        return View(
+            **{
```
RyanSkonnord, line 356: Rather than format this with `**`, could we do
```python
View(
    type="modal",
    title=...
)
```
?

If for some reason this is preferable, we can cut down on indentation by extracting the dict such as
```python
kwargs = {
    "type": "modal",
    ...
}
return View(**kwargs)
```

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -337,7 +338,51 @@ def build_resolve_modal_payload(self, callback_id):
...
+            formatted_resolve_options.append(
+                {
+                    "text": {
+                        "type": "plain_text",
+                        "text": text,
+                        "emoji": True,
+                    },
+                    "value": value,
+                }
+            )
```
RyanSkonnord, line 353: Consider rewriting with a comprehension:

```suggestion
        formatted_resolve_options = [
            {
                "text": {
                    "type": "plain_text",
                    "text": text,
                    "emoji": True,
                },
                "value": value,
            }
            for text, value in RESOLVE_OPTIONS.items()
        ]
```

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -382,6 +427,53 @@ def build_archive_modal_payload(self, callback_id):
             "callback_id": callback_id,
         }
 
+    def build_archive_modal_payload(self, callback_id) -> View:
```
RyanSkonnord, line 430: Let's look for ways to reduce code duplicated from `build_resolve_modal_payload`. Maybe we can extract a few helper functions.

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -399,32 +491,50 @@ def open_resolve_dialog(self, slack_request: SlackActionRequest, group: Group) -
...
+            except ApiError as e:
+                logger.exception(
+                    "slack.action.response-error",
+                    extra={
+                        "error": str(e),
+                        "organization_id": org.id,
+                        "integration_id": slack_request.integration.id,
+                        "trigger_id": slack_request.data["trigger_id"],
+                        "dialog": "resolve",
+                    },
```
RyanSkonnord, line 536: We should reduce some of the duplication between the two `except` blocks if we can. The common elements of `extra` could be extracted into a local variable before entering the if-else block (I think we're doing this elsewhere with variables named `logger_params`).

If we want to add an extra `"error"` value in the `ApiError` case, an easy way to write it is `dict(error=str(e), **logger_params)`.

Also, is there a particular reason we're capturing `str(e)` here but not from the `SlackApiError`? Does the `SlackApiError` not come with a meaningful message? Or are they too verbose? (If it's that thing where they include an entire stack trace jammed into the message field, maybe we could trim from the first line break like we're doing elsewhere. And also maybe create a centralized utility for that sort of thing, if we're doing it a lot.)

iamrajjoshi: This logic will eventually be deleted so I am not really changing much here.

iamrajjoshi: logger.exception should automatically grab the exception for us.

RyanSkonnord: Okay.

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -382,6 +383,78 @@ def build_archive_modal_payload(self, callback_id):
...
+
+    def build_archive_modal_payload(self, callback_id: str) -> View:
+        return self.build_modal_payload(
+            title="Archive",
+            action_text="Archive",
+            options=ARCHIVE_OPTIONS,
+            initial_option_text="Until escalating",
+            initial_option_value="ignored:archived_until_escalating",
+            callback_id=callback_id,
+        )
```
RyanSkonnord, line 456: Nice!

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -382,6 +383,78 @@ def build_archive_modal_payload(self, callback_id):
             "callback_id": callback_id,
         }
 
+    def build_format_options(self, options: dict[str, str]) -> list[dict[str, dict[str, str]]]:
```
RyanSkonnord, line 386: The return type hint should be `list[dict[str, Any]]`, I think. (The dict at `"text"` has a `bool` as one of its values, and `"value"` maps to a `str` rather than a `dict`.)

RyanSkonnord (approved):
LGTM, just one small note about type hints.

PR #73750, opened 2024-07-03
fix(icon-audit): remove unused icons

doralchan:
**Removed:**

IconFrozen - Was never used for frozen frames after all (Not added to stack trace)
IconOption - Was never used for search after all (Not added for a keyboard shortcut)
IconRobot - Was never used for LLM monitoring after all (Not added as a primary sidebar item)


colin-sentry (approved):
I'll miss you, mister robot.

PR #73760, opened 2024-07-03
fix(similarity): Call seer delete record before hash deletion

jangjodi:
Before:
- Call to delete seer records was only called [here](https://github.com/getsentry/sentry/pull/new/jodi/fix-delete-records-by-hash), however hashes corresponding to these groups are already deleted [here](https://github.com/getsentry/sentry/compare/jodi/fix-delete-records-by-hash?expand=1#diff-c6570c132b6f37f9f29502555693ca55c2440059eda7edc4bce455ddf607bdc1L49) if called by the group deletion endpoint
- This only worked for cases where group deletion was called from elsewhere

Now:
- Call delete seer records before we delete the hashes in the group deletion endpoint

### DIFF
--- a/src/sentry/tasks/delete_seer_grouping_records.py
+++ b/src/sentry/tasks/delete_seer_grouping_records.py
@@ -58,7 +58,8 @@ def call_delete_seer_grouping_records_by_hash(
             "calling seer record deletion by hash",
             extra={"project_id": project.id, "hashes": group_hashes},
         )
-        delete_seer_grouping_records_by_hash.apply_async(args=[project.id, group_hashes, 0])
+        if group_hashes:
+            delete_seer_grouping_records_by_hash.apply_async(args=[project.id, group_hashes, 0])
 
 
 @instrumented_task(
--- a/tests/sentry/api/helpers/test_group_index.py
+++ b/tests/sentry/api/helpers/test_group_index.py
@@ -5,6 +5,7 @@ import pytest
 from django.http import QueryDict
 
 from sentry.api.helpers.group_index import update_groups, validate_search_filter_permissions
+from sentry.api.helpers.group_index.delete import delete_groups
 from sentry.api.helpers.group_index.update import (
     handle_assigned_to,
     handle_has_seen,
@@ -18,6 +19,7 @@ from sentry.models.activity import Activity
 from sentry.models.group import Group, GroupStatus
 from sentry.models.groupassignee import GroupAssignee
 from sentry.models.groupbookmark import GroupBookmark
+from sentry.models.grouphash import GroupHash
 from sentry.models.groupinbox import GroupInbox, GroupInboxReason, add_group_to_inbox
 from sentry.models.groupseen import GroupSeen
 from sentry.models.groupshare import GroupShare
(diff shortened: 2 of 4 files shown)


src/sentry/tasks/delete_seer_grouping_records.py
```diff
@@ -58,7 +58,8 @@ def call_delete_seer_grouping_records_by_hash(
             "calling seer record deletion by hash",
             extra={"project_id": project.id, "hashes": group_hashes},
         )
-        delete_seer_grouping_records_by_hash.apply_async(args=[project.id, group_hashes, 0])
+        if group_hashes:
+            delete_seer_grouping_records_by_hash.apply_async(args=[project.id, group_hashes, 0])
```
lobsterkatie, line 62: Just curious: in what scenario would there ever not be hashes?

tests/sentry/api/helpers/test_group_index.py
```diff
@@ -1055,3 +1057,61 @@ def test_user_in_reassigned_team(self):
...
+            len(GroupHash.objects.filter(project_id=self.project.id, group_id__in=group_ids).all())
+            == 0
+        )
+        assert (
+            len(GroupInbox.objects.filter(project_id=self.project.id, group_id__in=group_ids).all())
+            == 0
+        )
+        assert send_robust.called
+
+    @with_feature("projects:similarity-embeddings-delete-by-hash")
```
lobsterkatie, line 1088: What's the thinking behind making this a separate flag (rather than relying on the similar issues flag, for example)?

tests/sentry/api/helpers/test_group_index.py
```diff
@@ -1055,3 +1057,61 @@ def test_user_in_reassigned_team(self):
...
+        )
+        assert (
+            len(GroupInbox.objects.filter(project_id=self.project.id, group_id__in=group_ids).all())
+            == 0
+        )
+        assert send_robust.called
+        mock_logger.info.assert_called_with(
+            "calling seer record deletion by hash",
+            extra={"project_id": self.project.id, "hashes": hashes},
+        )
```
lobsterkatie, line 1117: In addition to asserting on the logger, it might be be worthwhile to assert on the task being called.

lobsterkatie (approved):
LGTM!

src/sentry/tasks/delete_seer_grouping_records.py
jangjodi: this shouldn't happen anymore, and I'll remove this check once I confirm that the log above never shows no hashes

tests/sentry/api/helpers/test_group_index.py
jangjodi: I wanted to be able to test this separately first

tests/sentry/api/helpers/test_group_index.py
jangjodi: thanks! I'll follow up with this after just because I want to get this deployed faster

jangjodi: just followed up [here](https://github.com/getsentry/sentry/pull/73790)!

PR #73763, opened 2024-07-03
feat(replays): Add tooltip to replay timeline gaps

c298lee:
Adds tooltip to the timeline gap. Also prevents you from clicking on the timeline when there are gaps (only applies to timeline not scrubber)

image

### DIFF
--- a/static/app/components/replays/breadcrumbs/replayTimeline.tsx
+++ b/static/app/components/replays/breadcrumbs/replayTimeline.tsx
@@ -68,13 +68,13 @@ export default function ReplayTimeline() {
       >
         <MinorGridlines durationMs={durationMs} width={width} />
         <MajorGridlines durationMs={durationMs} width={width} />
+        <TimelineScrubber />
         <TimelineGaps
           durationMs={durationMs}
           frames={appFrames}
           totalFrames={chapterFrames.length}
           width={width}
         />
-        <TimelineScrubber />
         <TimelineEventsContainer>
           <ReplayTimelineEvents
             durationMs={durationMs}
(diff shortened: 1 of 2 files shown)


static/app/components/replays/breadcrumbs/replayTimeline.tsx
```diff
@@ -68,13 +68,13 @@ export default function ReplayTimeline() {
       >
         <MinorGridlines durationMs={durationMs} width={width} />
         <MajorGridlines durationMs={durationMs} width={width} />
+        <TimelineScrubber />
         <TimelineGaps
           durationMs={durationMs}
           frames={appFrames}
           totalFrames={chapterFrames.length}
           width={width}
         />
```
ryan953, line 77: This this re-ordering, just want to confirm that the Timeline can still be clicked on to jump to a specific timestamp? 

Can't click on the spots where the gray box is rendered, those eat clicks. But where no gray box is rendered clicks should still get handled by the TimelineScrubber

c298lee: Yes!

static/app/components/replays/breadcrumbs/timelineGaps.tsx
```diff
@@ -64,15 +66,23 @@ export default function TimelineGaps({durationMs, frames, totalFrames, width}: P
...
+const Column = styled(Timeline.Col)<{column: number}>`
   grid-column: ${p => p.column};
-  background: ${p => p.theme.gray400};
   line-height: 14px;
+`;
+
+const Gap = styled('span')`
+  background: ${p => p.theme.gray400};
   opacity: 16%;
+  height: 20px;
```
ryan953, line 87: this might work with `height: 100%` but i'm not sure, haven't gone to check.

ryan953 (approved):
Tooltips! looks good, come comments to explore or iterate on later, nbd either way.

static/app/components/replays/breadcrumbs/timelineGaps.tsx
c298lee: it does not :(

PR #73764, opened 2024-07-03
feat(insights): add hook for insights upsell page

DominikB2014:
Adds the hook for the insights upsell page and exports a few items that will be used in https://github.com/getsentry/getsentry/pull/14519

As https://github.com/getsentry/getsentry/pull/14519 is not merged, the upsell page won't work and no visual changes will occur as a result of this PR only. But we need to merge this one in first so that getsentry can consume it.

Once this PR and https://github.com/getsentry/getsentry/pull/14519 are merged, will also create a PR to remove the upsell tooltips


static/app/views/insights/common/components/modulePageProviders.tsx
```diff
@@ -30,19 +31,42 @@ export function ModulePageProviders({moduleName, pageTitle, children, features}:
...
+  'mobile-ui': 'performance-mobile-ui',
+  cache: 'performance-cache',
+  db: 'performance-database',
+  http: 'performance-http',
+  resource: 'performance-browser-resources',
+  screen_load: 'performance-mobile-screens',
+  app_start: 'performance-mobile-app-startup',
+  vital: 'performance-webvitals',
+  queue: 'performance-queues',
+};
```
gggritso, line 63: 🤔 since these are full page upsells, do you need to bother with this mapping and lookup? You don't need to pass these a sidebar ID, you can just pass through the module name, no?

static/app/types/hooks.tsx
```diff
@@ -658,6 +659,18 @@ type SidebarNavigationItemHook = () => React.ComponentType<{
...
+/**
+ * Insights upsell hook takes in a insights sidebar item id
+ * and (if applicable) passes in the module specific upsell page as props to its children.
+ */
+type InsightsUpsellHook = () => React.ComponentType<{
+  children: (opts: {
+    disabled: boolean;
+    upsellPage: React.ReactElement | null;
+  }) => React.ReactElement;
+  id: string;
```
gggritso, line 671: Out of curiosity, does this need to use the render props approach? Are there props that have to be passed down, or is it possible to use a "regular" hook that just substitutes in other content?

gggritso (commented):
A few questions! Treading carefully since hooks are confusing

static/app/views/insights/common/components/modulePageProviders.tsx
DominikB2014: You're right, I could just pass in a module name, and that would probably be my preference. 
But to share logic with the sidebar item hook, we'll have to map ids to module names (or vise versa) somewhere. 

Perhaps I do the mapping in getsentry and pass in module name here for a better api?

static/app/types/hooks.tsx
DominikB2014: I'll verify this more tmrw. But I believe the regular disabled feature hooks don't allow you to pass anything into them. In our case we want to pass in the module name to show the appropriate upsell page.

static/app/views/insights/common/components/modulePageProviders.tsx
DominikB2014: I've updated to accept moduleName instead!

static/app/types/hooks.tsx
DominikB2014: Ok i dug around, its possible with a "component hook", made the changes.

gggritso (approved):
👍🏻

PR #73769, opened 2024-07-03
ref(related_issues): Fix naming

armenzg:
My original code uses incorrect naming (subtitle <-> message).

This is a no-op change.

Issue Details ([code](https://github.com/getsentry/sentry/blob/fb4c0e3df5e80c06a0397fcaa4774477f43d1adc/static/app/components/eventOrGroupHeader.tsx#L126-L133), [code](https://github.com/getsentry/sentry/blob/fb4c0e3df5e80c06a0397fcaa4774477f43d1adc/static/app/components/eventOrGroupHeader.tsx#L61-L69) & [code](https://github.com/getsentry/sentry/blob/fb4c0e3df5e80c06a0397fcaa4774477f43d1adc/static/app/components/eventOrGroupTitle.tsx#L32)):
image

Trace Issue Event:
image

### DIFF
--- a/static/app/views/issueDetails/traceTimeline/traceIssue.tsx
+++ b/static/app/views/issueDetails/traceTimeline/traceIssue.tsx
@@ -19,7 +19,7 @@ export function TraceIssueEvent({event}: TraceIssueEventProps) {
   const organization = useOrganization();
   const project = useProjectFromSlug({organization, projectSlug: event['project.name']});
   const issueId = event['issue.id'];
-  const {title, subtitle} = getTitleSubtitle(event);
+  const {title, subtitle, message} = getTitleSubtitleMessage(event);
   const avatarSize = parseInt(space(4), 10);
 
   const referrer = 'issue_details.related_trace_issue';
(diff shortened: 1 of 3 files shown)


static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -59,26 +59,29 @@ export function TraceIssueEvent({event}: TraceIssueEventProps) {
         <TraceIssueDetailsContainer>
           <NoOverflowDiv>
             <TraceIssueEventTitle>{title}</TraceIssueEventTitle>
-            <TraceIssueEventTransaction>{event.transaction}</TraceIssueEventTransaction>
+            <TraceIssueEventSubtitle>{subtitle}</TraceIssueEventSubtitle>
```
armenzg, line 62: Right now, I'm making `subtitle` be `event.transaction` to prevent changing anything but the true source is `culprit` which we don't fetch

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -59,26 +59,29 @@ export function TraceIssueEvent({event}: TraceIssueEventProps) {
...
-          <NoOverflowDiv>{subtitle}</NoOverflowDiv>
+          <NoOverflowDiv>{message}</NoOverflowDiv>
         </TraceIssueDetailsContainer>
       </TraceIssueLinkContainer>
     </Fragment>
   );
 }
 
-function getTitleSubtitle(event: TimelineEvent) {
+function getTitleSubtitleMessage(event: TimelineEvent) {
```
vartec, line 71:
```suggestion
function getTitleSubtitleAndMessage(event: TimelineEvent) {
```

vartec (approved):

Issue #59320, opened 2023-11-02
Expand client outcomes on the stats page


cleptric:
Today, we only expose three categories to our users.

<img width="360" alt="Screenshot 2023-11-02 at 22 39 38" src="https://github.com/getsentry/sentry/assets/6617432/b454d73e-b1ce-450d-8500-0a5866fdd7a2">

This makes debugging issues with the SDKs too cumbersome.
We should expose some stats we show on `_admin`.

<img width="549" alt="Screenshot 2023-11-02 at 22 41 05" src="https://github.com/getsentry/sentry/assets/6617432/df617d57-1fb5-420b-8a23-f56209bccfef">

cleptric:
In https://github.com/getsentry/sentry/issues/67801, there was an ask to set up alerts on certain outcomes. We should consider this as well.

matejminar:
Related: https://github.com/getsentry/sentry/issues/42940

PR #73774, opened 2024-07-03
feat(stats): Display subseries in tooltip

priscilawebdev:
### Goal

On the stats page, we aim to provide more detailed information about the statistics. Currently, we display the outcomes: Accepted, Filtered, and Dropped, but we should also include specifics about what was filtered and what was dropped.

**Before:**

image

**After:**
image

### Note

Discarded will be included in a follow-up

Part of https://github.com/getsentry/sentry/issues/59320


scttcper (approved):

static/app/views/organizationStats/usageStatsOrg.tsx
```diff
@@ -115,7 +117,7 @@ class UsageStatsOrganization<
     return {
       ...queryDatetime,
       interval: getSeriesApiInterval(dataDatetime),
-      groupBy: ['category', 'outcome'],
+      groupBy: ['category', 'outcome', 'reason'],
```
scttcper, line 120: does this need to be behind a feature flag?

priscilawebdev: If I recall correctly, no, but let me check again to confirm.

vgrozdanic (approved):

matejminar (approved):
Looks good to me.

Could we do something about this huge tooltip? Perhaps only show non-zero values?
CleanShot 2024-07-04 at 14 45 14@2x

priscilawebdev:
@matejminar sure thing! I created the [issue](https://github.com/getsentry/sentry/issues/73829) and will work on it in a follow-up 😉

PR #73801, opened 2024-07-03
fix(replay): Include timestamps in the useExtractedPageHtml cache key

ryan953:


ryan953:
Duplicate of https://github.com/getsentry/sentry/pull/73752

PR #51074, opened 2023-06-15
feat(mute-metric-alerts): Add mute button to metric alert email

roggenkemper:
this pr adds in the mute button to the metric alert email . It attached a query parameter to the email which is used to automatically mute the alert if clicked.
Screenshot 2023-06-15 at 1 05 11 PM

PR #47650, opened 2023-04-19
feat(rulesnooze): mute alert given query parameter

roggenkemper:
this pr adds in the ability to mute an alert if the query parameter `mute` is passed in. We will add the query parameter to a link in the email which if a user clicks it, will bring them to the alert details page and automatically mute the alert for them. 

closes https://github.com/getsentry/sentry/issues/47508

PR #47361, opened 2023-04-13
feat(rulesnooze): Adds mute email link to issue alert email

roggenkemper:
this pr adds in the link to mute alerts to the issue alert email and slightly updates the banner of the issue alert email.

Banner Update: 
Screenshot 2023-04-17 at 2 51 45 PM

With Mute Alert Link: 
Screenshot 2023-04-17 at 2 54 19 PM

Closes https://github.com/getsentry/sentry/issues/46833

PR #73803, opened 2024-07-03
style(alerts): Change copy of mute alert link

### LINKED ISSUE #69792: Clarify Sentry alert email - Mute this alert button - confusion
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

1. create an alert that send an email
2. trigger that alert
3. view email

### Expected Result

The request and expectation here is that the Sentry Alert emails be more clear what the mute this alert button does. Or another acceptable solution is that the action is done completely on the sentry UI, so it is clear and straightforward to undo.

### Actual Result

We have had several user click the 'mute this alert button' in Sentry alert emails for an issue believing that this would mute the alerting for the specific issue they had received the email for only to later find out that all alerts have been muted for everyone.

image

The email isn't that clear, what the intension is. Are you muting for yourself only, or for everyone. Are you muting the specific issue or the alert itself and thus multiple issues.

If you go to the alert page you have options, but that is not clear on the email.
image

If you click the "mute this alert" button you do get redirect to the alert page. And it's only then (after the button is pressed) do you discover what you did via a banner. We have found that the banner isn't sufficient for most users to realize that their action and intention do not match.

### Product Area

Alerts

### Link

_No response_

### DSN

_No response_

### Version

_No response_

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2160) by [Unito](https://www.unito.io)

mifu67:
Change the link text for muting email alerts from "Mute this alert" to "Mute this alert for everyone."

Fixes GH-69792

### DIFF
--- a/src/sentry/templates/sentry/emails/digests/body.html
+++ b/src/sentry/templates/sentry/emails/digests/body.html
@@ -34,7 +34,7 @@
       <div class="container">
         {% with rule_details=rules_details|get_item:rule.id snooze_alert_url=snooze_alert_urls|get_item:rule.id %}
         {% if snooze_alert %}
-            <a class="mute" href="{% absolute_uri snooze_alert_url %}">Mute this alert</a>
+            <a class="mute" href="{% absolute_uri snooze_alert_url %}">Mute this alert for everyone</a>
         {% endif %}
             This email was triggered by <a href="{% absolute_uri rule_details.status_url %}">{{ rule_details.label }}</a>
         {% endwith %}
--- a/src/sentry/templates/sentry/emails/feedback.html
+++ b/src/sentry/templates/sentry/emails/feedback.html
@@ -191,7 +191,7 @@
 
     <p class="info-box">
       {% if snooze_alert %}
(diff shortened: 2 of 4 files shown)


schew2381 (commented):
lgtm!

schew2381 (approved):

ceorourke:
Hi I'm sorry I didn't see this before, but this text change is incorrect. You can see the original implementation [here](https://github.com/getsentry/sentry/pull/47650/files) for the front end, and [here](https://github.com/getsentry/sentry/pull/47361/files) for the issue alert email. ([digests](https://github.com/getsentry/sentry/pull/48044/files) and [metric alerts](https://github.com/getsentry/sentry/pull/51074/files)) We pass the query param `mute:1` in the link and the front end reads that as "mute for me". 

I don't really know what the person who opened that GH issue is talking about, it would need more investigation. As far as I can see this hasn't changed since it was originally added so all emails would have the `mute:1` query parameter and only mute for the person clicking the link in their email.

PR #73807, opened 2024-07-03
feat(seer grouping): Call seer deletion task when parent group not found

lobsterkatie:
In theory, the Seer database should never contain outdated hashes (that is, hashes whose corresponding group no longer exists on the Sentry side), because group deletion in Sentry triggers a call to the hash deletion endpoint in Seer. That said, things happen.

If/when they do, one potential outcome is that the Seer similarity endpoint will respond with an outdated hash rather than the closest still-existing one. In case this happens, we want to let Seer know so it can delete the hash. This PR therefore adds a call to the Seer hash-deletion task which will fire whenever `get_similarity_data_from_seer` encounters a `SimilarGroupNotFoundError`.

### DIFF
--- a/src/sentry/seer/similarity/similar_issues.py
+++ b/src/sentry/seer/similarity/similar_issues.py
@@ -13,6 +13,7 @@ from sentry.seer.similarity.types import (
     SimilarGroupNotFoundError,
     SimilarIssuesEmbeddingsRequest,
 )
+from sentry.tasks.delete_seer_grouping_records import delete_seer_grouping_records_by_hash
 from sentry.utils import json, metrics
 from sentry.utils.json import JSONDecodeError, apply_key_filter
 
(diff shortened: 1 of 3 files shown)


JoshFerge (approved):

src/sentry/seer/similarity/similar_issues.py
```diff
@@ -200,19 +201,19 @@ def get_similarity_data_from_seer(
...
-            # Sentry side) comes with an inherent slight delay - the request to Seer happens async, and
-            # requests between services aren't instantaneous. It's therefore possible for us to land
-            # in a race condition, where Seer recommends a hash it doesn't know it's about to be
-            # asked to delete, and we come up empty because on the Sentry side the hash has already
-            # been deleted.
             parent_hash = raw_similar_issue_data.get("parent_hash")
 
+            # Tell Seer to delete the hash from its database, so it doesn't keep suggesting a group
+            # which doesn't exist
+            delete_seer_grouping_records_by_hash.apply_async(args=[project_id, [parent_hash]])
```
vartec, line 208: Purely stylistic choice, feel free to ignore.

Personally I prefer `delay()` over `apply_async()` because it looks exactly like calling the original function and can be easily refactored to add or remove Celery by inserting or removing `.delay` between function name and its params. Also, because it preserves the signature, works better with MyPy.

vartec (approved):

src/sentry/seer/similarity/similar_issues.py
jangjodi, line 208: Could you add the feature check for `projects:similarity-embeddings-delete-by-hash` before making the call here? The reason I didn't have it in the function is because it calls itself and I didn't want it to be doing checks every time

lobsterkatie: Oh, good call. I was pattern-matching off of our other calls to `delete_seer_grouping_records_by_hash`, but what you're saying makes a lot of sense. I'll change it.

lobsterkatie: UPDATE: Had a convo IRL. The project itself isn't in scope in this function, and while we have the id, if we can avoid an extra DB call I'd like to, given this runs during ingest. So we agreed that for now I'm not going to add the flag check but also will wait to merge this until we've given the deletion a chance to run on our internal projects for a few days, to make sure it feels safe to make the deletion call without checking the flag.

PR #73810, opened 2024-07-04
feat(stats): Expose discarded outcome

### LINKED ISSUE #59320: Expand client outcomes on the stats page

priscilawebdev:
### Goal

Expose the discarded outcome on the stats page, so users can better debug.

**Before:**
image

**After:**
image

closes https://github.com/getsentry/sentry/issues/59320


static/app/views/organizationStats/usageStatsOrg.tsx
```diff
@@ -427,6 +440,24 @@ class UsageStatsOrganization<
             return;
           }
 
+          if (outcome === Outcome.CLIENT_DISCARD) {
+            usageStats[i].discarded += stat;
```
vgrozdanic, line 444: Nit, but due to consistency with other part of code inside the function maybe we could also use: `usageStats[i][outcome] += stat;` here.

On the next refactor Friday, we can refactor this function then, cause it seems to me that it can be written more concise, but that is outside of the scope of this PR :)

priscilawebdev: you have good eyes :) but we cannot do that, because we get 'client_discarded' from the backend and in the frontend we are naming it 'discarded'

priscilawebdev: the function can for sure be improved but as you said it is outside of the scope of this PR  - maybe in a follow-up

vgrozdanic: My bad, I missed that, makes sense now why are we using it like this, thanks :)

vgrozdanic (approved):

matejminar (approved):
I'd prefer this card to be last.
Accepted -> Filtered -> Dropped -> Discarded
Basically descending order of what's happening.

priscilawebdev:
closing this PR in favor of https://github.com/getsentry/sentry/pull/73869

PR #73812, opened 2024-07-04
Delete most of the Reprocessing code

Swatinem:
This removes tons of legacy reprocessing-related code. It does keep some of the API endpoints around and rather returns dummy results for now.


loewenheim (approved):

src/sentry/notifications/notifications/activity/__init__.py
```diff
@@ -3,9 +3,6 @@
 from sentry.notifications.notifications.activity.assigned import AssignedActivityNotification
 from sentry.notifications.notifications.activity.base import ActivityNotification
 from sentry.notifications.notifications.activity.escalating import EscalatingActivityNotification
-from sentry.notifications.notifications.activity.new_processing_issues import (
-    NewProcessingIssuesActivityNotification,
```
nhsiehgit, line 7: Can we actually delete this entire class as well then?

Looks like we're only deleting the imports and the template, but not the class nor the tuple enum? (maybe we can add a comment there that this is no longer used?)

nhsiehgit (approved):
rubber stamp 👍 

although 1 quick question

anonrig (approved):

src/sentry/notifications/notifications/activity/__init__.py
Swatinem: Will do, I believe I missed that and picked it up in the followup #73912, though I might just merge those two PRs together.

PR #73816, opened 2024-07-04
fix(metrics) Separate tags and attributes options

vgrozdanic:
Fix for a bug when span attributes which had extraction rule were showing as disabled in tags section.

## Before
image

## After
image

Order of items is different in images, but it can be seen that `browser.name` is not disabled anymore

### DIFF
--- a/static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
+++ b/static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
@@ -52,6 +52,7 @@ const KNOWN_NUMERIC_FIELDS = new Set([
   SpanIndexedField.MESSAGING_MESSAGE_BODY_SIZE,
   SpanIndexedField.MESSAGING_MESSAGE_RECEIVE_LATENCY,
   SpanIndexedField.MESSAGING_MESSAGE_RETRY_COUNT,
+  SpanIndexedField.TRANSACTION_ID,
 ]);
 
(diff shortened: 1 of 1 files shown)


static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
```diff
@@ -52,6 +52,7 @@ const KNOWN_NUMERIC_FIELDS = new Set([
   SpanIndexedField.MESSAGING_MESSAGE_BODY_SIZE,
   SpanIndexedField.MESSAGING_MESSAGE_RECEIVE_LATENCY,
   SpanIndexedField.MESSAGING_MESSAGE_RETRY_COUNT,
+  SpanIndexedField.TRANSACTION_ID,
```
vgrozdanic, line 55: Added this as a numeric field so that we don't show it as a tag option later in code when we filter out all the numeric fields

ArthurKnaus: maybe rename the set? `HIGH_CARDINALITY_TAGS`

ArthurKnaus (approved):
thx 🚀

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
vgrozdanic: Renamed, thanks for suggestion

PR #73822, opened 2024-07-04
Remove the "Processing Issues" settings page / hint

Swatinem:
As part of deprecating legacy reprocessing, this completely removes the "Processing Issues" page from settings, as well as the Processing Issues Hint from the issues stream.

It leaves other parts of the issues stream untouched, which I believe are still in-use for "reprocessing2".

---

I’m a bit unsure about completely removing the settings page directly. I believe users who still have links floating around might hit a 404 error? Not sure if its worth the effort to display a "this feature has been removed" page instead?


armenzg (approved):
For anyone coming along afterward, this removes this page: https://sentry.sentry.io/settings/projects/sentry/processing-issues/
which was already marked as deprecated and making a reference to this doc: https://docs.sentry.io/product/issues/reprocessing/
image

PR #73831, opened 2024-07-04
feat: Update Loader script to be v8 compatible

mydea:
This updates the Loader Script to ensure we also check for an already installed instance of the SDK under v8, where the global carrier `window.__SENTRY__` has a different shape.

Part of https://github.com/getsentry/sentry-javascript/issues/12187

* Tests for v8: https://github.com/getsentry/sentry-javascript/pull/12764
* Tests for v7: https://github.com/getsentry/sentry-javascript/pull/12765


src/sentry/templates/sentry/js-sdk-loader.ts
```diff
@@ -1,5 +1,5 @@
 declare const __LOADER__PUBLIC_KEY__: any;
-declare const __LOADER_SDK_URL__: any;
+declare const __LOADER__SDK_URL__: any;
```
mydea, line 2: This is not directly related but noticed that this was inconsistent before!

s1gr1d (approved):

lforst (approved):

PR #73835, opened 2024-07-04
feat: Allow to select v8 in Loader Script

mydea:
This adds a feature flag `organizations:js-sdj-loader-v8` that can be set that will add `8.x` to the selectable options for the Loader Script.

Part of https://github.com/getsentry/sentry-javascript/issues/12187

### DIFF
--- a/src/sentry/loader/browsersdkversion.py
+++ b/src/sentry/loader/browsersdkversion.py
@@ -8,6 +8,7 @@ from django.conf import settings
 from packaging.version import Version
 
 import sentry
+from sentry import features
 
 logger = logging.getLogger("sentry")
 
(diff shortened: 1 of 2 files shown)


src/sentry/loader/browsersdkversion.py
s1gr1d, line 1: Do we need to change line 73 in this file? It think is not giving back v8.

lforst: That is by design :) We don't want to put people who are currently on 7.x on 8.x because that may break them.

lforst (approved):
The way the feature flag is set up will not prevent anybody to manually put themselve onto v8 but that's probably fine.

mydea:
> The way the feature flag is set up will not prevent anybody to manually put themselve onto v8 but that's probably fine.

Jup, but I think this is OK - honestly if somebody does that manually, any potential problems (that we don't really forsee anyhow) are their problem 😅

PR #73840, opened 2024-07-04
feat(inbound-filters): Move filters to generic filters

### LINKED ISSUE #2850: Inbound Filter Outcome Reason
**Description**
We want to be able to categorize why errors dropped by a serverside inound filter are dropped. In this specific case we are trying to analyze the full impact of filtering out hydration errors, but it maybe usefuly to do this in a more scalable way

image

[link](https://sentryio.cloud.looker.com/explore/super_big_facts/outcomes_v3?qid=G287w19xEvHJKFC4g4n312&toggle=fil)

If the filter is based on hyrdation error. Is this included in the "error message" reason? If yes could that be added to the outcome reason? error "message: hydration"

the in product toggle found in project inbound filter settings

image

iambriccardo:
This PR ports the first two inbound filters to the new generic filters implementation, namely chunk load errors and hydration errors.

The definition of the new generic filter is very close to the previous hardcoded definition in Relay (for the `errorMessages` inbound filter); however, it differs in the matching of the `ty` and `value`. In Relay, there is an additional step in which, if both `ty` and `value` are set, Relay will try to match the string in the form `"{ty}: {value}"`, which is currently not expressible with `RuleCondition` constructs.

Our plan is to roll out this implementation and check how many occurrences of the error messages are still going through the old filter and act accordingly.

Closes: https://github.com/getsentry/relay/issues/2850


src/sentry/ingest/inbound_filters.py
```diff
@@ -290,3 +297,107 @@ class _LegacyBrowserFilterSerializer(_FilterSerializer):
...
+    """
+    generic_filters: list[GenericFilter] = []
+
+    for generic_filter_id, generic_filter_fn in ACTIVE_GENERIC_FILTERS:
+        if project.get_option(f"filters:{generic_filter_id}") not in ("1", True):
+            continue
+
+        generic_filter = generic_filter_fn()
+        if generic_filter is not None:
+            generic_filter["id"] = generic_filter_id
```
iambriccardo, line 394: Decided to put the `id` here. A bit risky but reduces duplication and arguments of the closures.

src/sentry/ingest/inbound_filters.py
```diff
@@ -290,3 +297,106 @@ class _LegacyBrowserFilterSerializer(_FilterSerializer):
...
+    419 - The server could not finish this Suspense boundary, likely due to an error during server rendering.
+        Switched to client rendering.
+    422 - There was an error while hydrating this Suspense boundary. Switched to client rendering.
+    423 - There was an error while hydrating. Because the error happened outside of a Suspense boundary, the entire
+        root will switch to client rendering.
+    425 - Text content does not match server-rendered HTML.
+    """
+    values = [
+        ("*", "*https://reactjs.org/docs/error-decoder.html?invariant={418,419,422,423,425}*"),
+        ("*", "*https://react.dev/errors/{418,419,422,423,425}*"),
```
jan-auer, line 359: Ideally, in this case we omit the check for `event.exceptions.ty`.

jan-auer (approved):

PR #73843, opened 2024-07-04
feat(profiling): add endpoint to generate flamegraph from span group for continuous profiling

viglia:
depends on https://github.com/getsentry/vroom/pull/482

**Request**:
```bash

curl --location 'http://127.0.0.1:8000/api/0/organizations/sentry/profiling/chunks-flamegraph/?project=1&span_group=e55389c43689f2c5&statsPeriod=2d&utc=false'
````

**Body Response**: here in this file an example of the response for the request above.
[body_response.json](https://github.com/user-attachments/files/16106426/flamegraph_response.json)

@JonasBa , for when you'll implement the UI: the flamegraph returned is the same exact struct we used for the older one. I didn't change it. 
You'll still find a field named `profile_ids`. 
In this case, though, this array contains a list of `chunk_ids` not `profile_ids` despite the name.
If this is confusing we can change it.

### DIFF
--- a/src/sentry/api/endpoints/organization_profiling_profiles.py
+++ b/src/sentry/api/endpoints/organization_profiling_profiles.py
@@ -16,9 +16,11 @@ from sentry.api.bases import NoProjects, OrganizationEventsV2EndpointBase
 from sentry.exceptions import InvalidSearchQuery
 from sentry.models.organization import Organization
 from sentry.profiles.flamegraph import (
+    get_chunks_from_spans_metadata,
     get_profile_ids,
     get_profile_ids_with_spans,
     get_profiles_with_function,
+    get_spans_from_group,
 )
 from sentry.profiles.profile_chunks import get_chunk_ids
 from sentry.profiles.utils import parse_profile_filters, proxy_profiling_service
@@ -122,4 +124,36 @@ class OrganizationProfilingChunksEndpoint(OrganizationProfilingBaseEndpoint):
                 "profiler_id": profiler_id,
                 "chunk_ids": [el["chunk_id"] for el in chunk_ids],
             },
+        )
+
+
+@region_silo_endpoint
+class OrganizationProfilingChunksFlamegraphEndpoint(OrganizationProfilingBaseEndpoint):
+    def get(self, request: Request, organization: Organization) -> HttpResponse:
+        if not features.has("organizations:profiling", organization, actor=request.user):
+            return Response(status=404)
+
+        params = self.get_snuba_params(request, organization, check_global_views=False)
+
+        project_ids = params.get("project_id")
+        if project_ids is None or len(project_ids) != 1:
+            raise ParseError(detail="one project_id must be specified.")
+
+        span_group = request.query_params.get("span_group")
+        if span_group is None:
+            raise ParseError(detail="span_group must be specified.")
+
+        spans = get_spans_from_group(
+            organization.id,
+            project_ids[0],
+            params,
+            span_group,
+        )
+
+        chunksMetadata = get_chunks_from_spans_metadata(organization.id, project_ids[0], spans)
+
+        return proxy_profiling_service(
+            method="POST",
+            path=f"/organizations/{organization.id}/projects/{project_ids[0]}/chunks-flamegraph",
+            json_data={"chunks_metadata": chunksMetadata},
         )
--- a/src/sentry/api/endpoints/organization_profiling_profiles.py
+++ b/src/sentry/api/endpoints/organization_profiling_profiles.py
@@ -65,7 +65,7 @@ class OrganizationProfilingFlamegraphEndpoint(OrganizationProfilingBaseEndpoint)
         if not features.has("organizations:continuous-profiling", organization, actor=request.user):
             return Response(status=404)
 
-        params = self.get_snuba_params(request, organization, check_global_views=False)
+        params = self.get_snuba_params(request, organization)
         project_ids = params["project_id"]
         if len(project_ids) > 1:
             raise ParseError(detail="You cannot get a flamegraph from multiple projects.")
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/sentry/api/endpoints/organization_profiling_profiles.py [hunk 1]
```diff
@@ -123,3 +125,35 @@ def get(self, request: Request, organization: Organization) -> HttpResponse:
                 "chunk_ids": [el["chunk_id"] for el in chunk_ids],
             },
         )
+
+
+@region_silo_endpoint
+class OrganizationProfilingChunksFlamegraphEndpoint(OrganizationProfilingBaseEndpoint):
+    def get(self, request: Request, organization: Organization) -> HttpResponse:
+        if not features.has("organizations:profiling", organization, actor=request.user):
```
Zylphrex, line 133:
```suggestion
        if not features.has("organizations:continuous-profiling", organization, actor=request.user):
```

src/sentry/api/endpoints/organization_profiling_profiles.py
```diff
@@ -123,3 +125,35 @@ def get(self, request: Request, organization: Organization) -> HttpResponse:
...
+            return Response(status=404)
+
+        params = self.get_snuba_params(request, organization, check_global_views=False)
```
Zylphrex, line 136: Why is this turning global views off here

src/sentry/api/endpoints/organization_profiling_profiles.py
```diff
@@ -123,3 +125,35 @@ def get(self, request: Request, organization: Organization) -> HttpResponse:
...
+
+        project_ids = params.get("project_id")
```
Zylphrex, line 138: `project_id` is not passed for my projects. This should check params to see if it resolved to a single project or be a projects endpoint.

src/sentry/api/endpoints/organization_profiling_profiles.py
viglia: honestly, this was just carried over from the other 2 endpoints defined in `organization_profiling_profiles.py`, which also define `check_global_views=False`. Should the default (`True`) be used instead?

src/sentry/api/endpoints/organization_profiling_profiles.py
viglia: @Zylphrex marking this as resolved as we've discussed and sorted this out this over Slack

src/sentry/profiles/flamegraph.py
```diff
@@ -176,3 +189,172 @@ def extract_profile_ids() -> list[str]:
...
+                }
+            )
+        else:
+            spans[row["profiler_id"]] = [
+                {
+                    "active_thread_id": row["active_thread_id"],
+                    "start": row["start_timestamp_precise"],
+                    "end": row["end_timestamp_precise"],
+                }
+            ]
```
Zylphrex, line 279: Nit: You could use a defaultdict for this


```python
spans = defaultdict(list)
for row in data:
  spans[row["profiler_id"].append(...)
```

src/sentry/profiles/flamegraph.py
```diff
@@ -176,3 +189,172 @@ def extract_profile_ids() -> list[str]:
...
+            )
+        cond.append(
+            And(
+                [
+                    Condition(Column("profiler_id"), Op.EQ, profiler_id),
+                    Or(chunk_range_cond) if len(chunk_range_cond) >= 2 else chunk_range_cond[0],
+                ]
+            )
+        )
+    return [Or(cond)] if len(cond) >= 2 else cond
```
Zylphrex, line 312: This should be okay for 100 spans but know that there is a maximum query size so if we were to raise the number of spans in the future, you may need to split the query into smaller blocks.

src/sentry/profiles/flamegraph.py
```diff
@@ -176,3 +189,172 @@ def extract_profile_ids() -> list[str]:
...
+    )
+    data = raw_snql_query(
+        request,
+        referrer=Referrer.API_PROFILING_PROFILE_FLAMEGRAPH.value,
+    )["data"]
+    chunks = []
+    for row in data:
+        intervals = [
+            {
+                "start": str(int(datetime.fromisoformat(el["start"]).timestamp() * 10**9)),
```
Zylphrex, line 347: Are we still expecting timestamps in nanoseconds in vroom? I thought the continuous profiling chunks stored timestamps in seconds so there shouldn't be a need to do additional conversion here.

src/sentry/api/endpoints/organization_profiling_profiles.py
```diff
@@ -60,10 +62,10 @@ def get(self, request: Request, organization: Organization) -> HttpResponse:
 @region_silo_endpoint
 class OrganizationProfilingFlamegraphEndpoint(OrganizationProfilingBaseEndpoint):
     def get(self, request: Request, organization: Organization) -> HttpResponse:
-        if not features.has("organizations:profiling", organization, actor=request.user):
+        if not features.has("organizations:continuous-profiling", organization, actor=request.user):
```
Zylphrex, line 65: Was this change in feature flags intentional? This endpoint is for the transaction based profiles right?

src/sentry/api/endpoints/organization_profiling_profiles.py
Zylphrex: I think you updated the wrong feature flag check.

Zylphrex (commented):
It'd be nice to use dedicated referrers for the various queries so we can understand them better.

src/sentry/profiles/flamegraph.py
viglia: This is for the callTrees we generate with the chunks. Basically instead of copying and pasting a ton of code that would be exactly the same and that the only diff would be the timestamp, I'm reusing all the existing Calltree/Node/Flamegraph code. The price to pay for reusing all this stuff is this conversion though.

Zylphrex (approved):

src/sentry/profiles/flamegraph.py
viglia: yep, one more thing I want to do next, is to replace the fixed amount of 100 with an option as we did for the old flamegraph. Then I'll play with that value a bit to see how far can we stretch it before we have to split it as you suggest. 👍🏻

PR #73845, opened 2024-07-04
chore(features) Remove mobile-ttid-ttfd-contribution feature

markstory:
This feature flag has inconsistent results between flagr/flagpole and isn't actually used anywhere.


shruthilayaj (approved):

aldy505:
A note: @hubertdeng123 this should be removed on self-hosted (I was browsing the FF per usual)

PR #73846, opened 2024-07-04
chore(sdk): Pass current scope to thread executor

Zylphrex:
Same as #73482. We need to pass the current scope too or the spans will be lost.


markstory (approved):
Looks good to me.

PR #73848, opened 2024-07-04
ref(trace): Switch over to SnubaParams instead

wmak:
- Instead of using the old ParamsType, switch over to the typed SnubaParams

### DIFF
--- a/src/sentry/api/endpoints/organization_events_trace.py
+++ b/src/sentry/api/endpoints/organization_events_trace.py
@@ -491,7 +491,6 @@ def create_transaction_params(
     snuba_params: SnubaParams,
 ) -> SnubaParams:
     """Can't use the transaction params for errors since traces can be errors only"""
-    transaction_params = snuba_params.copy()
     query_metadata = options.get("performance.traces.query_timestamp_projects")
     sentry_sdk.set_tag("trace_view.queried_timestamp_projects", query_metadata)
     if not query_metadata:
@@ -535,6 +534,7 @@ def create_transaction_params(
     project_ids = list(project_id_set)
     # Reusing this option for now
     time_buffer = options.get("performance.traces.span_query_timebuffer_hours")
+    transaction_params = snuba_params.copy()
     if min_timestamp:
         transaction_params.start = min_timestamp - timedelta(hours=time_buffer)
     if max_timestamp:
--- a/src/sentry/api/endpoints/organization_events_trace.py
+++ b/src/sentry/api/endpoints/organization_events_trace.py
@@ -25,7 +25,7 @@ from sentry.models.group import Group
 from sentry.models.organization import Organization
 from sentry.search.events.builder.discover import DiscoverQueryBuilder
 from sentry.search.events.builder.spans_indexed import SpansIndexedQueryBuilder
-from sentry.search.events.types import ParamsType, QueryBuilderConfig
+from sentry.search.events.types import QueryBuilderConfig, SnubaParams
 from sentry.snuba import discover
 from sentry.snuba.dataset import Dataset
 from sentry.snuba.referrer import Referrer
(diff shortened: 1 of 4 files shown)


mjq (approved):
💯

src/sentry/search/events/types.py
```diff
@@ -113,6 +114,9 @@ def interval(self) -> float | None:
             return (self.end - self.start).total_seconds()
         return None
 
+    def copy(self) -> "SnubaParams":
```
Zylphrex, line 117: Nit: could use `from __future__ import annotations` and you'll be able to do
```suggestion
    def copy(self) -> SnubaParams:
```

src/sentry/api/endpoints/organization_events_trace.py
```diff
@@ -276,7 +274,9 @@ def load_performance_issues(self, light: bool, snuba_params: ParamsType | None)
                 if self.nodestore_event is not None:
                     occurrence_query = DiscoverQueryBuilder(
                         Dataset.IssuePlatform,
-                        snuba_params,
+                        # Params is ignored if snuba_params is passed
+                        params={},
```
Zylphrex, line 278: I assume this is because `params` is a positional argument in the query builder still? It'd be nice if it was an optional one in the future to avoid this strange signature.

src/sentry/api/endpoints/organization_events_trace.py
```diff
@@ -487,18 +488,19 @@ def count_performance_issues(trace_id: str, params: Mapping[str, str]) -> int:
 @sentry_sdk.tracing.trace
 def create_transaction_params(
     trace_id: str,
-    params: Mapping[str, str],
-) -> Mapping[str, str]:
+    snuba_params: SnubaParams,
+) -> SnubaParams:
     """Can't use the transaction params for errors since traces can be errors only"""
-    transaction_params = params.copy()
+    transaction_params = snuba_params.copy()
```
Zylphrex, line 494: Nit: This copy doesn't need to happen until almost the end of this function as there's several early escapes in this function

src/sentry/api/utils.py
```diff
@@ -486,5 +490,13 @@ def update_snuba_params_with_timestamp(
         example_end = example_timestamp + timedelta(days=time_buffer)
         # If timestamp is being passed it should always overwrite the statsperiod or start & end
         # the client should just not pass a timestamp if we need to overwrite this logic for any reason
-        params["start"] = max(params["start"], example_start)
-        params["end"] = min(params["end"], example_end)
+        if isinstance(params, SnubaParams):
+            # Typing gets mad that start is optional still but has_dates has checked it
+            params.start = cast(datetime.datetime, params.start)
+            params.end = cast(datetime.datetime, params.end)
```
Zylphrex, line 496: Not a huge fan of casting if we avoid it as it makes assumptions in the preceding code. Can this be an assertion instead?

```python
assert params.start is not None
assert params.end is not None
```

Zylphrex (approved):

src/sentry/api/endpoints/organization_events_trace.py
wmak: Changing that breaks a buuuunch of stuff lol, will be doing that in the future but increases the impact of the change significantly

src/sentry/api/endpoints/organization_events_trace.py
wmak: oh good catch

src/sentry/api/utils.py
wmak: For sure, good idea

src/sentry/search/events/types.py
wmak: Ah I keep forgetting about that ty

PR #73852, opened 2024-07-04
ref(discover): Use the same helper for checking dataset

narsaynorath:
There were two helpers for evaluating the split decisions for Discover and Dashboards. The necessary logic for both was very similar so I figured there was an opportunity to turn some of the differences into parameters that we could pass in and have the same logic.

Aside from renaming and moving around code, I updated the `save_dashboard_widget_split_decision` function to have the same logic and telemetry as `save_discover_saved_query_split_decision`. The only difference is that since we haven't decided on a default condition for dashboard widgets, I've left filling out the field as `None`

I'm relying on the previous tests for validity on this refactor. I will add more tests as the functionality changes.

### DIFF
--- a/src/sentry/api/bases/organization_events.py
+++ b/src/sentry/api/bases/organization_events.py
@@ -234,32 +234,31 @@ class OrganizationEventsV2EndpointBase(OrganizationEventsEndpointBase):
 
         return use_on_demand_metrics, on_demand_metric_type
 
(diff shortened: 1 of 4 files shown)


narsaynorath:
Drafting this to fix tests and try an alternate approach. Decomposing the common logic in smaller helpers might be better than trying to fit it to a template, the dashboard function has more logic to handle on demand queries than I realized.

PR #73853, opened 2024-07-04
flag(sampling): Add DS killswitch for ST AM1->AM2

k-fish:
The sampling flags are having inconsistencies on STs, this will let us immediately kill sampling if plan changes are not instant.

⚠️  Would like careful eyes on the logic here since it can effect on the pipeline.

### DIFF
--- a/src/sentry/relay/config/__init__.py
+++ b/src/sentry/relay/config/__init__.py
@@ -351,6 +351,10 @@ def get_project_config(
 
 
 def get_dynamic_sampling_config(timeout: TimeChecker, project: Project) -> Mapping[str, Any] | None:
+    if options.get("dynamic-sampling.config.killswitch"):
+        # This killswitch will cause extra load, and should only be used for AM1->AM2 migration.
+        return None
+
     if features.has("organizations:dynamic-sampling", project.organization):
         return {"version": 2, "rules": generate_rules(project)}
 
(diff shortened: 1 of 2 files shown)


wmak (approved):

src/sentry/relay/config/__init__.py
```diff
@@ -351,6 +351,10 @@ def get_project_config(
 
 
 def get_dynamic_sampling_config(timeout: TimeChecker, project: Project) -> Mapping[str, Any] | None:
+    if options.get("dynamic-sampling.config.killswitch"):
+        # This killswitch will cause extra load, and should only be used for AM1->AM2 migration.
+        return None
```
k-fish, line 356: @wmak you double checks this ships without this tripping, yeah? Just want to make sure

wmak: 🤔 Yeah, I think if you're concerned about this change we could take the extra step to return a manually set config from options. But that might not be worth the extra complexity.

k-fish: Yeah I still don't know that a custom config as a signal will work since we'd have to confirm it doesn't break relay in some way.

wmak: :+1: Makes sense to me

PR #73855, opened 2024-07-04
ref(related_issues): Various improvements

armenzg:
These changes are in preparation for the next pull request.

Changes included:
* Fix typing for events
* Handle titles ending with a colon
* Report Sentry error if determining the metadata fails
  * Fallback to default values
* Add test for the error event type

PR #73860, opened 2024-07-04
feat(related_issues): Support default events

armenzg:
Based on #73855.

Some issues are created from default events rather than error events.

This change makes Trace-related issues use the same title, subtitle and message as issues created from the default events.

This shows a working default event:
image
instead of this:
image

NOTE: We can't use the same function we use for the Issue Details page since the data come from different endpoints and have different types.


static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -60,7 +60,9 @@ export function TraceIssueEvent({event}: TraceIssueEventProps) {
         <TraceIssueDetailsContainer>
           <NoOverflowDiv>
             <TraceIssueEventTitle>{title}</TraceIssueEventTitle>
-            <TraceIssueEventSubtitle>{subtitle}</TraceIssueEventSubtitle>
+            <TraceIssueEventSubtitle data-test-id="subtitle-span">
```
armenzg, line 63: `data-test-id` will be used in the tests.

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -81,14 +83,20 @@ export function TraceIssueEvent({event}: TraceIssueEventProps) {
...
     title = event.title.trimEnd();
     if (event['event.type'] === 'error') {
       if (title[title.length - 1] !== ':') {
         title = event.title.split(':')[0];
       }
+    } else if (event['event.type'] === 'default') {
+      // https://github.com/getsentry/sentry/blob/f08644a004f9980d48f93dec2d7cfc9eeecd9a9e/static/app/utils/events.tsx#L179-L184
+      subtitle = '';
+      // https://github.com/getsentry/sentry/blob/f08644a004f9980d48f93dec2d7cfc9eeecd9a9e/static/app/utils/events.tsx#L59-L60
+      const errorEvent = event as TimelineDiscoverEvent;
```
armenzg, line 98: `culprit` is not shared by both types, thus, I must distinguish them here.

static/app/views/issueDetails/traceTimelineOrRelatedIssue.spec.tsx
```diff
@@ -256,6 +269,33 @@ describe('TraceTimeline & TraceRelated Issue', () => {
...
+      url: `/organizations/${organization.slug}/projects/`,
+      body: [],
+    });
+
+    render(<TraceTimeLineOrRelatedIssue event={event} />, {organization});
+
+    // Check title, subtitle and message of default event
+    expect(await screen.findByText(defaultEvent.title)).toBeInTheDocument();
+    // The subtitle is empty for default events
+    expect(await screen.getByTestId('subtitle-span')).toHaveTextContent('');
```
armenzg, line 294: The subtitle is empty for default events.

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
malwilley: It's better to fix the types in this case so that typescript can discriminate them without you needing to use `as`.

If TimelineDiscoverEvent has `event.type': 'default'` in the type def, typescript should do this automatically.

malwilley (approved):

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
armenzg: I spoke with Malachi and I will be addressing this in a follow-up.

Issue #73717, opened 2024-07-01
Frontend is still using ``culprit`` to determine subtitle of issues


DanielNoord:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

Send an event to Sentry using the Python SDK on version 2.x.
Make sure to be using OpenTelemetry for tracing or just disable all tracing.
Add a transaction name using the functions the SDK exposes.

### Expected Result

The subtitle of the event should not be using the `culprit` key as it is [deprecated](https://github.com/getsentry/sentry/pull/73557).

I tried fixing it in https://github.com/getsentry/sentry/pull/73557, but it seems to be a larger refactor than I can take on right now. Please feel free to add commits to that branch to fix the issue if that is easier.

### Actual Result

The frontend still uses the deprecated key for the subtitle, with no way in the SDK to set it.

This makes triaging issues much harder than it needs to be.

### Product Area

Issues

### Link

/

### DSN

/

### Version

Not self hosted

vartec:
@armenzg is https://github.com/getsentry/sentry/pull/73861 fixing this?

DanielNoord:
@vartec That PR actually makes the situation worse by introducing even more uses of the deprecated key. According to your documentation no code should be using `culprit` anymore as it is deprecated.

vartec:
> @vartec That PR actually makes the situation worse by introducing even more uses of the deprecated key. According to your documentation no code should be using `culprit` anymore as it is deprecated.

@armenzg FYI

armenzg:
I will look into undoing what I added.

armenzg:
#74072 partially fixes this.

armenzg:
Hi @DanielNoord could you please provide STR?

I mentioned this in the PR but I will in here too:
[The [events.tsx](https://github.com/getsentry/sentry/blob/master/static/app/utils/events.tsx) file] is using `culprit` from the `Group` model:
https://github.com/getsentry/sentry/blob/ac83b15608a785ac60ae46231da83c8dee5bfe00/src/sentry/models/group.py#L532-L534

In my PR I am using `culprit` from an event which is a [generated value](https://github.com/getsentry/sentry/blob/23f2b42028b197415b1fc55d1f233c1e5414c89f/src/sentry/culprit.py#L15-L43).

Ideally, we would not use the same field name to prevent confusion.

DanielNoord:
[...]

PR #73861, opened 2024-07-04
fix(related_issues): Use culprit and last error.value

armenzg:
Up until now, we were using the `event.transaction` to generate the subtitle, however, `getTitle()` from utils/events.tsx uses the `culprit`.

We also use the last `error.value` since it's more accurate to what the group title would be.

image


static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -82,19 +82,21 @@ export function TraceIssueEvent({event}: TraceIssueEventProps) {
...
   let message = event.message;
   try {
     if (event['event.type'] === 'error') {
       if (title[title.length - 1] !== ':') {
         title = event.title.split(':')[0];
       }
+      // It uses metadata.value which could differ depending on what error.value is used in the event manager
+      // TODO: Add support for chained exceptions since we grab the value from the first stack trace
+      // https://github.com/getsentry/sentry/blob/a221f399d2b4190f2631fcca311bdb5b3748838b/src/sentry/eventtypes/error.py#L115-L134
+      message = event['error.value'][event['error.value'].length - 1];
```
malwilley, line 95: nit: can use `.at(-1)` to access the last index

malwilley (approved):

DanielNoord:
@armenzg did you see https://github.com/getsentry/sentry/issues/73717?

armenzg:
#74072 partially reverts some of what I did in this PR.

PR #73862, opened 2024-07-04
ref(snuba): Refactor utils/snuba to allow for multiple referrers

Zylphrex:
Snuba queries only allow for a single referrer for a batch of parallel queries. But sometimes, these queries have nothing to do with each other and using the same referrer is not ideal. This refactor prepares the utilities to allow callers to pass a referrer per query.

### DIFF
--- a/src/sentry/utils/snuba.py
+++ b/src/sentry/utils/snuba.py
@@ -849,13 +849,12 @@ def raw_query(
 
 
 Translator = Callable[[Any], Any]
-RequestQueryBody = tuple[Request, Translator, Translator]
 
 
 @dataclasses.dataclass(frozen=True)
 class SnubaRequest:
-    referrer: str | None  # this should use the referrer class
     request: Request
+    referrer: str | None  # this should use the referrer class
     forward: Translator
     reverse: Translator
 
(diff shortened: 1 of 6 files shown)


mjq (approved):
Nice!

src/sentry/utils/snuba.py
```diff
@@ -847,9 +848,31 @@ def raw_query(
...
 
-SnubaQuery = Union[Request, MutableMapping[str, Any]]
 Translator = Callable[[Any], Any]
-RequestQueryBody = tuple[Request, Translator, Translator]
+
+
+@dataclasses.dataclass(frozen=True)
+class SnubaRequest:
+    request: Request
+    referrer: str | None  # this should use the referrer class
```
wmak, line 857: Is this a TODO?

src/sentry/utils/snuba.py
```diff
@@ -847,9 +848,31 @@ def raw_query(
...
+    forward: Translator
+    reverse: Translator
+
+    def __post_init__(self) -> None:
+        self.validate()
+
+    def validate(self):
+        if self.referrer:
```
wmak, line 865: Huh, this matches previous behaviour but it feels so weird that referrer isn't required

tests/sentry/issues/test_attributes.py
```diff
@@ -237,14 +235,7 @@ def run_attr_test(
...
             request.validate()
-            identity = lambda x: x
-            isolation_scope = sentry_sdk.Scope.get_isolation_scope().fork()
-            current_scope = sentry_sdk.Scope.get_current_scope().fork()
-            resp = _snuba_query(
-                ((request, identity, identity), isolation_scope, current_scope, {}, "test_api")
-            )
-            assert resp[0].status == 200
-            stuff = json.loads(resp[0].data)
+            stuff = raw_snql_query(request)
```
wmak, line 238: ... `stuff`, I know you're reusing the existing variable names but what the heck

wmak (approved):
Small comments, lgtm

src/sentry/utils/snuba.py
Zylphrex: Yes, I'll prefix the comment with a TODO once I figure out if we can change this in all the call sites.

src/sentry/utils/snuba.py
Zylphrex: Yeah, I thought this was weird too but let's leave the behaviour unchanged in this PR

tests/sentry/issues/test_attributes.py
Zylphrex: Good point, let me rename all these to something more meaningful since I'm here anyways

PR #73881, opened 2024-07-05
ref(related_issues): Do not fetch message for error events

armenzg:
In PR #73861 we dropped the usage of the `message` field for error events, however, I separated the change here to reduce the size of that PR.


static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -98,9 +98,13 @@ export function getTitleSubtitleMessage(event: TimelineEvent) {
       subtitle = '';
       message = event.culprit;
     } else {
+      const issuePlatformEvent = event as TimelineIssuePlatformEvent;
```
malwilley, line 101: As mentioned in another PR review, if you change the types to include literals for `event.type` typescript will discriminate automatically here

malwilley (approved):

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -83,7 +83,7 @@ export function TraceIssueEvent({event}: TraceIssueEventProps) {
 export function getTitleSubtitleMessage(event: TimelineEvent) {
   let title = event.title.trimEnd();
   let subtitle = event.culprit;
-  let message = event.message;
+  let message = '';
```
armenzg, line 86: A default value rather than null in case the try/catch block fails.

static/app/views/issueDetails/traceTimeline/useTraceTimelineEvents.tsx
```diff
@@ -10,15 +10,16 @@ interface BaseEvent {
   culprit: string; // Used for default events & subtitles
   id: string;
   'issue.id': number;
-  message: string;
```
armenzg, line 13: It is now not a shared property between the two types.

static/app/views/issueDetails/traceTimelineOrRelatedIssue.spec.tsx
```diff
@@ -49,7 +49,7 @@ describe('TraceTimeline & TraceRelated Issue', () => {
         'project.name': project.name,
         title: 'Slow DB Query',
         id: 'abc',
-        transaction: '/api/slow/',
+        transaction: 'n/a',
```
armenzg, line 52: `transaction` is not used anymore when creating the message for issue platform events.

PR #73886, opened 2024-07-05
feat(related_issues): Support issues with no stats

armenzg:
The upcoming Uptime issues do not have stats, thus, they do not render the `GroupEventHeader`.

This change moves the rendering of the `TraceTimeLineOrRelatedIssues` higher up in the stack.

Screenshots of various issues (narrow window):
* With timeline:
image

* With a trace-related issue and suspect commit:
image


static/app/views/issueDetails/groupEventDetails/groupEventDetails.tsx
```diff
@@ -192,6 +193,11 @@ function GroupEventDetails(props: GroupEventDetailsProps) {
                     project={project}
                   />
                 )}
+                {eventWithMeta && (
```
armenzg, line 196: See the lines above and you can notice that `issueTypeConfig.stats.enabled` would prevent issues without stats from rendering.
image

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -89,7 +89,7 @@ const TraceIssueLinkContainer = styled(Link)`
   gap: ${space(2)};
   color: ${p => p.theme.textColor};
   padding: ${space(2)} ${space(2)} ${space(2)} ${space(2)};
-  margin: ${space(1)} 0 ${space(1)} 0;
```
armenzg, line 92: It looks a tiny bit better.

static/app/views/issueDetails/groupEventDetails/groupEventDetails.tsx
```diff
@@ -245,4 +251,11 @@ const StyledLayoutSide = styled(Layout.Side)`
   }
 `;
 
+const StyledDataSection = styled('div')`
```
armenzg, line 254: This aims to replicate the wrapper we had in `groupEventHeader`:
image

davidenwang (approved):

malwilley:
@armenzg do you think that this could be placed in `groupEventDetailsContent.tsx` instead? I think this would make more sense there

armenzg:
@malwilley I tested a few different cases and everything looks as expected
image

<hr />
image

<hr />
image

PR #73887, opened 2024-07-05
feat(snuba): New api to query snuba with multiple referrers

Zylphrex:
Adds a new api to that allows making parallel snuba queries but with different referrers per query.

### DIFF
--- a/src/sentry/utils/snuba.py
+++ b/src/sentry/utils/snuba.py
@@ -893,7 +893,7 @@ def raw_snql_query(
     # other functions do here. It does not add any automatic conditions, format
     # results, nothing. Use at your own risk.
     return bulk_snuba_queries(
-        requests=[request], referrer=referrer, query_source=query_source, use_cache=use_cache
+        requests=[request], referrer=referrer, use_cache=use_cache, query_source=query_source
     )[0]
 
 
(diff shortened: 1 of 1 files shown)


src/sentry/utils/snuba.py
```diff
@@ -904,18 +904,39 @@ def bulk_snuba_queries(
...
+    metrics.incr("snql.sdk.api", tags={"referrer": referrer or "unknown"})
+
+    return concurrent_snuba_queries(
+        [(request, referrer) for request in requests],
+        use_cache=use_cache,
+        query_source=query_source,
+    )
+
+
+def concurrent_snuba_queries(
```
wmak, line 921: 🤔 should we just have this fully replace `bulk_snuba_queries`? call the old one _deprecated or something

Zylphrex: It might be nice to keep `bulk_snuba_queries` around for compatibility and convenience unless you'd like to pass a referrer for each query you pass. I'm okay with getting rid of it too but it feels very verbose in most use cases.

wmak: 🤔 I think we could replace the old function with like (needs a better name lmao) `bulk_snuba_queries_with_same_referrer` that just calls this current function with
```python
concurrent_snuba_queries([(request, referrer) for request in requests], use_cache, query_source)
```

Zylphrex: That's what the old function is doing now. So you're proposing we rename both functions to make it clear which one uses 1 referrer and which one uses multiple referrers?

wmak: d'oh missed that, actually the more i think about it changing the name of the existing one will suck cause you have to change all the references

so maybe just rename the current function to `bulk_snuba_queries_with_referrers`? 🤔
I'm just worried with `concurrent_snuba_queries` and `bulk_snuba_queries` at a glance its not obvious which one to use

Zylphrex: Yep, that's precisely my concern. I can definitely rename the second one to `bulk_snuba_queries_with_referrers` if that makes the intent clear that it is to be used when you want to make parallel queries with different referrers.

wmak: yeah i feel like longterm we need to drop the other one, its easy enough to write the `with_same_referrer` version yourself.

Zylphrex: Sounds good, gonna rename it to `bulk_snuba_queries_with_referrers` to start and we can think about migrating to it everywhere in the future.

wmak: ~do it in this PR!~
:+1: sounds good to me

wmak (approved):
Looks good to me

PR #73888, opened 2024-07-05
ref(typing): Type the events-trace endpoint

wmak:
- This updates the trace endpoint so that its passing typing again

### DIFF
--- a/src/sentry/api/endpoints/organization_events_trace.py
+++ b/src/sentry/api/endpoints/organization_events_trace.py
@@ -1,9 +1,11 @@
+from __future__ import annotations
+
 import logging
 from collections import defaultdict, deque
 from collections.abc import Callable, Iterable, Mapping, Sequence
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from datetime import datetime, timedelta
-from typing import Any, Deque, Optional, TypedDict, TypeVar
+from typing import Any, Deque, Optional, TypedDict, TypeVar, cast
 
 import sentry_sdk
 from django.http import Http404, HttpRequest, HttpResponse
@@ -23,6 +25,8 @@ from sentry.eventstore.models import Event
 from sentry.issues.issue_occurrence import IssueOccurrence
 from sentry.models.group import Group
 from sentry.models.organization import Organization
+from sentry.models.project import Project
+from sentry.organizations.services.organization import RpcOrganization
 from sentry.search.events.builder.discover import DiscoverQueryBuilder
 from sentry.search.events.builder.spans_indexed import SpansIndexedQueryBuilder
 from sentry.search.events.types import QueryBuilderConfig, SnubaParams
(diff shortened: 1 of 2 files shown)


src/sentry/api/endpoints/organization_events_trace.py
```diff
@@ -371,10 +401,10 @@ def full_dict(
         event_id = self.event["id"]
         # We're in a loop!
```
k-fish, line 402: 🤔  Why is this comment here?

wmak: This refers to a loop in the trace structure lol

k-fish: Do we really need this here?

wmak: Hmm probably could reword it to be better

src/sentry/api/endpoints/organization_events_trace.py
```diff
@@ -41,124 +45,144 @@
...
-        "precise.start_ts": int,
+        "measurements": dict[str, int],
+        "occurrence_id": list[str],
+        "occurrence_spans": list[SnubaSpan],
+        "occurrence_to_issue_id": dict[str, list[int]],
         "precise.finish_ts": int,
-        "trace.span": str,
+        "precise.start_ts": int,
+        "profile.id": str,
+        "profile.profiler_id": str,
```
Zylphrex, line 71: The type here changed from before from

```python
"profile.profile_id" -> "profile.profiler_id"
"profiler.id" -> "profile.id"
```

Can you double check to make sure which one is correct?

src/sentry/api/endpoints/organization_events_trace.py
```diff
@@ -371,10 +399,10 @@ def full_dict(
         event_id = self.event["id"]
         # We're in a loop!
         if event_id in visited:
-            return
+            return None
         else:
             visited.add(self.event["id"])
-        result = self.to_dict()
+        result = cast(FullResponse, self.to_dict())
```
Zylphrex, line 405: Is this casting a `LightResponse` to a `FullResponse`? Is this right?

src/sentry/api/endpoints/organization_events_trace.py
```diff
@@ -1013,11 +1053,6 @@ def get(self, request: HttpRequest, organization: Organization, trace_id: str) -
         if event_id and not is_event_id(event_id):
             return Response({"detail": INVALID_ID_DETAILS.format("Event ID")}, status=400)
 
-        tracing_without_performance_enabled = features.has(
-            "organizations:performance-tracing-without-performance",
-            organization,
-            actor=request.user,
-        )
```
Zylphrex, line 1020: Is this flag fully GA'ed now? Are there environments where this can't be turned on like single tenant or self hosted?

src/sentry/api/endpoints/organization_events_trace.py
```diff
@@ -1063,11 +1096,23 @@ def get(self, request: HttpRequest, organization: Organization, trace_id: str) -
...
+        limit: int,
+        transactions: Sequence[SnubaTransaction],
+        errors: Sequence[SnubaError],
+        roots: Sequence[SnubaTransaction],
+        warning_extra: dict[str, str],
+        event_id: str | None,
+        detailed: bool = False,
+        use_spans: bool = False,
+    ) -> Any:
+        pass
```
Zylphrex, line 1114: Is this an abstract method that is overridden in the child classes? If so, can it

```python
raise NotImplementedError
```

Bonus points if you turn this into an abstract class and decorate this with `@abc.abstractmethod`

src/sentry/api/endpoints/organization_events_trace.py
```diff
@@ -1158,7 +1195,7 @@ def serialize(
 
         with sentry_sdk.start_span(op="building.trace", description="light trace"):
             # Check if the event is an orphan_error
-            if not snuba_event and not nodestore_event and allow_orphan_errors:
+            if not snuba_event or not nodestore_event:
```
Zylphrex, line 1198: Why did this `and` change into an `or`?

src/sentry/api/endpoints/organization_events_trace.py
```diff
@@ -1325,9 +1358,11 @@ def serialize(
                 warning_extra,
                 event_id,
                 detailed,
-                allow_orphan_errors,
             )
             return results
+
+        # Code past here is deprecated, but must continue to exist until sentry installs in every possible environment
+        # are able to support storing span data
```
Zylphrex, line 1365: We need to be more strict here. As long as any environment has the option to turn off spans, we have to keep this code around.

src/sentry/api/endpoints/organization_events_trace.py
wmak: New one is correct, these types weren't being used before so they were just wrong without being noticed

src/sentry/api/endpoints/organization_events_trace.py
wmak: Yeah, I a FullResponse shares a lot with a LightResponse so this is just extending it. Probably could rewrite this to be a bit cleaner, but we aren't using light responses anymore in environments where spans are enabled

src/sentry/api/endpoints/organization_events_trace.py
wmak: Fully GA'ed, the frontend already assumes this flag is on since it changes the response 😬

src/sentry/api/endpoints/organization_events_trace.py
wmak: fiiiine

src/sentry/api/endpoints/organization_events_trace.py
wmak: 👀 uh. no idea good catch

src/sentry/api/endpoints/organization_events_trace.py
wmak: Hmm what do you mean by being more strict?

Zylphrex: I guess what I'm trying to say is that it is not enough that every environment is capable of storing span data. Every environment must only use span data before this old code can be removed.

src/sentry/api/endpoints/organization_events_trace.py
wmak: Oh nevermind, this is a typing catch, the rest of this function needs both snuba_event and nodestore_event to be defined, so if either isn't defined we want to do the following block

src/sentry/api/endpoints/organization_events_trace.py
wmak: Reworded this slightly

k-fish (approved):
thanks for the cleanup, nice types Will 👏

PR #73891, opened 2024-07-05
feat(insights): module empty states

DominikB2014:
Everything here is internal only for now, this PR updates the module onboarding, and wraps the other modules with the onboarding component. Any old onboarding content will be removed in the release PR.

Text that should be used is found here https://www.notion.so/sentry/Insights-Pending-for-M-Launch-4c9162936fbc4a6897085851c503c0f1

Also note: there is duplicate images between this PR and the module upsells (https://github.com/getsentry/getsentry/pull/14519/files), I didn't realize they would share the same images. 
Once this is merged, i will delete the `getsentry` duplicates and use the ones from `sentry`

image


static/app/views/insights/cache/settings.ts
```diff
@@ -3,6 +3,8 @@ import type {SpanMetricsQueryFilters} from 'sentry/views/insights/types';
 
 export const MODULE_TITLE = t('Caches');
 export const BASE_URL = 'caches';
+export const DATA_TYPE = t('Cache');
+export const DATA_TYPE_PLURAL = MODULE_TITLE;
```
gggritso, line 7:
```suggestion
export const DATA_TYPE_PLURAL = t('Caches');
```

Safer, so they can be changed independently. If you change the `MODULE_TITLE` to `t('Cache Insights')` or something, it'd look weird in other places

static/app/views/insights/queues/settings.ts
```diff
@@ -2,6 +2,8 @@ import {t} from 'sentry/locale';
 
 export const MODULE_TITLE = t('Queues');
 export const BASE_URL = 'queues';
+export const DATA_TYPE = t('Queue');
```
gggritso, line 5: 🤔 not `t('Task')`?

gggritso (commented):
Looks good! Watch out for the plurals though, they shouldn't reference the module title!

static/app/views/insights/cache/settings.ts
DominikB2014: Yeah that's a good point, I'll change it! 

I was trying to decide between reusing module title or not, but your of point of it being safer to not reuse it makes sense

static/app/views/insights/queues/settings.ts
DominikB2014: Spoke with Ed, We're going to keep it as Queue

gggritso (approved):
👍🏻

PR #73893, opened 2024-07-05
feat(mep): Add fallback to transactions

narsaynorath:
To split the discover dataset in dashboards, we need to be able to run a query against transactions as the fallback for metrics enhanced queries instead of the discover dataset.

This adds a branch of code that will allow us to handle transaction-like requests when we cannot fulfill the query with metrics. Currently it isn't being used, but will be used when we need to remove our dependency on `discover`.

I added a test to check that the split occurs as expected. The class doesn't do much logic other than falling back and setting `meta` values, so I didn't bring over any tests for those. They're tested by other files specifically.

### DIFF
--- a/src/sentry/api/endpoints/organization_events.py
+++ b/src/sentry/api/endpoints/organization_events.py
@@ -23,7 +23,7 @@ from sentry.discover.models import DiscoverSavedQuery, DiscoverSavedQueryTypes
 from sentry.exceptions import InvalidParams
 from sentry.models.dashboard_widget import DashboardWidget, DashboardWidgetTypes
 from sentry.models.organization import Organization
-from sentry.snuba import discover, metrics_enhanced_performance, metrics_performance, transactions
+from sentry.snuba import discover, metrics_enhanced_performance, metrics_performance
 from sentry.snuba.metrics.extraction import MetricSpecType
 from sentry.snuba.referrer import Referrer
 from sentry.snuba.utils import dataset_split_decision_inferred_from_query, get_dataset
@@ -384,6 +384,8 @@ class OrganizationEventsEndpoint(OrganizationEventsV2EndpointBase):
                     organization,
                     actor=request.user,
                 )
+                has_errors = False
+                has_transactions = False
 
                 if does_widget_have_split and not has_override_feature:
                     # This is essentially cached behaviour and we skip the check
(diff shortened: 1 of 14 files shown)


src/sentry/api/endpoints/organization_events.py
```diff
@@ -432,12 +441,7 @@ def _dashboards_data_fn(scopedDataset, offset, limit, scoped_query, dashboard_wi
                 if has_errors and has_other_data and not using_metrics:
                     # In the case that the original request was not using the metrics dataset, we cannot be certain that other data is solely transactions.
                     sentry_sdk.set_tag("third_split_query", True)
-                    transactions_only_query = (
-                        f"({scoped_query}) AND event.type:transaction"
-                        if scoped_query
-                        else "event.type:transaction"
-                    )
```
Zylphrex, line 439: why was this removed?

src/sentry/api/endpoints/organization_events.py
```diff
@@ -369,6 +371,7 @@ def _data_fn(scopedDataset, offset, limit, query) -> dict[str, Any]:
                 use_metrics_layer=batch_features.get("organizations:use-metrics-layer", False),
                 on_demand_metrics_enabled=on_demand_metrics_enabled,
                 on_demand_metrics_type=on_demand_metrics_type,
+                fallback_to_transactions=fallback_to_transactions,
```
Zylphrex, line 374: Could this be a feature flag check within the `metrics_enhanced_performance`'s `.query` function? This parameter does not apply to any other dataset but is being added to every one.

src/sentry/api/endpoints/organization_events.py
narsaynorath: For this block of code we're evaluating the `transactions` only path without metrics, so I removed the call to the discover dataset and instead queried transactions directly. Is there something I'm missing the requires us to hit discover still in this case?

src/sentry/api/endpoints/organization_events.py
narsaynorath: I spoke to Shruthi about this and we wanted to control the rollout strictly so we can handle dashboards without affecting the performance overview page. I agree that adding the parameter to other datasets is pretty ugly but we do the same with other params currently and this one is intended to be temporary

tests/snuba/api/endpoints/test_organization_events_mep.py
```diff
@@ -3458,6 +3458,64 @@ def test_filtering_by_org_id_is_not_compatible(self):
         )
         assert response.status_code == 400, response.content
 
+    @mock.patch("sentry.snuba.transactions.query")
+    def test_mep_uses_transactions_dataset_when_fallback_is_specified(
```
shruthilayaj, line 3462: I propose we add extensive test cases now - ie, cover all the test cases that the `discover` arm catches. because once that's removed we might have holes in the tests.

src/sentry/api/endpoints/organization_events.py
shruthilayaj: Can we gate the using the transaction fallback behind the `split-dataset` flag?

tests/snuba/api/endpoints/test_organization_events_mep.py
shruthilayaj: We should also include tests for `timeseries_query`

shruthilayaj (commented):
I think there docstrings for the `query` and `timeseries_query` api, can we add some docstrings for the new param too?

src/sentry/api/endpoints/organization_events.py
shruthilayaj: Oh actually, this should be okay since we're explicitly trying to query transactions here. Can you send [this](https://github.com/getsentry/sentry/pull/73893/files#diff-084d1fb1e68307632ee00ae568af3c10195b19880bc8b28c49585c1e1f4d5beeR407-R416) query to the errors dataset too?

src/sentry/api/endpoints/organization_events.py
```diff
@@ -417,7 +420,13 @@ def _dashboards_data_fn(scopedDataset, offset, limit, scoped_query, dashboard_wi
                     has_errors = False
                     error_results = None
 
-                original_results = _data_fn(scopedDataset, offset, limit, scoped_query)
+                original_results = _data_fn(
+                    scopedDataset,
+                    offset,
+                    limit,
+                    scoped_query,
+                    fallback_to_transactions=save_discover_dataset_decision,
```
shruthilayaj, line 428: This is the original results we return if we're unable to make a decision. For now (since we haven't decided on a default, have we?) this should be `fallback_to_transaction=False`. The transactions results below should query metricsEnhanced with `fallback_to_transaction=True`.

So now you have 3 queries:
1. Errors
2. MEP with transactions
3. MEP with discover

If both 2 and 3 return results or neither return results (ie if it's ambiguous, use the results from 3).

shruthilayaj (approved):

PR #73897, opened 2024-07-05
chore(deletions) Fix project deletions

markstory:
Fix stuck deletions caused by foreign key constraints. The QuerySubscription -> Incident relation is blocking deletions. Setting Incident.subscription_id = null will unblock deletions and lets the incident outlive the project.

Fixes SENTRY-3AVB


nhsiehgit (approved):

nhsiehgit:
Oof thanks for this catch! 🙏

PR #73926, opened 2024-07-08
feat(uptime): Log urls that would be monitored

davidenwang:
If the urls make it this far we should log a message


evanpurkhiser (approved):
Will be useful to do some spot checks

davidenwang:
> Will be useful to do some spot checks

tru tru lemme instrument some more

PR #73931, opened 2024-07-08
ref(onboarding): Consolidate getting started docs hooks

priscilawebdev:
### Goal

This PR aims to consolidate the hooks `useLoadFeedbackOnboardingDoc`, `useLoadGettingStarted`, and `useLoadReplayOnboardingDoc` due to their similarities. By unifying these hooks, we can streamline the codebase and improve maintainability. 

Part of https://github.com/getsentry/sentry/pull/73496#pullrequestreview-2147510629


ArthurKnaus (approved):
Great work! 🚀

static/app/components/replaysOnboarding/sidebar.tsx
```diff
@@ -319,12 +321,10 @@ function OnboardingContent({currentProject}: {currentProject: Project}) {
...
-    return (
-      <Fragment>
-        {radioButtons}
-        <LoadingIndicator />
-      </Fragment>
-    );
+    <Fragment>
+      {radioButtons}
+      <LoadingIndicator />
+    </Fragment>;
```
ryan953, line 327: missing `return`

static/app/utils/useProjectKeys.tsx
```diff
@@ -1,49 +1,28 @@
...
+} from 'sentry/utils/queryClient';
 
-import type {Organization, Project, ProjectKey, RequestState} from 'sentry/types';
+interface ProjectKeysParameters {
+  orgSlug: string;
+  projSlug?: string;
+}
 
-import useApi from './useApi';
+export const makeProjectKeysQueryKey = ({
```
ryan953, line 13: exported but not used

static/app/components/onboarding/gettingStartedDoc/sdkDocumentation.tsx
```diff
@@ -25,40 +24,56 @@ export function SdkDocumentation({
...
-    isLoading: projectKeysIsLoading,
-    refetch: refetchProjectKeys,
-  } = useApiQuery<ProjectKey[]>([`/projects/${organization.slug}/${projectSlug}/keys/`], {
-    staleTime: Infinity,
-  });
-
-  const module = useLoadGettingStarted({
-    platformId: platform.id,
-    platformPath,
+  const loadGettingStarted = useLoadGettingStarted({
```
michellewzhang, line 31:
```suggestion
  const {isLoading, isError, dsn, docs} = useLoadGettingStarted({
```
doing some destructuring might make this file a tiny bit easier to read

static/app/components/onboarding/gettingStartedDoc/utils/useLoadGettingStarted.tsx
```diff
@@ -6,49 +6,73 @@ import {
...
     };
-  }, [platformPath, platformId, productType]);
+  }, [platformPath, platform.id, productType]);
 
-  return module;
+  return {
+    refetch: projectKeys.refetch,
+    isLoading: projectKeys.isLoading || module === undefined,
+    isError: projectKeys.isError,
+    docs: module === 'none' ? null : module?.default ?? null,
```
michellewzhang, line 74:
```suggestion
    docs: module?.default ?? null,
```
i think this should simplify the logic here

static/app/components/feedback/feedbackOnboarding/sidebar.tsx
```diff
@@ -192,32 +192,30 @@ function OnboardingContent({currentProject}: {currentProject: Project}) {
...
-      replayJsFrameworkOptions.find(p => p.id === jsFramework.value) ??
-      replayJsFrameworkOptions[0]
-    );
-  }
+  const jsFrameworkPlatform =
+    replayJsFrameworkOptions.find(p => p.id === jsFramework.value) ??
+    replayJsFrameworkOptions[0];
 
   const {
+    isLoading: isProjKeysLoading,
```
michellewzhang, line 200:
```suggestion
    isLoading,
```

can probably just change the use of `isProjKeysLoading` down below to avoid the alias here - it's only referenced once :)

static/app/components/replaysOnboarding/sidebar.tsx
```diff
@@ -354,7 +354,7 @@ function OnboardingContent({currentProject}: {currentProject: Project}) {
   }
 
   // No platform or no docs (new nor old)
-  if (!currentPlatform || (!newDocs && !hasOnboardingContents)) {
+  if (!currentPlatform || ((!newDocs || !dsn) && !hasOnboardingContents)) {
```
michellewzhang, line 357: i don't think we should add the `!dsn` check here because we still want to fallback to the old onboarding contents even if there isn't a `dsn`; however this logic means the entire block won't render if the `dsn` isn't available
```suggestion
  if (!currentPlatform || (!newDocs && !hasOnboardingContents)) {
```

static/app/components/updatedEmptyState.tsx
```diff
@@ -43,32 +28,26 @@ export default function UpdatedEmptyState({project}: {project?: Project}) {
     p => p.id === currentPlatformKey
   ) as PlatformIntegration;
 
-  const platformPath = getPlatformPath(currentPlatform);
-
-  const module = useLoadGettingStarted({
-    platformId: currentPlatform.id,
-    platformPath,
+  const loadGettingStarted = useLoadGettingStarted({
```
michellewzhang, line 31: same destructuring idea could be done here too!

static/app/components/updatedEmptyState.tsx
```diff
@@ -43,32 +28,26 @@ export default function UpdatedEmptyState({project}: {project?: Project}) {
...
-    !projectKeys ||
-    projectKeys.length === 0 ||
-    !module ||
-    !currentPlatform ||
-    module === 'none'
+    loadGettingStarted.isError ||
+    loadGettingStarted.isLoading ||
+    !loadGettingStarted.docs ||
+    !loadGettingStarted.cdn ||
+    !loadGettingStarted.dsn
```
michellewzhang, line 43: do we still need to check for `!currentPlatform` here? it was checked in the old code

static/app/components/updatedEmptyState.tsx
```diff
@@ -86,9 +65,9 @@ export default function UpdatedEmptyState({project}: {project?: Project}) {
     replayOptions: {block: true, mask: true},
   };
 
-  const install = docs.onboarding.install(docParams)[0];
-  const configure = docs.onboarding.configure(docParams);
-  const verify = docs.onboarding.verify(docParams);
+  const install = loadGettingStarted.docs.onboarding.install(docParams)[0];
+  const configure = loadGettingStarted.docs.onboarding.configure(docParams);
+  const verify = loadGettingStarted.docs.onboarding.verify(docParams);
```
michellewzhang, line 70: doing the destructuring means we can clean this up to be something like this:

```suggestion
  const {install, configure, verify} = docs.onboarding;
  const install = install(docParams)[0];
  const configure = configure(docParams);
  const verify = verify(docParams);
```

static/app/utils/useProjectKeys.tsx
```diff
@@ -1,49 +1,28 @@
...
+import type {ProjectKey} from 'sentry/types/project';
+import {
+  type ApiQueryKey,
+  useApiQuery,
+  type UseApiQueryOptions,
+} from 'sentry/utils/queryClient';
 
-import type {Organization, Project, ProjectKey, RequestState} from 'sentry/types';
+interface ProjectKeysParameters {
+  orgSlug: string;
```
michellewzhang, line 9: any reason why these are `string`s and the ones in other files are `Organization['slug']` and `Project['slug']`?

static/app/components/onboarding/gettingStartedDoc/sdkDocumentation.tsx
priscilawebdev: In this case, I do agree with you. However, when there are multiple `isLoading`, `isError`, etc  variables in the same file, I prefer this method as it makes it easier to see what each variable belongs to.

static/app/components/updatedEmptyState.tsx
priscilawebdev: In this case, I prefer to leave it as it is because we use multiple hooks with `isLoading`, etc in this file. I believe it's better to maintain the current approach rather than using aliases

static/app/utils/useProjectKeys.tsx
priscilawebdev: No, both are strings. However, I prefer using `Organization['slug']` over `string`. This way, if the type changes, it will be reflected here. I missed this one earlier.

static/app/components/updatedEmptyState.tsx
priscilawebdev: I have added it back, but if currentPlatform is not available, then there won't be any docs, DSN, etc., available either. So, it should work just fine

static/app/components/onboarding/gettingStartedDoc/utils/useLoadGettingStarted.tsx
priscilawebdev: That would be better, but we consider the module to be 'loading' when it is `undefined` and set it to `none` when certain conditions are met.

static/app/components/replaysOnboarding/sidebar.tsx
priscilawebdev: I updated the code because I was running into a strange state:

image

However, I just realized this issue arises because the old docs don't work for the de region.

image

should we look into this?

static/app/components/onboarding/gettingStartedDoc/sdkDocumentation.tsx
michellewzhang: yeah makes sense -- i didn't leave the same comment on the replay file for that exact reason!

static/app/components/onboarding/gettingStartedDoc/utils/useLoadGettingStarted.tsx
michellewzhang: if the module is set to `none`, then `module.default` doesn't exist right?

static/app/utils/useProjectKeys.tsx
michellewzhang: yeah sounds good! just noticed the inconsistency. thanks for changing

static/app/components/replaysOnboarding/sidebar.tsx
michellewzhang: hmm that's odd... maybe let's leave as-is for now then 🤔

cc @ArthurKnaus on this issue, maybe he knows more about it?

michellewzhang: looks like some old docs _do_ work, like react & svelte for example, but others don't

ArthurKnaus: I am not aware of any issue with the de region. Are we confident that it is de only? Or were the docs deleted / moved in the sentry-docs repo?

michellewzhang: ohh i think the old docs may have been deleted. old docs were the `.md` files we were using previously. honestly these docs are super old and we've migrated all of replay to the new docs structure so i'd be okay with cleaning up all references to the old docs! happy to make a PR to clean up the replays onboarding

michellewzhang: ah actually -- i think what @priscilawebdev is running into may be the fact that for some platforms, we didn't have old docs written for replay, so they didn't exist in the first place

michellewzhang: https://github.com/getsentry/sentry/pull/74025 here's the PR for cleaning up the references to the old onboarding docs. can wait to merge after this one gets merged in, because there'll be some merge conflicts

michellewzhang (approved):

PR #73936, opened 2024-07-08
feat(minimetrics): emit metric values as span attributes

### LINKED ISSUE #73930: Enrich sentry-internal metrics abstraction with span attributes

obostjancic:
- closes https://github.com/getsentry/sentry/issues/73930


vgrozdanic (approved):

tests/sentry/metrics/test_minimetrics.py
```diff
@@ -254,6 +267,7 @@ def __getattr__(self, name):
 @override_options(
     {
         "delightful_metrics.minimetrics_sample_rate": 1.0,
+        "delightful_metrics.enable_span_attributes": False,
```
vgrozdanic, line 270: Just a question, do we have to define this with `False` value if also default value of the flag is `False`?

obostjancic: yeah, tests start failing if it is not there :D

iambriccardo (approved):
LGTM, let's first test it in S4S for any problems.

PR #73939, opened 2024-07-08
Remove ProcessingIssue models, step 1

Swatinem:
This removes the last use of `RawEvent`, `ProcessingIssue`, `EventProcessingIssue` and `ReprocessingReport`. It also removes any DB foreign key constraints, which is the first step of removing those tables for good.


loewenheim (approved):
LGTM but please don't take that as gospel, I'm not super confident when it comes to migrations.

markstory (approved):
Looks good to me, I didn't find any other constraints that would need to be removed.

PR #73943, opened 2024-07-08
feat(tags): Make dataset a parameter in organization_tags endpoint 

### LINKED ISSUE #73699: Modify organization tags endpoint to accept a dataset parameter
The `OrganizationTagsEndpoint` will query the events dataset with `include_transactions=0` and the Discover dataset when `include_transactions` is enabled. This works for discover and performance issues, because both look at transactions. However, for issues, we need to be searching the errors and issue platform datasets. The way the endpoint works now, there is no way to search the issue platform dataset.

We should add an additional `dataset` param which will override the `dataset` used in the tags request. This param should eventually replace the `include_transactions` parameter. Callers should then be able to send `dataset=errors` or `dataset=issuePlatform` to query the required dataset.

MichaelSun48:
Fixes #73699

This PR replaces the `include_transaction` parameter with a `dataset` parameter in the `organization_tags` endpoint. Previously, the `include_transactions` was used to control whether to pull tags from snuba's `Discover` dataset (if `include_transactions=1`, or `Events` dataset (if `include_transactions=0`). Replacing this "toggle" with an explicit Dataset parameter will enable full control over which dataset is being queried, which is useful for Issues search since it needs Events and IssuePlatform tags, but not Performance tags (included in Discover dataset).

### DIFF
--- a/src/sentry/api/endpoints/organization_tags.py
+++ b/src/sentry/api/endpoints/organization_tags.py
@@ -10,8 +10,13 @@ from sentry.api.bases import NoProjects
 from sentry.api.bases.organization import OrganizationEndpoint
 from sentry.api.serializers import serialize
 from sentry.api.utils import handle_query_errors
+from sentry.snuba.dataset import Dataset
 from sentry.utils.numbers import format_grouped_length
 from sentry.utils.sdk import set_measurement
+
+
+class InvalidDatasetParam(Exception):
+    pass
 
 
 @region_silo_endpoint
@@ -27,6 +32,9 @@ class OrganizationTagsEndpoint(OrganizationEndpoint):
         except NoProjects:
             return Response([])
 
+        if request.GET.get("dataset") and request.GET.get("dataset") not in Dataset:
+            raise InvalidDatasetParam()
+
         with sentry_sdk.start_span(op="tagstore", description="get_tag_keys_for_projects"):
             with handle_query_errors():
                 results = tagstore.backend.get_tag_keys_for_projects(
@@ -35,6 +43,7 @@ class OrganizationTagsEndpoint(OrganizationEndpoint):
                     filter_params["start"],
                     filter_params["end"],
                     use_cache=request.GET.get("use_cache", "0") == "1",
+                    dataset=Dataset(request.GET.get("dataset", "events")),
                     # Defaults to True, because the frontend caches these tags globally
                     include_transactions=request.GET.get("include_transactions", "1") == "1",
                     tenant_ids={"organization_id": organization.id},
--- a/src/sentry/api/endpoints/organization_tags.py
+++ b/src/sentry/api/endpoints/organization_tags.py
@@ -38,7 +38,6 @@ class OrganizationTagsEndpoint(OrganizationEndpoint):
         with sentry_sdk.start_span(op="tagstore", description="get_tag_keys_for_projects"):
             with handle_query_errors():
                 results = tagstore.backend.get_tag_keys_for_projects(
-                    organization.id,
                     filter_params["project_id"],
                     filter_params.get("environment"),
                     filter_params["start"],
--- a/src/sentry/api/endpoints/organization_tags.py
+++ b/src/sentry/api/endpoints/organization_tags.py
@@ -33,7 +33,7 @@ class OrganizationTagsEndpoint(OrganizationEndpoint):
             try:
                 dataset = Dataset(request.GET.get("dataset", "discover"))
             except ValueError:
-                raise ParseError("Invalid dataset parameter")
+                raise ParseError(detail="Invalid dataset parameter")
 
         with sentry_sdk.start_span(op="tagstore", description="get_tag_keys_for_projects"):
             with handle_query_errors():
--- a/src/sentry/api/endpoints/organization_tags.py
+++ b/src/sentry/api/endpoints/organization_tags.py
@@ -1,4 +1,5 @@
 import sentry_sdk
+from rest_framework.exceptions import ParseError
 from rest_framework.request import Request
 from rest_framework.response import Response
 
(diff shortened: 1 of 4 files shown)


MichaelSun48:
If anyone would like to advise on how to best test these changes, I'm all ears. There doesn't seem to be a explicit way to check which dataset was queried from the organization_tags endpoint response, but I'm not sure what the best way to infer this is either.

src/sentry/api/endpoints/organization_tags.py
```diff
@@ -27,6 +32,9 @@ def get(self, request: Request, organization) -> Response:
         except NoProjects:
             return Response([])
 
+        if request.GET.get("dataset") and request.GET.get("dataset") not in Dataset:
+            raise InvalidDatasetParam()
```
malwilley, line 36: I think it would be better to raise a `ParseError` here instead, which sets the response as a 400 (and you'll want to include a message as well)

src/sentry/api/endpoints/organization_tags.py
```diff
@@ -35,6 +43,7 @@ def get(self, request: Request, organization) -> Response:
                     filter_params["start"],
                     filter_params["end"],
                     use_cache=request.GET.get("use_cache", "0") == "1",
+                    dataset=Dataset(request.GET.get("dataset", "events")),
                     # Defaults to True, because the frontend caches these tags globally
                     include_transactions=request.GET.get("include_transactions", "1") == "1",
```
malwilley, line 48: If this is the only caller of this function, you may be able to replace `include_transactions` with `dataset` actually. But you will still want the default behavior to be the same (dataset: Discover)

malwilley (commented):
> If anyone would like to advise on how to best test these changes, I'm all ears. There doesn't seem to be a explicit way to check which dataset was queried from the organization_tags endpoint response, but I'm not sure what the best way to infer this is either.

The _ideal_ test for this would be to add a few events with some tags, one in each different dataset. Then query the endpoint with different `dataset` params and see that the tags are correct for each response. I'm not sure if we have helpers for adding issue platform events or not, but @snigdhas might know more.

src/sentry/api/endpoints/organization_tags.py
MichaelSun48: It looks like this is indeed the only caller of this function, so i'll implement this, thanks!

src/sentry/tagstore/snuba/backend.py
```diff
@@ -218,7 +218,7 @@ def __get_tag_keys(
         limit=1000,
         keys=None,
         include_values_seen=True,
-        include_transactions=False,
```
MichaelSun48, line 221: Here and on line 428 of this file, we set the `dataset` default to be Events (instead of Discover), since `include_transaction`'s default was False

src/sentry/tagstore/snuba/backend.py
```diff
@@ -248,8 +248,8 @@ def __get_tag_keys_for_projects(
         keys=None,
         include_values_seen=True,
         use_cache=False,
-        include_transactions=False,
```
MichaelSun48, line 251: Removing include_transactions parameter here because it is redundant and strictly less customizable than the new dataset param. I have updated all usages of `__get_tag_keys_for_projects` to use the correct dataset as a result, feel free to double check.

src/sentry/api/endpoints/organization_tags.py
```diff
@@ -27,6 +29,12 @@ def get(self, request: Request, organization) -> Response:
         except NoProjects:
             return Response([])
 
+        if request.GET.get("dataset"):
+            try:
+                Dataset(request.GET.get("dataset"))
+            except ValueError:
+                raise ParseError("Invalid dataset parameter")
```
mrduncan, line 36:
```suggestion
        try:
            dataset = Dataset(request.GET.get("dataset", "discover"))
        except ValueError:
            raise ParseError("Invalid dataset parameter")
```
If you pull out the dataset here you should be able to avoid having to keep the fallback "discover" aligned in all the gets everywhere else - line 46 should be able to be just `dataset=dataset,` and line 60 is a lot easier too

MichaelSun48: Good call!

src/sentry/api/endpoints/organization_tags.py
```diff
@@ -27,6 +29,12 @@ def get(self, request: Request, organization) -> Response:
         except NoProjects:
             return Response([])
 
+        if request.GET.get("dataset"):
+            try:
+                dataset = Dataset(request.GET.get("dataset", "discover"))
+            except ValueError:
+                raise ParseError("Invalid dataset parameter")
```
malwilley, line 36: I believe you want to pass this string as a name parameter `detail="..."`

src/sentry/api/endpoints/organization_tags.py
```diff
@@ -35,8 +43,7 @@ def get(self, request: Request, organization) -> Response:
                     filter_params["start"],
                     filter_params["end"],
                     use_cache=request.GET.get("use_cache", "0") == "1",
-                    # Defaults to True, because the frontend caches these tags globally
-                    include_transactions=request.GET.get("include_transactions", "1") == "1",
```
malwilley, line 39: Just double checking that the frontend is not using `include_transactions` anywhere? If it was we would need to handle it sending `include_transactions=0`

MichaelSun48: AFAICT, at least for this endpoint there are no usages of `include_transactions` on the frontend. (A similar endpoint, `organization_tagkey_values` _does_ use this though)

malwilley (approved):
This looks good to me! If you want to be confident that this hasn't broken anything you could run the app locally and ensure that tags show up correctly in the search bar still.

PR #73949, opened 2024-07-08
fix: lint `.ts` and `.js` files in CI

gggritso:
We've been seeing issues lately where bad formatting in `.ts` files is making it into `master` and blocking deploys. This PR fixes the issue in `sentry`, `getsentry` PR coming separately but will look the same.

The flow is:

- someone makes a `.ts` change in an editor that doesn't run Prettier
- local `pre-commit` doesn't run Prettier on `.ts` files (Problem 1)
- the `pre-commit` PR check doesn't run Prettier on `.ts` files either, so there's no auto-fix
- the `prettier (changed files only)` task on the PR doesn't see any changed files, and doesn't run so there's no PR failure (Problem 2)
- the PR is merged into `master`. `typescript-and-lint` runs on _all_ files and finds the bad formatting

Problem 1: This PR updates the `pre-commit` config which will auto-lint the files before they're committed (in most cases) and make sure that `pre-commit` auto-fixes the files in the PR (in all cases)

Problem 2: This PR fixes a typo in the file names that prevented those names from being passed through to the tasks. With this fix, the `typescript-and-lint

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -24,11 +24,11 @@ jobs:
     timeout-minutes: 3
     # Map a step output to a job output
     outputs:
-      lintable_prettier_modified: ${{ steps.changes.output.lintable_prettier_modified }}
-      lintable_prettier_modified_files: ${{ steps.changes.output.lintable_prettier_modified_files }}
+      lintable_prettier_modified: ${{ steps.changes.outputs.lintable_prettier_modified }}
+      lintable_prettier_modified_files: ${{ steps.changes.outputs.lintable_prettier_modified_files }}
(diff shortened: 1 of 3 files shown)


evanpurkhiser (approved):
Nice work

joshuarli (approved):
Please see https://github.com/getsentry/sentry/pull/73508, which should fix these issues.

gggritso:
@joshuarli 👋🏻 looking! Should I hold off on merging these, and wait for you to merge _your_ PR, or merge anyway?

joshuarli:
> @joshuarli 👋🏻 looking! Should I hold off on merging these, and wait for you to merge _your_ PR, or merge anyway?

Merge anyway, you fixed a bug in `.github/workflows/frontend.yml`. The pre-commit changes I've fixed in my PR with `\.[jt]sx?$`.

PR #73951, opened 2024-07-08
feat(web-vitals): add analytics for browser filter select

KevinL10:
Adds an analytic event on browser dropdown selection.


Zylphrex (approved):
:shipit:

PR #73953, opened 2024-07-08
feat(trace-explorer): Use separate referrers for each traces related …

Zylphrex:
…query

These queries were sharing the same referrers even though they were querying different things thus consuming the same quota. Split them up to manage them independently.


k-fish:
@Zylphrex are we also considering upping the referrer limit in general for these? this may cut the load by 3x but we may need more... you should never run into timeouts in regular usage if more than one user is using the explorer

k-fish (approved):
Approving

Zylphrex:
> @Zylphrex are we also considering upping the referrer limit in general for these? this may cut the load by 3x but we may need more... you should never run into timeouts in regular usage if more than one user is using the explorer

I'm going to throw together a dashboard to monitor these referrers and will reach out to sns to bump them as needed if we find any.

PR #73959, opened 2024-07-08
feat(query-builder): Click and drag to select tokens

malwilley:
We already have the ability to `cmd + a` to select all, but this adds click and drag selection similar to a native textarea.

The bulk of the logic is contained in the `useSelectOnDrag` hook, which watches on mouse events to change the selection. Some changes in other files were necessary to get partial selection working.

https://github.com/getsentry/sentry/assets/10888943/2388f98a-1329-4e49-8adc-b0315a05c2fc


static/app/components/searchQueryBuilder/useQueryBuilderGrid.tsx
malwilley, line 1: Changes in this file allow partial selection to work well. When there is a selection, this file decides what happens on arrowleft/arrowright, cmd+c, etc.

roggenkemper (approved):
works well!! very cool feature

PR #73960, opened 2024-07-08
fix: remove appstore connect integration and tables

anonrig:
Followed the guideline available at: https://develop.sentry.dev/database-migrations/#deleting-tables as much as I can.

I basically created a migration that doesn't get executed unless we manually run it. Please let me know if this is not the preferred approach.

Once this gets merged in:
- we need to manually run the migration to remove the database tables in production.
- land another PR to remove the existing TODOs and noops.

Let's wait until July 10 to remove all data.

cc @kahest 

Ref: https://github.com/getsentry/sentry/issues/51994


src/sentry/migrations/0734_sunset_appstore_connect_integration.py
```diff
@@ -0,0 +1,42 @@
...
+        migrations.RemoveField(
+            model_name="latestappconnectbuildscheck",
+            name="project",
+        ),
+        migrations.DeleteModel(
+            name="AppConnectBuild",
+        ),
+        migrations.DeleteModel(
+            name="LatestAppConnectBuildsCheck",
+        ),
```
markstory, line 41: This migration isn't safe to deploy currently. Migrations are deployed many minutes before code is. Removing tables that django could be using will result in requests/tasks failing. Generally what we do is remove the table from django state first, and then in a follow up pull request remove the tables from schema.

src/sentry/tasks/app_store_connect.py
```diff
@@ -4,232 +4,13 @@
...
-
-
-# Untyped decorator would stop type-checking of entire function, split into an inner
-# function instead which can be type checked.
-@instrumented_task(
-    name="sentry.tasks.app_store_connect.refresh_all_builds",
-    queue="appstoreconnect",
-    ignore_result=True,
-)
-def refresh_all_builds() -> None:
```
markstory, line 185: If there are any inflight tasks when these changes are deployed you'll get errors captured in sentry.

src/sentry/migrations/0734_sunset_appstore_connect_integration.py
anonrig: If I set `is_post_deployment` to True, would it make it safe?

src/sentry/tasks/app_store_connect.py
anonrig: Should I just keep it as noop? This function gets called by a cronjob, and I already removed the cronjob declaration?

src/sentry/migrations/0734_sunset_appstore_connect_integration.py
anonrig: Actually, I updated the migration to use `migrations.SeparateDatabaseAndState`

src/sentry/tasks/app_store_connect.py
markstory: If you're confident there won't be any tasks in-flight when this is deployed removal is fine. Otherwise a no-op task will ensure you don't create an error.

src/sentry/migrations/0734_prepare_sunset_appstore_connect_integration.py
```diff
@@ -0,0 +1,36 @@
...
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
+            state_operations=[migrations.DeleteModel(name="LatestAppConnectBuildsCheck")],
+            database_operations=[],
+        ),
+        migrations.SeparateDatabaseAndState(
+            state_operations=[migrations.DeleteModel(name="AppConnectBuild")],
+            database_operations=[],
+        ),
```
markstory, line 35: You can combine these into one `SeparateDatabaseAndState` operation if you want. It can also be left this way.

markstory (approved):
Migration looks good.

Swatinem (approved):

src/sentry/migrations/0734_prepare_sunset_appstore_connect_integration.py
anonrig: It's better if we merge it. I updated the pull-request. Thanks for the review.

JoshFerge (approved):

wedamija (approved):
Migrations look good here, thanks for removing dead code

PR #73962, opened 2024-07-08
feat(query-builder): Add analytics when pressing enter to search

malwilley:
This follows the same logic that exists in SmartSearchBar, which executes some extra logic whenever the enter key is pressed:

- Records `searched.searched` analytic event  if all is valid
- Records `search.search_with_invalid` analytic event if there are any invalid tokens
- Saves the query as a recent search (if you've passed in a savedSearchType)


roggenkemper (approved):
"ruducer" is the best typo i've seen all year

PR #73963, opened 2024-07-08
feat(slack): add `/sentry support` and `/sentry docs` commands

cathteng:
https://github.com/getsentry/sentry/assets/70817427/e295cd0f-3121-4ba3-9a0b-94c732c8f5af


src/sentry/integrations/slack/webhooks/base.py
```diff
@@ -46,8 +46,8 @@ def post_dispatcher(self, request: SlackDMRequest) -> Response:
         """
         command, args = request.get_command_and_args()
 
-        if command in ["help", ""]:
-            return self.respond(SlackHelpMessageBuilder().build())
+        if command in ["help", "", "support", "docs"]:
```
iamrajjoshi, line 49: should we leave `""` currently if the user doesn't give us a command after `/sentry`, it will give the user the help response.

iamrajjoshi (commented):
LGTM, just that one question

src/sentry/integrations/slack/webhooks/base.py
iamrajjoshi: oop i didn't read

cathteng: this still works even with "" https://github.com/getsentry/sentry/pull/73963/files#diff-cd09b7d4ee99699ebb00a4fa5f4fb54620e0591930926469c7b9a25572160eafR65-R68

src/sentry/integrations/slack/message_builder/disconnected.py
```diff
@@ -1,17 +1,24 @@
...
     "Slack has been uninstalled from your Sentry organization, re-install it to continue."
 )
 
 
-class SlackDisconnectedMessageBuilder(SlackHelpMessageBuilder):
-    def __init__(self) -> None:
-        """Override parent constructor."""
-        super().__init__()
+class SlackDisconnectedMessageBuilder(BlockSlackMessageBuilder):
+    def get_docs_block(self) -> SlackBlock:
```
cathteng, line 10: this is only used here, so moved from the old parent class. we also don't need to subclass SlackHelpMessageBuilder anymore

src/sentry/integrations/slack/__init__.py
```diff
@@ -10,7 +10,6 @@
 from .message_builder.base.block import *  # noqa: F401,F403
 from .message_builder.disconnected import *  # noqa: F401,F403
 from .message_builder.discover import *  # noqa: F401,F403
-from .message_builder.event import *  # noqa: F401,F403
```
cathteng, line 13: this was unused

src/sentry/integrations/slack/message_builder/help.py
```diff
@@ -24,6 +24,13 @@
 CONTACT_MESSAGE = "Let us know if you have feedback: ecosystem-feedback@sentry.io"
 
 
+SUPPORT_HEADER_MESSAGE = "Need support? Check out these resources:"
+SUPPORT_OPTIONS_MESSAGE = "• Want help for your particular case? <https://docs.sentry.io/support|File a ticket in Zendesk>. \n• Want to report an issue, request a feature, or get support? <https://github.com/getsentry/sentry/issues/new/choose|File a GitHub issue>. \n• Have feedback? Email ecosystem-feedback@sentry.io."
```
cathteng, line 28: @leedongwei is this email correct

RyanSkonnord (approved):

PR #73967, opened 2024-07-08
feat(new-trace): Updated group hash for span summary route

Abdkhan14:
For the new span summary UI we pass the `group` hash from `sentry_tags` if it exists and nothing otherwise:
Screenshot 2024-07-08 at 1 57 01 PM


static/app/views/performance/newTraceDetails/traceDrawer/details/span/sections/generalInfo.tsx
```diff
@@ -72,6 +72,13 @@ export function GeneralInfo(props: GeneralnfoProps) {
     span.sentry_tags?.category
   );
 
+  const hasNewSpansUIFlag =
+    props.organization.features.includes('performance-spans-new-ui') &&
+    props.organization.features.includes('insights-initial-modules');
+
+  // The new spans UI relies on the group hash assigned by Relay, which is different from the hash available on the span itself.
+  const groupHash = hasNewSpansUIFlag ? span.sentry_tags?.group ?? '' : span.hash ?? '';
```
Zylphrex, line 80: `span.hash` refers to the old span hash for the old span summary page right? Does that still work? If not, we should just remove it.

Abdkhan14: @Zylphrex I just checked with the flag turned off, it works with the old ui.

Zylphrex (approved):

PR #73969, opened 2024-07-08
ref(typing): Type the org-event-stats endpoint

wmak:
- Types the org event stats endpoint


src/sentry/api/endpoints/organization_events_stats.py
```diff
@@ -390,10 +392,7 @@ def fn(
                         comparison_delta,
                     )
                     has_other_data = self.check_if_results_have_data(original_results)
-                    if isinstance(original_results, SnubaTSResult):
-                        dataset_meta = original_results.data.get("meta", {})
-                    else:
-                        dataset_meta = list(original_results.values())[0].data.get("meta", {})
```
Zylphrex, line 396: Why was the `else` block removed? This is to handle the multi-axis charts right?

wmak: lol probably should've drafted this sorry, I forgot what this block was and wanted CI to explain it

wmak: I'm not sure if its multi-axis though, when i ran the events_stats.py tests locally it passed 🤔

Zylphrex: Hm... could we just be missing test coverage? This code reads like it's taking the meta from the first timeseries.

wmak: 😨 if that's the case that's much worse lol

wmak: 🤔 finally tested this, yup its multi-axis but its not tested in events-stats 😬

src/sentry/api/endpoints/organization_events_stats.py
```diff
@@ -392,7 +394,7 @@ def fn(
                     has_other_data = self.check_if_results_have_data(original_results)
                     if isinstance(original_results, SnubaTSResult):
                         dataset_meta = original_results.data.get("meta", {})
-                    else:
+                    elif isinstance(original_results, dict):
```
Zylphrex, line 397: Can we add an else block for `dataset_meta = {}` otherwise, it's possible that `dataset_meta` is undefined.

src/sentry/api/endpoints/organization_events_stats.py
Zylphrex: Can we get a test added here? 😅

src/sentry/api/endpoints/organization_events_stats.py
wmak: Sure!

src/sentry/api/endpoints/organization_events_stats.py
wmak: I'll add them in a separate PR, don't want to overload this one

Zylphrex (approved):

PR #73970, opened 2024-07-08
feat(apis): PUT endpoint to rename user auth tokens

### LINKED ISSUE #73556: Allow User Auth Tokens to be renamed
### Problem Statement

I have existing user auth tokens. I want to assign names to them.

See https://YOU_ORG_SLUG.sentry.io/settings/account/api/auth-tokens/

Screenshot 2024-06-30 at 7 14 38 PM

### Solution Brainstorm

- Create a PUT/PATCH API endpoint that allows the name of the token to be modified.

### Product Area

Settings - Security & Privacy

ameliahsu:
closes https://github.com/getsentry/sentry/issues/73556

created a PUT API endpoint that allows the name of a token to be modified


src/sentry/api/endpoints/api_tokens.py
```diff
@@ -22,8 +22,11 @@
 from sentry.types.token import AuthTokenType
 
 
-class ApiTokenSerializer(serializers.Serializer):
+class ApiTokenNameSerializer(serializers.Serializer):
     name = CharField(max_length=255, allow_blank=True, required=False)
+
+
+class ApiTokenSerializer(ApiTokenNameSerializer):
```
cathteng, line 29: do we need to separate these out? if somebody tries to update another part of the auth token, we should probably give them some kind of error message

src/sentry/api/endpoints/api_tokens.py
```diff
@@ -119,3 +123,30 @@ def delete(self, request: Request):
...
+            with outbox_context(transaction.atomic(router.db_for_write(ApiToken)), flush=False):
+                token_to_rename: ApiToken | None = ApiToken.objects.filter(
+                    id=token_id, application__isnull=True, user_id=user_id
+                ).first()
+
+                if token_to_rename is None:
+                    return Response({"tokenId": token_id, "userId": user_id}, status=400)
+
+                token_to_rename.update(name=result.get("name", None))
+                token_to_rename.save()
```
cathteng, line 149: `update` does the job for you, you don't need to call `save` again unless you want to invoke `post_save` or `pre_save` Django signals (or have some special save logic for the model) https://docs.djangoproject.com/en/5.0/topics/db/queries/#updating-multiple-objects-at-once

although we have some special save logic here https://github.com/getsentry/sentry/blob/7d2cb635fe65068d0a2f1d1fbb0aec106d80f6d1/src/sentry/models/apitoken.py#L209-L225

so we should call `save`. so i'd change this to

```
token_to_rename.name = result.get("name", None)
token_to_rename.save()
```

src/sentry/api/endpoints/api_tokens.py
```diff
@@ -119,3 +123,30 @@ def delete(self, request: Request):
         analytics.record("api_token.deleted", user_id=request.user.id)
 
         return Response(status=204)
+
+    @method_decorator(never_cache)
+    def put(self, request: Request) -> Response:
```
cathteng, line 128: we can also error immediately if somebody tries to edit scopes
```
if “scopes” in request.data:
    return Response(403)
```

src/sentry/api/endpoints/api_tokens.py
```diff
@@ -119,3 +123,33 @@ def delete(self, request: Request):
...
+        if "scopes" in request.data:
+            return Response(status=403)
```
cathteng, line 130: oh, can we add a message to this saying you cannot edit auth token scopes after creation

tests/sentry/api/endpoints/test_api_tokens.py
```diff
@@ -153,6 +153,74 @@ def test_returns_400_when_no_token_param_is_sent(self) -> None:
...
+        assert token.name == "rename1"
+
+    def test_invalid_name(self):
+        token = ApiToken.objects.create(user=self.user)
+        self.login_as(self.user)
+        url = reverse("sentry-api-0-api-tokens")
+        response = self.client.put(
+            url,
+            data={
+                "name": "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in",
```
cathteng, line 203: nice

src/sentry/api/endpoints/api_tokens.py
```diff
@@ -119,3 +123,36 @@ def delete(self, request: Request):
...
 
         return Response(status=204)
+
+    @method_decorator(never_cache)
+    def put(self, request: Request) -> Response:
+        for key in request.data.keys():
+            if key != "name" and key != "tokenId":
+                return Response(
+                    {"error": "Only auth token name can be edited after creation"}, status=403
+                )
```
cathteng, line 133: i think we can write this a bit cleaner, we could also pull out ALLOWED_FIELDS into a top level variable

```suggestion
        keys = list(request.data.keys())
        allowed_fields = ["name", "tokenId"]
        
       if any(key not in allowed_fields for key in keys):
            return Response(
                {"error": "Only auth token name can be edited after creation"}, status=403
            )
```

cathteng (approved):

PR #73971, opened 2024-07-08
feat(new-trace): Updated logic for routing upon event id click from discover

Abdkhan14:
We should route to issue details for events that are not of type `transaction` or are not associated to a `trace` id

### DIFF
--- a/static/app/views/discover/table/tableView.tsx
+++ b/static/app/views/discover/table/tableView.tsx
@@ -190,7 +190,10 @@ function TableView(props: TableViewProps) {
 
       let target;
 
-      if (dataRow.trace === null || dataRow.trace === undefined) {
+      if (
+        !dataRow.trace ||
+        ['error', 'default'].includes(dataRow['event.type'].toString())
+      ) {
         if (dataRow['event.type'] === 'transaction') {
           throw new Error(
             'Transaction event should always have a trace associated with it.'
@@ -321,7 +324,10 @@ function TableView(props: TableViewProps) {
     if (columnKey === 'id') {
       let target;
 
-      if (dataRow.trace === null || dataRow.trace === undefined) {
+      if (
+        !dataRow.trace ||
+        ['error', 'default'].includes(dataRow['event.type'].toString())
+      ) {
         if (dataRow['event.type'] === 'transaction') {
           throw new Error(
             'Transaction event should always have a trace associated with it.'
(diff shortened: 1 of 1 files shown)


static/app/views/discover/table/tableView.tsx
```diff
@@ -190,7 +190,10 @@ function TableView(props: TableViewProps) {
 
       let target;
 
-      if (dataRow.trace === null || dataRow.trace === undefined) {
+      if (
+        !dataRow.trace ||
+        ['error', 'default'].includes(dataRow['event.type'].toString())
```
Zylphrex, line 195: I think this should be `dataRow['event.type'].toString() != 'transaction'` as we have other error event types such as csp and a few more that I don't recall.

static/app/views/discover/table/tableView.tsx
```diff
@@ -321,7 +324,10 @@ function TableView(props: TableViewProps) {
     if (columnKey === 'id') {
       let target;
 
-      if (dataRow.trace === null || dataRow.trace === undefined) {
+      if (
+        !dataRow.trace ||
+        ['error', 'default'].includes(dataRow['event.type'].toString())
```
Zylphrex, line 329: Same here

Zylphrex (commented):
Just double checking but `event.type` is always defined right?

static/app/views/discover/table/tableView.tsx
Abdkhan14: Ahh I se

static/app/views/discover/table/tableView.tsx
```diff
@@ -190,13 +190,7 @@ function TableView(props: TableViewProps) {
...
       let target;
 
-      if (dataRow.trace === null || dataRow.trace === undefined) {
-        if (dataRow['event.type'] === 'transaction') {
-          throw new Error(
-            'Transaction event should always have a trace associated with it.'
-          );
-        }
-
+      if (dataRow['event.type'] !== 'transaction') {
```
Zylphrex, line 193: Nit: This is more of a stylistic thing but can we invert this condition

```typescript
if (dataRow['event.type'] === 'transaction) {
  // the stuff current in side else block
} else {
  // the stuff currently in the if block
}
```

Makes this easier to reason about at a glance with fewer negations.

static/app/views/discover/table/tableView.tsx
```diff
@@ -321,13 +321,7 @@ function TableView(props: TableViewProps) {
...
```
Zylphrex, line 324: Same thing here.

Zylphrex (approved):

PR #73976, opened 2024-07-08
fix(similarity): Add retries to snuba query in backfill

jangjodi:
Add retries to snuba query in backfill if RateLimitExceeded occurs

### DIFF
--- a/src/sentry/tasks/embeddings_grouping/utils.py
+++ b/src/sentry/tasks/embeddings_grouping/utils.py
@@ -37,7 +37,7 @@ from sentry.utils import json, metrics
 from sentry.utils.iterators import chunked
 from sentry.utils.query import RangeQuerySetWrapper
 from sentry.utils.safe import get_path
-from sentry.utils.snuba import bulk_snuba_queries
+from sentry.utils.snuba import RateLimitExceeded, bulk_snuba_queries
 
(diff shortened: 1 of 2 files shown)


src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -236,14 +236,40 @@ def get_data_from_snuba(project, groups_to_backfill_with_no_embedding):
             snuba_requests.append(request)
 
         with metrics.timer(f"{BACKFILL_NAME}.bulk_snuba_queries", sample_rate=1.0):
```
jangjodi, line 238: note that this timer will now include these long attempts with retries

JoshFerge (approved):
nice!

src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -551,17 +579,18 @@ def lookup_group_data_stacktrace_bulk(
...
             return operation(*args, **kwargs)
-        except (ServiceUnavailable, DeadlineExceeded):
+        except exceptions:
             if attempt < retries - 1:
                 time.sleep(delay * (2**attempt))
             else:
                 raise
 
 
+# TODO: delete this and its tests
```
armenzg, line 593: Do you remember why you added this note?

PR #73977, opened 2024-07-08
feat(uptime): Add bin script to add uptime configs

evanpurkhiser:
I am planning to use this to load up a large number of configurations to start doing some load testing / tuning of the uptime-checker service in S4S.


evanpurkhiser:
Moving to getsentry

Issue #37410, opened 2022-08-03
Some internal project id instead of project name


pva:
### Environment

self-hosted (https://develop.sentry.dev/self-hosted/)

### Version

 22.8.0.dev0

### Steps to Reproduce

1. Cloned https://github.com/getsentry/self-hosted.git
2. Configured a bit and run
3. I created a new project named "auth". 

Instead of the project name, I see 097bd3cc54a2:
![изображение](https://user-images.githubusercontent.com/92356/182586645-0df59931-3c98-4866-a04b-efe10d1433ff.png)

Project name, that is auth, I see only in settings...

### Expected Result

I expect to see auth everywhere I see 097bd3cc54a2.

### Actual Result

![изображение](https://user-images.githubusercontent.com/92356/182586847-fd1ae5b7-7700-40e1-88f8-022a4721ee0b.png)
![изображение](https://user-images.githubusercontent.com/92356/182586914-8e7f40a3-c93d-4cbe-85d9-a616456f213a.png)

pva:
I've tried Sentry 22.7.0 and I still have the same issue.

getsentry-release:
Routing to @getsentry/open-source for [triage](https://develop.sentry.dev/processing-tickets/#3-triage). ⏲️

ethanhs:
Hm, I wonder if we treat "auth" specially. I will try to reproduce this later today.

ethanhs:
So I was able to reproduce this but I don't think it is related to open source. Routing to the frontend team to look at this. I'm guessing we show the hash of the name since auth is special?

getsentry-release:
Routing to @getsentry/app-frontend for [triage](https://develop.sentry.dev/processing-tickets/#3-triage). ⏲️

dashed:
@pva @ethanhs It's a reserved project slug. The reserved list for project slugs are here: https://github.com/getsentry/sentry/blob/67e9fb66b7b074009de46b7e63cb4bc799936463/src/sentry/constants.py#L157-L178

ethanhs:
@dashed do you know if it would be possible to show the name of a project even if it is reserved? Or just reject projects with those names? The current behavior is a bit un-intuitive.

dashed:
@ethanhs I agree. I think rejecting project slugs might be the right way.

billsbooth:
I renamed my project, and the entire build system has hopelessly imploded.

hubertdeng123:
[...]

PR #73979, opened 2024-07-08
fix(project-creation): Reject project names with reserved slugs (#37410)

Christinarlong:
Adds validation during project creation to check if the project name is any of the RESERVED_PROJECT_SLUGS

ref (#37410)

After:
image

### DIFF
--- a/tests/sentry/api/endpoints/test_team_projects.py
+++ b/tests/sentry/api/endpoints/test_team_projects.py
@@ -1,7 +1,7 @@
 from unittest.mock import Mock
 
 from sentry.api.endpoints.organization_projects_experiment import DISABLED_FEATURE_ERROR_STRING
-from sentry.constants import RESERVED_ORGANIZATION_SLUGS
+from sentry.constants import RESERVED_PROJECT_SLUGS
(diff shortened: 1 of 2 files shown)


cathteng (commented):
lgtm, will wait for @dashed though

dashed (approved):

tests/sentry/api/endpoints/test_team_projects.py
```diff
@@ -96,6 +97,18 @@ def test_invalid_platform(self):
         )
         assert response.data["platform"][0] == "Invalid platform"
 
+    def test_invalid_name(self):
+
+        invalid_name = list(RESERVED_PROJECT_SLUGS)[0]
```
dashed, line 102: You could make this a for loop.

Christinarlong: Yeah, I didn't know if there's a precedent or preference for that because the two other tests I could find only used the first index/member for their test. I guess if the list gets super long the test will take a while, but idk if it's a big concern ?

dashed (approved):

PR #73982, opened 2024-07-08
feat(uptime): Add additional logs for detection

davidenwang:
Add additional logs to get an idea of how many events -> urls get added to ranking buckets, and which projects + urls are being processed


evanpurkhiser (approved):
This might be too much logging once we bump the rollout higher, but I think this is fine for now.

PR #73985, opened 2024-07-08
chore(slack): remove old slack client from link commands

cathteng:
Default to using the new Slack SDK Client when somebody tries to link / unlink a team through the Slack command. Also adds metrics for the success / failures of these calls (either through the webhook client or chat_postMessage).

Added logic to ignore the error if the url has expired, that's on the user for not clicking soon enough.

### DIFF
--- a/src/sentry/integrations/slack/utils/notifications.py
+++ b/src/sentry/integrations/slack/utils/notifications.py
@@ -18,16 +18,16 @@ from sentry.integrations.repository.metric_alert import (
     NewMetricAlertNotificationMessage,
 )
 from sentry.integrations.services.integration import integration_service
-from sentry.integrations.slack.client import SlackClient
 from sentry.integrations.slack.message_builder.incidents import SlackIncidentsMessageBuilder
 from sentry.integrations.slack.metrics import (
+    SLACK_COMMANDS_LINK_IDENTITY_FAILURE_DATADOG_METRIC,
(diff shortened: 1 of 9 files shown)


src/sentry/integrations/slack/utils/notifications.py
```diff
@@ -169,14 +169,24 @@ def respond_to_slack_command(
...
             webhook_client = WebhookClient(params.response_url)
             webhook_client.send(text=text, replace_original=False, response_type="ephemeral")
+            metrics.incr(
+                SLACK_COMMANDS_LINK_IDENTITY_SUCCESS_DATADOG_METRIC,
+                sample_rate=1.0,
+                tags={"type": "webhook", "command": command},
+            )
         except (SlackApiError, SlackRequestError) as e:
-            logger.exception(log + "error", extra={"error": str(e)})
+            if "Expired url" not in str(e):
```
iamrajjoshi, line 183: i wish the sdk had more specific error types

iamrajjoshi (approved):

src/sentry/integrations/slack/utils/notifications.py
RyanSkonnord: We should spike on building a utility around this. I feel like our own list of expected error text (perhaps an enum class), and a common function for parsing them out, would go a long way.

RyanSkonnord (approved):

PR #73988, opened 2024-07-08
ref(replay): Move currentHoverTime into its own Context and hook

ryan953:
I'm slicing some non-critical stuff out of the main `useReplayContext()`, and the currentHoverTime has so few dependencies, it's a perfect candidate.

### DIFF
--- a/static/app/utils/replays/playback/providers/useCurrentHoverTime.tsx
+++ b/static/app/utils/replays/playback/providers/useCurrentHoverTime.tsx
@@ -1,0 +1,16 @@
+import type {Dispatch, ReactNode, SetStateAction} from 'react';
+import {createContext, useContext, useState} from 'react';
+
+type ContextType = [undefined | number, Dispatch<SetStateAction<number | undefined>>];
+
+const context = createContext<ContextType>([undefined, () => {}]);
+
+export function ReplayCurrentTimeContextProvider({children}: {children: ReactNode}) {
+  const state = useState<undefined | number>(undefined);
+
+  return <context.Provider value={state}>{children}</context.Provider>;
+}
+
+export default function useCurrentHoverTime() {
+  return useContext(context);
+}
--- a/static/app/utils/replays/replayPlayerTimestampEmitter.tsx
+++ b/static/app/utils/replays/replayPlayerTimestampEmitter.tsx
@@ -31,7 +31,7 @@ class ReplayPlayerTimestampEmitter {
 }
 
 /**
- * @deprecated This emitter is a singlton that sends some global state around.
+ * @deprecated This emitter is a singleton that sends some global state around.
  * If there are multiple replay instances on the page values will be confusing.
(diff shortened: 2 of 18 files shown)


static/app/utils/replays/playback/providers/useCurrentHoverTime.tsx
```diff
@@ -0,0 +1,16 @@
+import type {Dispatch, ReactNode, SetStateAction} from 'react';
+import {createContext, useContext, useState} from 'react';
+
+type ContextType = [undefined | number, Dispatch<SetStateAction<number | undefined>>];
+
+const context = createContext<ContextType>([undefined, () => {}]);
+
+export function ReplayCurrentTimeContextProvider({children}: {children: ReactNode}) {
+  const state = useState<undefined | number>(undefined);
```
ryan953, line 9: It's still using state, just like before.
The `currentTime` has the potential to update every 16ms and cause lots of re-renders... but i think for this one, because it's driven by mouse-hover, it'll still cause lots of re-renders as the mouse moves around, but it's directly related to what the user is doing. 
If we want to improve perf in the future we can add a slight debounce to the `setState` call, and/or we could have this accept a callback instead of return the `state` value, the callback would allow us to consume the value and transform it into something more stable/memoizeable depending on the callsite.

For now, i'm happy! `<ReplayContext>` is smaller

static/app/utils/replays/replayPlayerTimestampEmitter.tsx
```diff
@@ -30,4 +30,10 @@ class ReplayPlayerTimestampEmitter {
   }
 }
 
+/**
+ * @deprecated This emitter is a singlton that sends some global state around.
```
billyvg, line 34:
```suggestion
 * @deprecated This emitter is a singleton that sends some global state around.
```

static/app/views/replays/detail/trace/replayTransactionContext.tsx
```diff
@@ -82,6 +83,7 @@ const TxnContext = createContext<TxnContextProps>({
 function ReplayTransactionContext({children, replayRecord}: Options) {
   const api = useApi();
   const organization = useOrganization();
+  useEmitTimestampChanges();
```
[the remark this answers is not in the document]
ryan953: it couldn't stay in `replayContext` because it reads from the new `useCurrentTimeHook`, so it had to be a child of that. 

I only put it in this one spot because it's used to communicate between the trace tab and the replay, to make the vertical purple lines showup in the trace tab. 
There are a few places that create a `<ReplayContext>` provider, and this `replayTransactionContext` is a child of one of those callsites. I looked at the other callsites and I don't think this hook is needed there because those places are for the different, mostly static, replay rendering cases. For example there's the diff tools stuff, issue page snippet rendering, none of them show a trace-view and replay that need to be sync'd up.

michellewzhang (approved):

PR #73989, opened 2024-07-09
feat(replay): support case-insensitive substring queries from replay search bar

aliu39:
A follow-up on https://github.com/getsentry/sentry/pull/73537. This _could_ be implemented for issues search bar too, but isn't.
Right now:
Screenshot 2024-07-08 at 5 16 28 PM


src/sentry/replays/query.py
```diff
@@ -221,7 +223,9 @@ def query_replays_dataset_tagkey_values(
         aggregated_column = Function("identity", parameters=[grouped_column], alias="tag_value")
 
     if tag_like:
-        where.append(Condition(grouped_column, Op.LIKE, tag_like))
+        where.append(
+            Condition(Function("lower", parameters=[grouped_column]), Op.LIKE, tag_like.lower())
```
cmanallen, line 227: `ILIKE` is case insensitive.

cmanallen (changes requested):

src/sentry/replays/query.py
Zylphrex, line 227: Just putting this here as a FYI, wild card searches on strings for other datasets previously used ClickHouse's [match](https://clickhouse.com/docs/en/sql-reference/functions/string-search-functions#match) function but was timing out on the spans dataset. One workaround I found was to use [positionCaseInsensitive](https://clickhouse.com/docs/en/sql-reference/functions/string-search-functions#positioncaseinsensitive) for these types of queries which performed a lot better. (I also have special cases for startsWith and endsWith).

I haven't tested the performance of `LIKE` but just wanted to share this in case you find it useful.

aliu39: ILIKE doesn't seem to be a snuba op.. is there another way to use it in a condition? https://github.com/getsentry/sentry/blob/343fd08cd8baaa4de18351d49549da69e0c7e418/.venv/lib/python3.11/site-packages/snuba_sdk/conditions.py#L21-L21

aliu39: Thanks! Did you implement it with something like
`Condition(Function("positionCaseInsensitive", parameters=[]), Op.NEQ, 0)`?

cmanallen: Could be a good opportunity to open a PR on the snuba-sdk repo.

cmanallen: This is how we do case insensitive search for replays currently: https://github.com/getsentry/sentry/blob/d14d08fcadfc054a3261355db3ff7e2dd7883dcb/src/sentry/replays/lib/new_query/conditions.py#L133-L136

So you can ignore ilike and use `match`.

aliu39: Kk, we could also use positionCaseInsensitive like Tony suggests below https://github.com/getsentry/sentry/pull/73989#discussion_r1670964386

cmanallen (approved):

PR #73990, opened 2024-07-09
feat(useFeedbackWidget): Add a more flexible hook for opening feedback forms

malwilley:
Adds a new `useFeedbackForm()` hook which simply returns a `openForm()` function which opens the form.

While doing so, extracted the `feedback` and `options` creation to `useFeedback()` which is now used by both `useFeedbackForm()` and `useFeedbackWidget()`

### DIFF
--- a/static/app/components/feedback/widget/useFeedback.tsx
+++ b/static/app/components/feedback/widget/useFeedback.tsx
@@ -1,0 +1,36 @@
+import {useMemo} from 'react';
+import * as Sentry from '@sentry/react';
+
+import {t} from 'sentry/locale';
+import ConfigStore from 'sentry/stores/configStore';
+import {useLegacyStore} from 'sentry/stores/useLegacyStore';
+import useAsyncSDKIntegrationStore from 'sentry/views/app/asyncSDKIntegrationProvider';
+
+type FeedbackIntegration = NonNullable<ReturnType<typeof Sentry.getFeedback>>;
+
+export type UseFeedbackOptions = {
+  formTitle?: string;
+  messagePlaceholder?: string;
+};
+
+export function useFeedback({formTitle, messagePlaceholder}: UseFeedbackOptions) {
+  const config = useLegacyStore(ConfigStore);
+  const {state} = useAsyncSDKIntegrationStore();
+
+  // TODO(ryan953): remove the fallback `?? Sentry.getFeedback()` after
+  // getsentry is calling `store.add(feedback);`
+  const feedback =
+    (state.Feedback as FeedbackIntegration | undefined) ?? Sentry.getFeedback();
+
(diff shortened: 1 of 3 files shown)


static/app/components/feedback/widget/useFeedback.tsx
```diff
@@ -0,0 +1,36 @@
...
+};
+
+export function useFeedback({formTitle, messagePlaceholder}: UseFeedbackOptions) {
+  const config = useLegacyStore(ConfigStore);
+  const {state} = useAsyncSDKIntegrationStore();
+
+  // TODO(ryan953): remove the fallback `?? Sentry.getFeedback()` after
+  // getsentry is calling `store.add(feedback);`
+  const feedback =
+    (state.Feedback as FeedbackIntegration | undefined) ?? Sentry.getFeedback();
```
ryan953, line 23: This todo is to-done! Can drop the fallback now. https://github.com/getsentry/getsentry/pull/14390 was the one.

static/app/components/feedback/widget/useFeedbackForm.tsx
```diff
@@ -0,0 +1,22 @@
...
+    if (!feedback) {
+      return;
+    }
+
+    const form = await feedback.createForm(options);
+    form.appendToDom();
+    form.open();
+  }, [feedback, options]);
+
+  return {feedback, openForm};
```
ryan953, line 21: probably also want to cleanup the form sometime too right?

malwilley: Does the form need to be cleaned up? It should be removed from the dom when the user closes it, right?

ryan953: it'll render with `open={false}` so there'll be an empty div hanging around, but the object pointing to it with all the methods should be eligible for GC

malwilley: Ah yes I do see it leaving behind a div. I agree, added a cleanup on unmount. Will make using it in my use case a bit more difficult, but makes sense.

 I could actually see this hook living at the top level of our app and all consumers would use the same form instance, actually. It would work similarly to our global modal component then.

ryan953: > actually see this hook living at the top level of our app and all consumers would use the same form instance, actually. It would work similarly to our global modal component then.

We could probably set that up now: do like a Provider that is like `<context.Provider value={{openForm, removeForm}}>{children}</context.Provider>` and then this `useFeedbackForm()` is the Consumer for that. Then any callsites in the app would share the same `formRef`

ryan953 (approved):
fine refactor as-is

PR #73998, opened 2024-07-09
ref(hc): Refactor "convert to async response" tasks

RyanSkonnord:
Refactor duplicated code out of the two tasks (for Slack and Discord) into a new abstract base class. Replace typed dicts with a new `_AsyncResult` dataclass.

### DIFF
--- a/src/sentry/middleware/integrations/tasks.py
+++ b/src/sentry/middleware/integrations/tasks.py
@@ -53,6 +53,8 @@ class _AsyncRegionDispatcher(ABC):
         )
 
         if successes:
+            # Typically we expect only one request to be made or only one successful
+            # response. If there are multiple, forward one arbitrarily.
             return self._forward_response(successes[-1])
         else:
             return None
--- a/src/sentry/middleware/integrations/tasks.py
+++ b/src/sentry/middleware/integrations/tasks.py
@@ -1,5 +1,7 @@
 import logging
-from collections.abc import MutableMapping
+from abc import ABC, abstractmethod
+from collections.abc import Iterable
+from dataclasses import dataclass
 from typing import Any, cast
 
 import orjson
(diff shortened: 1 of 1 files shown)


src/sentry/middleware/integrations/tasks.py
```diff
@@ -11,11 +13,100 @@
...
+        logger.info(
+            self.log_message("async_region_response"),
+            extra={
+                "regions": [r.region.name for r in successes],
+                "response_map": {r.region.name: r.response.status_code for r in results},
+            },
+        )
+
+        if successes:
+            return self._forward_response(successes[-1])
```
RyanSkonnord, line 56: Using the last successful response replicates what the old code was doing in [lines 46-48](https://github.com/getsentry/sentry/pull/73998/files#diff-b2cbbebe9bbbe1cbfe72cb721b6efdd7c8fa211e0df2da77b39a5ecbccaf7947L46-L48). I'm not sure whether the original intent was

1. that we expect at most one successful response,
2. that we can forward any arbitrary one if we get more than one, or
3. that we intentionally should forward the last one.

I don't think it's 3 (because that would imply that the `region_names` argument is meaningfully ordered, which is not consistent with how this task is invoked AFAIK), but it was the easiest way to replicate the exact behavior. At least now it's a bit more explicit (`[-1]`) and we have a good place if we want to refine it.

markstory: I think it is a combination of 1 & 2.

markstory (approved):
Looks good to me.

dashed (approved):

src/sentry/middleware/integrations/tasks.py
RyanSkonnord: Thanks. [Commented](https://github.com/getsentry/sentry/pull/73998/commits/095fc9a7d039dd4ac0ebc15abcc1002143087d00) accordingly.

src/sentry/middleware/integrations/tasks.py
```diff
@@ -11,11 +13,102 @@
...
+            response_payload = self.unpack_payload(result.response)
+            if response_payload is None:
+                return None
+            integration_response = requests.post(self.response_url, json=response_payload)
+        except Exception as exc:
+            sentry_sdk.capture_exception(exc)
+            return None
+        else:
+            logger.info(
+                "discord.async_integration_response",
```
schew2381, line 92: tiny nit: I think the log msg needs to be generic here

PR #74000, opened 2024-07-09
feat(metrics-extraction): Create and edit rule dialogs

### LINKED ISSUE #73937: Move creation/edit modal to metrics page
**Problem**: Configuring and editing a span-based metric has quite some friction, because the user has to go to a separate page and perform a series of (potentially unnecessary) clicks:
- go from the metrics screen to the settings screen
- click on add new metric
- fill in the requested info and save
- find their way back to the metrics page

**Solution**: In order to remove some friction, we could have the "Add New Metric" and "Metric Settings" buttons just open the modal directly on the Metrics explorer page. In that case the user would have to:
- click one button
- fill in the form and save

Image

**Caveats**:
- if you have multiple projects selectors, you need you to specify a project
- if a metric is being edited/deleted and there are saved alerts/dashboard widgets we need to warn you about

ArthurKnaus:
https://github.com/getsentry/sentry/assets/7033940/956f1515-9d06-4947-b2c5-3a58b1937d19

Closes https://github.com/getsentry/sentry/issues/73937


static/app/views/metrics/metricQueryContextMenu.tsx
```diff
@@ -162,14 +160,17 @@ export function MetricQueryContextMenu({
...
-              router
-            );
+            const extractionRule = getExtractionRule(metricsQuery.mri);
+            if (extractionRule) {
+              openExtractionRuleEditModal(
+                {
+                  metricExtractionRule: extractionRule,
+                },
+                {
+                  onClose: refetch,
```
vgrozdanic, line 170: Shouldn't we refetch only if the "Save" action happened, there is no need to refetch all the rules if user opened the modal and closed it without performing any action?

If this is too complicated to do, I understand why are we refetching `onClose` and we can keep it simple for now, but in that case I would still add a TODO comment

ArthurKnaus: true! ideally, I would update and invalidate queries. Though I am not sure how good we are with partially matching query keys in our react-query setup.
Will check in a follow-up 👍

static/app/views/metrics/pageHeaderActions.tsx
```diff
@@ -146,12 +149,7 @@ export function PageHeaderActions({showAddMetricButton, addCustomMetric}: Props)
         (hasCustomMetricsExtractionRules(organization) ? (
           <Button
             priority="primary"
-            onClick={() =>
-              navigateTo(
-                `/settings/projects/:projectId/metrics/configure-metric/`,
-                router
-              )
-            }
+            onClick={() => openExtractionRuleCreateModal({}, {onClose: refetch})}
```
vgrozdanic, line 152: Just commenting so it is not missed in the follow up, if possible, here also a refetch should only happen if the new metric is created

static/app/views/metrics/metricQueryContextMenu.tsx
ArthurKnaus: nvm, it actually was really easy -> done 👍

vgrozdanic (approved):
Nice 💪

matejminar (approved):
Nice!

Issue #73179, opened 2024-06-24
Implement search of the metric extraction rules on /metrics/extraction-rules/


vgrozdanic:
search by span attribute

PR #74003, opened 2024-07-09
feat(metrics) enable query filter for metrics extraction rules

vgrozdanic:
Allow filtering of metric extraction rules on API endpoint using `query` parameter.

Part of: https://github.com/getsentry/sentry/issues/73179

### DIFF
--- a/src/sentry/api/endpoints/project_metrics_extraction_rules.py
+++ b/src/sentry/api/endpoints/project_metrics_extraction_rules.py
@@ -84,8 +84,12 @@ class ProjectMetricsExtractionRulesEndpoint(ProjectEndpoint):
         if not self.has_feature(project.organization, request):
             return Response(status=404)
 
+        query = request.GET.get("query")
+
         with handle_exceptions():
             configs = SpanAttributeExtractionRuleConfig.objects.filter(project=project)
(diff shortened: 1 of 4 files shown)


src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -84,8 +84,12 @@ def get(self, request: Request, project: Project) -> Response:
         if not self.has_feature(project.organization, request):
             return Response(status=404)
 
+        query = request.GET.get("query")
```
shellmayr, line 87: nit: can we find a better name for this, especially since query already has a meaning in the context of the UI?

vgrozdanic: Usually in our API for this type of filtering we use a `query` parameter. I agree that here it can be confusing, and maybe some other parameter name could be better, but we just have to be aware that in that case we are breaking a "convention".
Do you have any suggestion for the parameter name?

shellmayr: Ah yeah you're right

shellmayr (approved):

PR #74010, opened 2024-07-09
ref(related_issues): Improve typing

armenzg:
This change helps typing different types of events (see https://www.typescriptlang.org/play/?#example/discriminate-types for details).

Changes included:
* `TimelineEvent` becomes the union of three types (instead of two)
* We add `event.type` to the list of field we fetch from the `events` endpoint
* We do not fetch the `message` field for error and default events


static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -10,7 +10,7 @@ import {trackAnalytics} from 'sentry/utils/analytics';
 import useOrganization from 'sentry/utils/useOrganization';
 import useProjectFromSlug from 'sentry/utils/useProjectFromSlug';
 
-import type {TimelineEvent, TimelineIssuePlatformEvent} from './useTraceTimelineEvents';
+import type {TimelineEvent} from './useTraceTimelineEvents';
```
armenzg, line 13: We do not need to do this anymore:
`-      const issuePlatformEvent = event as TimelineIssuePlatformEvent;`

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
```diff
@@ -92,19 +92,15 @@ export function getTitleSubtitleMessage(event: TimelineEvent) {
       // It uses metadata.value which could differ depending on what error.value is used in the event manager
       // TODO: Add support for chained exceptions since we grab the value from the first stack trace
       // https://github.com/getsentry/sentry/blob/a221f399d2b4190f2631fcca311bdb5b3748838b/src/sentry/eventtypes/error.py#L115-L134
-      message = event['error.value'].at(-1);
+      message = event['error.value'].at(-1) || '';
```
armenzg, line 95: Otherwise, we get this error: `Type 'string | undefined' is not assignable to type 'string'.`

static/app/views/issueDetails/traceTimeline/traceLink.spec.tsx
armenzg, line 1: All changes in this file are because typing is now working properly.

static/app/views/issueDetails/traceTimeline/traceTimelineEvents.tsx
```diff
@@ -168,7 +168,7 @@ function NodeGroup({
                 groupEvents
                   .slice(0, 5)
                   .map(groupEvent =>
-                    'event.type' in groupEvent ? (
+                    'event.type' in ['error', 'default'] ? (
```
armenzg, line 171: Before, `event.type` would only be a field for events from the Discover dataset.

static/app/views/issueDetails/traceTimeline/useTraceTimelineEvents.tsx
```diff
@@ -17,16 +17,23 @@ interface BaseEvent {
...
+  'event.type': 'default';
+}
+export interface TimelineErrorEvent extends BaseEvent {
   'error.value': string[]; // Used for message for error events
-  'event.type': string;
+  'event.type': 'error';
   'stack.function': string[];
 }
 
-export type TimelineEvent = TimelineDiscoverEvent | TimelineIssuePlatformEvent;
```
armenzg, line 29: We now have the union of three interfaces.

static/app/views/issueDetails/traceTimeline/useTraceTimelineEvents.tsx
```diff
@@ -17,16 +17,23 @@ interface BaseEvent {
   transaction: string;
 }
 
-export interface TimelineIssuePlatformEvent extends BaseEvent {
+interface TimelineIssuePlatformEvent extends BaseEvent {
+  'event.type': '';
```
armenzg, line 21: This is a new field we're going to start fetching. It just happens to always be the empty string!

static/app/views/issueDetails/traceTimeline/useTraceTimelineEvents.tsx
```diff
@@ -17,16 +17,23 @@ interface BaseEvent {
...
   message: string; // Used for message for issue platform events
 }
-interface TimelineDiscoverEvent extends BaseEvent {
+interface TimelineDefaultEvent extends BaseEvent {
+  'event.type': 'default';
```
armenzg, line 25: This is now a literal type rather than `string`.

static/app/views/issueDetails/traceTimeline/useTraceTimelineEvents.tsx
```diff
@@ -17,16 +17,23 @@ interface BaseEvent {
   transaction: string;
 }
 
-export interface TimelineIssuePlatformEvent extends BaseEvent {
+interface TimelineIssuePlatformEvent extends BaseEvent {
+  'event.type': '';
   message: string; // Used for message for issue platform events
 }
-interface TimelineDiscoverEvent extends BaseEvent {
```
armenzg, line 23: This type is now split into two: `TimelineDefaultEvent` and `TimelineErrorEvent`

static/app/views/issueDetails/traceTimelineOrRelatedIssue.spec.tsx
```diff
@@ -51,25 +54,25 @@ describe('TraceTimeline & TraceRelated Issue', () => {
...
         transaction: 'n/a',
         culprit: '/api/slow/',
+        'event.type': '',
       },
     ],
     meta: {fields: {}, units: {}},
   };
-  const mainError = {
-    message: 'This is the message for the issue',
+  const mainError: TimelineErrorEvent = {
```
armenzg, line 62: It seems that this is needed to make it work.

static/app/views/issueDetails/traceTimelineOrRelatedIssue.spec.tsx
```diff
@@ -340,16 +344,17 @@ function createEvent({
...
   culprit: string;
   title: string;
   error_value?: string[];
-  event_type?: string;
+  event_type?: 'default' | 'error' | '';
   message?: string;
+  stack_function?: string[];
 }) {
   const event = {
-    message: message,
```
armenzg, line 352: This is now not a shared property for all three types.

static/app/views/issueDetails/traceTimelineOrRelatedIssue.spec.tsx
```diff
@@ -441,7 +460,7 @@ describe('getTitleSubtitleMessage()', () => {
           message: '/api/slow/ Slow DB Query SELECT "sentry_monitorcheckin"."monitor_id"',
           culprit: '/api/slow/',
           title: 'Slow DB Query',
-          event_type: 'transaction',
+          event_type: '',
```
armenzg, line 463: `transaction` was the incorrect value.

static/app/views/issueDetails/traceTimeline/traceIssue.tsx
malwilley: nit: it has the same behavior here, but we generally want to use `??` unless you really need `||`

static/app/views/issueDetails/traceTimeline/traceTimelineEvents.tsx
malwilley: Hmm this line will always return false as written. I think you want `['error', 'default'].includes(groupEvent.['event.type'])` instead

malwilley (commented):
This is looking like a great improvement!

static/app/views/issueDetails/traceTimeline/traceTimelineEvents.tsx
armenzg: I have no idea how I wrote that line of code!!

Anyway, I fixed it.

malwilley (approved):
Thanks for fixing this up!

PR #74012, opened 2024-07-09
ref(metrics): Remove permission warning

priscilawebdev:
### Goal

The banner shall be displayed only for the old experience, when users need the minium admin permission to edit the metrics configuration

### DIFF
--- a/static/app/views/settings/projectMetrics/projectMetrics.tsx
+++ b/static/app/views/settings/projectMetrics/projectMetrics.tsx
@@ -67,7 +67,9 @@ function ProjectMetrics({project}: Props) {
         )}
       </TextBlock>
 
-      <PermissionAlert project={project} />
+      {hasMetricsExtrapolationFeature(organization) ? null : (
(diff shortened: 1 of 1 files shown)


ArthurKnaus (approved):

static/app/views/settings/projectMetrics/projectMetrics.tsx
```diff
@@ -67,7 +67,9 @@ function ProjectMetrics({project}: Props) {
         )}
       </TextBlock>
 
-      <PermissionAlert project={project} />
+      {hasMetricsExtrapolationFeature(organization) ? null : (
```
matejminar, line 70: shouldn't this be `hasCustomMetricsExtractionRules`?

priscilawebdev: actually we want to check if `hasExtractionRules` is available. If `True` we do not display the permission warning. I have update the code. Thanks

PR #74015, opened 2024-07-09
feat(profiling) fetch continuous profile chunk

JonasBa:
Add fetching of continuous profile chunk

### DIFF
--- a/static/app/routes.tsx
+++ b/static/app/routes.tsx
@@ -1980,25 +1980,23 @@ function buildRoutes() {
         component={make(() => import('sentry/views/profiling/differentialFlamegraph'))}
       />
       <Route
-        path="profile/:projectId/:eventId/"
-        component={make(() => import('sentry/views/profiling/profilesProvider'))}
-      >
-        <Route
-          path="flamegraph/"
-          component={make(() => import('sentry/views/profiling/profileFlamechart'))}
-        />
-      </Route>
-      <Route
         path="profile/:projectId/"
-        component={make(
-          () => import('sentry/views/profiling/continuousProfilesProvider')
-        )}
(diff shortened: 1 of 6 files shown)


static/app/routes.tsx
```diff
@@ -1980,25 +1980,23 @@ function buildRoutes() {
         component={make(() => import('sentry/views/profiling/differentialFlamegraph'))}
       />
       <Route
-        path="profile/:projectId/:eventId/"
-        component={make(() => import('sentry/views/profiling/profilesProvider'))}
+        path="profile/:projectId/"
```
JonasBa, line 1983: I had to reorder this as the evenId param was intercepting the route. Since we dont have tests for routing I manually checked existing routes to ensure they are not being intercepted by this setup

static/app/views/profiling/continuousProfileProvider.tsx
```diff
@@ -0,0 +1,142 @@
...
+    setProfiles({type: 'loading'});
+
+    fetchContinuousProfileFlamegraph(api, chunkParams, project.id, organization.slug)
+      .then(p => {
+        setProfiles({type: 'resolved', data: p});
+      })
+      .catch(err => {
+        // XXX: our API client mock implementation does not mimick the real
+        // implementation, so we need to check for an empty object here. #sad
+        const isEmptyObject = err.toString() === '[object Object]';
```
Zylphrex, line 122: Is this used in tests somehow? The only test that errors I see doesn't reach this point.

Zylphrex (approved):

static/app/views/profiling/continuousProfileProvider.tsx
JonasBa: It is in the old test. Let me remove it here, we dont need it rn

PR #74019, opened 2024-07-09
chore(integrations) Remove escalating-issues-msteams feature

markstory:
This feature was built more than a year ago and was never finished or enabled for any customers. If we have no near term plans to finish it, I think we should remove the feature to reduce future maintenance effort.


ceorourke (approved):
LGTM assuming we don't intend to release this - I'm not sure who would have context around that.

markstory:
> LGTM assuming we don't intend to release this - I'm not sure who would have context around that.

My assumption is that the context has been lost and these changes have been forgotten about or are no longer a priority.

PR #74023, opened 2024-07-09
feat(uptime): Read robots txt before validating url

davidenwang:
Uses urllib.robot_parser which seems to handle all the logic for fetching/parsing. For the sake of testing, pulled out retrieving the parser into its own function.

### DIFF
--- a/src/sentry/uptime/detectors/tasks.py
+++ b/src/sentry/uptime/detectors/tasks.py
@@ -1,6 +1,8 @@
 import datetime
 import logging
 from datetime import timedelta
+from urllib.parse import urljoin
+from urllib.robotparser import RobotFileParser
 
 from django.utils import timezone
 
@@ -19,6 +21,7 @@ from sentry.utils import metrics
 from sentry.utils.hashlib import md5_text
 from sentry.utils.locking import UnableToAcquireLock
 
+UPTIME_USER_AGENT = "sentry.io_uptime_checker_v_1"
 LAST_PROCESSED_KEY = "uptime_detector_last_processed"
 SCHEDULER_LOCK_KEY = "uptime_detector_scheduler_lock"
 FAILED_URL_RETRY_FREQ = timedelta(days=7)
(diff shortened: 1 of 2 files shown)


tests/sentry/uptime/detectors/test_tasks.py
```diff
@@ -189,14 +190,28 @@ def test_failed_url(self):
...
+        url = "https://sentry.io"
+        test_robot_parser = RobotFileParser()
+        robots_txt = []
+        test_robot_parser.parse(robots_txt)
+        with mock.patch(
+            "sentry.uptime.detectors.tasks.get_robots_txt_parser",
+            return_value=test_robot_parser,
+        ):
+            assert process_candidate_url(self.project, 100, url, 50)
+
```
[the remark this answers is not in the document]
davidenwang: good call, lemme add

src/sentry/uptime/detectors/tasks.py
```diff
@@ -197,8 +206,13 @@ def get_failed_url_key(url: str) -> str:
 
 
 def check_url_robots_txt(url: str) -> bool:
-    # TODO: Implement this check
-    return True
+    return get_robots_txt_parser(url).can_fetch("*", url)
```
wedamija, line 209: We probably need to use a specific user agent here, maybe that should be a constant or something.

Will this attempt to hit an external url during tests?

davidenwang: That's true, would we just want to match the user agent here with whatever the user agent is for when we make the requests on the checker side?

Also, it won't call the external url, I override the `get_robots_txt_parser` function with a mock which returns a RobotFileParser that I manually pass a set of robot.txt lines to parse. Not sure if thats the best way to go about this, but it already feels a bit weird since I'm testing the functionality of the library a little bit here now

wedamija (approved):
lgtm, might be worth testing a `robots.txt` that uses our user agent

PR #74024, opened 2024-07-09
fix(web-vitals): do not truncate browser selector

KevinL10:
Before:
image

After:

image


static/app/views/insights/browser/webVitals/components/browserTypeSelector.tsx
```diff
@@ -16,6 +16,7 @@ import {SpanIndexedField} from 'sentry/views/insights/types';
 const LabelContainer = styled('div')`
   display: flex;
   gap: ${space(1)};
+  min-width: 130px;
```
narsaynorath, line 19: Instead of pinning a width here, we could also use `width: max-content` so additional options don't end up overflowing and have space.

narsaynorath (approved):
Going to approve, but try out the `max-content` value and see if that works too! it'll be more flexible long-term, just trying to avoid special numbers in CSS when possible

PR #74033, opened 2024-07-09
feat(query-builder): Handle all keyboard actions while there is a selection

malwilley:
Prior to this change, only a few keys were implemented like cmd+c, backspace, and the arrow keys. You weren't able to immediately replace a selection without first deleting it.

- Add handling for cut and paste when there is a selection
- Replace the selection with a typed character
- Rename `PASTE_FREE_TEXT` action to `REPLACE_TOKENS_WITH_TEXT` because the logic is now used across multiple actions

https://github.com/getsentry/sentry/assets/10888943/e2c6d969-159b-4d9b-8b28-4844540afd19


roggenkemper (approved):
thanks for adding the todo about needing permissions! was going to mention that but saw it was already there

PR #74034, opened 2024-07-09
ref(breadcrumbs): Update breadcrumbs drawer from feedback

leeandher:
Lots of fixes here:

- Clicking the icon from issue details will show some focus styles for the control. Currently we cant open the menu as the tooltip will be incorrectly positioned as it tries to appear while the drawer is animating onto the page. For now it at least draws attention to the correct part of the drawer.

focus icon

- The default crumbs have a new icon

new default icon

- The replay colors match those chosen for issues.

replay colors and time

- HTTP items will fully expand by default in the drawer

expand http items

- The header is sticky and now contains the nav crumbs in it

custom sticky header

- There is no excess space/line on the breadcrumb list for longer entries anymore

no excess line on breadcrumb list

-  The display time icon changes when toggled.

relative icon

- [Update] More accurate timestamps that respect user preferences (for 24h clocks, and timezone)

timestamps

### DIFF
--- a/static/app/components/events/breadcrumbs/breadcrumbsDataSection.tsx
+++ b/static/app/components/events/breadcrumbs/breadcrumbsDataSection.tsx
@@ -1,4 +1,4 @@
-import {useCallback, useMemo} from 'react';
+import {Fragment, useCallback, useMemo} from 'react';
 import styled from '@emotion/styled';
 
 import ProjectAvatar from 'sentry/components/avatar/projectAvatar';
@@ -33,6 +33,7 @@ import type {Event} from 'sentry/types/event';
 import type {Group} from 'sentry/types/group';
 import type {Project} from 'sentry/types/project';
 import {trackAnalytics} from 'sentry/utils/analytics';
+import {getShortEventId} from 'sentry/utils/events';
 import {useLocalStorageState} from 'sentry/utils/useLocalStorageState';
 import useOrganization from 'sentry/utils/useOrganization';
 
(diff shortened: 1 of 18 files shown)


michellewzhang (commented):
looks good from the replay side!

roggenkemper (approved):
looks good to me! this might be personal preference but i feel like the time button for the breadcrumbs on the details page could benefit from a tooltip or something so the user knows what's happening/what option they currently are on. right now you only know what's going to happen when you press it, since there's no dropdown like there is in the drawer.

static/app/components/globalDrawer/index.tsx
```diff
@@ -28,6 +28,10 @@ export interface DrawerOptions {
    * If true (default), closes the drawer when anywhere else is clicked
    */
   closeOnOutsideClick?: boolean;
+  /**
+   * Custom content for the header of the drawer
+   */
+  headerContent?: React.ReactNode;
```
malwilley, line 34: As we increase the customizability of this component, it might be better to give more control to the consumer rather than add a new prop. Perhaps `children` could render the drawer header and close button as well? That would give the greatest flexibility and simplify the interface a bit.

malwilley (commented):
Noticed this on an event: https://sentry-539of4nfw.sentry.dev/issues/2382970109/events/da370e7cf6674bbd829629d9507ca68b/

Seems confusing, not sure if this is expected behavior?

CleanShot 2024-07-10 at 10 21 11

static/app/components/events/breadcrumbs/breadcrumbsDataSection.tsx
```diff
@@ -70,14 +79,31 @@ export default function BreadcrumbsDataSection({
...
+                {
+                  label: (
+                    <CrumbContainer>
+                      <ProjectAvatar project={project} />
+                      <GroupShortId>{group.shortId}</GroupShortId>
+                    </CrumbContainer>
+                  ),
+                },
+                {
+                  label: <GroupShortId>{event.id.substring(0, 8)}</GroupShortId>,
```
malwilley, line 100: There should be a common util for this: `getShortEventId()`

static/app/components/events/breadcrumbs/breadcrumbsDrawerContent.tsx
```diff
@@ -41,27 +35,25 @@ export const enum BreadcrumbControlOptions {
...
   const theme = useTheme();
+
   const [search, setSearch] = useState('');
   const [filterSet, setFilterSet] = useState(new Set<string>());
   const [sort, setSort] = useLocalStorageState<BreadcrumbSort>(
     BREADCRUMB_SORT_LOCALSTORAGE_KEY,
     BreadcrumbSort.NEWEST
   );
+
+  const [focusControl, setFocusControl] = useState(initialFocusControl);
```
malwilley, line 55: nit: This is a bit disjointed from the useCallback below, it might be nice to combine them into a hook like `useFocusControl()` which return `getFocusProps`

leeandher: That makes sense, I never thought of grouping hooks in this way, thanks!

static/app/components/globalDrawer/index.tsx
leeandher: Good call, I changed it to pass in a `<Header />` component the same way it passes Body. I added a context provider as well so the header button would still fire the onClose handler specified in `openDrawer`. I'll have to update tests as well since they all expect a close button to be rendered.

malwilley: Thanks for making that change! 

And now, more unsolicited feedback! Not something to do in this PR surely, but something we may want to do in the future is stop passing in `Header` and `Body` in the `renderDrawer` function. That's how GlobalModal works, but I don't think it's necessary for this component since we are using context which didn't exist when. GlobalModal was originally created. They could directly import `Drawer.Body` and `Drawer.Header`, right?

leeandher:
> Noticed this on an event: https://sentry-539of4nfw.sentry.dev/issues/2382970109/events/da370e7cf6674bbd829629d9507ca68b/
> 
> Seems confusing, not sure if this is expected behavior?
> 
> CleanShot 2024-07-10 at 10 21 11

Long living bug! fixed it!

static/app/components/globalDrawer/index.tsx
leeandher: that's a very good point, makes sense. I'll defer to a later change, but thanks!

malwilley (approved):

PR #74040, opened 2024-07-09
ref(slack): Formalize SlackApiError message categories

RyanSkonnord:
Refine the way we unpack and parse `SlackApiError` objects. Formalize the set of expected error conditions.


RyanSkonnord:
Regarding https://github.com/getsentry/sentry/pull/74040/commits/d5c7de52f4b51825ebde15eedc4ec11d5a497d73: Based on discussion with @cathteng and @iamrajjoshi, "Expired url" may be an odd one out. It's a sign that it differs in capitalization and whitespace, and sometimes it shows up only after the "The server responded with" message. (See [SlackApiError.\_\_init\_\_](https://github.com/slackapi/python-slack-sdk/blob/main/slack_sdk/errors/__init__.py#L30-L33) in the Slack SDK source.)

I'm not sure if we necessarily need to be strict about scanning the string body for the other two slugs, but I prefer to avoid the possibility of false matches in case, say, the string `"ratelimited"` shows up in some random stack trace for an unrelated reason.

src/sentry/integrations/slack/utils/errors.py
```diff
@@ -0,0 +1,51 @@
...
+            pass
+
+    dump = str(exc)
+    category = _CATEGORIES_BY_MESSAGE.get(dump.split("\n")[0])
+    if category:
+        return category
+
+    for category in SLACK_SDK_ERROR_CATEGORIES:
+        if category.check_body and category.message in dump:
+            return category
```
cathteng, line 47: the "Expired url error" still seems iffy to me. maybe we log here to double check if we can rely on any of the other places instead?

cathteng (approved):

src/sentry/integrations/slack/utils/errors.py
RyanSkonnord: Discussed with @cathteng and @iamrajjoshi; resolved that we'll add metrics in a follow-up PR.

PR #74042, opened 2024-07-09
fix(spans): Validate id fields on spans query builders

Zylphrex:
Not sure when this validation disappeared but let's add it back to the spans query builders first.

### DIFF
--- a/src/sentry/search/events/builder/base.py
+++ b/src/sentry/search/events/builder/base.py
@@ -1212,21 +1212,21 @@ class BaseQueryBuilder:
         name = search_filter.key.name
         value = search_filter.value
 
-        if label := self.uuid_fields.get(name):
+        if name in self.uuid_fields:
             if value.is_wildcard():
-                raise InvalidSearchQuery(WILDCARD_NOT_ALLOWED.format(label))
+                raise InvalidSearchQuery(WILDCARD_NOT_ALLOWED.format(name))
             if not value.is_event_id():
-                raise InvalidSearchQuery(INVALID_ID_DETAILS.format(label))
+                raise InvalidSearchQuery(INVALID_ID_DETAILS.format(name))
 
     def validate_span_id_like_filters(self, search_filter: event_search.SearchFilter):
         name = search_filter.key.name
         value = search_filter.value
 
-        if label := self.span_id_fields.get(name):
+        if name in self.span_id_fields:
             if value.is_wildcard():
-                raise InvalidSearchQuery(WILDCARD_NOT_ALLOWED.format(label))
+                raise InvalidSearchQuery(WILDCARD_NOT_ALLOWED.format(name))
             if not value.is_span_id():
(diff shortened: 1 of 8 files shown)


src/sentry/search/events/builder/base.py
```diff
@@ -77,7 +77,6 @@
 class BaseQueryBuilder:
     requires_organization_condition: bool = False
     organization_column: str = "organization.id"
-    uuid_fields = {"id", "trace", "profile.id", "replay.id"}
```
wmak, line 80: Don't we still need the `uuid_fields` on the base query builder?

src/sentry/search/events/builder/spans_indexed.py
```diff
@@ -46,8 +57,49 @@ def time_column(self) -> SelectType:
...
+        if search_filter.value.is_wildcard():
+            raise InvalidSearchQuery(WILDCARD_NOT_ALLOWED.format(label))
+        if not search_filter.value.is_span_id():
+            raise InvalidSearchQuery(INVALID_SPAN_ID.format(label))
+
+    if label := UUID_FIELDS.get(name):
+        if search_filter.value.is_wildcard():
+            raise InvalidSearchQuery(WILDCARD_NOT_ALLOWED.format(label))
+        if not search_filter.value.is_event_id():
+            raise InvalidSearchQuery(INVALID_ID_DETAILS.format(label))
```
wmak, line 105: A block like this already exists in DiscoverQueryBuilder:127 should we move that into the base, and then override it in the SpanBuilder?

src/sentry/search/events/builder/base.py
Zylphrex: Technically no because the only thing that checks is in the discover query builder but I've moved all the logic back into the base query builder and put this back.

src/sentry/search/events/builder/base.py
```diff
@@ -72,18 +72,20 @@
...
     function_alias_prefix: str | None = None
     spans_metrics_builder = False
     profile_functions_metrics_builder = False
     entity: Entity | None = None
     config_class: type[DatasetConfig] | None = None
     duration_fields: set[str] = set()
+    uuid_fields: Mapping[str, str] = {}
```
wmak, line 87: Nice, i like that you made the value a label for more human readable messages, do you mind commenting that here on the base class too?

wmak (approved):

src/sentry/search/events/builder/base.py
Zylphrex: On second thought, the final error message does read as smoothly like this.

```python
"Wildcard conditions are not permitted on `trace.span` field"
```

vs

```python
"Wildcard conditions are not permitted on Filter Trace ID field"
```

Any thoughts on switching it just showing the column name surrounded in backticks?

wmak: Ah, that's a very good point. I think customers will still be able to figure it out either way, but its probably easier for both sides if we stick with the column name

PR #74044, opened 2024-07-09
Add signature auth to severity score seer calls.

corps:
Adds and test for signature auth handling in the severity score code paths.

### DIFF
--- a/tests/sentry/event_manager/test_severity.py
+++ b/tests/sentry/event_manager/test_severity.py
@@ -6,6 +6,7 @@ from unittest.mock import MagicMock, patch
 
 import orjson
 from django.core.cache import cache
+from django.test import override_settings
 from urllib3 import HTTPResponse
 from urllib3.exceptions import ConnectTimeoutError, MaxRetryError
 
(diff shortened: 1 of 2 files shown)


tests/sentry/event_manager/test_severity.py
```diff
@@ -75,6 +77,21 @@ def test_error_event_simple(self, mock_urlopen: MagicMock) -> None:
...
+        ):
+            _get_severity_score(event)
+            mock_urlopen.assert_called_with(
+                "POST",
+                "/v0/issues/severity-score",
+                body=orjson.dumps(payload),
+                headers={
+                    "content-type": "application/json;charset=utf-8",
+                    "Authorization": "Rpcsignature rpc0:8d982376e4e49ffe845ed39853f6f2cb9bf38564d2a8a325dcd88abba8c58564",
+                },
```
vartec, line 91: QQ: what is that signature format?

vartec (approved):

tests/sentry/event_manager/test_severity.py
corps: An arbitrary format that RPC has settled on for now: https://github.com/search?q=repo%3Agetsentry%2Fsentry%20Rpcsignature&type=code

I suspect in the future the mechanism and protocol of RPC will change to a standard and not an inhouse implementation, which will simplify things, but for the time being this will work.

🙏 thanks for the review

PR #74054, opened 2024-07-09
feat(query-builder): Allow pasting without requiring user permissions

malwilley:
Currently when there is a selection, the focus is on the top-level grid div (because there is nowhere else for the focus to go). We have been listening for keydown events here, which works for everything _except_ pasting, because divs do not have the `onPaste` event. Currently we are accessing the clipboard with `navigator.clipboard.readText()`, but that requires the user to grant permission which is not a good experience.

To solve this, we now render an invisible input within the grid (`<SelectionKeyHandler />`), which receives focus when there is a selection. Since this input needs to handle the paste event, I also moved all the selection-specific keydown logic there as well.

### DIFF
--- a/static/app/components/searchQueryBuilder/selectionKeyHandler.tsx
+++ b/static/app/components/searchQueryBuilder/selectionKeyHandler.tsx
@@ -1,0 +1,182 @@
+import {type ForwardedRef, forwardRef, useCallback} from 'react';
+import {VisuallyHidden} from '@react-aria/visually-hidden';
+import type {ListState} from '@react-stately/list';
+import type {Key} from '@react-types/shared';
+
+import {useSearchQueryBuilder} from 'sentry/components/searchQueryBuilder/context';
+import {type ParseResultToken, Token} from 'sentry/components/searchSyntax/parser';
+import {defined} from 'sentry/utils';
+import {isCtrlKeyPressed} from 'sentry/utils/isCtrlKeyPressed';
+
+type SelectionKeyHandlerProps = {
+  state: ListState<ParseResultToken>;
+  undo: () => void;
+};
+
+function findNearestFreeTextKey(
+  state: ListState<ParseResultToken>,
+  startKey: Key | null,
+  direction: 'right' | 'left'
+): Key | null {
+  let key: Key | null = startKey;
+  while (key) {
+    const item = state.collection.getItem(key);
+    if (!item) {
+      break;
+    }
+    if (item.value?.type === Token.FREE_TEXT) {
+      return key;
+    }
+    key = (direction === 'right' ? item.nextKey : item.prevKey) ?? null;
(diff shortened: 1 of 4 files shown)


static/app/components/searchQueryBuilder/tokenizedQueryGrid.tsx
```diff
@@ -40,14 +43,35 @@ function useApplyFocusOverride(state: ListState<ParseResultToken>) {
 
 function Grid(props: GridProps) {
   const ref = useRef<HTMLDivElement>(null);
+  const selectionKeyHandlerRef = useRef<HTMLInputElement>(null);
   const {size} = useSearchQueryBuilder();
-  const {state, gridProps} = useQueryBuilderGrid(props, ref);
-
+  const state = useListState<ParseResultToken>({
```
malwilley, line 48: Had to move some things around to pass the `undo` function into the new component.

static/app/components/searchQueryBuilder/useQueryBuilderGrid.tsx
```diff
@@ -17,59 +13,26 @@ interface UseQueryBuilderGridProps extends AriaGridListOptions<ParseResultToken>
 
 const noop = () => {};
 
-function findNearestFreeTextKey(
```
malwilley, line 20: Most of the changes in this file are moving the onKeyDown logic from there to the new component

static/app/components/searchQueryBuilder/selectionKeyHandler.tsx
```diff
@@ -0,0 +1,182 @@
...
+import {type ParseResultToken, Token} from 'sentry/components/searchSyntax/parser';
+import {defined} from 'sentry/utils';
+import {isCtrlKeyPressed} from 'sentry/utils/isCtrlKeyPressed';
+
+type SelectionKeyHandlerProps = {
+  state: ListState<ParseResultToken>;
+  undo: () => void;
+};
+
+function findNearestFreeTextKey(
```
malwilley, line 16: This function is unchanged from the file it was moved from

static/app/components/searchQueryBuilder/selectionKeyHandler.tsx
```diff
@@ -0,0 +1,182 @@
...
+        dispatch({
+          type: 'REPLACE_TOKENS_WITH_TEXT',
+          tokens: selectedTokens,
+          text,
+        });
+      },
+      [dispatch, selectedTokens]
+    );
+
+    const onKeyDown = useCallback(
```
malwilley, line 74: All this logic is copied over from `useQueryBuilderGrid`, the only new part of this component is the rendered input and the `onPaste` handler

roggenkemper (approved):
lgtm!

PR #74057, opened 2024-07-10
fix(metrics): Default metric selection

ArthurKnaus:
Pre-fill the correct condition and aggregation on setting the default selected metric.

---

Changes after revert of [original PR](https://github.com/getsentry/sentry/pull/74011):
- Extract and test logic for creating a new widget.
- Add logging for invalid data.
- Filter unresolvable virtual metrics from the `metrics/query` request.

### DIFF
--- a/static/app/views/metrics/utils/getNewMetricsWidget.tsx
+++ b/static/app/views/metrics/utils/getNewMetricsWidget.tsx
@@ -1,0 +1,37 @@
+import * as Sentry from '@sentry/react';
+
+import type {MetricMeta} from 'sentry/types/metrics';
+import {getDefaultAggregation, isVirtualMetric} from 'sentry/utils/metrics';
+import {emptyMetricsQueryWidget} from 'sentry/utils/metrics/constants';
+
+/**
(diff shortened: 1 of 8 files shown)


static/app/views/metrics/utils/getNewMetricsWidget.tsx
```diff
@@ -0,0 +1,37 @@
...
+    if (!defaultCondition) {
+      Sentry.captureMessage(
+        'Metrics: Trying to create widget from virtual MRI without condition'
+      );
+      // Invalid data -> falling back to default
+      return emptyMetricsQueryWidget;
+    }
+
+    condition = defaultCondition;
+    aggregation = metricsMeta.operations[0];
```
obostjancic, line 28: do we need to filter these operations with `isAllowedAggregate` somewhere?

obostjancic (approved):

static/app/views/metrics/utils/getNewMetricsWidget.tsx
ArthurKnaus: Added logging for that case 👍

PR #74060, opened 2024-07-10
Remove `ProcessingIssue` models, step 2

Swatinem:
This removes all the rest of the `RawEvent`, `ProcessingIssue`, `EventProcessingIssue` and `ReprocessingReport` models, and removes those from the migration state.

---

This is the second step after https://github.com/getsentry/sentry/pull/73939. The migration script initially auto-generated some removal of some `UniqueTogether` constraints, plus removal of the `project` columns. Though I assume that removing the DB-level constraints from those as was done in step 1 is sufficient here, and after removal from migration state, the tables will rather have some dangling `project` references instead.


wedamija (approved):
lgtm, thanks for cleaning this up

PR #74062, opened 2024-07-10
Remove `ProcessingIssue` models, step 3

Swatinem:
This is the final step, and fully drops the `RawEvent`, `ProcessingIssue`, `EventProcessingIssue` and `ReprocessingReport` tables.

---

This is the 3rd step after https://github.com/getsentry/sentry/pull/73939 and https://github.com/getsentry/sentry/pull/74060.

In particular the `RawEvent` and `EventProcessingIssue` still have some 50-60 mil rows, though I assume the reason for that is that the `clear_expired_raw_events` task was not properly doing its job.
You can see here that the task was constantly running at its time limit and presumably timing out: https://sentry.sentry.io/performance/summary/?project=1&query=&referrer=performance-transaction-summary&statsPeriod=7d&transaction=sentry.tasks.clear_expired_raw_events&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29

Either way, the cleanup task should have cleaned up rows older than 10/13 days, and we have stopped creating any new rows since longer than that.


wedamija (approved):

Swatinem:
Thanks for the reminder: https://github.com/getsentry/etl/pull/1518

markstory (approved):

PR #74070, opened 2024-07-10
ref: Prototype of messaging integration interface

RyanSkonnord:
Sentry devs, see ["Task-list for Messaging Vertical" on Notion](https://www.notion.so/sentry/Task-list-for-Messaging-Vertical-ca4f431d43b84e72b8a91643fe880b66) for context.

The big idea here is to give, via an in-code abstraction, a mile-high view of all the code that is necessary to make a messaging integration work. The `MessagingIntegrationSpec` class is a sketch of one way we could do that.

The existing `IntegrationProvider` class is a close analog to what we are doing here. The distinction is that `IntegrationProvider` is general to all integrations, where `MessagingIntegrationSpec` seeks to cover the additional features and business logic are particular to *messaging* integrations.

The potential benefits include:

1. Serving as internal documentaton. Developers new to our integration code can look at the `MessagingIntegrationSpec` (and `IntegrationProvider`) as a sort of "table of contents" for all the code in a messaging integration. (And the process of creating it will also force the Ecosystem team to wrap our own heads around it in the process.)
2. Providing a basis for improving code consistency and reducing technical debt. See the linked Notion doc above for details on these goals. At worst, `MessagingIntegrationSpec` will provide a common structure and vocabulary for organizing the work. At best, it will provide us a place to extract duplicated code and common patterns, and to apply a common set of automated tests. The best way to ensure that logging, metrics, error handling, etc., are all applied consistently is to do it in a common base class.
3. Providing a "fill in the blanks" template for adding new messaging services. "Implement this interface" is the simplest way to hand off the task (especially if it comes with a built-in test cases). In a perfect outcome, as much business logic as possible will live in the base class, and a new implementation would only have to provide the minimum amount of code that is unique to the new third-party service.

For the purpose of spiking on this task, I focused on the MsTeams and Slack integrations. My goal was to extract some broad chunks from MsTeams (the smaller of the two) that it had in common with Slack, and to implement the spec for both. There is nothing too interesting going on in `MsTeamsMessagingSpec` and `SlackMessagingSpec`; these are intended as proofs of concept and to give an idea of what the final product might look like.

One thing to notice about `MsTeamsMessagingSpec` is that it mostly just points to other classes. This is what I mean when I describe it as a "table of contents". Those classes are where the actual implementations live, and they themselves branch out with common patterns. As mentioned above, `IntegrationProvider` does a lot of heavy lifting already; `IntegrationEventAction` provides a `*NotifyServiceForm` class; and so on. Some of those classes are already rich interfaces that carry a lot of information about business logic, and hopefully we will create new ones (or tweak and extend the existing ones) as we discover the need to do so.

Assorted findings and notes from my spike follow. An investigation into tech debt in the MS Teams integration was a side quest.

## Notes on MessagingIntegrationSpec

Known things left to cover:

* ApiClients. Each integration has its own `sentry.integrations.client.ApiClient` instance but AFAICT it's just a utility internal to business logic, and we're getting rid of `SlackClient` in favor of `SlackSdkClient` anyway. So I'm assuming for now that we don't need to expose classes like `MsTeamsClient` directly from `MessagingIntegrationSpec`.
* Webhooks. I haven't really dove into these yet and it would probably save time to get a tour from someone with existing familiarity. A quick skim reveals very little commonality between Slack and MS Teams: Slack has its own `webhooks` package with lots of endpoints, whereas MS Teams has only `MsTeamsWebhookEndpoint` with four event handlers. Is this just historical accident? How much overlap is there in what the endpoints do?
* Middleware. See https://github.com/getsentry/sentry/pull/73998. We could fold it into `MessagingIntegrationSpec` pretty easily if we wanted to. Do we want/need to generalize this to all messaging integrations? (Why do we have ones for Slack and Discord but not MS Teams?)

Also, for the record, I'm assuming everything in the `sentry.integrations.msteams.card_builder` package is presentation details entirely unique to MS Teams and we have no expectation of generalizing it.

## Notes on MS Teams tech debt investigation

Mostly, not a lot of tech debt is jumping off the page to me? At least, not of the kind that was described in the Notion doc, as we encountered with Slack. (Things like: magic strings; values representable as enum classes; brittle exception handling.)

There may or may not be significant room for improvement in logging, metrics, and unit test coverage. I haven't been drilling down on that topic so far, but I can start if we think it's valuable. However, I'd much prefer to do that in common base code, as described above.


RyanSkonnord:
Rescoped to https://github.com/getsentry/sentry/pull/74743

PR #74072, opened 2024-07-10
fix(related_issues): Do not use culprit as it's deprecated

armenzg:
In #73861 I started using `culprit` to generate the subtitle and message values similarly to what [events.tsx](https://github.com/getsentry/sentry/blob/a2d7e9d2707f0e2e061600dc0daac9cdee06e858/static/app/utils/events.tsx) does.

Unfortunately, `culprit` is not a field we should be using since it is a deprecated value.

See https://github.com/getsentry/relay/pull/1082 for details.


static/app/views/issueDetails/traceTimeline/useTraceTimelineEvents.tsx
```diff
@@ -164,7 +160,9 @@ export function useTraceTimelineEvents({event}: UseTraceTimelineEventsOptions):
         'project.name': '',
         timestamp: event.dateCreated!,
         title: event.title,
+        // The event variable comes from a call to the issues endpoint
```
armenzg, line 163: The issues endpoint does not return the `transaction` field.

static/app/views/issueDetails/traceTimeline/useTraceTimelineEvents.tsx
```diff
@@ -164,7 +160,9 @@ export function useTraceTimelineEvents({event}: UseTraceTimelineEventsOptions):
...
         transaction: '',
+        // XXX: Typing is not working here; this should event.type
```
armenzg, line 165: This is a very similar situation as to what I noticed in #74010

I will not be fixing it in this PR.

static/app/views/issueDetails/traceTimelineOrRelatedIssue.spec.tsx
```diff
@@ -73,15 +71,14 @@ describe('TraceTimeline & TraceRelated Issue', () => {
...
-    culprit: 'billiard.pool in foo', // Used for subtitle
     'error.value': ['some-other-error-value', 'The last error value'],
     timestamp: '2024-01-24T09:09:04+00:00',
     'issue.id': 9999,
     project: project.slug,
     'project.name': project.name,
     title: 'someTitle',
     id: '12345',
-    transaction: 'foo',
+    transaction: '',
```
armenzg, line 81: The original reason to grab the `culprit` was this case where the event does not have `transaction` set to something.

This is [the issue](https://sentry.sentry.io/issues/2733785806/events/d72ca2999f1743aaa4e4c9cc21f3f4c4/?project=1&referrer=issue_details.related_trace_issue).

Not using `culprit` makes us lose `billiard.pool in mark_as_worker_lost` as the subtitle but I'm willing to do this if we're going to get rid of `culprit` in events.tsx.
image

The info is not in the events JSON:
image
or the Discover data:
image

Ideally, we would have made a 2nd call to the issues endpoint for `metadata` but here we are.

static/app/views/issueDetails/traceTimelineOrRelatedIssue.spec.tsx
```diff
@@ -340,29 +337,28 @@ describe('TraceTimeline & TraceRelated Issue', () => {
 });
 
 function createEvent({
-  culprit,
+  transaction,
   title,
-  error_value,
+  error_value = [''], // We always get a non-empty array
```
armenzg, line 342: Passing by improvement.

static/app/views/issueDetails/traceTimelineOrRelatedIssue.spec.tsx
```diff
@@ -409,14 +405,14 @@ describe('getTitleSubtitleMessage()', () => {
...
-          culprit: 'billiard.pool in foo',
+          transaction: '',
           title: 'WorkerLostError: ',
           error_value: ['some-other-error-value', 'The last error value'],
         })
       )
     ).toEqual({
       title: 'WorkerLostError:', // The colon is kept
-      subtitle: 'billiard.pool in foo',
+      subtitle: '', // XXX: Fix events.tsx to use transaction instead of culprit
```
armenzg, line 415: This is the only regression in this PR.

armenzg:
I've removed the review request since:
The [events.tsx](https://github.com/getsentry/sentry/blob/master/static/app/utils/events.tsx) file is using `culprit` from the `Group` model:
https://github.com/getsentry/sentry/blob/ac83b15608a785ac60ae46231da83c8dee5bfe00/src/sentry/models/group.py#L532-L534

In my code, I am using `culprit` from an event which is a [generated value](https://github.com/getsentry/sentry/blob/23f2b42028b197415b1fc55d1f233c1e5414c89f/src/sentry/culprit.py#L15-L43).

Ideally, we would not use the same field name to prevent confusion.

PR #74076, opened 2024-07-10
feat(profiling) construct profile from continuous chunk

JonasBa:
Construct a profile from a continuous chunk.

### DIFF
--- a/static/app/types/profiling.d.ts
+++ b/static/app/types/profiling.d.ts
@@ -37,6 +37,7 @@ declare namespace Profiling {
     stack_id: number;
     thread_id: string;
     timestamp: number;
+    queue_address?: string;
   };
 
   type SentrySampledProfileFrame = {
@@ -104,7 +105,7 @@ declare namespace Profiling {
     measurements?: Measurements;
   };
 
-  interface SentryContinousProfile {
+  interface SentryContinousProfileChunk {
     chunk_id: string;
     environment: string;
     project_id: number;
@@ -123,6 +124,7 @@ declare namespace Profiling {
       frames: SentrySampledProfileFrame[];
       stacks: SentrySampledProfileStack[];
       thread_metadata?: Record<string, {name?: string; priority?: number}>;
+      queue_metadata?: Record<string, {label: string}>;
     };
   }
 
--- a/static/app/types/profiling.d.ts
+++ b/static/app/types/profiling.d.ts
@@ -37,7 +37,6 @@ declare namespace Profiling {
     stack_id: number;
     thread_id: string;
     timestamp: number;
-    queue_address?: string;
   };
 
   type SentrySampledProfileFrame = {
@@ -124,7 +123,6 @@ declare namespace Profiling {
       frames: SentrySampledProfileFrame[];
       stacks: SentrySampledProfileStack[];
       thread_metadata?: Record<string, {name?: string; priority?: number}>;
-      queue_metadata?: Record<string, {label: string}>;
     };
   }
 
--- a/static/app/utils/profiling/profile/continuousProfile.tsx
+++ b/static/app/utils/profiling/profile/continuousProfile.tsx
@@ -157,30 +157,17 @@ export class ContinuousProfile extends Profile {
   }
 }
 
-const COCOA_MAIN_THREAD = 'com.apple.main-thread';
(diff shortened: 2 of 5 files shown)


static/app/types/profiling.d.ts
```diff
@@ -37,6 +37,7 @@ declare namespace Profiling {
     stack_id: number;
     thread_id: string;
     timestamp: number;
+    queue_address?: string;
```
Zylphrex, line 40: Continuous profiling doesnt use queue addresses

static/app/types/profiling.d.ts
```diff
@@ -123,6 +124,7 @@ declare namespace Profiling {
       frames: SentrySampledProfileFrame[];
       stacks: SentrySampledProfileStack[];
       thread_metadata?: Record<string, {name?: string; priority?: number}>;
+      queue_metadata?: Record<string, {label: string}>;
```
Zylphrex, line 127: Continuous profiling doesnt use queue addresses

static/app/utils/profiling/profile/profile.tsx
```diff
@@ -58,7 +61,7 @@ export class Profile {
     name: string;
     startedAt: number;
     threadId: number;
-    type: string;
+    type: 'flamechart' | 'flamegraph' | '';
```
Zylphrex, line 64: What is `''` supposed to represent?

static/app/utils/profiling/profile/continuousProfile.tsx
```diff
@@ -0,0 +1,187 @@
...
+      let size = 0;
+
+      for (let j = 0; j < stack.length; j++) {
+        frame = resolveFrame(stack[j]);
+        if (frame) resolvedStack[size++] = frame;
+      }
+
+      profile.appendSample(
+        resolvedStack,
+        (sample.timestamp - previousSampleTimestamp) * 1e3,
```
Zylphrex, line 59: Don't we normally take the timestamp from the current sample to the _next_ sample for the duration? Here, it's taking the timestamp from the current sample to the _previous_ sample instead.

static/app/utils/profiling/profile/continuousProfile.tsx
```diff
@@ -0,0 +1,187 @@
...
+  // is this the only platform string to check for?
+  if (profile.platform === 'cocoa') {
+    const queueName =
+      sample.queue_address && queue_metadata?.[sample.queue_address]?.label;
+
+    // if a queue has the main thread name, we discard it
+    if (queueName && queueName !== COCOA_MAIN_THREAD) {
+      return {threadId, threadName: queueName};
+    }
+  }
```
Zylphrex, line 184: Mentioned this earlier but I don't think we kept the queue names for continuous profiling. It might be mixed together with the thread data so maybe let's double-check with @armcknight

static/app/utils/profiling/profile/continuousProfile.tsx
JonasBa: Ah yes, I always get confused by the order

static/app/utils/profiling/profile/profile.tsx
JonasBa: Renaming this to 'empty'. We use the static empty instance and I didnt want to pick either as none really applies

static/app/types/profiling.d.ts
JonasBa: Removed all of this

static/app/utils/profiling/profile/continuousProfile.tsx
```diff
@@ -0,0 +1,174 @@
...
+      return resolvedFrame;
+    }
+
+    let frame: Frame | null = null;
+    const resolvedStack: Frame[] = new Array(256); // stack size limit
+
+    for (let i = 0; i < chunk.profile.samples.length; i++) {
+      const sample = chunk.profile.samples[i];
+      const nextSampleTimestamp =
+        chunk.profile.samples[i - 1]?.timestamp ?? sample.timestamp;
```
Zylphrex, line 47: Should this be `i + 1` for the next sample?

JonasBa: Yes, I renamed the variable without fixing the lookup 😶‍🌫️

Zylphrex (approved):

PR #74077, opened 2024-07-10
ref(discover): Simplify split query logic for dashboards

narsaynorath:
We can just query errors and transactions directly instead of appending the event.type:X filter to discover. I added a number of tests to validate the split decision routing.

During that testing, I realized I had a case that was erroring out and it just seems like we were trying to access the meta results inside the data, but usually they're keys at the same level. I removed the extra `.get("data")` in the first block, but I'm not sure what the second block is doing and whether or not the `.get("data")` is a bug


narsaynorath:
Converting to draft to add more test coverage for non-on-demand cases.

shruthilayaj (approved):

PR #74078, opened 2024-07-10
ref(similarity): Change location of deleted check in backfill

jangjodi:
The status index was sometimes being used in this query, causing it to take 30s +, however groups are very infrequently in this deletion status
Remove the filtering by status, and add the check to exclude it out of the postgres query

### DIFF
--- a/src/sentry/tasks/embeddings_grouping/utils.py
+++ b/src/sentry/tasks/embeddings_grouping/utils.py
@@ -17,7 +17,7 @@ from sentry.eventstore.models import Event
 from sentry.grouping.grouping_info import get_grouping_info
 from sentry.issues.grouptype import ErrorGroupType
 from sentry.issues.occurrence_consumer import EventLookupError
-from sentry.models.group import Group, GroupStatus
+from sentry.models.group import Group
 from sentry.models.project import Project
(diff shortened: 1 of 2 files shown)


JoshFerge (commented):
can we check to make sure its not in deleted status once we have the group pulled down from postgres?

src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -168,17 +167,33 @@ def get_current_batch_groups_from_postgres(project, last_processed_group_index,
             total_groups_to_backfill_length,
         )
 
+    groups_to_backfill_not_deleted_status = [
+        (group_id, data)
+        for (group_id, data, status) in groups_to_backfill_batch
+        if status not in [GroupStatus.PENDING_DELETION, GroupStatus.DELETION_IN_PROGRESS]
+    ]
```
jangjodi, line 174: this check could be combined with the following embedding check, but I kept it separate for logging purposes. let me know if you think we should combine them

JoshFerge (approved):

JoshFerge:
can wait to merge this if we determine it helps, still investigating the query plan

PR #74081, opened 2024-07-10
feat(query-builder): Focus filter when pressing backspace and value input is empty

malwilley:
When editing a tag value, pressing backspace with an empty input used to do nothing. This modifies the behavior to focus the whole token, allowing easier deletion if you change your mind while creating a filter.

https://github.com/getsentry/sentry/assets/10888943/4d31a032-eeed-4b83-ad91-307e25f19655

### DIFF
--- a/static/app/components/searchQueryBuilder/filter.tsx
+++ b/static/app/components/searchQueryBuilder/filter.tsx
@@ -23,11 +23,15 @@ import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
 import {defined} from 'sentry/utils';
 
-type SearchQueryTokenProps = {
+interface SearchQueryTokenProps {
   item: Node<ParseResultToken>;
   state: ListState<ParseResultToken>;
   token: TokenResult<Token.FILTER>;
-};
+}
+
+interface FilterValueProps extends SearchQueryTokenProps {
+  filterRef: React.RefObject<HTMLDivElement>;
+}
 
 function FilterValueText({token}: {token: TokenResult<Token.FILTER>}) {
   const {size} = useSearchQueryBuilder();
@@ -77,7 +81,7 @@ function FilterValueText({token}: {token: TokenResult<Token.FILTER>}) {
   }
 }
 
-function FilterValue({token, state, item}: SearchQueryTokenProps) {
+function FilterValue({token, state, item, filterRef}: FilterValueProps) {
   const ref = useRef<HTMLDivElement>(null);
   const {dispatch, focusOverride} = useSearchQueryBuilder();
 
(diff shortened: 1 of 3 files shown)


static/app/components/searchQueryBuilder/filter.tsx
```diff
@@ -23,11 +23,15 @@ import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
 import {defined} from 'sentry/utils';
 
-type SearchQueryTokenProps = {
+interface SearchQueryTokenProps {
```
leeandher, line 26: ❤️

static/app/components/searchQueryBuilder/valueCombobox.tsx
```diff
@@ -776,16 +778,11 @@ export function SearchQueryBuilderValueCombobox({
       }
 
       // If at the start of the input and backspace is pressed, delete the last selected value
```
leeandher, line 780: I don't think this comment is accurate anymore

static/app/components/searchQueryBuilder/valueCombobox.tsx
```diff
@@ -776,16 +778,11 @@ export function SearchQueryBuilderValueCombobox({
...
-      if (
-        e.key === 'Backspace' &&
-        e.currentTarget.selectionStart === 0 &&
-        e.currentTarget.selectionEnd === 0 &&
-        canSelectMultipleValues
-      ) {
-        dispatch({type: 'DELETE_LAST_MULTI_SELECT_FILTER_VALUE', token});
```
leeandher, line 785: I'm not the most familiar with the new search query code, but Is it expected that this action isn't being dispatched anymore? I see onDelete was modified but this call isn't in it's new definition.

static/app/components/searchQueryBuilder/valueCombobox.tsx
```diff
@@ -440,17 +441,17 @@ function cleanFilterValue(key: string, value: string): string {
...
     case FieldValueType.INTEGER:
       if (FILTER_VALUE_INT.test(value)) {
         return value;
       }
-      return '';
+      return null;
     case FieldValueType.DATE:
       const parsed = parseFilterValueDate(value);
 
       if (!parsed) {
```
leeandher, line 453: super nit: parsed is only used once so we can put it in the conditional

leeandher: also  🙌  to swapping `''` for `null` to be more explicit about unmatched values.

leeandher (approved):
One question, rest looks good though! so many quality of life features 💪

static/app/components/searchQueryBuilder/valueCombobox.tsx
malwilley: You are correct!

PR #74084, opened 2024-07-10
feat(uptime): Frontend Project Setting for Enabling/Disabling Uptime Checks

davidenwang:
image


davidenwang:
Following discussions from the uptime team, we no longer will support a project level setting. It will still exist on the backend, but will be automatically toggled on/off from alternative user action such as deleting the auto-instrumented uptime monitor

PR #74087, opened 2024-07-10
feat(profiling) import continuous profiles

JonasBa:


static/app/utils/profiling/guards/profile.tsx
```diff
@@ -30,3 +30,21 @@ export function isSentrySampledProfile(
...
+export function isSentryContinuousProfile(
+  profile: any
+): profile is Profiling.ContinuousProfile {
+  return (
+    'samples' in profile &&
+    'stacks' in profile &&
+    'frames' in profile &&
+    !('type' in profile) &&
+    !Array.isArray(profile.resources)
+  );
```
Zylphrex, line 43: Won't this also return true for the sentry sampled profile?

static/app/utils/profiling/profile/importProfile.tsx
```diff
@@ -219,11 +231,52 @@ export function importSchema(
...
+    | Profiling.EventedProfile
+    | Profiling.SampledProfile
+    | JSSelfProfiling.Trace,
+  frameIndex:
+    | ReturnType<typeof createFrameIndex>
+    | ReturnType<typeof createContinuousProfileFrameIndex>
+    | ReturnType<typeof createSentrySampleProfileFrameIndex>,
   {span, type, frameFilter, profileIds}: ImportOptions
 ): Profile {
+  if (isSentryContinuousProfile(profile)) {
```
Zylphrex, line 269: This is where things might break because the `Profiling.SampledProfile` I think also evaluates to true here.

static/app/utils/profiling/guards/profile.tsx
JonasBa: I think it wont, but this is brittle code as the formats are not very consistent (future changes might break this). Let me explicitly force it down this code path

static/app/utils/profiling/profile/importProfile.tsx
JonasBa: Yup, agree. Lets force it down this path explicitly, this is too brittle otherwise

Zylphrex (approved):

Issue #70042, opened 2024-05-01
Replace our usage of `react-inspector` in Replay with component used in Issue Details


billyvg:
When we first worked on Replay, we did not have a component to do pretty rendering of objects for our Console and Breadcrumbs. However, it looks like we now have one available that is currently in use in our Issue Breadcrumbs:

![Image](https://github.com/getsentry/sentry/assets/79684/73c07d24-3810-4bc8-a33c-9ad0d24ae24d)

Let's look into using this so that we have one less dependency!

ryan953:
Replay currently uses `<ObjectInspector>`

Issues has https://sentry.sentry.io/stories/?name=app/components/structuredEventData/index.stories.tsx

You can see them both side-by-side (with the same data even!) here https://sentry.sentry.io/stories/?name=app/utils/useOrganization.stories.tsx

michellewzhang:
note: i finished one half of this ticket. other half requires some non-trivial refactoring work with remembering expansion state, which the current `StructuredEventData` component does not have, so i'll leave it open for the future, for anyone who wants to tackle it.

DONE:
- [x] replace component in network panel

TODO:
- [ ] replace component in breadcrumbs panel
- [ ] replace component in console panel

https://github.com/getsentry/sentry/pull/74238 is a half-working PR that might give some inspiration about how to implement the expansion state functionality -- basically we need to do something similar to `ObjectInspector` which passes in `onExpand` and `expandPaths`, which give the component information about how to re-calculate row height when the component is expanded/collapsed, and info that helps remember the expansion state when re-rendered.

michellewzhang:
@c298lee are the instances fully replaced within the breadcrumbs tab? :D

c298lee:
@michellewzhang yes! just the console panel left

PR #74088, opened 2024-07-10
ref(ui): add copy button to `StructuredEventData`

michellewzhang:
**motivation:** 
we want to replace our replay `ObjectInspector` usages in replay details with the issues `StructuredEventData` component. in replay, our component makes use of a copy to clipboard button, so i added this same functionality onto the `StructuredEventData` component as an optional prop (with option to specify `onCopy` callback).

behavior (using `StructuredEventData` here):

https://github.com/getsentry/sentry/assets/56095982/4cafda13-c763-425d-9f36-3429b5b6299f

### DIFF
--- a/static/app/components/structuredEventData/index.stories.tsx
+++ b/static/app/components/structuredEventData/index.stories.tsx
@@ -1,5 +1,6 @@
 import {Fragment} from 'react';
 
+import {addSuccessMessage} from 'sentry/actionCreators/indicator';
 import JSXNode from 'sentry/components/stories/jsxNode';
 import JSXProperty from 'sentry/components/stories/jsxProperty';
 import StructuredEventData from 'sentry/components/structuredEventData';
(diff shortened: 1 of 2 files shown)


c298lee (approved):

static/app/components/structuredEventData/index.tsx
```diff
@@ -380,3 +398,10 @@ const ValueNumber = styled('span')`
 const ValueObjectKey = styled('span')`
   color: var(--prism-keyword);
 `;
+
+const StyledCopyButton = styled(CopyToClipboardButton)`
+  position: absolute;
+  right: ${space(1.5)};
+  z-index: 2;
```
malwilley, line 405: You should be able to avoid using `z-index` if you render this after the `<StucturedData />`

static/app/components/structuredEventData/index.tsx
```diff
@@ -380,3 +398,10 @@ const ValueNumber = styled('span')`
...
+  margin: -${space(0.5)} ${space(1)};
```
malwilley, line 406: If you're absolutely positioning this you shouldn't need a margin, should be able to do it with `top` and `right`

static/app/components/structuredEventData/index.stories.tsx
```diff
@@ -58,4 +59,23 @@ export default storyBook(StructuredEventData, story => {
       </Fragment>
     );
   });
+
+  story('Allow copy to clipboard', () => {
```
malwilley, line 63: Thanks for adding to the story!

static/app/components/structuredEventData/index.tsx
```diff
@@ -324,10 +329,12 @@ export default function StructuredEventData({
   data = null,
   withAnnotatedText = false,
   forceDefaultExpand,
+  showCopyButton,
+  onCopy,
   ...props
 }: StructuredEventDataProps) {
   return (
-    <pre {...props}>
+    <pre {...props} style={{position: 'relative'}}>
```
malwilley, line 337: One last thing - can we add these styles with CSS in a styled component instead?

malwilley (approved):
Looks good!

static/app/components/structuredEventData/index.tsx
michellewzhang: yup!

PR #74090, opened 2024-07-10
feat(feedback): rm unused post-process-group flag from test

aliu39:
No references anywhere else in sentry and getsentry


michellewzhang (approved):

JoshFerge (commented):
double checking this isn't a flag used for the issue platform?

aliu39:
Oops, looks like it is! Test passes without this flag though. Doesn't seem like we have post process tests

aliu39:
@JoshFerge ok if I remove this reference? Or was there a reason we need it for this test?

JoshFerge (approved):

PR #74097, opened 2024-07-10
ref: Move useMouseTracking into utils/*

ryan953:
This isn't replay specific, and indeed it's imported by perf as well as replay. 

Lets put it into the top level utils folder so others can find it. If there turn out to be more mouse-related, or similar things we should group them together into a happy utils/foo/* folder, but for now i can't see that group, so good enough.


static/app/utils/url/useSetQueryFieldInLocation.tsx
```diff
@@ -0,0 +1,19 @@
+import {useCallback} from 'react';
```
michellewzhang, line 1: intentional file addition?

ryan953: oops, that's from https://github.com/getsentry/sentry/pull/74096

michellewzhang (approved):
gif

PR #74098, opened 2024-07-10
feat(feedback-form): Add a global form instance and hook for opening feedback forms

malwilley:
Followup to https://github.com/getsentry/sentry/pull/73990#discussion_r1670897970

The current way we open feedback forms is by using `useFeedbackWidget()` which requires a button ref to attach listeners and removes the form component when the connected button unmounts (this is necessary because we need to clean up the created form elements on the page somehow).

This PR adds a top-level context and form context that can be used throughout the app to open forms. This avoids the need to clean up for every usage and allows the form to remain open even if the trigger no longer exists. The usage looks like so:

```tsx
const openForm = useFeedbackForm();
 
if (!openForm) {
  return null;
}

return <Button onClick={() => openForm({formTitle: 'foo'})}>{t('Give Feedback')}</button>;
```


static/app/components/feedback/widget/useFeedbackForm.tsx
```diff
@@ -10,20 +10,22 @@ export function useFeedbackForm(props: UseFeedbackOptions) {
   const formRef = useRef<FeedbackDialog | null>(null);
   const {feedback, options} = useFeedback(props);
 
-  const openForm = useCallback(async () => {
-    if (!feedback) {
-      return;
-    }
+  const openForm = useCallback(
```
malwilley, line 13: Mostly whitespace changes. The big difference is that it now accepts option overrides in `openForm()` and removes the old form when another one is opened.

static/app/components/globalFeedbackForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+ *
+ *  ```tsx
+ * const {feedback, openForm} = useFeedbackForm();
+ *
+ * if (!feedback) {
+ *   return null;
+ * }
+ *
+ * return <Button onClick={() => openForm()}>{t('Give Feedback')}</button>;
+ * ```
```
evanpurkhiser, line 51: Would be better to see this documented on the `GlobalFeedbackFormContextType` (or a better type name). The usage example is good

static/app/components/globalFeedbackForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+ */
+export function useFeedbackForm() {
+  return useContext(GlobalFeedbackFormContext);
+}
+
+function useInternalFeedbackForm(props: UseFeedbackOptions) {
```
malwilley, line 57: I moved the old `useFeedbackForm` into this component so that there isn't any confusion on which one to use. I don't think it's helpful outside this component anymore

ryan953, line 57: I would inline this into GlobalFeedbackForm

At the same time, don't pass `feedback` untouched through this provider. If people want to get a reference to that they should call `useFeedback()` directly imo. One way get the value, and it cleans up the useMemo stuff below.

static/app/components/globalFeedbackForm.tsx
ryan953, line 1: I'd name this file `useFeedbackForm.tsx` because it matches the export that'll be referenced most often.

 ie: makes sense to me to see "import A from A" instead of "import A from FOO" and i'm like, 'idk what FOO is'

static/app/components/globalFeedbackForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+interface GlobalFeedbackFormProps {
+  children: ReactNode;
+}
+
+interface GlobalFeedbackFormContextType {
+  feedback: FeedbackIntegration | undefined;
+  openForm: (options?: UseFeedbackOptions) => Promise<void>;
+}
+
+const GlobalFeedbackFormContext = createContext<GlobalFeedbackFormContextType>({
```
ryan953, line 27: +1 to simpler names. 

none of these are exported, so what if the names were `Props` `ContextType` and `const context = createContext(...);`

static/app/components/globalFeedbackForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+ *  Usage:
+ *
+ *  ```tsx
+ * const {feedback, openForm} = useFeedbackForm();
+ *
+ * if (!feedback) {
+ *   return null;
+ * }
+ *
+ * return <Button onClick={() => openForm()}>{t('Give Feedback')}</button>;
```
ryan953, line 50: What if `openForm` was nullable, whenever a form cannot be opened the function itself is unavailable. One less thing to import, and still gives callers the chance to check.


```suggestion
 * - `openForm`: A function that opens the feedback form. It accepts some
 *               option overrides to change how the form is displayed.  If undefined, the feedback
 *               integration is not available and feedback buttons should not
 *               be rendered.
 *
 *  Usage:
 *
 *  ```tsx
 * const openForm = useFeedbackForm();
 *
 * if (!openForm) {
 *   return null;
 * }
 *
 * return <Button onClick={() => openForm()}>{t('Give Feedback')}</button>;
```

static/app/components/globalFeedbackForm.tsx
malwilley: Good call - moved the description to the type def, but kept the usage on the hook since that's what users will be importing and looking at.

static/app/components/globalFeedbackForm.tsx
malwilley: Agree, moved to `useFeedbackForm.tsx`

static/app/components/globalFeedbackForm.tsx
malwilley: Changed to your suggestion below, so we aren't returning the feedback integration anymore. Agree that it's not necessary here.

Why inline the hook, though? I don't see any reasons for doing that.

static/app/components/globalFeedbackForm.tsx
malwilley: I like this suggestion! Does simplify it quite a bit.

static/app/components/globalFeedbackForm.tsx
ryan953: > Why inline the hook

i assumed GlobalFeedbackForm wouldn't have much content inside it anymore

static/app/components/globalFeedbackForm.spec.tsx
ryan953, line 1: the extra file

ryan953 (approved):

PR #74099, opened 2024-07-10
chore: Move normalizeUrl into its own file

ryan953:
`normalizeUrl()` was previously inside a file called `withDomainRequired.tsx` which also had a function `withDomainRequired()`. But these two functions are, implementation wise, unrelated. Conceptually aligned yes, but so are so many things.

Since `normalizeUrl()` is used in 100+ files, and `withDomainRequired()` used in only about 3 files, i moved `normalizeUrl` into it's own place so it could more easily be found by the next ~100 places that'll endup needing to import it.


michellewzhang (approved):

saponifi3d (approved):
lgtm!

PR #74103, opened 2024-07-10
metrics(feedback): additional user report counters

aliu39:
- Gives more detailed metrics on how many user reports are missing an event or blocked from new feedback. I tried querying gcp logs, but get different results each time I run the same query: https://cloudlogging.app.goo.gl/6MisKjx12nqxWCjV6
- Tracks the exact #reports updated by post process


michellewzhang (approved):
gif

PR #74104, opened 2024-07-10
chore(slack): remove old client from webhook handling code

cathteng:
Remove the old slack client completely from Slack code that handles updating a group, specifically removing from the code the processes updates from the archive and resolve modals (and updates the Slack message).

Also adds metrics for all of the group update code in Slack, we also have another place in the same function that updates the Slack message for group updates not through the modal (e.g. mark the issue as ongoing)

### DIFF
--- a/src/sentry/integrations/slack/webhooks/action.py
+++ b/src/sentry/integrations/slack/webhooks/action.py
@@ -633,42 +633,27 @@ class SlackActionEndpoint(Endpoint):
                 issue_details=True,
                 skip_fallback=True,
             ).build()
-            body = self.construct_reply(
-                blocks, is_message=slack_request.callback_data["is_message"]
-            )
-            # use the original response_url to update the link attachment
-            if not features.has(
-                "organizations:slack-sdk-webhook-handling", group.project.organization
-            ):
-                slack_client = SlackClient(integration_id=slack_request.integration.id)
-                try:
-                    private_metadata = orjson.loads(
-                        slack_request.data["view"]["private_metadata"],
-                    )
-                    slack_client.post(private_metadata["orig_response_url"], data=body, json=True)
-                except ApiError as e:
-                    logger.error("slack.action.response-error", extra={"error": str(e)})
 
-            else:
-                json_blocks = orjson.dumps(blocks.get("blocks")).decode()
-                view = View(**slack_request.data["view"])
-                try:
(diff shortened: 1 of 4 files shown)


src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -655,42 +660,37 @@ def _handle_group_actions(
...
+                webhook_client.send(
+                    blocks=json_blocks, delete_original=False, replace_original=True
+                )
+                logger.info(
+                    "slack.webhook.view_submission.success",
+                    extra={
+                        "integration_id": slack_request.integration.id,
+                        "blocks": json_blocks,
+                    },
+                )
```
ykamo001, line 679: Do we need the success log now that we're moving to enabled? Plus we have the metrics for it to monitor

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -655,42 +660,37 @@ def _handle_group_actions(
...
+                    tags={"type": "submit_modal"},
+                )
+            except SlackApiError as e:
+                metrics.incr(
+                    SLACK_WEBHOOK_GROUP_ACTIONS_FAILURE_DATADOG_METRIC,
+                    sample_rate=1.0,
+                    tags={"type": "submit_modal"},
+                )
+                logger.error(
+                    "slack.webhook.view_submission.response-error", extra={"error": str(e)}
```
ykamo001, line 692: do we want to log more information to help with debugging?

ykamo001 (approved):

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -655,42 +660,35 @@ def _handle_group_actions(
...
+                    tags={"type": "submit_modal"},
+                )
+                logger.error(
+                    "slack.webhook.view_submission.response-error",
+                    extra={
+                        "error": str(e),
+                        "integration_id": slack_request.integration.id,
+                        "organization_id": group.project.organization_id,
+                    },
+                )
```
RyanSkonnord, line 691: Thought for later: we repeat this logging and metrics in a lot of `except SlackApiError` blocks, right? We could try to abstract it into a utility function -- ideally a context manager that could take the place of the `try` block.

RyanSkonnord (approved):

PR #74105, opened 2024-07-10
feat(apis): add ApiTokenDetailsEndpoint

ameliahsu:
created ApiTokenDetailsEndpoint with GET and PUT endpoints to fetch and modify user auth tokens

https://github.com/getsentry/sentry/issues/73556


src/sentry/api/endpoints/api_token_details.py
```diff
@@ -0,0 +1,71 @@
...
+
+ALLOWED_FIELDS = ["name", "tokenId"]
+
+
+class ApiTokenNameSerializer(serializers.Serializer):
+    name = CharField(max_length=255, allow_blank=True, required=True)
+
+
+@control_silo_endpoint
+class ApiTokenDetailsEndpoint(Endpoint):
```
cathteng, line 25: i would recommend adding a `convert_args` function to this that fetches the token from the `token_id`, then you can pass the resource directly into the GET/PUT methods here. example: https://github.com/getsentry/sentry/blob/4652fc7032c182ba8ce50daa64cf39766b32158e/src/sentry/api/endpoints/project_details.py#L473

also, should we move the DELETE method here as well?

ameliahsu: added DELETE method (will need to change UI to call this details endpoint later)

src/sentry/api/endpoints/api_token_details.py
```diff
@@ -0,0 +1,92 @@
...
+            user_id = _get_appropriate_user_id(request=request)
+
+            try:
+                token_to_rename = ApiToken.objects.get(
+                    id=token_id, application__isnull=True, user_id=user_id
+                )
+            except ApiToken.DoesNotExist:
+                raise ResourceDoesNotExist
+
+            token_to_rename.name = result.get("name", None)
```
nhsiehgit, line 69: nit: Name is required right? (re: serializer?)
if so, then i think the `.get` is unnecessary, otherwise we're assuming `token_to_rename.name` _could_ be `None` - which can it?

src/sentry/api/endpoints/api_token_details.py
```diff
@@ -0,0 +1,92 @@
...
+from rest_framework.fields import CharField
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.request import Request
+from rest_framework.response import Response
+
+from sentry import analytics
+from sentry.api.api_owners import ApiOwner
+from sentry.api.api_publish_status import ApiPublishStatus
+from sentry.api.base import Endpoint, control_silo_endpoint
+from sentry.api.endpoints.api_tokens import _get_appropriate_user_id
```
nhsiehgit, line 13: prepended `_`methods should _only_ be used within the files they're defined.
If you're making this a global helper, then we should remove the `_`

src/sentry/api/endpoints/api_tokens.py
```diff
@@ -24,11 +24,28 @@
 ALLOWED_FIELDS = ["name", "tokenId"]
 
 
-class ApiTokenNameSerializer(serializers.Serializer):
+def _get_appropriate_user_id(request: Request) -> int:
```
nhsiehgit, line 27: Any reason why we moved this up?

It makes this PR seem like there's a lot more changes 😅

nhsiehgit: Oh i see.... to make it a global 🤔 

If it was a class method before - we could just call it on the class?

eg
```
ApiTokenSerializer._get_appropriate_user_id(request)
```
?

Is this not acceptable?

nhsiehgit (commented):
few prelim comments

src/sentry/api/endpoints/api_tokens.py
ameliahsu: Ah you're right, changed it to be a static method

src/sentry/api/endpoints/api_token_details.py
ameliahsu: fixed - updated the `.get` to remove the default value

src/sentry/api/endpoints/api_token_details.py
schew2381: good catch!

tests/sentry/api/endpoints/test_api_token_details.py
```diff
@@ -0,0 +1,154 @@
+from rest_framework import status
+
+from sentry.models.apitoken import ApiToken
+from sentry.testutils.cases import APITestCase
+from sentry.testutils.silo import control_silo_test
+
+
+@control_silo_test
+class ApiTokenDetailTest(APITestCase):
```
nhsiehgit, line 9: Maybe we can name the tests based on the methods theyre testing?

so this would be `ApiTokenGetTest`
then PUT
then DELETE

tests/sentry/api/endpoints/test_api_token_details.py
```diff
@@ -0,0 +1,154 @@
...
+        self.login_as(self.user)
+        response = self.get_success_response(
+            token.id, status_code=status.HTTP_204_NO_CONTENT, **payload
+        )
+        assert (
+            response.get("cache-control")
+            == "max-age=0, no-cache, no-store, must-revalidate, private"
+        )
+
+    def test_delete_name(self):
```
nhsiehgit, line 73: to not be confused with the `DELETE` method, lets call this `remove_name` instead

nhsiehgit (commented):
few more comments!

src/sentry/api/endpoints/api_token_details.py
```diff
@@ -0,0 +1,92 @@
...
+            user_id = ApiTokensEndpoint._get_appropriate_user_id(request=request)
+
+            try:
+                token_to_rename = ApiToken.objects.get(
+                    id=token_id, application__isnull=True, user_id=user_id
+                )
+            except ApiToken.DoesNotExist:
+                raise ResourceDoesNotExist
+
+            token_to_rename.name = result.get("name")
```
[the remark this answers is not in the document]
ameliahsu: Yeahh you can also do it that way but the org auth tokens uses `.get` so I wanted to be consistent 🥲

src/sentry/api/endpoints/api_tokens.py
```diff
@@ -39,13 +36,12 @@ class ApiTokensEndpoint(Endpoint):
...
         "POST": ApiPublishStatus.PRIVATE,
-        "PUT": ApiPublishStatus.PRIVATE,
     }
     authentication_classes = (SessionNoAuthTokenAuthentication,)
     permission_classes = (IsAuthenticated,)
 
-    @classmethod
-    def _get_appropriate_user_id(cls, request: Request) -> int:
+    @staticmethod
+    def _get_appropriate_user_id(request: Request) -> int:
```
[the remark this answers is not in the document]
ameliahsu: Cleanup would be done in a different PR, correct?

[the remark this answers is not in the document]
nhsiehgit: Yea yea unless you want to do it here.... 
but we can just get this one merged in and keep a ticket to restructure later 🤞

src/sentry/api/endpoints/api_token_details.py
```diff
@@ -0,0 +1,92 @@
...
+    publish_status = {
+        "GET": ApiPublishStatus.PRIVATE,
+        "PUT": ApiPublishStatus.PRIVATE,
+        "DELETE": ApiPublishStatus.PRIVATE,
+    }
+    owner = ApiOwner.SECURITY
+    permission_classes = (IsAuthenticated,)
+
+    @method_decorator(never_cache)
+    def get(self, request: Request, token_id: int) -> Response:
```
sentaur-athena, line 36: Is this used anywhere in the product? Or are you using it as part of your ticket. If not you can remove it.

src/sentry/api/endpoints/api_token_details.py
```diff
@@ -0,0 +1,92 @@
...
+                )
+            except ApiToken.DoesNotExist:
+                raise ResourceDoesNotExist
+
+            token_to_rename.name = result.get("name")
+            token_to_rename.save()
+
+            return Response(status=204)
+
+        return Response(serializer.errors, status=400)
```
sentaur-athena, line 74: Nit: You can change line 57 to if serializer is not valid and return early. It's more readable that way where serializer error is closer to the validation.

src/sentry/api/endpoints/api_token_details.py
```diff
@@ -0,0 +1,92 @@
...
+                token_to_rename = ApiToken.objects.get(
+                    id=token_id, application__isnull=True, user_id=user_id
+                )
+            except ApiToken.DoesNotExist:
+                raise ResourceDoesNotExist
+
+            token_to_rename.name = result.get("name")
+            token_to_rename.save()
+
+            return Response(status=204)
```
sentaur-athena, line 72: I recommend status code 200 with the updated token as response. Make sure to not serialize the token.

sentaur-athena (commented):
Looks good! Thank you for adding the new endpoint.

src/sentry/api/endpoints/api_token_details.py
ameliahsu: The GET method will be used for the UI to edit user auth token details (next PR)

sentaur-athena (approved):

src/sentry/api/endpoints/api_token_details.py
```diff
@@ -0,0 +1,92 @@
...
+        "PUT": ApiPublishStatus.PRIVATE,
+        "DELETE": ApiPublishStatus.PRIVATE,
+    }
+    owner = ApiOwner.SECURITY
+    permission_classes = (IsAuthenticated,)
+
+    @method_decorator(never_cache)
+    def get(self, request: Request, token_id: int) -> Response:
+
+        user_id = ApiTokensEndpoint.get_appropriate_user_id(request=request)
```
markstory, line 38: Having endpoints use each other for methods is not a pattern we frequently use. Instead you could promote this method to a module level function and have both endpoints use it that way.

ameliahsu: @nhsiehgit mentioned creating another ticket to move this into a separate file and group `api_token` files under a separate `endpoints/api_tokens/...` folder

nhsiehgit: @ameliahsu have/can you create a jira task to do this and add it to your sprint?

nhsiehgit: Oh wai t- i see some updated PRs referencing this ticket 🙏 
perfect, thank you!

schew2381:
> ## Suspect Issues
> 
> * ‼️ **ValueError: invalid literal for int() with base 10: 'pal-festin'** `/api/0/api-tokens/` [View Issue](https://sentry.sentry.io/issues/5601658294/?referrer=github-pr-bot)
> 

@ameliahsu Could we add a try-except around these lines to solve the error above?
https://github.com/getsentry/sentry/blob/f24951b7b119e80cea2bdc0eaa5bf15dd6f71517/src/sentry/api/endpoints/api_tokens.py#L58-L59

Also not sure if you've tried this before but you can put the `Closes {Sentry Issue Title` ([docs here](https://github.com/getsentry/sentry/blob/f24951b7b119e80cea2bdc0eaa5bf15dd6f71517/src/sentry/api/endpoints/api_tokens.py#L58-L59)) in your PR description and it'll automatically link and then resolve the sentry issue when the PR is merged

PR #74106, opened 2024-07-10
ref(typing): Type the snuba discover module

wmak:
- This types the discover snuba module so that it can be removed from the pyproject.toml
- Had to move the Span type from the endpoint to the search list to avoid a circular import
  - Its heavily used in the Discover module so i think it's a reasonable move
- Changed group_ids in the group model to an Iterable -- tbh I think we can swap a LOT of instances of Sequence to Iterable, but I rather keep impact of this change small
- Removed the _query etc. methods, they're unnecessary since the default param already handles the functionality


wmak:
Here's a gist of all the original mypy failures to help review: https://gist.github.com/wmak/77c1fa6950578dec27ea1315a34de9de

src/sentry/search/events/builder/base.py
```diff
@@ -1483,7 +1483,7 @@ def handle_invalid_float(cls, value: float | None) -> float | None:
             return None
         return value
 
-    def run_query(self, referrer: str, use_cache: bool = False) -> Any:
+    def run_query(self, referrer: str | None, use_cache: bool = False) -> Any:
```
wmak, line 1486: This should be a no-op 1487 was already handling the cases where referrer is None

src/sentry/search/events/types.py
```diff
@@ -55,14 +56,18 @@ class QueryFramework:
     entity: Entity
 
 
+SnubaRow = dict[str, Any]
```
wmak, line 59: afaik this is the best we can do for Snuba's response, we know the keys are strings matching our field names, but the values can literally be anything

src/sentry/snuba/discover.py
```diff
@@ -80,11 +93,6 @@ class EventsMeta(TypedDict):
 EventsMeta.__annotations__["isMetricsExtractedData"] = bool
 
 
-class EventsResponse(TypedDict):
-    data: list[dict[str, Any]]
-    meta: EventsMeta
-
-
```
wmak, line 87: Removed this, its duplicated in search.events.types

src/sentry/snuba/discover.py
```diff
@@ -105,106 +113,155 @@ def is_real_column(col):
     return True
 
 
-def format_time(data, start, end, rollup, orderby):
-    rv = []
-    start = int(to_naive_timestamp(naiveify_datetime(start)) / rollup) * rollup
-    end = (int(to_naive_timestamp(naiveify_datetime(end)) / rollup) * rollup) + rollup
-    data_by_time = {}
+def format_time(
+    data: SnubaData, start_param: datetime, end_param: datetime, rollup: int, orderby: list[str]
```
wmak, line 117: Needed to rename the start & end params since their type changes later on 121

src/sentry/snuba/discover.py
```diff
@@ -105,106 +113,155 @@ def is_real_column(col):
...
+) -> SnubaData:
+    """Format the time field from a snuba response so that we can easily use it in the rest of processing"""
+    return_value: SnubaData = []
+    start = int(to_naive_timestamp(naiveify_datetime(start_param)) / rollup) * rollup
+    end = (int(to_naive_timestamp(naiveify_datetime(end_param)) / rollup) * rollup) + rollup
+    data_by_time: dict[int, SnubaData] = {}
 
-    for obj in data:
+    for row in data:
```
wmak, line 125: renamed this just cause `obj` was bothering me haha, same with `rv` on 120

src/sentry/search/events/types.py
wmak: Tried typing this with:
`JSON: TypeAlias = dict[str, "JSON"] | list["JSON"] | str | int | float | bool | None`
Which then made everything using results break since we aren't typing the usage of SnubaResults much yet. Going to leave this as Any for now

src/sentry/snuba/discover.py
wmak: :+1: not used anywhere except `src/sentry/snuba/metrics_performance.py` which just passes params

src/sentry/snuba/discover.py
```diff
@@ -247,14 +304,44 @@ def _timeseries_query(
...
+    allow_metric_aggregates - Ignored here, only used in metric enhanced performance
+    """
+    assert dataset in [
+        Dataset.Discover,
+        Dataset.Transactions,
+    ], "A dataset is required to query discover"
+    assert (
+        "start" not in params or params["start"] is not None
+    ), "Start is required to query timeseries"
+    assert "end" not in params or params["end"] is not None, "End is required to query timeseries"
```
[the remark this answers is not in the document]
wmak: woops yeah good catch

src/sentry/snuba/discover.py
```diff
@@ -1040,7 +942,7 @@ def histogram_query(
         elif array_column == "span_op_breakdowns":
             key_column = "array_join(span_op_breakdowns_key)"
             histogram_function = get_span_op_breakdown_name
-        else:
+        elif array_column is None:
```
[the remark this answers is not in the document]
wmak: It shouldn't https://github.com/getsentry/sentry/pull/74106/files#diff-f9a8d9bbcb1bf33b928235196e41ca3666cc8c508bbb46d3238aae26d3e13a6cR938 only returns `"measurements", "span_op_breakdowns" or None`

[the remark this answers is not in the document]
Zylphrex: Isn't this strictly worse though? Because python doesn't force you to check all possible cases, it's entirely possible the function changes its return type (in this case, it's not likely to change), and then fails here.

So this specific use case is probably okay but is there a reason not to leave this as `else` as a catch-all?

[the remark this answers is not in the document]
wmak: That's a good point, i'll switch this to Else

src/sentry/api/endpoints/organization_traces.py
```diff
@@ -130,7 +130,7 @@ def get(self, request: Request, organization: Organization) -> Response:
         serialized = serializer.validated_data
 
         executor = TracesExecutor(
-            params=cast(ParamsType, params),
```
Zylphrex, line 133: Nice! 💅

src/sentry/dynamic_sampling/tasks/custom_rule_notifications.py
```diff
@@ -82,18 +82,24 @@ def get_num_samples(rule: CustomDynamicSamplingRule) -> int:
         # org rule get all projects for org
         projects = rule.organization.project_set.filter(status=ObjectStatus.ACTIVE)
 
-    params: dict[str, Any] = {
+    project_id = []
+    project_objects = []
```
Zylphrex, line 86: Isn't this list of project objects redundant?

wmak: Hmm if you're asking why i switched to a for loop over a list comprehension was that i'm worried about hitting the queryset twice
If you're asking why not just pass projects, its cause the type is a list[project] not queryset

Zylphrex: I see, not big deal so feel free to disregard.

src/sentry/snuba/discover.py
wmak: turns out, shouldn't have removed this, needed for the response typing of the events endpoint, moved this there instead

Zylphrex (approved):

PR #74109, opened 2024-07-10
ref(replay): replace some replay `ObjectInspector` references

michellewzhang:

replaces some instances of `ObjectInspector` in replay details with the issues `StructuredEventData` -- namely, instances on the network tab.

|  after `StructuredEventData` |  before `ObjectInspector`|
| ------------- | ------------- |
| SCR-20240710-odxm | SCR-20240710-oebz |
| SCR-20240710-oktd | SCR-20240710-olfl |

### DIFF
--- a/static/app/components/replays/breadcrumbs/breadcrumbItem.tsx
+++ b/static/app/components/replays/breadcrumbs/breadcrumbItem.tsx
@@ -8,11 +8,11 @@ import {CodeSnippet} from 'sentry/components/codeSnippet';
 import {Flex} from 'sentry/components/container/flex';
 import ErrorBoundary from 'sentry/components/errorBoundary';
 import Link from 'sentry/components/links/link';
+import ObjectInspector from 'sentry/components/objectInspector';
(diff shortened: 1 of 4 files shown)


aliu39:
where are other references?

michellewzhang:
> where are other references?

SCR-20240710-oogf

aliu39 (approved):

billyvg (approved):
Ugh don't love that this component doesn't quote the strings -- this means that copy/pasting isn't guaranteed to have valid JSON.

image

michellewzhang:
> Ugh don't love that this component doesn't quote the strings -- this means that copy/pasting isn't guaranteed to have valid JSON.
> 
> image

yeah, it looks like we can customize how we render only certain types of data types :( object keys for example, but not object values

https://github.com/getsentry/sentry/blob/1d8bcf330d7ea91be04df91579f958d9a46627e1/static/app/components/structuredEventData/index.tsx#L21-L30

PR #74111, opened 2024-07-10
nit(opsgenie): add config access requirement to error text

mifu67:
Users who set up their Opsgenie API keys via the Opsgenie API were experiencing problems due to a lack of configuration access. However, the error message that they encountered did not indicate that this was the problem, so this PR makes the issue more clear.


cathteng:
what is configuration access? do you have a link

mifu67:
This is specifically for key creation via the API; keys created in the Opsgenie UI seem to have configuration access enabled automatically

cathteng (approved):

cathteng:
do we need to point them to documentation?

mifu67:
The Opsgenie docs don't say anything about needing config access to authenticate. It could be helpful to point to the Sentry docs. What do you think?

PR #74112, opened 2024-07-10
feat(query-builder): Visually mark boolean ops and parens as invalid in the issue search

malwilley:
- Adds an option to the syntax parser to consider parentheses invalid (there already exists one for boolean ops)
- Adds a prop to SearchQueryBuilder to disable logical operations
- When invalid, shows the token as red and displays a tooltip when focused or hovered

CleanShot 2024-07-10 at 15 50 39

CleanShot 2024-07-10 at 15 50 20

### DIFF
--- a/static/app/components/searchQueryBuilder/boolean.tsx
+++ b/static/app/components/searchQueryBuilder/boolean.tsx
@@ -20,7 +20,13 @@ export function SearchQueryBuilderBoolean({
   token,
 }: SearchQueryBuilderBooleanProps) {
   return (
-    <DeletableToken item={item} state={state} token={token} label={token.value}>
+    <DeletableToken
+      item={item}
+      state={state}
+      token={token}
+      label={token.value}
+      invalid={token.invalid}
+    >
       {token.text}
     </DeletableToken>
   );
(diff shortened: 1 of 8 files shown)


roggenkemper (commented):
is there a way to prevent the search if we find the disallowed operators in the search query? we have an error message that shows up for and/or but not one for parentheses if someone does input an invalid search. i feel like streamlining this experience would be helpful

somewhat related - if i'm typing a query with and/or right now and immediately press enter after typing it, it's not highlighted red (this is probably harder to fix)

malwilley:
> is there a way to prevent the search if we find the disallowed operators in the search query? we have an error message that shows up for and/or but not one for parentheses if someone does input an invalid search. i feel like streamlining this experience would be helpful
> 
> somewhat related - if i'm typing a query with and/or right now and immediately press enter after typing it, it's not highlighted red (this is probably harder to fix)

I agree that preventing the search and showing the error message would be the best UX, so I'll add that to the feedback doc. Not sure if we'll do that right away though since there's some complexity to that and invalid searches aren't very common.

roggenkemper (approved):

PR #74115, opened 2024-07-10
feat(alert rules): Add columns for anomaly detection alerts

ceorourke:
Add new columns needed to support anomaly detection metric alerts.

Closes https://getsentry.atlassian.net/browse/ALRT-138


src/sentry/incidents/models/alert_rule.py
```diff
@@ -77,6 +78,28 @@ class AlertRuleStatus(Enum):
     DISABLED = 5
 
 
+class AlertRuleDetectionType(models.IntegerChoices):
```
ceorourke, line 81: I'm using a new (to us) thing here to avoid situations like https://github.com/getsentry/sentry/blob/master/src/sentry/incidents/models/alert_rule.py#L391-L409 where we reference an enum but require a comment to know which one.

Django docs are here: https://docs.djangoproject.com/en/5.0/ref/models/fields/#field-choices-enum-types
and we do have _one_ instance of this being used in Sentry today https://github.com/getsentry/sentry/blob/master/src/sentry/models/dashboard_widget.py#L161-L182

ceorourke: I could be convinced to change this to "static", "percent", and "dynamic" which would require a backfill, but could be done. The thinking here was just that it's a static threshold in the sense that it's user defined or it's dynamic in the sense that they've chosen anomaly detection.

src/sentry/incidents/models/alert_rule.py
```diff
@@ -251,24 +274,31 @@ class AlertRule(Model):
     excluded_projects = models.ManyToManyField(
         "sentry.Project", related_name="alert_rule_exclusions", through=AlertRuleExcludedProjects
     )  # NOTE: This feature is not currently utilized.
-    name = models.TextField()
```
ceorourke, line 254: I just moved things around so that `excluded_projects` and `include_all_projects` are next to each other

src/sentry/migrations/0734_metric_alert_anomaly_detection.py
```diff
@@ -0,0 +1,68 @@
...
+            model_name="alertrule",
+            name="sensitivity",
+            field=models.SmallIntegerField(null=True),
+        ),
+        migrations.AlterField(
+            model_name="alertrule",
+            name="threshold_period",
+            field=models.IntegerField(null=True),
+        ),
+        migrations.SeparateDatabaseAndState(
```
ceorourke, line 48: This is from the adding columns with a default section https://develop.sentry.dev/database-migrations/#adding-columns-with-a-default

ceorourke:
idk what this test failure is

src/sentry/incidents/models/alert_rule.py
nhsiehgit: Nice! Great find!

If we're gonna name these string values anyways, i think i'm down to make these `models.TextChoices`
makes it easier to see what the selection is when sifting through db data in redash or something

nhsiehgit: I'm also strongly 👍  on being explicit between static and percent.
Otherwise this is effectively a boolean value.

Static threshold to me means the _threshold_ is static - not that i've chosen a static value

src/sentry/incidents/models/alert_rule.py
```diff
@@ -77,6 +78,28 @@ class AlertRuleStatus(Enum):
...
+    DYNAMIC = 1, gettext_lazy("Dynamic")
+
+
+class AlertRuleSensitivity(models.IntegerChoices):
+    LOW = 0, gettext_lazy("Low")
+    MEDIUM = 1, gettext_lazy("Medium")
+    HIGH = 2, gettext_lazy("High")
+
+
+class AlertRuleSeasonality(models.IntegerChoices):
```
nhsiehgit, line 92: Can we add a note here to call out that we've iterated through all combinations and that we don't foresee expanding this?

src/sentry/incidents/models/alert_rule.py
```diff
@@ -251,24 +274,31 @@ class AlertRule(Model):
...
     include_all_projects = models.BooleanField(
         default=False
     )  # NOTE: This feature is not currently utilized.
+    name = models.TextField()
+    status = models.SmallIntegerField(default=AlertRuleStatus.PENDING.value)
     threshold_type = models.SmallIntegerField(null=True)
     resolve_threshold = models.FloatField(null=True)
     # How many times an alert value must exceed the threshold to fire/resolve the alert
-    threshold_period = models.IntegerField()
+    threshold_period = models.IntegerField(null=True)
```
nhsiehgit, line 287: Does this conflict with anomaly detection?
Why do we need to enable nullable here?

src/sentry/incidents/models/alert_rule.py
```diff
@@ -251,24 +274,31 @@ class AlertRule(Model):
...
     date_modified = models.DateTimeField(default=timezone.now)
     date_added = models.DateTimeField(default=timezone.now)
     monitor_type = models.IntegerField(default=AlertRuleMonitorTypeInt.CONTINUOUS)
     description = models.CharField(max_length=1000, null=True)
+    detection_type = models.SmallIntegerField(
+        default=AlertRuleDetectionType.STATIC, choices=AlertRuleDetectionType.choices
+    )
+    sensitivity = models.SmallIntegerField(choices=AlertRuleSensitivity.choices, null=True)
+    seasonality = models.SmallIntegerField(
+        default=AlertRuleSeasonality.AUTO, choices=AlertRuleSeasonality.choices, null=True
```
nhsiehgit, line 300: should we default this to null?
Otherwise we could have static detection w/ auto seasonality and that doesn't make much sense? (but also hopefully shoudln't break anything🤔 )

nhsiehgit (commented):
few initial comments

src/sentry/incidents/models/alert_rule.py
ceorourke: There will be no threshold period for anomaly detection since that's the "1 minute" dropdown in this: 
Screenshot 2024-07-11 at 2 06 35 PM

src/sentry/incidents/models/alert_rule.py
ceorourke: Yeah I could go either way on that, I can make the default null

src/sentry/migrations/0738_metric_alert_anomaly_detection.py
```diff
@@ -0,0 +1,93 @@
...
+    STATIC = "Static", gettext_lazy("Static")
+    PERCENT = "Percent", gettext_lazy("Percent")
+    DYNAMIC = "Dynamic", gettext_lazy("Dynamic")
+
+
+def _backfill_detection_type(apps: StateApps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    AlertRule = apps.get_model("sentry", "AlertRule")
+
+    # use RangeQuerySetWrapper to avoid loading all AlertRules into memory
+    for rule in RangeQuerySetWrapper(AlertRule.objects.all()):
```
ceorourke, line 22: not sure what I am missing but when I try to run this locally it's failing saying `AttributeError: type object 'AlertRule' has no attribute 'objects'. Did you mean: 'projects'?` huh??

src/sentry/incidents/models/alert_rule.py
```diff
@@ -251,24 +279,29 @@ class AlertRule(Model):
...
     # This represents a time delta, in seconds. If not null, this is used to determine which time
     # window to query to compare the result from the current time_window to.
     comparison_delta = models.IntegerField(null=True)
     date_modified = models.DateTimeField(default=timezone.now)
     date_added = models.DateTimeField(default=timezone.now)
     monitor_type = models.IntegerField(default=AlertRuleMonitorTypeInt.CONTINUOUS)
     description = models.CharField(max_length=1000, null=True)
+    detection_type = models.SmallIntegerField(
+        default=AlertRuleDetectionType.STATIC, choices=AlertRuleDetectionType.choices
+    )
```
markstory, line 302: Nice that Django's choice/enum models work well with fields now. We could start using these choice enums more as we have a bunch of duplicate code to define `choices` methods on Enums throughout the application.

src/sentry/migrations/0738_metric_alert_anomaly_detection.py
markstory: The `AlertRule` model defines a custom default manager (`Meta.default_manager_name`), you need to use `AlertRule.objects_with_snapshots`

src/sentry/migrations/0738_metric_alert_anomaly_detection.py
```diff
@@ -0,0 +1,93 @@
...
+    # of a table.
+    # Here are some things that make sense to mark as post deployment:
+    # - Large data migrations. Typically we want these to be run manually so that they can be
+    #   monitored and not block the deploy for a long period of time while they run.
+    # - Adding indexes to large tables. Since this can take a long time, we'd generally prefer to
+    #   run this outside deployments so that we don't block them. Note that while adding an index
+    #   is a schema change, it's completely safe to run the operation after the code has deployed.
+    # Once deployed, run these manually via: https://develop.sentry.dev/database-migrations/#migration-deployment
+
+    is_post_deployment = False
```
markstory, line 45: AlertRule is a biggish table, I don't think a backfill will fit within the deployment timeout. Splitting the migration up into two migrations could work well. One migration for the new columns, and a second for the backfill.

markstory:
> ALTER TABLE "sentry_alertrule" ALTER COLUMN "threshold_period" DROP NOT NULL;

Are we confident that there aren't any nulls left in single-tenants? I checked US/DE and they were clear.

ceorourke:
> > ALTER TABLE "sentry_alertrule" ALTER COLUMN "threshold_period" DROP NOT NULL;
> 
> Are we confident that there aren't any nulls left in single-tenants? I checked US/DE and they were clear.

This is changing it so that it can be nullable, how there could have been nulls in the past and why that would be a problem? How can I check data in st?

src/sentry/incidents/models/alert_rule.py
wedamija: That dropdown is the `time_window`, which is on the `SnubaQuery`. The `threshold_period` is used to set how many failures/successes in a row are used to trigger the alert

wedamija (changes requested):

src/sentry/incidents/models/alert_rule.py
```diff
@@ -77,6 +78,33 @@ class AlertRuleStatus(Enum):
...
+class AlertRuleDetectionType(models.TextChoices):
+    STATIC = "Static", gettext_lazy("Static")
+    PERCENT = "Percent", gettext_lazy("Percent")
+    DYNAMIC = "Dynamic", gettext_lazy("Dynamic")
+
+
+class AlertRuleSensitivity(models.TextChoices):
+    LOW = "Low", gettext_lazy("Low")
+    MEDIUM = "Medium", gettext_lazy("Medium")
+    HIGH = "High", gettext_lazy("High")
```
wedamija, line 90: Usually for db constants we'd use lower case

src/sentry/migrations/0740_metric_alert_anomaly_detection.py
```diff
@@ -0,0 +1,58 @@
...
+        migrations.AddField(
+            model_name="alertrule",
+            name="sensitivity",
+            field=models.CharField(null=True),
+        ),
+        migrations.SeparateDatabaseAndState(
+            database_operations=[
+                migrations.RunSQL(
+                    """
+                    ALTER TABLE "sentry_alertrule" ADD COLUMN "detection_type" VARCHAR(32) NOT NULL DEFAULT 'Static';
```
wedamija, line 42: Default should be updated here

src/sentry/migrations/0740_metric_alert_anomaly_detection.py
```diff
@@ -0,0 +1,58 @@
...
+                    ALTER TABLE "sentry_alertrule" DROP COLUMN "detection_type";
+                    """,
+                    hints={"tables": ["sentry_alertrule"]},
+                ),
+            ],
+            state_operations=[
+                migrations.AddField(
+                    model_name="alertrule",
+                    name="detection_type",
+                    field=models.CharField(default="Static"),
```
wedamija, line 54: Same here

wedamija (approved):

PR #74116, opened 2024-07-10
feat(query-builder): Add feedback button

malwilley:
- Modifies `useFeedback` to accept `tags` so we can pass that info into the form
- Adds a footer with a feedback button to the empty state key explorer 

CleanShot 2024-07-10 at 16 50 04

### DIFF
--- a/static/app/components/feedback/widget/useFeedback.tsx
+++ b/static/app/components/feedback/widget/useFeedback.tsx
@@ -6,14 +6,17 @@ import ConfigStore from 'sentry/stores/configStore';
 import {useLegacyStore} from 'sentry/stores/useLegacyStore';
 import useAsyncSDKIntegrationStore from 'sentry/views/app/asyncSDKIntegrationProvider';
 
+type ArgumentTypes<F extends Function> = F extends (...args: infer A) => any ? A : never;
+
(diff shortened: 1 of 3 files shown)


static/app/components/feedback/widget/useFeedback.tsx
```diff
@@ -6,14 +6,17 @@ import ConfigStore from 'sentry/stores/configStore';
 import {useLegacyStore} from 'sentry/stores/useLegacyStore';
 import useAsyncSDKIntegrationStore from 'sentry/views/app/asyncSDKIntegrationProvider';
 
+type ArgumentTypes<F extends Function> = F extends (...args: infer A) => any ? A : never;
```
billyvg, line 9: Is this the same as [`Parameters`](https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype)?

billyvg (approved):

roggenkemper (approved):

static/app/components/feedback/widget/useFeedback.tsx
malwilley: You're totally right - thank you!

PR #74117, opened 2024-07-11
feat(uptime): Add organization level setting for uptime autodetection

davidenwang:
Add organization level setting for uptime autodetection, similar to https://github.com/getsentry/sentry/pull/74052

### DIFF
--- a/src/sentry/api/serializers/models/organization.py
+++ b/src/sentry/api/serializers/models/organization.py
@@ -48,6 +48,7 @@ from sentry.constants import (
     SAFE_FIELDS_DEFAULT,
     SCRAPE_JAVASCRIPT_DEFAULT,
     SENSITIVE_FIELDS_DEFAULT,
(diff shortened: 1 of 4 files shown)


wedamija (approved):

src/sentry/api/serializers/models/organization.py
```diff
@@ -457,6 +458,7 @@ class DetailedOrganizationSerializerResponse(_DetailedOrganizationSerializerResp
     metricsActivateLastForGauges: bool
     extrapolateMetrics: bool
     requiresSso: bool
+    uptimeAutodetection: bool
```
isabellaenriquez, line 461: since adding this to the response is behind a flag, should this be in `_DetailedOrganizationSerializerResponseOptional` instead?

davidenwang: ahha yes it should

PR #74134, opened 2024-07-11
fix(metrics) Do not allow group by non existent tags

### LINKED ISSUE #73909: Allow grouping by non existing tag
When doing a group by on a tag that has not been yet ingested, return a single group with tag value (none) instead of hard failing the entire query

vgrozdanic:
image

Instead of reading available tags from extraction rule for span basaed (virtual) metric, based on selected aggregation fetch ingested tags and allow only group by the ones we have seen. Tags that exists in extraction rules, but are not seen in the selected time range, are disabled and it is not possible to group by them.

As a part of this PR, tags in a dropdown menu are no longer clickable, because it is not possible to nicely do the checks if they exist in the selected time range. This is the simplest option to deal with this problem:
image

Fixes and closes: https://github.com/getsentry/sentry/issues/73909


static/app/components/metrics/mriSelect/metricListItemDetails.tsx
```diff
@@ -207,12 +205,7 @@ export function MetricListItemDetails({
                 return (
                   <Fragment key={tag.key}>
                     <TagWrapper>
-                      <Button
```
ArthurKnaus, line 210: Shouldn't we do the same here? Only disable those that are not available in the time range?

ArthurKnaus: hmm, right. What is even more important, at this point the metric is most likely not yet selected. Therefore, we do not have a selected condition and aggregate that we could use for resolving.
k, let's keep it disabled 👍  But only for virtual metrics pls.

vgrozdanic: Yes, that was my idea also, only tags from virtual metrics might be disabled (have `isQueryable` field set to false). That check is happening in https://github.com/getsentry/sentry/pull/74134/files#diff-24ee1ab39c06e08aaf8bf035fcbd7d2256f8b056daf112bf8ec9070b94982dafR90-R107

ArthurKnaus: but in here is no check?

vgrozdanic: Conditional "clickability" added

ArthurKnaus (approved):

PR #74135, opened 2024-07-11
feat(profiling) render continuous profile chunk flamechart

JonasBa:
Renders continuous flamechart for a profile chunk.

Disclaimer:
This is one big copy paste of the flamechart component with removal of span tree code and any reference to the old txn model. Since the underlying components are not tied to this model anymore, it means they can be wired up to work with the continuous model. I opted for this to have cleaner separation and make it easier to make modifications without having to think of both models - since the main component mostly just wires up the different parts of the UI, I think that is a good approach to take


static/app/utils/profiling/profile/continuousProfile.tsx
```diff
@@ -48,7 +48,7 @@ export class ContinuousProfile extends Profile {
       const stack = chunk.stacks[sample.stack_id];
       let size = 0;
 
-      for (let j = 0; j < stack.length; j++) {
+      for (let j = stack.length - 1; j >= 0; j--) {
```
Zylphrex, line 51: This was what we talked about that caused the stack to be upside down right?

static/app/components/profiling/flamegraph/continuousFlamegraph.tsx
Zylphrex, line 1: Is there any part of this file that is a big change from what we had originally? Hard to parse what may have changed from this diff.

static/app/utils/profiling/profile/continuousProfile.tsx
JonasBa: Yup, I reversed it

static/app/components/profiling/flamegraph/continuousFlamegraph.tsx
JonasBa: Nope, no changes from flamegraph.tsx besides the removal of the spanTree, spanChart and passing the txn event to the drawer as opposed to calling useProfiles

Zylphrex (approved):

PR #74139, opened 2024-07-11
feat(discover): Add split decision in meta response

narsaynorath:
Adds the split decision to requests on widgets that already have a split decision made, and ones that are making the decision for the first time. I only tagged errors and transactions specifically, the other cases are ambiguous or we need the user to determine which dataset they want.

### DIFF
--- a/src/sentry/api/endpoints/organization_events.py
+++ b/src/sentry/api/endpoints/organization_events.py
@@ -398,7 +398,20 @@ class OrganizationEventsEndpoint(OrganizationEventsV2EndpointBase):
                     else:
                         split_dataset = discover
 
-                    return _data_fn(split_dataset, offset, limit, scoped_query)
+                    response = _data_fn(split_dataset, offset, limit, scoped_query)
+
+                    if widget.discover_widget_split == DashboardWidgetTypes.ERROR_EVENTS:
+                        response["meta"][
+                            "discoverSplitDecision"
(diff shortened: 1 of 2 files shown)


src/sentry/api/endpoints/organization_events.py
```diff
@@ -398,7 +398,20 @@ def _dashboards_data_fn(scopedDataset, offset, limit, scoped_query, dashboard_wi
...
+                            "discoverSplitDecision"
+                        ] = DashboardWidgetTypes.get_type_name(DashboardWidgetTypes.ERROR_EVENTS)
+                    elif widget.discover_widget_split == DashboardWidgetTypes.TRANSACTION_LIKE:
+                        response["meta"][
+                            "discoverSplitDecision"
+                        ] = DashboardWidgetTypes.get_type_name(
+                            DashboardWidgetTypes.TRANSACTION_LIKE
+                        )
+
+                    return response
```
shruthilayaj, line 414: nit: Would we need to add this if the widget split is already saved on the widget model? In the UI we should prioritize the data from the widget model and if that isn't available, rely on the split decision. IMO the "discoverSplitDecision" meta should only be added if we make a decision by running queries in that request.

shruthilayaj (approved):

src/sentry/api/endpoints/organization_events.py
narsaynorath: Good point, I can remove this then for split widgets, thanks for calling that out!

PR #74143, opened 2024-07-11
feat(breadcrumbs): Add feedback to drawer

leeandher:
Adds the feedback widget to the header of the drawer. Adding it here instead of the base page since if the drawer is open, they've seen the entire new experience prior to writing in.

To allow feedback to be submitted properly, I had to prevent outside clicks from closing the drawer. That said, the close button is sticky now, so users should always be able to quickly close the drawer if they need.

Also fixes a bug where HTTP crumbs would not expand the default depth size of 2 on the base page.

image

**todo**
- [x] Double check that this doesn't act funky with the drawer in getsentry

### DIFF
--- a/static/app/components/events/breadcrumbs/breadcrumbsDataSection.tsx
+++ b/static/app/components/events/breadcrumbs/breadcrumbsDataSection.tsx
@@ -1,4 +1,4 @@
-import {Fragment, useCallback, useMemo} from 'react';
+import {Fragment, useCallback, useMemo, useRef} from 'react';
 import styled from '@emotion/styled';
 
 import ProjectAvatar from 'sentry/components/avatar/projectAvatar';
@@ -18,11 +18,13 @@ import {
   getSummaryBreadcrumbs,
 } from 'sentry/components/events/breadcrumbs/utils';
 import {EventDataSection} from 'sentry/components/events/eventDataSection';
+import useFeedbackWidget from 'sentry/components/feedback/widget/useFeedbackWidget';
 import useDrawer from 'sentry/components/globalDrawer';
 import {
   IconClock,
   IconEllipsis,
   IconFilter,
+  IconMegaphone,
   IconSearch,
   IconSort,
   IconTimer,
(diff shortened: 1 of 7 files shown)


roggenkemper (approved):

static/app/components/events/breadcrumbs/breadcrumbsDataSection.tsx
```diff
@@ -194,6 +199,35 @@ export default function BreadcrumbsDataSection({
   );
 }
 
+function BreadcrumbsFeedback() {
+  const buttonRef = useRef<HTMLButtonElement>(null);
+  const feedback = useFeedbackWidget({
```
malwilley, line 204: @leeandher I actually just added a different way of doing this if you're interested. You can just use `useFeedbackForm()` which will return an `openForm()` function

static/app/components/events/breadcrumbs/breadcrumbsDataSection.tsx
```diff
@@ -194,6 +199,35 @@ export default function BreadcrumbsDataSection({
...
+    buttonRef,
+    messagePlaceholder: t('How can we make breadcrumbs more useful to you?'),
```
malwilley, line 206: We should pass `tags` here to differentiate this feedback from the other feedbacks. I did this [here](https://github.com/getsentry/sentry/pull/74116/files#diff-656d36abacc50a0e0713a3222927e65e2668506bfb6fe21314be7088fc55d582R296) (you may need to rebase to be able to add this prop)

If you follow the same naming convention it would be `feedback_source: 'issue_breadcrumbs'` (or similar)

static/app/components/events/breadcrumbs/breadcrumbsDataSection.tsx
malwilley: > To allow feedback to be submitted properly, I had to prevent outside clicks from closing the drawer.

You won't have to do this actually with `useFeedbackForm()`! I added it because I had the same problem with the search bar.

leeandher: I implemented the new hook (thanks!) but I still had the same problem with outside clicks. The drawer is set to close if a click is registered on any element that isn't a child of the drawer, so since the form is rendered outside it, it still has the same issue. Is there something I missed?

malwilley: Ah yes you're right that it doesn't prevent that. The thing useFeedbackForm avoids is the feedback form being closed when the trigger unmounts.

Something you may want to look at is using `useOverlayAria()` (which we use in useOverlay) for the drawer component. That accepts a `shouldCloseOnInteractOutside` prop which you could use here to prevent feedback interactions from closing the drawer.

PR #74148, opened 2024-07-11
ref(replay): render obj strings properly in StructuredEventData

michellewzhang:

render strings properly with quotations in the `StructuredEventData` component in the network tab.

strings in objects now look like this:
SCR-20240711-iwfj

### DIFF
--- a/static/app/components/structuredEventData/index.tsx
+++ b/static/app/components/structuredEventData/index.tsx
@@ -185,7 +185,7 @@ export function StructuredData({
     );
   }
 
-  if (typeof value === 'string' || config?.isString?.(value)) {
+  if (typeof value === 'string') {
     if (config?.isString?.(value)) {
       const stringValue = config.renderString?.(value) ?? value;
 
--- a/static/app/components/structuredEventData/index.tsx
+++ b/static/app/components/structuredEventData/index.tsx
@@ -185,7 +185,7 @@ export function StructuredData({
     );
   }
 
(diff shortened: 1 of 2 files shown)


static/app/components/structuredEventData/index.tsx
```diff
@@ -185,7 +185,7 @@ export function StructuredData({
     );
   }
 
-  if (typeof value === 'string') {
+  if (typeof value === 'string' || config?.isString?.(value)) {
```
malwilley, line 188: It looks like we make this check on the line below already?

static/app/views/replays/detail/network/details/sections.tsx
```diff
@@ -39,6 +39,12 @@ export type SectionProps = {
   startTimestampMs: number;
 };
 
+const config = {
```
malwilley, line 42: This config makes sense though! I think this is all you need

malwilley: It might make sense for this to be the default for the component and have the event details page pass in the custom config actually. But we can do that later

static/app/components/structuredEventData/index.tsx
michellewzhang: ahh you're right, i'll remove

malwilley (approved):

billyvg (approved):
Oooh nice

PR #74150, opened 2024-07-11
ref(similarity): Change flags to be be internal instead of options

jangjodi:
Change similarity flags to be internal instead of using options
Project list was too long to be applied in options automator

Requires https://github.com/getsentry/getsentry/pull/14597


vartec (approved):
FWIW, now that's its fully launched, you can also use Flagpole for this.

mrduncan:
> FWIW, now that's its fully launched, you can also use Flagpole for this.

Unfortunately this is one case where we can't - the list of opted-in projects is _massive_

PR #74153, opened 2024-07-11
chore(replay): Frontend timeline gap feature flag

c298lee:
Backend ff: https://github.com/getsentry/sentry/pull/74140


michellewzhang (approved):
not related to the ff but could we make the ReplayTimelineEvents component name more specific if it's just showing the gaps?

c298lee:
> not related to the ff but could we make the ReplayTimelineEvents component name more specific if it's just showing the gaps?

omg I put the flag in the wrong spot let me fix it

michellewzhang (changes requested):
wait nvm i think you did the ff check on the wrong component! should be on the gaps component up above

michellewzhang (approved):
WOO

PR #74155, opened 2024-07-11
chore(fe): Surface Id in Project/Org Settings

iamrajjoshi:
After we started supporting Ids in our APIs, we would like to make a similar change in our CLI. For our users to actually use CLI, we should surface the ids for their org and projects more prominently. 

Here I add it to their settings page. 
image
image
image

### DIFF
--- a/static/app/views/settings/organizationGeneralSettings/organizationSettingsForm.tsx
+++ b/static/app/views/settings/organizationGeneralSettings/organizationSettingsForm.tsx
@@ -57,7 +57,7 @@ function OrganizationSettingsForm({initialData, onSave}: Props) {
       setValue(_, _name) {
         return organization.id;
       },
-      help: `This is the organization's unique identifier. It cannot be changed.`,
+      help: `The unique identifier for this organization. It cannot be modified.`,
     };
 
     formsConfig[0].fields = [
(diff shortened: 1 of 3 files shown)


static/app/views/settings/organizationGeneralSettings/organizationSettingsForm.tsx
```diff
@@ -48,8 +49,21 @@ function OrganizationSettingsForm({initialData, onSave}: Props) {
...
     const formsConfig = cloneDeep(organizationSettingsFields);
+    const organizationIdField: FieldObject = {
+      name: 'organizationId',
+      type: 'string',
+      disabled: true,
+      label: t('Organization ID'),
+      setValue(_, _name) {
+        return organization.id;
+      },
+      help: `This is the organization's unique identifier. It cannot be changed.`,
```
schew2381, line 60: this is more subjective, but I think this matches the other settings text slightly better
```suggestion
      help: `The unique identifier for this organization. It cannot be modified.`,
```

iamrajjoshi: good call, i updated the project id description as well

schew2381:
What about for teams 🥺

schew2381:
what about for everything else with a slug 🥺

iamrajjoshi:
> what about for everything else with a slug 🥺

i want to prioritize the slugs that will be used in the CLI for now, which is project and org

schew2381 (approved):
alright fine

PR #74160, opened 2024-07-11
feat(query-builder): Add analytics for autocomplete actions

malwilley:
Adds the rest of the analytics events to match what we log in SmartSearchBar (plus a couple extra):

- When a filter key is autocompleted
- When an operator is changed in the dropdown
- When a filter value is autocompleted
- When a filter value is manually entered

### DIFF
--- a/static/app/components/searchQueryBuilder/filterKeyOperator.tsx
+++ b/static/app/components/searchQueryBuilder/filterKeyOperator.tsx
@@ -21,6 +21,8 @@ import {
 } from 'sentry/components/searchSyntax/parser';
 import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
+import {trackAnalytics} from 'sentry/utils/analytics';
+import useOrganization from 'sentry/utils/useOrganization';
 
 type FilterOperatorProps = {
   item: Node<ParseResultToken>;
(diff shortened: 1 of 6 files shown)


roggenkemper (approved):

MichaelSun48 (approved):

static/app/components/searchQueryBuilder/filterKeyOperator.tsx
```diff
@@ -206,6 +209,15 @@ export function FilterKeyOperator({token, state, item}: FilterOperatorProps) {
       options={options}
       value={operator}
       onChange={option => {
+        trackAnalytics('search.operator_autocompleted', {
+          organization,
+          query,
+          search_type: savedSearchType === 0 ? 'issues' : 'events',
```
mrduncan, line 215: Thoughts on a function to convert the enum to a display string? If other product areas adopt this `savedSearchType` will probably end up not being 0 or 1 and it seems probable we'll miss updating at least one of these

malwilley: You make a good point - I pulled this logic over from SmartSearchBar without really thinking about it, but this isn't the right way to be logging this with all the new use cases. I'll follow up with that function

PR #74165, opened 2024-07-11
ref(metric-extraction): Increase timeout for experimental config

k-fish:
### Summary
The only one going over is widget specs, looks to be by a small amount (1-2 seconds), but it's been creeping up over the last few weeks.

We need a bit of time to investigate, but in the meantime this should lower any trouble caused by timeouts.


Zylphrex (approved):
👍 I assume once the cause if identified, you'll look into lowering this again?

k-fish:
@Zylphrex yeah, we previously got this down to stop firing timeouts entirely, but it's crept up. A fix is coming but it'll take me a little bit.

PR #74166, opened 2024-07-11
feat(discover): Surface split decision through events-stats meta

narsaynorath:
Surfaces the split decision through the endpoint response meta for dashboard queries.

### DIFF
--- a/src/sentry/api/endpoints/organization_events_stats.py
+++ b/src/sentry/api/endpoints/organization_events_stats.py
@@ -435,16 +435,28 @@ class OrganizationEventsStatsEndpoint(OrganizationEventsV2EndpointBase):
                             comparison_delta,
                         )
                     elif decision == DashboardWidgetTypes.TRANSACTION_LIKE:
-                        original_results.data.get("meta")[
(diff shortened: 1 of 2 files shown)


mjq (approved):

shruthilayaj (approved):
looks good, though should we also update the dataset it queries?
ie instead of `transactions_only_query = f"({query}) AND event.type:transaction"`, query the transaction dataset etc?

narsaynorath:
@shruthilayaj I'll put up another PR for that! I was just keeping this for adding the meta value. Would you mind taking another quick glance? I had to fix a typing issue to handle top-events queries. My changes prior only handled one SnubaTS result.

shruthilayaj:
@narsaynorath let's make sure the test cases cover both result types?

narsaynorath:
@shruthilayaj thanks for calling that out. I put up the code change too quickly and assumed top events had the same access path. Turns out, `meta` might not exist so my code was incorrect. I've fixed it and added tests in the latest commit: 1769c52bfcbda9ad24ddebf58355a19d19006532

PR #74167, opened 2024-07-11
fix(insights): minor bugs/improvements to empty states

DominikB2014:
1. Don't show extra filters on modules if there is no module data to begin with. For example on the assets module, we won't show the domain filter. This makes sense because these filters will not change anything in this context anyways.
2. Use correct docs links from each modules `settings.ts`
3. Remove duplicated `read docs`
4. Don't show no query results banner in query module when no data has been sent to begin with.


gggritso (approved):
👍🏻

PR #74171, opened 2024-07-11
feat(profiling) construct profile from continuous chunk

JonasBa:
This PR implements measurements rendering for the continuous profiling UI.

The chart to render slow and frozen frames still needs to be updated to reflect this.


static/app/utils/profiling/flamegraphChart.tsx
```diff
@@ -115,14 +118,14 @@ export class FlamegraphChart {
         }
 
         // Track and update X domain max and min
-        if (m.elapsed_since_start_ns > this.domains.x[1]) {
-          this.domains.x[1] = m.elapsed_since_start_ns;
+        if (m.elapsed > this.domains.x[1]) {
+          this.domains.x[1] = m.elapsed;
```
JonasBa, line 122: this had previously shared the type with the measurement, but was now moved to a separate field

static/app/components/profiling/flamegraph/continuousFlamegraph.tsx
```diff
@@ -229,7 +232,11 @@ export function ContinuousFlamegraph(): ReactElement {
     }
     return new UIFrames(
       {
+        // @TODO
+        // @ts-expect-error
```
JonasBa, line 236: these need to be updated, I'm going to tackle them in the next PR

static/app/components/profiling/flamegraph/continuousFlamegraph.tsx
```diff
@@ -251,25 +258,31 @@ export function ContinuousFlamegraph(): ReactElement {
...
-            };
-          }),
-          // some versions of cocoa send byte so we need to correct it to watt
-          unit: 'watt',
-          name: 'CPU energy usage',
-        });
+        const measurements = profileGroup.measurements[key]!;
+        const values: ProfileSeriesMeasurement['values'] = [];
+
+        let offset = 0;
```
JonasBa, line 264: This computes the value offset to the start of the profile

Zylphrex: Might be good to put into a helper function that takes a formatted for the value? I see it's copied for all the other charts below.

Zylphrex (approved):

static/app/components/profiling/flamegraph/continuousFlamegraph.tsx
```diff
@@ -123,7 +126,7 @@ export function ContinuousFlamegraph(): ReactElement {
   const dispatch = useDispatchFlamegraphState();
 
   const profiles = useContinuousProfile();
-  const profileGroup = useProfileGroup();
+  const profileGroup = useProfileGroup() as ContinuousProfileGroup;
```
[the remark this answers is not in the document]
JonasBa: Let me tackle this in a separate PR with the approach we discussed (separate context and asserting data shape). We can get rid of calling useProfiles there too

PR #74174, opened 2024-07-11
feat(backup): Enable relocation hiding

azaslavsky:
This PR adds a new flag to `sentry import`, called `hide_organizations`, which places organizations in a special "hidden" mode after an import, rendering them invisible to the UI (org dropdowns, etc).

### DIFF
--- a/src/sentry/backup/helpers.py
+++ b/src/sentry/backup/helpers.py
@@ -90,7 +90,7 @@ class ImportFlags(NamedTuple):
     """
 
     # If a username already exists, should we re-use that user, or create a new one with a randomly
-    # suffixed username (ex: "some-user" would become "some-user-ad21")
+    # suffixed username (ex: "some-user" would become "some-user-ad21")?
     merge_users: bool = False
 
     # If a global configuration value `ControlOption`/`Option` (as identified by its unique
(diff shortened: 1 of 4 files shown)


src/sentry/receivers/outbox/control.py
```diff
@@ -127,3 +130,47 @@ def process_issue_email_reply(shard_identifier: int, payload: Any, **kwds):
...
         text=payload["text"],
     )
+
+
+# See the comment on /src/sentry/tasks/relocation.py::uploading_start for a detailed description of
+# how this outbox drain handler fits into the entire SAAS->SAAS relocation workflow.
+@receiver(process_control_outbox, sender=OutboxCategory.RELOCATION_EXPORT_REQUEST)
+def process_relocation_request_new_export(payload: Mapping[str, Any], **kwds):
+    encrypt_with_public_key = (
+        payload["encrypt_with_public_key"].encode("utf-8")
```
hubertdeng123, line 140: What would be the reason we expect other types of payloads besides strings?

hubertdeng123 (approved):

src/sentry/receivers/outbox/control.py
azaslavsky: The hybrid cloud (or pydantic?) machinery seems to serialize them to strings when sending them over the wire. Not sure why this happens, but this seems to resolve the issue.

PR #74178, opened 2024-07-11
nit(gitlab): quiet invalid token sentry error

cathteng:
This issue is not actionable on our side https://sentry.sentry.io/issues/3622378978/.

If Gitlab's webhook secret does not match, the user needs to refresh token (or re-install the integration) by following this https://docs.sentry.io/product/integrations/integration-platform/public-integration/#refreshing-tokens.

### DIFF
--- a/src/sentry/integrations/gitlab/webhooks.py
+++ b/src/sentry/integrations/gitlab/webhooks.py
@@ -31,7 +31,7 @@ from sentry.plugins.providers import IntegrationRepositoryProvider
 logger = logging.getLogger("sentry.webhooks")
 
 PROVIDER_NAME = "integrations:gitlab"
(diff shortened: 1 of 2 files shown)


iamrajjoshi (approved):

src/sentry/integrations/gitlab/webhooks.py
```diff
@@ -296,19 +297,12 @@ def post(self, request: Request) -> HttpResponse:
             },
         }
 
-        try:
-            if not constant_time_compare(secret, integration.metadata["webhook_secret"]):
-                # Summary and potential workaround mentioned here:
-                # https://github.com/getsentry/sentry/issues/34903#issuecomment-1262754478
-                # This forces a stack trace to be produced
-                raise Exception("The webhook secrets do not match.")
```
cathteng, line 304: the stack trace has a single frame

PR #74181, opened 2024-07-11
feat(profiling): Pass timestamps to profile chunks

Zylphrex:
We need to drop samples outside the selected timestamps so pass the start/end timestamps to vroom so we can do this processing.


phacops (approved):

wmak (approved):
lgtm

PR #74183, opened 2024-07-11
feat(uptime): Guard against processing duplicate results in the result consumer

wedamija:
We only want to process each result one time. Since we guarantee receiving results in order from the uptime checker we can just use the scheduled time of each result to maintain a watermark of the last processed result.

### DIFF
--- a/src/sentry/uptime/consumers/results_consumer.py
+++ b/src/sentry/uptime/consumers/results_consumer.py
@@ -19,7 +19,7 @@ from sentry.utils import metrics
 from sentry.utils.hashlib import md5_text
 
 logger = logging.getLogger(__name__)
-REDIS_TTL = timedelta(days=7)
+LAST_UPDATE_REDIS_TTL = timedelta(days=7)
 
 
 def build_last_update_key(project_subscription: ProjectUptimeSubscription) -> str:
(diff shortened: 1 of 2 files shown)


src/sentry/uptime/consumers/results_consumer.py
```diff
@@ -11,10 +12,18 @@
...
     ResultsStrategyFactory,
 )
+from sentry.uptime.detectors.ranking import _get_cluster
 from sentry.uptime.issue_platform import create_issue_platform_occurrence
 from sentry.uptime.models import ProjectUptimeSubscription, UptimeSubscription
+from sentry.utils import metrics
+from sentry.utils.hashlib import md5_text
 
 logger = logging.getLogger(__name__)
+REDIS_TTL = timedelta(days=7)
```
evanpurkhiser, line 22: Should we give this a more descriptive name?

evanpurkhiser, line 22:
```suggestion
LAST_UPDATE_REDIS_TTL = timedelta(days=7)
```

src/sentry/uptime/consumers/results_consumer.py
```diff
@@ -41,11 +51,42 @@ def handle_result(self, subscription: UptimeSubscription, result: CheckResult):
...
+        self,
+        project_subscription: ProjectUptimeSubscription,
+        result: CheckResult,
+        last_update_ms: int,
+    ):
+        cluster = _get_cluster()
+        try:
+            if result["scheduled_check_time_ms"] <= last_update_ms:
+                # If the scheduled check time is older than the most recent update then we've already processed it.
+                # Skip and log
```
evanpurkhiser, line 75: Can we explain _why_ we might have duplicates here?

evanpurkhiser (approved):

PR #74185, opened 2024-07-11
chore(slack): Cleanup Tests & Code for Culprit Block

iamrajjoshi:
We GA'ed culprit blocks in our slack notifications, so i am cleaning up tests and code here.

### DIFF
--- a/tests/sentry/integrations/slack/notifications/test_issue_alert.py
+++ b/tests/sentry/integrations/slack/notifications/test_issue_alert.py
@@ -104,7 +104,7 @@ class SlackIssueAlertNotificationTest(SlackActivityNotificationTest, Performance
         return_value=TEST_PERF_ISSUE_OCCURRENCE,
         new_callable=mock.PropertyMock,
     )
-    def test_performance_issue_alert_user_block_with_culprit_blocks(self, occurrence):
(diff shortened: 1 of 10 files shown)


tests/sentry/integrations/slack/notifications/test_issue_alert.py
```diff
@@ -389,106 +349,7 @@ def test_issue_alert_issue_owners_environment_block(self):
...
-        )
-        assert blocks[5]["elements"][0]["text"] == f"Suggested Assignees: #{self.team.slug}"
-        assert (
-            blocks[6]["elements"][0]["text"]
-            == f"{event.project.slug} | <http://testserver/settings/{event.organization.slug}/teams/{self.team.slug}/notifications/?referrer=issue_alert-slack-team&notification_uuid={notification_uuid}|Notification Settings>"
-        )
-
     @responses.activate
-    @with_feature("organizations:slack-culprit-blocks")
     def test_issue_alert_team_issue_owners_block_with_culprit_blocks(self):
```
cathteng, line 353: nit: if having culprit blocks is now the default, can we keep the name as `test_issue_alert_team_issue_owners_block`, etc

cathteng (approved):
nice

PR #74190, opened 2024-07-12
feat(metrics): numeric span attribute extraction

### LINKED ISSUE #73866: Extraction of tags with numeric values
```[tasklist]
- [ ] https://github.com/getsentry/sentry/pull/74118
- [ ] extend numeric conditions
```

In the example below, tags `temp` and `voltage` that always have numeric values are never available, while tag `label` whose value is always a string is normally ingested
```
{
   "category":"span",
   "mri":"c:custom/span_attribute_24@none",
   "field":null,
   "tags":[
      {
         "key":"environment",
         "field":"span.sentry_tags.environment"
      },
      {
         "key":"label",
         "field":"span.data.label"
      },
      {
         "key":"release",
         "field":"span.sentry_tags.release"
      },
      {
         "key":"span.description",
         "field":"span.description"
      },
      {
         "key":"temp",
         "field":"span.data.temp"
      },
      {
         "key":"voltage",
         "field":"span.data.voltage"
      }
   ],
   "condition":{
      "op":"not",
      "inner":{
         "name":"span.sentry_tags.sdk.name",
         "op":"eq",
         "value":null
      }
   }
}
```

related sentry error https://sentry.sentry.io/issues/5274121069/events/b7bbbcd5916b4cb4a2f57af3cefbc98a/

obostjancic:
- closes #73866 

Creating an extraction rule with a condition whose RHS is a numeric value or a sequence of numeric values will result in following conditions:
- `foo:123` -> `{"op": "eq", "name": "span.data.foo", "value": "123"}` 
- `responseStatus:[400,500]` -> `{"op": "eq", "name": "span.data.responseStatus", "value": ["200", "400"]}`

that will not match a span with an attribute `foo` and value `123` becuase `123 != "123"`

This PR fixes that by detecting such conditions and extending them such that  `foo:123` first gets converted to `foo:123 or foo:"123"` and then to a relay condition


src/sentry/snuba/metrics/span_attribute_extraction.py
```diff
@@ -126,26 +126,86 @@ def _flatten_query_tokens(conditions: Sequence[QueryToken]) -> list[SearchFilter
...
+    expands the original token into a paren expression if the value is a numeric string.
+
+    Example:
+    `key:'123'` -> `key:'123' OR key:123`
+    `key:['123', '456']` -> `key:['123', '456'] OR key:[123, 456]`
+
+    """
+    if token.operator == "=" or token.operator == "!=":
+        if not str(token.value.value).isdigit():
+            return token
```
shellmayr, line 166: .isdigit() only returns true if all characters are numeric literals - floats do not fall into this category and are therefore returned as-is - is this on purpose?

src/sentry/snuba/metrics/span_attribute_extraction.py
```diff
@@ -126,26 +126,86 @@ def _flatten_query_tokens(conditions: Sequence[QueryToken]) -> list[SearchFilter
...
+            "OR",
+            numeric_value_token,
+        ]
+    )
 
-    non_empty_conditions = [condition for condition in conditions if condition]
 
-    search_query = " or ".join([f"({condition})" for condition in non_empty_conditions])
-    return event_search.parse_search_query(search_query)
+# Conversion to RuleCondition
```
shellmayr, line 194: Nit: indentation of this comment is off - where does it belong?

shellmayr (approved):
LGTM - small caveat that this only pertains to ints, but maybe we only have that limitation there

src/sentry/snuba/metrics/span_attribute_extraction.py
```diff
@@ -126,26 +126,86 @@ def _flatten_query_tokens(conditions: Sequence[QueryToken]) -> list[SearchFilter
...
+        if not all(value.isdigit() for value in str_values):
+            return token
+
+        numeric_values = [int(value) for value in str_values]
+        numeric_value_token = SearchFilter(
+            key=token.key, operator=token.operator, value=SearchValue(numeric_values)
+        )
+
+    if not numeric_value_token:
+        return token
```
vgrozdanic, line 183: What does this check do? 
Unless I am missing on something, if we enter in one of the if/elif branches, this check will always be true - `numeric_value_token` will be a reference to a `SearchFilter` object.

If non of the conditions in if/elif are true (e.g. operator somehow ends being  '>' or smth similar), then an exception will be raised because we have tried to access variable that wasn't previously defined

vgrozdanic (approved):

Issue #73929, opened 2024-07-08
SPIKE Move virtual MRI resolution to backend API?


matejminar:
**What are we trying to solve?**

The settings UI for span based metrics requires us to branch out into multiple virtual metrics that are then combined in the front-end, such that we can offer a nice experience to the user, where they only select the span attribute and query, and can then choose from the configured aggregations. The metrics have an MRI in the shape of {type}:custom/span_attribute{numeric_id}@{type}, which of course contains no information about the contents, so there needs to be some inference between the span_attribute_{id} metrics, and the configuration object containing the associated queries and ids.

**What is the current implementation?**

Currently, the UI sends two requests to the backend: 
1. the metrics/meta call returns the plain metrics with the MRI {type}:custom/span_attribute{numeric_id}@{type}, returning a list of metrics meta objects
2. a call to fetch the configuration objects for span based metrics, returning a list of span based metric config objects
These two objects are then combined in the front-end into the correct representation.

**What are our options?**

Ideally, the front-end could call the backend and receive just the information required for display, but there is some complexity to this. 

There are now two different types of metrics that we want to still display: 

1. Custom Metrics
2. Span-based Metrics

The two types require a different return format, as the span-based metrics and custom metrics both need to be considered. 

1. Have two different API endpoints that each only return a specific set of objects (only old style & only new style each)
2. Include the span-based metrics into the metrics/meta and adapt the format to include additional info about span-based metrics

**Proposal**
Add a field to metrics/meta that allows for arbitrary custom metadata to be added. First use-case for this is to add metadata for span-based metric metadata, such that the front-end does not have to stitch together multiple endpoint responses on the fly, but this would be adaptable in case we have other types of metrics with different metadata as well.

In effect, the meta object would be extended by a sub-object with arbitrary keys based on the type of metadata added, whose existence can be checked on the frontend:
```
{
    "name": "count_clicks",
    "type": "generic_distribution",
    "operations": [
        "p50",
        "p75",
        "p90",
        "p95",
        "p99"
    ],
    "unit": "none",
    "mri": "d:custom/span_attribute_1@none",
    "projectIds": [
        4554336753811458
    ],
    "blockingStatus": null,
    "customMeta": {
        "spanBasedMeta": {
            "associatedSpanAttribute": "count_clicks",
            "associatedProjectId": 4554336753811458,
            "virtualMRI": "v:custom/count_clicks|4554336753811458@none"
        }
    }
}
```

[...]

PR #74194, opened 2024-07-12
feat(metrics): allow custom metadata to be added to metrics meta

shellmayr:
Contributes to https://github.com/getsentry/sentry/issues/73929

TLDR: add a field to metrics/meta that allows for arbitrary custom metadata to be added. First use-case for this is to add metadata for span-based metric metadata, such that the front-end does not have to stitch together multiple endpoint responses on the fly.

In effect, the meta object would be extended by a sub-object with arbitrary keys based on the type of metadata added, whose existence can be checked on the frontend: 
```json
{
    "name": "count_clicks",
    "type": "generic_distribution",
    "operations": [
        "p50",
        "p75",
        "p90",
        "p95",
        "p99"
    ],
    "unit": "none",
    "mri": "d:custom/span_attribute_1@none",
    "projectIds": [
        4554336753811458
    ],
    "blockingStatus": null,
    "customMeta": {
        "spanBasedMeta": {
            "associatedSpanAttribute": "count_clicks",
            "associatedProjectId": 4554336753811458,
            "virtualMRI": "v:custom/count_clicks|4554336753811458@none"
        }
    }
}
```


vgrozdanic (approved):

shellmayr:
Waiting to merge this to align with @ArthurKnaus on whether this makes sense from a front-end perspective.

ArthurKnaus:
Hmm, with this format I still need to stitch it together as I do not have any information about the user configured aggregates and conditions, right?

shellmayr:
Closing: currently not enough certainty around this - we will continue iterating on the experience and implement this when there is sufficient stability around functionality.

PR #74195, opened 2024-07-12
feat(relay): add span-normalization.allowed-hosts into relay config

aldy505:
Complementary relay config for https://github.com/getsentry/relay/pull/3813, based on the discussion on https://github.com/getsentry/relay/issues/3572

### Legal Boilerplate


jjbayer (commented):
We also need to register the option, see for example

https://github.com/getsentry/sentry/blob/23ccc70817cc6dbc43230ddd86c2f6329ecb34ab/src/sentry/options/defaults.py#L2347-L2354

jjbayer (approved):

aldy505:
Hey @jjbayer is there anything else that's blocking this PR? Since not all checks are green.

PR #74200, opened 2024-07-12
feat(features) Allow features to be excluded from API results

markstory:
We have many features that are not used in the UI code. Those features could be omitted from API responses to save time checking the features, and save response bloat.

### DIFF
--- a/src/sentry/features/manager.py
+++ b/src/sentry/features/manager.py
@@ -146,7 +146,7 @@ class FeatureManager(RegisteredFeatureManager):
         self._entity_handler: FeatureHandler | None = None
 
     def all(
-        self, feature_type: type[Feature] = Feature, api_expose: bool = True
+        self, feature_type: type[Feature] = Feature, api_expose: bool = False
     ) -> Mapping[str, type[Feature]]:
         """
         Get a mapping of feature name -> feature class, optionally specific to a
@@ -159,10 +159,7 @@ class FeatureManager(RegisteredFeatureManager):
             name: feature
             for name, feature in self._feature_registry.items()
             if issubclass(feature, feature_type)
-            and (
-                (api_expose and name in self.exposed_features)
-                or (not api_expose and name not in self.exposed_features)
-            )
+            and (not api_expose or name in self.exposed_features)
         }
 
     def add(
--- a/src/sentry/features/manager.py
+++ b/src/sentry/features/manager.py
@@ -138,17 +138,33 @@ class FeatureManager(RegisteredFeatureManager):
     def __init__(self) -> None:
         super().__init__()
         self._feature_registry: MutableMapping[str, type[Feature]] = {}
+        # Deprecated: Remove entity_features once flagr has been removed.
         self.entity_features: MutableSet[str] = set()
+        self.exposed_features: MutableSet[str] = set()
         self.option_features: MutableSet[str] = set()
         self.flagpole_features: MutableSet[str] = set()
         self._entity_handler: FeatureHandler | None = None
 
-    def all(self, feature_type: type[Feature] = Feature) -> Mapping[str, type[Feature]]:
+    def all(
+        self, feature_type: type[Feature] = Feature, api_expose: bool | None = None
+    ) -> Mapping[str, type[Feature]]:
         """
         Get a mapping of feature name -> feature class, optionally specific to a
         particular feature type.
+
+        :param feature_type: The feature class you want to filter by. eg. (OrganizationFeature | ProjectFeature | SystemFeature)
+        :param api_expose: Only include features that were registered with `api_expose`.
         """
-        return {k: v for k, v in self._feature_registry.items() if issubclass(v, feature_type)}
+        output: dict[str, type[Feature]] = {}
+        for name, feature in self._feature_registry.items():
+            match = False
+            if issubclass(feature, feature_type):
+                match = True
+            if match and api_expose and name not in self.exposed_features:
+                match = False
+            if match:
+                output[name] = feature
+        return output
 
     def add(
         self,
@@ -156,6 +172,7 @@ class FeatureManager(RegisteredFeatureManager):
         cls: type[Feature] = Feature,
         entity_feature_strategy: bool | FeatureHandlerStrategy = False,
         default: bool = False,
+        api_expose: bool = True,
     ) -> None:
         """
         Register a feature.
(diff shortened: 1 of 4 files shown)


src/sentry/features/temporary.py
```diff
@@ -61,7 +61,7 @@ def register_temporary_features(manager: FeatureManager):
     # Rollout of the new API rate limits for organization events
     manager.add("organizations:api-organization_events-rate-limit-reduced-rollout", OrganizationFeature, FeatureHandlerStrategy.INTERNAL)
     # Enables the cron job to auto-enable codecov integrations.
-    manager.add("organizations:auto-enable-codecov", OrganizationFeature, FeatureHandlerStrategy.INTERNAL)
+    manager.add("organizations:auto-enable-codecov", OrganizationFeature, FeatureHandlerStrategy.INTERNAL, api_expose=False)
```
markstory, line 64: I picked a couple flags to illustrate how this parameter could be used. I can add this option to the remaining features in a follow up pull request if folks like this approach.

src/sentry/features/manager.py
```diff
@@ -138,24 +138,41 @@ class FeatureManager(RegisteredFeatureManager):
...
+                output[name] = feature
+        return output
 
     def add(
         self,
         name: str,
         cls: type[Feature] = Feature,
         entity_feature_strategy: bool | FeatureHandlerStrategy = False,
         default: bool = False,
+        api_expose: bool = True,
```
wedamija, line 175: I think we'd be better off defaulting `api_expose=False`  and explicitly adding all existing features to have `api_expose=True`. That way we enforce people being explicit about exposing their features in the future.

src/sentry/features/manager.py
```diff
@@ -138,24 +138,41 @@ class FeatureManager(RegisteredFeatureManager):
...
-        return {k: v for k, v in self._feature_registry.items() if issubclass(v, feature_type)}
+        output: dict[str, type[Feature]] = {}
+        for name, feature in self._feature_registry.items():
+            match = False
+            if issubclass(feature, feature_type):
+                match = True
+            if match and api_expose and name not in self.exposed_features:
+                match = False
+            if match:
+                output[name] = feature
```
wedamija, line 166: Nit: `{k: v for k, v in self._feature_registry.items() if issubclass(v, feature_type) and (not api_expose or name in self.exposed_features)}`?

src/sentry/features/manager.py
```diff
@@ -138,24 +138,41 @@ class FeatureManager(RegisteredFeatureManager):
...
+        # Deprecated: Remove entity_features once flagr has been removed.
         self.entity_features: MutableSet[str] = set()
+        self.exposed_features: MutableSet[str] = set()
         self.option_features: MutableSet[str] = set()
         self.flagpole_features: MutableSet[str] = set()
         self._entity_handler: FeatureHandler | None = None
 
-    def all(self, feature_type: type[Feature] = Feature) -> Mapping[str, type[Feature]]:
+    def all(
+        self, feature_type: type[Feature] = Feature, api_expose: bool | None = None
```
wedamija, line 149: Nit: Should this be `api_expose: bool = False`?

markstory: I didn't want to change the default behavior just yet. Getting getsentry and sentry aligned is going to take a few pull requests.

src/sentry/features/manager.py
markstory: I would like to get to that point as well. Co-ordinating a change in the default value with features defined in getsentry is tricky though.

My thinking was to:

1. Introduce `api_expose` as True. This maintains the existing behavior
2. Update all feature flags that want to be exposed in both sentry & getsentry to have `api_expose=True`
3. change the default value of `api_expose` to `False`.

This sequencing takes a bit more time, but should avoid breaking compatibility.

wedamija: That makes sense to me!

src/sentry/features/manager.py
wedamija: I think defaulting it to either `None` or `False` ends up giving the same result though?

wedamija (approved):

dashed (approved):

src/sentry/features/manager.py
```diff
@@ -138,24 +138,40 @@ class FeatureManager(RegisteredFeatureManager):
...
+        # Deprecated: Remove entity_features once flagr has been removed.
         self.entity_features: MutableSet[str] = set()
+        self.exposed_features: MutableSet[str] = set()
         self.option_features: MutableSet[str] = set()
         self.flagpole_features: MutableSet[str] = set()
         self._entity_handler: FeatureHandler | None = None
 
-    def all(self, feature_type: type[Feature] = Feature) -> Mapping[str, type[Feature]]:
+    def all(
+        self, feature_type: type[Feature] = Feature, api_expose: bool = True
```
wedamija, line 149: Don't we want to default to `False` here?

src/sentry/features/manager.py
```diff
@@ -138,24 +138,40 @@ class FeatureManager(RegisteredFeatureManager):
...
         """
-        return {k: v for k, v in self._feature_registry.items() if issubclass(v, feature_type)}
+        return {
+            name: feature
+            for name, feature in self._feature_registry.items()
+            if issubclass(feature, feature_type)
+            and (
+                (api_expose and name in self.exposed_features)
+                or (not api_expose and name not in self.exposed_features)
+            )
```
wedamija, line 165: Oh I might have misunderstood the point of `api_expose` here. I thought we wanted to return only api expose when true, and everything when false. In that case I think the `None` makes sense for all?

markstory: I had been thinking about as a filter condition that had an on/off/undefined state. However, we don't really have any scenarios right now where we need all the features that are hidden from the API responses 🤔

src/sentry/features/manager.py
```diff
@@ -138,24 +138,37 @@ class FeatureManager(RegisteredFeatureManager):
...
+        # Deprecated: Remove entity_features once flagr has been removed.
         self.entity_features: MutableSet[str] = set()
+        self.exposed_features: MutableSet[str] = set()
         self.option_features: MutableSet[str] = set()
         self.flagpole_features: MutableSet[str] = set()
         self._entity_handler: FeatureHandler | None = None
 
-    def all(self, feature_type: type[Feature] = Feature) -> Mapping[str, type[Feature]]:
+    def all(
+        self, feature_type: type[Feature] = Feature, api_expose: bool = False
```
wedamija, line 149: Super nit: `api_expose_only` to clarify between that case you mentioned before

wedamija (approved):

PR #74203, opened 2024-07-12
feat(grouping): Method to upgrade deprecated configs

armenzg:
This will allow upgrading projects with a grouping configuration marked as deprecated. The project will also get the auto updates config set to `True` to prevent them from falling behind.

This [PR](https://github.com/getsentry/getsentry/pull/14616) is to run the upgrades.

### DIFF
--- a/src/sentry/grouping/ingest/config.py
+++ b/src/sentry/grouping/ingest/config.py
@@ -38,9 +38,9 @@ def _project_should_update_grouping(project: Project) -> bool:
     should_update_org = (
         project.organization_id % 1000 < float(settings.SENTRY_GROUPING_AUTO_UPDATE_ENABLED) * 1000
     )
-    deprecated_config = project.get_option("sentry:grouping_config") in CONFIGS_TO_DEPRECATE
+    is_deprecated_config = project.get_option("sentry:grouping_config") in CONFIGS_TO_DEPRECATE
     auto_update = bool(project.get_option("sentry:grouping_auto_update"))
-    return deprecated_config or (auto_update and should_update_org)
+    return is_deprecated_config or (auto_update and should_update_org)
 
 
 def _config_update_happened_recently(project: Project, tolerance: int) -> bool:
--- a/src/sentry/grouping/ingest/config.py
+++ b/src/sentry/grouping/ingest/config.py
@@ -18,7 +18,7 @@ logger = logging.getLogger("sentry.events.grouping")
 Job = MutableMapping[str, Any]
 
 # We are moving all projects off these configuration without waiting for events
-CONFIGS_TO_DEPRECATE = []
+CONFIGS_TO_DEPRECATE: list[str] = []
 
 
 def upgrade_deprecated_configs():
--- a/src/sentry/grouping/ingest/config.py
+++ b/src/sentry/grouping/ingest/config.py
@@ -85,7 +85,7 @@ def _auto_update_grouping(project: Project) -> None:
             "sentry:grouping_config": new_config,
         }
         # Any project on deprecated configs may be there only because auto updates are
-        # disabled (not because they want to use that specific config). This will reduce the 
+        # disabled (not because they want to use that specific config). This will reduce the
         # chance of that happening unintentionally.
         if current_config in CONFIGS_TO_DEPRECATE:
             changes["sentry:grouping_auto_update"] = True
--- a/src/sentry/grouping/ingest/config.py
+++ b/src/sentry/grouping/ingest/config.py
@@ -84,8 +84,9 @@ def _auto_update_grouping(project: Project) -> None:
             "sentry:secondary_grouping_expiry": expiry,
             "sentry:grouping_config": new_config,
         }
-        # Any project on deprecated configs may have disabled auto updates and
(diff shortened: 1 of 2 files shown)


src/sentry/grouping/ingest/config.py
```diff
@@ -17,6 +17,17 @@
 
 Job = MutableMapping[str, Any]
 
+# We are moving all projects off these configuration without waiting for events
+CONFIGS_TO_DEPRECATE = []
```
armenzg, line 21: When we're ready, we will add the mobile config here.

src/sentry/grouping/ingest/config.py
```diff
@@ -17,6 +17,17 @@
...
+
+
+def upgrade_deprecated_configs():
```
armenzg, line 24: Later I will be adding a script that calls this code.

src/sentry/grouping/ingest/config.py
```diff
@@ -17,6 +17,17 @@
...
+    queryset = Project.objects.filter(
+        projectoption__key="sentry:grouping_config", projectoption__value__in=CONFIGS_TO_DEPRECATE
+    )
+    for project in queryset:
+        _auto_update_grouping(project)
```
armenzg, line 29: This function is being modified to notice that the project is on a deprecated config, thus, upgrading it.

src/sentry/grouping/ingest/config.py
```diff
@@ -27,7 +38,9 @@ def _project_should_update_grouping(project: Project) -> bool:
     should_update_org = (
         project.organization_id % 1000 < float(settings.SENTRY_GROUPING_AUTO_UPDATE_ENABLED) * 1000
     )
-    return bool(project.get_option("sentry:grouping_auto_update")) and should_update_org
+    deprecated_config = project.get_option("sentry:grouping_config") in CONFIGS_TO_DEPRECATE
```
armenzg, line 41: This code is reached when we save an error event, thus, we will also upgrade during ingestion:
https://github.com/getsentry/sentry/blob/233ccd6aa3c807f6a4085ded5c191f761f7485fd/src/sentry/event_manager.py#L1400

src/sentry/grouping/ingest/config.py
```diff
@@ -27,7 +38,9 @@ def _project_should_update_grouping(project: Project) -> bool:
...
+    auto_update = bool(project.get_option("sentry:grouping_auto_update"))
+    return deprecated_config or (auto_update and should_update_org)
```
lobsterkatie, line 43: Nit: This makes it clearer the variable isn't pointed to a config itself.

```suggestion
    is_deprecated_config = project.get_option("sentry:grouping_config") in CONFIGS_TO_DEPRECATE
    auto_update = bool(project.get_option("sentry:grouping_auto_update"))
    return is_deprecated_config or (auto_update and should_update_org)
```

src/sentry/grouping/ingest/config.py
```diff
@@ -79,6 +92,11 @@ def _auto_update_grouping(project: Project) -> None:
             "sentry:secondary_grouping_expiry": expiry,
             "sentry:grouping_config": new_config,
         }
+        # Any project on deprecated configs may have disabled auto updates and
+        # this will reduce the chance of happening unintentionally.
```
lobsterkatie, line 96: The chance of _what_ happening unintentionally - being on a deprecated config, or having auto-update False?

src/sentry/grouping/ingest/config.py
lobsterkatie: So will this only be called from a script? If so, can we please put it into `grouping/utils.py`, since it's not called during ingest?

src/sentry/grouping/ingest/config.py
armenzg: Being on a deprecated config because they disabled updates. I _assume_ that for testing purposes they followed instructions on how to move to the mobile config and not being moved out of it.

src/sentry/grouping/ingest/config.py
armenzg: `_auto_update_grouping` is called during ingestion via `update_grouping_config_if_needed`

armenzg: I'm going to push back. All the code that is needed is here.

I can remove this function and put it directly on the script (which would make more sense).

lobsterkatie: > `_auto_update_grouping` is called during ingestion via `update_grouping_config_if_needed`

Yup. My thought was more, "it's going to be used by both ingest and this script-only function, so it should be in a neutral location."

But anyway, if the other function is only going to be used in the script, then yeah, I think moving it there is a better idea than putting it in `utils`.

src/sentry/grouping/ingest/config.py
lobsterkatie: Got it. Thanks for clarifying. In that case, how about 

```suggestion
        # Any project on deprecated configs may be there only because auto updates are
        # disabled (not because they want to use that specific config). This will reduce the 
        # chance of that happening unintentionally.
```

lobsterkatie (approved):

PR #74216, opened 2024-07-12
feat(ui): add edit auth token page

ameliahsu:
created UI to rename auth tokens (https://github.com/getsentry/sentry/issues/73556)

BEFORE (can not edit token name)
Screenshot 2024-07-12 at 2 01 17 PM

AFTER (token name links to page to edit token details)
image
\*\* note that tokens with no name are listed as _unnamed token_

New Edit User Auth Token page
image


iamrajjoshi:
quick nit question: how doe the UI work when a user deletes the name of a token?

ameliahsu:
@iamrajjoshi if the user delete the token name and clicks "Save Changes," on the User Auth Tokens page that lists all the API tokens the token will be listed as _unnamed token_

the AFTER photo I included has an example

saponifi3d (approved):
pretty much just nit picks / recommendations on ways to improve the code structure. i'd recommend adding a few more tests to ensure negative cases before merging.

static/app/utils/useMutateApiToken.tsx
```diff
@@ -0,0 +1,86 @@
...
+import {
+  getApiQueryData,
+  setApiQueryData,
+  useMutation,
+  useQueryClient,
+} from 'sentry/utils/queryClient';
+import type RequestError from 'sentry/utils/requestError/requestError';
+import useApi from 'sentry/utils/useApi';
+
+const API_TOKEN_QUERY_KEY = ['/api-tokens/'] as const;
```
saponifi3d, line 11: shouldn't need to type cast this, but you can type it a little more:
```suggestion
const API_TOKEN_QUERY_KEY: string[] = ['/api-tokens/'];
```

static/app/utils/useMutateApiToken.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+type UpdateTokenQueryVariables = {
+  name: string;
+};
+type FetchApiTokenParameters = {
+  tokenId: string;
+};
+export const makeFetchApiTokenKey = ({tokenId}: FetchApiTokenParameters) =>
+  [`/api-tokens/${tokenId}/`] as const;
```
saponifi3d, line 20:
```suggestion
export const makeFetchApiTokenKey = ({tokenId}: FetchApiTokenParameters): string[] => [`/api-tokens/${tokenId}/`];
```

static/app/utils/useMutateApiToken.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export default function useMutateApiToken({
+  token,
+  onSuccess,
+  onError,
+}: UseMutateApiTokenProps) {
+  const api = useApi();
+  const queryClient = useQueryClient();
+  return useMutation<{}, RequestError, UpdateTokenQueryVariables>({
```
saponifi3d, line 35: for `{}` type, that's the expected response type. will this return a new token or anything?

static/app/utils/useMutateApiToken.tsx
```diff
@@ -0,0 +1,86 @@
...
+          name,
+        },
+      }),
+
+    onSuccess: (_data, {name}) => {
+      // Update get by id query
+      setApiQueryData(
+        queryClient,
+        makeFetchApiTokenKey({tokenId: token.id}),
+        (oldData: InternalAppApiToken | undefined) => {
```
saponifi3d, line 49:
```suggestion
        (oldData?: InternalAppApiToken) => {
```

static/app/utils/useMutateApiToken.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+            if (existingToken) {
+              existingToken.name = name;
+            }
+
+            return oldData;
+          }
+        );
+      }
+      return onSuccess?.();
```
saponifi3d, line 80: make sure to pass along any data to the `onSuccess` handler here, it's expecting `onSuccess: (_data, {name})`, but you'll have to make sure to pass along any data modifications you made to the data.

saponifi3d (approved):
generally lgtm, just make sure to update the `onSuccess` function to pass the params to the callback function :tada:

static/app/utils/useMutateApiToken.tsx
ameliahsu: this won't return anything, but the new token will be passed to the `onSuccess` method if the token is successfully modified

Issue #74209, opened 2024-07-12
[User Feedback] Add Option to Hide Feedback Messages on Issue Details & Persist That As a User Setting


jas-kas:
## Problem

TL/DR: Issue Details page can show “User Feedback” section when the feedback comes from a crash report. 

Sometimes the feedback is vulgar and Sentry users don’t want to see that.

## Solution

Task: Make the feedback collapsable, and remember server-side that feedback is collapsed/hidden. The setting is applied to any feedback rendered directly on the Issue Details page, so if people hide the feedback once there, it’s hidden all the time for all feedbacks on Issue Details. Note: feedbacks on the the /feedback page and the Issue Details > User Feedback tab will still be visible as they are dedicated views for solely feedback messages.

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2222) by [Unito](https://www.unito.io)

jas-kas:
cc/ @ryan953 @aliu39 @michellewzhang

michellewzhang:
similar to what we did for the issue details onboarding banner:

- [ ] BE: add a new prompt config to the backend: ex. https://github.com/getsentry/sentry/pull/69513/files
- [ ] FE: add a chevron to the user feedback component in `userFeedback.tsx` that will allow the user to collapse/expand the feedback (maybe add a message like "feedback is hidden" when it's hidden)
- [ ] FE: remember the hidden state using the `usePrompt` hook on a **PROJECT** level: ex. https://github.com/getsentry/sentry/pull/69525/files

PR #74219, opened 2024-07-12
feat(feedback): add BE prompt for hiding feedback from issue details

aliu39:
Relates to https://github.com/getsentry/sentry/issues/74209

### DIFF
--- a/src/sentry/utils/prompts.py
+++ b/src/sentry/utils/prompts.py
@@ -19,6 +19,7 @@ DEFAULT_PROMPTS = {
     "data_consent_banner": {"required_fields": ["organization_id"]},
     "data_consent_priority": {"required_fields": ["organization_id"]},
     "issue_replay_inline_onboarding": {"required_fields": ["organization_id", "project_id"]},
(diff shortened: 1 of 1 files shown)


src/sentry/utils/prompts.py
```diff
@@ -19,6 +19,7 @@
     "data_consent_banner": {"required_fields": ["organization_id"]},
     "data_consent_priority": {"required_fields": ["organization_id"]},
     "issue_replay_inline_onboarding": {"required_fields": ["organization_id", "project_id"]},
+    "issue_feedback_hidden": {"required_fields": ["organization_id", "project_id"]},
```
michellewzhang, line 22: what if we renamed it to something like `issue_user_feedback`?  the trend seems to be that the names should be more related to the component rather than the action

michellewzhang: i think all `prompts` are used for hiding/dismissing something tho, it's implied but im not super adamant on it

michellewzhang (approved):

Issue #73856, opened 2024-07-04
App Modules - Consolidate User resources


markstory:
User and its direct relations are control_silo scoped resources. Currently the components that comprise the control silo aspects of users are spread throughout the repository. Consolidating those classes and functions into `sentry.users` will help clarify the application's design and make running subsets of tests easier and faster.

- Communicate these changes with the broader ecosystem team
- Move endpoints, serializers, models, and tasks to be co-located.

We'll need to provide import shims for any symbols used in getsentry until those imports can be updated separately. Furthremore, within `django shell` we should preserve `sentry.models.User` as an import.

cc @Christinarlong

markstory:
Closing as we've finished users off.

PR #74222, opened 2024-07-12
ref(control_silo): Consolidate User models (and authenticator) and tests

Christinarlong:
Moves all User and authenticator specific models and tests to their own directories.
    
ref(#73856)


bin/mock-user
```diff
@@ -8,7 +8,7 @@ import argparse
 
 
 def main(username, newsletter_consent_prompt=None):
-    from sentry.models.user import User
+    from sentry.users.models.users.user import User
```
markstory, line 11: Could this be `sentry.users.models.user`? Or is the additional layer of modules helpful?

src/sentry/models/__init__.py
```diff
@@ -1,3 +1,7 @@
+from ..users.models.auth.authenticator import *  # NOQA
+from ..users.models.users.userip import *  # NOQA
+from ..users.models.users.userreport import *  # NOQA
+from ..users.models.users.userrole import *  # NOQA
```
markstory, line 4: I would avoid relative imports that go up the module path. These could be `from sentry.users.models ...`.

src/sentry/models/authenticator.py
```diff
@@ -1,204 +1,3 @@
-from __future__ import annotations
+from ..users.models.auth.authenticator import Authenticator  # NOQA
```
markstory, line 1: Could we do `from sentry.users.models.authenticator import Authenticator`. The users app will have fewer than 10 models in it and I don't think the additional layer of directories adds much.

src/sentry/models/authenticator.py
```diff
@@ -1,204 +1,3 @@
...
-
-    @classmethod
-    def sanitize_relocation_json(
-        cls, json: Any, sanitizer: Sanitizer, model_name: NormalizedModelName | None = None
-    ) -> None:
-        model_name = get_model_name(cls) if model_name is None else model_name
-        super().sanitize_relocation_json(json, sanitizer, model_name)
-
-        sanitizer.set_string(json, SanitizableField(model_name, "config"), lambda _: '""')
+# flake8 dislikes imported but unused modules
```
markstory, line 3:
```suggestion
__all__ = ("Authentictator", )
```

Giving the module an `__all__` is what I've done in the past for these. It also lets you skip the `# NOQA` on the import.

src/sentry/models/avatars/__init__.py
```diff
@@ -1,9 +1,9 @@
+from ...users.models.users.user_avatar import UserAvatar
```
markstory, line 1: Could we use absolute imports when we're crossing modules that are children of `sentry`?

src/sentry/models/userrole.py
```diff
@@ -1,97 +1,7 @@
...
+from sentry.users.models.users.userrole import (  # NOQA
+    UserRole,
+    UserRoleUser,
     manage_default_super_admin_role,
-    dispatch_uid="manage_default_super_admin_role",
-    weak=False,
-    sender=SiloMode.MONOLITH,
 )
+
+# flake8 dislikes imported but unused modules
```
markstory, line 7: This is another place where I would put an `__all__` to appease tooling and humans doing `*` imports.

src/sentry/users/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+from ...users.models.users.user import *  # NOQA
```
markstory, line 1: Could we use absolute imports here? The `...` pattern is hard to search for in the future.

bin/mock-user
Christinarlong: Sure it definitely could go there. Just kinda worried that with all the other models, the models directory could grow p. large though we could do something like:
```
sentry.users.models
|
|--> auth
         |
         |authenticator.py
         
|--> emails
        
| user.py
| useremail.py
```
does this feel cleaner/makes more sense?

Christinarlong: nvm, read the rest of the review. 10 is p. few so I'm down to just flatten and put everything in models

Christinarlong:
Breaking down into smaller PRs

PR #74224, opened 2024-07-12
feat(issue-details): Add a foldable data section component

leeandher:
Super simple folding section. Didn't make progress on removing the margins for the independent data sections just yet, but at least checking this in will allow its use in the next week while I'm out

https://github.com/user-attachments/assets/2594daa9-e9fe-40ee-b83b-fd70313b01f9


static/app/views/issueDetails/streamline/foldSection.tsx
```diff
@@ -0,0 +1,84 @@
...
+}: FoldSectionProps) {
+  const [isCollapsed, setIsCollapsed] = useLocalStorageState(
+    `${LOCAL_STORAGE_PREFIX}${sectionKey}`,
+    initialCollapse
+  );
+
+  const toggleCollapse = useCallback(
+    (e: React.MouseEvent) => {
+      e.preventDefault();
+      setIsCollapsed(!isCollapsed);
```
malwilley, line 41: Should use the callback when the state change is dependent on the old state: `setIsCollapsed(collapsed => !collapsed)`

static/app/components/globalDrawer/index.stories.tsx
malwilley, line 1: These changes look unrelated?

malwilley (commented):
This might be planned future behavior, but I'd expect to see some hover/pressed states? You might be able to use `InteractionStateLayer` if you want the whole header to appear clickable.

static/app/views/issueDetails/streamline/foldSection.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <section {...props}>
+      <Details open={!isCollapsed || preventCollapse}>
+        <Summary
+          preventCollapse={preventCollapse}
+          onClick={preventCollapse ? e => e.preventDefault() : toggleCollapse}
+        >
+          <div>{title}</div>
+          <IconWrapper preventCollapse={preventCollapse}>
```
malwilley, line 54: There should be a button somewhere with an appropriate aria label, as-is there aren't any clickable elements

static/app/components/globalDrawer/index.stories.tsx
leeandher: Yeah, referenced it when putting together the other story book and corrected a typo.

static/app/views/issueDetails/streamline/foldSection.tsx
leeandher: `<summary>` is a conventional clickable HTML element that [most often gets implicit roles](https://www.w3.org/TR/html-aria/#docconformance), but there was a note on it:

image

That said, there's also [an example](https://www.w3.org/TR/html-aria/#side-effects) implying not to change set the role to button, so I guess no action needed here? 

image

Though I might be misunderstanding.

static/app/views/issueDetails/streamline/foldSection.stories.tsx
```diff
@@ -0,0 +1,166 @@
...
+          sectionKey={FoldSectionKey.HIGHLIGHTS}
+        >
+          <Lorem />
+        </FoldSection>
+        <FoldSection
+          title={
+            <Title>
+              <div>Header with Actions</div>
+              <ButtonBar gap={1}>
+                <Button size="xs" icon={<IconAdd />}>
```
roggenkemper, line 79: might be because it's a test button, but clicking these buttons collapses the section

static/app/views/issueDetails/streamline/foldSection.tsx
```diff
@@ -0,0 +1,170 @@
...
+                onClick={e => e.stopPropagation()}
+                onMouseEnter={() => setIsLayerEnabled(false)}
+                onMouseLeave={() => setIsLayerEnabled(true)}
+              >
+                {actions}
+              </div>
+            )}
+          </TitleWithActions>
+          <IconWrapper preventCollapse={preventCollapse}>
+            <IconChevron direction={isCollapsed ? 'down' : 'up'} size="xs" />
```
roggenkemper, line 129: the chevrons on the stories page are pointing in the wrong direction

leeandher: Good catch!  Updated the storybook video, but it should be accurate now.

static/app/views/issueDetails/streamline/foldSection.tsx
malwilley: Ah no you're right, I'm just unfamiliar with the summary element!

static/app/views/issueDetails/streamline/foldSection.tsx
```diff
@@ -0,0 +1,170 @@
...
+  sectionKey,
+  initialCollapse = false,
+  preventCollapse = false,
+  ...props
+}: FoldSectionProps) {
+  const [isCollapsed, setIsCollapsed] = useLocalStorageState(
+    `${LOCAL_STORAGE_PREFIX}${sectionKey}`,
+    initialCollapse
+  );
+  const [isLayerEnabled, setIsLayerEnabled] = useState(true);
```
malwilley, line 99: Why is `isLayerEnabled` necessary?

static/app/views/issueDetails/streamline/foldSection.stories.tsx
malwilley, line 1: Very nice 👏

static/app/views/issueDetails/streamline/foldSection.tsx
leeandher: I wasn't sure of another way to accomplish this, but it's used to disable the interaction state layer when hovering over the 'button area', to keep things legible.

https://github.com/user-attachments/assets/7f61aabc-94d8-4806-836a-81dcb5296d15

leeandher: Is this still the case, i wasn't able to repro

https://github.com/user-attachments/assets/4aed6e83-1e77-4773-a3d6-4a29e265b9e4

malwilley: Ahh I see! If you can't find another way around that I'd add a comment.

One other thing you could try would be styling the header with something like:

```css
&:has(button:hover) {
  ${InteractionStateLayer} {
    display: none;
  }
}
```

Not sure if that actually works or not (or is any better)

static/app/views/issueDetails/streamline/foldSection.stories.tsx
leeandher: Is this still the case, i wasn't able to repro

https://github.com/user-attachments/assets/4aed6e83-1e77-4773-a3d6-4a29e265b9e4

malwilley (approved):

static/app/views/issueDetails/streamline/foldSection.stories.tsx
roggenkemper: nope - it looks good now!

roggenkemper (approved):

static/app/views/issueDetails/streamline/foldSection.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+  const toggleCollapse = useCallback(
+    (e: React.MouseEvent) => {
+      e.preventDefault(); // Prevent browser summary/details behaviour
+      setIsCollapsed(collapsed => {
+        trackAnalytics('issue_details.section_fold', {
```
malwilley, line 105: I don't know if there is a rule for this or not, but I'd put the `trackAnalytics` call outside of the state updater function. It should be idempotent

PR #74228, opened 2024-07-12
feat(prompts): support a visible status for un-hiding components

aliu39:
Needed for https://github.com/getsentry/sentry/issues/74209. We're using the prompt backend to persist a user's decision to hide/unhide user feedback in the issue details page. The goal is to be able to toggle between [promptIsDismissed](https://github.com/getsentry/sentry/blob/539945981a00bc93a9525e61e052563df701757a/static/app/utils/promptIsDismissed.tsx#L18) = true/false.


evanpurkhiser (commented):
is there a test we can add?

markstory (approved):
Looks good to me.

PR #74229, opened 2024-07-12
ref(rules): Only fetch necessary data

ceorourke:
Optimize queries by only fetching the strictly necessary data, and add a metric around a rough estimate of how many groups we're processing.


src/sentry/rules/conditions/event_frequency.py
```diff
@@ -618,21 +626,21 @@ def batch_query_hook(
         self, group_ids: set[int], start: datetime, end: datetime, environment_id: int
     ) -> dict[int, int]:
         batch_percents: dict[int, int] = defaultdict(int)
-        groups = Group.objects.filter(id__in=group_ids)
-        project_id = groups[0].project.id
-        session_count_last_hour = self.get_session_count(project_id, environment_id, start, end)
+        groups = Group.objects.filter(id__in=group_ids).values_list("id", "type", "project")
+        project_ids = Project.objects.filter(id=groups[0][2]).values_list("id", flat=True)
```
ceorourke, line 630: Is there a way to get the project id in the group query above? I wasn't able to figure it out

src/sentry/rules/conditions/event_frequency.py
```diff
@@ -326,22 +327,27 @@ def get_snuba_query_result(
...
         start: datetime,
         end: datetime,
         environment_id: int,
         referrer_suffix: str,
     ) -> Mapping[int, int]:
+        org_id = (
+            Group.objects.filter(id=group_id)
+            .select_related("project__organization")
+            .values_list("id", flat=True)
+        )
```
wedamija, line 341: `.values_list("project__organization_id", flat=True)`. You can remove the `select_related`

Wondering if we should just pass the org id down though?

src/sentry/rules/conditions/event_frequency.py
```diff
@@ -396,27 +402,28 @@ def batch_query_hook(
         self, group_ids: set[int], start: datetime, end: datetime, environment_id: int
     ) -> dict[int, int]:
         batch_sums: dict[int, int] = defaultdict(int)
-        groups = Group.objects.filter(id__in=group_ids)
-        error_issues = [group for group in groups if group.issue_category == GroupCategory.ERROR]
-        generic_issues = [group for group in groups if group.issue_category != GroupCategory.ERROR]
+        groups = Group.objects.filter(id__in=group_ids).values_list("id", "type")
+        error_issue_ids = [group[0] for group in groups if group[1] == GroupCategory.ERROR.value]
+        generic_issue_ids = [group[0] for group in groups if group[1] != GroupCategory.ERROR.value]
```
wedamija, line 407: You'll need to copy this logic

https://github.com/getsentry/sentry/blob/3f6b07dbd53386c8b8bb44a84fbffcdd5d59f16f/src/sentry/models/group.py#L953-L959

src/sentry/rules/conditions/event_frequency.py
```diff
@@ -453,27 +460,28 @@ def batch_query_hook(
         self, group_ids: set[int], start: datetime, end: datetime, environment_id: int
     ) -> dict[int, int]:
         batch_totals: dict[int, int] = defaultdict(int)
-        groups = Group.objects.filter(id__in=group_ids)
-        error_issues = [group for group in groups if group.issue_category == GroupCategory.ERROR]
-        generic_issues = [group for group in groups if group.issue_category != GroupCategory.ERROR]
+        groups = Group.objects.filter(id__in=group_ids).values_list("id", "type")
+        error_issue_ids = [group[0] for group in groups if group[1] == GroupCategory.ERROR.value]
+        generic_issue_ids = [group[0] for group in groups if group[1] != GroupCategory.ERROR.value]
```
wedamija, line 465: I think we should move this into a function so the logic is shared

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -378,6 +378,20 @@ def get_group_to_groupevent(
...
+        return ">10000"
+    elif num_groups > 1000:
+        return ">1000"
+    elif num_groups > 100:
+        return ">100"
+    elif num_groups > 10:
+        return ">10"
+    elif num_groups > 1:
+        return ">1"
+    return "1"
```
saponifi3d, line 392: can simplify the logic for 1 -> Infinity with: `10 ** int(math.log10(num_groups))` (ty dan) -- that will return the magnitude of the number.

to mimic what you have:
```suggestion
  if num_groups > 1:
    magnitude = 10 ** int(math.log10(num_groups))
    return f">{magnitude}"
  return "1"
```

saponifi3d (commented):
mind adding a few unit tests to this as well? (i think we need it to get past codecov too)

saponifi3d (approved):
LGTM, hopefully helps! 🤞

src/sentry/rules/conditions/event_frequency.py
```diff
@@ -421,13 +421,18 @@ def batch_query_hook(
             "id", "type", "project__organization_id"
         )
         error_issue_ids, generic_issue_ids = self.get_error_and_generic_group_ids(groups)
+        organization_id = None
+        if groups:
+            group = groups[0]
+            if len(group) == 3:
+                organization_id = group[3]
```
saponifi3d, line 428: Should this be 2? (len 3 is 👍 -- good ol' off by 1s)
```suggestion
                organization_id = group[2]
```

ceorourke: lol oops yes

PR #74232, opened 2024-07-12
feat(api): Support Project ID and Slugs in `OrganizationArtifactBundleAssembleEndpoint`

iamrajjoshi:
I am working on getting all the commands in our Sentry CLI to support Ids as well as Slugs. [rough spec](https://www.notion.so/sentry/Supporting-IDs-and-Slugs-in-Sentry-CLI-35b805ea169e47348805672cdee41297?pm=c)

We have already gone through the effort of supporting ids and slugs in path parameters of all the APIs in our codebase. There are some endpoints that the CLI uses that pass in project slugs as _body parameters_, so we need to support Ids for those as well.

Here, I use the fact that ids are numeric and slugs aren't to process both type of identifiers.

### DIFF
--- a/src/sentry/api/endpoints/organization_artifactbundle_assemble.py
+++ b/src/sentry/api/endpoints/organization_artifactbundle_assemble.py
@@ -56,19 +56,41 @@ class OrganizationArtifactBundleAssembleEndpoint(OrganizationReleasesBaseEndpoin
         except Exception:
             return Response({"error": "Invalid json body"}, status=400)
 
-        projects = set(data.get("projects", []))
-        if len(projects) == 0:
+        input_projects = set(data.get("projects", []))
+        if len(input_projects) == 0:
             return Response({"error": "You need to specify at least one project"}, status=400)
 
-        project_ids = list(
+        input_project_slug = set()
+        input_project_id = set()
+        for project in input_projects:
+            # IDs are always numeric, slugs cannot be numeric
+            if str(project).isnumeric():
+                input_project_id.add(project)
+            else:
+                input_project_slug.add(project)
+
+        matched_projects_by_slug = set(
             Project.objects.filter(
-                organization=organization, status=ObjectStatus.ACTIVE, slug__in=projects
+                organization=organization,
+                status=ObjectStatus.ACTIVE,
+                slug__in=input_project_slug,
             ).values_list("id", flat=True)
         )
-        if len(project_ids) != len(projects):
+
+        matched_projects_by_id = set(
+            Project.objects.filter(
+                organization=organization,
+                status=ObjectStatus.ACTIVE,
+                id__in=input_project_id,
+            ).values_list("id", flat=True)
+        )
+
+        project_ids = matched_projects_by_slug.union(matched_projects_by_id)
+
+        if len(project_ids) != len(input_projects):
             return Response({"error": "One or more projects are invalid"}, status=400)
 
-        if not self.has_release_permission(request, organization, project_ids=set(project_ids)):
+        if not self.has_release_permission(request, organization, project_ids=project_ids):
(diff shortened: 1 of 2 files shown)


src/sentry/api/endpoints/organization_artifactbundle_assemble.py
```diff
@@ -56,19 +56,41 @@ def post(self, request: Request, organization) -> Response:
...
+        input_projects = set(data.get("projects", []))
+        if len(input_projects) == 0:
             return Response({"error": "You need to specify at least one project"}, status=400)
 
-        project_ids = list(
+        input_project_slug = set()
+        input_project_id = set()
+        for project in input_projects:
+            # IDs are always numeric, slugs cannot be numeric
+            if str(project).isnumeric():
```
schew2381, line 67:
```suggestion
            if str(project). isdecimal():
```

src/sentry/api/endpoints/organization_artifactbundle_assemble.py
```diff
@@ -56,19 +56,41 @@ def post(self, request: Request, organization) -> Response:
...
+                input_project_id.add(project)
+            else:
+                input_project_slug.add(project)
+
+        matched_projects_by_slug = set(
             Project.objects.filter(
-                organization=organization, status=ObjectStatus.ACTIVE, slug__in=projects
+                organization=organization,
+                status=ObjectStatus.ACTIVE,
+                slug__in=input_project_slug,
```
schew2381, line 76: I think you can make this is one query by doing:
```Python
filtered_queryset = YourModel.objects.filter(
    (Q(id__in=id_set) | Q(slug__in=slug_set)),
    organization=organization,
)
```

src/sentry/api/endpoints/organization_artifactbundle_assemble.py
```diff
@@ -131,6 +144,6 @@ def post(self, request: Request, organization) -> Response:
         )
 
         if is_org_auth_token_auth(request.auth):
-            update_org_auth_token_last_used(request.auth, project_ids)
+            update_org_auth_token_last_used(request.auth, list(project_ids))
```
schew2381, line 147: nit: if we don't convert to a set originally we can remove `set` here

src/sentry/api/endpoints/organization_artifactbundle_assemble.py
```diff
@@ -56,19 +57,31 @@ def post(self, request: Request, organization) -> Response:
         except Exception:
             return Response({"error": "Invalid json body"}, status=400)
 
-        projects = set(data.get("projects", []))
-        if len(projects) == 0:
+        input_projects = set(data.get("projects", []))
```
schew2381, line 60: nit: I think we can skip converting this to a set as I don't see it being used anywhere

tests/sentry/api/endpoints/test_organization_artifactbundle_assemble.py
```diff
@@ -141,6 +141,86 @@ def test_assemble_with_invalid_projects(self):
         assert response.status_code == 400, response.content
         assert response.data["error"] == "One or more projects are invalid"
 
+    def test_assemble_with_valid_project_slugs(self):
```
schew2381, line 144: subjective nit: maybe this test isn't needed b/c we're using project slug already for everything else? This is prob more subjective tho cuz maybe it's good to explicitly test anyways. Up to you

schew2381 (approved):
some nits but overall lgtm!

src/sentry/api/endpoints/organization_artifactbundle_assemble.py
schew2381: oh wait it's a queryset oops

PR #74236, opened 2024-07-12
feat(feedback): button to hide/show user feedback in issue details

aliu39:
Depends on https://github.com/getsentry/sentry/pull/74228

Relates to https://github.com/getsentry/sentry/issues/74209

Adds a button to the "User Feedback" section of issue details page. Allows members to hide/unhide feedback from **all** issue details in a project. Persists this decision across reloads using the prompt backend

Updated demo with chevron/drop-down style hiding:

https://github.com/user-attachments/assets/43a02a63-98c0-42bd-8651-f1baeeb22218

Updated loading indicator: (you can see for a split sec at :12)

https://github.com/user-attachments/assets/a2d57c51-e459-467b-9a9c-5e227050d9b7


aliu39:
This component, EventUserFeedback, is used in other places too. For ex the User Feedback tab of issues. So we'll have to move the usePrompt and isPromptDismissed check to one of its parents, probably `EventDataSection`.

static/app/views/issueDetails/groupEventDetails/groupEventDetailsContent.tsx
```diff
@@ -131,13 +149,31 @@ function DefaultGroupEventDetailsContent({
...
+                  isHidden={feedbackHidden}
+                  showFeedback={showPrompt}
+                  hideFeedback={dismissPrompt}
+                />
+              </ButtonBar>
+            </ErrorBoundary>
+          }
+        >
+          {feedbackHidden ? (
+            '(Hidden)'
```
michellewzhang, line 168: style thing but could we instead have something like `This feedback is hidden.` in grey text and italicized? i think that would match sentry style a bit more

static/app/actionCreators/prompts.tsx
```diff
@@ -215,12 +226,39 @@ export function usePrompt({
...
+  const showPrompt = useCallback(() => {
+    promptsUpdate(api, {
+      organization,
+      projectId,
+      feature,
+      status: 'visible',
+    });
+
+    // Update cached query data
+    // Will set prompt to snoozed
```
michellewzhang, line 238: why are we setting to snoozed?

static/app/views/issueDetails/groupEventDetails/groupEventDetailsContent.tsx
```diff
@@ -131,13 +149,31 @@ function DefaultGroupEventDetailsContent({
...
-            orgSlug={organization.slug}
-            issueId={group.id}
-            showEventLink={false}
-          />
+        <EventDataSection
+          title={t('User Feedback')}
+          type="user-feedback"
+          actions={
+            <ErrorBoundary mini>
+              <ButtonBar gap={1}>
```
michellewzhang, line 157: we don't need a `ButtonBar` here if it's just 1 button

static/app/actionCreators/prompts.tsx
aliu39: Forgot to update comment

static/app/components/events/userFeedback.tsx
```diff
@@ -11,6 +11,36 @@ import type {UserReport} from 'sentry/types/group';
 import {escape, nl2br} from 'sentry/utils';
 import useCopyToClipboard from 'sentry/utils/useCopyToClipboard';
 
+// used for issue details page (groupEventDetailsContent.tsx)
+export function HideUserFeedbackButton({
```
aliu39, line 15: I'm thinking of moving this to groupEventDetailsContent.tsx

static/app/views/issueDetails/groupEventDetails/groupEventDetailsContent.tsx
```diff
@@ -131,13 +148,33 @@ function DefaultGroupEventDetailsContent({
...
+                <HideUserFeedbackButton
+                  isLoading={promptLoading}
+                  isError={promptError || isPromptDismissed === undefined}
+                  isHidden={isPromptDismissed !== undefined ? isPromptDismissed : false}
```
michellewzhang, line 160:
```suggestion
                  isHidden={isPromptDismissed ?? false}
```

first part will automatically check for `undefined`

static/app/components/events/userFeedback.tsx
michellewzhang: might be good, could reduce a bunch of prop passing

static/app/views/issueDetails/groupEventDetails/groupEventDetailsContent.tsx
```diff
@@ -131,13 +148,33 @@ function DefaultGroupEventDetailsContent({
...
+                  showFeedback={showPrompt}
+                  hideFeedback={dismissPrompt}
+                />
+              </ButtonBar>
+            </ErrorBoundary>
+          }
+        >
+          {promptLoading || promptError || isPromptDismissed ? (
```
michellewzhang, line 168: should we have some different states for loading & error? could use `<LoadingIndicator>` or `<LoadingError>`

aliu39: I'm thinking of defaulting to displaying the feedback in either of these cases. So if the prompt API is having problems, users will still be able to see their feedback.

michellewzhang: that makes sense to me

static/app/views/issueDetails/groupEventDetails/groupEventDetailsContent.tsx
```diff
@@ -131,13 +177,33 @@ function DefaultGroupEventDetailsContent({
...
+          actions={
+            // Button is disabled if promptLoading or promptError.
+            <ErrorBoundary mini>
+              <HideUserFeedbackButton
+                isLoading={promptLoading}
+                isError={promptError || isPromptDismissed === undefined}
+                isHidden={isPromptDismissed ?? false}
+                showFeedback={showPrompt}
+                hideFeedback={dismissPrompt}
+              />
```
michellewzhang, line 192: would it be better to just move the button code here directly? it's not that much more code, and you'll have less prop passing

aliu39: Truu

static/app/views/issueDetails/groupEventDetails/groupEventDetailsContent.tsx
```diff
@@ -131,13 +148,38 @@ function DefaultGroupEventDetailsContent({
...
+              <Button
+                size="xs"
+                onClick={isPromptDismissed ? showPrompt : dismissPrompt}
+                title={
+                  isPromptDismissed
+                    ? t('Unhide feedback on all issue details')
+                    : t('Hide feedback on all issue details')
+                }
+                disabled={promptError || promptLoading || isPromptDismissed === undefined}
+              >
```
michellewzhang, line 166: what does it mean if `isPromptDismissed` undefined?

aliu39: https://github.com/getsentry/sentry/blob/138a24e16e507fdb66aa231a0f4300bcb37d62ce/static/app/actionCreators/prompts.tsx#L162 I think some kind of error with the backend. I'm assuming when this occurs promptError=true.

michellewzhang: i think it shouldn't be disabled if `isPromptDismissed === undefined` -- it looks like it's undefined anytime the prompt is not dismissed, even with no error


SCR-20240715-lqtw

aliu39: oh interesting.. thanks for testing this case!

michellewzhang: actually, hold on... `isPromptDismissed` is supposed to return `true` or `false` so the logic isn't adding up. maybe it's a vercel preview thing. does https://sentry-sdks.sentry.io/issues/5291936195/events/e24fd30c0f2b47a9b20bc5014de8e493/?project=-1&referrer=event-or-group-header&statsPeriod=90d work as expected for you?

aliu39: Using the latest commit, the button is disabled for me

michellewzhang:
might need a loading state here 

https://github.com/user-attachments/assets/68beca83-b09d-4f32-9176-5038628a0baf

static/app/views/issueDetails/groupEventDetails/groupEventDetailsContent.tsx
aliu39: Agreed on slack to show the feedback if error or `isPromptDismissed === undefined`, use `LoadingIndicator` when loading

michellewzhang (approved):
oo i like the new loading indicator! i know @billyvg said it adds noise to the page, so we could always do `Placeholder` instead but it definitely looks a lot better. cc @Jesse-Box for design thoughts

PR #74237, opened 2024-07-12
ref(query-builder): Reorganize files

malwilley:
Now that it's mostly feature complete, taking an opportunity to reorganize the file structure:

- Move all files which render token types into `/tokens` folder
- Move all files which render the filter token type into `/tokens/filter`
- Break up the base `utils.tsx` into ones that go into each subfolder
- Move all base-level hooks into `/hooks` 
- Change some component names to be more consistent


static/app/components/searchQueryBuilder/tokens/filter/utils.tsx
malwilley, line 1: Copied from the base `utils.tsx` file

static/app/components/searchQueryBuilder/tokens/utils.tsx
malwilley, line 1: Copied from the base `utils.tsx` file

roggenkemper (approved):
"Promise this is the last query builder pr" - aged like milk

PR #74238, opened 2024-07-12
ref(replay): replace console objectInspector with structuredEventData

michellewzhang:


static/app/views/replays/detail/network/details/sections.tsx
```diff
@@ -39,7 +39,7 @@ export type SectionProps = {
   startTimestampMs: number;
 };
 
-const config = {
+export const replayDataConfig = {
```
billyvg, line 42: we should move this somewhere common with other modules that need to import it

PR #74241, opened 2024-07-12
ref: Change ActionHandler.property from class var to abstract property

RyanSkonnord:
Make it clearer that `ActionHandler` expects subclasses to provide the value (the `abstractmethod` decorator will complain if it isn't there). Also, provide a better way to make a subclass where instances can serve different providers (where the provider value may depend on something passed to `__init__`).


RyanSkonnord:
This supports a further refactoring that I'd like to do in https://github.com/getsentry/sentry/pull/74070

cathteng (approved):
does putting it as a property in a function allow us to possibly return the provider as something passed into `__init__`? like via self?

dashed (approved):

RyanSkonnord:
> does putting it as a property in a function allow us to possibly return the provider as something passed into `__init__`? like via self?

It sure does! In fact, here's a preview of what I'm trying to write:

```python
class MessagingActionHandler(DefaultActionHandler):
    def __init__(self, action, incident, project, spec):
        super().__init__(action, incident, project)
        self.spec = spec

    @property
    def provider(self) -> str:
        return self.spec.integration_provider.key
```

(Although I'm actually having a slight problem right now, wanting to get the key from a `type[IntegrationProvider]` value and not an `IntegrationProvider` instance, which worked the old way but not with the new way. 😕 I'm going to hack on it a little longer before I merge this.)

PR #74244, opened 2024-07-12
feat(grouping): allow bun ai-grouping

trillville:
similar to JS, validated with one backfill and performance looks good


mrduncan (approved):

trillville:
realized that `bun` is not a valid project in this context

PR #74245, opened 2024-07-12
fix(replays): Fixes flickering when replaying a mobile replay

### LINKED ISSUE #70687: [Mobile Replay - Bug] Video player plays wrong clip

c298lee:
When replaying a replay, the video flickers between segments. When you slow down the video, you notice the flickering is caused by the last frame of the next video showing, before it quickly changes to the first frame of the next video when the video is played. 
To fix this, when a video ends, we reset it to the beginning to ensure that the last frame isn't shown when replaying.

Before:

https://github.com/user-attachments/assets/5ec8d1a0-03e6-4a26-b028-cc9d4cad02fd

After:

https://github.com/user-attachments/assets/eec76606-b36f-4d03-9478-2f7b3ba78a74

Closes https://github.com/getsentry/sentry/issues/70687

### DIFF
--- a/static/app/components/replays/videoReplayer.tsx
+++ b/static/app/components/replays/videoReplayer.tsx
@@ -412,6 +412,10 @@ export class VideoReplayer {
     // This is the soon-to-be previous video that needs to be hidden
     if (this._currentVideo) {
       this._currentVideo.style.display = 'none';
+      // resets the soon-to-be previous video to the beginning if it's ended so it starts from the beginning on replay
+      if (this._currentVideo.ended) {
+        this.setVideoTime(this._currentVideo, 0);
+      }
     }
 
     nextVideo.style.display = 'block';
--- a/static/app/components/replays/videoReplayer.tsx
+++ b/static/app/components/replays/videoReplayer.tsx
@@ -412,7 +412,7 @@ export class VideoReplayer {
     // This is the soon-to-be previous video that needs to be hidden
     if (this._currentVideo) {
       this._currentVideo.style.display = 'none';
-      // resets the soon-to-be previous video to the beginning if it's ended so it starts from the beginning on replay
+      // resets the soon-to-be previous video to the beginning if it's ended so it starts from the beginning on restart
       if (this._currentVideo.ended) {
         this.setVideoTime(this._currentVideo, 0);
       }


static/app/components/replays/videoReplayer.tsx
```diff
@@ -412,6 +412,10 @@ export class VideoReplayer {
     // This is the soon-to-be previous video that needs to be hidden
     if (this._currentVideo) {
       this._currentVideo.style.display = 'none';
+      // resets the soon-to-be previous video to the beginning if it's ended so it starts from the beginning on replay
```
michellewzhang, line 415:
```suggestion
      // resets the soon-to-be previous video to the beginning if it's ended so it starts from the beginning on restart
```

michellewzhang (approved):
did you happen to see this bug on other replays as well? do they also work as expected with this fix?

c298lee:
> did you happen to see this bug on other replays as well? do they also work as expected with this fix?

Yes, you can also see this bug in this replay: https://sentry-sdks.sentry.io/replays/2298b9c818db427cae385baf7945dc5b/ and it works as expected after the fix!

michellewzhang:
> https://sentry-sdks.sentry.io/replays/2298b9c818db427cae385baf7945dc5b/

this replay is weird bc i think it's also suffering from the timestamp being in seconds issue

c298lee:
> > https://sentry-sdks.sentry.io/replays/2298b9c818db427cae385baf7945dc5b/
> 
> this replay is weird bc i think it's also suffering from the timestamp being in seconds issue

Ohh maybe that's why it's buggy when you try to zoom in on the timeline. 
I found another replay with flickering on restart: https://sentry-sdks.sentry.io/replays/f9624587255447bcadd94d913b878d78

michellewzhang:
okay i tried testing with https://sentry-8moqv2a4f.sentry.dev/replays/f9624587255447bcadd94d913b878d78/ and after restart i think the flicker thing is still happening :( (around 15s in the video playback)

https://github.com/user-attachments/assets/b489e56f-ae42-4ce4-9f25-18d5f3b6de1f

c298lee:
> okay i tried testing with https://sentry-8moqv2a4f.sentry.dev/replays/f9624587255447bcadd94d913b878d78/ and after restart i think the flicker thing is still happening :( (around 15s in the video playback)
> 
>  Screen.Recording.2024-07-15.at.12.00.43.PM.mov 
> this is so hard to reproduce though. it sometimes doesn't flicker and sometimes does... i think this bug fix is an overall improvement though. might be good to get @billyvg 's eyes on this but i think we can merge 👍

yea i've noticed that flickering too, it's way more inconsistent tho and doesn't happen in every segment. I'll look into this too 😭

billyvg (approved):
I think this fix is worth getting out even if it doesn't completely fix the issue. We can follow-up with a more complete fix.

PR #74246, opened 2024-07-13
chore(hybrid-cloud): Moves majority of RPC Model TypedDicts to typing_extensions impl

GabeVillalobos:
This PR is in preparation for the Pydantic v2 upgrade. As per the [Pydantic documentation](https://docs.pydantic.dev/latest/api/standard_library_types/#typeddict), I've converted all `TypedDict` references for our RPC models to use the `typing_extensions` package implementation instead of the base `typing` `TypedDict`.

Once we upgrade to Python 3.12+, we should be able to safely use the `typing` package's implementation again.


markstory (approved):
Looks good to me.

PR #74247, opened 2024-07-13
Project Template API - POST

saponifi3d:
# Description
Add ability to create a Project Template in API

## Callouts
- Audit Logging when creating a new project template
- TODO - Rate Limiting, these numbers aren't clear yet so just leaving it as a TODO

### DIFF
--- a/src/sentry/api/endpoints/project_templates_index.py
+++ b/src/sentry/api/endpoints/project_templates_index.py
@@ -3,7 +3,7 @@ from functools import wraps
 from rest_framework.request import Request
 from rest_framework.response import Response
 
-from sentry import features
+from sentry import audit_log, features
 from sentry.api.api_owners import ApiOwner
 from sentry.api.api_publish_status import ApiPublishStatus
 from sentry.api.base import region_silo_endpoint
@@ -15,6 +15,7 @@ from sentry.models.organization import Organization
 from sentry.models.projecttemplate import ProjectTemplate
 
 PROJECT_TEMPLATE_FEATURE_FLAG = "organizations:project-templates"
+AUDIT_LOG_EVENT_ID = "PROJECT_TEMPLATE_CREATED"
 
 
 def ensure_rollout_enabled(flag):
@@ -60,3 +61,37 @@ class OrganizationProjectTemplatesIndexEndpoint(OrganizationEndpoint):
             on_results=lambda x: serialize(x, request.user, ProjectTemplateSerializer()),
             paginator_cls=OffsetPaginator,
         )
+
+    @ensure_rollout_enabled(PROJECT_TEMPLATE_FEATURE_FLAG)
+    def post(self, request: Request, organization: Organization) -> Response:
+        """
+        Create a new Project Template.
+
+        Create a new project template for the organization.
+        """
+        name = request.data.get("name", None)
+
+        if name is None:
+            return Response(status=400, data={"detail": "Template name is required"})
+
+        # TODO - Rate Limiting
+
+        project_template = ProjectTemplate.objects.create(
+            name=name,
+            organization=organization,
+        )
+
+        options = request.data.get("options", None)
+        if options:
+            for key, value in options.items():
+                project_template.options.create(key=key, value=value)
+
+        self.create_audit_entry(
+            request=request,
+            organization=organization,
+            target_object=project_template.id,
+            event=audit_log.get_event_id(AUDIT_LOG_EVENT_ID),
+            data=project_template.get_audit_log_data(),
+        )
+
+        return Response(serialize(project_template, request.user), status=201)
--- a/src/sentry/api/endpoints/project_templates_index.py
+++ b/src/sentry/api/endpoints/project_templates_index.py
@@ -70,9 +70,11 @@ class OrganizationProjectTemplatesIndexEndpoint(OrganizationEndpoint):
     @ensure_rollout_enabled(PROJECT_TEMPLATE_FEATURE_FLAG)
     def post(self, request: Request, organization: Organization) -> Response:
         """
-        Create a new Project Template.
+        Create a new Project Template for the organization.
 
-        Create a new project template for the organization.
+        The Request body should be a JSON object with the following keys:
+        - name: string
+        - options: {key: 'value'} - optional - creates a ProjectTemplateOption for each key-value pair
         """
         serializer = ProjectTemplateWriteSerializer(data=request.data)
         if not serializer.is_valid():
--- a/src/sentry/api/endpoints/project_templates_index.py
+++ b/src/sentry/api/endpoints/project_templates_index.py
@@ -42,7 +42,9 @@ class OrganizationProjectTemplatesIndexEndpoint(OrganizationEndpoint):
 
     publish_status = {
         "GET": ApiPublishStatus.PRIVATE,
+        "POST": ApiPublishStatus.PRIVATE,
     }
+
     permission_classes = (OrganizationPermission,)
 
     @ensure_rollout_enabled(PROJECT_TEMPLATE_FEATURE_FLAG)
--- a/src/sentry/api/endpoints/project_templates_index.py
+++ b/src/sentry/api/endpoints/project_templates_index.py
@@ -76,7 +76,7 @@ class OrganizationProjectTemplatesIndexEndpoint(OrganizationEndpoint):
         if name is None:
             return Response(status=400, data={"detail": "Template name is required"})
 
-        # TODO - Rate Limiting
+        # TODO @saponifi3d - Rate Limiting once we know the limits
 
         project_template = ProjectTemplate.objects.create(
             name=name,
--- a/src/sentry/api/endpoints/project_templates_index.py
+++ b/src/sentry/api/endpoints/project_templates_index.py
@@ -10,7 +10,10 @@ from sentry.api.base import region_silo_endpoint
 from sentry.api.bases.organization import OrganizationEndpoint, OrganizationPermission
 from sentry.api.paginator import OffsetPaginator
 from sentry.api.serializers import serialize
-from sentry.api.serializers.models.project_template import ProjectTemplateSerializer
+from sentry.api.serializers.models.project_template import (
+    ProjectTemplateSerializer,
+    ProjectTemplateWriteSerializer,
+)
 from sentry.models.organization import Organization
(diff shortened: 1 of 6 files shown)


src/sentry/api/endpoints/project_templates_index.py
```diff
@@ -41,7 +42,9 @@ class OrganizationProjectTemplatesIndexEndpoint(OrganizationEndpoint):
 
     publish_status = {
         "GET": ApiPublishStatus.PRIVATE,
+        "POST": ApiPublishStatus.PRIVATE,
```
schew2381, line 45: Should we change this to experimental for now? I feel like it could be useful to publish these in the future

src/sentry/api/endpoints/project_templates_index.py
```diff
@@ -60,3 +63,37 @@ def get(self, request: Request, organization: Organization) -> Response:
...
         )
+
+    @ensure_rollout_enabled(PROJECT_TEMPLATE_FEATURE_FLAG)
+    def post(self, request: Request, organization: Organization) -> Response:
+        """
+        Create a new Project Template.
+
+        Create a new project template for the organization.
+        """
+        name = request.data.get("name", None)
```
schew2381, line 74: I'm guessing project template will be have more body parameters eventually, would it be useful to move `name` and `options` into a serializer to setup for that?

tests/sentry/api/endpoints/test_project_templates_index.py
```diff
@@ -55,3 +58,70 @@ def test_get__no_access(self):
...
+
+
+class ProjectTemplateIndexPostTest(APITestCase):
+    endpoint = "sentry-api-0-organization-project-templates"
+
+    def setUp(self):
+        super().setUp()
+        self.user = self.create_user()
+        self.org = self.create_organization(owner=self.user)
+        self.team = self.create_team(organization=self.org, members=[self.user])
```
schew2381, line 70: nit: there's properties (on the base fixtures I think) that create these automatically for you

tests/sentry/api/endpoints/test_project_templates_index.py
```diff
@@ -55,3 +58,70 @@ def test_get__no_access(self):
...
+
+        self.project_template = self.create_project_template(organization=self.org)
+        self.project_template_two = self.create_project_template(organization=self.org)
+
+    def test_post__no_feature(self):
+        response = self.get_error_response(self.org.id, method="POST", status_code=404)
+        assert response.status_code == 404
+
+    @with_feature(PROJECT_TEMPLATE_FEATURE_FLAG)
+    def test_post__non_admin(self):
```
schew2381, line 82: nit: I think the test should be called `as_member` b/c there are multiple org roles (admin, manager, owner) that have `org:write` or `org:admin` to hit the POST

tests/sentry/api/endpoints/test_project_templates_index.py
saponifi3d: fyi - this didnt work for me, looking a little closer at the helpers it's cause `self.org` vs `self.organization` and the create_team helper doesn't add any members. i'll update the name for `org` => `organization` but `members` is correct.

src/sentry/api/endpoints/project_templates_index.py
saponifi3d: Just curious, what's the difference between them other than semantics? I feel like for the time being, this should likely stay private until we know that we want to eventually make this private. generally easier to roll stuff out than back.

src/sentry/api/endpoints/project_templates_index.py
saponifi3d: fair, i was being lazy 🤣

src/sentry/api/endpoints/project_templates_index.py
```diff
@@ -60,3 +66,35 @@ def get(self, request: Request, organization: Organization) -> Response:
             on_results=lambda x: serialize(x, request.user, ProjectTemplateSerializer()),
             paginator_cls=OffsetPaginator,
         )
+
+    @ensure_rollout_enabled(PROJECT_TEMPLATE_FEATURE_FLAG)
+    def post(self, request: Request, organization: Organization) -> Response:
+        """
+        Create a new Project Template.
+
+        Create a new project template for the organization.
```
ceorourke, line 75: Is it intentional to have this twice?

saponifi3d: NGL, ai wrote this. hahah i'll clean it up.

tests/sentry/api/endpoints/test_project_templates_index.py
```diff
@@ -55,3 +58,70 @@ def test_get__no_access(self):
...
+    def setUp(self):
+        super().setUp()
+        self.user = self.create_user()
+        self.organization = self.create_organization(owner=self.user)
+        self.team = self.create_team(members=[self.user])
+
+        self.login_as(self.user)
+
+        self.project_template = self.create_project_template(organization=self.organization)
+        self.project_template_two = self.create_project_template(organization=self.organization)
```
ceorourke, line 75: Since this is the same as the other test class could you put the setup in a base that they both inherit from?

tests/sentry/api/endpoints/test_project_templates_index.py
```diff
@@ -55,3 +58,70 @@ def test_get__no_access(self):
...
+
+    def test_post__no_feature(self):
+        response = self.get_error_response(self.organization.id, method="POST", status_code=404)
+        assert response.status_code == 404
+
+    @with_feature(PROJECT_TEMPLATE_FEATURE_FLAG)
+    def test_post__as_member(self):
```
ceorourke, line 82: I think this could benefit from a docstring - is this a 403 because the user isn't an orgmember?

tests/sentry/api/endpoints/test_project_templates_index.py
```diff
@@ -55,3 +58,70 @@ def test_get__no_access(self):
...
+        )
+
+        assert response.status_code == 201
+        new_template = ProjectTemplate.objects.get(id=response.data["id"])
+        created_options = {opt.key: opt.value for opt in new_template.options.all()}
+        assert created_options == test_options
+
+    @with_feature(PROJECT_TEMPLATE_FEATURE_FLAG)
+    def test_post__no_name(self):
+        response = self.get_error_response(self.organization.id, method="POST", status_code=400)
```
ceorourke, line 114: what kinds of names do we expect people to come up with for these? could it possibly benefit from autogenerated names?

saponifi3d: there's a test helper that creates and uses auto generated names. so basically everything that uses `self.create_template` kinda has that baked in 🎉

tests/sentry/api/endpoints/test_project_templates_index.py
saponifi3d: haha i had this as `non_admin` before but seiji asked me to change it to this.. i wanted to just convey that they don't have the correct permissions.

as far as the doc string.. idk, i feel like it's not sure helpful (am i the only dev that doesn't read inline docs? 😅). but since both you and seiji called this one out explicitly i feel like it's not clear and we should do something...

i wonder if it's something like `test_post__no_access` or something simple like that?

ceorourke: lol I read them 😢 maybe `test_post__permissions`?

saponifi3d: hahah okokok, i can add a quick note to outline the setup and how it's working.

for the name, i feel like permissions is kinda a piece of what i'm testing, but not the whole aspect to it.. it's like `test_post__as_member_without_permission`?

src/sentry/api/endpoints/project_templates_index.py
saponifi3d: was able to find a simpler example and updated. thanks for the push 🎉

tests/sentry/api/endpoints/test_project_templates_index.py
```diff
@@ -55,3 +60,61 @@ def test_get__no_access(self):
...
+        response = self.get_success_response(self.organization.id, name="Test Project Template")
+        assert response.status_code == 201
+
+    @with_feature(PROJECT_TEMPLATE_FEATURE_FLAG)
+    def test_post__as_member_without_permission(self):
+        """
+        Test that a member is part of the organization, but does not have the required
+        permissions to create a project template.
+
+        The user is a member of the organization, but does not have write access to the org.
```
ceorourke, line 83: ai 👀

ceorourke (approved):

src/sentry/api/endpoints/project_templates_index.py
schew2381: The enum comments are like:
https://github.com/getsentry/sentry/blob/d59dea9380eb57def2569285dedefe81e4ea425b/src/sentry/api/api_publish_status.py#L11-L13

Experimental also doesn't publish anything like private

tests/sentry/api/endpoints/test_project_templates_index.py
schew2381: I took a quick look at the `self.team` property, is looks like `self.user` should be included as a team member b/c it the user should be a part of `self.organization`.
https://github.com/getsentry/sentry/blob/dcec0930c33979f88929787fc1b43badf89f0df6/src/sentry/testutils/fixtures.py#L66-L71

src/sentry/api/serializers/models/project_template.py
```diff
@@ -63,3 +65,12 @@ def serialize(
...
         return response
+
+
+class ProjectTemplateWriteSerializer(serializers.Serializer):
+    id = serializers.IntegerField(required=False)
+    name = serializers.CharField(required=True)
+    options = serializers.DictField(required=False)
+
+    def validate(self, data):
+        return data
```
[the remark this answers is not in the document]
saponifi3d: sure, but do you have a recommendation or suggestion or anything? (it's also a pattern i saw in most serializers)

tests/sentry/api/endpoints/test_project_templates_index.py
saponifi3d: _shakes fist_ no fair!! the user wasn't part of the org when i left the comment! it changed when i created the base class for `setUp` for the comment colleen left.

PR #73869, opened 2024-07-05
feat(stats): Display subseries in tooltip

priscilawebdev:
~This change was previously approved in the [PR](https://github.com/getsentry/sentry/pull/73774), but we decided to revert it in order to merge it together with the [new doc changes](https://github.com/getsentry/sentry-docs/pull/10585)~

#### Goal

Debugging isn't the main focus of the stats page, but users can still use the information there for debugging. Based on [feedback](https://github.com/getsentry/sentry-docs/pull/10585#discussion_r1668595013), this PR makes the following changes to help with debugging:

1. **Showing `client_discard` Outcome for Debugging**:
   - The `client_discard` outcome is now available for debugging, but it is initially disabled in the chart because it can be very frequent due to sampling.

2. **Separating `invalid` Outcome and Updating the UI**:
   - The code now separates the `invalid` outcome from `abuse`, `cardinality_limited`, and `rate_limited`.
   - The UI now shows `invalid` as a separate outcome, with its own card at the top to highlight its importance.

3. **Replacing `dropped` with `rate_limited`**:
   - We have replaced `dropped` (which wasn't a real outcome) with `rate_limited`.
   - The `rate_limited` outcome now includes `rate_limited`, `abuse`, and `cardinality_limited` outcomes combined.

**Preview**

image

#### Note

We are working on the documentation in the [PR](https://github.com/getsentry/sentry-docs/pull/10585) 

requires https://github.com/getsentry/getsentry/pull/14653

PR #74248, opened 2024-07-15
feat(stats): Add switch button field for client discard legend

priscilawebdev:
### Goal

The "Client Discard" outcome is now shown in the stats chart. Since it can have a high volume, we want to disable it by default so customers can focus on other stats. See [feedback](https://github.com/getsentry/sentry/pull/73869#discussion_r1673847988).

While users can enable or disable it by clicking on the chart's legend, this option is not very noticeable. To make it easier, we are adding a more prominent switch button at the bottom of the chart.

This switch button updates the URL with the 'clientDiscard' parameter, which controls whether the chart shows or hides the outcome.

**Preview**

https://github.com/user-attachments/assets/d1182ac7-7e6c-49e7-945b-a592f268ca23

### Note

Customers are still able to hide/show the client discard outcome by clicking on the chart's legend

### DIFF
--- a/static/app/views/organizationStats/index.tsx
+++ b/static/app/views/organizationStats/index.tsx
@@ -344,7 +344,12 @@ export class OrganizationStats extends Component<OrganizationStatsProps> {
   }
 }
 
-export default withPageFilters(withOrganization(OrganizationStats));
+const HookOrgStats = HookOrDefault({
+  hookName: 'component:enhanced-org-stats',
+  defaultComponent: OrganizationStats,
+});
+
+export default withPageFilters(withOrganization(HookOrgStats));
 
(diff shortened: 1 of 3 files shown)


obostjancic (approved):

jan-auer (approved):

matejminar:
I find it a bit strange that the toggle turns on/off the visibility of the series, which is reflected in the legend, but not the other way around. Is that intended?
If changing the toggle value when clicking on the legend is hard to achieve, the toggle could remove that series completely.

priscilawebdev:
@matejminar it shall be fine now -> https://github.com/getsentry/sentry/pull/74248/commits/8abde32dfec6dd9c0af681b2a968db72f3ad4877

matejminar (approved):

PR #74272, opened 2024-07-15
ref(metrics) Filter out queryable operations from metrics/meta endpoint

### LINKED ISSUE #74257: Filter non queryable aggregates in metrics/meta endpoint
```
"histogram",
"max_timestamp",
"min_timestamp",
```

vgrozdanic:
Filters out non queryable operations from metrics/meta endpoint

Closes: https://github.com/getsentry/sentry/issues/74257


src/sentry/sentry_metrics/querying/metadata/metrics.py
```diff
@@ -136,11 +136,13 @@ def _build_metric_meta(
     blocking_status: Sequence[BlockedMetric],
     operations_config: OperationsConfiguration,
 ) -> MetricMeta:
+    NON_QUERYABLE_METRIC_OPERATIONS = {"histogram", "min_timestamp", "max_timestamp"}
```
shellmayr, line 139: Could we add this to the OperationsConfiguration object?

vgrozdanic: We could, but it seemed more fitting here, because this limit is only for custom metrics, and in theory we could reuse OperationsConfiguration for some other parts of the product.

If we add it to OperationsConfiguration, then we would attach (in my opinion) unnecessary business logic to it, and so far it is free of any business logic

shellmayr (approved):
Good stuff 🚢🚢🚢

PR #74277, opened 2024-07-15
metrics(replay): log rrweb_event_count to slow clicks table

aliu39:
@billyvg helps us query for a distribution chart of # of events in a replay

### DIFF
--- a/src/sentry/replays/usecases/ingest/__init__.py
+++ b/src/sentry/replays/usecases/ingest/__init__.py
@@ -264,10 +264,6 @@ def recording_post_processor(
 
         # Log # of rrweb events to bigquery.
         logger.info(
-            # Logging to the sentry.replays.slow_click namespace because
-            # its the only one configured to use BigQuery at the moment.
-            #
-            # NOTE: Needs an ops request if we want to create a new dataset.
             "sentry.replays.slow_click",
             extra={
                 "event_type": "rrweb_event_count",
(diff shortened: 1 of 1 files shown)


cmanallen (approved):
The comment inside the info block can be removed but otherwise LGTM.

src/sentry/replays/usecases/ingest/__init__.py
```diff
@@ -261,6 +261,23 @@ def recording_post_processor(
             message.replay_id,
             parsed_segment_data,
         )
+
+        # Log # of rrweb events to bigquery.
+        logger.info(
+            # Logging to the sentry.replays.slow_click namespace because
```
aliu39, line 267: Copied from log_canvas_size. What does the ops request comment mean, exactly?

src/sentry/replays/usecases/ingest/__init__.py
```diff
@@ -261,6 +261,23 @@ def recording_post_processor(
...
+            # its the only one configured to use BigQuery at the moment.
+            #
+            # NOTE: Needs an ops request if we want to create a new dataset.
```
aliu39, line 270: What's the ops request comment mean? @cmanallen

cmanallen: You can ignore it.  Its a note about the destination table name: `"sentry.replays.slow_click"`.  It can be removed.

PR #74278, opened 2024-07-15
chore(similarity): Change delete feature to use feature handler

jangjodi:
Change delete feature to use feature handler


jangjodi:
closing this cause we're going to remove the flag and use the same one as ingestion

PR #74280, opened 2024-07-15
feat(uptime): Handle uptime check results when subscription is in onboarding mode

wedamija:
This handles the lifecycle of a subscription when it's in onboarding mode. If successes are received, they're ignored unless the subscription has existed for long enough. If failures are received, we keep track of the number of failures, and if we hit too many we kill the subscription and disable checking the url for a while.

### DIFF
--- a/src/sentry/uptime/consumers/results_consumer.py
+++ b/src/sentry/uptime/consumers/results_consumer.py
@@ -31,23 +31,27 @@ from sentry.uptime.subscriptions.subscriptions import (
     remove_uptime_subscription_if_unused,
 )
 from sentry.utils import metrics
-from sentry.utils.hashlib import md5_text
 
 logger = logging.getLogger(__name__)
 LAST_UPDATE_REDIS_TTL = timedelta(days=7)
-ONBOARDING_FAILURE_THRESHOLD = 3
-ONBOARDING_FAILURE_REDIS_TTL = timedelta(days=2)
 ONBOARDING_MONITOR_PERIOD = timedelta(days=3)
+# When onboarding a new subscription how many total failures are allowed to happen during
+# the ONBOARDING_MONITOR_PERIOD before we consider the subscription to have failed onboarding.
+ONBOARDING_FAILURE_THRESHOLD = 3
+# The TTL of the redis key used to track the failure counts for a subscription in
+# `ProjectUptimeSubscriptionMode.AUTO_DETECTED_ONBOARDING` mode. Must be >= the
+# ONBOARDING_MONITOR_PERIOD.
+ONBOARDING_FAILURE_REDIS_TTL = ONBOARDING_MONITOR_PERIOD
 # How frequently we should run active auto-detected subscriptions
-AUTO_DETECTED_ACTIVE_SUBSCRIPTION_INTERVAL_SECONDS = int(timedelta(minutes=5).total_seconds())
+AUTO_DETECTED_ACTIVE_SUBSCRIPTION_INTERVAL = timedelta(minutes=5)
 
 
 def build_last_update_key(project_subscription: ProjectUptimeSubscription) -> str:
-    return f"project-sub-last-update:{md5_text(project_subscription.id).hexdigest()}"
+    return f"project-sub-last-update:{project_subscription.id}"
 
 
 def build_onboarding_failure_key(project_subscription: ProjectUptimeSubscription) -> str:
-    return f"project-sub-onboarding_failure:{md5_text(project_subscription.id).hexdigest()}"
+    return f"project-sub-onboarding_failure:{project_subscription.id}"
 
 
 class UptimeResultProcessor(ResultProcessor[CheckResult, UptimeSubscription]):
@@ -156,14 +160,16 @@ class UptimeResultProcessor(ResultProcessor[CheckResult, UptimeSubscription]):
                 metrics.incr("uptime.result_processor.autodetection.failed_onboarding")
         elif result["status"] == CHECKSTATUS_SUCCESS:
             assert project_subscription.date_added is not None
-            if (
-                datetime.now(timezone.utc) - ONBOARDING_MONITOR_PERIOD
-                > project_subscription.date_added
-            ):
+            scheduled_check_time = datetime.fromtimestamp(
+                result["scheduled_check_time_ms"] / 1000, timezone.utc
+            )
+            if scheduled_check_time - ONBOARDING_MONITOR_PERIOD > project_subscription.date_added:
+                # If we've had mostly successes throughout the onboarding period then we can graduate the subscription
+                # to active.
                 onboarding_subscription = project_subscription.uptime_subscription
                 active_subscription = create_uptime_subscription(
                     onboarding_subscription.url,
-                    AUTO_DETECTED_ACTIVE_SUBSCRIPTION_INTERVAL_SECONDS,
+                    int(AUTO_DETECTED_ACTIVE_SUBSCRIPTION_INTERVAL.total_seconds()),
                     onboarding_subscription.timeout_ms,
                 )
                 project_subscription.update(
--- a/src/sentry/uptime/consumers/results_consumer.py
+++ b/src/sentry/uptime/consumers/results_consumer.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
 import logging
-from datetime import timedelta
+from datetime import datetime, timedelta, timezone
 
 from django.conf import settings
 from sentry_kafka_schemas.schema_types.uptime_results_v1 import (
     CHECKSTATUS_FAILURE,
     CHECKSTATUS_MISSED_WINDOW,
+    CHECKSTATUS_SUCCESS,
     CheckResult,
 )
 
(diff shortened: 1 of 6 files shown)


src/sentry/uptime/consumers/results_consumer.py
```diff
@@ -17,19 +18,38 @@
...
+# How frequently we should run active auto-detected subscriptions
+AUTO_DETECTED_ACTIVE_SUBSCRIPTION_INTERVAL_SECONDS = int(timedelta(minutes=5).total_seconds())
 
 
 def build_last_update_key(project_subscription: ProjectUptimeSubscription) -> str:
     return f"project-sub-last-update:{md5_text(project_subscription.id).hexdigest()}"
 
 
+def build_onboarding_failure_key(project_subscription: ProjectUptimeSubscription) -> str:
+    return f"project-sub-onboarding_failure:{md5_text(project_subscription.id).hexdigest()}"
```
evanpurkhiser, line 50: Why med5 this? Can't we just use the hex value of the uuid?
```suggestion
    return f"project-sub-onboarding_failure:{project_subscription.id.hex}"
```

src/sentry/uptime/consumers/results_consumer.py
```diff
@@ -92,8 +116,13 @@ def handle_result_for_project(
                     extra={"project_id": project_subscription.project_id, **result},
                 )
 
-            if result["status"] == CHECKSTATUS_FAILURE:
-                create_issue_platform_occurrence(result, project_subscription)
+            if project_subscription.mode == ProjectUptimeSubscriptionMode.AUTO_DETECTED_ONBOARDING:
+                self.handle_result_for_project_auto_onboarding_mode(project_subscription, result)
+            elif project_subscription.mode in (
+                ProjectUptimeSubscriptionMode.AUTO_DETECTED_ACTIVE,
+                ProjectUptimeSubscriptionMode.MANUAL,
```
evanpurkhiser, line 123: Does this work or do you need to do `.value`?

wedamija: Yeah with `IntEnum` we don't need to use `.value`

src/sentry/uptime/consumers/results_consumer.py
wedamija: I'm just generally in the habit of md5ing things, but this will work too

wedamija: Wait no - this is just an int, not the uuid

evanpurkhiser: oh, can we just use the int then?

src/sentry/uptime/consumers/results_consumer.py
evanpurkhiser: python is the worst 😂

src/sentry/uptime/consumers/results_consumer.py
```diff
@@ -17,19 +18,38 @@
...
+    delete_project_uptime_subscription,
+    remove_uptime_subscription_if_unused,
+)
 from sentry.utils import metrics
 from sentry.utils.hashlib import md5_text
 
 logger = logging.getLogger(__name__)
 LAST_UPDATE_REDIS_TTL = timedelta(days=7)
+ONBOARDING_FAILURE_THRESHOLD = 3
+ONBOARDING_FAILURE_REDIS_TTL = timedelta(days=2)
```
evanpurkhiser, line 39: Shouldn't this be longer than `ONBOARDING_MONITOR_PERIOD`? If we have two failures day 1 and then 1 failure in the very last day, shouldn't that actually mean we met the threshold, but because of the expiry we won't have?

src/sentry/uptime/consumers/results_consumer.py
```diff
@@ -17,19 +18,38 @@
...
+ONBOARDING_MONITOR_PERIOD = timedelta(days=3)
+# How frequently we should run active auto-detected subscriptions
+AUTO_DETECTED_ACTIVE_SUBSCRIPTION_INTERVAL_SECONDS = int(timedelta(minutes=5).total_seconds())
```
evanpurkhiser, line 42: Can we add some comments?

```suggestion
ONBOARDING_FAILURE_THRESHOLD = 3
"""
When onboarding a new subscription how many TOTAL failures are allowed to happen during
the ONBOARDING_MONITOR_PERIOD before we consider the subscription to have failed onboarding.
"""

ONBOARDING_MONITOR_PERIOD = timedelta(days=3)
"""
How long does a subscription need to be in the `ProjectUptimeSubscriptionMode.AUTO_DETECTED_ONBOARDING` 
mode before it can graduate to `ProjectUptimeSubscriptionMode.AUTO_DETECTED_ACTIVE`?
During this time the subscription must not fail a check more than `ONBOARDING_FAILURE_THRESHOLD` times.
"""

# The TTL of the redis key used to track the failure counts for a subscription
# In `ProjectUptimeSubscriptionMode.AUTO_DETECTED_ONBOARDING` mode. Must be longer
# Than the ONBOARDING_MONITOR_PERIOD.
ONBOARDING_FAILURE_REDIS_TTL = ONBOARDING_MONITOR_PERIOD + timedelta(days=1)

AUTO_DETECTED_ACTIVE_SUBSCRIPTION_INTERVAL_SECONDS = int(timedelta(minutes=5).total_seconds())
"""
How frequently are active auto-detected subscriptions configured to run.
""" 
```

(feel free if you want to use normal comments, I just like the docstring comments since pyright shows them as editor hints on hover)

evanpurkhiser, line 42: IMO we should leave this as a timedelta here and convert closer to the usage site. It makes the configuration a bit easier to read, and at the call site it's just as obvious that we're passing seconds.

```suggestion
AUTO_DETECTED_ACTIVE_SUBSCRIPTION_INTERVAL = timedelta(minutes=5)
```

farther down it needs to be

```py
int(AUTO_DETECTED_ACTIVE_SUBSCRIPTION_INTERVAL.total_seconds())
```

src/sentry/uptime/consumers/results_consumer.py
wedamija: We bump the expiry on every result, so this should be fine unless we have a really extended downtime

evanpurkhiser: Seems like we only bump it on failures?

src/sentry/uptime/consumers/results_consumer.py
```diff
@@ -104,6 +133,52 @@ def handle_result_for_project(
...
+                    modes=[ProjectUptimeSubscriptionMode.AUTO_DETECTED_ONBOARDING],
+                )
+                # Mark the url as failed so that we don't attempt to auto-detect it for a while
+                set_failed_url(project_subscription.uptime_subscription.url)
+                redis.delete(key)
+                metrics.incr("uptime.result_processor.autodetection.failed_onboarding")
+        elif result["status"] == CHECKSTATUS_SUCCESS:
+            assert project_subscription.date_added is not None
+            if (
+                datetime.now(timezone.utc) - ONBOARDING_MONITOR_PERIOD
```
evanpurkhiser, line 160: you can also use `timezone.now()` from django, idk what we prefer

src/sentry/uptime/consumers/results_consumer.py
```diff
@@ -104,6 +133,52 @@ def handle_result_for_project(
...
+                > project_subscription.date_added
+            ):
```
evanpurkhiser, line 162: feels like it could be nice to name this condition
```suggestion
            has_graduated_onboarding = datetime.now(timezone.utc) - ONBOARDING_MONITOR_PERIOD
                > project_subscription.date_added

            if has_graduated_onboarding:
```

src/sentry/uptime/consumers/results_consumer.py
wedamija: Good point, I'll just bump it to 3 days

src/sentry/uptime/consumers/results_consumer.py
wedamija: Yeah that's probably preferable, I can't keep all the different ways of creating dates straight these days

src/sentry/uptime/consumers/results_consumer.py
evanpurkhiser: I called it out in the comment here https://github.com/getsentry/sentry/pull/74280#discussion_r1678166311

I gave it +1 day, but there's probably no reason, it can just be the same as `ONBOARDING_MONITOR_PERIOD`

src/sentry/uptime/subscriptions/subscriptions.py
```diff
@@ -67,21 +67,29 @@ def create_project_uptime_subscription(
...
+        mode__in=modes,
+    ):
         uptime_project_subscription.delete()
 
+    remove_uptime_subscription_if_unused(uptime_subscription)
+
+
+def remove_uptime_subscription_if_unused(uptime_subscription: UptimeSubscription):
+    """
+    Determines if an uptime subscription is no longer used and removes it if so
```
evanpurkhiser, line 91:
```suggestion
    Determines if an uptime subscription is no longer used by any project and removes it if so.
```

Adding `by any project` helps to hint people (who may have forgotten like me) that subscriptions are shared between projects.

tests/sentry/uptime/consumers/test_results_consumers.py
```diff
@@ -119,3 +129,138 @@ def test_missed(self):
...
+                [
+                    call(
+                        "uptime.result_processor.handle_result_for_project",
+                        tags={"status": CHECKSTATUS_FAILURE, "mode": "auto_detected_onboarding"},
+                    ),
+                ]
+            )
+        assert redis.get(key) == "1"
+
+        hashed_fingerprint = md5(str(self.project_subscription.id).encode("utf-8")).hexdigest()
```
evanpurkhiser, line 157: I wish there was a function that did this lol. Can't use `md5_text` here?

src/sentry/uptime/subscriptions/subscriptions.py
```diff
@@ -67,21 +67,29 @@ def create_project_uptime_subscription(
...
+        project=project,
+        uptime_subscription=uptime_subscription,
+        mode__in=modes,
+    ):
         uptime_project_subscription.delete()
 
+    remove_uptime_subscription_if_unused(uptime_subscription)
+
+
+def remove_uptime_subscription_if_unused(uptime_subscription: UptimeSubscription):
```
evanpurkhiser, line 89: nit: feels like this could be a different PR to reduce the size of this one

evanpurkhiser (commented):
Looking mostly good. I would like to see comments on those constants I think, but other than that looks good to go

tests/sentry/uptime/consumers/test_results_consumers.py
wedamija: Yeah maybe I should make a function for this since we use it a bunch, or build the check into some framework at some point.

src/sentry/uptime/consumers/results_consumer.py
wedamija: Actually, this should be the time the result was generated, not the wall clock time. Doesn't matter a lot probably, but it's more correct

evanpurkhiser: Makes sense

src/sentry/uptime/consumers/results_consumer.py
wedamija: Not sure if there's a lot of benefit here if it's used once, I can include a comment if the code is unclear though

evanpurkhiser: IMO the benifit is that the name documents it, so it's quicker to understand the condition.

I can just read

`if has_graduated_onboarding` vs having to understand the actual condition. It also gives a hint as to the value of `datetime.now(timezone.utc) - ONBOARDING_MONITOR_PERIOD > project_subscription.date_added` so when I'm reading that I already have a context clue for what I'm looking at.

A comment is fine too though sure

tests/sentry/uptime/consumers/test_results_consumers.py
wedamija: Can't change this unfortunately since it needs to match the method we use in issue platform

evanpurkhiser (approved):

PR #74281, opened 2024-07-15
ref(apis): promote get_appropriate_user to module level

ameliahsu:
resolving https://github.com/getsentry/sentry/pull/74105#discussion_r1676302422

moved `get_appropriate_user` function to the module level


src/sentry/api/base.py
```diff
@@ -582,6 +583,25 @@ def paginate(
         self.add_cursor_headers(request, response, cursor_result)
         return response
 
+    def get_appropriate_user_id(self, request: Request) -> int:
```
iamrajjoshi, line 586: is this method used by other endpoints? if not, maybe we can move it to `api_tokens.py` b/c `base.py` already has a lot of logic that lives there.

ameliahsu: fixed!

iamrajjoshi (approved):

PR #74173, opened 2024-07-11
ref(replay): replace replays/contextIcon import with PlatformIcon

aliu39:
Fixes https://github.com/getsentry/sentry/issues/73945

Bumping to the latest version of [the platformicons package](https://www.npmjs.com/package/platformicons?activeTab=code), since the current one doesn't have Chrome, Safari, and maybe some others.

I kinda eye-balled the size, when using 14 the new icons were clearly smaller than the old.
ContextIcon, 14px:
Screenshot 2024-07-12 at 10 29 24 AM
PlatformIcon (NEW), 18px: Didn't translate the name for edge here, but it should work in the preview deployment
Screenshot 2024-07-12 at 10 29 49 AM

TODO:
- [x] translate "mac-os-x" to "apple", and check other names.
    - [x] can maybe write a new function to replace `events/contexts/utils/generateIconName`
- [ ] use lazy loading like before. Is this necessary? I was getting type errors in the lazy() statement earlier

PR #74284, opened 2024-07-15
Aliu/context icon lazy

aliu39:
Forked from https://github.com/getsentry/sentry/pull/74173. The problem right now is none of the browser icons are loading, maybe because of the way I do the import? 
Screenshot 2024-07-15 at 10 03 51 AM


static/app/components/replays/contextIcon.tsx
```diff
@@ -16,18 +17,23 @@ type Props = {
   showVersion?: boolean;
 };
 
-const LazyContextIcon = lazy(
-  () => import('sentry/components/events/contexts/contextIcon')
-);
+async function loadPlatformIcon() {
```
aliu39, line 20: `import('platformicons')` has type 
Screenshot 2024-07-15 at 10 01 35 AM
so, I had to try this

PR #74285, opened 2024-07-15
feat(crons): Add new notification type

davidenwang:
Right now crons is borrowing from the "Nudges" category for broken monitor notifications, but we have received feedback from multiple sources that we should just separate into our own category for clarity.

### DIFF
--- a/src/sentry/notifications/types.py
+++ b/src/sentry/notifications/types.py
@@ -33,6 +33,7 @@ class NotificationSettingEnum(ValueEqualityEnum):
     SPIKE_PROTECTION = "spikeProtection"
     MISSING_MEMBERS = "missingMembers"
     REPORTS = "reports"
+    BROKEN_MONITORS = "brokenMonitors"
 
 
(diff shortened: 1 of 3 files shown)


src/sentry/notifications/types.py
```diff
@@ -33,6 +33,7 @@ class NotificationSettingEnum(ValueEqualityEnum):
     SPIKE_PROTECTION = "spikeProtection"
     MISSING_MEMBERS = "missingMembers"
     REPORTS = "reports"
+    CRON_MONITORS = "cronMonitors"
```
evanpurkhiser, line 36: Is it going to be confusing to customers that cron monitor notifications are JUST the broken notices and NOT the issues that get created?

davidenwang: dang yea you're right, lemme change the name to BROKEN_MONITORS, and update the help text to be very explicit here

evanpurkhiser (approved):

PR #74288, opened 2024-07-15
ref(metric-extraction): Clean up project threshold code

k-fish:
### Summary 
Project thresholds are ignored in on-demand in lieu of the dashboard always providing the actual number for apdex / user_misery functions. This also adds a function to be used by later caching work to check whether a MetricSpec is actually project dependent (if it uses the project metric for user_misery and apdex calculations).

### DIFF
--- a/src/sentry/snuba/metrics/extraction.py
+++ b/src/sentry/snuba/metrics/extraction.py
@@ -1621,7 +1621,7 @@ def _get_satisfactory_metric(project: Project) -> str:
     if len(result) == 0:
         metric = TransactionMetric.DURATION.value
     else:
-        metric = result[0]
+        metric = result[0][0]
 
     if metric == TransactionMetric.DURATION.value:
         metric_field = "transaction.duration"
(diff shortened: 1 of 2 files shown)


src/sentry/snuba/metrics/extraction.py
```diff
@@ -1593,24 +1607,21 @@ def _map_field_name(search_key: str) -> str:
...
 
     if len(result) == 0:
-        # We use the default threshold shown in the UI.
-        threshold = DEFAULT_PROJECT_THRESHOLD
         metric = TransactionMetric.DURATION.value
     else:
-        # We technically don't use this threshold since we extract it from the apdex(x) field
-        # where x is the threshold, however, we still return it in case a fallback is needed.
-        threshold, metric = result[0]
+        (metric,) = result[0]
```
Zylphrex, line 1624:
```suggestion
        metric = result[0]
```

Zylphrex (approved):

obostjancic (approved):

src/sentry/snuba/metrics/extraction.py
k-fish: Good catch thanks Tony

PR #74292, opened 2024-07-15
fix(flagpole): Makes system:multi-region flag internal only

GabeVillalobos:
This feature is being incorrectly checked as a remote flag, fix this to be internal only.


dashed (approved):

GabeVillalobos:
Tested this locally and verified it reads from config, as it should. I'll verify with @markstory that we don't intend this to be remote when he's back in.

PR #74295, opened 2024-07-15
fix(breadcrumbs): Close global drawer when clicking outside 

roggenkemper:
this pr updates the functionality of the breadcrumbs drawer so that it closes when a user clicks outside


malwilley:
FYI, Leander disabled this functionality on purpose here: https://github.com/getsentry/sentry/pull/74143

I'd make sure to run this in getsentry and test the feedback button to make sure the behavior is okay before merging this

PR #74296, opened 2024-07-15
ref(similarity): Use same feature flag for deletion

jangjodi:
Switch from using `projects:similarity-embeddings-delete-by-hash` flag to `projects:similarity-embeddings-grouping` flag so that it can be called in ingestion and to LA this feature
Add feature level killswitch check

Will remove old flag from code in follow up PR once the option is removed

### DIFF
--- a/src/sentry/tasks/delete_seer_grouping_records.py
+++ b/src/sentry/tasks/delete_seer_grouping_records.py
@@ -34,7 +34,7 @@ def delete_seer_grouping_records_by_hash(
     Task to delete seer grouping records by hash list.
     Calls the seer delete by hash endpoint with batches of hashes of size `BATCH_SIZE`.
     """
-    if killswitch_enabled(project_id):
+    if killswitch_enabled(project_id, check_delete_killswitch=True):
         return
 
     batch_size = options.get("embeddings-grouping.seer.delete-record-batch-size")
@@ -55,7 +55,7 @@ def call_delete_seer_grouping_records_by_hash(
     if (
         project
         and features.has("projects:similarity-embeddings-grouping", project)
-        and not killswitch_enabled(project.id)
+        and not killswitch_enabled(project.id, check_delete_killswitch=True)
     ):
         # TODO (jangjodi): once we store seer grouping info in GroupHash, we should filter by that here
         group_hash_objects = GroupHash.objects.filter(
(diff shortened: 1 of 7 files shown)


src/sentry/tasks/delete_seer_grouping_records.py
```diff
@@ -83,7 +86,9 @@ def call_seer_delete_project_grouping_records(
     *args: Any,
     **kwargs: Any,
 ) -> None:
-    if killswitch_enabled(project_id):
+    if killswitch_enabled(project_id) or options.get(
+        "seer.similarity-embeddings-delete-by-hash-killswitch.enabled"
+    ):
```
lobsterkatie, line 91: Any reason we couldn't add the second killswitch check to `killswitch_enabled`? IMHO it would be cleaner if we can - you'd only need to add it in one place.

lobsterkatie: Also, the purist in me says the killswitch bit should technically be in a separate PR, but I don't think it's a huge deal.

jangjodi: the `killswitch_enabled` function is used for the seer grouping changes, whereas this separate killswitch check if just for the deletion - do you think it'd make sense to add it to the function and run it based on an input parameter?

lobsterkatie: Yeah, my instinct would be to be able to pass in a project id or feature name or both and have it check whatever killswitches it needs to.

lobsterkatie: If we're going to have feature-specific (which is maybe just endpoint-specific?) killswitches, what I meant was more something like this:

```python
def killswitch_enabled(
    project_id: int, event: Event | None = None, feature: str | None = None
) -> bool:

	# <global and project killswitch checks>

	killswitches_by_feature = {
	    "delete-seer-records": "seer.similarity-embeddings-delete-by-hash-killswitch.enabled"
	}

	if feature and feature in killswitches_by_feature and options.get(killswitches_by_feature[feature]):
		# <logging and/or metrics>
		return True

	# <whatever other checks we might add>

	return False
```

That way, the signature won't change every time we add a new killswitch.

lobsterkatie: But, if you do the above, at that point I actually do think that the killswitch stuff should be in a separate PR, because it's not really about which feature flag to use.

jangjodi:
Talked with @lobsterkatie offline - we need the separate killswitch check so this feature can be turned off (and there are too many projects to use options automator). We will clean up the killswitch function in a follow up PR

lobsterkatie (approved):

PR #74298, opened 2024-07-15
feat(dashboards): Add errors dataset to dashboards

narsaynorath:
This adds the first part of the config for the errors dataset to dashboards. It focuses on the widget builder and accomplishes the following:

1. Surfacing the error radio button in the widget builder
2. When selected, a request is made using `dataset=errors`

The basis for the config was from the `errorsAndTransactions.tsx` config, but where applicable I've removed logic relating to on demand metrics and split decisions because with this dataset, the split decision is explicit and there's no need to send the dashboard widget ID.

I am aware that the fields being presented are not errors only, I want to make another PR to specifically split apart the errors from transactions fields/aggregates so this PR doesn't get too confusing.

### DIFF
--- a/static/app/views/dashboards/datasetConfig/errors.tsx
+++ b/static/app/views/dashboards/datasetConfig/errors.tsx
@@ -1,0 +1,334 @@
+import type {Client} from 'sentry/api';
+import Link from 'sentry/components/links/link';
+import {Tooltip} from 'sentry/components/tooltip';
+import {t} from 'sentry/locale';
+import type {
+  EventsStats,
+  MultiSeriesEventsStats,
+  Organization,
+  PageFilters,
+  TagCollection,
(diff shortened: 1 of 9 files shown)


static/app/views/dashboards/datasetConfig/errors.tsx
```diff
@@ -0,0 +1,240 @@
...
+  if (field === 'id') {
+    return renderEventIdAsLinkable;
+  }
+
+  if (field === 'trace') {
+    return renderTraceAsLinkable;
+  }
+
+  return getFieldRenderer(field, meta, false);
+}
```
shruthilayaj, line 186: For the function that overlap with errorsAndTransactions, can we just export those and reuse instead of re-implementing here?

shruthilayaj (approved):

PR #74301, opened 2024-07-15
feat(search, tags): Add new hooks to retrieve organization tags

### LINKED ISSUE #73700: Modify front end to query and store tags by dataset
Should be completed after https://github.com/getsentry/sentry/issues/73699

Background:

Currently, the front end keeps track of organization tags in [tagStore.tsx](https://github.com/getsentry/sentry/blob/fe8da9b3eeee698b2d17666165bced326bea4771/static/app/stores/tagStore.tsx). These are loaded by [loadOrganizationTags()](https://github.com/getsentry/sentry/blob/ff5efb59a42e5936507bcbc2a36888bc966742a8/static/app/actionCreators/tags.tsx#L31) which is called at the top of each page that requires them. They are used in issues, discover, and replays.

Another thing to note is that while this data is in a global store right now, it would be better for this to be cached using `useApiQuery` instead, which is something we have been slowly migrating towards ([see docs](https://develop.sentry.dev/frontend/network-requests/) for more info).

Requirements:

- Create a new hook to fetch organization tags: `useFetchOrganizationTags()` (or similar)
  - When fetching data, this should cache for as long as possible to avoid refetching unless the page selectors have changed (project, time range, environment)
  - When page selectors have changed, keep the previous data until the new request has finished (can use `keepPreviousData` option)
- Create a hook to be used in issue search: `useFetchIssueTags()` (or similar)
  - This hook should call `useFetchOrganizationTags` for the Events and IssuePlatform datasets and combine the tags into the returned list
  - Should also migrate the tag store logic contained in `getIssueTags()` which appends issue-specific tags.
- When the `issue-stream-search-query-builder` flag is enabled, disable the `loadOrganizationTags()` call on the issue stream and use the new hook instead to feed tags into the search bar.

MichaelSun48:
closes #73700

This PR alters the way organization tags are retrieved on the frontend to reflect the changes made in #73943. Specifically, it utilizes the new `dataset` parameter to retrieve only tags from the events and issue_platform datasets. In addition to this, this PR also creates new `useApiQuery` hooks to call the endpoints and cache the responses, rather than using TagStore. 

These changes are functionally gated by the `issue-stream-search-query-builder` flag until it is fully GA'd.


static/app/views/issueList/searchBar.tsx
```diff
@@ -64,6 +68,11 @@ const getFilterKeySections = (
     allTags.filter(tag => tag.kind === FieldKind.EVENT_FIELD),
     ['key']
   ).map(tag => tag.key);
+  const eventTags = orderBy(
+    allTags.filter(tag => tag.kind === FieldKind.TAG),
+    ['totalValues', 'key'],
+    ['desc', 'asc']
+  ).map(tag => tag.key);
```
MichaelSun48, line 75: Tags retrieved from the organization/org-slug/tags endpoint (which have `kind=FieldKind.TAG`) fall into the Event Tags section in the search dropdown

static/app/views/issueList/searchBar.tsx
```diff
@@ -84,6 +93,35 @@ const getFilterKeySections = (
   ];
 };
 
+const getAdditionalIssueTags = (
```
MichaelSun48, line 96: Add additional tags to the ones retrieved from the organization/org-slug/tags endpoint. These include the semver fields and the predefined builtIn tags.

static/app/views/issueList/issueSearchWithSavedSearches.spec.tsx
```diff
@@ -23,6 +23,10 @@ describe('IssueSearchWithSavedSearches', () => {
       url: '/organizations/org-slug/searches/',
       body: [savedSearch],
     });
+    MockApiClient.addMockResponse({
```
MichaelSun48, line 26: Ignore all of `MockApiClient.addMockResponse({...})` additions made in this PR - [this prevents CI from failing](https://develop.sentry.dev/development/testing/#api-fixtures)

static/app/components/modals/savedSearchModal/editSavedSearchModal.spec.tsx
```diff
@@ -27,6 +27,10 @@ describe('EditSavedSearchModal', function () {
       method: 'POST',
       body: [],
     });
+    MockApiClient.addMockResponse({
```
MichaelSun48, line 30: Ignore all of `MockApiClient.addMockResponse({...})` additions made in this PR - [this prevents CI from failing](https://develop.sentry.dev/development/testing/#api-fixtures)

static/app/components/modals/savedSearchModal/createSavedSearchModal.spec.tsx
```diff
@@ -47,6 +47,10 @@ describe('CreateSavedSearchModal', function () {
       method: 'POST',
       body: [],
     });
+    MockApiClient.addMockResponse({
```
MichaelSun48, line 50: Ignore all of `MockApiClient.addMockResponse({...})` additions made in this PR - [this prevents CI from failing](https://develop.sentry.dev/development/testing/#api-fixtures)

static/app/actionCreators/tags.tsx
```diff
@@ -192,3 +208,214 @@ export function fetchSpanFieldValues({
...
+  );
+
+  return {
+    tags: eventsTagCollection,
+    isLoading: eventsTagsQuery.isLoading || issuePlatformTagsQuery.isLoading,
+    isError: eventsTagsQuery.isError || issuePlatformTagsQuery.isError,
+  };
+};
+
+export function builtInIssuesFields(
```
MichaelSun48, line 292: This function defines and populates the builtin tags. It is nearly the same as the [getIssueAttributes](https://github.com/getsentry/sentry/blob/f66b6a9013dc827900f1f6fd087e51cfff06ce43/static/app/stores/tagStore.tsx#L69) function defined in `TagStore`. We cannot reuse that function because it depends on the TagStore's built in state to populate the `has` field values.

static/app/actionCreators/tags.tsx
```diff
@@ -192,3 +208,214 @@ export function fetchSpanFieldValues({
     query,
   });
 }
+
+type FetchOrganizationTagsParams = {
+  orgSlug: string;
+  // TODO: Change this to Dataset type once IssuePlatform is added
+  dataset?: string;
```
MichaelSun48, line 215: I initially made the type of this dataset parameter [Dataset](https://github.com/getsentry/sentry/blob/f66b6a9013dc827900f1f6fd087e51cfff06ce43/static/app/views/alerts/rules/metric/types.tsx#L29). However, we would need to add `IssuePlatform='search_issues'`, and this causes type errors elsewhere. Since there is only one place using this hook, I think it's reasonable to leave it as a String now and refactor it later in a different PR.

malwilley: Glad you called this out, this is a good followup after this gets merged.

static/app/actionCreators/tags.tsx
```diff
@@ -192,3 +208,214 @@ export function fetchSpanFieldValues({
...
+};
+
+export const makeFetchOrganizationTags = ({
+  orgSlug,
+  dataset,
+  projectIds,
+  useCache = true,
+}: FetchOrganizationTagsParams): ApiQueryKey => [
+  `/organizations/${orgSlug}/tags/`,
+  {query: {dataset, useCache, project: projectIds}},
```
malwilley, line 227: This should also accept time range parameters, right? like start, end, statsPeriod

static/app/actionCreators/tags.tsx
```diff
@@ -192,3 +208,214 @@ export function fetchSpanFieldValues({
...
+];
+
+export const useFetchOrganizationTags = (
+  params: FetchOrganizationTagsParams,
+  options: Partial<UseApiQueryOptions<Tag[]>>
+) => {
+  return useApiQuery<Tag[]>(makeFetchOrganizationTags(params), {
+    staleTime: Infinity,
+    keepPreviousData: true,
```
malwilley, line 236: For hooks in this file we'd like for them to be widely reusable, so I wouldn't add specific config like `keepPreviousData` here, I'd use that config when this gets consumed if the use case requires it.

static/app/actionCreators/tags.tsx
```diff
@@ -192,3 +208,214 @@ export function fetchSpanFieldValues({
...
+  });
+};
+
+type UseFetchIssueTagsParams = {
+  orgSlug: string;
+  projectIds: string[];
+  useCache?: boolean;
+};
+
+export const useFetchIssueOrganizationTags = ({
```
malwilley, line 247: Files in `actionCreators` should ideally be reusable across the app, so I'd suggest moving this hook to `views/issueList` since it should only be used by the components in that directory.

static/app/views/issueList/searchBar.tsx
```diff
@@ -93,6 +131,10 @@ const EXCLUDED_TAGS = ['environment'];
 function IssueListSearchBar({organization, tags, ...props}: Props) {
   const api = useApi();
   const {selection: pageFilters} = usePageFilters();
+  const {tags: issueTags} = useFetchIssueOrganizationTags({
+    orgSlug: organization.slug,
+    projectIds: pageFilters.projects.map(id => id.toString()),
```
malwilley, line 136: We should pass the option `enabled: organization.features.includes('...')` here so this request only gets made if the feature flag is enabled. You could also add a wrapper component for `SearchQueryBuilder` and add the hook there so it only gets called if that component is rendered.

static/app/views/issueList/searchBar.spec.tsx
malwilley, line 1: Can you add a test here that verifies the correct tags are being loaded?

malwilley:
I noticed while testing that `has` wasn't autocompleting the way it did before, so something might be missing there:

CleanShot 2024-07-16 at 09 40 41@2x

MichaelSun48:
> I noticed while testing that has wasn't autocompleting the way it did before, so something might be missing there: 

Good catch, just pushed a fix and confirmed its working again

static/app/views/issueList/utils/useFetchIssueTags.tsx
```diff
@@ -0,0 +1,303 @@
...
+    [FieldKey.FIRST_SEEN]: {
+      ...builtInTags[FieldKey.FIRST_SEEN],
+      name: 'First Seen',
+      values: [],
+      predefined: false,
+    },
+    [FieldKey.FIRST_RELEASE]: {
+      ...builtInTags[FieldKey.FIRST_RELEASE],
+      name: 'First Release',
+      values: ['latest'],
```
MichaelSun48, line 250: This is definitely out of the scope of this PR, but we should really consider if some of these predefined values are outdated. `"latest"` should not be the _only_ option for releases

static/app/views/issueList/utils/useFetchIssueTags.tsx
```diff
@@ -0,0 +1,303 @@
...
+  };
+};
+
+function builtInIssuesFields(
+  org: Organization,
+  currentTags: TagCollection,
+  assigneeFieldValues: SearchGroup[] | string[] = [],
+  bookmarksValues: string[] = []
+): TagCollection {
+  const semverFields: TagCollection = Object.values(SEMVER_TAGS).reduce<TagCollection>(
```
MichaelSun48, line 140: The `"RELEASE_STAGE"` field is part of both the builtInTags (line 159 in this file) and in semverFields, but the `getBuiltInTags` function categorizes it as an `"ISSUE_FIELD"` while this function categorizes it as an `"EVENT_FIELD"`. In the return statement, I have deconstructed semver fields after builtInTags, thus resolving the conflict by favoring semver's categorization as an `"EVENT_FIELD"`, but we should maybe resolve this in the `SEMVER_TAGS` or `ISSUE_PROPERTY_FIELDS` declarations

static/app/views/issueList/searchBar.spec.tsx
```diff
@@ -165,4 +166,68 @@ describe('IssueListSearchBar', function () {
...
+      defaultProps.organization.features = ['issue-stream-search-query-builder'];
+
+      render(<IssueListSearchBar {...defaultProps} />, {
+        router,
+      });
+
+      await userEvent.click(screen.getByRole('combobox', {name: 'Add a search term'}));
+      await userEvent.paste('is:', {delay: null});
+      await userEvent.click(
+        screen.getByRole('button', {name: 'Edit value for filter: is'})
```
malwilley, line 186: You might want to change this to a `await screen.findByRole(...)`. This might seem unnecessary since tests are passing, but for a complex component like this, it can lead to flaky tests in CI if we look for something without waiting.

static/app/views/issueList/searchBar.spec.tsx
```diff
@@ -165,4 +166,68 @@ describe('IssueListSearchBar', function () {
...
   });
+
+  describe('Tags and Fields', function () {
+    it('displays the correct options for the is tag', async function () {
+      MockApiClient.addMockResponse({
+        url: '/organizations/org-slug/tags/',
+        body: [],
+      });
+
+      defaultProps.organization.features = ['issue-stream-search-query-builder'];
```
malwilley, line 177: If you mutate this here, subsequent tests will have the same value which you don't want. I'd pass a new organization prop instead

static/app/views/issueList/searchBar.spec.tsx
```diff
@@ -165,4 +166,68 @@ describe('IssueListSearchBar', function () {
...
+
+      render(<IssueListSearchBar {...defaultProps} />, {
+        router,
+      });
+
+      await userEvent.click(screen.getByRole('combobox', {name: 'Add a search term'}));
+      await userEvent.click(screen.getByRole('button', {name: 'Event Tags'}));
+
+      expect(screen.getByRole('option', {name: 'someTag'})).toBeInTheDocument();
```
malwilley, line 209: This expectation you really want to wait for since there is an async request being made

static/app/views/issueList/searchBar.spec.tsx
```diff
@@ -165,4 +166,68 @@ describe('IssueListSearchBar', function () {
...
+      await userEvent.click(
+        screen.getByRole('button', {name: 'Edit value for filter: is'})
+      );
+
+      Object.values(IsFieldValues).forEach(value => {
+        expect(screen.getByRole('option', {name: value})).toBeInTheDocument();
+      });
+    });
+
+    it('displays the correct options under Issue Filters', async function () {
```
malwilley, line 194:
```suggestion
    it('displays the correct options under "Event Tags"', async function () {
```

static/app/views/issueList/searchBar.tsx
```diff
@@ -93,6 +98,12 @@ const EXCLUDED_TAGS = ['environment'];
 function IssueListSearchBar({organization, tags, ...props}: Props) {
   const api = useApi();
   const {selection: pageFilters} = usePageFilters();
+  const {tags: issueTags} = useFetchIssueTags({
```
malwilley, line 101: Can you pass in the time range filters here? It should work similarly to `tagValueLoader` below. You could also do this in a followup PR if you'd like

static/app/views/issueList/searchBar.tsx
```diff
@@ -93,6 +98,15 @@ const EXCLUDED_TAGS = ['environment'];
...
+    org: organization,
+    projectIds: pageFilters.projects.map(id => id.toString()),
+    keepPreviousData: true,
+    enabled: organization.features.includes('issue-stream-search-query-builder'),
+    start: pageFilters.datetime.start,
```
malwilley, line 106: Noticed a couple things here:

- Selecting an absolute date range doesn't work (need to use the `getUtcDateString()` function)
- Changing the time range seems to cause double requests, one with `start/end` as empty string and one with them undefined

MichaelSun48: Thanks for catching this - should be fixed now!

static/app/views/issueList/utils/useFetchIssueTags.tsx
```diff
@@ -0,0 +1,303 @@
...
+        ...nonMemberTeams.map(convertToSearchItem),
+      ],
+    },
+  ];
+  const eventsTags = [
+    ...(eventsTagsQuery.data || []),
+    ...(issuePlatformTagsQuery.data || []),
+  ];
+  const eventsTagCollection = eventsTags.reduce<TagCollection>((acc, tag) => {
+    acc[tag.key] = {...tag, kind: FieldKind.TAG};
```
malwilley, line 113: Sorry, one _last_ comment! We use the `totalValues` number on the event tag to sort keys in the "Event Tags" section (higher values go to the top). If the same tag key gets returned from both datasets, this will override it with the issue platform response. If we come across a tag that's already in the map, we should merge the `totalValues` value

malwilley (approved):
Works very well, great job!

PR #74307, opened 2024-07-15
feat(uptime): Add `uptime_status` column to `ProjectUptimeSubscription`

wedamija:
This adds the `update_status` column to `ProjectUptimeSubscription` so that we can keep track of the current state of the uptime monitor. This is probably a temporary column, and will be replaced with a more detailed table later on


src/sentry/uptime/models.py
```diff
@@ -50,6 +50,11 @@ class ProjectUptimeSubscriptionMode(enum.IntEnum):
     AUTO_DETECTED_ACTIVE = 3
 
 
+class UptimeStatus(enum.IntEnum):
+    OK = 1
+    FAILED = 2
```
evanpurkhiser, line 55: Let's leave a comment on this noting that it's very likely temporary.

evanpurkhiser (approved):

PR #74310, opened 2024-07-15
feat(anomaly detection): Update API with new fields

mifu67:
Update the serializer with the new fields needed for anomaly detection.

Closes https://getsentry.atlassian.net/browse/ALRT-139

### DIFF
--- a/src/sentry/incidents/endpoints/serializers/alert_rule.py
+++ b/src/sentry/incidents/endpoints/serializers/alert_rule.py
@@ -53,7 +53,7 @@ class AlertRuleSerializerResponseOptional(TypedDict, total=False):
     snooze: bool | None
     latestIncident: datetime | None
     errors: list[str] | None
-    sensitivity: str | None  # mifu67: I think these three should be here, but I will confirm
+    sensitivity: str | None
     seasonality: str | None
 
 
@@ -68,7 +68,7 @@ class AlertRuleSerializerResponseOptional(TypedDict, total=False):
         "totalThisWeek",
         "latestIncident",
         "description",  # TODO: remove this once the feature has been released to add to the public docs, being sure to denote it will only display in Slack notifications
-        "sensitivity",  # mifu67: the following three fields are for anomality detection, which is behind a feature flag
+        "sensitivity",  # The following three fields are for anomaly detection, which is behind a feature flag
         "seasonality",
         "detection_type",
     ]
--- a/src/sentry/incidents/endpoints/serializers/alert_rule.py
+++ b/src/sentry/incidents/endpoints/serializers/alert_rule.py
@@ -55,7 +55,6 @@ class AlertRuleSerializerResponseOptional(TypedDict, total=False):
     errors: list[str] | None
     sensitivity: str | None  # mifu67: I think these three should be here, but I will confirm
     seasonality: str | None
-    activationCondition: int | None
 
 
 @extend_schema_serializer(
@@ -95,6 +94,7 @@ class AlertRuleSerializerResponse(AlertRuleSerializerResponseOptional):
     createdBy: dict
     monitorType: int
     activations: list[dict]
+    activationCondition: int | None
     description: str
     detection_type: str
 
@@ -311,9 +311,9 @@ class AlertRuleSerializer(Serializer):
             "activationCondition": condition_type,
             "activations": attrs.get("activations", None),
             "description": obj.description if obj.description is not None else "",
-            "sensitivity": obj.get("sensitivity", None),
-            "seasonality": obj.get("seasonality", None),
-            "detection_type": obj.detection_type,  # mifu67: might need to get here
+            "sensitivity": obj.sensitivity,
+            "seasonality": obj.seasonality,
+            "detection_type": obj.detection_type,
         }
         rule_snooze = RuleSnooze.objects.filter(
             Q(user_id=user.id) | Q(user_id=None), alert_rule=obj
--- a/src/sentry/incidents/endpoints/serializers/alert_rule.py
+++ b/src/sentry/incidents/endpoints/serializers/alert_rule.py
@@ -68,9 +68,9 @@ class AlertRuleSerializerResponseOptional(TypedDict, total=False):
         "totalThisWeek",
         "latestIncident",
         "description",  # TODO: remove this once the feature has been released to add to the public docs, being sure to denote it will only display in Slack notifications
-        "sensitivity",  # The following three fields are for anomaly detection, which is behind a feature flag
-        "seasonality",
-        "detection_type",
+        "sensitivity",  # For anomaly detection, which is behind a feature flag
+        "seasonality",  # For anomaly detection, which is behind a feature flag
+        "detection_type",  # For anomaly detection, which is behind a feature flag
     ]
 )
 class AlertRuleSerializerResponse(AlertRuleSerializerResponseOptional):
--- a/src/sentry/incidents/endpoints/serializers/alert_rule.py
+++ b/src/sentry/incidents/endpoints/serializers/alert_rule.py
@@ -55,7 +55,7 @@ class AlertRuleSerializerResponseOptional(TypedDict, total=False):
     errors: list[str] | None
     sensitivity: str | None  # mifu67: I think these three should be here, but I will confirm
     seasonality: str | None
-    detection_type: str | None  # mifu67: this field has a default value, but not all alert rules have it
+    activationCondition: int | None
 
 
 @extend_schema_serializer(
@@ -95,8 +95,8 @@ class AlertRuleSerializerResponse(AlertRuleSerializerResponseOptional):
     createdBy: dict
     monitorType: int
     activations: list[dict]
-    activationCondition: int | None
     description: str
+    detection_type: str
 
 
 @register(AlertRule)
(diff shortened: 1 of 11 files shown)


src/sentry/incidents/logic.py
```diff
@@ -516,8 +519,12 @@ def create_alert_rule(
     monitor_type: AlertRuleMonitorTypeInt = AlertRuleMonitorTypeInt.CONTINUOUS,
     activation_condition: AlertRuleActivationConditionType | None = None,
     description: str | None = None,
+    sensitivity: AlertRuleSensitivity | None = None,
+    seasonality: AlertRuleSeasonality | None = None,
+    detection_type: AlertRuleDetectionType = AlertRuleDetectionType.STATIC,
```
ceorourke, line 524: I was just looking at this as a part of my backfill PR - I think we might infer the `detection_type` based on the presence of other fields rather than passing it in.  If there is a `comparison_delta` that means it's a percent type, and if there are `sensitivity` and `seasonality` values that means it's a dynamic type, otherwise it's static. Lmk if you disagree or think it should be handled higher up

src/sentry/incidents/endpoints/serializers/alert_rule.py
```diff
@@ -66,6 +68,9 @@ class AlertRuleSerializerResponseOptional(TypedDict, total=False):
         "totalThisWeek",
         "latestIncident",
         "description",  # TODO: remove this once the feature has been released to add to the public docs, being sure to denote it will only display in Slack notifications
+        "sensitivity",  # The following three fields are for anomaly detection, which is behind a feature flag
```
nhsiehgit, line 71: Lets not assume that these will stay in order here.
Best to add the note to each item if necessary

src/sentry/incidents/logic.py
```diff
@@ -553,6 +562,19 @@ def create_alert_rule(
 
     resolution = get_alert_resolution(time_window, organization)
 
+    if (sensitivity or seasonality) and not (sensitivity and seasonality):
```
nhsiehgit, line 565: Can we stick this behind the `detection_type` check?
having these set only makes sense for the dynamic detection type right?

src/sentry/incidents/logic.py
```diff
@@ -553,6 +562,19 @@ def create_alert_rule(
...
+        raise ValidationError("Must choose both sensitivity and seasonality")
+
+    if detection_type == AlertRuleDetectionType.STATIC:
+        if comparison_delta is not None or sensitivity is not None or seasonality is not None:
```
nhsiehgit, line 569: we're checking if any of them are set... 🤔 
this seems unnecessary.

can we not just have a validation here for dynamic, then `else` if any of them are set raise the error?

maybe something like
```
if detection_type == AlertRuleDetectionType.DYNAMIC:
    if not (sensitivity or seasonality):
        raise ...
else:
    if (sensitivity or seasonality):
        raise ...

if detection_type == AlertRuleDetectionType.PERCENT:
    if comparison_delta is None:
        raise ...
else:
    if comparison_delta is not None:
        raise ...
```

or something like this?

src/sentry/incidents/logic.py
```diff
@@ -732,6 +763,22 @@ def update_alert_rule(
...
+        updated_fields["sensitivity"] = sensitivity
+    if seasonality:
+        updated_fields["seasonality"] = seasonality
+    if detection_type:
+        updated_fields["detection_type"] = detection_type
+        # make sure we clear the incorrect fields for each detection type
+        if detection_type == AlertRuleDetectionType.STATIC:
+            updated_fields["sensitivity"] = None
+            updated_fields["seasonality"] = None
+            updated_fields["comparison_delta"] = None
```
nhsiehgit, line 776: is comparison delta being set anywhere in the update here?

nhsiehgit: Oh it's being set below.
shoudl this be inserted _after_ that?

src/sentry/incidents/models/alert_rule.py
```diff
@@ -424,6 +424,7 @@ def clear_alert_rule_trigger_cache(cls, instance, **kwargs):
 class AlertRuleThresholdType(Enum):
     ABOVE = 0
     BELOW = 1
+    BOTH = 2
```
nhsiehgit, line 427: ehhh?

what does both mean?

nhsiehgit (commented):
🤔 hmmmmm few comments

haven't reviewed the tests yet

src/sentry/incidents/models/alert_rule.py
mifu67: From the anomaly detection spec: detect anomaly if values are either above or below the threshold

src/sentry/incidents/logic.py
mifu67: This is a lot prettier 😅

src/sentry/incidents/models/alert_rule.py
nhsiehgit: Ahhhh yea i see that's right 🤔

Can we add a note for that here?

tests/sentry/incidents/test_logic.py
```diff
@@ -1267,6 +1275,94 @@ def test_update_alert_load_shedding_on_comparison_and_window(self):
             * 60
         )
 
+    @with_feature("organizations:anomaly-detection-alerts")
```
ceorourke, line 1278: love all these thorough tests!

src/sentry/incidents/models/alert_rule.py
ceorourke: Would naming it `ABOVE_AND_BELOW` be more clear? or is `BOTH` with the note fine?

ceorourke (approved):
looks great!! thanks!

tests/sentry/incidents/endpoints/test_organization_alert_rule_details.py
```diff
@@ -206,6 +211,113 @@ def test_with_description(self):
         resp = self.get_success_response(self.organization.slug, rule.id)
         assert rule.description == resp.data.get("description")
 
+    @with_feature("organizations:anomaly-detection-alerts")
+    @with_feature("organizations:incidents")
+    def test_static_detection_type(self):
```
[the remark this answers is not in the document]
nhsiehgit: I guess those validations actually live in the serializer too, so maybe the test belongs in the `test_serializers`?
🤔 not sure - here's fine also 🤷

tests/sentry/incidents/endpoints/test_serializers.py
```diff
@@ -701,6 +703,7 @@ def test_comparison_delta_below(self):
         params["resolveThreshold"] = 10
         params["triggers"][0]["alertThreshold"] = 50
         params["triggers"][1]["alertThreshold"] = 40
+        params["detection_type"] = AlertRuleDetectionType.PERCENT
         serializer = AlertRuleSerializer(context=self.context, data=params, partial=True)
         assert serializer.is_valid(), serializer.errors
         alert_rule = serializer.save()
```
nhsiehgit, line 709: in the serializer tests, you could test the validations raise here by wrapping the `serializer.save()` in the pytest raises as well

tests/sentry/incidents/test_logic.py
nhsiehgit: vnice! ditto colleen :D

nhsiehgit (commented):
awesome!
few more comments!

src/sentry/incidents/endpoints/serializers/alert_rule.py
```diff
@@ -91,6 +96,7 @@ class AlertRuleSerializerResponse(AlertRuleSerializerResponseOptional):
     activations: list[dict]
     activationCondition: int | None
     description: str
+    detection_type: str
```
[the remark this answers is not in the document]
mifu67: Nope, since this inherits the `AlertRuleSerializerResponseOptional`

src/sentry/incidents/logic.py
```diff
@@ -553,6 +563,26 @@ def create_alert_rule(
...
+            )
+
+    if detection_type == AlertRuleDetectionType.PERCENT:
+        if comparison_delta is None:
+            raise ValidationError("Percentage-based alerts require a comparison delta")
+    else:
+        if comparison_delta is not None:
+            raise ValidationError("Comparison delta is not a valid field for this alert type")
+
     if comparison_delta is not None:
```
ceorourke, line 586: I just remembered that since we've published these endpoints we need to be careful to not break customer's scripts/workflow/etc. and should make sure that if they're creating a percent type alert but not passing `detection_type` (because it's new and they didn't know to pass it) we set the detection type to percent for them.

src/sentry/incidents/serializers/alert_rule.py
```diff
@@ -94,6 +95,10 @@ class AlertRuleSerializer(CamelSnakeModelSerializer[AlertRule]):
     activation_condition = serializers.IntegerField(required=False, allow_null=True, min_value=0)
     description = serializers.CharField(required=False, allow_blank=True)
 
+    sensitivity = serializers.CharField(required=False, allow_null=True)
+    seasonality = serializers.CharField(required=False, allow_null=True)
+    detection_type = serializers.CharField(required=False, default=AlertRuleDetectionType.STATIC)
```
ceorourke, line 100: Should this be required? Or is this shared by the PUT endpoint and can't be for that reason?

mifu67: I followed the convention set by the owner field (set required=True once it starts being sent by the frontend)

PR #74311, opened 2024-07-15
fix(slack): get latest parent message to create a thread with

cathteng:
Fixes SENTRY-39PA

A parent message is a message you can create a thread with.

We are getting a `NotificationMessage.MultipleObjectsReturned` error when trying to fetch a single parent notification to start a thread with. We use the `rule_id` and `group_id` of the `rule_fire_history` object to get the `NotificationMessage` object, but this doesn't account for the case where a single rule can fire multiple times for an issue and create multiple `NotificationMessage` objects that match the filter.

If somebody doesn't have threads turned on, each firing of the same rule for 1 issue will be saved as a potential parent message. If somebody then turns the Slack threads feature on, this error will surface because we will now attempt to get 1 parent message from multiple. We need to pick one of these to start the thread with, here we can pick the _latest_ message.

### DIFF
--- a/src/sentry/integrations/repository/issue_alert.py
+++ b/src/sentry/integrations/repository/issue_alert.py
@@ -108,10 +108,14 @@ class IssueAlertNotificationMessageRepository:
         """
         try:
             base_filter = self._parent_notification_message_base_filter()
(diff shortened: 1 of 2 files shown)


RyanSkonnord (approved):
Thanks for the thorough explanation.

PR #74315, opened 2024-07-15
WIP new circuit breaker

lobsterkatie:
Notes:

- ~I think the config is too complicated. Am working on setting more of the values automatically. (I realized the defaults for recovery error limit and window were wrong/didn't do what I'd want them to do, and if I can mess that up after thinking about this stuff for a week, it's too confusing.)~
- I included the tests in case you're curious but really I mostly just want to make sure the two main methods `should_allow_request` and `record_error` pass a sanity check.
- ~There's obv still a bunch of cleanup I have to do.~
- Ugh, overloads.

### DIFF
--- a/src/sentry/utils/circuit_breaker2.py
+++ b/src/sentry/utils/circuit_breaker2.py
@@ -1,0 +1,6 @@
+"""
+NOTE: This is a replacement for the current circuit breaker implementation, which is why it is
+`circuit_breaker2`. It's first going to be used for the Seer similarity service, then once we're
+confident it works we can replace use of the original for the severity service with use of this one
+and get rid of the old one, at which point this can lose the `2`.
+"""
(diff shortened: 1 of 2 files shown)


src/sentry/utils/circuit_breaker2.py
```diff
@@ -0,0 +1,328 @@
...
+            else:
+                logger.warning("Request blocked by circuit breaker!")
+                return None
+        except TimeoutError:
+            breaker.record_error()
+            return None
+
+        if response.status == 500:
+            breaker.record_error()
+            return None
```
vartec, line 107:
```suggestion
        try:
            if breaker.should_allow_request():
                response = call_chase_simulation_service("/hall-of-fame", payload)
            else:
                logger.warning("Request blocked by circuit breaker!")
                return None
        except (InvalidInputDataError, PEBKACError):
              # explicitly list errors that are ignored by the breaker
              # for example errors caused by user data not validating
             return None        
        except Exception:
            # any exception not listed above should be counted by the breaker
            breaker.record_error()
            return None

        if response.status == 500:
            breaker.record_error()
            return None
```

lobsterkatie:
Broke this up into 3 PRs:
https://github.com/getsentry/sentry/pull/74557
https://github.com/getsentry/sentry/pull/74559
https://github.com/getsentry/sentry/pull/74560

Closing this in favor of those.

PR #74320, opened 2024-07-15
fix(empty-states): Fix auth token generation in empty states

roggenkemper:
this pr updates the updated empty state to fix a problem with auth token generation, which is needed to set up a few platforms


roggenkemper:
this doesn't actually change anything beyond adding the provider but that addition led to a few formatting changes which is why the diff is so large

malwilley (approved):

PR #74323, opened 2024-07-15
feat(rules): Backfill detection_type

ceorourke:
Backfill the `detection_type` column - the default is static but we can determine whether it's a static or percent based alert by checking for the existence of a value in the `comparison_delta` field which indicates that it's a percent based alert. At this time no anomaly detection alerts exist yet so we don't need to check for those in this migration.

Relies on https://github.com/getsentry/sentry/pull/74310 being merged first so we're setting the `detection_type` correctly on new / edited alert rules.

Closes https://getsentry.atlassian.net/browse/ALRT-160


tests/sentry/migrations/test_0742_backfill_alertrule_detection_type.py
```diff
@@ -0,0 +1,21 @@
...
+        self.metric_alert_rule = self.create_alert_rule(
+            organization=self.project.organization, projects=[self.project]
+        )
+        self.percent_metric_alert_rule = self.create_alert_rule(
+            organization=self.project.organization, projects=[self.project], comparison_delta=60
+        )
+
+    def test(self):
+        self.metric_alert_rule.refresh_from_db()
+        assert self.metric_alert_rule.detection_type == AlertRuleDetectionType.STATIC
```
ceorourke, line 19: This test is in a bit of a grey chicken/egg area - by the time the API changes are merged the detection type will be set when calling `create_alert_rule` so this is kind of moot, but we still need to backfill the rows for all metric alert rules created before that.

wedamija (approved):

src/sentry/migrations/0742_backfill_alertrule_detection_type.py
```diff
@@ -0,0 +1,58 @@
...
+    # of a table.
+    # Here are some things that make sense to mark as post deployment:
+    # - Large data migrations. Typically we want these to be run manually so that they can be
+    #   monitored and not block the deploy for a long period of time while they run.
+    # - Adding indexes to large tables. Since this can take a long time, we'd generally prefer to
+    #   run this outside deployments so that we don't block them. Note that while adding an index
+    #   is a schema change, it's completely safe to run the operation after the code has deployed.
+    # Once deployed, run these manually via: https://develop.sentry.dev/database-migrations/#migration-deployment
+
+    is_post_deployment = False
```
wedamija, line 45: We should make this a post deploy migration

PR #74325, opened 2024-07-15
chore(release): Adding 24.7.0 release test

IanWoodard:
Adding release test for 24.7.0 release.


hubertdeng123 (approved):

azaslavsky:
@IanWoodard quick note here - we need to delete the 24.5.X tests as well, since they have now fallen out of the support window (2 minor versions back).

PR #74326, opened 2024-07-15
feat(new-trace): Sorting issues by level.

Abdkhan14:


static/app/views/performance/newTraceDetails/traceDrawer/details/issues/issues.tsx
```diff
@@ -44,6 +45,20 @@ const TABLE_WIDTH_BREAKPOINTS = {
...
+    error: 1,
+    warning: 2,
+    sample: 3,
+    info: 4,
+    default: 5,
+    unknown: 6,
+  };
+
+  return issues.sort((a, b) => order[a.level] - order[b.level]);
+}
```
JonasBa, line 60: This will mutate the array reference which can be a tricky thing to find the root of. I would suggest you make a copy of the array and isolate the sorting action like so

```suggestion

  const issueOrderPriority: Record<keyof Theme['level'], number> = {
    fatal: 0,
    error: 1,
    warning: 2,
    sample: 3,
    info: 4,
    default: 5,
    unknown: 6,
  };
function sortIssuesByLevel(a: TraceError, b: TraceError): number {
  // Check for a.level and b.level in issueOrderPriority before doing int comparison
  return issueOrderPriority[a.level] - issueOrderPriority[b.level]);
}
```

Which means that you can then do 

```
const sortedErrorsByLevel = [...errors].sort(sortIssuesByLevel)
```

Which makes it clear that you are creating a copy and not mutating the arr reference inside some hidden function

Abdkhan14: Yeah agree, this is better.

JonasBa (approved):

PR #74328, opened 2024-07-16
chore(flagpole): Adds metrics to batch_has checks

GabeVillalobos:
Adds metric gathering to `batch_has` entity handler checks to begin gathering flagpole performance information.

### DIFF
--- a/src/sentry/features/manager.py
+++ b/src/sentry/features/manager.py
@@ -327,10 +327,13 @@ class FeatureManager(RegisteredFeatureManager):
         Will only accept one type of feature, either all ProjectFeatures or all
         OrganizationFeatures.
         """
+        sample_rate = 0.01
         if self._entity_handler:
-            return self._entity_handler.batch_has(
-                feature_names, actor, projects=projects, organization=organization
-            )
+            with metrics.timer("features.entity_batch_has", sample_rate=sample_rate):
+                return self._entity_handler.batch_has(
+                    feature_names, actor, projects=projects, organization=organization
+                )
+
         else:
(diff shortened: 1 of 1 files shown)


src/sentry/features/manager.py
```diff
@@ -327,10 +327,13 @@ def batch_has(
         Will only accept one type of feature, either all ProjectFeatures or all
         OrganizationFeatures.
         """
+        sample_rate = 0.01
```
mrduncan, line 330: Any reason to define this separately here instead of directly setting it on the metrics.timer?

src/sentry/features/manager.py
```diff
@@ -327,10 +327,13 @@ def batch_has(
...
         if self._entity_handler:
-            return self._entity_handler.batch_has(
-                feature_names, actor, projects=projects, organization=organization
-            )
+            with metrics.timer("features.entity_batch_has", sample_rate=sample_rate):
```
mrduncan, line 332: Doesn't need to change here since it doesn't particularly matter, but alternatively you should be able to annotate `batch_has` with `@metrics.wraps("features.entity_batch_has", sample_rate=0.01)` to keep the implementation a tiny bit cleaner

mrduncan (approved):
We're already sampling `has` checks with this same sample rate so this seems like a safe sample rate to me

src/sentry/features/manager.py
GabeVillalobos: I was mainly following the pattern `has` was using for this, but no strong reason to do it this way.

PR #74329, opened 2024-07-16
fix(alerts): fixes aggregate selector not functioning correctly for metric alerts

edwardgou-sentry:
Fixes an issue with the aggregate selector not handling change properly on Custom Metric based alerts. Updates to pass the correct MRI + Aggregate string to the onChange handler.

image


edwardgou-sentry:
I think the onChange logic had the mri field flipped with the aggregate field here.  The aggregate selector would kick me back to an empty state any time I tried to change the value.

ArthurKnaus (approved):
Thx for fixing and the test 🚀

PR #74330, opened 2024-07-16
Add PUT request for project templates

saponifi3d:
# Description
[Tech Spec](https://www.notion.so/sentry/Service-Settings-Templates-ca6fdb38b15f4a6db61d319bc342b86c?pvs=4#a49242f3a1094cc68f6fe1e893450e36)

- [x] Merge the POST request and update the target branch
- Setup the PUT request for `/api/0/organizations/:slug_or_id/project_templates/:id/` that supports updating the project template and it's options.

### DIFF
--- a/src/sentry/api/endpoints/project_template_detail.py
+++ b/src/sentry/api/endpoints/project_template_detail.py
@@ -81,17 +81,16 @@ class OrganizationProjectTemplateDetailEndpoint(OrganizationEndpoint):
             ProjectTemplate, id=template_id, organization=organization
         )
 
-        name = data.get("name", None)
+        name = data.get("name")
         if name is not None:
             project_template.name = name
             project_template.save()
 
         options = data.get("options", {})
-        if options:
-            for key, value in options.items():
-                project_template.options.create_or_update(
-                    project_template=project_template, key=key, value=value
-                )
+        for key, value in options.items():
+            project_template.options.create_or_update(
+                project_template=project_template, key=key, value=value
+            )
 
         return Response(
             serialize(
--- a/src/sentry/api/endpoints/project_template_detail.py
+++ b/src/sentry/api/endpoints/project_template_detail.py
@@ -25,6 +25,7 @@ class OrganizationProjectTemplateDetailEndpoint(OrganizationEndpoint):
 
     publish_status = {
         "GET": ApiPublishStatus.PRIVATE,
+        "PUT": ApiPublishStatus.PRIVATE,
         "DELETE": ApiPublishStatus.PRIVATE,
     }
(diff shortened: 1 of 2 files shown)


src/sentry/api/endpoints/project_template_detail.py
```diff
@@ -61,3 +63,39 @@ def delete(self, request: Request, organization: Organization, template_id: str)
...
+        if not serializer.is_valid():
+            return Response(status=400, data=serializer.errors)
+
+        data = serializer.validated_data
+
+        project_template = get_object_or_404(
+            ProjectTemplate, id=template_id, organization=organization
+        )
+
+        name = data.get("name", None)
```
ceorourke, line 83: the default is already `None` on `.get`

src/sentry/api/endpoints/project_template_detail.py
```diff
@@ -61,3 +63,39 @@ def delete(self, request: Request, organization: Organization, template_id: str)
...
+        if name is not None:
+            project_template.name = name
+            project_template.save()
+
+        options = data.get("options", {})
```
ceorourke, line 88: I don't think you need the default empty dict here if you're checking it on the next line

tests/sentry/api/endpoints/test_project_template_detail.py
```diff
@@ -64,6 +64,81 @@ def test_get__no_access(self):
...
+    method = "put"
+
+    def test_put__no_feature(self):
+        response = self.get_error_response(
+            self.organization.id, self.project_template.id, status_code=404
+        )
+        assert response.status_code == 404
+
+    @with_feature(PROJECT_TEMPLATE_FEATURE_FLAG)
+    def test_put__only_name(self):
```
ceorourke, line 78: The name here indicates that it's only updating the name, but it passes both the name and the options

src/sentry/api/endpoints/project_template_detail.py
saponifi3d: d'oh, looks like there was a merge conflict.

the way i had this before the merge conflict was to not have the `if` statement on the next line, and just use the `for` loop `.items` condition to limit the creation of options. thoughts on that approach vs removing the default value here?

tests/sentry/api/endpoints/test_project_template_detail.py
saponifi3d: ty - nice catch.

src/sentry/api/endpoints/project_template_detail.py
ceorourke: I like this better 👍

ceorourke (approved):

Issue #69387, opened 2024-04-22
Prevent suffix wildcard search as long as feature is not implemented


matejminar:
As described in https://github.com/getsentry/sentry/issues/65875, the Metrics API currently cannot deal with suffix wildcards (e.g. environment:prod-*). This leads to some weird graphs, because other APIs do support this:

![Image](https://github.com/getsentry/sentry/assets/6788060/ad2fac4e-4753-4a30-bf0f-dd34f5869666)

ArthurKnaus:
As there will not be support for wildcard search in the short term, we decided to disable it on the FE.

PR #74332, opened 2024-07-16
feat(metrics): allow wildcards in metrics search bar

shellmayr:
Wildcards were previously dis-allowed in https://github.com/getsentry/sentry/pull/69644 due to https://github.com/getsentry/sentry/issues/69387
With https://github.com/getsentry/snuba/pull/5972 being merged, wildcards now work and can be re-allowed in the UI

### DIFF
--- a/static/app/components/metrics/metricSearchBar.tsx
+++ b/static/app/components/metrics/metricSearchBar.tsx
@@ -129,7 +129,6 @@ export function MetricSearchBar({
       placeholder={t('Filter by tags')}
       query={query}
       savedSearchType={SavedSearchType.METRIC}
-      disallowWildcard
       {...searchConfig}
       {...props}
     />


vgrozdanic (approved):

static/app/components/metrics/metricSearchBar.tsx
```diff
@@ -129,7 +129,6 @@ export function MetricSearchBar({
       placeholder={t('Filter by tags')}
       query={query}
       savedSearchType={SavedSearchType.METRIC}
-      disallowWildcard
```
ArthurKnaus, line 132: Just to double check: This component is also used in alerts, where we use the event-stats endpoint.
Will is work also there?

ArthurKnaus (approved):

Issue #74258, opened 2024-07-15
Adjust cardinality checks


matejminar:
We should query for `c:metric_stats/volume@none{outcome.id:6} by (mri)`.
If the returned value of a metric is > 0 it was cardinality limited.

PR #74336, opened 2024-07-16
feat(metrics) Low cardinality limit for projects setup

vgrozdanic:
Limit projects cardinality limit for testing purposes.

Instead of hitting the real cardinality limits on relay, we can configure low cardinality limits for them, and test what is happening when cardinality limits are set and how to present it to the user.

### DIFF
--- a/src/sentry/relay/config/__init__.py
+++ b/src/sentry/relay/config/__init__.py
@@ -38,7 +38,7 @@ from sentry.relay.config.metric_extraction import (
     get_metric_extraction_config,
 )
 from sentry.relay.utils import to_camel_case_name
-from sentry.sentry_metrics.use_case_id_registry import CARDINALITY_LIMIT_USE_CASES, UseCaseID
+from sentry.sentry_metrics.use_case_id_registry import CARDINALITY_LIMIT_USE_CASES
 from sentry.sentry_metrics.visibility import get_metrics_blocking_state_for_relay_config
 from sentry.utils import metrics
 from sentry.utils.http import get_origins
@@ -228,6 +228,7 @@ class CardinalityLimit(TypedDict):
 class CardinalityLimitOption(TypedDict):
     rollout_rate: NotRequired[float]
     limit: CardinalityLimit
+    project_slugs_allowlist: NotRequired[list[str]]
 
 
 def get_metrics_config(timeout: TimeChecker, project: Project) -> Mapping[str, Any] | None:
@@ -241,15 +242,7 @@ def get_metrics_config(timeout: TimeChecker, project: Project) -> Mapping[str, A
     cardinality_limits: list[CardinalityLimit] = []
     for namespace in CARDINALITY_LIMIT_USE_CASES:
         timeout.check()
-        if (
-            features.has("organizations:low-cardinality-limits-on-project", project.organization)
-            and namespace == UseCaseID.CUSTOM
-        ):
-            option = options.get("sentry-metrics.cardinality-limiter.limits.custom.per-project.low")
-        else:
-            option = options.get(
-                f"sentry-metrics.cardinality-limiter.limits.{namespace.value}.per-org"
-            )
+        option = options.get(f"sentry-metrics.cardinality-limiter.limits.{namespace.value}.per-org")
 
         if not option or not len(option) == 1:
             # Multiple quotas are not supported
@@ -265,7 +258,7 @@ def get_metrics_config(timeout: TimeChecker, project: Project) -> Mapping[str, A
                 "granularitySeconds": quota["granularity_seconds"],
             },
             "limit": quota["limit"],
-            "scope": option.get("scope", "organization"),
+            "scope": "organization",
             "namespace": namespace.value,
         }
         if id in passive_limits:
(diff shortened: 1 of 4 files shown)


ArthurKnaus:
@vgrozdanic If we want to see cardinality warnings in the frontend, we need to adjust the constant:
https://github.com/getsentry/sentry/blob/aa98fdae48c0709f26e16657152084f671c7fbd3/static/app/utils/metrics/constants.tsx#L16

iambriccardo (approved):
LGTM, are we sure these numbers are what we want for testing? It might make the testing easier if we could configure the limit via an option, but only if we are not sure about `100`.

I will let @Dav1dde take a final look since he knows best cardinality limiting

jjbayer (commented):
Instead of overwriting limits defined in `sentry-metrics.cardinality-limiter.limits.{namespace.value}.per-org` with hardcoded values, should we instead define a second option `sentry-metrics.cardinality-limiter.limits.custom.per-org.low`, and then select the right option based on the feature flag?

vgrozdanic:
@ArthurKnaus we also want the relay to cardinality limits us so that we can query outcomes metric: [#74258](https://github.com/getsentry/sentry/issues/74258)

ArthurKnaus:
> @ArthurKnaus we also want the relay to cardinality limits us so that we can query outcomes metric: [#74258](https://github.com/getsentry/sentry/issues/74258)

Perfect, just wanted to make sure 👍

vgrozdanic:
@iambriccardo  @jjbayer initial plan was that this wouldn't be used for a long time, only for a short-lived development of the UI around cardinality limits. But when I gave it a second thought, it makes more sense to add extra option and based on the feature flag pick one, since we might need it in the future development :)

PR with options will be merged before this one: https://github.com/getsentry/sentry-options-automator/pull/1947/files

src/sentry/relay/config/__init__.py
```diff
@@ -241,7 +241,16 @@ def get_metrics_config(timeout: TimeChecker, project: Project) -> Mapping[str, A
     cardinality_limits: list[CardinalityLimit] = []
     for namespace in CARDINALITY_LIMIT_USE_CASES:
         timeout.check()
-        option = options.get(f"sentry-metrics.cardinality-limiter.limits.{namespace.value}.per-org")
+        if (
+            features.has("organizations:low-cardinality-limits-on-project", project.organization)
+            and namespace == UseCaseID.CUSTOM
+        ):
+            option = options.get("sentry-metrics.cardinality-limiter.limits.custom.per-org.low")
```
jjbayer, line 248: You'll need to register this option in `sentry.options.default.

nit: Does it make sense to save a per-project limit in an option called "per-org"? I like the idea of putting the scope in the option, but maybe we should remove `per-org` from the option name.

jjbayer:
> PR with options will be merged before this one

@vgrozdanic you'll need to register the option before setting it, but you can set the desired values as default when registering the option to make sure that you have sensible settings once you merge this PR (strictly speaking you don't even need the automator PR, but it's good to set the values explicitly).

src/sentry/relay/config/__init__.py
vgrozdanic: I agree, options is renamed and I have added it to defaults file

vgrozdanic:
@jjbayer I can close the other PR to add options to automator, but maybe it's better if we also have it as explicit option so that can more easily be configured without the need to deploy everything if we just want to change the limit to something smaller/larger

jjbayer (approved):

src/sentry/relay/config/__init__.py
```diff
@@ -256,7 +265,7 @@ def get_metrics_config(timeout: TimeChecker, project: Project) -> Mapping[str, A
                 "granularitySeconds": quota["granularity_seconds"],
             },
             "limit": quota["limit"],
-            "scope": "organization",
+            "scope": options.get("scope", "organization"),
```
Dav1dde, line 268: You query a global `scope` option for this?

Dav1dde (changes requested):
Let's not do this.

- The option your modifying is only there for legacy reasons (compatibility with the old cardinality limiter). You're adding more cruft to it.
- We don't need a feature flag for this

I showed you this yesterday, add your project allow list as another field/filter to here: https://github.com/getsentry/sentry/blob/5a93c4016808d1e935b073becc0856b05b200bc8/src/sentry/relay/config/__init__.py#L228-L231

Then you can freely configure limits as you please for testing.

src/sentry/relay/config/__init__.py
vgrozdanic: Nice catch, thank you

jjbayer: Whoops

jjbayer:
> The option your modifying is only there for legacy reasons (compatibility with the old cardinality limiter). You're adding more cruft to it.

@Dav1dde when reviewing this I wasn't aware that these are legacy options. Could you please clarify that in code somehow? E.g. split out into a function `_get_legacy_limits()` or just add a bunch of comments.

Dav1dde:
@jjbayer as with a lot of things this is on my todo to cleanup :(

vgrozdanic:
@jjbayer @Dav1dde changes applied

Dav1dde (approved):

PR #74337, opened 2024-07-16
fix(metrics): enable MetricsSummaries search by project

shellmayr:
The current implementation of project_slug_converter in filter_aliases.py converts a project:{slug} filter into a query condition on tags[project.id], which does not work. This PR fixes this by querying not for tags[project.id], but for the project_id column instead.

### DIFF
--- a/src/sentry/search/events/datasets/filter_aliases.py
+++ b/src/sentry/search/events/datasets/filter_aliases.py
@@ -102,7 +102,7 @@ def project_slug_converter(
         # Create a new search filter with the correct values
         converted_filter = builder.convert_search_filter_to_condition(
             SearchFilter(
-                SearchKey("project.id"),
+                SearchKey("project_id"),
                 search_filter.operator,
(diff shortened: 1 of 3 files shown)


src/sentry/search/events/datasets/filter_aliases.py
```diff
@@ -102,7 +102,7 @@ def project_slug_converter(
         # Create a new search filter with the correct values
         converted_filter = builder.convert_search_filter_to_condition(
             SearchFilter(
-                SearchKey("project.id"),
+                SearchKey(constants.PROJECT_ID_ALIAS),
```
Zylphrex, line 105: I'm not sure what other datasets may have `project_id` mapped the way you expect here. So I would recommend adding a mapping for `project.id` for metric summaries to `project_id` [here](https://github.com/getsentry/sentry/blob/master/src/sentry/utils/snuba.py#L175) to achieve the same effect without the possibility of impacting other datasets.

vgrozdanic (approved):

PR #74342, opened 2024-07-16
ref(insights): Remove unused `SpanOperationSelector`

gggritso:
Not used, and not even very good. It can go in the trash!


0Calories (approved):
🗑️

PR #74344, opened 2024-07-16
feat(discover): Save query dataset source

shruthilayaj:
Start tracking dataset source so we can distinguish between
user selected vs datasets inferred by us. Right now set everything
to `unknown`. Once the feature is released, new queries and updates
should set it to `user`.


narsaynorath (approved):
:+1:

PR #74345, opened 2024-07-16
fix(spans): Handle short span ids

Zylphrex:
Not a great solution but this relaxes the `is_span_id` check to allow for hex strings shorter than 16 chars. This is because it's very easy to accidentally get a hex string of 15 chars if the leading character is a 0. While we can be strict and require 16 chars, it's an easy to miss edge case so here we choose to allow it.

Fixes SENTRY-3BXV

### DIFF
--- a/src/sentry/utils/validators.py
+++ b/src/sentry/utils/validators.py
@@ -11,7 +11,7 @@ INVALID_SPAN_ID = (
     "`{}` must be a valid 16 character hex (containing only digits, or a-f characters)"
 )
 
-HEXADECIMAL_16_DIGITS = re.compile("^[0-9a-fA-F]{16}$")
+HEXADECIMAL_16_DIGITS = re.compile("^[0-9a-fA-F]{1,16}$")
 
 
 def normalize_event_id(value):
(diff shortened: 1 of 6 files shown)


src/sentry/utils/validators.py
```diff
@@ -11,7 +11,7 @@
     "`{}` must be a valid 16 character hex (containing only digits, or a-f characters)"
 )
 
-HEXADECIMAL_16_DIGITS = re.compile("^[0-9a-fA-F]{16}$")
+HEXADECIMAL_16_DIGITS = re.compile("^[0-9a-fA-F]{1,16}$")
```
wmak, line 14: Is this used anywhere else? Should we introduce a snuba_span_id regex, since this only applies to spans in the snuba datastore?

wmak (approved):
1 comment but lgtm

src/sentry/utils/validators.py
Zylphrex: So this is used by https://github.com/getsentry/sentry/blob/2cb46fe05401d7917e6a16c8963752eb95f4667d/src/sentry/utils/validators.py#L28-L29 which in turn is used by https://github.com/getsentry/sentry/blob/2cb46fe05401d7917e6a16c8963752eb95f4667d/src/sentry/api/event_search.py#L455-L464

The callers for this are
- the query builder: this is the case we're trying to fix here
- the old json query builder: this should be applicable there as well
- the endpoint for suspect spans to validate span groups: this should be applicable there as well

PR #74346, opened 2024-07-16
feat(profiling) link to span continuous profiles

JonasBa:
Link from all event types to continuous profiles

Also fixes https://sentry.sentry.io/issues/5585348781/?project=11276

### DIFF
--- a/static/app/components/profiling/flamegraph/continuousFlamegraph.tsx
+++ b/static/app/components/profiling/flamegraph/continuousFlamegraph.tsx
@@ -21,6 +21,8 @@ import {FlamegraphViewSelectMenu} from 'sentry/components/profiling/flamegraph/f
 import {FlamegraphZoomView} from 'sentry/components/profiling/flamegraph/flamegraphZoomView';
 import {FlamegraphZoomViewMinimap} from 'sentry/components/profiling/flamegraph/flamegraphZoomViewMinimap';
 import {t} from 'sentry/locale';
+import type {EntrySpans, EventTransaction} from 'sentry/types/event';
+import {EntryType} from 'sentry/types/event';
 import {defined} from 'sentry/utils';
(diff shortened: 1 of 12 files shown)


static/app/views/performance/newTraceDetails/traceDrawer/traceProfilingLink.ts
```diff
@@ -20,6 +24,20 @@ function toDate(value: unknown): Date | null {
   return dateObj;
 }
 
+function getTimeOffsetForDuration(duration: number): number {
+  return duration < 100 ? 100 : duration * 0.1;
```
JonasBa, line 28: 100ms window of time before/after short duration events and 10% of time for longer duration events

static/app/components/profiling/flamegraph/continuousFlamegraph.tsx
```diff
@@ -219,6 +250,28 @@ export function ContinuousFlamegraph(): ReactElement {
...
+  const spanTree: SpanTree = useMemo(() => {
+    if (segment.type === 'resolved' && segment.data) {
+      return new SpanTree(
+        segment.data,
+        collectAllSpanEntriesFromTransaction(segment.data)
+      );
+    }
+
+    return LOADING_OR_FALLBACK_SPAN_TREE;
+  }, [segment]);
```
Zylphrex, line 262: We talked about this but it'd be great if we could pick a different name other than `segment` for hits.

Zylphrex (approved):

static/app/components/profiling/flamegraph/continuousFlamegraph.tsx
JonasBa: Yeah, I went ahead and renamed it to eventId

PR #74347, opened 2024-07-16
fix(uptime): Catch errors when we fail to parse robots.txt

wedamija:
There are lots of cases where fetching robots can fail. If we can't fetch it, we just want to assume we shouldn't monitor the url.

Fixes SENTRY-3C3P
Fixes SENTRY-3C3Q
Fixes SENTRY-3C3S
Fixes SENTRY-3C42
Fixes SENTRY-3C3Z


davidenwang (approved):
good stuff ty!

PR #74351, opened 2024-07-16
fix(empty-states): Fix missing configuration instructions

roggenkemper:
this pr fixes a bug with missing configuration instructions that sometimes comes up in the issue stream empty states for java platforms

### DIFF
--- a/static/app/components/updatedEmptyState.tsx
+++ b/static/app/components/updatedEmptyState.tsx
@@ -75,11 +75,18 @@ export default function UpdatedEmptyState({project}: {project?: Project}) {
     replayOptions: {block: true, mask: true},
   };
 
+  if (currentPlatformKey === 'java' || currentPlatformKey === 'java-spring-boot') {
+    docParams.platformOptions = {
+      ...docParams.platformOptions,
+      packageManager: 'gradle',
+    };
+  }
+
   const install = loadGettingStarted.docs.onboarding.install(docParams)[0];
   const configure = loadGettingStarted.docs.onboarding.configure(docParams);
   const verify = loadGettingStarted.docs.onboarding.verify(docParams);
 
-  const {description: installDescription} = install;
+  const {description: installDescription, additionalInfo: installInfo} = install;
 
   const installConfigurations = install.configurations ?? [];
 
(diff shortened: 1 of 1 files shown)


static/app/components/updatedEmptyState.tsx
```diff
@@ -75,11 +75,18 @@ export default function UpdatedEmptyState({project}: {project?: Project}) {
     replayOptions: {block: true, mask: true},
   };
 
+  if (currentPlatformKey === 'java' || currentPlatformKey === 'java-spring-boot') {
+    docParams.platformOptions = {
+      ...docParams.platformOptions,
+      packageManager: 'gradle',
```
malwilley, line 81: Is there a better place for this config to live?

roggenkemper: i was thinking about it and im don't think there is yet, typically this changes depending on what set up you want to do , but since this is issues focused and tailored for the empty states experience, with some exceptions for platforms, i think this might be the best place for now

static/app/components/updatedEmptyState.tsx
```diff
@@ -180,6 +188,15 @@ export default function UpdatedEmptyState({project}: {project?: Project}) {
                               )
                             ) : null}
                           </CodeSnippetWrapper>
+                          <CodeSnippetWrapper>
+                            {configuration.configurations ? (
+                              Array.isArray(configuration.configurations[0].code) ? (
```
malwilley, line 193: Can `configurations` be an array with length 0? If so this would error out

roggenkemper: shouldn't be possible - but ill add an extra check to be safe

malwilley (approved):

static/app/components/updatedEmptyState.tsx
```diff
@@ -189,7 +189,8 @@ export default function UpdatedEmptyState({project}: {project?: Project}) {
                             ) : null}
                           </CodeSnippetWrapper>
                           <CodeSnippetWrapper>
-                            {configuration.configurations ? (
+                            {configuration.configurations &&
+                            configuration.configurations.length > 0 ? (
```
malwilley, line 193: nit: could also do `configuration.configurations?.length > 0`

PR #74352, opened 2024-07-16
feat(scm): pull request file issues API

cathteng:
API for users to upload their PR file information (filename, patch) to find issues associated with those files. This API can fetch issues for a single file.

This expands the scope of open PR issue comments beyond GitHub, and people can hit the API in their custom flows as well.

The API supports the ability to set a max limit on the number of issues to return, open PR comments only fetches a max of 5 per file -- using that as the default here.

### DIFF
--- a/src/sentry/api/endpoints/organization_pull_request_file_issues.py
+++ b/src/sentry/api/endpoints/organization_pull_request_file_issues.py
@@ -14,7 +14,7 @@ from sentry.snuba.referrer import Referrer
 from sentry.tasks.integrations.github.language_parsers import PATCH_PARSERS
 from sentry.tasks.integrations.github.open_pr_comment import (
     get_projects_and_filenames_from_source_file,
-    get_top_5_issues_by_count_for_file,
+    get_top_issues_by_count_for_file,
 )
 
 
@@ -22,6 +22,7 @@ class PullRequestFileSerializer(serializers.Serializer):
     filename = serializers.CharField(required=True)
     repo = serializers.CharField(required=True)
     patch = serializers.CharField(required=True)
+    limit = serializers.IntegerField(required=False, default=5)
 
     def validate_filename(self, value):
         if not value:
@@ -31,6 +32,12 @@ class PullRequestFileSerializer(serializers.Serializer):
         language_parser = PATCH_PARSERS.get(file_extension, None)
         if not language_parser:
             raise serializers.ValidationError("Unsupported file type")
+
+        return value
+
+    def validate_limit(self, value):
+        if value and value < 1 or value > 100:
+            raise serializers.ValidationError("Issue count must be between 1 and 100")
 
         return value
 
@@ -50,6 +57,7 @@ class OrganizationPullRequestFilesIssuesEndpoint(OrganizationEndpoint):
         filename = serializer.validated_data["filename"]
         repo_name = serializer.validated_data["repo"]
         patch = serializer.validated_data["patch"]
+        limit = serializer.validated_data["limit"]
 
         projects, sentry_filenames = get_projects_and_filenames_from_source_file(
             org_id=organization.id, repo_name=repo_name, pr_filename=filename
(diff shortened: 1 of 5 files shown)


src/sentry/api/endpoints/organization_pull_request_file_issues.py
```diff
@@ -0,0 +1,106 @@
...
+                    referrer=Referrer.TAGSTORE_GET_GROUPS_USER_COUNTS_OPEN_PR_COMMENT.value
+                ),
+                "event_count": group_id_to_info[issue.id]["event_count"],
+                "function_name": group_id_to_info[issue.id]["function_name"],
+            }
+            for issue in issues
+        ]
+        pr_file_issues.sort(key=lambda k: k.get("event_count", 0), reverse=True)
+
+        return Response(pr_file_issues)
```
iamrajjoshi, line 106: nit - if the end goal is for others to use this api, i would use a serializer here to return to make life less painful when ur documenting the endpoint. or u can just kick the can when u get there.

src/sentry/api/endpoints/organization_pull_request_file_issues.py
```diff
@@ -0,0 +1,106 @@
...
+from sentry.models.organization import Organization
+from sentry.snuba.referrer import Referrer
+from sentry.tasks.integrations.github.language_parsers import PATCH_PARSERS
+from sentry.tasks.integrations.github.open_pr_comment import (
+    get_projects_and_filenames_from_source_file,
+    get_top_issues_by_count_for_file,
+)
+
+
+class PullRequestFileSerializer(serializers.Serializer):
```
sentaur-athena, line 21: How is this gets related to a PR? Is it by `patch`? Or I can pass any file and get issues for it regardless of PR?

src/sentry/api/endpoints/organization_pull_request_file_issues.py
```diff
@@ -0,0 +1,106 @@
...
+
+        file_extension = value.split(".")[-1]
+        language_parser = PATCH_PARSERS.get(file_extension, None)
+        if not language_parser:
+            raise serializers.ValidationError("Unsupported file type")
+
+        return value
+
+    def validate_limit(self, value):
+        if value and value < 1 or value > 100:
```
RyanSkonnord, line 39: Should be `if value is not None`? It looks like this won't raise an error on an int value of 0.

Also, nit: `value < 1 or value > 100` can be written as `not (1 <= value <= 100)`.

src/sentry/api/endpoints/organization_pull_request_file_issues.py
```diff
@@ -0,0 +1,106 @@
...
+        filename = serializer.validated_data["filename"]
+        repo_name = serializer.validated_data["repo"]
+        patch = serializer.validated_data["patch"]
+        limit = serializer.validated_data["limit"]
+
+        projects, sentry_filenames = get_projects_and_filenames_from_source_file(
+            org_id=organization.id, repo_name=repo_name, pr_filename=filename
+        )
+
+        if not len(projects) or not len(sentry_filenames):
```
RyanSkonnord, line 66: Style nit: empty collections are falsy, so you don't need `len`. Just `if not projects or not sentry_filenames` will work.

I'd also consider writing `if not (projects and sentry_filenames)`, but that's a matter of taste.

src/sentry/api/endpoints/organization_pull_request_file_issues.py
```diff
@@ -0,0 +1,106 @@
...
+            projects=list(projects),
+            sentry_filenames=list(sentry_filenames),
+            function_names=list(function_names),
+            limit=limit,
+        )
+
+        group_id_to_info = {}
+        for issue in top_issues:
+            group_id = issue["group_id"]
+            group_id_to_info[group_id] = dict(filter(lambda k: k[0] != "group_id", issue.items()))
```
RyanSkonnord, line 87:
```suggestion
            group_id_to_info[group_id] = {k: v for (k, v) in issue.items() if k != "group_id"}
```

PR #74353, opened 2024-07-16
feat(uptime): Disable auto-detection for a project once we've starting monitoring a url for it

wedamija:
We want to disable detection for a project once we've started monitoring it. We may still use the ranking to keep track of suggested urls in the future, so we'll still keep a lot of this code around


src/sentry/uptime/detectors/ranking.py
```diff
@@ -126,3 +126,7 @@ def delete_project_bucket(bucket: datetime) -> None:
     key = get_project_bucket_key_for_datetime(bucket)
     cluster = _get_cluster()
     cluster.delete(key)
+
+
+def should_detect_for_project(project: Project) -> bool:
```
davidenwang, line 131: should we also replace the tests in `test_tasks.py:129`?

davidenwang (approved):

PR #74364, opened 2024-07-16
ref(alert): Register ActionHandler factories instead of classes

RyanSkonnord:
Refactor to no longer rely on there being a distinct `ActionHandler` subclass for each provider/service type. Replace the `AlertRuleTriggerAction.TypeRegistration` class with a new factory class that carries the same metadata and can provide different `ActionHandler`s on a per-instance basis.


RyanSkonnord:
The motivating change (i.e., why it's such a big deal that we don't have a unique `ActionHandler` subclass for each service) can be seen here: https://github.com/getsentry/sentry/pull/74070/commits/4a8860fb776bb7f929cfeaeb344f6f316fa8ab7b (context: https://github.com/getsentry/sentry/pull/74070)

src/sentry/incidents/models/alert_rule.py
```diff
@@ -496,6 +496,59 @@ def get_queryset(self) -> BaseQuerySet[AlertRuleTriggerAction]:
         return super().get_queryset().exclude(status=ObjectStatus.PENDING_DELETION)
 
 
+class AlertRuleActionHandlerFactory(abc.ABC):
```
RyanSkonnord, line 499: I know this name is something out of an OOP nightmare. I'm open to suggestions.

tests/sentry/incidents/test_tasks.py
```diff
@@ -208,7 +208,7 @@ def test_missing_project(self):
         )
 
     def test(self):
-        with patch.object(AlertRuleTriggerAction, "_type_registrations", new={}):
+        with patch.object(AlertRuleTriggerAction, "_factory_registrations", new={}):
```
RyanSkonnord, line 211: I'm not thrilled about the way these tests require visibility into the private registry. I might take a pass at it while I'm here.

src/sentry/incidents/endpoints/organization_alert_rule_available_action_index.py
```diff
@@ -29,7 +29,7 @@
 
 @region_silo_function
 def build_action_response(
-    registered_type: AlertRuleTriggerAction.TypeRegistration,
+    registered_type: AlertRuleActionHandlerFactory,
```
wedamija, line 32: Nit: `factory`?

src/sentry/incidents/models/alert_rule.py
wedamija: Maybe it can just be `ActionHandlerFactory` since it's already in the `incidents` project and it's specific to that?

src/sentry/incidents/models/alert_rule.py
```diff
@@ -496,6 +496,59 @@ def get_queryset(self) -> BaseQuerySet[AlertRuleTriggerAction]:
...
+    def build_handler(
+        self,
+        action: AlertRuleTriggerAction,
+        incident: Incident,
+        project: Project,
+    ) -> ActionHandler:
+        raise NotImplementedError
+
+
+class _AlertRuleActionHandlerClassFactory(AlertRuleActionHandlerFactory):
```
wedamija, line 528: The above name might make sense, since now this is a factory specific to alert rules?

wedamija (approved):

PR #74367, opened 2024-07-16
feat(query-builder): Add placeholder prop

### LINKED ISSUE #74242: SearchQueryBuilder: Support `placeholder` prop

malwilley:
Closes https://github.com/getsentry/sentry/issues/74242

Adds a `placeholder` prop and applies it on the issue stream.

CleanShot 2024-07-16 at 12 38 07

### DIFF
--- a/static/app/components/growingInput.tsx
+++ b/static/app/components/growingInput.tsx
@@ -50,7 +50,7 @@ export const GrowingInput = forwardRef<HTMLInputElement, InputProps>(
       if (isControlled && inputRef.current) {
         resize(inputRef.current);
       }
-    }, [props.value, isControlled]);
+    }, [props.value, props.placeholder, isControlled]);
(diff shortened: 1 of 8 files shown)


static/app/components/growingInput.tsx
```diff
@@ -50,7 +50,7 @@ export const GrowingInput = forwardRef<HTMLInputElement, InputProps>(
       if (isControlled && inputRef.current) {
         resize(inputRef.current);
       }
-    }, [props.value, isControlled]);
+    }, [props.value, props.placeholder, isControlled]);
```
malwilley, line 53: The `resize` function uses the input placeholder, so it makes sense to resize when the placeholder changes (required here since we change depending on the query value)

roggenkemper (approved):

PR #74368, opened 2024-07-16
chore(grouping): Remove deprecated configs

armenzg:
Deprecated configs.


vartec:
Hiding part ✅ 
Removing — did you validate that there aren't any projects still using it?

armenzg:
> Hiding part ✅ Removing — did you validate that there aren't any projects still using it?

I'm going to undo the hiding part just in case customer support uses it.

Nevertheless, customers cannot see this drop-down.

vartec (approved):
there's zero project using either of these, so it's all good!

PR #74371, opened 2024-07-16
feat(api): Added Project ID in Body Support for Couple APIs

iamrajjoshi:
I am working on getting all the commands in our Sentry CLI to support Ids as well as Slugs. [rough spec](https://www.notion.so/sentry/Supporting-IDs-and-Slugs-in-Sentry-CLI-35b805ea169e47348805672cdee41297?pm=c)

We have already gone through the effort of supporting ids and slugs in path parameters of all the APIs in our codebase. There are some endpoints that the CLI uses that pass in project slugs as body parameters, so we need to support Ids for those as well.

### DIFF
--- a/src/sentry/api/endpoints/organization_releases.py
+++ b/src/sentry/api/endpoints/organization_releases.py
@@ -466,16 +466,17 @@ class OrganizationReleasesEndpoint(
             result = serializer.validated_data
             scope.set_tag("version", result["version"])
 
+            # Get all projects that are available to the user/token
+            # Note: Does not use the "projects" data param from the request
             projects_from_request = self.get_projects(request, organization)
-            allowed_projects = {
-                k: project for project in projects_from_request for k in (project.slug, project.id)
-            }
+            allowed_projects = {project.slug: project for project in projects_from_request}
+            allowed_projects.update({project.id: project for project in projects_from_request})
 
             projects = []
-            for slug in result["projects"]:
-                if slug not in allowed_projects:
+            for id_or_slug in result["projects"]:
+                if id_or_slug not in allowed_projects:
                     return Response({"projects": ["Invalid project ids or slugs"]}, status=400)
-                projects.append(allowed_projects[slug])
+                projects.append(allowed_projects[id_or_slug])
 
             new_status = result.get("status")
             owner_id: int | None = None
--- a/src/sentry/api/endpoints/organization_releases.py
+++ b/src/sentry/api/endpoints/organization_releases.py
@@ -469,8 +469,10 @@ class OrganizationReleasesEndpoint(
             # Get all projects that are available to the user/token
             # Note: Does not use the "projects" data param from the request
             projects_from_request = self.get_projects(request, organization)
-            allowed_projects = {project.slug: project for project in projects_from_request}
-            allowed_projects.update({project.id: project for project in projects_from_request})
+            allowed_projects = {}
+            for project in projects_from_request:
+                allowed_projects[project.slug] = project
+                allowed_projects[project.id] = project
 
             projects = []
             for id_or_slug in result["projects"]:
--- a/src/sentry/api/endpoints/organization_releases.py
+++ b/src/sentry/api/endpoints/organization_releases.py
@@ -351,9 +351,11 @@ class OrganizationReleasesEndpoint(
                     organization.id,
                     filter_params["project_id"],
                     release_versions,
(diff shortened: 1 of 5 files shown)


src/sentry/api/endpoints/organization_releases.py
```diff
@@ -351,9 +351,11 @@ def qs_load_func(queryset, total_offset, qs_offset, limit):
...
                     release_versions,
-                    filter_params["start"]
-                    if filter_params["start"]
-                    else datetime.utcnow() - timedelta(days=90),
+                    (
+                        filter_params["start"]
+                        if filter_params["start"]
+                        else datetime.utcnow() - timedelta(days=90)
+                    ),
                     filter_params["end"] if filter_params["end"] else datetime.utcnow(),
```
iamrajjoshi, line 359: auto formatted

src/sentry/api/serializers/rest_framework/project.py
```diff
@@ -9,16 +9,22 @@
 
 @extend_schema_field(field=OpenApiTypes.STR)
 class ProjectField(serializers.Field):
-    def __init__(self, scope="project:write"):
+    def __init__(self, scope="project:write", id_allowed=False):
```
iamrajjoshi, line 12: we should eventually support id lookup everywhere this field is used, but didn't want to pollute this PR with the implications of that

src/sentry/api/endpoints/organization_releases.py
```diff
@@ -464,12 +466,15 @@ def post(self, request: Request, organization) -> Response:
...
 
-            allowed_projects = {p.slug: p for p in self.get_projects(request, organization)}
+            projects_from_request = self.get_projects(request, organization)
+            allowed_projects = {
+                k: project for project in projects_from_request for k in (project.slug, project.id)
+            }
 
             projects = []
             for slug in result["projects"]:
                 if slug not in allowed_projects:
```
schew2381, line 476: I think this logic would have to be updated to reflect the new mapping

src/sentry/api/endpoints/organization_releases.py
```diff
@@ -464,12 +466,15 @@ def post(self, request: Request, organization) -> Response:
             result = serializer.validated_data
             scope.set_tag("version", result["version"])
 
-            allowed_projects = {p.slug: p for p in self.get_projects(request, organization)}
+            projects_from_request = self.get_projects(request, organization)
```
schew2381, line 469: I took a look at `get_projects` but I couldn't figure out how this would work when the `projects` is a list of slugs and IDs. Did you take a look at the function?

iamrajjoshi: discussed offline

src/sentry/api/endpoints/organization_releases.py
```diff
@@ -466,16 +466,17 @@ def post(self, request: Request, organization) -> Response:
...
             scope.set_tag("version", result["version"])
 
+            # Get all projects that are available to the user/token
+            # Note: Does not use the "projects" data param from the request
             projects_from_request = self.get_projects(request, organization)
-            allowed_projects = {
-                k: project for project in projects_from_request for k in (project.slug, project.id)
-            }
+            allowed_projects = {project.slug: project for project in projects_from_request}
+            allowed_projects.update({project.id: project for project in projects_from_request})
```
schew2381, line 473: two loops smh

schew2381: not optimal enough

iamrajjoshi: smh i had something better before, what do u want me to do 😭

schew2381: how about:
```Python
allowed_projects = {}
for project in projects_from_request:
    allowed_projects[project.slug] = project
    allowed_projects[project.id] = project
```

schew2381: nice and simple

schew2381: nothing fancy

iamrajjoshi: traditional loops 🙄 , looks good to me

schew2381: kids these days 👴

schew2381 (approved):

PR #74372, opened 2024-07-16
ref(similarity): Use limit in backfill query

jangjodi:
Before:
- postgres query was on all groups and would sometimes time out
- groups were ordered based on times_seen
- calls to the next backfill used index of the returned groups from the query as an offset

Now:
- postgres query using limit to perform the query in chunks of `batch_size`, using id as offset
- groups are ordered by descending id
- calls to the next backfill use the last processed group id as an offset

Note: Ordering by descending id ends up prioritizing the last seen group of a set of Seer matches (whereas before this change we were prioritizing most frequently seen). Neither one matches ingest (which will by default prioritize first seen, since any later ones won't have come along yet), but over time records coming from the backfill will make up a smaller and smaller proportion of a project's Seer records. so the difference will effectively eventually come out in the wash.


src/sentry/tasks/embeddings_grouping/backfill_seer_grouping_records_for_project.py
```diff
@@ -223,50 +209,40 @@ def backfill_seer_grouping_records_for_project(
...
     cohort: str | list[int] | None = None,
     only_delete: bool = False,
 ):
-
-    if last_group_id is not None:
+    if last_processed_group_id is not None:
         redis_client.set(
-            f"{make_backfill_grouping_index_redis_key(project_id)}",
-            last_processed_group_index if last_processed_group_index is not None else 0,
+            f"{make_backfill_grouping_id_redis_key(project_id)}",
```
vartec, line 231:
```suggestion
            make_backfill_grouping_id_redis_key(project_id),
```
it's already a string

vartec (approved):
lgtm

src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -126,46 +126,48 @@ def initialize_backfill(
...
             last_seen__gt=(datetime.now(UTC) - timedelta(days=90)),
-        )
-        .exclude(status__in=[GroupStatus.PENDING_DELETION, GroupStatus.DELETION_IN_PROGRESS])
-        .values_list("id", "data")
-        .order_by("-times_seen", "id")
+        ).exclude(status__in=[GroupStatus.PENDING_DELETION, GroupStatus.DELETION_IN_PROGRESS]),
+        step=-1000,
+        limit=batch_size,
+    )
+    groups_to_backfill_batch = [(group.id, group.data) for group in groups_to_backfill_query]
```
wedamija, line 149: Since I think you're only using `group.id` and `group.data` you can use `values_list` to bring down only the fields you need.

jangjodi:
Thanks @wedamija! I've made the changes you suggested

src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -126,46 +126,56 @@ def initialize_backfill(
...
+    ]
 
-    batch_end_index = min(last_processed_group_index + batch_size, total_groups_to_backfill_length)
-    groups_to_backfill_batch = groups_to_backfill_query[last_processed_group_index:batch_end_index]
+    total_groups_to_backfill_length = len(groups_to_backfill_batch)
+    batch_end_group_id = (
+        groups_to_backfill_batch[total_groups_to_backfill_length - 1][0]
+        if total_groups_to_backfill_length
+        else None
+    )
```
wedamija, line 161: It might be a good idea to get this id before you filter out rows in memory, just in case you happen to fetch a bunch of things that end up filtered out.

wedamija (commented):
lgtm, I'll leave it to your team just to validate the rest of the logic

src/sentry/tasks/embeddings_grouping/utils.py
jangjodi: good catch, thanks!

src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -94,27 +95,26 @@ def filter_snuba_results(snuba_results, groups_to_backfill_with_no_embedding, pr
...
-    if last_processed_group_index is None:
-        last_processed_group_index_ret = int(
-            redis_client.get(make_backfill_grouping_index_redis_key(project_id)) or 0
-        )
+    last_processed_group_id_ret = redis_client.get(make_backfill_grouping_id_redis_key(project_id))
+    if last_processed_group_id is None and last_processed_group_id_ret is not None:
+        last_processed_group_id_ret = int(last_processed_group_id_ret)
     else:
-        last_processed_group_index_ret = last_processed_group_index
+        last_processed_group_id_ret = last_processed_group_id
```
lobsterkatie, line 117: Probably-for-a-different-PR nit: 

If I understand correctly what's happening in these lines, I think this would be easier to grock like this:

(it won't let me use the suggestion function for some reason)

```python
    last_processed_group_id = last_processed_group_id or int(
        redis_client.get(make_backfill_grouping_id_redis_key(project_id))
    )
```

and then just return `last_processed_group_id`.

lobsterkatie (approved):
The logic makes sense!

PR #74374, opened 2024-07-16
ref(replay): append DOM root inside hidden iframe in createHiddenPlayer

michellewzhang:
relates to performance issues described in https://github.com/getsentry/team-replay/issues/450

**context:**
we want to alleviate the amount of [layout thrashing](https://web.dev/articles/avoid-large-complex-layouts-and-layout-thrashing?utm_source=devtools#avoid-forced-synchronous-layouts) happening (purple gaps in pic below), which is causing performance slowdowns on the breadcrumbs tab and causing certain replays to crash. adding a hidden iframe means it won't affect the main DOM layout. we don't care how the DOM root/Replayer instance renders because in practice, we are just using it to generate the HTML to display in the breadcrumbs tab.

image
taken from the below image, from a crashed replay:
image

new iframe in the hierarchy:

SCR-20240716-mfqa

### DIFF
--- a/static/app/utils/replays/createHiddenPlayer.tsx
+++ b/static/app/utils/replays/createHiddenPlayer.tsx
@@ -15,11 +15,13 @@ export function createHiddenPlayer(rrwebEvents: RecordingFrame[]): Replayer {
 
   // create a hidden iframe
   const hiddenIframe = document.createElement('iframe');
-  hiddenIframe.style.visibility = 'hidden';
+  hiddenIframe.style.display = 'none';
   document.body.appendChild(hiddenIframe);
 
   // append the DOM root inside the iframe
-  hiddenIframe.appendChild(domRoot);
+  if (hiddenIframe.contentDocument) {
+    hiddenIframe.contentDocument.body.appendChild(domRoot);
+  }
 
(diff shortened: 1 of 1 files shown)


billyvg (approved):
can you double check that this works for the DOM Nodes chart in the Memory tab?

c298lee:
how does the hidden iframe help alleviate layout thrashing?

michellewzhang:
> how does the hidden iframe help alleviate layout thrashing?

i tried to explain that in the description

c298lee:
> > how does the hidden iframe help alleviate layout thrashing?
> 
> i tried to explain that in the description

i am slow 😞 i get it now tyyy

michellewzhang:
> > > how does the hidden iframe help alleviate layout thrashing?
> > 
> > 
> > i tried to explain that in the description
> 
> i am slow 😞 i get it now tyyy

@c298lee aw no worries !!

michellewzhang:
> can you double check that this works for the DOM Nodes chart in the Memory tab?
SCR-20240716-mwjr

looks like we have 2 of these when on the memory tab, seems like expected behavior

PR #74376, opened 2024-07-16
ref(profiling) remove percentile duration chart from continuous profiling UI

JonasBa:
Removes the profile duration chart that we have on the profiling landing page as it will be removed with the release of continuous profiling

### DIFF
--- a/static/app/views/profiling/content.tsx
+++ b/static/app/views/profiling/content.tsx
@@ -213,12 +213,16 @@ function ProfilingContent({location}: ProfilingContentProps) {
                     'profiling-global-suspect-functions'
                   ) ? (
                     <Fragment>
-                      <ProfilesChartWidget
-                        chartHeight={150}
-                        referrer="api.profiling.landing-chart"
-                        userQuery={query}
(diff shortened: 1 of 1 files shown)


static/app/views/profiling/content.tsx
```diff
@@ -213,12 +213,16 @@ function ProfilingContent({location}: ProfilingContentProps) {
...
                     <Fragment>
-                      <ProfilesChartWidget
-                        chartHeight={150}
-                        referrer="api.profiling.landing-chart"
-                        userQuery={query}
-                        selection={selection}
-                      />
+                      {organization.features.includes(
+                        'continuous-profiling-ui'
+                      ) ? null : (
```
Zylphrex, line 218: We can add a message here for people who are looking for it later

Zylphrex (approved):

static/app/views/profiling/content.tsx
JonasBa: Yeah, I'm going to add a banner separately and likely add this to the docs pages

PR #74378, opened 2024-07-16
chore(profiling): Clean up existing flamegraph endpoint for continuou…

Zylphrex:
…s profiling

Remove some unused code around generating flamegraph from spans and add tests. Will follow up to support generating flamegraphs from both transaction and continuous profiles. In the future, we'll add back the span based flamegraphs.

### DIFF
--- a/src/sentry/profiles/profile_chunks.py
+++ b/src/sentry/profiles/profile_chunks.py
@@ -35,7 +35,8 @@ def get_chunk_ids(
             Condition(Column("project_id"), Op.EQ, project_id),
             Condition(Column("profiler_id"), Op.EQ, profiler_id),
         ],
-        orderby=[OrderBy(Column("start_timestamp"), Direction.ASC)],
+        # We want the generate the flamegraph using the newest data
+        orderby=[OrderBy(Column("start_timestamp"), Direction.DESC)],
     ).set_limit(max_chunks)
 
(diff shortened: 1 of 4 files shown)


phacops (approved):

src/sentry/profiles/profile_chunks.py
```diff
@@ -35,7 +35,8 @@ def get_chunk_ids(
             Condition(Column("project_id"), Op.EQ, project_id),
             Condition(Column("profiler_id"), Op.EQ, profiler_id),
         ],
-        orderby=[OrderBy(Column("start_timestamp"), Direction.ASC)],
+        # We want the generate the flamegraph using the newest data
```
viglia, line 38: @Zylphrex, small clarification: `get_chunk_ids` is only used for `flamechart` (merging several chunks into a unique one), not for flamegraph. The idea of using DESC was to start from the oldest data so that we could then move forward as we added pagination support (assuming continuous scrolling is something we're still interested in supporting).

Anyway, since we don't have pagination yet, I think getting the most recent data is the right call. Just wanted to clarify.

Zylphrex: Yep, going to leave this for now to prefer newer chunks for now. We'll think about pagination when we get there.

PR #74380, opened 2024-07-16
fix(api): catch invalid user ID

ameliahsu:
closes SENTRY-3C1N

added try-except in case of invalid user ID

### DIFF
--- a/src/sentry/api/endpoints/api_token_details.py
+++ b/src/sentry/api/endpoints/api_token_details.py
@@ -37,8 +37,8 @@ class ApiTokenDetailsEndpoint(Endpoint):
 
         try:
             user_id = get_appropriate_user_id(request=request)
-        except ValueError as e:
-            return Response(e, status=400)
+        except ValueError:
+            return Response({"detail": "Invalid user ID"}, status=400)
 
         try:
             instance = ApiToken.objects.get(id=token_id, application__isnull=True, user_id=user_id)
@@ -64,8 +64,8 @@ class ApiTokenDetailsEndpoint(Endpoint):
 
         try:
             user_id = get_appropriate_user_id(request=request)
-        except ValueError as e:
-            return Response(e, status=400)
+        except ValueError:
+            return Response({"detail": "Invalid user ID"}, status=400)
 
         try:
             token_to_rename = ApiToken.objects.get(
@@ -83,8 +83,8 @@ class ApiTokenDetailsEndpoint(Endpoint):
     def delete(self, request: Request, token_id: int) -> Response:
         try:
             user_id = get_appropriate_user_id(request=request)
-        except ValueError as e:
-            return Response(e, status=400)
+        except ValueError:
+            return Response({"detail": "Invalid user ID"}, status=400)
 
         try:
             token_to_delete = ApiToken.objects.get(
--- a/src/sentry/api/endpoints/api_token_details.py
+++ b/src/sentry/api/endpoints/api_token_details.py
@@ -35,12 +35,15 @@ class ApiTokenDetailsEndpoint(Endpoint):
     @method_decorator(never_cache)
     def get(self, request: Request, token_id: int) -> Response:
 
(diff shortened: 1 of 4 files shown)


cathteng (commented):
should we be checking that `request.user.id` is numeric in `get_appropriate_user_id`, and error if it's not?

ameliahsu:
@cathteng would that check be in addition to the try-excepts or in replacement?

src/sentry/api/endpoints/api_token_details.py
```diff
@@ -35,12 +35,15 @@ class ApiTokenDetailsEndpoint(Endpoint):
...
+            user_id = get_appropriate_user_id(request=request)
+        except ValueError:
+            return Response({"detail": "Invalid user ID"}, status=400)
 
         try:
             instance = ApiToken.objects.get(id=token_id, application__isnull=True, user_id=user_id)
-        except ApiToken.DoesNotExist:
-            raise ResourceDoesNotExist
+        except ApiToken.DoesNotExist as e:
+            raise ResourceDoesNotExist from e
```
schew2381, line 46: Instead of exposing the error, we should write an explicit err msg like this example:
https://github.com/getsentry/sentry/blob/a1cd8f31af3e70e69943dd2e812ca519dc4cbdb0/src/sentry/plugins/providers/integration_repository.py#L80

cathteng:
@ameliahsu it's possible to raise an exception that also returns an API response https://github.com/getsentry/sentry/blob/a5d48aab48ad3dabc18068ef3e38658baddb68cc/src/sentry/api/exceptions.py#L15-L17, so you could get away with not having try excepts at all

src/sentry/api/endpoints/api_token_details.py
schew2381: Can also use the `ResourceDoesNotExist` exception with overriding the detail msg

ameliahsu: fixed!

schew2381 (commented):
could we copy the invalid user/token tests for each of the request methods (get, put, etc.) where it's applicable?

tests/sentry/api/endpoints/test_api_token_details.py
```diff
@@ -32,14 +32,21 @@ def test_never_cache(self):
...
-    def test_not_exists(self):
+    def test_invalid_token_id(self):
         self.login_as(self.user)
         self.get_error_response(-1, status_code=status.HTTP_404_NOT_FOUND)
 
     def test_no_auth(self):
         token = ApiToken.objects.create(user=self.user, name="token 1")
         self.get_error_response(token.id, status_code=status.HTTP_401_UNAUTHORIZED)
 
+    def test_unvalid_user_id(self):
```
schew2381, line 43: nit: small typo

tests/sentry/api/endpoints/test_api_token_details.py
```diff
@@ -120,6 +127,15 @@ def test_no_auth(self):
 
         self.get_error_response(token.id, status_code=status.HTTP_401_UNAUTHORIZED, **payload)
 
+    def test_unvalid_user_id(self):
```
schew2381, line 130: same

tests/sentry/api/endpoints/test_api_token_details.py
```diff
@@ -150,3 +166,11 @@ def test_no_auth(self):
         token = ApiToken.objects.create(user=self.user, name="token 1")
 
         self.get_error_response(token.id, status_code=status.HTTP_401_UNAUTHORIZED)
+
+    def test_unvalid_user_id(self):
```
schew2381, line 170: also same lol

schew2381 (approved):
small nit, but otherwise lgtm!

tests/sentry/api/endpoints/test_api_token_details.py
ameliahsu: brb crying

PR #74384, opened 2024-07-16
devenv: move install-js-dev to sync

joshuarli:
Frontend only devs can set `SENTRY_DEVENV_FRONTEND_ONLY=1` if they want to omit colima + python migrations.


scripts/lib.sh
```diff
@@ -89,18 +89,6 @@ node-version-check() {
...
-install-js-dev() {
-    node-version-check
-    echo "--> Installing Yarn packages (for development)"
-    # Use NODE_ENV=development so that yarn installs both dependencies + devDependencies
-    NODE_ENV=development yarn install --frozen-lockfile
-    # A common problem is with node packages not existing in `node_modules` even though `yarn install`
-    # says everything is up to date. Even though `yarn install` is run already, it doesn't take into
-    # account the state of the current filesystem (it only checks .yarn-integrity).
-    # Add an additional check against `node_modules`
-    yarn check --verify-tree || yarn install --check-files
```
joshuarli, line 101: removing this (added in https://github.com/getsentry/sentry/pull/13045) because it's probably no longer helpful, besides we don't build the production dockerfile this way

currently get `error "react-mentions#@babel/runtime" is wrong version: expected "7.4.5", got "7.24.7"` but neither `NODE_ENV=development yarn install --check-files` nor `NODE_ENV=development yarn install --frozen-lockfile resolves it`

joshuarli: billy says i can remove and see if anyone complains

IanWoodard (approved):

PR #74389, opened 2024-07-16
ref(replay): clean up replayer after stepping through `replayerStepper`

michellewzhang:
relates to https://github.com/getsentry/team-replay/issues/450

we create a `Replayer` instance every time we call `replayerStepper`. This PR cleans up the replayer & hidden iframe after we finish stepping through, which helps clean up some memory.

### DIFF
--- a/static/app/utils/replays/replayerStepper.tsx
+++ b/static/app/utils/replays/replayerStepper.tsx
@@ -49,6 +49,11 @@ export default function replayerStepper<
       resolve(collection);
       replayer.off('pause', handlePause);
       replayer.destroy();
+      Array.from(document.getElementsByTagName('iframe')).forEach(el => {
+        if (el.style.display === 'none') {
+          el.parentNode?.removeChild(el);
+        }
+      });
     };
 
     const nextOrDone = () => {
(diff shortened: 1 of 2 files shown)


static/app/utils/replays/replayerStepper.tsx
```diff
@@ -47,6 +47,8 @@ export default function replayerStepper<
 
     const onDone = () => {
       resolve(collection);
+      replayer.off('pause', handlePause);
```
michellewzhang, line 50: to avoid recursion, since `destroy()` calls `pause()`

static/app/utils/replays/replayerStepper.tsx
```diff
@@ -47,6 +47,13 @@ export default function replayerStepper<
...
+      replayer.destroy();
+      Array.from(document.getElementsByTagName('iframe')).forEach(el => {
```
michellewzhang, line 52: think i need a more specific way to clean up the desired iframe here

static/app/utils/replays/replayerStepper.tsx
billyvg: can you add this as a code comment?

billyvg (approved):

PR #74394, opened 2024-07-17
feat(snuba-search): Add a feature flag to snuba-first search logic

snigdhas:
This logic is still gated using the `useGroupSnubaDataset` url param. This PR adds an additional org-based feature flag so we can enable this internally for all users.


src/sentry/features/temporary.py
```diff
@@ -158,6 +158,7 @@ def register_temporary_features(manager: FeatureManager):
     manager.add("organizations:issue-stream-empty-state-additional-platforms", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE)
     # Enable issue stream performance improvements
     manager.add("organizations:issue-stream-performance", OrganizationFeature, FeatureHandlerStrategy.REMOTE)
+    manager.add("organizations:issue-search-snuba", OrganizationFeature, FeatureHandlerStrategy.REMOTE)
```
jangjodi, line 161: just curious, why is `REMOTE` used instead of `FLAGPOLE`?

snigdhas: oops old habits

jangjodi (approved):

PR #74397, opened 2024-07-17
feat(relocation): Add organization-fork endpoint

azaslavsky:
This endpoint triggers a SAAS->SAAS relocation. Since this relocation is really just a duplication of an existing org in another region (aka a "fork"), I've decided it made more sense to list under the org-slug of the org being exported, rather than under the `/relocation/` path.

### DIFF
--- a/src/sentry/api/endpoints/organization_fork.py
+++ b/src/sentry/api/endpoints/organization_fork.py
@@ -1,0 +1,158 @@
+import logging
+from string import Template
+
+from django.db import router
+from rest_framework import status
+from rest_framework.request import Request
+from rest_framework.response import Response
+
+from sentry.api.api_owners import ApiOwner
+from sentry.api.api_publish_status import ApiPublishStatus
+from sentry.api.base import Endpoint, region_silo_endpoint
+from sentry.api.endpoints.relocations.index import (
+    get_autopause_value,
+    validate_relocation_uniqueness,
+)
+from sentry.api.permissions import SuperuserOrStaffFeatureFlaggedPermission
+from sentry.api.serializers import serialize
+from sentry.hybridcloud.services.organization_mapping import organization_mapping_service
+from sentry.models.relocation import Relocation
+from sentry.organizations.services.organization import organization_service
+from sentry.tasks.relocation import uploading_start
+from sentry.types.region import get_local_region
+from sentry.utils.db import atomic_transaction
+
+ERR_DUPLICATE_ORGANIZATION_FORK = Template(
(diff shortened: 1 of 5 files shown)


src/sentry/api/endpoints/organization_fork.py
```diff
@@ -0,0 +1,158 @@
...
+from sentry.api.permissions import SuperuserOrStaffFeatureFlaggedPermission
+from sentry.api.serializers import serialize
+from sentry.hybridcloud.services.organization_mapping import organization_mapping_service
+from sentry.models.relocation import Relocation
+from sentry.organizations.services.organization import organization_service
+from sentry.tasks.relocation import uploading_start
+from sentry.types.region import get_local_region
+from sentry.utils.db import atomic_transaction
+
+ERR_DUPLICATE_ORGANIZATION_FORK = Template(
```
hubertdeng123, line 25: Will this api endpoint always be gated by superuser/staff? Curious because the wording around `fork` is pretty specific to developers and non-technical people may be confused by this terminology

src/sentry/api/endpoints/organization_fork.py
```diff
@@ -0,0 +1,158 @@
...
+
+        logger.info("relocations.fork.post.start", extra={"caller": request.user.id})
+
+        org_retrieval_args = {
+            "only_visible": True,
+            "include_projects": False,
+            "include_teams": False,
+        }
+        org_context = (
+            organization_service.get_organization_by_id(id=organization_id_or_slug)
```
hubertdeng123, line 71: Will this break if org slug is a number?

src/sentry/api/endpoints/organization_fork.py
azaslavsky: Yes, this process is not intended to be self-serve for the foreseeable future. If we ever change that decision, we can revisit the naming and/or make an alias API endpoint.

src/sentry/api/endpoints/organization_fork.py
azaslavsky: Valid org slugs cannot be a number:

https://github.com/getsentry/sentry/blob/d152e91d8ec0ed1a2816033ad87c9853ed0e0d40/src/sentry/slug/patterns.py#L12

This is checked upstream by the endpoint resolving regex, so if we're hitting this line of code, we can be sure that numerical == id, non-numerical == slug. This is also the check we use in other endpoints that take org ids or slugs interchangeably.

hubertdeng123 (approved):

PR #74398, opened 2024-07-17
feat(relocation): Add `relocation.forked` analytics event

azaslavsky:


src/sentry/analytics/events/relocation_forked.py
```diff
@@ -0,0 +1,17 @@
+from sentry import analytics
+
+
+class RelocationForkedEvent(analytics.Event):
```
hubertdeng123, line 4: What are we using for analytics? Google? Amplitude?

hubertdeng123 (approved):

src/sentry/analytics/events/relocation_forked.py
azaslavsky: Is there anything special we need to do for those? I remember you making a dashboard for org relocation some time ago, could we just have that pick this info up as well?

hubertdeng123: Oh I almost forgot I did that. If it's the same analytics as that, it is automatically populated in looker so it can pick this up

PR #74401, opened 2024-07-17
feat(replay): Fix replay restarting not working

### LINKED ISSUE #47963: Replay "restart" button does not correctly reset player.
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

watch a replay. then hit "restart replay"

### Expected Result

the replay is taken back to the beginning and starts over

### Actual Result

the replay is frozen on the last screen from the end of the replay, at least it looks like until the next full checkout happens in the replay player

### Product Area

Replays

### Link

_No response_

### DSN

_No response_

### Version

_No response_

billyvg:
The start of a replay does not always equal the start of the replay
_recording_. This is because we include buffered events in the replay
that occur before the recording is started. In order to align the start
and end times of the replay between the recording, the timeline, and
timestamps of the events inside of e.g. Network tab, we create a fake
rrweb event that occurs at relative timestamp "00:00" (when necessary).

However, this fake event would cause navigation to the start (00:00) and
"Restart Replay" functionality to be broken as the player expects the
snapshot event when this happens. Instead of the fake event, duplicate
the snapshot, but change timestamp to be at the start of the replay. It
is a bit misleading, but is actually similar to what happens now except
that restart works.

An alternative idea is to render a white screen which would illustrate
better that certain events occur before recording was started. However, I
anticipate more user confusion with this solution even though it is more
correct.

Fixes https://github.com/getsentry/sentry/issues/47963

### DIFF
--- a/static/app/utils/replays/hydrateRRWebRecordingFrames.tsx
+++ b/static/app/utils/replays/hydrateRRWebRecordingFrames.tsx
@@ -2,17 +2,6 @@ import type {RecordingFrame} from 'sentry/utils/replays/types';
 import {EventType} from 'sentry/utils/replays/types';
 import type {ReplayRecord} from 'sentry/views/replays/types';
 
-export function recordingStartFrame(replayRecord: ReplayRecord): RecordingFrame {
-  return {
-    type: EventType.Custom,
-    timestamp: replayRecord.started_at.getTime(),
-    data: {
-      tag: 'replay.start',
-      payload: {},
-    },
-  };
-}
-
 export function recordingEndFrame(replayRecord: ReplayRecord): RecordingFrame {
   return {
     type: EventType.Custom,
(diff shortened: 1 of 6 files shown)


ryan953 (approved):

static/app/utils/replays/hydrateRRWebRecordingFrames.tsx
```diff
@@ -2,17 +2,6 @@ import type {RecordingFrame} from 'sentry/utils/replays/types';
 import {EventType} from 'sentry/utils/replays/types';
 import type {ReplayRecord} from 'sentry/views/replays/types';
 
-export function recordingStartFrame(replayRecord: ReplayRecord): RecordingFrame {
-  return {
-    type: EventType.Custom,
-    timestamp: replayRecord.started_at.getTime(),
-    data: {
-      tag: 'replay.start',
```
ryan953, line 10: There should be other spots, like types.tsx, that reference `replay.start` and could be cleaned up

billyvg: Updated!

PR #74404, opened 2024-07-17
feat(dashboards): Adds public api docs for post and get /dashboards

edwardgou-sentry:
Documents the dashboards endpoint and makes GET and POST methods public API.


src/sentry/api/endpoints/organization_dashboards.py
```diff
@@ -43,28 +57,35 @@ def has_object_permission(self, request: Request, view, obj):
...
+        },
+        examples=DashboardExamples.DASHBOARDS_QUERY_RESPONSE,
+    )
     def get(self, request: Request, organization) -> Response:
         """
-        Retrieve an Organization's Dashboards
-        `````````````````````````````````````
-
         Retrieve a list of dashboards that are associated with the given organization.
         If on the first page, this endpoint will also include any pre-built dashboards
```
sentaur-athena, line 87: I don't quite understand this sentence `if on the first page`. Can you please expand on it?

sentaur-athena (commented):
For easier review, you can follow these steps and see how your published API looks like in docs: https://develop.sentry.dev/application/api/public/#building-and-testing-locally

src/sentry/api/endpoints/organization_dashboards.py
edwardgou-sentry: Hi, actually I think that sentence is no longer true since we've added dashboard templates to the product. I've removed this line! Also fixed some missing query parameters
image

lizokm (approved):
Looks good!

PR #74406, opened 2024-07-17
ref(replay): Replay timeline gaps using timestamp

c298lee:
Refactors replay timeline gaps to use frame timestamps and styles `left` and `width` instead of frame columns. Using timestamps instead of the frame column makes it line up with breadcrumbs better. Switching to styling `left` and `width` instead of the range of frames also fixed the gap bug when zooming in on the timeline.

Also modified the gap logic:
When there's a background frame, a gap is created until the next frame that's not a background frame or error frame. In most cases it will be the foreground frame, but in cases of bad data it could be any other replay frame.

Before:
image

After:
image
The gap is now a consistent `div` so there's no weird spots that the gap doesn't cover. Additionally, the gap ends as soon as there's a frame even if it isn't a foreground frame, which is more accurate.

### DIFF
--- a/static/app/components/replays/breadcrumbs/timelineGaps.tsx
+++ b/static/app/components/replays/breadcrumbs/timelineGaps.tsx
@@ -1,5 +1,4 @@
 import {Fragment} from 'react';
-import {css} from '@emotion/react';
 import styled from '@emotion/styled';
 
 import {Tooltip} from 'sentry/components/tooltip';
@@ -7,7 +6,7 @@ import {t} from 'sentry/locale';
 import toPercent from 'sentry/utils/number/toPercent';
 import {
   isBackgroundFrame,
-  isForegroundFrame,
+  isErrorFrame,
   type ReplayFrame,
 } from 'sentry/utils/replays/types';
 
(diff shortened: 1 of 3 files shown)


static/app/components/replays/breadcrumbs/timelineGaps.tsx
```diff
@@ -13,76 +14,71 @@ import {
...
-const Column = styled(Timeline.Col)<{column: number}>`
-  grid-column: ${p => p.column};
-  line-height: 14px;
+const Range = styled('div')`
+  position: absolute;
 `;
 
 const Gap = styled('span')`
   background: ${p => p.theme.gray400};
+  display: block;
```
billyvg, line 80: You can use `div` of a `span`, which is `block` element by default.

static/app/components/replays/breadcrumbs/timelineGaps.tsx
```diff
@@ -13,76 +14,71 @@ import {
...
-          <Tooltip title={t('App is suspended')} isHoverable containerDisplayMode="grid">
-            <Gap style={{width: `${markerWidth}px`}} />
-          </Tooltip>
-        </Column>
-      ))}
-    </Timeline.Columns>
+    <Fragment>
+      {ranges.map(rangeCss => {
+        return (
+          <Range key={JSON.stringify(rangeCss)} css={css``} style={rangeCss}>
```
billyvg, line 59: Empty `css` prop

```suggestion
          <Range key={JSON.stringify(rangeCss)} style={rangeCss}>
```

billyvg: Also, it likely won't have a meaningful impact here, but `JSON.stringify()` is going to be slower than string interpolation, I would create your stringed key

billyvg (approved):
Do we handle the case where (maybe due to bad data), we have a start, but no end and it's not the end of the replay?

e.g:


background....??..background....foreground

c298lee:
> Do we handle the case where (maybe due to bad data), we have a start, but no end and it's not the end of the replay?
> 
> e.g:
> 
> background....??..background....foreground

We do not at the moment. We assume that the app is in the background until we get a foreground frame. Hmmm I can add some logic to ignore the background frame if the next frame isn't a foreground!

static/app/components/replays/breadcrumbs/timelineGaps.tsx
```diff
@@ -1,9 +1,10 @@
+import {Fragment} from 'react';
+import {css} from '@emotion/react';
```
ryan953, line 2:
```suggestion
```

static/app/components/replays/breadcrumbs/timelineGaps.tsx
ryan953: with interpolation:
```suggestion
          <Range key={`${rangeCss.left]-${rangeCss,width}`} style={rangeCss}>
```

static/app/components/replays/breadcrumbs/timelineGaps.tsx
```diff
@@ -13,76 +14,71 @@ import {
...
-
-  const gapCol: number[] = [];
-
-  const gapFrames = framesByCol.entries();
+export default function TimelineGaps({durationMs, startTimestampMs, frames}: Props) {
+  const ranges: Array<{left: string; width: string}> = [];
+  const gapFrames = frames.entries();
   let currFrame = gapFrames.next();
 
   while (!currFrame.done) {
```
ryan953, line 27: This could probably become more like `for (const currFrame of gapFrames) {` and we can avoid doing the `.next()` and `.done` calls manually.

ryan953 (approved):

ryan953 (approved):

static/app/components/replays/breadcrumbs/timelineGaps.tsx
```diff
@@ -1,88 +1,86 @@
...
 
+  // create gap if we still have start (background frame) until end of replay
+  if (start !== -1) {
+    ranges.push({
+      left: toPercent(start / durationMs),
+      width: toPercent((durationMs - start) / durationMs),
+    });
+  }
+
+  // TODO: Fix tooltip position to follow mouse (it currently goes off the timeline when zoomed too much)
```
ryan953, line 56: ah, bugs :(

c298lee: yes :( we have designs for changing up the tooltip for the entire timeline and scrubber tho so hopefully when we implement that, it will fix this bug

PR #74410, opened 2024-07-17
feat(metrics): add unit normalization for span metrics

### LINKED ISSUE #74408: Inject span metric unit config in unit normalization

shellmayr:
Closes https://github.com/getsentry/sentry/issues/74408

### DIFF
--- a/src/sentry/sentry_metrics/querying/visitors/query_expression.py
+++ b/src/sentry/sentry_metrics/querying/visitors/query_expression.py
@@ -7,6 +7,7 @@ from snuba_sdk.conditions import ConditionGroup
 
 from sentry.models.environment import Environment
 from sentry.models.project import Project
+from sentry.sentry_metrics.models import SpanAttributeExtractionRuleCondition
 from sentry.sentry_metrics.querying.constants import COEFFICIENT_OPERATORS
 from sentry.sentry_metrics.querying.data.mapping.base import (
     Mapper,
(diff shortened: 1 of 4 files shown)


src/sentry/sentry_metrics/querying/visitors/query_expression.py
```diff
@@ -422,6 +424,10 @@ def _extract_unit(self, timeseries: Timeseries) -> str | None:
 
         parsed_mri = parse_mri(timeseries.metric.mri)
         if parsed_mri is not None:
+            if parsed_mri.name.startswith(SPAN_ATTRIBUTE_PREFIX):
+                rule_id = int(parsed_mri.name[len(SPAN_ATTRIBUTE_PREFIX) :])
+                return SpanAttributeExtractionRuleCondition.objects.get(id=rule_id).config.unit
```
iambriccardo, line 429: We would have to handle the lack of the condition by catching the `SpanAttributeExtractionRuleCondition.DoesNotExist` exception.

src/sentry/sentry_metrics/querying/visitors/query_expression.py
```diff
@@ -422,6 +424,10 @@ def _extract_unit(self, timeseries: Timeseries) -> str | None:
 
         parsed_mri = parse_mri(timeseries.metric.mri)
         if parsed_mri is not None:
+            if parsed_mri.name.startswith(SPAN_ATTRIBUTE_PREFIX):
+                rule_id = int(parsed_mri.name[len(SPAN_ATTRIBUTE_PREFIX) :])
```
iambriccardo, line 428: What if we extend the `ParsedMRI` to have a method which tries to get the rule id? Maybe in a subclass or just as standalone method, I feel like we might need this more than once. This method could also embed good error handling in case of failure of `int` conversion.

src/sentry/sentry_metrics/querying/visitors/query_expression.py
shellmayr: Good point, added the handling

src/sentry/sentry_metrics/querying/visitors/query_expression.py
shellmayr: Added a property to the class to do this

iambriccardo (approved):
PR LGTM, I would consider for a follow up, how to deal with a metric whose MRI can't be resolved because the rule id does not exist (right now it will silently not normalize the unit). For that though, I would use the visitors that perform query validation.

PR #74415, opened 2024-07-17
feat(crons): Backfill new broken monitor setting

davidenwang:
Backfills the new broken monitor setting according to the previous value we were using which was from "Nudges" (`approval`)


src/sentry/migrations/0742_backfill_broken_monitor_notification_setting_option.py
```diff
@@ -0,0 +1,52 @@
...
+from sentry.utils.query import RangeQuerySetWrapperWithProgressBar
+
+
+def backfill_broken_monitor_notification_setting_option(
+    apps: Apps, schema_editor: BaseDatabaseSchemaEditor
+) -> None:
+    NotificationSettingOption = apps.get_model("sentry", "NotificationSettingOption")
+
+    # Migrate all nudge settings to the new broken cron settings
+    approval_settings = NotificationSettingOption.objects.filter(type="approval")
```
markstory, line 17: There isn't an index on the `type` column, so you'll be stuck doing a table scan which could timeout. Scanning all rows and doing this filtering in python will be slower, but give you a query that will not timeout.

src/sentry/migrations/0742_backfill_broken_monitor_notification_setting_option.py
```diff
@@ -0,0 +1,52 @@
...
+    for setting in RangeQuerySetWrapperWithProgressBar(approval_settings):
+        NotificationSettingOption.objects.get_or_create(
+            user_id=setting.user_id,
+            scope_type=setting.scope_type,
+            scope_identifier=setting.scope_identifier,
+            type="brokenMonitors",
+            defaults={"value": setting.value},
+        )
```
markstory, line 25: Do we need to remove the `approval` type setting?

tests/sentry/migrations/test_0742_backfill_broken_monitor_notification_setting_option.py
```diff
@@ -0,0 +1,61 @@
...
+            value="always",
+        )
+        # Test that the migration doesn't create duplicate settings
+        NotificationSettingOption.objects.create(
+            user_id=self.user3.id,
+            scope_type="user",
+            scope_identifier=self.user.id,
+            type="brokenMonitors",
+            value="never",
+        )
```
markstory, line 38: If there is an existing row for `brokenMonitors` does the value need to be updated?

src/sentry/migrations/0742_backfill_broken_monitor_notification_setting_option.py
davidenwang: ahh makes sense tyty

src/sentry/migrations/0742_backfill_broken_monitor_notification_setting_option.py
davidenwang: nah, the approval setting still exists and will control other classes of notifications, just wanted to split out broken monitor emails into its own category with this migration

tests/sentry/migrations/test_0742_backfill_broken_monitor_notification_setting_option.py
davidenwang: its very unlikely for brokenMonitors to have a pre-existing value unless a user has manually called the API at this point, but if they did choose to do that I just opt to respect the setting they chose, rather than overriding the value from their "approval" setting

wedamija (approved):

Issue #74340, opened 2024-07-16
Cleanup manually sampled transactions


sl0thentr0py:
We need to cleanup sampling in sentry for manual transactions.

example:
https://github.com/getsentry/sentry/blob/2b4cd05450b5f5602d80f3758a5c42ee374656fa/src/sentry/ingest/consumer/processors.py#L44-L47

All such cases are overriding sdk sampling so the DSC sample rates are inaccurate.
Move the sampling logic to a rate that's passed to the sdk via `traces_sampler`.

CAUTION: Ingest consumers are very critical and overloaded so make sure SDK overhead is low when we make this change and rollout carefully with monitoring.

antonpirker:
A quick code search for the linked `SENTRY_INGEST_CONSUMER_APM_SAMPLING` above shows that is is only something other than `0` in the tests. So the sampling rate for ingest consumsers is always `0`? (or is this settingoverwritten somewhere)

I also was unable to find transactions with a name of for example `"ingest_consumer.process_event"`.

jan-auer:
The setting is set to `0.02` [in getsentry](https://github.com/getsentry/getsentry/blob/12b7a3aede21d041e612b30129a18904ea04370c/getsentry/conf/settings/defaults.py#L1146).

PR #74425, opened 2024-07-17
Move sampling decision into `traces_sampler` for backpressure monitor.

antonpirker:
Refs #74340

### DIFF
--- a/src/sentry/processing/backpressure/monitor.py
+++ b/src/sentry/processing/backpressure/monitor.py
@@ -118,6 +118,7 @@ def start_service_monitoring() -> None:
             time.sleep(options.get("backpressure.monitoring.interval"))
             continue
 
+        # TODO-anton: remove sampled here and let traces_sampler decide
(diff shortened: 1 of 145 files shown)


src/sentry/processing/backpressure/monitor.py
```diff
@@ -118,7 +118,10 @@ def start_service_monitoring() -> None:
             time.sleep(options.get("backpressure.monitoring.interval"))
             continue
 
-        with sentry_sdk.start_transaction(name="backpressure.monitoring", sampled=True):
+        with sentry_sdk.start_transaction(
+            name="backpressure.monitoring",
+            custom_sampling_context={"sample_rate": 1.0},
```
jan-auer, line 123: For now you can also leave `sampled=True` here. I'd change this only once we get rid of `sampled`.

antonpirker:
OK, then I will close this PR for now.

PR #74426, opened 2024-07-17
Move sampling decision into `traces_sampler` for occurence consumer.

antonpirker:
Refs #74340


src/sentry/issues/occurrence_consumer.py
```diff
@@ -351,7 +351,7 @@ def _process_message(
     with sentry_sdk.start_transaction(
         op="_process_message",
         name="issues.occurrence_consumer",
-        sampled=True,
+        custom_sampling_context={"sample_rate": 1.0},
```
jan-auer, line 354: Same as in https://github.com/getsentry/sentry/pull/74425

For now you can also leave `sampled=True` here. I'd change this only once we get rid of sampled.

antonpirker:
OK, then I will close this for now.

PR #74433, opened 2024-07-17
feat(query-builder): Allow field definition getter to be defined by props

### LINKED ISSUE #74309: SearchQueryBuilder: Support `fieldDefinitionGetter`

malwilley:
Closes https://github.com/getsentry/sentry/issues/74309

No behavioral changes, but this does allow modifying the field definitions by passing `fieldDefinitionGetter`, similar to how it works with SmartSearchBar currently. Certain use cases like releases and replays need this in order to modify field descriptions and other parts of the definition.

The bulk of the work here is replacing calls to the util `getFieldDefinition` with the function that gets passed in.

### DIFF
--- a/static/app/components/searchQueryBuilder/tokens/freeText.tsx
+++ b/static/app/components/searchQueryBuilder/tokens/freeText.tsx
@@ -16,6 +16,7 @@ import {
   useShiftFocusToChild,
 } from 'sentry/components/searchQueryBuilder/tokens/utils';
 import type {
+  FieldDefinitionGetter,
   FilterKeySection,
   FocusOverride,
 } from 'sentry/components/searchQueryBuilder/types';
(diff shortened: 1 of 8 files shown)


static/app/components/searchQueryBuilder/tokens/freeText.tsx
```diff
@@ -134,29 +134,7 @@ function replaceFocusedWordWithFilter(
   return value;
 }
 
-/**
- * Takes a string that contains a filter value `<key>:` and replaces with any aliases that may exist.
- *
- * Example:
- * replaceAliasedFilterKeys('foo issue: bar', {'status': 'is'}) => 'foo is: bar'
- */
-function replaceAliasedFilterKeys(value: string, aliasToKeyMap: Record<string, string>) {
```
malwilley, line 143: Instead of modifying this function to support the new getFieldDefinition, I deleted it because we are no longer using aliases.

roggenkemper (approved):

Issue #73859, opened 2024-07-04
App Modules - Consolidate Integrations


markstory:
Integration and its direct relations are control_silo scoped resources. Currently the components that comprise the control silo aspects of integrations are spread throughout the repository. Consolidating those classes and functions into `sentry.integrations` will help clarify the application's design and make running subsets of tests easier and faster.

- [ ] Move endpoints, serializers, models, and tasks to be co-located in sentry.integrations
- [ ] Communicate these changes with the broader ecosystem team
- [ ] Update getsentry references and remove import shims

We'll need to provide import shims for any symbols used in getsentry until those imports can be updated separately.

cc @Christinarlong

markstory:
Closing as we've finished off integration consolidation.

PR #74438, opened 2024-07-17
ref(control_silo): Consolidate integrations resources

Christinarlong:
Move all integrations related resources (models, endpoints, validators, bases, tests etc.) to be `sentry.integrations` and `tests.sentry.integrations`. 
Note: `tasks` not included

ref(#73859)


pyproject.toml
```diff
@@ -257,6 +245,18 @@ module = [
...
+    "sentry.integrations.api.endpoints.group_integration_details",
+    "sentry.integrations.api.endpoints.group_integrations",
+    "sentry.integrations.api.endpoints.integrations.install_request",
+    "sentry.integrations.api.endpoints.organization.organization_code_mapping_codeowners",
+    "sentry.integrations.api.endpoints.organization.organization_code_mappings",
+    "sentry.integrations.api.serializers.doc_integration",
+    "sentry.integrations.api.serializers.external_actor",
+    "sentry.integrations.api.serializers.integration",
+    "sentry.integrations.api.serializers.integration_feature",
+    "sentry.integrations.api.serializers.rest_framework.doc_integration",
```
asottile-sentry, line 259: as part of moving this can you fix the type errors?  this list isn't meant to be added to

asottile-sentry (commented):
can you add some context / rationale to this change?  this is going to conflict with a bunch of my in-flight branches and I'd like to make sure we're doing this for the right reason instead of creating a bunch of work

Christinarlong:
> can you add some context / rationale to this change? this is going to conflict with a bunch of my in-flight branches and I'd like to make sure we're doing this for the right reason instead of creating a bunch of work

For sure, this PR is a part of the `Application Modularization` effort - notion (https://www.notion.so/sentry/Consolidate-more-control-silo-code-71396b786fe944d5a4c643864d2b8e2e). 

Yeah I agree this PR is a bit large, and encompassing. An alternative could be splitting the PR to -> consolidate-integrations-...
- models
- api
- tasks
- etc.

Not sure if that's less disruptive, since it's still the same end result.

pyproject.toml
Christinarlong: I believe all these files were originally in the mypy ignore, it's just that because they were moved the routes needed to be updated.

asottile-sentry: I think one part of improving the organization would be to fix the typing while moving the files

asottile-sentry:
let's split this up then -- this is unreviewable.

pyproject.toml
markstory: Could we fix the typing after the file moves? I think improving the typing as we're shifting the files around.

markstory:
> can you add some context / rationale to this change? this is going to conflict with a bunch of my in-flight branches and I'd like to make sure we're doing this for the right reason instead of creating a bunch of work

I'd like to start encouraging teams to avoid the 'by type' directories and instead start organizing product functionality into 'domains' or 'apps' instead. This will help clarify what parts of the application are related to each other better than a directory with 400+ endpoint classes does. Organizing by domain will also help make running tests locally a bit easier as `pytest tests/sentry/integrations` or `pytests tests/sentry/users` are simple to use, while running all tests related to integrations today is complicated. I also see this a step to help break up our import tangle, as we can start removing the large barrel modules like `sentry.models`.

Christina and I will be working on the control silo related logic first, as it is operated separately and the domain boundaries are easier to define.

PR #74444, opened 2024-07-17
feat(autofix): Add status check for autofix runs and log an error if it fails

jennmueng:
Adds a celery task that will log an error to sentry if an autofix state has been processing and hasn't been updated in 5 minutes. Scheduled to run 15 minutes after the run has started.

This is in order to alert the ML team if Autofix is unresponsive again.

### DIFF
--- a/src/sentry/tasks/autofix.py
+++ b/src/sentry/tasks/autofix.py
@@ -1,0 +1,22 @@
+import logging
+from datetime import datetime, timedelta
+
+from sentry.autofix.utils import AutofixStatus, get_autofix_state
+from sentry.tasks.base import instrumented_task
+
+logger = logging.getLogger(__name__)
+
+
+@instrumented_task(
+    name="sentry.tasks.autofix.check_autofix_status",
+    max_retries=1,
+)
+def check_autofix_status(run_id: int):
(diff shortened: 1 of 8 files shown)


src/sentry/tasks/autofix.py
```diff
@@ -0,0 +1,25 @@
...
+    name="sentry.tasks.autofix.check_autofix_status",
+    max_retries=1,
+)
+def check_autofix_status(run_id: int):
+    state = get_autofix_state(run_id=run_id)
+
+    if (
+        state
+        and state.status == AutofixStatus.PROCESSING
+        and state.updated_at < datetime.now() - timedelta(minutes=5)
```
ram-senth, line 20: Nit: ideally 5 mins should be a configurable value.

ram-senth (approved):

src/sentry/tasks/autofix.py
jennmueng: Do you mean via the sentry config right?

I'll get this in for now so we monitor but can follow up

jennmueng: nvm just noticed the tests

PR #74447, opened 2024-07-17
feat(query-builder): Add support for duration filters

### LINKED ISSUE #74302: SearchQueryBuilder: Support duration filters

malwilley:
Closes https://github.com/getsentry/sentry/issues/74302

https://github.com/user-attachments/assets/2bf62070-bb0a-4d9c-af4e-bdd8e6ddcbe5

Adds support for duration filters, which don't exist in issues search but can in other contexts. You can test it on the stories page, where I've added `measurements.lcp` as an option.

### DIFF
--- a/static/app/components/searchQueryBuilder/tokens/filter/valueCombobox.tsx
+++ b/static/app/components/searchQueryBuilder/tokens/filter/valueCombobox.tsx
@@ -82,7 +82,7 @@ function isStringFilterValues(
 
 const NUMERIC_UNITS = ['k', 'm', 'b'] as const;
 const RELATIVE_DATE_UNITS = ['m', 'h', 'd', 'w'] as const;
-const DURATION_UNITS = ['ms', 's', 'm', 'h', 'd', 'w'] as const;
+const DURATION_UNIT_SUGGESTIONS = ['ms', 's', 'm', 'h'] as const;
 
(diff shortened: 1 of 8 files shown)


roggenkemper (approved):

static/app/components/searchQueryBuilder/tokens/filter/valueCombobox.tsx
```diff
@@ -93,7 +94,7 @@ const DEFAULT_NUMERIC_SUGGESTIONS: SuggestionSection[] = [
 const DEFAULT_DURATION_SUGGESTIONS: SuggestionSection[] = [
   {
     sectionText: '',
-    suggestions: [{value: '100'}, {value: '100k'}, {value: '100m'}, {value: '100b'}],
+    suggestions: DURATION_UNITS.map(unit => ({value: `100${unit}`})),
```
mrduncan, line 97: These defaults feel weird for the days/weeks - 100d/100w won't really ever find anything given our 90d retention. Aside from special casing all of them the only suggestion I have would be to maybe change to `10${unit}`

malwilley: That's true, I had similar thoughts. I doubt days or weeks will every be useful for the filters we have, but it also seems like we should show them as an option somewhere.

I think I'll change to 10 as you suggested, and perhaps remove d and w from the default suggestions

PR #74463, opened 2024-07-17
feat(profiling): Continuous profiling flamegraph compat

Zylphrex:
This tries to creates a backwards compatible endpoint that allows generating flamegraphs from transaction profiles and continuous profiles.


src/sentry/api/endpoints/organization_profiling_profiles.py
```diff
@@ -29,36 +31,90 @@ class OrganizationProfilingBaseEndpoint(OrganizationEventsV2EndpointBase):
...
+                query=serialized.get("query", ""),
+                fingerprint=serialized.get("fingerprint"),
+            )
+            profile_candidates = executor.get_profile_candidates()
 
         return proxy_profiling_service(
             method="POST",
-            path=f"/organizations/{organization.id}/projects/{project_ids[0]}/flamegraph",
-            json_data=profile_ids,
+            path=f"/organizations/{organization.id}/flamegraph",
```
Zylphrex, line 116: Added in getsentry/vroom#487

src/sentry/profiles/flamegraph.py
```diff
@@ -257,3 +263,278 @@ def get_chunks_from_spans_metadata(
...
+                }
+            )
+
+        return continuous_profile_candidates
+
+    def _query_chunks_for_profilers(self, query: Query) -> Mapping[str, Any]:
+        """This function is split out for mocking as we cannot write to the
+        profile chunks dataset in tests today"""
+
+        assert self.snuba_params.organization
```
wmak, line 527: Should we raise an error here? Otherwise an assertion error might bubble up into unexpected places

src/sentry/profiles/profile_chunks.py
```diff
@@ -57,21 +57,29 @@ def get_chunk_ids(
     return [row["chunk_id"] for row in result["data"]]
 
 
-def resolve_datetime64(value: datetime | None, precision: int = 6) -> Function | None:
+def resolve_datetime64(
+    raw_value: datetime | str | float | None, precision: int = 6
+) -> Function | None:
```
wmak, line 62: For a follow up, but let's move this somewhere that can be used by more than just profiles, Spans will be using a `datetime64(6)` soon too

wmak (approved):
Minor comments, looks good to me otherwise

src/sentry/profiles/profile_chunks.py
Zylphrex: Will follow up on this in another PR. Ideally this should live in the SDK but this will serve as a stop-gap solution.

Issue #73758, opened 2024-07-03
Redesigned event navigation


scttcper:
![image](https://github.com/getsentry/sentry/assets/1400464/18519fae-e386-4dd7-854d-b78459578239)

Worth thinking more about how this would look on smaller screens.

**Acceptance Criteria:**

- “Jump to” should smooth scroll
- We should find a place for some of the missing actions (from the design) like next/previous
- Should be not terrible on mobile

https://www.figma.com/design/p2DOzd1RwAHQbq0v6KaYmF/Q2-Issue-Details?node-id=430-6233&t=agIKtZo34vqdVzUM-4

PR #74466, opened 2024-07-17
feat(issue-details): Updated event navigation

roggenkemper:
this pr updates the event navigation as part of the issues detail streamline project. changes to the tab design will come in a different PR
Screenshot 2024-07-16 at 3 31 52 PM. 

All Events will still redirect to the All Events tab, until we update the functionality of that. 

Part of https://github.com/getsentry/sentry/issues/73758

### DIFF
--- a/static/app/components/tabs/tab.tsx
+++ b/static/app/components/tabs/tab.tsx
@@ -23,8 +23,6 @@ interface TabProps extends AriaTabProps {
    */
   overflowing: boolean;
   state: TabListState<any>;
-  hideSelection?: boolean;
-  showPressed?: boolean;
 }
 
 /**
@@ -45,14 +43,7 @@ function handleLinkClick(e: React.PointerEvent<HTMLAnchorElement>) {
  * usage in tabs.stories.js
  */
 function BaseTab(
-  {
-    item,
-    state,
-    orientation,
-    overflowing,
-    hideSelection = false,
-    showPressed = false,
-  }: TabProps,
+  {item, state, orientation, overflowing}: TabProps,
   forwardedRef: React.ForwardedRef<HTMLLIElement>
 ) {
   const ref = useObjectRef(forwardedRef);
@@ -94,13 +85,10 @@ function BaseTab(
         <StyledInteractionStateLayer
           orientation={orientation}
           higherOpacity={isSelected}
-          isPressed={showPressed ? isSelected : undefined}
         />
         <FocusLayer orientation={orientation} />
         {rendered}
-        {!hideSelection && (
-          <TabSelectionIndicator orientation={orientation} selected={isSelected} />
-        )}
+        {<TabSelectionIndicator orientation={orientation} selected={isSelected} />}
       </InnerWrap>
     </TabWrap>
   );
--- a/static/app/components/tabs/tab.tsx
+++ b/static/app/components/tabs/tab.tsx
@@ -88,7 +88,7 @@ function BaseTab(
         />
         <FocusLayer orientation={orientation} />
         {rendered}
-        {<TabSelectionIndicator orientation={orientation} selected={isSelected} />}
+        <TabSelectionIndicator orientation={orientation} selected={isSelected} />
       </InnerWrap>
     </TabWrap>
   );
--- a/static/app/components/tabs/tab.tsx
+++ b/static/app/components/tabs/tab.tsx
@@ -23,6 +23,8 @@ interface TabProps extends AriaTabProps {
    */
   overflowing: boolean;
   state: TabListState<any>;
+  hideSelection?: boolean;
+  showPressed?: boolean;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


static/app/views/issueDetails/eventNavigation.tsx
```diff
@@ -0,0 +1,214 @@
...
+  {section: '#breadcrumbs', label: t('Replay & Breadcrumbs')},
+  {section: '#tags', label: t('Tags')},
+  {section: '#context', label: t('Context')},
+  {section: '#hydration-diff', label: t('Hydration Error Diff')},
+  {section: '#user-feedback', label: t('User Feedback')},
+  {section: '#replay', label: t('Replay')},
+];
+
+function handleClick(section: string) {
+  scrollToElement(section);
```
malwilley, line 51: I need to look into why we have a library for this, you can do this natively with `scrollIntoView`: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

Should be a bit smoother than doing it with JS

static/app/views/issueDetails/eventNavigation.tsx
```diff
@@ -0,0 +1,214 @@
...
+          <EventID>
+            {tct('Event [eventId]', {eventId: getShortEventId(event.id)})}
+          </EventID>
+          <TimeSince date={event.dateCreated ?? event.dateReceived} />
+        </EventInfo>
+        <JumpTo>
+          <div>{t('Jump to:')}</div>
+          <ButtonBar>
+            {jumpToSections.map(jump => {
+              if (!document.getElementById(jump.section.replace('#', ''))) {
```
malwilley, line 140: This line seems a bit suspect. It may be possible for react to be rendering before this element has been painted to the page, which would cause this to fail. Might be better to check the event contents instead?

static/app/views/issueDetails/eventNavigation.tsx
```diff
@@ -0,0 +1,214 @@
...
+  {section: '#exception', label: t('Exception')},
+  {section: '#breadcrumbs', label: t('Replay & Breadcrumbs')},
+  {section: '#tags', label: t('Tags')},
+  {section: '#context', label: t('Context')},
+  {section: '#hydration-diff', label: t('Hydration Error Diff')},
+  {section: '#user-feedback', label: t('User Feedback')},
+  {section: '#replay', label: t('Replay')},
+];
+
+function handleClick(section: string) {
```
malwilley, line 50: nit: not very descriptive since there are multiple clickable elements

static/app/components/tabs/tab.tsx
```diff
@@ -23,6 +23,8 @@ interface TabProps extends AriaTabProps {
    */
   overflowing: boolean;
   state: TabListState<any>;
+  hideSelection?: boolean;
+  showPressed?: boolean;
```
malwilley, line 27: What does `showPressed` mean? Not sure that this prop makes sense. We might want to think of a more descriptive way to modify the styling here.

static/app/views/issueDetails/eventNavigation.spec.tsx
```diff
@@ -0,0 +1,85 @@
+import {EventFixture} from 'sentry-fixture/event';
+import {GroupFixture} from 'sentry-fixture/group';
+
+import {render, screen, userEvent} from 'sentry-test/reactTestingLibrary';
+
+import {browserHistory} from 'sentry/utils/browserHistory';
+import * as useMedia from 'sentry/utils/useMedia';
+import EventNavigation from 'sentry/views/issueDetails/eventNavigation';
+
+describe('EventNavigation', () => {
```
malwilley, line 10: Probably want some coverage for the jump to element logic?

static/app/views/issueDetails/eventNavigation.tsx
roggenkemper: yeah - let me try to come up with a better way of doing this

static/app/views/issueDetails/eventNavigation.spec.tsx
roggenkemper: will add tests for making sure that area renders correctly - hard to test this jump logic without the other components of the page but once we have integrated things more, will add tests for that

static/app/views/issueDetails/eventNavigation.tsx
```diff
@@ -0,0 +1,233 @@
...
+  OLDEST = 'oldest',
+}
+
+const EventNavLabels = {
+  [EventNavOptions.RECOMMENDED]: t('Recommended Event'),
+  [EventNavOptions.LATEST]: t('Last Event'),
+  [EventNavOptions.OLDEST]: t('First Event'),
+};
+
+const eventDataSections = [
```
malwilley, line 38: nit: it's usually good to type this at the source (e.g. `const eventDataSections: SectionDefinition[]`), makes it easier on the TS compiler and also will give better error messages if someone modifies one of the properties.

static/app/views/issueDetails/eventNavigation.tsx
```diff
@@ -0,0 +1,233 @@
...
+              size="sm"
+              icon={<IconChevron direction="right" />}
+              disabled={!hasNextEvent}
+              to={{
+                pathname: `${baseEventsPath}${event.nextEventID}/`,
+                query: {...location.query, referrer: 'next-event'},
+              }}
+            />
+          </Navigation>
+          <StyledButton
```
malwilley, line 131: Is there a reason why this isn't a `LinkButton` like the others?

roggenkemper: no good reason, had it before i switched over to using LinkButton but didn't update this one

malwilley (approved):

PR #74467, opened 2024-07-17
feat(insights): allow insight charts to be expanded

KevinL10:
Adds a full-screen button to insight charts to allow users to view in full-screen.

image
image

### DIFF
--- a/static/app/views/insights/common/components/chart.tsx
+++ b/static/app/views/insights/common/components/chart.tsx
@@ -1,5 +1,5 @@
 import type {RefObject} from 'react';
-import {useEffect, useMemo, useRef, useState} from 'react';
+import {createContext, useContext, useEffect, useMemo, useRef, useState} from 'react';
 import {useTheme} from '@emotion/react';
 import styled from '@emotion/styled';
 import type {LineSeriesOption} from 'echarts';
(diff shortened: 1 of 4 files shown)


static/app/views/insights/common/components/chart.tsx
```diff
@@ -143,6 +145,11 @@ function Chart({
   const pageFilters = usePageFilters();
   const {start, end, period, utc} = pageFilters.selection.datetime;
 
+  const chartHeightInContext = useContext(ChartHeightContext);
+  // override the height parameter if the chart height is re-specified in the context. for example, this may happen
+  // when the chart is expanded in full-screen.
+  const height = chartHeightInContext !== undefined ? chartHeightInContext : chartHeight;
```
DominikB2014, line 151:
```suggestion
  const height = chartHeightInContext ?? chartHeight;
```
You can shorten this a bit and using nullish coalescing (unless you accept `chartHeightInContext` to be null)

DominikB2014 (approved):
nice!

static/app/views/insights/common/components/chart.tsx
DominikB2014, line 151: You can probably combine all of this too
```suggestion
  const height = useContext(ChartHeightContext) ?? chartHeight;
```

PR #74473, opened 2024-07-17
ref(StructuredData): Move some helper components out of StructuredData and isolate internals

ryan953:
The big change in here is that there's a new internal helper called `<RecursiveStructuredData>` which uses a copy of `type StructuredDataProps`, right now it's the exact same as what `<StructuredData>` uses, but that'll change in the next PR.

Also i moved `<AnnotatedValue>` and `<LinkHint>` because they're really isolated and have a bunch of their own imports.


malwilley (approved):
Good change! Now that this is getting reused in more places, could definitely use some cleaning up 👍

PR #74474, opened 2024-07-17
feat(onboarding): add messaging onboarding button

ameliahsu:
added messaging integration onboarding button

Button on Alerts page
Screenshot 2024-07-17 at 3 46 56 PM

Modal (in progress, will be updated to include buttons to add integrations)
Screenshot 2024-07-17 at 3 46 45 PM

### DIFF
--- a/static/app/views/alerts/rules/issue/setupAlertIntegrationButton.tsx
+++ b/static/app/views/alerts/rules/issue/setupAlertIntegrationButton.tsx
@@ -1,9 +1,13 @@
+import styled from '@emotion/styled';
+
+import {openModal} from 'sentry/actionCreators/modal';
 import {Button} from 'sentry/components/button';
 import DeprecatedAsyncComponent from 'sentry/components/deprecatedAsyncComponent';
 import {Tooltip} from 'sentry/components/tooltip';
 import {t} from 'sentry/locale';
 import PluginIcon from 'sentry/plugins/components/pluginIcon';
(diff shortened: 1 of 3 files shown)


static/app/views/alerts/rules/issue/setupAlertIntegrationButton.tsx
```diff
@@ -54,6 +59,32 @@ export default class SetupAlertIntegrationButton extends DeprecatedAsyncComponen
       return null;
     }
 
+    if (organization.features.includes('messaging-integration-onboarding')) {
```
cathteng, line 62: should we only show this if somebody doesn't have a messaging integration installed?

iamrajjoshi: We should also check for plan right?

ameliahsu: @cathteng yup, `SetupAlertIntegrationButton` returns null if messaging integration is already installed:
https://github.com/getsentry/sentry/blob/master/static/app/views/alerts/rules/issue/setupAlertIntegrationButton.tsx#L53-L55

cathteng (commented):
can we add a quick test for this that the correct version of the button is rendered when the feature flag is active? in `static/app/views/alerts/rules/issue/setupAlertIntegrationButton.spec.tsx`

cathteng (approved):

Issue #74451, opened 2024-07-17
[DevToolbar] Move dev toolbar code from getsentry to sentry


ryan953:
In order to get more feedback from engineers, and better support the 'at dev time' use case, we should move the code from getsentry to sentry so that all engineers at sentry will be exposed to the toolbar and can give feedback. Feature flag rules should not be affected.

PR #74478, opened 2024-07-17
ref(devtoolbar): move code from getsentry

michellewzhang:
- this PR copies everything from the devtoolbar POC (https://github.com/getsentry/getsentry/pull/14420) but moves all files from `getsentry` to `sentry`. no files were changed, except for a few imports moved from `getsentry` to `sentry`, and a few rewordings in the readme. 
- this change means that sentry employees on the sentry org, running `dev-ui` from `sentry`, will be able to see the devtoolbar. previously, the devtoolbar was only visible when running `dev-ui` from `getsentry`. note that sentry employees on any other org will see the employee feedback button, which was the previous behavior.
- will follow up with a PR that deletes the duplicated code from `getsentry`. can also follow up with PRs to clean up this code/make any adjustments as needed.
- relates to https://github.com/getsentry/sentry/issues/74451

running `dev-ui` from `getsentry` and `sentry` on sentry org gives the devtoolbar:
SCR-20240717-obbs

on any other org gives the employee feedback button:
SCR-20240717-obsf

### DIFF
--- a/static/app/components/devtoolbar/README.md
+++ b/static/app/components/devtoolbar/README.md
@@ -2,10 +2,10 @@
 
 This folder contains a PoC for what a Dev Toolbar product would look like and what features it could have.
 
-This is not production ready. Only sentry employees should be able to see this code, while they're on sentry.io. In order for this to be production-ready for customers to install on their own website it would need to be bundled similar to the SDK or Spotlight. There are many steps remaining to get there.
+This is not production ready. Only Sentry employees should be able to see this code, while they're on sentry.io or running a development environment. In order for this to be production-ready for customers to install on their own website, it would need to be bundled similar to the SDK or Spotlight. There are many steps remaining to get there.
 
-Therefore, this is built with maximum portability in mind. It is a goal for this to have as few dependencies on sentry & getsentry as possible to make the majority of the code easy to port into a different repo when/if the PoC is successful. However, some code will not be portable, which is why we're leveraging getsentry to host the PoC. For example:
+Therefore, this is built with maximum portability in mind. It is a goal for this to have as few dependencies on sentry/getsentry as possible to make the majority of the code easy to port into a different repo when/if the PoC is successful. However, some code will not be portable, which is why we're leveraging sentry to host the PoC. For example:
 
-- The API calls themselves are not portable. We're using sentry & getsentry to explicitly avoid solving the API-Auth problem right now.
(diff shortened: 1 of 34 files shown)


billyvg (commented):
How does this affect self hosted?

michellewzhang:
> How does this affect self hosted?

@bruno-garcia @billyvg we have the `isEmployee` check and feature flag check to prevent anyone but Sentry employees from seeing this currently. the goal of this PR is to make the devtoolbar more available to employees, so that we can get more initial feedback for our POC. previously, it was only accessible on prod or via `dev-ui` from `getsentry`, which we saw from slack isn't a popular way to develop. the primary reason we're moving to `sentry` is to get more exposure/feedback since this is still in the POC phase!

the data we've hardcoded is not sensitive if that's a concern; we have the project slug / project id / org slug hardcoded in several places in our public repo, so it seems to be OK.

are there any other concerns about this change?

cc @ryan953

ryan953 (approved):

ryan953:
I believe that this will not affect self-hosted. @michellewzhang summarized the reason why we'd make this change, I'll dig in to look at the implications on self-hosted, and what they could do with this code. The relevant files to look at is static/app/utils/useDevToolbar.tsx and static/app/views/organizationLayout/index.tsx.

By default the `useIsSentryEmployee()` is a basic, but robust and will only allow folks with @sentry.io email addresses in. I did see that there will be iteration in the future to use the `user.isEmployee` flag, once that becomes reliable. 
https://github.com/getsentry/sentry/blob/516dbb40a8b38e19ba03688d00bb53bf4fed845d/static/app/utils/useIsSentryEmployee.ts#L11-L16

We also have the feature flag for it... which self-hosted could enable....

if they defeat the email address thing, and enable the feature flag they'll get 401 responses from the api when they open the current panels (in the future they might get those by default if the buttons start making api requests). This is because the browser session will not be logged into an account with access to the `sentry` org. 

They _could_ then change the hard-coded settings in `useDevToolbar()` to point to their own org, and the api will return valid responses from their own org. But it's really limited utility, because at this point they'd be doing queries for "issues matching the url: /issues", but the data inside their sentry instance would be from urls like `/checkout` or whatever their urls are. 

Shipping the devtoolbar like this is really sentry-in-sentry specific, and i don't think it poses a problem. Being open source people can checkout what were working on, which'll happen anyway too!

Issue #74456, opened 2024-07-17
[Hydration Errors] Add amplitude analytics to hydration error tools


ryan953:
I want to be able to have charts like:
- [x] ~~when the issue stream items are fetched, how many hydration errors does the list contain?~~ Can redash query for num issues per proj, org, etc. No need to track how many per issues stream.
- [x] when the issue details page is rendered, is it for a hydration error?
- [x] Log an event for each interaction (debounced) with the slider-diff-tool
- [x] Log an event each time the hydration diff modal is opened, include what page did the button click happen on
- [x] Log an event each time the hydration diff modal tabs are changed, include which tab was picked.

Also:
- [ ] Clone the rage click issue dashboard in Redash: https://redash.getsentry.net/dashboards/227-rage-click-issues

aliu39:
Looks like there's already an analytics event for #4: https://github.com/getsentry/sentry/pull/61888. Might not track the page though.

aliu39:
> Log an event for each interaction (debounced) with the slider-diff-tool
An approach would be to track _only the first time_ the user clicks or drags the slider, since they last visited the page

PR #74479, opened 2024-07-17
analytics(replay): click tracking analytics for hydration diff modal

aliu39:
Ref https://github.com/getsentry/sentry/issues/74456

1. Renames the "replay.details-hydration-modal-opened" event to "replay.hydration-modal.opened", and adds the `surface` param, which can be either "issue-details" or "replay-breadcrumbs"
2. Tracks analytics on changing tabs in the modal. Event name = "replay.hydration-modal.tab-change". Includes the tab type (html/slider/visual)

### DIFF
--- a/static/app/components/events/eventHydrationDiff/replayDiffContent.tsx
+++ b/static/app/components/events/eventHydrationDiff/replayDiffContent.tsx
@@ -44,6 +44,7 @@ export default function ReplayDiffContent({event, group, orgSlug, replaySlug}: P
           leftOffsetMs={leftOffsetMs}
           replay={replay}
           rightOffsetMs={rightOffsetMs}
+          surface={'issue-details'} // TODO: refactor once this component is used in more surfaces
           size="xs"
         >
           {t('Open Diff Viewer')}
--- a/static/app/components/events/eventHydrationDiff/replayDiffContent.tsx
+++ b/static/app/components/events/eventHydrationDiff/replayDiffContent.tsx
@@ -44,7 +44,7 @@ export default function ReplayDiffContent({event, group, orgSlug, replaySlug}: P
           leftOffsetMs={leftOffsetMs}
           replay={replay}
           rightOffsetMs={rightOffsetMs}
-          surface={'issue-details'} // TODO: refactor once this component is used in more surfaces
+          surface="issue-details" // TODO: refactor once this component is used in more surfaces
           size="xs"
         >
           {t('Open Diff Viewer')}
--- a/static/app/components/replays/breadcrumbs/openReplayComparisonButton.tsx
+++ b/static/app/components/replays/breadcrumbs/openReplayComparisonButton.tsx
@@ -34,8 +34,8 @@ export function OpenReplayComparisonButton({
     <Button
       role="button"
       size={size}
-      analyticsEventKey="replay.details-hydration-modal-opened"
-      analyticsEventName="Replay Details Hydration Modal Opened"
+      analyticsEventKey="replay.hydration-modal.opened"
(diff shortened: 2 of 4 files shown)


static/app/components/replays/breadcrumbs/openReplayComparisonButton.tsx
```diff
@@ -35,6 +36,7 @@ export function OpenReplayComparisonButton({
       size={size}
       analyticsEventKey="replay.details-hydration-modal-opened"
       analyticsEventName="Replay Details Hydration Modal Opened"
+      analyticsParams={{url: getLocationHref()}}
```
ryan953, line 39: This might be hard to group-by later on. something like a transaction name would be better, or when the button is created we could pass in a simple string that names the page. 

like `<OpenReplayComparisonButton surface="replay-breadcrumbs" ... />` or `<OpenReplayComparisonButton surface="issue-details" ... />`

aliu39: Got it, I think there are a lot of components we have to pass this down though.. maybe we can find a cleaner way to get the prefix/suffix of a URL?

aliu39: pathname is everything after the `sentry.io`

ryan953: we usually don't want to log the whole url. it's not important to know that the url was `/issues/3758928546/events/9e0d6bc120794875a08ec471ac8840c0` better to know it was `issues/<issue_id>/events/<event_id>`

aliu39:
The other analytics in the ticket will take some more work, since they should be tracked on the _first_ render since the last page load. I'll make a separate PR for those.

aliu39:
Oh, also note:
- tab change won't include the first tab when the modal's first opened (slider). I'm assuming this is ok. If we query by slider, it'll represent all the times the user _switched back_ to slider.
- org and project are excluded from both events. Lmk if this would be useful to track

static/app/components/events/eventHydrationDiff/replayDiffContent.tsx
```diff
@@ -44,6 +44,7 @@ export default function ReplayDiffContent({event, group, orgSlug, replaySlug}: P
           leftOffsetMs={leftOffsetMs}
           replay={replay}
           rightOffsetMs={rightOffsetMs}
+          surface={'issue-details'} // TODO: refactor once this component is used in more surfaces
```
ryan953, line 47:
```suggestion
          surface="issue-details" // TODO: refactor once this component is used in more surfaces
```

static/app/components/replays/diff/replayDiffChooser.tsx
```diff
@@ -29,8 +30,12 @@ export default function ReplayDiffChooser({
 }: Props) {
   const isSameTimestamp = leftOffsetMs === rightOffsetMs;
 
+  const onTabChange = (tabKey: DiffType) => {
+    trackAnalytics('replay.hydration-modal.tab-change', {tabKey, organization: null});
+  };
```
ryan953, line 35: it's important to have the organization for slicing and dicing the data later, and it's easy to get a reference
`import useOrganization from 'sentry/utils/useOrganization';`
then:

```suggestion
  const organization = useOrganization();
  const onTabChange = (tabKey: DiffType) => {
    trackAnalytics('replay.hydration-modal.tab-change', {tabKey, organization});
  };
```

ryan953 (approved):

PR #74484, opened 2024-07-17
ref(StructuredEventData): Rewrite expand/collapse state into a provider

ryan953:
Pushed the expand/collapse state into a provider, which accepts a callback so external systems can save the state and pass it back in on mount so that the component will render fresh with the state from an old component.

There's a `useRef` to track the overall state of things in the provider, and each `<CollapsibleValue>` has it's own `setState` so when something is toggled we don't endup having react look at the whole tree.

### DIFF
--- a/static/app/components/structuredEventData/collapsibleValue.tsx
+++ b/static/app/components/structuredEventData/collapsibleValue.tsx
@@ -24,7 +24,6 @@ export function CollapsibleValue({
 }: Props) {
   const {collapse, expand, isExpanded: isInitiallyExpanded} = useExpandedState({path});
   const [isExpanded, setIsExpanded] = useState(isInitiallyExpanded);
-  // const isExpanded = expandedPaths.includes(path);
 
   const numChildren = Children.count(children);
 
--- a/static/app/components/structuredEventData/collapsibleValue.tsx
+++ b/static/app/components/structuredEventData/collapsibleValue.tsx
@@ -1,4 +1,4 @@
-import {Children, type ReactNode} from 'react';
+import {Children, type ReactNode, useState} from 'react';
 import styled from '@emotion/styled';
 
 import {Button} from 'sentry/components/button';
@@ -22,8 +22,9 @@ export function CollapsibleValue({
   path,
   prefix = null,
 }: Props) {
-  const {collapse, expand, expandedPaths} = useExpandedState();
-  const isExpanded = expandedPaths.includes(path);
+  const {collapse, expand, isExpanded: isInitiallyExpanded} = useExpandedState({path});
+  const [isExpanded, setIsExpanded] = useState(isInitiallyExpanded);
+  // const isExpanded = expandedPaths.includes(path);
 
   const numChildren = Children.count(children);
 
@@ -40,7 +41,15 @@ export function CollapsibleValue({
         <ToggleButton
           size="zero"
           aria-label={isExpanded ? t('Collapse') : t('Expand')}
-          onClick={() => (isExpanded ? collapse(path) : expand(path))}
+          onClick={() => {
+            if (isExpanded) {
+              collapse();
+              setIsExpanded(false);
+            } else {
+              expand();
+              setIsExpanded(true);
+            }
+          }}
           icon={
             <IconChevron direction={isExpanded ? 'down' : 'right'} legacySize="10px" />
           }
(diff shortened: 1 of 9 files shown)


static/app/components/structuredEventData/index.tsx
```diff
@@ -29,73 +75,85 @@ interface StructuredDataProps {
...
+  withOnlyFormattedText = false,
 }: StructuredDataProps) {
+  const getInitialExpandedPaths = useCallback(() => {
+    return (
+      initialExpandedPaths ??
+      (forceDefaultExpand === false ||
+      (forceDefaultExpand === undefined && !maxDefaultDepth)
+        ? []
+        : getDefaultExpanded(Math.max(1, maxDefaultDepth), value))
+    );
```
ryan953, line 95: I didn't write unit tests for this, but the "Auto-Expanded items" story renders the same as before. See https://github.com/getsentry/sentry/pull/74481 as a reference.

static/app/components/structuredEventData/collapsibleValue.tsx
```diff
@@ -1,44 +1,35 @@
...
+  path,
   prefix = null,
-  depth,
-  maxDefaultDepth,
-  forceDefaultExpand,
-}: CollapsibleValueProps) {
+}: Props) {
+  const {collapse, expand, isExpanded: isInitiallyExpanded} = useExpandedState({path});
+  const [isExpanded, setIsExpanded] = useState(isInitiallyExpanded);
+  // const isExpanded = expandedPaths.includes(path);
```
malwilley, line 27: Left a little something here

static/app/utils/replays/countDomNodes.tsx
```diff
@@ -26,6 +27,7 @@ export default function countDomNodes({
   const frameStep = Math.max(Math.round(length * 0.007), 1);
 
   let prevIds: number[] = [];
+  useOrganization();
```
malwilley, line 30: And this seems unrelated

static/app/components/structuredEventData/useExpandedState.tsx
```diff
@@ -0,0 +1,66 @@
+import type {ReactNode} from 'react';
+import {createContext, useCallback, useContext, useMemo, useRef} from 'react';
+
+import {uniq} from 'sentry/utils/array/uniq';
+
+type State = 'expanded' | 'collapsed';
+
+const context = createContext<{
```
malwilley, line 8: nit: casing would usually be `Context`

static/app/components/structuredEventData/index.tsx
```diff
@@ -15,12 +18,55 @@ export type StructedEventDataConfig = {
   renderString?: (value: string) => string;
 };
 
-interface StructuredDataProps {
-  maxDefaultDepth: number;
-  withAnnotatedText: boolean;
+interface BaseProps {
```
malwilley, line 21: Thanks for the descriptions! 👏

static/app/components/structuredEventData/useExpandedState.tsx
```diff
@@ -0,0 +1,66 @@
...
+  initialExpandedPaths: () => string[];
+  onToggleExpand?: (expandedPaths: string[], path: string, state: State) => void;
+}
+
+export function ExpandedStateContextProvider({
+  children,
+  initialExpandedPaths,
+  onToggleExpand,
+}: Props) {
+  const expandedRef = useRef<string[]>(initialExpandedPaths());
```
malwilley, line 29: Is there a reason we need to keep this `expandedRef` up to date when expanding/collapsing? It seems a bit confusing to me if we have two sources of truth in the component. Could we just use it for the initial state and leave it be after that?

static/app/components/structuredEventData/index.tsx
malwilley: Should we write a test for this? Seems like a lot of extra logic that could regress

ryan953: tested!

static/app/components/structuredEventData/useExpandedState.tsx
ryan953: yes! it makes the callback api so much easier.

The main use-case for the callback is so we can keep track of everything that's expanded, and then re-render with that same state in the future. This is especially needed in breadcrumbs inside the replay details page where we use a virtual list, as the user scrolls list items are being unmounted and re-mounted all the time.

Without the callback a callsite would need to do the `.concat` and `.filter` calls (just a few lines below here). But with the ref the callback takes that on instead. Then we can shove the whole state into another ref that contains a list of states for our virtual list to use.

malwilley: Ahh okay, I see the utility now 👍

malwilley (approved):

PR #74492, opened 2024-07-18
feat(lookup-field) Add custom lookup field `element_contains` for ArrayField

vgrozdanic:
Current Django ORM is limited if there is a need for a check if the element of the array field contains some value.
This custom lookup fields works in a similar way to Django ORM `contains` lookup field, but it is applied to the every element of the array field.

## Use case
This is semi-pseudo code:
```
class CustomModel(models.Model):
    field = ArrayField()

CustomModel.objects.filter(field__element_contains="<value>")
```

### DIFF
--- a/src/sentry/db/models/fields/array.py
+++ b/src/sentry/db/models/fields/array.py
@@ -73,5 +73,4 @@ class ArrayField(models.Field):
         return [self.of.to_python(x) for x in value]
 
 
-ArrayField.register_lookup(ArrayElementContainsLookup)
 DjangoArrayField.register_lookup(ArrayElementContainsLookup)
--- a/src/sentry/db/models/fields/array.py
+++ b/src/sentry/db/models/fields/array.py
@@ -6,7 +6,7 @@ from django.contrib.postgres.fields import ArrayField as DjangoArrayField
 from django.db import models
 
 from sentry.db.models.utils import Creator
-from sentry.db.postgres.lookups import ArrayElementContainsLookup
+from sentry.db.postgres.lookups.array_element_contains import ArrayElementContainsLookup
 from sentry.utils import json
 
 
--- a/src/sentry/db/models/fields/array.py
+++ b/src/sentry/db/models/fields/array.py
@@ -2,9 +2,11 @@ from __future__ import annotations
 
 import ast
 
+from django.contrib.postgres.fields import ArrayField as DjangoArrayField
 from django.db import models
 
 from sentry.db.models.utils import Creator
+from sentry.db.postgres.lookups import ArrayElementContainsLookup
 from sentry.utils import json
 
 
@@ -69,3 +71,7 @@ class ArrayField(models.Field):
                     assert "\\" not in value, "Unexpected ArrayField format"
                     value = value[1:-1].split(",")
         return [self.of.to_python(x) for x in value]
+
+
+ArrayField.register_lookup(ArrayElementContainsLookup)
+DjangoArrayField.register_lookup(ArrayElementContainsLookup)
--- a/src/sentry/db/postgres/__init__.py
+++ b/src/sentry/db/postgres/__init__.py
@@ -1,1 +1,0 @@
-from .lookups import *  # NOQA
--- a/src/sentry/db/postgres/__init__.py
+++ b/src/sentry/db/postgres/__init__.py
@@ -1,0 +1,1 @@
+from .lookups import *  # NOQA
--- a/src/sentry/db/postgres/lookups/__init__.py
+++ b/src/sentry/db/postgres/lookups/__init__.py
@@ -1,3 +1,3 @@
-from .array_element_contains import ArrayElementContains  # NOQA
+from .array_element_contains import ArrayElementContainsLookup
 
-__all__ = ["ArrayElementContains"]
+__all__ = ["ArrayElementContainsLookup"]
--- a/src/sentry/db/postgres/lookups/__init__.py
+++ b/src/sentry/db/postgres/lookups/__init__.py
@@ -1,1 +1,3 @@
 from .array_element_contains import ArrayElementContains  # NOQA
+
+__all__ = ["ArrayElementContains"]
--- a/src/sentry/db/postgres/lookups/__init__.py
+++ b/src/sentry/db/postgres/lookups/__init__.py
@@ -1,0 +1,1 @@
+from .array_element_contains import ArrayElementContains  # NOQA
--- a/src/sentry/db/postgres/lookups/__init__.py
+++ b/src/sentry/db/postgres/lookups/__init__.py
@@ -1,3 +1,0 @@
-from .array_element_contains import ArrayElementContainsLookup
-
-__all__ = ["ArrayElementContainsLookup"]
--- a/src/sentry/db/postgres/lookups/array_element_contains.py
+++ b/src/sentry/db/postgres/lookups/array_element_contains.py
@@ -2,7 +2,7 @@ from django.db.backends.base.base import BaseDatabaseWrapper
 from django.db.models import Lookup
 from django.db.models.sql.compiler import SQLCompiler
 
-__all__ = ["ArrayElementContainsLookup"]
+__all__ = ("ArrayElementContainsLookup",)
 
 
 class ArrayElementContainsLookup(Lookup):
--- a/src/sentry/db/postgres/lookups/array_element_contains.py
+++ b/src/sentry/db/postgres/lookups/array_element_contains.py
@@ -1,15 +1,17 @@
-from typing import Any
-
 from django.contrib.postgres.fields import ArrayField
+from django.db.backends.postgresql.base import DatabaseWrapper
 from django.db.models import Lookup
+from django.db.models.sql.compiler import SQLCompiler
 
-__all__ = ["ArrayElementContains"]
+__all__ = ["ArrayElementContainsLookup"]
 
 
-class ArrayElementContains(Lookup):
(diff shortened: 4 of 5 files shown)


src/sentry/db/postgres/__init__.py
```diff
@@ -0,0 +1 @@
+from .lookups import *  # NOQA
```
evanpurkhiser, line 1: @asottile-sentry are we moving away from re-exports like this?

vgrozdanic: I had no better idea where to put this so that is picked up during start up, and that it gets registered on an `ArrayField`

markstory: You could put this lookup with our ArrayField in `sentry.db.models.fields.array` :shrug:

vgrozdanic: @markstory maybe that would also make sense, an to register it also to our ArrayField, but I need to use it on Django's `ArrayField` for Postgres. More specifically on `DashboardWidgetQuery` model (`aggregates` field) from `src/sentry/models/dashboard_widget.py`.

Would it be weird to register it also to Django's `ArrayField` in the same file/place where we define our own `ArrayField`?

I have no strong opinions here, and I am open to any proposal how to make it better and more in line with our codebase

src/sentry/db/postgres/lookups/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .array_element_contains import ArrayElementContains  # NOQA
```
asottile-sentry, line 1: noqa should not be needed here because it is listed in `__all__`

src/sentry/db/postgres/lookups/array_element_contains.py
```diff
@@ -0,0 +1,32 @@
...
+from django.contrib.postgres.fields import ArrayField
+from django.db.models import Lookup
+
+__all__ = ["ArrayElementContains"]
+
+
+class ArrayElementContains(Lookup):
+    lookup_name = "element_contains"
+
+    def as_sql(self, compiler: Any, connection: Any) -> Any:
```
asottile-sentry, line 12: please fill out the actual types here -- you can find the base class in `.venv/lib/python3.XX/site-packages/django-stubs/...`

typing things as `(Any) -> Any` makes it much harder to fix later since they look "typed" but really aren't

leaving them untyped results in the same typing

tests/sentry/db/postgres/lookups/test_array_element_contains.py
```diff
@@ -0,0 +1,30 @@
+from unittest.mock import Mock, patch
+
+from sentry.db.postgres.lookups import ArrayElementContains
+from sentry.testutils.cases import TestCase
+
+
+class TestArrayElementContains(TestCase):
+    def setUp(self) -> None:
+        self.compiler = Mock()
+        self.connection = Mock()
```
asottile-sentry, line 10: no reason to assign these to `self` -- they are only used in one place

and actually I don't think there's a reason to write a deprecated TestCase class either -- write a pytest function test please!

src/sentry/db/postgres/lookups/array_element_contains.py
```diff
@@ -0,0 +1,32 @@
...
+        rhs, rhs_params = self.process_rhs(compiler, connection)
+        params = lhs_params + rhs_params
+
+        return (
+            """EXISTS (
+            SELECT * FROM UNNEST({}) AS elem
+            WHERE elem LIKE '%%' || {} || '%%'
+        )""".format(
+                lhs, rhs
+            ),
```
asottile-sentry, line 27: should be an f string

though I'm a little nervous building manual queries using string formatting -- are we guarded against xss here?

src/sentry/db/postgres/__init__.py
asottile-sentry: yeah I just killed the star import here, would be a shame to bring it back.  I think there are better places for database initialization throughout `sentry.db` as mark points out

asottile-sentry (changes requested):

src/sentry/db/postgres/lookups/array_element_contains.py
vgrozdanic: We are guarded, since we are not inserting the direct value here:
- `lhs` will be column_name <- controlled by us, injection can't happen
- `rhs` will be `'%s'` (always)

`params` that are returned from this `as_sql` function will be the values which are going to be inserted in a query when `cursor.execute` is called, but when this call happens, all parameters inserted in the query are properly escaped ([django code](https://github.com/django/django/blob/cf03aa4e94625971852a09e869f7ee7c328b573f/django/db/models/sql/compiler.py#L1580)).

src/sentry/db/postgres/__init__.py
vgrozdanic: I have moved it to `src/sentry/db/models/fields/array.py` as Mark suggested and registered this custom lookup on both our custom `ArrayField` and Django's `ArrayField`

src/sentry/db/postgres/lookups/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .array_element_contains import ArrayElementContainsLookup
+
+__all__ = ["ArrayElementContainsLookup"]
```
asottile-sentry, line 3: please remove the contents of `__init__.py`

src/sentry/db/postgres/lookups/array_element_contains.py
```diff
@@ -0,0 +1,28 @@
...
+        rhs, rhs_params = self.process_rhs(compiler, connection)
+        params = lhs_params + rhs_params
+
+        return (
+            f"""EXISTS (
+            SELECT * FROM UNNEST({lhs}) AS elem
+            WHERE elem LIKE '%%' || {rhs} || '%%'
+        )""",
+            params,
+        )
```
asottile-sentry, line 28: I know this is black's doing -- but can you format this so it doesn't look so bad?

tests/sentry/db/postgres/lookups/test_array_element_contains.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def test_as_sql_basic_usage():
+    with (
+        patch(
+            "sentry.db.postgres.lookups.array_element_contains.ArrayElementContainsLookup.process_lhs"
+        ) as mock_process_lhs,
+        patch(
+            "sentry.db.postgres.lookups.array_element_contains.ArrayElementContainsLookup.process_rhs"
+        ) as mock_process_rhs,
```
asottile-sentry, line 13: prefer `mock.patch.object` to `patch` -- the latter has import side-effects and is more likely to lead to leaked patches

asottile-sentry: also it'll be way fewer characters here :)

tests/sentry/db/postgres/lookups/test_array_element_contains.py
```diff
@@ -0,0 +1,30 @@
...
+        sql, params = lookup.as_sql(Mock(), Mock())
+
+        assert (
+            sql
+            == """EXISTS (
+            SELECT * FROM UNNEST(column_name) AS elem
+            WHERE elem LIKE '%%' || %s || '%%'
+        )"""
+        )
+        assert params == ["value"]
```
asottile-sentry, line 30: this test doesn't actually test anything -- it's a copy paste of the implementation with mocks inserted.  the implementation could change in a compatible way and the test would break even if the functionality still worked

this should test the _behaviour_ and not the _implementation_

good rule of thumb: if you find yourself reaching for mock, mock, mock at the beginning and asserting exactly the implementation return value at the end: it's probably not a good test!

src/sentry/db/postgres/lookups/array_element_contains.py
asottile-sentry: maybe:

```python
        clause = f"""\
EXISTS (
    SELECT * FROM UNNEST({lhs}) AS elem
    WHERE elem LIKE '%%' || {rhs} || '%%'
)
"""
        return clause, params
```

src/sentry/db/postgres/lookups/array_element_contains.py
```diff
@@ -0,0 +1,28 @@
+from django.db.backends.base.base import BaseDatabaseWrapper
+from django.db.models import Lookup
+from django.db.models.sql.compiler import SQLCompiler
+
+__all__ = ["ArrayElementContainsLookup"]
```
asottile-sentry, line 5:
```suggestion
__all__ = ("ArrayElementContainsLookup",)
```

src/sentry/db/models/fields/array.py
```diff
@@ -69,3 +71,6 @@ def to_python(self, value):
                     assert "\\" not in value, "Unexpected ArrayField format"
                     value = value[1:-1].split(",")
         return [self.of.to_python(x) for x in value]
+
+
+DjangoArrayField.register_lookup(ArrayElementContainsLookup)
```
vgrozdanic, line 76: I haven't found better/more fitting place to register this lookup.

Newly added lookup isn't registered to our custom `ArrayField` because our implementation of the field (`get_prep_value` [method](https://github.com/getsentry/sentry/blob/67db76651c2ff49d07acb309022bf2554e3d8b09/src/sentry/db/models/fields/array.py#L44-L54)) would require changes in order to work with this new lookup. I am not even sure if that is needed, but if it will be, that can be done as a follow up PR

asottile-sentry (approved):

PR #74500, opened 2024-07-18
chore(insights): Remove Webvitals static weight feature check

edwardgou-sentry:
Removes unneeded feature check due to GA


gggritso (approved):
🧹

PR #74504, opened 2024-07-18
chore(insights): removes unused performance score flag checks 

edwardgou-sentry:
Removes `organizations:insights-browser-webvitals-optional-components` and `organizations:insights-default-performance-score-profiles` since these flags have been GA'd and no longer needed.

### DIFF
--- a/src/sentry/relay/config/__init__.py
+++ b/src/sentry/relay/config/__init__.py
@@ -403,14 +403,7 @@ def _should_extract_abnormal_mechanism(project: Project) -> bool:
     )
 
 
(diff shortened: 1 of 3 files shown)


Dav1dde (approved):
You didn't delete the options, but I assume that was intentional?

edwardgou-sentry:
> You didn't delete the options, but I assume that was intentional?

Yeah I have a separate [pr](https://github.com/getsentry/sentry/pull/74511) to delete the options here plus an extra one
Thanks for the review!

PR #74506, opened 2024-07-18
feat(issue-details): Add new search component to all events tab

malwilley:
Under the `issue-stream-search-query-builder` flag, replaces the old SmartSearchBar component with the new one.

To load the available tags, it uses the same group tags endpoint that loads the tags summary on issue details. This is an improvement over what existed prior, since it was loading all available organization tags event if they did not exist in the issue you're looking at.

CleanShot 2024-07-18 at 09 23 45@2x

### DIFF
--- a/static/app/views/issueList/searchBar.tsx
+++ b/static/app/views/issueList/searchBar.tsx
@@ -19,16 +19,12 @@ import {
   type TagCollection,
 } from 'sentry/types';
 import {getUtcDateString} from 'sentry/utils/dates';
-import {
-  DEVICE_CLASS_TAG_VALUES,
-  FieldKind,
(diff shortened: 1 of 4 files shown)


static/app/views/issueList/searchBar.tsx
```diff
@@ -145,23 +141,8 @@ function IssueListSearchBar({organization, tags, ...props}: Props) {
     ]
   );
 
-  const getTagValues = useCallback(
```
malwilley, line 148: Moved this function into the util file below so it could be reused in the all events component.

roggenkemper (approved):
good change - was looking at all events page yesterday and forgot how good we have it with the new search experience

PR #74512, opened 2024-07-18
fix(hybrid-cloud): Fix namespace collisions with pydantic V2 upgrade

GabeVillalobos:
Pydantic V2 has introduced the concept of [protected namespaces](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.protected_namespaces), which prevents us from using `model_` in pydantic models and RPC type hints. This PR renames the fields colliding with this new namespace.

Note: These features _are_ breaking for imports/exports, but the volume of these RPCs is low enough that this shouldn't be a concern.

### DIFF
--- a/src/sentry/backup/services/import_export/impl.py
+++ b/src/sentry/backup/services/import_export/impl.py
@@ -158,7 +158,7 @@ class UniversalImportExportService(ImportExportService):
         in_pk_map = pk_map.from_rpc()
         filters: list[Filter] = []
         for fb in filter_by:
-            if NormalizedModelName(fb.for_model) == batch_model_name:
+            if NormalizedModelName(fb.on_model) == batch_model_name:
                 filters.append(fb.from_rpc())
 
         import_chunk_type = (
@@ -479,7 +479,7 @@ class UniversalImportExportService(ImportExportService):
             out_pk_map = PrimaryKeyMap()
             filters: list[Filter] = []
             for fb in filter_by:
-                if NormalizedModelName(fb.for_model) == batch_model_name:
+                if NormalizedModelName(fb.on_model) == batch_model_name:
                     filters.append(fb.from_rpc())
 
             def filter_objects(queryset_iterator):
--- a/src/sentry/backup/services/import_export/impl.py
+++ b/src/sentry/backup/services/import_export/impl.py
@@ -158,7 +158,7 @@ class UniversalImportExportService(ImportExportService):
         in_pk_map = pk_map.from_rpc()
         filters: list[Filter] = []
         for fb in filter_by:
-            if NormalizedModelName(fb.import_model_name) == batch_model_name:
+            if NormalizedModelName(fb.for_model) == batch_model_name:
                 filters.append(fb.from_rpc())
 
         import_chunk_type = (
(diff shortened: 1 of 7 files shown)


src/sentry/backup/services/import_export/model.py
```diff
@@ -27,7 +27,7 @@ class RpcFilter(RpcModel):
     Shadows `sentry.backup.helpers.Filter` for the purpose of passing it over an RPC boundary.
     """
 
-    model_name: str
+    import_model_name: str
```
azaslavsky, line 30: Let's change this to `for_model`. `Filter`/`RpcFilter` is used both when importing and exporting, so calling this `import_...` will be confusing in the latter case. I think omitting the `name` suffix is okay in this case, since said in English it would be something like "this is a `Filter` *on model* X`"

src/sentry/backup/services/import_export/impl.py
```diff
@@ -404,29 +405,29 @@ def import_by_model(
...
-                on=InstanceID(model_name),
+                on=InstanceID(import_model_name),
                 reason=f"Unknown internal error occurred: {traceback.format_exc()}",
             )
 
     def export_by_model(
         self,
         *,
-        model_name: str = "",
+        import_model_name: str = "",
```
azaslavsky, line 430:
```suggestion
        export_model_name: str = "",
```

src/sentry/backup/services/import_export/service.py
```diff
@@ -87,7 +87,7 @@ def get_importer_for_model(
     def export_by_model(
         self,
         *,
-        model_name: str = "",
+        import_model_name: str = "",
```
azaslavsky, line 90: Ditto with above.

```suggestion
        export_model_name: str = "",
```

markstory (approved):
With relocations being low volume, I think this will be safe. Temporarily disabling relocations while the deploy containing these changes goes out like Alex suggested is a good way to avoid cross deploy issues as well.

src/sentry/backup/services/import_export/impl.py
```diff
@@ -106,59 +106,59 @@ class UniversalImportExportService(ImportExportService):
     def import_by_model(
         self,
         *,
-        model_name: str,
+        import_model_name: str = "",
         scope: RpcImportScope | None = None,
-        flags: RpcImportFlags,
+        flags: RpcImportFlags = DEFAULT_IMPORT_FLAGS,
```
azaslavsky, line 111: Ty for adding this (and the `= ""` below)!

src/sentry/backup/services/import_export/model.py
azaslavsky: Whoops, meant `on_model` in that first sentence!

src/sentry/backup/services/import_export/impl.py
GabeVillalobos: oops, good call 😅

azaslavsky:
LGTM. Let's make sure the `relocation.enabled = False` options change rolls out concurrently with this.

azaslavsky (approved):

PR #74515, opened 2024-07-18
chore(performance): Clean up device class feature checks in the front end due to the flag already being GA'd

edwardgou-sentry:
Clean up device class feature checks in the front end due to the flag already being GA'd


narsaynorath (approved):
If you can, can you merge master up and remove the flag I added in my widget builder test as well?

PR #74517, opened 2024-07-18
feat: implement IS_IN condition

nhsiehgit:
Implements an `IS_IN` condition for issue alerts

Applies to tagged events and event attributes
where `{key} tag IS_IN {comma separated value}`

Jul-18-2024 12-36-56


src/sentry/rules/conditions/event_attribute.py
```diff
@@ -243,69 +243,79 @@ def render_label(self) -> str:
         return self.label.format(**data)
 
     def _passes(self, attribute_values: Sequence[object | None]) -> bool:
-        match = self.get_option("match")
-        value = self.get_option("value")
+        option_match = self.get_option("match")
```
nhsiehgit, line 246: for clarity

src/sentry/rules/conditions/event_attribute.py
```diff
@@ -243,69 +243,79 @@ def render_label(self) -> str:
...
-            return bool(values)
+        elif option_match == MatchType.IS_SET:
+            return bool(option_value)
 
-        elif match == MatchType.NOT_SET:
-            return not values
+        elif option_match == MatchType.NOT_SET:
+            return not option_value
+
+        elif option_match == MatchType.IS_IN:
```
nhsiehgit, line 313: This is the main change

src/sentry/rules/conditions/tagged_event.py
```diff
@@ -64,71 +64,81 @@ def _passes(self, raw_tags: Sequence[tuple[str, Any]]) -> bool:
...
-        elif match == MatchType.NOT_CONTAINS:
-            for t_value in values:
-                if value in t_value:
+        elif option_match == MatchType.NOT_CONTAINS:
+            for t_value in tag_values:
+                if option_value in t_value:
                     return False
             return True
 
+        elif option_match == MatchType.IS_IN:
```
nhsiehgit, line 134: This is the main change

nhsiehgit:
Should this go behind a flag first?
🤷

tests/sentry/rules/conditions/test_event_attribute.py
```diff
@@ -844,3 +844,15 @@ def test_does_not_error_with_none(self):
             data={"match": MatchType.EQUAL, "attribute": "exception.type", "value": "SyntaxError"}
         )
         self.assertPasses(rule, event)
+
+    def test_attr_is_in(self):
+        event = self.get_event()
+        rule = self.get_rule(
+            data={"match": MatchType.IS_IN, "attribute": "platform", "value": "php, python"}
+        )
+        self.assertPasses(rule, event)
```
ceorourke, line 853: Is this saying the event's platform is php?

ceorourke:
I agree this should be behind a flag, and can you post a screenshot of what it looks like on the front end? Have you tested this locally end to end?

tests/sentry/rules/conditions/test_event_attribute.py
nhsiehgit: Ya - this is saying "if the platform attr of the event is in `['php', 'python']`

wedamija (commented):
It'd be nice if the ui made it clear how to enter values as a list. In our search ui, list values are like `[1, 2, 3]`, so this being different might be confusing otherwise

nhsiehgit:
@wedamija Yup - i added `(comma separated)` to the filter

tests/sentry/api/endpoints/test_project_rules_configuration.py
```diff
@@ -213,3 +214,37 @@ def test_high_priority_issue_condition_feature(self):
...
+                ),
+                None,
+            )
+            assert tagged_event_filter
+            filter_list = [
+                choice[0] for choice in tagged_event_filter["formFields"]["match"]["choices"]
+            ]
+            assert MatchType.IS_IN not in filter_list
+
+        # NOTE: failing because the False case above is removing the choice from the registry :think:
```
nhsiehgit, line 235: For some reason this is failing because it looks like the logic is modifying a cached response 🤔 

trying to figure out why this is...

tests/js/fixtures/projectAlertRuleConfiguration.ts
```diff
@@ -249,6 +249,7 @@ export function ProjectAlertRuleConfigurationFixture(
               ['ns', 'is not set'],
               ['nsw', 'does not start with'],
               ['sw', 'starts with'],
+              ['in', 'is in (comma separated)'],
```
nhsiehgit, line 252: cc @wedamija added `(comma separated)` here

src/sentry/api/endpoints/project_rules_configuration.py
```diff
@@ -88,6 +89,15 @@ def get(self, request: Request, project) -> Response:
                     and not has_latest_adopted_release
                 ):
                     continue
+                if (
+                    context["id"] == "sentry.rules.filters.tagged_event.TaggedEventFilter"
+                    or context["id"] == "sentry.rules.filters.event_attribute.EventAttributeFilter"
+                ) and not has_is_in:
+                    # Filter the form_fields to strip `is_in` choices
+                    match_choices = context["formFields"]["match"]["choices"]
+                    context["formFields"]["match"]["choices"] = [
```
nhsiehgit, line 98: hmmmmmm....

Ok - so in the tests this is failing because it seems like this is modifying the global registry for `rules` and i'm not entirley clear if htis is a test issue or whether this would actually affect the host registry or something 🤔 

not entirley sure how i would validate this...

@wedamija would you have any idea?

nhsiehgit: oof deepcopy 🙏 👍

tests/sentry/api/endpoints/test_project_rules_configuration.py
```diff
@@ -213,3 +214,36 @@ def test_high_priority_issue_condition_feature(self):
                 "sentry.rules.conditions.high_priority_issue.ExistingHighPriorityIssueCondition"
                 in [filter["id"] for filter in response.data["conditions"]]
             )
+
+    def test_has_in_feature(self):
+        with self.feature({"organizations:issues-alerts-is-in": False}):
```
saponifi3d, line 219: nit: `issue-alert-in-filter` might be a better name for the feature flag. the first time reading it i was like "what's it in?" 😅

src/sentry/rules/conditions/event_attribute.py
```diff
@@ -243,69 +243,79 @@ def render_label(self) -> str:
...
-        value = value.lower()
+        option_value = option_value.lower()
 
-        values = [str(v).lower() for v in attribute_values if v is not None]
+        attr_values = [str(v).lower() for v in attribute_values if v is not None]
 
-        if match == MatchType.EQUAL:
-            for a_value in values:
-                if a_value == value:
+        if option_match == MatchType.EQUAL:
```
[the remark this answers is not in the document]
nhsiehgit: eh - out of scope for this PR i think.
10yo cramer code 😬

src/sentry/api/endpoints/project_rules_configuration.py
```diff
@@ -88,6 +91,18 @@ def get(self, request: Request, project) -> Response:
                     and not has_latest_adopted_release
                 ):
                     continue
+                if (
+                    context["id"] == "sentry.rules.filters.tagged_event.TaggedEventFilter"
+                    or context["id"] == "sentry.rules.filters.event_attribute.EventAttributeFilter"
```
saponifi3d, line 96: future todo...

i feel like it'd be really cool if we changed this to be a registry pattern. that way we can just have each rule implement anything it wants to in this layer, if we created a registry and registered different life cycles of this processing, then we could move all the code that's specific to a condition into that specific condition.

src/sentry/rules/conditions/tagged_event.py
```diff
@@ -64,71 +64,80 @@ def _passes(self, raw_tags: Sequence[tuple[str, Any]]) -> bool:
...
+            values = option_value.replace(" ", "").split(",")
+            for t_value in tag_values:
+                if t_value in values:
+                    return True
+            return False
```
saponifi3d, line 139: Minor perf improvements to this logic:
```suggestion
            values = set(option_value.replace(" ", "").split(","))
            return any(t_value in values for t_value in tag_values)
```
This will ensure that if a tag is listed multiple times, we will only check it once. This also reduces the runtime from O(n * m) to O(n) because the `any` check will return true on the first match case vs the in check iterating `values`.

src/sentry/rules/conditions/event_attribute.py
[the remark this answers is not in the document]
wedamija: The only way this will improve is if people change it when they touch it though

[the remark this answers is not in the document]
saponifi3d: +1 would be awesome to turn this if/else tree into a reusable function.

src/sentry/rules/conditions/tagged_event.py
```diff
@@ -46,16 +46,16 @@ class TaggedEventCondition(EventCondition):
     }
 
     def _passes(self, raw_tags: Sequence[tuple[str, Any]]) -> bool:
-        key = self.get_option("key")
-        match = self.get_option("match")
-        value = self.get_option("value")
+        option_key = self.get_option("key")
```
saponifi3d, line 49: nitpick / general feedback for the future, it might be a little easier to review if it's only changes needed for the feature. in this case, changing the variable name makes it a little harder to tell that you just want to update the tag check at the end of the logic tree, vs modifying logic throughout.

the way i tend to handle this is to make commits and then stash these changes when i go to open the pr, and move them onto a new branch. that way i get all the wins for me while i'm coding everything and i can then make 2 prs that are a little easier to review.

wedamija (approved):

src/sentry/rules/conditions/event_attribute.py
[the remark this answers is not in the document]
nhsiehgit: hmmmm
ok one moment

saponifi3d (approved):
realized i p much just left nits, :shipit: whenever you're ready 🎉

src/sentry/rules/conditions/tagged_event.py
nhsiehgit: Nice, i'll update this for all condition match types

tests/sentry/api/endpoints/test_project_rules_configuration.py
nhsiehgit: ah yea - since the flag is already merged in options automator i'll keep as is

PR #74522, opened 2024-07-18
feat(anomaly detection): Send subscription update data to Seer

mifu67:
Update the subscription processor to send the updated data from Snuba to Seer’s `detect_anomalies` endpoint. Seer will respond with data about whether or not there has been an anomaly, and we fire an alert if an anomaly was detected.

Closes https://getsentry.atlassian.net/browse/ALRT-141


src/sentry/incidents/subscription_processor.py
```diff
@@ -500,6 +506,86 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
 
         aggregation_value = self.get_aggregation_value(subscription_update)
 
+        self.has_feature_flag = features.has(
+            "organizations:anomaly-detection-alerts", self.subscription.project.organization
+        )
+
+        if self.has_feature_flag:
+            # MF: I would prefer not to have this all be in an if statement. But what can you do ¯\_(ツ)_/¯
```
ceorourke, line 514: You could put this logic into a separate function - that actually might make this easier to test without being able to trigger metric alerts locally.

src/sentry/incidents/subscription_processor.py
```diff
@@ -522,31 +608,73 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
-                ) and not self.check_trigger_matches_status(trigger, TriggerStatus.ACTIVE):
-                    # If the value has breached our threshold (above/below)
-                    # And the trigger is not yet active
-                    metrics.incr("incidents.alert_rules.threshold", tags={"type": "alert"})
-                    # triggering a threshold will create an incident and set the status to active
-                    incident_trigger = self.trigger_alert_threshold(trigger, aggregation_value)
-                    if incident_trigger is not None:
-                        fired_incident_triggers.append(incident_trigger)
+                if self.has_feature_flag:
+                    if trigger.alert_rule.detection_type == AlertRuleDetectionType.DYNAMIC:
```
ceorourke, line 612: You can combine these two if statements

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,6 +690,31 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
 
+    def get_anomaly_detected(
+        self, anomaly
+    ) -> bool:  # TODO: add the anomaly type once that's added to Sentry
+        """
+        Helper function to determine whether we care about an anomaly based on the
+        anomaly type and the alert rule's threshold type.
+        """
+        # need to define constants for anomaly types
+        anomaly_type = anomaly.anomaly_type
```
ceorourke, line 701: Somewhere we need to handle if the user has opted into getting warning level alerts as well, I can't remember if that's in the Seer response currently or not

src/sentry/incidents/subscription_processor.py
```diff
@@ -522,31 +608,73 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+                                )
+                                incident_trigger = self.trigger_alert_threshold(
+                                    trigger, aggregation_value
+                                )
+                                if incident_trigger is not None:
+                                    fired_incident_triggers.append(incident_trigger)
+                            else:
+                                self.trigger_alert_counts[trigger.id] = 0
+
+                            # MF: There's no explicit "resolve threshold" for anomaly detection. I chose to resolve if no anomaly detected, but unsure if that's correct.
```
ceorourke, line 631: This makes sense to me - we might change it in the future but for now it's fine.

src/sentry/incidents/subscription_processor.py
mifu67: I don't think it is yet, so I'll add a note to come back to it

mifu67: I'll probably have the function return an enum to handle the various cases

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,6 +622,104 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
 
+    def get_anomaly_detected(
+        self, anomaly
+    ) -> bool:  # TODO: add the anomaly type once that's added to Sentry
+        """
+        Helper function to determine whether we care about an anomaly based on the
+        anomaly type and the alert rule's threshold type.
+        """
+        # need to define constants for anomaly types
+        anomaly_type = anomaly.get("anomaly").get("anomaly_type")
```
ceorourke, line 633: `.get()` has a default value of `None` (like if `anomaly'` isn't found) so to safely chain them you can add a default empty dict like `anomaly_type = anomaly.get("anomaly", {}).get("anomaly_type")` otherwise you risk the 2nd `.get()` erroring because you can't do `.get()` on `None`

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,6 +622,104 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+                "id": self.alert_rule.id,
+                "cur_window": {
+                    "timestamp": self.last_update,
+                    "value": aggregation_value,
+                },
+            }
+            response = make_signed_seer_api_request(
+                self.seer_anomaly_detection_connection_pool,
+                SEER_ANOMALY_DETECTION_URL,
+                json.dumps(
```
ceorourke, line 670: This should be the URL for the `detect_anomalies_endpoint`, e.g. https://github.com/getsentry/sentry/blob/master/src/sentry/conf/server.py#L3417-L3429 so for this it would be the URL from the decorator here: https://github.com/getsentry/seer/blob/f0b1cfd9cf9c04300307b549bd36c38fc1097c8a/src/seer/app.py#L248

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,6 +622,104 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+                    {
+                        "organization_id": self.subscription.project.organization.id,
+                        "project_id": self.subscription.project_id,
+                        "config": ad_config,
+                        "context": context,
+                    }
+                ),
```
ceorourke, line 677: I think this needs `.encode("utf-8")`

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,6 +622,104 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+            )
+        except (TimeoutError, MaxRetryError):
+            logger.warning(
+                "Timeout error when hitting anomaly detection endpoint",
+                extra={
+                    "subscription_id": self.subscription.id,
+                    "dataset": self.subscription.snuba_query.dataset,
+                    # TODO (MF): add other, relevant fields here
```
ceorourke, line 685: Maybe the alert rule id, project id and/or org id?

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -285,6 +291,18 @@ def comparison_rule_below(self):
...
 
+    @cached_property
+    def dynamic_rule(self):
+        rule = self.rule
+        rule.update(
+            detection_type=AlertRuleDetectionType.DYNAMIC,
+            sensitivity=AlertRuleSensitivity.HIGH,
+            seasonality=AlertRuleSeasonality.AUTO,
+        )
+        # dynamic alert rules have a threshold of 0.0
```
ceorourke, line 302: Do we check this in the serializer? If not, do you mind adding a ticket in the epic for us to make sure we do?

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -408,6 +426,123 @@ def test_no_alert(self):
         self.assert_trigger_does_not_exist(self.trigger)
         self.assert_action_handler_called_with_actions(None, [])
 
+    @mock.patch(
+        "sentry.incidents.subscription_processor.SubscriptionProcessor.seer_anomaly_detection_connection_pool.urlopen"
+    )
+    def test_seer_call(self, mock_seer_request: MagicMock):
+        with self.feature(["organizations:incidents", "organizations:anomaly-detection-alerts"]):
```
ceorourke, line 433: nbd but optionally you can also do this as a decorator like
```
from sentry.testutils.helpers.features import with_feature

@with_feature("organizations:anomaly-detection-alerts")
@with_feature("organizations:incidents")
```

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -408,6 +426,123 @@ def test_no_alert(self):
...
+            assert mock_seer_request.assert_called_once
+            assert mock_seer_request.call_args.args[0] == "POST"
+            assert mock_seer_request.call_args.args[1] == SEER_ANOMALY_DETECTION_URL
+            deserialized_body = json.loads(mock_seer_request.call_args.kwargs["body"])
+            assert deserialized_body["organization_id"] == self.sub.project.organization.id
+            assert deserialized_body["project_id"] == self.sub.project_id
+            assert deserialized_body["config"]["time_period"] == rule.threshold_period
+            assert deserialized_body["config"]["sensitivity"] == rule.sensitivity.value
+            assert deserialized_body["config"]["seasonality"] == rule.seasonality.value
+            assert deserialized_body["config"]["direction"] == rule.detection_type.value
```
ceorourke, line 460: `direction` should be `rule.threshold_type` I think

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,6 +622,104 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+                return True
+        return False
+
+    def get_anomaly_data_from_seer(self, aggregation_value: float | None):
+        try:
+            ad_config = {
+                "time_period": self.alert_rule.threshold_period,
+                "sensitivity": self.alert_rule.sensitivity,
+                "seasonality": self.alert_rule.seasonality,
+                "direction": self.alert_rule.detection_type,
```
ceorourke, line 657: Oop this too, `direction` is the `threshold_type`

src/sentry/incidents/subscription_processor.py
```diff
@@ -500,6 +511,15 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
 
         aggregation_value = self.get_aggregation_value(subscription_update)
 
+        self.has_feature_flag = features.has(
```
nhsiehgit, line 514: Lets name this something more explicit?
`self.has_anomaly_detection`?

src/sentry/incidents/subscription_processor.py
```diff
@@ -522,31 +542,71 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
-                    and self.check_trigger_matches_status(trigger, TriggerStatus.ACTIVE)
+                    self.has_feature_flag
+                    and trigger.alert_rule.detection_type == AlertRuleDetectionType.DYNAMIC
                 ):
-                    metrics.incr("incidents.alert_rules.threshold", tags={"type": "resolve"})
-                    incident_trigger = self.trigger_resolve_threshold(trigger, aggregation_value)
-
-                    if incident_trigger is not None:
-                        fired_incident_triggers.append(incident_trigger)
+                    # MF: There should only be one anomaly in the list
```
nhsiehgit, line 549: Is this a `NOTE:`? or a `TODO:`?

src/sentry/incidents/subscription_processor.py
```diff
@@ -522,31 +542,71 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+                    for anomaly in anomalies:
+                        if self.get_anomaly_detected(
```
nhsiehgit, line 551: `get_anomaly_detected`... sounds like an incomplete phrase?
maybe
`get_anomaly`?
or if this is being evalutated as a boolean - `has_anomaly`?

src/sentry/incidents/subscription_processor.py
```diff
@@ -522,31 +542,71 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+                            metrics.incr("incidents.alert_rules.threshold", tags={"type": "alert"})
+                            incident_trigger = self.trigger_alert_threshold(
+                                trigger, aggregation_value
+                            )
+                            if incident_trigger is not None:
+                                fired_incident_triggers.append(incident_trigger)
+                        else:
+                            self.trigger_alert_counts[trigger.id] = 0
+
+                        # MF: There's no explicit "resolve threshold" for anomaly detection. I chose to resolve if no anomaly detected, but unsure if that's correct.
```
nhsiehgit, line 563: for notes - you don't need to leave your initials here

If you mean for it to live in code, then just keep it as a `NOTE:`

nhsiehgit: Also yes - no anomaly i would assume means "resolved"

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,6 +622,106 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
         # before the next one then we might alert twice.
         self.update_alert_rule_stats()
 
+    def get_anomaly_detected(
```
nhsiehgit, line 625: `has_anomaly`?

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,6 +622,106 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+        self, anomaly
+    ) -> bool:  # TODO: add the anomaly type once that's added to Sentry
```
nhsiehgit, line 627: you can leave the `TODO:` in the docstring i guess?
Or maybe just a comment below the docstring

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,6 +622,106 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+        """
+        Helper function to determine whether we care about an anomaly based on the
+        anomaly type and the alert rule's threshold type.
+        """
+        # need to define constants for anomaly types
```
nhsiehgit, line 632: TODO?

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,6 +622,106 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
         # before the next one then we might alert twice.
         self.update_alert_rule_stats()
 
+    def get_anomaly_detected(
+        self, anomaly
```
nhsiehgit, line 626: Are we passing the `anomaly` here?

Or is it just the results from seer?
shoudl this be `seer_results` or something?

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,6 +622,106 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+        anomaly_type = anomaly.get("anomaly", {}).get("anomaly_type")
+
+        if anomaly_type == "none" or anomaly_type == "no_data":
+            return False
+        elif anomaly_type == "anomaly_high":
+            if (
+                self.alert_rule.threshold_type == AlertRuleThresholdType.ABOVE.value
+                or self.alert_rule.threshold_type == AlertRuleThresholdType.ABOVE_AND_BELOW.value
+            ):
+                return True
```
nhsiehgit, line 642: Is this logic correct? 🤔 

So - `anomaly_high` only triggers for above/a&b anomalies?
isn't `anomaly_high` just mean we have high confidence that htis is an anomaly? Not that it's 'above' any trheshold?

src/sentry/incidents/subscription_processor.py
```diff
@@ -500,6 +511,15 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
 
         aggregation_value = self.get_aggregation_value(subscription_update)
 
+        self.has_feature_flag = features.has(
+            "organizations:anomaly-detection-alerts", self.subscription.project.organization
+        )
+
+        if self.has_feature_flag:
+            anomalies = self.get_anomaly_data_from_seer(aggregation_value)
```
nhsiehgit, line 519: these are not necessarily 'anomalies' so maybe we should rename this?
`anomaly_check_results` or something? 🤔

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,6 +622,106 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+            if (
+                self.alert_rule.threshold_type == AlertRuleThresholdType.BELOW.value
+                or self.alert_rule.threshold_type == AlertRuleThresholdType.ABOVE_AND_BELOW.value
+            ):
+                return True
+        return False
+
+    def get_anomaly_data_from_seer(self, aggregation_value: float | None):
+        try:
+            ad_config = {
```
nhsiehgit, line 653: lets spell out the full words rather than using any acronyms

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,6 +622,106 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+                    "alert_rule_id": self.alert_rule.id,
+                },
+            )
+            return None
+
+        # TODO (MF): handle response codes if status code != 200
+        # MF: But it doesn't seem like the mock response has status codes. So.
+
+        try:
+            anomalies = json.loads(response.data.decode("utf-8")).get("anomalies")
```
nhsiehgit, line 696: `results`

nhsiehgit (commented):
comment comment comments

src/sentry/incidents/subscription_processor.py
mifu67: Ah, it was just for me to ctrl-f to find my notes easier! It should be deleted

src/sentry/incidents/subscription_processor.py
mifu67: We pass each anomaly object here.

src/sentry/incidents/subscription_processor.py
mifu67: Confirmed with Tillman that the logic is correct.

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -285,6 +292,18 @@ def comparison_rule_below(self):
...
+    @cached_property
+    def dynamic_rule(self):
+        rule = self.rule
+        rule.update(
+            detection_type=AlertRuleDetectionType.DYNAMIC,
+            sensitivity=AlertRuleSensitivity.HIGH,
+            seasonality=AlertRuleSeasonality.AUTO,
+        )
+        # dynamic alert rules have a threshold of 0.0
+        self.trigger.update(alert_threshold=0)
```
nhsiehgit, line 304: 🤔 

not quite sure, but firing updates within the cached property seems odd to me...
Are you seeing what you expect to from these tests?

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -408,6 +427,231 @@ def test_no_alert(self):
...
+                {
+                    "anomaly": {"anomaly_score": 0.7, "anomaly_type": "anomaly_low"},
+                    "timestamp": 1,
+                    "value": 5,
+                }
+            ]
+        }
+
+        mock_seer_request.return_value = HTTPResponse(orjson.dumps(seer_return_value_1), status=200)
+        processor = self.send_update(rule, 5, timedelta(minutes=-3))
```
nhsiehgit, line 457: personally - would prefer all calls be kwargs rather than implicit ordering

nhsiehgit (commented):
few more comments

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -408,6 +427,231 @@ def test_no_alert(self):
...
+        incident = self.assert_active_incident(rule)
+        self.assert_trigger_exists_with_status(incident, warning_trigger, TriggerStatus.ACTIVE)
+        self.assert_trigger_exists_with_status(incident, trigger, TriggerStatus.ACTIVE)
+        self.assert_actions_fired_for_incident(
+            incident,
+            [warning_action],
+            [(10, IncidentStatus.CRITICAL, mock.ANY)],
+        )
+
+        # trigger a resolution
```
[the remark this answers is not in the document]
mifu67: It should be the same test (can't trigger a resolution without having an active alert)

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -408,6 +427,231 @@ def test_no_alert(self):
...
+        incident = self.assert_active_incident(rule)
+        self.assert_trigger_exists_with_status(incident, warning_trigger, TriggerStatus.ACTIVE)
+        self.assert_trigger_does_not_exist(trigger)
+        self.assert_actions_fired_for_incident(
+            incident,
+            [warning_action],
+            [(5, IncidentStatus.WARNING, mock.ANY)],
+        )
+
+        # trigger critical
```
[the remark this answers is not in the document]
mifu67: I wanted to check that you can go from warning to critical without any issue. Being able to trigger the warning on its own confirms that triggering a single alert works.

src/sentry/incidents/subscription_processor.py
```diff
@@ -543,13 +543,13 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
-                    self.has_feature_flag
+                    self.has_anomaly_detection
                     and trigger.alert_rule.detection_type == AlertRuleDetectionType.DYNAMIC
                 ):
-                    # MF: There should only be one anomaly in the list
-                    for anomaly in anomalies:
-                        if self.get_anomaly_detected(
-                            anomaly
+                    # NOTE: There should only be one anomaly in the list
+                    for potential_anomaly in potential_anomalies:
```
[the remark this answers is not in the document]
mifu67: If we make it past the `get_anomaly_data_from_seer` function and return a list of potential anomalies, then the objects in the list are guaranteed to be potential anomaly objects.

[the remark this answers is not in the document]
mifu67: The checks for badly formatted responses are in `get_anomaly_data_from_seer`, so I didn't feel that it was necessary to have additional checks for badly formatted data here.

src/sentry/incidents/subscription_processor.py
```diff
@@ -522,31 +542,70 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
-
-                    if incident_trigger is not None:
-                        fired_incident_triggers.append(incident_trigger)
+                    # NOTE: There should only be one anomaly in the list
+                    for potential_anomaly in potential_anomalies:
+                        if self.has_anomaly(
+                            potential_anomaly
+                        ) and not self.check_trigger_matches_status(trigger, TriggerStatus.ACTIVE):
+                            metrics.incr("incidents.alert_rules.threshold", tags={"type": "alert"})
+                            incident_trigger = self.trigger_alert_threshold(
```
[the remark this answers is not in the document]
mifu67: The trigger threshold is zero, so threshold _should_ always be exceeded (unless the user defines <0 for some reason, which would never trigger anyway) and anomaly count is being used to detect whether we should fire here.

src/sentry/incidents/subscription_processor.py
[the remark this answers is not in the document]
mifu67: Actually, maybe it's best to have some safety code (even if it never gets triggered)

src/sentry/incidents/subscription_processor.py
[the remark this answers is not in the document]
mifu67: Hmm, wait, I see the problem here. Will fix.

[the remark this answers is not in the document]
nhsiehgit: Oh damn that's right 🤔 
ok!

nhsiehgit (approved):

tests/sentry/incidents/test_subscription_processor.py
mifu67: [https://getsentry.atlassian.net/browse/ALRT-165](url)

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -408,6 +427,232 @@ def test_no_alert(self):
...
+    )
+    @mock.patch("sentry.incidents.subscription_processor.logger")
+    def test_seer_call_bad_status(self, mock_logger, mock_seer_request):
+        processor = SubscriptionProcessor(self.sub)
+        mock_seer_request.return_value = HTTPResponse("You flew too close to the sun", status=403)
+        result = processor.get_anomaly_data_from_seer(10)
+        assert mock_logger.error.called_with(
+            f"Received 403 when calling Seer endpoint {SEER_ANOMALY_DETECTION_ENDPOINT_URL}.",  # noqa
+            extra={"response_data": "You flew too close to the sun"},
+        )
```
asottile-sentry, line 641: this assertion will always be true -- `called_with` will return a mock object which is always truthy -- I believe you're looking for `assert_called_with`  (same below)

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -408,6 +427,232 @@ def test_no_alert(self):
...
+        assert result is None
+
+    @with_feature("organizations:anomaly-detection-alerts")
+    @mock.patch(
+        "sentry.incidents.subscription_processor.SubscriptionProcessor.seer_anomaly_detection_connection_pool.urlopen"
+    )
+    @mock.patch("sentry.incidents.subscription_processor.logger")
+    def test_seer_call_failed_parse(self, mock_logger, mock_seer_request):
+        processor = SubscriptionProcessor(self.sub)
+        mock_seer_request.return_value = HTTPResponse(None, status=200)  # type: ignore[arg-type]
```
asottile-sentry, line 651: also please do not ignore the type checker -- fix the actual problem instead

Issue #73039, opened 2024-06-20
Insights - Mobile UI


philipphofmann:
### Problem Statement

TODO

### Product Area

Insights

philippsuess:
Closing this because it's out of date.

PR #74523, opened 2024-07-18
feat(insights): Switch to ratios for slow and frozen frames

markushi:
As described [here](https://github.com/getsentry/sentry/issues/73039), we want to move to ratios for slow and frozen frames.

Utilizes [division_if](https://github.com/getsentry/sentry/pull/72995) to calculate the ratios.

image


static/app/utils/discover/fieldRenderers.tsx
```diff
@@ -819,6 +819,10 @@ export function getSortField(
     return field;
   }
 
+  if (field.startsWith('division_if(')) {
```
markushi, line 822: This feels like a hack 🙈, an alternative would be adding them to aggregations, but then we'd need to support this for the discover as well (see count_if as an example, https://github.com/getsentry/sentry/blob/d10148cf0bc3bf203ba0781378956925cef68f4f/static/app/utils/discover/fields.tsx#L291-L318) - not too sure if this would blow stuff up, as the backend support for division_if is tightly bound to releases.

static/app/components/gridEditable/index.tsx
```diff
@@ -45,6 +45,7 @@ export type GridColumn<K = ObjectKey> = {
 
 export type GridColumnHeader<K = ObjectKey> = GridColumn<K> & {
   name: string;
+  tooltip?: string;
```
markushi, line 48: This is bad, as I'm adding the definition here, but `gridEditable` doesn't use this prop. Ideally it would support tooltips out of the box, I can take a look into this in a separate PR.

static/app/views/insights/mobile/common/components/tables/screensTable.tsx
```diff
@@ -129,6 +147,7 @@ export function ScreensTable({
             key: columnKey,
             name: columnNameMap[columnKey],
             width: COL_WIDTH_UNDEFINED,
+            toolTip: columnTooltipMap ? columnTooltipMap[columnKey] : undefined,
```
narsaynorath, line 150: Instead of having to parse the tooltip out from the column tooltip map here, is it possible to just use the key in the `renderHeadCell` helper and pull the tooltip out from the map there? That way we can avoid having to edit the column type

We do want tooltips in headers one day so I think that's a good PR to have in the future, but for now this should work

static/app/utils/discover/fieldRenderers.tsx
narsaynorath: I think there's a cleaner way to get around this, but I have to do a bit of investigation. I'm seeing some odd access paths when it comes to the structure of `tableMeta`. If you want, we can remove this until I get a chance to poke around a bit more and we can put up a separate PR to handle the sortability of the headers.

The endpoint returns in the meta data that the corresponding type to the `division_if` statements _should_ be percentages so they should be sortable by default when it gets to [this block](https://github.com/getsentry/sentry/pull/74523/files#diff-04dbefa42672572db5ea85f1a074f5eaf3c6df7703c0e73dfa88747a84335e18R833). I know that the key gets kind of mangled for equations but we should be able to work around that

narsaynorath (commented):
Just a few comments. I think the conversion for the table looks good but there's probably an alternate way to access the tooltips without editing the type and also fixing the sorting without hardcoding the `division_if` prefix.

I think we should pull out the field renderer part and have a separate PR. I can take a look deeper tomorrow on how to get around it

static/app/views/insights/mobile/common/components/tables/screensTable.tsx
markushi: Oh right, yeah that should do the trick, thanks!

narsaynorath (approved):
looks good to me! thanks for splitting out the sorting changes we discussed

PR #74524, opened 2024-07-18
feat(query-builder): Add support for percentage filters

### LINKED ISSUE #74305: SearchQueryBuilder: Support percentage filters

malwilley:
Closes https://github.com/getsentry/sentry/issues/74305

Turns out that percentage filters, as we use them now, are actually numeric filters under the hood. Therefore, the behavior I'm adding allows the user to input `%` values but converts them to a ratio. See this comment for more info: https://github.com/getsentry/sentry/issues/74305#issuecomment-2237207619

https://github.com/user-attachments/assets/05c99ccd-f1c0-48fc-a6e7-e6021f47b33b

### DIFF
--- a/static/app/components/searchQueryBuilder/index.spec.tsx
+++ b/static/app/components/searchQueryBuilder/index.spec.tsx
@@ -1,4 +1,5 @@
 import type {ComponentProps} from 'react';
+import {destroyAnnouncer} from '@react-aria/live-announcer';
 
 import {
   render,
@@ -99,6 +100,9 @@ describe('SearchQueryBuilder', function () {
   beforeEach(() => {
     // `useDimensions` is used to hide things when the component is too small, so we need to mock a large width
     Object.defineProperty(Element.prototype, 'clientWidth', {value: 1000});
+
+    // Combobox announcements will pollute the test output if we don't clear them
+    destroyAnnouncer();
   });
 
   afterEach(function () {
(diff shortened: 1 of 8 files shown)


static/app/components/searchQueryBuilder/index.spec.tsx
```diff
@@ -99,6 +100,9 @@ describe('SearchQueryBuilder', function () {
   beforeEach(() => {
     // `useDimensions` is used to hide things when the component is too small, so we need to mock a large width
     Object.defineProperty(Element.prototype, 'clientWidth', {value: 1000});
+
+    // Combobox announcements will pollute the test output if we don't clear them
+    destroyAnnouncer();
```
MichaelSun48, line 105: the most violent function name i've ever seen

MichaelSun48 (approved):

static/app/components/searchQueryBuilder/tokens/filter/valueCombobox.tsx
```diff
@@ -480,6 +483,18 @@ function cleanFilterValue(
...
     }
+    case FieldValueType.PERCENTAGE: {
+      const parsed = parseFilterValuePercentage(value);
+      if (!parsed) {
+        return null;
+      }
+      // If the user passes "50%", convert to 0.5
+      if (parsed.unit) {
+        const numericValue = parseFloat(parsed.value);
+        return isNaN(numericValue) ? parsed.value : (numericValue / 100).toString();
```
roggenkemper, line 494: this might be a dumb question but in what scenario would we pass the `parsed.unit` condition, but still end up with `numericValue` being NaN

malwilley: Not a dumb question, I don't think this should actually be possible. Probably not necessary, I am being overly safe here

PR #74525, opened 2024-07-18
feat(tags): Add dataset param to OrganizationTagKeyValues endpoint

MichaelSun48:
This PR adds a dataset parameter to the OrganizationTagKeyValues endpoint, allowing full control over which dataset will be queried. Our frontend currently queries the Discover dataset to retrieve tag values for the issue stream - we need this control to query only the Events and IssuePlatform datasets.

### DIFF
--- a/src/sentry/api/endpoints/organization_tagkey_values.py
+++ b/src/sentry/api/endpoints/organization_tagkey_values.py
@@ -10,6 +10,7 @@ from sentry.api.bases import NoProjects, OrganizationEventsEndpointBase
 from sentry.api.paginator import SequencePaginator
 from sentry.api.serializers import serialize
 from sentry.api.utils import handle_query_errors
+from sentry.snuba.dataset import Dataset
 from sentry.tagstore.base import TAG_KEY_RE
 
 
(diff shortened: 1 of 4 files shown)


vartec (approved):

src/sentry/api/endpoints/organization_tagkey_values.py
```diff
@@ -31,6 +32,13 @@ def get(self, request: Request, organization, key) -> Response:
 
         sentry_sdk.set_tag("query.tag_key", key)
 
+        dataset = None
+        if request.GET.get("dataset"):
```
vartec, line 36:
```suggestion
        if dataset := request.GET.get("dataset"):
```

src/sentry/api/endpoints/organization_tagkey_values.py
```diff
@@ -31,6 +32,12 @@ def get(self, request: Request, organization, key) -> Response:
 
         sentry_sdk.set_tag("query.tag_key", key)
 
+        if dataset := request.GET.get("dataset"):
```
malwilley, line 35: Downside of this is that `dataset` starts as a string and becomes type `Dataset` later on which breaks typing a bit

malwilley (approved):
Glad it turned out to be simpler than expected!

MichaelSun48:
Going to merge this on Monday rather than Today since these changes have no potential benefits until the respective frontend changes are made

Issue #74448, opened 2024-07-17
Regression version comparison incorrectly identified issue as regressed, when version was older than resolved-in version


jboulter11:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

See:
https://dropbox.sentry.io/issues/5597803787/activity/?project=4505546600087552

I do not know why this happened.

### Expected Result

Sentry should have compared the version numbers 388.0.6 vs 388.0.8 and concluded there was no regression.

### Actual Result

Sentry compared 388.0.6 and 388.0.8 and determined 388.0.6 to be equal to or greater than 388.0.8, which is incorrect.

### Product Area

Issues

### Link

https://dropbox.sentry.io/issues/5597803787/activity/?project=4505546600087552

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2226) by [Unito](https://www.unito.io)

armenzg:
Hey @jboulter11 👋 
Thanks for reporting. I will have a look.

armenzg:
I have placed the issue into the backlog after attempting a fix.

These are my notes from today:
* The developer had marked the issue to be resolved in [388.0 (388.0.8)](https://dropbox.sentry.io/releases/com.getdropbox.Dropbox%40388.0%2B388.0.8/?project=4505546600087552)
  * I have noticed that the releases differ in the name of the release `com.getdropbox.Dropbox` vs `com.getdropbox.DropboxTest`
![Image](https://github.com/user-attachments/assets/5b33f98a-112e-4452-b289-c8c91be123d5)
* The event that caused the regression was on release [388.0 (388.0.6)](https://dropbox.sentry.io/releases/com.dropbox.DropboxTest%40388.0%2B388.0.6/) at 2:24pm
![Image](https://github.com/user-attachments/assets/c7c9b437-8b26-445c-9ad8-a3be2de8929e)
  * Noticed that the patch version is 6 which is less than 8
* I have attempted to create a test case in #74526 
  * More time is needed to determine if the test case was written properly as it does not reproduce the reported issue

armenzg:
For anyone who ends up picking this up, this PR tries to reproduce this issue:
https://github.com/getsentry/sentry/pull/74526/files

jboulter11:
This happened again but this time with versions using `release date` instead of semver: https://dropbox.sentry.io/issues/5918496859/activity/?project=4505546831036416

[...]

PR #74526, opened 2024-07-18
test(event_manager): Test case 

armenzg:
This attempts to reproduce the test case for issue https://github.com/getsentry/sentry/issues/74448.

### DIFF
--- a/src/sentry/event_manager.py
+++ b/src/sentry/event_manager.py
@@ -2109,8 +2109,8 @@ def _handle_regression(group: Group, event: BaseEvent, release: Release | None) 
 
     # we only mark it as a regression if the event's release is newer than
     # the release which we originally marked this as resolved
-    elif GroupResolution.has_resolution(group, release):
-        return None
+    # elif GroupResolution.has_resolution(group, release):
+    #     return None
 
     elif has_pending_commit_resolution(group):
         return None
--- a/tests/sentry/event_manager/test_event_manager.py
+++ b/tests/sentry/event_manager/test_event_manager.py
@@ -628,11 +628,20 @@ class EventManagerTest(TestCase, SnubaTestCase, EventManagerTestMixin, Performan
         assert event.group is not None
         group = event.group
 
-        # XXX: The GH issue shows that it should have unresolved
-        assert group.status == GroupStatus.RESOLVED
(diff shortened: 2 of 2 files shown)


src/sentry/event_manager.py
```diff
@@ -2109,8 +2109,8 @@ def _handle_regression(group: Group, event: BaseEvent, release: Release | None)
 
     # we only mark it as a regression if the event's release is newer than
     # the release which we originally marked this as resolved
-    elif GroupResolution.has_resolution(group, release):
-        return None
+    # elif GroupResolution.has_resolution(group, release):
+    #     return None
```
armenzg, line 2113: I'm forcing the issue to happen.

In my test case, we end up reaching this comparison:
https://github.com/getsentry/sentry/blob/159d52c5273b0d6b326e3b2669635cd4a880d0a7/src/sentry/models/groupresolution.py#L156-L172

tests/sentry/event_manager/test_event_manager.py
```diff
@@ -564,6 +564,85 @@ def test_resolved_in_release_regression_activity_follows_semver(
         assert regressed_activity.data["follows_semver"] is True
         assert regressed_activity.data["resolved_in_version"] == "foo@1.0.0"
 
+    def resolve_in_release(self, group: Group, release: Release) -> Activity:
+        # Resolved in a future known release
```
armenzg, line 568:
```suggestion
```

tests/sentry/event_manager/test_event_manager.py
```diff
@@ -564,6 +564,85 @@ def test_resolved_in_release_regression_activity_follows_semver(
...
+        patch_version6 = "com.foo.FooTest@388.0+388.0.6"
+        release_06 = self.create_release(version=patch_version6, date_added=_date(70))
+
+        patch_version8 = "com.foo.Foo@388.0+388.0.8"
+        release_08 = self.create_release(version=patch_version8, date_added=_date(65))
+
+        self.create_release(version="com.foo.Foo@388.0+388.0.13", date_added=_date(60))
+
+        # Marking as resolved in the second latest release
+        self.resolve_in_release(group, release_08)
```
armenzg, line 615: All the releases above already exist once the developer calls "resolve in release X".

PR #74527, opened 2024-07-18
ref(replay/feedback): default frontend framework option is js

michellewzhang:
make javascript the default & first frontend platform for replay & feedback onboarding: 

SCR-20240718-mnqv

### DIFF
--- a/static/app/components/replaysOnboarding/utils.tsx
+++ b/static/app/components/replaysOnboarding/utils.tsx
@@ -3,9 +3,13 @@ import platforms from 'sentry/data/platforms';
 import type {PlatformIntegration} from 'sentry/types/project';
 
 export function replayJsFrameworkOptions(): PlatformIntegration[] {
+  // the platforms array is sorted alphabetically, but we want javascript to be
+  // at the front so that it shows up by default in the onboarding.
   const frameworks = platforms.filter(p => replayFrontendPlatforms.includes(p.id));
   const jsPlatformIdx = frameworks.findIndex(p => p.id === 'javascript');
   const jsPlatform = frameworks[jsPlatformIdx];
+
+  // move javascript to the front
   frameworks.splice(jsPlatformIdx, 1);
   frameworks.unshift(jsPlatform);
   return frameworks;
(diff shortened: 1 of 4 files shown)


static/app/components/replaysOnboarding/utils.tsx
```diff
@@ -2,6 +2,11 @@ import {replayFrontendPlatforms} from 'sentry/data/platformCategories';
...
-export const replayJsFrameworkOptions: PlatformIntegration[] = platforms.filter(p =>
-  replayFrontendPlatforms.includes(p.id)
-);
+export function replayJsFrameworkOptions(): PlatformIntegration[] {
+  const frameworks = platforms.filter(p => replayFrontendPlatforms.includes(p.id));
+  const jsPlatformIdx = frameworks.findIndex(p => p.id === 'javascript');
+  const jsPlatform = frameworks[jsPlatformIdx];
+  frameworks.splice(jsPlatformIdx, 1);
+  frameworks.unshift(jsPlatform);
+  return frameworks;
```
michellewzhang, line 11: had to do this logic bc `platforms` (used across the codebase so don't want to mess with it) is sorted alphabetically :(

static/app/data/platformCategories.tsx
```diff
@@ -346,13 +346,14 @@ export const replayBackendPlatforms: readonly PlatformKey[] = [
 
 // These are the frontend platforms that can set up replay.
 export const replayFrontendPlatforms: readonly PlatformKey[] = [
-  'capacitor',
-  'electron',
+  'javascript',
```
ryan953, line 349: some kind of comment like this will be helpful for the next person who comes by and wants to sort the list:
```suggestion
  'javascript', // First in the list so that it can be used as the default option in platform choosing dropdown
```

ryan953 (approved):

static/app/data/platformCategories.tsx
michellewzhang: this list actually does nothing :( what pushes `javascript` to the top is the logic i do inside `utils.tsx`

michellewzhang: can add a comment there though

PR #74528, opened 2024-07-18
analytics(replay): analytics for hydration issue details and slider interactions

aliu39:
Ref https://github.com/getsentry/sentry/issues/74456

- Tracks first render of a hydrate error on issue details, after page load
- Tracks first click on the slider's divider bar, after the last render of the slider diff

Current behavior:
https://github.com/user-attachments/assets/0a6431d1-00a8-43fd-befb-0807d432518c

### DIFF
--- a/static/app/components/replays/diff/replaySliderDiff.tsx
+++ b/static/app/components/replays/diff/replaySliderDiff.tsx
@@ -1,4 +1,4 @@
-import {Fragment, useRef} from 'react';
+import {Fragment, useCallback, useRef, useState} from 'react';
 import styled from '@emotion/styled';
 
 import NegativeSpaceContainer from 'sentry/components/container/negativeSpaceContainer';
@@ -6,9 +6,11 @@ import ReplayIFrameRoot from 'sentry/components/replays/player/replayIFrameRoot'
 import {StaticReplayPreferences} from 'sentry/components/replays/preferences/replayPreferences';
 import {Provider as ReplayContextProvider} from 'sentry/components/replays/replayContext';
 import {space} from 'sentry/styles/space';
+import {trackAnalytics} from 'sentry/utils/analytics';
 import toPixels from 'sentry/utils/number/toPixels';
 import type ReplayReader from 'sentry/utils/replays/replayReader';
 import {useDimensions} from 'sentry/utils/useDimensions';
+import useOrganization from 'sentry/utils/useOrganization';
 import {useResizableDrawer} from 'sentry/utils/useResizableDrawer';
 
 interface Props {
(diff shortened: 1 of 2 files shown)


static/app/components/replays/diff/replaySliderDiff.tsx
```diff
@@ -63,6 +66,23 @@ function DiffSides({leftOffsetMs, replay, rightOffsetMs, viewDimensions, width})
     },
   });
 
+  const organization = useOrganization();
+  const [dividerClicked, setDividerClicked] = useState(false); // never flips back to false
```
ryan953, line 70: I would create this with `useRef` instead, right now it'll cause a re-render which isn't needed as we're just tracking the value not rendering anything.

```
  const dividerClickedRef = useRef(false);
  ...
  if (organization && !dividerClickedRef.current) {
    ...
    dividerClickedRef.current = true
```

aliu39:
@ryan953 Got it. What are your thoughts on tracking first click after render vs after page load?

ryan953:
> @ryan953 Got it. What are your thoughts on tracking first click after render vs after page load?

first click the way it is now, after render, seems fine. the person can use it a bunch of times through the day, we just don't need every single little mouse move

static/app/components/replays/diff/replaySliderDiff.tsx
```diff
@@ -63,6 +66,23 @@ function DiffSides({leftOffsetMs, replay, rightOffsetMs, viewDimensions, width})
     },
   });
 
+  const organization = useOrganization();
+  const dividerClickedRef = useRef(false); // once set, never flips back to false
+
+  const onDividerMouseDownWithAnalytics: React.MouseEventHandler<HTMLElement> =
```
ryan953, line 72: nice types!

static/app/components/replays/diff/replaySliderDiff.tsx
```diff
@@ -44,7 +46,8 @@ export function ReplaySliderDiff({leftOffsetMs, replay, rightOffsetMs}: Props) {
 function DiffSides({leftOffsetMs, replay, rightOffsetMs, viewDimensions, width}) {
   const rightSideElem = useRef<HTMLDivElement>(null);
   const dividerElem = useRef<HTMLDivElement>(null);
-  const {onMouseDown} = useResizableDrawer({
+
+  const {onMouseDown: onDividerMouseDown} = useResizableDrawer({
```
ryan953, line 50: that is a better name `onDividerMouseDown`

static/app/views/issueDetails/header.tsx
```diff
@@ -83,6 +84,12 @@ function GroupHeaderTabs({
     group_has_replay: (replaysCount ?? 0) > 0,
   });
 
+  useEffect(() => {
+    if (organization && group.issueType === IssueType.REPLAY_HYDRATION_ERROR) {
+      trackAnalytics('replay.hydration-error.issue-details', {organization});
+    }
```
ryan953, line 90: typescript thinks that organization will always be set here, so we should be good without it in the check
```suggestion
    if (group.issueType === IssueType.REPLAY_HYDRATION_ERROR) {
      trackAnalytics('replay.hydration-error.issue-details', {organization});
    }
```

ryan953 (approved):

PR #74531, opened 2024-07-18
feat(issue-search): Check `max_candidates` to push Postgres fields to post-filtering

snigdhas:
When applying query filters from Postgres, pre-filtering the results before passing the query to Snuba can break if it hits the max size that ClickHouse is able to process ([sentry issue](https://sentry.sentry.io/issues/5419972487/events/e14ff2501904443eb1c38f7600509b4f/?project=1&referrer=slack)). This also manifests as a possible [bug](https://sentry.sentry.io/issues/5419971969/events/9f65f1ff493a47858d91c2f9951e59a7/?referrer=issue_details.related_trace_issue) in the serializer, though that's more of a red herring since ClickHouse won't be able to handle the query either way. 

To work around this, I'm applying the  pattern used in the `PostgresSnubaQueryExecutor` to determine whether to pre- or post-filter the results based on the number of group_ids in the filter. If the number of groups in the pre-filter exceeds the `max_candidates`, we'll apply the filter after the Snuba query in a post-filtering step right before the pagination happens.

The tests confirm that the `is:linked` and `is:unlinked` queries work as a pre- and post-filter dependent on the value configured in `snuba.search.max-pre-snuba-candidates`.

### DIFF
--- a/src/sentry/search/snuba/executors.py
+++ b/src/sentry/search/snuba/executors.py
@@ -1665,13 +1665,9 @@ class GroupAttributesPostgresSnubaQueryExecutor(PostgresSnubaQueryExecutor):
             group_ids_to_pass_to_snuba = list(
                 group_queryset.using_replica().values_list("id", flat=True)[: max_candidates + 1]
             )
-        if (
-            group_ids_to_pass_to_snuba is not None
-            and len(group_ids_to_pass_to_snuba) > max_candidates
-        ):
-            metrics.incr("snuba.search.too_many_candidates", skip_internal=False)
-            too_many_candidates = True
-            group_ids_to_pass_to_snuba = []
+            if too_many_candidates := (len(group_ids_to_pass_to_snuba) > max_candidates):
+                metrics.incr("snuba.search.too_many_candidates", skip_internal=False)
+                group_ids_to_pass_to_snuba = None
 
         # remove the search filters that are only for postgres
         search_filters = [
@@ -1706,7 +1702,7 @@ class GroupAttributesPostgresSnubaQueryExecutor(PostgresSnubaQueryExecutor):
 
             having = []
             # if we need to prefetch from postgres, we add filter by the group ids
-            if group_ids_to_pass_to_snuba is not None and not too_many_candidates:
+            if group_ids_to_pass_to_snuba is not None:
                 # will not find any matches, we can return early
                 if len(group_ids_to_pass_to_snuba) == 0:
                     return self.empty_result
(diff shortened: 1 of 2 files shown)


src/sentry/search/snuba/executors.py
```diff
@@ -1692,9 +1703,10 @@ def query(
                 Condition(Column("timestamp", joined_entity), Op.GTE, start),
                 Condition(Column("timestamp", joined_entity), Op.LT, end),
             ]
+
             having = []
             # if we need to prefetch from postgres, we add filter by the group ids
-            if group_ids_to_pass_to_snuba is not None:
+            if group_ids_to_pass_to_snuba is not None and not too_many_candidates:
```
vartec, line 1709: going back to the comment above, if you reset group_ids_to_pass_to_snuba to None when too_many_candidates, not need to change anything here

src/sentry/search/snuba/executors.py
```diff
@@ -1838,6 +1850,15 @@ def query(
                 count += bulk_result[k]["data"][0]["count"]
             k += 1
 
+        if too_many_candidates and group_ids_to_pass_to_snuba == []:
```
vartec, line 1853:
```suggestion
        if too_many_candidates:
```

No need to check group_ids_to_pass_to_snuba, when too_many_candidates it's not relevant.

src/sentry/search/snuba/executors.py
```diff
@@ -1659,8 +1659,19 @@ def query(
...
+            group_ids_to_pass_to_snuba = list(
+                group_queryset.using_replica().values_list("id", flat=True)[: max_candidates + 1]
+            )
+        if (
+            group_ids_to_pass_to_snuba is not None
+            and len(group_ids_to_pass_to_snuba) > max_candidates
+        ):
+            metrics.incr("snuba.search.too_many_candidates", skip_internal=False)
+            too_many_candidates = True
+            group_ids_to_pass_to_snuba = []
```
vartec, line 1674:
```suggestion
            if (too_many_candidates := (len(group_ids_to_pass_to_snuba) > max_candidates)):
              metrics.incr("snuba.search.too_many_candidates", skip_internal=False)
              group_ids_to_pass_to_snuba = None
```

You can keep this within same if that starts in line 1663, because that's the only way you'll have any candidates. 

Reseting to None when there are too many candidates will simplify some checks below.

vartec (approved):
LGTM, I've added some suggestion to simplify the logic a bit.

PR #74532, opened 2024-07-18
fix(slack): fix bad revert

cathteng:
Fix the part of the revert that broke CI on master


iamrajjoshi (approved):

wedamija (approved):
I ran the tests from this file and they pass now, +1

vartec (approved):

PR #74534, opened 2024-07-18
ref(multiple-checkbox): Make MultipleCheckbox props types generic

malwilley:
Prior to this change, you would get `string | number` in the `onChange` handler despite passing in `string[]`. Now the types match what you put in.


ryan953 (approved):
seems legit to me

PR #74535, opened 2024-07-18
feat(uptime): Add name and owner columns to the `ProjectUptimeSubscription` table

wedamija:
We want to treat uptime monitors like other alert rules. To do so, we need a name and the ability to assign owners, so adding these columns in preparation for that.


src/sentry/uptime/migrations/0006_projectuptimesubscription_name_owner.py
```diff
@@ -0,0 +1,66 @@
...
+        ("sentry", "0742_backfill_alertrule_detection_type"),
+        ("uptime", "0005_uptime_status"),
+    ]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
+            database_operations=[
+                migrations.RunSQL(
+                    """
+                    ALTER TABLE "uptime_projectuptimesubscription" ADD COLUMN "name" integer NOT NULL;
```
wedamija, line 36: There are no rows in this table anywhere in prod, so I'm comfortable not providing a default here

markstory, line 36: In the state operations you have the column as a TextField() but the DDL is an integer. Should those types be aligned?

wedamija: Starting to think we should submit a patch to the zerodowntime library so we don't need to do this manually!

markstory: That's a great idea! We have to do this quite a bit.

markstory (approved):

PR #74536, opened 2024-07-18
fix(tests): fixing broken test on main

vartec:
seems to be caused by https://github.com/getsentry/sentry/commit/14676fd593b30b2e46be071533d90f55c20fbde5


cathteng:
i already fixed this here https://github.com/getsentry/sentry/pull/74532

PR #74537, opened 2024-07-18
ref(useRAF): add an enabled option to useRAF so it can be turned off at times.

ryan953:


static/app/utils/replays/hooks/useRAF.tsx
```diff
@@ -1,9 +1,16 @@
 import {useEffect} from 'react';
 
 // TODO: move into app/utils/*
-export default function useRAF(callback: () => unknown) {
+export default function useRAF(
+  callback: () => unknown,
+  opts?: {enabled: undefined | null | boolean}
```
michellewzhang, line 6: why both `undefined` and `null`?

michellewzhang (approved):

static/app/utils/replays/hooks/useRAF.tsx
ryan953: In my use-case i was planning to write `{enabled: replayer && playerState === 'playing'}` in which case `replayer` could be `null`, so `undefined` seemed like a good addition too. 

but now i realize i can write `{enabled: replayer ? playerState === 'playing' : false}` and set the types here to be `boolean` only, and that's fine with me too. seems better.

PR #74540, opened 2024-07-18
ref(replay): refactor `replayerStepper` to be called less

michellewzhang:
- relates to https://github.com/getsentry/team-replay/issues/450
- TLDR: big experimental refactor of `replayerStepper` and all the functions relating to it. following up with a slightly better implementation (https://github.com/getsentry/sentry/pull/74606), but this PR gives context behind it all.

**current state of things**:
we have distinct `replayerStepper` calls in the breadcrumbs + memory tabs. in total, we could be calling it 3 times because it's used by
(1) the breadcrumbs tab to render the HTML snippet, 
(2) for hydration error text diffs,
(3) and the DOM nodes chart.

**the plan**:
improve things by calling `replayerStepper` once (or at least less) and memoize the results. side note: it was hard to combine the hydration error text diff `replayerStepper` call with the others due to the types, so we had to keep that call separate. 

**the question**:
is it faster to run one `replayerStepper` instance, and collect all the data at one time (iterate over the frames exactly once)? this means the speed of any tab will always be limited by the slower call, since we're going through ALL the data at once, even some that we might not need yet. Like this:
https://github.com/getsentry/sentry/blob/70c90544f39dcc3d87dea8c85c0a7669c9585403/static/app/utils/replays/replayReader.tsx#L363-L366

or is it better to continue to have 2 stepper instances that iterate over 2 different sets of frames (breadcrumbs loops over hundreds of frames, and DOM node count iterates over thousands, which means the speeds could be drastically different). in this situation, each tab would handle their own array of necessary frames.

**this PR**:
this PR tried to explore a refactor where `replayerStepper` is called once for the two DOM node actions (counting, for the memory chart, and extracting, for the breadcrumbs HTML). 

**results**:
in this PR, changes were made to the stepper so it could accept a list of callbacks, and return all types of data in one shot, so we only need one hidden `replayer` instance on the page for the DOM node functions. unscientifically, it seemed like loading breadcrumbs + memory took the same amount of time as loading the memory chart. (see video below -- as soon as the breadcrumbs tab is done loading, the memory tab is already done).

https://github.com/user-attachments/assets/2c882bcb-c8df-409c-8e2a-69d75c279735

this makes sense because we’re iterating over thousands of frames for the DOM nodes chart, so that’s the bottleneck. this means the breadcrumbs tab loads slower.

compared to that is the approach where we have 1 stepper instance for breadcrumbs, and one for DOM node count, each iterating over their own lists (will make a separate PR for that one shortly: https://github.com/getsentry/sentry/pull/74606). this approach showed breadcrumb data on the screen about 2x as fast as the approach above. therefore 2 stepper instances on the screen is better for users, especially since breadcrumbs tab is more popular than the memory tab.


static/app/views/replays/detail/memoryPanel/index.tsx
```diff
@@ -6,26 +6,13 @@ import Placeholder from 'sentry/components/placeholder';
...
-import countDomNodes from 'sentry/utils/replays/countDomNodes';
+import useCountDomNodes, {
+  type DomNodeChartDatapoint,
+} from 'sentry/utils/replays/hooks/useCountDomNodes';
 import useCurrentHoverTime from 'sentry/utils/replays/playback/providers/useCurrentHoverTime';
-import type ReplayReader from 'sentry/utils/replays/replayReader';
 import DomNodesChart from 'sentry/views/replays/detail/memoryPanel/domNodesChart';
 import MemoryChart from 'sentry/views/replays/detail/memoryPanel/memoryChart';
 
-function useCountDomNodes({replay}: {replay: null | ReplayReader}) {
```
michellewzhang, line 16: moved to  its own file,`useCountDomNodes.tsx`

static/app/utils/replays/hooks/useExtractedDomNodes.tsx
michellewzhang, line 1: renamed to `useExtractDomNodes.tsx`

static/app/utils/replays/hooks/useExtractedPageHtml.tsx
michellewzhang, line 1: renamed to `useExtractPageHtml.tsx`

static/app/utils/replays/hooks/useExtractPageHtml.tsx
michellewzhang, line 1: same code (no changes) as `useExtractedPageHtml.tsx`, just renamed

static/app/utils/replays/extractPageHtml.tsx
michellewzhang, line 1: moved to `useExtractPageHtml.tsx`

static/app/utils/replays/extractHtml.tsx
```diff
@@ -1,62 +1,14 @@
...
-      const nodeId = getNodeId(frame);
-      const html = extractHtml(nodeId as number, mirror);
-      collection.set(frame as ReplayFrame, {
-        frame,
-        html,
-        timestamp: frame.timestampMs,
-      });
-    },
-  });
-}
```
michellewzhang, line 57: moved to `replayReader`

static/app/utils/replays/countDomNodes.tsx
```diff
@@ -1,54 +0,0 @@
...
-        added,
-        removed,
-        timestampMs: frame.timestamp,
-        startTimestampMs: frame.timestamp,
-        endTimestampMs: frame.timestamp,
-      });
-      prevIds = ids;
-    },
-  });
-}
```
michellewzhang, line 54: moved to `replayReader`

static/app/utils/replays/countDomNodes.tsx
michellewzhang, line 1: moved to `replayReader`

static/app/utils/replays/extractPageHtml.tsx
michellewzhang, line 1: moved to `useExtractPageHtml.tsx`

static/app/utils/replays/hooks/useExtractedDomNodes.tsx
michellewzhang, line 1: renamed to `useExtractDomNodes.tsx`

static/app/utils/replays/hooks/useExtractedPageHtml.tsx
michellewzhang, line 1: renamed to `useExtractPageHtml.tsx`

PR #74542, opened 2024-07-19
fix(metrics-extraction): Alert details

### LINKED ISSUE #74419: Alert details fixes
- Open in Metrics leads to Deleted Metric
- There's nowhere mentioned that this is a `browser.name` filtered down to `browser.name:Chrome OR browser.name:Firefox`

https://sentry.sentry.io/alerts/rules/details/262469/

ArthurKnaus:
Map extracted metrics to virtual ones when reading the URL params on the metrics page.
Display the filter condition on alert details.
Screenshot 2024-07-19 at 07 37 14

Closes https://github.com/getsentry/sentry/issues/74419

### DIFF
--- a/static/app/utils/metrics/extractionRules.tsx
+++ b/static/app/utils/metrics/extractionRules.tsx
@@ -1,4 +1,10 @@
-import type {MetricAggregation, MetricType} from 'sentry/types/metrics';
+import type {
+  MetricAggregation,
+  MetricsExtractionCondition,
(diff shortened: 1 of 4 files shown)


matejminar (approved):
Looks good to me.
I wonder if, for completeness, we should show the query also here, or no need because we have it in the inputs lower on the page. What do you think?
CleanShot 2024-07-19 at 11 04 54

PR #74551, opened 2024-07-19
feat(metrics-summary-table): Dynamic aggreagate columns

### LINKED ISSUE #74543: Improve summary table
Users seem to still be confused about "value", despite adding the tooltip.

We can try a more obvious approach where:
- we remove the column when the selected aggregate is already avg/min/max/sum
- displaying it with the correct label when it's not (count_unique, p50, p99, etc) 
- display multiple columns if there are multiple queries with different aggregations selected ("-" for the query where you don’t have it selected)

ArthurKnaus:
Add missing p90 aggregate in type and definitions.
Display additional aggregate columns for percentiles.

https://github.com/user-attachments/assets/03ba318f-3fac-4814-8f8a-307a366e896f

Closes https://github.com/getsentry/sentry/issues/74543

### DIFF
--- a/static/app/types/metrics.tsx
+++ b/static/app/types/metrics.tsx
@@ -10,6 +10,7 @@ export type MetricAggregation =
   | 'min'
   | 'p50'
   | 'p75'
+  | 'p90'
   | 'p95'
   | 'p99';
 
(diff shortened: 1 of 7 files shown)


vgrozdanic (approved):

vgrozdanic:
Arthur is on vacation for the next two weeks, if he doesn't merge this today (Fri July 19th), I can take over and merge this next week

PR #74555, opened 2024-07-19
feat: Ignore Keyboard layout files from iOS IO in the main thread

### LINKED ISSUE #62080: File IO on Main Thread is triggered whenever user types in Textfield using iPhone default Keyboard
### Discussed in https://github.com/getsentry/sentry-cocoa/discussions/3412

<div type='discussions-op-text'>

<sup>Originally posted by **sumit11007** November 15, 2023</sup>
Sentry is logging lot of Most slow frames when ever user is typing something in Textfield using iPhone default keyboard.
To Verify I have also added a default Textfield and when I try to type File IO is logged in Sentry.
"File IO on Main Thread
UITextSelectionInteraction._handleMultiTapGesture"
image
</div>

brustolin:
Ignores IO in the main thread for a specific case that the user has no control over.

Close https://github.com/getsentry/sentry/issues/62080


.vscode/settings.json
```diff
@@ -22,7 +22,7 @@
   },
 
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
brustolin, line 25:
```suggestion
    "source.fixAll.eslint": "true"
```

.vscode/settings.json
```diff
@@ -60,8 +60,8 @@
   "[python]": {
     "editor.formatOnSave": true,
     "editor.codeActionsOnSave": {
-      "source.organizeImports": true,
-      "source.fixAll.eslint": false
+      "source.organizeImports": "explicit",
+      "source.fixAll.eslint": "never"
```
brustolin, line 64:
```suggestion
      "source.organizeImports": true,
      "source.fixAll.eslint": false
```

.vscode/settings.json
```diff
@@ -22,7 +22,7 @@
   },
 
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "true"
```
brustolin, line 25:
```suggestion
    "source.fixAll.eslint": true
```

src/sentry/utils/performance_issues/detectors/io_main_thread_detector.py
```diff
@@ -193,8 +193,8 @@ def _is_io_on_main_thread(self, span: Span) -> bool:
         data = span.get("data", {})
         if data is None:
             return False
-        _, fileext = os.path.splitext(data.get("file.path", ""))
-        if fileext in self.IGNORED_EXTENSIONS:
+        file_path = data.get("file.path", "")
```
brustolin, line 196:
```suggestion
        file_path = data.get("file.path", "").lower()
```

src/sentry/utils/performance_issues/detectors/io_main_thread_detector.py
```diff
@@ -117,7 +117,7 @@ class FileIOMainThreadDetector(BaseIOMainThreadDetector):
 
     __slots__ = ("stored_problems",)
 
-    IGNORED_EXTENSIONS = {".nib", ".plist"}
+    IGNORED_LIST = {".nib", ".plist", "KBLayout_iPhone.dat"}
```
brustolin, line 120:
```suggestion
    IGNORED_LIST = {".nib", ".plist", "kblayout_iphone.dat"}
```

src/sentry/utils/performance_issues/detectors/io_main_thread_detector.py
```diff
@@ -193,8 +192,8 @@ def _is_io_on_main_thread(self, span: Span) -> bool:
         data = span.get("data", {})
         if data is None:
             return False
-        _, fileext = os.path.splitext(data.get("file.path", ""))
-        if fileext in self.IGNORED_EXTENSIONS:
+        file_path = data.get("file.path", "").lower()
+        if any(file_path.endswith(ext) for ext in self.IGNORED_LIST):
```
narsaynorath, line 196: If we're doing full filenames instead of extensions, can we do something more strict than `endswith`? I looked in the codebase and found a util called `glob_match` in `src/sentry/utils/glob.py`. Can we use change the `IGNORED_LIST` to a glob format (e.g. `*.nib`, `*.plist`, `kblayout_iphone.dat`) and use the util like:
```
if any(glob_match(file_path, ignored_pattern, ignore_case=True))
```
and potentially any other options that would be relevant here? There are some tests in `tests/sentry/utils/test_glob.py` that describes it behaviour more

narsaynorath (commented):
Changes look good to me but I had a suggestion on possibly being more strict on the pattern matching here.
Can we also add a test in `tests/sentry/utils/performance_issues/test_file_io_on_main_thread_detector.py`? There's a test for similarly ignoring `.nib` files

narsaynorath (approved):

PR #74556, opened 2024-07-19
feat(dashboards): Use inferred split logic in events endpoint

narsaynorath:
Supports inferring the widget type based off the query for `events` endpoints which should cut down on rate limits. I'll put up a similar PR for events-stats.


narsaynorath:
Drafting this to fix up some tests.

src/sentry/api/endpoints/organization_events.py
```diff
@@ -43,6 +43,10 @@
     DiscoverSavedQueryTypes.TRANSACTION_LIKE: get_dataset("discover"),
     DiscoverSavedQueryTypes.ERROR_EVENTS: get_dataset("errors"),
 }
+DASHBOARDS_DATASET_MAP = {
+    DashboardWidgetTypes.TRANSACTION_LIKE: get_dataset("metricsEnhanced"),
```
shruthilayaj, line 47: okay so this is fine for now because the codepath that currently runs this is scoped to metricsEnhanced dataset but I realized we don't attempt to split the dataset for this codepath 
https://github.com/getsentry/sentry/blob/master/src/sentry/api/endpoints/organization_events.py#L590-L591

this would have to hit the `transactions` dataset instead of `metricsEnhanced`

shruthilayaj (approved):

narsaynorath:
Drafting until we fix up the errors dataset for use here

PR #74557, opened 2024-07-19
feat(utils): Add stub for rate-limit-based `CircuitBreaker` class

lobsterkatie:
This adds a new `CircuitBreaker` class, which will eventually contain the methods necessary for a rate-limit-based circuit breaker implementation. Our current circuit breaker implementation has four drawbacks that this is aiming to solve:

- The standalone `circuit_breaker_activated` function checks how many errors have been tallied, but doesn't have an accompanying way to actually do the tallying, forcing a reimplementation of the error tracking piece in each scenario in which it's used.
- The `with_circuit_breaker` wrapper fixes that, but only works if the errors are allowed to bubble from the (potentially fairly low-level) spot where we're actually making the request all the way up to the (potentially much higher-level) spot where we're deciding whether or not to make the request in the first place. Handle the errors gracefully anywhere below the `with_circuit_breaker` call and the breaker has no idea they happened, negating the entire purpose of keeping track.
- By catching errors with a simple `try-except`, `with_circuit_breaker` a) counts every kind of exception as an errored request, when that may or may not be accurate, and b) can only track instances of actual `Exception`s, not any other problem (like a 500 response) which may occur.
- Both methods reset the breaker from broken to closed the moment a single request goes through. That works well if the issue is a service having a complete outage, but not if the service is instead overloaded such that it's timing out on, say, 95% of requests. In that case, a single request succeeding isn't in fact a good indicator that the service is working and/or has been fixed.

To solve these problems, the new implementation:

- Will include a way to track errors, so that particular wheel doesn't have to be reinvented over and over (solving problem 1) _and_ so that it can be called selectively and in non-`Exception` spots (solving both halves of problem 3).
- Will do the tracking in a separate method from the one checking if requests should be made, so those two functions can happen in different levels of the call stack (solving problem 2).
- Will be based on a rate limiter rather than an immediately-resetting tally, so errors don't have to be consecutive to trip the breaker (solving be initial-outage-detection part of problem 4) and so that we don't  conclude that an overloaded service has caught back up when it hasn't (solving the service-restoration-detection part of problem 4).

This PR adds the class (along with a `CircuitBreakerConfig` class), documentation of how to use it, a constructor which sets up the appropriate rate limits, and tests of said constructor. Future PRs will add other helper methods and add and test the core error-tracking and request-gating methods.

The initial use of the circuit breaker will be for Seer-based grouping during ingest. Once we're happy with how it's working, we can go back and refactor the places which use the current circuit breaker to use this new one.

### DIFF
--- a/src/sentry/utils/circuit_breaker2.py
+++ b/src/sentry/utils/circuit_breaker2.py
@@ -9,7 +9,18 @@ import logging
 from enum import Enum
 from typing import NotRequired, TypedDict
 
+from django.conf import settings
+
+from sentry.ratelimits.sliding_windows import Quota, RedisSlidingWindowRateLimiter
+
 logger = logging.getLogger(__name__)
+
+# XXX: If either of these values changes, the `CircuitBreakerConfig` docstrings
+# need to be updated
+# How many times stricter to be with the error limit during recovery
+DEFAULT_RECOVERY_STRICTNESS = 10
+# How many times the length of the error window to make the recovery window
+DEFAULT_RECOVERY_WINDOW_MULTIPLIER = 2
 
 
 class CircuitBreakerState(Enum):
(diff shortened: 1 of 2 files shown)


src/sentry/utils/circuit_breaker2.py
```diff
@@ -0,0 +1,185 @@
...
+            logger.warning(
+                "Circuit breaker '%s' has BROKEN and RECOVERY state durations (%d and %d sec, respectively)"
+                + " which together are less than the main error limit window (%d sec). This can lead to the"
+                + " breaker getting tripped unexpectedly, until the original spike in errors clears the"
+                + " main time window.",
+                key,
+                self.broken_state_duration,
+                self.recovery_duration,
+                self.window,
+            )
```
vartec, line 185: Rather than just a warning, IMO `if settings.DEBUG == True` (so in dev and CI) this should raise an exception

vartec (approved):
some nitpicks, but LGTM!

src/sentry/utils/circuit_breaker2.py
```diff
@@ -0,0 +1,185 @@
...
+# values, we'd rather have the circuit breaker not engage when maybe it should (we also have
+# killswitches, rate limiters, feature flags, and the like upon which we can rely to shut things
+# down) than have it engage prematurely.
+#
+# TODO: Consider making this percentage-based rather than having an absolute limit
+CIRCUIT_BREAKER_DEFAULT_CONFIG: CircuitBreakerConfig = {
+    "error_limit": 11211231,
+    "error_limit_window": 3600,  # 1 hr
+    "broken_state_duration": 300,  # 5 min
+}
```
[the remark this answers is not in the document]
lobsterkatie: My theory was that it's both. (See [here](https://github.com/getsentry/sentry/pull/74563/files#diff-c4bf5bc2eae9efe5e948b5deda98c34fafd95674ae376aaf790c21a039e59f91R895-R909) where the defaults for Seer similarity are set.) My thinking was that if the config were a dataclass rather than a typeddict, I'd've included default values, and so this basically doing that. I guess there's no reason we couldn't have the default-defaults as well as the service-specific defaults in options, though. I'll make the change.

src/sentry/utils/circuit_breaker2.py
lobsterkatie: Oh, yeah - I can do that. Good idea!

src/sentry/utils/circuit_breaker2.py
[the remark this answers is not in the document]
lobsterkatie: UPDATE: In the end I decided that the best option was not to have a default at all, so that if you're going to use the circuit breaker, you have to at least think about what numbers you want.

src/sentry/utils/circuit_breaker2.py
lobsterkatie: UPDATE: Done.

PR #72729, opened 2024-06-13
chore(seer grouping): Add `seer.similarity.supported_platforms` option

lobsterkatie:
This changes our list of Seer similarity supported platforms to be configurable via the options automator. As we canonically support new platforms, we should still add them to the default value here (rather than adding them in the automator), but this will let us selectively enable or disable platforms on the fly as we're testing things out.

PR #74563, opened 2024-07-19
ref(seer grouping): Use new `CircuitBreaker` class for circuit breaking

lobsterkatie:
The current implementation of circuit breaking for Seer similarity requests, using `with_circuit_breaker`, doesn't work because any errors the breaker might track are caught much farther down the call stack and never make it to the breaker. This fixes that by switching from using `with_circuit_breaker` to using an instance of the new `CircuitBreaker` class, which allows us to track the errors at a low level, before they're caught, while still using the breaker to decide whether to make the request at a much higher level.

### DIFF
--- a/src/sentry/event_manager.py
+++ b/src/sentry/event_manager.py
@@ -1528,6 +1528,13 @@ def _save_aggregate(
                 seer_matched_group = None
 
                 if should_call_seer_for_grouping(event, primary_hashes):
+                    metrics.incr(
+                        "grouping.similarity.did_call_seer",
+                        # TODO: Consider lowering this (in all the spots this metric is
+                        # collected) once we roll Seer grouping out more widely
+                        sample_rate=1.0,
+                        tags={"call_made": True, "blocker": "none"},
+                    )
                     try:
                         # If the `projects:similarity-embeddings-grouping` feature is disabled,
                         # we'll still get back result metadata, but `seer_matched_group` will be None
@@ -1543,21 +1550,8 @@ def _save_aggregate(
                                 "seer_similarity"
                             ] = seer_response_data
 
-                        metrics.incr(
-                            "grouping.similarity.did_call_seer",
-                            # TODO: Consider lowering this (in all the spots this metric is
-                            # collected) once we roll Seer grouping out more widely
-                            sample_rate=1.0,
-                            tags={"call_made": True, "blocker": "none"},
-                        )
-
                     # Insurance - in theory we shouldn't ever land here
                     except Exception as e:
-                        metrics.incr(
-                            "grouping.similarity.did_call_seer",
-                            sample_rate=1.0,
-                            tags={"call_made": True, "blocker": "none"},
-                        )
                         sentry_sdk.capture_exception(
                             e, tags={"event": event.event_id, "project": project.id}
                         )
--- a/src/sentry/event_manager.py
+++ b/src/sentry/event_manager.py
@@ -129,9 +129,7 @@ from sentry.utils.cache import cache_key_for_event
 from sentry.utils.circuit_breaker import (
     ERROR_COUNT_CACHE_KEY,
     CircuitBreakerPassthrough,
-    CircuitBreakerTripped,
     circuit_breaker_activated,
-    with_circuit_breaker,
 )
 from sentry.utils.dates import to_datetime
 from sentry.utils.event import has_event_minified_stack_trace, has_stacktrace, is_handled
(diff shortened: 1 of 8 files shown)


src/sentry/event_manager.py
```diff
@@ -1555,18 +1551,6 @@ def _save_aggregate(
                             tags={"call_made": True, "blocker": "none"},
                         )
 
-                    except CircuitBreakerTripped:
```
armenzg, line 1558: Is this change so the error can go higher up the call stack?

src/sentry/grouping/ingest/seer.py
```diff
@@ -147,6 +154,32 @@ def _ratelimiting_enabled(event: Event, project: Project) -> bool:
     return False
 
 
+def _circuit_breaker_broken(event: Event, project: Project) -> bool:
+    circuit_broken = not seer_similarity_circuit_breaker.should_allow_request()
```
armenzg, line 158: For some reason, GitHub was not helping me get to the source code for this.

Here's a link in case it happens to anyone else after me:
https://github.com/getsentry/sentry/blob/e88dfb903ad1428c5d44e2b20aef172562222c0d/src/sentry/utils/circuit_breaker2.py#L269-L281

src/sentry/grouping/ingest/seer.py
```diff
@@ -147,6 +154,32 @@ def _ratelimiting_enabled(event: Event, project: Project) -> bool:
...
+
+    if circuit_broken:
+        logger.warning(
+            "should_call_seer_for_grouping.circuit_breaker_tripped",
+            extra={
+                "event_id": event.event_id,
+                "project_id": project.id,
+                **options.get("seer.similarity.circuit-breaker-config"),
```
armenzg, line 166: Good idea 👍🏻

src/sentry/grouping/ingest/seer.py
```diff
@@ -147,6 +154,32 @@ def _ratelimiting_enabled(event: Event, project: Project) -> bool:
...
+            },
+        )
+        metrics.incr(
+            "grouping.similarity.circuit_breaker_tripped",
+        )
+        metrics.incr(
+            "grouping.similarity.did_call_seer",
```
armenzg, line 173: Do you want to increment this metric outside of this if block? That way you can also count when it does get called.

src/sentry/grouping/ingest/seer.py
```diff
@@ -147,6 +154,32 @@ def _ratelimiting_enabled(event: Event, project: Project) -> bool:
...
+            sample_rate=1.0,
+            tags={"call_made": False, "blocker": "circuit-breaker"},
```
armenzg, line 175: Suggestion: You could use a centralized constant to keep track of all the reasons that could block

e.g.
```
class SEER_BLOCKING_REASONS(Enum):
  CIRCUIT_BREAKER = "circuit-breaker"
  ...
```

src/sentry/grouping/ingest/seer.py
```diff
@@ -147,6 +154,32 @@ def _ratelimiting_enabled(event: Event, project: Project) -> bool:
...
+        )
+
+        return True
+
+    return False
```
armenzg, line 180: Nit: you could do `return circuit_broken`

armenzg (approved):

src/sentry/event_manager.py
lobsterkatie: No, it's because that's part of the old circuit breaker apparatus, and this PR switches us to using the new one, which doesn't throw that error.

src/sentry/grouping/ingest/seer.py
lobsterkatie: No - calling the metric with that combo of tags is specific to this case. The affirmative case gets logged in `event_manager`, once we've already passed the `should_call_seer` test. But your question is making me realize that the reason I had to do it that way - because passing `should_call_seer` didn't guarantee the call was made, since the circuit breaker check was outside of `should_call_seer` - no longer applies, which means I can simplify the affirmative case, and move all calls to the metric into this file.

src/sentry/grouping/ingest/seer.py
lobsterkatie: I could, it's true. It'd have to be imported from `utils`, since different versions of the metric are recorded there, too. 

I'm not sure there's a huge advantage there if I'm only using each one once, though, is there?

src/sentry/grouping/ingest/seer.py
lobsterkatie: Yeah, you’re right - unlike the other similar functions this is patterned off of, there's only one `return True` off-ramp. Can do.

src/sentry/seer/similarity/similar_issues.py
```diff
@@ -30,6 +36,11 @@
     timeout=settings.SEER_GROUPING_TIMEOUT,
 )
 
+seer_similarity_circuit_breaker = CircuitBreaker(
+    SEER_SIMILARITY_CIRCUIT_BREAKER_KEY,
+    options.get("seer.similarity.circuit-breaker-config"),
```
asottile-sentry, line 41: an options call at the module is only reflected at import time (and is always the default because the database connection isn't set up)

this is breaking mypy locally because the database isn't set up

lobsterkatie: > and is always the default because the database connection isn't set up

Oh, very good to know. Thanks for catching that! I'll switch it to initializing on the fly.

asottile-sentry: ah yeah -- thought it would've stuck from [the last time](https://github.com/getsentry/sentry/pull/72729#discussion_r1639858737)

lobsterkatie: Ah, missed that - thanks for pointing it out!

PR #74570, opened 2024-07-19
ref(querybuilder): Remove ParamsType from tasks

wmak:
- This removes the usage of ParamsType from tasks in favour of SnubaParams
- Added stats_period as an option to SnubaParams as a way to define start and end
- Added a filter_params property on SnubaParams to make this switchover easy so I'm impacting less code (eg. don't have to change the params references in `format_top_events_timeseries_results` even though `params={}` now)

### DIFF
--- a/src/sentry/search/events/builder/discover.py
+++ b/src/sentry/search/events/builder/discover.py
@@ -294,6 +294,7 @@ class TopEventsQueryBuilder(TimeseriesQueryBuilder):
         super().__init__(
             dataset,
             params,
+            snuba_params=snuba_params,
             interval=interval,
             query=query,
             selected_columns=list(set(selected_columns + timeseries_functions)),
--- a/src/sentry/search/events/builder/discover.py
+++ b/src/sentry/search/events/builder/discover.py
@@ -32,6 +32,7 @@ from sentry.search.events.types import (
     ParamsType,
     QueryBuilderConfig,
     SelectType,
+    SnubaParams,
     WhereType,
 )
 from sentry.snuba.dataset import Dataset
@@ -159,6 +160,7 @@ class TimeseriesQueryBuilder(UnresolvedQuery):
         dataset: Dataset,
         params: ParamsType,
         interval: int,
+        snuba_params: SnubaParams | None = None,
         query: str | None = None,
         selected_columns: list[str] | None = None,
         equations: list[str] | None = None,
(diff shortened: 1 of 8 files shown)


src/sentry/tasks/statistical_detectors.py
```diff
@@ -924,16 +924,14 @@ def query_functions_timeseries(
...
-        "start": end - timedelta(days=14),
-        "end": end,
-        "project_id": project_ids,
-        "project_objects": projects,
-    }
+    params = SnubaParams(
+        start=end - timedelta(days=14),
+        end=end,
+        projects=projects,
+    )
```
wmak, line 934: Didn't use `SnubaParams(stats_period=` here since the existing functionality sets end to 1hr in the future

src/sentry/search/events/builder/discover.py
```diff
@@ -275,6 +278,7 @@ def __init__(
         params: ParamsType,
         interval: int,
         top_events: list[dict[str, Any]],
+        snuba_params: SnubaParams | None = None,
```
Zylphrex, line 281: Should this be passed to `super().__init__`?

src/sentry/search/events/types.py
```diff
@@ -74,24 +75,34 @@ class EventsResponse(TypedDict):
...
+        if self.start is None and self.end is None:
+            self.parse_stats_period()
 
         # Only used in the trend query builder
         self.aliases: dict[str, Alias] | None = {}
 
+    def parse_stats_period(self) -> None:
+        if self.stats_period is not None:
+            self.end = django_timezone.now()
+            from sentry.api.utils import get_datetime_from_stats_period
```
Zylphrex, line 102: Does this import have to be in here?

src/sentry/search/events/types.py
```diff
@@ -74,24 +75,34 @@ class EventsResponse(TypedDict):
...
+
+            self.start = get_datetime_from_stats_period(self.stats_period, self.end)
```
Zylphrex, line 104: Nit: Thoughts on using `get_date_range_from_stats_period` that doesnt require you to set `self.end` here?

src/sentry/search/events/builder/discover.py
wmak: woops good catch

src/sentry/search/events/types.py
wmak: 🤔 those require passing a params object no? I rather not make this more circular than it already is

Zylphrex (approved):

PR #74575, opened 2024-07-19
fix(query-builder): Fix filter value being reset on rerenders

malwilley:
A few changes to prevent the filter value from resetting when the page rerenders:

- Modifies a `useEffect` which had `token` as a dependency to use `token.text` instead
- Wrapped inputs to the component in `useMemo` and `useCallback`

### DIFF
--- a/static/app/views/issueList/utils/useFetchIssueTags.tsx
+++ b/static/app/views/issueList/utils/useFetchIssueTags.tsx
@@ -1,3 +1,5 @@
+import {useMemo} from 'react';
+
 import {useFetchOrganizationTags} from 'sentry/actionCreators/tags';
 import {ItemType, type SearchGroup} from 'sentry/components/smartSearchBar/types';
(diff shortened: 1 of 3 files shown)


static/app/views/issueList/utils/useFetchIssueTags.tsx
```diff
@@ -73,67 +75,71 @@ export const useFetchIssueTags = ({
     {}
   );
 
-  const userTeams = teams.filter(team => team.isMember).map(team => `#${team.slug}`);
-  const usernames: string[] = members.map(getUsername);
-  const nonMemberTeams = teams
-    .filter(team => !team.isMember)
-    .map(team => `#${team.slug}`);
+  const allTags = useMemo(() => {
```
malwilley, line 78: Large diff here but I've just wrapped all the lines below this in a `useMemo`

roggenkemper (approved):

PR #74576, opened 2024-07-19
test(processing): Add test for nonexistent project from buffer

schew2381:
Closes [ALRT-162](https://getsentry.atlassian.net/browse/ALRT-162)

[ALRT-162]: https://getsentry.atlassian.net/browse/ALRT-162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -432,7 +432,16 @@ def apply_delayed(project_id: int, *args: Any, **kwargs: Any) -> None:
     Grab rules, groups, and events from the Redis buffer, evaluate the "slow" conditions in a bulk snuba query, and fire them if they pass
     """
     # STEP 1: Fetch the rulegroup_to_event_data mapping for the project from redis
-    project = Project.objects.get_from_cache(id=project_id)
+    try:
+        project = Project.objects.get_from_cache(id=project_id)
+    except Project.DoesNotExist:
+        # The TTL of the buffer is 1 hr so the rule_group to event data for the
+        # nonexistent project will eventually be cleaned up.
+        logger.info(
+            "delayed_processing.project_does_not_exist",
+            extra={"project_id": project_id},
+        )
+        return
     rulegroup_to_event_data = buffer.backend.get_hash(
         model=Project, field={"project_id": project.id}
     )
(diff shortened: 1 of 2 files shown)


nhsiehgit (approved):

mifu67 (approved):
lgtm

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -432,7 +432,16 @@ def apply_delayed(project_id: int, *args: Any, **kwargs: Any) -> None:
     Grab rules, groups, and events from the Redis buffer, evaluate the "slow" conditions in a bulk snuba query, and fire them if they pass
     """
     # STEP 1: Fetch the rulegroup_to_event_data mapping for the project from redis
-    project = Project.objects.get_from_cache(id=project_id)
+    try:
+        project = Project.objects.get_from_cache(id=project_id)
+    except Project.DoesNotExist:
+        # The TTL of the buffer is 1 hr so the rule_group to event data for the
+        # nonexistent project will eventually be cleaned up.
+        logger.info(
```
saponifi3d, line 440: i think a metric here might be more useful, then tagged with the project_id and "reason: project_does_not_exist"

then we could have a single metric `delayed_process.error` with those other attributes to slice the data on.

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -432,7 +432,16 @@ def apply_delayed(project_id: int, *args: Any, **kwargs: Any) -> None:
     Grab rules, groups, and events from the Redis buffer, evaluate the "slow" conditions in a bulk snuba query, and fire them if they pass
     """
     # STEP 1: Fetch the rulegroup_to_event_data mapping for the project from redis
-    project = Project.objects.get_from_cache(id=project_id)
+    try:
+        project = Project.objects.get_from_cache(id=project_id)
+    except Project.DoesNotExist:
```
saponifi3d, line 437: does `get_from_cache` update the cache as well? (looking at the code, it looks like it's only updating an in memory `local_cache` and a django cache `cache`).

should we manually eject the project id from the buffer instead of relying on the TTL In redis? (my fear is that we're going to be creating 60 * projects logs per hour, until these are cleaned up in the TTL)

schew2381: I didn't realize this but we do actually delete the `project_id` after the task for that project is fired off
https://github.com/getsentry/sentry/blob/af5beb5dfc533a91485311721a7a2b5f0fbf4b3e/src/sentry/rules/processing/delayed_processing.py#L417

As for caching, if the project isn't found it doesn't update the cache. Is that what you mean by update the cache?

PR #74577, opened 2024-07-19
fix(dashboards): Revert querying dataset directly

narsaynorath:
Uses discover and the event.type filter to choose the correct dataset. This is because there may be bugs with the querying the errors dataset directly.

Reverts https://github.com/getsentry/sentry/pull/74077/ but keeps some of the tests that are still compatible.


shruthilayaj (approved):

shruthilayaj:
thanks!

PR #74578, opened 2024-07-19
Delayed Processing Unit Tests::get_group_to_groupevent

saponifi3d:
# Description
- ~Fixes the testutils return type to be an `Event` - this then correctly breaks other typing issues.~ will do this separately as it broke lots of stuff.
- Fix the GroupEvent -> Event typing
- Add tests to ensure the `get_group_to_groupevent` works as expected in more scenarios.

### DIFF
--- a/tests/sentry/rules/processing/test_delayed_processing.py
+++ b/tests/sentry/rules/processing/test_delayed_processing.py
@@ -313,18 +313,17 @@ class GetGroupToGroupEventTest(CreateEventTestCase):
                 **self.occurrence2,
             },
         }
+        self.group_ids = {self.group1.id, self.group2.id}
 
     def test_basic(self):
-        group_ids = {self.group1.id, self.group2.id}
         self.parsed_data.pop((self.rule.id, self.group2.id))
-        result = get_group_to_groupevent(self.parsed_data, self.project.id, group_ids)
+        result = get_group_to_groupevent(self.parsed_data, self.project.id, {self.group1.id})
 
         assert len(result) == 1
         assert result[self.group1] == self.event1
 
     def test_many(self):
-        group_ids = {self.group1.id, self.group2.id}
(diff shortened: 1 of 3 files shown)


tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -94,8 +96,12 @@ def create_event(
         if tags:
             data["tags"] = tags
 
-        return self.store_event(
-            data=data, project_id=project_id, assert_no_errors=False, event_type=EventType.ERROR
+        # Cast the event to GroupEvent to avoid type errors
```
nhsiehgit, line 99: 😬 
what type does `store_event` typically produce?

are there any concerns with casting this?

nhsiehgit (approved):
small nit question but LGTM

tests/sentry/rules/processing/test_delayed_processing.py
saponifi3d: `store_event` returns an `Event` type. i went down the rabbit hole of correctly typing that function, but it created hundreds / thousands of type issues 😭 then... i created this helper to return a group event (which broke lots of things) instead of just casting it, so then i updated this function to return a tuple of the event / group event -- which worked, but it was kinda a nightmare to look at in code, and i updated everywhere outside of this function to be a GroupEvent (but that made my actual changes really hard to grok). 😵 

finally i landed on just wrapping this with the cast, so it will be explicit that it's _not_ actually a group event (i spent a long time trying to figure out what was wrong with my test instead because the return type was just set, and it was implicitly casting it.. so at least now it's explicit.

schew2381: It returns the `Event` type. The reason why we're casting is if you don't specify the `GroupEvent`, you have to assert event.group before you can fetch the value b/c it's typed as `Group | None`. We check that everywhere so there'd be a million asserts all over the place

PR #74586, opened 2024-07-19
ref(similarity): Add project option to enable ingestion after backfill

jangjodi:
Add enable ingestion flag to backfill endpoint
Add enable ingestion flag in backfill
If the flag is enabled, add `sentry:similarity_backfill_completed` to project options (to be checked in ingestion)

### DIFF
--- a/src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
+++ b/src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
@@ -33,7 +33,7 @@ class ProjectBackfillSimilarIssuesEmbeddingsRecords(ProjectEndpoint):
 
         last_processed_id = None
         only_delete = False
-        enable_ingestion = True
+        enable_ingestion = False
 
         if request.data.get("last_processed_id"):
             last_processed_id = int(request.data["last_processed_id"])
--- a/src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
+++ b/src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
@@ -33,15 +33,21 @@ class ProjectBackfillSimilarIssuesEmbeddingsRecords(ProjectEndpoint):
 
         last_processed_id = None
         only_delete = False
+        enable_ingestion = True
+
         if request.data.get("last_processed_id"):
             last_processed_id = int(request.data["last_processed_id"])
 
         if request.data.get("only_delete"):
             only_delete = True
 
+        if request.data.get("enable_ingestion"):
+            enable_ingestion = request.data["enable_ingestion"] == "true"
+
         backfill_seer_grouping_records_for_project.delay(
             current_project_id=project.id,
             last_processed_group_id_input=last_processed_id,
             only_delete=only_delete,
+            enable_ingestion=enable_ingestion,
         )
         return Response(status=204)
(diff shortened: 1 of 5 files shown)


src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -174,6 +176,13 @@ def get_current_batch_groups_from_postgres(project, last_processed_group_id, bat
             "backfill_seer_grouping_records.no_more_groups",
             extra={"project_id": project.id},
         )
+        if enable_ingestion:
+            logger.info(
+                "backfill_seer_grouping_records.enable_ingestion",
+                extra={"project_id": project.id},
+            )
+            project.update_option("sentry:similarity_backfill_completed", int(time.time()))
+
```
lobsterkatie, line 185: Shouldn't we wait to set this option until we know the backfill hasn't errored out?

tests/sentry/tasks/test_backfill_seer_grouping_records.py
```diff
@@ -1534,3 +1539,55 @@ def test_backfill_seer_grouping_records_killswitch_enabled(self, mock_logger):
...
         )
+
+    @with_feature("projects:similarity-embeddings-backfill")
+    @patch("sentry.tasks.embeddings_grouping.utils.logger")
+    @patch("sentry.tasks.embeddings_grouping.utils.post_bulk_grouping_records")
+    def test_backfill_seer_grouping_records_enable_ingestion(
+        self, mock_post_bulk_grouping_records, mock_logger
+    ):
+        """
+        Test that the metadata is set for all groups showing that the record has been created.
```
lobsterkatie, line 1550: Isn't the point of this test (and the next one) to show what happens with the project option, not with the metadata?

src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
```diff
@@ -33,15 +33,21 @@ def post(self, request: Request, project) -> Response:
 
         last_processed_id = None
         only_delete = False
+        enable_ingestion = True
```
trillville, line 36: is there a good reason to default this to `True`? May be less risky to default false

trillville (approved):

src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
jangjodi: I figured we'd typically want to have it be true, but makes sense either way

src/sentry/tasks/embeddings_grouping/utils.py
jangjodi: once we reach this point, we've completed backfilling for that project. we could error out later, but that project has been completed

tests/sentry/tasks/test_backfill_seer_grouping_records.py
jangjodi: haha thanks, copy and paste error. I'll fix it

src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
trillville: yeah unless it greatly simplifies things id prefer we default to false so we can make it more explicit whether or not its happening when we do the backfills

trillville: (worst case for example we mean to do a test backfill of a random language and brick their ingestion bc we forget to pass the flag)

src/sentry/tasks/embeddings_grouping/utils.py
lobsterkatie: Ah, I see. I missed the `if total_groups_to_backfill_length == 0` which was hidden in the fold.

lobsterkatie (approved):

PR #74593, opened 2024-07-19
chore(api): Publish team member endpoint

schew2381:
Discovered a bug with using `NotRequired` for nested TypedDicts (where a field is a TypedDict itself). It marks the nested TypedDict field as required when it's not, kinda interesting.

Also doing this again makes me realize there are def some improvements, like marking fields to exclude from TypedDict fields so we don't have to show them in the examples (such as team/org role configs which are hundreds of lines long and only used by FE). They would still show up in the response, but we can at least not annoy the customer with them on our API page.

BTW the image has a typo when it says ` ... member's with pending ...`, I fixed it but too lazy to regenerate image
Screenshot 2024-07-19 at 1 21 12 PM

Closes [ALRT-161](https://getsentry.atlassian.net/browse/ALRT-161)

[ALRT-161]: https://getsentry.atlassian.net/browse/ALRT-161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/sentry/api/endpoints/team_members.py
```diff
@@ -37,21 +48,43 @@ def get_attrs(self, item_list, user, **kwargs):
...
+            403: RESPONSE_FORBIDDEN,
+            404: RESPONSE_NOT_FOUND,
+        },
+        examples=TeamExamples.LIST_TEAM_MEMBERS,
+    )
     def get(self, request: Request, team) -> Response:
+        """
+        List all members on a team.
+
+        The response will not include member's with pending invites.
```
sentaur-athena: nit: members

sentaur-athena (approved):
The nested dict thing can happen in other endpoints. Is it from our swagger code? If so can you please create a ticket to fix it at some point?

sentaur-athena:
@narsaynorath here's an example with user serializer.

src/sentry/api/endpoints/team_members.py
```diff
@@ -8,10 +9,24 @@
...
+from sentry.apidocs.parameters import CursorQueryParam, GlobalParams
+from sentry.apidocs.utils import inline_sentry_response_serializer
 from sentry.models.organizationmember import InviteStatus
 from sentry.models.organizationmemberteam import OrganizationMemberTeam
 
 
+class OrganizationMemberOnTeamResponse(OrganizationMemberResponse):
+    # We override users to be required b/c team members must already have a user
+    # account to be part of a team.
+    user: UserSerializerResponse
```
schew2381: Narrow typing from `OrganizationMemberResponse`

schew2381:
> The nested dict thing can happen in other endpoints. Is it from our swagger code? If so can you please create a ticket to fix it at some point?

No I'm pretty sure it's from drf-spectacular directly b/c we don't generate the required part on our end, it's given to us in post-processing hook where we can then play around with it

PR #74595, opened 2024-07-19
implement `not in` issue alert filter

nhsiehgit:
corollary to https://github.com/getsentry/sentry/pull/74517

### DIFF
--- a/src/sentry/api/endpoints/project_rules_configuration.py
+++ b/src/sentry/api/endpoints/project_rules_configuration.py
@@ -101,7 +101,7 @@ class ProjectRulesConfigurationEndpoint(ProjectEndpoint):
                     copied_context["formFields"]["match"]["choices"] = [
                         choice
                         for choice in match_choices
-                        if (choice[0] != MatchType.IS_IN or choice[0] != MatchType.NOT_IN)
+                        if (choice[0] != MatchType.IS_IN and choice[0] != MatchType.NOT_IN)
                     ]
                     filter_list.append(copied_context)
                     continue
--- a/src/sentry/api/endpoints/project_rules_configuration.py
+++ b/src/sentry/api/endpoints/project_rules_configuration.py
@@ -99,7 +99,9 @@ class ProjectRulesConfigurationEndpoint(ProjectEndpoint):
                     match_choices = context["formFields"]["match"]["choices"]
                     copied_context = copy.deepcopy(context)
                     copied_context["formFields"]["match"]["choices"] = [
-                        choice for choice in match_choices if choice[0] != MatchType.IS_IN
+                        choice
+                        for choice in match_choices
+                        if (choice[0] != MatchType.IS_IN or choice[0] != MatchType.NOT_IN)
                     ]
                     filter_list.append(copied_context)
                     continue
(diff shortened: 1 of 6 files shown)


tests/sentry/rules/conditions/test_event_attribute.py
```diff
@@ -856,3 +856,15 @@ def test_attr_is_in(self):
             data={"match": MatchType.IS_IN, "attribute": "platform", "value": "python"}
         )
         self.assertDoesNotPass(rule, event)
+
+    def test_attr_not_in(self):
```
saponifi3d, line 860: i'd make sure to have a positive and negative test case for these.

tests/sentry/rules/conditions/test_event_attribute.py
```diff
@@ -856,3 +856,15 @@ def test_attr_is_in(self):
...
+        event = self.get_event()
+        rule = self.get_rule(
+            data={"match": MatchType.NOT_IN, "attribute": "platform", "value": "php, python"}
```
saponifi3d, line 863: thoughts on adding some test cases around different `values` here? thinking like empty string, strings with spaces / w/o etc -- i'd just do it with like a pymark.paramterize too

tests/sentry/rules/conditions/test_event_attribute.py
nhsiehgit: the tests include both the positive and negative test already

tests/sentry/rules/conditions/test_event_attribute.py
nhsiehgit: I don't see that as particularly useful atm.

since it's a text-input, the values always going to be a string.
the different values will all just be unmatching strings 🤷

src/sentry/api/endpoints/project_rules_configuration.py
```diff
@@ -99,7 +99,9 @@ def get(self, request: Request, project) -> Response:
                     match_choices = context["formFields"]["match"]["choices"]
                     copied_context = copy.deepcopy(context)
                     copied_context["formFields"]["match"]["choices"] = [
-                        choice for choice in match_choices if choice[0] != MatchType.IS_IN
+                        choice
+                        for choice in match_choices
+                        if (choice[0] != MatchType.IS_IN or choice[0] != MatchType.NOT_IN)
```
schew2381, line 104: Why do we need to strip `is_in` and `not_is_in`? (like it says on line 98)

schew2381 (commented):
I know there's some conflict resolution we have between conditions, although I don't know exactly where the logic for that is. It'd be cool if we could catch duplicate tags btwn `in` and `not in` somehow

src/sentry/api/endpoints/project_rules_configuration.py
nhsiehgit: Oh oops - let me update the note

this is now stripping both of them

tests/sentry/rules/conditions/test_event_attribute.py
saponifi3d: the value add is really in resiliency in the code and to reduce number of issues we see in production. if there are additional test cases we can reduce the number of issues while you still have strong context over the feature.

tests/sentry/rules/conditions/test_event_attribute.py
saponifi3d: in that case, it might be nice to split them out as separate test cases then to illustrate how this code works in a positive and negative case. it's really meant to help people understand the code and what expected / unexpected cases are handled.

---

in general my theory with this is that if we can improve our tests and readability of code - we can reduce the number of issues and magnitude of issues in the future. since we can't know when / where that will happen i think it's important we're diligent with our testing.

schew2381 (approved):

PR #74598, opened 2024-07-19
feat(dev-toolbar): Add organization derived API applications

cmanallen:
# Overview

To enable the dev-toolbar feature we want to allow public, third-party applications (think "Login with CodeCov" but self-serve and for all organizations) to fetch Sentry data on behalf of a Sentry user.  To accommodate this I've:

- [x] Added a "scoping_organization_id" column to the `ApiToken` model and replica.
- [x] Added a "organization_id" column to the `ApiApplication` model.
- [x] Updated the oauth flow to set the "scoping_organization_id" on the `ApiToken` when an "organization_id" is present on the `ApiApplication`.
- [x] Updated the middleware responsible for gating access. 
- [x] Added a resource for managing organization `ApiApplication` instances.

# Goal

The goal is to allow customers to authenticate with Sentry _from a third-party origin_.

# How It Works

1. A customer defines an [ApiApplication](https://github.com/getsentry/sentry/blob/8c4f17f93dd6d4a94df689d840180dfe491da78e/src/sentry/models/apiapplication.py#L43) instance tied to their organization.
2. This ApiApplication contains the valid origin and redirect locations for an authorization request for that organization.
3. The ApiApplication generates a client-id value.  This value is public and is passed to the Javascript SDK's initializer (e.g. `sdk.init(..., client_id=xyz)`.
4. The SDK reads this client-id and initiates Sentry's oauth flow (`/oauth/authorize/`) either through a hard-redirect or a pop-up window.  Making sure to specify the client-id in the request parameters.
5. Sentry appends a url fragment containing the access token and redierects back to the requester's application.
6. The token is now accessible by the third-party program and can be used to make API requests to Sentry.

## CORS

To manage CORS issues, the login flow will be initiated from an iframe with a sentry.io origin.

# Supporting Documentation

PRD: https://www.notion.so/sentry/FY-25-Q2-Dev-Toolbar-e2a259c063634f93a6c3d89584e812d8

# Security

Public clients present two problems:

1. First there is a phishing risk where a malicious third-party is given an `access_token` and uses that token to extract sensitive data from Sentry servers.
8. Second there is a risk that a legitimate third-party provider may use an `access_token` to read data from other organizations.

To address these issues I've done two things:

1. ApiApplications will only generate credentials if the user is a member of the organization.
2. ApiTokens will be scoped to the ApiApplication's organization_id.
    a. Requests for data outside of the organization will be denied.


tests/sentry/web/frontend/test_oauth_authorize.py
```diff
@@ -346,3 +346,79 @@ def test_minimal_params_code_deny_flow(self):
...
+            owner=self.user,
+            redirect_uris="https://example.com",
+            organization_id=self.organization.id,
+        )
+
+        resp = self.client.get(
+            f"{self.path}?client_id={self.application.client_id}&response_type=token"
+            "&redirect_uri=https://example.com"
+        )
+        assert resp.status_code == 200
```
mdtro, line 368: Shouldn't this fail without a `state` parameter?

mdtro:
I'd be more comfortable if we put this behind a feature flag limiting it to specific organizations until we're sure the authn/authz flow is ready for production.

src/sentry/web/frontend/oauth_authorize.py
mdtro, line 1: I'm not entirely sure where this should go. We need to use a tighter expiration time on these tokens. By default, they'll be 30 days. No more than 12 hours seems acceptable and would help mitigate risk.

I _think_ the `self.approve` call can support an `expires_in` arg?

cmanallen: Yes it can 👍

cmanallen:
@mdtro They're supposed to be able to revoke the token with a logout request.  I'll have to find exactly where that happens.

tests/sentry/web/frontend/test_oauth_authorize.py
cmanallen: I believe the answer is yes to prevent CSRF.  But it wasn't required prior to my being here.

>    The client MUST implement CSRF protection for its redirection URI.
   This is typically accomplished by requiring any request sent to the
   redirection URI endpoint to include a value that binds the request to
   the user-agent's authenticated state (e.g., a hash of the session
   cookie used to authenticate the user-agent).  The client SHOULD
   utilize the "state" request parameter to deliver this value to the
   authorization server when making an authorization request.

cmanallen:
@evanpurkhiser @nhsiehgit @JoshFerge Ya'll worked on the enterprise team right?  Do you have any opinions on this auth flow?

wedamija (commented):
Migration looks ok, but since this is a large pr might splitting out the model changes?

static/app/data/controlsiloUrlPatterns.ts
ryan953, line 1: bunch of stuff got re-generated that's unrelated to the oauth code. we should send that along in it's own PR

PR #74601, opened 2024-07-19
ref(tags): Use new useApiQuery hooks to retrieve tag values using new dataset param

MichaelSun48:
This PR creates the new `useFetchOrganizationTagKeyValues` hook using `useApiQuery`. The reason for these changes are: 
1.  Replace the old `fetchTagValues` function in favor of a `useApiQuery` implementation
2. Utilize the new `Dataset` parameter to make faster and more accurate requests for tag values.

### DIFF
--- a/static/app/views/issueDetails/groupEvents.tsx
+++ b/static/app/views/issueDetails/groupEvents.tsx
@@ -4,7 +4,6 @@ import styled from '@emotion/styled';
 import orderBy from 'lodash/orderBy';
 
 import {useFetchIssueTags} from 'sentry/actionCreators/group';
-import {fetchTagValues} from 'sentry/actionCreators/tags';
 import EventSearchBar from 'sentry/components/events/searchBar';
 import * as Layout from 'sentry/components/layouts/thirds';
 import {SearchQueryBuilder} from 'sentry/components/searchQueryBuilder';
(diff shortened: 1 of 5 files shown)


static/app/views/issueList/utils/getIssueTagValues.tsx
```diff
@@ -1,29 +1,94 @@
...
+    return await Promise.resolve(
+      allTagValues.map(({value}) => {
+        // Truncate results to 5000 characters to avoid exceeding the max url query length
+        // The message attribute for example can be 8192 characters.
+        if (typeof value === 'string' && value.length > 5000) {
+          return value.substring(0, 5000);
+        }
+        return value;
+      })
+    );
```
MichaelSun48, line 92: This `await Promise.resolve(...)` is pretty janky, but is a workaround to force the return type of this function to be `Promise<string[]>`, which is what both the new and old search components expect in their `getTagValue` prop. In the future when the old search component is completely phased out, we should refactor this and get rid of this jankyness, but I don't think it'll cause much harm until then.

static/app/views/issueDetails/groupEvents.tsx
```diff
@@ -100,25 +99,9 @@ function UpdatedSearchBar({
...
-      });
-    },
-    [api, group.project.id, organization.slug]
-  );
-
   const getTagValues = useMemo(
-    () => makeGetIssueTagValues(tagValueLoader),
-    [tagValueLoader]
+    () => makeGetIssueTagKeyValues({organization, pageFilters}),
+    [organization, pageFilters]
```
MichaelSun48, line 104: Updated issue details with the same logic since it also uses the `makeGetIssueTagValues` function. I have assumed with this change that issue details also pulls tags from `events` and `issuePlatform`, lmk if that is incorrect

PR #74606, opened 2024-07-19
ref(replay): refactor replayerStepper calls to be memoized

michellewzhang:
- closes https://github.com/getsentry/team-replay/issues/450
- this is the branch off from https://github.com/getsentry/sentry/pull/74540, the experimental refactor of `replayerStepper`
- TLDR: memoize the `replayerStepper` calls so that we remember the results and don't have to reload the data every time.

**current state of things**:
we have distinct `replayerStepper` calls in the breadcrumbs + memory tabs. in total, we could be calling it 3 times because it's used by
(1) the breadcrumbs tab to render the HTML snippet, 
(2) for hydration error text diffs,
(3) and the DOM nodes chart.

**the plan**:
improve things by calling `replayerStepper` once (or at least less) and memoize the results. side note: it was hard to combine the hydration error text diff `replayerStepper` call with the others due to the types, so we had to keep that call separate. 

**the question**:
is it faster to run one `replayerStepper` instance, and collect all the data at one time (iterate over the frames exactly once)? this means the speed of any tab will always be limited by the slower call, since we're going through ALL the data at once, even some that we might not need yet. like this:
https://github.com/getsentry/sentry/blob/70c90544f39dcc3d87dea8c85c0a7669c9585403/static/app/utils/replays/replayReader.tsx#L363-L366

or is it better to continue to have 2 stepper instances that iterate over 2 different sets of frames (breadcrumbs loops over hundreds of frames, and DOM node count iterates over thousands, which means the speeds could be drastically different). in this situation, each tab would handle their own array of necessary frames.

**the experiments**:
the first PR (https://github.com/getsentry/sentry/pull/74540) tried to explore a refactor where `replayerStepper` is called once for the two DOM node actions (counting, for the memory chart, and extracting, for the breadcrumbs HTML). changes were made to the stepper so it could accept a list of callbacks, and return all types of data in one shot, so we only need one hidden `replayer` instance on the page for the DOM node functions.

unscientifically, it seemed like loading breadcrumbs + memory took the same amount of time as loading the memory chart. (see video below -- as soon as the breadcrumbs tab is done loading, the memory tab is already done).

https://github.com/user-attachments/assets/2c882bcb-c8df-409c-8e2a-69d75c279735

this makes sense because we’re iterating over thousands of frames for the DOM nodes chart, so that’s the bottleneck. this means the breadcrumbs tab loads slower.

**this PR**:
compared to that is the approach where we have 1 stepper instance for breadcrumbs, and one for DOM node count, each iterating over their own lists (which is what i'm doing in this PR). this approach showed breadcrumb data on the screen about 2x as fast as the approach above. therefore 2 stepper instances on the screen is better for users, especially since breadcrumbs tab is more popular than the memory tab.

the video below demonstrates the memoization in action. once the breadcrumbs or memory tabs are loaded, switching back to them does not cause re-loading, because the results are cached. notice that the breadcrumbs tab loading is not dependent on the loading of the DOM tab, unlike the video above where the breadcrumbs tab has to "wait" for the memory tab. (i'm also on slower wifi in this clip below)

https://github.com/user-attachments/assets/be71add1-63b5-4308-9d26-356d544980ef


static/app/views/replays/detail/memoryPanel/index.tsx
```diff
@@ -6,26 +6,11 @@ import Placeholder from 'sentry/components/placeholder';
...
 import {space} from 'sentry/styles/space';
-import {useQuery} from 'sentry/utils/queryClient';
-import countDomNodes from 'sentry/utils/replays/countDomNodes';
+import useCountDomNodes from 'sentry/utils/replays/hooks/useCountDomNodes';
 import useCurrentHoverTime from 'sentry/utils/replays/playback/providers/useCurrentHoverTime';
-import type ReplayReader from 'sentry/utils/replays/replayReader';
 import DomNodesChart from 'sentry/views/replays/detail/memoryPanel/domNodesChart';
 import MemoryChart from 'sentry/views/replays/detail/memoryPanel/memoryChart';
 
-function useCountDomNodes({replay}: {replay: null | ReplayReader}) {
```
michellewzhang, line 16: moved to its own file, `useCountDomNodes.tsx`

static/app/utils/replays/hooks/useExtractedPageHtml.tsx
michellewzhang, line 1: renamed to `useExtractPageHtml.tsx`

static/app/utils/replays/hooks/useExtractedDomNodes.tsx
michellewzhang, line 1: renamed to `useExtractDomNodes.tsx`

static/app/utils/replays/extractHtml.tsx
```diff
@@ -1,62 +1,14 @@
...
-      const nodeId = getNodeId(frame);
-      const html = extractHtml(nodeId as number, mirror);
-      collection.set(frame as ReplayFrame, {
-        frame,
-        html,
-        timestamp: frame.timestampMs,
-      });
-    },
-  });
-}
```
michellewzhang, line 57: moved to `replayReader.tsx`

static/app/utils/replays/countDomNodes.tsx
michellewzhang, line 1: moved to `replayReader.tsx`

ryan953 (approved):

billyvg (approved):
What a great PR write-up!

billyvg:
It might be nice to have a loading indicator for the dom node chart

michellewzhang:
> It might be nice to have a loading indicator for the dom node chart

@billyvg we had a placeholder before but seems our `isFetching` check no longer works. pushing a fix now! lmk if you have a strong preference for loading indicator vs placeholder

static/app/utils/replays/replayerStepper.tsx
```diff
@@ -62,25 +62,25 @@ export default function replayerStepper<
...
 
     const considerFrame = (frame: Frame) => {
       if (shouldVisitFrame(frame, replayer)) {
-        frameRef.frame = frame;
+        frameRef.current = frame;
         window.setTimeout(() => {
           const timestamp =
             'offsetMs' in frame ? frame.offsetMs : frame.timestamp - startTimestampMs;
           replayer.pause(timestamp);
         }, 0);
```
billyvg, line 75: would RAF be better here @ryan953?

billyvg:
> > It might be nice to have a loading indicator for the dom node chart
> 
> @billyvg we had a placeholder before but seems our `isFetching` check no longer works. pushing a fix now! lmk if you have a strong preference for loading indicator vs placeholder
> 
>  Screen.Recording.2024-07-23.at.12.12.23.PM.mov

Placeholder looks great!

static/app/utils/replays/replayerStepper.tsx
ryan953: we could try it, and see if the memory chart renders the same as before. 

also, we could experiment with different granularity in the memory chart.

ryan953: Here's an answer i found on the differences: https://stackoverflow.com/questions/38709923/why-is-requestanimationframe-better-than-setinterval-or-settimeout

michellewzhang: seems to be the same behavior after using RAF so i pushed up the change

PR #74613, opened 2024-07-19
chore(alerts): Stop sending unused count headers from combined rules endpoint

wedamija:
These headers don't look to be used in our codebase and are undocumented. They're likely contributing to slowness on this endpoint, so just removing.


saponifi3d (approved):
🥇

PR #74614, opened 2024-07-19
feat(alerts): Send historical snuba data to Seer

ceorourke:
When an anomaly detection metric alert is created, query Snuba for 28 days of historical data and send it to Seer.

TODO (as separate follow ups):
[ ] [Add to edited anomaly detection alert](https://getsentry.atlassian.net/browse/ALRT-168)
[ ] [Warn users if less than a week of historical data is available](https://getsentry.atlassian.net/browse/ALRT-169)
[ ] [Support async Slack channel lookup _and_ anomaly detection alert - currently it's one or the other](https://getsentry.atlassian.net/browse/ALRT-170)

Closes https://getsentry.atlassian.net/browse/ALRT-140

### DIFF
--- a/src/sentry/incidents/endpoints/organization_alert_rule_index.py
+++ b/src/sentry/incidents/endpoints/organization_alert_rule_index.py
@@ -123,7 +123,7 @@ class AlertRuleIndexMixin(Endpoint):
             else:
                 alert_rule = serializer.save()
                 if alert_rule.detection_type == AlertRuleDetectionType.DYNAMIC.value:
-                    resp = send_historical_data_to_seer(rule=alert_rule, user=request.user)
+                    resp = send_historical_data_to_seer(alert_rule=alert_rule, user=request.user)
                     if resp.status != 200:
                         alert_rule.delete()
                         return Response({"detail": resp.reason}, status=status.HTTP_400_BAD_REQUEST)
--- a/src/sentry/incidents/endpoints/organization_alert_rule_index.py
+++ b/src/sentry/incidents/endpoints/organization_alert_rule_index.py
@@ -1,4 +1,3 @@
-import logging
 from copy import deepcopy
 from datetime import UTC, datetime
 
@@ -9,7 +8,6 @@ from drf_spectacular.utils import extend_schema, extend_schema_serializer
 from rest_framework import serializers, status
 from rest_framework.request import Request
 from rest_framework.response import Response
-from urllib3.exceptions import MaxRetryError, TimeoutError
 
 from sentry import features
 from sentry.api.api_owners import ApiOwner
@@ -30,7 +28,6 @@ from sentry.apidocs.constants import RESPONSE_FORBIDDEN, RESPONSE_NOT_FOUND, RES
 from sentry.apidocs.examples.metric_alert_examples import MetricAlertExamples
 from sentry.apidocs.parameters import GlobalParams
 from sentry.apidocs.utils import inline_sentry_response_serializer
-from sentry.conf.server import SEER_ANOMALY_DETECTION_URL
 from sentry.constants import ObjectStatus
 from sentry.exceptions import InvalidParams
 from sentry.incidents.endpoints.serializers.alert_rule import (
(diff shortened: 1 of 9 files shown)


src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -0,0 +1,180 @@
...
+            },
+        )
+        return None
+    snuba_query = SnubaQuery.objects.get(id=rule.snuba_query_id)
+    time_period = int(snuba_query.time_window / 60)
+    historical_data = fetch_historical_data(rule, snuba_query)
+    formatted_data = format_historical_data(historical_data)
+
+    if is_data_empty(formatted_data):
+        return BaseHTTPResponse(
```
ceorourke, line 102: I'm using this here (docs: https://urllib3.readthedocs.io/en/latest/reference/urllib3.response.html) because it matches the return type used by `make_signed_seer_api_request` so it's consistent. It requires all these things are passed even though we only care about the `status` and `reason` here.

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -0,0 +1,190 @@
...
+    formatted_data = []
+    for datum in data.data.get("data", []):
+        ts_point = TimeSeriesPoint(timestamp=datum.get("time"), value=datum.get("count", 0))
+        formatted_data.append(ts_point)
+    return formatted_data
+
+
+def translate_direction(direction: int) -> str:
+    """
+    Temporary translation map to Seer's expected values
```
ceorourke, line 49: temp because we can ask the ai/ml team to update what the expect :) but for now we'll just translate

src/sentry/incidents/endpoints/organization_alert_rule_index.py
```diff
@@ -118,8 +119,15 @@ def create_metric_alert(self, request, organization, project=None):
                 }
                 find_channel_id_for_alert_rule.apply_async(kwargs=task_args)
                 return Response({"uuid": client.uuid}, status=202)
+                # TODO handle slack channel lookup AND anomaly detection
             else:
                 alert_rule = serializer.save()
+                if alert_rule.detection_type == AlertRuleDetectionType.DYNAMIC.value:
+                    resp = send_historical_data_to_seer(rule=alert_rule, user=request.user)
+                    if resp.status != 200:
+                        alert_rule.delete()
```
nhsiehgit, line 128: Why delete the alert rule if the call fails???

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -0,0 +1,187 @@
...
+    """
+    Format Snuba data into the format the Seer API expects.
+    If there are no results, it's just the timestamp
+    {'time': 1719012000}, {'time': 1719018000}, {'time': 1719024000}
+
+    If there are results, the count is added
+    {'time': 1721300400, 'count': 2}
+    """
+    formatted_data = []
+    for datum in data.data.get("data", []):
```
nhsiehgit, line 41: `data.data.data` lol

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -0,0 +1,187 @@
...
+            extra={
+                "rule_id": rule.id,
+                "project_id": project_id,
+            },
+        )
+        base_error_response.reason = "No project associated with rule. Cannot create rule."
+        return base_error_response
+
+    snuba_query = SnubaQuery.objects.get(id=rule.snuba_query_id)
+    time_period = int(snuba_query.time_window / 60)
```
nhsiehgit, line 99: is this a period?
or is this just a `window_min`?

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -0,0 +1,187 @@
...
+    formatted_data = format_historical_data(historical_data)
+
+    if not rule.sensitivity or not rule.seasonality or rule.threshold_type is None:
+        # this won't happen because we've already gone through the serializer, but mypy insists
+        base_error_response.reason = (
+            "Cannot create rule - missing expected configuration for a dynamic alert."
+        )
+        return base_error_response
+
+    ad_config = ADConfig(
```
nhsiehgit, line 115: Lets spell out full words rather than abbreviate

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -0,0 +1,187 @@
...
+        time_period=time_period,
```
nhsiehgit, line 116: a time period i expect to have a start/end
this is just a window size right?

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -0,0 +1,187 @@
...
+        sensitivity=rule.sensitivity,
+        direction=translate_direction(rule.threshold_type),
+        expected_seasonality=rule.seasonality,
+    )
+    alert = Alert(id=rule.id)
```
nhsiehgit, line 121: Is this an `AlertRule`? or an `IssueAlert`? 🤔 

Can we name this something more distinct?

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -0,0 +1,187 @@
...
+        if not project:
+            return None
+
+    if project:
+        return project.id
+
+    return None
+
+
+def send_historical_data_to_seer(rule: AlertRule, user: User) -> BaseHTTPResponse:
```
nhsiehgit, line 73: This is doing quite a bit more than just sending historical data to seer right?

I see it's querying snuba for the data too?

Can you add a docstring here to expand on this method?

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -0,0 +1,187 @@
...
+    dataset_label = snuba_query.dataset
+    if dataset_label == "events":
+        # DATSET_OPTIONS expects the name 'errors'
+        dataset_label = "errors"
+    dataset = get_dataset(dataset_label)
+    project_id = get_project_id_from_rule(rule)
+    if not project_id or not dataset:
+        return None
+
+    historical_data = dataset.timeseries_query(
```
nhsiehgit, line 173: This query should match teh subscription query right?

src/sentry/seer/anomaly_detection/types.py
```diff
@@ -0,0 +1,25 @@
+from typing import TypedDict
+
+
+class Alert(TypedDict):
```
nhsiehgit, line 4: re: above

can we name this something more descriptive and less saturated?

src/sentry/seer/anomaly_detection/types.py
```diff
@@ -0,0 +1,25 @@
...
+    id: int
+
+
+class TimeSeriesPoint(TypedDict):
+    timestamp: float
+    value: float
+
+
+class ADConfig(TypedDict):
```
nhsiehgit, line 13: No abbreviations - lets spell out the full thing

src/sentry/seer/anomaly_detection/types.py
```diff
@@ -0,0 +1,25 @@
...
+    time_period: int
```
nhsiehgit, line 14: `window_min`?

nhsiehgit (commented):
few comments

src/sentry/incidents/endpoints/organization_alert_rule_index.py
ceorourke: I discussed the best way to handle this with Josh - if the call fails that means we now have an alert rule that can't fire because we haven't send the historical data needed to make predictions about anomalies to Seer. It's a little awkward but Seer needs to store it along with the rule ID so I can't do it before it's saved.

src/sentry/seer/anomaly_detection/store_data.py
ceorourke: lol take it up with snuba team

src/sentry/seer/anomaly_detection/store_data.py
ceorourke: this was named by the ai/ml team - I can rename the variable but the typeddict should match as closely to the seer data as possible

src/sentry/seer/anomaly_detection/store_data.py
ceorourke: `fetch_historical_data` is doing the actual query, this just calls that function and sends it. I can add a docstring though

src/sentry/seer/anomaly_detection/store_data.py
ceorourke: isn't a window of minutes a time period?

src/sentry/seer/anomaly_detection/store_data.py
ceorourke: yes

src/sentry/seer/anomaly_detection/store_data.py
ceorourke: no, this has to match the data types in Seer. I can ask them to rename it and the update this later

src/sentry/seer/anomaly_detection/store_data.py
ceorourke: yeah it's the same metric and dataset, but we're getting it for 28 days

src/sentry/seer/anomaly_detection/types.py
ceorourke: again no, this matches the data in the Seer repo

src/sentry/incidents/endpoints/organization_alert_rule_index.py
saponifi3d: i think there's basically a data normalization thing going on deeper with seer (it seems like it needs a snuba query id, but it's requiring the alert stuff). buuutttt considering that's likely a lot to refactor and would probably need updates in seer, this seemed like an explicit way to say "if we can't create a seer data set for the alert, we can't create the alert"

nhsiehgit: 👍 ah i see yea, this is the `create_alert_rule` method?

Do we need to `serializer.save()` above then?
Can we check `serializer.detection_type` and construct the query from the serializer before creating the alert rule?

might be better to do this so we can avoid 2 requests to the db?

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+    direction_map = {
+        AlertRuleThresholdType.ABOVE: "up",
+        AlertRuleThresholdType.BELOW: "down",
+        AlertRuleThresholdType.ABOVE_AND_BELOW: "both",
+    }
+    return direction_map[AlertRuleThresholdType(direction)]
+
+
+def get_project_id_from_rule(rule: AlertRule) -> int | None:
```
nhsiehgit, line 64: an AlertRule also has a fk to `Project`'s.

https://github.com/getsentry/sentry/blob/master/src/sentry/incidents/models/alert_rule.py#L280

probably better to go through this join table rather than through subscriptions?

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -0,0 +1,195 @@
...
+        if not project:
+            return None
+
+    if project:
+        return project.id
+
+    return None
+
+
+def send_historical_data_to_seer(rule: AlertRule, user: User) -> BaseHTTPResponse:
```
nhsiehgit, line 78: nit: lets just name the argument `alert_rule`?

If we update the API to ping seer _before_ creating the alert rule though, we can pass the serialized object instead?

src/sentry/incidents/endpoints/organization_alert_rule_index.py
ram-senth: On seer side we do not need Snuba query id. We do need an identifier for the alert so that future anomaly detection calls can retrieve the historic metric data. Currently, we are reusing the db generated id for that. We can avoid the save-delete-if-fail loop by switching to a GUID as GUIDs can be generated before storing in the table. However, it will need some changes in the alerting system:
- store the GUID in the alert table
- retrieve and pass it to seerwhen triggering anomaly detection.

nhsiehgit: Ok hmmm.

actually i think i understand the need now.
we need the alert_rule.id for seer to be able to associate this backfilled data with the alert rule it will be monitoring.

this flow seems fine then 👍

src/sentry/seer/anomaly_detection/store_data.py
mifu67: Should I copy this over to the subscription processor? If we expect that this is a permanent fix, should I import it from this file?

nhsiehgit (approved):
🫡

src/sentry/seer/anomaly_detection/store_data.py
ceorourke: Maybe we should just ask Ram to update the Seer side to expect ints and import the `AlertRuleThresholdType`

PR #74626, opened 2024-07-22
feat(apple-symbols): Improve metrics collection

iambriccardo:
This PR makes the `both` and `old`/`symx` symbols collection mutually exclusive, which gives us the ability to know how many eligible symbols have `symx` or `old` data. Before, a metric was emitted in a way that made the calculation impossible due to double counting.


Swatinem (approved):
nice catch

jernejstrasner (approved):

PR #74631, opened 2024-07-22
chore(replay): Log mobile replay usage

cmanallen:


src/sentry/replays/consumers/recording_buffered.py
```diff
@@ -247,7 +247,19 @@ def process_message(buffer: RecordingBuffer, message: bytes) -> None:
     )
 
     if replay_video := decoded_message.get("replay_video"):
+        # Logging org info for bigquery
+        logger.info(
+            "sentry.replays.slow_click",
```
c298lee, line 252: why are we logging as "sentry.replays.slow_click"?

cmanallen: That's the name of our big query table.  Bad name I know :\

c298lee (approved):

PR #74634, opened 2024-07-22
feat(dashboards): Surface split decision in widget type field

narsaynorath:
Surface the split decision in the widget type field. This is because we don't return the split decision on subsequent requests when we detect that the `discover_widget_split` field is populated, but the frontend still needs to know which dataset to show on subsequent loads.


shruthilayaj (approved):
lgtm

PR #74637, opened 2024-07-22
feat(performance): enable long animation frame support for internal t…

KevinL10:
Enables long animation frame support in the browser SDK for internal testing.

See https://github.com/getsentry/sentry-javascript/issues/11719 for more information.


gggritso (approved):
:shipit:

PR #74640, opened 2024-07-22
ref: fix subscription_id None check

asottile-sentry:
looks like this might've been skipping some of these as well (always passing `None` when deleting)


wedamija (approved):
Thanks, this isn't in production yet but we'd have skipped the deletion

PR #74643, opened 2024-07-22
chore(features) Remove FeatureManager.has_for_batch

markstory:
This method is rarely used and `batch_has` fulfils the same role and is more widely used.

### DIFF
--- a/src/sentry/api/serializers/models/project.py
+++ b/src/sentry/api/serializers/models/project.py
@@ -179,7 +179,7 @@ def get_environments_by_projects(projects: Sequence[Project]) -> MutableMapping[
 def get_features_for_projects(
     all_projects: Sequence[Project], user: User, filter_unused_on_frontend_features: bool = False
 ) -> MutableMapping[Project, list[str]]:
-    # Arrange to call features.has_for_batch rather than features.has
+    # Arrange to call features.batch_has rather than features.has
     # for performance's sake
     projects_by_org = defaultdict(list)
     for project in all_projects:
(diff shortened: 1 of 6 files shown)


src/sentry/api/serializers/models/project.py
```diff
@@ -220,9 +220,8 @@ def get_features_for_projects(
             continue
         abbreviated_feature = feature_name[len(_PROJECT_SCOPE_PREFIX) :]
         for organization, projects in projects_by_org.items():
-            result = features.has_for_batch(feature_name, organization, projects, user)
-            for project, flag in result.items():
-                if flag:
+            if features.has(feature_name, organization, projects, user):
```
GabeVillalobos, line 223: Shouldn't this be `batch_has` instead of `has`? As written this will only return the first project's feature results. Also, I don't think we currently detect this case so we might need to update the flagpole handler to error when multiple projects are passed to a non-batched feature check 🤔

markstory: There is a call to `batch_has()` on line 203. The implementation for `has_for_batch` is odd, as the base class implements a naive loop over the 'batch' which is a collection of projects.

https://github.com/getsentry/sentry/blob/30d0a96140f68f91a17d77503776ab6601187cd9/src/sentry/features/handler.py#L34-L39

There is also a `BatchFeatureHandler` which also implements `has_for_batch` but it only checks features with the organization only and doesn't even pass the projects along 🫠 . Instead it does a single check with the organization and uses that result for all projects. 🙃 

https://github.com/getsentry/sentry/blob/30d0a96140f68f91a17d77503776ab6601187cd9/src/sentry/features/handler.py#L59-L71

While there are several sub-classes of `BatchFeatureHandler` none of them get a list of projects as a parameter.

I'll go through a few more of the project feature handlers in depth, but on a cursory read of them they should be compatible with this change as `BatchFeatureHandler.has` will still behave the same as before, and the single check result will be applied to all projects in the collection.

GabeVillalobos (approved):

src/sentry/api/serializers/models/project.py
GabeVillalobos: Ah I missed that this was the secondary feature check after the initial batch. I'm good with this as long as the behavior is still consistent.

markstory:
After doing some local testing, this change will not be as easy as I had originally thought. The secondary `feature.has` check in the project serializer isn't compatible with the existing `BatchFeatureHandler` implementations in getsentry. Removing `has_for_batch` is going to be more complicated and needs a different solution.

PR #74645, opened 2024-07-22
fix(issues): allow rendering of annotations in both formats

oioki:


malwilley:
@oioki it looks like there is one more test file you'll need to make changes in: `issueFieldRenderers.spec.tsx`. I think the test is actually written wrong - it is sending an array of strings, but really it should be sent as a string with commas in between the items.

malwilley (approved):

PR #74648, opened 2024-07-22
fix(issues): structured issue annotations

oioki:


src/sentry/integrations/mixins/issues.py
```diff
@@ -348,7 +348,7 @@ def map_external_issues_to_annotations(self, external_issues):
         for ei in external_issues:
             link = self.get_issue_url(ei.key)
             label = self.get_issue_display_name(ei) or ei.key
-            annotations.append(f'<a href="{link}">{label}</a>')
+            annotations.append({"url": link, "display_name": label})
```
malwilley, line 351: Already mentioned, but since we changed this to `displayName` in the frontend we need to change it here as well

src/sentry/integrations/mixins/issues.py
```diff
@@ -348,7 +348,7 @@ def map_external_issues_to_annotations(self, external_issues):
         for ei in external_issues:
             link = self.get_issue_url(ei.key)
             label = self.get_issue_display_name(ei) or ei.key
-            annotations.append(f'<a href="{link}">{label}</a>')
+            annotations.append({"url": link, "displayName": label})
```
armenzg, line 351: I see no more usage of `a href` in the APIs 👍🏻

armenzg (approved):
🚀

PR #74649, opened 2024-07-22
feat(replays): Emit mobile replay outcomes

cmanallen:
Emitting replay video outcome for tracking purposes.

Closes: https://github.com/getsentry/team-replay/issues/452
Related: https://github.com/getsentry/relay/pull/3847


michellewzhang (approved):

src/sentry/replays/usecases/ingest/__init__.py
```diff
@@ -176,7 +176,19 @@ def track_initial_segment_event(
         first_replay_received.send_robust(project=project, sender=Project)
 
     # Replay videos are not billed for now.
```
asottile-sentry, line 178: comment is outdated?

cmanallen: No this outcome is not billed.  We're adding it so the PMs can access the data in looker.

asottile-sentry: additionally: should there be a test covering this change?

PR #74650, opened 2024-07-22
feat(onboarding): create messaging integration picker modal

ameliahsu:
Created modal that opens when "Connect to messaging" button is selected (see https://github.com/getsentry/sentry/pull/74474)

Screenshot 2024-07-22 at 3 45 35 PM

Installation buttons have the same behavior as those on the Integrations page in Settings

If the modal content can't be loaded, the modal is closed and a small error message appears in the corner:

https://github.com/user-attachments/assets/621725cd-8ffd-4c91-af0c-8e91483bfcaa


static/app/views/alerts/rules/issue/addIntegrationRow.tsx
```diff
@@ -0,0 +1,116 @@
...
+      .catch(error => {
+        console.error('Error fetching data:', error);
+        setHasError(true);
+      });
+  }, [integrationSlug, api, organization.slug]);
+
+  useEffect(() => {
+    fetchData();
+  }, [fetchData]);
+
```
sentaur-athena, line 50: You can add `console.log("provider", integrations)` here or any other log you like to see why it's not loading. They will show up in the terminal when you run the test.

cathteng:
is it just me or is the `Add Workspace` button just a tiny bit off being vertically centered?

ameliahsu:
@cathteng it's cuz "workspace" and "installation" are diff lengths 😭  the buttons are right justified, I can add a min length to the buttons

static/app/views/alerts/rules/issue/addIntegrationRow.tsx
```diff
@@ -0,0 +1,116 @@
...
+import PluginIcon from 'sentry/plugins/components/pluginIcon';
+import {space} from 'sentry/styles/space';
+import type {IntegrationProvider, Project} from 'sentry/types';
+import type {Organization} from 'sentry/types/organization';
+import useApi from 'sentry/utils/useApi';
+import {AddIntegrationButton} from 'sentry/views/settings/organizationIntegrations/addIntegrationButton';
+
+type Props = {
+  closeModal: () => void;
+  integrationSlug: string;
```
sentaur-athena, line 14: Not 100% sure about this but I don't think integration slug is always the same as provider key and I believe what your code here gets is the provider key. For slack, msteams and discord they seem to be but it's not enforced anywhere in the code to be so I would rename it to avoid confusion.

static/app/views/alerts/rules/issue/addIntegrationRow.tsx
```diff
@@ -0,0 +1,116 @@
...
+    fetchData();
+  }, [fetchData]);
+
+  if (!provider) {
+    return null;
+  }
+
+  const {metadata} = provider;
+  const size = 'sm' as const;
+  const priority = 'primary' as const;
```
sentaur-athena, line 56: Please use them inline just like style since they're used only once.

static/app/views/alerts/rules/issue/addIntegrationRow.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  const buttonProps = {
+    style: {margin: 0},
+    size,
+    priority,
+    'data-test-id': 'install-button',
+    organization,
+  };
+
+  // TODO(Mia): disable button if user does not have necessary permissions
```
sentaur-athena, line 66: Just to make sure we're aligned, if user doesn't have permission we render instead of install button the request button.

static/app/views/alerts/rules/issue/setupAlertIntegrationModal.tsx
```diff
@@ -1,16 +0,0 @@
-import {Fragment} from 'react';
-
-import type {ModalRenderProps} from 'sentry/actionCreators/modal';
-
-function SetupAlertIntegrationModal({Header, Body}: ModalRenderProps) {
```
sentaur-athena, line 5: Would the old slack button open this modal? In this case since it now opens your new modal you need to check feature flag in `MessagingIntegrationModal` because the old button has access to it. Maybe I'm misunderstanding.

ameliahsu: The old slack button still directs to [docs](https://develop.sentry.dev/integrations/slack/), the new button that opens this modal is only returned when the user has the feature flag

static/app/views/alerts/rules/issue/addIntegrationRow.tsx
```diff
@@ -0,0 +1,114 @@
...
+  }, [fetchData]);
+
+  if (!provider) {
+    return null;
+  }
+
+  const {metadata} = provider;
+
+  const buttonProps = {
+    style: {margin: 0},
```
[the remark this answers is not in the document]
ameliahsu: Used inline styles since this Button is an imported Sentry component with other styling

static/app/views/alerts/rules/issue/addIntegrationRow.tsx
```diff
@@ -0,0 +1,114 @@
...
+    size: 'sm' as const,
+    priority: 'primary' as const,
+    'data-test-id': 'install-button',
+    organization,
+  };
+
+  // TODO(Mia): show request installation button if user does not have necessary permissions
+  const integrationButton = metadata.aspects.externalInstall ? (
```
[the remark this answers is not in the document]
ameliahsu: Found similar code in another file, will refactor to create a separate reusable component in another PR

static/app/views/alerts/rules/issue/messagingIntegrationModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+}: Props) {
+  const [hasError, setHasError] = useState(false);
+  const integrationValues = ['slack', 'discord', 'msteams'];
+
+  useEffect(() => {
+    if (hasError) {
+      closeModal();
+      addErrorMessage(t('Failed to load integration data'));
+    }
+  }, [hasError, closeModal]);
```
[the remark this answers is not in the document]
ameliahsu: If the modal content can't be loaded, the modal is closed and a small error message appears in the corner:

https://github.com/user-attachments/assets/621725cd-8ffd-4c91-af0c-8e91483bfcaa

static/app/views/alerts/rules/issue/addIntegrationRow.tsx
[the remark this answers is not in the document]
saponifi3d: You should still be able to use styled components, just need to pass that component into the styled helper like this:
```
styled(Button)`
  margin: 0;
`;
```
That should do the trick (and can do the same thing for the integration button).

mixing inline styles with styled components can make things a little tricky because it mixes css selector priorities with classes / element styles. (there's also implications on what happens during the render cycle in browsers based on that - in the element styles are applied, then the class styles. since this is changing spacing it could cause reflows / layout thrashes)

static/app/views/alerts/rules/issue/messagingIntegrationModal.tsx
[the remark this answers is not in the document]
saponifi3d: hmm, the ux of this feels a little awkward. i wonder if we could block opening the popup until we've confirmed the integration can load instead? that way the UX would just show the toast instead of flashing the popup momentarily.

to do that, maybe have the button click trigger the xhr request, then on success have the modal open. This introduces a new problem of a "slow to open" modal, so maybe changing the staet of `connect to messaging` to a loading spinner while the xhr request is open? (should be able to get all the state of the request from react-query state)

saponifi3d (approved):
i think the code lgtm, the remaining feedback is really just taking it to the next level :)

PR #74652, opened 2024-07-22
feat(uptime): Use organization level autodetection flag to gate whether we should run detection

wedamija:
Respect the org level flag that controls whether to run detection or not.

### DIFF
--- a/src/sentry/uptime/detectors/detector.py
+++ b/src/sentry/uptime/detectors/detector.py
@@ -3,7 +3,11 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from sentry import features
-from sentry.uptime.detectors.ranking import add_base_url_to_rank, should_detect_for_project
+from sentry.uptime.detectors.ranking import (
+    add_base_url_to_rank,
+    should_detect_for_organization,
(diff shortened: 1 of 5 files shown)


davidenwang (approved):
lgtm, for this setting is there a possibility that the following could happen
1. We process events for Org A, and add base urls to redis
2. The org owner turns off the setting, but base urls are still in redis, just not processed
3. we process that org A's project's bucket, and then create an auto-detected uptime monitor

do we need to check again for should_detect_for_organization before creating a monitor in candidate mode?

wedamija:
> lgtm, for this setting is there a possibility that the following could happen
> 
> 1. We process events for Org A, and add base urls to redis
> 2. The org owner turns off the setting, but base urls are still in redis, just not processed
> 3. we process that org A's project's bucket, and then create an auto-detected uptime monitor
> 
> do we need to check again for should_detect_for_organization before creating a monitor in candidate mode?

Yep, we handle that here and skip processing if the key was disabled https://github.com/getsentry/sentry/blob/f42956f4b8e3d5525261d1557e91236b8854f58d/src/sentry/uptime/detectors/tasks.py#L116

PR #74656, opened 2024-07-22
feat(insights): add release markings to fullscreen insight charts

KevinL10:
Adds release markings to the insights charts on fullscreen. This is similar to how we currently display releases on the performance page charts.

image

### DIFF
--- a/static/app/components/modals/insightChartModal.tsx
+++ b/static/app/components/modals/insightChartModal.tsx
@@ -5,7 +5,7 @@ import styled from '@emotion/styled';
 
 import type {ModalRenderProps} from 'sentry/actionCreators/modal';
 import {space} from 'sentry/styles/space';
-import {ChartHeightContext} from 'sentry/views/insights/common/components/chart';
+import {ChartIsFullScreenContext} from 'sentry/views/insights/common/components/chart';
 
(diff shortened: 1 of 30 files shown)


gggritso (commented):
Makes sense! I think it needs a few small changes to simplify and clarify a bit

1. `ReleaseSeries` should be rendered conditionally, only in full-screen mode. Otherwise you're loading release series every time for every chart even if it hasn't been made fullscreen. If you render conditionally, you'll save a network request most of the time _and_ you won't need to update all those specs
2. `ChartHeightContext` is good, but `ChartRenderingContext` would be better. It should hold the height and the _grid_ and in fact all properties that might be affected by full-screen status so ... height and grid. Otherwise  we'll have to rename this context every time we add something to it
3. `chartShowLegend` is redundant. IMO keep it `showLegend` and change the variable inside the component to `isLegendVisible` so you don't have to update all usages of `Chart`
4. Adding an extra series to every chart is a little risky because the number of series and their order matters for charts that have sample point hover states. Could you check that the span sample icons and row still work correctly on hover?

KevinL10:
closing this (too many changes to deal with), replaced by https://github.com/getsentry/sentry/pull/75012.

PR #74657, opened 2024-07-22
feat(toolbar): Add a panel to show active alerts

ryan953:
New panel to reveal active alerts in your sentry instance:

SCR-20240722-kpuw

### DIFF
--- a/static/app/components/devtoolbar/components/alerts/alertsPanel.tsx
+++ b/static/app/components/devtoolbar/components/alerts/alertsPanel.tsx
@@ -152,8 +152,8 @@ function AlertListItem({item}: {item: Incident}) {
           }}
           onClick={() => {
             trackAnalytics?.({
-              eventKey: `devtoolbar.feedback-list.item.click`,
-              eventName: `devtoolbar: Click feedback-list item`,
+              eventKey: `devtoolbar.alert-list.item.click`,
+              eventName: `devtoolbar: Click alert-list item`,
             });
           }}
         >
--- a/static/app/components/devtoolbar/components/alerts/alertsPanel.tsx
+++ b/static/app/components/devtoolbar/components/alerts/alertsPanel.tsx
@@ -1,0 +1,185 @@
+import {css} from '@emotion/react';
+
+import ActorAvatar from 'sentry/components/avatar/actorAvatar';
+import AlertBadge from 'sentry/components/badge/alertBadge';
+import ProjectBadge from 'sentry/components/idBadge/projectBadge';
+import Placeholder from 'sentry/components/placeholder';
+import TextOverflow from 'sentry/components/textOverflow';
+import TimeSince from 'sentry/components/timeSince';
+import type {Actor} from 'sentry/types/core';
+import type {Organization} from 'sentry/types/organization';
+import AlertRuleStatus from 'sentry/views/alerts/list/rules/alertRuleStatus';
+import type {Incident, MetricAlert} from 'sentry/views/alerts/types';
+import {CombinedAlertType} from 'sentry/views/alerts/types';
+import {alertDetailsLink} from 'sentry/views/alerts/utils';
+
+import useConfiguration from '../../hooks/useConfiguration';
+import {
+  badgeWithLabelCss,
+  gridFlexEndCss,
+  listItemGridCss,
+  listItemPlaceholderWrapperCss,
+} from '../../styles/listItem';
+import {panelInsetContentCss, panelSectionCss} from '../../styles/panel';
+import {resetFlexColumnCss, resetFlexRowCss} from '../../styles/reset';
+import {smallCss, xSmallCss} from '../../styles/typography';
+import InfiniteListItems from '../infiniteListItems';
+import InfiniteListState from '../infiniteListState';
+import PanelLayout from '../panelLayout';
+import SentryAppLink from '../sentryAppLink';
+import useTeams from '../teams/useTeams';
+
+import useInfiniteAlertsList from './useInfiniteAlertsList';
+
+export default function AlertsPanel() {
+  const {projectId, projectSlug, trackAnalytics} = useConfiguration();
+  const queryResult = useInfiniteAlertsList();
+
+  const estimateSize = 108;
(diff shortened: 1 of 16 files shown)


static/app/components/devtoolbar/components/teams/useTeams.tsx
```diff
@@ -0,0 +1,30 @@
...
+  idOrSlug?: string;
+}
+
+export default function useTeams({idOrSlug}: Props, opts?: {enabled: boolean}) {
+  const {organizationSlug} = useConfiguration();
+
+  return useFetchApiData<Team[]>({
+    queryKey: useMemo(
+      () => [
+        `/organizations/${organizationSlug}/teams/`,
```
michellewzhang, line 19: can we replace `organizationSlug` here with `idOrSlug`? (do we need both?)

static/app/components/devtoolbar/components/alerts/alertsPanel.tsx
```diff
@@ -0,0 +1,185 @@
...
+import useTeams from '../teams/useTeams';
+
+import useInfiniteAlertsList from './useInfiniteAlertsList';
+
+export default function AlertsPanel() {
+  const {projectId, projectSlug, trackAnalytics} = useConfiguration();
+  const queryResult = useInfiniteAlertsList();
+
+  const estimateSize = 108;
+  const placeholderHeight = `${estimateSize - 8}px`; // The real height of the items, minus the padding-block value
```
michellewzhang, line 39: should we make these constants/export them somewhere? they're used in the feedback & issues panels too

static/app/components/devtoolbar/components/alerts/alertsPanel.tsx
```diff
@@ -0,0 +1,185 @@
...
+                resetFlexColumnCss,
+                panelSectionCss,
+                panelInsetContentCss,
+                listItemPlaceholderWrapperCss,
+              ]}
+            >
+              <Placeholder height={placeholderHeight} />
+              <Placeholder height={placeholderHeight} />
+              <Placeholder height={placeholderHeight} />
+              <Placeholder height={placeholderHeight} />
```
michellewzhang, line 85: can probably reuse this too!

static/app/components/devtoolbar/components/alerts/alertsPanel.tsx
```diff
@@ -0,0 +1,185 @@
...
+        <SentryAppLink
+          to={{
+            url: alertDetailsLink({slug: organizationSlug} as Organization, item),
+            query: {alert: item.identifier},
+          }}
+          onClick={() => {
+            trackAnalytics?.({
+              eventKey: `devtoolbar.feedback-list.item.click`,
+              eventName: `devtoolbar: Click feedback-list item`,
+            });
```
michellewzhang, line 157: need to update analytics name for alerts

static/app/components/devtoolbar/types.ts
```diff
@@ -27,9 +27,12 @@ type QueryKeyEndpointOptions<
   query?: Query;
 };
 
+// Prefix the key with a namespace, to avoid key collisions with other tanstack/query
+// cache requests that imported sentry modules make within the toolbar scope.
 export type ApiQueryKey =
-  | readonly [url: string]
+  | readonly ['io.sentry.toolbar', url: string]
   | readonly [
+      'io.sentry.toolbar',
```
ryan953, line 35: This is annoying, but for now it's the only solution i've got.

What's happening is we're importing components from sentry. Those components can fetch data using `static/app/utils/queryClient.tsx` like usual. The problem with calling `queryClient.tsx` is that it depends on ConfigStore and a bunch of internals. So, long term the toolbar shouldn't depend on those (by default shouldn't), which is why we have `static/app/components/devtoolbar/hooks/useApiEndpoint.tsx` for the toolbar, it's much simpler and skips a bunch of stuff, but should be ok for what we're doing.

The problem is that sentry components (that are imported into the toolbar), and toolbar components must share the same query cache. They could have the same cache-key in there, with different results cached. This is not good.

So what i've done is added a cache-busting namespace string to the start of all new keys that the toolbar uses. Fixed!

static/app/components/devtoolbar/styles/reset.ts
```diff
@@ -15,6 +15,12 @@ export const resetFlexColumnCss = css`
   flex-grow: 1;
 `;
 
+export const resetFlexRowCss = css`
+  display: flex;
+  flex-direction: row;
```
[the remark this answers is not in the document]
ryan953: it's true!

static/app/components/devtoolbar/components/teams/useTeams.tsx
ryan953: `organizationSlug` could be a slug or id, but it's different from `idOrSlug` in this same file.. 

`idOrSlug` is meant to be the teamId or teamSlug. I tried to make the method signature read well: ie `useTeams({idOrSlug: 'replay-team'})` but maybe it's more confusing inside the method because there are multiple slugs available.

static/app/components/devtoolbar/components/alerts/alertsPanel.tsx
ryan953: oops, these should probably be unique values

static/app/components/devtoolbar/components/alerts/alertsPanel.tsx
ryan953: i'll come back and see about some of the copy+paste in a followup. but playing with this `css={}` prop is interesting so far.

static/app/components/devtoolbar/styles/reset.ts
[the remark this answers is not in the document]
ryan953: we need this because i was messing around and added this global:

https://github.com/getsentry/sentry/blob/8cfb85f3b9759e7486020924c6533fb281260184/static/app/components/devtoolbar/styles/global.ts#L57-L60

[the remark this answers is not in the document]
michellewzhang: ahh i see

michellewzhang (approved):

PR #74660, opened 2024-07-22
feat(dashboards): Add split decision source field

narsaynorath:
Adds a dataset_source field to show how the source was set. Will be
useful when showing alerts in the UI for the user when a dataset was
automatically selected for them.

To be handled in another PR: We don't currently force a dataset for
dashboard widgets. Once I get a confirmation on which widgets
should be used as default, then I will update the logic to handle 
ambiguous cases like data from both errors and transactions.


src/sentry/models/dashboard_widget.py
```diff
@@ -76,6 +77,31 @@ class DashboardWidgetTypes(TypesClass):
     TYPE_NAMES = [t[1] for t in TYPES]
 
 
+class DatasetSourcesTypes(Enum):
```
narsaynorath, line 80: I had to copy this from Discover Saved Queries because if I tried to import it, I was getting circular import errors.

narsaynorath:
Closing this to split up the migration + endpoint logic

PR #74665, opened 2024-07-22
feat(dashboards): Save widget type to split decision if explicit

narsaynorath:
If the widget type is explicitly one of our split types (i.e. error-events, transaction-like) then save that split type to the `discover_widget_split` field as well for compatibility with on-demand.

### DIFF
--- a/tests/sentry/api/endpoints/test_organization_dashboards.py
+++ b/tests/sentry/api/endpoints/test_organization_dashboards.py
@@ -861,3 +861,76 @@ class OrganizationDashboardsTest(OrganizationDashboardWidgetTestCase):
         }
         response = self.do_request("post", f"{self.url}?environment=mock_env", data=data)
         assert response.status_code == 201, response.data
+
+    def test_post_dashboard_with_widget_split_datasets(self):
+        mock_project = self.create_project()
+        self.create_environment(project=mock_project, name="mock_env")
+        data = {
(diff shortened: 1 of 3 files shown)


tests/sentry/api/endpoints/test_organization_dashboards.py
```diff
@@ -861,3 +861,76 @@ def test_post_dashboard_with_widget_filter_requiring_environment(self):
         }
         response = self.do_request("post", f"{self.url}?environment=mock_env", data=data)
         assert response.status_code == 201, response.data
+
+    def test_post_dashboard_with_widget_split_datasets(self):
```
shruthilayaj, line 865: should we write a test for `put` as well?

shruthilayaj (approved):

tests/sentry/api/endpoints/test_organization_dashboards.py
narsaynorath: The list endpoint only supports `GET` and `POST`! The `PUT` test is on the detail endpoint where you're working with a specific dashboard instance (in `tests/sentry/api/endpoints/test_organization_dashboard_details.py`), it's basically the same test

narsaynorath:
Going to split this up

narsaynorath:
Closed the wrong PR 🤦

PR #74667, opened 2024-07-22
fix(insights): transaction names with quotes break caches

### LINKED ISSUE #74554: Cannot load Caches page
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

1. the customer tries to open the Caches page in production
2. the endpoint gives the following error: 

```
{
"detail": "Invalid quote at '[\"BackOfficeClient': quotes must enclose text or be escaped."
}
```
3. the page does not load properly

### Expected Result

the customer should be able to load the page

### Actual Result

the page is not loading

### Product Area

Insights

### Link

_No response_

### DSN

_No response_

### Version

_No response_

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2227) by [Unito](https://www.unito.io)

DominikB2014:
fixes #74554 

The above issue was a result in the transaction duration query sending the list of transactions without the quotes escaped .

### DIFF
--- a/static/app/views/insights/cache/views/cacheLandingPage.tsx
+++ b/static/app/views/insights/cache/views/cacheLandingPage.tsx
@@ -134,7 +134,7 @@ export function CacheLandingPage() {
     isFetching: isTransactionDurationFetching,
   } = useMetrics(
     {
-      search: `transaction:[${transactionsList.map(({transaction}) => `"${transaction}"`).join(',')}]`,
+      search: `transaction:[${transactionsList.map(({transaction}) => `"${transaction.replaceAll('"', '\\"')}"`).join(',')}]`,
       fields: [`avg(transaction.duration)`, 'transaction'],
       enabled: !isTransactionsListFetching && transactionsList.length > 0,
     },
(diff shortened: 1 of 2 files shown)


static/app/views/insights/cache/views/cacheLandingPage.tsx
```diff
@@ -134,7 +134,7 @@ export function CacheLandingPage() {
     isFetching: isTransactionDurationFetching,
   } = useMetrics(
     {
-      search: `transaction:[${transactionsList.map(({transaction}) => `"${transaction}"`).join(',')}]`,
+      search: `transaction:[${transactionsList.map(({transaction}) => `"${transaction.replaceAll('"', '\\"')}"`).join(',')}]`,
```
gggritso, line 137: 🤔 `useMetrics` uses `MutableQuery` under the hood, which should escape double quotes. Is that not working?

DominikB2014: We ended up having to bypass the whole `MutableQuery` thing, because it didn't handle having spaces in the transaction name correctly.
see https://github.com/getsentry/sentry/pull/70363#discussion_r1592903273

DominikB2014: `useMetrics` `search` is of type `search?: MutableSearch | string`

gggritso (approved):

PR #74669, opened 2024-07-22
add tests for parse rulegroup to event data

saponifi3d:
## Description
Moar tests for 666, this is for the `parse_rulegroup_to_event_data` method - one more to go after this!

### DIFF
--- a/tests/sentry/rules/processing/test_delayed_processing.py
+++ b/tests/sentry/rules/processing/test_delayed_processing.py
@@ -575,19 +575,17 @@ class ParseRuleGroupToEventDataTest(TestCase):
             "occurrence_id": "1",
         }
 
-        self.input_data = {
+    def test_parse_rulegroup(self):
+        input_data = {
             f"{self.rule.id}:{self.group.id}": json.dumps(self.event_data),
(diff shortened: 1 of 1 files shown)


tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -562,8 +563,46 @@ def test_get_slow_conditions__only_fast(self):
...
+            f"{self.rule_two.id}:{self.group_two.id}": json.dumps(self.event_data),
+        }
+
+        self.expected = {
+            (str(self.rule.id), str(self.group.id)): self.event_data,
+            (str(self.rule_two.id), str(self.group_two.id)): self.event_data,
+        }
+
+    def test_parse_rulegroup(self):
+        result = parse_rulegroup_to_event_data(self.input_data)
```
nhsiehgit, line 589: nit - do we need to add the success case input_data to `self`?
Can we just define those here since we're not using them in other tests?

nhsiehgit (approved):

tests/sentry/rules/processing/test_delayed_processing.py
saponifi3d: 👍 i did it first expecting more cases 🤣

PR #74672, opened 2024-07-22
feat(issue-details): Update group header

roggenkemper:
this pr introduces changes to the group header on issue details. In addition to styling changes, it also combines the first/last seen, participant, assigned, and viewer sections from the sidebar into the header. for now, it hides the elements that it will replace. to limit the scope of this PR, additional functionality changes will be made in future PRs (expanding partipants/viewers, fine tuning first/last seen for example)

Screenshot 2024-07-24 at 2 08 53 PM


static/app/views/issueDetails/updatedHeader.tsx
```diff
@@ -0,0 +1,367 @@
...
+  display: flex;
+  color: ${p => p.theme.gray300};
+  gap: ${space(1)};
+`;
+
+const InfoWrapper = styled('div')<{isResolved: boolean}>`
+  padding-top: ${space(1)};
+  padding-bottom: ${space(1)};
+  display: flex;
+  justify-content: space-between;
```
leeandher, line 339: We'll also need a gap to ensure they don't collide:
`gap: ${space(1)}` will match the button bar if you like

image

leeandher (commented):
Left probably too many comments, feel free to chat offline if you need clarification, looks 🔥 btw, nice job!

static/app/components/events/eventMessage.tsx
```diff
@@ -47,14 +48,17 @@ function EventMessage({
...
+  hideLevel = false,
 }: Props) {
   return (
     <LevelMessageContainer className={className}>
-      <EventOrGroupLevel
-        level={level}
-        levelIndicatorSize={levelIndicatorSize}
-        type={type}
-      />
+      {!hideLevel && (
```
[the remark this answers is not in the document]
roggenkemper: good catch  - will not pass in level for now, will deal with an updated design for level in a different PR (since it also likely will affect the issue stream)

static/app/views/issueDetails/groupSidebar.tsx
```diff
@@ -301,8 +308,8 @@ export default function GroupSidebar({
       {issueTypeConfig.regression.enabled && event && (
         <EventThroughput event={event} group={group} />
       )}
-      {renderParticipantData()}
-      {renderSeenByList()}
+      {!hasStreamlinedUI && renderParticipantData()}
+      {!hasStreamlinedUI && renderSeenByList()}
```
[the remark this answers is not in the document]
roggenkemper: i already followed up to figure out how we want to view all for this, but decided to leave that for a future PR to limit scope. i think in this scenario because the designs are so different it might make more sense to rebuild them in header (since it really is just a `StyledAvatarList`)

static/app/views/issueDetails/updatedHeader.tsx
```diff
@@ -0,0 +1,367 @@
...
+      if (newAssignee) {
+        return;
+      }
+    },
+    onError: () => {
+      addErrorMessage('Failed to update assignee');
+    },
+  });
+
+  const disabledTabs = useMemo(() => {
```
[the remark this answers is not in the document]
roggenkemper: definitely - will help clean things up and if we need to diverge that can always be done outside the hook

static/app/components/stream/group.tsx
```diff
@@ -542,34 +539,11 @@ function BaseGroupRow({
           ) : null}
           {withColumns.includes('assignee') && (
             <AssigneeWrapper narrowGroups={narrowGroups}>
-              <AssigneeSelectorDropdown
+              <AssigneeSelector
```
leeandher, line 542: This looks so much cleaner 🔥

static/app/views/issueDetails/issueDetailsHeader.tsx
leeandher, line 1: nit: You can add `use` to the filename since this is only a hook, otherwise it implies this is a component.

leeandher (approved):
Looks awesome thanks for addressing everything, I know it was a lot! 🚢 🚀

static/app/components/assigneeSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+import type {User} from 'sentry/types/user';
+
+interface AssigneeSelectorProps {
+  assigneeLoading: boolean;
+  group: Group;
+  handleAssigneeChange: (assignedActor: AssignableEntity | null) => void;
+  memberList?: User[];
+}
+
+export function AssigneeSelector({
```
malwilley, line 19: What is the difference between this and `AssigneeSelectorDropdown` (which this uses?). Might be good to explain that since this is a global component.

static/app/views/issueDetails/updatedHeader.spec.tsx
```diff
@@ -0,0 +1,101 @@
...
+import {UserFixture} from 'sentry-fixture/user';
+
+import {render, screen} from 'sentry-test/reactTestingLibrary';
+
+import type {TeamParticipant, UserParticipant} from 'sentry/types';
+import {IssueCategory} from 'sentry/types';
+import StreamlinedGroupHeader from 'sentry/views/issueDetails/streamlinedHeader';
+import {ReprocessingStatus} from 'sentry/views/issueDetails/utils';
+
+describe('UpdatedGroupHeader', () => {
```
malwilley, line 14: Thanks for adding these tests!

One thing to remember here is that we are testing a lot of behavior in the old components like `AssignedTo` and the old sidebar/header. We will want to copy those tests over to this one or else we'll lose a lot of test coverage when we delete the old compoents.

roggenkemper: makes sense! ill do that in a separate PR, especially since im still working through some of the functionality of the header

malwilley (approved):

PR #74675, opened 2024-07-22
feat(tracing): reparent ssr spans under pageload txn and under browser request span

JonasBa:
Implements required changes for https://github.com/getsentry/rfcs/pull/138

### DIFF
--- a/static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
+++ b/static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
@@ -218,11 +218,11 @@ function isJavascriptSDKTransaction(transaction: TraceTree.Transaction): boolean
   );
 }
 
-function isPageloadTransaction(node: TraceTreeNode<TraceTree.NodeValue>): boolean {
+function isPageloadTransactionNode(node: TraceTreeNode<TraceTree.NodeValue>): boolean {
   return isTransactionNode(node) && node.value['transaction.op'] === 'pageload';
 }
 
-function isServerRequestHandlerTransaction(
+function isServerRequestHandlerTransactionNode(
   node: TraceTreeNode<TraceTree.NodeValue>
 ): boolean {
   return isTransactionNode(node) && node.value['transaction.op'] === 'http.server';
@@ -590,8 +590,8 @@ export class TraceTree {
 
       if (
         childrenCount === 1 &&
-        isPageloadTransaction(node) &&
-        isServerRequestHandlerTransaction(parent)
+        isPageloadTransactionNode(node) &&
+        isServerRequestHandlerTransactionNode(parent)
       ) {
         // The swap can occur at a later point when new transactions are fetched,
         // which means we need to invalidate the tree and re-render the UI.
@@ -936,7 +936,7 @@ export class TraceTree {
         firstTransaction &&
         !childTransactions.length &&
         isBrowserRequestSpan(spanNodeValue) &&
-        isServerRequestHandlerTransaction(firstTransaction)
+        isServerRequestHandlerTransactionNode(firstTransaction)
       ) {
         childTransactions = [firstTransaction];
         spanNodeValue.childTransactions = childTransactions;
--- a/static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
+++ b/static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
@@ -525,7 +525,8 @@ export class TraceTree {
 
     function visit(
       parent: TraceTreeNode<TraceTree.NodeValue | null>,
-      value: TraceTree.Transaction | TraceTree.TraceError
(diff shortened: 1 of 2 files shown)


static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
```diff
@@ -218,6 +218,42 @@ function isJavascriptSDKTransaction(transaction: TraceTree.Transaction): boolean
...
+}
+
+function isBrowserRequestSpan(value: TraceTree.Span): boolean {
+  return value.op === 'browser' && value.description === 'request';
+}
+
+/**
+ * Swaps the two nodes in the graph.
+ */
+function childParentSwap({
```
JonasBa, line 236: This is a bit more annoying than I would like. Tldr, but because the child nodes keep a reference to the parent, it means we need to filter out the children part of the tree so that we dont end up with circular references.

static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
```diff
@@ -218,6 +218,42 @@ function isJavascriptSDKTransaction(transaction: TraceTree.Transaction): boolean
   );
 }
 
+function isPageloadTransaction(node: TraceTreeNode<TraceTree.NodeValue>): boolean {
+  return isTransactionNode(node) && node.value['transaction.op'] === 'pageload';
+}
+
+function isSSRTransaction(node: TraceTreeNode<TraceTree.NodeValue>): boolean {
```
lforst, line 225: Idk, maybe we want to name this something like `isServerRequestHandlerTransaction` instead because `op: http.server` doesn't really qualify SSR transactions. It's more generic.

static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
```diff
@@ -218,6 +218,42 @@ function isJavascriptSDKTransaction(transaction: TraceTree.Transaction): boolean
...
+  parent,
+  child,
+}: {
+  child: TraceTreeNode<TraceTree.NodeValue>;
+  parent: TraceTreeNode<TraceTree.NodeValue>;
+}) {
+  const parentOfParent = parent.parent!;
+
+  parent.invalidate(parent);
+  child.invalidate(child);
```
lforst, line 246: Dumb question: Wouldn't it make more sense to invalidate _after_ the mutation?

static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
```diff
@@ -871,12 +913,25 @@ export class TraceTree {
...
         event: data as EventTransaction,
         childTransactions,
       };
+
+      if (
+        firstTransaction &&
+        !childTransactions.length &&
+        isBrowserRequestSpan(spanNodeValue) &&
+        isSSRTransaction(firstTransaction)
+      ) {
```
lforst, line 929: I think a small comment of what is going on here would go a long way.

static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
```diff
@@ -1833,6 +1888,26 @@ export class TraceTreeNode<T extends TraceTree.NodeValue = TraceTree.NodeValue>
     return clone;
   }
 
+  filter(
+    node: TraceTreeNode<TraceTree.NodeValue>,
```
lforst, line 1892: Just a thought: Would it make sense to pass the node itself (or rather its value) through the predicate as well?

lforst (approved):
Looks great! All of my comments are non-blocking.

static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
JonasBa: I've moved this out of the swap fn and added a comment. The only reason we need it is because we compute and cache some properties (like the depth of each node) and because reparenting might happen due to incremental trace loading (like in replays), we need to invalidate those fields.

static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
JonasBa: I dont know, right now I dont think there is a need, but it should be a simple change if we need it

static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
```diff
@@ -218,6 +218,41 @@ function isJavascriptSDKTransaction(transaction: TraceTree.Transaction): boolean
   );
 }
 
+function isPageloadTransaction(node: TraceTreeNode<TraceTree.NodeValue>): boolean {
```
Abdkhan14, line 221: Minor: We should use `TraceTree.Transaction` to type the node params here. Same for line 226.

Abdkhan14 (approved):

PR #74686, opened 2024-07-22
feat(devtoolbar): add feature flag panel

### LINKED ISSUE #74459: [DevToolbar] Read the array of enabled feature flags and list them inside a new feature flag panel

michellewzhang:
- closes https://github.com/getsentry/sentry/issues/74459
- adds a basic panel with the org's enabled feature flags

https://github.com/user-attachments/assets/7a05438e-937c-419a-8464-139054901a96


ryan953 (approved):
Nice! Lots of ways to follow up on this too. We could make a search/filter I out at the top. Also, idk if we can link to anything? Probably not yet...
Maybe link to issue search with the feature name as a tag?!

PR #74688, opened 2024-07-22
chore(seer): Add timing metric to requests to Seer

lobsterkatie:
This wraps our calls to Seer with a timing metric, tagged with the endpoint being called.

### DIFF
--- a/src/sentry/seer/signed_seer_api.py
+++ b/src/sentry/seer/signed_seer_api.py
@@ -2,11 +2,13 @@ import hashlib
 import hmac
 from random import random
 from typing import Any
+from urllib.parse import urlparse
 
 from django.conf import settings
 from urllib3 import BaseHTTPResponse, HTTPConnectionPool
 
 from sentry import options
+from sentry.utils import metrics
 
 
 def make_signed_seer_api_request(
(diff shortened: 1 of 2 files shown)


src/sentry/seer/signed_seer_api.py
```diff
@@ -21,13 +22,20 @@ def make_signed_seer_api_request(
...
-        path,
-        body=body,
-        headers={"content-type": "application/json;charset=utf-8", **auth_headers},
-        **timeout_options,
-    )
+    with metrics.timer(
+        "seer.request_to_seer",
+        tags={
+            # Pull off query params, if any
+            "endpoint": path.split("?")[0],
```
JoshFerge, line 29: can do `urlparse(url).path` instead

JoshFerge, line 29: should be able to use `urlparse(url1).path` which might be a tad safer

src/sentry/seer/signed_seer_api.py
```diff
@@ -21,13 +22,20 @@ def make_signed_seer_api_request(
...
         timeout_options["timeout"] = timeout
 
-    return connection_pool.urlopen(
-        "POST",
-        path,
-        body=body,
-        headers={"content-type": "application/json;charset=utf-8", **auth_headers},
-        **timeout_options,
-    )
+    with metrics.timer(
```
JoshFerge, line 25: let's make the sample rate 100% while we're in EA / still debugging this?

src/sentry/seer/signed_seer_api.py
lobsterkatie: Didn't know you could use `urlparse` when all you had was a path! Just tested it out and it seems to work, though, so I made the change.

src/sentry/seer/signed_seer_api.py
lobsterkatie: This metric applies to all requests to Seer, not just grouping. Is it worth making it configurable (i.e., making it so you can pass in a higher sample rate)? Or are we fine with collecting 100% for all Seer features?

JoshFerge: i think its fine

JoshFerge (approved):

PR #74695, opened 2024-07-22
ref(tags): Update tag key-value request to use new dataset param

MichaelSun48:
This PR updates the fetch made to the `organization/<slug>/tags/<key>/values/` endpoint on the frontend within the issue stream search bar and the issue details page. Specifically, it makes the request twice now, and utilizes the extra `dataset` query parameter (which was added in [this PR](https://github.com/getsentry/sentry/pull/74525)) to query only the tag values from the Errors and IssuePlatform datasets. This is in an effort to make the requests faster and more accurate. 

Should be merged after #74696 so we can track the performance improvements of these changes.

### DIFF
--- a/static/app/views/issueDetails/groupEvents.tsx
+++ b/static/app/views/issueDetails/groupEvents.tsx
@@ -19,7 +19,7 @@ import useApi from 'sentry/utils/useApi';
 import useCleanQueryParamsOnRouteLeave from 'sentry/utils/useCleanQueryParamsOnRouteLeave';
 import useOrganization from 'sentry/utils/useOrganization';
 import {Dataset} from 'sentry/views/alerts/rules/metric/types';
-import {mergeTagValues} from 'sentry/views/issueDetails/utils';
+import {mergeAndSortTagValues} from 'sentry/views/issueDetails/utils';
 import {makeGetIssueTagValues} from 'sentry/views/issueList/utils/getIssueTagValues';
 
 import AllEventsTable from './allEventsTable';
@@ -124,7 +124,7 @@ function UpdatedSearchBar({
         projectIds,
         dataset: Dataset.ISSUE_PLATFORM,
       });
-      return mergeTagValues(eventsDatasetValues, issuePlatformDatasetValues);
+      return mergeAndSortTagValues(eventsDatasetValues, issuePlatformDatasetValues);
     },
     [api, group.project.id, organization.slug]
   );
--- a/static/app/views/issueDetails/groupEvents.tsx
+++ b/static/app/views/issueDetails/groupEvents.tsx
@@ -107,7 +107,7 @@ function UpdatedSearchBar({
       const orgSlug = organization.slug;
       const projectIds = [group.project.id];
 
-      const eventsDatasetValues = await fetchTagValues({
+      const eventsDatasetValues = fetchTagValues({
         api,
         orgSlug,
         tagKey: key,
@@ -116,7 +116,7 @@ function UpdatedSearchBar({
         dataset: Dataset.ERRORS,
       });
 
-      const issuePlatformDatasetValues = await fetchTagValues({
+      const issuePlatformDatasetValues = fetchTagValues({
         api,
         orgSlug,
         tagKey: key,
@@ -124,7 +124,10 @@ function UpdatedSearchBar({
         projectIds,
         dataset: Dataset.ISSUE_PLATFORM,
       });
-      return mergeAndSortTagValues(eventsDatasetValues, issuePlatformDatasetValues);
+
+      return await Promise.all([eventsDatasetValues, issuePlatformDatasetValues]).then(
+        tagValues => mergeAndSortTagValues(tagValues[0], tagValues[1])
+      );
     },
     [api, group.project.id, organization.slug]
   );
--- a/static/app/views/issueDetails/groupEvents.tsx
+++ b/static/app/views/issueDetails/groupEvents.tsx
@@ -107,27 +107,26 @@ function UpdatedSearchBar({
       const orgSlug = organization.slug;
       const projectIds = [group.project.id];
 
-      const eventsDatasetValues = fetchTagValues({
-        api,
-        orgSlug,
-        tagKey: key,
-        search,
-        projectIds,
-        dataset: Dataset.ERRORS,
-      });
+      const [eventsDatasetValues, issuePlatformDatasetValues] = await Promise.all([
+        fetchTagValues({
+          api,
+          orgSlug,
+          tagKey: key,
+          search,
+          projectIds,
+          dataset: Dataset.ERRORS,
+        }),
+        fetchTagValues({
+          api,
+          orgSlug,
+          tagKey: key,
+          search,
+          projectIds,
+          dataset: Dataset.ISSUE_PLATFORM,
+        }),
+      ]);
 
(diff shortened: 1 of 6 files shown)


static/app/views/issueDetails/groupEvents.tsx
```diff
@@ -101,17 +103,28 @@ function UpdatedSearchBar({
...
+
+      const issuePlatformDatasetValues = await fetchTagValues({
+        api,
+        orgSlug,
+        tagKey: key,
+        search,
+        projectIds,
+        dataset: Dataset.ISSUE_PLATFORM,
+      });
+      return mergeTagValues(eventsDatasetValues, issuePlatformDatasetValues);
```
MichaelSun48, line 127: I have updated issue details to query the Errors and IssuePlatform datasets under the assumption issue details, like issue stream, also pulls tags from events and issuePlatform, but I'm not 100% sure if that's correct. Lmk if that isn't and I can revert this change

malwilley: This is right! I can see that `monitor.slug` returns results after your change, whereas before it returned nothing. So it's working as intended!

static/app/views/issueList/searchBar.tsx
```diff
@@ -121,15 +123,27 @@ function IssueListSearchBar({organization, tags, onClose, ...props}: Props) {
           : undefined,
         statsPeriod: pageFilters.datetime.period,
       };
+      const eventsDatasetValues = await fetchTagValues({
```
malwilley, line 126: Instead of making sequential requests, we can make them in parallel with `Promise.all`. Since the requests aren't dependent on each other, it makes sense to do it in parallel.

static/app/views/issueList/searchBar.spec.tsx
```diff
@@ -238,4 +236,44 @@ describe('IssueListSearchBar', function () {
...
+      });
+
+      await userEvent.click(screen.getByRole('combobox', {name: 'Add a search term'}));
+      await userEvent.paste(tagKey, {delay: null});
+      await userEvent.click(screen.getByRole('row', {name: tagKey}));
+
+      waitFor(() => {
+        // Expected twice since we make one request for values in events dataset
+        // and another for values in IssuePlatform dataset.
+        expect(tagValueMock).toHaveBeenCalledTimes(2);
```
malwilley, line 275: Might make sense to return a tag from each mock response and see that they are actually rendered

static/app/views/issueDetails/utils.tsx
```diff
@@ -48,6 +48,30 @@ export function useDefaultIssueEvent() {
...
+/**
+ * Combines two TagValue arrays and combines TagValue.count upon conflict
+ */
+export function mergeTagValues(
+  tagValues1: TagValue[],
+  tagValues2: TagValue[]
+): TagValue[] {
+  const tagValueCollection = tagValues1.reduce<Record<string, TagValue>>(
+    (acc, tagValue) => {
+      acc[tagValue.name] = tagValue;
```
malwilley, line 60: I don't think it really matters, but I'd use `value` over `name`

static/app/views/issueDetails/utils.tsx
```diff
@@ -48,6 +48,30 @@ export function useDefaultIssueEvent() {
...
+      return acc;
+    },
+    {}
+  );
+  tagValues2.forEach(tagValue => {
+    if (tagValueCollection[tagValue.name]) {
+      tagValueCollection[tagValue.name].count += tagValue.count;
```
malwilley, line 67: Tag values also have `lastSeen` which will need to be merged as well (take the more recent one)

static/app/views/issueDetails/utils.tsx
```diff
@@ -48,6 +48,30 @@ export function useDefaultIssueEvent() {
...
+    } else {
+      tagValueCollection[tagValue.name] = tagValue;
+    }
+  });
+
+  return Object.values(tagValueCollection);
```
malwilley, line 73: We should also sort the tags here. They should be sorted according to `lastSeen` or `count`, whichever we are using in the original requests. (I think I'd prefer count, but I the default is probably last seen)

MichaelSun48: maybe I'm missing something, but it looks like neither request specifies a sort variable even though the endpoint does accept a sort parameter. I've added a sort param nonetheless and set the default to lastSeen since this is what the backend sorts by default when there is nothing passed in.

static/app/views/issueDetails/utils.tsx
```diff
@@ -51,26 +51,40 @@ export function useDefaultIssueEvent() {
...
     },
     {}
   );
   tagValues2.forEach(tagValue => {
-    if (tagValueCollection[tagValue.name]) {
-      tagValueCollection[tagValue.name].count += tagValue.count;
+    if (tagValueCollection[tagValue.value]) {
+      tagValueCollection[tagValue.value].count += tagValue.count;
+      if (
+        new Date(tagValue.lastSeen) >
```
malwilley, line 70: No need to use the Date class here, ISO format is sortable alphabetically, so you can just compare the string values (very convenient!)

static/app/views/issueList/searchBar.tsx
```diff
@@ -121,15 +123,29 @@ function IssueListSearchBar({organization, tags, onClose, ...props}: Props) {
...
         api,
         orgSlug,
         tagKey: key,
         search,
         projectIds,
         endpointParams,
+        dataset: Dataset.ISSUE_PLATFORM,
       });
+
+      return await Promise.all([eventsDatasetValues, issuePlatformDatasetValues]).then(
```
malwilley, line 146: Look like this is mixing async/await and the old `.then()` syntax. Not the worst thing, but might be a bit cleaner to do 

```
const [eventTagValues, issuePlatformTagValues] = await Promise.all([...]);

return mergeAndSortTagValues(eventTagValues, issuePlatformTagValues);
```

static/app/views/issueDetails/utils.tsx
```diff
@@ -48,6 +48,41 @@ export function useDefaultIssueEvent() {
...
+    } else {
+      tagValueCollection[tagValue.value] = tagValue;
+    }
+  });
+  const allTagValues: TagValue[] = Object.values(tagValueCollection);
+  if (sort === 'count') {
+    allTagValues.sort((a, b) => b.count - a.count);
+  } else {
+    allTagValues.sort(
+      (a, b) => new Date(b.lastSeen).getTime() - new Date(a.lastSeen).getTime()
```
malwilley, line 81: Can also remove the Date class here

static/app/views/issueList/searchBar.spec.tsx
malwilley: @MichaelSun48 looks like the test got fixed a bit, but still missing this at the end

`expect(await screen.findByRole('option', { name: 'randomValue' })).toBeInTheDocument()`

MichaelSun48: @malwilley 
Unless I'm misunderstanding, I believe this is done on line 291?

malwilley: Oops. I misread `tagValue` as `tagKey`. Looks good!

malwilley (approved):

PR #74698, opened 2024-07-22
Add tests for bulk fetch events test

saponifi3d:
# Description
The last of the delayed processing unit tests before looking at perf updates 🎉 

This tests the `bulk_fetch_events` method.

### DIFF
--- a/tests/sentry/rules/processing/test_delayed_processing.py
+++ b/tests/sentry/rules/processing/test_delayed_processing.py
@@ -197,14 +197,14 @@ class BulkFetchEventsTest(CreateEventTestCase):
     def test_more_than_limit_event_ids(self, mock_retry_policy):
         """
         Test that when the number of event_ids exceeds the EVENT_LIMIT,
-        will batch into groups of EVENT_LIMIT and merge the results.
+        batches into groups based on the EVENT_LIMT, and then merges results.
         """
         event_ids: list[str] = ["-1", self.event.event_id, "0", self.event_two.event_id]
         mock_retry_instance = MagicMock()
         mock_retry_policy.return_value = mock_retry_instance
 
-        def mock_return_value(lamda_func):
-            return lamda_func()
+        def mock_return_value(lambda_func):
+            return lambda_func()
 
         mock_retry_instance.side_effect = mock_return_value
 
--- a/tests/sentry/rules/processing/test_delayed_processing.py
+++ b/tests/sentry/rules/processing/test_delayed_processing.py
@@ -3,7 +3,7 @@ from collections.abc import Sequence
 from copy import deepcopy
 from datetime import datetime, timedelta
 from typing import DefaultDict, cast
-from unittest.mock import Mock, patch
+from unittest.mock import MagicMock, Mock, patch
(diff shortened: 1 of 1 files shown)


tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -142,9 +143,66 @@ def test_build_group_to_groupevent(self):
...
+    def test_basic(self):
+        event_ids: list[str] = [self.event.event_id, self.event_two.event_id]
+        result = bulk_fetch_events(event_ids, self.project.id)
+
+        assert len(result) == len(self.expected)
+        for expected, fetched in zip(self.expected, result):
+            assert expected == fetched
+
+    def test_with_invalid_event_id(self):
+        event_ids: list[str] = ["0", "1"]
```
ceorourke, line 183: do we know with certainty these will never be the correct event ids?

tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -142,9 +143,76 @@ def test_build_group_to_groupevent(self):
...
+    def test_more_than_limit_event_ids(self, mock_retry_policy):
+        """
+        Test that when the number of event_ids exceeds the EVENT_LIMIT,
+        will batch into groups of EVENT_LIMIT and merge the results.
+        """
+        event_ids: list[str] = ["0", self.event.event_id, "1", self.event_two.event_id]
+        mock_retry_instance = MagicMock()
+        mock_retry_policy.return_value = mock_retry_instance
+
+        def mock_return_value(lamda_func):
```
ceorourke, line 206: nit: lambda

tests/sentry/rules/processing/test_delayed_processing.py
saponifi3d: `0` yes because of the db schema, but `1` no.. i'll update to be `-1` for that case as well as the others. 👍

tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -142,9 +143,76 @@ def test_build_group_to_groupevent(self):
...
+        assert len(result) == len(self.expected)
+        for expected, fetched in zip(self.expected, result):
+            assert expected == fetched
+
+    @patch("sentry.rules.processing.delayed_processing.ConditionalRetryPolicy")
+    @patch("sentry.rules.processing.delayed_processing.EVENT_LIMIT", 2)
+    def test_more_than_limit_event_ids(self, mock_retry_policy):
+        """
+        Test that when the number of event_ids exceeds the EVENT_LIMIT,
+        will batch into groups of EVENT_LIMIT and merge the results.
```
ceorourke, line 200: will -> we'll? or we will?

ceorourke (approved):
lgtm assuming it's not flaky due to the event ids

tests/sentry/rules/processing/test_delayed_processing.py
saponifi3d: :yay: it's an ai one.

PR #74700, opened 2024-07-22
feat(data-secrecy): Migration to Add `prevent_superuser_access` Bit Flag

iamrajjoshi:
Created migration to add bit flag to Organization. Will need to follow this up with a migration to add the flag so Hybrid Cloud services can handle syncing.

**Glossary**
**Data secrecy mode:** Disallows any kind of superuser access into an organization

**Enable/Disable Data secrecy mode:** Persistently enable/disable data secrecy for an organization

**Waive Data secrecy mode:** Temporarily disable data secrecy for an organizations
**Reinstate Data secrecy mode:** Re-enable data secrecy after a temporary waiver

This flag handles the enable/disable function.

[spec](https://www.notion.so/sentry/Superuser-Data-Secrecy-Mode-b9f7fdfd8b564615ae1f91d3d981bc1a)


tests/sentry/models/test_organization.py
```diff
@@ -228,9 +230,11 @@ def test_handle_2fa_required__compliant_and_non_compliant_members(self):
         self.assert_org_member_mapping(org_member=compliant_member)
         self.assert_org_member_mapping(org_member=non_compliant_member)
 
-        with self.options(
-            {"system.url-prefix": "http://example.com"}
-        ), self.tasks(), outbox_runner():
+        with (
```
iamrajjoshi, line 233: A lot of these diffs are auto-formatting from IDE

src/sentry/models/organization.py
```diff
@@ -354,7 +357,7 @@ def _get_bulk_owner_ids(cls, organizations: Collection[Organization]) -> dict[in
                 organization_id__in=org_ids_to_query, role=roles.get_top_dog().id
             ).values_list("organization_id", "user_id")
 
-            for (org_id, user_id) in queried_owner_ids:
+            for org_id, user_id in queried_owner_ids:
```
iamrajjoshi, line 360: Auto formatting from IDE

src/sentry/models/organization.py
```diff
@@ -198,6 +198,9 @@ class flags(TypedClassBitField):
         # Disable org-members from creating new projects
         disable_member_project_creation: bool
 
+        # Disable superuser access to an organization
+        disable_superuser_access: bool
```
iamrajjoshi, line 202: This is a negative since the default is False and we **do not** want to have data secrecy enabled by default.

src/sentry/api/endpoints/organization_details.py
```diff
@@ -528,6 +531,7 @@ def save(self):
                 "require_2fa": org.flags.require_2fa.is_set,
                 "codecov_access": org.flags.codecov_access.is_set,
                 "disable_member_project_creation": org.flags.disable_member_project_creation.is_set,
+                "disable_superuser_access": org.flags.disable_superuser_access.is_set,
```
cathteng, line 534: i've had trouble with options using the negative before... i would suggest calling this `allow_superuser_access`

src/sentry/api/serializers/models/organization.py
```diff
@@ -353,6 +353,7 @@ def serialize(
             ),
             "avatar": avatar,
             "allowMemberProjectCreation": not obj.flags.disable_member_project_creation,
+            "allowSuperuserAccess": not obj.flags.disable_superuser_access,
```
cathteng, line 356: would customers understand what this means? can we specify that this means sentry employees?

cathteng (commented):
the wording for this project is difficult for sure

src/sentry/api/endpoints/organization_details.py
iamrajjoshi: This is a negative since the default is False and we do not want to have data secrecy enabled by default.
If i make this allow, I will need to write a script to make everyone's allow_data_secrecy True by default and we would have to do this for every new org creation.

src/sentry/api/serializers/models/organization.py
iamrajjoshi: we can add a description in the frontend for this

src/sentry/api/endpoints/organization_details.py
markstory: What about `prevent_superuser_access`? That would avoid needing to use negation 🤷

wedamija (commented):
migration lgtm, bitflag changes are a no-op

cathteng (approved):

PR #74710, opened 2024-07-23
ref(traces): Add 'open in metrics' button

### LINKED ISSUE #74622: Check for existing extraction rules in trace view
- if we already have a rule for that attribute, show Open in Metrics - otherwise keep the + icon
- open in metrics will adjust the time window to +/- 30 mins of that span

priscilawebdev:
### Goal
On the traces pages, we display a '+' button that allows users to add a new metric based on the span attribute when rendering the additional data card. This PR introduces a new button, 'Open in Metrics', which appears when an existing metrics rule is already defined for the span attribute. By clicking this button, users will be directed to the metrics page, pre-filtered with the first condition (query) and the first aggregation (agg by).

**Preview**

https://github.com/user-attachments/assets/cab3cd06-7087-43e9-b777-5ecfbc96c743

### Note:

~We typically render a very large button for 'Open in Discover'. However, this might be too much. I have configured the 'Open in Metrics' button to be 'xs' but feedback is welcome :)~ Now its a link 

closes https://github.com/getsentry/sentry/issues/74622

### DIFF
--- a/static/app/components/keyValueData/index.tsx
+++ b/static/app/components/keyValueData/index.tsx
@@ -1,5 +1,6 @@
 import {Children, isValidElement, type ReactNode, useRef, useState} from 'react';
 import React from 'react';
+import {css} from '@emotion/react';
 import styled from '@emotion/styled';
 
 import {useIssueDetailsColumnCount} from 'sentry/components/events/eventTags/util';
@@ -89,12 +90,11 @@ export function Content({
         {hasSuffix && (
           <div>
             {hasErrors && <AnnotatedTextErrors errors={errors} />}
-            {actionButton &&
-              (actionButtonAlwaysVisible ? (
-                actionButton
-              ) : (
-                <ActionButtonWrapper>{actionButton}</ActionButtonWrapper>
-              ))}
+            {actionButton && (
+              <ActionButtonWrapper actionButtonAlwaysVisible={actionButtonAlwaysVisible}>
+                {actionButton}
+              </ActionButtonWrapper>
+            )}
           </div>
         )}
       </ValueSection>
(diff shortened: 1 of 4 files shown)


static/app/views/performance/newTraceDetails/traceDrawer/details/span/sections/keys.tsx
```diff
@@ -64,6 +76,93 @@ function partitionSizes(data: RawSpanType['data']): {
...
+                displayType: MetricDisplayType.BAR,
+                aggregation: extractionRule.aggregates[0],
+                query: '',
+                groupBy: undefined,
+                condition: extractionRule.conditions[0].id,
+              },
+            ],
+          })}
+        >
+          {t('Open in Metrics')}
```
matejminar, line 137: We used to have this as a blue link. It's way less heavy that way. See "View Similar Spans" button in your video. 👍

static/app/components/keyValueData/index.tsx
```diff
@@ -82,7 +89,12 @@ export function Content({
         {hasSuffix && (
           <div>
             {hasErrors && <AnnotatedTextErrors errors={errors} />}
-            {actionButton && <ActionButtonWrapper>{actionButton}</ActionButtonWrapper>}
+            {actionButton &&
+              (actionButtonAlwaysVisible ? (
+                actionButton
+              ) : (
+                <ActionButtonWrapper>{actionButton}</ActionButtonWrapper>
```
vgrozdanic, line 96: Why do we wrap it in one occasion, and not in the other?

If I understand this correctly, `actionButtonAlwaysVisible` is set to true when there is custom metric available for the span attribute, and false when we want to offer the user possibility to create a new one. Can you maybe rename the variable or add a comment explaining what does the check `actionButtonAlwaysVisible ?` do?

priscilawebdev: sure thing!

static/app/views/performance/newTraceDetails/traceDrawer/details/span/sections/keys.tsx
JonasBa: +1, the links in these section are all inline and we avoid the button as there is already a lot going on in this entire UI here :)

static/app/views/performance/newTraceDetails/traceDrawer/details/span/sections/keys.tsx
```diff
@@ -1,4 +1,5 @@
 import {Fragment, useMemo} from 'react';
+import {Link} from 'react-router';
```
priscilawebdev, line 2:
```suggestion
import Link from 'sentry/components/links/link';
```

vgrozdanic (approved):

PR #74714, opened 2024-07-23
ref: defer opensudo import

JonasBa:
Lets get this out of the import graph for our users since it's only being used by us


JonasBa:
Doesnt seem like its worth it. We probably have everything there already and the modal is already async

PR #74715, opened 2024-07-23
feat(dart): add dart raw stacktrace representation

buenaflor:
For raw stacktraces we don't have a Dart stacktrace representation and we currently fall back to using `getPythonFrame`.

Related issue: https://github.com/getsentry/sentry-dart/issues/2040


malwilley (approved):
Looks good!

PR #74718, opened 2024-07-23
fix(metrics): Fix formula not working

priscilawebdev:
**Before:**

https://github.com/user-attachments/assets/f884cd25-fe7a-46c2-b45d-fc6198754e78

**After:**

https://github.com/user-attachments/assets/0a7f92a6-2e29-47ff-a2a3-ead16dbb7192


obostjancic:
Just a general remark that dashboard widgets that contain equations have them persisted like `equation|$a + $b` so we need to make sure that that still works.

shellmayr:
Only thing I'm noticing is that the function in the summary table is still lower-case, and should be upper-case

priscilawebdev:
> Only thing I'm noticing is that the function in the summary table is still lower-case, and should be upper-case
@shellmayr it reflects what the user types

obostjancic:
> Only thing I'm noticing is that the function in the summary table is still lower-case, and should be upper-case

@priscilawebdev i think if you switch the entire `getQuerySymbol` to use uppercase, this should also work ootb

priscilawebdev:
good catch @shellmayr ! fixed

priscilawebdev:
@obostjancic it seems to be working as expected 🙂

obostjancic (approved):

PR #74723, opened 2024-07-23
feat(dashboards): Default widget split decision to errors

narsaynorath:
In the case that neither side has data, or both sides have data, force
the dataset to errors and record the reason in the source field. This
way we can ensure the split occurs and also warn the user that they may
need to select a dataset.

Note: To avoid changing behaviour for existing on-demand users by forcing the split to use errors, I feature flagged the default scenario so without the feature flag the default is still discover.

### DIFF
--- a/src/sentry/api/bases/organization_events.py
+++ b/src/sentry/api/bases/organization_events.py
@@ -240,33 +240,28 @@ class OrganizationEventsV2EndpointBase(OrganizationEventsEndpointBase):
 
         return use_on_demand_metrics, on_demand_metric_type
 
-    def get_split_decision(self, has_errors, has_transactions_data):
-        """This can be removed once the discover dataset has been fully split"""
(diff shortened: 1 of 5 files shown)


src/sentry/api/bases/organization_events.py
```diff
@@ -240,33 +240,39 @@ def handle_on_demand(self, request: Request) -> tuple[bool, MetricSpecType]:
 
         return use_on_demand_metrics, on_demand_metric_type
 
-    def get_split_decision(self, has_errors, has_transactions_data):
+    def save_split_decision(self, widget, has_errors, has_transactions_data, organization, user):
```
narsaynorath, line 243: The diff looks weird, but I deleted the `get_split_decision` helper just so we can have more control over setting the values when analyzing the decision.

shruthilayaj (approved):

PR #74737, opened 2024-07-23
fix(insights): Cache page alert re-rendering

narsaynorath:
~We should only be triggering the set state for the alert to `undefined` if we're not already `undefined`, i.e. prevents an infinite loop of renders.~

The page alert context provides setters that were getting re-created each time the state changed. Wrap the setters in `useCallback` so the reference to the functions are always the same. This way, hooks that list these functions as dependencies don't go rampant.

### DIFF
--- a/static/app/utils/performance/contexts/pageAlert.tsx
+++ b/static/app/utils/performance/contexts/pageAlert.tsx
@@ -1,5 +1,5 @@
 import type React from 'react';
-import {createContext, Fragment, useContext, useState} from 'react';
+import {createContext, Fragment, useCallback, useContext, useState} from 'react';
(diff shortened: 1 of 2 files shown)


DominikB2014 (approved):

0Calories (approved):

DominikB2014:
@narsaynorath shouldnt setting state to the same value yield no rerenders?

narsaynorath:
@DominikB2014 I agree and thought it was weird so I dug a bit deeper and it looks like it's because of referential inequality. Basically when we trigger the `setPageInfo` function, it creates a new `setPageInfo` callback. I think we could potentially fix this by wrapping the settings with `useCallback`

PR #74738, opened 2024-07-23
fix: split samples by threadid

JonasBa:
Samples need to be grouped by threadId

### DIFF
--- a/static/app/utils/profiling/profile/importProfile.tsx
+++ b/static/app/utils/profiling/profile/importProfile.tsx
@@ -290,7 +290,7 @@ export function importSentryContinuousProfileChunk(
         () => ContinuousProfile.FromProfile(profile, frameIndex),
         {
           op: 'profile.import',
-          description: 'sampled',
+          description: 'continuous',
         }
       )
     );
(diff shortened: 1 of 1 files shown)


static/app/utils/profiling/profile/importProfile.tsx
```diff
@@ -259,13 +258,51 @@ export function importSentryContinuousProfileChunk(
...
+      samples: samplesByThread[key],
+    };
+
+    profiles.push(
+      wrapWithSpan(
+        options.span,
+        () => ContinuousProfile.FromProfile(profile, frameIndex),
+        {
+          op: 'profile.import',
+          description: 'sampled',
```
Zylphrex, line 293: Can we give this a different description, perhaps `continuous`?

Zylphrex (approved):

static/app/utils/profiling/profile/importProfile.tsx
JonasBa: I thought that since it's sampled we could just leave it, but better to distinguish this since these might be a lot longer than the old sampled profiles

PR #74741, opened 2024-07-23
feat(issue-search): Add metrics for new snuba search

snigdhas:
Add metrics to track performance for the new GroupAttributes search.

### DIFF
--- a/src/sentry/search/snuba/executors.py
+++ b/src/sentry/search/snuba/executors.py
@@ -1616,6 +1616,7 @@ class GroupAttributesPostgresSnubaQueryExecutor(PostgresSnubaQueryExecutor):
         end = max([retention_date, end])
         return start, end, retention_date
 
+    @metrics.wraps("snuba.search.group_attributes.query")
     def query(
         self,
         projects: Sequence[Project],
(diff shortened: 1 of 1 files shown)


src/sentry/search/snuba/executors.py
```diff
@@ -1877,5 +1892,10 @@ def query(
 
         groups = group_queryset.in_bulk(paginator_results.results)
         paginator_results.results = [groups[k] for k in paginator_results.results if k in groups]
+        metrics.timing(
```
JoshFerge, line 1895: why not just use @metrics.wraps on the whole function?

snigdhas: smart!

src/sentry/search/snuba/executors.py
```diff
@@ -1660,14 +1661,22 @@ def query(
...
-        if any(sf.key.name in POSTGRES_ONLY_SEARCH_FIELDS for sf in search_filters):
-            max_candidates = options.get("snuba.search.max-pre-snuba-candidates")
-            group_ids_to_pass_to_snuba = list(
-                group_queryset.using_replica().values_list("id", flat=True)[: max_candidates + 1]
-            )
-            if too_many_candidates := (len(group_ids_to_pass_to_snuba) > max_candidates):
-                metrics.incr("snuba.search.too_many_candidates", skip_internal=False)
-                group_ids_to_pass_to_snuba = None
+        max_candidates = options.get("snuba.search.max-pre-snuba-candidates")
+        with sentry_sdk.start_span(op="snuba_group_attributes_query") as span:
```
JoshFerge, line 1665: you could also consider doing `@sentry_sdk.tracing.trace` instead this

JoshFerge: (which is another decorater on the function)

JoshFerge: oh i guess you want the span to set data on it. this makes sense.

JoshFerge (approved):

PR #74743, opened 2024-07-23
ref(integration): Introduce MessagingIntegrationSpec

RyanSkonnord:
Begin unifying the three messaging-related integrations (Slack, MS Teams, and Discord) with an abstract base class. The base class is a prototype that extracts duplicated code in the short term, and in the long term may serve as a comprehensive "feature inventory" of all Sentry's features related to messaging integrations.


RyanSkonnord:
Previously spiked on and discussed at https://github.com/getsentry/sentry/pull/74070

My goal with this PR is to get a minimal but stable version of the "spec" class into master, for us to build off of.

One major order of business: are we happy with `MessagingIntegrationSpec` and all other naming choices? I struggled a bit to come up with a metaphor for this new base class; "spec" reflects its purpose as a "feature inventory" and as a framework for scoping unimplemented features, but ideally it would also hold as much of the business logic as we can put there.

cathteng (commented):
i am understanding what is happening, which is a great sign for me!

src/sentry/integrations/messaging.py
```diff
@@ -0,0 +1,215 @@
...
+        `get_identity_view_set_url_patterns` for things *not* covered by this method,
+        which require additional boilerplate.
+        """
+
+        AlertRuleTriggerAction.register_factory(_MessagingHandlerFactory(self))
+
+        if self.notify_service_action:
+            rules.add(self.notify_service_action)
+        if self.notification_sent:
+            analytics.register(self.notification_sent)
```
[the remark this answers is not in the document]
RyanSkonnord: Yes. However, since they are currently implemented for all messaging integrations, we could go ahead and make them required. I'm trying to remember my rationale here -- I think I intended it as a conceptual demonstration that they don't *have* to be required unless we want?

I should at least go ahead and make `notify_service_action` required -- the integration is pretty useless without it.

src/sentry/integrations/messaging.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+@dataclass(frozen=True)
+class MessagingIdentityLinkViewSet:
+    """An integration's set of view classes for linking and unlinking identities."""
+
+    link_personal_identity: type[View]
+    unlink_personal_identity: type[View]
+
+    # TODO: Allow these to be optional? Require for all messaging integrations?
```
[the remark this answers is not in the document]
RyanSkonnord: Yes! For anything that we *want* to be a consistent feature across all messaging integrations, it would be good to represent it in the spec and to communicate whether it's optional. Filling in each unimplemented method can be self-contained tasks, and when they're all implemented we can switch it to be required.

src/sentry/integrations/discord/__init__.py
```diff
@@ -11,4 +10,4 @@
 from .utils import *  # noqa: F401,F403
 from .views import *  # noqa: F401,F403
 
-rules.add(DiscordNotifyServiceAction)
+DiscordMessagingSpec().initialize()
```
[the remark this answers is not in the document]
RyanSkonnord: I would like that, especially in coordination with the work being done on the SCM and ticketing verticals.

I'm 👍 to nuking the `__init__` imports. Moreover, I'd like to move the `initialize()` call from here (initializing as a side effect of package import == 🤮) to somewhere like `sentry/runner/initializer.py` where it can be called at a predictable time during server start-up.

cathteng (approved):
🚢

src/sentry/integrations/msteams/spec.py
```diff
@@ -0,0 +1,59 @@
...
+from sentry.integrations.base import IntegrationProvider
+from sentry.integrations.messaging import MessagingIdentityLinkViewSet, MessagingIntegrationSpec
+from sentry.models.notificationaction import ActionService
+from sentry.rules.actions import IntegrationEventAction
+
+
+class MsTeamsMessagingSpec(MessagingIntegrationSpec):
+    @property
+    def provider_slug(self) -> str:
+        return "msteams"
```
GabeVillalobos, line 13: Should these eventually live in an enum somewhere vs just being hardcoded on the spec?

src/sentry/integrations/messaging.py
```diff
@@ -0,0 +1,218 @@
...
+        self,
+        metric_value: int | float,
+        new_status: IncidentStatus,
+        notification_uuid: str | None = None,
+    ) -> None:
+        success = self._spec.send_incident_alert_notification(
+            self.action, self.incident, metric_value, new_status, notification_uuid
+        )
+        if success:
+            self.record_alert_sent_analytics(self.action.target_identifier, notification_uuid)
```
GabeVillalobos, line 202: Do we have any place that captures metrics or sentry issues around alert failures? If not, this might be a good thing to add here.

GabeVillalobos (approved):

src/sentry/integrations/msteams/spec.py
RyanSkonnord: Maybe. My problem for now is that there are too many enums already that map onto integration providers, none of which represent quite the same thing: `ActionService`, `ExternalProviders`, `ExternalProviderEnum`. I'd rather unify them than add another one.

Furthermore, I'd actually prefer not to have enums like that at all. It's a bit of an antipattern to have places in the code where we have to list every integration of a certain kind, rather than having each spec subclass stand on its own. (Imagine adding a new messaging integration. How many places do we need to update?)

I tend to think of a hardcoded return value of a property being similar in spirit to a class-level constant, in that it's declared in one place and can't be (normally) mutated. In other words, we could have written

```python
class MsTeamsMessagingSpec(MessagingIntegrationSpec):
    PROVIDER_SLUG = "msteams"
```

but referring to `MsTeamsMessagingSpec().provider_slug` is basically the same as `MsTeamsMessagingSpec.PROVIDER_SLUG`. Does it feel this way to folks, or is it just me?

src/sentry/integrations/messaging.py
RyanSkonnord: I'd be happy to add one. Would we want it to go to `analytics` as `record_alert_sent_analytics` does? Or just `metrics` or `logger`?

src/sentry/integrations/msteams/spec.py
GabeVillalobos: I agree that having enums for everything is an antipattern, so I wonder if these different values are things we should be consolidating some of these enums in static configs for each provider vs just placing them on the classes themselves? I guess my main concern is if we have to copy the provider slug to each distinct spec implementation for a given provider, which seems brittle.

src/sentry/integrations/messaging.py
GabeVillalobos: What's the main use case for `analytics`? I sort of just assumed it was mainly for bizops tracking 😅 

I typically use metrics + optionally logger (depending on the volume)

PR #74744, opened 2024-07-23
ref(integration): Extract base class for identity linking

RyanSkonnord:
Extract common code for the "link identity" and "unlink identity" operations on messaging integrations.


RyanSkonnord:
This PR builds on https://github.com/getsentry/sentry/pull/74743. I'm keeping it in a draft state as I continue to pull more code into the base class.

- The **present** state of the PR (as I write this comment) is a basic proof of concept showing how the linking and unlinking views for MS Teams and Discord can be unified.
- Slack's view classes are structured differently, and also have more logging and such. I'll be working on unifying Slack with the other two, hopefully imposing some consistency along the way.
- Once Slack's identity-linking business logic is unified, I hope it will serve as a template for how to extract an abstract base view for the team-linking views. That can then "specify" the implementation of the team-linking feature for MS Teams and Discord (at which point we'll hopefully learn if it's suitable for handing off to a developer as a self-contained task, or if the abstraction process makes the remaining steps trivial).

Visual representation:

| | Slack | MS Teams | Discord |
|-|-|-|-|
| Link Identity | ⌛ (1) | ✅ | ✅ |
| Unlink Identity | ⌛ (1) | ✅ | ✅ |
| Link Team | ⌛ (2) | ❓ (3) | ❓ (3) |
| Link Team | ⌛ (2) | ❓ (3) | ❓ (3) |

- ✅ = has been extracted into abstract base classes
- ⌛ = to be extracted into abstract base classes (numbers show intended order)
- ❓ = not yet implemented at all; abstract base class will hopefully show the way

RyanSkonnord:
I'm setting this as ready for review. Only the "✅" items as described in my previous comment are in scope right now. The next steps will entail a level of complexity that seem to be worth handling separately.

asottile-sentry:
imo inheritance makes this way harder to reason about and debug -- would prefer free functions which are also much easier to test

src/sentry/integrations/messaging.py
```diff
@@ -216,3 +237,174 @@ def build_handler(
...
+    @property
+    @abstractmethod
+    def success_template(self) -> str:
+        """Path to the HTML template to show when an identity has been linked."""
+        raise NotImplementedError
+
+    @property
+    def success_metric(self) -> str | None:
+        """Optional analytics key to record on success."""
+        return None
```
cathteng, line 292: should we use something like this for success / failure of linking / unlinking?

src/sentry/integrations/messaging.py
```diff
@@ -216,3 +237,174 @@ def build_handler(
...
+        self, idp: IdentityProvider | None, params: Mapping[str, Any], request: HttpRequest
+    ) -> HttpResponse | None:
+        if idp is None:
+            raise ValueError('idp is required for linking (params must include "integration_id")')
+
+        user = request.user
+        if isinstance(user, AnonymousUser):
+            raise TypeError("Cannot link identity without a logged-in user")
+
+        Identity.objects.link_identity(user=user, idp=idp, external_id=params[self.user_parameter])
```
cathteng, line 374: should this be put in a try/except and should we track metrics for it? i'm not sure how much identity linking errors relate to system or user errors

cathteng (commented):
just questions about metrics, but i think this is looking good :)

src/sentry/integrations/messaging.py
RyanSkonnord: We got this from the Slack code when https://github.com/getsentry/sentry/pull/75899 brought it in. Failures on this query are now captured as metrics for all providers.

src/sentry/integrations/messaging.py
RyanSkonnord: I generalized this in https://github.com/getsentry/sentry/pull/75899 (now merged into this branch) and cleaned it up a bit in https://github.com/getsentry/sentry/pull/74744/commits/76d43b7b307fae6b9ffd6ad6a9d03eb04eb885f8. The level of log coverage that @iamrajjoshi implemented for Slack is now present for all providers.

RyanSkonnord:
> imo inheritance makes this way harder to reason about and debug -- would prefer free functions which are also much easier to test

@asottile-sentry Could you elaborate/give examples? I think the OO approach is good here. For one thing, the end product needs to be `View` objects anyway. For another, I think it's a pretty natural way to express the set of inter-provider differences and inter-operation differences. The abstract properties document the values that would need to be filled in when support is added for an operation with a new provider, which goes along with Ecosystem dev goals.

My gut says that, if I tried to refactor the operations into free functions, they would need a number of parameters proportional to the current number of abstract properties -- so, a lot. I picked the somewhat complex class structure over functions with obnoxiously long arg lists. Happy to hear other perspectives, though -- thanks.

RyanSkonnord:
A sub-branch that replaced the ABC's implementation with Slack's was reviewed in https://github.com/getsentry/sentry/pull/75899 and is now merged into here. This branch should be ready to ship.

Updated table:

| | Slack | MSTeams | Discord |
|-|-|-|-|
| Link identity | ✅ | ✅ | ✅ |
| Unlink identity | ✅ | ✅ | ✅ |
| Link team | ⌛ |❓ | ❓ |
| Link team | ⌛ |❓ | ❓ |

✅ = in the scope of this PR
⌛ = in progress at https://github.com/getsentry/sentry/pull/76214
❓ = targeted for future hand-off

src/sentry/integrations/messaging.py
```diff
@@ -199,3 +225,283 @@ def build_handler(
...
+        metrics.incr(event, tags=(tags or {}), sample_rate=1.0)
+        return event
+
+    @property
+    def has_analytics(self) -> bool:
+        return False
+
+    def record_analytic(self, event_tag: str, actor_id: int) -> None:
+        if not self.has_analytics:
+            return
```
RyanSkonnord, line 301: Previously only Discord had the `analytics.record` statement; this is to preserve that behavior without adding a new analytics key.

I'm not opposed to either adding it to the others or removing it from Discord, which would obviate the `has_analytics` property. But for now I'm trying to minimize behavior changes.

src/sentry/integrations/discord/views/link_identity.py
```diff
@@ -27,41 +23,44 @@ def build_linking_url(integration: RpcIntegration, discord_id: str) -> str:
...
-            "integrations.discord.identity_linked",
-            provider="discord",
-            actor_id=request.user.id,
-            actor_type=ActorType.USER,
-        )
-        return render_to_response(
-            "sentry/integrations/discord/linked.html",
-            request=request,
-        )
+class DiscordLinkingView(LinkingView, ABC):
```
sentaur-athena, line 26: Can you move this to its own file since this one is called link identity?

src/sentry/integrations/messaging.py
sentaur-athena: I agree with minimizing the behavior change. But makes sense to have it in the parent class, can you please add a TODO here and also add it to the backlog you have for future work.

src/sentry/integrations/messaging.py
```diff
@@ -199,3 +225,283 @@ def build_handler(
...
+        if organization is not None:
+            self._send_nudge_notification(organization, request)
+
+        success_template, success_context = self.get_success_template_and_context(
+            params_dict, integration
+        )
+        return render_to_response(success_template, request=request, context=success_context)
+
+    def _send_nudge_notification(self, organization: RpcOrganization, request: Request):
+        # TODO: Delete this if no longer needed
```
sentaur-athena, line 405: Is it not used?

src/sentry/integrations/slack/utils/notifications.py
```diff
@@ -163,51 +164,67 @@ def send_incident_alert_notification(
     return success
 
 
+class SlackCommand(Enum):
```
sentaur-athena, line 167: Would another potential backlog item be to move commands to the parent class?

src/sentry/integrations/slack/views/link_identity.py
```diff
@@ -49,118 +32,60 @@ def build_linking_url(
     )
 
 
-@control_silo_view
-class SlackLinkIdentityView(BaseView):
-    """
-    Django view for linking user to slack account. Creates an entry on Identity table.
-    """
+class SlackLinkingView(LinkingView, ABC):
```
sentaur-athena, line 35: nit: move this to its own file

src/sentry/templates/sentry/integrations/generic-error.html
```diff
@@ -0,0 +1,14 @@
+{% extends "sentry/bases/modal.html" %}
```
sentaur-athena, line 1: Is this a new view? Do you have a screenshot of it?

src/sentry/integrations/discord/views/link_identity.py
RyanSkonnord: Sure. I made that change in https://github.com/getsentry/sentry/pull/76214 and I've pulled it down into this branch.

I'm also adopting the term "linkage" in class names, to try to encompass both linking and unlinking.

src/sentry/templates/sentry/integrations/generic-error.html
RyanSkonnord: No, it's an exact duplicate of `slack/link-team-error.html`, which was never specific to Slack nor to team-linking. I just copied it to a new path for clarity.

Deleting the file from the old path is another todo.

src/sentry/integrations/slack/utils/notifications.py
RyanSkonnord: Although the current set of commands is limited to identity linkage, I think `respond_to_slack_command` wants to remain in `slack/utils/notifications.py` because it was intended to be used for commands in general.

However, your question did give me the idea of moving the values associated with each command into the parent classes, and passing a dataclass here instead of an enum value. I think it's much nicer that way. See https://github.com/getsentry/sentry/pull/74744/commits/3e9926d0739524f657b4219fb659a197c12ba32c

src/sentry/integrations/messaging.py
RyanSkonnord: I don't think it is. Per the comment on line 422, I think this has been broken since the Hybrid Cloud split.

sentaur-athena: This is not blocker for your PR since it was existing behavior but can you follow-up with the team in slack and figure out? Basically, is it important? -> add fix to backlog, if not -> delete

sentaur-athena (approved):

PR #74745, opened 2024-07-23
ref(control_silo): Consolidate integrations models and models' tests

Christinarlong:
Moves all integrations models and related tests to the sentry/integrations/models or tests/sentry/integrations/models directories

ref(#73859)


src/sentry/integrations/models/external_issue.py
```diff
@@ -0,0 +1,98 @@
...
+        unique_together = (("organization", "integration_id", "key"),)
+
+    __repr__ = sane_repr("organization_id", "integration_id", "key")
+
+    def get_installation(self) -> Any:
+        from sentry.integrations.services.integration import integration_service
+
+        integration = integration_service.get_integration(integration_id=self.integration_id)
+
+        return (
```
Christinarlong, line 94: I changed this btw from `return integration.get_installation(organization_id=self.organization_id)` to deal with mypy saying `None doesn't have .get_integration()`

markstory: Makes sense. Another solution for these kinds of error where the types indicate that something could be missing, but it rarely is in practice is to use assertions

```python
assert integration, "integration is required to get an installation"
```

an assertion like this will narrow the variable's type removing `None`. Raising an exception can be helpful in identifiying if this problem occurs in production more than returning a `None` will.

src/sentry/models/avatars/__init__.py
```diff
@@ -1,14 +1,12 @@
...
 from .control_base import ControlAvatarBase
-from .doc_integration_avatar import DocIntegrationAvatar
 from .organization_avatar import OrganizationAvatar
 from .sentry_app_avatar import SentryAppAvatar
 from .user_avatar import UserAvatar
 
 __all__ = (
     "AvatarBase",
     "ControlAvatarBase",
-    "DocIntegrationAvatar",
```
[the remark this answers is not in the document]
Christinarlong: The only place I could find `DocIntegrationAvatar` being used was in the now deleted `migrate_control_files` script where we also saw the `UserAvatar` from last week

markstory (approved):

PR #74750, opened 2024-07-23
Fix creating crash free session alerts

### LINKED ISSUE #74716: Unable to save Crash Free Session rate alert
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

1. Go to the Alerts page
2. Create new Crash free session rate alert
3. Add dummy data
4. Try saving the rule - it will return an internal error

Example error: id: 064a31d78deb46978f5938fa094a47ec

### Expected Result

it should be possible to create a Crash Free Session rate alert

### Actual Result

it is not currently possible to create the alert

### Product Area

Alerts

### Link

_No response_

### DSN

_No response_

### Version

_No response_

saponifi3d:
# Description
Fixes #74716 

A user reported an issue making crash free session rate alerts, upon investigation it was using a deprecated dataset. This updates the alert UI to use the correct dataset (which fixes the save issue)


nhsiehgit (approved):
🙌

PR #74761, opened 2024-07-23
feat(discover): Add top-n support for errors dataset

shruthilayaj:
Adds top-n support for errors dataset
Allows errors dataset to pass aggregates as filter conditions - this is needed to maintain backwards compatibility with existing `discover` dataset queries that will query the `errors` dataset going forward
Also adds missing test coverage for errors dataset

### DIFF
--- a/src/sentry/search/events/builder/errors.py
+++ b/src/sentry/search/events/builder/errors.py
@@ -44,6 +44,7 @@ class ErrorsQueryBuilderMixin:
             self.params.user,
             list(filter(None, self.params.environments)),
             value_converters=value_converters,
+            allow_aggregate_filters=True,
         )
         return parsed_terms
 
--- a/src/sentry/search/events/builder/errors.py
+++ b/src/sentry/search/events/builder/errors.py
@@ -18,7 +18,11 @@ from snuba_sdk import (
 )
 
 from sentry.api.issue_search import convert_query_values, convert_status_value
-from sentry.search.events.builder.discover import DiscoverQueryBuilder, TimeseriesQueryBuilder
+from sentry.search.events.builder.discover import (
+    DiscoverQueryBuilder,
+    TimeseriesQueryBuilder,
+    TopEventsQueryBuilder,
+)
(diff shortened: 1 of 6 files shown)


src/sentry/search/events/builder/errors.py
```diff
@@ -40,6 +44,7 @@ def parse_query(self, query: str | None) -> ParsedTerms:
             self.params.user,
             list(filter(None, self.params.environments)),
             value_converters=value_converters,
+            allow_aggregate_filters=True,
```
shruthilayaj, line 47: We're splitting queries to discover dataset to explicitly query errors or transactions datasets.

The `convert_query_values` doesn't allow aggregate filters but it's allowed in discover queries. Added this allow_aggregate_filters param to get around it so the ErrorsQueryBuilder is allowed to have aggregate filters and all other references to `convert_query_values` continue to not allow it. Want to make sure this is okay (@wedamija perhaps you'd know?)

src/sentry/snuba/errors.py
```diff
@@ -179,3 +188,165 @@ def timeseries_query(
...
+
+
+def top_events_timeseries(
+    timeseries_columns: list[str],
+    selected_columns: list[str],
+    user_query: str,
+    params: ParamsType,
+    orderby: list[str],
+    rollup: int,
+    limit: int,
```
wmak, line 200: I'm moving to add snuba_params everywhere, do you mind adding it here too?
reference here; https://github.com/getsentry/sentry/compare/wmak/ref/change-snuba-modules-to-include-snuba-params?expand=1

wmak (approved):
As mentioned in Slack, if possible we should try to use inheritance so there's less code repeat in the tests, but not a huge blocker

shruthilayaj:
> As mentioned in Slack, if possible we should try to use inheritance so there's less code repeat in the tests, but not a huge blocker

I'll address this in a follow up! I was just being lazy to re-do all overlapping discover/errors/transaction test cases with events and conditions that would work for all of them

PR #74766, opened 2024-07-23
fix(hybrid-cloud): Adds defaults to provisioning model fields, lost_password_hash model

GabeVillalobos:
Fixes a few issues that crop up as errors in the Pydantic upgrade work:
- Optional fields without defaults now cause failures instead of assuming a value of None
- Type mismatches between fields are no longer automatically coerced without an additional config setting

### DIFF
--- a/src/sentry/users/services/lost_password_hash/model.py
+++ b/src/sentry/users/services/lost_password_hash/model.py
@@ -3,7 +3,10 @@
 # in modules such as this one where hybrid cloud data models or service classes are
 # defined, because we want to reflect on type annotations and avoid forward references.
 
-import datetime
+from datetime import datetime
+
+from django.utils import timezone
+from pydantic import Field
 
 from sentry.hybridcloud.rpc import RpcModel
 from sentry.models.lostpasswordhash import LostPasswordHash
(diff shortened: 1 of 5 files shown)


src/sentry/users/services/lost_password_hash/model.py
```diff
@@ -13,7 +16,7 @@ class RpcLostPasswordHash(RpcModel):
     id: int = -1
     user_id: int = -1
     hash: str = ""
-    date_added = datetime.datetime
+    date_added: datetime = Field(default_factory=timezone.now)
```
GabeVillalobos, line 19: I'm not super sure we want this to be the default. I could also remove this entirely and just assume the caller needs to provide one, but I this was failing tests for whatever reason 🤔

corps: I think it's a reasonable default.  Forcing the caller to provide one is reasonable in tests, but not in production, due to the reality that there will be (increasing) version drift between callers and receivers.

corps: Whatever default does it exist, it should be safe for the purpose of backwards compatibility, which depends on the use case.

RyanSkonnord (approved):

src/sentry/users/services/lost_password_hash/model.py
markstory: This default seems fine to me. Looking at the existing usage, we only create `RpcLostPasswordHash` by serializing an ORM object, which would have `date_added` set as well.

markstory (approved):

PR #74768, opened 2024-07-23
build: Start uploading test results to Codecov

joseph-sentry:
This enhances the Codecov coverage comment when test failures occur to show information pertaining to test failures instead of coverage results. This also enables Codecov to start running flaky test detection in the background on Sentry's test results to see how it performs in terms of latency and accuracy.


bukzor (commented):
+1

I just need you to look at the effect on CI time.

.github/actions/artifacts/action.yml
```diff
@@ -20,13 +24,22 @@ inputs:
...
+        curl -Os https://cli.codecov.io/latest/linux/codecov.SHA256SUM
+        curl -Os https://cli.codecov.io/latest/linux/codecov.SHA256SUM.sig
+        gpgv codecov.SHA256SUM.sig codecov.SHA256SUM
+        shasum -a 256 -c codecov.SHA256SUM
+        chmod +x codecov
+    - name: Upload Coverage
+      shell: bash
       continue-on-error: true
+      run: |
+        files=$(echo ${{ inputs.files }} | sed 's/^/--file /;s/,/ --file /')
```
bukzor, line 41: This is fraught, whenever the variable is likely to contain whitespace, quotes, other shell characters (and this is). The `${{` syntax will textually substitute the filenames so all characters in filesnames become part of the _source_ of the shell script -- an escaping nightmare.

Instead, pass all values from GHA to sh as environment variables. This looks similar but behaves very differently as the shell now has no confusion about which characters are the shell script and which are parts of the parameters.

```bash
env:
  INPUT_FILES: ${{ inputs.test_result_files }}
```

.github/actions/artifacts/action.yml
```diff
@@ -20,13 +24,22 @@ inputs:
...
+        test_result_files=$(echo ${{ inputs.test_result_files }} | sed 's/^/--file /;s/,/ --file /')
+        ./codecov --verbose upload-process -t ${{ inputs.token }} --commit-sha ${{ inputs.commit_sha }} --plugin noop --flag ${{ inputs.type }} $files &
+        ./codecov --verbose do-upload --report-type test_results -t ${{ inputs.token }} --commit-sha ${{ inputs.commit_sha }} --plugin noop --flag ${{ inputs.type }} $test_result_files  &
+        wait
```
bukzor, line 45: Please put anything even slightly hard to read into its own shell script. Then we can take advantage of syntax highlighting and shellcheck, etc. You can create as many little scripts as you like under `.github/actions/artifacts/`.

.github/actions/artifacts/action.yml
```diff
@@ -20,13 +24,22 @@ inputs:
...
-      uses: codecov/codecov-action@e0b68c6749509c5f83f984dd99a76a1c1a231044 # v4.0.1
-      with:
-        token: ${{ inputs.token }}
-        flags: ${{ inputs.type }}
-        files: ${{ inputs.files }}
-        override_commit: ${{ inputs.commit_sha}}
-        plugin: noop
-        verbose: true
+    - name: Download and Verify Codecov CLI
+      shell: bash
```
bukzor, line 28: Please use the `shell` seen in actions/setup-sentry to ensure that these commands are properly logged and properly stop on error:

`actions/setup-sentry/action.yml`
```
# the default for "bash" is:
#      bash --noprofile --norc -eo pipefail {0}
shell: bash --noprofile --norc -eo pipefail -ux {0}
```

This becomes unnecessary if you move all shell code to their own files (just don't forget to `set -euxo pipefail` on the first line).

bukzor (changes requested):

bukzor (commented):
I still want you to convince me, with details, that this doesn't add a new long-pole to CI.
I could do it but it seems like it should be the author's burden.

joseph-sentry:
@bukzor 
Here's this PRs `backend test (0,14)` step times:
image
Handle artifacts took 5s.

Here are some commits on main:
image
image
image

.github/actions/artifacts/download-codecov-cli.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -euo pipefail
+
+curl https://keybase.io/codecovsecurity/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import # One-time step
+curl -Os https://cli.codecov.io/latest/linux/codecov
```
bukzor, line 5: It would be better to see these commands in the log.

```suggestion

set -x
curl https://keybase.io/codecovsecurity/pgp_keys.asc |
    gpg --no-default-keyring --keyring trustedkeys.gpg --import
curl -Os https://cli.codecov.io/latest/linux/codecov
```

.github/actions/artifacts/action.yml
bukzor: Resolved by `set -euo pipefail` in scripts, instead.

bukzor (approved):
Shipit!

.github/actions/artifacts/do-upload.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
asottile-sentry, line 1: please do not write new shell scripts -- this should be written in a language which can have proper type checking and tests.  python is sentry's language of choice for such things

asottile-sentry (dismissed):
same for the other shell scripts

bukzor (approved):

.github/actions/artifacts/do-upload.sh
joseph-sentry: Fixed this if you want to take another look!

PR #74770, opened 2024-07-23
chore(hybrid-cloud): Upgrade Pydantic to version 2.7

GabeVillalobos:
Based on PR #74766, which should be merged first.

This PR upgrades pydantic and updates the model configs we use for RPC methods to conform with the new version.

This PR _does not_ address deprecation warnings, which will be handled in a series of follow-up commits to both Sentry and GetSentry.


.github/workflows/backend.yml
```diff
@@ -10,7 +10,7 @@ on:
 # https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
-  cancel-in-progress: true
+  cancel-in-progress: false
```
GabeVillalobos, line 13: This is only set for obtaining test results. This will be removed prior to merging.

RyanSkonnord (approved):

requirements-base.txt
```diff
@@ -45,7 +45,7 @@ python-rapidjson>=1.4
 psutil>=5.9.2
 psycopg2-binary>=2.9.9
 PyJWT>=2.4.0
-pydantic>=1.10.17,<2
+pydantic>=2.5.0
```
markstory, line 48:
```suggestion
pydantic>=2.5.0,<3
```

Could we pin to the 2.x range so that we don't accidentally bump a major?

markstory (approved):

requirements-base.txt
asottile-sentry: there's no risk to uncapping here -- in fact it's encouraged

this file isn't installed from -- it's used to generate the pinned file

asottile-sentry (approved):
fwiw I usually split out the compatible changes and land them first -- so that the core changes can be focused on

PR #74772, opened 2024-07-23
subscription processor groupby pseudo

nhsiehgit:
pseudo code of what changes might need to be made to the subscription processor in order to handle group by

### DIFF
--- a/src/sentry/incidents/models/alert_rule.py
+++ b/src/sentry/incidents/models/alert_rule.py
@@ -51,7 +51,8 @@ logger = logging.getLogger(__name__)
 
 
 class SubscriptionCallback(Protocol):
-    def __call__(self, subscription: QuerySubscription, *args: Any, **kwargs: Any) -> bool: ...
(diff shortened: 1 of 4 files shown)


src/sentry/incidents/models/alert_rule.py
```diff
@@ -734,6 +734,12 @@ class Meta:
 def update_alert_activations(
     subscription: QuerySubscription, alert_rule: AlertRule, value: float
 ) -> bool:
+    # TODO: handle multiple values?
+    # metric_value is on the activations model...
+    # Do we have multiple activations per alert rule then?
+    # We _already_ have multiple activations per alert rule...
+    # they're filtered per subscription though....
+    # So - if a subscription can have multiple values... then this is NOT supported :thinking:
```
nhsiehgit, line 742: OR - we just get rid of `Activations` since this was mainly implemented to track values for activated alerts

If we get rid of activated alerts then we can get rid of activations as well (unless we actually want explicit "activations" rather than implicitly derived incidents)

PR #74777, opened 2024-07-23
feat(uptime): Add function to count number of uptime monitors active in an organization

wedamija:
This adds a funtion to count active uptime monitors for an org. Since this will be called from high volum areas it also adds caching around the function.

We want the cache to be actively invalidated when the count changes in the org, so we also want to listen to `post_save` and `post_delete` signals. This is a pattern we've used a few times in sentry, so I also generalized this cache so that we don't have to manually write it every time.

### DIFF
--- a/tests/sentry/utils/test_function_cache.py
+++ b/tests/sentry/utils/test_function_cache.py
@@ -36,7 +36,6 @@ class CacheFuncForModelsTest(TestCase):
     def test(self):
         mock_test_func = create_autospec(count_func)
         mock_test_func.side_effect = count_func
-        # mock_test_func =  MagicMock(side_effect=count_func)
         decorated_test_func = cache_func_for_models([(CacheModel, arg_extractor)])(mock_test_func)
         self.assert_called_with_count(mock_test_func, "test", 0)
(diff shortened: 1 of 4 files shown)


tests/sentry/utils/test_function_cache.py
```diff
@@ -0,0 +1,82 @@
...
+class CacheFuncForModelsTest(TestCase):
+    def assert_called_with_count(self, mock_test_func, text_search: str, count: int):
+        assert (
+            len([ca for ca in mock_test_func.call_args_list if ca.args[0] == text_search]) == count
+        )
+
+    def test(self):
+        mock_test_func = create_autospec(count_func)
+        mock_test_func.side_effect = count_func
+        # mock_test_func =  MagicMock(side_effect=count_func)
```
davidenwang, line 39: nit: is this comment intentional?

davidenwang (approved):

tests/sentry/utils/test_function_cache.py
wedamija: Good catch, removed

PR #74779, opened 2024-07-23
ref(control_silo): Consolidate integrations bases

Christinarlong:
Moves all integrations related bases to integrations/api/bases . Updates related path changes and adds import shims for bases accessed in getsentry.

ref(#73859)


markstory (approved):
Looks good to me :clap:

markstory (approved):
Looks good to me.

PR #74780, opened 2024-07-23
ref(querybuilder): Add snuba_params to all the snuba query modules

wmak:
- This updates all the snuba query modules so they can take the
  snuba_params instead of params
- Should also allow params={} since we can infer it from the
  snuba_params


src/sentry/api/endpoints/organization_events_histogram.py
```diff
@@ -79,11 +79,11 @@ def get(self, request: Request, organization) -> Response:
...
-                        data["field"],
-                        data.get("query"),
-                        params,
-                        data["numBuckets"],
-                        data["precision"],
+                        fields=data["field"],
+                        user_query=data.get("query"),
+                        params=params,
+                        num_buckets=data["numBuckets"],
+                        precision=data["precision"],
```
wmak, line 86: Didn't really want to fiddle with this more than I needed to, just added all the args instead of reordering it

src/sentry/snuba/discover.py
```diff
@@ -322,6 +326,10 @@ def timeseries_query(
         Dataset.Discover,
         Dataset.Transactions,
     ], "A dataset is required to query discover"
+
+    if len(params) == 0 and snuba_params is not None:
+        params = snuba_params.filter_params
+
```
wmak, line 332: This is a pattern i reuse repeatedly, its so in future PRs i can pass `params={}` and just pass snuba_params, I'll then follow up with another PR where i'll start deleting `params={}` everywhere

src/sentry/snuba/metrics_enhanced_performance.py
```diff
@@ -21,27 +22,31 @@
...
+    sample: float | None = None,
+    has_metrics: bool = False,
     use_metrics_layer: bool = False,
+    skip_tag_resolution: bool = False,
+    extra_columns: list[Column] | None = None,
     on_demand_metrics_enabled: bool = False,
-    on_demand_metrics_type=None,
-    fallback_to_transactions=False,
+    on_demand_metrics_type: MetricSpecType | None = None,
+    fallback_to_transactions: bool = False,
```
wmak, line 49: Just typing these args since i'm in this neck of the woods anyways

mjq (approved):

Zylphrex (approved):
This pattern
```python
if len(params) == 0 and snuba_params is not None:
    params = snuba_params.filter_params
```
still seems a little strange because you don't always pass `snuba_params` into the query builders in those functions and still rely on `params` in some places. But this exposes `snuba_params` in all these places in preparation for removing `params` so this is good. Just feels like it adds another step of fixing all these places instead of just fixing it now.

PR #74785, opened 2024-07-23
ref(breadcrumbs): Address early feedback

leeandher:
This PR addresses a wave a feedback for the new breadcrumbs design
- Larger, darker text for descriptions
- Monospace font and preserving whitespace for description
- Show message crumbs as blue
- Remove Sort/Filter on issue details
- Allow categories to contain upper case letters
- Unknown categories will be unchanged
- Absolute timestamp is being kept as the default
- Respect sort order on summary crumbs,

image

**todo**
- [x] Fix tests 
- [ ] Fix virtualization (deferring to future PR)
- [x] Remove text from control on main page

### DIFF
--- a/static/app/components/timeline/index.tsx
+++ b/static/app/components/timeline/index.tsx
@@ -73,7 +73,7 @@ const Row = styled('div')<{showLastLine?: boolean}>`
   &:last-child {
     margin-bottom: 0;
     /* Show/hide connecting line from the last element of the timeline */
-    background: ${p => (p.showLastLine ? 'transparent' : p.theme.background)} !important;
+    background: ${p => (p.showLastLine ? 'transparent' : p.theme.background)};
   }
 `;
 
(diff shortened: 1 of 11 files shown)


static/app/components/timeline/index.tsx
```diff
@@ -87,7 +72,8 @@ const Row = styled('div')`
   }
   &:last-child {
     margin-bottom: 0;
-    background: ${p => p.theme.background};
+    /* Show/hide connecting line from the last element of the timeline */
+    background: ${p => (p.showLastLine ? 'transparent' : p.theme.background)} !important;
```
malwilley, line 76: Is there a way to avoid this `!important`?

malwilley (approved):

static/app/components/timeline/index.tsx
leeandher: Sure is, I actually refactored it to remove it and then forgot to do that, thanks!

PR #74786, opened 2024-07-23
feat(uptime): Enforce a basic quota of 1 uptime monitor per organization

wedamija:
This limits each organization to 1 uptime monitor. We're not using the official quota system yet, but will just keep track of how many uptime monitors exist in the org.


src/sentry/tasks/post_process.py
```diff
@@ -1515,6 +1514,8 @@ def detect_new_escalation(job: PostProcessJob):
 
 
 def detect_base_urls_for_uptime(job: PostProcessJob):
+    from sentry.uptime.detectors.detector import detect_base_url_for_project
```
wedamija, line 1517: I've been running into a few import issues, seems simpler to just import inside the function here since no other systems refer to uptime directly

davidenwang (approved):

Issue #74276, opened 2024-07-15
Traces: Add definitions for all static filter keys


malwilley:
## Background

Many of the search keys on the traces page do not have field definitions. These field definitions are how the search bar knows what type the filters are (i.e. string, date, duration, etc) and what to display for descriptions. Without this info, the search is less useful, as seen in the screenshots below.

The missing `valueType` is a blocker for the new search component. It relies on the type config in order to help create the correct filter values.

![CleanShot 2024-07-15 at 11 50 45](https://github.com/user-attachments/assets/67566a89-aed7-4be0-9cf1-9830e3676979)

![CleanShot 2024-07-15 at 11 49 09](https://github.com/user-attachments/assets/4355a34b-f10b-4182-a327-ccbe42b6ce88)

## Requirements

- Add definitions for search keys in [/app/utils/fields/index.ts](https://github.com/getsentry/sentry/blob/c40b88517bd0f342343beb6ade4dbeb8f705d21a/static/app/utils/fields/index.ts). These definitions should include a description and the correct value type.
- Update `getFieldDefinition()` to return trace search definitions
- Add a `fieldDefinitionGetter` prop to the current search bar to use the new definitions

When done, the search bar should show descriptions and the correct autocomplete operators.

malwilley:
Done in https://github.com/getsentry/sentry/pull/74787

PR #74787, opened 2024-07-23
feat(traces): Add definitions for span field search keys

0Calories:
Adds definitions for span search fields. This affects mainly the search in the Traces page, but also several searchbars in Insights modules that allow for searching on spans.

Initially, I planned on referencing `SpanIndexedField` and `SpanMetricsField` when setting the definitions, but importing it results in a circular dependency due to the file they reside in also importing from `static/app/utils/fields/index.ts`. I decided to create a `TraceFields` enum instead which results in some code duplication, but the alternative is a massive and fairly messy refactor.

Related to https://github.com/getsentry/sentry/issues/74276

image

Side note; I noticed that most of these definitions are missing in our documentation, we should update them there as well once we've decided on adequate descriptions for these fields.

### DIFF
--- a/static/app/utils/discover/fields.tsx
+++ b/static/app/utils/discover/fields.tsx
@@ -11,7 +11,7 @@ import {
   SESSIONS_OPERATIONS,
 } from 'sentry/views/dashboards/widgetBuilder/releaseWidget/fields';
 import {STARFISH_FIELDS} from 'sentry/views/insights/common/utils/constants';
-import {STARFISH_AGGREGATION_FIELDS} from 'sentry/views/insights/types';
+import {STARFISH_AGGREGATION_FIELDS} from 'sentry/views/insights/constants';
 
 import {
   AGGREGATION_FIELDS,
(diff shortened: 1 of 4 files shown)


gggritso (commented):
>  importing it results in a circular dependency

Ouch! If I'm reading this right, I think you can resolve the circular dependency by extracting `STARFISH_AGGREGATION_FIELDS` into its own file! That will eliminate the imports for 

```tsx
import type {AggregationOutputType} from 'sentry/utils/discover/fields';
import {DiscoverDatasets} from 'sentry/utils/discover/types';
import {FieldKind, FieldValueType} from 'sentry/utils/fields';
```

You can also delete `StarfishDatasetFields`, it's unused. That will remove the import for

```tsx
import {DiscoverDatasets} from 'sentry/utils/discover/types';
```

That should remove the circular dependency

0Calories:
Thanks for the suggestion @gggritso, that refactor was a lot cleaner

gggritso (approved):
Nice! Although, I have a nit. "i.e.," and "e.g.,"` are different! 

Correct:

>i.e. the tables involved in a `db` span or the host name in an `http` span'

Not correct, should be "e.g.,"

> i.e `http.client`, `middleware`
> i.e `cache`, `db`, `http`, etc.

See [Chicago Manual of Style](https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Abbreviations/faq0047.html) for deets

Zylphrex (approved):
Nice! 🎉 This is a good start

0Calories:
@gggritso learned something new today, thanks! But this also made me realize we have a lot of field definitions (aside from the ones I added) that incorrectly use i.e. 😅

PR #74788, opened 2024-07-23
Make sure end_timestamp_precise and organization_id are sent to snuba

colin-sentry:
… from tests

This lets us remove some Optionals from snuba


phacops:
Adding the fields here is what really allows us to remove the `Optional` from Snuba.

https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/snuba-spans.v1.schema.json#L101

phacops (changes requested):
It'd be wise to test this against your consumer branch before we merge.

phacops (approved):

PR #74791, opened 2024-07-23
feat(toolbar): Add a badge showing the number of active alerts

### LINKED ISSUE #74568: [DevToolbar] Update Alerts panel button to have an indicator/count of how many alerts are currently active
Depends on https://github.com/getsentry/sentry/issues/74567

We should do something to preiodically refetch data so that the counter is always up to date (useQuery has cache timeout options for this), then people will know if there are active alerts right now, and can stop what they're doing to investigate immediatly.

ryan953:
The badge is looking like this:

SCR-20240723-nkva

also, i'm pretty happy about getting the types in shape to be able to use `select` from useQuery. This makes things easier going forward.

Fixes https://github.com/getsentry/sentry/issues/74568

### DIFF
--- a/static/app/components/devtoolbar/styles/global.ts
+++ b/static/app/components/devtoolbar/styles/global.ts
@@ -34,6 +34,26 @@ export const globalCss = css`
     --blue200: rgba(60, 116, 221, 0.5);
     --blue100: rgba(60, 116, 221, 0.09);
 
+    --green400: #207964;
+    --green300: #2ba185;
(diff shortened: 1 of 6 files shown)


static/app/components/devtoolbar/styles/global.ts
ryan953, line 1: I used `--red400` but added all the other missing colors just because.

michellewzhang (approved):

PR #74792, opened 2024-07-23
ref: add threadId to profile link and initialize the profile to thread id

JonasBa:
Pass tid from spans and transactions so that it is initialized to the correct value


Zylphrex (commented):
👍 There's some failing tests but otherwise LGTM

Zylphrex (approved):

PR #74793, opened 2024-07-23
ref(control_silo): Consolidate integrations api endpoints and tests

Christinarlong:
Moves all integrations related endpoints to new space sentry/integrations/api/endpoints and updates any import paths. In addition to any integrations endpoint tests. Includes import shims for modules that are imported in getsentry

ref(#73859)


src/sentry/api/endpoints/codeowners/external_actor/team_details.py
```diff
@@ -1,81 +1,5 @@
-import logging
-from typing import Any
+from sentry.integrations.api.endpoints.codeowners.external_actor.team_details import (
```
markstory, line 1: Do we need all these directories? We could flatten this to `sentry.integrations.api.endpoints.codeowners` and rename endpoint module to `external_actor_team_details`.

src/sentry/api/urls.py
```diff
@@ -113,6 +112,57 @@
...
+    OrganizationIntegrationDetailsEndpoint,
+    OrganizationIntegrationRequestEndpoint,
+    OrganizationIntegrationsEndpoint,
+    OrganizationPluginsConfigsEndpoint,
+    OrganizationPluginsEndpoint,
+)
+from sentry.integrations.api.endpoints.internal.integration_proxy import (
+    InternalIntegrationProxyEndpoint,
+)
+from sentry.integrations.api.endpoints.organization.organization_code_mapping_codeowners import (
```
markstory, line 136: These endpoints could be moved up to `sentry.integrations.api.endpoints` There aren't *that* many endpoints within integrations yet.

src/sentry/api/endpoints/codeowners/external_actor/team_details.py
Christinarlong: 👍 that sounds cleaner

markstory (approved):
Looking good :clap:

PR #74794, opened 2024-07-23
ref(similarity): Remove ids from being saved in redis in backfill

jangjodi:
Remove use of redis to save group/project ids in similarity backfill
We don't need to save it since re-running for projects will fast since most of the work is done in bigtable and seer


JoshFerge (approved):
awesome!!

PR #74796, opened 2024-07-23
fix(replay): Remove browser from table if all mobile replays

### LINKED ISSUE #74747: [Mobile Replay] Issue Detail Tab: Browser showing as `null null` with `<>` icon.
Image

For mobile projects we should hide the browser tab. Similar to how we did for the Replay Index.

Image

Example:

https://sentry-sdks.sentry.io/issues/5607602092/replays/?project=4507596011798528&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&selected_replay_index=0&statsPeriod=30d&stream_index=0

c298lee:
Fixes https://github.com/getsentry/sentry/issues/74747

### DIFF
--- a/static/app/views/issueDetails/groupReplays/groupReplays.tsx
+++ b/static/app/views/issueDetails/groupReplays/groupReplays.tsx
@@ -21,6 +21,7 @@ import useReplayReader from 'sentry/utils/replays/hooks/useReplayReader';
 import {useLocation} from 'sentry/utils/useLocation';
 import useOrganization from 'sentry/utils/useOrganization';
 import useUrlParams from 'sentry/utils/useUrlParams';
+import useAllMobileProj from 'sentry/views/replays/detail/useAllMobileProj';
 import ReplayTable from 'sentry/views/replays/replayTable';
 import {ReplayColumn} from 'sentry/views/replays/replayTable/types';
 import type {ReplayListLocationQuery, ReplayListRecord} from 'sentry/views/replays/types';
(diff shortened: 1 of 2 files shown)


ryan953 (approved):

static/app/views/issueDetails/groupReplays/groupReplays.tsx
```diff
@@ -80,7 +88,7 @@ function GroupReplays({group}: Props) {
           isFetching={isFetching}
           replays={[]}
           sort={undefined}
-          visibleColumns={VISIBLE_COLUMNS}
+          visibleColumns={allMobileProj ? VISIBLE_COLUMNS_MOBILE : VISIBLE_COLUMNS}
```
michellewzhang, line 91: this ternary seems to be used a lot throughout this file, can we make it a constant up top?

static/app/views/issueDetails/groupReplays/groupReplays.tsx
```diff
@@ -52,6 +62,7 @@ function GroupReplays({group}: Props) {
     location,
     organization,
   });
+  const {allMobileProj} = useAllMobileProj();
```
michellewzhang, line 65: oo sorry, i meant, can we move the `visibleColumns` definition to under here so that we don't need a function? (we don't need to pass in `allMobileProj` then, can just case directly on `allMobileProj`

michellewzhang: actually, what do you think is better🤔 because i see with that method we'd have to define the const twice

c298lee: there's three functional components using `VISIBLE_COLUMNS` in this file so I don't think we can do it without a function :(

michellewzhang (approved):

PR #74802, opened 2024-07-23
ref(feedback): add visualization for non-image attachments and allow download

### LINKED ISSUE #74731: [userFeedback] attachments that are not screenshots are very hard for support to download
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

add non image attachments such as csv/text to feedback

also see related issue i opened in the wrong subproject
https://github.com/getsentry/sentry-javascript/issues/13009#issuecomment-2244497464

### Expected Result

- display images inline, offer download for non-image files
- **or** offer one big zip for download
- **or** offer download links for each item including images

### Actual Result

Image

it tries to create an image preview of a file that is not an image 

### Product Area

User Feedback

### Link

https://zentra.sentry.io/feedback/?feedbackSlug=zentra-frontend%3A5633115988&project=4505371415216128&referrer=feedback_list_page&statsPeriod=30d

### DSN

_No response_

### Version

_No response_

michellewzhang:
- closes https://github.com/getsentry/sentry/issues/74731
- adds support for downloading non-image files
- attempt to load all attachments as images first. if `` gives an error, then send that to our screenshot renderer, and tell it to use the default "no preview found" renderer instead.

before: 
SCR-20240723-nrth

after: "no preview found" text along with option to download the item:
SCR-20240723-nrjc

SCR-20240723-nuzb

all the attachments show up in the modal as well and work as expected with the pagination.
SCR-20240723-obyr
SCR-20240723-obxj

works with the example @c298lee brought up where the `mimetype` was incorrectly set:
SCR-20240724-jsua

### DIFF
--- a/static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
+++ b/static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
@@ -50,7 +50,7 @@ export default function FeedbackScreenshot({
       {onClick ? <StyledImageButton onClick={onClick}>{img}</StyledImageButton> : img}
     </StyledPanel>
   ) : (
-    <File>
+    <File onClick={onClick}>
       <NoPreviewFound>
         <IconImage />
         {t('No preview found')}
--- a/static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
+++ b/static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
@@ -27,7 +27,8 @@ export default function FeedbackScreenshot({
   onClick,
 }: Props) {
   const [isLoading, setIsLoading] = useState(true);
-  const isImage = screenshot.mimetype.includes('image');
+  const isImage =
+    screenshot.mimetype.includes('image') || screenshot.mimetype.includes('png');
 
   const img = (
     <StyledImageVisualization
--- a/static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
+++ b/static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
@@ -27,8 +27,9 @@ export default function FeedbackScreenshot({
   onClick,
 }: Props) {
   const [isLoading, setIsLoading] = useState(true);
-  const isImage =
-    screenshot.mimetype.includes('image') || screenshot.mimetype.includes('png');
+  // since we can't trust mimetype, we'll try to load all attachments as images
+  // if it fails, then set that here, and render a default preview instead.
+  const [imgLoadError, setImgLoadError] = useState(false);
 
   const img = (
     <StyledImageVisualization
@@ -36,12 +37,17 @@ export default function FeedbackScreenshot({
       orgId={organization.slug}
       projectSlug={projectSlug}
       eventId={screenshot.event_id}
-      onLoad={() => setIsLoading(false)}
-      onError={() => setIsLoading(false)}
+      onLoad={() => {
+        setIsLoading(false);
+      }}
+      onError={() => {
+        setIsLoading(false);
+        setImgLoadError(true);
+      }}
     />
   );
 
-  return isImage ? (
+  return !imgLoadError ? (
     <StyledPanel className={className}>
       {isLoading && (
         <StyledLoadingIndicator>
--- a/static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
+++ b/static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
@@ -4,6 +4,11 @@ import styled from '@emotion/styled';
 import ImageVisualization from 'sentry/components/events/eventTagsAndScreenshot/screenshot/imageVisualization';
 import LoadingIndicator from 'sentry/components/loadingIndicator';
 import Panel from 'sentry/components/panels/panel';
+import TextOverflow from 'sentry/components/textOverflow';
+import {Tooltip} from 'sentry/components/tooltip';
+import {IconImage} from 'sentry/icons';
+import {t} from 'sentry/locale';
+import {space} from 'sentry/styles/space';
 import type {EventAttachment, Organization, Project} from 'sentry/types';
 
 type Props = {
@@ -22,6 +27,8 @@ export default function FeedbackScreenshot({
   onClick,
 }: Props) {
   const [isLoading, setIsLoading] = useState(true);
+  const isImage = screenshot.mimetype.includes('image');
+
   const img = (
     <StyledImageVisualization
       attachment={screenshot}
@@ -33,7 +40,7 @@ export default function FeedbackScreenshot({
     />
   );
 
-  return (
+  return isImage ? (
     <StyledPanel className={className}>
       {isLoading && (
         <StyledLoadingIndicator>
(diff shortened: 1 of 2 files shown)


static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
```diff
@@ -42,6 +49,20 @@ export default function FeedbackScreenshot({
...
+  ) : (
+    <File onClick={onClick}>
+      <NoPreviewFound>
+        <IconImage />
+        {t('No preview found')}
+      </NoPreviewFound>
+      <Tooltip position="right" title={t('Click to download')}>
+        <FileDownload
+          href={`/api/0/projects/${organization.slug}/${projectSlug}/events/${screenshot.event_id}/attachments/${screenshot.id}/?download=1`}
+        >
```
ryan953, line 61: dogfooding, but i wonder if we should set an extra attr so this doesn't get detected as a dead click

ryan953 (approved):

static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
michellewzhang: added a `className` so we can add it to our `slowClickIgnoreSelectors` if we want

ryan953 (approved):

c298lee:
Some images that were previewed before are no longer previewable :( 
Example: https://sentry-emerging-tech.sentry.io/feedback/?feedbackSlug=catherine-test-a7%3A5530677928&mailbox=ignored&project=4506508390105088&referrer=feedback_list_page&statsPeriod=90d

static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
```diff
@@ -22,6 +27,8 @@ export default function FeedbackScreenshot({
   onClick,
 }: Props) {
   const [isLoading, setIsLoading] = useState(true);
+  const isImage = screenshot.mimetype.includes('image');
```
bruno-garcia, line 30: based on @c298lee's review I was wonder what could be. Mime type isnt' a required field on envelopes. So might be wroth having a fallback to file extension.

We could have some reusable logic in issue detail, by attachment preview

michellewzhang: ahh @bruno-garcia the case @c298lee brought up is because the mimetype is `application/png` instead of `image/png` -- maybe a wrong mimetype setting? but i'll make a case for that

michellewzhang:
> Some images that were previewed before are no longer previewable :( Example: https://sentry-emerging-tech.sentry.io/feedback/?feedbackSlug=catherine-test-a7%3A5530677928&mailbox=ignored&project=4506508390105088&referrer=feedback_list_page&statsPeriod=90d

should be fixed with the latest change

static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
bruno-garcia: got it.

What fills out the `mimetype` property? It's not then the value from attachment, is it?
Does it align with Issue details attachment preview?

bruno-garcia: `application/png` seems incorrect to me, did we add that by hand?

I think the file extension will work better with data coming from the wild. Since we expose a `addAttachment` API that doesn't require mimetype let alone users adding the right value

bruno-garcia (commented):
I added a couple more thoughts on the "outdated" thread

c298lee:
> > Some images that were previewed before are no longer previewable :( Example: https://sentry-emerging-tech.sentry.io/feedback/?feedbackSlug=catherine-test-a7%3A5530677928&mailbox=ignored&project=4506508390105088&referrer=feedback_list_page&statsPeriod=90d
> 
> should be fixed with the latest change @c298lee

I found another case where of an image that isn't previewable that previously was :( 
Preview: https://sentry-5w2s9trmc.sentry.dev/feedback/?feedbackSlug=pokemart%3A5540685980&mailbox=ignored&project=4506072436572160&referrer=feedback_list_page&statsPeriod=30d
Santry: https://sentry-test.sentry.io/feedback/?feedbackSlug=pokemart%3A5540682689&mailbox=ignored&project=4506072436572160&referrer=feedback_list_page&statsPeriod=30d

michellewzhang:
> https://sentry-test.sentry.io/feedback/?feedbackSlug=pokemart%3A5540682689&mailbox=ignored&project=4506072436572160&referrer=feedback_list_page&statsPeriod=30d

@c298lee do you know why the content mimetype is `application/octet-stream` instead of the normal image mimetype, `image/jpeg` or `image/png`? is that something the SDK is setting?

static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
michellewzhang: yeah @bruno-garcia i'm not totally sure, `application/png` doesn't seem right to me either. because it was from a test feedback, maybe @c298lee knows? maybe we manually set that at some point?

michellewzhang: i think the issue details preview renderer requires the right type to set, otherwise it returns `undefined1
https://github.com/getsentry/sentry/blob/f2003b5ed3eff14b1481fe99fa154398fa94435a/static/app/components/events/eventAttachments.tsx#L38-L56

bruno-garcia: I believe we should consider the file extension for images since the mine type is added "by hand" and can and is wrong often

c298lee:
> > https://sentry-test.sentry.io/feedback/?feedbackSlug=pokemart%3A5540682689&mailbox=ignored&project=4506072436572160&referrer=feedback_list_page&statsPeriod=30d
> 
> @c298lee do you know why the content mimetype is `application/octet-stream` instead of the normal image mimetype, `image/jpeg` or `image/png`? is that something the SDK is setting?

@michellewzhang The SDK isn't setting anything, I think the content mimetype is encoded in the attachment data that's sent and if the widget is used for screenshotting, the preview works. However, if someone uses the feedback api and adds the attachment themselves, they could set the type. The default  is `application/octet-stream`, docs here: https://docs.sentry.io/platforms/javascript/enriching-events/attachments/#uploading-attachments

joranbeasley:
> > > https://sentry-test.sentry.io/feedback/?feedbackSlug=pokemart%3A5540682689&mailbox=ignored&project=4506072436572160&referrer=feedback_list_page&statsPeriod=30d
> > 
> > 
> > @c298lee do you know why the content mimetype is `application/octet-stream` instead of the normal image mimetype, `image/jpeg` or `image/png`? is that something the SDK is setting?
> 
> @michellewzhang The SDK isn't setting anything, I think the content mimetype is encoded in the attachment data that's sent and if the widget is used for screenshotting, the preview works. However, if someone uses the feedback api and adds the attachment themselves, they could set the type. The default is `application/octet-stream`, docs here: https://docs.sentry.io/platforms/javascript/enriching-events/attachments/#uploading-attachments


If someone sets the incorrect mime type thats kind of on them ... all the examples show how to set contentType I think ... that said you could do something like the following and just be super forgiving for "image types"

```python
is_image = re.match("\.(png|jpg|gif)$",filename) or re.match("(image|png|gif|jpg|jpeg)",mimetype)
```

alternatively i think all images bytes always start with some magic numbers

static/app/components/feedback/feedbackItem/feedbackScreenshot.tsx
c298lee: I agree with Bruno, it could be added "by hand", and the `addAttachment` API defaults to `application/octet-stream` even if it's an image. However I'm not sure how well using file extension for images would work since the name of the file could also be added "by hand"

michellewzhang: @c298lee @bruno-garcia i figured out a way around casing on the `mimetype` -- instead, i'm using the image error state to tell me whether an image failed or not. we'll attempt to load all attachments as errors first, and if `` gives an error because the `src` it tries to load is not in fact an image, then we tell that to our image renderer and render the default preview instead.

c298lee (approved):

aliu39 (approved):

ryan953 (approved):

PR #74807, opened 2024-07-23
ref: adjust typing so rate_limits can be a callable

asottile-sentry:
this fixes an error in mypy 1.11 -- it also fixes an error in a currently-ignored file which has a callable `rate_limits`


joshuarli (approved):

tests/sentry/ratelimits/utils/test_get_rate_limit_value.py
```diff
@@ -88,31 +88,3 @@ class ChildEndpoint(ParentEndpoint):
         assert get_rate_limit_value(
             "GET", RateLimitCategory.IP, rate_limit_config
         ) == get_default_rate_limits_for_group("foo", RateLimitCategory.IP)
-
-    def test_multiple_inheritance(self):
```
asottile-sentry, line 92: this test wasn't actually testing anything beyond "does python do inheritance" so I removed it

PR #74808, opened 2024-07-23
fix(query-builder): Correctly place selection background behind only text portion of input

malwilley:
Uses the invisible text div to correctly place the selection background. Prior to this, we were changing the background of the input, which was sometimes much larger that its text contents (the last input takes up the remaining space for usability reasons). 

Before:

CleanShot 2024-07-23 at 16 01 18@2x

After:

CleanShot 2024-07-23 at 15 58 59@2x


roggenkemper (approved):
very happy selecting all no longer highlights the entire search bar!

PR #74811, opened 2024-07-23
feat(issue-details): Add event actions to updated event navigation

roggenkemper:
this pr adds in the actions for copying event id + a dropdown for json and copy the link to the event, which are no longer buttons in the updated design


malwilley (approved):
Looks good! No need to do so in this PR, but we should make sure to get this looking good on narrow viewports:

CleanShot 2024-07-24 at 10 47 09@2x

PR #74813, opened 2024-07-23
feat(replay): add feedback button to HE modal

### LINKED ISSUE #74732: [Hydration Errors] Embed Feedback Widget in the the Diff Viewer
## Opportunity 

The Issues team has embedded the Feedback widget on new and improved UI components and it has been successful in garnering feedback from customers specifically on those new components (I see them come thru on the #all-user-feedback Slack channel).

Image

Image

^ Examples above for event highlights and breadcrumbs.

## Solution

Embed the Feedback widget in the Diff Viewer Modal as opposed to the discussion from last year to garner feedback from our user base in a quick and easy manner on our new HE issues.

Image

aliu39:
Fixes https://github.com/getsentry/sentry/issues/74732

### DIFF
--- a/static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
+++ b/static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
@@ -1,3 +1,4 @@
+import {useEffect, useState} from 'react';
 import styled from '@emotion/styled';
 
 import type {ModalRenderProps} from 'sentry/actionCreators/modal';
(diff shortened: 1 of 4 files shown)


static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
```diff
@@ -25,40 +28,37 @@ export default function ReplayComparisonModal({
...
-            label={t('Discussion')}
-            title={null}
-            analyticsEventKey="replay.details-hydration-discussion-clicked"
-            analyticsEventName="Replay Details Hydration Discussion Clicked"
-            priority="primary"
+      <AsyncSDKIntegrationContextProvider outerContext={integrationsContext}>
+        <Header closeButton>
+          <ModalHeader>
+            <h4>
+              Hydration Error!!
```
michellewzhang, line 39: exclamations just for fun?🤓

aliu39:
Some notes for debugging:
This is a bit tricky since we're using a button in a modal, which we don't do anywhere else in sentry. At least I couldn't find any examples. 
I got the button working, but both the diff modal and feedback widget stop responding after that. Console shows `Autofocus processing was blocked because a document already has a focused element.` There's some kind of conflict between the modal and widget's focus.

Uploading Screen Recording 2024-07-23 at 4.29.43 PM.mov…

aliu39:
References I found online: 

https://stackoverflow.com/questions/58830133/autofocus-on-input-when-opening-modal-does-not-work-react-bootstrap

Google AI:
> The HTML5 autofocus attribute can cause issues in some browsers, so custom JavaScript can be used to achieve the same effect. For example, PMACS Frontend uses custom JavaScript to autofocus on an element when a modal opens. To use this feature, the data-autofocus attribute can be set instead of the HTML5 autofocus attribute. This attribute can also be used for elements inside a Bootstrap modal.

Maybe we shouldn't use `autofocus` for the widget? But this would require changing the SDK, right? Is there a way to solve this by changing the modal?

static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
aliu39: This was from a test to make sure I had the right component lol

ryan953:
the problem with the focus is related to this FocusTrap feature:

https://github.com/getsentry/sentry/blob/5914796731d8a32ee7789b0ff31b392e441e1a53/static/app/components/globalModal/index.tsx#L148-L154

aliu39:
Discussed in a team sync a solution, to move the feedback's shadowRoot to inside the modal element. Depends on https://github.com/getsentry/sentry-javascript/pull/13047/files.

Unfortunately this is still not working... need to look deeper into what focusTrap is doing here.

PR #74815, opened 2024-07-23
ref(control_silo): Consolidate integrations api serializers and tests

Christinarlong:
Moves all integrations related serializers and their tests to sentry/integrations/api/serializers.

ref(#73859)


src/sentry/integrations/api/serializers/models/integration_feature.py
```diff
@@ -10,18 +10,21 @@
...
-            IntegrationFeature.objects.get_descriptions_as_dict(item_list) if has_target else {}
+            IntegrationFeature.objects.get_descriptions_as_dict(list(item_list))
+            if has_target
+            else {}
         )
         return {item: {"description": description_attrs.get(item.id)} for item in item_list}
 
-    def serialize(self, obj: IntegrationFeature, attrs, user: User, has_target: bool = True):
+    def serialize(self, obj: IntegrationFeature, attrs: Mapping[Any, Any], user: User, **kwargs):
+        has_target = kwargs.get("has_target", True)
```
markstory, line 27: Could we keep the formal parameter here as well?

markstory (approved):

PR #74823, opened 2024-07-24
feat(issue-priority): Check flag before calculating issue severity

### LINKED ISSUE #74757: Apply feature flag to first event severity and priority calculations

snigdhas:
Check the seer-based-priority flag to determine if we need to calculate first-event severity for incoming events. If the flag is not set, we default to the current behavior which determines severity for python and javascript projects.

Once the flag is GA, we should remove this platform check and only calculate the severity if the flag (subscription-based) is enabled for an org.

Fixes https://github.com/getsentry/sentry/issues/74757


src/sentry/event_manager.py
```diff
@@ -2374,17 +2374,24 @@ def _get_severity_metadata_for_group(
         metrics.incr("issues.severity.seer_killswitch_enabled")
         return {}
 
+    seer_based_priority_enabled = features.has(
+        "organizations:seer-based-priority", event.project.organization, actor=None
+    )
     feature_enabled = features.has("projects:first-event-severity-calculation", event.project)
-    if not feature_enabled:
+    if not seer_based_priority_enabled and not feature_enabled:
```
jangjodi, line 2381: nit: you can do `not (A or B)` to make this a bit cleaner

jangjodi (approved):

PR #74825, opened 2024-07-24
ref(grouping): Strip querystrings from stacktrace filenames

lobsterkatie:
In cases where a stacktrace frame's `filename` is the basename of a URL, we sometimes end up with querystring nonsense on the end of it. This both makes the stacktrace hard to read and forces Seer to tokenize a whole bunch of meaningless stuff. Here's a recent example of such a stacktrace:

```python
"frames": [
    {
      "filename": "index.html?__geo_region=jp&loc=eyjrawqioiiydks4rjniyvrlwekwovb5yxdrrno4iiwiywxnijoirvmyntyifq.eyjzdwiioijhmvpdmevvnuuilcjhdwqioijndxj1z3vydsisimnvdw50cnkioijkucisimnyzwf0zwqioje3mtk4otu0mzisimlzcyi6imcxmjmtyxv0acisimn1cnjlbmn5ijoislbziiwizxhwijoxnzixmjy5nzuylcjyzwdpb24ioijkucisimxhbmcioijqysisimlhdci6mtcymta5njk1miwianrpijoicgnfnelbovpovel1cfrhsllncemyce9lwij9.wefd0fvomovr_gjrcquzatrsmstgrvzqew7uhuyiibajhas7m_hyceqkigikwyybvlsqxhdqrwywsrxqthmjeq&lang=jp&platform=jorp1&mode=0",
      "function": "t",
      "context_line": '<!DOCTYPE HTML><html><head><meta charset="utf-8"><title></title><meta name="viewport" content="width=device-width,initial-scale=1,minimum-sc {snip}'
    },
    # and about 50 more frames just like this
]
```

This PR therefore adds querystring stripping for JavaScript events to `normalize_stacktraces_for_grouping`, along with a metric tracking how often we do it.

It's technically possible that this could end up occasionally biting us - if the querystring is both meaningful and enough to change grouping, and now we're ditching it - but given that most often querystrings show up in the names of HTML files, which tend to contain ungroupable garbage in any case, the risk seems low. If we discover it's causing problems, we can tighten the criteria for filename stripping beyond just "be from a JS event" and "have a querystring."

### DIFF
--- a/src/sentry/stacktraces/processing.py
+++ b/src/sentry/stacktraces/processing.py
@@ -4,12 +4,14 @@ import logging
 from collections.abc import Callable, Mapping, MutableMapping, Sequence
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any, NamedTuple
+from urllib.parse import urlparse
 
 import sentry_sdk
 
 from sentry.models.project import Project
 from sentry.models.release import Release
 from sentry.stacktraces.functions import set_in_app, trim_function_name
+from sentry.utils import metrics
 from sentry.utils.cache import cache
 from sentry.utils.hashlib import hash_values
 from sentry.utils.safe import get_path, safe_execute
@@ -298,7 +300,7 @@ def normalize_stacktraces_for_grouping(
 ) -> None:
     """
     Applies grouping enhancement rules and ensure in_app is set on all frames.
-    This also trims functions if necessary.
+    This also trims functions and pulls query strings off of filenames if necessary.
     """
 
     stacktrace_frames = []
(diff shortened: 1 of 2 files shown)


src/sentry/stacktraces/processing.py
```diff
@@ -323,10 +325,21 @@ def normalize_stacktraces_for_grouping(
     # otherwise stored in `function` to not make the payload larger
     # unnecessarily.
     with sentry_sdk.start_span(op=op, description="iterate_frames"):
+        stripped_querystring = False
         for frames in stacktrace_frames:
             for frame in frames:
                 _update_frame(frame, platform)
 
+                if platform == "javascript":
+                    parsed_filename = urlparse(frame.get("filename", ""))
```
JoshFerge, line 334: is there any possible filename that could cause this to throw?

JoshFerge (approved):

src/sentry/stacktraces/processing.py
lobsterkatie: I guess technically it could raise a `ValueError`: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse

That said, I don't think the SDK would/could actually send the kind of corrupt values it would take to cause that. Can't hurt to wrap it in a try-catch anyway, though.

JoshFerge: 👍🏼

PR #74833, opened 2024-07-24
fix(getting-started): Add missing Node SDK import to verification snippet

Lms24:
This PR just adds a missing `require` call to the verification snippet
in the Node onboarding. Came up in internal feedback.

Also, I shortened the errors-only verification snippet. I don't think we
need the `setTimeout` call in there.

### DIFF
--- a/static/app/gettingStartedDocs/node/node.tsx
+++ b/static/app/gettingStartedDocs/node/node.tsx
@@ -103,6 +103,8 @@ const onboarding: OnboardingConfig = {
           language: 'javascript',
           code: isPerformanceSelected
             ? `
+const Sentry = require("@sentry/node");
+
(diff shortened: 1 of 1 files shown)


priscilawebdev (approved):

static/app/gettingStartedDocs/node/node.tsx
```diff
@@ -114,13 +116,13 @@ Sentry.startSpan({
...
-    foo();
-  } catch (e) {
-    Sentry.captureException(e);
-  }
-}, 99);`,
+const Sentry = require("@sentry/node");
+
+try {
+  foo();
+} catch (e) {
```
priscilawebdev, line 123:
```suggestion
} catch {
```

PR #74835, opened 2024-07-24
feat(profiling): remove profile id from transaction id endpoint

viglia:
related to https://github.com/getsentry/vroom/pull/491

@JonasBa @Zylphrex , please could you confirm if this is still used anywhere in the front-end?

If not I can simply merge this and remove it altogether, otherwise I'll refactor this to read directly from Sentry and we can still proceed with removing the endpoint in `vroom` (see linked PR)


Zylphrex (approved):
This should be safe to remove now

PR #74837, opened 2024-07-24
fix(metrics): Remove colon when single value

priscilawebdev:
**Before:**

image

**After:**

image

### DIFF
--- a/static/app/views/metrics/summaryTable.tsx
+++ b/static/app/views/metrics/summaryTable.tsx
@@ -587,6 +587,6 @@ const Row = styled('div')`
   }
 `;
 
-const SerieNamePrefix = styled('span')<{uppercaseText?: boolean}>`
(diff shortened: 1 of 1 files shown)


shellmayr (approved):

static/app/views/metrics/summaryTable.tsx
```diff
@@ -27,6 +27,31 @@ import useOrganization from 'sentry/utils/useOrganization';
...
+  const prefix = seriesName.split(':')[0];
+  const sufix = seriesName.split(':')[1] ?? null;
+
+  return (
+    <TextOverflow>
+      <SerieNamePrefix
+        uppercaseText={hasMetricsNewInputs(organization) && !isSingleSeries}
+      >
+        {prefix}
+        {sufix && ':'}
```
priscilawebdev, line 48: this is a string, so even if it is '0' it will pass 😉

PR #74838, opened 2024-07-24
fix(onboarding): Add 'other' platform to all platforms list

priscilawebdev:
**Before:**

https://github.com/user-attachments/assets/a8a27ca3-b6cb-41b5-8c31-7cfd52f96c3e

**After:**

https://github.com/user-attachments/assets/a02622ff-2d2d-4633-8b81-893583b5b942

### DIFF
--- a/static/app/components/platformPicker.spec.tsx
+++ b/static/app/components/platformPicker.spec.tsx
@@ -106,4 +106,16 @@ describe('PlatformPicker', function () {
       expect(platform).toHaveTextContent(alphabeticallyOrderedPlatformNames[index]);
     });
   });
+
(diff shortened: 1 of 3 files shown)


matejminar:
Could we make "other" not visible in the list of All? So basically, only show it when you search for something non-existent and click "Other" in the warning. We want to discourage creating Other projects.

priscilawebdev:
@matejminar If a user doesn't find a platform under the 'all' tab, I think we should provide an alternative on the same page without requiring extra typing. However, I understand your point and am happy to include your suggestion as well

matejminar (approved):

Issue #74333, opened 2024-07-16
Alert wizard copy and illustration for span based metrics


ArthurKnaus:
Custom Metrics should have `(deprecated)` next to them if you have rules feature flag enabled.

![Image](https://github.com/user-attachments/assets/a6bd58a9-8b2f-4fc0-b17f-51c4fb572b15)

Jesse-Box:
Here's the revised svg
![Image](https://github.com/user-attachments/assets/5107ca78-cc3a-4ea8-8339-ae986470afcd)

PR #74843, opened 2024-07-24
ref(metrics): Add deprecated tag and other updates

priscilawebdev:
**Before:**

image
image

**After:**
image
image

### Note

- There are still a few instances where we display 'Custom Metric' without the deprecated tag, such as in the dropdown list when creating an alert. Should we add the deprecated tag to all of these instances?
image

- This PR does not close the issue yet, as we still need to include the link to the docs in the description.

part of https://github.com/getsentry/sentry/issues/74333


static/app/views/alerts/wizard/index.tsx
```diff
@@ -156,11 +158,18 @@ function AlertWizard({organization, params, location, projectId}: AlertWizardPro
...
+                          alertType,
+                          AlertWizardAlertNames[alertType],
+                          alertType === 'custom_metrics' &&
+                          hasCustomMetricsExtractionRules(organization) ? (
+                            <Tag type="warning">{t('deprecated')}</Tag>
+                          ) : null,
+                        ];
                       })}
-                      onChange={handleChangeAlertOption}
+                      onChange={option => handleChangeAlertOption(option as AlertType)}
```
priscilawebdev, line 172: typescript was not happy with the change

vgrozdanic (approved):

PR #74847, opened 2024-07-24
chore: separate salt for slack integration signatures, step 1

oioki:
Slack integration features will use a separate salt value for signing payloads.
Uses https://github.com/getsentry/sentry/pull/74839.

### DIFF
--- a/src/sentry/integrations/slack/views/__init__.py
+++ b/src/sentry/integrations/slack/views/__init__.py
@@ -10,6 +10,8 @@ from sentry.utils.signing import sign
 from sentry.web.decorators import EndpointFunc
 from sentry.web.helpers import render_to_response
 
+SALT = "sentry-slack-integration"
+
 
 def never_cache(view_func: EndpointFunc) -> EndpointFunc:
     """TODO(mgaeta): Remove cast once Django has a typed version."""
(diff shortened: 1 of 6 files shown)


src/sentry/integrations/slack/views/__init__.py
```diff
@@ -19,7 +21,7 @@ def never_cache(view_func: EndpointFunc) -> EndpointFunc:
 
 def build_linking_url(endpoint: str, **kwargs: Any) -> str:
     """TODO(mgaeta): Remove cast once sentry/utils/http.py is typed."""
-    url: str = absolute_uri(reverse(endpoint, kwargs={"signed_params": sign(**kwargs)}))
+    url: str = absolute_uri(reverse(endpoint, kwargs={"signed_params": sign(salt=SALT, **kwargs)}))
```
RyanSkonnord, line 24: Is there any chance of this causing requests to fail if they happen mid-deployment? That is, a linking URL gets emitted signed by the new salt, but the webhook request lands on a host that's still trying to unsign it with the old salt, or vice versa.

oioki: You are absolutely right. There are two cases:
1. Sign with the old salt, unsign with the new salt -- covered by previous PR https://github.com/getsentry/sentry/pull/74839 (compatibility during transitional period)
2. Sign with the new salt, unsign with the old salt -- however, this one would fail during deployments. I have fixed the code so that all hosts will be ready to unsign with the new one, but don't sign payloads with new salt just yet. I will add new salt to `sign` function in the follow-up PR.

RyanSkonnord (approved):

PR #74854, opened 2024-07-24
ref: remove unused get_annotations plugin method

asottile-sentry:
last referenced in 7a31fa2cabb0eec202f9a6c562cb711c1eef4333


IanWoodard (approved):

asottile-sentry:
seems it's _intended_ to use for "v2 plugins" but was never migrated to -- going to fix the signatures instead in a different PR

PR #74860, opened 2024-07-24
ref(similarity): Add retries with backoff to seer call in backfill

jangjodi:
Add retries with backoff to seer call in similarity backfill if the service is unavailable
Does not do exponential retries for other exceptions

### DIFF
--- a/src/sentry/tasks/embeddings_grouping/utils.py
+++ b/src/sentry/tasks/embeddings_grouping/utils.py
@@ -371,13 +371,9 @@ def _make_seer_call(
             exceptions=ServiceUnavailable,
         )
     except ServiceUnavailable:
-        extra = {
-            "project_id": project_id,
-            "error": "Service Unavailable",
-        }
         logger.exception(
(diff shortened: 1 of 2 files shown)


src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -358,33 +359,63 @@ def get_events_from_nodestore(
...
+        )
+    except ServiceUnavailable:
+        extra = {
+            "project_id": project_id,
+            "error": "Service Unavailable",
+        }
+        logger.exception(
+            "tasks.backfill_seer_grouping_records.seer_service_unavailable",
+            extra=extra,
+        )
```
lobsterkatie, line 381: Nit: Is there are reason to include `service unavailable` in the tags when it's in the log event name?

lobsterkatie (approved):

src/sentry/tasks/embeddings_grouping/utils.py
jangjodi: nope, I can remove it

PR #74864, opened 2024-07-24
feat(dashboards): Update the dashboard widget state when split decision arrives

narsaynorath:
When there's a split decision, we should call the dashboard updating method to update the internal state to reflect the decision (since it's saved, it's just not re-fetched).

This:
- handles both events/events-stats requests
- only submits the `dashboardWidgetId` for evaluation if we're on the widget builder page
- Sets the widget builder state to use the `discoverSplitDecision` result in the response if it's there
- Submitting the widget builder updates the dashboard widget
- Fixes a number of the default datasets and default widget types used in the builder when the feature flag is enabled

⚠️ Needs to be followed up on:
- timeseries charts end up being a small size, need to figure out why they're not full sized
- cancelling editing the widget won't update the dashboard UI to query the correct dashboard


static/app/views/dashboards/datasetConfig/errorsAndTransactions.tsx
```diff
@@ -688,15 +688,16 @@ export function filterAggregateParams(
...
 
 const getQueryExtraForSplittingDiscover = (
   widget: Widget,
   organization: Organization,
-  useOnDemandMetrics: boolean
+  _useOnDemandMetrics: boolean
 ) => {
-  if (!useOnDemandMetrics || !shouldSendWidgetForSplittingDiscover(organization)) {
+  // TODO: What's the correct condition here for sending the split?
+  if (!shouldSendWidgetForSplittingDiscover(organization)) {
```
shruthilayaj, line 700:
```suggestion
  if (!organization.features.includes('performance-discover-dataset-selector')) && (!useOnDemandMetrics || !organization.features.includes('performance-discover-widget-split-ui')) {
  ```
  
  something like this so we don't clobber over the old condition if the dataset selector flag isn't enabled

shruthilayaj (approved):

PR #74865, opened 2024-07-24
analytics(replay): track render of missing replay alert

michellewzhang:
track render of alert banner, which we currently only show on issue details. 

context: 
> I’m concerned this can happen more often than customers are comfortable with and we might need to dig into reasons and improve things somehow

### DIFF
--- a/static/app/utils/analytics/replayAnalyticsEvents.tsx
+++ b/static/app/utils/analytics/replayAnalyticsEvents.tsx
@@ -92,6 +92,9 @@ export type ReplayEventParameters = {
     platform: string | undefined;
     project_id: string | undefined;
   };
+  'replay.render-missing-replay-alert': {
+    location: string;
+  };
   'replay.render-player': {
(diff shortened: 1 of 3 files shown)


c298lee (approved):
too fast

aliu39 (approved):

static/app/utils/analytics/replayAnalyticsEvents.tsx
```diff
@@ -92,6 +92,9 @@ export type ReplayEventParameters = {
     platform: string | undefined;
     project_id: string | undefined;
   };
+  'replay.render-missing-replay-alert': {
+    location: string;
```
ryan953, line 96: This might be misinterpreted as a `Location` object. Maybe it could be renamed to like `surface` or something. It's not fully a 'page' value because there are multiple spots within issue details, so which spot exactly is what i'm aiming for

ryan953 (approved):

PR #74867, opened 2024-07-24
feat(tabs): Add new tab variant for draggable tabs

MichaelSun48:
This PR makes non-visible changes to the tabs component in anticipation of future changes being made for custom views. Specifically, it separates the `<Tab/>` and `<BaseTab/>` variants and adds a new `variant` prop that controls which group of styles are being applied to the baseTab component. 

It also reorganizes the component structure of the `<TabList>` to allow for the overflow menu to be reusable, since it will be used in the future draggable tabs component.

### DIFF
--- a/static/app/components/interactionStateLayer.tsx
+++ b/static/app/components/interactionStateLayer.tsx
@@ -7,6 +7,10 @@ import {defined} from 'sentry/utils';
 interface StateLayerProps extends React.HTMLAttributes<HTMLSpanElement> {
   as?: React.ElementType;
   color?: string;
+  /**
+   * Controls if the opacity is increased when the element is in a
+   * selected or expanded state (aria-selected='true' or aria-expanded='true')
+   */
   hasSelectedBackground?: boolean;
   higherOpacity?: boolean;
   isHovered?: boolean;
--- a/static/app/components/interactionStateLayer.tsx
+++ b/static/app/components/interactionStateLayer.tsx
@@ -7,6 +7,7 @@ import {defined} from 'sentry/utils';
 interface StateLayerProps extends React.HTMLAttributes<HTMLSpanElement> {
   as?: React.ElementType;
   color?: string;
+  hasSelectedBackground?: boolean;
   higherOpacity?: boolean;
   isHovered?: boolean;
   isPressed?: boolean;
@@ -65,11 +66,16 @@ const InteractionStateLayer = styled(
         `
       : // If isPressed is undefined, then fallback to default press selectors
         css`
-          *:active > &&,
-          *[aria-expanded='true'] > &&,
-          *[aria-selected='true'] > && {
+          *:active > && {
             opacity: ${p.higherOpacity ? 0.12 : 0.09};
           }
+          ${p.hasSelectedBackground &&
+          css`
+            *[aria-expanded='true'] > &&,
+            *[aria-selected='true'] > && {
+              opacity: ${p.higherOpacity ? 0.12 : 0.09};
+            }
+          `}
         `}
 
 
@@ -78,5 +84,9 @@ const InteractionStateLayer = styled(
     opacity: 0;
   }
 `;
+
+InteractionStateLayer.defaultProps = {
+  hasSelectedBackground: true,
+};
 
 export default InteractionStateLayer;
--- a/static/app/components/tabs/tab.tsx
+++ b/static/app/components/tabs/tab.tsx
@@ -3,9 +3,14 @@ import type {Theme} from '@emotion/react';
 import styled from '@emotion/styled';
 import type {AriaTabProps} from '@react-aria/tabs';
 import {useTab} from '@react-aria/tabs';
-import {useObjectRef} from '@react-aria/utils';
+import {mergeProps, useObjectRef} from '@react-aria/utils';
 import type {TabListState} from '@react-stately/tabs';
-import type {Node, Orientation} from '@react-types/shared';
+import type {
+  DOMAttributes,
+  FocusableElement,
+  Node,
+  Orientation,
+} from '@react-types/shared';
 
 import InteractionStateLayer from 'sentry/components/interactionStateLayer';
 import Link from 'sentry/components/links/link';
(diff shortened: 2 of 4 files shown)


static/app/components/tabs/tab.tsx
```diff
@@ -37,64 +42,165 @@ function handleLinkClick(e: React.PointerEvent<HTMLAnchorElement>) {
...
+   * Additional props to be merged with `tabProps`. This is used
+   * by <DraggableTab> to pass in props used for drag-and-drop functionality.
+   */
+  additionalProps?: React.HTMLAttributes<HTMLElement>;
+  borderStyle?: 'solid' | 'dashed';
+  to?: string;
+  variant?: 'vanilla' | 'draggable';
+}
+
+export const BaseTab = forwardRef(
```
MichaelSun48, line 62: This `<BaseTab/>` component is functionally exactly the same the previous `<BaseTab/>` comopnent, except its being wrapped by a new `<Tab/>` component.

static/app/components/tabs/tab.tsx
```diff
@@ -37,64 +42,165 @@ function handleLinkClick(e: React.PointerEvent<HTMLAnchorElement>) {
...
+  overflowing: boolean;
+  tabProps: DOMAttributes<FocusableElement>;
+  /**
+   * Additional props to be merged with `tabProps`. This is used
+   * by <DraggableTab> to pass in props used for drag-and-drop functionality.
+   */
+  additionalProps?: React.HTMLAttributes<HTMLElement>;
+  borderStyle?: 'solid' | 'dashed';
+  to?: string;
+  variant?: 'vanilla' | 'draggable';
```
MichaelSun48, line 59: Rather than making this a boolean, i've left it as a union type since I heard there are _even more_ tab variants that are coming.

MichaelSun48: also open to naming these something else in case the draggable tabs design is planned to be used elsewhere

malwilley: A union is great! I would ask Vu about naming, he did seem to think that this new tab design would be used elsewhere. I _believe_ it is meant to be used when it is at the top of the page, whereas the current design would only be used for lower, self-contained elements.

static/app/components/tabs/tab.tsx
```diff
@@ -37,64 +42,165 @@ function handleLinkClick(e: React.PointerEvent<HTMLAnchorElement>) {
...
+}>`
+  &[aria-selected='true'] {
+    ${p =>
+      `
+        border-radius: 6px 6px 1px 1px;
+        border-top: 1px ${p.borderStyle} ${p.theme.border};
+        border-left: 1px ${p.borderStyle} ${p.theme.border};
+        border-right: 1px ${p.borderStyle} ${p.theme.border};
+        background-color: ${p.theme.white};
+        color: ${p.theme.fontWeightBold};
```
malwilley, line 177: This should be on `font-weight`?

static/app/components/tabs/tab.tsx
```diff
@@ -37,64 +42,165 @@ function handleLinkClick(e: React.PointerEvent<HTMLAnchorElement>) {
...
+  }
+
+  &[aria-selected='false'] {
+    border-top: 1px solid transparent;
+  }
+
+  transform: translateY(1px);
+  padding: 5px 10px;
+
+  opacity: 0px;
```
malwilley, line 189: `px` for opacity?

malwilley (commented):
Could you add a section to the `.stories` file so we can see what the new variant looks like?

static/app/components/tabs/tab.tsx
MichaelSun48: Spoke with Vu and decided on renaming it to "flat" (old, existing design) and "filled" (new design, used for draggable tabs)

MichaelSun48:
> Could you add a section to the .stories file so we can see what the new variant looks like?
Done!
image

static/app/components/tabs/tab.tsx
```diff
@@ -37,64 +43,167 @@ function handleLinkClick(e: React.PointerEvent<HTMLAnchorElement>) {
...
+        font-weight: ${p.theme.fontWeightBold};
+      `}
+  }
+
+  &[aria-selected='false'] {
+    border-top: 1px solid transparent;
+  }
+
+  transform: translateY(1px);
+  padding: 5px 10px;
```
malwilley, line 192: Can this use the `space()` constants instead?

static/app/components/tabs/tab.tsx
```diff
@@ -37,64 +43,167 @@ function handleLinkClick(e: React.PointerEvent<HTMLAnchorElement>) {
...
+  overflowing: boolean;
+}>`
+  &[aria-selected='true'] {
+    ${p =>
+      `
+        border-radius: 6px 6px 1px 1px;
+        border-top: 1px ${p.borderStyle} ${p.theme.border};
+        border-left: 1px ${p.borderStyle} ${p.theme.border};
+        border-right: 1px ${p.borderStyle} ${p.theme.border};
+        background-color: ${p.theme.white};
```
malwilley, line 182: Probably want `background` instead, as `white` will be white regardless of theme 

CleanShot 2024-07-24 at 13 24 01

malwilley (commented):
Is it purposeful that there is a large gap between tabs? 

CleanShot 2024-07-24 at 13 25 40

MichaelSun48:
> Is it purposeful that there is a large gap between tabs? 

Nope, this was caused by the fact that the padding is still applied to the tab names even if they are not selected. This is fixed in the next commit!

MichaelSun48:
Default behavior for selected is broken - working on a fix, will delete this when its out

static/app/components/interactionStateLayer.tsx
```diff
@@ -79,4 +85,8 @@ const InteractionStateLayer = styled(
   }
 `;
 
+InteractionStateLayer.defaultProps = {
+  hasSelectedBackground: true,
+};
```
MichaelSun48, line 90: This is how default props have to be defined apparently 🫠

static/app/components/interactionStateLayer.tsx
```diff
@@ -7,6 +7,7 @@ import {defined} from 'sentry/utils';
 interface StateLayerProps extends React.HTMLAttributes<HTMLSpanElement> {
   as?: React.ElementType;
   color?: string;
+  hasSelectedBackground?: boolean;
```
malwilley, line 10: Since it isn't obvious, could you add a `/** */` comment here explaining what this does?

static/app/components/interactionStateLayer.tsx
malwilley: Oh I see, because it's a styled component I guess

static/app/components/tabs/tab.tsx
```diff
@@ -37,64 +43,171 @@ function handleLinkClick(e: React.PointerEvent<HTMLAnchorElement>) {
...
+  hidden: boolean;
+  isSelected: boolean;
+  orientation: Orientation;
+  overflowing: boolean;
+  tabProps: DOMAttributes<FocusableElement>;
+  /**
+   * Additional props to be merged with `tabProps`. This is used
+   * by <DraggableTab> to pass in props used for drag-and-drop functionality.
+   */
+  additionalProps?: React.HTMLAttributes<HTMLElement>;
```
malwilley, line 57: Can we avoid this by prop by merging in `DraggableTab` instead?

malwilley (approved):
Looks good!

PR #74868, opened 2024-07-24
fix(profiling): take max width of the two spaces when rendering a flamegraph

JonasBa:
We need to take the max space of the two views so that both can be fully rendered (in practice txn are likely always going to be longer than profiles, but we shouldnt trim the duration)

### DIFF
--- a/static/app/components/profiling/flamegraph/continuousFlamegraph.tsx
+++ b/static/app/components/profiling/flamegraph/continuousFlamegraph.tsx
@@ -61,6 +61,7 @@ import {
   fromNanoJoulesToWatts,
   type ProfilingFormatterUnit,
 } from 'sentry/utils/profiling/units/units';
+import {formatTo} from 'sentry/utils/profiling/units/units';
 import {useDevicePixelRatio} from 'sentry/utils/useDevicePixelRatio';
 import {useMemoWithPrevious} from 'sentry/utils/useMemoWithPrevious';
 import {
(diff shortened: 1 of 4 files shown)


static/app/components/profiling/flamegraph/continuousFlamegraph.tsx
```diff
@@ -97,10 +98,27 @@ function collectAllSpanEntriesFromTransaction(
   return allSpans;
 }
 
-function getMaxConfigSpace(profileGroup: ContinuousProfileGroup): Rect {
-  // We have a transaction, so we should do our best to align the profile
-  // with the transaction's timeline.
+function getMaxConfigSpace(
+  profileGroup: ContinuousProfileGroup,
+  transaction: EventTransaction | null,
```
Zylphrex, line 103: I think this is fine for now because we're aiming for feature parity with the transaction profiles but there's a world where we won't have a transaction for a continuous profile so we should rely on the start/end timestamps passed to determine the config space instead of the transaction.

Zylphrex (approved):

static/app/components/profiling/flamegraph/continuousFlamegraph.tsx
JonasBa: Yup, I realized that. It'll need some more tweaking so I'm punting it till we get to adapt this to work without perf

PR #74872, opened 2024-07-24
fix(relocation): Only use OrganizationMapping for forking

azaslavsky:


src/sentry/api/endpoints/organization_fork.py
```diff
@@ -141,7 +121,7 @@ def post(self, request: Request, organization_id_or_slug) -> Response:
 
         # Identify who will be the owner of the newly forked organization, and ensure that they
         # don't already have relocations in flight.
-        owners = organization.get_owners()
+        owners = organization_mapping_service.get_owners(organization_id=org_mapping.id)
```
markstory, line 124: Using new RPC methods in the same changes as the methods are added can be unsafe. It would be safer to merge and deploy the new RPC methods separately, and once the RPC methods are deployed do changes to use those RPC methods in a follow up pull request.

azaslavsky: I agree in the general case, and would usually do that, but:

1. The only use of these new methods is triggered via the /organization/[ID_OR_SLUG]/fork/.
2. This API is only accessible to admins, and has no UI enabled (yet).
3. This means that realistically, only I will be hitting the API while this PR rolls out. I can just avoid doing so until both control and the regions have deployed properly, or otherwise triage the relevant sentry errors if I forget.

Does that seem reasonable?

markstory: :+1: Ok. That sounds like a good enough guard rail to avoid breakage.

markstory (approved):
Looks good to me.

PR #74876, opened 2024-07-24
chore(slack): correctly type issue message builder

cathteng:
Fixing typing for the issues Slack message builder.

### DIFF
--- a/src/sentry/integrations/message_builder.py
+++ b/src/sentry/integrations/message_builder.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from collections.abc import Mapping, Sequence
+from collections.abc import Callable, Mapping, Sequence
 from typing import Any
 
 from sentry import features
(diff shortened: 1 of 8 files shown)


src/sentry/integrations/slack/message_builder/issues.py
```diff
@@ -584,7 +545,8 @@ def build(self, notification_uuid: str | None = None) -> SlackBlock:
         # If an event is unspecified, use the tags of the latest event (if one exists).
         event_for_tags = self.event or self.group.get_latest_event()
 
-        obj = self.event if self.event is not None else self.group
+        event_or_group: Group | GroupEvent = self.event if self.event is not None else self.group
```
asottile-sentry, line 548: I don't think this should need an annotation?  mypy should be able to infer this union I think

src/sentry/integrations/slack/message_builder/issues.py
```diff
@@ -243,7 +223,7 @@ def get_context(group: Group) -> str:
     return context_text.rstrip()
 
 
-def get_option_groups_block_kit(group: Group) -> Sequence[Mapping[str, Any]]:
+def get_option_groups(group: Group) -> Sequence[Mapping[str, Any]]:
```
asottile-sentry, line 226: ~generally return types should use concrete types (`list[...]`, `dict[...]`, etc.) -- this may also benefit from a `TypedDict` ?

src/sentry/integrations/slack/message_builder/issues.py
```diff
@@ -109,9 +109,9 @@ def build_assigned_text(identity: RpcIdentity, assignee: str) -> str | None:
     except ObjectDoesNotExist:
         return None
 
-    if actor.is_team:
+    if actor.is_team and isinstance(assigned_actor, Team):
         assignee_text = f"#{assigned_actor.slug}"
-    elif actor.is_user:
+    elif actor.is_user and isinstance(assigned_actor, RpcUser):
```
asottile-sentry, line 114: could probably just use the isinstance checks here since they should do the same thing

asottile-sentry (commented):
nice!

src/sentry/integrations/slack/message_builder/issues.py
cathteng: it gets angry if i take it away

asottile-sentry: what's the error? -- does it narrow it to `object` (probably)? :(

src/sentry/integrations/message_builder.py
```diff
@@ -51,8 +51,6 @@ def format_actor_option(actor: Team | RpcUser, use_block_kit: bool = False) -> M
             }
         return {"text": f"#{actor.slug}", "value": f"team:{actor.id}"}
 
-    raise NotImplementedError
```
asottile-sentry, line 54: I think I would probably change the `if`s to `elifs` and `else: raise AssertionError('unreachable')`

src/sentry/integrations/slack/message_builder/issues.py
asottile-sentry: nice -- now with your TypedDict this should be `-> list[OptionGroup]`!

iamrajjoshi (approved):

iamrajjoshi (approved):
lgtm, i wonder if we can consolidate these pure functions in message builder into a class eventually

asottile-sentry:
> lgtm, i wonder if we can consolidate these pure functions in message builder into a class eventually

consolidating pure functions into a class is a bad idea -- they are pure functions and there is no reason to bring unnecessary object-orientation into this.  classes make it harder to reason about, test, type, and work with.  most of the mess we have today is due to excessive OO / inheritance / cuteness with classes and we'd be much better off without them

src/sentry/integrations/slack/message_builder/issues.py
cathteng: yes

PR #74879, opened 2024-07-24
feat(query-builder): Update the savedSearchType prop to be more descriptive and inclusive of other types

malwilley:
To be more descriptive, renames `savedSearchType` to `recentSearches` because that is actually what it configures. 

Additionally, when recording analytics, it will use the correct label for each type (the current search bar only records 'issues' and 'events').


static/app/components/searchQueryBuilder/tokens/freeText.tsx
```diff
@@ -541,7 +542,7 @@ function SearchQueryBuilderInputInternal({
             resetInputValue();
             trackAnalytics('search.key_manually_typed', {
               organization,
-              search_type: savedSearchType === 0 ? 'issues' : 'events',
+              search_type: recentSearches === 0 ? 'issues' : 'events',
```
roggenkemper, line 545: is there a reason why this isn't `recentSearchTypeToLabel(recentSearches)`

malwilley: Missed this one - thanks!

roggenkemper (approved):

PR #74880, opened 2024-07-24
ref: unify signature of plugin get_config

asottile-sentry:
mypy 1.11 points out these are inconsistent

### DIFF
--- a/src/sentry/api/serializers/models/plugin.py
+++ b/src/sentry/api/serializers/models/plugin.py
@@ -125,7 +125,7 @@ class PluginWithConfigSerializer(PluginSerializer):
                 "config": [
                     serialize_field(self.project, item, c)
                     for c in item.get_config(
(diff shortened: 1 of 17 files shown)


src/sentry/api/serializers/models/plugin.py
```diff
@@ -125,7 +125,7 @@ def get_attrs(self, item_list, user, **kwargs):
                 "config": [
                     serialize_field(self.project, item, c)
                     for c in item.get_config(
-                        project=self.project, user=user, add_additial_fields=True
```
joshuarli, line 128: lol

joshuarli (approved):

src/sentry/api/serializers/models/plugin.py
asottile-sentry: lolright

PR #74885, opened 2024-07-24
Refactor alert_rules to be a single query

saponifi3d:
# Description
Perf improvements to delayed processing series.

This will change from using 2 queries and then filtering in memory to instead handle this in sql.

### DIFF
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -458,9 +458,7 @@ def apply_delayed(project_id: int, *args: Any, **kwargs: Any) -> None:
 
     # STEP 3: Fetch the Rule models we need to check
     alert_rules_queryset = Rule.objects.filter(id__in=list(rules_to_groups.keys())).exclude(
-        id__in=Subquery(
-            RuleSnooze.objects.filter(rule_id=OuterRef("id"), user_id=None).values("rule_id")
-        )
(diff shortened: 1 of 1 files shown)


ceorourke (approved):

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -456,13 +457,10 @@ def apply_delayed(project_id: int, *args: Any, **kwargs: Any) -> None:
     rules_to_groups = get_rules_to_groups(rulegroup_to_event_data)
 
     # STEP 3: Fetch the Rule models we need to check
-    alert_rules_qs = Rule.objects.filter(id__in=list(rules_to_groups.keys()))
-    snoozed_rules = set(
-        RuleSnooze.objects.filter(rule__in=alert_rules_qs, user_id=None).values_list(
-            "rule", flat=True
-        )
+    alert_rules_queryset = Rule.objects.filter(id__in=list(rules_to_groups.keys())).exclude(
+        id__in=Subquery(RuleSnooze.objects.filter(rule_id=OuterRef("id")).values("rule_id"))
```
ceorourke, line 461: Wait josh no lol the `user_id=None` is from the original query, we still need that.

ceorourke: the lack of a `user_id` in this table means that it's snoozed for _everyone_ as opposed to being snoozed for an individual

saponifi3d: lol 😅 - i'll add it back.

just curious, what case we would have for that?

PR #74890, opened 2024-07-24
ref(breadcrumbs): Remove drawer components from render props

leeandher:
Since the drawer components can access any state they'll need  through the context provider, there's no need to pass them as props. Anyone implementing `openDrawer` can just import the components they care about. See https://github.com/getsentry/sentry/pull/74034#discussion_r1672804668

Also made the change to the one call site with a bit of a refactor.

### DIFF
--- a/static/app/components/events/breadcrumbs/breadcrumbsDrawer.tsx
+++ b/static/app/components/events/breadcrumbs/breadcrumbsDrawer.tsx
@@ -2,6 +2,8 @@ import {Fragment, useCallback, useMemo, useState} from 'react';
 import {useTheme} from '@emotion/react';
 import styled from '@emotion/styled';
 
+import ProjectAvatar from 'sentry/components/avatar/projectAvatar';
+import {Breadcrumbs as NavigationBreadcrumbs} from 'sentry/components/breadcrumbs';
 import {Button} from 'sentry/components/button';
 import ButtonBar from 'sentry/components/buttonBar';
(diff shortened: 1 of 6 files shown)


static/app/components/events/breadcrumbs/breadcrumbsDrawer.tsx
```diff
@@ -19,11 +21,16 @@ import {
   BREADCRUMB_SORT_OPTIONS,
   BreadcrumbSort,
 } from 'sentry/components/events/interfaces/breadcrumbs';
+import {DrawerBody, DrawerHeader} from 'sentry/components/globalDrawer/components';
```
malwilley, line 24: I don't have a strong opinion on it for this use case specifically, but for other compound components (like Tab, MultipleCheckbox, GuidedSteps) we have named the components like `Drawer.Body` to denote that they need to be used inside of the parent component. It's a bit different here though since you aren't actually rendering `<Drawer />` as the wrapper, so it could go either way for me. Just wanted to point it out

malwilley (approved):
Thanks for coming back to this! I like it much more this way!

static/app/components/events/breadcrumbs/breadcrumbsDrawer.tsx
leeandher: I thought about it but I think because there isn't the Drawer wrapper I'll retain this as is. They're entirely optional as well since the drawer renders empty, so I think it's okay as is.

PR #74891, opened 2024-07-24
feat(data-secrecy): Migration to Add Bit Flags to Hybrid Cloud Services

iamrajjoshi:
I recently merged a [migration](https://github.com/getsentry/sentry/pull/74700) to add `prevent_superuser_access` to the Organization bitflag and now need to add it to Hybrid Cloud services to sync since I need it to exist on `RpcOrganization`. 

In this PR, I also add the `disable_member_project_creation` flag since it existed above the new flag I created. 

My flag is still not used in our logic, but the `disable_member_project_creation` is currently being used, so we need to make sure this pr doesn't mess around with its value.


wedamija (commented):
Migration lgtm

iamrajjoshi:
> Migration lgtm

could i pytest.skip the failing tests? i think they are happening because the function i updated expects my migration's columns to exist.

wedamija:
> > Migration lgtm
> 
> could i pytest.skip the failing tests? i think they are happening because the function i updated expects my migration's columns to exist.

Yeah, it's fine to skip migration tests

tests/sentry/migrations/test_0729_backfill_groupsearchviews_with_pinned_searches.py
```diff
@@ -19,6 +21,7 @@ def setup_initial_state(self):
             sort="date",
         )
 
+    @pytest.mark.skip(reason="old migration test")
```
markstory, line 24: If these tests don't work anymore, why not delete them?

markstory (approved):
Looks safe to deploy to me. As regions are updated, they will start sending the new flag keys, and once control updates the flag state will start being updated. On the read side, the default values in `RpcOrganizationMappingFlags` will the columns in.

tests/sentry/migrations/test_0729_backfill_groupsearchviews_with_pinned_searches.py
wedamija: I think it's helpful to keep some around as an example, but we could delete old skipped tests after a while

PR #74905, opened 2024-07-24
fix(settings): show new secondary email after saving

oioki:
Fixes a bug where a newly added secondary email was not visible on the Email Addresses page till page reload.
Now we invalidate the query client cache and make sure to fetch the correct list of email addresses from the backend.

### DIFF
--- a/static/app/views/settings/account/accountEmails.spec.tsx
+++ b/static/app/views/settings/account/accountEmails.spec.tsx
@@ -106,29 +106,30 @@ describe('AccountEmails', function () {
     render(<AccountEmails />);
     expect(mock).not.toHaveBeenCalled();
 
+    const mockGetResponseBody = [
+      ...AccountEmailsFixture(),
+      {
+        email: 'test@example.com',
+        isPrimary: false,
+        isVerified: false,
+      },
+    ];
+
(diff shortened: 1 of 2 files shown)


static/app/views/settings/account/accountEmails.spec.tsx
```diff
@@ -106,7 +106,30 @@ describe('AccountEmails', function () {
...
+    });
+
+    const textbox = await screen.findByRole('textbox');
+    const expectedSecondaryEmails = AccountEmailsFixture().length - 1;
+    expect(screen.getAllByLabelText('Remove email')).toHaveLength(
+      expectedSecondaryEmails
+    );
+
+    await userEvent.type(textbox, 'test@example.com{enter}');
+    expect(screen.getAllByLabelText('Remove email')).toHaveLength(
```
saponifi3d, line 130: This might be a little flakey because of the GET request to the API request. 🤔 i wonder if we can tie this to the `mockGet` request you already have? (I guess because that's mocked it's unlikely to be flakey too 🎉)

saponifi3d (approved):
lgtm! the next level for this then is to update the data in `react-query`'s cache. to do that it'll take a bit of a refactor to make sure the data in the store is `UserEmail[]`

PR #74908, opened 2024-07-24
ref(scm): abstraction for repository integrations

cathteng:
Adds a `RepositoryIntegration` abstract base class that represents the same set of functionality as [RepositoryMixin](https://github.com/getsentry/sentry/blob/b5265c353916f10b1ea3e915cfef126e8d02e790/src/sentry/integrations/mixins/repositories.py). Also ties `RepositoryIntegration` to a new `RepositoryClient` abstract base class that represents the expected functionality of a SCM integration client tied to the SCM integration installation class.


src/sentry/integrations/github/integration.py
```diff
@@ -155,7 +155,7 @@ def get_document_origin(org) -> str:
 # Github App docs and list of available endpoints
 # https://docs.github.com/en/rest/apps/installations
 # https://docs.github.com/en/rest/overview/endpoints-available-for-github-apps
-class GitHubIntegration(IntegrationInstallation, GitHubIssueBasic, RepositoryMixin, CommitContextMixin):  # type: ignore[misc]
+class GitHubIntegration(IntegrationInstallation, RepositoryIntegration, GitHubIssueBasic, CommitContextMixin):  # type: ignore[misc]
```
cathteng, line 158: i've only replaced the GitHub integration's `RepositoryMixin` with `RepositoryIntegration`, but i've added `RepositoryClient` as a parent class of all the SCM integration clients

cathteng: in this class i've also grouped the methods by the parent class methods they are overriding

src/sentry/integrations/source_code_management/repository.py
```diff
@@ -0,0 +1,200 @@
...
+from sentry.shared_integrations.exceptions import ApiError, IntegrationError
+
+
+class RepositoryIntegration(ABC):
+    repo_search = False
+    model: RpcIntegration | Integration
+
+    # @abstractmethod
+    @property
+    def codeowners_locations(self) -> Sequence[str] | None:
```
cathteng, line 26: could probably make this abstract and return None everywhere that doesn't support codeowners

src/sentry/integrations/source_code_management/repository.py
```diff
@@ -0,0 +1,200 @@
...
+            'identifier': external_repo_id,
+        }]
+
+        The shape of the `identifier` should match the data
+        returned by the integration's
+        IntegrationRepositoryProvider.repository_external_slug()
+        """
+        raise NotImplementedError
+
+    def get_unmigratable_repositories(self) -> Collection[RpcRepository]:
```
cathteng, line 136: the default, if unimplemented, is []. again this could possibly be an abstract method and return [] everywhere that doesn't have an overriding implementation

src/sentry/integrations/source_code_management/repository.py
```diff
@@ -0,0 +1,200 @@
...
+    def codeowners_locations(self) -> Sequence[str] | None:
+        """
+        A list of possible locations for the CODEOWNERS file.
+        """
+        return None
+
+    @abstractmethod
+    def get_client(self) -> RepositoryClient:
+        """Returns the client for the integration. The client must be a subclass of RepositoryClient."""
+        raise NotImplementedError
```
GabeVillalobos, line 35: Is there a way we could maybe pass a client instance or factory to the class instead of having the subclass method to create one?

src/sentry/integrations/source_code_management/repository.py
```diff
@@ -0,0 +1,200 @@
...
+    def reinstall_repositories(self) -> None:
+        """Reinstalls repositories associated with the integration."""
+        result = integration_service.organization_contexts(integration_id=self.model.id)
+
+        for install in result.organization_integrations:
+            repository_service.reinstall_repositories_for_integration(
+                organization_id=install.organization_id,
+                integration_id=self.model.id,
+                provider=f"integrations:{self.model.provider}",
+            )
```
GabeVillalobos, line 152: Should this be wrapped in a try-catch, or are we okay with only some of the repositories being resintalled?

src/sentry/integrations/source_code_management/repository.py
```diff
@@ -0,0 +1,200 @@
...
+        try:
+            response = client.check_file(repo, filepath, branch)
+            if response is None:
+                return None
+        except IdentityNotValid:
+            return None
+        except ApiError as e:
+            if e.code != 404:
+                sentry_sdk.capture_exception()
+                raise
```
GabeVillalobos, line 85: These might be good areas to gather some metrics around failures in case we hit issues with the specific integrations/providers.

src/sentry/integrations/github/integration.py
```diff
@@ -164,34 +164,53 @@ def get_client(self) -> GitHubBaseClient:
...
-    def get_trees_for_org(self, cache_seconds: int = 3600 * 24) -> dict[str, RepoTree]:
-        trees: dict[str, RepoTree] = {}
-        domain_name = self.model.metadata["domain_name"]
-        extra = {"metadata": self.model.metadata}
-        if domain_name.find("github.com/") == -1:
-            logger.warning("We currently only support github.com domains.", extra=extra)
-            return trees
+    def message_from_error(self, exc: Exception) -> str:
+        if not isinstance(exc, ApiError):
+            return ERR_INTERNAL
```
GabeVillalobos, line 179: I wonder if this sort of error handling/extraction is something we could also define a new method for in the ABC

src/sentry/integrations/source_code_management/repository.py
cathteng: i copied this from the old mixin, this is wrapped in an atomic transaction?

src/sentry/integrations/source_code_management/repository.py
cathteng: i've actually been typing some classes in relation to this and found that some of them have different signatures (e.g. VSTS's `get_client` takes an optional `base_url` parameter), would a factory or client instance be better in that case?

src/sentry/integrations/source_code_management/repository.py
cathteng: nvm, seems like this is unused. removed!

src/sentry/integrations/github/integration.py
cathteng: this is from `IntegrationInstallation`, we can make it an `@abstractmethod` but it also has a default implementation

src/sentry/integrations/source_code_management/repository.py
cathteng: i'm adding some logging to make sure we don't need to add an overloaded function to handle VSTS's case https://github.com/getsentry/sentry/pull/75277

src/sentry/integrations/source_code_management/repository.py
```diff
@@ -0,0 +1,200 @@
...
+        if self.codeowners_locations is None:
+            raise NotImplementedError("Implement self.codeowners_locations to use this method.")
+
+        for filepath in self.codeowners_locations:
+            html_url = self.check_file(repo, filepath, ref)
+            if html_url:
+                try:
+                    contents = self.get_client().get_file(repo, filepath, ref, codeowners=True)
+                except ApiError:
+                    continue
```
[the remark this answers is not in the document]
cathteng: would it be better to wrap functions in the client? or would we catch it here specifically to be able to isolate the feature that is erroring?

src/sentry/integrations/source_code_management/repository.py
```diff
@@ -0,0 +1,213 @@
...
+from sentry.shared_integrations.exceptions import ApiError, IntegrationError
+from sentry.utils import metrics
+
+REPOSITORY_INTEGRATION_CHECK_FILE_METRIC = "repository_integration.check_file.{result}"
+REPOSITORY_INTEGRATION_GET_FILE_METRIC = "repository_integration.get_file.{result}"
+
+
+class RepositoryIntegration(ABC):
+    repo_search = False
+    model: RpcIntegration | Integration
```
RyanSkonnord, line 25: `model` should also be an abstract property.

For consistency's sake, let's go ahead and make `repo_search` a property as well (but a non-abstract one that returns `False` unless overridden). The class-level variable works for now, but I like a property method because (1) it allows subclasses to depend on instance state if they want to and (2) it's a better place to put a docstring.

src/sentry/api/serializers/models/projectcodeowners.py
```diff
@@ -38,8 +39,8 @@ def get_attrs(self, item_list, user, **kwargs):
                 organization_id=item.repository_project_path_config.organization_id,
             )
             codeowners_url = "unknown"
-            if item.repository_project_path_config.organization_integration_id and isinstance(
-                install, RepositoryMixin
+            if item.repository_project_path_config.organization_integration_id and (
+                isinstance(install, RepositoryMixin) or isinstance(install, RepositoryIntegration)
```
RyanSkonnord, line 43: Nit: this (and all similar `isinstance` checks) can be written as `isinstance(install, (RepositoryMixin, RepositoryIntegration))`. (Assuming we don't just delete `RepositoryMixin`.)

RyanSkonnord (commented):
Progress looks good.

`RepositoryIntegration` is meant to completely replace `RepositoryMixin`, isn't it? I'd recommend going ahead and deleting `RepositoryMixin` and replacing it with `RepositoryIntegration` in this PR. Making unit tests pass after doing that is the best way to ensure that we've designed `RepositoryIntegration` correctly.

src/sentry/integrations/source_code_management/repository.py
RyanSkonnord: 👍 but nit: `()` is preferable.

src/sentry/integrations/source_code_management/repository.py
RyanSkonnord: Afterthought: unless we care about reading those attributes directly from subclass type tokens, not from instances. But even in that case I think it may be possible to use properties with `@classmethod`. LMK if you want help with that, or else I'm fine just leaving them as class variables if it saves trouble.

cathteng: `repo_search` looks to always be `True` in the installation 🤔

cathteng: @RyanSkonnord `model` is also set, should i also implement a setter for it?

src/sentry/integrations/source_code_management/repository.py
```diff
@@ -9,31 +10,92 @@
...
 
+REPOSITORY_INTEGRATION_CHECK_FILE_METRIC = "repository_integration.check_file.{result}"
+REPOSITORY_INTEGRATION_GET_FILE_METRIC = "repository_integration.get_file.{result}"
 
-class RepositoryMixin:
-    # whether or not integration has the ability to search through Repositories
-    # dynamically given a search query
-    repo_search = False
 
+class RepositoryIntegration(ABC):
```
RyanSkonnord, line 21: Since this class has an "is-a" relationship to `IntegrationInstallation`, I think it would be good to extend it (i.e., `class RepositoryIntegration(IntegrationInstallation, ABC)`).

Then we won't have to inherit both separately in the subclasses. (That is, instead of writing `class VstsIntegration(IntegrationInstallation, RepositoryIntegration, VstsIssueSync)` it can just be `class VstsIntegration(RepositoryIntegration, VstsIssueSync)`, no longer using the mixin pattern.)

src/sentry/integrations/github/integration.py
```diff
@@ -155,43 +155,65 @@ def get_document_origin(org) -> str:
...
 # https://docs.github.com/en/rest/overview/endpoints-available-for-github-apps
-class GitHubIntegration(IntegrationInstallation, GitHubIssueBasic, RepositoryMixin, CommitContextMixin):  # type: ignore[misc]
-    repo_search = True
+class GitHubIntegration(IntegrationInstallation, RepositoryIntegration, GitHubIssueBasic, CommitContextMixin):  # type: ignore[misc]
     codeowners_locations = ["CODEOWNERS", ".github/CODEOWNERS", "docs/CODEOWNERS"]
 
+    @property
+    def integration_name(self) -> str:
+        return "github"
+
```
cathteng, line 164: everything besides this in this file is the same, i just grouped methods together

src/sentry/integrations/source_code_management/repository.py
```diff
@@ -1,39 +1,98 @@
...
 
+REPOSITORY_INTEGRATION_CHECK_FILE_METRIC = "repository_integration.check_file.{result}"
+REPOSITORY_INTEGRATION_GET_FILE_METRIC = "repository_integration.get_file.{result}"
 
-class RepositoryMixin:
-    # whether or not integration has the ability to search through Repositories
-    # dynamically given a search query
-    repo_search = False
 
+class RepositoryIntegration(IntegrationInstallation, ABC):
```
cathteng, line 22: this is the important new addition!

src/sentry/integrations/github/client.py
```diff
@@ -632,10 +633,14 @@ def get_labels(self, repo: str) -> Sequence[Any]:
         """
         return self.get(f"/repos/{repo}/labels", params={"per_page": 100})
 
-    def check_file(self, repo: Repository, path: str, version: str) -> BaseApiResponseX:
+    def check_file(
+        self, repo: Repository, path: str, version: str | None
+    ) -> BaseApiResponseX | None:
```
sentaur-athena, line 638: It seems like on a cache miss it sets the cache for that key and returns result so this can't be None. I don't think you will get type failure is the child type is subset of the parent.

src/sentry/integrations/github/integration.py
```diff
@@ -155,43 +154,65 @@ def get_document_origin(org) -> str:
...
     def get_client(self) -> GitHubBaseClient:
         if not self.org_integration:
             raise IntegrationError("Organization Integration does not exist")
         return GitHubApiClient(integration=self.model, org_integration_id=self.org_integration.id)
 
+    # IntegrationInstallation methods
```
sentaur-athena, line 169: Is this a TODO?

src/sentry/integrations/gitlab/client.py
```diff
@@ -319,14 +323,16 @@ def check_file(self, repo: Repository, path: str, ref: str) -> str | None:
             encoded_path = quote(path, safe="")
 
             request_path = GitLabApiClientPath.file.format(project=project_id, path=encoded_path)
-            return self.head_cached(request_path, params={"ref": ref})
+            return self.head_cached(request_path, params={"ref": version})
         except ApiError as e:
             # Gitlab can return 404 or 400 if the file doesn't exist
             if e.code != 400:
```
sentaur-athena, line 329: Should this include 404 too then because of the comment?

src/sentry/integrations/gitlab/client.py
```diff
@@ -61,7 +63,7 @@ def get_group(self, group):
         return self.get(path)
 
 
-class GitLabApiClient(IntegrationProxyClient):
+class GitLabApiClient(IntegrationProxyClient, RepositoryClient):
     integration_name = "gitlab"
```
sentaur-athena, line 67: This can be set during the init from the integration. Just for enforcing consistency.

src/sentry/integrations/source_code_management/repository.py
```diff
@@ -1,39 +1,98 @@
...
+            'identifier': external_repo_id,
+        }]
+
+        The shape of the `identifier` should match the data
+        returned by the integration's
+        IntegrationRepositoryProvider.repository_external_slug()
+        """
+        raise NotImplementedError
+
+    def get_unmigratable_repositories(self) -> Collection[RpcRepository]:
```
sentaur-athena, line 88: Can this be list or some other tighter type than collection?

src/sentry/integrations/github/client.py
cathteng: you're right, thanks for noticing!

src/sentry/integrations/gitlab/client.py
cathteng: ah yes, i can fix this here i guess

src/sentry/integrations/source_code_management/repository.py
cathteng: yes this can be a list

src/sentry/integrations/github/integration.py
```diff
@@ -155,43 +154,67 @@ def get_document_origin(org) -> str:
 # Github App docs and list of available endpoints
 # https://docs.github.com/en/rest/apps/installations
 # https://docs.github.com/en/rest/overview/endpoints-available-for-github-apps
-class GitHubIntegration(IntegrationInstallation, GitHubIssueBasic, RepositoryMixin, CommitContextIntegration):  # type: ignore[misc]
-    repo_search = True
+
+
+class GitHubIntegration(RepositoryIntegration, CommitContextIntegration, GitHubIssueBasic):  # type: ignore[misc]
     codeowners_locations = ["CODEOWNERS", ".github/CODEOWNERS", "docs/CODEOWNERS"]
```
cathteng, line 160: making this a property rather than leaving it here results in mypy issues

src/sentry/integrations/source_code_management/repository.py
```diff
@@ -1,39 +1,98 @@
...
+        """Extracts the source path from the source code url. Used for stacktrace linking."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def has_repo_access(self, repo: RpcRepository) -> bool:
+        """Used for migrating repositories. Checks if the installation has access to the repository."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_repositories(self, query: str | None = None) -> Sequence[dict[str, Any]]:
```
GabeVillalobos, line 70: Possible future improvement: Could we define a proper type to encapsulate an Integration Repository in the future? Having `Any` for this isn't great :/

src/sentry/integrations/source_code_management/repository.py
```diff
@@ -144,7 +183,29 @@ def get_codeowner_file(
             if html_url:
                 try:
                     contents = self.get_client().get_file(repo, filepath, ref, codeowners=True)
+                    metrics.incr(
+                        REPOSITORY_INTEGRATION_GET_FILE_METRIC.format(result="success"),
+                        tags={"integration": self.integration_name},
+                    )
                 except ApiError:
+                    metrics.incr(
```
GabeVillalobos, line 191: It might also be good to include some additional tag data here, like HTTP status code, just so we have some rough idea of the types of failures we're getting.

GabeVillalobos: Though I also understand that cardinality is a concern, so perhaps we could restrict it to a small number of the most important codes.

src/sentry/integrations/example/integration.py
```diff
@@ -65,14 +66,21 @@ def dispatch(self, request: Request, pipeline) -> HttpResponse:
 )
 
 
-class ExampleIntegration(IntegrationInstallation, IssueSyncMixin, RepositoryMixin):
```
sentaur-athena, line 68: Noob question, how is Example used?

src/sentry/integrations/github/integration.py
sentaur-athena: Do other integrations have this too? I don't think mypy should be a blocker about making design decisions. Can you put up a PR with the issue and then maybe we can help fixing the typing?

src/sentry/integrations/example/integration.py
cathteng: it's used in tests, e.g. https://github.com/getsentry/sentry/blob/4642ebb073341ee976c3cf2f4550c93865749691/tests/sentry/models/test_groupassignee.py#L119-L135

src/sentry/integrations/source_code_management/repository.py
cathteng: after a quick glance, it seems possible to define a typed dict with some optional elements, but i'd have to look closer

sentaur-athena (commented):
Looks good to me. Just wondering if we can fix the mypy issue

src/sentry/integrations/github_enterprise/integration.py
```diff
@@ -172,9 +176,27 @@ def get_repositories(self, query=None):
...
+        # Must format the url ourselves since `check_file` is a head request
+        # "https://github.example.org/octokit/octokit.rb/blob/master/README.md"
+        return f"{repo.url}/blob/{branch}/{filepath}"
+
+    def extract_branch_from_source_url(self, repo: Repository, url: str) -> str:
+        raise IntegrationFeatureNotImplementedError
+
+    def extract_source_path_from_source_url(self, repo: Repository, url: str) -> str:
+        raise IntegrationFeatureNotImplementedError
+
```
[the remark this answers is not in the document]
cathteng: i believe so

src/sentry/integrations/example/integration.py
sentaur-athena: Cool 🫘 
Do you plan to eventually complete its methods to have more complete tests then?

cathteng: i'm actually not sure what the benefit of having this is versus just using an existing integration, i would need to dig more

src/sentry/integrations/github/integration.py
cathteng: hmm i can tackle the mypy stuff after i get this in, if not i'll do what Gabe suggested

sentaur-athena (approved):

src/sentry/integrations/example/integration.py
RyanSkonnord: For dummy implementations that we expect not to get called, it might be better to raise `NotImplementedError`, maybe with a message like "assumed not to be needed for test cases".

It would also be good to move this to `testutils` to help communicate its purpose (but that doesn't need to be part of this PR).

RyanSkonnord (approved):

PR #74915, opened 2024-07-24
fix(anomaly detection): add validation for dynamic alert time window and threshold

mifu67:
Fixes https://getsentry.atlassian.net/browse/ALRT-165

### DIFF
--- a/src/sentry/incidents/logic.py
+++ b/src/sentry/incidents/logic.py
@@ -736,8 +736,8 @@ def update_alert_rule(
     comparison_delta=NOT_SET,
     monitor_type: AlertRuleMonitorTypeInt | None = None,
     description: str | None = None,
-    sensitivity: AlertRuleSensitivity | None = None,
-    seasonality: AlertRuleSeasonality | None = None,
+    sensitivity: AlertRuleSensitivity | None | object = NOT_SET,
+    seasonality: AlertRuleSeasonality | None | object = NOT_SET,
     detection_type: AlertRuleDetectionType | None = None,
     **kwargs,
 ):
@@ -778,9 +778,9 @@ def update_alert_rule(
         updated_fields["name"] = name
     if description:
         updated_fields["description"] = description
-    if sensitivity:
+    if sensitivity is not NOT_SET:
         updated_fields["sensitivity"] = sensitivity
-    if seasonality:
+    if seasonality is not NOT_SET:
         updated_fields["seasonality"] = seasonality
     if query is not None:
         updated_query_fields["query"] = query
@@ -815,16 +815,34 @@ def update_alert_rule(
 
         updated_fields["comparison_delta"] = comparison_delta
 
-    # if we modified the comparison_delta or the time_window, we should update the resolution accordingly
-    if "comparison_delta" in updated_fields or "time_window" in updated_query_fields:
+    # if we're updating these fields, then we may be changing the detection type
+    # print("COMPARISON DELTA:", comparison_delta)
+    # print("SEASONALITY:", seasonality)
+    # print("SENSITIVITY:", sensitivity)
+
+    # TODO: it seems that we can't check for all of these conditions. We may need to separate the seasonality and sensitivity checks
+    if "comparison_delta" in updated_fields:  # some value changed -> update type if necessary
         if comparison_delta is not None:
             detection_type = AlertRuleDetectionType.PERCENT
         else:
-            if seasonality:
-                detection_type = AlertRuleDetectionType.DYNAMIC
-            else:
-                detection_type = AlertRuleDetectionType.STATIC
+            detection_type = AlertRuleDetectionType.STATIC
 
+    if "seasonality" in updated_fields or "sensitivity" in updated_fields:
+        if seasonality is not None or seasonality is not None:
+            detection_type = AlertRuleDetectionType.DYNAMIC
+            # if the alert wasn't a dynamic alert, ensure that both seasonality and sensitivity are present
+            if alert_rule.detection_type != AlertRuleDetectionType.DYNAMIC and not (
+                "seasonality" in updated_fields and "sensitivity" in updated_fields
+            ):
+                raise ValidationError(
+                    "Dynamic alert rules must be defined with both sensitivity and seasonality"
+                )
+        else:
+            detection_type = AlertRuleDetectionType.STATIC
+
+    # print("NEW DETECTION TYPE:", detection_type)
+    # if we modified the comparison_delta or the time_window, we should update the resolution accordingly
+    if "comparison_delta" in updated_fields or "time_window" in updated_query_fields:
         window = int(
             updated_query_fields.get(
                 "time_window", timedelta(seconds=alert_rule.snuba_query.time_window)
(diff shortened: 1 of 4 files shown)


src/sentry/incidents/logic.py
```diff
@@ -854,6 +856,12 @@ def update_alert_rule(
             updated_fields["seasonality"] = None
         elif detection_type == AlertRuleDetectionType.DYNAMIC:
             updated_fields["comparison_delta"] = None
+            if (
+                time_window not in {15, 30, 60}
+                or time_window is None
+                and alert_rule.snuba_query.time_window not in {15 * 60, 30 * 60, 60 * 60}
+            ):
+                raise ValidationError("Invalid time window for dynamic alert")
```
ceorourke, line 864: tiny nit but since the validation error strings are repeated we could put them in a const to avoid potential future drift

tests/sentry/incidents/test_logic.py
```diff
@@ -1364,6 +1368,59 @@ def test_update_detection_type(self):
...
+            seasonality=AlertRuleSeasonality.AUTO,
+            time_window=60,
+            detection_type=AlertRuleDetectionType.DYNAMIC,
+        )
+
+        updated_rule = update_alert_rule(
+            rule, comparison_delta=None, sensitivity=None, seasonality=None
+        )
+
+        assert updated_rule.detection_type == AlertRuleDetectionType.STATIC
```
ceorourke, line 1422: I think this is missing dynamic to dynamic, static to dynamic, and percent to dynamic. Also slightly out of scope but I wonder if it would be useful to create a `self.create_dynamic_alert_rule` with some default values or update `self.create_alert_rule` to reduce the boilerplate

mifu67: Since dynamic requires detection type, these static to dynamic and percent to dynamic are tested in test_update_detection_type. I'll add a test for dynamic to dynamic there.

src/sentry/incidents/logic.py
```diff
@@ -566,6 +570,8 @@ def create_alert_rule(
     if detection_type == AlertRuleDetectionType.DYNAMIC:
         if not (sensitivity and seasonality):
             raise ValidationError("Dynamic alerts require both sensitivity and seasonality")
+        if time_window not in {15, 30, 60}:
```
nhsiehgit, line 573: can we make these values constants somewhere rather than arbitrary numbers?

src/sentry/incidents/logic.py
```diff
@@ -812,17 +818,17 @@ def update_alert_rule(
...
-
-    # if we modified the comparison_delta or the time_window, we should update the resolution accordingly
-    if "comparison_delta" in updated_fields or "time_window" in updated_query_fields:
-        if comparison_delta is not None:
-            detection_type = AlertRuleDetectionType.PERCENT
-        else:
-            if seasonality:
-                detection_type = AlertRuleDetectionType.DYNAMIC
+    if detection_type is not None:
+        updated_fields["detection_type"] = detection_type
```
nhsiehgit, line 822: should users be able to swap between detection types?
we should still clear comparison delta if the type is _not_ percent right?

src/sentry/incidents/logic.py
```diff
@@ -854,6 +860,12 @@ def update_alert_rule(
             updated_fields["seasonality"] = None
         elif detection_type == AlertRuleDetectionType.DYNAMIC:
             updated_fields["comparison_delta"] = None
+            if (
+                time_window not in {15, 30, 60}
```
nhsiehgit, line 864: lets make these windows constants
maybe something like
`window_s`, `window_m`, `window_l`
and `window_s_min`, `window_m_min`, `window_l_min`?
I dunno - maybe something else if we ever anticipate more windows than these

nhsiehgit (commented):
few comments

src/sentry/incidents/logic.py
mifu67: `we should still clear comparison delta if the type is not percent right?`
I believe we do this ^^
Users are able to swap detection types.

src/sentry/incidents/logic.py
```diff
@@ -98,6 +98,12 @@
 
 CRITICAL_TRIGGER_LABEL = "critical"
 WARNING_TRIGGER_LABEL = "warning"
+INVALID_TIME_WINDOW = (
+    "Invalid time window for dynamic alert (valid windows are 15, 30, and 60 minutes)"
+)
+INVALID_ALERT_THRESHOLD = "Dynamic alerts cannot have a nonzero alert threshold"
+DYNAMIC_TIME_THRESHOLDS = {15, 30, 60}
+DYNAMIC_TIME_THRESHOLDS_SECONDS = {15 * 50, 30 * 60, 60 * 60}
```
ceorourke, line 106: 🤔 is this 15 * 50 on purpose instead of 15 * 60?

ceorourke: and maybe `DYNAMIC_TIME_WINDOWS` instead of threshold? I keep getting the 2 mixed up so maybe you're right

mifu67: no it is not 😅

src/sentry/incidents/logic.py
```diff
@@ -98,6 +98,12 @@
 
 CRITICAL_TRIGGER_LABEL = "critical"
 WARNING_TRIGGER_LABEL = "warning"
+INVALID_TIME_WINDOW = (
+    "Invalid time window for dynamic alert (valid windows are 15, 30, and 60 minutes)"
+)
+INVALID_ALERT_THRESHOLD = "Dynamic alerts cannot have a nonzero alert threshold"
+DYNAMIC_TIME_WINDOWS = {15, 30, 60}
+DYNAMIC_TIME_WINDOWS_SECONDS = {15 * 60, 30 * 60, 60 * 60}
```
ceorourke, line 106: Could we make this populate based on `DYNAMIC_TIME_WINDOWS` like `DYNAMIC_TIME_WINDOWS_SECONDS {window * 60 for window in DYNAMIC_TIME_WINDOWS}` so we're not repeating the numbers in multiple places?

Same for `INVALID_TIME_WINDOW`, might read slightly worse but like 
`INVALID_TIME_WINDOW = f"Invalid time window for dynamic alert. Valid windows are {', '.join(map(str, DYNAMIC_TIME_WINDOWS))} minutes"`

src/sentry/incidents/logic.py
```diff
@@ -812,17 +820,15 @@ def update_alert_rule(
...
-
-    # if we modified the comparison_delta or the time_window, we should update the resolution accordingly
-    if "comparison_delta" in updated_fields or "time_window" in updated_query_fields:
-        if comparison_delta is not None:
-            detection_type = AlertRuleDetectionType.PERCENT
-        else:
-            if seasonality:
-                detection_type = AlertRuleDetectionType.DYNAMIC
+    if detection_type is None:
+        if "comparison_delta" in updated_fields:  # some value changed -> update type if necessary
```
ceorourke, line 824: Can we combine this into one line?

tests/sentry/incidents/endpoints/test_organization_alert_rule_index.py
```diff
@@ -223,10 +223,34 @@ def test_with_description(self):
     )
     def test_anomaly_detection_alert(self, mock_seer_request):
         data = {
-            **self.alert_rule_dict,
+            "aggregate": "count()",
+            "query": "",
+            "time_window": 30,
```
ceorourke, line 228: Is the `time_window` the only other difference between this and `self.alert_rule_dict`? You can just update that rather than define it all again here

mifu67: I also changed the alert thresholds to 0; happy to update instead if you think that's better

src/sentry/incidents/logic.py
mifu67: I tried, and it causes tests to fail, so apparently not

tests/sentry/incidents/endpoints/test_organization_alert_rule_index.py
ceorourke: Oh I see agh the threshold makes it trickier, this is fine

ceorourke (approved):

PR #74920, opened 2024-07-24
ref(onboarding): add/request integrations button

ameliahsu:
refactored `renderTopButton` method from `integrationsDetailedView` to be a separate shared `IntegrationButton` component which renders the appropriate button (add integration, request installation, or external installation)

### DIFF
--- a/static/app/views/alerts/rules/issue/messagingIntegrationModal.tsx
+++ b/static/app/views/alerts/rules/issue/messagingIntegrationModal.tsx
@@ -1,15 +1,14 @@
-import {Fragment, useCallback, useEffect, useState} from 'react';
+import {Fragment} from 'react';
 import styled from '@emotion/styled';
 
 import {addErrorMessage} from 'sentry/actionCreators/indicator';
 import type {ModalRenderProps} from 'sentry/actionCreators/modal';
-import Access from 'sentry/components/acl/access';
 import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
 import type {IntegrationProvider} from 'sentry/types/integrations';
-import type {Project} from 'sentry/types/project';
 import type {Organization} from 'sentry/types/organization';
-import useApi from 'sentry/utils/useApi';
+import type {Project} from 'sentry/types/project';
+import {useApiQuery} from 'sentry/utils/queryClient';
 import AddIntegrationRow from 'sentry/views/alerts/rules/issue/addIntegrationRow';
 import {IntegrationContext} from 'sentry/views/settings/organizationIntegrations/integrationContext';
 
@@ -21,38 +20,6 @@ type Props = ModalRenderProps & {
   bodyContent?: React.ReactElement<any, any>;
 };
 
-function getProvider(
-  providerKey: string,
-  provider: IntegrationProvider | null,
-  setProvider: (IntegrationProvider) => void,
-  organization: Organization,
-  api: any,
-  setHasError: (boolean) => void
-) {
-  const fetchData = useCallback(() => {
-    if (!providerKey) {
-      return Promise.resolve();
-    }
-
-    const endpoint = `/organizations/${organization.slug}/config/integrations/?provider_key=${providerKey}`;
-    return api
-      .requestPromise(endpoint)
-      .then(integrations => {
-        setProvider(integrations.providers[0]);
-      })
-      .catch(() => {
-        setHasError(true);
-        return null;
-      });
-  }, [providerKey, api, organization.slug, setHasError]);
-
-  useEffect(() => {
-    fetchData();
-  }, [fetchData]);
-
-  return provider;
-}
-
 function MessagingIntegrationModal({
   closeModal,
   Header,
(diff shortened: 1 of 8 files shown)


static/app/views/settings/organizationIntegrations/integrationDetailedView.tsx
```diff
@@ -242,54 +240,31 @@ class IntegrationDetailedView extends AbstractIntegrationDetailedView<
...
-
-    const size = 'sm' as const;
-    const priority = 'primary' as const;
 
     const buttonProps = {
       style: {marginBottom: space(1)},
-      size,
-      priority,
+      size: 'sm' as const,
+      priority: 'primary' as const,
```
saponifi3d, line 247: TypeScript will infer the type, so you can just do:
```suggestion
      size: 'sm',
      priority: 'primary',
```

static/app/views/settings/organizationIntegrations/integrationDetailedView.tsx
```diff
@@ -242,54 +240,31 @@ class IntegrationDetailedView extends AbstractIntegrationDetailedView<
   renderTopButton(disabledFromFeatures: boolean, userHasAccess: boolean) {
```
saponifi3d, line 240: i'd try to replace this function with the new component. Each of these `render*` functions shuold likely be different components.

static/app/views/settings/organizationIntegrations/integrationButton.tsx
```diff
@@ -0,0 +1,98 @@
...
+  onExternalClick,
+  organization,
+  provider,
+  type,
+  userHasAccess,
+  installStatus,
+  analyticsParams,
+  externalInstallText,
+  modalParams = {},
+  buttonProps,
```
saponifi3d, line 53: I'd recommend adding a React.context to access the properties that aren't intended to be directly an attribute of a "button" - by that i mean things like data or adding to the parent state should ideally be accessed via a react.context. this allows you to share all of that data with all the children components of a single parent, and it means that the interface to this component will just look like a button (making it easier to reuse in the future)

static/app/views/settings/organizationIntegrations/integrationButton.tsx
```diff
@@ -0,0 +1,98 @@
...
+        external
+        {...buttonProps}
+      >
+        {externalInstallText
+          ? externalInstallText
+          : metadata.aspects.externalInstall.buttonText}
+      </Button>
+    );
+  }
+  return <Fragment />;
```
saponifi3d, line 95: could we return null instead of fragment here? (a fragment will render to nothing in the html, so it might be better to tell react to render nothing right away)

saponifi3d (commented):
the way i'd think about trying to remove this button from the view class, is to create a drop in component for the `render` function. eventually if we can turn each render function into a component, we can then simplify this whole view to a functional component instead of class based.

It may also help to create a `react.context` ([docs](https://react.dev/reference/react/createContext)) to share all of this complex state to the children components, and allow those child components to easily modify / access top level state.

static/app/views/alerts/rules/issue/messagingIntegrationModal.tsx
```diff
@@ -43,15 +81,37 @@ function MessagingIntegrationModal({
...
-                onClickHandler={closeModal}
-                setHasError={setHasError}
-              />
+                key={value}
+              >
+                {({hasAccess}) => {
+                  return (
+                    <IntegrationContext.Provider
+                      value={{
+                        provider: provider,
```
saponifi3d, line 98: it seems like most of the data we would want is really in the `provider` key. things like `organization` should likely be passed as a property for now.

I'd recommend looking at this context and making sure it's just the API response information needed for this component. higher level things like organization and project should likely be refactored into their own providers later.

static/app/views/settings/organizationIntegrations/integrationButton.tsx
```diff
@@ -0,0 +1,83 @@
...
+  onExternalClick,
+  externalInstallText,
+  buttonProps,
+}: Props) {
+  const integration = useContext(IntegrationContext);
+  if (!integration) {
+    return null;
+  }
+  const {provider, type, installStatus, analyticsParams, modalParams} = integration;
+  const {metadata} = provider;
```
saponifi3d, line 40: nit: can safely destruct `integration` like this:
```suggestion
  const {provider, type, installStatus, analyticsParams, modalParams } = useContext(IntegrationContext) ?? {};
  if (!provider) return null;
  const {metadata} = provider;
```

static/app/views/settings/organizationIntegrations/integrationContext.tsx
```diff
@@ -0,0 +1,22 @@
...
+import type {IntegrationProvider, IntegrationType} from 'sentry/types/integrations';
+
+export type IntegrationContextProps = {
+  analyticsParams: {
+    already_installed: boolean;
+    view:
+      | 'integrations_directory_integration_detail'
+      | 'integrations_directory'
+      | 'onboarding'
+      | 'project_creation';
```
saponifi3d, line 12: i like how exact this is.. but i wonder if we could also just use `string` here to make it a bit more reusable for the future?

saponifi3d (approved):
🎉 thanks for diving into the context and improving the composition of the changes!

static/app/views/settings/organizationIntegrations/integrationContext.tsx
ameliahsu: Unfortunately we can't use `string` here because the analyticsParams type is defined by [AddIntegration](https://github.com/getsentry/sentry/blob/master/static/app/views/settings/organizationIntegrations/addIntegration.tsx#L22-L29)

PR #74921, opened 2024-07-24
fix(typing): fix typing on BaseApiClient

cathteng:


iamrajjoshi (approved):
lgtm, thanks for doing this 🙌

PR #74922, opened 2024-07-24
feat(toolbar): add latest release panel

### LINKED ISSUE #74453: [DevToolbar] Add latest release into section
We should add a section to the toolbar which simply pulls the latest release info from the API and shows it either in a panel, or show the basics like deploy-time & commit in the button bar itself. 

This will let engineers know (when in prod) what version of the code they're looking at.

michellewzhang:
- barebones release panel showing the latest release for one specific project.
- closes [#74453](https://github.com/getsentry/sentry/issues/74453)
- can follow up with adding crash free rate, issues related to the release, adoption chart, etc! 
- matches the release card on the releases page somewhat:

SCR-20240724-oirj

### DIFF
--- a/static/app/components/devtoolbar/components/releases/useToolbarRelease.tsx
+++ b/static/app/components/devtoolbar/components/releases/useToolbarRelease.tsx
@@ -1,24 +1,28 @@
+import {useMemo} from 'react';
+
+import type {ApiEndpointQueryKey} from 'sentry/components/devtoolbar/types';
 import type {Release} from 'sentry/types/release';
-import {useApiQuery} from 'sentry/utils/queryClient';
 
 import useConfiguration from '../../hooks/useConfiguration';
+import useFetchApiData from '../../hooks/useFetchApiData';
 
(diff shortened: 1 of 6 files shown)


static/app/components/devtoolbar/components/releases/useToolbarRelease.tsx
```diff
@@ -0,0 +1,24 @@
+import type {Release} from 'sentry/types/release';
+import {useApiQuery} from 'sentry/utils/queryClient';
```
ryan953, line 2: If this switches to `from '../../hooks/useFetchApiData'` then this can also change in releasesPanel.tsx:
```
- <ReleaseHeader release={data[0]} orgSlug={organizationSlug} />
+ <ReleaseHeader release={data} orgSlug={organizationSlug} />
```

eventually we'll have to remove any `from 'sentry/...` imports. This one needs to be replaced because `queryClient.tsx` imports from stores and organization and lots of stuff like that.

ryan953 (approved):

PR #74924, opened 2024-07-24
feat(toolbar): Iterate on the toolbar color/style and close button

ryan953:
The main floating panel has some new colors now, but it also fades away after 3seconds so it's not fulling in people's face

This is to address the feedback that:
a) it's too blended in when the app has a white background
b) it stands out too much when there's a red badge on a button

We can style it better with some more effort, i was trying to invert the background when it's faded but had trouble because of the gradient and transitions not playing together or something. We can get fancier too and make it like slide away and be replaced with only a sentry logo or something idk.

| With an open panel | Without a panel open, after 3s |
| --- | --- |
| SCR-20240724-ouox | SCR-20240724-owiz


michellewzhang (approved):
gif

c298lee (approved):
no pink :(

PR #74925, opened 2024-07-25
ref(control_silo): Consolidate integrations tasks and tasks' tests

Christinarlong:
Moves all integrations tasks and related tests to sentry/integrations/tasks or the specific integration provider/tasks (i.e github/tasks). Includes tasks shims for moved tasks.

ref(#73859)


src/sentry/integrations/tasks/sync_assignee_outbound_impl.py
```diff
@@ -0,0 +1,47 @@
...
+from sentry.models.integrations.integration import Integration
+from sentry.models.organization import Organization
+from sentry.models.user import User
+from sentry.silo.base import SiloMode
+from sentry.tasks.base import instrumented_task, retry
+from sentry.users.services.user.service import user_service
+
+
+@instrumented_task(
+    name="sentry.integrations.tasks.sync_assignee_outbound_impl",
```
markstory, line 13: Did this task name change? I don't see this name in any current tasks.

src/sentry/integrations/api/bases/external_actor.py
```diff
@@ -0,0 +1,165 @@
...
+    ExternalProviders.CUSTOM,
+}
+
+STRICT_NAME_PROVIDERS = {
+    ExternalProviders.GITHUB,
+    ExternalProviders.GITLAB,
+}
+
+
+class ExternalActorSerializerBase(CamelSnakeModelSerializer):
```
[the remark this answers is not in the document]
Christinarlong: agh I thought I got all of these. I think this is an artifact from making changes on one branch, moving branches, and then the changes get added to the current branches if a new file from the prev. branch gets updated.

src/sentry/integrations/tasks/sync_assignee_outbound_impl.py
Christinarlong: yes, in the previous/old module the task was named `sentry.tasks.integrations.sync_assignee_outbound`. In the new task/location it's updated to have the full module name. 

Will this lead to conflicts? It shouldn't (?), because all the new module paths are being added to the celery_imports

markstory: There shouldn't be any conflicts as you've kept a task with the old name around to handle the deploy boundary. Since we're moving/renaming this task could we skip the `_impl` suffix? We have a different namespace which will give us unique names.

src/sentry/integrations/slack/tasks/find_channel_id_for_rule.py
```diff
@@ -0,0 +1,115 @@
...
+from sentry.models.rule import Rule, RuleActivity, RuleActivityType
+from sentry.shared_integrations.exceptions import ApiRateLimitedError, DuplicateDisplayNameError
+from sentry.silo.base import SiloMode
+from sentry.tasks.base import instrumented_task
+
+logger = logging.getLogger("sentry.integrations.slack.tasks")
+
+
+@instrumented_task(
+    name="sentry.integrations.slack.search_channel_id",
```
markstory, line 25: I'm seeing two tasks with this name. One in this file and the other in `src/sentry/tasks/integrations/slack/find_channel_id_for_rule`. We should have unique names for celery tasks.

src/sentry/integrations/slack/tasks/post_message.py
```diff
@@ -0,0 +1,55 @@
...
+        integration_id=integration_id,
+        payload=payload,
+        log_error_message=log_error_message,
+        log_params=log_params,
+    )
+
+
+# TODO: add retry logic
+@instrumented_task(
+    name="sentry.integrations.slack.post_message_control",
```
markstory, line 38: This task and `sentry.integrations.slack.tasks.post_message` both have duplicate tasks registered. We could remove the task defined in `src/sentry/tasks/integrations/slack/post_message` and only do an import shim to avoid the duplicate task names.

src/sentry/integrations/tasks/kick_off_status_syncs_impl.py
```diff
@@ -0,0 +1,26 @@
+from sentry.models.grouplink import GroupLink
+from sentry.silo.base import SiloMode
+from sentry.tasks.base import instrumented_task, retry, track_group_async_operation
+
+
+@instrumented_task(
+    name="sentry.integrations.tasks.kick_off_status_syncs_impl",
```
markstory, line 7: We could rename this task and remove the `_impl`. We'd need a shim task with the old task name for deploy compatibility.

src/sentry/integrations/tasks/link_all_repos.py
```diff
@@ -0,0 +1,88 @@
...
+def get_repo_config(repo, integration_id):
+    return {
+        "external_id": repo["id"],
+        "integration_id": integration_id,
+        "identifier": repo["full_name"],
+    }
+
+
+@instrumented_task(
+    name="sentry.integrations.tasks.link_all_repos",
```
markstory, line 26: Should these tasks be within the `sentry.integrations.github` ? We have some tasks within their integration (some slack tasks) and other tasks are in `sentry.integrations`. Perhaps we can settle on having integration specific tasks within that integration?

src/sentry/integrations/tasks/migrate_issues.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+@instrumented_task(
+    name="sentry.integrations.tasks.migrate_issues",
+    queue="integrations",
+    default_retry_delay=60 * 5,
+    max_retries=5,
+)
+@retry(exclude=(Integration.DoesNotExist))
+def migrate_issues(integration_id: int, organization_id: int) -> None:
```
markstory, line 21: We could put this task in the jira integration?

src/sentry/integrations/tasks/migrate_opsgenie_plugins.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@instrumented_task(
+    name="sentry.integrations.tasks.migrate_opsgenie_plugins",
+    queue="integrations",
+    default_retry_delay=60 * 5,
+    max_retries=5,
+)
+@retry(exclude=(Integration.DoesNotExist, OrganizationIntegration.DoesNotExist))
+def migrate_opsgenie_plugin(integration_id: int, organization_id: int) -> None:
```
markstory, line 28: This module could be moved to `sentry.integrations.opsgenie.tasks` 🤷

src/sentry/integrations/tasks/sync_metadata.py
```diff
@@ -0,0 +1,25 @@
...
+
+@instrumented_task(
+    name="sentry.integrations.tasks.sync_metadata",
+    queue="integrations.control",
+    default_retry_delay=20,
+    max_retries=5,
+    silo_mode=SiloMode.CONTROL,
+)
+@retry(on=(IntegrationError,), exclude=(Integration.DoesNotExist,))
+def sync_metadata(integration_id: int) -> None:
```
markstory, line 15: Could move this task into the jira integration.

src/sentry/integrations/tasks/vsts/subscription_check.py
```diff
@@ -0,0 +1,90 @@
...
+
+@instrumented_task(
+    name="sentry.integrations.tasks.vsts.subscription_check",
+    queue="integrations.control",
+    default_retry_delay=60 * 5,
+    max_retries=5,
+    silo_mode=SiloMode.CONTROL,
+)
+@retry(exclude=(ApiError, ApiUnauthorized, Integration.DoesNotExist))
+def vsts_subscription_check(integration_id: int, organization_id: int) -> None:
```
markstory, line 21: Same here with putting these tasks in `sentry.integrations.vsts.tasks`. We could skip renaming the task for now and do that in a followup change.

src/sentry/integrations/slack/tasks/post_message.py
Christinarlong: what does duplicate task registered mean? Is that like two separate tasks are named the same ? 

Also for the import shim, would that look something like the class/symbol import shims ? I'm kinda confused rn, don't the task import shims need the `@instrumented_task` decorator to be valid?

src/sentry/integrations/tasks/kick_off_status_syncs_impl.py
Christinarlong: `src/sentry/tasks/integrations/kick_off_status_syncs_impl.py` has a shim task with `"sentry.tasks.integrations.kick_off_status_syncs"` would that work for compatibility?

src/sentry/integrations/tasks/link_all_repos.py
Christinarlong: sounds good to me 👍

src/sentry/integrations/tasks/vsts/__init__.py
```diff
@@ -0,0 +1,9 @@
+from sentry.integrations.tasks.vsts.kickoff_subscription_check import (
+    kickoff_vsts_subscription_check,
+)
+from sentry.integrations.tasks.vsts.subscription_check import vsts_subscription_check
```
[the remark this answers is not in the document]
Christinarlong: Agreege, though there's an issue when having the vsts tasks in the vsts namespace (i.e sentry/integrations/vsts/tasks) where there's a cyclical dependency for the plugins module (1s lemme get a screenshot to remember what exactly was the problem)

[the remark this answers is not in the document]
Christinarlong: Screenshot below. But it seems like the path is 
- setup -> plugins -> start loading other modules .... -> integrations/tasks/init.py -> integrations/vsts/init.py -> rules -> plugins
image

src/sentry/integrations/slack/tasks/post_message.py
markstory: > what does duplicate task registered mean? Is that like two separate tasks are named the same ?

Yes, two tasks have the same `name` parameter provided to `instrumented_task`. Celery will get confused when trying to execute these tasks and we should avoid that.

src/sentry/integrations/tasks/kick_off_status_syncs_impl.py
markstory: Yeah that would work, I mostly want to avoid having `_impl` in the task name as that will require another round of renames to fix.

src/sentry/integrations/tasks/sync_status_inbound.py
```diff
@@ -0,0 +1,261 @@
...
+        raise Integration.DoesNotExist
+
+    organizations = Organization.objects.filter(id=organization_id)
+    affected_groups = Group.objects.get_groups_by_external_issue(
+        integration, organizations, issue_key
+    )
+    if not affected_groups:
+        return
+
+    installation = integration.get_installation(organization_id=organization_id)
```
Christinarlong, line 196: Would it be bad, to leave off the type hint for the return type of `get_installation()` ? Trying to import `IssueSyncMixin` in the `get_installation()` file (to typehint, IntegrationInstallation | IssueSyncMixin) causes an explosion of circular import errors. Currently alot of the mypy errors are solved by hasattr() which, doesn't seem as clean as saying,  installation is X | Y type.

src/sentry/integrations/tasks/sync_status_inbound.py
```diff
@@ -0,0 +1,261 @@
...
+        "integration_id": integration_id,
+    }
+    if action == ResolveSyncAction.RESOLVE:
+        (
+            resolutions_by_group_id,
+            activity_type,
+            activity_data,
+        ) = get_resolutions_and_activity_data_for_groups(
+            list(affected_groups),
+            str(config.get("resolution_strategy")),
```
Christinarlong, line 224: coerced for mypy

asottile-sentry: `str(x.get(...))` is going to end up with `"None"` which is almost certainly wrong -- probably the correct response to this error should be `config['resolution_strategy']` -- or the caller should accept `None`?

Christinarlong: gave caller `None`typehint for this param 👍

src/sentry/integrations/tasks/slack/send_notifications_on_activity.py
```diff
@@ -10,15 +10,15 @@
...
+_FAILURE_METRIC = "sentry.integrations.tasks.slack.send_notifications_on_activity.send_activity_notifications.failure"
 _TASK_QUEUED_METRIC = (
-    "sentry.integrations.slack.tasks.send_notifications_on_activity.activity_created_receiver"
+    "sentry.integrations.tasks.slack.send_notifications_on_activity.activity_created_receiver"
 )
 
 
 @instrumented_task(
-    name="sentry.integrations.slack.tasks.send_activity_notifications_to_slack_threads",
+    name="sentry.integrations.slack.send_activity_notifications_to_slack_threads",
```
asottile-sentry, line 21: I'm pretty sure changing the `name` of a task is a breaking change -- old code won't be able to queue up for the new code and new code won't be able to queue up for the old code (in the case of a rollback)

markstory: Yes, changing celery task names will cause failures. I'll restore this and the other changed task names, as we don't want to drop any tasks during deploys. A rollback would cause failures with the changes as they stand right now, as we're introducing new task names that wouldn't be present in a previous release 🤔

asottile-sentry: probably need a multi-stage rollout then (1) introduce new tasks (2) cut over to new tasks (3) remove old tasks

markstory: Yeah. We'll split this up into multiple stages so that we can bridge rollbacks and forwards deploys.

Christinarlong:
Closing as PR has been split :^)

PR #74927, opened 2024-07-25
feat(replay): add feedback button to HE modal, using pause/unpause focus

### LINKED ISSUE #72377: [Hydration Errors] Remove "Discussion" button from existing Modal and close github discussion topic
https://github.com/getsentry/sentry/discussions/62097

### LINKED ISSUE #74732: [Hydration Errors] Embed Feedback Widget in the the Diff Viewer

aliu39:
Alternative to https://github.com/getsentry/sentry/pull/74813.
Fixes https://github.com/getsentry/sentry/issues/74732
Fixes https://github.com/getsentry/sentry/issues/72377

Avoids feedback SDK changes. Instead we can use the pause/unpause fxs of the global modal's focus trap, exposed in the ModalStore (😮‍💨).

To add this button, this PR makes changes to:
- The modalStore config. Adds nullable pause/unpause-focus-trap callbacks to the state, and sets them when the global modal first renders.
- FeedbackWidgetButton and useFeedbackWidget. Adds an optional optionOverrides prop that gets passed to the widget fxs. This could be pretty useful in the future too.

https://github.com/user-attachments/assets/c656d3b3-e159-47cf-9cfb-6bdec5f3b1be


static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
```diff
@@ -25,40 +28,37 @@ export default function ReplayComparisonModal({
...
-            <FeatureBadge type="beta" />
-          </h4>
-          <GithubFeedbackButton
-            href="https://github.com/getsentry/sentry/discussions/62097"
-            label={t('Discussion')}
-            title={null}
-            analyticsEventKey="replay.details-hydration-discussion-clicked"
-            analyticsEventName="Replay Details Hydration Discussion Clicked"
-            priority="primary"
+      <AsyncSDKIntegrationContextProvider outerContext={integrationsContext}>
```
ryan953, line 35: Because `useAsyncSDKIntegrationStore` and `<AsyncSDKIntegrationContextProvider>` are so close together here... i wonder if we need this extra step at all, it might just work without the sub-provider.

static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
```diff
@@ -25,40 +28,37 @@ export default function ReplayComparisonModal({
...
+                boldBefore: <strong />,
+                boldAfter: <strong />,
+              }
+            )}
+          </StyledParagraph>
+          <ReplayDiffChooser
+            replay={replay}
+            leftOffsetMs={leftOffsetMs}
+            rightOffsetMs={rightOffsetMs}
           />
```
ryan953, line 43: This is for https://github.com/getsentry/sentry/issues/72377

static/app/components/feedback/widget/feedbackPauseFocusButton.tsx
ryan953, line 1: I wonder if we can write this in terms of `<FeedbackWidgetButton>`. That one might need a few more props, like passing in more options to the feedbackIntegration, but it should be possible to avoid re-writing this kind of thing: `<Button ref={buttonRef} size="sm" icon={<IconMegaphone />}>`

static/app/stores/modalStore.tsx
```diff
@@ -44,9 +50,23 @@ const storeConfig: ModalStoreDefinition = {
   },
 
   openModal(renderer: Renderer, options: ModalOptions) {
-    this.state = {renderer, options};
+    this.state = {...this.state, renderer, options};
     this.trigger(this.state);
   },
+
+  setPauseFocusTrap(fx: State['pauseFocusTrap'] | null) {
```
ryan953, line 57: `null` is part of the State definition

```suggestion
  setPauseFocusTrap(fx: State['pauseFocusTrap']) {
```

static/app/stores/modalStore.tsx
```diff
@@ -44,9 +50,23 @@ const storeConfig: ModalStoreDefinition = {
...
+    this.state = {
+      ...this.state,
+      pauseFocusTrap: fx,
+    };
+  },
+
+  setUnpauseFocusTrap(fx: State['unpauseFocusTrap'] | null) {
```
ryan953, line 64: `null` is part of the State definition

```suggestion
  setUnpauseFocusTrap(fx: State['unpauseFocusTrap']) {
```

static/app/components/feedback/widget/feedbackPauseFocusButton.tsx
aliu39: Gotcha. I can take a look, I mainly made this component to get stuff out fast!

static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
aliu39: Oh we don't, thanks for the catch. I'm a little confused cause I remembered discussing the problem with you which led to the teachings on how to use context providers... maybe the real problem was not running getsentry? 😆

static/app/components/globalModal/index.tsx
```diff
@@ -151,6 +151,8 @@ function GlobalModal({onClose}: Props) {
       escapeDeactivates: false,
       fallbackFocus: portal,
     });
+    ModalStore.setPauseFocusTrap(focusTrap.current.pause);
+    ModalStore.setUnpauseFocusTrap(focusTrap.current.unpause);
```
ryan953, line 155: These 2 are only called in one spot, here. And they both depend on `focusTrap.current`, what if this was one setter like `MocalStore.setFocusTrap(focusTrap.current);`

static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
```diff
@@ -33,14 +39,18 @@ export default function ReplayComparisonModal({
...
-          />
+          {pauseFocusTrap && unpauseFocusTrap ? (
+            <FeedbackWidgetButton
+              optionOverrides={{
+                onFormOpen: () => {
+                  pauseFocusTrap();
+                },
+                onFormClose: () => {
+                  unpauseFocusTrap();
+                },
```
ryan953, line 50: This can be compressed. Instead of wrapping with an extra function call, we can just pass the functions themselves:

```suggestion
                onFormOpen: pauseFocusTrap,
                onFormClose: unpauseFocusTrap,
```

ryan953 (approved):

static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
aliu39: The functions return focustrap though, while onformclose needs to return void.

ryan953: ah, good point!

Issue #74255, opened 2024-07-15
Delete connected alerts when a span metric is deleted


matejminar:

vgrozdanic:
Cascade deletion of alerts is not possible, due to the possible problem with permissions that could occur.

Alert deletion requires `project:write` permission, and span extraction rule requires only `project:read` permission. If we would allow alert to be deleted via span extraction rule, we would be breaking our permission system, and allow deletion of the alerts to users who shouldn't have permission to do it

vgrozdanic:
Cascade deletion of alerts is not possible, due to the possible problem with permissions that could occur.

Alert deletion requires `project:write` permission, and span extraction rule requires only `project:read` permission. If we would allow alert to be deleted via span extraction rule, we would be breaking our permission system, and allow deletion of the alerts to users who shouldn't have permission to do it

PR #74932, opened 2024-07-25
fix(metrics): wrap span metric alerts formatting in try/except for model DoesNotExist

shellmayr:
Fixes [SENTRY-3CCT](https://sentry.sentry.io/issues/5642599130/)
Fixes [SENTRY-3CCV](https://sentry.sentry.io/issues/5642599365/)

### DIFF
--- a/src/sentry/snuba/metrics/naming_layer/mri.py
+++ b/src/sentry/snuba/metrics/naming_layer/mri.py
@@ -264,9 +264,12 @@ def format_mri_field(field: str) -> str:
 
         if parsed:
             if condition_id := parsed.mri.span_attribute_rule_id:
-                condition = SpanAttributeExtractionRuleCondition.objects.get(id=condition_id)
-                config = condition.config
-                return f'{parsed.op}({config.span_attribute}) filtered by "{condition.value}"'
+                try:
+                    condition = SpanAttributeExtractionRuleCondition.objects.get(id=condition_id)
+                    config = condition.config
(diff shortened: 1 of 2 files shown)


src/sentry/snuba/metrics/naming_layer/mri.py
```diff
@@ -264,9 +264,12 @@ def format_mri_field(field: str) -> str:
...
             if condition_id := parsed.mri.span_attribute_rule_id:
-                condition = SpanAttributeExtractionRuleCondition.objects.get(id=condition_id)
-                config = condition.config
-                return f'{parsed.op}({config.span_attribute}) filtered by "{condition.value}"'
+                try:
+                    condition = SpanAttributeExtractionRuleCondition.objects.get(id=condition_id)
+                    config = condition.config
+                    return f'{parsed.op}({config.span_attribute}) filtered by "{condition.value}"'
+                except SpanAttributeExtractionRuleCondition.DoesNotExist:
+                    return field
```
vgrozdanic, line 272: Can you also add a metric or some logging, because, if I am not mistaken, this situation shouldn't be happening?

vgrozdanic (approved):

src/sentry/snuba/metrics/naming_layer/mri.py
shellmayr: Yeah, at the moment it can still work because you can delete a metric without it deleting an alert. (when https://github.com/getsentry/sentry/issues/74255 is done this problem goes away)

PR #74936, opened 2024-07-25
ref(incidents): Factor out dict of slugs to alert rule trigger actions

RyanSkonnord:
Delete `sentry.incidents.serializers.STRING_TO_ACTION_TYPE`. Fix bugs that depend on import order by always looking at the set of registered actions rather than building the dict too early.

Needed for https://github.com/getsentry/sentry/pull/74743


src/sentry/incidents/models/alert_rule.py
```diff
@@ -564,7 +578,7 @@ class AlertRuleTriggerAction(AbstractNotificationAction):
     Type = ActionService
     TargetType = ActionTarget
 
-    _factory_registrations: dict[ActionService, ActionHandlerFactory] = {}
+    factory_registrations = FactoryRegistry()
```
RyanSkonnord, line 581: Previously, the privacy-by-convention of the leading underscore was being honored only in prod code but ignored by several unit tests. Now those unit tests interact with it only through the new `TemporaryAlertRuleTriggerActionRegistry` class from testutils.

I'm on the fence about whether to keep this as underscore-prefixed, and to just let `TemporaryAlertRuleTriggerActionRegistry` be an exception (since it's only in one place and relatively easy to fix if we break it again).

src/sentry/incidents/logic.py
```diff
@@ -1809,13 +1809,16 @@ def get_filtered_actions(
     alert_rule_data: Mapping[str, Any],
     action_type: ActionService,
 ):
-    from sentry.incidents.serializers import STRING_TO_ACTION_TYPE
+    def is_included(action: Mapping[str, Any]) -> bool:
+        type_slug = action.get("type")
+        factory = AlertRuleTriggerAction.look_up_factory_by_slug(type_slug)
+        return factory is not None and factory.service_type == action_type
```
GabeVillalobos, line 1815: Is there any benefit to defining this as a nested closure vs just implementing the logic here?

src/sentry/incidents/models/alert_rule.py
```diff
@@ -550,6 +550,20 @@ def build_handler(
...
+class FactoryRegistry:
+    def __init__(self) -> None:
+        self.by_action_service: dict[ActionService, ActionHandlerFactory] = {}
+        self.by_slug: dict[str, ActionHandlerFactory] = {}
+
+    def register(self, factory: ActionHandlerFactory) -> None:
+        if factory.service_type in self.by_action_service:
+            raise Exception(f"Handler already registered for type {factory.service_type}")
+        if factory.slug in self.by_slug:
+            raise Exception(f"Handler already registered with slug={factory.slug!r}")
```
cathteng, line 562: what does the `!r` do here?

tests/sentry/incidents/test_tasks.py
```diff
@@ -208,7 +209,8 @@ def test_missing_project(self):
         )
 
     def test(self):
-        with patch.object(AlertRuleTriggerAction, "_factory_registrations", new={}):
+        suspended_registrations = TemporaryAlertRuleTriggerActionRegistry.suspend()
```
cathteng, line 212: what is the difference between patching, and temporarily overriding the registrations in tests?

src/sentry/incidents/models/alert_rule.py
```diff
@@ -550,6 +550,20 @@ def build_handler(
         return self.trigger_action_class(action, incident, project)
 
 
+class FactoryRegistry:
+    def __init__(self) -> None:
+        self.by_action_service: dict[ActionService, ActionHandlerFactory] = {}
+        self.by_slug: dict[str, ActionHandlerFactory] = {}
```
cathteng, line 556: could we possibly add a comment about the difference between these two? when do we use one over the other perhaps?

src/sentry/incidents/models/alert_rule.py
GabeVillalobos: Are we using this class member variable as a singleton? What's the benefit of this over a module-level singleton we reference instead?

Regardless, I think keeping the `_` prefix here makes sense just to make it clear that mutating this is discouraged.

cathteng (commented):
just some questions for my understanding, i see what's happening here and it's cool that we can remove a random floating constant

src/sentry/incidents/models/alert_rule.py
RyanSkonnord: The only difference is which key you want to use for a lookup. If that isn't clear enough from the names, I'll add a comment.

src/sentry/incidents/models/alert_rule.py
RyanSkonnord: In an f-string, it's shorthand for `f"slug={repr(factory.slug)}"`, so that it prints `slug="slack"` rather than `slug=slack`.

src/sentry/incidents/logic.py
RyanSkonnord: It's for readability. By "implementing the logic here", do you mean inlining it into the comprehension below (line 1821)?

Having `factory` as an intermediate variable is necessary for the null check on line 1815. I'm not sure if we need to anticipate unmatched type slugs, but if we do receive one, we don't want to get back a null value for `factory` and then choke on `factory.service_type`.

By comparison, this was not a problem when it was written as
```python
STRING_TO_ACTION_TYPE.get(action.get("type")) == action_type
```
because `STRING_TO_ACTION_TYPE.get(action.get("type"))` would just be `None` and then `None == action_type` evaluates to false.

To inline it without the helper function, we'd have to write it as
```python
    return [
        rewrite_trigger_action_fields(action)
        for trigger in alert_rule_data.get("triggers", [])
        for action in trigger.get("actions", [])
        if (
            AlertRuleTriggerAction.look_up_factory_by_slug(action.get("type")) is not None
            and AlertRuleTriggerAction.look_up_factory_by_slug(action.get("type")).service_type == action_type
        )
    ]
```

tests/sentry/incidents/test_tasks.py
RyanSkonnord: They're functionally equivalent. My gripe with the old way wasn't about `patch.object`, but that the thing we were patching was underscore-prefixed and should have been private. IMHO, the privacy convention applies to unit tests too, and we were breaking it here.

My approach with `TemporaryAlertRuleTriggerActionRegistry` is equally kludgey, but my goal was to isolate it into one helper class that lives in a `testutils` package, so as to more clearly communicate "this is some special test-only magic".

As for the try-finally structure versus the `with` block, I actually like the `with` better, but the `suspend` and `restore` methods were necessary because of the other unit tests that want to call them during `setUp` and `tearDown`. If we value the `with` structure, I could put add a helper method with `@contextmanager` that lets us continue to use it.

src/sentry/incidents/logic.py
GabeVillalobos: Oh, I don't know how I missed the list comprehension in the return line 😅 This makes sense now. Thanks for clarifying!

src/sentry/incidents/models/alert_rule.py
RyanSkonnord: > Are we using this class member variable as a singleton?

Yes. We previously had a single global `dict[ActionService, ActionHandlerFactory]` object, and now a `FactoryRegistry` instance is replacing it (and the only reason I made `FactoryRegistry` its own class is because I wanted to have a single `register` method responsible for populating two dicts).

> Regardless, I think keeping the `_` prefix here makes sense just to make it clear that mutating this is discouraged.

I agree. I think I'll put the underscore back and add a comment that `TemporaryAlertRuleTriggerActionRegistry` is a testutil with special dispensation to violate privacy.

> What's the benefit of this over a module-level singleton we reference instead?

The distinction matters less without the underscore prefix. If it's private to the class, that communicates that it should be modified only by calling `classmethod`s on `AlertRuleTriggerAction`.

However (is this what you were getting at?) `FactoryRegistry` would make sense as a *public* module-level singleton. I think the real problem is that (prior to any of my factoring) the `AlertRuleTriggerAction` class was pulling double duty as a registry object. Methods like `register_type` want to be instance methods on a registry object, not class methods on anything.

Hence, a good next refactoring step would be to extract *all* the class methods from `AlertRuleTriggerAction` into a new public class, which means going around to all the calls like `@AlertRuleTriggerAction.register_type(...)` and change them to something like `@action_handler_factory_registry.register_type(...)` where `action_handler_factory_registry` is a global, static singleton. (That still leaves us with the pre-existing problems of having global, static singletons in the first place... which, lest we forget, is why I wanted to nuke `STRING_TO_ACTION_TYPE` in the first place. 😜)

GabeVillalobos: > However (is this what you were getting at?) FactoryRegistry would make sense as a public module-level singleton. I think the real problem is that (prior to any of my factoring) the AlertRuleTriggerAction class was pulling double duty as a registry object.

Yep, exactly what I was getting at. I see your point about this being very similar to the dictionary you were refactoring away in the first place, but I don't see a great way around having _some_ sort of registration singleton at the moment, unless we pursue a more dynamic way of registering `ActionHandlerFactory` classes similar to what we did with RPC services 🤔

RyanSkonnord: https://github.com/getsentry/sentry/pull/74936/commits/44aae2b8bb28bd7a9f52a7fae13661965d79d6a3

tests/sentry/incidents/test_tasks.py
RyanSkonnord: https://github.com/getsentry/sentry/pull/74936/commits/0e792376bfd8e74b3f2aad40a964d00f4dde24ca

src/sentry/incidents/models/alert_rule.py
RyanSkonnord: https://github.com/getsentry/sentry/pull/74936/commits/1f33e57458097c433a0c3ed0fc0d7c7fd2f5e718

GabeVillalobos (approved):

PR #74943, opened 2024-07-25
chore(features) Expand documentation for sentry.features

markstory:
Expand the documentation for features.handler and features.manager.

### DIFF
--- a/src/sentry/features/handler.py
+++ b/src/sentry/features/handler.py
@@ -22,6 +22,9 @@ class FeatureHandler:
 
     Subclasses should implement `has` and contain the logic
     necessary for the feature check.
+
+    Generally FeatureHandlers are only implemented in `getsentry.features`
+    as we don't programatically release features in self-hosted.
     """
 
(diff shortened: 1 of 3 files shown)


src/sentry/features/handler.py
```diff
@@ -17,6 +17,13 @@
 
 
 class FeatureHandler:
+    """
+    Base class for defining custom logic for feature decisions.
+
+    Subclasses should implement `has` and contain the logic
+    necessary for the feature check.
+    """
+
```
cmanallen, line 26: Since this is the root it may be helpful to enumerate common use-cases, any recommended do's and don'ts, file locations to place flags, links to enabling flags in relay and other services, etc.

cmanallen (approved):
Yay documentation!

src/sentry/features/handler.py
markstory: I've added a bit more context, and can go through the getsentry code and expand documentation there as well.

PR #74944, opened 2024-07-25
ref(metrics): Make visualize field wider

priscilawebdev:
**Before:**

https://github.com/user-attachments/assets/dd0fa41b-bb87-4c56-8bd5-4f47822e452d

**After:**

https://github.com/user-attachments/assets/2bb7e432-78fb-4482-ba1d-c24432c2c44c

### DIFF
--- a/static/app/components/metrics/queryBuilder.tsx
+++ b/static/app/components/metrics/queryBuilder.tsx
@@ -216,9 +216,25 @@ export const QueryBuilder = memo(function QueryBuilder({
   const projectIdStrings = useMemo(() => projectIds.map(String), [projectIds]);
 
   return (
-    <QueryBuilderWrapper>
+    <QueryBuilderWrapper hasMetricsNewInputs={hasMetricsNewInputs(organization)}>
+      {hasMetricsNewInputs(organization) && (
+        <GuideAnchor target="metrics_selector" position="bottom" disabled={index !== 0}>
+          <QueryFieldGroup>
+            <QueryFieldGroup.Label>{t('Visualize')}</QueryFieldGroup.Label>
+            <MRISelect
+              onChange={handleMRIChange}
+              onTagClick={handleMetricTagClick}
+              onOpenMenu={handleOpenMetricsMenu}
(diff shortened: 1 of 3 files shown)


vgrozdanic:
When you select very long metric name, currently this box would re-adjust and become larger, with this changes, that is no longer the case, and full metric name can't fit in the box anymore (metric name from the screenshot is: allocation_policy.db_request_rejected)
image

static/app/components/metrics/queryBuilder.tsx
```diff
@@ -216,13 +216,45 @@ export const QueryBuilder = memo(function QueryBuilder({
...
-              <QueryFieldGroup>
-                <QueryFieldGroup.Label>{t('Visualize')}</QueryFieldGroup.Label>
+        {!hasMetricsNewInputs(organization) && (
+          <FlexBlock>
+            <GuideAnchor
+              target="metrics_selector"
+              position="bottom"
+              disabled={index !== 0}
+            >
+              {hasMetricsNewInputs(organization) ? (
```
vgrozdanic, line 244: This check is redundant, it will always be false, because of the check on line 237

priscilawebdev: oh yes! good catch

vgrozdanic:
There is still regression from the current state:
image

new state:
image

The metric name still has 3 dots, but it is not shown the same way as it was before

priscilawebdev:
@vgrozdanic thanks for the great feedback! This issue only happens on very small viewports and the ellipsis you're referencing isn't coming from the styles, but we do trim the string if it's too long via the function https://github.com/getsentry/sentry/blob/0067c2809f76a9dc9b3885d3181cb42b909e4740/static/app/utils/string/middleEllipsis.tsx#L15 . I can create a separate issue, so we can allocate some time to fix it, but merging this PR is fine. This is not a regression, in the old experience we did not respect the max-width of the parent element and the content is overflowing

matejminar (approved):

Issue #74254, opened 2024-07-15
Delete connected dashboard widgets when a span metric is deleted


matejminar:

vgrozdanic:
Cascade deletion of widgets is not possible, due to the possible problem with permissions that could occur.

Widget deletion requires `org:write` permission, and span extraction rule requires only `project:read` permission. If we would allow widget to be deleted via span extraction rule, we would be breaking our permission system, and allow deletion of the widgets to users who shouldn't have permission to do it

vgrozdanic:
Cascade deletion of widgets is not possible, due to the possible problem with permissions that could occur.

Widget deletion requires `org:write` permission, and span extraction rule requires only `project:read` permission. If we would allow widget to be deleted via span extraction rule, we would be breaking our permission system, and allow deletion of the widgets to users who shouldn't have permission to do it

PR #74945, opened 2024-07-25
feat(metrics) Metric usage endpoint

vgrozdanic:
Endpoint which returns all the Alerts and Dashboard widgets that use MRIs given as a request parameter

Part of https://github.com/getsentry/sentry/issues/74254 and https://github.com/getsentry/sentry/issues/74255


src/sentry/incidents/models/alert_rule.py
```diff
@@ -223,6 +223,20 @@ def conditionally_subscribe_project_to_alert_rules(
...
+        """
+        Fetches AlertRules associated with the metric MRIs
+        """
+
+        alert_query = Q()
+        for metric_mri in metric_mris:
+            alert_query |= Q(snuba_query__aggregate__contains=metric_mri)
+
+        queryset = self.filter(organization=organization).filter(alert_query)
+        return queryset
```
obostjancic, line 238: nit: I'd expect this to be in `MetricsExtractionRule` since we only need this for Metric Extraction Rules, unless there is some other use case where we would want to query "which MRI is this specific alert depending on?", but then it would make sense to rewrite this into a method like `def get_metric(self) -> MRI`. 
Same for the widgets btw.

src/sentry/testutils/fixtures.py
```diff
@@ -627,6 +627,15 @@ def create_request_access(self, *args, **kwargs):
...
         return Factories.create_webhook_payload(*args, **kwargs)
 
+    def create_dashboard(self, *args, **kwargs):
+        return Factories.create_dashboard(*args, **kwargs)
+
+    def create_dashboard_widget(self, *args, **kwargs):
+        return Factories.create_dashboard_widget(*args, **kwargs)
+
+    def create_dashboard_widget_query(self, *args, **kwargs):
+        return Factories.create_dashboard_widget_query(*args, **kwargs)
```
obostjancic, line 637: these are great! 😍 Can we create a followup ticket to use them in other tests? AFAIK there are a couple of helpers that create widgets and dashboards that could all be streamlined to this.

tests/sentry/api/endpoints/test_organization_metrics_usage.py
```diff
@@ -0,0 +1,151 @@
...
+
+class TestOrganizationMetricsUsage(APITestCase):
+    endpoint = "sentry-api-0-organization-metrics-usage"
+
+    def setUp(self) -> None:
+        super().setUp()
+        self.login_as(user=self.user)
+
+        self.mri1 = "c:foo/1"
+        self.mri2 = "c:bar/2"
```
obostjancic, line 12: nit: these are actually not valid MRIs. 😅
```suggestion
        self.mri1 = "c:custom/foo1@none"
        self.mri2 = "c:custom/bar2@none"
```

obostjancic (approved):
Feature wise, LGTM. Regarding implementation, maybe go through the comments i left together with @shellmayr.

src/sentry/api/urls.py
```diff
@@ -1502,6 +1503,11 @@ def create_group_urls(name_prefix: str) -> list[URLPattern | URLResolver]:
         OrganizationMetricsExtractionRulesEndpoint.as_view(),
         name="sentry-api-0-organization-metrics-extraction-rules",
     ),
+    re_path(
+        r"^(?P<organization_id_or_slug>[^\/]+)/metrics-usage/$",
```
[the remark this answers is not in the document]
vgrozdanic: Makes sense, I opted in for this way to make this API endpoint more robust, so that we can query it and get all the widgets and alerts where specific metric MRI is used (which doesn't have to come from span extraction rule).

We can also offer both options:
- if the `ruleId` is present in URL - API returns all the alerts and widgets for the rule
- if there is no `ruleId` - API returns all the alerts and widgets connected to MRIs from `metricMRIs` parameter

src/sentry/testutils/fixtures.py
vgrozdanic: I was planning to do it, I have it on my refactor TODO list :)

src/sentry/incidents/models/alert_rule.py
vgrozdanic: Actually here we are querying: "give me all alerts that depend on this MRIs", which is basically a special type of filtering on `AlertRule` queryset, and that is why I have added it as a method on the `AlertRuleManager`.

I have no strong opinions here, but from my point of view, these complex filtering options should be in the custom Manager, that's why the Managers exist in Django

src/sentry/api/urls.py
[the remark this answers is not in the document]
shellmayr: I agree with @obostjancic here, and would add that MRIs are not user-facing entities, so we should not make them queriable to the user. Span metrics are the future, we should treat the endpoint as such: how can we infer usage of a span metric? As far as I'm concerned, MRI is an implementation detail at this point and should therefore not be exposed to the user.
There is an additional question here I think - is the ID really the identifer we want to use? I can imagine this being more idiomatic by querying by the span attribute, but I haven't fully thought this through at the moment.

src/sentry/incidents/models/alert_rule.py
```diff
@@ -223,6 +223,20 @@ def conditionally_subscribe_project_to_alert_rules(
             )
         return []
 
+    def get_for_metrics(
+        self, organization: Organization, metric_mris: list[str]
```
shellmayr, line 227: Since span metrics are defined on a per-project level, shouldn't this take project instead of organization as a parameter?

tests/sentry/incidents/models/test_alert_rule.py
```diff
@@ -211,6 +211,35 @@ def test_conditionally_subscribing_project_initializes_activation(self):
             assert current_activation.query_subscription == sub
             assert current_activation.is_complete() is False
 
+    def test_get_for_metrics(self):
```
shellmayr, line 214: We should test for prohibited access patterns here - test whether this is guarded against cross-org/cross-project access.

src/sentry/api/urls.py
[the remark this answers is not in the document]
vgrozdanic: Span attribute might be tricky, since it can be the same for multiple projects inside the organization.

To me, it makes the most sense if we for now use the id of the config.

tests/sentry/incidents/models/test_alert_rule.py
vgrozdanic: That is already being done in the test, unless I am missing smth?

vgrozdanic: Implemented in the test with endpoints

src/sentry/incidents/models/alert_rule.py
vgrozdanic: After our conversation, I have decided to keep organization as an argument since we filter only on it later in the method

src/sentry/api/urls.py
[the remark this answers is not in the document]
vgrozdanic: After the discussion, it was concluded that the access pattern will be (org.slug, project.slug, span_attribute) in an URL

src/sentry/api/endpoints/project_metrics_usage.py
```diff
@@ -0,0 +1,98 @@
...
+from sentry.api.base import region_silo_endpoint
+from sentry.api.bases.project import ProjectEndpoint
+from sentry.api.serializers.rest_framework.dashboard import DashboardWidgetQuerySerializer
+from sentry.incidents.models.alert_rule import AlertRule
+from sentry.models.dashboard_widget import DashboardWidget
+from sentry.models.project import Project
+from sentry.sentry_metrics.models import SpanAttributeExtractionRuleConfig
+
+
+class MetricsUsageAlertSerializer(serializers.Serializer):
```
shellmayr, line 17: Serializers should be located with the others at https://github.com/getsentry/sentry/blob/master/src/sentry/api/serializers/models/metrics_extraction_rules.py or in the appropriate file w.r.t. widgets/alerts.

src/sentry/api/urls.py
```diff
@@ -1502,6 +1503,11 @@ def create_group_urls(name_prefix: str) -> list[URLPattern | URLResolver]:
         OrganizationMetricsExtractionRulesEndpoint.as_view(),
         name="sentry-api-0-organization-metrics-extraction-rules",
     ),
+    re_path(
+        r"^(?P<organization_id_or_slug>[^\/]+)/(?P<project_id_or_slug>[^\/]+)/metrics-usage/(?P<span_attribute>[^\/]+)/$",
+        ProjectMetricsUsageEndpoint.as_view(),
+        name="sentry-api-0-project-metrics-usage",
+    ),
```
shellmayr, line 1510: This should probably be moved to the PROJECT_URLS instead of ORGANIZATION_URLS: https://github.com/getsentry/sentry/blob/7756b739c0bdedc8840fcc799999a5da4e54999f/src/sentry/api/urls.py#L2180

src/sentry/api/endpoints/project_metrics_usage.py
```diff
@@ -0,0 +1,98 @@
...
+    publish_status = {
+        "GET": ApiPublishStatus.EXPERIMENTAL,
+    }
+
+    def get(self, request: Request, project: Project, span_attribute: str) -> Response:
+        """
+        Retrieve the metrics usage inside the organization based on the span attribute extraction rule.
+        """
+
+        span_extraction_rule_config = get_object_or_404(
```
shellmayr, line 72: This is a nice method, did not know this existed

shellmayr (changes requested):

src/sentry/api/urls.py
vgrozdanic: True, good catch

src/sentry/api/endpoints/project_metrics_usage.py
vgrozdanic: Moved

shellmayr (approved):

PR #74950, opened 2024-07-25
fix(profiling): offset profile from txn based on the first sample timestamp

JonasBa:
Offset both txn and spans view from the startTime we receive in the qs


Zylphrex (approved):

static/app/components/profiling/flamegraph/flamegraphSpans.tsx
```diff
@@ -127,8 +127,8 @@ export function FlamegraphSpans({
 
     const drawSpans = () => {
       spansRenderer.draw(
-        spansView.configView.transformRect(spansView.configSpaceTransform),
```
JonasBa, line 130: This wasnt taking into account the offset config space and was rendering the span bars slightly differently

PR #74961, opened 2024-07-25
fix(metrics): Uppercase for new styles only

priscilawebdev:
### Goal

- The symbols shall be rendered as uppercase when the feature flag 'metrics-new-inputs' is true
- The symbols shall be rendered as lowercase when the feature flag 'metrics-new-inputs' is false
- Users shall be able to type in lowercase and uppercase independently of the feature flag


shellmayr (approved):
Looks good to me from a usage point of view, didn't find anything that breaks anymore. (can't comment on the code here)

obostjancic (approved):

PR #74981, opened 2024-07-25
feat(priority-alerts): Support alert previews for high priority alert conditions

### LINKED ISSUE #74819: Add support for alert preview for new and existing high priority issue conditions

snigdhas:
Add support for the alert preview for new and existing high priority alert conditions. The preview is populated using the `get_activity` method for these conditions, which needed some updates to accurately filter to the appropriate priority. 

The preview logic also has some short-circuit logic around mutually exclusive conditions. The high priority alert conditions are not mutually exclusive with all the previously supported conditions, so we're adding a map of explicitly supported pairings `VALID_CONDITION_PAIRS`. The mutual exclusivity check now checks if the multiple conditions are valid before possibly short circuiting to an empty result.

Fixes https://github.com/getsentry/sentry/issues/74819

### DIFF
--- a/src/sentry/rules/conditions/existing_high_priority_issue.py
+++ b/src/sentry/rules/conditions/existing_high_priority_issue.py
@@ -33,19 +33,24 @@ class ExistingHighPriorityIssueCondition(EventCondition):
                 project=self.project,
                 datetime__gte=start,
                 datetime__lt=end,
-                type__in=[ActivityType.SET_ESCALATING.value],
+                type__in=[ActivityType.SET_PRIORITY.value],
                 user_id=None,
             )
             .order_by("-datetime")[:limit]
             .values_list("group", "datetime", "data")
         )
 
+        activities = [
+            (group_id, timestamp, data)
+            for group_id, timestamp, data in activities
+            if data.get("priority") == "high"
+        ]
+
         return [
             ConditionActivity(
                 group_id=group_id,
-                type=ConditionActivityType.REAPPEARED,
+                type=ConditionActivityType.EXISTING_HIGH_PRIORITY_ISSUE,
                 timestamp=timestamp,
-                data=data,
             )
             for group_id, timestamp, data in activities
             if group_id is not None
--- a/src/sentry/rules/conditions/existing_high_priority_issue.py
+++ b/src/sentry/rules/conditions/existing_high_priority_issue.py
@@ -33,18 +33,13 @@ class ExistingHighPriorityIssueCondition(EventCondition):
                 project=self.project,
                 datetime__gte=start,
                 datetime__lt=end,
+                data={"priority": "high", "reason": "escalating"},
                 type__in=[ActivityType.SET_PRIORITY.value],
                 user_id=None,
             )
             .order_by("-datetime")[:limit]
             .values_list("group", "datetime", "data")
         )
-
-        activities = [
-            (group_id, timestamp, data)
-            for group_id, timestamp, data in activities
-            if data.get("priority") == "high"
-        ]
 
         return [
             ConditionActivity(
--- a/src/sentry/rules/conditions/new_high_priority_issue.py
+++ b/src/sentry/rules/conditions/new_high_priority_issue.py
@@ -2,11 +2,10 @@ from collections.abc import Sequence
 from datetime import datetime
 
 from sentry.eventstore.models import GroupEvent
-from sentry.models.activity import Activity
+from sentry.models.group import Group
 from sentry.receivers.rules import has_high_priority_issue_alerts
 from sentry.rules import EventState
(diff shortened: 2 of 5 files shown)


MichaelSun48 (approved):

src/sentry/rules/conditions/existing_high_priority_issue.py
```diff
@@ -33,19 +33,24 @@ def get_activity(
                 project=self.project,
                 datetime__gte=start,
                 datetime__lt=end,
-                type__in=[ActivityType.SET_ESCALATING.value],
+                type__in=[ActivityType.SET_PRIORITY.value],
```
JoshFerge, line 36: why was this escalating previously?

JoshFerge: oh are we making this more general? so all high priority activites will trigger this now?

src/sentry/rules/conditions/existing_high_priority_issue.py
```diff
@@ -33,19 +33,24 @@ def get_activity(
...
                 user_id=None,
             )
             .order_by("-datetime")[:limit]
             .values_list("group", "datetime", "data")
         )
 
+        activities = [
```
JoshFerge, line 43: any reason not to just add this filter in the above query?

src/sentry/rules/conditions/existing_high_priority_issue.py
```diff
@@ -33,19 +33,24 @@ def get_activity(
...
+            (group_id, timestamp, data)
+            for group_id, timestamp, data in activities
+            if data.get("priority") == "high"
+        ]
+
         return [
             ConditionActivity(
                 group_id=group_id,
-                type=ConditionActivityType.REAPPEARED,
+                type=ConditionActivityType.EXISTING_HIGH_PRIORITY_ISSUE,
```
JoshFerge, line 52: why is this changing?

src/sentry/rules/conditions/new_high_priority_issue.py
```diff
@@ -34,26 +33,19 @@ def passes(self, event: GroupEvent, state: EventState) -> bool:
...
-                datetime__lt=end,
-                type__in=[ActivityType.SET_UNRESOLVED.value],
-                user_id=None,
+                first_seen__gte=start,
+                first_seen__lt=end,
+                priority=PriorityLevel.HIGH,
             )
-            .order_by("-datetime")[:limit]
-            .values_list("group", "datetime", "data")
+            .order_by("-first_seen")[:limit]
```
JoshFerge, line 43: why are we changing from `-datetime` to `first_seen`?

src/sentry/rules/history/preview.py
```diff
@@ -62,13 +70,27 @@ def preview(
...
+
+        # all the issue state conditions are mutually exclusive
+        if not any(condition in VALID_CONDITION_PAIRS for condition in condition_ids):
+            return {}
+
+        # if there are multiple issue state conditions, they must be one of the valid pairs
+        for condition in condition_ids:
+            if (
+                condition in VALID_CONDITION_PAIRS
+                and VALID_CONDITION_PAIRS[condition] not in condition_ids
```
JoshFerge, line 91: for `VALID_CONDITION_PAIRS`, is there a guaranteed sorting? do you need the reverse in the dictionary as well?

JoshFerge (commented):
is this change behind a feature flag?

src/sentry/rules/conditions/existing_high_priority_issue.py
snigdhas: We're not generalizing the alert, but this was previous activity filter was incorrect. 

`SET_ESCALATING` is a superset of `SET_PRIORITY` since not all escalating groups have high priority. If we have an Activity entry with `SET_PRIORITY` and a priority of high though, that means the group escalated and has high priority and would trigger the alert.

src/sentry/rules/conditions/existing_high_priority_issue.py
snigdhas: REAPPEARED is used for escalating errors, but this condition has more nuance so I've added two new ConditionActivityType values.

src/sentry/rules/conditions/new_high_priority_issue.py
snigdhas: We're querying the Group table to get new groups with priority - new groups don't have an activity set for this. The Group table does not have a `datetime` column.

src/sentry/rules/history/preview.py
snigdhas: No guaranteed order and I think this would've worked since we're iterating through the entire list of conditions, but added the reverse to make it explicit and clearer.

PR #74992, opened 2024-07-25
feat(trace): expand tree if there are less than 3 transaction

JonasBa:
When coming to small traces, the UI looks empty as we are not expanding the transactions by default. This implements automatic expanding of traces with less than 3 transactions


mjq (approved):

Abdkhan14 (approved):
Tested it locally lets 🚢

PR #74999, opened 2024-07-25
fix(insights): move useHasData to ModulePageProviders

KevinL10:
Fixes issue where useHasData does not send any analytic events when has_data_ever_sent is False.


gggritso (approved):
👍🏻

PR #75012, opened 2024-07-25
feat(insights): add release markings to fullscreen insight charts

KevinL10:
let's try https://github.com/getsentry/sentry/pull/74656 again

releases on span samples look like:
image


gggritso (approved):
👍🏻

PR #75014, opened 2024-07-25
chore: Update pegjs-loader dependency

ryan953:
`yarn add pegjs-loader@^0.5.8`

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
     "moment-timezone": "0.5.44",
     "papaparse": "^5.3.2",
     "pegjs": "^0.10.0",
-    "pegjs-loader": "^0.5.6",
+    "pegjs-loader": "^0.5.8",
     "platformicons": "^6.0.1",
     "po-catalog-loader": "2.0.0",
     "prettier": "3.3.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -263,7 +263,9 @@
       "last 3 iOS major versions",
       "Firefox ESR"
     ],
-    "test": ["current node"]
+    "test": [
+      "current node"
+    ]
```
ryan953, line 268: love this linter battle playing out

yarn.lock
```diff
@@ -8659,15 +8659,14 @@ loader-utils@1.1.0:
     emojis-list "^2.0.0"
     json5 "^0.5.0"
 
-loader-utils@^0.2.5:
```
ryan953, line 8662: dropping this will fix https://github.com/getsentry/sentry/security/dependabot/117

gggritso (approved):
🚢 👍🏻

webpack.config.ts
```diff
@@ -269,8 +269,8 @@ const appConfig: webpack.Configuration = {
         },
       },
       {
-        test: /\.pegjs/,
-        use: {loader: 'pegjs-loader'},
+        test: /\.pegjs$/,
+        use: ['pegjs-loader?cache=false&optimize=speed'],
```
ryan953, line 273: These are default value (See https://github.com/eploko/pegjs-loader?tab=readme-ov-file#pegjs-options)

I set them in here so that they can be parsed by loader-utils here: https://github.com/webpack/loader-utils/blob/v1.4.2/lib/parseQuery.js#L12

PR #75017, opened 2024-07-25
fix(relocation): Change relocation_export RPC type

azaslavsky:
Fixes: SENTRY-3CDM

Previously, we tried to send a raw `bytes` object. While pydantic allows this from a type checking perspective, it fails when we try to encode the data as a JSON payload for transport over the wire, which requires UTF-8 encoding.

To get around this, we change the type of `encrypted_contents` to `list[int]`, then cast the `bytes` to/from a `list[int]` when alternating between sending over the wire and saving to a filestore.

### DIFF
--- a/src/sentry/receivers/outbox/control.py
+++ b/src/sentry/receivers/outbox/control.py
@@ -172,5 +172,5 @@ def process_relocation_reply_with_export(payload: Mapping[str, Any], **kwds):
             requesting_region_name=payload["requesting_region_name"],
             replying_region_name=payload["replying_region_name"],
             org_slug=payload["org_slug"],
-            encrypted_contents=encrypted_contents.read(),
+            encrypted_contents=[int(byte) for byte in encrypted_contents.read()],
         )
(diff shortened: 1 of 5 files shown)


src/sentry/receivers/outbox/control.py
```diff
@@ -162,15 +162,17 @@ def process_relocation_reply_with_export(payload: Mapping[str, Any], **kwds):
     relocation_storage = get_relocation_storage()
     path = f"runs/{uuid}/saas_to_saas_export/{slug}.tar"
     try:
-        encrypted_contents = relocation_storage.open(path)
+        encrypted_bytes = relocation_storage.open(path)
```
hubertdeng123, line 165: What happens if you do something like relocation_sorage.open(path, encoding="utf-8") ? This seems like a potentially risky change

azaslavsky: It will fail, which is fine, because the file we're storing is not a string? It's an encrypted tarball - opening it incorrectly will result in it not opening.

hubertdeng123: ah ok, so we need to ensure that the bytes are preserved as is in order to be decrypted later

hubertdeng123 (approved):

PR #75026, opened 2024-07-25
chore(similarity): Turn on ML grouping for new EA projects

jangjodi:
Add project option to turn on ML similarity grouping for newly created projects that belong to an EA org


JoshFerge (approved):
is there anywhere else a project can be created? or is this the only one

src/sentry/api/endpoints/team_projects.py
```diff
@@ -214,4 +217,16 @@ def post(self, request: Request, team) -> Response:
                 sender=self,
             )
 
+            # Create project option to turn on ML similarity feature for new EA projects
+            seer_eligible_platform = (
+                project.platform and project.platform in SEER_ELIGIBLE_PLATFORMS
+            )
+            if (
+                project
```
mrduncan, line 225: I think this should already be truthy by this point since line 222 is looking at `project.platform`?

src/sentry/api/endpoints/team_projects.py
```diff
@@ -214,4 +217,16 @@ def post(self, request: Request, team) -> Response:
                 sender=self,
             )
 
+            # Create project option to turn on ML similarity feature for new EA projects
+            seer_eligible_platform = (
+                project.platform and project.platform in SEER_ELIGIBLE_PLATFORMS
```
mrduncan, line 222:
```suggestion
                project.platform in SEER_ELIGIBLE_PLATFORMS
```
nit: shouldn't need to check `project.platform` here unless I'm missing something

src/sentry/api/endpoints/team_projects.py
```diff
@@ -214,4 +217,16 @@ def post(self, request: Request, team) -> Response:
                 sender=self,
             )
 
+            # Create project option to turn on ML similarity feature for new EA projects
+            seer_eligible_platform = (
```
lobsterkatie, line 221: Nit:

```suggestion
            is_seer_eligible_platform = (
```

Just to make it obvious it's a boolean.

tests/sentry/api/endpoints/test_team_projects.py
```diff
@@ -228,3 +237,61 @@ def test_default_inbound_filters(self):
         }
         assert javascript_filter_states["web-crawlers"]
         assert javascript_filter_states["filtered-transaction"]
+
+    def test_similarity_project_option_valid(self):
+        """
+        Test that project option for similarity grouping is created for EA organizations
+        where the project platform is python and seer eligible.
```
lobsterkatie, line 244: Nit:

```suggestion
        where the project platform is Seer-eligible.
```

tests/sentry/api/endpoints/test_team_projects.py
```diff
@@ -228,3 +237,61 @@ def test_default_inbound_filters(self):
...
+            self.team.slug,
+            **self.data,
+            status_code=201,
+        )
+
+        project = Project.objects.get(id=response.data["id"])
+        assert project.name == "foo"
+        assert project.slug == "bar"
+        assert project.platform == "python"
+        assert project.teams.first() == self.team
```
lobsterkatie, line 260: Why do we need to assert on these? Do they have anything to do with the projectoption? (Same question for the test below.)

src/sentry/api/endpoints/team_projects.py
jangjodi: good catches, thanks!

jangjodi:
> is there anywhere else a project can be created? or is this the only one

As far as I'm aware, this is the only way. There's a OrganizationProjectsEndpoint, but it doesn't have a post function

tests/sentry/api/endpoints/test_team_projects.py
jangjodi: I just included these to ensure that the change doesn't mess up project creation at all

PR #75029, opened 2024-07-25
feat(tabs): Add new variant to be used on issue details

roggenkemper:
this pr adds a new variant of tabs - "floating" , which will be used on the updated issue details page

### DIFF
--- a/static/app/components/tabs/index.stories.tsx
+++ b/static/app/components/tabs/index.stories.tsx
@@ -209,6 +209,22 @@ export default storyBook(Tabs, story => {
           </TabPanels>
         </Tabs>
       </SizingWindow>
+      <br />
+      <p>You can also use the "issue-details" variant, which is used below</p>
+      <SizingWindow>
+        <Tabs>
+          <TabList variant="issue-details" hideBorder>
+            {TABS.map(tab => (
+              <TabList.Item key={tab.key}>{tab.label}</TabList.Item>
+            ))}
+          </TabList>
+          <TabPanels>
+            {TABS.map(tab => (
+              <TabPanels.Item key={tab.key}>{tab.content}</TabPanels.Item>
+            ))}
+          </TabPanels>
+        </Tabs>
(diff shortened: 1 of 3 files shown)


static/app/components/tabs/tab.tsx
```diff
@@ -99,13 +99,28 @@ export const BaseTab = forwardRef(
...
 
+    if (variant === 'floating') {
+      return (
+        <FloatingTabWrap
+          {...tabProps}
+          hidden={hidden}
+          overflowing={overflowing}
+          ref={ref}
+        >
+          <VariantStyledInteractionStateLayer />
```
malwilley, line 117: One thing I'm not seeing are pressed styles (when you hold the mouse down) and hover styles on the selected tab. I think you can fix that by adding `hasSelectedBackground={false}` and adding a background in the CSS when it's selected.

static/app/components/tabs/tab.tsx
```diff
@@ -164,6 +179,43 @@ export const Tab = forwardRef(
...
+    `}
+  }
+
+  &[aria-selected='false'] {
+    border-top: 1px solid transparent;
+  }
+
+  color: ${p => p.theme.gray300};
+
+  border-radius: 6px;
```
malwilley, line 199: nit: lots of whitespace here between styles

malwilley (approved):

static/app/components/tabs/index.stories.tsx
```diff
@@ -209,6 +209,22 @@ export default storyBook(Tabs, story => {
           </TabPanels>
         </Tabs>
       </SizingWindow>
+      <br />
+      <p>You can also use the "floating" variant, which is used below</p>
+      <SizingWindow>
+        <Tabs>
+          <TabList variant="floating" hideBorder>
```
ryan953, line 216: This could use a Matrix to better show the different values along side each other, and in both orientations

PR #75030, opened 2024-07-25
chore: Guard against bad imports related to devtoolbar

ryan953:
The toolbar is a little world, we shouldn't import from it (but it's re-using things from all over sentry!)

Also, there's some specific spots that it shouldn't be importing from. Right now just queryClient, but over time that list will probably grow as they get further apart.

The getsentry version of this is https://github.com/getsentry/getsentry/pull/14747

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,6 +26,17 @@ module.exports = {
       'error',
       {additionalHooks: '(useEffectAfterFirstRender|useMemoWithPrevious)'},
     ],
+    'no-restricted-imports': [
+      'error',
+      {
+        patterns: [
+          {
+            group: ['sentry/components/devtoolbar/*'],
+            message: 'Do not depend on toolbar internals',
+          },
+        ],
+      },
+    ],
 
     // TODO(@anonrig): Remove this from eslint-sentry-config
     'space-infix-ops': 'off',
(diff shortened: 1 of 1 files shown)


priscilawebdev:
we have a few places in sentry that use devtoolbars imports. Everything seems to pass now, but I wonder if we will get lint errors after merging this PR. Also, can we add a note to the devtoolbars exports so people see them when trying to import?

evanpurkhiser (approved):

ryan953:
> we have a few places in sentry that use devtoolbars imports. Everything seems to pass now, but I wonder if we will get lint errors after merging this PR. Also, can we add a note to the devtoolbars exports so people see them when trying to import?

I don't see anything external to the toolbar itself doing an import, so if there is something and linter starts complaining i'll go fix it right away because it shouldn't do that. 

I don't think it's going to be practical to annotate all the exports from within the toolbar folder, just because it's any file really that's inside there. I think the most likely thing to get accidentally imported is one of the hooks, specifically if someone is letter vscode choose where to import from. I think i can followup and rename some things so they're less appealing. 

I'm mostly just trying to add a layer that helps prevent accidents. I've been saying it for a bit longer than I expected, but this folder is not forever, once we solve some auth related stuff we'll be unblocked and can move this to a new repo.

PR #75031, opened 2024-07-25
fix(autofix): update markdown rendering to be more readable

roaga:
Remove big headers from markdown in logs, remove huge padding, and allow horizontal scrolling for code snippets and logs that overflow off screen.

### DIFF
--- a/static/app/components/events/autofix/autofixSteps.tsx
+++ b/static/app/components/events/autofix/autofixSteps.tsx
@@ -111,13 +111,33 @@ function Progress({
   progress: AutofixProgressItem | AutofixStep;
   runId: string;
 }) {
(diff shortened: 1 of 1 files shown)


static/app/components/events/autofix/autofixSteps.tsx
```diff
@@ -111,13 +111,33 @@ function Progress({
   progress: AutofixProgressItem | AutofixStep;
   runId: string;
 }) {
+  function replaceHeadersWithBold(markdown) {
```
jennmueng, line 114: minor nit: let's type this
```suggestion
  function replaceHeadersWithBold(markdown: string) {
```

jennmueng (approved):
minor nit otherwise LGTM

PR #75035, opened 2024-07-26
ref(alerts): Refactor out alert badge and tooltip

davidenwang:
Refactor out the alert badge and tooltip between the `activatedRuleRow.tsx` and `row.tsx`. Eventually will extend that component to accomodate an uptime alert badge/tooltip.

Also added tests to the component

related to: https://github.com/getsentry/team-core-product-foundations/issues/289

### DIFF
--- a/static/app/views/alerts/list/rules/combinedAlertBadge.spec.tsx
+++ b/static/app/views/alerts/list/rules/combinedAlertBadge.spec.tsx
@@ -1,0 +1,73 @@
+import {MetricRuleFixture} from 'sentry-fixture/metricRule';
+import {MetricRuleActivationFixture} from 'sentry-fixture/metricRuleActivation';
+import {ProjectAlertRuleFixture} from 'sentry-fixture/projectAlertRule';
+
+import {render, screen, userEvent} from 'sentry-test/reactTestingLibrary';
+
+import {MonitorType} from 'sentry/types/alerts';
+import CombinedAlertBadge from 'sentry/views/alerts/list/rules/combinedAlertBadge';
+import {CombinedAlertType} from 'sentry/views/alerts/types';
+
+describe('CombinedAlertBadge', function () {
+  it('Renders correctly for waiting metric alert rules', async function () {
+    const rule = {
+      ...MetricRuleFixture({monitorType: MonitorType.ACTIVATED}),
+      // Cast here to inform typescript that this will always be a metric type rule
+      type: CombinedAlertType.METRIC as CombinedAlertType.METRIC,
+    };
+
+    render(<CombinedAlertBadge rule={rule} />);
+
+    await userEvent.hover(screen.getByTestId('alert-badge'));
+
+    // Renders tooltip with correct text
+    expect(
+      await screen.findByText('Metric Alert Status: Ready to monitor')
+    ).toBeInTheDocument();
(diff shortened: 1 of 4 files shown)


static/app/views/alerts/list/rules/combinedAlertBadge.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+import {CombinedAlertType, type MetricAlert} from 'sentry/views/alerts/types';
+
+describe('CombinedAlertBadge', function () {
+  it('Renders correctly for waiting metric alert rules', async function () {
+    const rule = {
+      ...MetricRuleFixture({monitorType: MonitorType.ACTIVATED}),
+      type: CombinedAlertType.METRIC,
+    };
+
+    render(<CombinedAlertBadge rule={rule as MetricAlert} />);
```
evanpurkhiser, line 18: Why do we need the cast here?

static/app/views/alerts/list/rules/combinedAlertBadge.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+interface Props {
+  rule: CombinedMetricIssueAlerts;
+}
+
+const IncidentStatusText: Record<IncidentStatus, string> = {
+  [IncidentStatus.CRITICAL]: t('Critical'),
+  [IncidentStatus.WARNING]: t('Warning'),
+  [IncidentStatus.CLOSED]: t('Resolved'),
+  [IncidentStatus.OPENED]: t('Resolved'),
```
evanpurkhiser, line 20: Is this right?

evanpurkhiser: I know it's copy paste but, seems weird?

static/app/views/alerts/list/rules/combinedAlertBadge.spec.tsx
davidenwang: it's because typescript is scared that `type` can change in the future and then the `rule` object wouldn't fit into any of the `CombinedAlert` alerts.

Another way to write it could be like

```
    const rule = {
      ...MetricRuleFixture({monitorType: MonitorType.ACTIVATED}),
      type: CombinedAlertType.METRIC as CombinedAlertType.METRIC,
    };

    render(<CombinedAlertBadge rule={rule} />);
```

or declaring the `rule` object `as const`, which I also ended up using. Not sure which one is preferred here

static/app/views/alerts/list/rules/combinedAlertBadge.tsx
davidenwang: yea I saw that too and was confused but didn't want to change any functionality with this refactor PR. Judging from the backend this could be a bug.
https://github.com/getsentry/sentry/blob/5b00015693fdc5dada4f7916853e04bf560c2139/src/sentry/incidents/models/incident.py#L168


 I'll open another PR to correct the text and maybe tag alerts team on it or dan if they have more context.

davidenwang: https://github.com/getsentry/sentry/pull/75456

evanpurkhiser (approved):

PR #75036, opened 2024-07-26
ref(alerts): Reconcile activatedRuleRow into row.tsx, remove activatedRuleRow

davidenwang:
Much of the code in `activatedRuleRow` was duplicated from `row.tsx` 

However, following the refactors from:
https://github.com/getsentry/sentry/pull/75035
https://github.com/getsentry/sentry/pull/74969

The `activatedRuleRow` component can rejoin its sister component `row.tsx` after these changes


static/app/views/alerts/list/rules/row.tsx
```diff
@@ -57,7 +59,24 @@ function RuleListRow({
   const [assignee, setAssignee] = useState<string>('');
   const activeIncident = hasActiveIncident(rule);
 
-  function renderLastIncidentDate(): React.ReactNode {
+  const isActivatedAlertRule =
+    !isIssueAlert(rule) && rule.monitorType === MonitorType.ACTIVATED;
+
+  // TODO(davidenwang): Decompose this further
```
davidenwang, line 65: not sure how i feel about this yet, but i think my idea would be to refactor it out similar to the `CombinedAlertBadge` and make the logic here clearer since it feels a bit fragmented

davidenwang: https://github.com/getsentry/sentry/pull/75309 ended up doing it here

evanpurkhiser (approved):

PR #75041, opened 2024-07-26
feat(discover): Expose trace.client_sample_rate

jan-auer:
The trace's client sample rate is indexed by Snuba but was not available
to Discover. This PR lists the field so that it can be shown as a column
and used in filters.

Snuba stores the client sample rate in the common contexts column as a
string value. Therefore, numeric operations cannot be supported yet.

### DIFF
--- a/src/sentry/api/event_search.py
+++ b/src/sentry/api/event_search.py
@@ -1207,6 +1207,7 @@ default_config = SearchConfig(
         "stack.lineno",
         "stack.stack_level",
         "transaction.duration",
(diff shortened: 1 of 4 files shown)


jan-auer:
Changes can be deployed in any order, a single PR is fine in this case.

priscilawebdev (approved):
can confirm that it works 

https://github.com/user-attachments/assets/49ccf872-7098-42d9-b40f-81fc9038071f

Issue #74938, opened 2024-07-25
Stop using `sentry_sdk.push_scope`


szokeasaurusrex:
We intend to deprecate `sentry_sdk.push_scope` (see [getsentry/sentry-python#3347](https://github.com/getsentry/sentry-python/issues/3347)), so we should stop using it in Sentry.

PR #75042, opened 2024-07-26
ref(snuba): Replace `push_scope` in `mri.py`

szokeasaurusrex:
`sentry_sdk.push_scope` is soft-deprecated, and will soon be hard- deprecated (see [sentry-python#3347](https://github.com/getsentry/sentry-python/issues/3347)), so replace the `push_scope` calls with appropriate new API. Here, we use `sentry_sdk.new_scope`, since we are setting tags for a `capture_message` call.

See #74938


shellmayr (approved):

shellmayr:
Thanks for noticing that!

PR #75043, opened 2024-07-26
ref(metrics): Change `push_scope` to `new_scope`

### LINKED ISSUE #74938: Stop using `sentry_sdk.push_scope`

szokeasaurusrex:
`sentry_sdk.push_scope` is soft-deprecated, and will soon be hard- deprecated (see [sentry-python#3347](https://github.com/getsentry/sentry-python/issues/3347)), so replace the `push_scope` calls with appropriate new API. Here, we use `sentry_sdk.new_scope`, since we are setting tags for a `capture_message` call.

Closes #74938


shellmayr (approved):
LGTM

PR #75053, opened 2024-07-26
feat(grouping): Remove ability to pick different project configs

armenzg:
**NOTE**: Putting this on hold until we upgrade all project configs. See #75092 for what we're removing for now.

This feature is only available to staff, we do not use it and do not intend to use it.

Before:
image

After:
image


static/app/components/events/eventEntries.tsx
```diff
@@ -131,14 +130,7 @@ function EventEntries({
         />
       )}
       {!isShare && event.groupID && (
-        <EventGroupingInfo
-          projectSlug={projectSlug}
-          event={event}
-          showGroupingConfig={
-            orgFeatures.includes('set-grouping-config') && 'groupingConfig' in event
```
armenzg, line 138: Other usage of this flag can be seen here:
https://github.com/search?q=org%3Agetsentry+set-grouping-config+NOT+repo%3Agetsentry%2Fhackweek-profiling-vscode+NOT+repo%3Agetsentry%2Fsentry-mirror-suggested-fix+NOT+repo%3Agetsentry%2Fsuggested-fix-pr-demo&type=code

I will follow up with a different PR for that.

PR #75061, opened 2024-07-26
Add install and verify to Rails

smeubank:
Add install and verify to Rails

### DIFF
--- a/static/app/gettingStartedDocs/ruby/rails.tsx
+++ b/static/app/gettingStartedDocs/ruby/rails.tsx
@@ -19,11 +19,7 @@ type Params = DocsParams;
 const getInstallSnippet = (
   params: Params
 ) => `${params.isProfilingSelected ? 'gem "stackprof"\n' : ''}gem "sentry-ruby"
-gem "sentry-rails"
-
-bin/rails generate sentry
-
-bundle install`;
+gem "sentry-rails"`;
 
 const getConfigureSnippet = (params: Params) => `
 Sentry.init do |config|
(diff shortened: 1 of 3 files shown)


static/app/gettingStartedDocs/ruby/rails.tsx
```diff
@@ -19,7 +19,8 @@ type Params = DocsParams;
 const getInstallSnippet = (
   params: Params
 ) => `${params.isProfilingSelected ? 'gem "stackprof"\n' : ''}gem "sentry-ruby"
-gem "sentry-rails"`;
+gem "sentry-rails"
+bundle install`;
```
benvinegar, line 23: Not sure if I'm reading this right, but I think you have to separate `bundle install` into its own step. It's not a line you add to the Gemfile. It's a terminal command you run afterwards.

(Basically, it's like adding something to `package.json` then running `npm install` after.)

static/app/gettingStartedDocs/ruby/rails.tsx
```diff
@@ -57,7 +64,7 @@ const onboarding: OnboardingConfig = {
     {
       type: StepType.INSTALL,
       description: tct(
-        'Add [sentryRubyCode:sentry-ruby] and [sentryRailsCode:sentry-rails] to your [sentryGemfileCode:Gemfile]:',
+        'Add [sentryRubyCode:sentry-ruby] and [sentryRailsCode:sentry-rails] and generate your [sentryGemfileCode:Gemfile]:',
```
priscilawebdev, line 67: nit: instead of `and... and` shouldn't the first 'and' be ','?

static/app/gettingStartedDocs/ruby/rails.tsx
```diff
@@ -19,7 +19,11 @@ type Params = DocsParams;
 const getInstallSnippet = (
   params: Params
 ) => `${params.isProfilingSelected ? 'gem "stackprof"\n' : ''}gem "sentry-ruby"
-gem "sentry-rails"`;
+gem "sentry-rails"
```
priscilawebdev, line 22: this did not work for me, I had to use `gem install sentry-rails` instead.

sl0thentr0py (approved):
fixed some stuff and made the 3 guides consistent.

PR #75063, opened 2024-07-26
ref(querybuilder): Change facets to use snuba params

wmak:
- This updates the facets endpoint to use snuba params
- Also missed updating the discover module to use snuba params as well


tests/snuba/api/endpoints/test_organization_events_facets.py
```diff
@@ -340,6 +340,8 @@ def test_no_projects(self):
     def test_multiple_projects_without_global_view(self):
         self.store_event(data={"event_id": uuid4().hex}, project_id=self.project.id)
         self.store_event(data={"event_id": uuid4().hex}, project_id=self.project2.id)
+        self.organization.flags.allow_joinleave = False
+        self.organization.save()
```
Zylphrex, line 344: Is this related to this PR somehow?

wmak: Yeah, the check for projects is looser on the new params function. If you can join/leave you're allowed to get projects that you're not a part of. Since you can freely join and leave those projects anyways.

wmak: As discussed in slack this doesn't make any sense. Going to remove this and we'll have to merge this PR once the base get_snuba_dataclass function is fixed

Zylphrex (approved):

PR #75066, opened 2024-07-26
ref(querybuilder): Change root cause to use snuba_params

wmak:
- This updates the root cause endpoint to get its params via the dataclass

### DIFF
--- a/src/sentry/api/endpoints/organization_events_root_cause_analysis.py
+++ b/src/sentry/api/endpoints/organization_events_root_cause_analysis.py
@@ -32,9 +32,11 @@ RESPONSE_KEYS = [
 ]
 
 
-def init_query_builder(params, transaction, regression_breakpoint, limit, span_score_threshold):
-    before_minutes = int((regression_breakpoint - params["start"]).total_seconds() // 60)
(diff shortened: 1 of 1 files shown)


src/sentry/api/endpoints/organization_events_root_cause_analysis.py
```diff
@@ -190,14 +193,15 @@ def get(self, request, organization):
...
+        snuba_params, _ = self.get_snuba_dataclass(request, organization)
 
         with handle_query_errors():
             transaction_count_query = metrics_query(
                 ["count()"],
                 f'event.type:transaction transaction:"{transaction_name}"',
-                params,
+                params={},
                 referrer=BASE_REFERRER,
+                snuba_params=snuba_params,
```
Zylphrex, line 204: nit: move `snuba_params` next to `params`

Zylphrex (approved):

src/sentry/api/endpoints/organization_events_root_cause_analysis.py
wmak: :+1: I'll do that in a follow up

PR #75067, opened 2024-07-26
feat(discover): Maintain saved query dataset state

shruthilayaj:
Maintain state properly so editing column modals,
adding filters etc work as expected


narsaynorath (approved):
looks good to me, also ran it locally and the bug was fixed :+1:

PR #75069, opened 2024-07-26
fix(insights): overlay insights charts on top of detail panel

KevinL10:
Overlays the full-screen insights chart on top of the side panel.

image

### DIFF
--- a/static/app/components/slideOverPanel.tsx
+++ b/static/app/components/slideOverPanel.tsx
@@ -85,7 +85,7 @@ const _SlideOverPanel = styled(motion.div, {
   pointer-events: auto;
   overscroll-behavior: contain;
 
-  z-index: ${p => p.theme.zIndex.modal + 1};
+  z-index: ${p => p.theme.zIndex.modal - 1};
 
   box-shadow: ${p => p.theme.dropShadowHeavy};
   background: ${p => p.theme.background};


static/app/components/modals/insightChartModal.tsx
```diff
@@ -13,7 +13,20 @@ export type InsightChartModalOptions = {
...
-export default function InsightChartModal({Header, title, children}: Props) {
+export default function InsightChartModal({
+  Header,
+  title,
+  children,
+  modalContainerRef,
+}: Props) {
+  useEffect(() => {
+    // display the fullscreen insight charts on top of any sidebar elements
+    // with z-index 10000 or 10001
```
DominikB2014, line 24: I'm kind of suprised we have to do this, typically a modal should always be on top, no matter what what. If we have to manually update the zIndex for the modal, i feel as though the sidebar's zIndex is too high?

KevinL10: the modal has z-index 10000 but the sidebar has z-index 10001. I think this is the only place in the product where we possibly display both a modal and a sidebar

DominikB2014: Could we make the sidebar zIndex 9999?

DominikB2014: We would be future proofing ourselves here, there may be more cases for sidebar and modal at some point

DominikB2014 (approved):
👍

PR #75070, opened 2024-07-26
ref(querybuilder): Change profiling profiles to use snuba params

wmak:
- This updates the profiling profiles endpoint to use snuba params over the paramstype
- Needed to add quantize_date_params as a param to be compatible


tests/sentry/api/endpoints/test_organization_profiling_profiles.py
```diff
@@ -48,7 +48,7 @@ def test_more_than_one_project(self):
         assert response.status_code == 400, response.data
         assert response.data == {
             "detail": ErrorDetail(
-                "You cannot view events from multiple projects.",
```
wmak, line 51: Based on the name of this test, i think this is the correct error message?

src/sentry/search/events/types.py
```diff
@@ -105,6 +106,19 @@ def parse_stats_period(self) -> None:
 
             self.start = get_datetime_from_stats_period(self.stats_period, self.end)
 
+    @property
+    def start_date(self) -> datetime:
+        # This and end_date are helper functions so callers don't have to check if either are defined for typing
+        if self.start is None:
+            raise InvalidSearchQuery("start is required")
+        return self.start
```
wmak, line 114: I'm not hugely in love with this, might revisit it later to see if i can make it cleaner

Zylphrex (approved):

PR #75078, opened 2024-07-26
fix(query-builder): Improve selection behavior and consistency

malwilley:
This fixes a couple problems:

1. Sometimes the wrapper div would take focus from the selection key handler, which was causing all onKeyDown events to be ignored.
2. When there was a selection, arrow keys were not focusing the right elements because the events were not being prevented from propagating

### DIFF
--- a/static/app/components/searchQueryBuilder/hooks/useQueryBuilderGrid.tsx
+++ b/static/app/components/searchQueryBuilder/hooks/useQueryBuilderGrid.tsx
@@ -88,8 +88,11 @@ export function useQueryBuilderGrid({
       onKeyDownCapture: noop,
       onKeyDown,
       onFocus: (e: FocusEvent) => {
-        // If this element gets focused and there is a selection, focus the SelectionKeyHandler instead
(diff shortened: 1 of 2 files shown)


static/app/components/searchQueryBuilder/hooks/useQueryBuilderGrid.tsx
```diff
@@ -88,8 +88,11 @@ export function useQueryBuilderGrid({
       onKeyDownCapture: noop,
       onKeyDown,
       onFocus: (e: FocusEvent) => {
-        // If this element gets focused and there is a selection, focus the SelectionKeyHandler instead
-        if (e.target === ref.current && state.selectionManager.selectedKeys.size > 0) {
```
malwilley, line 92: The old logic looked at the selected keys state which was sometimes stale. We know that we never want this element to take focus away from the selection key handler, so we just look at the focus event instead to refocus the right element.

roggenkemper (approved):

PR #75084, opened 2024-07-26
ref(similarity): Add option to skip processed and chosen projects in backfill

jangjodi:
Add flags in similarity backfill to
- Allow already backfilled projects to be skipped
- Skip manually included projects that are causing failures that we're fixing

Admin endpoint change [here](https://github.com/getsentry/getsentry/pull/14752)

### DIFF
--- a/src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
+++ b/src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
@@ -34,6 +34,8 @@ class ProjectBackfillSimilarIssuesEmbeddingsRecords(ProjectEndpoint):
         last_processed_id = None
         only_delete = False
         enable_ingestion = False
+        skip_processed_projects = False
+        skip_project_ids = None
 
         if request.data.get("last_processed_id"):
             last_processed_id = int(request.data["last_processed_id"])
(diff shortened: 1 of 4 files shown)


src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
```diff
@@ -44,10 +46,18 @@ def post(self, request: Request, project) -> Response:
         if request.data.get("enable_ingestion"):
             enable_ingestion = request.data["enable_ingestion"] == "true"
 
+        if request.data.get("skip_processed_projects"):
+            skip_processed_projects = request.data["skip_processed_projects"] == "true"
```
JoshFerge, line 50: why use the string `"true"` instead of the boolean `True`?

JoshFerge (commented):
looks good other than note on string `true`!

src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
```diff
@@ -44,10 +46,18 @@ def post(self, request: Request, project) -> Response:
         if request.data.get("enable_ingestion"):
             enable_ingestion = request.data["enable_ingestion"] == "true"
```
JoshFerge, line 47: just noticing this here as well

src/sentry/api/endpoints/project_backfill_similar_issues_embeddings_records.py
jangjodi: good catch, thanks!

JoshFerge (approved):

PR #75087, opened 2024-07-26
fix(replay/issues): don't show onboarding banner if project already sent a replay

michellewzhang:
was trying to debug the onboarding panel state bug, but also realized that we're showing the onboarding banner even if the event has sent a replay already.

### DIFF
--- a/static/app/components/events/eventReplay/index.tsx
+++ b/static/app/components/events/eventReplay/index.tsx
@@ -7,6 +7,7 @@ import type {Event} from 'sentry/types/event';
 import type {Group} from 'sentry/types/group';
 import useEventCanShowReplayUpsell from 'sentry/utils/event/useEventCanShowReplayUpsell';
 import {getReplayIdFromEvent} from 'sentry/utils/replays/getReplayIdFromEvent';
+import {useHaveSelectedProjectsSentAnyReplayEvents} from 'sentry/utils/replays/hooks/useReplayOnboarding';
 
 interface Props {
   event: Event;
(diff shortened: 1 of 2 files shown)


static/app/components/events/eventReplay/index.tsx
```diff
@@ -28,7 +30,7 @@ export default function EventReplay({event, group, projectSlug}: Props) {
     return <ReplayClipSection event={event} replayId={replayId} group={group} />;
   }
 
-  if (canShowUpsell) {
+  if (canShowUpsell && !hasSentOneReplay) {
```
c298lee, line 33: what does canShowUpsell check?

c298lee (approved):

static/app/components/events/eventReplay/index.tsx
michellewzhang: https://github.com/getsentry/sentry/blob/0887140eb635c990b2b64f3ec2218feee3669fb9/static/app/utils/event/useEventCanShowReplayUpsell.tsx#L53-L55 

it checks
1) the platform is supported by replay
2) either the org has not sent any replays or the platform is part of the replay backend platforms

c298lee: oh why doesn't the check for the org not sending any replays work in this case?

michellewzhang: bc it's an or (`||`) so any backend platform (e.g. the platforms that we show the banner for) will automatically be true

c298lee: oh is there a reason why we can't remove the backend platform condition?

michellewzhang: the check targets whether the entire org has sent replays, but we want to show the backend banner even if the org has sent replays bc that specific project might not have sent replays yet, so we always override it with the backend platform check

PR #75089, opened 2024-07-26
chore(replay): Add HE events to replayAnalyticsEvents definition

aliu39:
Ref https://github.com/getsentry/sentry/issues/74456. The `trackAnalytics` calls were added last week in the 2 PRs linked in this issue. We have a dashboard in redash but events don't get sent to amplitude unless we add these

### DIFF
--- a/static/app/utils/analytics/replayAnalyticsEvents.tsx
+++ b/static/app/utils/analytics/replayAnalyticsEvents.tsx
@@ -60,6 +60,16 @@ export type ReplayEventParameters = {
   'replay.frame-after-background': {
     frame: string;
   };
+
+  'replay.hydration-error.issue-details-opened': {};
+  'replay.hydration-modal.opened': {
+    surface: string;
+  };
+  'replay.hydration-modal.slider-interaction': {};
+  'replay.hydration-modal.tab-change': {
+    tabKey: string;
+  };
+
   // similar purpose as "replay.details-viewed", however we're capturing the navigation action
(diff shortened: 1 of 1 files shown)


static/app/utils/analytics/replayAnalyticsEvents.tsx
```diff
@@ -134,6 +144,10 @@ export const replayEventMap: Record<ReplayEventKey, string | null> = {
   'replay.details-tab-changed': 'Changed Replay Details Tab',
   'replay.details-time-spent': 'Time Spent Viewing Replay Details',
   'replay.frame-after-background': 'Replay Frame Following Background Frame',
+  'replay.hydration-error.issue-details-opened': 'Hydration Issue Details Opened',
+  'replay.hydration-modal.opened': 'Hydration Modal Opened',
```
aliu39, line 148: This is duplicated from the button prop at https://github.com/getsentry/sentry/blob/a997487d02475a32b0200842e84def606dde9fd8/static/app/components/replays/breadcrumbs/openReplayComparisonButton.tsx#L39. Not sure what overrides what here. Or if the button's eventName can be removed after adding this definition.

michellewzhang: i don't think you need to add this one to this file (`replay.hydration-modal.opened`), since it's already defined with the button analytics. and you mentioned it's already being sent to amplitude

michellewzhang (approved):

PR #75090, opened 2024-07-26
fix(issues): Don't treat %s as unique

mrduncan:
If an event already has a parameterized `logentry.message`, ensure that uniq_id doesn't try to treat the %s as unique.

Example event payload:
```
{
  "logentry": {
    "message": "Hard time limit (%ss) exceeded for %s[%s]",
    "formatted": "Hard time limit (123s) exceeded for sentry.tasks.example[01234567-890a-bcde-f012-34567890abcd]",
    "params": [123, "sentry.tasks.example", "01234567-890a-bcde-f012-34567890abcd"]
  }
}
```

Prior to this change, these would have been parameterized as `Hard time limit (%ss) exceeded for <uniq_id>` ([internal example](https://sentry.sentry.io/issues/5137501296/events/dbe48edc9fa7485dbb5a0c8e095b2717/?project=1)).

This is currently pretty hard to test end-to-end and is also only catching a subset of possible `%` replacements, will follow up with a future PR to improve that but for now this will address the most common case.

Reminder: this parameterization is still only enabled experimentally and is not widely released.


src/sentry/grouping/parameterization.py
```diff
@@ -240,11 +240,11 @@ def num_tokens_from_string(token_str: str) -> int:
...
         if len(token_str) < _UniqueId.TOKEN_LENGTH_MINIMUM:
-            return False
-        if (
-            token_str[0] == "<" and token_str[-1] == ">"
-        ):  # Don't replace already-parameterized tokens
-            return False
+            return False  # Don't replace short tokens
+        if token_str[0] == "<" and token_str[-1] == ">":
+            return False  # Don't replace already-parameterized tokens
+        if "%s" in token_str:
```
lobsterkatie, line 246: Is there a reason not to add `%d` in here?

lobsterkatie (approved):

src/sentry/grouping/parameterization.py
mrduncan: Mostly a combination of the difficulty in testing this and I want to limit the amount of additional work we have do do on each `token_str` until we have a better way to do better performance testing as well

PR #75094, opened 2024-07-26
ref(onboarding): fix initial project state

michellewzhang:
Fixes an onboarding sidebar bug where multiple projects or no projects being selected leads the wrong platform to be displayed in the sidebar.

Specifically on issues, if you have multiple projects selected and click into a specific event details, the global project selection will persist, and there's a chance the wrong platform will be displayed when you jump into the replay onboarding side. This PR sets the URL `project` param so that it's always the platform of the event being showed.

Before: sidebar displays instructions for bun when the platform is go, since bun is the first project in the global selection:

https://github.com/user-attachments/assets/0b98ff2f-04fc-4bc5-929d-fe4c69669a19

After: sidebar displays correct instructions for the platform even though there are many projects in the global selection:

https://github.com/user-attachments/assets/f4c9155c-1448-4626-b1bc-2dc8d5769705


static/app/components/onboarding/gettingStartedDoc/utils/useCurrentProjectState.tsx
```diff
@@ -21,12 +21,11 @@ function useCurrentProjectState({
...
   allPlatforms,
 }: Props) {
-  const params = useParams<{projectId?: string}>();
-  const projectSlug = params.projectId;
   const {projects, initiallyLoaded: projectsLoaded} = useProjects();
   const {selection, isReady} = useLegacyStore(PageFiltersStore);
   const [currentProject, setCurrentProject] = useState<Project | undefined>(undefined);
-
+  const {getParamValue: projectIds} = useUrlParams('project');
+  const projectId = projectIds()?.split('&')[0];
```
ryan953, line 28: with `.at()` typescript will tell you that it could be undefined

```suggestion
  const projectId = projectIds()?.split('&').at(0);
```

ryan953 (approved):

PR #75102, opened 2024-07-26
Minor cleanup / changes to delayed_processing

saponifi3d:
# Description

Some clean up moving code around for delayed_processing, this was mostly really minor perf improvements or code structure changes.

Accidentally lost some comments along the refactor way, please let me know if there are any comments that would be helpful to add.


src/sentry/rules/processing/delayed_processing.py
```diff
@@ -142,84 +149,183 @@ def get_condition_query_groups(
...
-                # the same UniqueConditionQuery, the condition JSON in
-                # DataAndGroups will be incorrect for one of those types.
-                # The JSON will either have or be missing a comparisonInterval
-                # which only applies to percent conditions, and have the incorrect
-                # comparisonType for one type. This is not a concern because
-                # when we instantiate the exact condition class with the JSON,
-                # the class ignores both fields when calling get_rate_bulk.
-
-                # Add to set of group_ids if there are already group_ids
-                # that apply to the unique condition query.
```
schew2381, line 160: Without this comment I think people could get tripped up on this in the future

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -142,84 +149,183 @@ def get_condition_query_groups(
...
+    condition_groups: dict[UniqueConditionQuery, DataAndGroups], project: Project
 ) -> dict[UniqueConditionQuery, dict[int, int]] | None:
-    condition_group_results: dict[UniqueConditionQuery, dict[int, int]] = {}
+    condition_group_results = {}
     current_time = datetime.now(tz=timezone.utc)
-    for unique_condition, (condition_data, group_ids) in condition_groups.items():
-        condition_cls = rules.get(unique_condition.cls_id)
+    project_id = project.id
 
+    for unique_condition, data_and_groups in condition_groups.items():
```
schew2381, line 291: subjective nit: I think unpacking `data_and_groups` here is more readable

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -244,7 +350,6 @@ def _passes_comparison(
         return False
 
     calculated_value = query_values[0]
-    # If there's a second query we must have a percent comparison condition.
```
schew2381, line 247: can this comment also be added back?

src/sentry/rules/processing/delayed_processing.py
saponifi3d: 👍 agree - my bad, the comments got destroyed by an editor plugin when i tried to do some automation to the file 😅 -- lemme know if you notice any other ones you think are important as well. (i tried to pull back as many as i could 😅)

src/sentry/rules/processing/delayed_processing.py
saponifi3d: 👍 yeah - a can do.

That said, i don't see a lot of value in this component as it's saying the same thing as the code. (query length is 2, than it's a percent_increase comparison). i wonder if there's a better comment about why we're making this decision or why this is being handled in this part of the code (vs elsewhere)

schew2381: Hmm that's true, we could replace the comment and code with just:
```Python
if condition_data.get("comparisonType") == ComparisonType.PERCENT:
```

saponifi3d: LOVE that. updating.

ceorourke (approved):
seems ok I think, I find it hard to review this type of PR where things get moved around _and_ changed though

schew2381 (approved):
lgtm

schew2381:
> ## Suspect Issues
> 
> * ‼️ **UnboundLocalError: cannot access local variable 'rules_to_fire' where it is not associated with a value** `sentry.rules.processing.delayed_processing` [View Issue](https://sentry.sentry.io/issues/5658441415/?referrer=github-pr-bot)
> 

Amazing feature

saponifi3d:
@schew2381 right!? there was also an issue alert created for it! anywho pushed a fix here: https://github.com/getsentry/sentry/pull/75280

PR #75104, opened 2024-07-26
ref(similarity): Catch all seer exceptions in backfill

jangjodi:
Catch all seer exceptions in retry in backfill
We do not have a definitive list of exceptions we want to catch, we want to retry them all


tests/sentry/tasks/test_backfill_seer_grouping_records.py
```diff
@@ -240,15 +240,15 @@ def test_lookup_group_data_stacktrace_bulk_exceptions(
...
-                mock_logger.exception.assert_called_with(
-                    "tasks.backfill_seer_grouping_records.bulk_event_lookup_exception",
-                    extra={
-                        "organization_id": self.project.organization.id,
-                        "project_id": self.project.id,
-                        "group_data": json.dumps(rows),
-                        "error": exception.message,
-                    },
-                )
+            mock_logger.exception.assert_called_with(
```
jangjodi, line 243: fix some assertions that weren't running properly, same for line 457

JoshFerge (approved):

PR #75105, opened 2024-07-26
nit(scm): consolidate api client class naming

cathteng:
Give all of the SCM API client classes the name `__ApiClient`.


GabeVillalobos (approved):
🎉

PR #75107, opened 2024-07-26
ref(slack-image): Add Issue Type to Function Mapping & Update Metrics

iamrajjoshi:
We had slightly skewed metrics b/c we were emitting the `chartcuterie.issue_alert.attempt` metric without checking if we even supported that issue type. We would then emit `chartcuterie.issue_alert.success` only if we had an image (which would be when we actually supported the issue type). Because of this, we had a large difference in attempts vs success.

In this pr, i added a mapping from issue type to the function used to generate the image, and an early exist block before collecting the attempt metric to make sure we can correlate attempts to successes.


cathteng (commented):
i think this looks good but will wait for @sentaur-athena to review. should we have tests so that we're sure the metrics are being emitted when/where we want them to be?

sentaur-athena (approved):

PR #75108, opened 2024-07-26
ref: fix override of get_or_create

asottile-sentry:
mypy 1.11 + checked model types pointed out this inconsistency

### DIFF
--- a/src/sentry/models/commitauthor.py
+++ b/src/sentry/models/commitauthor.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, ClassVar
+from collections.abc import Mapping
+from typing import TYPE_CHECKING, Any, ClassVar
 
 from django.db import models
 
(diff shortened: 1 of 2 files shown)


src/sentry/models/commitauthor.py
```diff
@@ -13,13 +14,14 @@
...
+        self, defaults: Mapping[str, Any] | None = None, **kwargs: Any
+    ) -> tuple[CommitAuthor, bool]:
         # Force email address to lowercase because many providers do this. Note though that this isn't technically
         # to spec; only the domain part of the email address is actually case-insensitive.
         # See: https://stackoverflow.com/questions/9807909/are-email-addresses-case-sensitive
-        return super().get_or_create(
-            organization_id=organization_id, email=email.lower(), defaults=defaults, **kwargs
-        )
+        email = kwargs.pop("email").lower()
+        return super().get_or_create(defaults=defaults, email=email, **kwargs)
```
wedamija, line 24: Should we also check for `organization_id` here?

asottile-sentry: it's non-null so the underlying call will do that for us

joshuarli (approved):

PR #75110, opened 2024-07-26
ref: fix num_comments typing in migrations again

asottile-sentry:


src/sentry/migrations/0739_backfill_group_info_to_group_attributes.py
```diff
@@ -73,7 +73,7 @@ def _bulk_retrieve_group_values(group_ids: list[int], Group: type[Group]) -> lis
                 status=group_values["status"],
                 substatus=group_values["substatus"],
                 first_seen=group_values["first_seen"],
-                num_comments=group_values["num_comments"],
+                num_comments=group_values["num_comments"] or 0,
```
asottile-sentry, line 76: the model requires `int` but the schema here is nullable

wedamija (approved):

IanWoodard (approved):

PR #75111, opened 2024-07-26
ref: fix erroneous `project=` passed to `ProjectTemplate`

asottile-sentry:
once models are checked mypy notices that the `project=` argument goes nowhere here.  I'm surprised this works at runtime tbh


wedamija (approved):
I was a bit confused about this, it's because there's an fk from Project -> ProjectTemplate, and I guess that causes this to slip by validation

IanWoodard (approved):

PR #75114, opened 2024-07-26
feat(draggableTabs): Add query count and tab menu button 

### LINKED ISSUE #73224: Custom Views: Add tab menu and query count to dnd tabs
Once dnd tabs have been created, we will need to add the tab menu button and query count badge to each tab. This menu will be used to contain delete, duplicate, rename, and save functionality. This button will need to also need to display a small purple dot on the top right of the menu button to be a visual cue for when the current tab has been modified. ([Internal Mockups](https://www.figma.com/design/MM42MEY0OiZRVzQlI8XHej/Q1-Issue-Stream?node-id=1025-40250&t=1ZGFIH1VYoEC9Had-4))

This ticket can be marked as completed once the tab menu button, the menu itself, and the button's modified state has been completed. As of now, the available menu options will be: 
- `Save Changes` (if tab is in modified state)
- `Discard Changes` (if tab is in modified state)
- `Rename`
- `Duplicate`
- `Delete` 

Implementing the functionality for each of these options is not within the scope of this ticket. 

Direct Dependencies: 
- #73223

MichaelSun48:
closes #73224

This PR adds the query count and tab dropdown and button to the draggable tabs component. These changes are purely visual, no functionality for the dropdown menu options have been implemented. 

image

As an added bonus, I also have cleaned up the DraggableTabList and DraggableTab components so that they are a lot  more generic than the previous implementation of this feature

Known Issues:
- Dragging a tab to the left makes it disappear into the overflow menu for some reason
- Hovering over the elements in the dropdown menu causes the tab itself to increase in opacity as if its being hovered over. Clicking on an option in the dropdown menu also increases the opacity of the tab


static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -210,7 +212,7 @@ const TabDivider = styled('div')`
   width: 1px;
   border-radius: 6px;
   background-color: ${p => p.theme.gray200};
-  margin: 9px auto;
+  margin: 8px 4px;
```
MichaelSun48, line 215: This is applied to the tab divider element - it ensures that there's a little bit of gap between the shadow of a tab and the divider:
image

Without the margin, it looks like this:

image

static/app/views/issueList/draggableTabBar.tsx
```diff
@@ -25,7 +26,13 @@ export function DraggableTabBar(props: DraggableTabBarProps) {
...
       <DraggableTabList tabs={tabs} setTabs={setTabs} orientation="horizontal">
         {tabs.map(tab => (
-          <DraggableTabList.Item key={tab.key}>{tab.label}</DraggableTabList.Item>
+          <DraggableTabList.Item
+            key={tab.key}
+            count={tab.queryCount}
+            hasUnsavedChanges={tab.hasUnsavedChanges}
+          >
+            {tab.label}
+          </DraggableTabList.Item>
```
MichaelSun48, line 35: I really do not like how the TabList is consuming the tabs in two ways - the `tabs` prop and the `<DraggableTabList.Item>` and I'm currently thinking about the best way to resolve this

static/app/views/issueList/draggableTabBar.tsx
```diff
@@ -25,7 +26,13 @@ export function DraggableTabBar(props: DraggableTabBarProps) {
...
         ))}
       </DraggableTabList>
```
MichaelSun48, line 37: I really do not like how the TabList is consuming the tabs in two ways – the `tabs` prop and the `<DraggableTabList.Item>` – and I'm currently thinking about the best way to resolve this

MichaelSun48: nvm fixed this in the latest commit

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -15,24 +15,20 @@ import {OverflowMenu, useOverflowTabs} from 'sentry/components/tabs/tabList';
...
-  setTabs: (tabs: Tab[]) => void;
-  tabs: Tab[];
 }
 
 function BaseDraggableTabList({
   hideBorder = false,
   className,
   outerWrapStyles,
-  tabs,
-  setTabs,
```
MichaelSun48, line 35: These tabs are already being consumed as children elements, it's cleaner to remove them as props here and handle the reordering in the parent component.

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -132,6 +133,7 @@ function BaseDraggableTabList({
                 }
                 ref={element => (tabItemsRef.current[item.key] = element)}
               />
+
               {state.selectedKey !== item.key &&
                 state.collection.getKeyAfter(item.key) !== state.selectedKey && (
                   <TabDivider />
```
MichaelSun48, line 139: This part of the DraggableTabListComponent is the only part that is sort of  specific to this implementation (or more specifically, the "filled" tab variant).

static/app/components/draggableTabs/draggableTabList.tsx
malwilley: Agreed!

static/app/views/issueList/draggableTabBar.tsx
```diff
@@ -19,13 +32,46 @@ export interface DraggableTabBarProps {
 }
 
 export function DraggableTabBar(props: DraggableTabBarProps) {
-  const [tabs, setTabs] = useState<Tab[]>([...props.tabs]);
+  const [tabs, setTabs] = useState<Tab[]>(props.tabs);
+  const [selectedTabKey, setSelectedTabKey] = useState<Key>(props.tabs[0].key);
+
+  const onReorder: (newOrder: Node<DraggableTabListItemProps>[]) => void = newOrder => {
+    let newTabOrder: Tab[] = [];
```
malwilley, line 39: Nit: this might be a bit simpler?

```tsx
return newOrder.map(node => tab.key === node.key ? tab : null).filter(defined);
```

static/app/views/issueList/draggableTabMenuButton.tsx
```diff
@@ -0,0 +1,164 @@
...
+        offset={[-10, 5]}
+      />
+    </TriggerIconWrap>
+  );
+}
+
+function IconCompactEllipsis() {
+  return (
+    <SvgIcon>
+      <circle cx="8" cy="8" r="1.11" />
```
malwilley, line 131: Is this different from the `IconEllipsis` we already have? If this is necessary, we probably want to either add a prop to the existing icon or add this to the icons folder

static/app/views/issueList/draggableTabBar.tsx
```diff
@@ -1,15 +1,28 @@
...
 import {TabPanels, Tabs} from 'sentry/components/tabs';
+import {space} from 'sentry/styles/space';
+import {DraggableTabMenuButton} from 'sentry/views/issueList/draggableTabMenuButton';
 
 export interface Tab {
   content: React.ReactNode;
   key: Key;
   label: string;
+  hasUnsavedChanges?: boolean;
+  onDelete?: (key: MenuItemProps['key']) => void;
```
malwilley, line 21: Since these action callbacks accept a `key` parameter, they don't need to live on the `Tab` object, right? It will probably be simpler and less error-prone to  pass these in `DraggableTabProps` instead.

static/app/views/issueList/draggableTabMenuButton.tsx
MichaelSun48: Yup - the existing IconEllipsis had too much spacing between the dots, which made it look a little cramped inside the dropdown button. This one just squeezes them a little closer. Agreed that this should just be controlled with a prop rather than a whole new component. 

`<IconEllipsis/>` (existing component):
image

`<IconCompactEllipsis/>` (new component): 
image

static/app/views/issueList/draggableTabBar.tsx
MichaelSun48: You're completely right, good catch!

malwilley (approved):

PR #75117, opened 2024-07-27
feat(discover): Query specific datasets in events

narsaynorath:
Returns the changes to support querying specific datasets in the events endpoint.


shruthilayaj (approved):
errors, and transactions datasets both now have ton n support and a lot of test coverage, so this change should be good to go

PR #75119, opened 2024-07-29
fix(metrics) visualize filed max length fix

vgrozdanic:
Fixes text (metric name) shown in the visualize field depending on the width of the screen

### DIFF
--- a/static/app/components/metrics/mriSelect/index.tsx
+++ b/static/app/components/metrics/mriSelect/index.tsx
@@ -308,7 +308,6 @@ export const MRISelect = memo(function MRISelect({
         size="md"
         sizeLimit={100}
         value={value}
-        css={comboBoxCss}
       />
     );
   }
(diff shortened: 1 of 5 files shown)


static/app/components/metrics/mriSelect/index.tsx
```diff
@@ -214,6 +216,14 @@ export const MRISelect = memo(function MRISelect({
     [onChange]
   );
 
+  const maxLength = useMemo(() => {
```
priscilawebdev, line 219: https://github.com/user-attachments/assets/cef3f4de-ba9c-425e-b66b-c5dae722ab1e

I am afraid we still have to tweak this function a little more

priscilawebdev: This is hard to achieve with a single number because small devices have different viewport sizes. Ideally, CSS should handle this. However, since we treat the string differently based on its format, we need a different approach 🤔

static/app/components/metrics/mriSelect/index.tsx
```diff
@@ -214,6 +216,14 @@ export const MRISelect = memo(function MRISelect({
...
+    if (breakpoints.small) {
+      // at least small size screen, no problem with fitting 100 characters
+      return 100;
+    }
+    return 35; // xs size screen, width less than theme.breakpoints.small
```
priscilawebdev, line 224:
```suggestion
    return 22; // xs size screen, width less than theme.breakpoints.small
```

I believe that 22 would be better here

static/app/components/metrics/mriSelect/index.tsx
vgrozdanic: Added new breakpoint to deal with this use case

vgrozdanic:
Closing this, I will open a new PR form a new branch for this without previous changes...

PR #75141, opened 2024-07-29
feat(alerts): add idiomatic time duration formatting for metric alerts

### LINKED ISSUE #75048: Better time formatting in alerts
For daily metric alerts, the formatting is off, displaying 1440 minutes instead of 1 day.

Image

Code that formats this text can be found here:
https://github.com/getsentry/sentry/blob/5e05ce5d1e44aa3d1f788e6db9b1b09736055de0/src/sentry/integrations/metric_alerts.py#L61-L103

shellmayr:
Fixes https://github.com/getsentry/sentry/issues/75048

Adds a new duration formatter function that is slightly smarter to format things like:
- "in the last hour" instead of "in the last 60 minutes" (metric alerts) or "in the last 1 hour" (general alerts)
- "in the last day" instead of "in the last 1440 minutes" (metric alerts) or "in the last 24 hours" (general alerts)
etc.

### DIFF
--- a/src/sentry/incidents/utils/format_duration.py
+++ b/src/sentry/incidents/utils/format_duration.py
@@ -29,3 +29,4 @@ def format_duration_idiomatic(minutes: int) -> str:
             literal = f"{value:d}" if value != 1 else ""
             unit_str = f" {unit}{pluralize(value)}"
             return f"{literal}{unit_str}".strip()
+    return f"{minutes} minutes"
--- a/src/sentry/incidents/utils/format_duration.py
+++ b/src/sentry/incidents/utils/format_duration.py
@@ -1,0 +1,31 @@
+from django.template.defaultfilters import pluralize
+
+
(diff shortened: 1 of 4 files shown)


obostjancic (approved):

src/sentry/incidents/utils/format_duration.py
```diff
@@ -0,0 +1,31 @@
...
+    "in the past month"
+    """
+
+    literal = ""
+    unit = ""
+    limits = {"month": 30 * 24 * 60, "week": 7 * 24 * 60, "day": 24 * 60, "hour": 60, "minute": 1}
+
+    for unit, limit in limits.items():
+        if minutes >= limit:
+            value = int(minutes // limit)
```
vgrozdanic, line 27: very small nit: both `minutes` and `limit` here are ints, so the result of `//` division will always be int

vgrozdanic (approved):

nhsiehgit (approved):

PR #75143, opened 2024-07-29
fix(metrics): Check for feature flag when uppercasing variable

priscilawebdev:
We should check if the feature flag `metrics-new-input` is `true`. If it is, we should convert the variables to uppercase to ensure the formula works.


obostjancic (approved):

shellmayr (approved):
LGTM

PR #75144, opened 2024-07-29
feat(metrics) Delete connected widgets and alerts to an extraction rule

vgrozdanic:
Backend part of https://github.com/getsentry/sentry/issues/74255 and https://github.com/getsentry/sentry/issues/74254

### DIFF
--- a/src/sentry/api/endpoints/project_metrics_extraction_rules.py
+++ b/src/sentry/api/endpoints/project_metrics_extraction_rules.py
@@ -17,11 +17,12 @@ from sentry.api.serializers import serialize
 from sentry.api.serializers.models.metrics_extraction_rules import (
     SpanAttributeExtractionRuleConfigSerializer,
 )
+from sentry.incidents.models.alert_rule import AlertRule
+from sentry.models.dashboard_widget import DashboardWidget
 from sentry.models.project import Project
 from sentry.sentry_metrics.models import SpanAttributeExtractionRuleConfig
 from sentry.sentry_metrics.span_attribute_extraction_rules import (
     create_extraction_rule_config,
-    delete_extraction_rule_config,
     update_extraction_rule_config,
 )
 from sentry.tasks.relay import schedule_invalidate_project_config
@@ -85,11 +86,29 @@ class ProjectMetricsExtractionRulesEndpoint(ProjectEndpoint):
 
         with handle_exceptions():
             with transaction.atomic(router.db_for_write(SpanAttributeExtractionRuleConfig)):
-                delete_extraction_rule_config(project, config_update)
-                schedule_invalidate_project_config(
-                    project_id=project.id, trigger="span_attribute_extraction_configs"
-                )
                 for config in config_update:
+                    try:
+                        span_attribute = SpanAttributeExtractionRuleConfig.objects.get(
+                            project=project, span_attribute=config["spanAttribute"]
+                        )
+                    except SpanAttributeExtractionRuleConfig.DoesNotExist:
+                        # ignore non existing extraction rules
+                        continue
+
+                    config_mris = []
+                    for condition in span_attribute.conditions.all():
+                        config_mris.extend(condition.generate_mris())
+
+                    span_attribute.delete()
+                    # delete connected alert rules and dashboards widgets to an extraction rule
+                    AlertRule.objects.get_for_metrics(project.organization, config_mris).delete()
+                    DashboardWidget.objects.get_for_metrics(
+                        project.organization, config_mris
+                    ).delete()
+
+                    schedule_invalidate_project_config(
+                        project_id=project.id, trigger="span_attribute_extraction_configs"
+                    )
                     self._create_audit_entry(
                         "SPAN_BASED_METRIC_DELETE", project, config["spanAttribute"]
                     )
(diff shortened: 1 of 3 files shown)


src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -85,11 +86,29 @@ def delete(self, request: Request, project: Project) -> Response:
...
-                    project_id=project.id, trigger="span_attribute_extraction_configs"
-                )
                 for config in config_update:
+                    try:
+                        span_attribute = SpanAttributeExtractionRuleConfig.objects.get(
+                            project=project, span_attribute=config["spanAttribute"]
+                        )
+                    except SpanAttributeExtractionRuleConfig.DoesNotExist:
+                        # ignore non existing extraction rules
+                        continue
```
obostjancic, line 96: Does this actually happen or are we trying to please mypy here 😅  If it does happen this request should fail with either 400 or 404.

obostjancic (approved):

src/sentry/api/endpoints/project_metrics_extraction_rules.py
obostjancic: Or at least, lets do a `logger.error` so that we get pinged if this happens

vgrozdanic: I think we are trying to be idempotent here, and allow "deletion" as many times as it is requested.
Simon has implemented this, and there is a test that checks this, so I assumed that there is a reason behind it.

@shellmayr should we handle these differently or leave it as it is?

shellmayr: For posterity: we talked about this and the right way to go is to make this idempotent, return 204 in line with the HTTP RFC, but log any instance of this in order to know when e.g. some front-end change starts sending too many requests with the same payload.

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -85,11 +86,33 @@ def delete(self, request: Request, project: Project) -> Response:
...
+                        config_mris.extend(condition.generate_mris())
+
+                    span_attribute.delete()
+                    # delete connected alert rules and dashboards widgets to an extraction rule
+                    AlertRule.objects.get_for_metrics(project.organization, config_mris).delete()
+                    DashboardWidget.objects.get_for_metrics(
+                        project.organization, config_mris
+                    ).delete()
+
+                    schedule_invalidate_project_config(
```
shellmayr, line 113: We should definitely add audit entries for deletion of the AlertRule and DashboardWidget here.

src/sentry/api/endpoints/project_metrics_extraction_rules.py
```diff
@@ -85,11 +86,33 @@ def delete(self, request: Request, project: Project) -> Response:
...
+                        )
+                        continue
+
+                    config_mris = []
+                    for condition in span_attribute.conditions.all():
+                        config_mris.extend(condition.generate_mris())
+
+                    span_attribute.delete()
+                    # delete connected alert rules and dashboards widgets to an extraction rule
+                    AlertRule.objects.get_for_metrics(project.organization, config_mris).delete()
```
shellmayr, line 108: When an alert rule is deleted, there's multiple dependent actions that need to be taken - see https://github.com/getsentry/sentry/blob/21aefaf841727b1f6929b6627e4ad1fccc46b82f/src/sentry/incidents/logic.py#L1005 - is there a reason we're not using this function here?

tests/sentry/api/endpoints/test_project_metrics_extraction_rules.py
```diff
@@ -4,7 +4,10 @@
```
shellmayr, line 4: Did you do some due-diligence concerning the permission levels of 
1. deletion of extraction rules
2. delection of alert rules
3. deletion of dashboard widgets

Do they all have the same permissions, or could some privilege escalation happen here somehow? (e.g. user being allowed to delete the extraction rule, but not alertrules)

tests/sentry/api/endpoints/test_project_metrics_extraction_rules.py
```diff
@@ -772,3 +775,137 @@ def test_query_filter_rules(self):
...
+        ).exists()
+
+        assert not AlertRule.objects.filter(id=alert_rule1.id).exists()
+        assert not AlertRule.objects.filter(id=alert_rule2.id).exists()
+        assert not DashboardWidget.objects.filter(id=dashboard_widget1.id).exists()
+        assert not DashboardWidget.objects.filter(id=dashboard_widget2.id).exists()
+
+    @django_db_all
+    @with_feature("organizations:custom-metrics-extraction-rule")
+    def test_alerts_and_widgets_connected_to_other_extraction_rules_are_not_affected(self):
```
shellmayr, line 838: One of these tests verifies correct deletion and the other verifies unaffected other extraction rules. Could you add another test here that verifies that when changing an extraction rule, only those rules are deleted that the specific combination of spanAttribute, aggregate and condition is targeting?

shellmayr (changes requested):

vgrozdanic:
Closing this, we will handle deletion in the other way. Current decision is not to delete connected alerts and widgets, but just to show warning that those widgets and alerts will be affected

vgrozdanic:
Closing this, current decision is not to delete connected widgets and alerts, but just to show the warning during a deletion

PR #75145, opened 2024-07-29
fix(profiling): align everything to txn space

JonasBa:


static/app/components/profiling/flamegraph/flamegraph.tsx
```diff
@@ -90,15 +90,7 @@ function getMaxConfigSpace(
...
-    // On most platforms, profile duration < transaction duration, however
-    // there is one beloved platform where that is not true; android.
-    // Hence, we should take the max of the two to ensure both the transaction
-    // and profile are fully visible to the user.
-    const duration = Math.max(
-      formatTo(transactionDuration, 'seconds', unit),
-      maxProfileDuration
-    );
-    return new Rect(0, 0, duration, 0);
+    return new Rect(0, 0, formatTo(transactionDuration, 'seconds', unit), 0);
```
Zylphrex, line 93: Small nit: the `maxProfileDuration` is now unnecessary if `transaction` is defined so we can move that to after the `if`.

Zylphrex (approved):

static/app/components/profiling/flamegraph/flamegraph.tsx
JonasBa: Good eye.

viglia (approved):

PR #75147, opened 2024-07-29
fix: deprecate old landing page and create a new wrapper around it

JonasBa:
Deprecate the old so we can start picking and replacing components on the landing page


Zylphrex (approved):
🚢

PR #75149, opened 2024-07-29
ref: make create or update typesafe

asottile-sentry:

### DIFF
--- a/src/sentry/models/release.py
+++ b/src/sentry/models/release.py
@@ -5,7 +5,7 @@ import logging
 import re
 from collections.abc import Mapping, Sequence
 from time import time
-from typing import ClassVar
+from typing import ClassVar, TypedDict
 
 import orjson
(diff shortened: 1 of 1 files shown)


src/sentry/models/release.py
```diff
@@ -768,14 +774,8 @@ def set_commits(self, commit_list):
...
                         defaults=commit_data,
                     )
-                    if not created:
-                        commit_data = {
-                            key: value
-                            for key, value in commit_data.items()
-                            if getattr(commit, key) != value
-                        }
-                        if commit_data:
-                            commit.update(**commit_data)
```
asottile-sentry, line 778: this is ever-so-slightly slower in that it sometimes updates more things -- but greatly simplifies the types here (`commit_data` being reassigned to `dict[str, object]` here)

joshuarli (approved):

PR #75150, opened 2024-07-29
ref: fix QuerySet soundness of event_frequency

asottile-sentry:

### DIFF
--- a/src/sentry/rules/conditions/event_frequency.py
+++ b/src/sentry/rules/conditions/event_frequency.py
@@ -6,7 +6,7 @@ import logging
 from collections import defaultdict
 from collections.abc import Callable, Mapping
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


src/sentry/rules/conditions/event_frequency.py
```diff
@@ -426,7 +438,7 @@ def batch_query_hook(
     ) -> dict[int, int]:
         batch_sums: dict[int, int] = defaultdict(int)
         groups = Group.objects.filter(id__in=group_ids).values(
-            "id", "type", "project__organization_id"
+            "id", "type", "project_id", "project__organization_id"
```
asottile-sentry, line 441: this (and the one below) adds an extra column to the select query -- necessary to make the calls below safe though (matching TypedDict)

joshuarli (approved):

PR #75151, opened 2024-07-29
ref: allow SnubaParams.teams to take a QuerySet

asottile-sentry:


wmak (approved):
Thanks for doing this!

🤔 Shouldn't this have been failing mypy before?

asottile-sentry:
> Thanks for doing this!
> 
> 🤔 Shouldn't this have been failing mypy before?

I am currently working through 700+ errors in django-stubs to get our models checked -- due to the way we make our `BaseQuerySet` ~most model methods are currently unchecked

down to the last ~15 or so :)

armenzg (approved):

PR #75155, opened 2024-07-29
fix(metrics): min/max queries without data

### LINKED ISSUE #75129: Min agg bugs
- min of span.duration looks janky
- min of empty breaks, max does not

obostjancic:
Handles min/max aggregations when querying metrics without data
- closes #75129 

Before:
image
After:
image


priscilawebdev (approved):
I tested it locally and can confirm that the changes in this PR resolve the issue

PR #75159, opened 2024-07-29
ref(metrics): Make span.duration built in default

### LINKED ISSUE #75132: span.duration should be selected by default
Select built-in one.

priscilawebdev:
### Goal

The 'visualization' field on the metrics page should default to the built-in 'span.duration' metric if it is available. However, if the URL already specifies a different metric, the field should select the metric from the URL.

**Preview**

https://github.com/user-attachments/assets/869521ac-9322-49af-892d-b2993d070045

https://github.com/user-attachments/assets/398442ee-8368-482b-be30-0cfaffe88ab3

closes https://github.com/getsentry/sentry/issues/75132


shellmayr (approved):
From a usage perspective: looks good to me

matejminar (approved):
Looks good to me 🚀

PR #75160, opened 2024-07-29
ref(metric-extraction) Timeout split from on-demand metric extraction config

### LINKED ISSUE #74620: Split the timeout from on-demand metrics extraction
(vgrozdanic 19.7.2024.) We will wait for a couple more days to see if there is any progress with optimization of the project config generation, if not, we will split the timeouts
Thread to follow updates: https://sentry.slack.com/archives/C05DDR7KLPP/p1720728460738409

vgrozdanic:
Splits timeout for on-demand metric extraction config builder from other (span-based extraction rules) metric extraction config creations by adding a new function `run_experimental_config_builder`, which runs function passed as an argument with a timeout, and catches possible `TimeoutException` from the function call.

By splitting the timeout, even if the on-demand metric extraction config generation fails, other metric extraction configs will still be generated, unlike before.

Closes https://github.com/getsentry/sentry/issues/74620

### DIFF
--- a/src/sentry/relay/config/experimental.py
+++ b/src/sentry/relay/config/experimental.py
@@ -79,7 +79,7 @@ def build_safe_config(
     """
     timeout = TimeChecker(_FEATURE_BUILD_TIMEOUT)
 
-    with sentry_sdk.start_span(op=f"project_config.time_constrained_config_builder.{key}"):
+    with sentry_sdk.start_span(op=f"project_config.build_safe_config.{key}"):
         try:
             return function(timeout, *args, **kwargs)
         except TimeoutException as e:
--- a/src/sentry/relay/config/experimental.py
+++ b/src/sentry/relay/config/experimental.py
@@ -79,18 +79,22 @@ def add_experimental_config(
 
 
 R = TypeVar("R")
+R_default = TypeVar("R_default")
 
 
-def run_experimental_config_builder(
-    function: Callable[..., R], *args: Any, **kwargs: Any
-) -> R | None:
+def run_time_constrained_config_builder(
+    function: Callable[..., R], *args: Any, default_return: R_default = None, **kwargs: Any
+) -> R | R_default:
     """
-    Runs an experimental config builder function with a timeout.
-    If the function call raises an exception, we log it to sentry and return None.
+    Runs a config builder function with a timeout.
+    If the function call raises an exception, we log it to sentry and return value passed as
+    `default_return` parameter (by default this is `None`).
     """
     timeout = TimeChecker(_FEATURE_BUILD_TIMEOUT)
 
-    with sentry_sdk.start_span(op="project_config.experimental_config_builder"):
+    with sentry_sdk.start_span(
+        op=f"project_config.time_constrained_config_builder.{function.__name__}"
+    ):
         try:
             return function(timeout, *args, **kwargs)
         except TimeoutException as e:
@@ -101,4 +105,4 @@ def run_experimental_config_builder(
         except Exception:
             logger.exception("Exception while building Relay project config field")
 
-    return None
+    return default_return
--- a/src/sentry/relay/config/experimental.py
+++ b/src/sentry/relay/config/experimental.py
@@ -1,7 +1,7 @@
 import logging
-from collections.abc import MutableMapping
+from collections.abc import Callable, MutableMapping
 from datetime import datetime, timedelta, timezone
-from typing import Any, Protocol
+from typing import Any, Protocol, TypeVar
 
 import sentry_sdk
 
@@ -76,3 +76,27 @@ def add_experimental_config(
         else:
             if subconfig is not None:
                 config[key] = subconfig
+
+
+R = TypeVar("R")
+
+
+def run_experimental_config_builder(
+    function: Callable[..., R], *args: Any, **kwargs: Any
+) -> R | None:
+    """
+    Runs an experimental config builder function with a timeout.
+    If the function call raises an exception, we log it to sentry and return None.
+    """
+    timeout = TimeChecker(_FEATURE_BUILD_TIMEOUT)
+
+    with sentry_sdk.start_span(op="project_config.experimental_config_builder"):
+        try:
+            return function(timeout, *args, **kwargs)
+        except TimeoutException as e:
+            logger.exception(
+                "Project config feature build timed out",
+                extra={"hard_timeout": e._timeout, "elapsed": e._elapsed},
+            )
+        except Exception:
+            logger.exception("Exception while building Relay project config field")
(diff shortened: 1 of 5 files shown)


tests/sentry/relay/config/test_metric_extraction.py
vgrozdanic, line 1: Due to the refactor of the `get_metric_extraction_config`, `TimeChecker` is removed as an argument of the function. In this tests, timeout check was never enforced anyways, so eventhough it is removed from here, functionality that is being tested hasn't changed.

At the end I have added the test which checks what happens when the timeout exception is raised, and that that is properly handled.

src/sentry/relay/config/metric_extraction.py
```diff
@@ -113,14 +111,19 @@ def get_metric_extraction_config(
     sentry_sdk.set_tag("organization_id", project.organization_id)
 
     with sentry_sdk.start_span(op="get_on_demand_metric_specs"):
-        alert_specs, widget_specs = get_on_demand_metric_specs(timeout, project)
+        on_demand_specs = run_experimental_config_builder(get_on_demand_metric_specs, project)
```
jan-auer, line 114: Optional suggestion for ergonomics: You could add an optional argument for a default that is used instead of `None` when the experimental builder fails. Then, pass `([], [])` or `(None, None)` as default. This would make the signature larger, however.

src/sentry/relay/config/metric_extraction.py
```diff
@@ -113,14 +111,19 @@ def get_metric_extraction_config(
...
+        if on_demand_specs is not None:
+            # unpacking the tuple returned from get_on_demand_metric_specs
+            alert_specs, widget_specs = on_demand_specs
+        else:
+            # exception occurred during the get_on_demand_metric_specs run
+            alert_specs, widget_specs = [], []
     with sentry_sdk.start_span(op="generate_span_attribute_specs"):
         span_attr_specs = _generate_span_attribute_specs(project)
```
jan-auer, line 122: At least `_generate_span_attribute_specs` should also be wrapped by our guarding function, as it accesses the database and might lead to similar slowdowns as the on-demand function.

src/sentry/relay/config/experimental.py
```diff
@@ -76,3 +76,29 @@ def add_experimental_config(
         else:
             if subconfig is not None:
                 config[key] = subconfig
+
+
+R = TypeVar("R")
+
+
+def run_experimental_config_builder(
```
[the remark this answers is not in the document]
vgrozdanic: I have chosen not to touch `add_experimental_config` because it would require far larger refactoring, and potentially break a lot of things, that is why new function was created.

I wasn't aware of the reasons behind the name, I will rename my function 😅

src/sentry/relay/config/experimental.py
```diff
@@ -60,11 +60,28 @@ def add_experimental_config(
...
+    if subconfig := build_safe_config(key, function, *args, **kwargs):
+        config[key] = subconfig
+
+
+R = TypeVar("R")
+R_default = TypeVar("R_default")
+
+
+def build_safe_config(
+    key, function: Callable[..., R], *args: Any, default_return: R_default = None, **kwargs: Any
```
jan-auer, line 73:
```suggestion
    key: str, function: Callable[..., R], *args: Any, default_return: R_default = None, **kwargs: Any
```

jan-auer (approved):

src/sentry/relay/config/metric_extraction.py
```diff
@@ -113,14 +111,17 @@ def get_metric_extraction_config(
...
-        alert_specs, widget_specs = get_on_demand_metric_specs(timeout, project)
+        alert_specs, widget_specs = build_safe_config(
+            "on_demand_metric_specs", get_on_demand_metric_specs, project, default_return=([], [])
+        )  # type: ignore[misc]
     with sentry_sdk.start_span(op="generate_span_attribute_specs"):
-        span_attr_specs = _generate_span_attribute_specs(project)
+        span_attr_specs = build_safe_config(
+            "span_attribute_specs", _generate_span_attribute_specs, project, default_return=[]
+        )
     with sentry_sdk.start_span(op="merge_metric_specs"):
```
k-fish, line 121: Assuming the span time is similar, this actually increases the total time a task can take from `1 * Timeout` to `2 * Timeout`, and based on the bumped up timeout we may be exceeding the softtimeout for this task. 

I'd also recommend you change the soft timeout to match the outer bounds of all these checks, just to be safe, confirm with other Relay folks that the increased soft timeout won't cause issues. We'll be bringing these down, but  in the meantime the addition of span specs def needs us to bump up the total task time.

k-fish (commented):
You already fix the repeated code from when I last looked, so 👍, but we also need to check the task time here since in splitting the timeouts you're adding to the maximum possible time the sum of all the sub configs can take. We might not need the full subconfig timeout for each of these, so we should allow the time to be passed in per sub-config I think to keep the total time low.

Edit: https://github.com/getsentry/sentry/blob/cd04b3ba5a272c52ff9f1796baaa65feee19925d/src/sentry/tasks/relay.py#L20-L26

src/sentry/relay/config/metric_extraction.py
vgrozdanic: Great catch, I have checked with Ingest team, we don't expect generation of span based metrics to take long enough to make problems, so for now we will leave soft time limit as it is, and I will keep an eye on it if it needs to be increased, if some problems happen

src/sentry/relay/config/experimental.py
```diff
@@ -60,11 +60,32 @@ def add_experimental_config(
...
+R = TypeVar("R")
+R_default = TypeVar("R_default")
+
+
+def build_safe_config(
+    key: str,
+    function: Callable[..., R],
+    *args: Any,
+    default_return: R_default | None = None,
+    **kwargs: Any,
```
asottile-sentry, line 77: this should be `Callabe[P, R]` and `*args: P.args` and `**kwargs: P.kwargs` (`P` is the conventional name for `ParamSpec`) such that the signature is preserved in callers

vgrozdanic: I did this in a first place, but maybe I did something wrong since I had some problems with mypy because I am inserting extra parameter (`timeout`) when calling a function (line 88 in this PR):

This was an error:
```
src/sentry/relay/config/experimental.py:88: error: Argument 1 has incompatible type "TimeChecker"; expected "P.args"  [arg-type]
```

asottile-sentry: is this file in the ignore list?  I suspect you wrote `*args: *P.args` which mypy should be flagging separately as a mistake

PR #75163, opened 2024-07-29
Revert "fix(related_issues): Orgs with open membership should always show related issues (#72961)

wmak:
- This reverts commit 7a350510e73642cc34fa20f6d5a543116919550d.
- Commit 7a350510e is currently allowing anyone regardless of the global_views flag to view multiple projects as long as they have enabled the joinleave org setting
- @armenzg Does this break any critical flows in the UI?


wmak:
@armenzg is going to fix this in a different PR

PR #75166, opened 2024-07-29
chore(features) Remove FeatureHandlerStrategy.REMOTE

markstory:
We don't need two aliases for dynamic features and we can use more concrete terms now as flagpole is part of sentry.


cmanallen (approved):
Are we able to update the enum numbers to be sequential or are they stored somewhere?

PR #75167, opened 2024-07-29
feat(partner migration): Add new Fly sso class

brendanhsentry:
closes: https://getsentry.atlassian.net/browse/RV-1803

### DIFF
--- a/src/sentry/auth/partnership_configs.py
+++ b/src/sentry/auth/partnership_configs.py
@@ -3,7 +3,7 @@ from enum import Enum
 
 class ChannelName(Enum):
     FLY_IO = "fly"
-    FLY_DISABLEABLE = "fly-disableable"
+    FLY_NON_PARTNER = "fly-non-partner"
 
 
-SPONSOR_OAUTH_NAME = {ChannelName.FLY_IO: "Fly.io", ChannelName.FLY_DISABLEABLE: "Fly.disableable"}
+SPONSOR_OAUTH_NAME = {ChannelName.FLY_IO: "Fly.io", ChannelName.FLY_DISABLEABLE: "Fly.non-partner"}
--- a/src/sentry/auth/partnership_configs.py
+++ b/src/sentry/auth/partnership_configs.py
@@ -6,4 +6,4 @@ class ChannelName(Enum):
     FLY_NON_PARTNER = "fly-non-partner"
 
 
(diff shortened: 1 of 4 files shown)


src/sentry/auth/partnership_configs.py
```diff
@@ -3,6 +3,7 @@
 
 class ChannelName(Enum):
     FLY_IO = "fly"
+    FLY_DEACTIVATED = "fly-deactivated"
 
 
 SPONSOR_OAUTH_NAME = {ChannelName.FLY_IO: "Fly.io"}
```
isabellaenriquez, line 9: could we add something here for `FLY_DEACTIVATED` so that the second SSO class has a different name?

src/sentry/auth/partnership_configs.py
```diff
@@ -3,6 +3,7 @@
 
 class ChannelName(Enum):
     FLY_IO = "fly"
+    FLY_DEACTIVATED = "fly-deactivated"
```
isabellaenriquez, line 6: nit: maybe name this something like `fly-disableable` to make it clear what the difference between the two providers

src/sentry/auth/providers/fly/provider.py
```diff
@@ -82,3 +82,7 @@ def build_identity(self, state):
             "data": self.get_oauth_data(data),
             "email_verified": False,
         }
+
+
+class FlyOAuthPartnershipProvider(FlyOAuth2Provider):
```
isabellaenriquez, line 87: nit: `DisableableFlyOAuth2Provider` maybe or `ConfigurableFlyOAuth2Provider`? idk names are hard but ideally something that'd make it clear that this provider can be disabled

isabellaenriquez (commented):
just nits, lgtm overall

isabellaenriquez (approved):

src/sentry/auth/partnership_configs.py
```diff
@@ -3,6 +3,7 @@
 
 class ChannelName(Enum):
     FLY_IO = "fly"
+    FLY_DISABLEABLE = "fly-disableable"
```
dashed, line 6: Instead of "disableable", how about `FLY_NON_PARTNER` as a name?

dashed (approved):

PR #75169, opened 2024-07-29
feat(issue-details): First pass implementation of folding sections

leeandher:
Requires https://github.com/getsentry/sentry/pull/74224

This PR implements the foldable data section across _almost_ all the data sections on the default issue details page. There are a few that aren't addressed or have some weird quirks:

- View full trace section is has no header so rather than implement one here, I'll circle back with design, same goes for the suspect commit section
- Tags section doesn't have a line underneath and becomes confusing when a screenshot is attached to it. I'd like to be able to remove a bunch of wrapper components and make the screenshot accessible elsewhere some how, but if not, I will create a new DataSection called `Tags & Screenshot` to replace it in a future PR.
- Any sections like debug meta that have a `Show Details` button bug out with the folding, and will need to be individually removed since the component is agnostic to the type of buttons it renders as actions.

image

**todo**
- [ ] Double check with some different issue types that the default layout without the flag is not changing.


static/app/components/events/eventEntry.tsx
```diff
@@ -80,11 +84,16 @@ function EventEntryContent({
     case EntryType.EXPECTCT:
     case EntryType.EXPECTSTAPLE:
       const {data, type} = entry;
-      return <Generic type={type} data={data} />;
+      return <Generic sectionKey={FoldSectionKey.EXPECTCT} type={type} data={data} />;
```
roggenkemper, line 87: do you know if `EXPECTSTAPLE` is a case we need to cover here

static/app/components/events/breadcrumbs/breadcrumbsDataSection.tsx
```diff
@@ -143,12 +144,13 @@ export default function BreadcrumbsDataSection({
   const hasViewAll = summaryCrumbs.length !== enhancedCrumbs.length;
 
   return (
-    <EventDataSection
+    <InterimSection
       key="breadcrumbs"
```
roggenkemper, line 148: we seem to be pretty inconsistent about passing in key in as a prop for `EventDataSection`. might be useful to just standardize that while we are making this change?

static/app/components/events/breadcrumbs/breadcrumbsDataSection.tsx
```diff
@@ -143,12 +144,13 @@ export default function BreadcrumbsDataSection({
...
       type="breadcrmbs"
```
roggenkemper, line 149: not from your pr but `breadcrmbs`

static/app/components/events/breadcrumbs/breadcrumbsDataSection.tsx
roggenkemper: Building off this - since we are already passing in key?/type `InterimSection` i wonder if we need a full additional prop (FoldSectionKey) for `InterimSection`. is there a reason we can't just have one enum for the sections and expand the local storage key to add in the additional "issue-details" prefix? it just seems a bit repetitive now

roggenkemper:
Screenshot 2024-07-30 at 1 18 19 PM
some weird behavior in `issue-details-request` section since it adds this toggle only when it is expanded

static/app/components/events/breadcrumbs/breadcrumbsDataSection.tsx
leeandher: Makes a lot of sense, didn't notice the overlap. I'll use the `type` key and change the enum a bit. Good call!

static/app/components/events/eventEntry.tsx
leeandher: Missed it, good catch!

leeandher:
> some weird behavior in `issue-details-request` section since it adds this toggle only when it is expanded

The formatting on that header is especially bad, I'll fix it in a future PR, but hiding the controls of the section is expected when the header is closed. The same happens for highlights, tags and stack trace, for example.

roggenkemper (approved):
some typescript failures because of the section key change but the rest of it looks good!

PR #75171, opened 2024-07-29
ref(control_silo): Consolidate integrations web resources

Christinarlong:
Moves all integrations web resources and related tests to sentry/integrations/web and tests/sentry/integrations/web

ref(#73859)


markstory (approved):

src/sentry/integrations/web/integration_extension_configuration.py
```diff
@@ -81,7 +93,9 @@ def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponseBase:
                 )
                 if org_member and "org:integrations" in org_member.scopes:
                     try:
-                        pipeline = self.init_pipeline(request, organization, request.GET.dict())
+                        pipeline = self.init_pipeline(
+                            Request(request=request), organization, request.GET.dict()
```
markstory, line 97: I'm not sure this is going to be safe. Perhaps we an expand the types on the relevant `Pipeline` logic instead of attempting to convert the request types.

markstory (approved):
Looks good.

src/sentry/integrations/web/integration_extension_configuration.py
asottile-sentry: fwiw it's very likely the type for `init_pipeline` is already wrong -- a good percentage of `request: Request` annotations were done sweepingly before mypy was set up and they're almost all wrong (when most of the time they should be `request: HttpRequest`)

PR #75174, opened 2024-07-29
fix(trace_related): Use global_views whitelist for events endpoint

armenzg:
In #72961 I added a change to a base class which I should have not.

This PR reverts my original change and _only_ adjusts the events endpoint and requires a referrer whitelist.

### DIFF
--- a/src/sentry/api/bases/organization_events.py
+++ b/src/sentry/api/bases/organization_events.py
@@ -125,16 +125,8 @@ class OrganizationEventsEndpointBase(OrganizationEndpoint):
                     "organizations:global-views", organization, actor=request.user
                 )
                 fetching_replay_data = request.headers.get("X-Sentry-Replay-Request") == "1"
-                if not any(
-                    [
-                        has_global_views,
-                        len(params.projects) <= 1,
-                        fetching_replay_data,
-                        # If a developer can view issues of a project they do not belong to
-                        # via open membership, we will also allow the endpoint to return events for it
-                        organization.flags.allow_joinleave,
-                    ]
-                ):
+
+                if not has_global_views and len(params.projects) > 1 and not fetching_replay_data:
                     raise ParseError(detail="You cannot view events from multiple projects.")
 
             # Return both for now
--- a/src/sentry/api/bases/organization_events.py
+++ b/src/sentry/api/bases/organization_events.py
@@ -126,7 +126,9 @@ class OrganizationEventsEndpointBase(OrganizationEndpoint):
                 )
                 fetching_replay_data = request.headers.get("X-Sentry-Replay-Request") == "1"
 
-                if not has_global_views and len(params.projects) > 1 and not fetching_replay_data:
+                if not any(
+                    [has_global_views and len(params.projects) > 1 and not fetching_replay_data]
+                ):
                     raise ParseError(detail="You cannot view events from multiple projects.")
 
             # Return both for now
--- a/src/sentry/api/bases/organization_events.py
+++ b/src/sentry/api/bases/organization_events.py
@@ -119,12 +119,12 @@ class OrganizationEventsEndpointBase(OrganizationEndpoint):
                 teams=self.get_teams(request, organization),
                 organization=organization,
             )
+
             if check_global_views:
                 has_global_views = features.has(
                     "organizations:global-views", organization, actor=request.user
                 )
                 fetching_replay_data = request.headers.get("X-Sentry-Replay-Request") == "1"
-
                 if not any(
                     [
                         has_global_views,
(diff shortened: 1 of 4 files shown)


src/sentry/api/bases/organization_events.py
```diff
@@ -125,16 +125,8 @@ def get_snuba_dataclass(
...
-                        has_global_views,
-                        len(params.projects) <= 1,
-                        fetching_replay_data,
-                        # If a developer can view issues of a project they do not belong to
-                        # via open membership, we will also allow the endpoint to return events for it
-                        organization.flags.allow_joinleave,
-                    ]
-                ):
+
+                if not has_global_views and len(params.projects) > 1 and not fetching_replay_data:
```
armenzg, line 129: Back to what it was.

src/sentry/api/endpoints/organization_events.py
```diff
@@ -316,7 +316,14 @@ def get(self, request: Request, organization) -> Response:
...
 
         try:
-            snuba_params, params = self.get_snuba_dataclass(request, organization)
+            # If the organization allows joining and leaving projects, it means that their
+            # developers can visit issues from any projects. In the case of trace related issues,
+            # we need the endpoint to let us see events from all projects when this flag is enabled.
+            snuba_params, params = self.get_snuba_dataclass(
+                request,
+                organization,
+                check_global_views=not bool(organization.flags.allow_joinleave),
```
armenzg, line 325: This skips this block:
https://github.com/getsentry/sentry/blob/a190038cadc6f4deced6369fb36cfffe71694892/src/sentry/api/bases/organization_events.py#L123-L130

tests/snuba/api/endpoints/test_organization_events.py
```diff
@@ -165,6 +165,17 @@ def test_performance_view_feature(self):
         assert response.status_code == 200
         assert len(response.data["data"]) == 1
 
+    def test_multi_project_feature_gate_rejection(self):
```
armenzg, line 168: This adds the original test back.

wmak (commented):
As discussed, since there's currently a DACI to change this behaviour lets not go with the original suggestion to add new endpoints yet.

Instead please add (with a TODO to remove and a comment explaining why its a bad pattern to use more) a check that the referrer is in an allowlist before allowing global-views

src/sentry/api/bases/organization_events.py
```diff
@@ -130,9 +130,6 @@ def get_snuba_dataclass(
                         has_global_views,
                         len(params.projects) <= 1,
                         fetching_replay_data,
-                        # If a developer can view issues of a project they do not belong to
-                        # via open membership, we will also allow the endpoint to return events for it
-                        organization.flags.allow_joinleave,
```
armenzg, line 135: This was added to my original PR.

This has been moved down below:
image

src/sentry/api/endpoints/organization_events.py
```diff
@@ -315,8 +318,20 @@ def get(self, request: Request, organization) -> Response:
...
-            snuba_params, params = self.get_snuba_dataclass(request, organization)
+            # If the organization allows joining and leaving projects, it means that their
+            # developers can visit issues from any projects. In the case of trace related issues,
+            # we need the endpoint to let us see events from all projects when this flag is enabled.
+            snuba_params, params = self.get_snuba_dataclass(
+                request,
+                organization,
+                check_global_views=(
+                    referrer in GLOBAL_VIEW_WHITELIST
+                    and not bool(organization.flags.allow_joinleave)
```
wmak, line 332:
```suggestion
                # This is only temporary until we come to a decision on global views
                # checking for referrer for an allowlist is a brittle check since referrer
                # can easily be set by the caller
                check_global_views=(
                    referrer in GLOBAL_VIEW_WHITELIST
                    and not bool(organization.flags.allow_joinleave)
```

src/sentry/api/endpoints/organization_events.py
```diff
@@ -43,6 +43,9 @@
     DiscoverSavedQueryTypes.TRANSACTION_LIKE: get_dataset("transactions"),
     DiscoverSavedQueryTypes.ERROR_EVENTS: get_dataset("errors"),
 }
+# TODO: Adjust this once we make a decision in the DACI for global views restriction
+# Do not add more referres to this list as it is a temporary solution
+GLOBAL_VIEW_WHITELIST = ("api.issues.issue_events",)
```
wmak, line 48: nit: I believe the terminology we use for this type of thing should be `allowlist`
https://develop.sentry.dev/development/inclusive-language/#whitelistblacklist

wmak (commented):
Minor comments but this looks like its on the right track

wmak (approved):
Looks good

wmak:
let's make sure we revisit this within a few weeks since that's the current expectation for the DACI to have a decision

PR #75176, opened 2024-07-29
ref: remove indirection through stringly typed message_builder

asottile-sentry:
first step to making these typed properly


joshuarli (approved):

asottile-sentry:
ugh this causes a cross-repo import cycle, going to abandon this for now and do the minimal change to fix the typing without the strictly better steps before and come back to this later

PR #75188, opened 2024-07-29
feat(custom-views): Create FE hooks for group search view endpoint 

MichaelSun48:
This PR creates the `useApiQuery` and `useMutation` hooks to call the GET and PUT groupsearchview endpoints, respectively. The endpoints can be found [here](https://github.com/getsentry/sentry/blob/master/src/sentry/issues/endpoints/organization_group_search_views.py)

### DIFF
--- a/static/app/types/views.tsx
+++ b/static/app/types/views.tsx
@@ -1,6 +1,0 @@
-export type View = {
-  name: string;
-  query: string;
-  querySort: 'date' | 'new' | 'trend' | 'freq' | 'user' | 'inbox';
-  id?: string;
-};
--- a/static/app/types/views.tsx
+++ b/static/app/types/views.tsx
@@ -1,0 +1,6 @@
+export type View = {
+  name: string;
+  query: string;
+  querySort: 'date' | 'new' | 'trend' | 'freq' | 'user' | 'inbox';
+  id?: string;
+};
--- a/static/app/views/issueList/mutations/useUpdateGroupSearchViews.tsx
+++ b/static/app/views/issueList/mutations/useUpdateGroupSearchViews.tsx
@@ -1,6 +1,5 @@
 import {addErrorMessage} from 'sentry/actionCreators/indicator';
 import {t} from 'sentry/locale';
-import type {View} from 'sentry/types/views';
 import {
   setApiQueryData,
   useMutation,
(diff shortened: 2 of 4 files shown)


static/app/views/issueList/mutations/useUpdateGroupSearchViews.tsx
```diff
@@ -0,0 +1,59 @@
...
+    mutationFn: ({orgSlug, views: data}: UpdateGroupSearchViewsVariables) =>
+      api.requestPromise(`/organizations/${orgSlug}/group-search-views/`, {
+        method: 'PUT',
+        data,
+      }),
+    onSuccess: (views, parameters, context) => {
+      setApiQueryData<View[]>(
+        queryClient,
+        makeFetchGroupSearchViewsKey({orgSlug: parameters.orgSlug}),
+        () => views // Update the cache with the new views
```
MichaelSun48, line 50: Not sure if this updater function is correct - I'm assuming we just reset the cache with the new set of views we get from the backend?

static/app/types/views.tsx
```diff
@@ -0,0 +1,6 @@
+export type View = {
+  name: string;
+  query: string;
+  querySort: 'date' | 'new' | 'trend' | 'freq' | 'user' | 'inbox';
```
malwilley, line 4: Could use use the existing enum `IssueSortOptions`?

static/app/types/views.tsx
malwilley, line 1: Since the types for issue views are only relevant for the issues list, I'd add them in that folder instead of the global types folder.

static/app/views/issueList/mutations/useUpdateGroupSearchViews.tsx
malwilley, line 50: I believe you can just pass `views` here without the function

static/app/types/views.tsx
```diff
@@ -0,0 +1,6 @@
+export type View = {
```
malwilley, line 1: I'd make this name more similar to the backend model `GroupSearchView`, `View` is a bit too generic

static/app/types/views.tsx
MichaelSun48: yup! did not see that one my bad

malwilley (approved):

PR #75198, opened 2024-07-29
fix(api) Don't overflow memcache key length

markstory:
When a release involves many projects the cache key can overflow the 250 char limit that memcache has for keys. Using a hash will give us deterministic sizing

Fixes SENTRY-3CMQ


JoshFerge (approved):

cmanallen (approved):
Do we know what kind of volume org auth tokens get?  No issue with a large scale cache invalidation is there?

markstory:
> Do we know what kind of volume org auth tokens get? No issue with a large scale cache invalidation is there?

Without the cache the updates will be high volume until the cache rebuilds. However, with a 5min TTL we're mostly protecting against bursts of traffic when CI/deploys run which we'll still be able to handle reasonably well.

PR #75200, opened 2024-07-29
feat(profiling): Generate flamegraph from latest transaction or conti…

Zylphrex:
…nuous profiles

This adds a new flamegraph source for `profiles` where we always generate flamegraphs using the latest transaction or continuous profiles. This does not require that the user has tracing enabled, just continuous profiling.

### DIFF
--- a/src/sentry/profiles/flamegraph.py
+++ b/src/sentry/profiles/flamegraph.py
@@ -2,7 +2,7 @@ from collections import defaultdict
 from collections.abc import Mapping
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Any, TypedDict
+from typing import Any, Literal, NotRequired, TypedDict
 
 from snuba_sdk import (
     And,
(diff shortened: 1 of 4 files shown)


src/sentry/profiles/flamegraph.py
```diff
@@ -559,3 +561,96 @@ def _query_chunks_for_profilers(self, query: Query) -> Mapping[str, Any]:
...
+    def get_profile_candidates_from_profiles(self) -> ProfileCandidates:
+        if self.snuba_params.organization is None:
+            raise ValueError("`organization` is required and cannot be `None`")
+
+        max_profiles = options.get("profiling.flamegraph.profile-set.size")
+
+        referrer = Referrer.API_PROFILING_PROFILE_FLAMEGRAPH_PROFILE_CANDIDATES.value
+
+        transaction_profiles_builder = DiscoverQueryBuilder(
+            dataset=Dataset.Discover,
```
wmak, line 574: 🤔 do you get a bit more perf by doing `Dataset.Transactions` or are there profiles on Error events too?

wmak (approved):
Minor comment on dataset selection, lgtm otherwise

src/sentry/profiles/flamegraph.py
Zylphrex: No profiles on error events. Not right now anyways. So I guess I could switch it and it'd be the same 🤔 I'm going to merge this as is and follow up to see how that works.

wmak: 👍 I *think* its slightly faster, cause snuba doesn't need to make any dataset decision, and then the transactions table has less rows

PR #75201, opened 2024-07-29
feat(profiling) split flamegraph and transaction data to separate tabs

JonasBa:


static/app/views/profiling/content.tsx
```diff
@@ -277,11 +278,20 @@ function ProfilingContentLegacy({location}: ProfilingContentProps) {
   );
 }
 
+function validateTab(tab: unknown): tab is 'flamegraph' | 'transactions' {
+  return tab === 'flamegraph' || tab === 'transactions';
+}
+
+function decodeTab(tab: unknown): 'flamegraph' | 'transactions' {
+  return validateTab(tab) ? tab : 'transactions';
```
Zylphrex, line 286: A comment explaining why we're choosing transactions as the default would be helpful here

static/app/views/profiling/content.tsx
```diff
@@ -388,18 +419,22 @@ function ProfilingContent({location}: ProfilingContentProps) {
...
                 </PageFilterBar>
-                <SearchBar
-                  searchSource="profile_landing"
-                  organization={organization}
-                  projectIds={selection.projects}
-                  query={query}
-                  onSearch={handleSearch}
-                  maxQueryLength={MAX_QUERY_LENGTH}
-                />
+                {tab === 'transactions' ? (
```
Zylphrex, line 422: Nit: seems like instead of littering `tab === 'transactions' and `tab === 'flamegraph` everywhere, we can just do a single top level one and render the rest of the page appropriately. There will be some code duplication but I think that's cleaner and easier to maintain as we make changes.

Zylphrex (approved):

static/app/views/profiling/content.tsx
JonasBa: Yeah, I was on the fence and didnt do it, but I agree splitting is will make it nicer to work with. I've split it to two separate components now and took the opp to extract some functionality to a separate component too

static/app/views/profiling/content.tsx
```diff
@@ -277,63 +279,42 @@ function ProfilingContentLegacy({location}: ProfilingContentProps) {
   );
 }
 
+function validateTab(tab: unknown): tab is 'flamegraph' | 'transactions' {
+  return tab === 'flamegraph' || tab === 'transactions';
+}
+
+function decodeTab(tab: unknown): 'flamegraph' | 'transactions' {
+  // Fallback to transactions if tab is invalid. We default to transactions
+  // becuase that is going to be the most common perf setup when we release.
```
Zylphrex, line 288:
```suggestion
  // because that is going to be the most common perf setup when we release.
```

Zylphrex (approved):

PR #75202, opened 2024-07-29
nit(integrations): fix typing in integrations base file

cathteng:
Fix typing in `sentry.integrations.base` and `sentry.integrations.manager`


src/sentry/integrations/base.py
```diff
@@ -451,7 +443,7 @@ def is_rate_limited_error(self, exc: Exception) -> bool:
         raise NotImplementedError
 
     @property
-    def metadata(self) -> IntegrationMetadata:
+    def metadata(self) -> dict[str, Any]:
```
cathteng, line 446: this was typed incorrectly, the `IntegrationMetadata` is not associated with each individual integration instance but rather the overall integration implementation (we list the features, the author, a description for the integration,etc). the integration instance's metadata is a dict that varies per integration

src/sentry/integrations/base.py
```diff
@@ -87,14 +79,14 @@ def feature_flag_name(f: str | None) -> str | None:
             return f"integrations-{f}"
         return None
 
-    def _asdict(self) -> dict[str, Sequence[Any]]:
-        metadata = super()._asdict()
+    def asdict(self) -> dict[str, Sequence[Any]]:
```
asottile-sentry, line 82: I think this is probably `dict[str, Any]` -- or could be a `TypedDict` if you want to write one out for the class (though maybe overkill for now?)

src/sentry/integrations/manager.py
```diff
@@ -35,17 +35,17 @@ def exists(self, key: str) -> bool:
         return key in self.__values
 
     def register(self, cls: type[IntegrationProvider]) -> None:
-        self.__values[cls.key] = cls
+        self.__values[str(cls.key)] = cls
```
asottile-sentry, line 38: hmmm the `str(...)` calls are suspicious here -- is this defined as `str | None` somewhere? should it be `: str` and that'd prevent the `str` calls?

src/sentry/utils/audit.py
```diff
@@ -25,7 +26,7 @@
 
 
 def create_audit_entry(
-    request: HttpRequest,
+    request: HttpRequest | Request,
```
asottile-sentry, line 29: hmmm so I'm suspicious of almost every `: Request` in the codebase -- that they usually should be `HttpRequest` -- would it be possible to fix the caller here?

src/sentry/integrations/manager.py
cathteng: i think it's because `cls` is `type[IntegrationProvider]` rather than just `IntegrationProvider`? if i don't wrap it with `str` mypy thinks `cls.key` is `Callable[[PipelineProvider], str]`

src/sentry/utils/audit.py
cathteng: oh actually this is passes typing when i remove it 🤔 , thanks for calling out

src/sentry/integrations/base.py
cathteng: `dict[str, Any]` it is

src/sentry/integrations/manager.py
asottile-sentry: ah, yeah it isn't understanding the `abstractmethod` + `property`

maybe this will fix it?
```
diff --git a/src/sentry/pipeline/provider.py b/src/sentry/pipeline/provider.py
index b08693aef9..3f9df77320 100644
--- a/src/sentry/pipeline/provider.py
+++ b/src/sentry/pipeline/provider.py
@@ -18,13 +18,12 @@ class PipelineProvider(abc.ABC):
     def __init__(self) -> None:
         self.config: dict[str, Any] = {}
 
-    @property
-    @abc.abstractmethod
-    def key(self) -> str:
-        """
-        A unique identifier (e.g. 'slack'). Used to lookup sibling classes and
-        the `key` used when creating Integration objects.
-        """
+    # abstract
+    """
+    A unique identifier (e.g. 'slack'). Used to lookup sibling classes and
+    the `key` used when creating Integration objects.
+    """
+    key: str
 
     @property
     @abc.abstractmethod
 ```

cathteng: it worked 🤯

asottile-sentry (approved):

PR #75206, opened 2024-07-29
feat(custom-views): Add new add view button and temporary tab designs

### LINKED ISSUE #73230: Custom Views: Add "Add View" button to tab bar
Users will need to be able to add new views to the tab bar. We will need to add a "Add View" button to the right of all permanent tabs that enables this feature. Once clicked, a new tab should be created, and the name of the tab should be highlighted to encourage users to name all of their tabs something unique. We will need to debounce the request once the user has completed creating the new view

This feature can be marked as completed once the "Add View" button has been styled according to design mockups and functionality has been completed.

### LINKED ISSUE #73338: Custom Views: Add temporary view component
Custom views will need to support temporary views as part of its feature set. A temporary view appears as a tab after the `Add View` button, and do not persist permanently like normal views. They are opened when a user clicks on a shared link. 

This ticket can be marked as completed once the visual component for the temporary views has been created and matches the design of the internal mockups ([here](https://www.figma.com/design/MM42MEY0OiZRVzQlI8XHej/Q1-Issue-Stream?node-id=1197-43161&t=9HxVvwbscu26O6ma-4)). The functionality to save temporary views is not within the scope of this PR.

MichaelSun48:
closes #73230
closes #73338

This PR adds the `Add View` button to the DraggableTabBar component as well as the design for temporary tabs. It also makes some refactors to the existing draggable tabs components, such as creating the dropdown menu options directly in the draggableTabBar rather than the menu component. 

image

### DIFF
--- a/static/app/components/draggableTabs/draggableTabList.tsx
+++ b/static/app/components/draggableTabs/draggableTabList.tsx
@@ -9,11 +9,13 @@ import {useTabListState} from '@react-stately/tabs';
 import type {Node, Orientation} from '@react-types/shared';
 import {Reorder} from 'framer-motion';
 
+import {Button} from 'sentry/components/button';
 import type {SelectOption} from 'sentry/components/compactSelect';
 import {TabsContext} from 'sentry/components/tabs';
 import {type BaseTabProps, Tab} from 'sentry/components/tabs/tab';
 import {OverflowMenu, useOverflowTabs} from 'sentry/components/tabs/tabList';
 import {tabsShouldForwardProp} from 'sentry/components/tabs/utils';
+import {IconAdd} from 'sentry/icons';
 import {space} from 'sentry/styles/space';
 import {browserHistory} from 'sentry/utils/browserHistory';
 
@@ -29,6 +31,8 @@ function BaseDraggableTabList({
   className,
   outerWrapStyles,
   onReorder,
+  onAddView,
+  showTempTab = false,
   tabVariant = 'filled',
   ...props
 }: BaseDraggableTabListProps) {
@@ -103,6 +107,11 @@ function BaseDraggableTabList({
     });
   }, [state.collection, overflowTabs]);
 
+  const persistentTabs = [...state.collection].filter(
+    item => item.key !== 'temporary-tab'
+  );
+  const tempTab = [...state.collection].find(item => item.key === 'temporary-tab');
+
   return (
     <TabListOuterWrap style={outerWrapStyles}>
       <Reorder.Group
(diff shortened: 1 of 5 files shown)


static/app/components/draggableTabs/index.stories.tsx
```diff
@@ -37,24 +37,37 @@ export default storyBook(DraggableTabBar, story => {
...
+            <DraggableTabBar
+              tabs={TABS}
+              showTempTab={showTempTab}
+              tempTabContent={
+                <TabPanelContainer>This is a Temporary view</TabPanelContainer>
+              }
+              // The add view button should NOT toggle the temp tab normally.
+              // This is a very temporary way to show off the temp tab design to PR reviewers,
+              // and it will be removed in the very near future
+              onAddView={() => setShowTempTab(!showTempTab)}
```
MichaelSun48, line 64: This is a convenient way to show off the design of the temp tab, but this will be removed in a future PR when the addView button's functionality is actually implemented correctly.

static/app/views/issueList/draggableTabBar.tsx
```diff
@@ -24,17 +25,76 @@ export interface Tab {
...
   onRename?: (key: MenuItemProps['key']) => void;
+  /**
+   * Callback function to be called when user clicks the 'Save' button.
+   * Note: The `Save` button only appears for persistent views when `isChanged=true`
+   */
   onSave?: (key: MenuItemProps['key']) => void;
+  /**
+   * Callback function to be called when user clicks the 'Save View' button for temporary views.
+   */
+  onSaveTempView?: () => void;
```
MichaelSun48, line 62: Moving these to this component since there are three different possible variations of the tab menu options, and it makes more sense for this component to handle the logic that picks between them rather than directly inside the tabMenuButton component.

static/app/views/issueList/draggableTabBar.tsx
```diff
@@ -24,17 +25,74 @@ export interface Tab {
...
+    content: tempTabContent,
+  };
+  const [tabs, setTabs] = useState<Tab[]>([...props.tabs, tempTab]);
   const [selectedTabKey, setSelectedTabKey] = useState<Key>(props.tabs[0].key);
 
+  useEffect(() => {
+    if (!showTempTab && selectedTabKey === 'temporary-tab') {
+      setSelectedTabKey(props.tabs[0].key);
+    }
+  }, [showTempTab, selectedTabKey, props.tabs]);
```
MichaelSun48, line 94: If the temp tab goes away, switch to the first tab. (Ideally this would be the most recent tab, but that's out of the scope of this PR)

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -136,12 +146,32 @@ function BaseDraggableTabList({
...
+              {(state.selectedKey === 'temporary-tab' ||
+                (state.selectedKey !== item.key &&
+                  state.collection.getKeyAfter(item.key) !== state.selectedKey)) && (
+                <TabDivider />
+              )}
             </Reorder.Item>
           ))}
+          <AddViewButton borderless size="zero" onClick={onAddView}>
+            <StyledIconAdd size="xs" />
+            Add View
```
malwilley, line 158:
```suggestion
            t('Add View')
```

malwilley (approved):

PR #75207, opened 2024-07-29
chore(similarity): Turn on ML grouping for new EA projects in us region

jangjodi:
On new project creation for EA orgs in the us region, create the similarity backfill completed option

Related options automator PR to turn it on in US region: https://github.com/getsentry/sentry-options-automator/pull/2036


src/sentry/api/endpoints/team_projects.py
```diff
@@ -214,4 +218,15 @@ def post(self, request: Request, team) -> Response:
...
             )
 
+        # Create project option to turn on ML similarity feature for new EA projects
+        if project:
+            is_seer_eligible_platform = project.platform in SEER_ELIGIBLE_PLATFORMS
+            if (
+                project.organization.flags
+                and project.organization.flags.early_adopter
+                and is_seer_eligible_platform
+                and get_local_region().name == "us"
```
JoshFerge, line 228: cc @getsentry/hybrid-cloud is there a preferred way of doing this?

mrduncan: In general, we've done this by using options which can be applied on a per-region basis (and also work nicely for enabling in ST)

JoshFerge: @jangjodi can we use an option for this? maybe make a boolean that is default false across all regions, then `true` for us?

jangjodi: thanks! I'll make this an option instead

JoshFerge (approved):

PR #75209, opened 2024-07-29
ref(seer grouping): Handle Seer only sending matching groups

lobsterkatie:
Seer no longer sends back similar but non-matching groups when called during ingest, which means on the Sentry side, we no longer need to handle that case. This PR therefore removes the `should_group` check on retrieving a parent group, since Seer will only send back results if `should_group` is true. It also stops storing the Seer results on the group, since a) the only time there's a new group on which to store the results is when a match isn't found, and b) in cases where a match isn't found, there now won't be any similar groups to store, only an empty results list. Finally, since the metadata is now only going to be stored on the event, and the event knows its own hash, `request_hash` has been removed from the metadata we store.

### DIFF
--- a/src/sentry/event_manager.py
+++ b/src/sentry/event_manager.py
@@ -1551,13 +1551,6 @@ def _save_aggregate(
                         )
                         event.data["seer_similarity"] = seer_response_data
 
-                        # We only want to add this data to new groups, while we're testing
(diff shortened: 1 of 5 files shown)


JoshFerge (approved):
awesome!

PR #75210, opened 2024-07-29
feat(toolbar): add red indicator badge if crashes >> healthy sessions

### LINKED ISSUE #75184: [DevToolbar] Show an indicator on the Open/Close Releases button that indicates whether the latest release has more or less crashes than the prev release
We could do like a subtle green indicator with a checkbox icon when things are good, and a red one with an arrow when things are bad. 

Add a tooltip so we know if it's crash-free rate (which is what's inside the panel) or crash-count. 
- Crash-free rate, higher is better, up to 100%
- Crash-count, lower is better, down to zero.

michellewzhang:
- closes https://github.com/getsentry/sentry/issues/75184
- add red indicator badge with fire icon if over 10% of sessions are crashes (this threshold can be changed; i picked it arbitrarily)
- for ratios < 10%, no indicator badge

SCR-20240729-ocbq

### DIFF
--- a/static/app/components/devtoolbar/components/releases/sessionStatusBadge.tsx
+++ b/static/app/components/devtoolbar/components/releases/sessionStatusBadge.tsx
@@ -1,0 +1,24 @@
+import IndicatorBadge from 'sentry/components/devtoolbar/components/indicatorBadge';
+import useSessionStatus from 'sentry/components/devtoolbar/components/releases/useSessionStatus';
+import {IconFire} from 'sentry/icons';
+
+export default function CrashCountBadge() {
+  const {data} = useSessionStatus();
+  const healthySessions = data?.json.groups.find(
+    g => g.by['session.status'] === 'healthy'
+  )?.totals['sum(session)'];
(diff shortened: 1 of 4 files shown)


static/app/components/devtoolbar/components/releases/sessionStatusBadge.tsx
```diff
@@ -0,0 +1,24 @@
...
+    g => g.by['session.status'] === 'healthy'
+  )?.totals['sum(session)'];
+  const crashSessions = data?.json.groups.find(g => g.by['session.status'] === 'crashed')
+    ?.totals['sum(session)'];
+
+  if (!crashSessions || !healthySessions) {
+    return null;
+  }
+
+  const crashPercent = (crashSessions / healthySessions) * 100;
```
c298lee, line 17: Since it's percentage of sessions that were crashes, should we be using total sessions instead?
```suggestion
  const crashPercent = (crashSessions / (crashSessions + healthySessions)) * 100;
```

michellewzhang: oh yes good catch

ryan953 (approved):

PR #75212, opened 2024-07-29
chore(slack): actually type issues message builder correctly

cathteng:
and don't break Slack in the process

Re-up of #74876

### DIFF
--- a/src/sentry/integrations/message_builder.py
+++ b/src/sentry/integrations/message_builder.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from collections.abc import Callable, Iterable, Mapping, Sequence
+from collections.abc import Callable, Mapping, Sequence
 from typing import Any
 
 from sentry import features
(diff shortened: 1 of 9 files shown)


src/sentry/integrations/slack/message_builder/base/block.py
```diff
@@ -176,7 +176,7 @@ def _build_blocks(
         *args: SlackBlock,
         fallback_text: str | None = None,
         color: str | None = None,
-        block_id: dict[str, int] | None = None,
+        block_id: dict[str, int] | str | None = None,
```
GabeVillalobos, line 179: This is a bit of an unusual type. Are we calling this with both dictionaries and strings?

src/sentry/integrations/slack/message_builder/issues.py
```diff
@@ -147,40 +147,20 @@ def build_action_text(identity: RpcIdentity, action: MessageAction) -> str | Non
...
-                }
-            )
-    return fields
-
-
-def format_release_tag(value: str, event: GroupEvent | Group):
+def format_release_tag(value: str, event: GroupEvent | None) -> str:
     """Format the release tag using the short version and make it a link"""
+    if not event:
+        return ""
```
GabeVillalobos, line 153: Would it be better if we return `None` here instead of `""`? Otherwise the caller will have to perform falsey checks to validate whether this was successful or not.

src/sentry/utils/committers.py
```diff
@@ -293,7 +293,7 @@ def get_event_file_committers(
 
 
 def get_serialized_event_file_committers(
-    project: Project, event: Event, frame_limit: int = 25
+    project: Project, event: Event | GroupEvent, frame_limit: int = 25
```
GabeVillalobos, line 296: Since both of these are subclasses of `BaseEvent`, could we loosen the type to the parent by chance?

src/sentry/integrations/slack/webhooks/options_load.py
```diff
@@ -69,15 +74,17 @@ def get_filtered_option_groups(
...
-            member_options = format_actor_options(filtered_members, True)
-            option_groups.append(
-                {"label": {"type": "plain_text", "text": "People"}, "options": member_options}
-            )
+            member_options_group: OptionGroup = {
+                "label": {"type": "plain_text", "text": "People"},
+                "options": format_actor_options(filtered_members, True),
+            }
+            option_groups.append(member_options_group)
         return option_groups
```
GabeVillalobos, line 88: It might be better just to update this to `list[OptionGroup]` since we now have a concrete type. Having `Any` in the return type might mask more type issues in the callers.

src/sentry/integrations/slack/message_builder/base/block.py
cathteng: i'm not entirely sure, i can try to remove it and see if anything fails. we're inserting whatever `block_id` is into a dict though

src/sentry/integrations/slack/message_builder/issues.py
cathteng: this will show up as a release tag like the below, i believe `None` gets rendered as "null" when dumped into json. i'm not sure which is preferable tbh

Screenshot 2024-07-30 at 15 06 55

src/sentry/integrations/slack/message_builder/base/block.py
cathteng: ok nothing fails, we're good to go

asottile-sentry (approved):

iamrajjoshi (approved):

GabeVillalobos (approved):

schew2381 (approved):
for good measure

asottile-sentry:
oops looks like this got caught in a merge race:

```
src/sentry/integrations/msteams/card_builder/issues.py:187: error: Argument 1 to "format_actor_options" has incompatible type "BaseQuerySet[Team, Team]"; expected "Sequence[Team | RpcUser]"  [arg-type]
src/sentry/integrations/slack/message_builder/issues.py:240: error: Argument 1 to "format_actor_options" has incompatible type "BaseQuerySet[Team, Team]"; expected "Sequence[Team | RpcUser]"  [arg-type]
```

src/sentry/integrations/message_builder.py
```diff
@@ -20,17 +20,17 @@
 
 
 def format_actor_options(
-    actors: Iterable[Team | RpcUser], use_block_kit: bool = False
+    actors: Sequence[Team | RpcUser], is_slack: bool = False
```
asottile-sentry, line 23: this particular line -- the `Iterable` should have stayed -- probably a merge conflict sorry about that!

cathteng: oopsie, thanks for letting me know

PR #75214, opened 2024-07-30
feat(anomaly detection): warn users if less than one week of historical data is available (backend)

mifu67:
If snuba returns fewer than 7 days worth of historical data, then surface a warning.


mifu67:
ack accidentally branched off of a branch ;_;

tests/sentry/incidents/endpoints/test_organization_alert_rule_index.py
```diff
@@ -309,6 +312,60 @@ def test_anomaly_detection_alert_seer_timeout_max_retry(self, mock_logger, mock_
         assert resp.data["detail"] == "Timeout error when hitting Seer store data endpoint"
         assert mock_seer_request.call_count == 1
 
+    @with_feature("organizations:anomaly-detection-alerts")
+    @with_feature("organizations:incidents")
+    @patch(
+        "sentry.seer.anomaly_detection.store_data.seer_anomaly_detection_connection_pool.urlopen"
+    )
+    def test_anomaly_detection_alert_not_enough_data(self, mock_seer_request):
```
mifu67, line 320: This test mysteriously started failing and just as mysteriously began working again... nothing about it looks flaky, as far as I can tell, but maybe there's something to do with time here that I'm missing?

tests/sentry/incidents/endpoints/test_organization_alert_rule_index.py
```diff
@@ -309,6 +312,60 @@ def test_anomaly_detection_alert_seer_timeout_max_retry(self, mock_logger, mock_
...
+        data = {
+            **self.alert_rule_dict,
+            "detection_type": AlertRuleDetectionType.DYNAMIC,
+            "sensitivity": AlertRuleSensitivity.LOW,
+            "seasonality": AlertRuleSeasonality.AUTO,
+        }
+        mock_seer_request.return_value = HTTPResponse(status=202)
```
nhsiehgit, line 327: chatted in person - but i think we actually want this to be a mocked 200
the helper method that this is being pulled through is actually modifying the response which isn't ideal, but for now we should test the existing implementation.

lets add a note here to why we're mocking the one status and asserting the other though

src/sentry/incidents/endpoints/organization_alert_rule_index.py
```diff
@@ -121,10 +121,15 @@ def create_metric_alert(self, request, organization, project=None):
                 return Response({"uuid": client.uuid}, status=202)
                 # TODO handle slack channel lookup AND anomaly detection
             else:
+                http_status = 201
                 alert_rule = serializer.save()
                 if alert_rule.detection_type == AlertRuleDetectionType.DYNAMIC.value:
                     resp = send_historical_data_to_seer(alert_rule=alert_rule, user=request.user)
```
nhsiehgit, line 127: Can we label this something more descriptive rather than `resp`?
Was a bit confused since this isn't the straight Seer response either, but a modified response

Might be out of scope for this pr though :/ 
since we're only using it to determine whether the seer backfill was successful or not, we could just have `send_historical_data_to_seer` raise an exception if there's not enough data instead 🤔

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -50,6 +52,21 @@ def format_historical_data(data: SnubaTSResult) -> list[TimeSeriesPoint]:
...
+def get_start_and_end(data: SnubaTSResult) -> tuple[int, int]:
+    start, end = -1, -1
+    indices_with_results = []
+    for i, datum in enumerate(data.data.get("data", [])):
+        if "count" in datum:
+            indices_with_results.append(i)
+    if not indices_with_results:
+        return start, end
+    else:
+        start = indices_with_results[0]
```
nhsiehgit, line 65: Ahhh grab the first and the last because the data is already ordered is that right?
Maybe we should add some quick validations in here for that?

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -50,6 +52,21 @@ def format_historical_data(data: SnubaTSResult) -> list[TimeSeriesPoint]:
     return formatted_data
 
 
+# return the start and end indices
+def get_start_and_end(data: SnubaTSResult) -> tuple[int, int]:
```
nhsiehgit, line 56: Maybe we make this an `_get_start_end` underscored helper?
Since this is generically named, i wouldn't want anyone else to pull and try to use this helper

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -136,7 +153,19 @@ def send_historical_data_to_seer(alert_rule: AlertRule, user: User) -> BaseHTTPR
         base_error_response.status = status.HTTP_408_REQUEST_TIMEOUT
         return base_error_response
 
-    # TODO warn if there isn't at least 7 days of data
+    MIN_DAYS = 7
+    data_start_index, data_end_index = get_start_and_end(historical_data)
+    if data_start_index == -1:
+        resp.status = 202
```
nhsiehgit, line 159: Ah ok - RE: comments above, lets just throw an exception if there's not enough data.
This _should_ be a loud failure, rather than a seeming success

nhsiehgit (commented):
few comments

src/sentry/seer/anomaly_detection/store_data.py
mifu67: I think that we do create the alert rule—it just won't detect any anomalies until Seer has at least a week of data. Should we still have an exception in this case?

src/sentry/seer/anomaly_detection/store_data.py
mifu67: Yeah, we're looping through the data in order, so start is guaranteed to be <= end

mifu67: I wouldn't be able to test the validation—should I still add it?

tests/sentry/incidents/endpoints/test_organization_alert_rule_index.py
```diff
@@ -97,6 +100,39 @@ def alert_rule_dict(self):
             "activations": [],
         }
 
+    @cached_property
+    def dynamic_alert_rule_dict(self):
```
mifu67, line 104: Just cleanup because we use the dynamic rule dict in four different tests now

src/sentry/incidents/endpoints/organization_alert_rule_index.py
```diff
@@ -106,6 +106,7 @@ def create_metric_alert(self, request, organization, project=None):
             },
             data=data,
         )
+
```
nhsiehgit, line 109: 🤷

src/sentry/incidents/logic.py
```diff
@@ -652,7 +652,10 @@ def create_alert_rule(
                 )
 
             try:
-                send_historical_data_to_seer(alert_rule=alert_rule, project=projects[0])
+                resp = send_historical_data_to_seer(alert_rule=alert_rule, project=projects[0])
```
nhsiehgit, line 655: nit: Can we instead update `send_historical_data_to_seer` to return a boolean?
Or maybe we can just have it return an alert rule status?
if bool, we could rename this variable to be `backfill_success` or something

I just find it odd needing to inspect a `response` status here 🤔

src/sentry/incidents/logic.py
```diff
@@ -652,7 +652,10 @@ def create_alert_rule(
                 )
 
             try:
```
nhsiehgit, line 654: 👌 

So we've covered `create_alert_rule`,
Lets also be sure to capture `update_alert_rule` as well

src/sentry/incidents/subscription_processor.py
```diff
@@ -637,6 +648,14 @@ def has_anomaly(self, anomaly, label: str) -> bool:
             return True
         return False
 
+    def has_enough_data(self, anomaly) -> bool:
```
nhsiehgit, line 651: `has_enough_data` sounds pretty generic
Can we name this something more explicit? maybe like `anomaly_has_confidence`?

src/sentry/seer/anomaly_detection/store_data.py
nhsiehgit: maybe a quick assertion here so that it would raise an exception 🤷 

could just be a quick `assert start < end`
This way it will raise in case this is ever false

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -49,6 +51,21 @@ def format_historical_data(data: SnubaTSResult) -> list[TimeSeriesPoint]:
     return formatted_data
 
 
+# return the start and end indices
+def _get_start_and_end(data: SnubaTSResult) -> tuple[int, int]:
```
nhsiehgit, line 55: If we're returning indices, can we update the name to reflect that?
Also the note can go into a docstring `"""` so it shows up on hover

src/sentry/seer/anomaly_detection/store_data.py
nhsiehgit: got it, that makes sense!

Commented above - rather than modifying the response and returning the response,
Thoughts on simply returning a boolean for successful backfill or not?

we're not utilizing the rest of the response, and modifying the response is unexpected (for ex. mocking seer response w/ 200, but expecting a 202?)

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -620,6 +621,106 @@ def test_seer_call_timeout_error(self, mock_logger, mock_seer_request):
 
         assert result is None
 
+    @with_feature("organizations:anomaly-detection-alerts")
+    @mock.patch(
+        "sentry.incidents.subscription_processor.SubscriptionProcessor.seer_anomaly_detection_connection_pool.urlopen"
+    )
+    def test_dynamic_alert_rule_not_enough_data(self, mock_seer_request):
+        rule = self.dynamic_rule
+        rule.update(status=AlertRuleStatus.NOT_ENOUGH_DATA.value)
```
nhsiehgit, line 630: Do we need to refresh the rule after this update as well?

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -620,6 +621,106 @@ def test_seer_call_timeout_error(self, mock_logger, mock_seer_request):
...
+        assert rule.status == AlertRuleStatus.NOT_ENOUGH_DATA.value
+        self.assert_trigger_counts(processor, self.trigger, 0, 0)
+
+    @with_feature("organizations:anomaly-detection-alerts")
+    @mock.patch(
+        "sentry.incidents.subscription_processor.SubscriptionProcessor.seer_anomaly_detection_connection_pool.urlopen"
+    )
+    def test_enable_dynamic_alert_rule(self, mock_seer_request):
+        rule = self.dynamic_rule
+        rule.update(status=AlertRuleStatus.NOT_ENOUGH_DATA.value)
```
nhsiehgit, line 661: Do we need to refresh the rule after this update as well?

tests/sentry/incidents/test_subscription_processor.py
```diff
@@ -620,6 +621,106 @@ def test_seer_call_timeout_error(self, mock_logger, mock_seer_request):
...
+        assert rule.status == AlertRuleStatus.PENDING.value
+        self.assert_trigger_counts(processor, self.trigger, 0, 0)
+
+    @with_feature("organizations:anomaly-detection-alerts")
+    @mock.patch(
+        "sentry.incidents.subscription_processor.SubscriptionProcessor.seer_anomaly_detection_connection_pool.urlopen"
+    )
+    def test_enable_dynamic_alert_rule_and_fire(self, mock_seer_request):
+        rule = self.dynamic_rule
+        rule.update(status=AlertRuleStatus.NOT_ENOUGH_DATA.value)
```
nhsiehgit, line 692: Do we need to refresh the rule after this update as well?

nhsiehgit (commented):
few comments!

src/sentry/incidents/subscription_processor.py
```diff
@@ -550,6 +551,17 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+                        # check to see if we have enough data for the dynamic alert rule now
+                        if self.alert_rule.status == AlertRuleStatus.NOT_ENOUGH_DATA.value:
+                            if self.has_enough_data(potential_anomaly):
+                                # NOTE: this means "enabled," and it's the default alert rule status.
+                                # TODO: change these status labels to be less confusing
+                                self.alert_rule.status = AlertRuleStatus.PENDING.value
+                                self.alert_rule.save()
+                            else:
+                                # we don't need to check if the alert should fire if the alert can't fire yet
+                                continue
```
ceorourke, line 563: Do we need to continue iterating through this - like would there be a situation in which the next `potential_anomaly` would change the result of `self.has_enough_data()`? If not, we can just break here (which I realize is also a little silly since we expect one anomaly, but 🤷‍♀️

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -49,6 +51,21 @@ def format_historical_data(data: SnubaTSResult) -> list[TimeSeriesPoint]:
     return formatted_data
 
 
+# return the start and end indices
```
ceorourke, line 54: is this supposed to be here?

tests/sentry/incidents/test_subscription_processor.py
mifu67: Apparently not? I checked the rule status and it is "NOT_ENOUGH_DATA" when we enter the subscription processor logic. ¯\_(ツ)_/¯

src/sentry/seer/anomaly_detection/store_data.py
mifu67: The comment? I can remove it.

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -115,7 +132,19 @@ def send_historical_data_to_seer(alert_rule: AlertRule, project: Project) -> Bas
...
 
-    # TODO warn if there isn't at least 7 days of data
+    MIN_DAYS = 7
+    data_start_index, data_end_index = _get_start_and_end(historical_data)
+    if data_start_index == -1:
+        resp.status = 202
+        resp.reason = NOT_ENOUGH_DATA  # this reason value isn't read anywhere, but I think it's good to justify the 202
+
+    data_start_time = datetime.fromtimestamp(
+        historical_data.data.get("data")[data_start_index]["time"]
```
ceorourke, line 142: not a huge deal but since we're checking after the data has been formatted you could use `formatted_data` instead and do `formatted_data[data_start_index].get("timestamp")`

src/sentry/incidents/subscription_processor.py
mifu67: I don't *think* so, but I also don't know, which is why I did continue instead of break. If only one anomaly is returned, like how it's spec'd in Seer, then they accomplish the same thing. I think continue is just a little safer.

src/sentry/seer/anomaly_detection/store_data.py
mifu67: I think we can do that

src/sentry/incidents/subscription_processor.py
ceorourke: It could have "none" and still not "have confidence" (aka be either `HIGH_CONFIDENCE` or `LOW_CONFIDENCE`) which is different from no data so to me that makes it more confusing. Would it help if this were moved to `store_data.py`?

src/sentry/incidents/logic.py
mifu67: I'll work in that in the next PR after Colleen's changes to `update_alert_rule` land

src/sentry/incidents/subscription_processor.py
nhsiehgit: 🤔 

Would Seer ever actually return any potential anomalies if we don't have enough data?
This check may actually be unnecessary? (but safe to have anyways)

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -32,6 +31,8 @@
     timeout=settings.SEER_ANOMALY_DETECTION_TIMEOUT,
 )
 
+NOT_ENOUGH_DATA = "Fewer than seven days of historical data available"
```
nhsiehgit, line 34: nit: can we update this to be `NOT_ENOUGH_DATA_MSG` to make it more explicit from the status?

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -49,26 +50,35 @@ def format_historical_data(data: SnubaTSResult) -> list[TimeSeriesPoint]:
...
-    )
-
     snuba_query = SnubaQuery.objects.get(id=alert_rule.snuba_query_id)
     window_min = int(snuba_query.time_window / 60)
     historical_data = fetch_historical_data(alert_rule, snuba_query, project)
 
     if not historical_data:
-        base_error_response.reason = "No historical data available. Cannot create alert_rule."
-        return base_error_response
+        raise ValidationError("No historical data available. Cannot create alert_rule.")
```
nhsiehgit, line 81: nit: Since we're not actually modifying the alert_rule here, lets not tie this msg explicitly to the alert_rule
I think just `No historical data available` is good enough for a msg?

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -79,10 +89,9 @@ def send_historical_data_to_seer(alert_rule: AlertRule, project: Project) -> Bas
         or alert_rule.organization is None
     ):
         # this won't happen because we've already gone through the serializer, but mypy insists
-        base_error_response.reason = (
+        raise ValidationError(
             "Cannot create alert_rule - missing expected configuration for a dynamic alert."
```
nhsiehgit, line 93: oops - missed this too
nit: lets not tie this msg explicitly to the alert_rule either, just the second half is good 👍

nhsiehgit (approved):
hell yea!

a few nit comments, but this is looking sharp!

src/sentry/seer/anomaly_detection/store_data.py
mifu67: We don't use it anymore, so I'll just get rid of it

PR #75215, opened 2024-07-30
ref(rules): Move Snuba query and Seer send to create_alert_rule

ceorourke:
Move the call to fetch 28 days of historical data from Snuba and send it to Seer to `create_metric_alert` in `logic.py`.
Original implementation in the endpoint is [here](https://github.com/getsentry/sentry/pull/74614).

Closes https://getsentry.atlassian.net/browse/ALRT-170


tests/sentry/incidents/endpoints/test_organization_alert_rule_index.py
```diff
@@ -103,6 +103,25 @@ class AlertRuleIndexBase(AlertRuleBase):
 
     endpoint = "sentry-api-0-organization-alert-rules"
 
+    def setUp(self):
+        super().setUp()
```
ceorourke, line 107: This is just a little refactoring as I realized we were using the same data in multiple tests

tests/sentry/incidents/endpoints/test_organization_alert_rule_index.py
```diff
@@ -796,36 +883,21 @@ def test_kicks_off_slack_async_job(
...
+                    "actions": self.slack_action,
                 },
             ],
         }
         with self.feature(["organizations:incidents"]):
             resp = self.get_success_response(
                 self.organization.slug, status_code=202, **valid_alert_rule
             )
-        resp.data["uuid"] = "abc123"
+        assert resp.data["uuid"] == "abc123"
```
ceorourke, line 900: This was supposed to be an assertion but [2023 me](https://github.com/getsentry/sentry/commit/697e6385d69ed3004df5c31ece1a41cc30c784b0) made a mistake

mifu67 (commented):
This looks good to me, but I'll defer approval for Nathan since he has more context.

src/sentry/tasks/integrations/slack/find_channel_id_for_alert_rule.py
```diff
@@ -104,14 +105,31 @@ def find_channel_id_for_alert_rule(
...
+        else:
+            success = True
+
+        # XXX: this is not the ideal location to do this, but because we need the rule id at this stage
+        # it's the only option outside of major changes. see blame PR description for more details
+        if success and alert_rule:
+            if alert_rule.detection_type == AlertRuleDetectionType.DYNAMIC.value:
+                if not user:
+                    redis_rule_status.set_value("failed")
+                    return
```
nhsiehgit, line 124: should we delete the alert if failed?

nhsiehgit (commented):
:squint: Oh...

the big change here is that we're fetching and sending the data to seer...
in the slack find_channel_id_for_alert_rule task?!? :face_with_open_mouth_vomiting: 

Question

If we're fetching/rendering a preview before submitting the form and creating the alert rule

Can we not just send the data from the FE?
This way we don't need to tie in to an unrelated task?

nhsiehgit:
Responded in slack, but adding comments on the PR here

We may actually want to add the `send_historical_data_to_seer` method to the `post_save` of the Alertrule model.

This way - we don't need to implement this in 2 separate spots, and we can avoid coupling this logic closely with the Slack action creation logic

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -50,7 +49,7 @@ def format_historical_data(data: SnubaTSResult) -> list[TimeSeriesPoint]:
     return formatted_data
 
 
-def send_historical_data_to_seer(alert_rule: AlertRule, user: User) -> BaseHTTPResponse:
```
ceorourke, line 53: I moved the feature check to `logic.py` which is the only thing I was using `user` for

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -157,8 +138,8 @@ def fetch_historical_data(alert_rule: AlertRule, snuba_query: SnubaQuery) -> Snu
         # DATSET_OPTIONS expects the name 'errors'
         dataset_label = "errors"
     dataset = get_dataset(dataset_label)
-    project = alert_rule.projects.get()
```
ceorourke, line 160: projects get added to the join table `AlertRuleProjects` _after_ `send_historical_data_to_seer` so I just pass it in instead. In case it fails I didn't want to have to delete more stuff.

tests/sentry/incidents/endpoints/test_organization_alert_rule_index.py
```diff
@@ -291,72 +310,6 @@ def test_anomaly_detection_alert(self, mock_seer_request):
         assert alert_rule.sensitivity == resp.data.get("sensitivity")
         assert mock_seer_request.call_count == 1
 
-    @with_feature("organizations:anomaly-detection-alerts")
-    @with_feature("organizations:incidents")
-    @patch(
-        "sentry.seer.anomaly_detection.store_data.seer_anomaly_detection_connection_pool.urlopen"
-    )
-    @patch("sentry.seer.anomaly_detection.store_data.logger")
-    def test_anomaly_detection_alert_seer_timeout_max_retry(self, mock_logger, mock_seer_request):
```
ceorourke, line 300: These got moved and rewritten in `test_logic.py`

src/sentry/incidents/logic.py
```diff
@@ -629,6 +631,13 @@ def create_alert_rule(
             detection_type=detection_type,
         )
 
+        if alert_rule.detection_type == AlertRuleDetectionType.DYNAMIC.value:
+            if not features.has("organizations:anomaly-detection-alerts", organization, actor=user):
+                alert_rule.delete()
+                raise ResourceDoesNotExist
```
nhsiehgit, line 637: Can we add details to this exception?

Maybe just a `Feature is not available`?

src/sentry/incidents/logic.py
```diff
@@ -629,6 +631,13 @@ def create_alert_rule(
             detection_type=detection_type,
         )
 
+        if alert_rule.detection_type == AlertRuleDetectionType.DYNAMIC.value:
```
nhsiehgit, line 634: Can we move this check below the project checks?

I don't think multiple projects is utilized atm, but since the logic exists maybe we implement this assuming that multiple projects per rule is valid today?

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -125,23 +106,23 @@ def send_historical_data_to_seer(alert_rule: AlertRule, user: User) -> BaseHTTPR
...
+            "Timeout error when hitting Seer store data endpoint",
             extra={
                 "rule_id": alert_rule.id,
                 "project_id": project.id,
             },
         )
-        base_error_response.reason = timeout_text
-        base_error_response.status = status.HTTP_408_REQUEST_TIMEOUT
-        return base_error_response
+        alert_rule.delete()
```
nhsiehgit, line 116: rather than deleting the rule here, can we add a try catch in `logic.py` and delete if we encounter an exception?

nhsiehgit (commented):
nice nice nice

few small comments

src/sentry/incidents/logic.py
ceorourke: The `include_all_projects` part or the `AlertRuleProjects` creation or the `alert_rule.subscribe_projects` part?

ceorourke: I moved it to below `include_all_projects` but lmk if you wanted it lower

tests/sentry/seer/anomaly_detection/test_store_data.py
```diff
@@ -0,0 +1,62 @@
...
+        )
+        time_2 = datetime.timestamp(
+            before_now(days=1).replace(hour=11, minute=0, second=0, microsecond=0)
+        )
+        expected_return_value = [
+            {"timestamp": time_1, "value": 0},
+            {"timestamp": time_2, "value": 1},
+        ]
+        snuba_raw_data = [{"time": time_1}, {"time": time_2, "count": 1}]
+        # SnubaTSResult = namedtuple("SnubaTSResult", "data")
```
nhsiehgit, line 24: remove?

tests/sentry/seer/anomaly_detection/test_store_data.py
```diff
@@ -0,0 +1,62 @@
...
+
+from sentry.seer.anomaly_detection.store_data import fetch_historical_data, format_historical_data
+from sentry.snuba.models import SnubaQuery
+from sentry.testutils.factories import EventType
+from sentry.testutils.helpers.datetime import before_now, iso_format
+from sentry.utils.snuba import SnubaTSResult
+from tests.sentry.incidents.endpoints.test_organization_alert_rule_index import AlertRuleBase
+
+
+class AnomalyDetectionStoreDataTest(AlertRuleBase):
```
nhsiehgit, line 11: If we're `store_event`'ing data, then i think we'll want to inherit from the SnubaTestCase as well

nhsiehgit (approved):
Thank you for the changes!

few nit comments, but +1 once they've been updated

tests/sentry/seer/anomaly_detection/test_store_data.py
ceorourke: I don't think there'd be any data if it wasn't actually stored.. ?

PR #75058, opened 2024-07-26
ref(metrics): Replace SmartSearch with SearchQueryBuilder component

priscilawebdev:
### Goal

Replace the component `SmartSearchBar` with `SearchQueryBuilder` based on the [instructions](https://sentry.sentry.io/stories/?name=app/components/searchQueryBuilder/index.stories.tsx). 

The changes are behind the feature flag `organizations:search-query-builder-metrics` as defined in [the file](https://github.com/getsentry/sentry-options-automator/blob/831a7b55cf3c3be2d57eb6cab161571f08fe226c/options/default/flagpole.yml#L489)

PR #75221, opened 2024-07-30
ref(metrics): Replace SmartSearch with SearchQueryBuilder component

priscilawebdev:
### Goal

This [PR](https://github.com/getsentry/sentry/pull/75058) was initially reverted because the onChange event was firing too frequently in our main organization compared to our test organization. However, with the [fix](https://github.com/getsentry/sentry/pull/75082) now merged, everything looks good.

https://github.com/user-attachments/assets/92ef5772-da0c-4cb5-8472-92a2249d1ca6

### Note

- The feature flag is not available for our test organization, but I enabled it temporarily (hardcoded) to test the search bar
- Changes were tested in our main org but there is no preview


static/app/components/metrics/metricSearchBar.tsx
```diff
@@ -40,7 +47,7 @@ export function MetricSearchBar({
   id: idProp,
   ...props
 }: MetricSearchBarProps) {
-  const org = useOrganization();
+  const organization = useOrganization();
```
priscilawebdev, line 50: minor change, please let me know if it is ok to introduce this here

shellmayr (approved):
From a usage perspective: LGTM

obostjancic (approved):
Works great! 🚀

PR #75222, opened 2024-07-30
fix(metrics) fix max length on small width screens

### LINKED ISSUE #75052: Visualize field and ellipisis value
On the metrics page, we decided to make the visualization field wider than it previously was to avoid an odd-looking empty space. This [PR](https://github.com/getsentry/sentry/pull/74944) introduces these changes, but on small viewports, the selected value does not display ellipses (three dots) when it overflows the space. The fix is not straightforward, as selecting another value from the dropdown causes the ellipses to disappear, possibly due to the layout calculations performed for this field in the Combobox component.

Additionally, we are using the function from [this link](https://github.com/getsentry/sentry/blob/0067c2809f76a9dc9b3885d3181cb42b909e4740/static/app/utils/string/middleEllipsis.tsx#L15) to trim very long values. This means that even if the field takes up the whole available space, it can still display with ellipses, even though it would have enough space to render the full value.

vgrozdanic:
Fix breaking of the metric name on smaller screens.

Closes https://github.com/getsentry/sentry/issues/75052

### DIFF
--- a/static/app/components/comboBox/index.tsx
+++ b/static/app/components/comboBox/index.tsx
@@ -240,8 +240,6 @@ function ControlledComboBox<Value extends string>({
     );
   });
 
-  // Sync input value with value prop
-  const previousValue = useRef(value);
   useEffect(() => {
     if (value) {
       const selectedLabel = options
@@ -250,7 +248,6 @@ function ControlledComboBox<Value extends string>({
       if (selectedLabel) {
         setInputValue(selectedLabel);
       }
-      previousValue.current = value;
     }
   }, [value, options]);
 
--- a/static/app/components/comboBox/index.tsx
+++ b/static/app/components/comboBox/index.tsx
@@ -242,15 +242,17 @@ function ControlledComboBox<Value extends string>({
 
   // Sync input value with value prop
   const previousValue = useRef(value);
-  if (previousValue.current !== value) {
-    const selectedLabel = options
-      .flatMap(item => ('options' in item ? item.options : [item]))
-      .find(option => option.value === value)?.label;
-    if (selectedLabel) {
-      setInputValue(selectedLabel);
+  useEffect(() => {
+    if (value) {
+      const selectedLabel = options
+        .flatMap(item => ('options' in item ? item.options : [item]))
+        .find(option => option.value === value)?.label;
+      if (selectedLabel) {
+        setInputValue(selectedLabel);
+      }
+      previousValue.current = value;
     }
-    previousValue.current = value;
-  }
+  }, [value, options]);
 
   const items = useMemo(() => {
     return getItemsWithKeys(options) as ComboBoxOptionOrSectionWithKey<Value>[];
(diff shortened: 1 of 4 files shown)


static/app/components/comboBox/index.tsx
```diff
@@ -242,15 +242,17 @@ function ControlledComboBox<Value extends string>({
 
   // Sync input value with value prop
   const previousValue = useRef(value);
-  if (previousValue.current !== value) {
-    const selectedLabel = options
-      .flatMap(item => ('options' in item ? item.options : [item]))
-      .find(option => option.value === value)?.label;
-    if (selectedLabel) {
-      setInputValue(selectedLabel);
+  useEffect(() => {
```
obostjancic, line 245: is this `useEffect` needed? I'm not very familiar with this code but I assume that we are storing value in the ref in order to avoid a re-render every time the value changes.

obostjancic (approved):

static/app/components/comboBox/index.tsx
vgrozdanic: But I have missed to remove `useRef` here, now that is fixed

static/app/components/metrics/mriSelect/index.tsx
```diff
@@ -214,6 +216,19 @@ export const MRISelect = memo(function MRISelect({
...
+    if (breakpoints.small) {
+      // at least small size screen, no problem with fitting 100 characters
+      return 100;
+    }
+    if (breakpoints.xsmall) {
+      return 35; // at least xsmall size screen, no problem with fitting 35 characters
+    }
+
+    // screen smaller than xsmall, 22 characters is optimal
+    return 22;
```
priscilawebdev, line 229: please let's change it to '20' so the string fits for the iPhone see as well 

image

image

static/app/components/metrics/mriSelect/index.tsx
```diff
@@ -214,6 +216,19 @@ export const MRISelect = memo(function MRISelect({
     [onChange]
   );
 
+  const maxLength = useMemo(() => {
+    if (breakpoints.small) {
+      // at least small size screen, no problem with fitting 100 characters
+      return 100;
+    }
+    if (breakpoints.xsmall) {
+      return 35; // at least xsmall size screen, no problem with fitting 35 characters
```
priscilawebdev, line 225: yes 35 looks good here :)

image

static/app/components/comboBox/index.tsx
priscilawebdev, line 243: did we decide to use `useEffect` because the change of the values is a side effect?

vgrozdanic: Yes, value comes from `options` list, and it should be updated if that list is updated. And that list might get updated if the screen size changes

priscilawebdev (approved):
great work 🚀 tested the changes and it looks good to me!

PR #75231, opened 2024-07-30
Experiment with simplifications

smeubank:
Want feedback from the team, extra for example I though wasn't necessary

Screenshot 2024-07-30 at 16 38 29
//the dsn is useless ignore


antonpirker (approved):
Looks good!

The django extra is just there to make sure that only supported Django versions are installed with the sentry-sdk. The minimum version we support Django 1.8 with an end of life date almost 10 years ago so it is safe to remove the extra!

sentrivana (approved):
I'd also remove the "You need a minimum version 1.18.0 of the sentry-python SDK for the profiling feature." part -- if you follow the doc, you're installing the SDK from scratch, so the likelihood of getting an ancient version like that is about 0.0000000001%. (It can only happen if you had it installed previously and it's still cached. New installs always pull the latest SDK.)

static/app/gettingStartedDocs/python/django.tsx
```diff
@@ -39,31 +39,15 @@ sentry_sdk.init(
...
-          description: params.isProfilingSelected
-            ? tct(
-                'You need a minimum version [codeVersion:1.18.0] of the [codePackage:sentry-python] SDK for the profiling feature.',
-                {
-                  codeVersion: <code />,
-                  codePackage: <code />,
-                }
-              )
-            : undefined,
+          description: params.isProfilingSelected,
```
priscilawebdev, line 50: this will be now 'true' or 'false', do we want that?

priscilawebdev, line 50: I tested this locally and it seems fine, but maybe we want to remove this since we do not have a description anymore?
```suggestion
```

priscilawebdev:
I’m always in favor of simplicity 🙂 The new onboarding document looks great to me, though I can’t confirm if the new installation instructions work and will leave it to the SDK Team. I have only 2x questions: 

- is it sure that users will have the `sentry-debug/` page when they are done installing django? I’ve noticed that in other frameworks, this is sometimes an opt-in feature when using CLI.
- shall we remove the following comment from the code snippet?

Screenshot 2024-07-31 at 11 36 51

smeubank:
@priscilawebdev 
> * is it sure that users will have the `sentry-debug/` page....
they will if they copy paste the verify snippet into their code. So if they don't do that then the docs and onbaording have never made sense, but I think this an OK requirement

```
# urls.py
from django.urls import path

def trigger_error(request):
    division_by_zero = 1 / 0

urlpatterns = [
    path('sentry-debug/', trigger_error),
    # ...
]
```

> * shall we remove the following comment from the code snippet?
This is actually an improvement we could make to a lot of sentry sites. For now let's leave it, but feature request is that code snippets could have file names associated with them. Same as docs

Sentry blogs have the same problem. Scroll to the snippets on this [blog](https://blog.sentry.io/sentry-javascript-v8-sdk-otel-and-node-support/), the file name associated looks terrible, while this is a standard feature on many sites.....actually feature request incoming :D  

So i would merge this PR, and we clean those up when the Sentry onboarding UI has a component feature that allows this.

https://github.com/getsentry/sentry/issues/75445

priscilawebdev (approved):

PR #75238, opened 2024-07-30
ref: incident is nullable

asottile-sentry:
mypy notices the assignment is incompatible once it can follow foreign keys

### DIFF
--- a/src/sentry/monitors/logic/mark_failed.py
+++ b/src/sentry/monitors/logic/mark_failed.py
@@ -140,6 +140,7 @@ def mark_failed_threshold(
 
         starting_checkin = previous_checkins[0]
 
+        incident: MonitorIncident | None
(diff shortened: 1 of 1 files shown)


src/sentry/monitors/logic/mark_failed.py
```diff
@@ -140,6 +140,7 @@ def mark_failed_threshold(
 
         starting_checkin = previous_checkins[0]
 
+        incident: MonitorIncident | None
```
wedamija, line 143: Would it be a bit clearer if this was defined outside the if statement?

wedamija (approved):

src/sentry/monitors/logic/mark_failed.py
asottile-sentry: potentially -- though I put it next to the first assignment so it didn't get potentially orphaned

PR #75248, opened 2024-07-30
fix(similarity): Fix backfill skip project option

jangjodi:
Fix backfill skip project option by properly sending the options to the next backfill calls
Previously would only work on the first backfill batch
Also add options to log message


tests/sentry/tasks/test_backfill_seer_grouping_records.py
```diff
@@ -1036,6 +1059,8 @@ def test_backfill_seer_grouping_records_multiple_batches(
                     "cohort": None,
                     "last_processed_project_index": 0,
                     "only_delete": False,
+                    "skip_processed_projects": True,
+                    "skip_project_ids": None,
```
jangjodi, line 1063: This `test_backfill_seer_grouping_records_multiple_batches` shows we're sending the option between batches now

JoshFerge (approved):

PR #75253, opened 2024-07-30
feat(discover): Get samples works with dataset selector

shruthilayaj:
Only show get samples when transactions dataset is selected
Append `event.type:transaction` to `/dynamic-sampling/custom-rules/` call
if transactions dataset is selected (since event.type:transaction condition
is redundant and we don't expect the user to add it) otherwise, it fails validation.

### DIFF
--- a/static/app/views/discover/table/tableActions.tsx
+++ b/static/app/views/discover/table/tableActions.tsx
@@ -171,21 +171,20 @@ function TableActions(props: Props) {
   const numSamples = tableData?.data?.length ?? null;
   const totalNumSamples = numSamples === null ? null : numSamples + cursorOffset;
 
-  const isNotTransactionsView =
-    hasDatasetSelector(organization) &&
-    queryDataset &&
-    queryDataset !== SavedQueryDatasets.TRANSACTIONS;
+  const isTransactions =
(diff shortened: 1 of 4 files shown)


static/app/views/discover/table/tableActions.tsx
```diff
@@ -159,14 +163,22 @@ function FeatureWrapper(props: FeatureWrapperProps) {
...
+
+  const isNotTransactionsView =
+    hasDatasetSelector(organization) &&
+    queryDataset &&
+    queryDataset !== SavedQueryDatasets.TRANSACTIONS;
+
   return (
     <Fragment>
-      {props.supportsInvestigationRule && (
+      {supportsInvestigationRule && !isNotTransactionsView && (
```
edwardgou-sentry, line 181: nit: Up to you, but I found this is a little confusing to read, because the `!` and variable (`isNotTransactionsView`) forms a double negation. Should we simplify this a bit, or break up the condition a little to be more readable? Maybe something like this:
```
const datasetUnselected = !hasDatasetSelector(organization) || !queryDataset;
const isTransactionsView = queryDataset === SavedQueryDatasets.TRANSACTIONS;
...
{supportsInvestigationRule && (datasetUnselected || isTransactionsView) && (
```

edwardgou-sentry (approved):

matejminar (approved):

PR #75257, opened 2024-07-30
ref(seer grouping): Prepare for calling Seer in optimized grouping flow

lobsterkatie:
This PR makes two changes to prepare for adding a Seer call to the optimized grouping flow in ingest.

- Pull the check for whether to call Seer and the call itself into a new function, `maybe_check_seer_for_matching_grouphash`, so that the same logic can be used in both the current and optimized flows.
- Refactor `get_seer_similar_issues` so that it returns a `GroupHash` object rather than a group, to make it compatible with `handle_existing_grouphash` (which is used in the new flow).

### DIFF
--- a/src/sentry/event_manager.py
+++ b/src/sentry/event_manager.py
@@ -1545,11 +1545,16 @@ def _save_aggregate(
                     )
                     try:
                         # If no matching group is found in Seer, we'll still get back result
-                        # metadata, but `seer_matched_group` will be None
-                        seer_response_data, seer_matched_group = get_seer_similar_issues(
+                        # metadata, but `seer_matched_grouphash` will be None
+                        seer_response_data, seer_matched_grouphash = get_seer_similar_issues(
                             event, primary_hashes
                         )
                         event.data["seer_similarity"] = seer_response_data
+                        seer_matched_group = (
+                            Group.objects.filter(id=seer_matched_grouphash.group_id).first()
+                            if seer_matched_grouphash
+                            else None
+                        )
 
                     # Insurance - in theory we shouldn't ever land here
                     except Exception as e:
--- a/src/sentry/event_manager.py
+++ b/src/sentry/event_manager.py
@@ -71,7 +71,7 @@ from sentry.grouping.ingest.metrics import (
     record_hash_calculation_metrics,
     record_new_group_metrics,
 )
-from sentry.grouping.ingest.seer import get_seer_similar_issues, should_call_seer_for_grouping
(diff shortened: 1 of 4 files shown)


jangjodi (approved):

src/sentry/event_manager.py
```diff
@@ -1725,7 +1702,7 @@ def _save_aggregate_new(
 
     # If we've found one, great. No need to do any more calculations
     if primary.existing_grouphash:
-        group_info = handle_existing_grouphash(
+        group_info = handle_existing_group_or_grouphash(
```
JoshFerge, line 1705: it seems confusing to me that this and line 1716 both pass in `existing_grouphash` -- but this can sometimes be a group? where does that happen?

src/sentry/event_manager.py
```diff
@@ -1820,24 +1797,27 @@ def handle_existing_grouphash(
...
-    #    (otherwise they would not operate on the same rows)
-    # 3) yet somehow also retrieve different groups here
-    #    (otherwise the update would not change anything)
-    #
-    # We think this is a very unlikely situation. A previous version of
-    # _save_aggregate had races around group creation which made this race
-    # more user visible. For more context, see 84c6f75a and d0e22787, as
-    # well as GH-5085.
-    group = Group.objects.get(id=existing_grouphash.group_id)
+    if isinstance(existing_group_or_grouphash, GroupHash):
```
JoshFerge, line 1800: why not lift this logic to callers and instead simply have this method use `Group`? could be a bit more straightforward

src/sentry/event_manager.py
lobsterkatie: It doesn't yet - this is prep for the https://github.com/getsentry/sentry/pull/75266, where it does.

JoshFerge (approved):

src/sentry/event_manager.py
lobsterkatie: I could. Three of the four times it's used, it's used on grouphashes rather than groups, though. Perhaps better would be to change the one spot where it gets passed a group to pass it a grouphash instead?

JoshFerge: yeah that would also make sense! but i'll defer to you on how you'd like to implement

lobsterkatie: I ended up going that route. I've updated the PR description.

src/sentry/event_manager.py
lobsterkatie: This is now moot, given the change discussed in the convo below.

PR #75265, opened 2024-07-30
feat(relocation): Forbid forking from `de` region

azaslavsky:
This is mostly a legal safeguard: we don't want to allow easy export out of DE, and it's hard to imagine a (legal) case where this would be allowed anyway. If such a case ever does come up, we can easily modify the `CANNOT_FORK_FROM_REGION` set.

### DIFF
--- a/src/sentry/api/endpoints/organization_fork.py
+++ b/src/sentry/api/endpoints/organization_fork.py
@@ -34,6 +34,12 @@ ERR_ORGANIZATION_INACTIVE = Template(
 ERR_CANNOT_FORK_INTO_SAME_REGION = Template(
     "The organization already lives in region `$region`, so it cannot be forked into that region."
 )
+ERR_CANNOT_FORK_FROM_REGION = Template(
+    "Forking an organization from region `$region` is forbidden."
+)
+
+# For legal reasons, there are certain regions from which forking is disallowed.
+CANNOT_FORK_FROM_REGION = {"de"}
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


src/sentry/relocation/services/relocation_export/impl.py
```diff
@@ -181,7 +181,7 @@ def reply_with_export(
             "replying_region_name": replying_region_name,
             "org_slug": org_slug,
             # TODO(azaslavsky): finish transfer from `encrypted_contents` -> `encrypted_bytes`.
-            "encrypt_with_public_key_size": len(encrypted_bytes or []),
+            "encrypted_bytes_size": len(encrypted_bytes or []),
```
azaslavsky, line 184: Fixes a typo from a previous PR.

src/sentry/api/endpoints/organization_fork.py
```diff
@@ -34,6 +34,12 @@
 ERR_CANNOT_FORK_INTO_SAME_REGION = Template(
     "The organization already lives in region `$region`, so it cannot be forked into that region."
 )
+ERR_CANNOT_FORK_FROM_REGION = Template(
+    "Forking an organization from region `$region` is forbidden."
+)
+
+# For legal reasons, there are certain regions from which forking is disallowed.
+CANNOT_FORK_FROM_REGION = {"de"}
```
markstory, line 42: So once you're in de you can never relocate somewhere else?

azaslavsky: Yes. IANAL, but my understanding is that's what we want?

markstory: Ok. That could make for some tricky support issues, but we'll figure that out later.

markstory (approved):

Issue #74312, opened 2024-07-15
SearchQueryBuilder: Support aggregate filters


malwilley:
See functions defined in the docs: https://docs.sentry.io/concepts/search/searchable-properties/events/

Functions also [defined in code here](https://github.com/getsentry/sentry/blob/6add1d3dc8f0e2d7af211f936071c611290ce2d0/src/sentry/search/events/fields.py#L1511).

![CleanShot 2024-07-16 at 14 16 31](https://github.com/user-attachments/assets/5b317b13-441e-45f4-856e-863964114647)

- `avg(field)` - accepts any numeric or duration field, returns the same type
- `count_if(column,operator,value)` - accepts any column, op, value and returns an integer
  - Example: `count_if(transaction.duration,equals,300):>10`
- `apdex(threshold)` - accepts a duration number (in ms) that is used as the apdex threshold. Returns a number between 0 and 1.
- `count_unique(field)` - Accepts any field and returns an integer
- `count_web_vitals(vital, threshold)` - Accepts a web vital and a categorical threshold. Returns an integer.
  - Example: `count_web_vitals(measurements.lcp, good):>10`
- `epm()` - Events per minute, accepts no parameters
- `eps()` - Events per second, accepts no parameters
- `count_miserable(field, threshold)` - Count of miserable users given a threshold (despite accepting an arbitrary field, user is the only acceptable one)
  - Example: `count_miserable(user, 300)>10`

PR #75267, opened 2024-07-30
feat(query-builder): Add ability to add a basic aggregate filter

malwilley:
Ref https://github.com/getsentry/sentry/issues/74312

This PR adds the building blocks of aggregate filters. More functionality will be added in subsequent PRs.

- Adds `parameters` type to the field definition to allow users to define what is accepted
- Adds default parameters when an aggregate filter is selected from the key menu
- Separates the key from the operator for aggregate filters (because clicking the key will allow you to edit the parameters)

CleanShot 2024-07-30 at 09 32 06

### DIFF
--- a/static/app/components/searchQueryBuilder/hooks/useHandleSearch.tsx
+++ b/static/app/components/searchQueryBuilder/hooks/useHandleSearch.tsx
@@ -8,6 +8,7 @@ import {
   tokenIsInvalid,
 } from 'sentry/components/searchQueryBuilder/utils';
 import {type ParseResult, Token} from 'sentry/components/searchSyntax/parser';
+import {getKeyName} from 'sentry/components/searchSyntax/utils';
 import type {SavedSearchType} from 'sentry/types';
 import type {Organization} from 'sentry/types/organization';
 import {trackAnalytics} from 'sentry/utils/analytics';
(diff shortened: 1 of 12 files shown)


roggenkemper (commented):
A few miscellaneous things i noticed:
1 - the operators for count/count_if are a mix of "is"/"is not" and math symbols which is slightly confusing? 
2- clicking the value for count + count_if opens a dropdown with a spinning loader which then goes away
3- something about having `count()` seems a little weird to me? i know it's a function and maybe it is because the rest of the search experience is much more readable but based of of everything else i would expect this to be `count is 100` 
4 - i can't reproduce it but i saw twice while playing around that adding a new filter re-opened the dropdown for a previous filter 
5- `UnstyledButton` is used quite a bit - would it make sense to just make that it's own component so we don't have to keep redefining it

malwilley:
@roggenkemper 

> 1 - the operators for count/count_if are a mix of "is"/"is not" and math symbols which is slightly confusing?

Fair point, that's how it works for other numeric filters too. I think it makes sense to change "is" to "=" for those (will do in a separate PR)

> 2- clicking the value for count + count_if opens a dropdown with a spinning loader which then goes away
 
Good catch, we were looking at `count_if(...)` instead of `count_if`. Using `getKeyName` everywhere now so that should be fixed.

3- something about having count() seems a little weird to me? i know it's a function and maybe it is because the rest of the search experience is much more readable but based of of everything else i would expect this to be count is 100

Yeah it's not the most intuitive. The functions are only applicable when you have grouped results, like in discover, which is why they are special. Definitely more we can do to make that self evident

> 5- UnstyledButton is used quite a bit - would it make sense to just make that it's own component so we don't have to keep redefining it

Good call, will extract that

roggenkemper (approved):
looks good!

PR #75270, opened 2024-07-30
feat(query-builder): Add basic parameter editing for aggregate keys

malwilley:
Ref: https://github.com/getsentry/sentry/issues/74312

Depends on https://github.com/getsentry/sentry/pull/75267

This PR adds the ability to edit the function parameters (with no validation or suggestions).

https://github.com/user-attachments/assets/01b19c36-832a-4671-8e64-7eb0071b55da


roggenkemper (approved):
the parameter is pretty hard to read (took me a bit to notice that it was actually 3 comma separated values). not a blocker for this PR but i wonder if the legibility can be improved

PR #75271, opened 2024-07-30
chore(ops): add backfill url

JoshFerge:
use the backfill url for bulk grouping requests!


src/sentry/conf/server.py
```diff
@@ -3136,6 +3136,9 @@ def custom_parameter_sort(parameter: dict) -> tuple[str, int]:
 SEER_GROUPING_URL = SEER_DEFAULT_URL  # for local development, these share a URL
 SEER_GROUPING_TIMEOUT = 1
 
+SEER_GROUPING_BACKFILL_URL = SEER_DEFAULT_URL  # for local development, these share a URL
+SEER_GROUPING_BACKFILL_TIMEOUT = 1
```
snigdhas, line 3140: Should this be used in the request below?

snigdhas (approved):

src/sentry/conf/server.py
JoshFerge: I think its just a default, and we override it. maybe we don't need it

PR #75275, opened 2024-07-30
feat(onboarding) add profiling to onboarding docs

JonasBa:
Add document header onboarding step and enable profiling onboarding for js projects

### DIFF
--- a/static/app/components/onboarding/gettingStartedDoc/utils/profilingOnboarding.tsx
+++ b/static/app/components/onboarding/gettingStartedDoc/utils/profilingOnboarding.tsx
@@ -1,0 +1,12 @@
+import type {StepProps} from 'sentry/components/onboarding/gettingStartedDoc/step';
+
+export function getProfilingDocumentHeaderConfigurationStep(): StepProps {
+  return {
+    title: 'Add Document-Policy: js-profiling header',
+    description: `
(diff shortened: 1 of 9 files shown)


static/app/components/onboarding/gettingStartedDoc/utils/profilingOnboarding.tsx
```diff
@@ -0,0 +1,12 @@
+import type {StepProps} from 'sentry/components/onboarding/gettingStartedDoc/step';
+
+export function getProfilingDocumentHeaderConfigurationStep(): StepProps {
+  return {
+    title: 'Add Document-Policy: js-profiling header',
+    description: `
```
Zylphrex, line 6: should `Document-Policy: js-profiling` be formatted as a code block for better legibility?

Zylphrex (approved):
Just a comment around formatting for the header but otherwise LGTM

static/app/components/onboarding/gettingStartedDoc/utils/profilingOnboarding.tsx
JonasBa: Hmm, I'm not sure, but I think it would look weird if we had code in the title. Let me see if there is some better formatting here

priscilawebdev:
I would have loved to have had the chance to review this PR 🙂

PR #75276, opened 2024-07-30
fix(issue-search): Fix bug when searching for non-error issues

snigdhas:
There was a bug in the logic to search for group categories with a negation, specifically searches like `!issue.category:error`. This PR fixes the logic to filter out the negated issue categories and adds a test to validate that negation searches work as expected.

### DIFF
--- a/src/sentry/issues/search.py
+++ b/src/sentry/issues/search.py
@@ -81,14 +81,11 @@ def group_categories_from(
     for search_filter in search_filters or ():
         if search_filter.key.name in ("issue.category", "issue.type"):
             if search_filter.is_negation:
+                # get all group categories except the ones in the negation filter
                 group_categories.update(
                     get_group_type_by_type_id(value).category
-                    for value in list(
-                        filter(
-                            lambda x: x not in get_all_group_type_ids(),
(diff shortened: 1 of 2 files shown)


roggenkemper (approved):

MichaelSun48 (approved):

src/sentry/issues/search.py
```diff
@@ -81,14 +81,11 @@ def group_categories_from(
...
                 group_categories.update(
                     get_group_type_by_type_id(value).category
-                    for value in list(
-                        filter(
-                            lambda x: x not in get_all_group_type_ids(),
-                            search_filter.value.raw_value,
-                        )
-                    )
+                    for value in get_all_group_type_ids()
+                    if value not in search_filter.value.raw_value
```
malwilley, line 88: Will this work for a query like `!issue.category:[performance,cron]`?

snigdhas: yup, added a test

malwilley: Thanks for confirming!

malwilley (approved):

PR #75277, opened 2024-07-30
chore(azure): check if repo config instance differs from integration domain name

cathteng:
The `get_client` function signature for Azure DevOps differs from all the other integrations because we are passing an optional `base_url` parameter. We might not need to do this if the repository `instance` is always the same as the integration `domain_name` (both are passed as base urls). According to the code, the repository instance should always be the same as the domain name (aka `installation.instance` in the code).

https://github.com/getsentry/sentry/blob/3ba7201f3ec899add45e3478a98d93b88d1447e9/src/sentry/integrations/vsts/repository.py#L26-L37

https://github.com/getsentry/sentry/blob/3ba7201f3ec899add45e3478a98d93b88d1447e9/src/sentry/integrations/vsts/integration.py#L366-L368

This PR adds logging to check if we can remove the `base_url` parameter or if we should add an overloaded function declaration in the `IntegrationInstallation` base class. In `VstsIntegration`, we only ever pass `base_url=self.instance` to `get_client` which is already the default if we don't pass `base_url`.

### DIFF
--- a/src/sentry/integrations/vsts/repository.py
+++ b/src/sentry/integrations/vsts/repository.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 from collections.abc import Mapping, MutableMapping, Sequence
 from typing import Any
 
@@ -10,6 +11,8 @@ from sentry.organizations.services.organization.model import RpcOrganization
 from sentry.plugins.providers import IntegrationRepositoryProvider
 
 MAX_COMMIT_DATA_REQUESTS = 90
+
+logger = logging.getLogger(__name__)
 
 
 class VstsRepositoryProvider(IntegrationRepositoryProvider):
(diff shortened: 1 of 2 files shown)


src/sentry/integrations/vsts/repository.py
```diff
@@ -73,7 +76,18 @@ def zip_commit_data(
         self, repo: Repository, commit_list: Sequence[Commit], organization_id: int
     ) -> Sequence[Commit]:
         installation = self.get_installation(repo.integration_id, organization_id)
-        client = installation.get_client(base_url=repo.config["instance"])
+        instance = repo.config["instance"]
+        if installation.instance != instance:
```
sentaur-athena, line 80: what type object is instance?

cathteng: it's a string that is a url

iamrajjoshi (approved):
lgtm, just wondering if there are tests for the functions where you removed the base_url from the get_client call

GabeVillalobos (approved):
LGTM!

PR #75281, opened 2024-07-30
ref: fix nullability of Relocation.want_usernames in sentry.tasks.relocation

asottile-sentry:
discovered after mypy follows foreign keys and understands nullable foreign keys

### DIFF
--- a/src/sentry/tasks/relocation.py
+++ b/src/sentry/tasks/relocation.py
@@ -242,8 +242,6 @@ def uploading_start(uuid: UUID, replying_region_name: str | None, org_slug: str 
         sequence (`uploading_complete`) is scheduled.
     """
 
-    relocation: Relocation | None
-    attempts_left: int
     (relocation, attempts_left) = start_relocation_task(
         uuid=uuid,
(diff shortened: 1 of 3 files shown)


joshuarli (approved):

markstory (approved):
Looks good to me.

src/sentry/tasks/relocation.py
```diff
@@ -425,7 +423,7 @@ def cross_region_export_timeout_check(
     """
 
     try:
-        relocation: Relocation = Relocation.objects.get(uuid=uuid)
+        relocation = Relocation.objects.get(uuid=uuid)
```
azaslavsky, line 426: Just to be clear: `objects.get()` now returns a typed value as far as mypy is concerned? Or are we removing the case here for other reasons?

azaslavsky (approved):

src/sentry/tasks/relocation.py
asottile-sentry: from the date this code was written it always has -- this annotation was not needed

PR #75292, opened 2024-07-30
ref: handle nullability of AuditLogEntry.target_user

asottile-sentry:
discovered after mypy follows foreign keys and understands nullable foreign keys


joshuarli (approved):
nice!

PR #75293, opened 2024-07-30
feat(custom-views): Add delete, duplicate, add view, save, and temp menu functionality

### LINKED ISSUE #73225: Custom Views:

### LINKED ISSUE #73228: Custom Views: Add delete functionality to tab menu
Once the tab menu and options have been created, we will need to implement functionality for the delete option. When clicked, the tab should be deleted (possibly after a confirmation modal?). ~The request to update our backend should trigger the debounce trigger when the button is pressed.~ Making the request to our backend is no longer part of this issue - connecting the backend to the frontend will be its own ticket. 

This ticket can be marked as completed once the delete option in the tab menu works as described above.

**Questions:**
* Should we have a confirmation modal for this? 

**Direct dependencies:**

* #73224

### LINKED ISSUE #73229: Custom Views: Add Save Changes and Discard Changes functionality to tabs menu
Once the tab menu and options have been created, we will need to implement functionality for the `Save Changes` and `Discard Changes` options. When clicked, the changes to the tab (query or query sort) should be saved or discarded, respectively. ~The request made to our backend to save changes will need to trigger the debounce timer.~ Making the request to our backend is no longer part of this issue - connecting the backend to the frontend will be its own ticket. 

This ticket can be marked as completed once the `Save Changes` and `Discard Changes` option in the tab menu works as described above.

**Direct dependencies:**

* #73224

### LINKED ISSUE #73339: Custom Views: Add temporary view functionality
Custom views will need to support temporary views as part of its feature set. A temporary view appears as a tab after the `Add View` button, and do not persist permanently like normal views. They are opened when a user clicks on a shared link. 

This ticket can be marked as completed once all of the functionality for temporary views has been completed. This includes the implementing the `Save View` and `Discard` options in the dropdown menu, and modifying the frontend state logic to reflect when these options have been selected. 

Direct Dependencies:
* #73338

MichaelSun48:
closes #73339
closes #73225
closes #73229
closes #73228

This PR implements default functionality for the save changes, discard changes, duplicate, delete, add view, and discard/save temp view functionality. This does **not** include the calls made to the api, nor the debouncing logic that is planned to be included. 

Rename is the only option that has not been implemented in this PR, since that one is the most complex and requires design work. It will be coming in the subsequent PR.

### DIFF
--- a/static/app/views/issueList/draggableTabBar.tsx
+++ b/static/app/views/issueList/draggableTabBar.tsx
@@ -17,14 +17,18 @@ import {DraggableTabMenuButton} from 'sentry/views/issueList/draggableTabMenuBut
 
 export interface Tab {
   content: React.ReactNode;
-  key: Key;
+  key: string;
   label: string;
   hasUnsavedChanges?: boolean;
   queryCount?: number;
 }
 
 export interface DraggableTabBarProps {
+  defaultNewTab: Tab;
+  setTabs: (tabs: Tab[]) => void;
+  showTempTab: boolean;
   tabs: Tab[];
+  tempTab: Tab;
   onAddView?: React.MouseEventHandler;
   /**
    * Callback function to be called when user clicks the `Delete` button.
@@ -60,15 +64,16 @@ export interface DraggableTabBarProps {
    * Callback function to be called when user clicks the 'Save View' button for temporary views.
    */
   onSaveTempView?: () => void;
-  showTempTab?: boolean;
   tempTabContent?: React.ReactNode;
   tempTabLabel?: string;
 }
 
 export function DraggableTabBar({
-  showTempTab = false,
-  tempTabContent,
-  tempTabLabel = 'Unsaved',
+  tabs,
+  setTabs,
+  tempTab,
+  defaultNewTab,
+  showTempTab,
   onAddView,
   onDelete,
   onDiscard,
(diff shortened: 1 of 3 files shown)


static/app/views/issueList/draggableTabBar.tsx
```diff
@@ -130,18 +161,18 @@ export function DraggableTabBar({
         onSelectionChange={setSelectedTabKey}
         selectedKey={selectedTabKey}
         showTempTab={showTempTab}
-        onAddView={onAddView}
+        onAddView={e => handleOnAddView(e)}
         orientation="horizontal"
       >
-        {tabs.map(tab => (
+        {[...tabs, tempTab].map(tab => (
```
MichaelSun48, line 167: This might be a little controversial - I was deciding between two options to handle temporary tabs:
1. Pass in the temp tab as its own prop. It is not included in the top level tabs state, but it is included in the aria collection so it still has accessibly features. (The current implementation)
2. Have the temporary tab be passed in with the `tabs` prop, and make a note that the temp tab must have `key = 'temporary-tab'`.

Number 1 seemed to be more intuitive and less error prone than number 2, so that's what I went with. But pls lmk if there might be a better design than this.

static/app/views/issueList/draggableTabBar.tsx
```diff
@@ -177,13 +216,16 @@ const makeDefaultMenuOptions = ({onRename, onDuplicate, onDelete}): MenuItemProp
       label: t('Duplicate'),
       onAction: onDuplicate,
     },
-    {
+  ];
+  if (onDelete) {
```
MichaelSun48, line 220: Delete option now only shows up if there are more than 2 tabs to prevent a user from deleting all tabs. The backend also enforces this, but we should prevent this on the frontend as well.

static/app/views/issueList/draggableTabBar.tsx
malwilley: I agree that separating the temp tab from the list of saved views makes sense and is probably the better way to go here 👍. Conceptually it is very different and has different behavior.

malwilley (approved):

PR #75294, opened 2024-07-30
feat(autofix): Switch root cause to show relevant code context for issue instead of fixes

roaga:
Backend PR: https://github.com/getsentry/seer/pull/979 

This reduces redundancy between the root cause and planning steps and helps the user understand and choose a root cause more confidently.

This PR contains the frontend changes to match the backend changes in Seer.
Screenshot 2024-07-31 at 12 43 50 PM


static/app/components/codeSnippet.tsx
```diff
@@ -80,6 +82,11 @@ export function CodeSnippet({
 }: CodeSnippetProps) {
   const ref = useRef<HTMLModElement | null>(null);
 
+  // https://prismjs.com/plugins/line-highlight/
+  useEffect(() => {
+    import('prismjs/plugins/line-highlight/prism-line-highlight');
```
malwilley, line 87: You could add a `if (linesToHeightlight)` condition to avoid importing if we aren't using the feature.

static/app/components/codeSnippet.tsx
```diff
@@ -36,6 +36,7 @@ interface CodeSnippetProps {
    */
   isRounded?: boolean;
   language?: string;
+  linesToHighlight?: number[];
```
malwilley, line 39: Since this is a highly used component, could you do a couple extra things:

1. Add a comment here explaining what the prop does
2. Add section to the `codeSnippet.stories.tsx` file that demos the functionality. You can view the story book at `/stories`

roaga: Done in next commit

static/app/components/codeSnippet.tsx
roaga: Done in next commit

static/app/components/events/autofix/autofixRootCause.tsx
```diff
@@ -123,6 +123,50 @@ function useSelectCause({groupId, runId}: {groupId: string; runId: string}) {
...
+        }
+      }
+    });
+
+    return result;
+  }
+
+  const lineNumbersToHighlight = findLinesWithSubstrings(
+    suggestedFix.snippet?.snippet,
+    getBacktickSubstrings(suggestedFix.description)
```
jennmueng, line 165: UX wise, would it make more sense to highlight a specific line that is causing the issue rather than highlighting the lines from the description's backticks?

roaga: Done in next commit

static/app/components/events/autofix/autofixRootCause.tsx
```diff
@@ -224,41 +215,45 @@ function CauseOption({
...
           }}
         />
-        <Button
-          icon={<IconChevron size="xs" direction={selected ? 'down' : 'right'} />}
-          aria-label={t('Select root cause')}
-          aria-expanded={selected}
-          size="zero"
-          borderless
-        />
+        <div style={{display: 'flex', flexDirection: 'row'}}>
```
malwilley, line 218: Could you apply these styles with a styled component instead? The `style` attribute cannot be overridden so we generally avoid using it for static styles.

malwilley (approved):
Otherwise looks good!

static/app/components/events/autofix/autofixRootCause.tsx
roaga: Done!

malwilley (approved):

PR #75300, opened 2024-07-30
chore(types) Improve types for File classes

markstory:
The File classes were relying on unsound types. Extract the varying parts of the logic into abstract methods and use a Generic to tie relations together.

Django doesn't allow `Generic` to be a super class for ORM records, so I've used a `TYPE_CHECKING` guard. I also attempted a `Protocol` based solution but ran into compatibility issues with how it overlapped with Django.


asottile-sentry (approved):
very nice!

PR #75302, opened 2024-07-30
Process delayed alert conditions in batches of 10,000

saponifi3d:
# Description
Some orgs are sending 100k+ events per minute, and the processing is taking to long for a single task.

This PR will look at the size of the hash and determine if it needs to be batched.

There's some restrictions around the celery task / redis, info is outlined in a code comment here: https://github.com/getsentry/sentry/pull/75302/files#diff-f906e75a0e4419db4870fa45ca5a1608ca79beaa052c8bc50b4805607a665d27R482-R486

### DIFF
--- a/src/sentry/buffer/redis.py
+++ b/src/sentry/buffer/redis.py
@@ -81,6 +81,7 @@ class BufferHookRegistry:
 redis_buffer_registry = BufferHookRegistry()
 
 
+# Note HMSET is not supported after redis 4.0.0, after updating we can use HSET directly.
 class RedisOperation(Enum):
     SORTED_SET_ADD = "zadd"
     SORTED_SET_GET_RANGE = "zrangebyscore"
--- a/src/sentry/buffer/redis.py
+++ b/src/sentry/buffer/redis.py
@@ -88,6 +88,7 @@ class RedisOperation(Enum):
     HASH_ADD = "hset"
     HASH_GET_ALL = "hgetall"
     HASH_DELETE = "hdel"
+    HASH_LENGTH = "hlen"
 
 
 class PendingBuffer:
@@ -310,6 +311,12 @@ class RedisBuffer(Buffer):
             decoded_hash[k] = v
 
         return decoded_hash
+
+    def get_hash_length(
+        self, model: type[models.Model], field: dict[str, models.Model | str | int]
+    ) -> int:
+        key = self._make_key(model, field)
+        return self._execute_redis_operation(key, RedisOperation.HASH_LENGTH)
 
     def process_batch(self) -> None:
         try:
--- a/src/sentry/buffer/redis.py
+++ b/src/sentry/buffer/redis.py
@@ -86,6 +86,7 @@ class RedisOperation(Enum):
     SORTED_SET_GET_RANGE = "zrangebyscore"
     SORTED_SET_DELETE_RANGE = "zremrangebyscore"
     HASH_ADD = "hset"
+    HASH_ADD_BULK = "hmset"
     HASH_GET_ALL = "hgetall"
     HASH_DELETE = "hdel"
     HASH_LENGTH = "hlen"
@@ -296,6 +297,15 @@ class RedisBuffer(Buffer):
     ) -> None:
         key = self._make_key(model, filters)
         self._execute_redis_operation(key, RedisOperation.HASH_ADD, field, value)
+
+    def push_to_hash_bulk(
+        self,
+        model: type[models.Model],
+        filters: dict[str, models.Model | str | int],
+        data: dict[str, str],
+    ) -> None:
+        key = self._make_key(model, filters)
+        self._execute_redis_operation(key, RedisOperation.HASH_ADD_BULK, data)
 
     def get_hash(
         self, model: type[models.Model], field: dict[str, models.Model | str | int]
--- a/src/sentry/options/defaults.py
+++ b/src/sentry/options/defaults.py
@@ -2636,3 +2636,7 @@ register(
     default=1,
     flags=FLAG_AUTOMATOR_MODIFIABLE,
 )
+register(
+    "delayed_processing.batch_size",
+    default=10000,
+)
--- a/src/sentry/options/defaults.py
+++ b/src/sentry/options/defaults.py
@@ -2639,4 +2639,5 @@ register(
 register(
     "delayed_processing.batch_size",
     default=10000,
+    flags=FLAG_AUTOMATOR_MODIFIABLE,
 )
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -481,8 +481,9 @@ def process_rulegroups_in_batches(project_id: int):
     The batches are replicated into a new redis hash with a unique filter (a uuid) to identify the batch.
     We need to use a UUID because these batches can be created in multiple processes and we need to ensure
     uniqueness across all of them for the centralized redis buffer. The batches are stored in redis because
-    we shouldn't pass complex objects in the celery task arguments, and we can't send a page of data in the
-    batch becaues redis may not maintain the sort of the hash response.
+    we shouldn't pass objects that need to be pickled and 10k items could be problematic in the celery tasks
+    as arguments could be problematic. Finally, we can't use a pagination system on the data because
+    redis doesn't maintain the sort order of the hash keys.
 
     `apply_delayed` will fetch the batch from redis and process the rules.
     """
--- a/src/sentry/rules/processing/delayed_processing.py
+++ b/src/sentry/rules/processing/delayed_processing.py
@@ -466,41 +466,38 @@ def bucket_num_groups(num_groups: int) -> str:
 
 
 def process_rulegroups_in_batches(project_id: int, batch_size=CHUNK_BATCH_SIZE):
-    # TODO REMOVE THIS
-    rulegroup_to_event_data = fetch_rulegroup_to_event_data(project_id)
-
     event_count = buffer.backend.get_hash_length(Project, {"project_id": project_id})
 
-    print("REHYDRAAATTE", batch_size, event_count, project_id, rulegroup_to_event_data)
     if event_count < batch_size:
-        print("SINGLE EXECUTE!")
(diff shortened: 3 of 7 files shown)


src/sentry/rules/processing/delayed_processing.py
```diff
@@ -463,6 +464,39 @@ def bucket_num_groups(num_groups: int) -> str:
...
+
+    with metrics.timer("delayed_processing.process_batch.duration"):
+        items = iter(rulegroup_to_event_data.items())
+        while True:
+            batch = dict(islice(items, batch_size))
+            if not batch:
+                break
+
+            batch_key = f"{project_id}:{uuid.uuid4()}"
+            batch_values = json.dumps(batch)
```
saponifi3d, line 487: this is likely to not work.. more likely i'll need to iterate each item in the batch, adding each value to the new redis store, and removing it from the old. :sob:

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -463,6 +473,55 @@ def bucket_num_groups(num_groups: int) -> str:
...
+    rulegroup_to_event_data = fetch_rulegroup_to_event_data(project_id)
+
+    with metrics.timer("delayed_processing.process_batch.duration"):
+        items = iter(rulegroup_to_event_data.items())
+
+        while batch := dict(islice(items, batch_size)):
+            batch_key = str(uuid.uuid4())
+
+            for field, value in batch.items():
+                buffer.backend.push_to_hash(
```
saponifi3d, line 510: looking at a way to do this in bulk rn - i am prreeettty sure that `redis.hset` supports this (since we're just setting a key / value on the hash). 🔬

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -463,6 +473,53 @@ def bucket_num_groups(num_groups: int) -> str:
...
+def process_rulegroups_in_batches(project_id: int, batch_size=CHUNK_BATCH_SIZE):
+    """
+    This will check the number of rulegroup_to_event_data items in the Redis buffer for a project.
+
+    If the number is larger than the batch size, it will chunk the items and process them in batches.
+
+    The batches are replicated into a new redis hash with a unique filter (a uuid) to identify the batch.
+    We need to use a UUID because these batches can be created in multiple processes and we need to ensure
+    uniqueness across all of them for the centralized redis buffer. The batches are stored in redis because
+    we shouldn't pass complex objects in the celery task arguments, and we can't send a page of data in the
```
wedamija, line 485: Jfyi: Mostly we don't like to pass things that would have to be pickled, like Django models etc. That said, this approach is probably better because I'm not sure if there are downsides to storing 10k ints per task in rabbitmq

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -45,6 +47,7 @@
 logger = logging.getLogger("sentry.rules.delayed_processing")
 EVENT_LIMIT = 100
 COMPARISON_INTERVALS_VALUES = {k: v[1] for k, v in COMPARISON_INTERVALS.items()}
+CHUNK_BATCH_SIZE = 10000
```
wedamija, line 50: It could be interesting to push this higher, or even make it configurable via flagpole (that would also allow us to customize it per project if that was every necessary)

wedamija (approved):
Nice fix, I was concerned that this batching might end up pretty complex, but this is easy to reason about

src/sentry/rules/processing/delayed_processing.py
saponifi3d: 👍 makes sense to set via a configuration!

any chance you know of an example of setting a project scope'd int to do that? (i was reading the docs and got a little 😵, seems like it's something like...
```
default_manager.add('projects:delayed-processing-chunk-size', ProjectFeature, FeatureHandlerStrategy.FLAGPOLE)
```

then to set the value i add it to [flagpole.yml](https://github.com/getsentry/sentry-options-automator/blob/main/options/default/flagpole.yml#L735) or is that just to define the conditions when to change the value?

wedamija: Hmm... actually I'm not sure how you'd do this via flagpole right now, I guess it'd just be a raw options automator thing? Since flagpole is just for verifying whether a feature is enabled.

Not a huge deal either way, although maybe we should just make this configurable via options automator with a default, and not worry about being able to customize per project.

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -47,7 +47,7 @@
 logger = logging.getLogger("sentry.rules.delayed_processing")
 EVENT_LIMIT = 100
 COMPARISON_INTERVALS_VALUES = {k: v[1] for k, v in COMPARISON_INTERVALS.items()}
-CHUNK_BATCH_SIZE = 10000
+CHUNK_BATCH_SIZE = options.get("delayed_processing.batch_size")
```
wedamija, line 50: We can't set this as a module variable - it'll never update (and might cause problems on load). We just need to perform the check in function

src/sentry/buffer/redis.py
```diff
@@ -86,8 +86,10 @@ class RedisOperation(Enum):
     SORTED_SET_GET_RANGE = "zrangebyscore"
     SORTED_SET_DELETE_RANGE = "zremrangebyscore"
     HASH_ADD = "hset"
+    HASH_ADD_BULK = "hmset"
```
schew2381, line 89: According to https://redis.io/docs/latest/commands/hmset/, this command is deprecated after 4.0 and can be replaced with `hset` which can take multiple key value pairs.
I'm not what version of redis we use tho

src/sentry/rules/processing/delayed_processing.py
saponifi3d: heh, ty, i was just about to ask to check if did that right. thanks!

src/sentry/rules/processing/delayed_processing.py
schew2381: +1 to raw option :)

src/sentry/rules/processing/delayed_processing.py
```diff
@@ -85,8 +88,15 @@ def fetch_project(project_id: int) -> Project | None:
         return None
 
 
-def fetch_rulegroup_to_event_data(project_id: int) -> dict[str, str]:
-    return buffer.backend.get_hash(model=Project, field={"project_id": project_id})
+def fetch_rulegroup_to_event_data(project_id: int, batch_key: str | None = None) -> dict[str, str]:
+    field: dict[str, models.Model | int | str] = {
```
schew2381, line 91: Where does this models.Model typing come from on the key 🤔

src/sentry/rules/processing/delayed_processing.py
saponifi3d: okay, just updated again, anything else i'm missing with the options? (I was looking at the docs here: https://develop.sentry.dev/backend/options/ to add it). ps, thanks for the help!

src/sentry/options/defaults.py
```diff
@@ -2636,3 +2636,7 @@
     default=1,
     flags=FLAG_AUTOMATOR_MODIFIABLE,
 )
+register(
+    "delayed_processing.batch_size",
+    default=10000,
+)
```
wedamija, line 2642: You just need to add `flags=FLAG_AUTOMATOR_MODIFIABLE` and then you'll be good to go

schew2381 (commented):
One thing is in `appled_delayed`, I think we need to edit `cleanup_redis_buffer` b/c it uses the `project_id` when deleting the hash values rather than the generated `uuid`

wedamija (approved):

src/sentry/buffer/redis.py
saponifi3d: yeah, i had initially tried the `hset` with `mapping` syntax and it threw errors, so we'll need to use hmset until then.

schew2381: ah ok, could we add a comment somewhere then referencing the future deprecation?

src/sentry/rules/processing/delayed_processing.py
saponifi3d: It's like a weird nested definition on the function call to `get_hash` for the field variable.

ideally i could _not_ type this at all, and mypy would evaluate the field data type as `dict[str, int | str]` which adheres to `dict[str, models.Model | int | str]` (at least, that's how typescript works). unfortunately, it was throwing errors and if i typed it explicitly it threw the same errors.. since the dict _does_ adhere to the type definition with models.Model, i just added it to appease the `mypy` overlords.

any recommendations on cleanup or better ways to appease the mypy?

saponifi3d:
> One thing is in appled_delayed, I think we need to edit cleanup_redis_buffer b/c it uses the project_id when deleting the hash values rather than the generated uuid

@schew2381 🙏 💯 thanks for catching that. i'll update to pass the batch_key as well!

src/sentry/rules/processing/delayed_processing.py
schew2381: that makes sense, I would think the implicit typing would work but I guess not lol
mypy is just a mystery sometimes 🔍

tests/sentry/rules/processing/test_delayed_processing.py
```diff
@@ -1326,6 +1332,66 @@ def test_apply_delayed_process_count_then_percent(self, safe_execute_callthrough
...
+        self.group_three = self.create_group(self.project)
+        self.rule = self.create_alert_rule()
+
+    @patch("sentry.rules.processing.delayed_processing.apply_delayed")
+    def test_no_redis_data(self, mock_apply_delayed):
+        mock_delayed = Mock()
+        mock_apply_delayed.delayed = mock_delayed
+        process_rulegroups_in_batches(self.project.id)
+
+        mock_delayed.assert_called_once_with(self.project.id)
```
schew2381, line 1351: nit: I think any calls on mocks generate more mocks, so you can do:
```suggestion
        process_rulegroups_in_batches(self.project.id)

        mock_apply_delayed.delayed.assert_called_once_with(self.project.id)
```

schew2381 (approved):
overall lgtm!

PR #75303, opened 2024-07-30
ref: use a TypedDict to represent filter params

asottile-sentry:


src/sentry/search/events/filter.py
```diff
@@ -3,7 +3,7 @@
 from collections.abc import Callable, Mapping, Sequence
```
asottile-sentry, line 3: this file is ignored which is why this wasn't caught before -- will follow up to hopefully get this file unignored

snigdhas (approved):

PR #75306, opened 2024-07-30
feat(query-builder): Add aggregate parameter suggestions

malwilley:
Ref https://github.com/getsentry/sentry/issues/74312

Adds function parameter suggestions for column and predefined value types.

https://github.com/user-attachments/assets/0e25e2ae-9429-4fea-871b-039e636ac02e

### DIFF
--- a/static/app/components/searchQueryBuilder/tokens/filter/aggregateKey.tsx
+++ b/static/app/components/searchQueryBuilder/tokens/filter/aggregateKey.tsx
@@ -1,4 +1,4 @@
-import {useRef, useState} from 'react';
+import {useLayoutEffect, useRef, useState} from 'react';
 import styled from '@emotion/styled';
 import {useFocusWithin} from '@react-aria/interactions';
 import {mergeProps} from '@react-aria/utils';
@@ -34,9 +34,21 @@ export function AggregateKey({
   filterRef,
 }: AggregateKeyProps) {
   const ref = useRef<HTMLDivElement>(null);
-  const {disabled} = useSearchQueryBuilder();
+  const {dispatch, focusOverride, disabled} = useSearchQueryBuilder();
 
   const [isEditing, setIsEditing] = useState(false);
+
+  useLayoutEffect(() => {
+    if (
+      !isEditing &&
+      focusOverride?.itemKey === item.key &&
+      focusOverride.part === 'key'
+    ) {
+      setIsEditing(true);
+      onActiveChange(true);
+      dispatch({type: 'RESET_FOCUS_OVERRIDE'});
+    }
+  }, [dispatch, focusOverride, isEditing, item.key, onActiveChange]);
 
   const {focusWithinProps} = useFocusWithin({
     onBlurWithin: () => {
@@ -124,10 +136,6 @@ const UnfocusedText = styled('span')`
 
 const Parameters = styled('span')`
   height: 100%;
-
-  &:not(:empty) {
-    padding: 0 ${space(0.25)};
-  }
 `;
 
 const KeyEditing = styled('div')`
--- a/static/app/components/searchQueryBuilder/tokens/filter/aggregateKey.tsx
+++ b/static/app/components/searchQueryBuilder/tokens/filter/aggregateKey.tsx
@@ -38,6 +38,7 @@ export function AggregateKey({
 
   const [isEditing, setIsEditing] = useState(false);
 
+  // Enters edit mode when focusOverride is set to this item
   useLayoutEffect(() => {
     if (
       !isEditing &&
--- a/static/app/components/searchQueryBuilder/tokens/filter/parametersCombobox.tsx
+++ b/static/app/components/searchQueryBuilder/tokens/filter/parametersCombobox.tsx
@@ -9,7 +9,6 @@ import {replaceCommaSeparatedValue} from 'sentry/components/searchQueryBuilder/t
 import type {AggregateFilter} from 'sentry/components/searchSyntax/parser';
 import {t} from 'sentry/locale';
 import {FieldValueType} from 'sentry/utils/fields';
-import {isCtrlKeyPressed} from 'sentry/utils/isCtrlKeyPressed';
 
 type ParametersComboboxProps = {
   onCommit: () => void;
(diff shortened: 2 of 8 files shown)


static/app/components/searchQueryBuilder/tokens/filter/aggregateKey.tsx
```diff
@@ -34,10 +34,22 @@ export function AggregateKey({
   filterRef,
 }: AggregateKeyProps) {
   const ref = useRef<HTMLDivElement>(null);
-  const {disabled} = useSearchQueryBuilder();
+  const {dispatch, focusOverride, disabled} = useSearchQueryBuilder();
 
   const [isEditing, setIsEditing] = useState(false);
 
+  useLayoutEffect(() => {
```
leeandher, line 41: It might be my lack of familiarity with the search query builder, but what is this hook doing?

static/app/components/searchQueryBuilder/tokens/filter/parametersCombobox.tsx
```diff
@@ -57,28 +176,70 @@ function useParameterSuggestions(): SelectOptionWithKey<string>[] {
...
+    inputValue,
+    selectionIndex
+  );
+
+  const items = useParameterSuggestions({token, parameterIndex});
+
+  const onKeyDown = useCallback(
+    (e: KeyboardEvent) => {
+      // Default combobox behavior stops events from propagating outside of input
+      // Certain keys like ctrl+z should be handled handled in useQueryBuilderGrid()
```
leeandher, line 199: typo: `handled handled`

static/app/components/searchQueryBuilder/tokens/filter/parametersCombobox.tsx
```diff
@@ -89,19 +250,22 @@ export function SearchQueryBuilderParametersCombobox({
...
       inputLabel={t('Edit function parameters')}
       onInputChange={e => setInputValue(e.target.value)}
+      onKeyDown={onKeyDown}
+      onKeyUp={updateSelectionIndex}
+      onClick={updateSelectionIndex}
       autoFocus
-      maxOptions={50}
-      openOnFocus
+      maxOptions={20}
+      isOpen
```
leeandher, line 262: I saw your comment about the state being controlled by this parent, is there a reason why this implementation is always open?

static/app/components/searchQueryBuilder/tokens/filter/utils.tsx
```diff
@@ -68,3 +68,28 @@ export function formatFilterValue(token: TokenResult<Token.FILTER>['value']): st
       return token.text;
   }
 }
+
+/**
+ * Replaces the focused parametr (at cursorPosition) with the new value.
+ *
+ * Example:
+ * replaceCommaSeparatedValue('foo,bar,baz', 5, 'new') => 'foo,new,baz'
```
leeandher, line 76: This is an awesome doc string 📝, the example helps a lot.

Thought two things:

1. Typo: `parametr`
2. Is the `cursorPosition` falling back to the end of the string standard for other of these helpers? If not it might help to add it to the doc string.

static/app/components/searchQueryBuilder/tokens/filter/utils.tsx
```diff
@@ -68,3 +68,28 @@ export function formatFilterValue(token: TokenResult<Token.FILTER>['value']): st
...
+  replacement: string
+) {
+  const items = value.split(',');
+
+  let characterCount = 0;
+  for (let i = 0; i < items.length; i++) {
+    characterCount += items[i].length + 1;
+    if (characterCount > (cursorPosition ?? value.length + 1)) {
+      const newItems = [...items.slice(0, i), replacement, ...items.slice(i + 1)];
+      return newItems.map(item => item.trim()).join(',');
```
leeandher, line 90: I'm not sure where else this may get used, but trimming the item seems like a side effect compared to the function name and doc string. If it's possible to remove it that might be a good call but mentioning it works as well.

static/app/components/searchQueryBuilder/tokens/filter/valueCombobox.tsx
```diff
@@ -174,30 +175,6 @@ function getValueAtCursorPosition(text: string, cursorPosition: number | null) {
 
   return '';
 }
-/**
- * Replaces the focused filter value (at cursorPosition) with the new value.
- *
- * Example:
- * replaceValueAtPosition('foo,bar,baz', 5, 'new') => 'foo,new,baz'
- */
-function replaceValueAtPosition(
```
leeandher, line 183: OOPS now i feel bad since this function was just moved and I left like 3 comments, still though!

leeandher (approved):
Looks good! Left a few question comments, clarified some stuff, but approving 🔥

static/app/components/searchQueryBuilder/tokens/filter/aggregateKey.tsx
malwilley: No you're right it is not obvious at all - added a comment to explain that it enters edit mode when the focus override is set to this item (the focus override is used when we specifically want to set the focus elsewhere after a state change)

static/app/components/searchQueryBuilder/tokens/filter/utils.tsx
malwilley: Fixed the typo and added info about the fallback

static/app/components/searchQueryBuilder/tokens/filter/utils.tsx
malwilley: I do agree with this actually, but I'll leave it for now since this is used for the string filter multi-select as well, which may need that trimming behavior (I forget if it was actually necessary or not). I may come back to this and remove that trim later on

static/app/components/searchQueryBuilder/tokens/filter/parametersCombobox.tsx
malwilley: I forced it to be open because it felt more natural. Since pressing escape exits edit mode, there isn't really a good way to close the menu anyway, so this felt like the best way to maintain some consistency

PR #75307, opened 2024-07-30
fix(ecosystems): Disables window close for integration pipeline dialog

GabeVillalobos:
Prevents the dialog window from immediately closing when the integration pipeline is finished. This currently happens, regardless of the success of the operation which results in us dropping errors messages when we should be displaying them to the customer.

### DIFF
--- a/src/sentry/templates/sentry/integrations/dialog-complete.html
+++ b/src/sentry/templates/sentry/integrations/dialog-complete.html
@@ -14,7 +14,6 @@
   if (window.opener) {
     window.opener.postMessage({{ payload|to_json }}, {{ document_origin|safe }});
   }
-  window.close();
 </script>
 {% endscript %}
 


src/sentry/templates/sentry/integrations/dialog-complete.html
```diff
@@ -14,7 +14,6 @@
   if (window.opener) {
     window.opener.postMessage({{ payload|to_json }}, {{ document_origin|safe }});
   }
-  window.close();
```
iamrajjoshi, line 17: do you have any idea why this was here in the first place?

iamrajjoshi (approved):

src/sentry/templates/sentry/integrations/dialog-complete.html
GabeVillalobos: I have no clue why. I'm guessing as a convenience thing when the pipeline succeeds, but it was never intended to close for failures.

PR #75309, opened 2024-07-30
ref(alerts): Refactor out alert latest incident info

davidenwang:
Refactors out the component that renders subtext here ex: `Triggered 6 days ago`
image

In my opinion the logic before was hard to follow as the alert rule type would trigger different branches of code and return statements. The new component separates each alert type's sub info into its own component with an appropriate comment and also adds tests

Depends on:
https://github.com/getsentry/sentry/pull/75308


evanpurkhiser (approved):

evanpurkhiser (approved):
FE seems good to me

PR #75319, opened 2024-07-31
ref(metrics): Render info when creating new span attribute

### LINKED ISSUE #75124: Poor man's autocomplete
Image
We should tell users here that if you've created an attribute with setAttribute/setData and don't see it in autocomplete, it's ok, they can just click enter.

Notion example:
Image

Vjeran:
Note next to autocomplete missing which would say: “There might be more things…”

priscilawebdev:
### Goal

- In the metrics modal, an extra message shall be displayed above the 'Create "new-span-attribute"' option in the dropdown, with a link to the docs for instrumenting custom metrics

### Notes
- This PR also updates the blue alert copy shown when the form changes
- This PR changes 'Custom: new-value' to 'Create new-value'

**Preview**

https://github.com/user-attachments/assets/dcb9c94d-4dc1-4c36-90a9-e194155aabc0

closes https://github.com/getsentry/sentry/issues/75124


vgrozdanic:
image

When selecting group and filter by, should we also show the message about creating new tag if it isn't already present there (which is very likely scenario)?

priscilawebdev:
hey @shellmayr yes I saw that 🙂 I did not want to give the same left padding to the new text, because it would look even more strange and I did not remove it from the Create because it is like that everywhere. Let me discuss this internally and work on it in a follow-up.

priscilawebdev:
@vgrozdanic is that is a good idea and I can create an extra issue for that. Do you know if we already have a doc describing it?

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
```diff
@@ -331,13 +338,38 @@ export function MetricsExtractionRuleForm({
...
+                Option: containerProps => {
+                  if (containerProps.data.__isNew__) {
+                    return (
+                      <Fragment>
+                        <OptionInfo>
+                          {tct(
+                            'Select an attribute or create one. [link:See how to instrument a custom attribute.]',
+                            {
+                              link: (
+                                <ExternalLink href="https://docs.sentry.io/product/explore/metrics/metrics-set-up/" />
```
vgrozdanic, line 356: I don't think we should be using this link, or maybe the docs should be updated, because currently old `metrics` API is mentioned on those docs, and we are trying to move away from it. There is no mention here about new span based attributes, or attaching an attribute to a span

Jesse-Box:
@priscilawebdev one option would be to add a plus icon to the left of the “Create...”. My thinking is that initially the users mental model might be to “select something from this list,” but eventually, they're “Adding something”...

Let me know if that jives with you or not.

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
priscilawebdev: Afaik @ale-cota is working on updating this doc

vgrozdanic:
> @vgrozdanic that is a good idea and I can create an extra issue. Do you know if we already have a doc describing it?

I don't know if we have a doc describing it, maybe @ale-cota has more details.

But I think that can be done as a part of this PR, because the same problem is happening with both "Filter and group by" and "Span attribute" fields, and that is the creation of something that we don't have indexed yet, and it shouldn't feel weird to create something that user knows should be there

priscilawebdev:
> > @vgrozdanic that is a good idea and I can create an extra issue. Do you know if we already have a doc describing it?
> 
> I don't know if we have a doc describing it, maybe @ale-cota has more details.
> 
> But I think that can be done as a part of this PR, because the same problem is happening with both "Filter and group by" and "Span attribute" fields, and that is the creation of something that we don't have indexed yet, and it shouldn't feel weird to create something that user knows should be there

All right! I can introduce this change as soon as we clarify the link to the docs 😉

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
vgrozdanic: Nice, ok, got it. Should we wait with this PR until the docs are updated, maybe we can link the specific section of those docs?

priscilawebdev: yes, I think we should but I will leave it up to @ale-cota  🙂

vgrozdanic: Perfect, thank you :)

ale-cota: We don't have to wait for the docs update to merge this, since it's behind a feature flag and only available to Sentry internal. Thanks!

ale-cota (approved):

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
```diff
@@ -410,12 +436,31 @@ export function MetricsExtractionRuleForm({
...
+                      {
+                        link: (
+                          <ExternalLink href="https://docs.sentry.io/product/explore/metrics/metrics-set-up/" />
+                        ),
+                      }
+                    )}
+                  />
+                );
+              },
+            }}
```
priscilawebdev, line 463: @vgrozdanic I have introduced the changes you suggested 😉

static/app/components/forms/controls/selectOption.tsx
```diff
@@ -56,17 +56,28 @@ function SelectOption(props: Props) {
...
-                  />
-                )}
-              </CheckWrap>
-              {data.leadingItems}
-            </Fragment>
+            itemProps.__isNew__ ? (
+              <Fragment>
+                <IconAdd size="sm" />
+                {data.leadingItems}
+              </Fragment>
```
priscilawebdev, line 63: @Jesse-Box I added your suggestion here, but since I updated a core component, I would like to wait for @vuluongj20 opinion as well. 

it looks like this 
image

static/app/components/forms/controls/selectOption.tsx
```diff
@@ -56,17 +56,28 @@ function SelectOption(props: Props) {
...
+            ) : (
+              isSelected && (
+                <Fragment>
+                  <CheckWrap isMultiple={false} isSelected>
+                    <IconCheckmark size="sm" />
+                  </CheckWrap>
+                  {data.leadingItems}
+                </Fragment>
```
priscilawebdev, line 78: I've implemented this change because the values looked odd when no item was selected—they were shifted to the right. @vuluongj20, could you please review and let me know if this works for you?

**Before:**
image


**After:**

image
image

priscilawebdev: ^ @shellmayr

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
```diff
@@ -331,13 +339,32 @@ export function MetricsExtractionRuleForm({
...
+                          ),
+                        }
+                      )}
+                    />
+                  );
+                },
+              }}
+              placeholder={t('Select span attribute')}
+              creatable
+              formatCreateLabel={value => tct('Create "[value]"', {value})}
```
obostjancic, line 367: I checked this out and we probably don't need this prop anymore since the default is `Create [value]` (same for other occurences)

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
```diff
@@ -306,12 +308,18 @@ export function MetricsExtractionRuleForm({
     [onSubmit]
   );
 
+  const isCustomSpanAttribute = useCallback((value: string) => {
```
obostjancic, line 311: This will return false for custom span attributes that are already present in the dropdown so i would propose renaming it to:
```suggestion
  const isNewCustomSpanAttribute = useCallback((value: string) => {
```

static/app/components/forms/controls/selectOption.tsx
shellmayr: Looks much better now 🙏

obostjancic (approved):

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
priscilawebdev: the new version has double quotes, but I don't have a strong opinion on this.

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
```diff
@@ -555,6 +606,29 @@ export function MetricsExtractionRuleForm({
...
+  ...props
+}: React.ComponentProps<typeof components.MenuList> & {info: React.ReactNode}) {
+  const theme = useTheme();
+  return (
+    <components.MenuList {...props}>
+      <div
+        css={css`
+          /* The padding must align with the values specified for the option in the forms/controls/selectOption component */
+          padding: ${space(1)};
+          padding-left: calc(${space(0.5)} + ${space(1.5)});
```
vgrozdanic, line 621: Is there a reason why this has to be done like this?

static/app/components/forms/controls/selectOption.tsx
```diff
@@ -56,17 +56,28 @@ function SelectOption(props: Props) {
...
-                {isSelected && (
-                  <IconCheckmark
-                    size={isMultiple ? 'xs' : 'sm'}
-                    color={isMultiple ? 'white' : undefined}
-                  />
-                )}
-              </CheckWrap>
-              {data.leadingItems}
-            </Fragment>
+            itemProps.__isNew__ ? (
```
priscilawebdev, line 59: I’m not fond of using "__isNew__", as it doesn’t feel quite right, but I don’t have a better alternative at the moment. For what it’s worth, newer versions of react-select still use it. You can see an example here: [react-select GitHub](https://github.com/JedWatson/react-select/blob/06e34882638d1526b9f5a1238bb567a3e9460ce5/packages/react-select/src/useCreatable.ts#L98-L102).

priscilawebdev: We could create our own prop to check if a value is new, but this would require manually updating every place we use it. Otherwise, the "create" option won’t show the plus icon. That might be fine as well

vgrozdanic: I would rather not show plus icon, than using the `__isNew__` attribute over which we have no control, and is not a public property, but I have no strong opinion here

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
priscilawebdev: I could use `${space(2)}`, but I prefer to be more explicit about the value. However, it might not be clear enough

Screenshot 2024-08-01 at 09 34 30
Screenshot 2024-08-01 at 09 34 37

priscilawebdev: the options are wrapped with elements that have different paddings. 1 element has `${space(0.5)} ` and the other `${space(1.5)}`

vgrozdanic: Ok, got it, makes sense then

static/app/components/forms/controls/selectOption.tsx
priscilawebdev: the whole data prop does not have a type, but we are using props from it 🙂 the `__isNew__` feels odd only because of the `__`

vgrozdanic (approved):

static/app/components/forms/controls/selectOption.tsx
priscilawebdev: I removed this change from this PR and created a [new one](https://github.com/getsentry/sentry/pull/75434) 😉

PR #75339, opened 2024-07-31
feat(metric): Add menu footer to the visualize field

### LINKED ISSUE #75127: Add create metric button to drop-down
**Problem**: Customers used to the old metrics solution are used to the top purple button for "create metric" to pull up instructions, so they ignore it. Also a customer that just instrumented a new span attribute, is likely to click on the dropdown and expect to see it already (as we've seen through early user testing).

**Design**: https://www.figma.com/design/gTSzrls85h2wW0Gfct9gM3/Specs%3A-Search-Experience-v3?node-id=14-1827&t=aDikrlFgsQviLiz6-0

priscilawebdev:
**preview**

https://github.com/user-attachments/assets/69b60100-e8e5-4c9f-ae10-0ad9f70871aa

closes https://github.com/getsentry/sentry/issues/75127

### DIFF
--- a/static/app/components/comboBox/index.tsx
+++ b/static/app/components/comboBox/index.tsx
@@ -45,6 +45,13 @@ interface ComboBoxProps<Value extends string>
   hiddenOptions?: Set<string>;
   isLoading?: boolean;
   loadingMessage?: string;
+  /**
+   * Footer to be rendered at the bottom of the menu.
+   * @closeOverlay is a function that closes the menu
+   */
+  menuFooter?:
+    | React.ReactNode
+    | ((actions: {closeOverlay: () => void}) => React.ReactNode);
   menuSize?: FormSize;
   menuWidth?: string;
   size?: FormSize;
@@ -67,6 +74,7 @@ function ComboBox<Value extends string>({
   menuWidth,
   hiddenOptions,
   hasSearch,
+  menuFooter,
   ...props
 }: ComboBoxProps<Value>) {
   const theme = useTheme();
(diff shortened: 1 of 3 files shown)


static/app/components/comboBox/index.tsx
```diff
@@ -464,4 +479,11 @@ const StyledLoadingIndicator = styled(LoadingIndicator)`
     width: 12px;
   }
 `;
+
+const MenuFooter = styled('div')`
```
priscilawebdev, line 483: I copied this from [this file on GitHub](https://github.com/getsentry/sentry/blob/3ee19817686bed81ec94843d05bef798db7e21f6/static/app/components/compactSelect/control.tsx#L708). I chose not to export the style because it felt not correct to do so, but I would still like to find a way to reuse it. @vuluongj20 do you have an opinion?

static/app/components/metrics/mriSelect/index.tsx
```diff
@@ -315,7 +319,7 @@ export const MRISelect = memo(function MRISelect({
         isLoading={isLoading}
         loadingMessage={t('Loading\u2026')}
         menuSize="sm"
-        menuWidth="450px"
+        menuWidth="450px" // TODO(priscilawebdev): update this value for small screens
```
priscilawebdev, line 322: since now we are using `useBreakpoints()` we can do better here and make this value more dynamic. Right now the menu does not look nice on small devices

obostjancic (approved):

PR #75345, opened 2024-07-31
ref: use new onboarding component

JonasBa:
Uses the new onboarding component that does not fetch onboarding from the docs.


static/app/utils/profiling/platforms.tsx
```diff
@@ -72,6 +77,21 @@ export function getDocsPlatformSDKForPlatform(
   if (platform === 'javascript-angular') {
     return 'javascript-angular';
   }
+  if (platform === 'javascript-angular') {
+    return 'javascript-angular';
+  }
```
Zylphrex, line 82: This is repeated from above.

Zylphrex (approved):

PR #75348, opened 2024-07-31
ref: fix mypy ignored errors checker in CI

asottile-sentry:
bash sucks, these are the errors that snuck through in ignored files that the ci check was supposed to prevent:

```
src/sentry/tagstore/snuba/backend.py:1179: error: Incompatible default for argument "dataset" (default has type "None", argument has type "Dataset")  [assignment]
src/sentry/testutils/cases.py:3253: error: Incompatible default for argument "scheduled_check_time" (default has type "None", argument has type "datetime")  [assignment]
src/sentry/testutils/cases.py:3290: error: Incompatible default for argument "timestamp" (default has type "None", argument has type "datetime")  [assignment]
src/sentry/testutils/cases.py:3291: error: Incompatible default for argument "duration" (default has type "None", argument has type "timedelta")  [assignment]
src/sentry/integrations/github/integration.py:482: error: Incompatible return value type (got "HttpResponseBase", expected "HttpResponse")  [return-value]
src/sentry/integrations/github/integration.py:507: error: Incompatible return value type (got "HttpResponseBase", expected "HttpResponse")  [return-value]
src/sentry/web/frontend/auth_login.py:280: error: Incompatible return value type (got "HttpResponseBase", expected "HttpResponse")  [return-value]
```

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -325,6 +325,7 @@ jobs:
 
       - run: |
           # mypy does not have granular codes so don't allow specific messages to regress
+          set -euo pipefail
           ! grep "'Settings' object has no attribute" .artifacts/mypy-all
           ! grep 'Cannot override class variable' .artifacts/mypy-all
           ! grep 'Exception type must be derived from BaseException' .artifacts/mypy-all
--- a/src/sentry/integrations/github/integration.py
+++ b/src/sentry/integrations/github/integration.py
@@ -6,7 +6,7 @@ from collections.abc import Collection, Mapping, Sequence
 from typing import Any
 from urllib.parse import parse_qsl
 
-from django.http import HttpResponse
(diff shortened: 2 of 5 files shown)


.github/workflows/backend.yml
```diff
@@ -325,6 +325,7 @@ jobs:
 
       - run: |
           # mypy does not have granular codes so don't allow specific messages to regress
+          set -euo pipefail
```
joshuarli, line 328: would be even better to put this at top level everywhere? i can slap this everywhere if you want

```
defaults:
  run:
    shell: bash -euo pipefail {0}
```

joshuarli (approved):

.github/workflows/backend.yml
asottile-sentry: it probably should -- but I wasn't ready to do that big of a change in this PR

PR #75358, opened 2024-07-31
feat(insights): reload projects when missing data

KevinL10:
Refetch the projects when the user arrives on an Insights Module or the Performance page with no data.

Users can now directly navigate to the Performance or Insights tabs and see their data during onboarding, without needing an additional refresh to synchronize the projects hasData flags.


DominikB2014 (approved):

static/app/views/insights/common/components/modulesOnboarding.tsx
```diff
@@ -42,8 +43,19 @@ export function ModulesOnboarding({
...
+  // Refetch the project metadata if the selected project does not have insights data, because
+  // we may have received insight data (and subsequently updated `Project.hasInsightxx`)
+  // after the initial project fetch.
+  useEffect(() => {
+    if (!hasData) {
+      reloadProjects();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [hasData]);
+
```
0Calories, line 58: How does the page know when to refetch the data? I see that the `useEffect` hook here is reliant on `hasData` changing, but I don't think this value changes unless the page filters are modified. Testing locally as well, I don't see any periodic network requests being sent out. 

Maybe I misunderstand the intent of the change, but do we not want to periodically check to see if data has come in?

KevinL10: Good question, we're only polling at the start if there's no data instead of periodically. When the user visits a page with no data we'll perform an initial refetch, but no additional requests beyond that.

PR #75363, opened 2024-07-31
ref(controlsilo) Add forwards compatibility for integration tasks

markstory:
Add forwards compatible tasks for github, jira and opsgenie plugins, so that when we update callers we won't drop any tasks. These aliases will also ensure continuity should we need to rollback.

The next step of this work is to move the implementation of these tasks into the new modules and have the old modules call the new ones. After that is complete, we'll remove the old module locations.

Split off of #74925


Christinarlong (approved):
lgtm 👍

src/sentry/integrations/jira/tasks.py
Christinarlong, line 1: curious q, where is this file and `opsgenie/tasks.py` named `tasks.py` vs `migrate_issues` etc.

markstory: With only one or two tasks, we can skip a layer of directories and have a single file as the module. For larger integrations we can & should have multiple modules for tasks, and we can use a `__init__.py` to make make the celery imports simpler.

PR #75366, opened 2024-07-31
feat(profiling): Do not require transactions for landing page flamegraph

Zylphrex:
The landing page should not depend on transactions so switch to the profiles data source to generate flamegraphs for it.


static/app/utils/profiling/hooks/useAggregateFlamegraphQuery.ts
```diff
@@ -56,3 +85,21 @@ export function useAggregateFlamegraphQuery({
...
+  props: AggregateFlamegraphQueryParameters
+): props is FunctionsAggregateFlamegraphQueryParameters {
+  return 'fingerprint' in props;
+}
+
+function isDataSourceTransactions(
+  props: AggregateFlamegraphQueryParameters
+): props is TransactionsAggregateFlamegraphQueryParameters {
+  return !isDataSourceProfiles(props) && isDataSourceFunctions(props);
+}
```
JonasBa, line 105: :catjam:

JonasBa (approved):

PR #75370, opened 2024-07-31
chore(similarity): Add option rerank seer records in backfill

jangjodi:
Add use_reranking option to send in similarity backfill to allow seer to use reranking logic


JoshFerge (approved):
might just be better / simpler to control this through an option, but this also works

vartec (approved):

trillville (approved):

jangjodi:
> might just be better / simpler to control this through an option, but this also works

good suggestion, thank you!

PR #75372, opened 2024-07-31
chore(insights): Remove empty domain sample search hack

gggritso:
The hack was added in https://github.com/getsentry/sentry/pull/69129 but the data was fixed in https://github.com/getsentry/snuba/pull/5780. The hack is no longer needed because it's been over 90 days, so the old data has fallen out of retention.


0Calories (approved):
🧹

PR #75375, opened 2024-07-31
feat(onboarding): refresh "perform these actions" dropdown after installation

ameliahsu:
once user installs integration:
- modal is closed
-  "Connect to messaging" button is hidden
-  "THEN perform these actions" dropdown is refreshed to include actions for the newly installed messaging integration

https://github.com/user-attachments/assets/06c60700-ed3c-4e8a-a987-096c50b1def5

also added a new dropdown option to manually refresh the dropdown actions, in case of external installations such as MS Teams
Screenshot 2024-08-01 at 4 08 08 PM

### DIFF
--- a/static/app/views/alerts/rules/issue/ruleNodeList.tsx
+++ b/static/app/views/alerts/rules/issue/ruleNodeList.tsx
@@ -306,7 +306,9 @@ class RuleNodeList extends Component<Props> {
           placeholder={placeholder}
           value={null}
           onChange={obj => {
-            additionalAction ? additionalAction.onClick() : onAddRow(obj.value);
+            additionalAction && obj === additionalAction.option
+              ? additionalAction.onClick()
+              : onAddRow(obj.value);
           }}
           options={options}
           disabled={disabled}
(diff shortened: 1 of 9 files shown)


static/app/views/alerts/rules/issue/ruleNodeList.tsx
```diff
@@ -254,10 +264,19 @@ class RuleNodeList extends Component<Props> {
...
+    let options;
+    if (selectType === 'grouped') {
+      options = groupSelectOptions(enabledNodes);
+      if (additionalAction) {
+        for (const option of options) {
+          if (option.label === additionalAction.label) {
+            option.options.push(additionalAction.option);
+          }
+        }
+      }
```
saponifi3d, line 276: Rather than using a `for` loop here, I'd recommend using [`.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) to get the option, and then modify it.

```js
if (additionalAction) {
  const optionToModify = options.find((option) => option.label === additionalAction.label)
  optionToMoidfy.options.push(additionalAction.option)
} else {
  ...
```
since javascript uses references to variables, this will modify the option in the option list.

---

Using find will also be slightly more efficient as will break out of the loop execution after the option is found.

saponifi3d (approved):
lgtm! just the recommendation about .find 🎉

PR #75378, opened 2024-07-31
fix(issue-priority): Re-add platform check

snigdhas:
Enable the platform check regardless of feature flags when calling Seer since the service only supports severity calculation for Python and JS projects.

### DIFF
--- a/src/sentry/event_manager.py
+++ b/src/sentry/event_manager.py
@@ -2360,14 +2360,13 @@ def _get_severity_metadata_for_group(
     if not seer_based_priority_enabled and not feature_enabled:
         return {}
 
-    if not seer_based_priority_enabled:
-        is_supported_platform = (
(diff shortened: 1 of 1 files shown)


src/sentry/event_manager.py
```diff
@@ -2360,14 +2360,13 @@ def _get_severity_metadata_for_group(
...
 
-    if not seer_based_priority_enabled:
-        is_supported_platform = (
-            any(event.platform.startswith(platform) for platform in PLATFORMS_WITH_PRIORITY_ALERTS)
-            if event.platform
-            else False
-        )
-        if not is_supported_platform:
-            return {}
+    is_supported_platform = (
```
snigdhas, line 2363: Mistakenly put this behind a feature flag, but this platform check is separate from the default-alerts platform check which we're getting rid of. In this case, we can skip calling Seer if we know the service won't support severity for those platforms.

roggenkemper (approved):

PR #75379, opened 2024-07-31
ref(frontend): Convert DebugMeta to FC

leeandher:
To change this component for some issue details work, I needed to either add a prop to get a hooks value, or change it to a functional component. Way more work this way but I think its a bit better. Few notes:

- Still uses previous version of list virtualizer
- No longer uses query params for managing modal values

### DIFF
--- a/static/app/components/events/eventEntry.tsx
+++ b/static/app/components/events/eventEntry.tsx
@@ -122,7 +122,6 @@ function EventEntryContent({
           event={event}
           projectSlug={projectSlug}
           groupId={group?.id}
-          organization={organization as Organization}
           data={entry.data}
         />
       );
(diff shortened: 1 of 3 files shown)


malwilley (approved):
Thanks for doing this! Great to be rid of another class component.

Merging this is a bit scary because of the lack of test coverage for this file, but for what it's worth I tried what I could and it seems to work the same as before.

leeandher:
> Thanks for doing this! Great to be rid of another class component.
> 
> Merging this is a bit scary because of the lack of test coverage for this file, but for what it's worth I tried what I could and it seems to work the same as before.

Good point, I'll add a test and double check all the existing assumptions and paste a run of it passing with the previous component.

PR #75383, opened 2024-07-31
chore(profiling): Change default to left heavy flamegraph

Zylphrex:
Change the default to a left heavy flamegraph to make the areas of focus more obvious.


JonasBa (approved):
Much better 💯

PR #75384, opened 2024-07-31
feat(perf): Upgrade Transaction Summary searchbar to `SearchQueryBuilder`

0Calories:
Upgrades the search bar in Transaction Summary to the new `SearchQueryBuilder` component. We now have various categorizations for search filter keys, which are modular so they can be reused when upgrading search bars on other Insights / Performance pages.

The default ordering of the filter categories is TBD, feel free to let me know if you think they should be re-ordered in a certain way. I also referred to the tags that are fetched from `TagStore` as "Custom Tags", but not sure how accurate this is, so I'm open to renaming this section 

In a followup PR, I'd like to introduce the ability to suggest certain categories **only** if it makes sense to for the currently selected project. For example, if the current transaction summary is for a backend Python transaction, we won't surface categories like `Web Vitals` or `Mobile Vitals`.

### Before
image

### After
image

**NOTE**: This new search bar is gated behind the `search-query-builder-performance` feature flag, which is an EA feature. When the new `SearchQueryBuilder` component is shipped to GA, this will become the default search experience for all customers.


static/app/views/performance/transactionSummary/transactionOverview/content.tsx
```diff
@@ -344,6 +378,83 @@ function SummaryContent({
     organization.features.includes('performance-spans-new-ui') &&
     organization.features.includes('insights-initial-modules');
 
+  // This is adapted from the `getEventFieldValues` function in `events/searchBar.tsx`
+  const getTransactionFilterTagValues = memoize(
```
malwilley, line 382: There's no need to use `memoize` anymore, the component should cache these results automatically (since it uses `useQuery` under the hood)

static/app/views/performance/transactionSummary/transactionOverview/content.tsx
```diff
@@ -344,6 +378,83 @@ function SummaryContent({
...
+        return Promise.resolve([]);
+      }
+
+      // device.class is stored as "numbers" in snuba, but we want to suggest high, medium,
+      // and low search filter values because discover maps device.class to these values.
+      if (isDeviceClass(tag.key)) {
+        return Promise.resolve(DEVICE_CLASS_TAG_VALUES);
+      }
+
+      const fetchPromise = fetchTagValues({
```
malwilley, line 398: Not totally necessary, but the results should be improved (and return more quickly) if you pass time range parameters with `endpointParams`.

You can also pass `sort: '-count'` for a more useful order (it defaults to last_seen)

static/app/views/performance/transactionSummary/transactionOverview/content.tsx
```diff
@@ -344,6 +378,83 @@ function SummaryContent({
...
+        tagKey: tag.key,
+        search: queryString,
+        projectIds: projectIdStrings,
+        // allows searching for tags on transactions as well
+        includeTransactions: true,
+        // allows searching for tags on sessions as well
+        includeSessions: true,
+      });
+
+      return fetchPromise.then(
```
malwilley, line 410: nit: could use async/await instead

static/app/views/performance/transactionSummary/transactionOverview/content.tsx
```diff
@@ -344,6 +378,83 @@ function SummaryContent({
...
+              actionBarItems={generateActionBarItems(
+                organization,
+                location,
+                mepDataContext
+              )}
+            />
+          )}
+        >
+          <SearchQueryBuilder
+            filterKeys={getTransactionFilterTags(tagStoreCollection)}
```
malwilley, line 443: Since it's a pretty heavy component, I'd recommend putting `filterKeys` and other non-primitive inputs in `useMemo` or `useCallback` to prevent excess rerenders

static/app/views/performance/transactionSummary/transactionOverview/content.tsx
```diff
@@ -344,6 +378,83 @@ function SummaryContent({
...
+        }
+      );
+    },
+    ({key}, queryString) => `${key}-${queryString}`
+  );
+
+  function renderSearchBar() {
+    return (
+      <StyledSearchBarWrapper>
+        <Feature
```
malwilley, line 423: nit: in most cases we aren't using this component anymore, it's simple enough to just do `organization.features.includes('search-query-builder-performance')`

malwilley (commented):
This looks really good!! Thanks for implementing!

static/app/views/performance/transactionSummary/transactionOverview/content.tsx
0Calories: Oh right! Good callout, I'll change that

0Calories:
Thanks for the suggestions @malwilley! Just finished adding everything you suggested

static/app/views/performance/transactionSummary/transactionOverview/content.tsx
```diff
@@ -344,6 +383,75 @@ function SummaryContent({
...
+        const results = await fetchTagValues({
+          api,
+          orgSlug: organization.slug,
+          tagKey: tag.key,
+          search: queryString,
+          projectIds: projectIdStrings,
+          includeTransactions: true,
+          includeSessions: true,
+          sort: '-count',
+          endpointParams: {start: eventView.start, end: eventView.end},
```
malwilley, line 413: This is missing `statsPeriod`, so if you have say 14d selected this doesn't limit by time range

malwilley (approved):

static/app/views/insights/constants.tsx
gggritso, line 1: This is great! Generic definitions are awesome. I think these field definitions can go into `static/app/utils/discover/fields.tsx`. They're not specific to Insights, since they're most transaction fields, so it's good to lift these out of the `/insights/` directory where everyone can collaborate on them.

@malwilley unless there another good place to keep these nice generic definitions? I want to avoid people re-inventing these, and I want to avoid having to update multiple places if we add a new field

gggritso (commented):
Very nice! I think it'd be good to figure out where to put those field definitions (I'm pretty confident that `/insights/constants.tsx` is not the right place) but otherwise this rips 👍🏻

static/app/views/insights/constants.tsx
0Calories: That's fair, these definitions are for sure not specific to Insights. I'll move them over to discover!

0Calories:
Going to merge this for now, can follow up with an additional PR if we feel that `discover/fields.tsx` isn't the right place to hold these generic definitions

static/app/views/insights/constants.tsx
malwilley: Discover fields is an alright place to put them I think. I don't have a better place to put this sort of config, it really depends on whether or not other instances of the search bar will organize fields this way or not.

PR #75385, opened 2024-07-31
feat(custom-views): Support renaming tab titles

MichaelSun48:
This PR adds support for the Rename button that appears in a tab menu. Some implementation specific logic:

- Pressing escape discards your current change
- Pressing enter confirms your current change
- Clicking away from the tab is effectively the same as pressing enter
- Empty strings are not allowed. Attempting to save one will revert back to the original value 

https://github.com/user-attachments/assets/ef211987-a3b4-4f34-9318-1c548a416c51

### DIFF
--- a/static/app/views/issueList/draggableTabBar.tsx
+++ b/static/app/views/issueList/draggableTabBar.tsx
@@ -1,6 +1,6 @@
 import 'intersection-observer'; // polyfill
 
-import {useEffect, useState} from 'react';
+import {useEffect, useRef, useState} from 'react';
 import styled from '@emotion/styled';
 import type {Key, Node} from '@react-types/shared';
 
@@ -14,6 +14,7 @@ import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
 import {defined} from 'sentry/utils';
 import {DraggableTabMenuButton} from 'sentry/views/issueList/draggableTabMenuButton';
+import EditableTabTitle from 'sentry/views/issueList/editableTabTitle';
 
 export interface Tab {
   content: React.ReactNode;
@@ -51,11 +52,6 @@ export interface DraggableTabBarProps {
    */
   onDuplicate?: (key: MenuItemProps['key']) => void;
   /**
-   * Callback function to be called when user clicks the 'Rename' button.
-   * Note: The `Rename` button only appears for persistent views
-   */
-  onRename?: (key: MenuItemProps['key']) => void;
-  /**
    * Callback function to be called when user clicks the 'Save' button.
    * Note: The `Save` button only appears for persistent views when `isChanged=true`
    */
@@ -64,6 +60,11 @@ export interface DraggableTabBarProps {
    * Callback function to be called when user clicks the 'Save View' button for temporary views.
    */
   onSaveTempView?: () => void;
+  /**
+   * Callback function to be called when user renames a tab.
+   * Note: The `Rename` button only appears for persistent views
+   */
+  onTabRenamed?: (key: MenuItemProps['key'], newLabel: string) => void;
   tempTabContent?: React.ReactNode;
   tempTabLabel?: string;
 }
@@ -78,12 +79,18 @@ export function DraggableTabBar({
   onDelete,
   onDiscard,
   onDuplicate,
-  onRename,
+  onTabRenamed,
   onSave,
   onDiscardTempView,
   onSaveTempView,
 }: DraggableTabBarProps) {
   const [selectedTabKey, setSelectedTabKey] = useState<Key>(tabs[0].key);
+  const [isEditingTabTitle, setIsEditingTabTitle] = useState<{
+    isEditing: boolean;
+    editingTabKey?: string; // Should be undefined when isEditing = false
+  }>({isEditing: false, editingTabKey: undefined});
+
+  const tabLabelRef = useRef<HTMLInputElement>(null);
 
   useEffect(() => {
     if (!showTempTab && selectedTabKey === 'temporary-tab') {
@@ -131,6 +138,21 @@ export function DraggableTabBar({
     onAddView?.(e);
   };
 
+  const handleOnRename = (tab: Tab) => {
+    setIsEditingTabTitle({
+      isEditing: true,
+      editingTabKey: tab.key,
+    });
+  };
+
+  const handleOnTabRenamed = (newLabel: string, tabKey: string) => {
+    const tab = tabs.find(tb => tb.key === tabKey);
+    if (tab && newLabel !== tab.label) {
+      setTabs(tabs.map(tb => (tb.key === tab.key ? {...tb, label: newLabel} : tb)));
+      onTabRenamed?.(tab.key, newLabel);
+    }
+  };
+
   const makeMenuOptions = (tab: Tab): MenuItemProps[] => {
     if (tab.key === 'temporary-tab') {
       return makeTempViewMenuOptions({
@@ -140,7 +162,7 @@ export function DraggableTabBar({
     }
     if (tab.hasUnsavedChanges) {
       return makeUnsavedChangesMenuOptions({
-        onRename,
+        onRename: () => handleOnRename(tab),
         onDuplicate: () => handleOnDuplicate(tab),
         onDelete: tabs.length > 1 ? () => handleOnDelete(tab) : undefined,
         onSave,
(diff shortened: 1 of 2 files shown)


static/app/views/issueList/draggableTabBar.tsx
```diff
@@ -171,7 +193,23 @@ export function DraggableTabBar({
...
+                setIsEditing={isEditing =>
+                  setIsEditingTabTitle({
+                    isEditing,
+                    editingTabKey: isEditing ? tab.key : undefined,
+                  })
+                }
+                onChange={newLabel => handleOnTabRenamed(newLabel.trim(), tab.key)}
+                inputRef={
+                  isEditingTabTitle.editingTabKey === tab.key ? tabLabelRef : undefined
+                }
```
MichaelSun48, line 211: I'm attaching the single reference to _only_ the title that is being renamed, rather than creating a reference for each tab title from the start. This feels like it could be one of those things that has unforeseen consequences, lmk if that's the case

static/app/views/issueList/draggableTabBar.tsx
```diff
@@ -64,6 +60,11 @@ export interface DraggableTabBarProps {
    * Callback function to be called when user clicks the 'Save View' button for temporary views.
    */
   onSaveTempView?: () => void;
+  /**
+   * Callback function to be called when user renames a tab.
+   * Note: The `Rename` button only appears for persistent views
+   */
+  onTabRenamed?: (key: MenuItemProps['key'], newLabel: string) => void;
```
MichaelSun48, line 67: Changed `onRename` -> `onTabRenamed` because the parent component probably cares when the tab has already been renamed, not when the `Rename` menu option is clicked.

static/app/views/issueList/editableTabTitle.tsx
MichaelSun48, line 1: This class is mostly a pared down version of <EditableText/>, made specifically for this use case. I'm also not really sure if this is the right place to put this component, it feels like it should be part of a "helper" folder, or combined with `<DraggableTabBar/>` into one component.

static/app/views/issueList/editableTabTitle.tsx
```diff
@@ -0,0 +1,120 @@
...
+  useEffect(() => {
+    if (isEditing) {
+      onEnter();
+      onEsc();
+    }
+  }, [onEnter, onEsc, isEditing]);
+
+  useEffect(() => {
+    setTimeout(() => {
+      inputRef?.current?.focus();
```
malwilley, line 80: Can you use the `autoFocus` attribute (on the input) instead?

static/app/views/issueList/editableTabTitle.tsx
```diff
@@ -0,0 +1,120 @@
...
+  label: string;
+  onChange: (newLabel: string) => void;
+  setIsEditing: (isEditing: boolean) => void;
+  inputRef?: React.RefObject<HTMLInputElement>;
+}) {
+  const [inputValue, setInputValue] = useState(label);
+
+  const wrapperRef = useRef<HTMLDivElement>(null);
+
+  const enter = useKeyPress('Enter', undefined, true);
```
malwilley, line 24: Looks like this is how the existing `EditableText` component does this, but it would be better I think to do this in the `onKeyDown` handler on the input

static/app/views/issueList/editableTabTitle.tsx
```diff
@@ -0,0 +1,120 @@
...
+      />
+    </div>
+  ) : (
+    label
+  );
+}
+
+export default EditableTabTitle;
+
+const StyledInput = styled('input')`
```
malwilley, line 106: We do have a reusable `GrowingInput` component that you could use. Though this does seem to work well

static/app/views/issueList/editableTabTitle.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const [inputValue, setInputValue] = useState(label);
+
+  const wrapperRef = useRef<HTMLDivElement>(null);
+
+  const enter = useKeyPress('Enter', undefined, true);
+  const esc = useKeyPress('Escape', undefined, true);
+
+  const isEmpty = !inputValue.trim();
+
+  useOnClickOutside(wrapperRef, () => {
```
malwilley, line 29: Is this necessary? This might work fine with just the `onBlur`

malwilley (commented):
DraggableTabBar could use some tests for this

static/app/views/issueList/draggableTabBar.tsx
malwilley: Do you need the ref for anything in the future? If not I wouldn't keep track of any refs

static/app/views/issueList/draggableTabBar.tsx
```diff
@@ -78,12 +79,18 @@ export function DraggableTabBar({
...
   onDuplicate,
-  onRename,
+  onTabRenamed,
   onSave,
   onDiscardTempView,
   onSaveTempView,
 }: DraggableTabBarProps) {
   const [selectedTabKey, setSelectedTabKey] = useState<Key>(tabs[0].key);
+  const [isEditingTabTitle, setIsEditingTabTitle] = useState<{
+    isEditing: boolean;
```
malwilley, line 89: Hmm I don't think you need this `isEditing` state - could just have `editingTabKey: string | null` where null means nothing is being edited.

One other way you could simplify things (if you wanted, not required since this works well enough), is put this state as low as possible. You could extract the tab item into its own component and just have `const [isEditing, setIsEditing] = useState(false)`. Your new component would look something like:

```tsx
return (
  <Tabs>
    <DraggableTabList>
      {tabs.map(tab => <TabItem {...} />)}
    </DraggableTabList>
  </Tabs>
)

// in another file

function TabItem(...) {
  const [isEditing, setIsEditing] = useState(false)

  return (
   <DraggableTabList.Item>
     {...}
   </DraggableTabList.Item>
  )
}
```

It's always preferred to place the state as low as you can because it reduces rerenders and generally makes things simpler. The one downside is that you _could_ technically have multiple tabs editing at once, but I don't think that is actually possible since you exit editing on blur.

static/app/views/issueList/editableTabTitle.tsx
MichaelSun48: This doesn't work for some reason :/

static/app/views/issueList/draggableTabBar.tsx
MichaelSun48: Good point about the state! I also agree that the editing state makes more sense to apply to just the TabItem, but for the sake of getting this out to internal as soon as possible, I'll leave that out of this PR for now.

static/app/views/issueList/editableTabTitle.tsx
MichaelSun48: GrowingInput also (And everything derived from Input) seems to have a few non-configurable opinions on how an editable state should look (like having a background, fixed height, etc.) that we don't want here

MichaelSun48:
> DraggableTabBar could use some tests for this

You're absolutely right, but after trying to write some tests for this, I think it makes the most sense to make those their own PR since there are several small changes/tweaks that need to be made but are not related to this PR. I'll work on that now and It'll be the next PR that comes out.

static/app/views/issueList/editableTabTitle.tsx
malwilley: Hmm it also seems like you don't need the `setTimeout` either. I wonder if something attempts to steal focus when clicking edit

static/app/views/issueList/editableTabTitle.tsx
malwilley: You can do this later on as necessary, but might make sense to put all the related components into a `groupSearchViewTabs` folder (or something similar)

static/app/views/issueList/editableTabTitle.tsx
MichaelSun48: Oh interesting - the setTimeout solved a previous issue I had, but it sounds like that's no longer an issue. Same thing happened with the `onInputChange = {e => e.stopPropogation()}` - I deleted it after a few other changes, and then the issue just didn't appear anymore.

static/app/views/issueList/editableTabTitle.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const isEmpty = !inputValue.trim();
+
+  const handleOnBlur = () => {
+    if (!isEditing) {
+      return;
+    }
+    if (isEmpty) {
+      setInputValue(label);
```
malwilley, line 26: You could avoid this line if the input (and corresponding state) was extracted to its own component. That way the state would be reset whenever the edit state was entered.

malwilley (approved):
Overall looks good!

PR #75386, opened 2024-07-31
fix(hybrid-cloud): Pushes a temporary fix for pydantic unpickling issues with the v2 upgrade

GabeVillalobos:
Adds a `__setstate__` override for `RpcModel` to allow it to correctly unpickle digest payloads from Redis.

This is a temporary workaround while the upgrade settles and will be removed once all digests from the previous version have been processed.

### DIFF
--- a/src/sentry/hybridcloud/rpc/__init__.py
+++ b/src/sentry/hybridcloud/rpc/__init__.py
@@ -49,6 +49,16 @@ class RpcModel(pydantic.BaseModel):
         from_attributes=True, use_enum_values=True, coerce_numbers_to_str=True
     )
 
+    def __setstate__(self, state: dict[Any, Any]):
+        """
+        __setstate__ override to alleviate an unpickling issue in production with the pydantic version upgrade.
+        """
+        state.setdefault("__pydantic_extra__", {})
+        state.setdefault("__pydantic_private__", {})
+
+        if "__pydantic_fields_set__" not in state:
+            state["__pydantic_fields_set__"] = state.get("__fields_set__")
+
     @classmethod
     def get_field_names(cls) -> Iterable[str]:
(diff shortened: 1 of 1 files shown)


wedamija (approved):

markstory (approved):

GabeVillalobos:
Tested this locally between versions 1 and 2 and caught a bug with the implementation. For reference here's how I validated this:

```python
def test_pickle():
    user, organization, project, user_email = generate_base_data()
    codec = CompressedPickleCodec()

    with open("pickle_tester_user.txt", "wb") as f:
        f.write(codec.encode(user))
    with open("pickle_tester_organization.txt", "wb") as f:
        f.write(codec.encode(organization))
    with open("pickle_tester_project.txt", "wb") as f:
        f.write(codec.encode(project))
    with open("pickle_tester_user_email.txt", "wb") as f:
        f.write(codec.encode(user_email))

def test_depickle():
    codec = CompressedPickleCodec()
    expected_user, expected_organization, expected_project, expected_user_email = generate_base_data()
    with open("pickle_tester_user.txt", "rb") as f:
        read_bytes = f.read()
        user = codec.decode(read_bytes)
        assert user == expected_user

    with open("pickle_tester_organization.txt", "rb") as f:
        read_bytes = f.read()
        org = codec.decode(read_bytes)
        assert org == expected_organization

    with open("pickle_tester_project.txt", "rb") as f:
        read_bytes = f.read()
        project = codec.decode(read_bytes)
        assert project == expected_project

    with open("pickle_tester_user_email.txt", "rb") as f:
        read_bytes = f.read()
        user_email = codec.decode(read_bytes)
        assert user_email == expected_user_email

```

The object contents were identical, but there must be some internal field differences because equality checks failed for some reason.

PR #75390, opened 2024-07-31
fix(discover): Crash when event view requires trace id

shruthilayaj:
The component hierarchy for the table in discover goes
Results > Table > TableView

Results uses `getDerivedStateFromProps` to generate `eventView`
which is passed down to `Table` and `TableView`

`Table` uses eventView to fetch table data and passes down
`tableData` and `isLoading` state down to `TableView`

`TableView` uses `eventView`, `tableData` and `isLoading` to
render the table. If no aggregates are present in eventView, we prepend
an event ID column to eventView. Rendering the event ID column requires
an event ID and a trace ID and throws an error when trace id does not exist.

When eventView switches from a state where trace id isn't required
to one where trace id is required, that prop change immediately gets
propagated down to `Table` and `TableView` causing both components
to re-render. `Table` only refetches data on `componentDidUpdate` which
happens after a render. So `TableView` has stale tableData and `isLoading`
still set to false until fetchData is called by `Table`. 
This state mismatch trips the "no trace ID" error since the new data isn't
available yet.

We're introducing `getDerivedStateFromProps` on `Table` to force `isLoading`
to true when fields or datasets change. 

fixes JAVASCRIPT-2TT6

### DIFF
--- a/static/app/views/discover/savedQuery/datasetSelector.tsx
+++ b/static/app/views/discover/savedQuery/datasetSelector.tsx
@@ -60,6 +60,9 @@ export function DatasetSelector(props: Props) {
       value={value}
       options={options}
       onChange={newValue => {
+        if (newValue.value === value) {
+          return;
+        }
         let nextEventView: EventView;
         if (eventView.id) {
           nextEventView = eventView.withDataset(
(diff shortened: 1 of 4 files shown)


static/app/views/discover/savedQuery/datasetSelector.tsx
```diff
@@ -60,6 +60,9 @@ export function DatasetSelector(props: Props) {
       value={value}
       options={options}
       onChange={newValue => {
+        if (newValue.value === value) {
```
shruthilayaj, line 63: minor unrelated change to no op if the same dataset is selected

mjq (approved):
Thanks for the detailed description, made this much easier to follow 🙏

PR #75395, opened 2024-07-31
ref: strongly type get_configure_view

asottile-sentry:
the "configure view" is more of a snippet that's injected via `src/sentry/web/frontend/organization_auth_settings.py` into `organization-auth-provider-settings.html`

as such it's not really a "view" in the django sense (which returns an instance of `HttpResponseBase`) -- separating this from the inheritance hierarchy of `BaseView` allows this to be properly typed (since the original signature of `ConfigureView.dispatch` is incompatible with `BaseView`)

this allows it to be strongly typed -- and fixes a bunch of adjacent mypy errors


src/sentry/plugins/base/response.py
```diff
@@ -1,20 +1,24 @@
...
 
-__all__ = ("Response",)
+from sentry.web.helpers import render_to_string
 
+__all__ = ("DeferredResponse",)
 
-class Response:
-    def __init__(self, template, context=None):
+
+class DeferredResponse:
```
asottile-sentry, line 13: I renamed this because it was getting confused in several places with `rest_framework.response.Response` and `requests.models.Response` -- this name should also be more clear as to what it's trying to do

joshuarli (approved):

PR #75079, opened 2024-07-26
chore(api): Unpublish ProjectUserReportsEndpoint

iamrajjoshi:
The API is deprecated and Feedback team wants to unpublish it from the api docs. cc @aliu39

PR #75396, opened 2024-07-31
chore(feedback): Add deprecation notes to user report endpoint apidocs

aliu39:
Ref https://github.com/getsentry/sentry-docs/issues/10643
The replay team decided on a call to revert https://github.com/getsentry/sentry/pull/75079 because some users may still need this doc as a reference.

What it looks like:

Screenshot 2024-07-31 at 3 18 44 PM
Screenshot 2024-07-31 at 3 27 25 PM

### DIFF
--- a/api-docs/paths/projects/user-feedback.json
+++ b/api-docs/paths/projects/user-feedback.json
@@ -1,7 +1,7 @@
 {
   "get": {
     "tags": ["Projects"],
-    "description": "Return a list of user feedback items within this project.",
+    "description": "Return a list of user feedback items within this project.\n\n*This list does not include submissions from the [User Feedback Widget](https://docs.sentry.io/product/user-feedback/#user-feedback-widget). This is because it is based on an older format called User Reports - read more [here](https://develop.sentry.dev/application/feedback-architecture/#user-reports).*",
     "operationId": "List a Project's User Feedback",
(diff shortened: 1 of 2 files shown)


JoshFerge (approved):

api-docs/paths/projects/user-feedback.json
```diff
@@ -1,7 +1,7 @@
 {
   "get": {
     "tags": ["Projects"],
-    "description": "Return a list of user feedback items within this project.",
+    "description": "Return a list of user feedback items within this project.\n\n*This list does not include submissions from the [User Feedback Widget](https://docs.sentry.io/product/user-feedback/#user-feedback-widget). This is because it is based on an older format called User Reports - read more [here](https://develop.sentry.dev/application/feedback-architecture/#user-reports).*",
```
michellewzhang, line 4: was this manually modified?👀

aliu39: Yeah - turns out the json files is what api-docs actually uses. But they want the docstrings to stay consistent since they might switch to using those in the future.

michellewzhang: oh cool i did not know - TIL!

michellewzhang (approved):

PR #75403, opened 2024-07-31
feat(toolbar): add a replay panel for start/stop current replay

### LINKED ISSUE #74452: [DevToolbar] Add 'start replay' buttons to the toolbar
We should add a button to the toolbar to start a replay, it one isn't already running. This way users will be able to guarantee that a replay is running and will capture whatever it is their doing as the dogfood the site

Related to https://github.com/getsentry/sentry/issues/74583

### LINKED ISSUE #74583: [DevToolbar] When a replay is running, link to the Replay Details page for that replay
This will allow users to be doing some dogfooding, and if there's something to share they can find and share it quickly

If we can't deeplink to the exact replay, then link out to a search page that has a useful search term inserted.

Related to https://github.com/getsentry/sentry/issues/74580

aliu39:
Closes #74583 
Closes #74452 

`getReplay` returns undefined:
Screenshot 2024-08-08 at 10 28 18 AM

SentrySDK doesn't have `getReplay` method:
Screenshot 2024-08-08 at 10 22 25 AM

SentrySDK is falsey (failed to import the package):
Screenshot 2024-08-08 at 10 22 02 AM

If you want to checkout this branch to test it, you need to run dev-ui in getsentry and make local changes: 
- Comment out https://github.com/getsentry/getsentry/blob/2a1da081f3a9e4e4111b577d5551fa24691da374/static/getsentry/gsApp/utils/useReplayInit.tsx#L85-L87
- Right below that, manually control the sample rates by commenting out/overriding ^. Best to test with 1.0/0.0, 0.0/1.0, 0/0.

Notes:
- "last recorded replay" is persisted with sessionStorage
- stopping then starting will make a new replay
- uses some try-catch logic to handle older SDK versions where the recording fxs throw.

Follow-ups before merging
- [x] Analytics context provider and start/stop button analytics (todo in this PR)
- [x] comment on SDK versioning
   - exact release of `getReplay`/public API is unknown, but it was ~2yr ago near the release of the whole product
- [x] test with v8.18
- [x] remove the debug flag
- [x] make the links work for dev mode (can hard-code the sentry-test/app-frontend org/proj)

Follow-ups after merging
- [ ] test the links work in prod
- [ ] account for minimum replay duration (so users don't stop too early)
- [ ] add more content to the panel! Open to ideas. Can do this in a separate PR
- [ ] keep dogfooding for bugs, w/different sample rates, sdk versions


package.json
```diff
@@ -59,13 +59,13 @@
     "@sentry-internal/rrweb-player": "2.25.0",
     "@sentry-internal/rrweb-snapshot": "2.25.0",
     "@sentry/babel-plugin-component-annotate": "^2.16.1",
-    "@sentry/core": "^8.18.0",
-    "@sentry/node": "^8.18.0",
-    "@sentry/react": "^8.18.0",
+    "@sentry/core": "^8.20.0",
```
aliu39, line 62: Needed for most up-to-date behavior of the SDK functions (older version throws, newer logs to debug)

static/app/bootstrap/initializeSdk.tsx
```diff
@@ -175,6 +175,8 @@ export function initializeSdk(config: Config, {routes}: {routes?: Function} = {}
 
       return event;
     },
+
+    debug: true,
```
aliu39, line 179: Need to remove before merging.

static/app/components/devtoolbar/components/panelRouter.tsx
```diff
@@ -44,6 +45,8 @@ export default function PanelRouter() {
           <PanelReleases />
         </AnalyticsProvider>
       );
+    case 'replay':
+      return <PanelReplay />;
```
michellewzhang, line 49: let's add an `<AnalyticsProvider>` wrapper around here so we can track when users press record!

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+export default function ReplayPanel() {
+  const {projectSlug, projectId, projectPlatform, trackAnalytics, SentrySDK} =
+    useConfiguration();
+  const replay =
+    SentrySDK && 'getReplay' in SentrySDK ? SentrySDK.getReplay() : undefined;
+
+  const [replayId, setReplayId] = useState<string | undefined>(() =>
+    replay?.getReplayId()
+  );
```
michellewzhang, line 28:
```suggestion
  const [replayId, setReplayId] = useState<string | undefined>(
    replay?.getReplayId()
  );
```
does this work?

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,126 @@
...
+  // Polls periodically since a replay could be started by sessionSampleRate
+  useEffect(() => {
+    const intervalId = setInterval(
+      () => setReplayId(replay?.getReplayId()),
+      POLL_INTERVAL_MS
+    );
+    return () => clearInterval(intervalId);
```
michellewzhang, line 35:
```suggestion
    return clearInterval(intervalId);
```
same idea

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,126 @@
...
+  );
+  useEffect(() => {
+    if (replayId) {
+      setLastReplayId(replayId);
+    }
+  }, [replayId]);
+
+  return (
+    <PanelLayout title="Session Replay">
+      <Fragment>
```
michellewzhang, line 56: is this `<Fragment>` needed?

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,126 @@
...
+              <SentryAppLink
+                to={{
+                  url: `/replays/${lastReplayId}`,
+                  query: {project: projectId},
+                }}
+                onClick={() => {
+                  trackAnalytics?.({
+                    eventKey: `devtoolbar.current-replay-link.click`,
+                    eventName: `devtoolbar: Current replay link clicked`,
+                  });
```
michellewzhang, line 91: can remove this because we already have `trackAnalytics` being called inside `<SentryAppLink>` already, we just need to set the context name by using `<AnalyticsProvider>`

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,126 @@
...
+          }}
+        >
+          {isRecording
+            ? t('In progress. Click to stop recording')
+            : t('Start recording the current session')}
+        </Button>
+        <div css={[smallCss, panelSectionCss, panelInsetContentCss]}>
+          {lastReplayId ? (
+            <span css={[resetFlexRowCss, {gap: 'var(--space50)'}]}>
+              {isRecording ? t('Current replay: ') : t('Last recorded replay: ')}
```
michellewzhang, line 81: probably don't need `t('...')` here (or anywhere in the devtoolbar code)

yarn.lock
```diff
@@ -7446,6 +7447,16 @@ import-in-the-middle@1.7.1:
     cjs-module-lexer "^1.2.2"
     module-details-from-path "^1.0.3"
 
+import-in-the-middle@^1.10.0:
```
michellewzhang, line 7450: try running `npx yarn-deduplicate && yarn install` and see if this one goes away

aliu39:
I'll be changing up the logic a good bit, since Billy pointed out this version will count a replay in buffering mode as isRecording.

static/app/components/devtoolbar/components/replay/replayPanel.tsx
aliu39: Oops no

static/app/components/devtoolbar/components/replay/replayPanel.tsx
aliu39: useEffect can optionally return a cleanup function

static/app/components/devtoolbar/components/replay/replayPanel.tsx
aliu39: These are equivalent, but passing in a callback causes `getReplayId` to be called on less re-renders. https://medium.com/@jscodelover/using-functions-as-the-initial-state-in-the-usestate-hook-in-react-7069cdfcc6ed

I want this to refresh every time the panel is _first_ rendered (e.g. switching panels). Any thoughts @ryan953 ?

package.json
ryan953: What's different in the new version? We should be building in a way that older SDK versions have a chance to work too, and if not then we should build-in checks and messages so the app doesn't throw when users access it!

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,136 @@
...
+    useConfiguration();
+  const replay =
+    SentrySDK && 'getReplay' in SentrySDK ? SentrySDK.getReplay() : undefined;
+
+  // sessionId is undefined iff we are recording in session OR buffer mode.
+  const [sessionId, setSessionId] = useState<string | undefined>(() =>
+    replay?.getReplayId()
+  );
+  const [recordingMode, setRecordingMode] = useState<ReplayRecordingMode>(
+    () => replay?._replay.recordingMode
```
ryan953, line 31: Could create a helper to get this `_replay` value, that'll help so we only have to squash the lint warning in one place.

To start it'll be something like this: 
```
function useFoo() {
  const {SentrySDK} = useConfiguration();
  const replay =
    SentrySDK && 'getReplay' in SentrySDK ? SentrySDK.getReplay() : undefined;
  return replay?._replay;
```

But we'll probably actually need to return some other information in cases like:
- the user has too old of an SDK, idk what the problem is with 8.18, but v7 is also something they could have
- there is no replayIntegration at all and we want to show a message about that

static/app/components/devtoolbar/components/replay/replayPanel.tsx
ryan953: the useEffect cleanup is needed here. `return () => clearInterval(interfaceId);` is the cleanup. 👍

static/app/components/devtoolbar/components/replay/replayPanel.tsx
ryan953: Two things to disambiguate: `useState()` and `setState()`

`useState()` can accept a value or a callback into the [`initialState` argument](https://react.dev/reference/react/useState#usestate). These examples are the same: `const state = useState('foo')` OR `const state = useState(() => {return 'foo';})`.
The benefit of the 2nd example is that if you're doing something expensive the callback will only be executed on the first render, once the initialState has been set whatever is passed into this argument isn't 'needed' anymore by react. 

`setState()` accepts a `nextState` argument, which can be a value or updater function. [The updater is not always necessary.](https://react.dev/reference/react/useState#is-using-an-updater-always-preferred)

package.json
aliu39: The new version is more "lenient" with errors. For example getReplayId used to throw if no recording session is active - now it returns undefined. `start()` and `stop()` would also throw if called in the wrong state. And `flush()` would start a replay if in session mode.

I'll double check all this with a diff on the SDK code

static/app/components/devtoolbar/components/replay/replayPanel.tsx
aliu39: Gotcha, I think in this case I don't want the `() =>` because the current `replayId` (renamed to `sessionId`) should be as up-to-date as possible on every initial render.

static/app/components/devtoolbar/components/replay/replayPanel.tsx
aliu39: Could the helper accept the `replay` (replay integration) as an argument, instead of calling useConfiguration and getReplay every time?

ryan953: ya, i wasn't trying to be too prescriptive. 

The notes you made about 8.18 needing try/catch too could also factor in here, we should add those!

aliu39: Sounds good, 8.18 will not be disabled but some of the fx calls will probably throw, so I need to handle that..
The button won't be disabled unless the SDK is older than the `getReplay` function. Need to check which version it was released.

aliu39: Got it - just want to see which one it is for myself!

yarn.lock
aliu39: Just ran

aliu39:
Hm actually I think the context is unnecessary.. just need some kind of store for the lastReplayId.

package.json
aliu39: Left a comment in the code. The changes were made in 8.20: https://github.com/getsentry/sentry-javascript/pull/13000

static/app/components/devtoolbar/components/replay/replayPanel.tsx
aliu39: I haven't found when `getReplay()`/the public API was first released. If we need the exact version, maybe @billyvg can help?

static/app/components/devtoolbar/hooks/useReplayRecorder.tsx
```diff
@@ -0,0 +1,118 @@
...
+  isRecording: boolean;
+  lastReplayId: string | undefined;
+  start(): Promise<boolean>;
+  stop(): Promise<boolean>;
+};
+
+function getIsRecording(
+  replay: ReturnType<typeof replayIntegration> | undefined
+): boolean {
+  return replay?._replay.isEnabled() ?? false;
```
ryan953, line 19: probably want one helper `getReplayInternal(replay: ReturnType<typeof replayIntegration> | undefined) { return replay?._replay; }` instead of these three.

The purpose of that single helper would be so there's only one line that needs the lint error silenced.

aliu39: gotcha, that's the type of the public api though, we probably don't have access to the private type

package.json
aliu39: This PR will work without bumping the SDK.

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </ExternalLink>
+    );
+  }
+
+  return (
+    <PanelLayout title="Session Replay">
+      <AnalyticsProvider
+        keyVal={`replay-button-${isRecording ? 'stop' : 'start'}`}
+        nameVal={`replay button ${isRecording ? 'stop' : 'start'}`}
+      >
```
ryan953, line 60: Whether the replay is started or stopped is more of a property, the button is the toggle button
```suggestion
      <AnalyticsProvider
        keyVal="toggle-replay-button"
        nameVal="toggle replay button"
      >
```

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+const TRUNC_ID_LENGTH = 16;
+
+export default function ReplayPanel() {
+  const {projectSlug, projectId, projectPlatform, trackAnalytics} = useConfiguration();
+
+  const {disabledReason, isDisabled, isRecording, lastReplayId, start, stop} =
+    useReplayRecorder();
+
+  function ReplayLink({children}: {children: React.ReactNode}) {
```
ryan953, line 29: move components to the top-level of the file

static/app/components/devtoolbar/hooks/useReplayRecorder.tsx
```diff
@@ -0,0 +1,116 @@
...
+      ? 'Your SDK version is too outdated to use the Replay integration.'
+      : !replay
+        ? "Failed to load your SDK's Replay integration."
+        : undefined;
+
+  const [isRecording, setIsRecording] = useState<boolean>(
+    () => replayInternal?.isEnabled() ?? false
+  );
+  const [lastReplayId, setLastReplayId] = useState<string | undefined>(
+    () => sessionStorage.getItem(LAST_REPLAY_STORAGE_KEY) || undefined
```
aliu39, line 60: Ah, can probably use `useSessionStorage` here

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,111 @@
...
+        to={{
+          url: `/replays/${lastReplayId}`,
+          query: {project: projectId},
+        }}
+      >
+        {children}
+      </SentryAppLink>
+    ) : (
+      <ExternalLink
+        href={`https://sentry-test.sentry.io/replays/${lastReplayId}/?project=5270453`}
```
ryan953, line 42: can you add links to the code that shows why this is needed? i suspect that we're conflating the env with the settings that relate to the env and the real if-statement should be inside [`useDevToolbar.tsx`](https://github.com/getsentry/sentry/blob/c9065b923e634ea79f788d4fce4f43d658590ae6/static/app/utils/useDevToolbar.tsx#L25-L32)

static/app/components/devtoolbar/hooks/useReplayRecorder.tsx
```diff
@@ -0,0 +1,116 @@
...
+  );
```
ryan953, line 61: we have a `useSessionStorage` that can be helpful here

static/app/components/devtoolbar/components/replay/replayPanel.tsx
aliu39: > can you add links to the code that shows why this is needed?

Yeah, I don't have any atm but can take a look

> i suspect that we're conflating the env with the settings that relate to the env and the real if-statement should be inside [useDevToolbar.tsx](https://github.com/getsentry/sentry/blob/c9065b923e634ea79f788d4fce4f43d658590ae6/static/app/utils/useDevToolbar.tsx#L25-L32)

What do you mean?

static/app/components/devtoolbar/hooks/useReplayRecorder.tsx
```diff
@@ -0,0 +1,116 @@
...
+    let success = false;
+    if (replay && isRecording) {
+      try {
+        await replay.stop();
+        success = true;
+        // eslint-disable-next-line no-empty
+      } catch {}
+    }
+    refreshState();
+    return success;
```
ryan953, line 105: Could be a good candidate for a [`finally` block](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#the_finally_block)

static/app/components/devtoolbar/components/replay/replayPanel.tsx
ryan953: `useDevToolbar.tsx` hard-coded the project name and id's and stuff. If the DSN is configured then the orgSlug, projectSlug, etc should be updated to match

aliu39: Gotcha, I can move the `process.env.NODE_ENV === 'production'` check to useDevToolbar then and hard code the test org + project.

Also, to get the integration locally, I needed to comment out a line in getsentry's useReplayInit. We need to look at that file too, before sentry employees can use this

package.json
billyvg: It's generally a good idea to keep them as separate PRs

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const {disabledReason, isDisabled, isRecording, lastReplayId, start, stop} =
+    useReplayRecorder();
+
+  const {eventName, eventKey} = useContext(AnalyticsContext);
+  const [buttonLoading, setButtonLoading] = useState(false);
+  return (
+    <PanelLayout title="Session Replay">
+      <Button
+        size="sm"
+        icon={isRecording ? <IconPause /> : <IconPlay />}
```
billyvg, line 34: Would stop button be better?

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,108 @@
...
+            trackAnalytics?.({
+              eventKey: eventKey + `${type}-button-failed`,
+              eventName: eventName + `${type} button failed`,
+            });
+          }
+        }}
+      >
+        {isRecording
+          ? 'In progress. Click to stop recording'
+          : 'Start recording the current session'}
```
billyvg, line 56: There's potentially a third state where it's buffering

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,108 @@
...
+          if (!success) {
+            trackAnalytics?.({
+              eventKey: eventKey + `${type}-button-failed`,
+              eventName: eventName + `${type} button failed`,
+            });
+          }
+        }}
+      >
+        {isRecording
+          ? 'In progress. Click to stop recording'
```
billyvg, line 55:
```suggestion
          ? 'Recording in progress, click to stop'
```

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,108 @@
...
+    </PanelLayout>
+  );
+}
+
+function ReplayLink({lastReplayId}: {lastReplayId: string}) {
+  const {projectSlug, projectId, projectPlatform} = useConfiguration();
+  return (
+    <SentryAppLink
+      to={{
+        url: `/replays/${lastReplayId}`,
```
billyvg, line 79: Our URLs always have trailing slashes

```suggestion
        url: `/replays/${lastReplayId}/`,
```

static/app/components/devtoolbar/hooks/useReplayRecorder.tsx
```diff
@@ -0,0 +1,117 @@
...
+  // @ts-ignore:next-line
+  return replay._replay;
+}
+
+const LAST_REPLAY_STORAGE_KEY = 'devtoolbar.last_replay_id';
+
+export default function useReplayRecorder(): ReplayRecorderState {
+  const {SentrySDK} = useConfiguration();
+  const replay =
+    SentrySDK && 'getReplay' in SentrySDK ? SentrySDK.getReplay() : undefined;
```
billyvg, line 37: you can use `SentrySDK.getIntegrationByName()`, it's just a bit more annoying dealing with types

static/app/utils/useDevToolbar.tsx
```diff
@@ -25,10 +25,12 @@ export default function useDevToolbar({enabled}: {enabled: boolean}) {
...
         environment: ['prod'],
-        organizationSlug: 'sentry',
-        projectId: 11276,
+        organizationSlug:
+          process.env.NODE_ENV !== 'development' ? 'sentry' : 'sentry-test',
+        projectId: process.env.NODE_ENV !== 'development' ? 11276 : 5270453,
         projectPlatform: 'javascript',
-        projectSlug: 'javascript',
+        projectSlug:
+          process.env.NODE_ENV !== 'development' ? 'javascript' : 'app-frontend',
```
billyvg, line 33: Should this be in here?

static/app/components/devtoolbar/components/replay/replayPanel.tsx
aliu39: I couldn't find one..

static/app/components/devtoolbar/hooks/useReplayRecorder.tsx
aliu39: I'd like it to cover the outer condition too

static/app/components/devtoolbar/components/replay/replayPanel.tsx
aliu39: Should we surface this to users? Whether it's buffering or not

static/app/utils/useDevToolbar.tsx
aliu39: While this is PoC we are hard coding the org/project slugs, this allows the links to work in dev mode (assuming the replay integration is enabled)

static/app/components/devtoolbar/hooks/useReplayRecorder.tsx
aliu39: Hm my VSCode can't find this method in browser or react packages..

static/app/components/devtoolbar/hooks/useReplayRecorder.tsx
ryan953:
```suggestion
try {
  if (replay && isRecording) {
    await replay.stop();
    return true;
  }
  return false;
} catch {
  return false;
} finally {
  refreshState();
}
```

static/app/utils/useDevToolbar.tsx
ryan953: can you test with `dev-ui`. my workflow depends on that command

aliu39: It works, but you need to comment out this condition in useReplayInit: https://github.com/getsentry/getsentry/blob/2a1da081f3a9e4e4111b577d5551fa24691da374/static/getsentry/gsApp/utils/useReplayInit.tsx#L85-L87

I think it's disabled so we don't record everyone's dev-ui sessions. Maybe we could uncomment but set both sample rates to 0?

aliu39: I'll make a ticket and add it to the dev toolbar project

static/app/components/devtoolbar/hooks/useReplayRecorder.tsx
aliu39: Got it, does this have any improvements over the current approach? Right now I'm using these conditions to get the `disabledReason` (see screenshots I just added in the description)

billyvg: Ah just as a fallback if the SDK is too old to support `getReplay` as it was a relatively recent addition, but it's fine not to support older

static/app/utils/useDevToolbar.tsx
ryan953: probably hold off on a ticket to change `useReplayInit`. The toolbar should support whatever env it finds itself inside of. So if sometimes there's no `replayIntegration` at all, that should be ok. We'll be able to test that case in dev, and on prod there will be a `replayIntegration` so we can test that case as well!

aliu39: I wasn't thinking about the other panels, this change would break them for dev. I'm done testing, let's just leave this panel disabled in dev.

static/app/components/devtoolbar/components/replay/replayPanel.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const [buttonLoading, setButtonLoading] = useState(false);
+  return (
+    <PanelLayout title="Session Replay">
+      <Button
+        size="sm"
+        icon={isDisabled ? undefined : isRecordingSession ? <IconPause /> : <IconPlay />}
+        disabled={isDisabled || buttonLoading}
+        onClick={async () => {
+          setButtonLoading(true);
+          isRecordingSession ? await stopRecording() : await startRecordingSession();
```
aliu39, line 46: Not allowing users to stop buffering.

static/app/components/devtoolbar/hooks/useReplayRecorder.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  const refreshState = useCallback(() => {
+    setIsRecording(replayInternal?.isEnabled() ?? false);
+    setSessionId(replayInternal?.getSessionId());
+    setRecordingMode(replayInternal?.recordingMode);
+  }, [replayInternal]);
+
+  const startRecordingSession = useCallback(async () => {
+    let success = false;
+    if (replay) {
```
aliu39, line 79: Don't want to try-catch here because we can already condition to make sure we're in a safe state.

billyvg (approved):

PR #75408, opened 2024-07-31
feat(prevent-superuser-acces): Create Prevent Superuser Access Service

iamrajjoshi:
Made a Lazy Wrapper Service to Implement Prevent Superuser Access logic. The logic will live in `getsentry` gated by a plan, so the service will return  `True` when `should_allow_superuser_access` is called in `sentry`. 

There will be a separate PR in `getsentry` which actually implements the business logic. 

https://github.com/getsentry/getsentry/pull/14779 for more context

### DIFF
--- a/src/sentry/superuser_access/prevent_superuser_access_backend.py
+++ b/src/sentry/superuser_access/prevent_superuser_access_backend.py
@@ -1,0 +1,19 @@
+from sentry.models.organization import Organization
+from sentry.organizations.services.organization import RpcUserOrganizationContext
+from sentry.utils.services import Service
+
+
+class PreventSuperuserAccessBackend(Service):
+    def should_prevent_superuser_access(
+        self,
+        organization_context: Organization | RpcUserOrganizationContext,
+    ) -> bool:
+        raise NotImplementedError
+
+
+class PreventSuperuserAccess(Service):
+    def should_prevent_superuser_access(
+        self,
+        organization_context: Organization | RpcUserOrganizationContext,
+    ) -> bool:
+        return False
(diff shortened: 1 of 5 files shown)


src/sentry/superuser_access/prevent_superuser_access_backend.py
```diff
@@ -0,0 +1,19 @@
...
+class PreventSuperuserAccessBackend(Service):
+    def should_allow_superuser_access(
+        self,
+        organization_context: Organization | RpcUserOrganizationContext,
+    ) -> bool:
+        raise NotImplementedError
+
+
+class PreventSuperuserAccess(Service):
+    def should_allow_superuser_access(
```
markstory, line 15: It would be good to have a docstring on why there are two classes. You could skip those docs by having `PreventSuperuserAccess(PreventSuperuserAccessBackend)` which makes the relation between the classes clear.

Having a docstring explaining that this is a saas only component used in getsentry could be helpful for future us as well.

src/sentry/superuser_access/prevent_superuser_access_backend.py
```diff
@@ -2,6 +2,13 @@
 from sentry.organizations.services.organization import RpcUserOrganizationContext
 from sentry.utils.services import Service
 
+"""
+This class defines the interface for a backend that determines whether a superuser should be allowed access to an organization.
+The backend should implement the `should_allow_superuser_access` method, which takes an organization context and returns a boolean.
+This is a SAAS-only feature, so the default implementation should always return True.
+getsentry contains buisness logic.
```
markstory, line 9:
```suggestion
getsentry contains business logic.
```

markstory (approved):

PR #75409, opened 2024-08-01
fix(endpoints): Fix requests with invalid statuses leading to errors

Christinarlong:
Prev changes led to some integrations that set status to an invalid value to be an empty queryset leading to errors for certain providers (i.e terraform-provider)

See: https://github.com/jianyuan/terraform-provider-sentry/issues/470

### DIFF
--- a/src/sentry/integrations/api/endpoints/organization_repositories.py
+++ b/src/sentry/integrations/api/endpoints/organization_repositories.py
@@ -85,7 +85,7 @@ class OrganizationRepositoriesEndpoint(OrganizationEndpoint):
 
             return Response(serialize(repos, request.user))
 
-        else:
+        elif status:
             queryset = queryset.none()
 
         return self.paginate(


asottile-sentry:
can you add a test covering this case?

benlangfeld:
The change which broke this would be https://github.com/getsentry/sentry/commit/e2fa60a843ceec6c2eaf662ec706f6f0bad76242, correct?

benlangfeld:
@asottile-sentry Test coverage added in https://github.com/getsentry/sentry/pull/75455. I would prefer not to go through setting up a local development environment to validate the test works, so would appreciate if you would label that PR appropriately such that CI will run the tests.

Christinarlong:
> can you add a test covering this case?

Sorry was out for doctor's this morning, closing this as https://github.com/getsentry/sentry/pull/75455. includes both

PR #75414, opened 2024-08-01
ref(stats): Expose org quota reason

priscilawebdev:
### Goal

On the stats chart, we currently display the 'key_quota' reason, which represents the rate limit defined for individual projects, but we do not display the 'org_quota' reason, which is the rate limit for the entire organization. Currently, the "org_quota" reason is shown as "internal," which is confusing for users. This PR addresses the issue.

### DIFF
--- a/static/app/views/organizationStats/getReasonGroupName.ts
+++ b/static/app/views/organizationStats/getReasonGroupName.ts
@@ -140,7 +140,7 @@ function getRateLimitedReasonGroupName(reason: RateLimitedReason | string): stri
     case RateLimitedReason.PROJECT_QUOTA:
       return 'global limit';
     case RateLimitedReason.KEY_QUOTA:
-      return 'DSN limit"';
+      return 'DSN limit';
     case RateLimitedReason.SPIKE_PROTECTION:
     case RateLimitedReason.SMART_RATE_LIMIT:
       return 'spike protection';
--- a/static/app/views/organizationStats/getReasonGroupName.ts
+++ b/static/app/views/organizationStats/getReasonGroupName.ts
@@ -72,6 +72,7 @@ enum ClientDiscardReason {
 }
 
 enum RateLimitedReason {
+  ORG_QUOTA = 'org_quota',
   KEY_QUOTA = 'key_quota',
   SPIKE_PROTECTION = 'spike_protection',
   SMART_RATE_LIMIT = 'smart_rate_limit',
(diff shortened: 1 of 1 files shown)


static/app/views/organizationStats/getReasonGroupName.ts
```diff
@@ -130,6 +131,8 @@ function getRateLimitedReasonGroupName(reason: RateLimitedReason | string): stri
   }
 
   switch (reason) {
+    case RateLimitedReason.ORG_QUOTA:
+      return 'org quota';
```
priscilawebdev, line 135: @jan-auer what do we want to call it? I was not sure if we want to differentiate or display it as 'quota' as we do in line 129 https://github.com/getsentry/sentry/blob/3efa8cf306b9650d8129ffeaca77601b833b1009/static/app/views/organizationStats/getReasonGroupName.ts#L128-L130

jan-auer: With the word "quota", in the product we usually refer to the monthly quota that the user pays for. It may be a bit inconsistent, that the word "quota" is used for this internal reason, but we should refer to it as a (rate) limit. 

I took another look at all rate limit reasons and would suggest the following:
- `endsWith('_usage_exceeded')` -> "quota"
- `endsWith('_disabled')` -> "disabled"
- `org_quota`, `project_quota` -> "global limit" (in docs explain that this is an old setting available to some orgs to configure rate limits on organization- and project level)
- `key_quota` -> "DSN limit" (better than "key limit")
- `spike_protection`, `smart_rate_limit` -> "spike protection"
- all other can remain "internal", these are abuse limits and when the organization is disabled.

Potentially we could think of a better description than "disabled"

priscilawebdev: @jan-auer can you please give me a description for disabled? I will update the docs

priscilawebdev: here is the [PR](https://github.com/getsentry/sentry-docs/pull/10947) for the docs

static/app/views/organizationStats/getReasonGroupName.ts
```diff
@@ -129,9 +131,16 @@ function getRateLimitedReasonGroupName(reason: RateLimitedReason | string): stri
...
+    return 'disabled';
+  }
+
   switch (reason) {
+    case RateLimitedReason.ORG_QUOTA:
+    case RateLimitedReason.PROJECT_QUOTA:
+      return 'global limit';
     case RateLimitedReason.KEY_QUOTA:
-      return 'key limit';
+      return 'DSN limit"';
```
jan-auer, line 143:
```suggestion
      return 'DSN limit';
```

jan-auer (approved):

static/app/views/organizationStats/getReasonGroupName.ts
priscilawebdev: ohh good catch ...thanks!

PR #75415, opened 2024-08-01
chore(metrics) Clean up of `useMetricsTags`

vgrozdanic:
Clean up of `useMetricsTags` since API endpoint is not even using the parameters that we are sending: https://github.com/getsentry/sentry/blob/master/src/sentry/api/endpoints/organization_metrics_tags.py


priscilawebdev (approved):
lgtm

Issue #75120, opened 2024-07-29
Insights - Mobile Vitals


markushi:
A mobile module providing an entry point, top level metrics and the ability to drill deeper into app starts, screen loads and UI jank.

Internal [notion doc for feedback](https://www.notion.so/sentry/Mobile-Vitals-Feedback-0e136e5cd0784c3192beeae3dd4db085) here.

![Image](https://github.com/user-attachments/assets/71922621-a4bc-495a-a215-10c1133846db)

```[tasklist]
### Tasks
- [ ] Add tooltips for table columns to clearly explain what they represent.
- [ ] Switch to slow and frozen frame rates
- [ ] Add platform switcher for RN projects
- [ ] Remove screens label
- [ ] Sub-Modules: Fix padding
- [ ] Sub-Modules: Have proper routing
- [ ] Fix missing unit for frame delay
- [ ] Add release selectors
- [ ] Update docs / sales
```

kahest:
possible GA/post GA item: https://github.com/getsentry/sentry/issues/77560

kahest:
we aligned internally that we will expand the scope and adapt the timeline a little:
- include screen rendering in EA/GA (currently LA)
- rename to Mobile Vitals
- public comms push (blog post, email outreach)
- SE training

updated timeline:
 - EA: Early Feb
 - GA: Mid-late Feb

markushi:
There's now a new project for post GA topics and tasks: https://github.com/getsentry/projects/issues/703

kahest:
GA rollout currently at 50%

markushi:
Rollout is now at 100%, Docs have been merged as well.

Issue #75318, opened 2024-07-31
Switch to ratio for frame delay (frame_delay / span.duration)


markushi:

philippsuess:
Closing because this is out of date. If it's still relevant, we should open a new issue with up-to-date context.

PR #75427, opened 2024-08-01
feat(insights): Add Mobile Screens insights module

markushi:
Implements basis for https://github.com/getsentry/sentry/issues/75120


static/app/views/insights/mobile/vitals/components/vitalScreens.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const orderby = decodeScalar(locationQuery.sort, '-count');
+  const newQuery: NewQuery = {
+    name: '',
+    fields: [
+      SpanMetricsField.PROJECT_ID,
+      'transaction',
+      `count()`,
+      `avg(mobile.slow_frames)`,
+      `avg(mobile.frozen_frames)`,
+      `avg(mobile.frames_delay)`,
```
narsaynorath, line 53: Considering the sub-screens have release filters applied, should this page likewise query for the avg by release? Maybe not for this PR, but I think maybe drilling down into values and seeing the release filter appear (if it does, I wasn't able to test this on the `sentry-sdks` org yet) might be misleading to the user.

narsaynorath: Also in the future, are we more concerned about the slow frame _rate_ instead of the specific value? I know now the rate is more important because it normalizes the value across transaction durations

static/app/views/insights/mobile/vitals/components/vitalsScreensTable.tsx
```diff
@@ -0,0 +1,106 @@
...
+    if (field === 'transaction') {
+      return (
+        <Fragment>
+          <TopResultsIndicator count={TOP_SCREENS} index={index} />
+          <Link
+            to={`${moduleURL}/screens/?${qs.stringify({
+              ...location.query,
+              project: row['project.id'],
+              transaction: row.transaction,
+            })}`}
```
narsaynorath, line 54: Because of hybrid cloud, we wrap these in `normalizeUrl`, would you be able to add this here and confirm it maintains the same behaviour?

It might not be as applicable now that we've done most of the migration to the `org.sentry.io` format but afaik we should still be wrapping URLs with it!

narsaynorath (approved):
Just some small comments but nothing blocking. This is looking great I'm excited to see this stuff brought together in a more cohesive flow :pray:

static/app/views/insights/mobile/vitals/components/vitalScreens.tsx
markushi: Yes, fully agree - we should switch to frame rates. sub-task here: https://github.com/getsentry/sentry/issues/75318

static/app/components/sidebar/index.spec.tsx
```diff
@@ -396,6 +396,25 @@ describe('Sidebar', function () {
...
+
+      await waitFor(function () {
+        expect(apiMocks.broadcasts).toHaveBeenCalled();
+      });
+
+      ['App Starts', 'Screen Loads', /Mobile UI/].forEach(title => {
+        expect(screen.queryByText(title)).not.toBeInTheDocument();
+      });
+
+      expect(screen.queryByText(/Mobile Vitals/)).toBeInTheDocument();
```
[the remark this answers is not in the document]
markushi: 🙏 thanks, feel free to be very picky about stuff like this - the code I'm contributing should match the existing style!

static/app/views/insights/mobile/screens/components/screensOverview.tsx
```diff
@@ -0,0 +1,235 @@
...
+    sortedBy,
+    true,
+    []
+  );
+
+  const {
+    data: secondaryData,
+    // isLoading: secondaryLoading,
+    // pageLinks: secondaryLinks,
+    // eventView: secondaryEventView,
```
narsaynorath, line 137: Not sure if you wanted to keep these around for the future, but if not we could just remove them

static/app/views/insights/mobile/screens/components/screensOverview.tsx
```diff
@@ -0,0 +1,235 @@
...
+      });
+      setVisibleScreens(screens);
+    }
+  }, [primaryData]);
+
+  const tableSearchFilters = new MutableSearch(['transaction.op:ui.load']);
+  const derivedQuery = getTransactionSearchQuery(location, primaryEventView.query);
+
+  const combinedData = useMemo((): TableData | undefined => {
+    if (primaryData !== undefined && secondaryData !== undefined) {
```
narsaynorath, line 165: nit: We have a `defined` helper that makes this kind of check look a little neater

static/app/views/insights/mobile/screens/components/screensOverview.tsx
```diff
@@ -0,0 +1,235 @@
...
+            ...row,
+          };
+        }
+        return row;
+      });
+
+      return {
+        data: data,
+        meta: meta,
+      };
```
narsaynorath, line 192: I'm surprised this didn't auto-format but if the key and value are the same, you can just write it like this:

```suggestion
      return {
        data,
        meta,
      };
```

static/app/views/insights/mobile/screens/components/screensOverview.tsx
```diff
@@ -0,0 +1,235 @@
...
+        organization={organization}
+        query={getFreeTextFromQuery(derivedQuery)}
+        placeholder={t('Search for Screen')}
+        additionalConditions={new MutableSearch(tableSearchFilters.formatString())}
+      />
+      <Container>
+        <ScreensOverviewTable
+          eventView={primaryEventView}
+          data={combinedData}
+          isLoading={loading}
```
narsaynorath, line 223: Looks like this is the only spot where `loading` is used, so maybe we should just inline it if it's strictly `primaryLoading`. Just wondering, we don't want to wait for the results of both queries (i.e. `primaryLoading || secondaryLoading`)? The combined data makes me think we should wait for both requests to finish

static/app/views/insights/mobile/screens/components/screensOverview.tsx
```diff
@@ -0,0 +1,235 @@
...
+              ...location.query,
+              cursor: undefined,
+              query: String(search).trim() || undefined,
+            },
+          });
+        }}
+        organization={organization}
+        query={getFreeTextFromQuery(derivedQuery)}
+        placeholder={t('Search for Screen')}
+        additionalConditions={new MutableSearch(tableSearchFilters.formatString())}
```
narsaynorath, line 217: Is it possible to just pass along `tableSearchFilters` here? I think it's of type `MutableSearch` so we could avoid formatting it to a string to re-instantiate it

narsaynorath (approved):
I re-reviewed the portion of code related to the multiple queries for the screens table and it looks good to me. Just some small nits and a question about the loading state prop! 🔥

PR #75429, opened 2024-08-01
fix(metrics) Show tooltip left when tag is disabled

### LINKED ISSUE #75430: High cardinality tooltip position
Image

vgrozdanic:
Moves tooltip left (as it is in component where we use it)

Closes https://github.com/getsentry/sentry/issues/75430

Before
image

After
image


vgrozdanic:
Closing for now, something is off with the diff

PR #75434, opened 2024-08-01
ref(selectoption): Tweak list items alignment

priscilawebdev:
**Before:**

When no item is selected, the alignment of the items looks odd — they are shifted to the right

image
image

**After:**

image
image
image


static/app/components/forms/controls/selectOption.tsx
```diff
@@ -56,17 +56,28 @@ function SelectOption(props: Props) {
...
-                {isSelected && (
-                  <IconCheckmark
-                    size={isMultiple ? 'xs' : 'sm'}
-                    color={isMultiple ? 'white' : undefined}
-                  />
-                )}
-              </CheckWrap>
-              {data.leadingItems}
-            </Fragment>
+            itemProps.__isNew__ ? (
```
priscilawebdev, line 59: I’m not fond of using `__isNew__`, as it doesn’t feel quite right, but I don’t have a better alternative at the moment. For what it’s worth, newer versions of react-select still use it. You can see an example here: [react-select GitHub](https://github.com/JedWatson/react-select/blob/06e34882638d1526b9f5a1238bb567a3e9460ce5/packages/react-select/src/useCreatable.ts#L98-L102).

priscilawebdev: Plus if it breaks, the maximum that can happen is that our UI won't render the "+" icon

dashed (approved):

vuluongj20:
I think the current version looks normal. It's that way due to a combination of 2 design constraints. The first is establishing and maintaining minimal [keylines](https://m1.material.io/layout/metrics-keylines.html#metrics-keylines-keylines-spacing). The new version breaks that by introducing a second keyline to the menu, making the menu less tidy and skimmable:
image
The second is avoiding layout/location shifts. The new version also breaks that by shifting the selected element's x-position when it becomes selected. 
A hypothetical, alternative solution could be: no left padding when no element is selected, but left paddings for all elements when at least one element is selected. This solution would satisfy the first constraint about key lines, but still fail the second.
That's why we ended up with the current version, which again I don't think is abnormal.

priscilawebdev:
@vuluongj20  thanks for your review! I wasn’t aware that we were following the Material Design guidelines, so it was great to read through them. I understand now why there’s empty space on the left of the menu when nothing is selected. However, it still feels a bit strange to have that space.

what about we display the icon next to an empty row as the following:


https://github.com/user-attachments/assets/d40139d7-1aef-48e9-a3da-f91a3790d288

I also would be interesting to know you opinion on the '+' icon next to the 'Create' word

vuluongj20:
@priscilawebdev Hmm I'd be supportive of that if we also added a label "None"/"None selected" label:
image

priscilawebdev:
@vuluongj20 what do you think? 

https://github.com/user-attachments/assets/020daf99-f915-4355-a0fb-92808eac7011

ArthurKnaus:
@priscilawebdev If we go with an additional option for `no value`, I would render it as disabled.

vuluongj20:
@priscilawebdev I'd prefer "None selected"/"None" over "Select span attribute". Having "None" as a the first makes it extra clear what it stands for. "Select span attribute" sounds like a call to action rather than a value state.

vuluongj20:
Looks great otherwise!

priscilawebdev:
@ArthurKnaus @vuluongj20 Thank you for the reviews 🙏 I have applied your suggestions! 

https://github.com/user-attachments/assets/6814d745-50ab-4b79-9d91-bbe8a1febbfd

ArthurKnaus (approved):

vuluongj20 (approved):

PR #75438, opened 2024-08-01
Fix(metrics) Filter autocomplete fix

### LINKED ISSUE #75437: Environment is not searchable in autocomplete

vgrozdanic:
Autocomplete wasn't showing all the data it could have because of the bug in one custom hook.
Also some fields were excluded by default (e.g. environment) which is now also fixed.

Closes https://github.com/getsentry/sentry/issues/75437

Before
image

Now
image

### DIFF
--- a/static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
+++ b/static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
@@ -526,6 +526,7 @@ export function MetricsExtractionRuleForm({
                               placeholder={t('Add span attributes')}
                               organization={organization}
                               supportedTags={supportedTags}
+                              excludedTags={[]}
                               dataset={DiscoverDatasets.SPANS_INDEXED}
                               projectIds={[Number(projectId)]}
                               hasRecentSearches={false}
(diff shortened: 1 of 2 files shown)


static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
```diff
@@ -526,6 +526,7 @@ export function MetricsExtractionRuleForm({
                               placeholder={t('Add span attributes')}
                               organization={organization}
                               supportedTags={supportedTags}
+                              excludedTags={[]}
```
priscilawebdev, line 529: do we need to inform an empty array here?

priscilawebdev:
love that you added 'before' and 'after' screenshots <3

static/app/views/settings/projectMetrics/metricsExtractionRuleForm.tsx
vgrozdanic: Yes, to override the default from the component. E.g. component filters out "environment" tag by default, and we don't want that behaviour

mjq (approved):

priscilawebdev:
it is not related to your PR but the search feels slow + when I add a '.', for example, 'browser.name' ...the "." position looks odd 

https://github.com/user-attachments/assets/ec0bbc81-6cea-4317-a147-2bd6c1084d41

priscilawebdev (approved):
lgtm

PR #75441, opened 2024-08-01
fix(buffer): Do not propagate traces for batched incr

jan-auer:
Stops propagation of traces from `buffer.process_pending` down to
`buffer.process_incr`. The first task tends to spawn thousands of sub
tasks, which creates enormeous traces. These traces are not useful to
debug, since their connection has no meaning and they are only sent
sporadically.

In theory, direct use of `buffer.incr` should not have these issues.
However, in getsentry it is also used in hot loops in the outcomes
consumer, so we also update this instance.

SDK Documentation: https://docs.sentry.io/platforms/python/integrations/celery/#distributed-traces

### DIFF
--- a/src/sentry/buffer/base.py
+++ b/src/sentry/buffer/base.py
@@ -102,7 +102,8 @@ class Buffer(Service):
                 "filters": filters,
                 "extra": extra,
                 "signal_only": signal_only,
-            }
+            },
+            headers={"sentry-propagate-traces": False},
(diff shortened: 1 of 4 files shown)


jan-auer:
cc @jjbayer @Dav1dde

jjbayer (approved):

antonpirker (approved):

PR #75455, opened 2024-08-01
fix(endpoints): Fetching organisation repos regardless of status

benlangfeld:
Previous to https://github.com/getsentry/sentry/commit/e2fa60a843ceec6c2eaf662ec706f6f0bad76242, the behaviour of the API was that sending `?status=` would return all entities regardless of their status. This restores that behaviour.

See https://github.com/getsentry/sentry/pull/75409 where the actual fix comes from. My contribution here is adding test coverage.

closes #75409 

### Legal Boilerplate


asottile-sentry:
technically your new test already passed (since it's part of the normal `sentry` suite and not the private `getsentry` suite -- thanks for the patch!

asottile-sentry (approved):

PR #75456, opened 2024-08-01
ref(alerts): Correct metric alert tooltip text

davidenwang:
Match the text in the backend here:

https://github.com/getsentry/sentry/blob/5b00015693fdc5dada4f7916853e04bf560c2139/src/sentry/incidents/models/incident.py#L168

depends on: https://github.com/getsentry/sentry/pull/75035


ceorourke (approved):
😬

PR #75459, opened 2024-08-01
feat(issue-details): Add dropdown for participants and viewers

roggenkemper:
this pr adds a dropdown to view all participants and viewers for an issue. the design is slightly different than what it was before when it was in the sidebar. now, it is an overlay since this information is in the group header.


static/app/components/group/streamlinedParticipantList.tsx
```diff
@@ -0,0 +1,113 @@
...
+import useOverlay from 'sentry/utils/useOverlay';
+
+interface DropdownListProps {
+  users: User[];
+  teams?: Team[];
+}
+
+export default function ParticipantList({users, teams}: DropdownListProps) {
+  const {overlayProps, isOpen, triggerProps} = useOverlay({
+    position: 'bottom-start',
```
malwilley, line 21: I'd add `isKeyboardDismissable: true` and `shouldCloseOnBlur: true` here, it seems a bit too sticky as is

static/app/components/group/streamlinedParticipantList.tsx
```diff
@@ -0,0 +1,113 @@
...
+  });
+
+  const theme = useTheme();
+  const showHeaders = users.length > 0 && teams && teams.length > 0;
+
+  return (
+    <div>
+      <div {...triggerProps}>
```
malwilley, line 29: This should be a button, it should be focusable

static/app/components/group/streamlinedParticipantList.tsx
```diff
@@ -0,0 +1,113 @@
...
+        <StyledAvatarList users={users} avatarSize={24} maxVisibleAvatars={3} />
+      </div>
+      {isOpen && (
+        <FocusScope contain restoreFocus autoFocus>
```
malwilley, line 33: I don't think this will actually do anything since there are no focusable elements in the menu afaict

malwilley (commented):
Not sure if this is purposeful or not, but looks a bit weird when there is nobody:

CleanShot 2024-08-02 at 10 07 08

roggenkemper:
> Not sure if this is purposeful or not, but looks a bit weird when there is nobody:
> 
> CleanShot 2024-08-02 at 10 07 08

good catch - this was a strange bug, there's already code to prevent it from showing up if there's nobody, but i wasn't passing `teams` into the avatarlist so in the case that there was only a team participant, the avatarlist wouldn't show while the dropdown was still there

malwilley (approved):

PR #75460, opened 2024-08-01
ref(issue-details): Reimplements summary icons

leeandher:
Re-implements the summary icons in a more concise matter and hiding unknown values, with new design.

I doubt there'd ever be that many, but in case the page is constrained, it overflows into a scroll with a gradient to imply the extra are still there.

**Note: Repeated to show scroll, only displays one of each**
Screenshot 2024-08-08 at 1 21 00 PM

**When no icons are interpreted**
Screenshot 2024-08-08 at 1 22 31 PM


static/app/components/events/contexts/contextIcon.tsx
```diff
@@ -151,9 +150,6 @@ function ContextIcon({name, size: providedSize = 'xl', hideUnknown = false}: Pro
   const extraCass = isDarkmode && INVERT_IN_DARKMODE.includes(name) ? darkCss : null;
 
   const imageName = getLogoImage(name);
-  if (hideUnknown && imageName === logoUnknown) {
-    return null;
-  }
```
leeandher, line 156: Removing the `hideUnknown` param because the caller will always receive a <ContextIcon />, it will just render null if specified. If the caller needs to know whether or not the icon will render null, it can just check it manually with `getLogoImage` like I did in `getContextIcon`

static/app/components/events/contexts/utils.tsx
```diff
@@ -382,17 +385,31 @@ export function getContextMeta(event: Event, contextType: string): Record<string
   }
 }
 
+const iconSizesNumber = {
```
roggenkemper, line 388: is there a reason we are redefining iconSizes here instead of using `iconSizes` from `utils/theme.tsx`

leeandher: There's probably a better way to do this, but it's since the ones in theme are suffixed with `px` and the `UserAvatar` component only accepts numbers, but now that I say that aloud I'll just change the `UserAvatar` component

roggenkemper (approved):

PR #22800, opened 2020-12-18
feat(alerts): Add new scopes for editing/viewing alert rules

davidenwang:
This PR adds alerts permission group with `alerts:read` and `alerts:write`. These scopes are added to all roles within the app except owners and also added to the relevant issue alert and metric alert endpoints. Eventually we will have a setting to turn on `alerts:write` for owners

There are also some frontend changes to enable buttons which were previously behind `projects:write`

PR #75464, opened 2024-08-01
nit(scopes): Remove nonexistent alert scopes

schew2381:
`alert_rule:read/write` were originally added in https://github.com/getsentry/sentry/pull/22800, but don't exist anywhere else in sentry or getsentry.

They are replaced be `alerts:read/write`


davidenwang (approved):
good stuff

PR #75465, opened 2024-08-01
fix(insights): cache page doesn't load with long transactions

### LINKED ISSUE #74554: Cannot load Caches page

DominikB2014:
Fixes https://github.com/getsentry/sentry/issues/74554 (the first issue mentioned is fixed already, this fixes the second page)

In cache landing page we fetch the transaction durations for the 20 transactions that appear. The query looks something like `query=transaction:[transaction1, transaction2, ....]`. If the transaction names are really long a 502 will occur on the request originating from ngnix. This occurs because the response headers contains the next page url for pagination, and with large transaction names, the header was too large for the nginx request buffer.

We actually don't need pagination at all for this request, as we always return 20 results, therefore we can add `noPagination` to the query parms and not worry about the nginx buffer at all!


static/app/views/insights/common/queries/useDiscover.ts
```diff
@@ -19,8 +19,10 @@ interface UseMetricsOptions<Fields> {
   enabled?: boolean;
   fields?: Fields;
   limit?: number;
+  noPagination?: boolean;
   pageFilters?: PageFilters;
-  search?: MutableSearch | string; // TODO - ideally this probably would be only `Mutable Search`, but it doesn't handle some situations well
+  search?: MutableSearch | string;
+  // TODO - ideally this probably would be only `Mutable Search`, but it doesn't handle some situations well
```
gggritso, line 25: Stray whitespace change?

gggritso (approved):
Right on!

static/app/views/insights/common/queries/useDiscover.ts
DominikB2014: oops, fixed!

PR #75466, opened 2024-08-01
ci: replace setup-volta with setup-node

joshuarli:
setup-node now supports .volta.json and always good to have something simpler + more supported


joshuarli:
44s cold https://github.com/getsentry/sentry/actions/runs/10203836276/job/28231006759
11s hawt https://github.com/getsentry/sentry/actions/runs/10203836276/job/28231113319

joshuarli:
ok now, 42s cold and 8s hot: https://github.com/getsentry/sentry/actions/runs/10203974732/job/28231483278

.github/workflows/openapi.yml
```diff
@@ -42,8 +42,10 @@ jobs:
           path: sentry-api-schema
           token: ${{ steps.getsentry.outputs.token }}
 
-      - uses: getsentry/action-setup-volta@e4939d337b83760d13a9d7030a6f68c9d0ee7581 # v2.0.0
-        if: steps.changes.outputs.api_docs == 'true'
```
asottile-sentry, line 46: this is missing here

.github/workflows/sync-labels.yml
```diff
@@ -22,7 +22,11 @@ jobs:
       pull-requests: write
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
-      - uses: getsentry/action-setup-volta@e4939d337b83760d13a9d7030a6f68c9d0ee7581 # v2.0.0
+
+      - uses: actions/setup-node@v4
```
asottile-sentry, line 26: we should freeze the versions

.github/workflows/self-hosted.yml
```diff
@@ -49,7 +52,7 @@ jobs:
         id: nodemodulescache
         with:
           path: node_modules
-          key: ${{ runner.os }}-self-hosted-node-modules-${{ hashFiles('yarn.lock') }}
+          key: ${{ runner.os }}-self-hosted-node-modules-${{ hashFiles('yarn.lock', 'api-docs/yarn.lock', '.volta.json') }}
```
asottile-sentry, line 55: I'm not sure all of the installs are the same -- wouldn't want the cache to flipflop based on the apidocs usage -- double check this (it might be fine already)

joshuarli: good catch i meant to remove this

asottile-sentry (approved):

PR #75470, opened 2024-08-01
feat(autofix): Add UI to mask lengthy logs in Autofix

roaga:
Makes long logs expandable so they're not too overwhelming.

Screenshot 2024-08-02 at 10 21 17 AM

### DIFF
--- a/static/app/components/events/autofix/autofixSteps.tsx
+++ b/static/app/components/events/autofix/autofixSteps.tsx
@@ -502,7 +502,6 @@ const LogText = styled('div')<{expanded: boolean; isExpandable: boolean}>`
   -webkit-line-clamp: ${props => (props.expanded ? 'unset' : '2')};
   -webkit-box-orient: vertical;
   overflow-y: hidden;
-  cursor: ${props => (props.isExpandable ? 'pointer' : 'default')};
   max-height: ${props => (props.expanded ? 'none' : '3em')};
 `;
 
--- a/static/app/components/events/autofix/autofixSteps.tsx
+++ b/static/app/components/events/autofix/autofixSteps.tsx
@@ -452,56 +452,54 @@ const ProgressStepContainer = styled('div')`
   grid-column: 1/-1;
 `;
 
-function LogComponent({html}) {
-  const logRef = useRef<any>(null);
-  const [fadeTop, setFadeTop] = useState(false);
-  const [fadeBottom, setFadeBottom] = useState(false);
+function LogComponent({html}: {html: string}) {
+  const [expanded, setExpanded] = useState(false);
+  const [isExpandable, setIsExpandable] = useState(false);
+  const logRef = useRef(null);
 
   useEffect(() => {
-    const element = logRef.current;
-
-    const updateFades = () => {
-      const {scrollTop, clientHeight, scrollHeight} = element;
-      const threshold = 1;
-      const contentScrollable = scrollHeight > clientHeight;
-      if (contentScrollable) {
-        const atBottom = scrollHeight - scrollTop <= clientHeight + threshold;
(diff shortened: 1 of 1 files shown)


static/app/components/events/autofix/autofixSteps.tsx
```diff
@@ -456,3 +451,57 @@ const ProgressContainer = styled('div')`
 const ProgressStepContainer = styled('div')`
   grid-column: 1/-1;
 `;
+
+function LogComponent({html}) {
```
malwilley, line 455: This component should have a type for its props

static/app/components/events/autofix/autofixSteps.tsx
```diff
@@ -456,3 +451,57 @@ const ProgressContainer = styled('div')`
...
+  const logRef = useRef<any>(null);
```
malwilley, line 456: Type should be `HTMLDivElement` instead of `any`

malwilley (commented):
Hmm so I do think this looks pretty good! But I'm not sure we should go forward with this.

This looks nice and clean on macs that have scrollbars automatically hidden, but for many users it will look like this (or even worse on windows or linux):

CleanShot 2024-08-01 at 13 44 37

Another problem is that we just changed our breadcrumbs to avoid stealing scroll focus when scrolling down the page. This would be adding the problem back.

Instead of adding scroll containers, I would find a solution that expands the log text when clicked, if we need to truncate these logs.

static/app/components/events/autofix/autofixSteps.tsx
roaga: Done in next commit

static/app/components/events/autofix/autofixSteps.tsx
roaga: Done in next commit

roaga:
Next commit switches to expandable logs instead of scrolling!

static/app/components/events/autofix/autofixSteps.tsx
```diff
@@ -456,3 +451,55 @@ const ProgressContainer = styled('div')`
...
+    };
+
+    checkExpandable();
+    window.addEventListener('resize', checkExpandable);
+    return () => window.removeEventListener('resize', checkExpandable);
+  }, [html]);
+
+  const toggleExpand = () => {
+    if (isExpandable) {
+      setExpanded(!expanded);
```
malwilley, line 475: nit: better to do `setExpanded(oldState => !oldState)` - react prefers this so you can't use a stale state value

static/app/components/events/autofix/autofixSteps.tsx
```diff
@@ -456,3 +451,55 @@ const ProgressContainer = styled('div')`
...
+    }
+  };
+
+  return (
+    <LogText
+      ref={logRef}
+      expanded={expanded}
+      isExpandable={isExpandable}
+      onClick={toggleExpand}
```
malwilley, line 484: We want to avoid putting `onClick` handlers on non-interactable elements since it is bad for keyboard usage. It should ideally be on a button or at least on something with `role="button"` and a tabIndex=0. (You should be able to expand these without using your mouse by tabbing into it and pressing enter)

malwilley (commented):
This approach is much better! Thanks for making that change 💯

static/app/components/events/autofix/autofixSteps.tsx
roaga: Done in next commit

static/app/components/events/autofix/autofixSteps.tsx
roaga: Done in next commit

malwilley (approved):
Looks good!

PR #75473, opened 2024-08-01
ref: use TypeGuard to fix a few type ignores

asottile-sentry:


asottile-sentry:
cc @isabellaenriquez as an fyi -- [TypeGuard is pretty neat!](https://youtu.be/Zv2K2AVy5lA)

isabellaenriquez (approved):
thanks for cleaning this up!

PR #75475, opened 2024-08-01
ref: fix typing for sentry.tsdb.dummy

asottile-sentry:

### DIFF
--- a/src/sentry/tsdb/redis.py
+++ b/src/sentry/tsdb/redis.py
@@ -878,7 +878,7 @@ class RedisTSDB(BaseTSDB):
         limit: int | None = None,
         environment_id: int | None = None,
         tenant_ids: dict[str, int | str] | None = None,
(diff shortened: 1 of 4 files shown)


src/sentry/tsdb/redis.py
```diff
@@ -878,7 +878,7 @@ def get_most_frequent_series(
         limit: int | None = None,
         environment_id: int | None = None,
         tenant_ids: dict[str, int | str] | None = None,
-    ) -> dict[str, Iterable[dict[str, float]]]:
```
joshuarli, line 881: how was this passing before?

joshuarli (approved):

src/sentry/tsdb/redis.py
asottile-sentry: I think the redis bits are `Any` 😬

PR #75476, opened 2024-08-01
ref(alerts): Remove metric alert load shedding flag

ceorourke:
This has been GA'd for some time now and we don't need the flag anymore. Removing from options automator in https://github.com/getsentry/sentry-options-automator/pull/2051


nhsiehgit (approved):

saponifi3d (approved):
🙏 thanks!

PR #75478, opened 2024-08-01
feat(dashboards): Open in discover button for dashboard widgets

shruthilayaj:
Open in discover is disabled if dataset-selector flag is enabled
and widget is discover type. This should be a fairly short-lived state
since we plan to migrate all existing discover widgets to either errors or
transactions once this feature is GA'ed

### DIFF
--- a/static/app/components/modals/widgetViewerModal.tsx
+++ b/static/app/components/modals/widgetViewerModal.tsx
@@ -66,6 +66,7 @@ import {
   getWidgetIssueUrl,
   getWidgetMetricsUrl,
   getWidgetReleasesUrl,
+  hasDatasetSelector,
 } from 'sentry/views/dashboards/utils';
 import {
   SESSION_DURATION_ALERT,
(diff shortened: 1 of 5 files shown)


static/app/components/modals/widgetViewerModal.tsx
```diff
@@ -1022,29 +1022,6 @@ function WidgetViewerModal(props: Props) {
...
-                          //       )}
-                          //     >
-                          //       <IconWarning color="warningText" size="md" />
-                          //     </Tooltip>
-                          //   );
-                          // }
-
-                          return null;
-                        }}
-                      </DashboardsMEPConsumer>
```
shruthilayaj, line 1047: this is code from 15 months ago that was never reenabled

static/app/utils/discover/fields.tsx
```diff
@@ -749,6 +749,7 @@ export const ERROR_ONLY_FIELDS: (FieldKey | SpanOpBreakdown)[] = [
   FieldKey.STACK_COLNO,
   FieldKey.STACK_LINENO,
   FieldKey.STACK_STACK_LEVEL,
+  FieldKey.EVENT_TYPE,
```
narsaynorath, line 752: Just to confirm, this is because we can have `default`, `csp`, etc as error types right? Good catch.

narsaynorath (approved):

PR #75480, opened 2024-08-01
chore(github): fix typing for github client

cathteng:
bufo-dont-even-see-the-code-anymore

### DIFF
--- a/src/sentry/integrations/github/client.py
+++ b/src/sentry/integrations/github/client.py
@@ -7,7 +7,7 @@ from typing import Any, cast
 
 import orjson
 import sentry_sdk
-from requests import PreparedRequest
+from requests import PreparedRequest, Response
 
 from sentry.constants import ObjectStatus
 from sentry.integrations.github.blame import (
@@ -45,12 +45,14 @@ MINIMUM_REQUESTS = 200
 
 class GithubRateLimitInfo:
     def __init__(self, info: dict[str, int]) -> None:
-        self.limit = info["limit"]
-        self.remaining = info["remaining"]
-        self.reset = info["reset"]
-        self.used = info["used"]
+        self.limit = info.get("limit")
+        self.remaining = info.get("remaining", 0)
+        self.reset = info.get("reset")
+        self.used = info.get("used")
 
     def next_window(self) -> str:
+        if not self.reset:
+            return "unknown"
         return datetime.fromtimestamp(self.reset).strftime("%H:%M:%S")
 
     def __repr__(self) -> str:
@@ -58,8 +60,9 @@ class GithubRateLimitInfo:
 
 
 class GithubProxyClient(IntegrationProxyClient):
+    integration: RpcIntegration | Integration
+
     def _get_installation_id(self) -> str:
-        self.integration: RpcIntegration
         """
         Returns the Github App installation identifier.
         This is necessary since Github and Github Enterprise integrations store the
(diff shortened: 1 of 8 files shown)


src/sentry/integrations/github/client.py
```diff
@@ -45,21 +45,24 @@
...
 class GithubRateLimitInfo:
     def __init__(self, info: dict[str, int]) -> None:
-        self.limit = info["limit"]
-        self.remaining = info["remaining"]
-        self.reset = info["reset"]
-        self.used = info["used"]
+        self.limit = info.get("limit")
+        self.remaining = info.get("remaining", 0)
+        self.reset = info.get("reset")
+        self.used = info.get("used")
```
asottile-sentry, line 51: would rather not weaken these -- the brackets are preferrable since we know the structure (whereas `.get` would allow a programming error to slip through unnoticed on a typo)

src/sentry/shared_integrations/client/base.py
```diff
@@ -304,7 +303,7 @@ def _request(
         self.record_response_for_disabling_integration(resp)
 
         if resp.status_code == 204:
-            return {}
+            return BaseApiResponse(status_code=204)
```
asottile-sentry, line 306: I've been meaning to make this change for a while -- but it's not so simple

some things test `if not response: ...` and assume that's a 204 so it needs quite a lot of work to safely unwind this (probably needs some sort of opt-in `dont_special_case_204s=True` which can then be targetted with `@overload` and not return the empty dict here -- but it's quite a bit of work!)

this probably needs some more involved tracing to nuke properly

also callers really shouldn't need `BaseApiResponseX` since it should be one of the two types specifically (based on the `raw_response=...`) -- so I suspect the github client doesn't need this union either and will always get specifically one or the other response type

cathteng: i see... i put the `{}` back then. is it possible to define `__bool__` to determine if an object is truthy? for 204s we could return `False`

src/sentry/integrations/github/client.py
```diff
@@ -58,8 +58,9 @@ def __repr__(self) -> str:
 
 
 class GithubProxyClient(IntegrationProxyClient):
+    integration: RpcIntegration | Integration
```
asottile-sentry, line 61: I think this is due to it being *expected* from a subclass -- but it turns out there's only one subclass!  so let's merge `GithubProxyClient` and the class below and then I think this annotation may be able to be removed?

src/sentry/integrations/github/client.py
```diff
@@ -750,13 +758,9 @@ def __init__(
...
-        kwargs = {}
-        if hasattr(self.integration, "id"):
-            kwargs["integration_id"] = integration.id
-
         super().__init__(
             org_integration_id=org_integration_id,
             verify_ssl=verify_ssl,
             logging_context=logging_context,
-            **kwargs,
+            integration_id=getattr(self.integration, "id", None),
```
asottile-sentry, line 765: this `getattr` shouldn't be necessary if the `integration: Integration` type is correct -- since that always has an `id: int`

asottile-sentry (commented):
I'm a little torn because I think this _is_ an improvement -- but I also think that the whole api client stuff needs to be torn out and rethought (like I think ideally there would be some response schemas with typeddicts for the structure -- and not have to guess whether it's a requests.Response, the weird dict subclass, or empty dict from 204s)

it'd be a larger chunk of work but it would make everything way easier to work with and have actual proper types

PR #75485, opened 2024-08-01
ref(am1): Collect span counts for orgs prior to migrating

k-fish:
### Summary
Some organizations are not yet on AM2, without span extraction running, which makes it hard to estimate usage for future plan changes to AM3. This allows us to collect metrics for only those specific customers who we can't estimate through outcomes.

Flag I'm using in lieu of a new option is from the subscription plan handler, should be similarly fast to an option in region mode.

### DIFF
--- a/src/sentry/ingest/consumer/processors.py
+++ b/src/sentry/ingest/consumer/processors.py
@@ -187,7 +187,11 @@ def process_event(
                 event_id=event_id,
                 project_id=project_id,
             )
-            collect_span_metrics(project, data)
+
+            try:
+                collect_span_metrics(project, data)
+            except Exception:
+                pass
 
         elif data.get("type") == "feedback":
             if features.has("organizations:user-feedback-ingest", project.organization, actor=None):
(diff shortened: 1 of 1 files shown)


phacops (approved):

src/sentry/ingest/consumer/processors.py
```diff
@@ -187,6 +187,8 @@ def process_event(
                 event_id=event_id,
                 project_id=project_id,
             )
+            collect_span_metrics(project, data)
```
jjbayer, line 190: Can we move this into `save_event_transaction`, and / or wrap it in a try / catch?

This line is part of a `try` block, but it only DLQs `KeyErrors`. Any other failure might block the consumer and cause a P0 incident.

https://github.com/getsentry/sentry/blob/503314b8deabd129e94599c5f3fb36b127d0da6a/src/sentry/ingest/consumer/processors.py#L141-L142

k-fish: Wrap in a try/catch is a good idea, sure 👍

mwarkentin:
@k-fish this was reverted for INC-842, please review when you're back.

PR #75490, opened 2024-08-01
chore(custom-views): Let USERS pass the flag gate 

MichaelSun48:
Adds the ability for individual users with the custom views flag to pass the feature flag gate. We are planning to roll this out internally first so this is necessary.


malwilley (approved):
I think there's one more in `group_index` right?

MichaelSun48:
> I think there's one more in group_index right?

Yes good catch! adding that now

PR #75492, opened 2024-08-01
chore(issue-details): Route feedback to our team

leeandher:
Routes the feedback on some recent issue team changes to us with tags.

### DIFF
--- a/static/app/components/searchQueryBuilder/tokens/combobox.tsx
+++ b/static/app/components/searchQueryBuilder/tokens/combobox.tsx
@@ -323,6 +323,7 @@ function FeedbackFooter() {
             tags: {
               feedback_source: 'search_query_builder',
               search_source: searchSource,
+              ['feedback.owner']: 'issues',
             },
           })
         }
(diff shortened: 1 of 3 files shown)


static/app/components/searchQueryBuilder/tokens/combobox.tsx
```diff
@@ -323,6 +323,7 @@ function FeedbackFooter() {
             tags: {
               feedback_source: 'search_query_builder',
               search_source: searchSource,
+              ['feedback.owner']: 'issues',
```
leeandher, line 326: @malwilley The tags UI makes `.` separated tags look pretty, its the only reason im not using the `_` pattern you started. Is it cool if I update the `feedback_source` tag here? I wanted to have the feedback alert check for the `feedback.owner` anyway.

leeandher: and by pretty I just mean they'll display as nested under `feedback` and I made that component so i think its pretty 🤣

malwilley: Haha I agree it is prettier! Go ahead and change them up!

malwilley (approved):

PR #75496, opened 2024-08-01
ref: delete sentry.tsdb.inmemory

asottile-sentry:
last referenced in 1449643f60404c3ec50ec4eab11bc1c3b3bfe1ab


src/sentry/tsdb/base.py
```diff
@@ -644,7 +644,7 @@ def get_most_frequent_series(
         limit: int | None = None,
         environment_id: int | None = None,
         tenant_ids: dict[str, str | int] | None = None,
-    ) -> dict[str, Iterable[tuple[int, dict[str, float]]]]:
+    ) -> dict[str, list[tuple[int, dict[str, float]]]]:
```
asottile-sentry, line 647: I started typing the inmemory tsdb -- found this inconsistency

asottile-sentry:
a bunch of the inmemory functions always returned the wrong data (always 0.0) or resulted in TypeError -- and hasn't been used in a while so deleting!

wedamija (approved):

PR #75497, opened 2024-08-01
Hmac auth for options webhook

kneeyo1:
### Legal Boilerplate

### DIFF
--- a/src/sentry/runner/commands/presenters/webhookpresenter.py
+++ b/src/sentry/runner/commands/presenters/webhookpresenter.py
@@ -1,3 +1,5 @@
+import hashlib
+import hmac
 from typing import Any
 
 import requests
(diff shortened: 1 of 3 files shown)


src/sentry/runner/commands/presenters/webhookpresenter.py
```diff
@@ -125,9 +127,18 @@ def invalid_type(
...
-            headers = {"Content-Type": "application/json"}
+
+            json_data_str = json.dumps(json_data)
+
+            secret_key = settings.OPTIONS_AUTOMATOR_HMAC_SECRET.encode()
+            signature = hmac.new(secret_key, json_data_str.encode(), hashlib.sha256).hexdigest()
+
+            headers = {
+                "Content-Type": "application/json",
+                "X-Hub-Signature-256": f"sha256={signature}",
```
mdtro, line 138: Verify this is the correct header name. `X-Hub-Signature-256` is what GitHub uses for its webhooks. I suspect we might use something different for options-automator? 🙂

markstory (approved):
Looks good. When the option is set an hmac header will be set that the receiver can use to validate the source of a request.

PR #75499, opened 2024-08-01
ref: add platform to frame and store projectId

JonasBa:
Store platform on frame and projectID in continuous profile metadata


static/app/types/profiling.d.ts
```diff
@@ -185,28 +187,28 @@ declare namespace Profiling {
   };
 
   type FrameInfo = {
-    key: string | number;
```
JonasBa, line 188: I went ahead and sorted these to make it easier to parse given that we are starting to have quite a few

Zylphrex (approved):

PR #75500, opened 2024-08-01
chore(slack): Add metrics to see if "Expired url" Error ever occurs

iamrajjoshi:
i have a hunch that this error never actually happens, so i am adding a metric to see if it actually happens.


cathteng (approved):
i'm not sure you need both the metric and the logger

PR #75509, opened 2024-08-01
ref(alerts): Send historical data to Seer on update

ceorourke:
A follow up to https://github.com/getsentry/sentry/pull/75215 which sent historical data to Seer when a dynamic metric alert rule is created, this PR does the same when a dynamic alert's query or aggregate is edited, or when a static or dynamic alert rule is changed to be a dynamic alert rule.

Closes https://getsentry.atlassian.net/browse/ALRT-168

### DIFF
--- a/src/sentry/incidents/logic.py
+++ b/src/sentry/incidents/logic.py
@@ -654,8 +654,8 @@ def create_alert_rule(
 
             rule_data = DynamicAlertRuleData(
                 alert_rule_id=alert_rule.id,
-                seasonality=alert_rule.seasonality.value,
-                sensitivity=alert_rule.sensitivity.value,
+                seasonality=alert_rule.seasonality,
+                sensitivity=alert_rule.sensitivity,
                 threshold_type=alert_rule.threshold_type,
             )
             snuba_query_data = SnubaQueryData(
--- a/src/sentry/incidents/logic.py
+++ b/src/sentry/incidents/logic.py
@@ -761,13 +761,11 @@ def should_update_seer(
     query: str,
     aggregate: str,
 ) -> bool:
-    should_update_seer = False
     if previous_detection_type != existing_detection_type or query or aggregate:
-        should_update_seer = True
+        return True
     elif updated_detection_type is not None:
         if previous_detection_type != updated_detection_type:
-            should_update_seer = True
-    return should_update_seer
+            return True
 
 
 def update_dynamic_alert_data(
--- a/src/sentry/incidents/logic.py
+++ b/src/sentry/incidents/logic.py
@@ -652,7 +652,7 @@ def create_alert_rule(
                 )
 
             try:
-                send_historical_data_to_seer(alert_rule, projects[0])
+                send_historical_data_to_seer(alert_rule=alert_rule, project=projects[0])
             except (TimeoutError, MaxRetryError):
                 alert_rule.delete()
                 raise TimeoutError
@@ -899,21 +899,20 @@ def update_alert_rule(
                     "Your organization does not have access to this feature."
                 )
 
-            previous_detection_type = alert_rule.detection_type
             if updated_fields.get("detection_type") == AlertRuleDetectionType.DYNAMIC and (
-                previous_detection_type != AlertRuleDetectionType.DYNAMIC or query or aggregate
+                alert_rule.detection_type != AlertRuleDetectionType.DYNAMIC or query or aggregate
             ):
                 for k, v in updated_fields.items():
                     alert_rule.k = v
                 try:
                     send_historical_data_to_seer(
-                        alert_rule, projects[0] if projects else alert_rule.projects.get()
+                        alert_rule=alert_rule,
+                        project=projects[0] if projects else alert_rule.projects.get(),
                     )
                 except (TimeoutError, MaxRetryError):
                     raise TimeoutError("Failed to send data to Seer - cannot update alert rule.")
 
         alert_rule.update(**updated_fields)
-
         AlertRuleActivity.objects.create(
             alert_rule=alert_rule,
             user_id=user.id if user else None,
--- a/src/sentry/incidents/logic.py
+++ b/src/sentry/incidents/logic.py
@@ -755,7 +755,11 @@ def snapshot_alert_rule(alert_rule, user=None):
 
 
 def should_update_seer(
-    previous_detection_type, existing_detection_type, updated_detection_type, query, aggregate
+    previous_detection_type: AlertRuleDetectionType,
+    existing_detection_type: AlertRuleDetectionType,
+    updated_detection_type: AlertRuleDetectionType,
+    query: str,
+    aggregate: str,
 ) -> bool:
     should_update_seer = False
     if previous_detection_type != existing_detection_type or query or aggregate:
@@ -767,15 +771,15 @@ def should_update_seer(
 
 
 def update_dynamic_alert_data(
-    alert_rule,
-    seasonality,
-    sensitivity,
-    threshold_type,
-    query,
-    time_window,
-    dataset,
-    aggregate,
-    projects,
+    alert_rule: AlertRule,
+    seasonality: AlertRuleSeasonality,
+    sensitivity: AlertRuleSensitivity,
+    threshold_type: int,
+    query: str,
+    time_window: int,
+    dataset: str,
+    aggregate: str,
+    projects: list[Project],
 ) -> None:
     updated_seasonality = seasonality if seasonality is not NOT_SET else alert_rule.seasonality
     updated_sensitivity = sensitivity if sensitivity is not NOT_SET else alert_rule.sensitivity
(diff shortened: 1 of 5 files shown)


src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -49,7 +48,12 @@ def format_historical_data(data: SnubaTSResult) -> list[TimeSeriesPoint]:
     return formatted_data
 
 
-def send_historical_data_to_seer(alert_rule: AlertRule, project: Project) -> BaseHTTPResponse:
+def send_historical_data_to_seer(
```
ceorourke, line 51: I had to refactor this to handle the case where we're updating an alert but the Seer call fails - if I let the rule update and _then_ did the Snuba query and Seer call, we risk having an alert in a bad state. I decided to pass all the data needed to the function instead - for edited alerts we already have the rule ID so that's fine, and the rest I just put into new typeddicts.

src/sentry/incidents/logic.py
```diff
@@ -735,6 +754,67 @@ def snapshot_alert_rule(alert_rule, user=None):
...
+    previous_detection_type: AlertRuleDetectionType,
+    updated_detection_type: AlertRuleDetectionType,
+    query: str,
+    aggregate: str,
+) -> bool:
+    """
+    Update Seer data if a dynamic alert's query or aggregate has changed OR
+    a static or percent alert has changed to become a dynamic alert
+    """
+    if (
```
nhsiehgit, line 767: Could we consolidate this to
```
return (
    updated_type == Dynamic AND (
        previous_type != Dynamic OR
        query OR
        aggregate
    )
)
```
?

src/sentry/incidents/logic.py
```diff
@@ -893,7 +973,31 @@ def update_alert_rule(
             updated_fields["user_id"] = alert_rule.user_id
             updated_fields["team_id"] = alert_rule.team_id
 
+        if detection_type == AlertRuleDetectionType.DYNAMIC:
+            if not features.has("organizations:anomaly-detection-alerts", alert_rule.organization):
+                raise ResourceDoesNotExist(
+                    "Your organization does not have access to this feature."
+                )
+            if should_update_seer(
```
nhsiehgit, line 981: Is this helper used anywhere else?
Does it need to be a helper?

src/sentry/incidents/logic.py
```diff
@@ -735,6 +754,67 @@ def snapshot_alert_rule(alert_rule, user=None):
...
+    )
+    snuba_query_data = SnubaQueryData(
+        query=query or alert_rule.snuba_query.query,
+        time_window=time_window or alert_rule.snuba_query.time_window,
+        dataset=dataset or alert_rule.snuba_query.dataset,
+        aggregate=aggregate or alert_rule.snuba_query.aggregate,
+    )
+    project = projects[0] if projects else alert_rule.projects.get()
+    # do we need to tell seer when sensitivity, seasonality, or threshold_type changes?
+    # or can they get that from the periodic updates?
```
nhsiehgit, line 806: What are periodic updates? Does this mean from the subscription processor ingest?

Ideally we be explicit with the updates - when a customer updates the model, we propagate that through to Seer

src/sentry/incidents/logic.py
```diff
@@ -893,7 +973,31 @@ def update_alert_rule(
...
+                raise ResourceDoesNotExist(
+                    "Your organization does not have access to this feature."
+                )
+            if should_update_seer(
+                alert_rule.detection_type,
+                updated_fields.get("detection_type"),
+                query,
+                aggregate,
+            ):
+                update_dynamic_alert_data(
```
nhsiehgit, line 987: Does this need to be a separate helper?
the method above seems fairly straight forwards and i don't know if we gain much from componentizing this?

src/sentry/seer/anomaly_detection/store_data.py
nhsiehgit: Hmm...

If the concern here is that the alert_rule could be saved in a bad state, rather than using `alert_rule.update`, we could just set the attrs on the `alert_rule` and just not `alert_rule.save()` it until we've properly updated seer?

> having an alert in a bad state
could you explain more of what state the alert rule could fall into?

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -62,38 +66,37 @@ def send_historical_data_to_seer(alert_rule: AlertRule, project: Project) -> Bas
...
         return base_error_response
 
-    formatted_data = format_historical_data(historical_data)
-
     if (
-        not alert_rule.sensitivity
-        or not alert_rule.seasonality
-        or alert_rule.threshold_type is None
-        or alert_rule.organization is None
+        not rule_data["sensitivity"]
```
nhsiehgit, line 76: Is this necessary now that you've built a typeddict?

nhsiehgit (commented):
I'm not entirely sure how i feel about these typeddicts 🤔 

So the purpose we're breaking this up is so that we can either utilize the `alert_rule` to fetch our historical snuba data, or we can pass in the updated configs to query for the historical data?

Could we instead split the `fetch_historical_snuba_data` and the `backfill_seer` methods here?
Then we can be explicit with the args to fetch the data with, backfill_seer with, and on success, we can update the alert rule?

src/sentry/incidents/logic.py
ceorourke: yeah just the regular metric alert updates

src/sentry/incidents/logic.py
ceorourke: `create_alert_rule` and `update_alert_rule` are just SO long

src/sentry/seer/anomaly_detection/store_data.py
ceorourke: The bad state would be that we've already updated it but the Seer call fails so we now have a dynamic alert rule with a metric that has no historical data, or outdated data for a different metric

src/sentry/seer/anomaly_detection/store_data.py
ceorourke: idk i'll ask mypy

src/sentry/seer/anomaly_detection/store_data.py
ceorourke: I changed this to loop through and set the attrs for dynamic alerts only before we call `send_historical_data_to_seer` and then `update()`

src/sentry/incidents/logic.py
```diff
@@ -893,6 +893,25 @@ def update_alert_rule(
...
+            if not features.has("organizations:anomaly-detection-alerts", alert_rule.organization):
+                raise ResourceDoesNotExist(
+                    "Your organization does not have access to this feature."
+                )
+
+            if updated_fields.get("detection_type") == AlertRuleDetectionType.DYNAMIC and (
+                alert_rule.detection_type != AlertRuleDetectionType.DYNAMIC or query or aggregate
+            ):
+                for k, v in updated_fields.items():
+                    alert_rule.k = v
```
nhsiehgit, line 906: 👌 nice.
This feels much more succinct

tests/sentry/incidents/test_logic.py
```diff
@@ -1569,6 +1569,151 @@ def test_update_infer_detection_type(self, mock_seer_request):
...
+
+        dynamic_rule = self.create_alert_rule(
+            sensitivity=AlertRuleSensitivity.HIGH,
+            seasonality=AlertRuleSeasonality.AUTO,
+            time_window=60,
+            detection_type=AlertRuleDetectionType.DYNAMIC,
+        )
+        assert mock_seer_request.call_count == 1
+        mock_seer_request.reset_mock()
+        update_alert_rule(dynamic_rule, time_window=30)
```
nhsiehgit, line 1587: Just to make each bit clear, can we leave some notes on what we're updating on the alert rules, and whether we expect/don't expect them to re-query seer?

tests/sentry/incidents/test_logic.py
```diff
@@ -1569,6 +1569,151 @@ def test_update_infer_detection_type(self, mock_seer_request):
...
+            static_rule,
+            time_window=30,
+            sensitivity=AlertRuleSensitivity.HIGH,
+            seasonality=AlertRuleSeasonality.AUTO,
+            detection_type=AlertRuleDetectionType.DYNAMIC,
+        )
+        assert mock_seer_request.call_count == 1
+        mock_seer_request.reset_mock()
+
+        percent_rule = self.create_alert_rule(
```
nhsiehgit, line 1623: Can we make this a separate test?
`test_update_alert_rule_from_percent_to_dynamic`?

nhsiehgit (approved):
:chefs_kiss:

very nice
a few nit comments, but +1 after!

tests/sentry/incidents/test_logic.py
ceorourke: it seems redundant to me to do
```
# update time_window
update_alert_rule(dynamic_rule, time_window=30)
```

nhsiehgit: For the other updates though it's not as useless?

it's difficult to see exactly what's being updated where.
Adding notes at least helps visually differentiate where each "test" is happening.
alternatively we could also split each of these into their own separate test

PR #75511, opened 2024-08-01
ref(control_silo): Move vsts and general integrations tasks to new location

Christinarlong:
Moves logic/body of old tasks to new task locations, updates any places calling the old task to use new location.

ref(issue # here)


markstory (approved):
Looks good to me.

src/sentry/conf/server.py
```diff
@@ -814,6 +814,7 @@ def SOCIAL_AUTH_DEFAULT_USERNAME() -> str:
     "sentry.uptime.detectors.tasks",
     "sentry.uptime.subscriptions.tasks",
     "sentry.integrations.vsts.tasks",
+    "sentry.integrations.vsts.tasks.kickoff_subscription_check",
```
Christinarlong, line 817: It seems like this is needed to pass the test `tests/sentry/celery/test_app.py`, even though the 'folder' with an __init__ is specified above (?). Is my understanding of celery off that, celery would be able to find this task (even though the test said otherwise) ?

markstory: Yeah, the test is naive and assumes that a module import doesn't pull in all of the child modules (which is a safe assumption to make imo). Additional celery imports don't hurt anything.

PR #75516, opened 2024-08-01
fix(trace) check for pageload count instead of child count

JonasBa:
We had a test to checking that we skip reparenting cases where server handlers have multiple direct pageload children, however the actual condition was checking all children types and not only pageload children, making the tests pass, but the reparenting to fail on cases where it should have worked.

### DIFF
--- a/static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
+++ b/static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
@@ -233,6 +233,21 @@ function isBrowserRequestSpan(value: TraceTree.Span): boolean {
   return value.op === 'browser' && value.description === 'request';
 }
 
+function getPageloadTransactionChildCount(
+  node: TraceTreeNode<TraceTree.NodeValue>
+): number {
+  if (!isTransactionNode(node)) {
+    return 0;
+  }
+  let count = 0;
+  for (const txn of node.value.children) {
+    if (txn && txn['transaction.op'] === 'pageload') {
+      count++;
+    }
+  }
+  return count;
+}
+
 /**
  * Swaps the two nodes in the graph.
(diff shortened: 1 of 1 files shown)


static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
```diff
@@ -532,8 +547,7 @@ export class TraceTree {
 
     function visit(
       parent: TraceTreeNode<TraceTree.NodeValue | null>,
-      value: TraceTree.Transaction | TraceTree.TraceError,
-      childrenCount: number
```
JonasBa, line 536: count is no longer needed

static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
```diff
@@ -597,9 +611,9 @@ export class TraceTree {
       }
 
       if (
-        childrenCount === 1 &&
         isPageloadTransactionNode(node) &&
-        isServerRequestHandlerTransactionNode(parent)
+        isServerRequestHandlerTransactionNode(parent) &&
+        getPageloadTransactionChildCount(parent) === 1
```
Lms24, line 616: just for me to understand: We don't do anything if the parent (http.server) txn has more than one `op: pageload` child txns, right?

Lms24 (approved):
Thanks for fixing this so quickly!

static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
JonasBa: Exactly (afaik this should never happen), but the main reason being that we dont do anything is because we dont know which pageload to reparent the server handler under.

PR #75523, opened 2024-08-02
ref(relay): Write project config revision to Redis

Dav1dde:
See: https://github.com/getsentry/relay/issues/3887

Writes the revision to Redis as a separate key. Allowing for optimized reads by revision.

#skip-changelog

### DIFF
--- a/src/sentry/relay/projectconfig_cache/redis.py
+++ b/src/sentry/relay/projectconfig_cache/redis.py
@@ -36,7 +36,7 @@ class RedisProjectConfigCache(ProjectConfigCache):
     def set_many(self, configs: dict[str, Mapping[str, Any]]):
         metrics.incr("relay.projectconfig_cache.write", amount=len(configs), tags={"action": "set"})
 
-        # Note: Those are multiple pipelines, one per cluster node
+        # Note: Those are multiple pipelines, one per cluster node.
         p = self.cluster.pipeline(transaction=False)
         for public_key, config in configs.items():
             serialized = json.dumps(config).encode()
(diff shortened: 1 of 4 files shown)


src/sentry/relay/projectconfig_cache/redis.py
```diff
@@ -63,6 +68,11 @@ def get(self, public_key):
                 rv = zstandard.decompress(rv_b).decode()
             except (TypeError, zstandard.ZstdError):
                 # assume raw json
-                pass
+                rv = rv_b
```
Dav1dde, line 71: Drive by fix ... That would have just blown up before because `rv` is not defined.

src/sentry/relay/projectconfig_cache/redis.py
```diff
@@ -41,7 +43,10 @@ def set_many(self, configs):
             )
             metrics.distribution("relay.projectconfig_cache.size", len(compressed), unit="byte")
 
-            p.setex(self.__get_redis_key(public_key), REDIS_CACHE_TIMEOUT, compressed)
+            key = self.__get_redis_key(public_key)
+            if rev := config.get("rev"):
+                p.setex(f"{key}.rev", REDIS_CACHE_TIMEOUT, rev)
```
jjbayer, line 48: nit: 
```suggestion
                p.setex(self.__get_rev_key(public_key), REDIS_CACHE_TIMEOUT, rev)
```

jjbayer (approved):

src/sentry/relay/projectconfig_cache/redis.py
```diff
@@ -28,11 +30,14 @@ def validate(self):
...
-    def set_many(self, configs):
+    def __get_redis_rev_key(self, public_key):
+        return f"{self.__get_redis_key(public_key)}.rev"
+
+    def set_many(self, configs: dict[str, Mapping[str, Any]]):
         metrics.incr("relay.projectconfig_cache.write", amount=len(configs), tags={"action": "set"})
 
         # Note: Those are multiple pipelines, one per cluster node
-        p = self.cluster.pipeline()
+        p = self.cluster.pipeline(transaction=False)
```
Dav1dde, line 40: There is no need for this pipeline to be transactional (default value), all keys are independent.

PR #75528, opened 2024-08-02
chore(replays): Include compressed rrweb in video size metric

cmanallen:
Including the compressed rrweb data to get a full picture on the amount of data stored in GCS.


michellewzhang (approved):

c298lee (approved):

aliu39:
Why are we adding them instead of using a separate metric?

PR #75529, opened 2024-08-02
chore(controlsilo) Move github tasks to sentry.integrations.github.tasks

markstory:
- Move implementation github tasks from sentry.tasks.integrations.github to sentry.integrations.github.tasks
- Move related constants and utilities modules to sentry.integrations.github as well.
- Stabilize a flaky test by adding order_by to the queries in application and test code.

Part of the #74925


src/sentry/conf/server.py
```diff
@@ -738,6 +738,7 @@ def SOCIAL_AUTH_DEFAULT_USERNAME() -> str:
     "sentry.hybridcloud.tasks.deliver_webhooks",
     "sentry.incidents.tasks",
     "sentry.integrations.github.tasks",
+    "sentry.integrations.github.tasks.pr_comment",
```
Christinarlong, line 741: copy pasting my comment from vsts PR here:

It seems like this is needed to pass the test tests/sentry/celery/test_app.py, even though the 'folder' with an init is specified above (?). Is my understanding of celery off that, celery would be able to find this task (even though the test said otherwise) ?

markstory: Hrm, this shouldn't be necessary as the import for `sentry.integrations.github.tasks` should pull in all the child modules as well. I'll look into the test and see if it is faulty or if celery would be missing tasks.

markstory: Looked at the test, and the test is looking for exact module matches. It assumes that a module like `sentry.integrations.github.tasks` may not pull in all of the child modules (which is a safer assumption).

Christinarlong (approved):

PR #75532, opened 2024-08-02
fix(openapi): update openapi server url

DominikB2014:
Since we moved to regional apis, our server url is now `us.sentry.io`.

Someone can correct me if i'm wrong, but the europe server url is `de.sentry.io`? In which case, we can add it as well.

For context, when you feed the api spec into a doc generator such as swagger UI, the server url is used as a base url to all the endpoints.


schew2381:
We could also need to change the url for drf-spec in `server.py`, as the change in the JSON file only affects APIs documented with manual JSON
https://github.com/getsentry/sentry/blob/45304f379adb4768372aeabef8d161f29d9ad744/src/sentry/conf/server.py#L1421 

One thing I'm curious about though is what having multiple urls will actually look like in the UI

DominikB2014:
@schew2381 
> We could also need to change the url for drf-spec in `server.py`, as the change in the JSON file only affects APIs documented with manual JSON

Thanks for the pointer!!

> One thing I'm curious about though is what having multiple urls will actually look like in the UI

In swagger, it shows up as another server in "servers" dropdown
image

schew2381:
I think it would be good to add the EU url as well, I think you're right that it is `de.sentry ...` although I'm not 100% sure

DominikB2014:
@schew2381 It appears the europe server is `de.sentry.io`, i made the changes. If there are more servers to add, we can add them in upcoming PRs!

schew2381 (approved):
lgtm!

DominikB2014:
@schew2381 Thanks for the reviews!! One last thing, I'm getting failures on the open api diff, but a diff is expected. 
https://github.com/getsentry/sentry/actions/runs/10376091159/job/28727210659
Any ideas if there's anything else I need to do here?

schew2381:
> @schew2381 Thanks for the reviews!! One last thing, I'm getting failures on the open api diff, but a diff is expected. https://github.com/getsentry/sentry/actions/runs/10376091159/job/28727210659 Any ideas if there's anything else I need to do here?
> 
> Edit: nvm, seems to have merged! I think the error is more of a warning that the spec is different?

Oh yeah that test is non-blocking. We've been meaning to make it a warning rather than a failure for a while now 😅

PR #75534, opened 2024-08-02
chore(statistical-detectors): Add option to configure timeseries days

Zylphrex:
Looking to experiment with more historical data for breakpoint detection. This will allow us to control the length of the timeseries.

### DIFF
--- a/src/sentry/options/defaults.py
+++ b/src/sentry/options/defaults.py
@@ -2000,6 +2000,12 @@ register(
     flags=FLAG_PRIORITIZE_DISK | FLAG_AUTOMATOR_MODIFIABLE,
 )
 register(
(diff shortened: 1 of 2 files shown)


trillville (approved):

src/sentry/options/defaults.py
```diff
@@ -1999,6 +1999,12 @@
     default=14,
     flags=FLAG_PRIORITIZE_DISK | FLAG_AUTOMATOR_MODIFIABLE,
 )
+register(
+    "statistical_detectors.query.functions.timeseries_days",
```
Zylphrex, line 2003: the option for transactions already exist above, just replicating it for functions too

PR #75536, opened 2024-08-02
feat(breadcrumbs): Add level filtering and previous search logic

leeandher:
Tried to simplify search but it might be behaving differently in a way I haven't been able to test. Since there was really nothing wrong with the old one, I've made it work for the new UI and removed 'new search' which wasn't markedly better (maybe faster? but I'm not sure).

Also added back level filtering, which was removed when we had decided to omit levels in a previous iteration. It was mistakenly not added back then.
image

### DIFF
--- a/static/app/components/events/breadcrumbs/breadcrumbsDrawer.tsx
+++ b/static/app/components/events/breadcrumbs/breadcrumbsDrawer.tsx
@@ -9,14 +9,14 @@ import ButtonBar from 'sentry/components/buttonBar';
 import {CompactSelect} from 'sentry/components/compactSelect';
 import BreadcrumbsTimeline from 'sentry/components/events/breadcrumbs/breadcrumbsTimeline';
 import {
-  applyBreadcrumbSearch,
   BREADCRUMB_TIME_DISPLAY_LOCALSTORAGE_KEY,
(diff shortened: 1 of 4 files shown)


static/app/components/events/breadcrumbs/breadcrumbsDrawer.tsx
```diff
@@ -85,20 +85,15 @@ export function BreadcrumbsDrawer({
     BREADCRUMB_TIME_DISPLAY_LOCALSTORAGE_KEY,
     BreadcrumbTimeDisplay.ABSOLUTE
   );
-  const filterOptions = useMemo(
-    () => getBreadcrumbFilterOptions(breadcrumbs),
-    [breadcrumbs]
-  );
+  const {filterOptions, applyFilters} = useBreadcrumbFilters(breadcrumbs);
```
leeandher, line 88: Decided to do it this way since the options should actually be applied in two different ways, though they show up in the same list of filters. 

Rather than having the caller know that `filterOptions` will contain `level-Error` and `type-Exception` and should treat those differently, I just made it spit out a function for applying the filter so all that weird `level-` and `type-` logic is self contained.

roggenkemper (approved):

PR #75539, opened 2024-08-02
feat(api): Publish ExternalActorEndpoint

iamrajjoshi:
image

image

image

### DIFF
--- a/src/sentry/apidocs/parameters.py
+++ b/src/sentry/apidocs/parameters.py
@@ -171,6 +171,14 @@ Valid fields include:
 """,
     )
 
+    EXTERNAL_USER_ID = OpenApiParameter(
+        name="external_user_id",
+        location="path",
+        required=True,
+        type=int,
+        description="The ID of the External User Object. This is returned when creating an external user.",
+    )
+
 
 class ReleaseParams:
     VERSION = OpenApiParameter(
--- a/src/sentry/apidocs/parameters.py
+++ b/src/sentry/apidocs/parameters.py
@@ -176,7 +176,7 @@ Valid fields include:
         location="path",
         required=True,
         type=int,
-        description="The ID of the External User Object. This is returned when creating an external user.",
+        description="The ID of the external user object. This is returned when creating an external user.",
     )
 
 
--- a/src/sentry/integrations/api/endpoints/external_user_details.py
+++ b/src/sentry/integrations/api/endpoints/external_user_details.py
@@ -64,7 +64,7 @@ class ExternalUserDetailsEndpoint(OrganizationEndpoint, ExternalActorEndpointMix
         self, request: Request, organization: Organization, external_user: ExternalActor
     ) -> Response:
         """
-        Update an External User
+        Updates a linked user in an external provider to a user in Sentry.
         """
         self.assert_has_feature(request, organization)
 
@@ -98,7 +98,7 @@ class ExternalUserDetailsEndpoint(OrganizationEndpoint, ExternalActorEndpointMix
         self, request: Request, organization: Organization, external_user: ExternalActor
     ) -> Response:
         """
-        Delete an External Team
+        Delete a linked user in an external provider to a user in Sentry.
         """
         external_user.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
--- a/src/sentry/integrations/api/endpoints/external_user_details.py
+++ b/src/sentry/integrations/api/endpoints/external_user_details.py
@@ -13,13 +13,13 @@ from sentry.api.api_publish_status import ApiPublishStatus
 from sentry.api.base import region_silo_endpoint
 from sentry.api.bases.organization import OrganizationEndpoint
 from sentry.api.serializers import serialize
+from sentry.apidocs.constants import RESPONSE_BAD_REQUEST, RESPONSE_FORBIDDEN, RESPONSE_NO_CONTENT
+from sentry.apidocs.examples.organization_examples import OrganizationExamples
+from sentry.apidocs.parameters import GlobalParams, OrganizationParams
 from sentry.integrations.api.bases.external_actor import (
     ExternalActorEndpointMixin,
     ExternalUserSerializer,
 )
-from sentry.apidocs.constants import RESPONSE_BAD_REQUEST, RESPONSE_FORBIDDEN, RESPONSE_NO_CONTENT
-from sentry.apidocs.examples.organization_examples import OrganizationExamples
-from sentry.apidocs.parameters import GlobalParams, OrganizationParams
 from sentry.integrations.models.external_actor import ExternalActor
 from sentry.models.organization import Organization
 
--- a/src/sentry/integrations/api/endpoints/external_user_details.py
+++ b/src/sentry/integrations/api/endpoints/external_user_details.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 from typing import Any
 
+from drf_spectacular.utils import extend_schema
 from rest_framework import status
 from rest_framework.request import Request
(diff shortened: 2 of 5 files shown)


src/sentry/api/bases/external_actor.py
```diff
@@ -106,7 +122,7 @@ def update(
 class ExternalUserSerializer(ExternalActorSerializerBase):
     _actor_key = "user_id"
 
-    user_id = serializers.IntegerField(required=True)
+    user_id = serializers.IntegerField(required=True, help_text="The User ID in Sentry.")
```
schew2381, line 125:
```suggestion
    user_id = serializers.IntegerField(required=True, help_text="The user ID in Sentry.")
```

src/sentry/api/bases/external_actor.py
```diff
@@ -40,11 +41,26 @@
...
+        required=False, allow_null=True, help_text="The associated User ID for provider."
+    )
+    external_name = serializers.CharField(
+        required=True, help_text="The associated username for the provider."
+    )
+    provider = serializers.ChoiceField(
+        choices=get_provider_choices(AVAILABLE_PROVIDERS),
+        help_text="The provider. ('github', 'gitlab', 'slack')",
+    )
+    integration_id = serializers.IntegerField(required=True, help_text="The Integration ID.")
```
schew2381, line 63:
```suggestion
    integration_id = serializers.IntegerField(required=True, help_text="The integration ID.")
```

src/sentry/api/bases/external_actor.py
```diff
@@ -40,11 +41,26 @@
...
+    external_id = serializers.CharField(
+        required=False, allow_null=True, help_text="The associated User ID for provider."
+    )
+    external_name = serializers.CharField(
+        required=True, help_text="The associated username for the provider."
+    )
+    provider = serializers.ChoiceField(
+        choices=get_provider_choices(AVAILABLE_PROVIDERS),
+        help_text="The provider. ('github', 'gitlab', 'slack')",
+    )
```
schew2381, line 62: This is a bit hard to document b/c we have the setting that automatically generates bullet points from choices, so it'll include more than github, gitlab, and slack which you can see in your screenshot

Not sure how to avoid this tbh

src/sentry/api/bases/external_actor.py
```diff
@@ -40,11 +41,26 @@
...
+    integration_id: int
+
+
 class ExternalActorSerializerBase(CamelSnakeModelSerializer):
-    external_id = serializers.CharField(required=False, allow_null=True)
-    external_name = serializers.CharField(required=True)
-    provider = serializers.ChoiceField(choices=get_provider_choices(AVAILABLE_PROVIDERS))
-    integration_id = serializers.IntegerField(required=True)
+    external_id = serializers.CharField(
+        required=False, allow_null=True, help_text="The associated User ID for provider."
```
schew2381, line 54:
```suggestion
        required=False, allow_null=True, help_text="The associated user ID for provider."
```

src/sentry/apidocs/parameters.py
```diff
@@ -171,6 +171,14 @@ class OrganizationParams:
 """,
     )
 
+    EXTERNAL_USER_ID = OpenApiParameter(
+        name="external_user_id",
+        location="path",
+        required=True,
+        type=int,
+        description="The ID of the External User Object. This is returned when creating an external user.",
```
schew2381, line 179:
```suggestion
        description="The ID of the external user object. This is returned when creating an external user.",
```

src/sentry/integrations/api/endpoints/external_user_index.py
```diff
@@ -10,30 +11,37 @@
...
-        `````````````
-
-        :pparam string organization_id_or_slug: the id or slug of the organization the
-                                          user belongs to.
-        :param required string provider: enum("github", "gitlab", "slack")
-        :param required string external_name: the associated username for this provider.
-        :param required int user_id: the User ID in Sentry.
-        :param string external_id: the associated user ID for this provider
-        :auth: required
+        Links a user in an external provider to a user in Sentry.
```
schew2381, line 44:
```suggestion
        Links a user from an external provider to a Sentry user.
```

src/sentry/integrations/api/endpoints/external_user_details.py
```diff
@@ -74,11 +80,22 @@ def put(
...
+            403: RESPONSE_FORBIDDEN,
+        },
+        examples=OrganizationExamples.EXTERNAL_USER_CREATE,
+    )
     def delete(
         self, request: Request, organization: Organization, external_user: ExternalActor
     ) -> Response:
         """
-        Delete an External Team
+        Delete a linked user in an external provider to a user in Sentry.
```
schew2381, line 98:
```suggestion
        Delete the link between a user from an external provider and a Sentry user.
```

src/sentry/integrations/api/endpoints/external_user_details.py
```diff
@@ -41,21 +46,22 @@ def convert_args(
...
-
-        :pparam string organization_id_or_slug: the id or slug of the organization the
-                                          user belongs to.
-        :pparam int user_id: the User id.
-        :pparam string external_user_id: id of external_user object
-        :param string external_id: the associated user ID for this provider
-        :param string external_name: the Github/Gitlab user name.
-        :param string provider: enum("github","gitlab")
-        :auth: required
+        Updates a linked user in an external provider to a user in Sentry.
```
schew2381, line 64:
```suggestion
        Update a user in an external provider that is currently linked to a Sentry user.
```

src/sentry/api/bases/external_actor.py
iamrajjoshi: actually ill remove the list i sent altogether.  even though we don't really officially use the ExternalActor table for other providers than the 3 i listed, technically we can add this linking for all of it. this will eventually help ecosystem anyway as we start supporting team linking in the other messaging and scm integrations.

schew2381: so the docs will include all providers, even tho we only support some of them?

iamrajjoshi: we support all of them, but we just don't do anything with them.

schew2381: Hmm should we change the `get_provider_choices` to be limited to those 3 providers then?

iamrajjoshi: good point, let me do that

schew2381: also can the choicefield just accept a regular list of strings, instead of a list of tuples? That way what's generated won't include the second item in the tuple which doesn't provide any useful info

iamrajjoshi: actually i don't want to limit it to those 3. in code technically we support other integrations as [well](https://github.com/getsentry/sentry/blob/27195bd74d86d77bbb37e7f962aabfd2a3c398b7/src/sentry/integrations/msteams/notifications.py#L88). the comment that existed on the old endpoint predated the new changes, it should have been deleted.

schew2381: actually I don't know why it's autogenerating those bullet points.

It's controlled by the `ENUM_GENERATE_CHOICE_DESCRIPTION` setting in `SPECTACULAR_SETTINGS` in `server.py`, but it's not there anymore

schew2381: oh actually I think it just defaults to True always. If the choicefield just has a regular list, it copies the first value as both the choice value - choice name

schew2381 (approved):
lgtm, just merge after/same time as https://github.com/getsentry/sentry/pull/75657

PR #75541, opened 2024-08-02
ref(incidents): Add type hints to sentry/incidents/logic.py

RyanSkonnord:
Add type hints to the `sentry.incidents.logic` module and some related functions.

Remove unused function parameters. Change default collection args from `None` to `()` where they are treated interchangeably, to minimize nullable arg types. Add explicit null checks where required by the type checker.

Add underscore prefix to functions used privately. Move functions that were used only by a single module.


src/sentry/incidents/logic.py
```diff
@@ -1849,7 +1841,7 @@ def is_included(action: Mapping[str, Any]) -> bool:
     ]
 
 
-def schedule_update_project_config(alert_rule: AlertRule, projects: Sequence[Project]):
+def schedule_update_project_config(alert_rule: AlertRule, projects: Collection[Project]) -> None:
```
cathteng, line 1844: is `Collection` looser than `Sequence`?

cathteng (approved):
the `src/sentry/incidents/logic.py` file is huge :/

src/sentry/incidents/logic.py
RyanSkonnord: Yes. `Sequence` implies that it has a consistent iteration order and can be indexed with brackets, usually meaning a `list` or `tuple`. `Collection` includes `set` and `frozenset`.

src/sentry/incidents/logic.py
```diff
@@ -2,14 +2,16 @@
```
asottile-sentry, line 2: this file is in the ignore list so none of this work is checked -- fix that first and *then* improve the type annotations

asottile-sentry (changes requested):

src/sentry/incidents/logic.py
RyanSkonnord: This started out smaller in scope. I was checking the work as I went with local `mypy` runs and PyCharm warnings, not necessarily intending to get as far as removing the file from the ignore list. When I did try that (earlier in this branch's history), there were about ~100 more failures, which seemed like too much to tackle at the time. I'd love to devote more time later to getting `logic.py` fully checked, but I was hoping this branch would be mergeable in its current state as intermediate progress.

Would you agree with that? Or are you saying that we shouldn't be adding type annotations without CI coverage? They may contain some mistakes, but from the offline checks, I'd say I have a high degree of confidence that these will help clarity more than harm it in the short term. But I'll understand if you don't think it's worth the risk.

asottile-sentry: without the progress being checked I have no confidence in the improvements here (and no confidence that they won't regress).  we have a mechanism to make incremental improvements -- just need to use it

src/sentry/incidents/logic.py
```diff
@@ -404,7 +389,7 @@ def calculate_incident_time_range(incident, start=None, end=None, windowed_stats
 
             start = end - timedelta(seconds=time_window * WINDOWED_STATS_DATA_POINTS)
 
-    retention = quotas.get_event_retention(organization=incident.organization) or 90
+    retention = quotas.get_event_retention(organization=incident.organization) or 90  # type: ignore[attr-defined]
```
RyanSkonnord, line 392: This `type:ignore` is needed because of the hack in `src/sentry/quotas/__init__.py`:

```python
backend.expose(locals())
```

Not sure there's a way around it without factoring out that hack and converting it to an actual method call.

src/sentry/incidents/logic.py
```diff
@@ -698,17 +687,17 @@ def create_alert_rule(
...
+def snapshot_alert_rule(alert_rule: AlertRule, user: RpcUser | None = None) -> None:
     # Creates an archived alert_rule using the same properties as the passed rule
     # It will also resolve any incidents attached to this rule.
     with transaction.atomic(router.db_for_write(AlertRuleActivity)):
         triggers = AlertRuleTrigger.objects.filter(alert_rule=alert_rule)
         incidents = Incident.objects.filter(alert_rule=alert_rule)
-        snuba_query_snapshot = deepcopy(alert_rule.snuba_query)
-        snuba_query_snapshot.id = None
+        snuba_query_snapshot: SnubaQuery = deepcopy(_unpack_snuba_query(alert_rule))
+        snuba_query_snapshot.id = None  # type: ignore[assignment]
```
RyanSkonnord, line 697: This and other `type:ignore`s seem to be unavoidable on various `*.id = None` statements. I don't think I fully grasp the point of nulling out the `id` fields, but it seems to be incompatible with the base model class's declared attributes.

src/sentry/incidents/logic.py
```diff
@@ -1557,22 +1627,31 @@ def get_alert_rule_trigger_action_slack_channel_id(
...
+        # validate_channel_id, it looks like this can only lead to an error. The unit
+        # test case originates in https://github.com/getsentry/sentry/pull/58005,
+        # where it looks like it was targeting a spurious failure mode that occurs
+        # while building the description string.
+        #
+        # I'd like to move the null check upstream, clean up the associated nullable
+        # function params, remove the piecemeal checks from
+        # _get_target_identifier_display_from_target_value, and modify
+        # test_discord_channel_id_none to assert for an explicit null check (or just
+        # delete it).
```
RyanSkonnord, line 1652: Assuming no one objects to modifying the test case, I'd be ready to make these changes right away. I just didn't want to do it in the same PR.

src/sentry/incidents/logic.py
```diff
@@ -882,9 +872,9 @@ def update_alert_rule(
             # This must be done after backfill PR lands
             updated_fields["comparison_delta"] = None
             if (
-                time_window not in DYNAMIC_TIME_WINDOWS
-                or time_window is None
-                and alert_rule.snuba_query.time_window not in DYNAMIC_TIME_WINDOWS_SECONDS
+                (time_window not in DYNAMIC_TIME_WINDOWS)
+                if time_window is not None
+                else (snuba_query.time_window not in DYNAMIC_TIME_WINDOWS_SECONDS)
```
RyanSkonnord, line 877: This is a functional change, but the original check seems to have been bugged because of the precedence order between `or` and `and`. (Mypy points out that the `alert_rule.snuba_query.time_window not in DYNAMIC_TIME_WINDOWS_SECONDS` can't ever be evaluated.)

I rewrote it in this slightly awkward way to preserve what I think was the original intent as closely as I could. I suspect that it could be rewritten as

```python
            if not (
                (time_window is None or time_window in DYNAMIC_TIME_WINDOWS)
                and snuba_query.time_window in DYNAMIC_TIME_WINDOWS_SECONDS
            ):
```

but I didn't want to assume there isn't some corner case where `snuba_query.time_window` is allowed to be invalid as long as `time_window` is non-null.

The validation was added in https://github.com/getsentry/sentry/pull/74915. @mifu67, let me know if you have any input.

RyanSkonnord (commented):
Type-checking is now enforced and passing on `logic.py`.

src/sentry/incidents/logic.py
asottile-sentry: use `quotas.backend` directly

src/sentry/incidents/logic.py
```diff
@@ -598,13 +611,16 @@ def create_alert_rule(
             sensitivity=sensitivity,
             seasonality=seasonality,
             detection_type=detection_type,
-        )
+        )  # type: ignore[misc]
```
asottile-sentry, line 614: should be fixed before shipping -- what's the bug here?

src/sentry/incidents/logic.py
asottile-sentry: hmmm yeah these are a little suspicious -- I believe it's trying to duplicate the row (but get a fresh id) which seems strange -- pretty sure django gives a better way to do this but I don't know it myself

asottile-sentry: if we absolutely have to ignore these errors I would prefer if these were localized somehow instead -- sprinkling type ignores without comments is not great

src/sentry/incidents/logic.py
```diff
@@ -1290,18 +1302,16 @@ def deduplicate_trigger_actions(
     """
     actions = _prioritize_actions(triggers=triggers)
 
-    deduped = {}
+    deduped: dict[tuple[Any, ...], AlertRuleTriggerAction] = {}
```
asottile-sentry, line 1305: `tuple[Any, ...]` should be replaced with the real type

src/sentry/incidents/logic.py
```diff
@@ -1385,12 +1400,13 @@ def create_alert_rule_trigger_action(
...
-        target_identifier=target_identifier,
-        target_display=target_display,
+        target_identifier=target.identifier,
+        target_display=target.display,
         integration_id=integration_id,
         sentry_app_id=sentry_app_id,
         sentry_app_config=sentry_app_config,
-    )
+    )  # type: ignore[misc]
+    # Ignore assigning sentry_app_config (dict[str, Any]) as JSONField
```
asottile-sentry, line 1409: this means that the annotation for JSONField is incorrect -- there should be a bunch of examples for this throughout the codebase (tl;dr it needs a manual `Field` annotation because our JSONField is a huge hack!)

src/sentry/incidents/logic.py
```diff
@@ -1556,15 +1627,31 @@ def _get_alert_rule_trigger_action_slack_channel_id(
...
+        # while building the description string.
+        #
+        # I'd like to move the null check upstream, clean up the associated nullable
+        # function params, remove the piecemeal checks from
+        # _get_target_identifier_display_from_target_value, and modify
+        # test_discord_channel_id_none to assert for an explicit null check (or just
+        # delete it).
         validate_channel_id(
-            channel_id=name,
+            channel_id=name,  # type: ignore[arg-type]
```
asottile-sentry, line 1654: hmmm maybe split this out and do it first rather than following up later?

src/sentry/incidents/logic.py
RyanSkonnord: Done: https://github.com/getsentry/sentry/pull/75541/commits/a33de845d89b1bfe4d6ea2f249eb883fab54aa35

I'd just as soon define a dataclass as list out all that stuff, if it weren't a one-off.

src/sentry/incidents/logic.py
RyanSkonnord: Done: https://github.com/getsentry/sentry/pull/75541/commits/677791c53200e26fccc5acb4226bf76a15a2200e

src/sentry/incidents/logic.py
RyanSkonnord: Done: https://github.com/getsentry/sentry/pull/75541/commits/678a63f51f3837f98876cd837809e62bf6807c64

src/sentry/incidents/logic.py
RyanSkonnord: I believe you're correct about the intent. I agree that there's probably a Django utility lurking somewhere to do this and would be glad to go hunting for it later, but prefer to avoid the functional change in this PR.

This calls it out a bit better in the meantime: https://github.com/getsentry/sentry/pull/75541/commits/214ce93d1b4ca95a1769a4661c1f2d0a74af3552

src/sentry/incidents/logic.py
RyanSkonnord: I considered that but it's much easier to understand the motivation with the type annotations in place.

Here's a follow-up PR: https://github.com/getsentry/sentry/pull/76553. We could merge it into this PR or rebase it onto master after this one is merged; lmk if you have a preference between the two.

src/sentry/incidents/logic.py
asottile-sentry: wow I didn't even know django could do that

a typesafe alternative would be something like:


```python
class _OwnerKwargs(TypedDict):
    team_id: int | None
    user_id: int | None

def actor_to_alert_rule_owner_kwargs(actor: Actor | None) -> _OwnerKwargs: ...
```

and then splat that into the caller

RyanSkonnord: Nice one. Done: https://github.com/getsentry/sentry/pull/75541/commits/adf5a0d2864369bca9ca9ab567936b7cb5ae7de1

src/sentry/incidents/logic.py
RyanSkonnord: Now merged into this one.

src/sentry/incidents/logic.py
```diff
@@ -702,17 +704,26 @@ def create_alert_rule(
...
-def snapshot_alert_rule(alert_rule, user=None):
+def snapshot_alert_rule(alert_rule: AlertRule, user: RpcUser | None = None) -> None:
+    def nullify_id(model: Model) -> None:
+        """Set the id field to null.
+
+        This coerces the `save` method to create a new object.
+
+        TODO: Refactor to not violate the type system
+        """
+        model.id = None  # type: ignore[assignment]
```
RyanSkonnord, line 715: I looked into it, and apparently nulling out the primary key is the normal and recommended way to copy a model in this way ([official docs](https://docs.djangoproject.com/en/4.2/topics/db/queries/#copying-model-instances); [Stack Overflow](https://stackoverflow.com/a/4736172)).

I think the reason Mypy doesn't like it is because of our `Model` class:

https://github.com/getsentry/sentry/blob/0a4fdfd29225757dacf713d66bff4225764470f4/src/sentry/db/models/base.py#L315-L316

Changing it to
```python
    id: models.Field[int | None, int] = BoundedBigAutoField(primary_key=True)
```
seems to fix it. I couldn't find any docs that explain the difference between the two type params to `Field`; I found the above through trial and error. But I'm guessing/hoping it satisfies the meaning of "it can be nullable during a model's lifecycle as a Python object, but is always a non-null int when persisted". Should we go ahead with such a change to the `Model` class? (If so, I'd lean toward doing it as a follow-up PR and reverting this then.)

src/sentry/incidents/logic.py
RyanSkonnord: See https://github.com/getsentry/sentry/pull/75541/files#r1733575385

src/sentry/incidents/logic.py
asottile-sentry: it's actually an intentional django-stubs thing -- you almost never want `id = None` (and making it nullable causes all sorts of _other_ problems)

asottile-sentry (commented):
this patch has gotten pretty large -- can we split out parts of those and land them separately to minimize risk and increase reviewability?

RyanSkonnord:
> this patch has gotten pretty large -- can we split out parts of those and land them separately to minimize risk and increase reviewability?

It got large because that's what was necessary to remove `"sentry.incidents.logic"` from `pyproject.toml`. You want me to put it back in?

asottile-sentry:
> > this patch has gotten pretty large -- can we split out parts of those and land them separately to minimize risk and increase reviewability?
> 
> It got large because that's what was necessary to remove `"sentry.incidents.logic"` from `pyproject.toml`. You want me to put it back in?

no.  there are logical parts of this patch that can be landed separately which will reduce risk and improve reviewability

RyanSkonnord:
@asottile-sentry Functional changes broken out into:

1. https://github.com/getsentry/sentry/pull/77197
2. https://github.com/getsentry/sentry/pull/77199
3. https://github.com/getsentry/sentry/pull/77200
4. https://github.com/getsentry/sentry/pull/77201

asottile-sentry (approved):
much easier to review now -- thanks!

PR #75543, opened 2024-08-02
chore(azure): remove base_url param from get_client

cathteng:
The logger merged in this PR https://github.com/getsentry/sentry/pull/75277 [has not fired](https://cloudlogging.app.goo.gl/y6wBUPCpW7VFJgSr8). According to the code, `repo.config["instance"]` is only ever set to the installation instance, so I believe the extra `base_url` parameter from the VSTS Integration (Azure DevOps) can be removed. We also emitted the logger to check if this these two are ever mismatched and it hasn't fired.

https://github.com/getsentry/sentry/blob/76160de98ae0655baffa6a45388ea1dbc110884f/src/sentry/integrations/vsts/repository.py#L26-L43

### DIFF
--- a/src/sentry/integrations/vsts/issues.py
+++ b/src/sentry/integrations/vsts/issues.py
@@ -32,13 +32,13 @@ class VstsIssueSync(IssueSyncMixin):
 
     def create_default_repo_choice(self, default_repo: str) -> tuple[str, str]:
         # default_repo should be the project_id
-        project = self.get_client(base_url=self.instance).get_project(default_repo)
+        project = self.get_client().get_project(default_repo)
         return (project["id"], project["name"])
 
     def get_project_choices(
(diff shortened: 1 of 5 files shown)


src/sentry/integrations/vsts/issues.py
```diff
@@ -32,13 +32,13 @@ def get_persisted_default_config_fields(self) -> Sequence[str]:
 
     def create_default_repo_choice(self, default_repo: str) -> tuple[str, str]:
         # default_repo should be the project_id
-        project = self.get_client(base_url=self.instance).get_project(default_repo)
```
cathteng, line 35: all of the `get_client` calls in here pass `self.instance `which is the default in `get_client`

RyanSkonnord (approved):

Issue #74861, opened 2024-07-24
[user feedback] ASANA Integration


elizabethmeter:
### Problem Statement

The ASANA integration is terrible. Please include a summary of the feedback.
Reason: I have a team and we are all in ASANA and not developers, I would rather not have them all in Sentry as well as ASANA and we have decided as a company that ASANA is our primary issue SAAS.

### Solution Brainstorm

The Click up integration has a nice summery that would solve all my problem if you did the same thing or similar 

### Product Area

User Feedback

leedongwei:
Sorry for the subpar experience. My team is starting to take ownership of project management integrations. We'll start work on feature parity between Asana vs Jira vs other integrations in Q4 this year.

joranbeasley:
i think this would just take the following 2 lines



![Image](https://github.com/user-attachments/assets/58d2d7d0-5817-43a7-b93a-09210f901f09)

vitor-neurodelabs:
I'm having a problem with the Asana integration as well for my team, when I click to authorise the integration it leads to an error page where it says the `redirect_uri` parameter is not valid.

Would like to be able to open issues in an Asana project based on the alerts we get from Sentry.

Do you guys have any deadline set-up for when this is going to be delivered?

Much appreciated :D

sentaur-athena:
@vitor-neurodelabs asana is built over our deprecated integration platform called plugins. We don't provide support for plugins anymore but for this specific issue we opened a ticket with Asana to fix the redirect_uri, so when they get to it it will be resolved. Unfortunately, it cannot be edited from our side since the app is already published.

sentaur-athena:
Asana responded to our ticket that this issue is fixed.

PR #75546, opened 2024-08-02
feat(asana-integration):default description and name

joranbeasley:
add default values for name and notes

resoves issue #74861

### Legal Boilerplate


leedongwei:
@joranbeasley Thanks! I'll get someone to review this to unblock https://github.com/getsentry/sentry/issues/74861

cathteng:
i added a test, not sure if that's exactly what this PR fixes so feel free to fix it up

joranbeasley:
> i added a test, not sure if that's exactly what this PR fixes so feel free to fix it up

I think it supposed to make it so that the actual issue description and title is present in the Asana task

I based my fix on the custom UI for integration works ... Which might not be how the fields are populated with built-in python integration... Ie it might not do the same string substitutions... But I'm assuming it does..

PR #75547, opened 2024-08-02
chore(onboarding): Cleanup ruby/rails onboarding a bit more

sl0thentr0py:


priscilawebdev:
When selecting the rack platform, we display the following under the configure section:

image

We highlight the addition of `use` but shouldn't we say something like: 

"You can then change the Sentry configuration as follows. Please make sure to add use Sentry::Rack::CaptureExceptions...."

priscilawebdev:
nit: by selecting the ruby platform, we give more information about the DSN, such as, where to find it. Do we need to provide this information during onboarding? Users are already provided with the DSN of the new created project.

image

priscilawebdev (approved):
left a few more suggestions, but the PR looks good to me..Thanks for doing that :)

smeubank (approved):

PR #75548, opened 2024-08-02
feat(performance): track has data on performance table

KevinL10:
Tracks whether the transactions table on the performance overview page contains data. The metric is in one of 3 states:
- `has_data`: table has data
- `no_data`: table loads but is empty
- `empty`: the selected project(s) have no data and we display the empty state screen

### DIFF
--- a/static/app/utils/analytics/performanceAnalyticsEvents.tsx
+++ b/static/app/utils/analytics/performanceAnalyticsEvents.tsx
@@ -79,8 +79,7 @@ export type PerformanceEventParameters = {
     metric: string;
   };
   'performance_views.overview.has_data': {
-    // either "has_data", "no_data", or "empty" (when the user only sees the onboarding screen)
-    table_data_state: string;
+    table_data_state: 'has_data' | 'no_data' | 'onboarding';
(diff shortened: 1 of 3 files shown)


static/app/utils/analytics/performanceAnalyticsEvents.tsx
```diff
@@ -77,6 +78,11 @@ export type PerformanceEventParameters = {
   'performance_views.overview.change_chart': {
     metric: string;
   };
+  'performance_views.overview.has_data': {
+    // either "has_data", "no_data", or "empty" (when the user only sees the onboarding screen)
+    table_data_state: string;
```
mjq, line 83: If there are only 3 possible values, might be worth using a [string literal type union](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types) here. That almost makes the comment superfluous, but the difference between `no_data` and `empty` might not be clear without it.

I don't know if we're already using `empty` elsewhere to mean the onboarding state, but calling it `onboarding` instead might make the meaning more obvious so you don't need to explain it in a comment at all? Just a thought.

mjq (approved):

PR #75549, opened 2024-08-02
ref(control_silo): move logic from old to new slack tasks

Christinarlong:
Moves logic and updates import routes to the new slack tasks located in sentry/integrations/slack/tasks

ref(issue # here)
Split of (https://github.com/getsentry/sentry/pull/74925)


src/sentry/integrations/slack/tasks/find_channel_id_for_rule.py
```diff
@@ -23,6 +34,82 @@ def find_channel_id_for_rule(
...
+    except Project.DoesNotExist:
+        redis_rule_status.set_value("failed")
+        return
+
+    organization = project.organization
+    integration_id: int | None = None
+    channel_name: str | None = None
+
+    # TODO: make work for multiple Slack actions
+    for action in actions:
```
markstory, line 50: If you get mypy errors here, remember that the type for `actions` is wrong, and the parameter is actually a `Sequence[dict[str, Any]]` based on the call sites.

markstory (approved):
Looks good.

markstory:
reverted because we missed entries in celery_imports.

PR #75552, opened 2024-08-02
ref(replay): migrate SmartSearchbar to SearchQueryBuilder

### LINKED ISSUE #75006: SearchQueryBuilder: Replace SmartSearchBar in replays page
## Background

`<SearchQueryBuilder />` is meant to be the next generation of `<SmartSearchBar />` and should ideally be replaced in all app surfaces. At this time, it is only implemented on issues pages. Because different parts of the app use this search bar differently, we need to implement and test it thoroughly ensure that is works for all use cases.

## Requirements

- Under a feature flag for this product area (already created), render `SearchQueryBuilder` instead of `SmartSearchBar`
-  If your search bar has a large number of filter keys, implement the `filterKeySections` prop to display them in an organized fashion. How your filter keys get categorized is up to you! 
- If you come across any missing feature that are critical to your use case, please note them and bring this up to the issues team.

## Resources

- See [here for SearchQueryBuilder documentation](https://sentry.sentry.io/stories/?name=app/components/searchQueryBuilder/index.stories.tsx) which should help you understand how to use the component and how `SmartSearchBar` props map to the new ones.
- A list of feature flags has been generated for each product surface. [See here for the full list](https://github.com/getsentry/sentry-options-automator/pull/2018/files). These flags are already enabled for internal orgs.

aliu39:
Closes https://github.com/getsentry/sentry/issues/75006
See https://sentry.sentry.io/stories/?name=app/components/searchQueryBuilder/index.stories.tsx
Used the [issue's search bar](https://github.com/getsentry/sentry/blob/f74ff05a30510bbd0ecb60fb1bfa0d912e6f5a5d/static/app/views/issueList/searchBar.tsx#L46) as a reference.

~~One thing I had trouble with was recent searches~~ Recent searches isn't supported yet.

The flag is enabled for internal orgs only, maybe we can merge this and dogfood a bit (the new FF toggles will help :)

Demo .mov: https://drive.google.com/file/d/1DXksPM0QURpzeKqVzH8ih6ir7pS8Vhk7/view?usp=sharing

Follow-up: 
- [ ] `useFetchOrganizationTags`
- [ ] Try sorting by `-totalValues` for most popular tags

Omitted props:
From SearchQueryBuilder:
// label
// onChange
// disallowLogicalOperators

From SmartSearchBar (unsupported in new component)
// maxMenuHeight={500}
// maxQueryLength={MAX_QUERY_LENGTH}
// pageFilters
// prepareQuery={prepareQuery}
// projectIds={projectIds} (only used for multi-project search analytics, which I don't think we use)


static/app/views/replays/list/replaySearchBar.tsx
```diff
@@ -63,23 +69,70 @@ function getSupportedTags(supportedTags: TagCollection) {
...
+  }
+
+  const nonFieldTags: string[] = Object.values(supportedTags)
+    .map(tag => tag.key)
+    .filter(
+      key =>
+        !REPLAY_FIELDS.includes(key as ReplayFieldKey | FieldKey) &&
+        !REPLAY_CLICK_FIELDS.includes(key as ReplayClickFieldKey)
+    );
+  nonFieldTags.sort();
```
malwilley, line 90: Something which we did on the issues page is sort these tags by `totalValues`. That way the top tags you see are ones which are used the most.

static/app/views/replays/list/replaySearchBar.tsx
```diff
@@ -63,23 +69,70 @@ function getSupportedTags(supportedTags: TagCollection) {
...
-  placeholder?: string;
 };
 
 function ReplaySearchBar(props: Props) {
-  const {organization, pageFilters, placeholder} = props;
+  const {organization, pageFilters} = props;
   const api = useApi();
   const projectIds = pageFilters.projects;
-  const tags = useTags();
+  const organizationTags = useTags();
```
malwilley, line 120: This is how the tags were being fetched before so it doesn't need to be changed in this PR, but I did want to point out that this uses tag data from the global store which fetches them from _all_ datasets. That means that it takes a long time to load and also may contain tags which aren't actually in the replay dataset.

For issues, we are using `useFetchOrganizationTags` and sending the applicable dataset, which returns much faster and is more accurate. Just throwing that out there!

malwilley (approved):
Thanks for migrating! The search bar props look good 👍

PR #75555, opened 2024-08-02
fix(custom-views): Draggable Tabs bug fixes and improvements

MichaelSun48:
This PR makes several necessary improvements/fixes to the DraggableTabs family of components. These fixes are somewhat disconnected, so I'll annotate the diff with what each change fixes.

### DIFF
--- a/static/app/components/draggableTabs/draggableTabList.tsx
+++ b/static/app/components/draggableTabs/draggableTabList.tsx
@@ -1,4 +1,4 @@
-import {useContext, useEffect, useMemo, useRef} from 'react';
+import {Fragment, useContext, useEffect, useMemo, useRef} from 'react';
 import styled from '@emotion/styled';
 import type {AriaTabListOptions} from '@react-aria/tabs';
 import {useTabList} from '@react-aria/tabs';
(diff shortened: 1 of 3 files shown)


static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -130,30 +130,31 @@ function BaseDraggableTabList({
...
+                  variant={tabVariant}
+                />
+              </Reorder.Item>
               {(state.selectedKey === 'temporary-tab' ||
                 (state.selectedKey !== item.key &&
                   state.collection.getKeyAfter(item.key) !== state.selectedKey)) && (
                 <TabDivider />
               )}
-            </Reorder.Item>
+            </Fragment>
```
MichaelSun48, line 157: Separate the `<TabDivider/>` from the `<Reorder.Item/>` component. This previously allowed users to drag the tab from the tab divider, which is visually separate from the tab.

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -268,7 +269,6 @@ const TabListWrap = styled('ul', {
   margin: 0;
   list-style-type: none;
   flex-shrink: 0;
-  padding-left: 15px;
```
MichaelSun48, line 271: This padding was here to make the stories component match the figma designs, but this padding is already accounted for in the issue stream. 

image

static/app/views/issueList/groupSearchViewTabs/draggableTabBar.tsx
```diff
@@ -166,10 +170,9 @@ export function DraggableTabBar({
...
 
   return (
-    <Tabs>
+    <Tabs value={selectedTabKey} onChange={setSelectedTabKey}>
       <DraggableTabList
         onReorder={onReorder}
-        onSelectionChange={setSelectedTabKey}
         selectedKey={selectedTabKey}
         showTempTab={showTempTab}
         onAddView={e => handleOnAddView(e)}
```
MichaelSun48, line 178: `setSelectedTabKey` was mistakenly put in the  `onSelectionChange` prop - this overrode the default behavior that controls link switching. The correct place to put this is in the onChange handler in the top level Tabs component.

static/app/views/issueList/groupSearchViewTabs/draggableTabBar.tsx
```diff
@@ -188,9 +192,14 @@ export function DraggableTabBar({
                 setIsEditing={isEditing => setEditingTabKey(isEditing ? tab.key : null)}
                 onChange={newLabel => handleOnTabRenamed(newLabel.trim(), tab.key)}
               />
-              {tab.key !== 'temporary-tab' && tab.queryCount && (
+              {tab.key !== 'temporary-tab' && tab.queryCount !== undefined && (
                 <StyledBadge>
-                  <QueryCount hideParens count={tab.queryCount} max={1000} />
+                  <QueryCount
+                    hideParens
+                    hideIfEmpty={false}
```
MichaelSun48, line 199: This was causing some weird behavior if queryCount was 0 - it would show the 0 but not within the badge, so the tab title would look like "Prioritized0". This change fixes that bug, and it also will display a 0 in the badge, though we can hide the 0 query count by setting `hideIfEmpty` to true in the future if we want

malwilley (approved):

PR #75569, opened 2024-08-03
ref(replay): Refactor currentTime context field

ryan953:
Before we had our `currentTime` stored as state inside ReplayContext. This meant that whenever any state inside ReplayContext changed, everything would re-render. This was bad because there's two kinds of things on the page: those that depend on currentTime, and everything else; they should re-render at different frequencies but` couldn't because of the shared Context and massive state.

So now we have a new hook for accessing `currentTime`, called: `useReplayCurrentTime`. 
- The first thing to notice is that it's moving the state value that we had before out of ReplayContext and into all the callsites around the codebase. This benefits anything that depends on ReplayContext because changes to `currentTime` will not force re-renders. 
- The second thing to notice is that the API for `useReplayCurrentTime` provides a callback, as well as a stable getter function. So we can (in a followup PR) avoid setting state in all the callsites by either: a) calling the getting as needed (see `useShareReplayAtTimestamp` for an example) or b) we can calculate a more complex state that changes less often inside the callback (no example for this right now) and store that instead of a frequently changing timestamp. Using those techniques means we can reduce re-renders for places that depend on `currentTime`.

This PR focuses on changing over the interfaces of things, and is very mechanical. There's no render performance improvement for anything because we still have that `const isBuffering` statement inside of ReplayContext which depends on `useCurrentTime(getCurrentPlayerTime)` which itself has state inside it :(

But! once that `useCurrentTime` internal hook is gone we'll be able to see render improvements. Here's how changes to `useShareReplayAtTimestamp` will result in better re-render perf for the whole header:
| Before | After |
| --- | --- |
| before | after |

### DIFF
--- a/static/app/components/events/eventReplay/replayPreviewPlayer.tsx
+++ b/static/app/components/events/eventReplay/replayPreviewPlayer.tsx
@@ -20,6 +20,7 @@ import EventView from 'sentry/utils/discover/eventView';
 import getRouteStringFromRoutes from 'sentry/utils/getRouteStringFromRoutes';
 import {TabKey} from 'sentry/utils/replays/hooks/useActiveReplayTab';
 import useMarkReplayViewed from 'sentry/utils/replays/hooks/useMarkReplayViewed';
+import useReplayCurrentTime from 'sentry/utils/replays/playback/hooks/useReplayCurrentTime';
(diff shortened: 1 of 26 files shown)


ryan953:
damn clips and startOffset timestamps are messing this up. The logic for that is all over the place :(

PR #75575, opened 2024-08-05
fix(tooltips): Make tooltips within a modal interactive

priscilawebdev:
### Problem

Tooltips rendered inside a modal are currently non-interactive. Users cannot select content or click on links within these tooltips.

https://github.com/user-attachments/assets/e30ff3e8-4f8a-4825-ab10-1659c4fac03d

### Solution

This PR updates the `createPortal` function for tooltips. It now checks if the modal is visible; if so, it renders the tooltip within the modal's DOM node, solving the issue.

https://github.com/user-attachments/assets/a3bc5f69-75b2-4b2b-a337-4bce73c329a5


ArthurKnaus (approved):

aliu39 (approved):

aliu39:
Nice!

priscilawebdev:
got feedback from @evanpurkhiser and will work soon on the ticket https://github.com/getsentry/sentry/issues/75624 to improve the logic - For now, I’m merging this to resolve the issue

Issue #57828, opened 2023-10-10
Ignore core modules' CODEOWNERS in traces


DavidCain:
### Problem Statement

Duplicate of https://github.com/getsentry/sentry/issues/54012, which was closed (and my requests to re-open haven't been seen).

Broadly, we'd like to have Sentry issues (whose stacktraces include some core application logic) assigned based on the faulty code, not the core modules involved.

## The problem
Certain core modules have a tendency to appear in _every_ stack trace (e.g. a module which provides a decorator to all asynchronous tasks will show up in all async task failures). The `CODEOWNERS` feature sees these core modules in the stack trace, and assigns all possible errors pertaining to async tasks to the team which maintains the core module. This has the effect of never getting the right team's attention. We'd like to consider these core modules "safe," exactly as if they were third-party modules.

## An illustrative example

Consider this simple stacktrace:
```
Traceback (most recent call last):
  File "example.py", line 3, in <module>
    bad_ideas.main()
  File "src/bad_ideas.py", line 5, in main
    numbers.divide(numerator=10, denominator=0)
  File "src/numbers.py", line 2, in divide
    return numerator / denominator
ZeroDivisionError: division by zero
```

And couple that trace with a simple `.github/CODEOWNERS`:
```
src/bad_ideas.py  @engineering
src/numbers.py    @mathematicians
```

When this stacktrace becomes a Sentry issue (in a project that uses the `CODEOWNERS` feature), the `@mathematicians` team (owners of `numbers.py`) is assigned the issue. However, the `divide` method works just fine -- the real bug is in `bad_ideas.py` (those silly engineers tried to divide by zero!).

With any given stacktrace, it's not possible to know which frame of the trace is _responsible_ for the bug (maybe the owners of `example.py` should be assigned; who knows). However, we'd really like the ability to consider certain core modules (in this case, `numbers.py`) to be presumed "safe" and for ownership assignment to be applied based on any other files in the trace.

### In-app frames
Sentry already exposes the ability to mark certain parts of a stack-trace as coming from code that is not in-app: https://docs.sentry.io/product/data-management-settings/event-grouping/stack-trace-rules/

If the `CODEOWNERS` assignment feature ignored frames which are not "in-app," that would solve this nicely. In other words, it would work perfectly fine if we treated this core modules like third-party dependencies (if `numpy` is in your stacktrace, it's presumed that `numpy` probably isn't the source of the problem -- it's the code that _calls_ `numpy` methods).

[...]

PR #75588, opened 2024-08-05
Filter out in-app frames when matching for CODEOWNERS

alxckn:
In the [documentation for ownership rules](https://docs.sentry.io/product/issues/ownership-rules/#evaluation-flow), about CODEOWNERS file evaluation on an event's frames, it is stated that "Sentry evaluates an event’s in-app frames against the rules".

As discussed in [this issue](https://github.com/getsentry/sentry/issues/57828), this does not seem to be in fact the case (see [this comment with a reproduction scenario specifically](https://github.com/getsentry/sentry/issues/57828#issuecomment-2052442150)).

This PR proposes changes to ensure the clause is respected for CODEOWNERS matchers. When matching frames, if they are explicitely not in app (`frame.get("in_app" == False`), they will be ignored. 

### Legal Boilerplate


JoshFerge:
thank you for the contribution @alxckn! this change looks good to me. I added one test case and will go ahead and merge after CI passes!

JoshFerge (approved):

PR #75594, opened 2024-08-05
Implement search bar for releases index

nhsiehgit:
per https://github.com/getsentry/sentry/issues/74991

### DIFF
--- a/static/app/views/releases/list/index.tsx
+++ b/static/app/views/releases/list/index.tsx
@@ -21,6 +21,7 @@ import PageFiltersContainer from 'sentry/components/organizations/pageFilters/co
 import {ProjectPageFilter} from 'sentry/components/organizations/projectPageFilter';
 import Pagination from 'sentry/components/pagination';
 import Panel from 'sentry/components/panels/panel';
+import {SearchQueryBuilder} from 'sentry/components/searchQueryBuilder';
 import SmartSearchBar from 'sentry/components/smartSearchBar';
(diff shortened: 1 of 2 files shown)


static/app/views/releases/list/index.tsx
```diff
@@ -563,23 +579,27 @@ class ReleasesList extends DeprecatedAsyncView<Props, State> {
 
               {this.shouldShowQuickstart ? null : (
                 <SortAndFilterWrapper>
-                  <StyledSmartSearchBar
-                    searchSource="releases"
-                    query={this.getQuery()}
-                    placeholder={t('Search by version, build, package, or stage')}
-                    hasRecentSearches={false}
-                    supportedTags={{
```
malwilley, line 571: Looks like we lost this in the conversion - SmartSearchBar should still send `supportedTags`

static/app/views/releases/list/index.tsx
```diff
@@ -285,6 +286,21 @@ class ReleasesList extends DeprecatedAsyncView<Props, State> {
     });
   };
 
+  filterKeys = () => {
```
malwilley, line 289: Looks like the filter keys are constant. Could we define this as a property (`FILTER_KEYS`) outside of the component to prevent unnecessary rerenders?

static/app/views/releases/list/index.tsx
nhsiehgit: oops thanks

malwilley (approved):
Thanks for implementing!

PR #75597, opened 2024-08-05
chore(control_silo): Move User model to new home in users package

Christinarlong:
Part of consolidating (user) control silo resources to be located with other user related models.

Ref(#73856)


src/sentry/users/models/user.py
Christinarlong, line 1: Added a lot of return & parameter typing for this file, pls double check if the types are wrong. I'll point out some specific ones I'm unsure on

src/sentry/users/models/user.py
```diff
@@ -0,0 +1,605 @@
...
+
+# HACK(dcramer): last_login needs nullable for Django 1.8
+User._meta.get_field("last_login").null = True
+
+
+# When a user logs out, we want to always log them out of all
+# sessions and refresh their nonce.
+@receiver(user_logged_out, sender=User)
+def refresh_user_nonce(
+    sender: User | RpcUser, request: HttpRequest, user: User | None, **kwargs: Any
```
Christinarlong, line 587: unsure if request can be `None` too ?

markstory: All of the current callers provide the request, *but* signals make no promises on providing all of their parameters. Annotating this as `HttpRequest | None` seems wise.

src/sentry/users/models/user.py
```diff
@@ -0,0 +1,606 @@
...
+) -> None:
+    if user is None:
+        return
+    user.refresh_session_nonce()
+    user.save(update_fields=["session_nonce"])
+
+
+@receiver(user_logged_out, sender=RpcUser)
+def refresh_api_user_nonce(
+    sender: RpcUser, request: HttpRequest, user: User | None, **kwargs: Any
```
markstory, line 598: Same here, I would assume that all parameters could be `None`.

markstory (approved):
Looks good to me. Most of the changes are import updates. I didn't spot any logic differences in the `User` model.

PR #75598, opened 2024-08-05
feat(custom-views): Add drag barriers to tab bar

MichaelSun48:
This PR adds drag constraints to the draggable tablist component. That is, it defines a container that the tabs can be dragged within. ~I've set this container to be the entire bar, including the empty area to the right of the add view button, though dragging the tab beyond the add view button will just snap it back to the last position. If necessary, we can change this container to be just within the section left of the Add View button.~

Update: Due to conflicting behavior with the overflow menu, I have change the behavior to only allow tabs to be dragged within the persistent tabs container (to the left of the add view). I have also removed the overflow menu entirely from this PR due to incompatibility with the draggable tabs component. This logic was copied directly over from the existing tabs component, but it needs to be refactored for this component since there are more elements in the tab bar now. This refactor will come in a future PR. 

Before:  

https://github.com/user-attachments/assets/0bdc21d6-53e0-45ca-850a-1984f02189d1

After: 

https://github.com/user-attachments/assets/6f49218e-3823-426f-bd53-b19e8be8779d

### DIFF
--- a/static/app/components/draggableTabs/draggableTabList.tsx
+++ b/static/app/components/draggableTabs/draggableTabList.tsx
@@ -6,15 +6,12 @@ import {useCollection} from '@react-stately/collections';
 import {ListCollection} from '@react-stately/list';
 import type {TabListStateOptions} from '@react-stately/tabs';
 import {useTabListState} from '@react-stately/tabs';
-import type {Node, Orientation} from '@react-types/shared';
+import type {Node} from '@react-types/shared';
 import {motion, Reorder} from 'framer-motion';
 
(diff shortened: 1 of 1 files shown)


malwilley (approved):
Works well! Looks like you removed the overflow logic which we will need in some form, but that may never have been working?

MichaelSun48:
> Works well! Looks like you removed the overflow logic which we will need in some form, but that may never have been working?

@malwilley  Yup exactly that - in hindsight I should have removed it from the beginning since its either not compatible with our code or needs to be refactored. That's the next thing I'm working on

PR #75600, opened 2024-08-05
chore(uptime): Include status reason in the `handle_result_for_project` stat

wedamija:
This helps us understand the reasons for failures in the result consumer graph

### DIFF
--- a/src/sentry/uptime/consumers/results_consumer.py
+++ b/src/sentry/uptime/consumers/results_consumer.py
@@ -99,6 +99,7 @@ class UptimeResultProcessor(ResultProcessor[CheckResult, UptimeSubscription]):
         metrics.incr(
             "uptime.result_processor.handle_result_for_project",
             tags={"status_reason": status_reason, **metric_tags},
+            sample_rate=1.0,
         )
         cluster = _get_cluster()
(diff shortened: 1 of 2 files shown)


src/sentry/uptime/consumers/results_consumer.py
```diff
@@ -91,7 +91,15 @@ def handle_result_for_project(
             "status": result["status"],
             "mode": ProjectUptimeSubscriptionMode(project_subscription.mode).name.lower(),
         }
-        metrics.incr("uptime.result_processor.handle_result_for_project", tags=metric_tags)
+
+        status_reason = "none"
+        if result["status_reason"]:
+            status_reason = result["status_reason"]["type"]
```
evanpurkhiser, line 97: Should we just shove it into `metric_tags`?

evanpurkhiser (approved):

evanpurkhiser (approved):

src/sentry/uptime/consumers/results_consumer.py
wedamija: We could... it didn't seem relevant to the other stats though

PR #75605, opened 2024-08-05
ref(similarity): Limit stacktrace to 30 chained exceptions

jangjodi:
Limit stacktrace to 30 chained exceptions in similarity stacktrace parsing
Add distribution metric for number of exceptions


src/sentry/seer/similarity/utils.py
```diff
@@ -43,9 +43,12 @@ def get_stacktrace_string(data: dict[str, Any]) -> str:
     found_non_snipped_context_line = False
     result_parts = []
 
+    metrics.distribution("seer.grouping.exceptions.length", len(exceptions))
+
     # Reverse the list of exceptions in order to prioritize the outermost/most recent ones in cases
     # where there are chained exceptions and we end up truncating
-    for exception in reversed(exceptions):
+    # Limit the number of chained exceptions to 30
+    for exception in reversed(exceptions[:30]):
```
snigdhas, line 51: Can we define this as a constant?

snigdhas (approved):

src/sentry/seer/similarity/utils.py
```diff
@@ -43,9 +44,12 @@ def get_stacktrace_string(data: dict[str, Any]) -> str:
     found_non_snipped_context_line = False
     result_parts = []
 
+    metrics.distribution("seer.grouping.exceptions.length", len(exceptions))
+
     # Reverse the list of exceptions in order to prioritize the outermost/most recent ones in cases
     # where there are chained exceptions and we end up truncating
-    for exception in reversed(exceptions):
+    # Limit the number of chained exceptions to 30
```
lobsterkatie, line 51: Nit: Since the number is now controlled by a variable, leaving it out of the comment means we don't have to remember to update the comment if/when we update the value.

```suggestion
    # Limit the number of chained exceptions
```

tests/sentry/seer/similarity/test_utils.py
```diff
@@ -634,6 +634,22 @@ def test_chained_too_many_frames_minified_js_frame_limit(self):
                 == expected_frame_count
             )
 
+    def test_chained_too_many_exceptions(self):
+        """Test that we restrict number of chained exceptions to 30."""
+        data_chained_exception = copy.deepcopy(self.CHAINED_APP_DATA)
+        data_chained_exception["app"]["component"]["values"][0]["values"] = [
+            self.create_exception(
+                exception_type_str="Exception",
+                exception_value=f"value {i}!",
```
lobsterkatie, line 643: Nit: This makes it clearer down in  the assertions what we're asserting about.

```suggestion
                exception_value=f"exception {i} message!",
```

lobsterkatie (approved):
LGTM!

PR #75608, opened 2024-08-05
feat(issue-details): Updated suspect commit design

roggenkemper:
this pr updates the design of the suspect commit banner. for now, the updated design does not support multiple suspect commits, which is something that will be handled in a future pr.

before:
Screenshot 2024-08-05 at 2 25 45 PM

after:
Screenshot 2024-08-07 at 3 27 01 PM


static/app/components/events/suspectCommits.tsx
```diff
@@ -33,6 +39,13 @@ export function SuspectCommits({group, eventId, project, commitRow: CommitRow}:
   });
   const committers = data?.committers ?? [];
 
+  const hasStreamlinedUI = useHasStreamlinedUI();
+
+  const [suspectCommitDismissed, setSuspectCommitDismissed] = useLocalStorageState(
+    `suspect-commit-dismissed-${eventId}`,
+    false
+  );
```
leeandher, line 47: TBH i don't really get why we'd let people dismiss suspect commits, its pretty different than the rest of the promos. Maybe if they hadn't set it up already? Either way right now what happens is the dismiss doesn't actually hide the commits, just displays the old UI.

leeandher: Also event IDs aren't guarenteed unique for different projects, so the local storage keys could collide, though I think we just shouldn't allow dismissing it in general, just the promo banner (which wouldn't need to be keyed on anything i think)

static/app/components/events/suspectCommits.tsx
```diff
@@ -80,7 +93,34 @@ export function SuspectCommits({group, eventId, project, commitRow: CommitRow}:
 
   const commitHeading = tn('Suspect Commit', 'Suspect Commits (%s)', commits.length);
 
-  return (
+  const firstCommit = commits[0];
```
leeandher, line 96: Hmm, I wonder we show the rest of the commits without falling back to the old UI. You may have to follow up with @vuluongj20

static/app/components/events/suspectCommits.tsx
```diff
@@ -80,7 +93,34 @@ export function SuspectCommits({group, eventId, project, commitRow: CommitRow}:
...
+
+  return hasStreamlinedUI && !suspectCommitDismissed ? (
+    <StreamlinedPanel>
```
leeandher, line 99: Looks fire 🔥

leeandher (commented):
Looks good and 🚀 👏  for writing tests, but falling back to the old UI feels a bit odd as a user, especially since it reveals more info with the other commits. I think we should follow up with the team for how to handle that. Components look good though!

scttcper (approved):

static/app/components/events/suspectCommits.tsx
roggenkemper: will add project to make the key more unique, showing the old UI on dismiss is just a bug in my conditional. will fix. i think for now, especially since it's not a binding change (it's easy to un-dismiss as we test), ill leave dismissing in. checked with vu and that seems to be the direction we want to move in.

static/app/components/events/suspectCommits.tsx
roggenkemper: showing the rest of the commits will be handled in a future PR, the scope for this PR is limited to just showing the first commit, this fallback is a bug that ill fix

PR #75609, opened 2024-08-05
feat(anomaly detection): Add check for seven days of data to update_alert_rule

mifu67:
Updates the alert rule status to NOT_ENOUGH_DATA in `update_alert_rule` if changing to a dynamic detection type and 7+ days of data are not present.

Closes https://getsentry.atlassian.net/browse/ALRT-191

### DIFF
--- a/src/sentry/incidents/logic.py
+++ b/src/sentry/incidents/logic.py
@@ -911,12 +911,18 @@ def update_alert_rule(
                     setattr(alert_rule, k, v)
 
                 try:
-                    send_historical_data_to_seer(
+                    rule_status = send_historical_data_to_seer(
                         alert_rule=alert_rule,
                         project=projects[0] if projects else alert_rule.projects.get(),
                     )
+                    if rule_status == AlertRuleStatus.NOT_ENOUGH_DATA:
+                        # if we don't have at least seven days worth of data, then the dynamic alert won't fire
+                        alert_rule.update(status=AlertRuleStatus.NOT_ENOUGH_DATA.value)
                 except (TimeoutError, MaxRetryError):
                     raise TimeoutError("Failed to send data to Seer - cannot update alert rule.")
+                except (ValidationError):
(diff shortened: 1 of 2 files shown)


src/sentry/incidents/logic.py
```diff
@@ -911,12 +911,18 @@ def update_alert_rule(
...
+                    rule_status = send_historical_data_to_seer(
                         alert_rule=alert_rule,
                         project=projects[0] if projects else alert_rule.projects.get(),
                     )
+                    if rule_status == AlertRuleStatus.NOT_ENOUGH_DATA:
+                        # if we don't have at least seven days worth of data, then the dynamic alert won't fire
+                        alert_rule.update(status=AlertRuleStatus.NOT_ENOUGH_DATA.value)
                 except (TimeoutError, MaxRetryError):
                     raise TimeoutError("Failed to send data to Seer - cannot update alert rule.")
+                except (ValidationError):
```
ceorourke, line 923: Unless you have more than one exception you don't need the parens here

tests/sentry/incidents/test_logic.py
```diff
@@ -1662,6 +1662,56 @@ def test_update_dynamic_alert_percent_to_dynamic(self, mock_seer_request):
         )
         assert mock_seer_request.call_count == 1
 
+    @with_feature("organizations:anomaly-detection-alerts")
+    @patch(
+        "sentry.seer.anomaly_detection.store_data.seer_anomaly_detection_connection_pool.urlopen"
+    )
+    def test_update_alert_rule_static_to_dynamic_enough_data(self, mock_seer_request):
+        # assert that the status is PENDING if enough data exists
```
ceorourke, line 1670: Typically for test docstrings they're formatted like
```
"""
Assert that the status is PENDING if enough data exists
"""
```

ceorourke (approved):
Left some stylistic nits but otherwise LGTM

PR #75614, opened 2024-08-05
WIP: update searchbar in alert rule form w/ searchQueryBuilder

nhsiehgit:


nhsiehgit:
🤔 

I think modifying the `SearchBar` component was the wrong move...
didn't quite realize just how much was using this.

instead it looks like i'll want to _just_ update the `ruleConditionForm`?
but that means i'll need to duplicate quite a bit of logic from the `SearchBar` component that's _not_ being passed in?

thoughts? @malwilley

PR #75615, opened 2024-08-05
feat(data-secrecy): Create DataSecrecyWaiver Model

iamrajjoshi:
Based on the conversation here, https://github.com/getsentry/getsentry/pull/14779#issuecomment-2264102902, we decided to move all the data secrecy logic to sentry. i already deleted the model & references from getsentry.

here, i build the model again. the model is essentially the clone of https://github.com/getsentry/getsentry/pull/13952, with the major change being relocation scope set to Organization


markstory (approved):
Schema looks good.

PR #75617, opened 2024-08-05
ref(alerts): Catch more errors for anomaly detection

ceorourke:
While doing prod testing for anomaly detection and expecting to hit Seer timeouts, nothing prevented the alerts from being created or edited despite the service not being available yet. I suspect there are errors other than the specific ones we're watching for, so we should check for what those could be and handle them appropriately.

### DIFF
--- a/src/sentry/seer/anomaly_detection/store_data.py
+++ b/src/sentry/seer/anomaly_detection/store_data.py
@@ -32,6 +32,10 @@ seer_anomaly_detection_connection_pool = connection_from_url(
 )
 
 
+class UnexpectedSeerError(Exception):
+    pass
+
+
 def format_historical_data(data: SnubaTSResult) -> list[TimeSeriesPoint]:
     """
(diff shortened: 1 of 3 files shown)


src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -118,6 +122,18 @@ def send_historical_data_to_seer(alert_rule: AlertRule, project: Project) -> Ale
...
+            "Unexpected error sending data to Seer",
+            exc_info=True,
+            extra={"rule_id": alert_rule.id, "project_id": project.id},
+        )
+        raise UnexpectedSeerError("Unexpected error from Seer. Cannot store historical data.")
+    else:
+        logger.info(
+            "Data successfully sent to Seer",
+            extra={"rule_id": alert_rule.id, "project_id": project.id},
+        )
```
schew2381, line 136: Do we need this success log? If nothing fails you could assume everything passed as expected

PR #75625, opened 2024-08-06
various date/time-related fixes

rockdrilla:


wedamija (changes requested):
Sorry, these migrations can't be run on our production systems because they'll require a full table rewrite. Is there a reason you'd like to increase the size of these columns?

rockdrilla:
Unfortunately, we're hitting integer overflows on these columns in our self-hosted setup which leads to sentry workers getting stuck and [other] queues growing with blazing speed.

rockdrilla:
@wedamija can't access anymore your reviewed changes after force-pushing branch. May I ask you to review once more?

wedamija:
> Unfortunately, we're hitting integer overflows on these columns in our self-hosted setup which leads to sentry workers getting stuck and [other] queues growing with blazing speed.

Does this mean you have group(s) in your table that have seen over 2 biliion events? In our own production setup, there are none so far with over 1 billion.

Are you able to tell me the maximum value of both the `times_seen` and `score` column in your `sentry_groupedmessages` table?

wedamija:
> @wedamija can't access anymore your reviewed changes after force-pushing branch. May I ask you to review once more?

My main comment is that we can't merge this as is without a lot of work on our end - upgrading the size of a column from and int to a bigint requires a full table lock and will take down our production systems until the rewrite is completed. For us to manage this safely we'd need to figure out a way to manage this change without locking the table.

I'm happy to work with you to try and fix the problems in your own system though.

PR #75626, opened 2024-08-06
feat(source-maps-debugger): Use new user feedback

### LINKED ISSUE #75623: Use the official User Feedback modal for source map debugger
While we're at it we can also remove the beta tag.

lforst:
Resolves https://github.com/getsentry/sentry/issues/75623

- Adds an option to the old feedback modal to use the new user feedback API.
- Removes the beta tag from the source map debugger.


billyvg (approved):
We should probably get rid of this modal eventually

PR #75641, opened 2024-08-06
feat(alerts): Open in discover button goes to the right dataset

shruthilayaj:
Open in the right dataset based on alert rule dataset


narsaynorath (approved):
lgtm, tested it on dev-ui with various alerts

PR #75646, opened 2024-08-06
chore(anr): Remove `anr-improvements` flag

leeandher:
The flag is already GA and this would simplify the refactors that are going to be done for the stack trace sections with threads.

Next PR will be to remove the flag from flagpole, then lastly remove the registration from manager.py.


scttcper (approved):

shruthilayaj (approved):
thank you for cleaning this up!

PR #75651, opened 2024-08-06
ref(grouping): Improve config transition metrics

lobsterkatie:
This makes a few small improvements to our metrics tracking the change from the legacy to the new grouping config transition logic:

- Use an option to control the sample rate, so it's uniform across metrics (allowing them to be compared to each other) and easy to change if we want to.

- Set that sample rate to 100% for now, as the number of projects transitioning is small and we want to get good signal.

- Add a new metric, `grouping.config_updated`, tagged with a source, either ingest or the getsentry script. (The script has been updated to pass the `source` parameter in https://github.com/getsentry/getsentry/pull/14839.)

### DIFF
--- a/src/sentry/grouping/ingest/config.py
+++ b/src/sentry/grouping/ingest/config.py
@@ -8,10 +8,11 @@ from typing import Any
 from django.conf import settings
 from django.core.cache import cache
 
-from sentry import features
+from sentry import features, options
 from sentry.locks import locks
 from sentry.models.project import Project
 from sentry.projectoptions.defaults import BETA_GROUPING_CONFIG, DEFAULT_GROUPING_CONFIG
+from sentry.utils import metrics
 
 logger = logging.getLogger("sentry.events.grouping")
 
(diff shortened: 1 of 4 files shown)


jangjodi (approved):

src/sentry/grouping/ingest/config.py
```diff
@@ -28,7 +29,7 @@
 
 # Used by getsentry script. Remove it once the script has been updated to call update_grouping_config_if_permitted
 def _auto_update_grouping(project: Project) -> None:
```
armenzg, line 31: I don't think we call this function anymore. We can remove it.

src/sentry/grouping/ingest/config.py
```diff
@@ -94,6 +95,11 @@ def update_grouping_config_if_permitted(project: Project) -> None:
             event=audit_log.get_event_id("PROJECT_EDIT"),
             data={**changes, **project.get_audit_log_data()},
         )
+        metrics.incr(
+            "grouping.config_updated",
+            sample_rate=options.get("grouping.config_transition.metrics_sample_rate"),
+            tags={"current_config": current_config, "source": source},
```
armenzg, line 101: Do we need to track `source`? What are we trying to measure?

armenzg (approved):

src/sentry/grouping/ingest/config.py
lobsterkatie: Not strictly necessary, but I figured it'd be since to be able to distinguish upgrades done by the script from upgrades which happen organically.

PR #75652, opened 2024-08-06
ref: type get_event_by_id and related functions

asottile-sentry:


asottile-sentry (commented):
almost all of this is annotations-only -- I've highlighted the few functional changes to aid in reviewing

armenzg (approved):
Thank you @asottile-sentry ! 🥔

PR #75653, opened 2024-08-06
chore(hybridcloud) Start moving outbox logic to sentry.hybridcloud

markstory:
Start moving outbox related logic into sentry.hybridcloud.outbox. I've split out the enums and signals into separate modules so that importing `OutboxCategory` won't pull in all the django models in the longer term.


src/sentry/hybridcloud/outbox/category.py
```diff
@@ -0,0 +1,359 @@
...
+    RELOCATION_SCOPE = scope_categories(
+        10, {OutboxCategory.RELOCATION_EXPORT_REQUEST, OutboxCategory.RELOCATION_EXPORT_REPLY}
+    )
+
+    def __str__(self) -> str:
+        return self.name
+
+    @classmethod
+    def scope_has_category(cls, shard_scope: int, category: int) -> bool:
+        return OutboxCategory(category) in _outbox_categories_for_scope[shard_scope]
```
markstory, line 320: This method is new and was added to avoid importing `_outbox_categories_for_scope` outside of this module.

GabeVillalobos (approved):
LGTM!

PR #75655, opened 2024-08-06
fix(issue-details): Fix overflow issue with jump to section

roggenkemper:
this pr fixes the overflow issue with the jump to section by limiting it's width and making the button bar scrollable if there's too many items.

### DIFF
--- a/static/app/views/issueDetails/streamline/eventNavigation.tsx
+++ b/static/app/views/issueDetails/streamline/eventNavigation.tsx
@@ -377,4 +377,13 @@ const EventTitle = styled('div')`
 
 const StyledButtonBar = styled(ButtonBar)`
   overflow-x: auto;
+  overflow-y: hidden;
+
+  &:after {
+    position: sticky;
+    padding: ${space(1)};
+    content: '';
+    inset: 0;
+    background: linear-gradient(90deg, transparent, ${p => p.theme.background});
+  }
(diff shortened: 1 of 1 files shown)


static/app/views/issueDetails/streamline/eventNavigation.tsx
```diff
@@ -332,6 +332,9 @@ const JumpTo = styled('div')`
   align-items: center;
   color: ${p => p.theme.subText};
   font-size: ${p => p.theme.fontSizeSmall};
+  white-space: nowrap;
+  max-width: 50%;
+  width: 400px;
```
roggenkemper, line 337: the width i chose is pretty arbitrary - happy to adjust as people see fit

malwilley:
I'm not sure scrolling is the solution we want - users that have scrollbars shown (anything that's not a mac) will see this:

CleanShot 2024-08-07 at 09 55 50

roggenkemper:
> I'm not sure scrolling is the solution we want - users that have scrollbars shown (anything that's not a mac) will see this:
> 
> CleanShot 2024-08-07 at 09 55 50

yeah - that's a good point, this is what the most recent designs/discussions landed on but I can make sure we follow up before we roll out to make sure this case is considered

scttcper (approved):

PR #75657, opened 2024-08-06
nit(api): Skip adding choice name to bullet point

schew2381:
For choicefields, DRF has a setting called [ENUM_GENERATE_CHOICE_DESCRIPTION](https://drf-spectacular.readthedocs.io/en/latest/settings.html) where in that field description, it adds a bullet point for each choice like
`* (choice value) - (choice name)`

This is nice and helpful, but often times in the choicefield code we either don't use the second name part of the tuple and don't pass a list of tuples to the choicefield, or have the name be identical to the value.

This PR fixes this by omitting the choice name from the bullet point in such cases, so that it looks like
`* (choice value)`

Closes [ALRT-194](https://getsentry.atlassian.net/browse/ALRT-194?atlOrigin=eyJpIjoiNzBjMTk5M2ZiMzc2NDA0NzhlZWZkYTYzMjA5MWZhNDAiLCJwIjoiaiJ9)

[ALRT-194]: https://getsentry.atlassian.net/browse/ALRT-194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/sentry/conf/server.py
+++ b/src/sentry/conf/server.py
@@ -1419,25 +1419,28 @@ if os.environ.get("OPENAPIGENERATE", False):
     from sentry.apidocs.build import OPENAPI_TAGS, get_old_json_components, get_old_json_paths
 
     SPECTACULAR_SETTINGS = {
-        "DEFAULT_GENERATOR_CLASS": "sentry.apidocs.hooks.CustomGenerator",
-        "PREPROCESSING_HOOKS": ["sentry.apidocs.hooks.custom_preprocessing_hook"],
-        "POSTPROCESSING_HOOKS": ["sentry.apidocs.hooks.custom_postprocessing_hook"],
-        "DISABLE_ERRORS_AND_WARNINGS": False,
-        "COMPONENT_SPLIT_REQUEST": False,
-        "COMPONENT_SPLIT_PATCH": False,
+        "APPEND_COMPONENTS": get_old_json_components(OLD_OPENAPI_JSON_PATH),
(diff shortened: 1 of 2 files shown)


src/sentry/conf/server.py
```diff
@@ -1419,25 +1419,28 @@ def custom_parameter_sort(parameter: dict) -> tuple[str, int]:
...
-        "PREPROCESSING_HOOKS": ["sentry.apidocs.hooks.custom_preprocessing_hook"],
-        "POSTPROCESSING_HOOKS": ["sentry.apidocs.hooks.custom_postprocessing_hook"],
+        "DESCRIPTION": "Sentry Public API",
         "DISABLE_ERRORS_AND_WARNINGS": False,
-        "COMPONENT_SPLIT_REQUEST": False,
-        "COMPONENT_SPLIT_PATCH": False,
-        "AUTHENTICATION_WHITELIST": ["sentry.api.authentication.UserAuthTokenAuthentication"],
+        # We override the default behavior to skip adding the choice name to the bullet point if
+        # it's identical to the choice value by monkey patching build_choice_description_list.
+        "ENUM_GENERATE_CHOICE_DESCRIPTION": True,
```
schew2381, line 1433: Only change here is explicitly adding this setting (which defaults to True) so that people know it's here and sorting the settings

iamrajjoshi (approved):

PR #75659, opened 2024-08-06
ref: fix types in slicing-related modules

asottile-sentry:


joshuarli (approved):
ooooh typeguard and final are nice, i'm liking this

asottile-sentry:
> ooooh typeguard and final are nice, i'm liking this

heh, `Final` is just to get automatic `: Literal['foo'] = 'foo'`

PR #75665, opened 2024-08-06
fix(events): Global views was enabled accidentally

wmak:
- This fixes a bug where global views was enabled for anyone with join leave accidentally


tests/sentry/api/endpoints/test_organization_traces.py
```diff
@@ -22,7 +22,7 @@ def setUp(self):
 
     def do_request(self, query, features=None, **kwargs):
         if features is None:
-            features = ["organizations:performance-trace-explorer"]
+            features = ["organizations:performance-trace-explorer", "organizations:global-views"]
```
Zylphrex, line 25: totally fair to enable this by default in these tests given what its testing 👍

Zylphrex (approved):

PR #69860, opened 2024-04-29
feat(sessions): Remove unused sessions ingestion feature flags

Dav1dde:
Relay can no longer ingest sessions into Kafka after https://github.com/getsentry/relay/pull/3271 making these feature flags obsolete.

See also: https://github.com/getsentry/relay/issues/3368 and https://github.com/getsentry/relay/issues/3492

PR #75669, opened 2024-08-06
chore(dashboards): Remove usage of `dashboards-rh-widget` flag

gggritso:
This PR removes the UI flag added in https://github.com/getsentry/sentry/pull/45828/. The "Release Health" widgets were gated behind a feature because the ingestion of Session metrics wasn't turned on for everyone. As of https://github.com/getsentry/sentry/pull/69860 the feature that controls Session metrics ingestion (`organizations:metrics-extraction`) doesn't even exist anymore, the logic is enabled for everyone. So, it's safe to show everyone release health widgets!

### DIFF
--- a/static/app/views/dashboards/addWidget.tsx
+++ b/static/app/views/dashboards/addWidget.tsx
@@ -148,14 +148,12 @@ export function AddWidgetButton({onAddWidget, ...buttonProps}: Props & ButtonPro
       onAction: () => handleAction(DataSet.ISSUES),
     });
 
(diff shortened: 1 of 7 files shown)


gggritso:
@jjbayer @shruthilayaj did I understand the situation correctly?

shruthilayaj (approved):
yup, (on the product side) we had to keep release health widgets gated till release health metrics were available on ST & SH (since directly querying the metrics backend instead of the sessions duplex for some things because reasons)
this clean up just kinda fell off the radar, so thanks George!

PR #75671, opened 2024-08-06
ref(control_silo): Move slack logic for tasks to new location in integrations module

Christinarlong:
Moves logic and updates import routes to the new slack tasks located in sentry/integrations/slack/tasks

ref(issue # here)
Split of (https://github.com/getsentry/sentry/pull/74925)

### DIFF
--- a/src/sentry/conf/server.py
+++ b/src/sentry/conf/server.py
@@ -773,6 +773,7 @@ CELERY_IMPORTS = (
     "sentry.tasks.files",
     "sentry.tasks.groupowner",
     "sentry.tasks.integrations",
+    "sentry.tasks.integrations.slack",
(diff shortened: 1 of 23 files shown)


src/sentry/conf/server.py
```diff
@@ -773,6 +773,7 @@ def SOCIAL_AUTH_DEFAULT_USERNAME() -> str:
     "sentry.tasks.files",
     "sentry.tasks.groupowner",
     "sentry.tasks.integrations",
+    "sentry.tasks.integrations.slack",
```
Christinarlong, line 776: only change

markstory (approved):
Looks good with the additional celery_imports

PR #75672, opened 2024-08-06
fix(discover): Update measurements.inp field type to duration

edwardgou-sentry:
Fix an issue where `measurements.inp` response field type was number instead of duration.


0Calories (approved):
👌

PR #75673, opened 2024-08-06
Update uptime checker user agent

gaprl:
Updates the uptime checker `User-Agent` to match what we have described in our [documentation PR.](https://github.com/getsentry/sentry-docs/pull/10810)


evanpurkhiser (approved):

wedamija (approved):
Note that this won't fix the agent for already created subscriptions. We'll need a backfill

PR #65116, opened 2024-02-13
chore(grouping): Add log for origin of grouping config used

lobsterkatie:
While I was working on https://github.com/getsentry/sentry/pull/65113, the question came up about why we get the primary grouping config from the event for new events and from the project for reprocessed events - why not get the config from the project where it lives in both cases, since this would simplify the logic? How does the grouping config even get onto the event before any of our grouping code runs? 

The answer to the latter question turns out to be that Relay [adds the config information](https://github.com/getsentry/relay/blob/1a780e2831c1e6e8b3cfda97e7345c3e3279c367/relay-event-normalization/src/normalize/mod.rs#L369-L372) (which it [gets from the project](https://github.com/getsentry/relay/blob/1a780e2831c1e6e8b3cfda97e7345c3e3279c367/relay-server/src/services/processor/event.rs#L383)) to the event before passing it to the processing pipeline. 

The answer to the former question is... not as clear. According to [the docstring in Relay](https://github.com/getsentry/relay/blob/1a780e2831c1e6e8b3cfda97e7345c3e3279c367/relay-event-normalization/src/lib.rs#L87-L89), it's for idempotency. But the only way that would matter is if an event somehow ran through Relay once and the processing pipeline twice, and in between the two trips through the pipeline had its project's config updated. In that case it would guarantee the hash result wouldn't change. But a) legit, how would the second trip through the pipeline even happen for a non-reprocessed event?, b) even if it did, what are the chances the project's grouping config is going to be updated in the meantime?, and c) if the project's config _were_ to be updated in the meantime, would we even want the result from the previous config, or would we want the updated result? But then if idempotency isn't the reason... what _is_ the reason? _Is_ there even a reason? Could we simplify it to always look on the project?

Regardless of _why_ we have Relay add the data and then pull it from the event for new events, the difference between that and getting it from the project doesn't matter if both ways always result in the same value. This therefore adds a log which will record any time the configs are different.

PR #75674, opened 2024-08-06
ref(grouping): Always pull grouping config from project

lobsterkatie:
Currently, when deciding what grouping config to use for an event, we pull the primary config off of the event itself (where it's been added by Relay) and the secondary config from the event's project... unless the event is being reprocessed, in which case we grab both configs from the project.  A while back, [I added a log](https://github.com/getsentry/sentry/pull/65116) so we could see if anything would be different if we were to always just grab both configs from the project, and lo and behold, it's exactly the same. This means we can simplify the logic (and remove the log), which this PR does.

As a bonus, it also solves another problem: Sometimes we calculate the primary and secondary hashes using the same config, totally negating the point of the secondary. This happens when a config update causes the project options cached in Relay to become out of date - Relay still thinks config A is primary (and adds that information to the event), whereas the project now thinks of config A as the secondary and the newer config B as the primary. We then use A for calculating the primary hash (because we read it off of the event), and A for calculating the secondary hash (because we pulled it from the project). Whoops. This fixes that problem by always getting both the primary and secondary configs from the same single source of truth, which then allows us to remove all of the logic (and logging) handling the problem (which this PR also does).

Fixes (the issues discussed in) https://github.com/getsentry/sentry/pull/65116.

### DIFF
--- a/src/sentry/grouping/ingest/hashing.py
+++ b/src/sentry/grouping/ingest/hashing.py
@@ -19,7 +19,6 @@ from sentry.grouping.api import (
     apply_server_fingerprinting,
     detect_synthetic_exception,
     get_fingerprinting_config_for_project,
-    get_grouping_config_dict_for_event_data,
     get_grouping_config_dict_for_project,
     load_grouping_config,
 )
@@ -29,7 +28,6 @@ from sentry.grouping.ingest.utils import extract_hashes
 from sentry.grouping.result import CalculatedHashes
 from sentry.models.grouphash import GroupHash
 from sentry.models.project import Project
-from sentry.reprocessing2 import is_reprocessed_event
 from sentry.utils import metrics
 from sentry.utils.metrics import MutableTags
 from sentry.utils.tag_normalization import normalized_sdk_tag_from_event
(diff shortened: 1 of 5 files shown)


src/sentry/grouping/ingest/hashing.py
```diff
@@ -174,43 +172,8 @@ def run_primary_grouping(
...
-                    logger.info(
-                        "Event grouping config different from project grouping config",
-                        extra={
-                            "project": project.id,
-                            "relay_config": config_from_relay,
-                            "project_config": config_from_project,
-                        },
-                    )
+        grouping_config = get_grouping_config_dict_for_project(project)
+        job["data"]["grouping_config"] = grouping_config
```
armenzg, line 176: Is there value in keeping this comment?
```
            # Write back grouping config because it might have changed since the
            # event was ingested.
```

Your call.

armenzg (approved):

src/sentry/grouping/ingest/hashing.py
lobsterkatie: I debated that. In the context of "Normally we don't write config data to the event because Relay has already done it for us, but in this specific case we want to, and here's why," it definitely made sense to have it there. But in this new world it's just "When we do grouping, we record on the event the config being used," and so I felt like it was actually over-complicating things to call out the reprocessing case specifically.

armenzg: 👍🏻

PR #75678, opened 2024-08-06
feat(data-secrecy): Create Data Secrecy API

iamrajjoshi:
`sentry` version of https://github.com/getsentry/getsentry/pull/14679


src/sentry/data_secrecy/api/waive_data_secrecy.py
```diff
@@ -0,0 +1,146 @@
...
+            defaults={
+                "access_start": access_start,
+                "access_end": access_end,
+            },
+        )
+
+        self.create_audit_entry(
+            request=request,
+            organization=organization,
+            event=audit_log.get_event_id("DATA_SECRECY_WAIVED"),
```
cathteng, line 114: do we care that this audit log entry will be emitted every time somebody adjusts the access period?

iamrajjoshi: not sure? i think that waiving shouldn't happen that often and customers should know who is adjusting this.

cathteng (approved):

PR #75679, opened 2024-08-06
feat(onboarding): add logging

ameliahsu:
added logging to monitor the impact of new messaging integration onboarding button

also converted setup button from a class component to a functional component which can replace the existing "Set up Slack Now" button

### DIFF
--- a/static/app/views/alerts/rules/issue/index.tsx
+++ b/static/app/views/alerts/rules/issue/index.tsx
@@ -69,6 +69,7 @@ import normalizeUrl from 'sentry/utils/url/normalizeUrl';
 import withOrganization from 'sentry/utils/withOrganization';
 import withProjects from 'sentry/utils/withProjects';
 import {PreviewIssues} from 'sentry/views/alerts/rules/issue/previewIssues';
+import SetupMessagingIntegrationButton from 'sentry/views/alerts/rules/issue/setupMessagingIntegrationButton';
 import {
   CHANGE_ALERT_CONDITION_IDS,
   CHANGE_ALERT_PLACEHOLDERS_LABELS,
@@ -1228,12 +1229,20 @@ class IssueRuleEditor extends DeprecatedAsyncView<Props, State> {
               </SettingsContainer>
             </ContentIndent>
             <SetConditionsListItem>
-              <StepHeader>{t('Set conditions')}</StepHeader>
-              <SetupAlertIntegrationButton
-                projectSlug={project.slug}
-                organization={organization}
-                refetchConfigs={this.refetchConfigs}
-              />
+              <StepHeader>{t('Set conditions')}</StepHeader>{' '}
+              {organization.features.includes('messaging-integration-onboarding') ? (
+                <SetupMessagingIntegrationButton
+                  projectSlug={project.slug}
+                  organization={organization}
+                  refetchConfigs={this.refetchConfigs}
+                />
+              ) : (
+                <SetupAlertIntegrationButton
+                  projectSlug={project.slug}
+                  organization={organization}
+                  refetchConfigs={this.refetchConfigs}
+                />
+              )}
             </SetConditionsListItem>
             <ContentIndent>
               <ConditionsPanel>
@@ -1435,18 +1444,33 @@ class IssueRuleEditor extends DeprecatedAsyncView<Props, State> {
                               </StyledAlert>
                             )
                           }
-                          additionalAction={{
-                            label: 'Notify integration\u{2026}',
-                            option: {
-                              label: 'Missing an integration? Click here to refresh',
-                              value: {
-                                enabled: true,
-                                id: 'refresh_configs',
-                                label: 'Refresh Integration List',
-                              },
-                            },
-                            onClick: this.refetchConfigs,
-                          }}
+                          {...(organization.features.includes(
+                            'messaging-integration-onboarding'
+                          )
+                            ? {
+                                additionalAction: {
+                                  label: 'Notify integration\u{2026}',
+                                  option: {
+                                    label:
+                                      'Missing an integration? Click here to refresh',
+                                    value: {
+                                      enabled: true,
+                                      id: 'refresh_configs',
+                                      label: 'Refresh Integration List',
+                                    },
+                                  },
+                                  onClick: () => {
+                                    trackAnalytics(
+                                      'onboarding.messaging_integration_steps_refreshed',
+                                      {
+                                        organization: this.props.organization,
+                                      }
+                                    );
+                                    this.refetchConfigs();
+                                  },
+                                },
+                              }
+                            : {})}
                         />
                         <TestButtonWrapper>
                           <Button
--- a/static/app/views/alerts/rules/issue/index.tsx
+++ b/static/app/views/alerts/rules/issue/index.tsx
@@ -1177,6 +1177,9 @@ class IssueRuleEditor extends DeprecatedAsyncView<Props, State> {
     const disabled = loading || !(canCreateAlert || isActiveSuperuser());
     const displayDuplicateError =
       detailedError?.name?.some(str => isExactDuplicateExp.test(str)) ?? false;
+    const hasMessagingIntegrationOnboarding = organization.features.includes(
+      'messaging-integration-onboarding'
+    );
 
     // Note `key` on `<Form>` below is so that on initial load, we show
     // the form with a loading mask on top of it, but force a re-render by using
@@ -1230,10 +1233,9 @@ class IssueRuleEditor extends DeprecatedAsyncView<Props, State> {
             </ContentIndent>
             <SetConditionsListItem>
               <StepHeader>{t('Set conditions')}</StepHeader>{' '}
-              {organization.features.includes('messaging-integration-onboarding') ? (
+              {hasMessagingIntegrationOnboarding ? (
                 <SetupMessagingIntegrationButton
                   projectSlug={project.slug}
-                  organization={organization}
                   refetchConfigs={this.refetchConfigs}
                 />
               ) : (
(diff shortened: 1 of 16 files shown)


static/app/views/alerts/rules/issue/setupMessagingIntegrationButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+    data: project,
+    isLoading,
+    isError,
+    refetch,
+  } = useApiQuery<{}>(
+    [`/projects/${organization.slug}/${projectSlug}/?expand=hasAlertIntegration`],
+    {staleTime: Infinity}
+  );
+
+  const detailedProject = project as Project & {
```
malwilley, line 44: You can define this type outside the component as:


```ts
interface ProjectWithAlertIntegrationInfo extends Project {
  hasAlertIntegrationInstalled: boolean;
}
```

Using `interface` and `extends` is better when adding properties to an existing type since it is more performant and it works more how you would expect it to (type intersection can result in some confusing types in some situations)

static/app/views/alerts/rules/issue/setupMessagingIntegrationButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+    refetchConfigs();
+  };
+
+  const {
+    data: project,
+    isLoading,
+    isError,
+    refetch,
+  } = useApiQuery<{}>(
+    [`/projects/${organization.slug}/${projectSlug}/?expand=hasAlertIntegration`],
```
malwilley, line 40: Not that big of a deal, but you can format it like this instead:

```ts
[`/projects/${organization.slug}/${projectSlug}/`, {query: { expand: 'hasAlertIntegration' }}],
```

static/app/views/alerts/rules/issue/setupMessagingIntegrationButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+              return <PluginIcon key={value} pluginId={value} size={16} />;
+            })}
+          </IconWrapper>
+        }
+        onClick={() =>
+          openModal(
+            deps => (
+              <MessagingIntegrationModal
+                {...deps}
+                headerContent={<h1>Connect with a messaging tool</h1>}
```
malwilley, line 84: Static strings should be wrapped in `t('...')` so that they can be translated

static/app/views/alerts/rules/issue/setupMessagingIntegrationButton.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  it('does not render button if alert integration installed when feature flag is on', function () {
+    MockApiClient.addMockResponse({
+      url: `/projects/${organization.slug}/${project.slug}/?expand=hasAlertIntegration`,
+      body: {
+        ...project,
+        hasAlertIntegrationInstalled: true,
+      },
+    });
+    const {container} = render(getComponent());
```
malwilley, line 45: Instead of checking the container, I'd check that no button exists:

```tsx
expect(screen.queryByRole('button')).not.toBeInTheDocument()
```

static/app/views/alerts/rules/issue/messagingIntegrationModal.tsx
```diff
@@ -32,6 +33,13 @@ function MessagingIntegrationModal({
   project,
   onAddIntegration,
 }: Props) {
+  useEffect(() => {
+    trackAnalytics('onboarding.messaging_integration_modal_rendered', {
```
malwilley, line 37: Usually we don't track analytics when something it rendered - it can result in a lot of excess analytic events if done wrong. Generally it's better to record the analytic on an action instead. I'd recommend firing this event when the button is clicked to open this modal.

If you haven't seen it yet, you can add them on the button very easily: https://develop.sentry.dev/development/analytics/#button-click-analytics

static/app/views/alerts/rules/issue/setupMessagingIntegrationButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+import MessagingIntegrationModal from 'sentry/views/alerts/rules/issue/messagingIntegrationModal';
+
+type Props = {
+  organization: Organization;
+  projectSlug: string;
+  refetchConfigs: () => void;
+};
+
+function SetupMessagingIntegrationButton({
+  organization,
```
malwilley, line 23: Instead of passing in `organization`, you can use `useOrganization()` inside this component. Makes it so that you don't need to pass that prop around everywhere.

In your test, you can add your test org to the context by doing:

```tsx
render(<Component />, { organization: mockOrganization })
```

static/app/views/alerts/rules/issue/index.tsx
```diff
@@ -1435,18 +1444,33 @@ class IssueRuleEditor extends DeprecatedAsyncView<Props, State> {
...
-                                enabled: true,
-                                id: 'refresh_configs',
-                                label: 'Refresh Integration List',
-                              },
-                            },
-                            onClick: this.refetchConfigs,
-                          }}
+                          {...(organization.features.includes(
+                            'messaging-integration-onboarding'
+                          )
```
saponifi3d, line 1449: i'd recommend pulling the `organization.features.includes('messaging-integration-onboarding')` into a variable like: `hasMessagingIntegration` or `isMessagingIntegrationEnabled` and use that variable instead of checking for the features each time.

static/app/views/alerts/rules/issue/index.tsx
```diff
@@ -1435,18 +1444,33 @@ class IssueRuleEditor extends DeprecatedAsyncView<Props, State> {
...
+                            ? {
```
saponifi3d, line 1450: rather than making this a ternary, could we just use `&&` logic since there isn't an else case?

static/app/views/alerts/rules/issue/messagingIntegrationModal.tsx
saponifi3d: +1 to the button clicks - just wanted to also thank you for that link! i've been hoping to find something like that! 🎉

ameliahsu: @malwilley @saponifi3d We need to track how many users are actually shown this setup button so that we can compare this number to how many people actually click the button and install an integration, I'm not sure if there's a better way than tracking when this button is rendered.

saponifi3d: 🤔 maybe trigger the `onboarding.messaging_integration_modal_rendered` onClick of the modal?

That way you could visualize the usage through:
- Page view
- Modal Open
- Setup Integration

and we'd be able to understand where a user might get stuck trying to setup the integration.

ameliahsu: Ah I think I misread this earlier-- I can definitely change when `onboarding.messaging_integration_modal_rendered` is logged to be onClick! The issue is with tracking `onboarding.setup_messaging_integration_button_rendered`, because I believe this can only be logged when the button is rendered.

static/app/views/alerts/rules/issue/setupMessagingIntegrationButton.tsx
```diff
@@ -0,0 +1,112 @@
...
+    [
+      `/projects/${organization.slug}/${projectSlug}/`,
+      {query: {expand: 'hasAlertIntegration'}},
+    ],
+    {staleTime: Infinity}
+  );
+
+  useEffect(() => {
+    if (project && !project.hasAlertIntegrationInstalled) {
+      trackAnalytics('onboarding.messaging_integration_button_rendered', {
```
sentaur-athena, line 49: @malwilley I saw your other comment about not having logs in modal render but this button we really want analytics for. The way I read this code is that because it's in useEffect and only depends on project and organization this useEffect will be only called once. What would make it be double logged?

malwilley: Since `project` is an object, if it were to be refetched the reference would change (even if the props didn't) which would trigger this useEffect again. Just pointing out that `useEffect` can easily result in more runs than you would expect.

Since there is no action the user takes to get to this state, another way you could log this data is by using `useRouteAnalyticsParams` to append a new property to the page view analytic. This is a reliable way of tracking things that are shown to the user: https://develop.sentry.dev/development/analytics/#route-based-frontend-analytics

sentaur-athena: Thank you. Will look into it ❤️

static/app/views/alerts/rules/issue/messagingIntegrationModal.tsx
```diff
@@ -13,11 +13,11 @@ import AddIntegrationRow from 'sentry/views/alerts/rules/issue/addIntegrationRow
 import {IntegrationContext} from 'sentry/views/settings/organizationIntegrations/integrationContext';
 
 type Props = ModalRenderProps & {
-  headerContent: React.ReactElement<any, any>;
+  headerContent: string;
```
saponifi3d, line 16: nit: `React.ReactNode` - which is: `string | null | JSX.Element` (basically, it's a catch all for how you're using this variable and the `bodyContent` variable. and is closer to the `React.ReactElement` type from before)

static/app/views/alerts/rules/issue/setupMessagingIntegrationButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  if (isLoading || isError) {
+    return null;
+  }
+
+  if (!shouldRenderSetupButton) {
+    return null;
+  }
+
+  // TODO(Mia): only render if organization has team plan and above
```
saponifi3d, line 61: just want to make sure it's okay to merge w/ this todo.

saponifi3d (approved):
LGTM 🎉

static/app/views/alerts/rules/issue/setupMessagingIntegrationButton.tsx
ameliahsu: Yup! That will be in my next PR, it's fine for now since this button is hidden behind a feature flag :)

saponifi3d:
^ thanks for the quick action and investigation on that failure. is unrelated. 🎉

PR #75682, opened 2024-08-06
feat(autofix): Allow user to reselect root cause and re-run

roaga:
Add a button to unselected root causes that re-triggers the root cause selection flow.

Screenshot 2024-08-06 at 1 29 06 PM

### DIFF
--- a/static/app/components/events/autofix/autofixRootCause.tsx
+++ b/static/app/components/events/autofix/autofixRootCause.tsx
@@ -410,7 +410,7 @@ function AutofixRootCauseDisplay({
                     size="xs"
                     onClick={() => handleSelectFix({causeId: cause.id})}
                     busy={isLoading}
-                    analyticsEventName="Autofix: Root Cause Fix Selected"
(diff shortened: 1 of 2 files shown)


static/app/components/events/autofix/autofixRootCause.tsx
```diff
@@ -398,11 +400,24 @@ function AutofixRootCauseDisplay({
...
+                      __html: singleLineRenderer(t('Cause: %s', cause.title)),
+                    }}
+                  />
+                  <Button
+                    size="xs"
+                    onClick={() => handleSelectFix({causeId: cause.id})}
+                    busy={isLoading}
+                    analyticsEventName="Autofix: Root Cause Fix Selected"
+                    analyticsEventKey="autofix.root_cause_fix_selected"
+                    analyticsParams={{group_id: groupId}}
```
malwilley, line 415: I would add some extra data here that would let us determine how many people are changing the root cause after the fact

malwilley (approved):

PR #75685, opened 2024-08-06
ref: add onRouteLeave shim

JonasBa:


static/app/utils/reactRouter6Compat.tsx
```diff
@@ -227,3 +230,46 @@ export function location6ToLocation3<Q extends Query = DefaultQuery>(
...
+        currentLocation.pathname !== nextLocation.pathname,
+    });
+  }
+}
+
+type ReactRouterV6RouteLeaveCallback = (state: {
+  currentLocation: Location;
+  nextLocation: Location;
+}) => boolean;
+type ReactRouterV3RouteLeaveCallback = () => string | undefined;
```
evanpurkhiser, line 249: Let's add a comment that explains how the old one worked vs the new one, just so we remember what the difference is in the future without having to re-understand it

PR #75687, opened 2024-08-06
ref: searchbar alert rule condition form

nhsiehgit:
per: https://github.com/getsentry/sentry/issues/75011
Aug-06-2024 14-00-48

### DIFF
--- a/static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
+++ b/static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
@@ -726,7 +726,7 @@ class RuleConditionsForm extends PureComponent<Props, State> {
                     />
                   ) : (
                     <SearchContainer>
-                      {!organization.features.includes('search-query-builder-alerts') ? (
+                      {organization.features.includes('search-query-builder-alerts') ? (
                         <SearchQueryBuilder
                           initialQuery={initialData?.query ?? ''}
                           getTagValues={this.getEventFieldValues}
--- a/static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
+++ b/static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
@@ -7,6 +7,7 @@ import omit from 'lodash/omit';
 import pick from 'lodash/pick';
 
 import {addErrorMessage} from 'sentry/actionCreators/indicator';
+import {fetchTagValues} from 'sentry/actionCreators/tags';
 import type {Client} from 'sentry/api';
 import {
   OnDemandMetricAlert,
@@ -19,15 +20,20 @@ import FormField from 'sentry/components/forms/formField';
 import IdBadge from 'sentry/components/idBadge';
 import ListItem from 'sentry/components/list/listItem';
 import {MetricSearchBar} from 'sentry/components/metrics/metricSearchBar';
+import {normalizeDateTimeParams} from 'sentry/components/organizations/pageFilters/parse';
 import Panel from 'sentry/components/panels/panel';
 import PanelBody from 'sentry/components/panels/panelBody';
+import {SearchQueryBuilder} from 'sentry/components/searchQueryBuilder';
 import {InvalidReason} from 'sentry/components/searchSyntax/parser';
 import {SearchInvalidTag} from 'sentry/components/smartSearchBar/searchInvalidTag';
 import {t, tct} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
 import type {Environment, Organization, Project, SelectValue} from 'sentry/types';
 import {ActivationConditionType, MonitorType} from 'sentry/types/alerts';
+import {defined} from 'sentry/utils';
+import {isAggregateField, isMeasurement} from 'sentry/utils/discover/fields';
 import {getDisplayName} from 'sentry/utils/environment';
+import {DEVICE_CLASS_TAG_VALUES, isDeviceClass} from 'sentry/utils/fields';
 import {hasCustomMetrics} from 'sentry/utils/metrics/features';
 import {getMRI} from 'sentry/utils/metrics/mri';
 import {getOnDemandKeys, isOnDemandQueryString} from 'sentry/utils/onDemandMetrics';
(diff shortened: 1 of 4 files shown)


static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
```diff
@@ -141,6 +218,36 @@ class RuleConditionsForm extends PureComponent<Props, State> {
     }
   }
 
+  getEventFieldValues = async (tag, query): Promise<string[]> => {
```
nhsiehgit, line 221: This is essentially a rewrite of the `searchBar.tsx` method `getEventFieldValues`

static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
```diff
@@ -118,6 +146,55 @@ class RuleConditionsForm extends PureComponent<Props, State> {
     this.fetchData();
   }
 
+  getFilterKeys = () => {
```
nhsiehgit, line 149: This is essentially a rewrite of the `searchBar.tsx` method `getTagList`

static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
```diff
@@ -118,6 +147,55 @@ class RuleConditionsForm extends PureComponent<Props, State> {
...
+          kind: FieldKind.MEASUREMENT,
+        };
+        return measurement_tags;
+      },
+      {}
+    );
+    const orgHasPerformanceView = organization.features.includes('performance-view');
+    const combinedTags: TagCollection =
+      dataset === Dataset.ERRORS
+        ? Object.assign({}, [], STATIC_FIELD_TAGS_WITHOUT_TRANSACTION_FIELDS)
```
scttcper, line 166: i think i'd use spread instead, also i'm not sure the array does anything here
```suggestion
        ? Object.assign({}, STATIC_FIELD_TAGS_WITHOUT_TRANSACTION_FIELDS)
```

static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
```diff
@@ -99,15 +121,22 @@ type Props = {
...
     environments: null,
+    measurements: {},
+    filterKeys: {},
   };
 
   componentDidMount() {
     this.fetchData();
+    const measurements = getMeasurements();
+    const filterKeys = this.getFilterKeys();
+    this.setState({measurements, filterKeys});
```
malwilley, line 139: Do these need to be in state? Are you doing this to memoize them?

static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
```diff
@@ -618,76 +726,103 @@ class RuleConditionsForm extends PureComponent<Props, State> {
...
-                          if (e.key === 'Enter') {
-                            e.preventDefault();
-                            e.stopPropagation();
+                      {organization.features.includes('search-query-builder-alerts') ? (
+                        <SearchQueryBuilder
+                          initialQuery={initialData?.query ?? ''}
+                          getTagValues={this.getEventFieldValues}
+                          placeholder={this.searchPlaceholder}
+                          searchSource="alert_builder"
+                          filterKeys={filterKeys}
```
malwilley, line 735: This file has a `getSupportedAndOmittedTags` function that it was using before. Does your new function do the same thing or do we will need that?

static/app/views/alerts/rules/metric/ruleForm.tsx
```diff
@@ -1212,36 +1212,36 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
         >
           <List symbol="colored-numeric">
             <RuleConditionsForm
-              project={project}
+              activationCondition={activationCondition}
```
nhsiehgit, line 1215: this is just alphabetizing the props

static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
```diff
@@ -618,76 +726,103 @@ class RuleConditionsForm extends PureComponent<Props, State> {
...
+                          {...getSupportedAndOmittedTags(dataset, organization)}
+                          includeSessionTagsValues={dataset === Dataset.SESSIONS}
+                          disabled={disabled || isErrorMigration}
+                          useFormWrapper={false}
+                          organization={organization}
+                          placeholder={this.searchPlaceholder}
+                          onChange={onChange}
+                          query={initialData.query}
+                          // We only need strict validation for Transaction queries, everything else is fine
+                          highlightUnsupportedTags={
```
malwilley, line 797: Looks like you probably want to pass this same thing to `disallowUnsupportedFilters` in SearchQueryBuilder?

malwilley:
Something might be wrong with the tag fetching function - the old experience returned some values for `release.build` for example:

CleanShot 2024-08-06 at 16 41 40

static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
nhsiehgit: Ya pretty much

nhsiehgit:
🤔 
Ah hmm i see, so that's what this function does 👀

static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
nhsiehgit: Ah got it 👍 
yes, i'll add this to the filterKeys

nhsiehgit:
@malwilley I still see it making the appropriate request to fetch the tags?
Is there any better way to validate this?

malwilley:
We discussed on slack already, but just verifying that the tag value function _does_ work, and actually works much better now that it's sending the project ID 🙌

static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
malwilley: Okay so I do like that `filterKeys` is memoized, but it looks like this is preventing custom tags from being shown. `this.getFilterKeys` depends on `tags` which are passed in as a prop and those can change depending on what project is displayed. So I think you'll need to make sure that this reacts to changes in `tags`

malwilley (commented):
This looks good aside from custom tags not being shown (see screenshots):

CleanShot 2024-08-07 at 10 22 31

CleanShot 2024-08-07 at 10 22 18

The only other thing that could be improved is using `filterKeySections` to organize the keys, but that could be done in a subsequent PR

static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
```diff
@@ -118,6 +147,61 @@ class RuleConditionsForm extends PureComponent<Props, State> {
...
+    );
+    const orgHasPerformanceView = organization.features.includes('performance-view');
+    const combinedTags: TagCollection =
+      dataset === Dataset.ERRORS
+        ? Object.assign({}, STATIC_FIELD_TAGS_WITHOUT_TRANSACTION_FIELDS)
+        : dataset === Dataset.TRANSACTIONS
+          ? Object.assign(
+              {},
+              measurementsWithKind,
+              [],
```
scttcper, line 171: i don't think array here does anything

static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
```diff
@@ -118,6 +147,61 @@ class RuleConditionsForm extends PureComponent<Props, State> {
...
+              )
+            : Object.assign({}, STATIC_FIELD_TAGS_WITHOUT_TRACING);
+
+    const tagsWithKind = Object.keys(tags).reduce((acc, key) => {
+      acc[key] = {
+        ...tags[key],
+        kind: FieldKind.TAG,
+      };
+      return acc;
+    }, {});
```
scttcper, line 191: types here could be a bit better i think right now the result is `{}`

```suggestion
    const tagsWithKind = Object.keys(tags).reduce<Record<string, something>>((acc, key) => {
      acc[key] = {
        ...tags[key],
        kind: FieldKind.TAG,
      };
      return acc;
    }, {});
```

static/app/views/alerts/rules/metric/ruleConditionsForm.tsx
```diff
@@ -118,6 +147,61 @@ class RuleConditionsForm extends PureComponent<Props, State> {
...
+
+    const {omitTags} = {
+      ...getSupportedAndOmittedTags(dataset, organization),
+    };
```
scttcper, line 195: i think spread can be removed
```suggestion
    const {omitTags} = getSupportedAndOmittedTags(dataset, organization);
```

scttcper (approved):

PR #75688, opened 2024-08-06
feat(trace): Add children count to the trace meta

wmak:
- This updates the trace meta endpoint to also get the counts of spans for each transaction id so we can know ahead of time whether they have any children


Abdkhan14 (approved):
🚢

tests/snuba/api/endpoints/test_organization_events_trace.py
```diff
@@ -1713,3 +1713,6 @@ def test_with_default(self):
         assert data["transactions"] == 8
         assert data["errors"] == 1
         assert data["performance_issues"] == 2
+        assert len(data["transaction_child_map"]) == 8
+        for item in data["transaction_child_map"]:
```
Abdkhan14, line 1717: Nit: transaction_children_count_map? Smth with count in it?

PR #75690, opened 2024-08-06
fix(issue-details): Fix buggy event navigation

roggenkemper:
this pr fixes the buggy event navigation that would sometimes happen with the updated event navigation. the new "custom event" tab will help alleviate this problem by redirecting generic events ( not recommended, first, last) so that it won't get in the way. it will also not respect the default setting for what event to show based on the user option


static/app/views/issueDetails/streamline/eventNavigation.tsx
```diff
@@ -153,16 +175,21 @@ export const EventNavigation = forwardRef<HTMLDivElement, EventNavigationProps>(
...
+                      };
                 return (
                   <TabList.Item
-                    to={{
-                      pathname: normalizeUrl(baseEventsPath + label + '/'),
-                      query: {...location.query, referrer: `${label}-event`},
-                    }}
+                    to={eventPath}
                     key={label}
+                    hidden={label === 'custom' && selectedOption !== 'custom'}
```
leeandher, line 192: nit: Just so they don't drift down the line, you can replace all these one off strings of `'custom'` with `EventNavOptions.CUSTOM.`

leeandher (approved):
Looks good!

static/app/views/issueDetails/streamline/eventNavigation.tsx
roggenkemper: good idea!

PR #75691, opened 2024-08-06
fix(anomaly detection): Surface errors to the user on API alert create and update

mifu67:
If an error occurs on serializer.save(), catch the error and surface it to the user instead of giving an InternalError.

Fixes https://getsentry.atlassian.net/browse/ALRT-193


src/sentry/incidents/endpoints/organization_alert_rule_index.py
```diff
@@ -125,7 +127,23 @@
...
+                )
+            except ValidationError:
+                raise
+            except forms.ValidationError as e:
+                # if we fail in create_metric_alert, then only one message is ever returned
+                raise ValidationError(e.error_list[0].message)
+            # catch-all for other errors
+            except Exception as e:
+                return Response(
+                    data={"detail": str(e)}, status=status.HTTP_400_BAD_REQUEST, exception=True
```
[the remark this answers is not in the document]
mifu67: Do we want to write a generic "something went wrong!" here?

tests/sentry/incidents/endpoints/test_organization_alert_rule_details.py
```diff
@@ -60,6 +64,39 @@ class AlertRuleDetailsBase(AlertRuleBase):
 
     endpoint = "sentry-api-0-organization-alert-rule-details"
 
+    @cached_property
+    def dynamic_alert_rule_dict(self):
```
ceorourke, line 68: Is this the same as in `test_organization_alert_rule_index`? If so could you put it in a shared place so it's not duplicated?

ceorourke (commented):
lgtm, the comment about the test fixture isn't blocking, just a nice to have

tests/sentry/incidents/endpoints/test_organization_alert_rule_details.py
mifu67: We actually don't need to duplicate it since `AlertRuleDetailsBase` inherits `AlertRuleBase` 😅

ceorourke (approved):

PR #75694, opened 2024-08-06
feat(data-secrecy): Create Data Secrecy RPC Service

iamrajjoshi:
`sentry` version for https://github.com/getsentry/getsentry/pull/14776

### DIFF
--- a/src/sentry/data_secrecy/service/impl.py
+++ b/src/sentry/data_secrecy/service/impl.py
@@ -6,11 +6,11 @@ from sentry.data_secrecy.service.service import DataSecrecyService
 
 class DatabaseBackedDataSecrecyService(DataSecrecyService):
     def get_data_secrecy_waiver(self, *, organization_id: int) -> RpcDataSecrecyWaiver | None:
-        data_secrecy_waiver = DataSecrecyWaiver.objects.filter(
-            organization_id=organization_id
-        ).first()
-
-        if not data_secrecy_waiver:
+        try:
+            data_secrecy_waiver = DataSecrecyWaiver.objects.filter(
+                organization_id=organization_id
+            ).get()
+        except DataSecrecyWaiver.DoesNotExist:
             return None
 
         return serialize_data_secrecy_waiver(data_secrecy_waiver=data_secrecy_waiver)
--- a/src/sentry/data_secrecy/service/impl.py
+++ b/src/sentry/data_secrecy/service/impl.py
@@ -1,0 +1,16 @@
+from sentry.data_secrecy.models.datasecrecywaiver import DataSecrecyWaiver
+from sentry.data_secrecy.service.model import RpcDataSecrecyWaiver
+from sentry.data_secrecy.service.serial import serialize_data_secrecy_waiver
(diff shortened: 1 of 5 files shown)


src/sentry/data_secrecy/service/impl.py
```diff
@@ -0,0 +1,16 @@
+from sentry.data_secrecy.models.datasecrecywaiver import DataSecrecyWaiver
+from sentry.data_secrecy.service.model import RpcDataSecrecyWaiver
+from sentry.data_secrecy.service.serial import serialize_data_secrecy_waiver
+from sentry.data_secrecy.service.service import DataSecrecyService
+
+
+class DatabaseBackedDataSecrecyService(DataSecrecyService):
+    def get_data_secrecy_waiver(self, organization_id: int) -> RpcDataSecrecyWaiver | None:
```
markstory, line 8:
```suggestion
    def get_data_secrecy_waiver(self, *, organization_id: int) -> RpcDataSecrecyWaiver | None:
```

markstory (approved):

src/sentry/data_secrecy/service/impl.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class DatabaseBackedDataSecrecyService(DataSecrecyService):
+    def get_data_secrecy_waiver(self, *, organization_id: int) -> RpcDataSecrecyWaiver | None:
+        data_secrecy_waiver = DataSecrecyWaiver.objects.filter(
+            organization_id=organization_id
+        ).first()
+
+        if not data_secrecy_waiver:
+            return None
```
RyanSkonnord, line 14:
```suggestion
        try:
            data_secrecy_waiver = DataSecrecyWaiver.objects.filter(
                organization_id=organization_id
            ).get()
        except DataSecrecyWaiver.DoesNotExist:
            return None
```

I dislike using `.first()` this way as it implies that more than one `DataSecrecyWaiver` may exist and we're okay with choosing one arbitrarily. The try-except structure also communicates (before your eyes reach the null check) that we don't necessarily expect to find it.

src/sentry/data_secrecy/service/model.py
```diff
@@ -0,0 +1,13 @@
...
+from django.utils import timezone
+from pydantic import Field
+
+from sentry.hybridcloud.rpc import RpcModel
+
+
+class RpcDataSecrecyWaiver(RpcModel):
+    organization_id: int = -1
+    access_start: datetime = Field(default_factory=timezone.now)
+    access_end: datetime = Field(default_factory=timezone.now)
```
RyanSkonnord, line 12: Oof, not ideal that we're going to call `timezone.now()` twice during object initialization and that we might get unequal timestamps (potentially with the end before the start). It's very unlikely to matter; just bugs me.

If you're interested in a bit of hacking, you could look into fixing it by overriding `__init__` (and manually calling `timezone.now()` to provide a common value to both attrs). I don't recommend this unless it sounds intrinsically rewarding; just thought it was interesting.

RyanSkonnord (approved):
LGTM, just the one suggestion.

PR #75696, opened 2024-08-06
fix(alerts): Add validation for missing alertThreshold

ceorourke:
If you forget to add `alertThreshold` to the payload when creating a metric alert through the API we don't currently throw a validation error, but we run into a key error. This PR checks that it's not None and raises an error if so.

Closes https://getsentry.atlassian.net/browse/ALRT-192 and SENTRY-3D4Q


saponifi3d (approved):
Yay! Throw(/raise) early and often!

PR #75697, opened 2024-08-06
fix(trace): compute start and end from new view

JonasBa:


static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
```diff
@@ -1661,9 +1661,10 @@ export class TraceTree {
           const new_end = view[1];
 
           // Update the space of the tree and the trace root node
+          const start = Math.min(new_start * node.multiplier, this.root.space[0]);
           this.root.space = [
-            Math.min(new_start * node.multiplier, this.root.space[0]),
-            Math.max(new_end * node.multiplier - prev_start, this.root.space[1]),
+            start,
+            Math.max(new_end * node.multiplier - start, this.root.space[1]),
```
JonasBa, line 1667: This was assuming that the start value would not change significantly, which it apparently can

Abdkhan14 (approved):

PR #75698, opened 2024-08-06
ref: fix types of sentry.testutils.asserts

asottile-sentry:

### DIFF
--- a/src/sentry/incidents/logic.py
+++ b/src/sentry/incidents/logic.py
@@ -658,7 +658,7 @@ def create_alert_rule(
                 raise TimeoutError
             except ValidationError:
                 alert_rule.delete()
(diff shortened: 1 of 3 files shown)


src/sentry/incidents/logic.py
```diff
@@ -658,7 +658,7 @@ def create_alert_rule(
                 raise TimeoutError
             except ValidationError:
                 alert_rule.delete()
-                raise ValidationError
+                raise
```
asottile-sentry, line 661: oops this snuck in from another branch -- the error it fixes here is "not enough parameters" (this would always raise `TypeError` instead of the expected reraise of `ValidationError`

wedamija (approved):

PR #75702, opened 2024-08-06
chore(issues): Backfill IGNORED groups with missing substatuses

### LINKED ISSUE #75684: Backfill IGNORED issues to set substatus

snigdhas:
We need to run a backfill to fix the missing substatuses for IGNORED groups. This [redash query](https://redash.getsentry.net/queries/6898) shows we have ~120K ignored groups that have no substatus. Further digging shows that we _should_ have the info needed to backfill via Activity and GroupSnooze tables. 

Marking this as a post-deploy migration for safety.

Fixes https://github.com/getsentry/sentry/issues/75684


src/sentry/migrations/0748_fix_substatus_for_ignored_groups.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def backfill_substatus_for_ignored_groups(
+    apps: Apps, schema_editor: BaseDatabaseSchemaEditor
+) -> None:
+    Group = apps.get_model("sentry", "Group")
+    Activity = apps.get_model("sentry", "Activity")
+    GroupSnooze = apps.get_model("sentry", "GroupSnooze")
+
+    groups = Group.objects.filter(status=GroupStatus.IGNORED, substatus=None)
```
armenzg, line 53: Do we need to use any RangeQuerySetWrapper?
https://develop.sentry.dev/development/database-migrations/#filters

What happens if the migration fails? Does this get scheduled outside of a normal deployment? (I have not done migrations in such a long time)

Should we use the new GoCD script to run this update?

src/sentry/migrations/0748_fix_substatus_for_ignored_groups.py
```diff
@@ -0,0 +1,107 @@
...
+            if snooze.exists():
+                # If snooze exists, we should set the substatus to UNTIL_CONDITION_MET
+                new_substatus = GroupSubStatus.UNTIL_CONDITION_MET
+            else:
+                # If we have no other information stored about the group's status conditions, the group is ignored forever
+                new_substatus = GroupSubStatus.FOREVER
+
+        group.substatus = new_substatus
+
+    Group.objects.bulk_update(groups, ["substatus"])
```
wedamija, line 79: How many groups are we going to have here? We'll end up ooming if it's too many

src/sentry/migrations/0748_fix_substatus_for_ignored_groups.py
armenzg: No script as we should run this for self-hosted as well (Dan explained this to me)

snigdhas: I was hoping to avoid using `RangeQuerySetWrapper` since we wouldn't want to pair that with the `.filter` on substatus. 

This will be run manually with the GoCD migrations job, and is safe to rerun if it fails.

src/sentry/migrations/0748_fix_substatus_for_ignored_groups.py
wedamija: We already have an index on `(status, substatus, id)` so that'll actually be fine with the `RangeQuerySetWrapperWithProgressBarApprox`, so I think we should use that. If you want to make it simpler, you can just process a row at a time and skip batching - it'll take less than an hour for 123k rows.

src/sentry/migrations/0748_fix_substatus_for_ignored_groups.py
```diff
@@ -0,0 +1,107 @@
...
+
+def backfill_substatus_for_ignored_groups(
+    apps: Apps, schema_editor: BaseDatabaseSchemaEditor
+) -> None:
+    Group = apps.get_model("sentry", "Group")
+    Activity = apps.get_model("sentry", "Activity")
+    GroupSnooze = apps.get_model("sentry", "GroupSnooze")
+
+    groups = Group.objects.filter(status=GroupStatus.IGNORED, substatus=None)
+    activity = Activity.objects.filter(group__in=groups, type=ActivityType.SET_IGNORED)
```
wedamija, line 54: I'd probably recommend removing `group__in` here and just rely on `group_id` like you are later

src/sentry/migrations/0753_fix_substatus_for_ignored_groups.py
```diff
@@ -0,0 +1,107 @@
...
+            if group_activity.data.get("ignoreUntilEscalating", False):
+                new_substatus = GroupSubStatus.UNTIL_ESCALATING
+            # If any other field in the activity data is set, we should set the substatus to UNTIL_CONDITION_MET
+            elif any(group_activity.data.get(field) for field in ACTIVITY_DATA_FIELDS):
+                new_substatus = GroupSubStatus.UNTIL_CONDITION_MET
+
+        # If no activity is found or the activity data is not set, check the group snooze table
+        if not new_substatus:
+            snooze = GroupSnooze.objects.filter(group=group)
+            if snooze.exists():
```
wedamija, line 71: Nit: Could just be `if GroupSnooze....exists()`

wedamija (approved):

PR #75703, opened 2024-08-06
chore(substatus): Cleanup constants defined for issue substatus

snigdhas:
These fields were deprecated during development of the feature and never cleaned up. We have no saved searches using the deprecated fields so this is safe to remove.


src/sentry/models/groupsnooze.py
```diff
@@ -39,8 +39,6 @@ class GroupSnooze(Model):
     - If ``user_count`` is set, the snooze is lfited when unique users match.
     - If ``user_window`` is set (in addition to count), the snooze is lifted
       when the rate unique users matches.
-    - If ``until_escalating`` is set, the snooze is lifted when the Group's occurrences
```
snigdhas, line 42: This comment exists but the field does not 😅

armenzg (approved):

PR #75708, opened 2024-08-07
perf(replay/feedback): use pageFilter time filters to narrow tag value queries in search bar

aliu39:
Noticed this is something issues does. Personally I've experienced this query timing out a lot, causing spinners for the typeahead/nothing returned. We already pass in page filters, maybe this will improve performance.

Can be done for feedback too


billyvg (approved):
There's a Sentry issue somewhere that this can be linked to

aliu39:
> There's a Sentry issue somewhere that this can be linked to

https://sentry.sentry.io/issues/4345176691/?project=11276 is one I saw a while ago

PR #75709, opened 2024-08-07
ref(feedback): migrate to SearchQueryBuilder

### LINKED ISSUE #75008: SearchQueryBuilder: Replace SmartSearchBar in feedbacks page
## Background

`<SearchQueryBuilder />` is meant to be the next generation of `<SmartSearchBar />` and should ideally be replaced in all app surfaces. At this time, it is only implemented on issues pages. Because different parts of the app use this search bar differently, we need to implement and test it thoroughly ensure that is works for all use cases.

## Requirements

- Under a feature flag for this product area (already created), render `SearchQueryBuilder` instead of `SmartSearchBar`
-  If your search bar has a large number of filter keys, implement the `filterKeySections` prop to display them in an organized fashion. How your filter keys get categorized is up to you! 
- If you come across any missing feature that are critical to your use case, please note them and bring this up to the issues team.

## Resources

- See [here for SearchQueryBuilder documentation](https://sentry.sentry.io/stories/?name=app/components/searchQueryBuilder/index.stories.tsx) which should help you understand how to use the component and how `SmartSearchBar` props map to the new ones.
- A list of feature flags has been generated for each product surface. [See here for the full list](https://github.com/getsentry/sentry-options-automator/pull/2018/files). These flags are already enabled for internal orgs.

aliu39:
Closes https://github.com/getsentry/sentry/issues/75008
See https://sentry.sentry.io/stories/?name=app/components/searchQueryBuilder/index.stories.tsx

Very similar to https://github.com/getsentry/sentry/pull/75552. Recent searches isn't supported yet.

Demo .mov: https://drive.google.com/file/d/15LK9LqL1pXv7O0Zxa7CWuZBlNXmsXoMc/view?usp=sharing

Follow-up: 
- [ ] `useFetchOrganizationTags`
- [ ] We should probably also clarify the tags we suggest aren't always valid for feedback/replays. From the code it looks like these are just any tags from the organization.
- [ ] Try sorting by `-totalValues` for most popular tags


static/app/components/feedback/feedbackSearch.tsx
```diff
@@ -64,6 +69,36 @@ function getSupportedTags(supportedTags: TagCollection) {
...
+  organization: Organization
+): FilterKeySection[] {
+  if (!organization.features.includes('search-query-builder-user-feedback')) {
+    return [];
+  }
+
+  const nonFieldTags: string[] = Object.values(supportedTags)
+    .map(tag => tag.key)
+    .filter(key => !FEEDBACK_FIELDS.includes(key as FeedbackFieldKey | FieldKey));
+  nonFieldTags.sort();
```
malwilley, line 86: Same comment as on the other PR, can sort these by `totalValues` to be a bit more useful

static/app/components/feedback/feedbackSearch.tsx
```diff
@@ -64,6 +69,36 @@ function getSupportedTags(supportedTags: TagCollection) {
...
+
+  return [
+    {
+      value: 'feedback_field',
+      label: t('Feedback fields'),
```
malwilley, line 91:
```suggestion
      label: t('Feedback Fields'),
```

static/app/components/feedback/feedbackSearch.tsx
```diff
@@ -73,11 +108,21 @@ export default function FeedbackSearch({className, style}: Props) {
   const projectIds = usePageFilters().selection.projects;
   const {pathname, query} = useLocation();
   const organization = useOrganization();
-  const tags = useTags();
+  const organizationTags = useTags();
```
malwilley, line 111: And same comment about this again, that this can be fetched from the issue platform dataset instead

malwilley: Actually, there is one additional problem with using `useTags()` on feedback, which is that the request to fetch them isn't being made if you refresh on the feedback page. It looks like tags never load 😕 . But that's a problem in the existing implementation already

malwilley (approved):
❤️

static/app/components/feedback/feedbackSearch.tsx
aliu39: Oh, is it because we don't call loadOrganizationTags like in replays?

malwilley: Yeah that must be the missing call

aliu39: So we could use `useFetchIssueTags` instead right? Which directly queries issue platform instead of the store

aliu39: Thanks for pointing this out! I'm thinking we do it in a follow-up, so this PR only adds the new component with no other changes

malwilley: You will want to use `useFetchOrganizationTags` directly since you only need the issue platform dataset. `useFetchIssueTags` also fetches tags from errors which you don't need here

aliu39: I see, thanks!!

PR #75715, opened 2024-08-07
fix(metrics) Include projectIds when fetching metrics tags

vgrozdanic:
Fixes the problem of group by not working mentiond in the thread: https://sentry.slack.com/archives/C05TB55ETN1/p1722892801710349

Partially reverts what was done in the this PR: https://github.com/getsentry/sentry/pull/75415
In the PR by accident projectIds were removed from the request.

### DIFF
--- a/static/app/utils/metrics/useMetricsTags.tsx
+++ b/static/app/utils/metrics/useMetricsTags.tsx
@@ -18,10 +18,19 @@ const ALLOWED_SPAN_DURATION_TAGS = [
   'span.op',
 ];
 
-export function getMetricsTagsQueryKey(organization: Organization, mri: MRI | undefined) {
-  const queryParams = {
-    metric: mri,
-  };
+export function getMetricsTagsQueryKey(
+  organization: Organization,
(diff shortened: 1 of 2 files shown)


shellmayr (approved):

static/app/utils/metrics/useMetricsTags.tsx
```diff
@@ -19,10 +19,19 @@ const ALLOWED_SPAN_DURATION_TAGS = [
...
+  pageFilters: Partial<PageFilters>
+) {
+  const queryParams = pageFilters.projects?.length
+    ? {
+        metric: mri,
+        project: pageFilters.projects,
+      }
+    : {
+        metric: mri,
+      };
```
ArthurKnaus, line 34: Sending the empty array or undefined should have the same effect, right?
Asking as for other API calls we simply pass the pageFilters object without any checks.
```suggestion
  const queryParams = {
    metric: mri,
    project: pageFilters.projects,
  };
```

ArthurKnaus (approved):

PR #75722, opened 2024-08-07
feat(metrics): Add syntax highlighting for conditions

### LINKED ISSUE #75419: Consolidate Where clause styles
This uses a monospaced font, but the rest doesn't. Can we use the same font in WHERE like in the rest?
Image

Consider parsing & rendering the conditions in the same as the new searchbar:
Image

ArthurKnaus:
Screenshot 2024-08-07 at 14 06 27

Closes https://github.com/getsentry/sentry/issues/75419


shellmayr (approved):
LGTM - but can we do something about truncated conditions?

obostjancic (approved):

PR #75731, opened 2024-08-07
fix(event-user): Set event user id to be user_ident if None

jangjodi:
EventUser id can be None based on how the model is [defined](https://github.com/getsentry/sentry/blob/bd9a15d70d0f7bd68e43cea0246e3150afb17050/src/sentry/utils/eventuser.py#L97-L98)
Set the serialized id to be user_ident if id is None

Backend fix for this https://github.com/getsentry/sentry/pull/75732


armenzg (approved):
Do you need any test changes for this?

PR #75735, opened 2024-08-07
chore(control_silo): Delete old integration task shims

Christinarlong:
After the new tasks have marinated, the old tasks are no longer needed so let's delete them.

ref(#73859 )


asottile-sentry (approved):

markstory (approved):
:clap:

PR #75737, opened 2024-08-07
chore(projects): Increase project slug limit from 50 -> 100

### LINKED ISSUE #73032: Project slug length limit
### Problem Statement

Hey 👋 
Is there any opportunity to change max size of project's slug field?
This request is based on UX, 50 chars is not quite enough.
By project convention, used in our company, some project's names can be longer than 50 chars, so we face char limit in "slug" field.

Thanks

### Solution Brainstorm

_No response_

### Product Area

Settings - Projects

schew2381:
Increase project slug limit from 50 -> 100. There are no FE changes needed b/c we don't validate length on frontend, but show the error received from the failed POST request.
Screenshot 2024-08-07 at 10 59 23 AM

Also switch to using a constant for default slug length.

Closes https://github.com/getsentry/sentry/issues/73032
Fixes [ALRT-195](https://getsentry.atlassian.net/browse/ALRT-195?atlOrigin=eyJpIjoiZDA1MzllNWQ5MDY5NDY5YWEwOGI1NTMyOWE1ZDc3NzciLCJwIjoiaiJ9)

[ALRT-195]: https://getsentry.atlassian.net/browse/ALRT-195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nhsiehgit (approved):

src/sentry/migrations/0748_increase_project_slug_max_length.py
```diff
@@ -0,0 +1,34 @@
...
+from sentry.new_migrations.migrations import CheckedMigration
+
+
+class Migration(CheckedMigration):
+    # This flag is used to mark that a migration shouldn't be automatically run in production.
+    # This should only be used for operations where it's safe to run the migration after your
+    # code has deployed. So this should not be used for most operations that alter the schema
+    # of a table.
+    # Here are some things that make sense to mark as post deployment:
+    # - Large data migrations. Typically we want these to be run manually so that they can be
```
saponifi3d, line 15: should we make this a post deployment migration? not quite sure how many projects there are or the scale of this migration.

wedamija: No, post deployment is for backfills and large indexes only, not thing that modify the schema.

Increasing the side of this column should be fast. If it's not, we have a statement timeout that will cause the migration to break within a few seconds.

I'd say that just make sure you're around to monitor this and revert it if it won't run - it should be ok though.

wedamija (approved):
lgtm, increasing the size of a varchar has been fast for a long time: https://www.postgresql.org/docs/9.2/release-9-2.html#AEN114949

src/sentry/migrations/0748_increase_project_slug_max_length.py
markstory: We might get hit with an index rebuild which could throw off query performance for slug lookups. The [postgres docs](https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-DESC-SET-DATA-TYPE) state

> When this form is used, the column's statistics are removed, so running [ANALYZE](https://www.postgresql.org/docs/current/sql-analyze.html) on the table afterwards is recommended.

We have had [an incident](https://www.notion.so/sentry/INC-654-Database-migration-on-ApiTokenReplica-resulted-in-down-Sentry-7188774d88a848ffb05d7e092e0234f5) where a column resize invalided stats and caused an outage. We could have a similar outcome here.

schew2381: How do you run ANALYZE, I can't figure it out from the [example in docs](https://www.notion.so/sentry/PostgreSQL-d2be3f78723d4b6fad12001f3522a5d2?pvs=4#701c3bab6cc74ee58309f773d361de1f)

wedamija: Good call @markstory, I'd forgotten about that. Maybe in that case, we should make this a post_deploy migration and trigger the analyze as part of it, or even have ops run the entire thing manually for us.

I doubt we can run the ANALYZE as part of the deploy, it'll very likely time out.

wedamija (changes requested):
Blocking for now until we figure out the best way to run

src/sentry/migrations/0748_increase_project_slug_max_length.py
schew2381: I switched it to post deploy, do I have to include the ANALYZE as part of the migration code?

wedamija (commented):
This is a little tricky to run safely, just because we've seen an incident when widening a column before.

Running the analyze will fix the problem, however it takes a while to run, and if we end up locking on sentry_project then we'll be hard down until the analyze runs.

Mark and I discussed this on slack, and we have a couple of options that you could try:

Firstly, we could ask ops to duplicate `sentry_project` and then we could run a test on it. The steps here would be
 - duplicate `sentry_project` to `sentry_project_test`
 - Run an `ANALYZE` on `sentry_project_test` to make sure the stats are in a good state
 - Take a copy of the stats for this column on the table https://redash.getsentry.net/queries/7220/source
 - Get query plans for queries that use `slug` related indexes. This would be something like `explain select * from sentry_project_test where organization_id = 1 and slug='sentry'`
 - Run the sql to widen the column
 - Re-run the stats query and compare to the original, see if they have significantly changed
 - Re-run the explain from above and compare it to before, see if it has significantly changed.

If things look good here, then we could ask ops to run these commands manually for us:
- Take a copy of the stats row from this query https://redash.getsentry.net/queries/7220/source. Convert it into an update statement, so that we can set the stats back onto the row if we see a problem
- ALTER TABLE "sentry_project" ALTER COLUMN "slug" TYPE varchar(100);
- Monitor sentry for any problems. If we see them, run the update statement from the first step. Note: We've never tried to perform stats surgery before, so this is a little risky. We should probably test it out on the test table from above first.
- If we're still seeing problems, run an analyze on the table

schew2381:
> This is a little tricky to run safely, just because we've seen an incident when widening a column before.
> 
> Running the analyze will fix the problem, however it takes a while to run, and if we end up locking on sentry_project then we'll be hard down until the analyze runs.
> 
> Mark and I discussed this on slack, and we have a couple of options that you could try:
> 
> Firstly, we could ask ops to duplicate `sentry_project` and then we could run a test on it. The steps here would be
> 
> * duplicate `sentry_project` to `sentry_project_test`
> * Run an `ANALYZE` on `sentry_project_test` to make sure the stats are in a good state
> * Take a copy of the stats for this column on the table https://redash.getsentry.net/queries/7220/source
> * Get query plans for queries that use `slug` related indexes. This would be something like `explain select * from sentry_project_test where organization_id = 1 and slug='sentry'`
> * Run the sql to widen the column
> * Re-run the stats query and compare to the original, see if they have significantly changed
> * Re-run the explain from above and compare it to before, see if it has significantly changed.
> 
> If things look good here, then we could ask ops to run these commands manually for us:
> 
> * Take a copy of the stats row from this query https://redash.getsentry.net/queries/7220/source. Convert it into an update statement, so that we can set the stats back onto the row if we see a problem
> * ALTER TABLE "sentry_project" ALTER COLUMN "slug" TYPE varchar(100);
> * Monitor sentry for any problems. If we see them, run the update statement from the first step. Note: We've never tried to perform stats surgery before, so this is a little risky. We should probably test it out on the test table from above first.
> * If we're still seeing problems, run an analyze on the table

ok let me ask ops tmr

schew2381:
Progress Tracking

- [ ] Make a copy of `sentry_project` table called `sentry_project_test` - [Ops request](https://getsentry.atlassian.net/servicedesk/customer/portal/18/INF-211?created=true)
- [ ] duplicate sentry_project to sentry_project_test
- [ ] Run an ANALYZE on sentry_project_test to make sure the stats are in a good state
- [ ] Take a copy of the stats for this column on the table https://redash.getsentry.net/queries/7220/source
- [ ] Get query plans for queries that use slug related indexes. This would be something like explain select * from sentry_project_test where organization_id = 1 and slug='sentry'
- [ ] Run the sql to widen the column
- [ ] Re-run the stats query and compare to the original, see if they have significantly changed
- [ ] Re-run the explain from above and compare it to before, see if it has significantly changed.

- [ ] ask ops to run these commands manually for us:
    - [ ] Take a copy of the stats row from this query https://redash.getsentry.net/queries/7220/source. Convert it into an update statement, so that we can set the stats back onto the row if we see a problem
    - [ ] ALTER TABLE "sentry_project" ALTER COLUMN "slug" TYPE varchar(100);
    - [ ] Monitor sentry for any problems. If we see them, run the update statement from the first step. Note: We've never tried to perform stats surgery before, so this is a little risky. We should probably test it out on the test table from above first.
    - [ ] If we're still seeing problems, run an analyze on the table

wedamija (approved):
Approving so that we can merge this on Monday

PR #75739, opened 2024-08-07
chore(hybridcloud) Move outbox tasks to sentry.hybridcloud

markstory:
Consolidate more outbox subsystem logic into sentry.hybridcloud.


src/sentry/conf/server.py
```diff
@@ -737,6 +737,8 @@ def SOCIAL_AUTH_DEFAULT_USERNAME() -> str:
     "sentry.data_export.tasks",
     "sentry.discover.tasks",
     "sentry.hybridcloud.tasks.deliver_webhooks",
+    "sentry.hybridcloud.tasks.backfill_outboxes",
+    "sentry.hybridcloud.tasks.deliver_from_outbox",
```
markstory, line 741: The tasks didn't have their names changed, so updating `CELERY_IMPORTS` should be all that is necessary for tasks to continue executing without interruption.

tests/sentry/hybridcloud/tasks/test_backfill_outboxes.py
```diff
@@ -78,7 +80,7 @@ def test_region_processing(task_runner):
 
 @django_db_all
 @control_silo_test
-def test_control_processing(task_runner):
+def test_control_processing(task_runner: Callable[..., Any]) -> None:
```
GabeVillalobos, line 83: Do we have any type exemptions on these files that we need to clear now that we're explicitly typing things?

GabeVillalobos (approved):

tests/sentry/hybridcloud/tasks/test_backfill_outboxes.py
markstory: In their new location they get opted into our standard mypy rules which led to additional types being added.

PR #75741, opened 2024-08-07
fix(insights): project selection tooltip interferes with project selector

DominikB2014:
The expected behaviour of the "change project" tooltip (see below) is to appear when the selected project does not have data for the given module, but another project does. It is then supposed to disappear after 5 seconds OR after a click anywhere on the screen.

image

There was a bug where clicking on any of the pagefilter controls did not close the tooltip. This is due to events not propagating to the document level click handler. To fix this two things were done.
1. We add `{capture: true}` which effectively ignores any stop propagation's from any handler.
2. We wrap the tooltip in a invisible div instead of the `ProjectPageFilter`. Without this, changing projects would require 2 clicks, one to close the tooltip, and one to open the panel. I spent a while trying to figure a less hacky way around this, but i figured its best to get a quick fix out the door to improve the experience. But if anyone has any ideas on how to fix this lmk!

### DIFF
--- a/static/app/views/insights/common/components/modulePageFilterBar.tsx
+++ b/static/app/views/insights/common/components/modulePageFilterBar.tsx
@@ -44,9 +44,9 @@ export function ModulePageFilterBar({moduleName, onProjectChange, extraFilters}:
   }, [hasDataWithAllProjects]);
 
   useEffect(() => {
-    document.body.addEventListener('click', handleClickAnywhereOnPage);
+    document.addEventListener('click', handleClickAnywhereOnPage, {capture: true});
     return () => {
-      document.body.removeEventListener('click', handleClickAnywhereOnPage);
+      document.removeEventListener('click', handleClickAnywhereOnPage);
     };
   }, []);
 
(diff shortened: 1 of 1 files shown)


static/app/views/insights/common/components/modulePageFilterBar.tsx
```diff
@@ -59,8 +59,11 @@ export function ModulePageFilterBar({moduleName, onProjectChange, extraFilters}:
           position="bottom-start"
           disabled={!showTooltip}
         >
-          <ProjectPageFilter onChange={onProjectChange} />
+          {/* TODO: Placing a DIV here is a hack, it allows the tooltip to close and the ProjectPageFilter to close at the same time,
+          otherwise two clicks are required because of some rerendering/event propogation issues into the children */}
+          <div style={{width: '100px', position: 'absolute', height: '100%'}} />
```
edwardgou-sentry, line 64: 🤔 I tested this out locally, and the fix still seems to work if I just removed the invisible div element, ie no children in the tooltip. Maybe check if it also works for you locally as well?

edwardgou-sentry (approved):

static/app/views/insights/common/components/modulePageFilterBar.tsx
DominikB2014: Ohh interesting, iirc the tooltip didn't point to the right element (the project dropdown selector) when i tried this, but i could be wrong.
Did you notice that?

DominikB2014: I double checked, and the tooltip is pointing in a slightly different spots.
Without div
image

With div
image

With that in mind, i'll just keep the div for now just to add the mocks a bit better.

edwardgou-sentry: Ah good point, thanks for checking!

PR #75742, opened 2024-08-07
chore(control_silo): Remove integrations shims and fix loose imports

Christinarlong:
After updating the paths in getsentry to use new locations, we can delete the import shims. Also fix-up some leftover paths that were missed in the first pass.

ref(# issue here)


src/sentry/api/endpoints/organization_code_mapping_codeowners.py
armenzg, line 1: Is it intentional that these APIs are being deleted?

markstory: The endpoints live on in `sentry.integrations.api.endpoints`. We've been doing some refactoring to co-locate related product features and use django-apps/application modules better, as our current code layout results in directories with 500+ endpoints which is hard to navigate and run tests for.

These endpoints had import references in getsentry that needed to be resolved first, and we're making a second pass to remove the previous aliases.

markstory (approved):

PR #75747, opened 2024-08-07
feat(discover): Create Alert sets up the correct alert type

narsaynorath:
Clicking the "Create Alert" button should:
- Open up the number of errors alert page if the discover query is "Errors"
- Open up the throughput alert page if the discover query is "Transactions"

The requests to events-stats and events-meta should send along `dataset=errors` if the user has the split dataset feature flag.

### DIFF
--- a/static/app/views/alerts/rules/metric/incompatibleAlertQuery.tsx
+++ b/static/app/views/alerts/rules/metric/incompatibleAlertQuery.tsx
@@ -96,7 +96,6 @@ export function checkMetricAlertCompatiablity(
 
 interface IncompatibleAlertQueryProps {
   eventView: EventView;
-  orgSlug: string;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


static/app/views/alerts/rules/metric/incompatibleAlertQuery.tsx
```diff
@@ -111,15 +110,6 @@ export function IncompatibleAlertQuery(props: IncompatibleAlertQueryProps) {
...
 
-  const eventTypeError = props.eventView.clone();
-  eventTypeError.query += ' event.type:error';
-  const eventTypeTransaction = props.eventView.clone();
-  eventTypeTransaction.query += ' event.type:transaction';
-  const eventTypeDefault = props.eventView.clone();
-  eventTypeDefault.query += ' event.type:default';
-  const eventTypeErrorDefault = props.eventView.clone();
-  eventTypeErrorDefault.query += ' event.type:error or event.type:default';
-
```
shruthilayaj, line 122: Seems like this just wasn't used anywhere?

static/app/views/alerts/rules/utils.tsx
```diff
@@ -101,6 +102,14 @@ export function getAlertRuleActionCategory(rule: MetricRule) {
...
+export function shouldUseErrorsDiscoverDataset(
+  query: string,
+  dataset: Dataset,
+  organization: Organization
+) {
+  if (!hasDatasetSelector(organization)) {
+    return dataset === Dataset.ERRORS && query?.includes('is:unresolved');
+  }
+
+  return dataset === Dataset.ERRORS;
```
shruthilayaj, line 114: I'm going to approve the PR since the rest of it seems fine to me. This is the only thing i'm kinda unsure of. It seems safe, but I don't know what the context around the extra check for `query?.includes('is:unresolved')` in the original code is

shruthilayaj (approved):

static/app/views/alerts/rules/metric/incompatibleAlertQuery.tsx
narsaynorath: Yeah, these values we're being used, there was a refactor that happened here and they likely got left behind

PR #75748, opened 2024-08-07
feat(insights): add issue list to db module

KevinL10:
Adds a "Related Issues" table to query-specific pages. Specifically, this table surfaces all `Slow DB Query` and `N + 1 Query` issues due to the given query.

image

### DIFF
--- a/static/app/views/insights/common/components/issues.tsx
+++ b/static/app/views/insights/common/components/issues.tsx
@@ -1,0 +1,278 @@
+import styled from '@emotion/styled';
+
+import ActorAvatar from 'sentry/components/avatar/actorAvatar';
+import Count from 'sentry/components/count';
+import EventOrGroupExtraDetails from 'sentry/components/eventOrGroupExtraDetails';
+import LoadingIndicator from 'sentry/components/loadingIndicator';
+import {normalizeDateTimeParams} from 'sentry/components/organizations/pageFilters/parse';
+import Panel from 'sentry/components/panels/panel';
+import PanelHeader from 'sentry/components/panels/panelHeader';
+import PanelItem from 'sentry/components/panels/panelItem';
+import {IconWrapper} from 'sentry/components/sidebarSection';
+import GroupChart from 'sentry/components/stream/groupChart';
+import {IconUser} from 'sentry/icons';
+import {t} from 'sentry/locale';
+import {space} from 'sentry/styles/space';
+import type {Group} from 'sentry/types';
+import {useApiQuery} from 'sentry/utils/queryClient';
+import useOrganization from 'sentry/utils/useOrganization';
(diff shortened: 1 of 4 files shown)


static/app/views/insights/database/views/databaseSpanSummaryPage.tsx
```diff
@@ -69,6 +79,20 @@ export function DatabaseSpanSummaryPage({params}: Props) {
 
   const sort = decodeSorts(sortField).filter(isAValidSort).at(0) ?? DEFAULT_SORT;
 
+  const {data: indexedSpansByGroupId, isLoading: areIndexedSpansByGroupIdLoading} =
```
KevinL10, line 82: this api request is needed to get the raw database query on the span page, so that we can then search for issues containing that database query

static/app/views/insights/database/views/databaseSpanSummaryPage.tsx
```diff
@@ -228,6 +252,20 @@ export function DatabaseSpanSummaryPage({params}: Props) {
...
 
+            <Feature features="insights-related-issues-table">
+              {!areIndexedSpansByGroupIdLoading && (
+                <ModuleLayout.Full>
+                  <InsightIssuesList
+                    issueTypes={[
+                      'performance_slow_db_query',
+                      'performance_n_plus_one_db_queries',
+                    ]}
+                    message={indexedSpansByGroupId[0]['span.description']}
```
mjq, line 263: We _should_ get an indexed span here, but we should probably protect against the case where we don't.
```suggestion
                    message={indexedSpansByGroupId[0]?.['span.description']}
```

static/app/views/insights/common/components/issues.tsx
```diff
@@ -0,0 +1,264 @@
...
+  const organization = useOrganization();
+  const {selection} = usePageFilters();
+
+  let query = `issue.type:[${issueTypes.join(',')}]`;
+  // note: backend supports a maximum number of characters for message (seems to vary).
+  // so, we query the first 200 characters of `message`, then filter for exact `message`
+  // matches in application code
+  if (message) {
+    query += ` message:"${message?.slice(0, 200).replaceAll('"', '\\"')}"`;
+  }
```
mjq, line 96: I'm not sure that we should proceed when `message` is falsey. I don't think it makes sense to make a query without filtering on the message, because it will just return a random performance issue. I'd suggest returning early when `message` is falsey instead.

mjq (changes requested):

static/app/views/insights/common/components/issues.tsx
KevinL10: Couldn't return early here because the useApiQuery hook has to run at the top-level, but added a return and disabled the api query when message is falsey.

mjq (approved):

PR #75751, opened 2024-08-07
ref(compact-select): Use key instead of value for hidden/disabled options

malwilley:
CompactSelect uses `hiddenOptions` and `disabledOptions` to determine which options are hidden or disabled. Currently, they track the option `value`, not the option `key`. For most use cases, the key is an escaped version of the value.

I want to change this to track the `key` for a couple reasons:
1. Because keys are guaranteed to be unique, it makes more sense to use that instead of the value
2. For the search bar, I want options that have the same `value`, but different `key`s, which is not possible with the current implementation of these utils.

### DIFF
--- a/static/app/components/compactSelect/utils.tsx
+++ b/static/app/components/compactSelect/utils.tsx
@@ -40,7 +40,10 @@ export function getItemsWithKeys<Value extends SelectKey>(
       };
     }
 
-    return {...item, key: getEscapedKey(item.value)};
+    const existingKey =
+      'key' in item && typeof item.key === 'string' ? item.key : undefined;
+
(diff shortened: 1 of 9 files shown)


static/app/components/compactSelect/utils.tsx
```diff
@@ -40,7 +40,10 @@ export function getItemsWithKeys<Value extends SelectKey>(
       };
     }
 
-    return {...item, key: getEscapedKey(item.value)};
+    const existingKey =
```
malwilley, line 43: Checking for an existing key here to prevent the key from being escaped more than once if this function were called multiple times. Not sure if that is actually done, but wanted to be careful

scttcper (approved):
key makes sense

PR #75757, opened 2024-08-07
fix(data-secrecy): Fix API Url Add FF

iamrajjoshi:
2 things in this pr:
1. fix the api path which i incorrectly copied over from `getsentry` 😶‍🌫️ 
2. add data secrecy ff

### DIFF
--- a/src/sentry/data_secrecy/api/waive_data_secrecy.py
+++ b/src/sentry/data_secrecy/api/waive_data_secrecy.py
@@ -79,7 +79,7 @@ class WaiveDataSecrecyEndpoint(OrganizationEndpoint):
             )
         except Http404:
             return Response(
-                {"detail": "No data secrecy waiver in place."}, status=status.HTTP_200_OK
+                {"detail": "No data secrecy waiver in place."}, status=status.HTTP_404_NOT_FOUND
             )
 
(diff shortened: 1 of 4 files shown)


src/sentry/data_secrecy/api/waive_data_secrecy.py
```diff
@@ -79,7 +79,7 @@ def get(self, request: Request, organization: Organization) -> Response:
             )
         except Http404:
             return Response(
-                {"detail": "No data secrecy waiver in place."}, status=status.HTTP_404_NOT_FOUND
+                {"detail": "No data secrecy waiver in place."}, status=status.HTTP_200_OK
```
cathteng, line 82: can you not allow 404 in useApiQuery? you are looking for something that doesn't exist here

iamrajjoshi: ill hendle this in the frontend

cathteng (approved):

iamrajjoshi:
@cathteng  making u review again cause i updated delete to actually delete. the method i used doesn't work if the user makes a waiver for the future.

cathteng (approved):

PR #75765, opened 2024-08-07
feat(alerts): Skip metrics_layer when creating a PerformanceMetricsEntitySubscription using insights related aggregates/queries

edwardgou-sentry:
For context, we want to enable alerting on metrics from the Insights modules. However, there is a blocker that this PR attempts to resolve:

Most metrics displayed in the Insights modules are based on the Discover `SpansMetricsQueryBuilder` **without** using the Metrics Layer, while Alerts uses `AlertMetricsQueryBuilder` **with** the Metrics Layer. 

To support Insights metrics in Alerts, a possible solution involves the following two changes:
1. Port over any fields and functions from the `SpansMetricsQueryBuilder` config to the `AlertMetricsQueryBuilder` config
2. Disable using the Metrics Layer on the `AlertMetricsQueryBuilder` when the alert is based on Insights metrics. (We cannot use the Metrics Layer because many of the Insights fields and functions cannot be reproduced with the Metrics Layer's capabilities)

This PR involves point 2 from the above. To detect whether an Alert query is Insights based, parse the `aggregate` and `query`. We do it this way because there are no other attributes on the Alert Rule that we can leverage to directly tell us if an Alert is based on an Insight metric (technically they're just generic metrics).


wmak (commented):
As discussed in slack, lets try to use the query parsing in the builder to make this determination instead of using regexes on the raw query and aggregate strings

src/sentry/search/events/builder/metrics.py
```diff
@@ -151,7 +151,7 @@ def load_config(self) -> DatasetConfig:
             return super().load_config()
 
         if self.dataset in [Dataset.Metrics, Dataset.PerformanceMetrics]:
-            if self.builder_config.use_metrics_layer:
+            if self.use_metrics_layer:
```
edwardgou-sentry, line 154: Seems like `load_config` was using the `use_metrics_layer` attribute directly from the builder config. Not sure if there is a specific reason for this, but I had to update it to use the `self.use_metrics_layer` property which has additional checks

edwardgou-sentry:
> As discussed in slack, lets try to use the query parsing in the builder to make this determination instead of using regexes on the raw query and aggregate strings

Updated! I moved the logic into `MetricsQueryBuilder` and added a `insights_metrics_override_metric_layer` builder config option to control this

edwardgou-sentry:
Fixed some issues with use_case_id not using SPAN when appropriate

src/sentry/search/events/builder/metrics.py
```diff
@@ -376,14 +377,51 @@ def validate_aggregate_arguments(self) -> None:
...
+                return True
+            argument = match.group("columns") if match else None
+            if argument in constants.SPAN_METRICS_MAP.keys() - constants.METRICS_MAP.keys():
+                return True
+        return False
+
+    # Some fields and functions cannot be translated to metrics layer queries.
+    # This property is used to determine if a query is using at least one of these fields or functions, and if so, we must not use the metrics layer.
+    @property
+    def is_unsupported_metrics_layer_query(self) -> bool:
```
wmak, line 405: Could you cache this property? I'm pretty sure `use_metrics_layer` is called all over the place and we don't want to recalculate this determination repeatedly

src/sentry/search/events/builder/metrics.py
```diff
@@ -376,14 +377,51 @@ def validate_aggregate_arguments(self) -> None:
         if not self.use_metrics_layer:
             super().validate_aggregate_arguments()
 
+    # This property is used to determine if a query is using at least one of the fields in the spans namespace.
+    @property
+    def is_spans_metrics_query(self) -> bool:
```
wmak, line 382: nit can you make these docstrings for both functions
```suggestion
    @property
    def is_spans_metrics_query(self) -> bool:
    """This property is used to determine if a query is using at least one of the fields in the spans namespace."""
```

wmak (approved):
Minor comments, but please fix them before merging (especially caching both functions)

src/sentry/search/events/builder/metrics.py
edwardgou-sentry: 👍 added!

PR #75776, opened 2024-08-07
ref(form): Update the logic for the submit button

priscilawebdev:
**Before:**

https://github.com/user-attachments/assets/38036467-108a-452e-838f-7d56fe69d681

**After:**

https://github.com/user-attachments/assets/9dfce3fe-d463-486c-803a-568878a5c6d1

requires https://github.com/getsentry/getsentry/pull/14903


static/app/components/forms/model.tsx
```diff
@@ -354,8 +365,18 @@ class FormModel {
...
     const isRequired = this.getDescriptor(id, 'required');
+
+    if (!isRequired) {
+      return true;
+    }
+
     const value = this.getValue(id);
-    return !isRequired || (value !== '' && defined(value));
+
+    if (Array.isArray(value)) {
```
priscilawebdev, line 375: we were not validating fields with value of type array

static/app/components/forms/model.tsx
```diff
@@ -259,6 +268,8 @@ class FormModel {
       this.initialData[id] = props.setValue(this.initialData[id], props);
       this.fields.set(id, this.initialData[id]);
     }
+
+    this.formState = this.isFormInvalid() ? FormState.INVALID : FormState.READY;
```
priscilawebdev, line 272: we have to check if the form's initial state is valid.

static/app/components/forms/form.tsx
```diff
@@ -235,21 +235,29 @@ function Form({
...
-                      submitDisabled ||
-                      (requireChanges ? !formModel.formChanged : false)
-                    }
-                    type="submit"
-                  >
-                    {submitLabel ?? t('Save Changes')}
-                  </Button>
-                )}
+                {() => {
+                  return (
```
evanpurkhiser, line 239: minor nit: any reason you addd the return? Makes indenting deeper

static/app/components/forms/model.tsx
```diff
@@ -135,11 +135,13 @@ class FormModel {
       fieldState: observable,
       formState: observable,
 
+      isInvalid: computed,
       isError: computed,
       isSaving: computed,
       formData: computed,
       formChanged: computed,
 
+      isFormInvalid: action,
```
evanpurkhiser, line 144: Actions should start with verbs, how about `markFormInvalid`?

static/app/components/forms/model.tsx
```diff
@@ -200,6 +202,13 @@ class FormModel {
     return this.fields;
   }
 
+  /**
+   * Is form invalid
+   */
+  get isInvalid() {
+    return this.formState === FormState.INVALID;
+  }
```
evanpurkhiser, line 210: What is the difference between this and the form having errors?

static/app/components/forms/form.tsx
```diff
@@ -235,21 +235,29 @@ function Form({
...
+                    <Button
+                      title={
+                        formModel.isInvalid || formModel.isError
+                          ? t('Required fields must be filled out')
```
evanpurkhiser, line 243: Is this always why a form would be invalid though?

evanpurkhiser: What if there's a regex for a text field that doesn't match or something?

static/app/components/forms/model.tsx
evanpurkhiser, line 272: I'm not sure I understand why this function is an action and not computed?

evanpurkhiser:
So this is basically adding a new concept of "validness" vs "error". Error is saying you tried to do something and something was wrong" where as valid is saying "you can't do this because there is going to be an error"

Does that mean a user will never see an error anymore because you will not be able to submit the form until it is valid?

Should invalid be strictly constrained to "all fields are not filled out" in which case maybe `INVALID` is the wrong state and it should be `INCOMPLETE`?

static/app/components/forms/form.tsx
priscilawebdev: I was probably debugging 🙂 let me remove it

static/app/components/forms/model.tsx
priscilawebdev: When the form contains errors, they are shown within the fields. However, we prefer not to display these errors immediately upon opening the form

static/app/components/forms/form.tsx
priscilawebdev: that makes sense. We can also allow custom labels later, but for now, how about 'Required fields must be filled out and contain valid inputs'?

static/app/components/forms/model.tsx
priscilawebdev: honestly, I'm a bit unclear on the logic. I thought we should mark the form as invalid if it’s not ready. We can definitely create a function to check the form’s validity. I’ve updated the PR - let me know if this is what you had in mind

priscilawebdev:
> Does that mean a user will never see an error anymore because you will not be able to submit the form until it is valid?
If the user interacts with the field, we also provide hints if the field's value is invalid. 

basically we don't want to show errors immediately when opening a modal. Instead, we prefer to display a disabled button with a message like 'Hey... you still have to fill out things before you can click here.' Does that make sense

priscilawebdev:
Uploading Screen Recording 2024-08-08 at 12.37.00.mov…

@evanpurkhiser it looks like this ^

static/app/components/forms/form.tsx
```diff
@@ -199,6 +199,34 @@ function Form({
...
+          return t('Required fields must be filled out');
+        }
+
+        return t('Required fields must be filled out and inputs must be valid');
+      }
+
+      return t('Fields must contain valid inputs');
+    }
+    return undefined;
+  }, []);
```
JonasBa, line 228: This looks like its trying to be something inbetween a react component and a function that returns the string value title (useCallback seems like a weird choice)

Have you considered taking this fn definition outside of the component or maybe creating a component for this title?

```ts
const SubmitButtonTitle = ({form}: {form: FormModel}) => {...}
title={() => <SubmitButtonTitle form={formModel} />

```

priscilawebdev: It feels weird to create a component just to return a string, so I’ve moved the function outside of `Form` 😉

static/app/components/forms/form.tsx
```diff
@@ -80,6 +80,34 @@ export interface FormProps
...
+      if (allRequiredFieldErrors) {
+        return t('Required fields must be filled out');
+      }
+
+      return t('Required fields must be filled out and inputs must be valid');
+    }
+
+    return t('Fields must contain valid inputs');
+  }
+  return undefined;
```
evanpurkhiser, line 108: nit (sorry should have commented earlier): you can use guard statements to reduce the nesting here

```suggestion
  if (!form.isError) {
    return undefined;
  }
  
  const errors = form.getErrors();
  const hasRequiredFieldError = [...errors].some(
    ([_field, message]) => message === fieldIsRequiredMessage
  );

  if (hasRequiredFieldError) {
    const allRequiredFieldErrors = [...errors].every(
      ([_field, message]) => message === fieldIsRequiredMessage
    );

    if (allRequiredFieldErrors) {
      return t('Required fields must be filled out');
    }

    return t('Required fields must be filled out and inputs must be valid');
  }

  return t('Fields must contain valid inputs');
```

evanpurkhiser (approved):

evanpurkhiser (approved):

static/app/views/issueList/savedIssueSearches.spec.tsx
```diff
@@ -43,6 +43,7 @@ describe('SavedIssueSearches', function () {
     name: 'Last 4 Hours',
     query: 'age:-4h',
     visibility: SavedSearchVisibility.ORGANIZATION,
+    sort: 'date',
```
evanpurkhiser, line 46: What's this doing here?

priscilawebdev: the form was invalid because this was missing in the initialData object, so to make the test pass, I had to add this value for the form to be complete. Does that make sense?

evanpurkhiser: ah got it

PR #75778, opened 2024-08-07
ref(scm): abstraction for commit context integrations

cathteng:
Convert `CommitContextMixin` to an abstract base class called `CommitContextIntegration`. Also ties `CommitContextIntegration` to a new `CommitContextClient` abstract base class that represents the expected functionality of a SCM integration client tied to the ABC.

### DIFF
--- a/src/sentry/integrations/source_code_management/commit_context.py
+++ b/src/sentry/integrations/source_code_management/commit_context.py
@@ -36,10 +36,6 @@ class FileBlameInfo(SourceLineInfo):
 
 
 class CommitContextIntegration(ABC):
-    # whether or not integration has the ability to search through Repositories
-    # dynamically given a search query
-    repo_search = False
-
     @abstractmethod
     def get_client(self) -> CommitContextClient:
         raise NotImplementedError
@@ -77,4 +73,5 @@ class CommitContextClient(ABC):
     def get_blame_for_files(
         self, files: Sequence[SourceLineInfo], extra: Mapping[str, Any]
     ) -> list[FileBlameInfo]:
+        """Get the blame for a list of files. This method should include custom metrics for the specific integration implementation."""
         raise NotImplementedError
--- a/src/sentry/integrations/source_code_management/commit_context.py
+++ b/src/sentry/integrations/source_code_management/commit_context.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
+from abc import ABC, abstractmethod
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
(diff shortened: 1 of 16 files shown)


src/sentry/integrations/source_code_management/commit_context.py
```diff
@@ -34,23 +35,15 @@ class FileBlameInfo(SourceLineInfo):
...
-class GetBlameForFile(Protocol):
-    def get_blame_for_files(
-        self, files: Sequence[SourceLineInfo], extra: Mapping[str, Any]
-    ) -> list[FileBlameInfo]:
-        ...
-
-
-class GetClient(Protocol):
-    def get_client(self) -> GetBlameForFile:
-        ...
```
RyanSkonnord, line 46: Nice. `ABC` is preferable to `Protocol` IMHO.

src/sentry/integrations/source_code_management/commit_context.py
```diff
@@ -34,23 +35,15 @@ class FileBlameInfo(SourceLineInfo):
...
-
-
-class CommitContextMixin(GetClient):
+class CommitContextIntegration(ABC):
```
RyanSkonnord, line 38: We might consider making this `class CommitContextIntegration(IntegrationInstallation, ABC)`, with subclasses extending only `CommitContextIntegration` instead of both.

That might lead to an awkward "diamond inheritance" structure if we do that, and then the concrete classes extend both `CommitContextIntegration` and `RepositoryIntegration` (the changed `RepositoryMixin`). But I think it would be fine under Python's MRO rules, since none of the overrides overlap (and, if they did, it would be a problem regardless).

One nice thing about extending `IntegrationInstallation` would be that we would inherit its abstract `get_client` method and wouldn't have to declare a redundant one here. Except that the one we declare here specifies `CommitContextClient` as the return type, where the parent method returns `Any`. This suggests to me that `IntegrationInstallation` actually wants to have the client class as a generic type param, i.e. something like

```python
C = TypeVar("C")

class IntegrationInstallation(Generic[C], ABC):
    @abstractmethod
    def get_client(self) -> C:
        raise NotImplementedError

class CommitContextIntegration(IntegrationInstallation[CommitContextClient], ABC):
    ...
```

If all that sounds too complicated, feel free to ignore this entire comment and do it the easier way until we tackle it in a later PR.

cathteng: chatted in person, i'll keep what i have for now since the resolving of types this way is also complicated 😞

RyanSkonnord (approved):

PR #75783, opened 2024-08-07
feat(onboarding): disable button for below team plan

ameliahsu:
disabled setup messaging integration button for users without a team plan or above

Screenshot 2024-08-07 at 2 09 48 PM

### DIFF
--- a/static/app/views/alerts/rules/issue/messagingIntegrationModal.tsx
+++ b/static/app/views/alerts/rules/issue/messagingIntegrationModal.tsx
@@ -6,15 +6,14 @@ import type {ModalRenderProps} from 'sentry/actionCreators/modal';
 import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
 import type {IntegrationProvider} from 'sentry/types/integrations';
-import type {Organization} from 'sentry/types/organization';
 import type {Project} from 'sentry/types/project';
 import {useApiQueries} from 'sentry/utils/queryClient';
+import useOrganization from 'sentry/utils/useOrganization';
 import AddIntegrationRow from 'sentry/views/alerts/rules/issue/addIntegrationRow';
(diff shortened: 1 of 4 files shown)


static/app/views/alerts/rules/issue/messagingIntegrationModal.tsx
```diff
@@ -28,10 +27,10 @@ function MessagingIntegrationModal({
   headerContent,
   bodyContent,
   providerKeys,
-  organization,
   project,
   onAddIntegration,
 }: Props) {
+  const organization = useOrganization();
```
sentaur-athena, line 33: 🥳

static/app/views/alerts/rules/issue/setupMessagingIntegrationButton.tsx
```diff
@@ -27,77 +29,99 @@ function SetupMessagingIntegrationButton({projectSlug, refetchConfigs}: Props) {
...
       `/projects/${organization.slug}/${projectSlug}/`,
       {query: {expand: 'hasAlertIntegration'}},
     ],
     {staleTime: Infinity}
   );
 
-  const shouldRenderSetupButton = project && !project.hasAlertIntegrationInstalled;
+  const integrationQuery = useApiQuery<{providers: IntegrationProvider[]}>(
+    [
+      `/organizations/${organization.slug}/config/integrations/?provider_key=${providerKeys[0]}`,
```
sentaur-athena, line 46: Can you add a comment here that checking only the first one is enough because they will all return the same response?
Just for code clarity, otherwise ideally in the future we should build things more like vertical by vertical vs integration by integration.

sentaur-athena (commented):
Looks good. Can you add a test for an org with developer plan? Don't have to go all the way to actually create the plan but just mocking integration feature to return disabled.

sentaur-athena (approved):

PR #75787, opened 2024-08-07
fix(alerts): Clean up querysubscription and snubaquery when alert is deleted

ceorourke:
We have a huge number of events on https://sentry.sentry.io/issues/2382970109 which happens when an alert rule has been deleted but the associated `QuerySubscription` and `SnubaQuery` are not deleted. This PR cleans up those rows when we find that the alert rule is missing. I left the metric so we can see it drop and even out (I expect it'll drop since we won't be repeatedly receiving subscription updates for a rule that doesn't exist anymore, and then even out at a much lower rate when we are just receiving an update right after a rule is deleted). Once it evens out we could move this logic to the deletions flow for the `AlertRule` model, but first we need to remove all the old stuff.

Closes https://sentry.sentry.io/issues/2382970109

### DIFF
--- a/src/sentry/incidents/subscription_processor.py
+++ b/src/sentry/incidents/subscription_processor.py
@@ -58,7 +58,8 @@ from sentry.snuba.entity_subscription import (
     get_entity_key_from_query_builder,
     get_entity_subscription_from_snuba_query,
 )
-from sentry.snuba.models import QuerySubscription
+from sentry.snuba.models import QuerySubscription, SnubaQuery
+from sentry.snuba.subscriptions import delete_snuba_subscription
 from sentry.utils import json, metrics, redis
 from sentry.utils.dates import to_datetime
 from sentry.utils.json import JSONDecodeError
@@ -483,13 +484,12 @@ class SubscriptionProcessor:
 
         if not hasattr(self, "alert_rule"):
             # QuerySubscriptions must _always_ have an associated AlertRule
-            # If the alert rule has been removed then just skip
+            # If the alert rule has been removed then clean up associated tables and return
             metrics.incr("incidents.alert_rules.no_alert_rule_for_subscription")
-            logger.error(
-                "Received an update for a subscription, but no associated alert rule exists"
-            )
-            # TODO: Delete QuerySubscription here
-            # TODO: Delete SnubaQuery here
+            delete_snuba_subscription(self.subscription)
+            query_subscription = QuerySubscription.objects.get(id=self.subscription.id)
+            SnubaQuery.objects.filter(id=query_subscription.snuba_query.id).delete()
+            query_subscription.delete()
             return
 
         if subscription_update["timestamp"] <= self.last_update:
--- a/src/sentry/incidents/subscription_processor.py
+++ b/src/sentry/incidents/subscription_processor.py
@@ -58,7 +58,7 @@ from sentry.snuba.entity_subscription import (
     get_entity_key_from_query_builder,
     get_entity_subscription_from_snuba_query,
 )
-from sentry.snuba.models import QuerySubscription, SnubaQuery
+from sentry.snuba.models import QuerySubscription
 from sentry.snuba.subscriptions import delete_snuba_subscription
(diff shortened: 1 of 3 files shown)


nhsiehgit (commented):
+1 approved

but gonna give dan teh seal of approval

ceorourke:
@wedamija lmk if I'm doing something silly here

src/sentry/incidents/subscription_processor.py
```diff
@@ -483,13 +484,12 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+            # If the alert rule has been removed then clean up associated tables and return
             metrics.incr("incidents.alert_rules.no_alert_rule_for_subscription")
-            logger.error(
-                "Received an update for a subscription, but no associated alert rule exists"
-            )
-            # TODO: Delete QuerySubscription here
-            # TODO: Delete SnubaQuery here
+            delete_snuba_subscription(self.subscription)
+            query_subscription = QuerySubscription.objects.get(id=self.subscription.id)
+            SnubaQuery.objects.filter(id=query_subscription.snuba_query.id).delete()
```
wedamija, line 491: I think it would be better to build this `SnubaQuery` deletion logic into `delete_snuba_subscription`, and that can be in a separate pr. The `SnubaQuery` hanging around doesn't really hurt, other than there being a lot of orphaned rows in the table.

If you do want to build it, you should check that there are no subscriptions left related to the SnubaQuery before deleting. I think the best spot for it would be here: https://github.com/getsentry/sentry/blob/1a06bc7b19e61e5d41156f845b8321d0b63e3178/src/sentry/snuba/tasks.py#L187 after we delete the subscription

src/sentry/incidents/subscription_processor.py
```diff
@@ -483,13 +484,12 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
...
+            query_subscription.delete()
```
wedamija, line 492: This is handled as part of `delete_snuba_subscription`

src/sentry/incidents/subscription_processor.py
ceorourke: Ok thank you! I moved the logic into that task and added a test  - the existing one handled the case where there is a related subscription left so we don't delete, then I added one for when there isn't and we do.

src/sentry/snuba/tasks.py
```diff
@@ -184,7 +184,16 @@ def delete_subscription_from_snuba(query_subscription_id, **kwargs):
         )
 
     if subscription.status == QuerySubscription.Status.DELETING.value:
+        snuba_query_id = subscription.snuba_query.id
```
wedamija, line 187: You could do `snuba_query = subscription.snuba_query` here and skip the fetch after, since this line is already making the query. Alternatively, `snuba_query_id = subscription.snuba_query_id`

src/sentry/snuba/tasks.py
```diff
@@ -184,7 +184,16 @@ def delete_subscription_from_snuba(query_subscription_id, **kwargs):
...
         subscription.delete()
+        # check that there are no subscriptions left related to the SnubaQuery before deleting
+        try:
+            snuba_query = SnubaQuery.objects.get(id=snuba_query_id)
+        except SnubaQuery.DoesNotExist:
+            return
+
+        if snuba_query and not QuerySubscription.objects.filter(snuba_query=snuba_query).count():
```
wedamija, line 195: Nit: `.exists` instead of `.count`

wedamija (approved):

wedamija:
> ## Suspect Issues
> 
> * ‼️ **ProtectedError: ("Cannot delete some instances of model 'SnubaQuery' because they are referenced through protecte...** `sentry.snuba.tasks.delete_subscription_from_snuba` [View Issue](https://sentry.sentry.io/issues/5692082124/?referrer=github-pr-bot)
> 

Haha, damnit, you need to run the query with `AlertRule.objects_with_snapshots.filter`

PR #75789, opened 2024-08-07
fix(slack): fix disappearing tags for issue button actions

cathteng:
Fix tags disappearing when resolving/archiving/unarchiving an issue through a Slack message. Pass the tags to the `View`'s `private_metadata`, so when somebody takes an action through the modal, we can read the data from the `view_submission` event.

Fix

https://github.com/user-attachments/assets/4302c02f-a2db-4c4f-8efc-c1c802bcb22e

Before

https://github.com/user-attachments/assets/413fb417-f6ed-45a4-89b0-fa793f25e05c


iamrajjoshi (approved):
lgtm, but this has me thinking how would we even abstract this away into the spec? i feel like these are such fine details that might work differently between integrations

cathteng:
@iamrajjoshi i don't think we'd abstract this far deep into the details, you'd have to go more broad

PR #75791, opened 2024-08-07
feat(data-secrecy): Data Secrecy Settings UI

iamrajjoshi:
Here I add the settings for Data Secrecy in the frontend. There are two:
1. Org settings to disable and enable superuser access
2. Allow users to temporarily waive data secrecy so sentry employees and help

## Updated Demo: https://github.com/getsentry/sentry/pull/75791#issuecomment-2309088811

<details>
<summary> Old Demo </summary>

https://github.com/user-attachments/assets/2e2cb5e1-d8b9-46a0-9b65-64228c99e03d

https://github.com/user-attachments/assets/a2541ccc-eae1-478a-bc99-216514e282fc

Rewatch Links for Demo:
https://sentry.rewatch.com/video/ispnhixepw165zl9-screen-recording-2024-08-07-at-1-28-31-pm
https://sentry.rewatch.com/video/2jucpfwxswxs0dv6-screen-recording-2024-08-07-at-2-45-14-pm

Update:
I moved the waive setting so that its under the original toggle
image
</details>


cathteng (commented):
thoughts about the UI:
can we group the data secrecy toggles in one section? or move the waiver section right underneath the main toggle?

iamrajjoshi:
> thoughts about the UI:
> 
> can we group the data secrecy toggles in one section? or move the waiver section right underneath the main toggle?

grouping is what i was going for, but the form component we have is very finicky where it only lets us add 1 endpoint for it to call to update values when we use onBlur to update different settings individually. i don't want to refactor the component cause it's used by majority of the UI and I don't have the context for it.

Let me look into moving it tho

iamrajjoshi:
> thoughts about the UI: can we group the data secrecy toggles in one section? or move the waiver section right underneath the main toggle?

i moved the section!

leedongwei:
Major pieces are in place. I can help to prettify this on Fri.

leedongwei:
Screenshot 2024-08-22 at 11 45 05 AM

Screenshot 2024-08-22 at 11 44 57 AM

Screenshot 2024-08-22 at 11 44 41 AM

- Note the function `allowTempAccessProps.onChange` is still buggy. We are trying to display the datetime in the picker in localtime, but the picker assumes it is in UTC all the time.

iamrajjoshi:
> Note the function allowTempAccessProps.onChange is still buggy. We are trying to display the datetime in the picker in localtime, but the picker assumes it is in UTC all the time.

fixed the issue

https://github.com/user-attachments/assets/4c29e9b7-daa9-43eb-bc2c-3b755d15b1b0

cathteng:
what does it look like when you have an access period in the past and you open the page? should we render an info message for that and/or when you don't have access to toggle it?

iamrajjoshi:
> what does it look like when you have an access period in the past and you open the page? should we render an info message for that and/or when you don't have access to toggle it?

how it looks when the access period was in past and u open the page
image


this is how it looks when u don't have access to toggle
image


it doesn't say anything, but since all the settings on the page can't be edited if u don't have access, it should have
image

sentaur-athena:
When the toggle is off and the time is in the past my suggestion is to clear the time field, showing basically there's no access going on at the moment.

static/app/views/settings/components/dataSecrecy/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  const [allowAccess, setAllowAccess] = useState(organization.allowSuperuserAccess);
+  const [allowDate, setAllowDate] = useState<WaiverData>();
+  const [allowDateFormData, setAllowDateFormData] = useState<string>('');
+
+  const {data, refetch} = useApiQuery<WaiverData>(
+    [`/organizations/${organization.slug}/data-secrecy/`],
+    {
+      staleTime: 3000,
+      retry: (failureCount, error) => failureCount < 3 && error.status !== 404,
```
sentaur-athena, line 37: Why do we retry? I just don't see this often in our frontend code so wondering how this endpoint is different from others that needs retrying.

static/app/views/settings/components/dataSecrecy/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+import PanelAlert from 'sentry/components/panels/panelAlert';
+import PanelBody from 'sentry/components/panels/panelBody';
+import PanelHeader from 'sentry/components/panels/panelHeader';
+import {t, tct} from 'sentry/locale';
+import {useApiQuery} from 'sentry/utils/queryClient';
+import useApi from 'sentry/utils/useApi';
+import useOrganization from 'sentry/utils/useOrganization';
+
+type WaiverData = {
+  access_end: string | null;
```
sentaur-athena, line 21: nit: camelCase

static/app/views/settings/components/dataSecrecy/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+      'Sentry employees will not have access to your data unless granted permission'
+    ),
+    value: allowAccess,
+    disabled: organization.access.includes('org:write'),
+    onBlur: updateAllowedAccess,
+  };
+
+  const allowTempAccessProps: DateTimeFieldProps = {
+    name: 'allowTemporarySuperuserAccess',
+    label: t('Allow temporary access for Sentry employees'),
```
sentaur-athena, line 122: nit: My english is not good so I don't know which one is wrong but one of them has to be `Allow temporary access for Sentry employees` vs `Allow access to Sentry employees`. Is it to? or is it for? 😛

static/app/views/settings/components/dataSecrecy/index.tsx
iamrajjoshi: most endpoints are able to use the default retry, but since the api will return a 404 if an org doesn't have a waiver currently, the default behavior of the api would be to keep refreshing then render a failure state. here we prevent this by not refreshing on 404s.

iamrajjoshi:
> When the toggle is off and the time is in the past my suggestion is to clear the time field, showing basically there's no access going on at the moment.

i agree with when the toggle is off, but my concern with hiding when its in the past is we don't currently show when it was past easily from the UI, so i could see someone waiving then coming back and being confused why it disappeared and think its a bug.

static/app/views/settings/components/dataSecrecy/index.tsx
iamrajjoshi: got opinion from docs and am switching to "to" lol

sentaur-athena:
> so i could see someone waiving then coming back and being confused why it disappeared

It's fine as is. I don't feel strongly for it.

sentaur-athena (approved):

static/app/views/settings/components/dataSecrecy/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+      accessStart: moment().utc().toISOString(),
+      accessEnd: moment.tz(allowDateFormData, moment.tz.guess()).utc().toISOString(),
+    };
+
+    try {
+      await await api.requestPromise(
+        `/organizations/${organization.slug}/data-secrecy/`,
+        {
+          method: 'PUT',
+          data: nextData,
```
sentaur-athena, line 93: wait a minute. Is this working? Now that you camelcased the frontend code I don't think passing this directly works.

iamrajjoshi: i updated the endpoint to support camel case and now this is working

PR #75793, opened 2024-08-07
feat(anomaly detection): Add anomaly detection alerts option + preliminary work for new form fields

mifu67:
Add the option for anomaly detection alerts. When the user clicks on the option, render something different as a placeholder. (The orange border is just to highlight the change.)

https://github.com/user-attachments/assets/a5f55b1d-c7c8-4d54-bdd4-d072ea918aa2

### DIFF
--- a/static/app/views/alerts/rules/metric/thresholdTypeForm.tsx
+++ b/static/app/views/alerts/rules/metric/thresholdTypeForm.tsx
@@ -35,7 +35,9 @@ function ThresholdTypeForm({
     return null;
   }
 
-  const hasAnomalyDetection = organization.features.includes('organizations:anomaly-detection-alerts')
+  const hasAnomalyDetection = organization.features.includes(
+    'organizations:anomaly-detection-alerts'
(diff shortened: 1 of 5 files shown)


mifu67:
The backend change here is just the update to the feature flag. Nothing is interdependent.

static/app/views/alerts/rules/metric/thresholdTypeForm.tsx
```diff
@@ -35,57 +35,62 @@ function ThresholdTypeForm({
...
+            options={COMPARISON_DELTA_OPTIONS}
+            required={comparisonType === AlertRuleComparisonType.CHANGE}
+          />
+        </ComparisonContainer>
+      ),
+    ],
+  ];
+
+  if (hasAnomalyDetection) {
+    thresholdTypeChoices.push([
```
nhsiehgit, line 82: 👌 nice

nhsiehgit (approved):
👏

PR #75794, opened 2024-08-08
feat(uptime): Add api for creating new uptime monitors

wedamija:
This api is currently gated behind a feature flag, so that we can just allow this on orgs we want to test on. We shouldn't enable this on other orgs until we have billing related concerns sorted out.

### DIFF
--- a/src/sentry/uptime/endpoints/project_uptime_alert_index.py
+++ b/src/sentry/uptime/endpoints/project_uptime_alert_index.py
@@ -24,7 +24,7 @@ from sentry.uptime.endpoints.validators import UptimeMonitorValidator
 @extend_schema(tags=["Uptime Monitors"])
 class ProjectUptimeAlertIndexEndpoint(ProjectEndpoint):
     publish_status = {
-        "POST": ApiPublishStatus.PUBLIC,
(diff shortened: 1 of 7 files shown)


gaprl (approved):

src/sentry/uptime/endpoints/project_uptime_alert_index.py
```diff
@@ -0,0 +1,63 @@
...
+from sentry.models.project import Project
+from sentry.uptime.endpoints.serializers import ProjectUptimeSubscriptionSerializer
+from sentry.uptime.endpoints.validators import UptimeMonitorValidator
+
+
+@region_silo_endpoint
+@extend_schema(tags=["Uptime Monitors"])
+class ProjectUptimeAlertIndexEndpoint(ProjectEndpoint):
+    publish_status = {
+        "POST": ApiPublishStatus.PUBLIC,
```
isabellaenriquez, line 27: I know the flag will bar customers from creating the monitors if they don't have it, but is it necessary to publish to our docs right now?

wedamija: Moved this over to experimental

PR #75795, opened 2024-08-08
fix(tests): Fix `test_get_api_names` so that it doesn't need to be updated when adding new events

wedamija:
Whenever we register a new event we have to add it to the test here, which seems redundant. Mocking the internal data structures so we can just register our own events and not have to update the test


wedamija:
@mbauer404 I think this is fine to do, let me know if this will cause any problems

mbauer404 (approved):

PR #75829, opened 2024-08-08
fix(breadcrumbs): Handle exceedingly large values properly

leeandher:
Two changes here:
- Long message values would clutter the rest of the page. Now they get trimmed at 200 characters in the issue details page preview. They are still fully accessible in the drawer.
- Long categories would flow outside of their dedicated space causing formatting issues but also obscuring data (like timestamps) in the drawer

**before**

image

**after**
image
image

### DIFF
--- a/static/app/components/events/breadcrumbs/breadcrumbItemContent.tsx
+++ b/static/app/components/events/breadcrumbs/breadcrumbItemContent.tsx
@@ -19,10 +19,12 @@ import {isUrl} from 'sentry/utils/string/isUrl';
 import {usePrismTokens} from 'sentry/utils/usePrismTokens';
 
 const DEFAULT_STRUCTURED_DATA_PROPS = {
-  maxDefaultDepth: 2,
+  maxDefaultDepth: 1,
   withAnnotatedText: true,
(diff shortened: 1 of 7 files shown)


static/app/components/events/breadcrumbs/breadcrumbItemContent.tsx
```diff
@@ -45,7 +47,24 @@ export default function BreadcrumbItemContent({
...
+        <StructuredData
+          value={bc.message}
+          meta={meta?.message}
+          {...structuredDataProps}
+        />
+      ) : (
+        <StructuredData
+          value={
+            bc.message.length > MESSAGE_PREVIEW_CHAR_LIMIT
+              ? bc.message.substring(0, MESSAGE_PREVIEW_CHAR_LIMIT) + '...'
```
scttcper, line 60: i've heard to prefer the unicode character "it provides a semantically meaningful symbol rather than just three generic periods"
```suggestion
              ? bc.message.substring(0, MESSAGE_PREVIEW_CHAR_LIMIT) + '\u2026'
```

scttcper (approved):

PR #75831, opened 2024-08-08
fix(discover): Clone eventView when injecting event.type for alert

narsaynorath:
By directly updating the eventView, we're modifying the state of the page. Clone it to avoid updating the event view.


shruthilayaj (approved):
lgtm

PR #75840, opened 2024-08-08
feat(explore): Create skeleton for new explore page

Zylphrex:
Just laying out the skeleton for the page and what the top level components that are needed are.


Abdkhan14 (approved):
Lgtm, it's all behind a flag, we'll iterate

PR #75849, opened 2024-08-08
feat(new-trace): Updating canFetch property of transactions based on spans count

Abdkhan14:
Note: For testing, try out next.js traces, they usually have the Missing Spans Case.

Customers are confused by the `Missing spans `scenario, for which we currently render a `No Data` node and the following CTA:
Screenshot 2024-08-12 at 1 14 48 PM

We've decided to make transactions that don't have embedded spans unable to fetch: We wont render the '+' icons, they will just be leaf nodes. We still need to decide if we want a CTA for this case, as the sdk team doesn't always see Missing spans as a problem. 

The `events-trace-meta` endpoint now gives us an array of 'transaction_id' -> 'span_children_count' called `transaction_child_count_map` that I used to achieve this new functionality. This means that we need to wait for the meta endpoint response for the tree, which is fine since the endpoint is mostly faster than the trace endpoint: [comparing durations](https://sentry.sentry.io/discover/homepage/?dataset=discover&display=top5&environment=prod&field=transaction&field=p99%28transaction.duration%29&id=16351&interval=1d&name=&project=1&query=transaction%3A%2Fapi%2F0%2Forganizations%2F%7Borganization_id_or_slug%7D%2Fevents-trace%2F%7Btrace_id%7D%2F+or+transaction%3A%2Fapi%2F0%2Forganizations%2F%7Borganization_id_or_slug%7D%2Fevents-trace-meta%2F%7Btrace_id%7D%2F&queryDataset=error-events&sort=-transaction&statsPeriod=14d&topEvents=5&yAxis=p99%28transaction.duration%29)

I will put up two stacked PRs:
- Pass the correct timestamp to the `events-trace-meta` endpoint for replay traces. Otherwise we are currently not getting the correct `transaction_child_count_map` for replay traces.
- Remove all the `No data` node code.


static/app/views/performance/newTraceDetails/traceApi/useTraceMeta.tsx
```diff
@@ -158,15 +184,13 @@ export function useTraceMeta(traceSlugs: string[]): TraceMetaQueryResults {
         performance_issues: 0,
         projects: 1,
         transactions: 1,
+        transactiontoSpanChildrenCount: {'1': 0},
```
JonasBa, line 187: Is this needed?

JonasBa (approved):
Looks good. My one concern is just that we should have some link for transactions without spans somewhere where we either link users to manual instrumentation docs or allow them to provide feedback (this could be global and not related to the clicked node). The reason I think it'd be helpful is because we've received feedback from users who dont know what to do in this case and need guidance(taking that away without providing a fallback seems like a bad choice)

static/app/views/performance/newTraceDetails/traceModels/traceTree.tsx
```diff
@@ -562,7 +568,13 @@ export class TraceTree {
               parent.parent_transaction?.metadata.project_slug,
       });
 
-      node.canFetch = true;
+      // We check for at least 2 spans because the first span is the transaction itself.
+      if (isTransactionNode(node)) {
+        const spanChildrenCount =
+          metaResults?.data?.transactiontoSpanChildrenCount[node.value.event_id];
+        node.canFetch = spanChildrenCount ? spanChildrenCount >= 2 : false;
+      }
```
JonasBa, line 576: If the result of metaResults.data lookup is undefined, then node.canFetch has to return true. It can only return false if the key is present in the lookup and the value is >= 2

PR #75853, opened 2024-08-08
fix(discover): Allow for errors and transactions in data export

narsaynorath:
Registers the errors and transactions dataset as supported datasets for export for the discover split.

### DIFF
--- a/src/sentry/data_export/endpoints/data_export.py
+++ b/src/sentry/data_export/endpoints/data_export.py
@@ -29,6 +29,8 @@ from ..tasks import assemble_download
 SUPPORTED_DATASETS = {
     "discover": Dataset.Discover,
     "issuePlatform": Dataset.IssuePlatform,
+    "transactions": Dataset.Transactions,
+    "errors": Dataset.Events,
 }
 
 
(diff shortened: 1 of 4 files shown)


src/sentry/data_export/endpoints/data_export.py
```diff
@@ -29,6 +29,8 @@
 SUPPORTED_DATASETS = {
     "discover": Dataset.Discover,
     "issuePlatform": Dataset.IssuePlatform,
+    "transactions": Dataset.Transactions,
+    "errors": Dataset.Events,
```
shruthilayaj, line 33: Works for transactions, but I think for errors, we should use the ErrorsQueryBuilder? It inherits DiscoverQueryBuilder but has an ErrorsQueryBuilderMixin 👀

narsaynorath: :+1: added a condition for the ErrorsQueryBuilder here
https://github.com/getsentry/sentry/pull/75853/commits/69678d3de6f0ad880fad8bb7eb03cd3b535acba8

shruthilayaj (approved):

PR #75857, opened 2024-08-08
ref(control_silo): Move over Authenticator model to Users module

Christinarlong:
Part of moving control silo user related resources into the users module. Includes adding of types for functions.

Apart of (#73856)


src/sentry/models/__init__.py
```diff
@@ -1,3 +1,4 @@
+from ..users.models.authenticator import *  # NOQA
```
markstory, line 1:
```suggestion
from sentry.users.models.authenticator import *  # NOQA
```

Lets stick to absolute imports. They are easier to understand and more portable.

markstory (approved):
Looks good to me. :clap:

markstory (approved):
Looks good to me.

Issue #75860, opened 2024-08-08
Events only getting halfway through ingest causing ML grouping weirdness


lobsterkatie:
Documenting for posterity:

We recently ran into a situation where it seemed as if a) we were ignoring Seer results and creating a new group even when Seer found a match, and b) sending the same hash to Seer multiple times. 

But TL;DR, here is what _actually_ happened:

- event A (with hash 1) put hash 1 into Seer, but disappeared before it could create a group for hash 1
- event B (with hash 2) matched hash 1 in Seer, but disappeared before it could create a group for hash 2 or be put into a group for hash 1
- event C (with hash 2) matched hash 1 in Seer, but no hash 1 group existed, so it created a group for hash 2
- event D (with hash 1) had no Seer match, so it created a group for hash 1 

The more detailed explanation, with links below:

The sequence of events _appeared to be_ this (where the starred lines seemed to indicate bugs):

- Event A with hash 1 
  - sent to seer
  - seer found no match
  - hash 1 added to seer DB
  - sentry created group 1 for hash 1 events
- Event B with hash 2
  - sent to seer
  - seer matched hash 2 to hash 1
  - based on seer result, sentry should have put event into group 1, but instead created group 2 for hash 2 events *****
- Event C with hash 2
  - based on hash, sentry should have put event into group 2, but instead sent it to seer *****
  - seer again matched hash 2 to hash 1
  - based on seer result, sentry should have put event into group 1, but instead put it in group 2 *****
- Event D with hash 1
  - based on hash, sentry should have put event into group 1, but instead sent it to seer *****
  - seer found no match (seems you can't match to yourself?)
  - sentry put event into group 1

But it turned out the sequence was actually this (where the starred lines indicate the actual problem):

- Event A with hash 1 
  - sent to seer
  - seer found no match
  - hash 1 added to seer DB
  - event got lost partway through ingest, so group 1 wasn't created *****
- Event B with hash 2
  - sent to seer
  - seer matched hash 2 to hash 1
  - event got lost partway through ingest, so it was neither assigned to group 1 (which would have been hard in any case, since group 1 didn't exist at that point) nor used to create group 2 *****
- Event C with hash 2
  - group 2 didn't yet exist, so event sent to seer
  - seer again matched hash 2 to hash 1
  - based on seer result, sentry should have put event into group 1, but group 1 still didn't yet exist, so event was used to create group 2
- Event D with hash 1
  - group 1 didn't yet exist, so event sent to seer
[...]

PR #75861, opened 2024-08-08
fix(seer grouping): Handle Seer returning grouphash with no group

lobsterkatie:
We have lately run into [an issue](https://github.com/getsentry/sentry/issues/75860) where Seer is recommending hashes with no group attached to them, which has highlighted a) the fact that this is even a possibility, and b) the fact that though we've been talking about Seer sometimes recommending non-existent _groups_, in fact all we check for is the existence of the grouphash. (The assumption there being that all grouphashes point to a group. If everything else is working correctly that should be true, but clearly stuff happens.)

This PR therefore makes two changes:

- Change the `SimilarGroupNotFoundError` to be a `SimilarHashNotFoundError`, to more accurately reflect the situation when it arises.

- Add a new `SimilarHashMissingGroupError` to denote the case described above, so that we can track it rather than silently swallowing the problem.

### DIFF
--- a/src/sentry/seer/similarity/types.py
+++ b/src/sentry/seer/similarity/types.py
@@ -13,7 +13,7 @@ class IncompleteSeerDataError(Exception):
     pass
 
 
-class SimilarGroupNotFoundError(Exception):
+class SimilarHashNotFoundError(Exception):
     pass
 
 
@@ -70,7 +70,7 @@ class SeerSimilarIssueData:
         similar issue in the Seer response.
 
         Throws an `IncompleteSeerDataError` if given data with any required keys missing, and a
-        `SimilarGroupNotFoundError` if the data points to a group which no longer exists. The latter
+        `SimilarHashNotFoundError` if the data points to a group which no longer exists. The latter
         guarantees that if this successfully returns, the parent group id in the return value points
         to an existing group.
 
@@ -99,7 +99,7 @@ class SeerSimilarIssueData:
 
         if not parent_grouphash:
             # TODO: Report back to seer that the hash has been deleted.
-            raise SimilarGroupNotFoundError("Similar group suggested by Seer does not exist")
+            raise SimilarHashNotFoundError("Similar group suggested by Seer does not exist")
 
         # TODO: The `Any` casting here isn't great, but Python currently has no way to
         # relate typeddict keys to dataclass properties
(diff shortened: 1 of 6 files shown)


src/sentry/seer/similarity/types.py
```diff
@@ -99,7 +103,11 @@ def from_raw(cls, project_id: int, raw_similar_issue_data: Mapping[str, Any]) ->
 
         if not parent_grouphash:
             # TODO: Report back to seer that the hash has been deleted.
-            raise SimilarGroupNotFoundError("Similar group suggested by Seer does not exist")
+            raise SimilarHashNotFoundError("Similar hash suggested by Seer does not exist")
+
+        if not parent_grouphash.group_id:
```
armenzg, line 108: I have confirmed that we have ~114M group hashes without a group ID.

```
select
  count(*)
from
  getsentry.sentry_grouphash as gh
where
  gh.group_id is null;
```

I have checked that a system hash and an in-app hash both point to a group_id:
image

We have an example case in the Sentry project:
image

src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -494,15 +495,20 @@ def update_groups(project, seer_response, group_id_batch_filtered, group_hashes_
...
+                SimilarHashNotFoundError,
+                SimilarHashMissingGroupError,
+            ) as err:
                 logger.exception(
                     "tasks.backfill_seer_grouping_records.invalid_parent_group",
                     extra={
                         "project_id": project.id,
                         "group_id": group.id,
                         "parent_hash": groups_with_neighbor[str(group.id)]["parent_hash"],
+                        "error": err,
```
armenzg, line 511: `logger.exception` already displays the stack trace. [This issue](https://sentry.sentry.io/issues/5570694624/events/7370b02a1ee943098868ad3688d05da2/?project=1) shows it:
image

armenzg (approved):

src/sentry/seer/similarity/types.py
lobsterkatie: Wow - IIRC there are ~800 M grouphashes, so that's ~12% - much higher than I would have expected!

I think the only way they could get into Seer is through the disappearing-events bug I mentioned in the PR description, as backfill should never send them (given that they have no group and therefore no event to get a stacktrace from). That means that the `SimilarHashMissingGroupError` added here should always indicate a disappearing event. So... we'll see how often it comes up.

src/sentry/tasks/embeddings_grouping/utils.py
lobsterkatie: I was more concerned about the GCP logs, but I just checked and you're right, it shows up there, too. Thanks for pointing that out. I can remove it.

PR #75862, opened 2024-08-08
inc-847: modify transactions consumer to support reingesting ingest-transactions

lynnagara:


fpacifici (commented):
Please let's filter out events we have already successfully processed. 
You can do that by checking if the event key is already in rc-processing. 
If it is already in rc-processing and you are in rcovery mode, it is an event to recover. 
If it is not in rc-processing it means the event is good and you can skip it.

fpacifici:
Actually this is already implemented 
https://github.com/getsentry/sentry/blob/master/src/sentry/ingest/consumer/processors.py#L144-L148

fpacifici (commented):
Also I think it would be a smart idea to log the event id that we are reprocessing once in a while, so we can check if the missing event ids actually disappear from redis and appear in clickhouse. 

If we do not do that, verifying the operaiton is working would be quite hard.

src/sentry/ingest/consumer/processors.py
```diff
@@ -149,10 +153,22 @@ def process_event(
         if reprocess_only_stuck_events and not event_processing_store.exists(data):
             return
 
+        is_reprocessed = event_processing_store.exists(data)
```
fpacifici, line 157: please put this in an else block of the previous block so you do not read from redis twice (exists).
```
event_exists = event_processing_store.exists(data)
if reprocess_only_stuck_events and not event_exists:
     return
if event_exists:
   ...the rest ...
```

src/sentry/ingest/consumer/processors.py
```diff
@@ -149,10 +153,22 @@ def process_event(
...
+        # TODO: make the 100 an option
+        if len(reprocessed_ids) > 100:
+            logger.info(reprocessed_ids)
+            reprocessed_ids.clear()
+            reprocessed_last_logged_time = now
+
+        if reprocessed_last_logged_time is None or now > reprocessed_last_logged_time + 10:
+            logger.info(reprocessed_ids)
+            reprocessed_ids.clear()
+            reprocessed_last_logged_time = now
```
fpacifici, line 171: This should be there only if the event exists. We do not want to do this for all the other events in the main consumer

victoria-yining-huang: also should change to 500? @fpacifici you said it could be more

src/sentry/ingest/consumer/processors.py
lynnagara: oh actually, this can be removed entirely right? since we'll never proceed past line 154 unless we're reprocessing

src/sentry/ingest/consumer/processors.py
fpacifici: 250 it is

victoria-yining-huang (approved):

fpacifici (approved):

PR #75863, opened 2024-08-08
ref(issue-details): Variety of fixes/refactors for issue details

leeandher:
List of fixes:
- No longer looks awful on safari

fix safari

- Trace now has its own data section

image

- All thread sections are nested in a parent

image

- All stack traces use the same section key (`exception`, `threads`, and `stacktrace` event entries)
- If first and last release match, the header no longer shows both

image

- The search input now uses options from all events filter. The filter is persisted in the new context provider (gonna be useful for updating graph)
image

Some other refactors:
- The search bar on all events was extracted into its own component to be used in both areas while working on the new UI
- There's a context provider surrounding the data sections for issue details now. It currently only stores the search query, but the goal is to store open/close states for different sections and other useful info at a page level.


static/app/views/issueDetails/streamlinedHeader.tsx
```diff
@@ -125,26 +125,26 @@ export default function StreamlinedGroupHeader({
             <Fragment>
               <Divider />
               <ReleaseWrapper>
-                {t('Releases')}
+                {firstRelease.id === lastRelease.id ? t('Release') : t('Releases')}
```
roggenkemper, line 128: thanks for making this change!!

roggenkemper (approved):

PR #75864, opened 2024-08-08
feat(insights): add insights related issues table flag

KevinL10:
https://github.com/getsentry/sentry-options-automator/pull/2099


colin-sentry (approved):
LGTM but this is a flagpole flag, not a sentry-options-automator flag

PR #75871, opened 2024-08-08
ref(insights): remove unused All module

mjq:
The "All" module is a holdover from the very early days of Starfish and not used anywhere. Remove it.


gggritso (approved):
🚮

PR #75875, opened 2024-08-08
ref(similarity): Add postgres query retry with decreased batch size

jangjodi:
Postgres query in similarity backfill is sometimes timing out
The first time this exception occurs, retry with a smaller batch size
If it occurs again, raise it

### DIFF
--- a/src/sentry/tasks/embeddings_grouping/utils.py
+++ b/src/sentry/tasks/embeddings_grouping/utils.py
@@ -3,6 +3,7 @@ import time
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import asdict
 from datetime import UTC, datetime, timedelta
+from sqlite3 import OperationalError
 from typing import Any, TypedDict
 
 import sentry_sdk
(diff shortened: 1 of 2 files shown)


src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -3,6 +3,7 @@
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import asdict
 from datetime import UTC, datetime, timedelta
+from sqlite3 import OperationalError
```
mrduncan, line 6: This import looks wrong to me since we don't use sqlite3, though it might be unintentionally working due to some aliasing

jangjodi: good catch, thanks! I've changed it to import from `django.db.utils import OperationalError`

tests/sentry/tasks/test_backfill_seer_grouping_records.py
```diff
@@ -1724,3 +1725,26 @@ def test_backfill_seer_grouping_records_manually_skip_project(self, mock_logger)
...
         assert mock_logger.info.call_args_list == expected_call_args_list
+
+    @with_feature("projects:similarity-embeddings-backfill")
+    @patch("sentry.tasks.embeddings_grouping.utils.logger")
+    @patch("sentry.tasks.embeddings_grouping.utils._make_postgres_call")
+    def test_backfill_seer_grouping_records_postgres_exception(
+        self, mock_make_postgres_call, mock_logger
+    ):
+        """
+        Test log after postgres query retries with decreased batch size
```
lobsterkatie, line 1736: Is there a reason to test the logging rather than the behavior itself?

tests/sentry/tasks/test_backfill_seer_grouping_records.py
```diff
@@ -1724,3 +1725,26 @@ def test_backfill_seer_grouping_records_manually_skip_project(self, mock_logger)
...
+        """
+        mock_make_postgres_call.side_effect = OperationalError
```
lobsterkatie, line 1738: Tiny nit: Do you know that you can set this straight in the `@patch` decorator? (If you're choosing not to, that's fine (and I'd be curious to hear why), but have seen this pattern in a few tests and keep meaning to bring it up.)

lobsterkatie (approved):
LGTM!

tests/sentry/tasks/test_backfill_seer_grouping_records.py
jangjodi: since the end behavior is still raising the error, testing the logging allows me to check that it was retried with the lower batch size

tests/sentry/tasks/test_backfill_seer_grouping_records.py
jangjodi: no reason! I'll change this and follow up to change the others as well

tests/sentry/tasks/test_backfill_seer_grouping_records.py
lobsterkatie: Ah, I see. Up to you, but might be worth adding a comment to that effect.

PR #75877, opened 2024-08-08
feat(query-builder): Add recent filter keys to menu

malwilley:
We will now fetch recent searches and find the top 5 used filters to suggest when adding a new filter key. Filters already present in the search will not be shown.

https://github.com/user-attachments/assets/8bcfc57d-7b7a-4e5c-b24b-c8029912944a


static/app/components/searchQueryBuilder/tokens/filterKeyListBox/useFilterKeyListBox.tsx
```diff
@@ -1,64 +1,201 @@
...
+    );
+  };
+
+  const shouldShowExplorationMenu =
+    filterValue.length === 0 && filterKeySections.length > 0;
+
+  // This logic allows us to treat the recent filters as a separate section.
+  // If we have 5 recent filters, we want arrow up/down to cylce from the
+  // first recent filter to the first non-recent filter and vice versa.
+  const handleArrowUpDown = useCallback(
```
malwilley, line 106: Most of the changes are straightforward (adding items to the list and rendering them), but this was a bit hacky. By default, arrow up/down will take you through each of the recent filters. But since we render them horizontally, we  are modifying the arrow key behavior to make left/right go through the recent filters.

static/app/components/searchQueryBuilder/tokens/filterKeyListBox/index.tsx
```diff
@@ -125,7 +210,7 @@ export function FilterKeyListBox<T extends SelectOptionOrSectionWithKey<string>>
               ref={listBoxRef}
               listState={state}
               hasSearch={false}
-              hiddenOptions={hiddenOptions}
+              hiddenOptions={modifiedHiddenOptions}
```
leeandher, line 213: Is this so they don't appear twice? TBH I don't think it'd be a bad thing if they did, kinda like sticky notes in a phone book or something, I wouldn't expect them to be removed from the bigger list.

static/app/components/searchQueryBuilder/tokens/filterKeyListBox/index.tsx
```diff
@@ -155,18 +240,62 @@ const SectionedOverlayFooter = styled('div')`
...
+  font-weight: ${p => p.theme.fontWeightNormal};
+  font-size: ${p => p.theme.fontSizeMedium};
+  padding: 0 ${space(1.5)} 0 ${space(0.75)};
+  background: ${p => p.theme.background};
+  box-shadow: inset 0 0 0 1px ${p => p.theme.innerBorder};
+  border-radius: ${p => p.theme.borderRadius} 0 0 ${p => p.theme.borderRadius};
+  cursor: pointer;
+
+  /* Fade out on right side to represent that this is a filter key only */
+  &::before {
```
leeandher, line 279: This is just so clever! 👏 👏

static/app/components/searchQueryBuilder/tokens/filterKeyListBox/useFilterKeyListBox.tsx
leeandher: Is there a limit to the quantity of recent filters? In theory this'd be fine if the user wants to use one of their recent filters for the query they're typing but if they don't they'd have to 'down'-press past them all to get to the list. I mean it's not that big a deal but more of a question (they can always type or use the cursor)

leeandher (approved):
Looks good! Great job with this component btw 🔥

static/app/components/searchQueryBuilder/tokens/filterKeyListBox/index.tsx
leeandher: I'll actually say I would find it odd if they were removed from the greater list, since if someone were to ignore suggestions, they'd scroll down to the lists and find their options missing.

malwilley: I'll clarify this in the code with a comment since it is confusing, but this `hiddenOptions` simply prevents the recent filters items from showing in the `<ListBox />` component. They have keys that look like `__recent_filter__<name>` so the keys don't conflict with the normal items. We use `hiddenOptions` because the whole component shares a collection state so it's difficult to just remove the items from the list here.

static/app/components/searchQueryBuilder/tokens/filterKeyListBox/index.tsx
malwilley: All credit to @vuluongj20 💯

static/app/components/searchQueryBuilder/tokens/filterKeyListBox/useFilterKeyListBox.tsx
malwilley: Yeah, right now this only displays a max of 5 recent filters at the top (for space reasons). This arrow key behavior does help save a few keystrokes if they don't care about recent filters. And I also thought it seemed more natural behavior-wise since they are horizontal and in their own container

static/app/components/searchQueryBuilder/tokens/filterKeyListBox/index.tsx
leeandher: I see, so were hiding rendering them twice, not omitting them from the list itself, makes sense. Thanks for clarifying!

PR #75878, opened 2024-08-08
fix(replay/feedback): use dataset specific tags for search typeaheads

### LINKED ISSUE #75745: Fetch tags from issue platform to improve replay/feedback search bar performance
From: https://github.com/getsentry/sentry/pull/75552 and https://github.com/getsentry/sentry/pull/75709

Image

Note that tags are currently not fetched/loaded for feedback:
Image

aliu39:
Fixes https://github.com/getsentry/sentry/issues/75745
Previously we were using the legacy global TagStore, which includes every tag in the org and loads slower. With this hook we can fetch just the tags we need, from the right dataset

Before: all these _ tags on top are irrelevant to replay. There's a limit on how many items the typeahead can display, and right now there are so many that the actual replay fields (url, user.email...) get crowded out.
Screenshot 2024-08-09 at 9 56 08 AM

I haven't made use of the isLoading or isError states since I couldn't find such states for the search bar. Like before, the tags will appear in the list once loaded, no transition (hopefully these will be faster).


malwilley (commented):
This is a great change for the feedback searchbar! But for the replay one, I believe you would want the replay dataset instead

aliu39:
> This is a great change for the feedback searchbar! But for the replay one, I believe you would want the replay dataset instead

Oh I see! I misunderstood, replays isn't in the Dataset enum but I guess it's not complete

aliu39:
I made a new `Dataset` enum in tag actionCreators, so we don't overload the alerts type. This was previously causing errors in https://github.com/getsentry/sentry/blob/041e1ab95fbe154ab638d2d3e8722fa72dfecd0f/static/app/views/alerts/wizard/utils.tsx#L11.

Let me know if you'd like this last commit to be split to a separate PR.

static/app/utils/replays/hooks/useFetchDatasetTags.tsx
```diff
@@ -0,0 +1,60 @@
...
+  keepPreviousData?: boolean;
+  start?: string;
+  statsPeriod?: string | null;
+  useCache?: boolean;
+};
+
+/**
+ * Fetches tags from exclusively one dataset. Based on useFetchIssueTags.
+ */
+export default function useFetchDatasetTags({
```
malwilley, line 22: Is it worth having this function instead of just using `useFetchOrganizationTags`? Seems like the params are pretty much the same

malwilley (commented):
Hmm testing this I'm getting an error from the backend:

> "Missing >= condition with a datetime literal on column timestamp for entity replays. Example: timestamp >= toDateTime('2023-05-16 00:00')"

The [endpoint](https://github.com/getsentry/sentry/blob/c23347e5cba990e2378b101ab9cd8c9b14e6a9cc/src/sentry/api/endpoints/organization_tags.py#L42) may need to be modified to support replays, not sure what's going on there

aliu39:
> Hmm testing this I'm getting an error from the backend:
> 
> > "Missing >= condition with a datetime literal on column timestamp for entity replays. Example: timestamp >= toDateTime('2023-05-16 00:00')"
> 
> The [endpoint](https://github.com/getsentry/sentry/blob/c23347e5cba990e2378b101ab9cd8c9b14e6a9cc/src/sentry/api/endpoints/organization_tags.py#L42) may need to be modified to support replays, not sure what's going on there

Tracked it to this Snuba query in [__get_tag_keys_for_projects](https://github.com/getsentry/sentry/blob/041e1ab95fbe154ab638d2d3e8722fa72dfecd0f/src/sentry/tagstore/snuba/backend.py#L318-L318).

And in the snuba parser: https://github.com/getsentry/snuba/blob/e6509ee9eac8e36f85679a08fcdf1d9c988984bb/snuba/query/snql/parser.py#L1381-L1393

Seems like the replays dataset has a required_time_column, queries to it need to have a >= and < for `timestamp`. It might be good to consult @cmanallen on the _best_ way to deal with this case, but he's out this week. I can give it a shot with a backend PR tomorrow.

static/app/utils/replays/hooks/useFetchDatasetTags.tsx
aliu39: It's mainly to avoid duplicating the useMemo/reduce pattern

aliu39:
> > Hmm testing this I'm getting an error from the backend:
> > > "Missing >= condition with a datetime literal on column timestamp for entity replays. Example: timestamp >= toDateTime('2023-05-16 00:00')"
> > 
> > 
> > The [endpoint](https://github.com/getsentry/sentry/blob/c23347e5cba990e2378b101ab9cd8c9b14e6a9cc/src/sentry/api/endpoints/organization_tags.py#L42) may need to be modified to support replays, not sure what's going on there
> 
> Tracked it to this Snuba query in [__get_tag_keys_for_projects](https://github.com/getsentry/sentry/blob/041e1ab95fbe154ab638d2d3e8722fa72dfecd0f/src/sentry/tagstore/snuba/backend.py#L318-L318).
> 
> And in the snuba parser: https://github.com/getsentry/snuba/blob/e6509ee9eac8e36f85679a08fcdf1d9c988984bb/snuba/query/snql/parser.py#L1381-L1393
> 
> Seems like the replays dataset has a required_time_column, queries to it need to have a >= and < for `timestamp`. It might be good to consult @cmanallen on the _best_ way to deal with this case, but he's out this week. I can give it a shot with a backend PR tomorrow.

Update: it seems like the backend query is more complicated than just adding >= < timestamp conditions. I think we should wait for after hackweek to do this so I can ask @cmanallen for help, and so we don't merge before hackweek.

We could merge the change for feedback though!

JoshFerge:
we should be able to use the normal tagstore as the tags that are on replay are usually the same as errors / transactions. at least that was the idea when originally implemented

aliu39:
> we should be able to use the normal tagstore as the tags that are on replay are usually the same as errors / transactions. at least that was the idea when originally implemented

@JoshFerge are the tags in the screenshot valid for replays?

aliu39:
> the tags that are on replay are usually the same as errors / transactions.
Maybe we can use issue platform for now

JoshFerge:
> > we should be able to use the normal tagstore as the tags that are on replay are usually the same as errors / transactions. at least that was the idea when originally implemented
> 
> @JoshFerge are the tags in the screenshot valid for replays?

i think if you only have the js project selected, any of those tags could be on a replay.

aliu39:
Here's an example of a tag that's not applicable for replays and shouldn't be showing up: `transaction`
https://sentry.slack.com/archives/C03USURCFBJ/p1723673755567199

PR #75879, opened 2024-08-08
[CapMan self-serve] Don't emit Sentry warnings for throttled queries to Snuba

xurui-c:
Allocation policies are our mechanism for doing traffic management for Snuba queries. Currently, we see a lot of warnings (Warning: Query from referrer ... is throttled) on Sentry Issues. This is because these queries are throttled due to being in the "warning zone"; however, we got feedback that this isn't actually actionable, so we're getting rid of them. We will only emit Sentry *errors* for rejected queries.

### DIFF
--- a/src/sentry/utils/snuba.py
+++ b/src/sentry/utils/snuba.py
@@ -4,7 +4,6 @@ import dataclasses
 import functools
 import logging
 import os
-import random
 import re
 import time
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


src/sentry/utils/snuba.py
```diff
@@ -1118,19 +1126,6 @@ def _bulk_snuba_query(snuba_requests: Sequence[SnubaRequest]) -> ResultSet:
...
-                    and quota_allowance_summary["throttled_by"]
-                ):
-                    metrics.incr("snuba.client.query.throttle", tags={"referrer": referrer})
-                    if random.random() < 0.01:
-                        logger.warning(
-                            "Warning: Query is throttled", extra={"response.data": response.data}
-                        )
-                        sentry_sdk.capture_message(
-                            f"Warning: Query from referrer {referrer} is throttled", level="warning"
-                        )
```
onewland, line 1132: What if, instead of deleting this, we made it configurable-by-referrer and default off?

So if there is a product team interested in monitoring throttles we can quickly turn on warnings for their referrer, but we don't annoy certain people by filling the issue dashboard with throttle warnings that won't be dealt with.

onkar (approved):
LGTM, let's include the suggestion in the scoping of the larger work.

PR #75881, opened 2024-08-08
ref(space) refactor template literals to use theme value

JonasBa:
Refactors template literals to use the theme value


biome.json
```diff
@@ -22,6 +22,7 @@
         "useHookAtTopLevel": "error",
         "useIsNan": "error",
         "noUnusedPrivateClassMembers": "error",
+        "noUnusedImports": "error",
```
malwilley, line 25: Did you mean to change this here? Or was it to help with the codemod

malwilley (commented):
Seems like a good omen

CleanShot 2024-08-08 at 14 36 04

biome.json
JonasBa: Nah, I'm removing these, I just set them so that biome could go and fix them. I'm going to go through all the rules to make sure nothing is borked.

JonasBa:
Let me close this until I get it into a functioning state so it wont spam folks

PR #75885, opened 2024-08-08
feat(data-secrecy): Implement Data Secrecy Backend

iamrajjoshi:
Sentry Version of https://github.com/getsentry/getsentry/pull/14779

Here, I create a new function to determine if we should prevent superuser access based on the new data secrecy logic. It lives behind the `organizations:data-secrecy` feature flag. I add it to `get_superuser_scopes` since if the person is accessing via superuser, it will go into the function.

[spec](https://www.notion.so/sentry/Superuser-Data-Secrecy-Mode-b9f7fdfd8b564615ae1f91d3d981bc1a)

### DIFF
--- a/src/sentry/data_secrecy/data_secrecy_logic.py
+++ b/src/sentry/data_secrecy/data_secrecy_logic.py
@@ -25,7 +25,7 @@ def should_allow_superuser_access(
     if not features.has("organizations:data-secrecy", organization):
         return True
 
-    # If organization's prevent_superuser_access bitflag is not set, return True
+    # If organization's prevent_superuser_access bitflag is False, return True
     if not organization.flags.prevent_superuser_access:
         return True
 
--- a/src/sentry/data_secrecy/data_secrecy_logic.py
+++ b/src/sentry/data_secrecy/data_secrecy_logic.py
@@ -1,0 +1,39 @@
+from django.conf import settings
+from django.utils import timezone
+
+from sentry import features
+from sentry.data_secrecy.service.service import data_secrecy_service
+from sentry.models.organization import Organization
+from sentry.organizations.services.organization import RpcOrganization, RpcUserOrganizationContext
+
+
+def should_allow_superuser_access(
+    organization_context: Organization | RpcUserOrganizationContext,
+) -> bool:
+
+    # If self hosted installation, superuser access is allowed
+    if settings.SENTRY_SELF_HOSTED:
+        return True
+
+    organization: Organization | RpcOrganization
+    if isinstance(organization_context, RpcUserOrganizationContext):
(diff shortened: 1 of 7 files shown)


src/sentry/data_secrecy/data_secrecy_logic.py
```diff
@@ -0,0 +1,39 @@
...
+    if isinstance(organization_context, RpcUserOrganizationContext):
+        organization = organization_context.organization
+    else:
+        organization = organization_context
+
+    # If organization does not have data-secrecy feature, return True
+    if not features.has("organizations:data-secrecy", organization):
+        return True
+
+    # If organization's prevent_superuser_access bitflag is not set, return True
```
cathteng, line 28: nit: clarify not set = False?

tests/sentry/api/test_data_secrecy.py
```diff
@@ -34,3 +34,41 @@ def test_non_member_no_access(self):
...
+    @with_feature("organizations:data-secrecy")
+    def test_superuser_no_access(self):
+        superuser = self.create_user(is_superuser=True)
+        self.login_as(superuser, superuser=True)
+
+        # superuser cannot access orgs with data secrecy
+        with self.settings(SENTRY_SELF_HOSTED=False):
+            self.get_error_response(self.organization.slug, status_code=401)
+
+    def test_superuser_has_access(self):
```
cathteng, line 58: should you have the feature flag here too?

tests/sentry/api/test_data_secrecy.py
```diff
@@ -34,3 +34,41 @@ def test_non_member_no_access(self):
 
     def test_member_has_access(self):
         self.get_success_response(self.organization.slug)
+
+
+class DataSecrecyV2TestCase(APITestCase):
```
cathteng, line 39: should we test that if an org has data secrecy and you're staff, you can access org endpoints used in admin, but if you're superuser at all, you can't?

cathteng (commented):
i think this looks good, just some suggestions for tests

tests/sentry/api/test_data_secrecy.py
```diff
@@ -72,3 +75,19 @@ def test_non_member_no_access(self):
     def test_member_has_access(self):
         with self.settings(SENTRY_SELF_HOSTED=False):
             self.get_success_response(self.organization.slug)
+
+    @with_feature("organizations:data-secrecy")
+    @override_options({"staff.ga-rollout": True})
+    def admin_access_when_superuser_no_access(self):
```
cathteng, line 81: nit: should this method begin with the word `test`?

cathteng (approved):

PR #75888, opened 2024-08-08
fix(alerts): Update QuerySubscription filter

ceorourke:
Follow up to https://github.com/getsentry/sentry/pull/75787 and https://github.com/getsentry/sentry/pull/75872

Fixes [SENTRY-3DDN](https://sentry.sentry.io/issues/5692294203/)


nhsiehgit (commented):
🤔 is this necessary?
the FK shouldn't this still target the same model instance?

nhsiehgit:
I'll let Dan get the final word tho 😅

wedamija (approved):

PR #75890, opened 2024-08-08
fix(slack): fix disappearing tags for issue button actions attempt 2

cathteng:
See #75789

Adds a fix to the above PR to account for someone clicking on an old message that doesn't have tags in the `View` `metadata`.


src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -530,6 +532,10 @@ def _handle_group_actions(
             except client.ApiError as error:
                 return self.api_error(slack_request, group, identity_user, error, "status_dialog")
 
+            view = View(**slack_request.data["view"])
+            private_metadata = orjson.loads(view.private_metadata)
+            original_tags_from_request = set(private_metadata.get("tags", {}))
```
cathteng, line 537: this is the only difference

sentaur-athena (approved):

PR #75893, opened 2024-08-08
feat(discover): add http_response_rate to metrics discover dataset

edwardgou-sentry:
Ports over the `http_response_rate` to discover metrics dataset


src/sentry/search/events/datasets/metrics.py
```diff
@@ -2018,3 +2045,28 @@ def _resolve_rate(
...
+            Function(
+                "equals",
+                [
+                    Column("metric_id"),
+                    self.resolve_metric("span.self_time"),
+                ],
+            ),
+            condition,
+            alias,
+        )
```
DominikB2014, line 2072: This looks quite identical to https://github.com/getsentry/sentry/blob/3860aba65a6d22152316f304bb2e13de18036eda/src/sentry/search/events/datasets/spans_metrics.py#L895-L918

Does it make sense to just share this function between metrics and spans_metrics?

edwardgou-sentry: Good point. I took a look at consolidating this and ran into some issues since this function uses other functions that are further duplicated in both datasets. I think I might save this for another PR since the code changes involved are more than expected

DominikB2014 (approved):

PR #75899, opened 2024-08-09
ref(messaging): Adapt other identity-linking to work like Slack

RyanSkonnord:
Generalize the recent refactoring in Slack's link and unlink identity views into the abstract linking view, so that it shares a path with MSTeams and Discord.

Change metric keys for MSTeams and Discord to be more consistent with Slack. Otherwise, behavior should be identical after refactoring.


src/sentry/integrations/slack/utils/notifications.py
```diff
@@ -163,25 +163,40 @@ def send_incident_alert_notification(
...
+            webhook_client = WebhookClient(response_url)
             webhook_client.send(text=text, replace_original=False, response_type="ephemeral")
             metrics.incr(
                 SLACK_LINK_IDENTITY_MSG_SUCCESS_DATADOG_METRIC,
                 sample_rate=1.0,
                 tags={"type": "webhook", "command": command},
             )
         except (SlackApiError, SlackRequestError) as e:
-            if unpack_slack_api_error(e) != EXPIRED_URL:
+            if "Expired url" not in str(e):
```
scefali, line 199: @RyanSkonnord should this be a constant?

src/sentry/integrations/messaging.py
```diff
@@ -280,72 +288,122 @@ def expired_link_template(self) -> str:
...
+            recipients=[request.user],
+            organization_id=organization.id,
+            provider=self.external_provider_enum,
+        )
+        has_provider_settings = controller.user_has_any_provider_settings(
+            self.external_provider_enum
+        )
+
+        if not has_provider_settings:
+            IntegrationNudgeNotification(organization, request.user, self.provider).send()
```
scefali, line 396: @RyanSkonnord Do we need this nudge still? I think the nudge was to get people to enable Slack back when Slack was disabled by default. But now Slack is enabled by default.

src/sentry/integrations/slack/utils/notifications.py
RyanSkonnord: Oh, I didn't do this on purpose. This was how it was in an earlier revision -- I must have goofed when I was diffing against Git history. I'll revert.

src/sentry/integrations/messaging.py
RyanSkonnord: I don't know, but I'll check. This is only here because I moved it from [`slack/views/link_identity.py` lines 155-158](https://github.com/getsentry/sentry/pull/75899/files/bd31b63b14da8b82c9994465a71c9b56ad8803c6#diff-2e2302a8e883ab01eb7334c5e15f596cdcb311e289c89aee74b6444d60e4571bL155-L158).

RyanSkonnord: I'm pretty sure this is dead code. I kludged out some type-checking errors (and explained in a comment) in https://github.com/getsentry/sentry/pull/75899/commits/f9e21028d9806f20511469416a4537e1ec7d3d9a. I'll open a PR after this one to delete it.

cathteng (approved):

src/sentry/integrations/messaging.py
```diff
@@ -355,25 +421,40 @@ def execute(
         """
         raise NotImplementedError
 
+    def notify_on_success(
+        self, external_id: str, params: Mapping[str, Any], integration: Integration | None
```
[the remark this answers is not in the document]
RyanSkonnord: Same meaning as the `external_id` field on the `Identity` model, so yes, the ID used by the external service of the entity being linked.

I'll add a docstring.

src/sentry/integrations/messaging.py
```diff
@@ -310,42 +330,96 @@ def handle(
...
-        self.notify_on_success(integration, params)
-
-        if self.success_metric:
-            provider_slug = self.parent_messaging_spec.provider_slug
-            analytics.record(
-                self.success_metric,
-                provider=provider_slug,
-                actor_id=request.user.id,
-                actor_type=ActorType.USER,
+        logger.info("get_identity_success", extra={"integration_id": integration_id})
```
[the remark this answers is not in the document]
RyanSkonnord: Sure, I'll add that.

PR #75901, opened 2024-08-09
feat(issue-resolve): stop using feature flag organizations:resolve-in-upcoming-release on the backend

scefali:
The feature flag `organizations:resolve-in-upcoming-release` has been there for some time (it was added while I was still a Sentry employee, lol). I'm going to assume it's been rolled out internally at least. But it's time to enable the feature globally so that people can start using it. I've tested it locally and seems to work fine. The purpose of this feature is that `Resolve in Next Release` doesn't work well if the latest event for an issue isn't the current release because if a new event comes on that new release, it automatically regresses. This was done because that sort of logic is needed for semver releases where current release might have different branches (e.x: an issue that appears in release 7.0 shouldn't be marked as resolved if release 6.1 happens, only if 7.1 happens). Here is the rollout plan.

1. Stop gating the `Resolve in Upcoming Release` feature on the backend
2. Stop gating the `Resolve in Upcoming Release` feature on the frontend and update 
3. Remove the feature flag altogether
4. Update docks

Here is what the feature looks like in the UI when it's enabled:
Screenshot 2024-08-09 at 3 10 09 PM

Here is what happens when you resolve the feature in a release then a new release gets released:
Screenshot 2024-08-09 at 2 57 46 PM

Related PR from @roggenkemper: https://github.com/getsentry/sentry/pull/70990

### DIFF
--- a/src/sentry/api/helpers/group_index/validators/status_details.py
+++ b/src/sentry/api/helpers/group_index/validators/status_details.py
@@ -1,6 +1,5 @@
 from rest_framework import serializers
 
-from sentry import features
 from sentry.models.release import Release
 
 from . import InCommitValidator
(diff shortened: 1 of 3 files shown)


scttcper (approved):

scefali:
@scttcper could you merge this? I don't have permission to do it.

PR #75903, opened 2024-08-09
feat: enable updating webhook events for published integration

scefali:
This PR enables updating events for published apps. There is a bug where we would prevent updating the webhook events because the scopes were out of order which actually turned out to be a blessing in disguise because the current implementation for updating the service hooks was not correct for published integrations which are installed by multiple integrations. As it happens, for that to work we have to update _every_ `ServiceHook` row that is attached to that integration. Because there could be thousands of them, I put that logic in a task because it would almost certainly time out if there are many installations.


src/sentry/sentry_apps/apps.py
```diff
@@ -185,30 +185,29 @@ def _update_events(self) -> None:
             self.sentry_app.events = expand_events(self.events)
 
     def _update_service_hooks(self) -> None:
-        hooks = hook_service.update_webhook_and_events(
-            organization_id=self.sentry_app.owner_id,
-            application_id=self.sentry_app.application_id,
+        if self.sentry_app.is_published:
+            # if it's a published integration, we need to do many updates so we have to do it in a task so we don't time out
+            # the client won't know it succeeds but there's not much we can do about that unfortunately
+            create_or_update_service_hooks_for_sentry_app(
```
scefali, line 191: @markstory So, I want this to run as an asynchronous control silo task but when I use `delay`, it doesn't work even though I specify the `silo_mode` as `CONTROL`. How do I do that?

markstory: Odd. The task looks like it is correctly defined, and the module containing the task is already part of celery's imports. Do you get an error when the task runs, or when it is scheduled?

scefali: @markstory well, I got it working now so I guess I'm good.

scefali:
@markstory I think there is some weird hybrid cloud shenanigans going on in this test which is causing it to fail. I added some print statements for the failure:
```
Service hook rpc: added hook for organization 4554523754102784 and application 76
Service hook rpc: added hook for organization 4554523754168325 and application 73
Service hook rpc: added hook for organization 4554523754233856 and application 73
start test with organization ids:  4554523754168325 4554523754233856 and application id:  73
running create_or_update_service_hooks_for_installation with application id 73
running create_or_update_service_hooks_for_installation with application id 73
task and outbox runner should be done by now with organization ids: 4554523754168325, 4554523754233856, application id: 73
```
Which has this test failure:
```
AssertionError: No service hooks found for Org1 (ID: 4554523754168325), App ID: 73
```

There are a number of things going on that don't make sense:
1. The print statements for the service hooks being updated are happening before the API call starts
2. There is an extraneous application and organization that' getting a service hook update
3. There are log statements that indeed the service hooks were created before we did the assertion

Here is the link to the test: https://github.com/getsentry/sentry/pull/75903/files#diff-253b2368a9e5befb9458c117ee39b0dda4a30afbd997dd320d0aff1d01ebb051R276

Any idea what's going on? I'm really out of clues and I can't replicate this locally. But I'm pretty sure this has to do with Hybrid Cloud because the project service hooks exist in the region silo while this is a control silo test.

src/sentry/tasks/sentry_apps.py
```diff
@@ -416,8 +416,10 @@ def send_webhooks(installation, event, **kwargs):
 def create_or_update_service_hooks_for_sentry_app(
     sentry_app_id: int, webhook_url: str, events: list[str], **kwargs: dict
 ) -> None:
+    print("create_or_update_service_hooks_for_sentry_app")
```
markstory: This will need to be removed to get through CI.

scefali:
@corps Maybe you have some idea about my last comment?

markstory:
I was not able to reproduce the failure when running tests in isolation. However, I was able to reproduce the problem by running the same slice of CI with

```shell
TEST_GROUP=2 TOTAL_TEST_GROUPS=11 MATRIX_INSTANCE_TOTAL=11 pytest tests --ignore tests/acceptance --ignore tests/apidocs --ignore tests/js --ignore tests/tools -x
```

tests/sentry/api/endpoints/test_sentry_app_details.py
```diff
@@ -219,12 +267,185 @@ def test_cannot_update_scopes_published_app(self):
...
+            self.published_app.webhook_url == "https://newurl.com"
+        ), f"Unexpected webhook URL: {self.published_app.webhook_url}"
+        # Check service hooks for each organization
+        with assume_test_silo_mode(SiloMode.REGION):
+            service_hooks_org1 = ServiceHook.objects.filter(
+                organization_id=org1.id, application_id=self.published_app.id
+            )
+            service_hooks_org2 = ServiceHook.objects.filter(
+                organization_id=org2.id, application_id=self.published_app.id
+            )
```
markstory, line 326: Shouldn't these conditions be `self.published_app.application_id` ?

scefali: @markstory You're a genius! And that was the problem. I guess what happened is that in my local env, the ids just happened to be the same so I didn't catch it. I bet there are a lot of bugs caused by that sort of behavior not being caught in tests.

src/sentry/sentry_apps/apps.py
```diff
@@ -185,30 +187,31 @@ def _update_events(self) -> None:
             self.sentry_app.events = expand_events(self.events)
 
     def _update_service_hooks(self) -> None:
-        hooks = hook_service.update_webhook_and_events(
-            organization_id=self.sentry_app.owner_id,
-            application_id=self.sentry_app.application_id,
+        if self.sentry_app.is_published:
+            # if it's a published integration, we need to do many updates so we have to do it in a task so we don't time out
+            # the client won't know it succeeds but there's not much we can do about that unfortunately
```
sentaur-athena, line 192: Please add logs on failure, so at least we have a way of monitoring if something goes wrong

sentaur-athena (approved):

src/sentry/sentry_apps/apps.py
scefali: @sentaur-athena are you talking about the task? Don't we automatically capture errors for every task?

PR #75905, opened 2024-08-09
ref(performance): Move PerformanceSearchQueryBuilder into its own com…

Zylphrex:
…ponent

This component will be reused anywhere we want to search for transactions.

### DIFF
--- a/static/app/components/performance/performanceSearchQueryBuilder.tsx
+++ b/static/app/components/performance/performanceSearchQueryBuilder.tsx
@@ -1,0 +1,126 @@
+import {useCallback, useMemo} from 'react';
+
+import {fetchTagValues} from 'sentry/actionCreators/tags';
+import {getHasTag} from 'sentry/components/events/searchBar';
+import {
+  STATIC_FIELD_TAGS_WITHOUT_ERROR_FIELDS,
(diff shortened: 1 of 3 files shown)


static/app/components/performance/performanceSearchQueryBuilder.tsx
gggritso, line 1: Earnest question, is this a "performance search query builder" or a "transaction search query builder"?

Zylphrex: Good question, this is definitely targeted for searching transactions so not limited to just the performance product. I can definitely rename this to `TransactionSearchQueryBuilder` if you'd like.

gggritso: I think that would be clearer and more likely to be re-used, but also, there's nowhere else to re-use this, so maybe moot? I leave it to you

gggritso (approved):

static/app/components/performance/performanceSearchQueryBuilder.tsx
Zylphrex: I'll be reusing this in some profiling pages as it searches transactions so that's a good call. Let me do that right now.

PR #75907, opened 2024-08-09
chore(profiling): Use transaction search query builder for profiles landing

Zylphrex:
This adopts the TransactionSearchQueryBuilder for the profiles landing


malwilley (approved):
Not sure if it matters, but the old landing searchbar had a placeholder and the new one does not. Don't see any other differences though!

Zylphrex:
> Not sure if it matters, but the old landing searchbar had a placeholder and the new one does not. Don't see any other differences though!

Good point, I'll fix this in a follow up as it'll be the default placeholder for all transaction searches

PR #75915, opened 2024-08-09
ref(control_silo): Move Email model to Users module

Christinarlong:
Move email model to the users module and leave shim for getsentry.

ref(#73856)


src/sentry/models/__init__.py
```diff
@@ -1,3 +1,4 @@
+from ..users.models.email import *  # NOQA
```
markstory, line 1:
```suggestion
from sentry.users.models.email import *  # NOQA
```

Lets stick to absolute imports. They are more portable and easier to understand.

markstory (approved):

src/sentry/models/__init__.py
Christinarlong: guh thank you for catching this, I need to make a checklist so I don't keep repeating this 😭

PR #75916, opened 2024-08-09
chore(profiling): Use transaction search query builder for transactio…

Zylphrex:
…n summary

Last 2 pages on transaction summary that can use this component.

### DIFF
--- a/static/app/views/performance/transactionSummary/transactionTags/content.tsx
+++ b/static/app/views/performance/transactionSummary/transactionTags/content.tsx
@@ -1,4 +1,4 @@
-import {useCallback, useEffect, useRef, useState} from 'react';
+import {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 import styled from '@emotion/styled';
 import type {Location} from 'history';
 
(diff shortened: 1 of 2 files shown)


static/app/views/performance/transactionSummary/transactionTags/content.tsx
```diff
@@ -181,6 +182,8 @@ function InnerContent(
 
   const query = decodeScalar(location.query.query, '');
 
+  const projectIds = useMemo(() => eventView.project?.slice(), [eventView.project]);
```
0Calories, line 185: nit: we should just pass in the projects as a prop to `TransactionSearchQueryBuilder` and memoize the ids within the component so we don't have to duplicate this line each time we use the component

0Calories (approved):
🙏

static/app/views/performance/transactionSummary/transactionTags/content.tsx
Zylphrex: the reason this is here is due to the incompatible types in `EventView.project` which is `readonly number[]` while `PageFilter['projects']` has a type of `number[]`.

0Calories: ah gotcha. can we not just cast it as `number[]` instead of cloning the array?

PR #75925, opened 2024-08-09
fix(ecosystems): Fixes missing integration proxy metrics after rename

GabeVillalobos:
Integration proxy stats metrics disappeared after #74793 landed, since the new path for this was not added to the metrics allowed paths list.

Updated the list to reinclude these metrics.

### Legal Boilerplate


leeandher (approved):

Christinarlong (approved):
tys! 👍

PR #75928, opened 2024-08-09
feat(stories): Add basic form fields story

evanpurkhiser:
image

### DIFF
--- a/static/app/components/forms/fields/index.stories.tsx
+++ b/static/app/components/forms/fields/index.stories.tsx
@@ -1,0 +1,130 @@
+import storyBook from 'sentry/stories/storyBook';
+import useProjects from 'sentry/utils/useProjects';
+
+import Form from '../form';
+
+import BlankField from './blankField';
+import BooleanField from './booleanField';
+import CheckboxField from './checkboxField';
+import DateTimeField from './dateTimeField';
+import EmailField from './emailField';
+import FileField from './fileField';
+import HiddenField from './hiddenField';
+import NumberField from './numberField';
+import RadioField from './radioField';
+import RangeField from './rangeField';
+import SecretField from './secretField';
+import SelectField from './selectField';
+import SentryMemberTeamSelectorField from './sentryMemberTeamSelectorField';
+import SentryProjectSelectorField from './sentryProjectSelectorField';
+import SeparatorField from './separatorField';
+import TextareaField from './textareaField';
+import TextField from './textField';
+
+export default storyBook(Form, story => {
+  story('Available fields', () => {
+    const {projects} = useProjects();
+
+    return (
+      <Form>
+        <HiddenField name="hidden" defaultValue="itsHidden" />
+        <BlankField label="My Blank Field" help="This is a blank field" />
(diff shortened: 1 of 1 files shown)


static/app/components/forms/index.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+import RadioField from './fields/radioField';
+import RangeField from './fields/rangeField';
+import SecretField from './fields/secretField';
+import SelectField from './fields/selectField';
+import SentryMemberTeamSelectorField from './fields/sentryMemberTeamSelectorField';
+import SentryProjectSelectorField from './fields/sentryProjectSelectorField';
+import SeparatorField from './fields/separatorField';
+import TextareaField from './fields/textareaField';
+import TextField from './fields/textField';
+import Form from './form';
```
ryan953, line 21: Lets make stories for each input field type on its own. A `<Form>` is not the same as a field and having everything in a giant file will not make it easier for people to find things in the future, or iterate on the examples

evanpurkhiser: For this I just want to see them all together.

I'm not going to spend more time on this now.

evanpurkhiser: IMO this is a very good starting point for seeing "what can I do with a form"

ryan953: I'd rename it `static/app/components/forms/fields/index.stories.tsx` at the very least. It's really not showing anything to do with `<Form>` specifically.

static/app/components/forms/index.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+export default storyBook(Form, story => {
```
ryan953, line 23: And change this
```suggestion
export default storyBook('Form fields', story => {
```

static/app/components/forms/index.stories.tsx
evanpurkhiser: Let me have it render the form state real quick, then it will be more like a form

ryan953: The "default" example for `<Form>` probably could just be without the fields, then it'll be easier to see the Submit button gets auto-inserted.

Then another example with fields and state is helpful

evanpurkhiser: I'll move it over. Will add the form state thing later

static/app/components/forms/fields/index.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+import RangeField from './rangeField';
+import SecretField from './secretField';
+import SelectField from './selectField';
+import SentryMemberTeamSelectorField from './sentryMemberTeamSelectorField';
+import SentryProjectSelectorField from './sentryProjectSelectorField';
+import SeparatorField from './separatorField';
+import TextareaField from './textareaField';
+import TextField from './textField';
+
+export default storyBook(Form, story => {
```
ryan953, line 24:
```suggestion
export default storyBook('Form Fields', story => {
```

ryan953 (approved):

PR #75937, opened 2024-08-09
chore(seer grouping): Log when seer event creates a group

lobsterkatie:
As part of debugging https://github.com/getsentry/sentry/issues/75860 (wherein events get sent to Seer and then seemingly disappear before they can be saved in the DB), this adds logs to the beginning and end of `_create_group`, so we can see how far such disappearing events make it before vanishing. Volume-wise this shouldn't be an issue, as it only affects events which are both sent to Seer and for which Seer doesn't find a match, which ends up being < 1/sec.


jangjodi (approved):
lgtm! thanks for checking the volume that this would occur in

PR #75942, opened 2024-08-09
feat(insights): Database system selector in Queries module

0Calories:
Adds a new dropdown selector to the Queries module that is responsible for changing the view depending on the currently selected database system,

While the component itself is simple, there is a lot of logic happening behind the scenes that is difficult to follow, here it is broken down;

- Queries the current project's spans to finds all database systems being used, and populates the dropdown selector options with compatible systems, in descending order based on number of spans per system
  - defaults to choosing the first option in the list (the system with the most spans) 
- Component is disabled when only one database system is being used (consistent with our design philosophy to not hide information / components)
- The most recent option that was selected is saved to `localStorage`, so next time you view the page it will maintain your selection
- The currently selected system is saved in the URL's query params
  - Following a URL with a non-default system selected will maintain this selection, but will not override your selection that is saved to your localStorage

**TODO in followup PR**
- [ ] Update the view according to which system was selected (mongodb terminology and formatting will be different from SQL)

image


0Calories:
I'm going to open this PR as is for now since it is getting lengthy, it is hidden behind a feature flag so it will only be visible for internal orgs. In a followup PR I will add the functionality to change the view based on the system that was selected

mjq (approved):
Awesome. Thanks for the detailed explanation, made the logic easier to follow. Looks great!

PR #75944, opened 2024-08-09
fix(slack): only pass tags in metadata, not callback_id

cathteng:
Fixes SENTRY-3DEY

The `callback_id` was used as both the `callback_id` and `private_metadata`. If we pass tags in the `callback_id`, sometimes it gets too long (has max length of 255 chars). Separate out creating the `callback_id` and `private_metadata` strings and only put the tags in the metadata will fix this.

https://github.com/getsentry/sentry/blob/22e0b9819a9af4518e3c55d74b6bb87854e314da/src/sentry/integrations/slack/webhooks/action.py#L351-L352

(^ getting rid of this!)

### DIFF
--- a/src/sentry/integrations/slack/webhooks/action.py
+++ b/src/sentry/integrations/slack/webhooks/action.py
@@ -321,11 +321,9 @@ class SlackActionEndpoint(Endpoint):
         initial_option_text: str,
         initial_option_value: str,
         callback_id: str,
-        metadata: str | None = None,
+        metadata: str,
     ) -> View:
         formatted_options = self.build_format_options(options)
-        if not metadata:
-            metadata = callback_id
 
         return View(
             type="modal",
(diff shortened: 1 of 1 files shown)


src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -321,8 +321,11 @@ def build_modal_payload(
         initial_option_text: str,
         initial_option_value: str,
         callback_id: str,
+        metadata: str | None = None,
```
schew2381, line 324: Is this ever `None`?

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -379,19 +384,22 @@ def open_resolve_dialog(self, slack_request: SlackActionRequest, group: Group) -
...
         if slack_request.data.get("channel"):
-            callback_id["channel_id"] = slack_request.data["channel"]["id"]
-            callback_id["rule"] = slack_request.callback_data.get("rule")
-        callback_id = orjson.dumps(callback_id).decode()
+            callback_id_dict["channel_id"] = slack_request.data["channel"]["id"]
+            callback_id_dict["rule"] = slack_request.callback_data.get("rule")
+        callback_id = orjson.dumps(callback_id_dict).decode()
+
+        # only add tags to metadata
+        callback_id_dict["tags"] = list(slack_request.get_tags())
```
sentaur-athena, line 398: not: for code clarity I suggest we copy callback_id_dict in metadata dict and then add tags to it instead of appending them to callback_id_dict.

sentaur-athena (approved):
Lol! Third time is the charm.

Naveen-C-Ramachandrappa (approved):

PR #75946, opened 2024-08-09
ref(profiling): update profile reference type

JonasBa:
Updates continuous profiling links to support linking to continuous profiles and txn based profiles

### DIFF
--- a/static/app/components/profiling/flamegraph/flamegraphContextMenu.tsx
+++ b/static/app/components/profiling/flamegraph/flamegraphContextMenu.tsx
@@ -610,20 +610,22 @@ function ProfileIdsSubMenu(props: {
             <ProfilingContextMenuGroup>
               <ProfilingContextMenuHeading>{t('Profiles')}</ProfilingContextMenuHeading>
               {props.profileIds.map((profileId, i) => {
+                const projectSlug =
+                  typeof profileId !== 'string' && 'project_id' in profileId
+                    ? projectLookupTable[profileId.project_id]?.slug ?? props.projectSlug
+                    : props.projectSlug;
+
+                if (!projectSlug) {
+                  return null;
+                }
+
                 const to = generateProfileRouteFromProfileReference({
                   orgSlug: props.organizationSlug,
-                  projectSlug:
-                    typeof profileId !== 'string' && 'project_id' in profileId
-                      ? projectLookupTable[profileId.project_id]?.slug
-                      : undefined,
+                  projectSlug,
                   reference: profileId,
                   frameName: props.frameName,
                   framePackage: props.framePackage,
                 });
-
-                if (!to) {
-                  return null;
-                }
 
                 return (
                   <ProfilingContextMenuItemButton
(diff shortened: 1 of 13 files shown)


static/app/utils/profiling/profile/importProfile.tsx
```diff
@@ -245,7 +247,7 @@ export function importSchema(
     profiles: input.profiles.map(profile =>
       importSingleProfile(profile, frameIndex, {
         ...options,
-        profileIds: input.shared.profile_ids,
+        profileIds: input.shared.profile_ids ?? input.shared.profiles,
```
JonasBa, line 250: This will infer the reference from either of the two fields and store it on the frame obj under a single type

static/app/components/profiling/flamegraph/flamegraphContextMenu.tsx
```diff
@@ -613,15 +609,35 @@ function ProfileIdsSubMenu(props: {
...
               <ProfilingContextMenuHeading>{t('Profiles')}</ProfilingContextMenuHeading>
-              {props.profileIds.map(profileId => {
-                const to = generateFlamechartLink(profileId);
+              {props.profileIds.map((profileId, i) => {
+                const to = generateProfileRouteFromProfileReference({
+                  orgSlug: props.organizationSlug,
+                  projectSlug:
+                    typeof profileId !== 'string' && 'project_id' in profileId
+                      ? projectLookupTable[profileId.project_id]?.slug
+                      : undefined,
```
Zylphrex, line 618: I think you need to add a fallback here to `props.projectSlug` for the old flamegraph format.

static/app/utils/profiling/routes.tsx
```diff
@@ -200,3 +218,60 @@ export function generateProfileFlamechartRouteWithHighlightFrame({
...
+  projectSlug,
+  frameName,
+  framePackage,
+  reference,
+  query,
+}: {
+  frameName: string;
+  framePackage: string | undefined;
+  orgSlug: Organization['slug'];
+  projectSlug: Project['slug'] | undefined;
```
Zylphrex, line 233: Should this be required? That will force the caller to pass a project slug instead of silently failing

static/app/utils/profiling/routes.tsx
```diff
@@ -200,3 +218,60 @@ export function generateProfileFlamechartRouteWithHighlightFrame({
...
+}): LocationDescriptor | null {
+  if (!projectSlug) return null;
+
+  if (typeof reference === 'string') {
+    return generateProfileFlamechartRouteWithHighlightFrame({
+      orgSlug: orgSlug,
+      projectSlug: projectSlug,
+      profileId: reference,
+      frameName: frameName,
+      framePackage: framePackage,
```
Zylphrex, line 245: Is it a TODO to add this highlight frame to the continuous profiles?

static/app/components/profiling/flamegraph/flamegraphContextMenu.tsx
JonasBa: Let me add it to both places actually

static/app/utils/profiling/routes.tsx
JonasBa: No, we should just add it. If its available it should always link. I missed the reference

static/app/utils/profiling/routes.tsx
JonasBa: Yeah, lets add it. It makes the return easier to work with as we dont need the null anymore

Zylphrex (approved):

PR #75950, opened 2024-08-09
fix(form): Always render FormField label component

evanpurkhiser:
Without this we don't actually provide a proper accessibility label


JonasBa (approved):

priscilawebdev (approved):
👏

PR #75951, opened 2024-08-09
feat(uptime): Add uptime edit form

evanpurkhiser:
image

Will follow up with tests. Need to fix something with forms to test properly.


scttcper (approved):
good form

PR #75955, opened 2024-08-09
feat(alerts): Backend for default metric alerts

schew2381:
Backend for default metric alerts. We do this by having the `project_created` djagno signal also create a metric alert.
Note that I held off on writing tests for now.

### Demo in Action
https://github.com/user-attachments/assets/1ef47275-e767-479e-bf18-7915ce4120c3

Closes [ALRT-207](https://getsentry.atlassian.net/browse/ALRT-207)

[ALRT-207]: https://getsentry.atlassian.net/browse/ALRT-207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/sentry/incidents/endpoints/organization_alert_rule_index.py
```diff
@@ -59,9 +63,79 @@
     ProjectUptimeSubscriptionMode,
     UptimeStatus,
 )
+from sentry.users.models.user import User
 from sentry.utils.cursors import Cursor, StringCursor
 
 
+def create_metric_alert(
```
schew2381, line 70: Need to change this from class method to standalone helper method so we can re-use in the project signal

tests/sentry/receivers/test_rules.py
```diff
@@ -0,0 +1,20 @@
+from sentry.analytics.events import project_created
+from tests.sentry.auth.test_access import AccessFactoryTestCase
+
+
+class DefaultRulesTest(AccessFactoryTestCase):
+    def test_include_all_projects_enabled(self):
```
ceorourke, line 6: What is this test doing?

schew2381: I still need to write tests, it's just a placeholder for now

ceorourke (approved):
LGTM besides the lack of tests but I trust you will add those

src/sentry/receivers/onboarding.py
```diff
@@ -63,7 +63,7 @@
 def record_new_project(project, user=None, user_id=None, **kwargs):
     if user_id is not None:
         default_user_id = user_id
-    elif user.is_authenticated:
+    elif user and user.is_authenticated:
```
schew2381, line 66: Small bug fix, if `user` and `user_id` are both None then this will error. We need to check if `user` exists first

PR #75963, opened 2024-08-12
ref(metrics): Make input fields same size

### LINKED ISSUE #75424: Metric input width
Image

We should aim for having as much consistent metric name input width as possible so that all other inputs are not shifted. In this scenario, it's extra strange as the length (and name) of all three metrics is the same.

priscilawebdev:
**Problem**

Adding one or more metric or equation may cause the inputs to vary in width, which could result in other inputs shifting and potentially appearing misaligned. For example:

image

**Solution**

To keep the "visualization", 'agg by' and 'group by' fields aligned, we have two options:

1. **Perform Layout Calculations:** Every time a query is added or changed, we would need to adjust the maximum width of each field to match others in different queries. This can be expensive and complex.
2. **Use CSS Grid:** This method creates a table-like layout and is more efficient and simple.

A downside of using CSS Grid is that the "Filter By" field does not automatically move below the other fields if it overflows, unlike with display: flex; flex-wrap: wrap.

With the changes within this PR, the "Filter By" field is positioned below the other fields, but on `xxlarge` screens, it remains in the same row. If the search input overflows, the field's height increases, which might not look ideal. However, this is likely a minor issue since long search queries are unlikely.

image

**Previews**

https://github.com/user-attachments/assets/facc8eb4-a6a2-433f-88ce-052525e347d0

closes https://github.com/getsentry/sentry/issues/75424


static/app/views/metrics/queries.tsx
```diff
@@ -371,7 +414,8 @@ const Row = styled('div')`
 const ButtonBar = styled('div')<{addQuerySymbolSpacing: boolean}>`
   align-items: center;
   display: flex;
-  padding-bottom: ${space(2)};
+  flex-wrap: wrap;
+  padding: ${space(2)} 0;
```
priscilawebdev, line 418: **Before**

image

**After**

image

static/app/components/metrics/equationInput/index.tsx
```diff
@@ -171,8 +172,27 @@ export function EquationInput({
...
+
+      @media (min-width: ${p.theme.breakpoints.large}) {
+        grid-column-end: 5;
+      }
+
+      @media (min-width: ${p.theme.breakpoints.xxlarge}) {
+        grid-column-end: 6;
+      }
+    `}
 `;
```
ArthurKnaus, line 196: This wrapper could be `QueryBuilder`, so `EquationInput` stays re-usable.

static/app/components/metrics/queryFieldGroup.tsx
```diff
@@ -153,6 +158,9 @@ const FieldGroup = styled('div')`
 `;
 
 const Label = styled('span')`
+  width: 95px;
+  min-width: 95px;
+  white-space: nowrap;
```
ArthurKnaus, line 163: Same thing here. I would defined the overrides for label in `QueryBuilder`.

ArthurKnaus (approved):
I like the new behaviour!

static/app/components/metrics/queryBuilder.tsx
```diff
@@ -345,14 +353,17 @@ export const QueryBuilder = memo(function QueryBuilder({
           >
             {selectedMeta && isVirtualMetric(selectedMeta) ? (
               <QueryFieldGroup>
-                <Label>{t('Where')}</Label>
+                <QueryFieldGroup.Label css={labelCss({theme})}>
```
ArthurKnaus, line 356: I wonder if calling this creates a new style sheet & `className` every render 🤔 
`labelCss({theme})`

ArthurKnaus (approved):

static/app/components/metrics/queryBuilder.tsx
priscilawebdev: I will improve it in a follow-up

Issue #71733, opened 2024-05-29
Unreal Engine onboarding wizard has wrong value for `DataRouterUrl`


bruno-garcia:
On a new project or `/projects/{name}/getting-started/` we show the install instructions.

At the top we talk about the [Unreal SDK](https://github.com/getsentry/sentry-unreal), with the DSN. Which is correct.

But at the bottom, we show the DataRouterUrl, for the [built-in crash reporter](https://docs.sentry.io/platforms/unreal/configuration/setup-crashreporter/). And there, it should be the specific Unreal endpoint instead of the DSN:

![image](https://github.com/getsentry/sentry/assets/1633368/9306604f-aa27-4d21-8055-13898d5b4d51)

Should be this URL instead:

![image](https://github.com/getsentry/sentry/assets/1633368/5bab12ba-3ac7-4ecf-a66e-c25ad289305b)

[Came up on Discord](https://discord.com/channels/621778831602221064/1245397411812348004/1245397411812348004), thanks @JasonDeArte

bruno-garcia:
cc @stephanie-anderson @bitsandfoxes

PR #75969, opened 2024-08-12
ref(onboarding): Expose DSN object

ArthurKnaus:
Expose the whole `dsn` object in the doc parameters, so all the entries can be accessed if necessary.

Part of https://github.com/getsentry/sentry/issues/71733


priscilawebdev (approved):
lgtm

PR #75971, opened 2024-08-12
chore(hybridcloud) Remove import shims for outbox model

markstory:
With getsentry updated we don't need these import shims anymore.

Requires https://github.com/getsentry/getsentry/pull/14877


GabeVillalobos (approved):
LGTM!

PR #75972, opened 2024-08-12
chore(derived_code_mappings): Split errors into different groups

armenzg:
This will allow the creation of separate groups based on the error, thus, making it easier to triage and provide fixes.

Currently, a lot of issues are being coalesced into [the same error](https://sentry.sentry.io/issues/4705496067/), thus, making it harder to investigate.

It currently requires scrolling through the breadcrumbs to find the root issue:
image

### DIFF
--- a/src/sentry/integrations/github/client.py
+++ b/src/sentry/integrations/github/client.py
@@ -386,41 +386,43 @@ class GitHubBaseClient(GithubProxyClient):
         """
         msg = "Continuing execution."
         should_count_error = False
-        txt = error.text
+        error_message = error.text
         if error.json:
             json_data: Any = error.json
-            txt = json_data.get("message")
+            error_message = json_data.get("message")
 
         # TODO: Add condition for  getsentry/DataForThePeople
         # e.g. getsentry/nextjs-sentry-example
-        if txt == "Git Repository is empty.":
+        if error_message == "Git Repository is empty.":
             logger.warning("The repository is empty. %s", msg, extra=extra)
-        elif txt == "Not Found":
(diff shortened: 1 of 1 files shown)


src/sentry/integrations/github/client.py
```diff
@@ -386,41 +386,43 @@ def _populate_trees_process_error(self, error: ApiError, extra: dict[str, str])
         """
         msg = "Continuing execution."
         should_count_error = False
-        txt = error.text
+        error_message = error.text
```
armenzg, line 389: Minor renaming to make it more clear as to the purpose of it.

src/sentry/integrations/github/client.py
```diff
@@ -386,41 +386,43 @@ def _populate_trees_process_error(self, error: ApiError, extra: dict[str, str])
...
+        ):
             logger.warning("Github has blocked this org. We will not continue.", extra=extra)
-            # Raising the error will be handled at the task level
+            # Raising the error will about the task and be handled at the task level
             raise error
         else:
             # We do not raise the exception so we can keep iterating through the repos.
             # Nevertheless, investigate the error to determine if we should abort the processing
-            logger.error("Investigate if to raise error. An error happened. %s", msg, extra=extra)
+            logger.error("Continuing execution. Investigate: %s", error_message, extra=extra)
```
armenzg, line 425: Using the `error_message` will hopefully split the grouping and if not at least make it easier to triage in Discover via the `title`.

The current value is not useful and groups all errors together:
image

jangjodi (approved):

PR #75976, opened 2024-08-12
fix(menulistitem): Add scroll-margin styles to menu list item

malwilley:
Prevents dropdown options from hitting the edge of the menu when navigating with the keyboard:

Before:

CleanShot 2024-08-12 at 08 47 34

After:

CleanShot 2024-08-12 at 08 47 13


leeandher (approved):
it's the little things 🔥

PR #75977, opened 2024-08-12
fix(replays): Trim clicked element whitespace backend

c298lee:
Fixes case where we could have extra whitespace in classes, resulting in extra `.`
image

### DIFF
--- a/src/sentry/replays/usecases/ingest/issue_creation.py
+++ b/src/sentry/replays/usecases/ingest/issue_creation.py
@@ -161,7 +161,7 @@ def _make_clicked_element(node):
             if key == "id":
                 element += f"#{value}"
             elif key == "class":
-                element = element + ".".join(value.split(" "))
+                element = element + ".".join(value.strip().split(" "))
             elif key == "role":
                 element += f'[role="{value}"]'
             elif key == "alt":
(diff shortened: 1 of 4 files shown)


aliu39 (approved):

src/sentry/replays/usecases/ingest/issue_creation.py
```diff
@@ -161,7 +161,7 @@ def _make_clicked_element(node):
             if key == "id":
                 element += f"#{value}"
             elif key == "class":
-                element = element + "." + ".".join(value.split(" "))
+                element = element + "." + ".".join(value.strip().split(" "))
```
aliu39, line 164: Maybe we could strip `value` in all cases?

michellewzhang (commented):
could we add a test?

src/sentry/replays/usecases/ingest/issue_creation.py
```diff
@@ -176,3 +177,38 @@ def _make_clicked_element(node):
                 element += f'[data-sentry-component="{value}"]'
 
     return element
+
+
+def test_make_clicked_element():
```
michellewzhang, line 182: can we move this to a separate test file?

michellewzhang (approved):

PR #75978, opened 2024-08-12
chore(onboarding): add view param to modal render analytics

ameliahsu:
adding a `view` param to the `messaging_integration_modal_rendered` analytics for when the modal is added to other product surfaces (ex. project creation)


static/app/views/alerts/rules/issue/setupMessagingIntegrationButton.tsx
```diff
@@ -22,9 +22,16 @@ interface ProjectWithAlertIntegrationInfo extends Project {
 type Props = {
   projectSlug: string;
   refetchConfigs: () => void;
+  analyticsParams?: {
+    view: string;
```
saponifi3d, line 26: nit: better to make this an enum rather than a string, so you can say "this will always be a string with this value".

iamrajjoshi (approved):
🚢

PR #75980, opened 2024-08-12
[WIP] Draft `grouphashmetadata` table

lobsterkatie:
NOTE: This is an overview of what the eventual state of the table might look like, but is not meant to be merged. It's just a place to have conversation about the overall plan.

----------

My first stab at creating the new table. General feedback welcome, but also, I have a two general questions:

- This is all of the fields that - as of now - are likely to be added in this first go-round. I'm sure there'll be changes, though. Do you think it's better to create the table with all the fields and possibly write migrations to change fields, or just add the fields as I add code related to them?

- Are there any bits I'm missing, things besides just the fields?


src/sentry/models/grouphashmetadata.py
```diff
@@ -0,0 +1,64 @@
...
+    class GroupingMethod(models.TextChoices):
+        HASH = "hash"
+        SEER = "seer"
+
+    class HashBasis(models.TextChoices):
+        MESSAGE = "msg"
+        FULL_STACKTRACE = "stack"
+        APP_STACKTRACE = "app_stack"
+        FINGERPRINT = "fingerprint"
+        # TODO: Are there others?
```
armenzg, line 23: Custom fingerprint.

Today I've found [this](https://sentry.sentry.io/issues/4956614130/?project=1&query=issue.category%3A%5Bperformance%2Ccron%2Creplay%5D&referrer=issue-stream&statsPeriod=7d&stream_index=0/) new way of grouping
image

src/sentry/models/grouphashmetadata.py
```diff
@@ -0,0 +1,64 @@
...
+    # GENERAL
+    grouphash = FlexibleForeignKey("sentry.GroupHash", unique=True, related_name="metadata")
+    date_added = models.DateTimeField(default=timezone.now)
+    grouping_method = models.PositiveIntegerField(choices=GroupingMethod)
+
+    # HASHING
+    # Most recent config to produce this hash
+    # TODO: do we need first, also?
+    # TODO: Is there a way to get choices for this from grouping.strategies.configurations.CONFIGURATIONS?
+    grouping_config = models.CharField()
```
armenzg, line 34: Beware: I'm handwaving here!

Maybe something like this:
```
from foo import CONFIGURATIONS

class ProjectConfig(models.TextChoices):
        CONFIG_0 = CONFIGURATIONS[0]
        CONFIG_1 = CONFIGURATIONS[1]
        ...
        CONFIG_N = CONFIGURATIONS[N]
```

Also, we should add an assertion for the size of it somewhere.

src/sentry/models/grouphashmetadata.py
```diff
@@ -0,0 +1,64 @@
...
+from django.db import models
+from django.utils import timezone
+
+from sentry.backup.scopes import RelocationScope
+from sentry.db.models import FlexibleForeignKey, JSONField, Model, region_silo_model
+
+
+@region_silo_model
+class GroupHashMetadata(Model):
+    __relocation_scope__ = RelocationScope.Excluded
```
armenzg, line 12: I do not understand the implications of using this.

src/sentry/models/grouphashmetadata.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # HASHING
+    # Most recent config to produce this hash
+    # TODO: do we need first, also?
+    # TODO: Is there a way to get choices for this from grouping.strategies.configurations.CONFIGURATIONS?
+    grouping_config = models.CharField()
+    # Eg. message, in-app stacktrace, full stacktrace, fingerprint, etc
+    hash_basis = models.CharField(choices=HashBasis, null=True)
+    # Eg. if message parameterized, what type of parameterization?, etc)
+    hashing_metadata: models.Field[dict[str, Any] | None, dict[str, Any] | None] = JSONField(
```
armenzg, line 38: Do you have an example to help me understand?

src/sentry/models/grouphashmetadata.py
```diff
@@ -0,0 +1,64 @@
...
+        null=True
+    )
+    # Built-in and custom stacktrace rules used
+    enhancements: models.Field[dict[str, Any] | None, dict[str, Any] | None] = JSONField(null=True)
+    # Client fingerprint, value, matched rule, rule source
```
armenzg, line 43: Do you have examples of each property you are referring to?

src/sentry/models/grouphashmetadata.py
```diff
@@ -0,0 +1,64 @@
...
+    fingerprint: models.Field[dict[str, Any] | None, dict[str, Any] | None] = JSONField(null=True)
+    # If this hash was associated to its group via a grouping config transition, what secondary hash
+    # made the link?
```
armenzg, line 46: I'm having trouble picturing when this would happen.

src/sentry/models/grouphashmetadata.py
```diff
@@ -0,0 +1,64 @@
...
+    secondary_hash_match = FlexibleForeignKey(
+        "sentry.GroupHash", null=True, related_name="secondary_matches"
+    )
+
+    # SEER
+    seer_model = models.CharField(null=True)
```
armenzg, line 52: What could the values be?

src/sentry/models/grouphashmetadata.py
```diff
@@ -0,0 +1,64 @@
...
+    # This will be different than `date_added` if we send it to Seer as part of a backfill
+    seer_date_sent = models.DateTimeField(null=True)
+    # TODO Figure out how to best reuse the types in seer.similarity.types
+    seer_results: models.Field[list[Any] | None, list[Any] | None] = JSONField(
```
armenzg, line 56: What can the seer results be?

armenzg (commented):
This is pretty cool to see. Good start 👍🏻

src/sentry/models/grouphashmetadata.py
lobsterkatie: Yeah, I actually think I need to go through all of the strategies, variants, and components to figure out which of them can turn into an over-arching reason for grouping. Will definitely take some more research, but I'll cross that bridge when I come to it.

src/sentry/models/grouphashmetadata.py
lobsterkatie: That makes sense. I think the big issue here is going to be circular dependencies - lots of the grouping modules import from `models`, and in this model I'd have to be importing from one of those modules. Even if it technically works right now (because whatever I need doesn't happen to come from modules which import from `models`), I'd be nervous to rely on that continuing to the the case. So move it to a neutral location? But then it's populated by grouping code...

src/sentry/models/grouphashmetadata.py
lobsterkatie: This is part of that backup/relocation project Alex has been working on. When you migrate an org from one region to another, or from on-prem to SAAS, not all of your data comes along for the ride, and in particular none of your event/group data does.

src/sentry/models/grouphashmetadata.py
lobsterkatie: There are examples of all of these in [the Notion doc](https://www.notion.so/sentry/GroupHash-metadata-a5a41b2a7e36401182a8eaa76f66c192?pvs=4), but one possibility for this field might be:

```json
{
  "message_parametrized": true,
  "parameterization_types": ["int", "date"]
}
```

src/sentry/models/grouphashmetadata.py
lobsterkatie: Yup - in the Notion doc linked in the above comment.

src/sentry/models/grouphashmetadata.py
lobsterkatie: In a config transition, if the secondary config calculates hash A and the new config calculates hash B, we use the group associated with hash A for hash B as well. This field would exist on hash B's metadata, and would point to hash A.

src/sentry/models/grouphashmetadata.py
lobsterkatie: Right now we're going with `vX`-type versioning:

https://github.com/getsentry/sentry/blob/3c02b743625293ae5b9d94b3390bda961856ae76/src/sentry/conf/server.py#L3443

src/sentry/models/grouphashmetadata.py
lobsterkatie: It'll be a list of dict versions of `SeerSimilarIssueData` instances. For now there'll only ever be one item in the list, since we're only planning to store the single best match (if any), but here and elsewhere in the code it's a list just in case that changes in the future.

https://github.com/getsentry/sentry/blob/3c02b743625293ae5b9d94b3390bda961856ae76/src/sentry/seer/similarity/types.py#L51-L56

PR #75986, opened 2024-08-12
ref(ui): Simplify form FieldGroup

evanpurkhiser:
Remove the option of using a render function as the child of a FieldGroup. There is no need to for this as the only time we use this is in the FormField itself where we use it to render the same ControlWrapper (renamed from FieldControl) but with a different controlState component.

The FieldGroup already had types that allowed for the controlState to be overridden in the FieldGroup.

Now the FormField just passes it's custom FormFieldControlState to the FieldGroup's controlState prop instead of rendering the whole ControlWrapper with the custom control statet pr

### DIFF
--- a/static/app/components/forms/fieldGroup/index.tsx
+++ b/static/app/components/forms/fieldGroup/index.tsx
@@ -1,7 +1,7 @@
 import QuestionTooltip from 'sentry/components/questionTooltip';
 
 import ControlState from './controlState';
-import FieldControl from './fieldControl';
+import ControlWrapper from './controlWrapper';
 import FieldDescription from './fieldDescription';
 import FieldHelp from './fieldHelp';
 import FieldLabel from './fieldLabel';
@@ -33,13 +33,13 @@ function FieldGroup({
   const {
     alignRight,
     children,
-    controlClassName,
-    disabledReason,
     error,
     flexibleControlStateSize,
     help,
     hideLabel,
+    hideControlState,
     highlighted,
+    controlState,
     id,
     isSaved,
     isSaving,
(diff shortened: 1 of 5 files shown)


static/app/components/forms/fieldGroup/index.tsx
```diff
@@ -58,28 +58,13 @@ function FieldGroup({
     return null;
   }
 
+  const controlStateElement = controlState ?? (
```
JonasBa, line 61: Drive-by PSA that we should start inlining this. It's annoying to read the render functions of components by having to jump to these definitions (especially as components get larger)

vuluongj20 (approved):
This is a good update! In general we can be more opinionated and not accommodate unnecessary customizations like the one you're removing here — I don't see why we would ever want to customize `ControlState`.

static/app/components/forms/fieldGroup/types.tsx
```diff
@@ -115,20 +110,5 @@ export interface FieldGroupProps {
 }
 
 /**
- * The children render props mostly pass down FieldGroupProps, with some slight
- * differences for properities that were resolved.
+ * Proops provided to the FieldGroupProps['controlWrapper']
```
priscilawebdev, line 113:
```suggestion
 * Props provided to the FieldGroupProps['controlWrapper']
```

static/app/components/forms/fieldGroup/index.tsx
priscilawebdev: I agree with @JonasBa  it would be better if this were inline

PR #75988, opened 2024-08-12
chore(custom-views): Make id required in group serach view response type

MichaelSun48:
This PR makes a minor update to the frontend functions that call the groupsearchviews endpoints. Specifically - it makes the id parameter required in the _response_ type of both the GET and PUT, but leaves it optional in the _request_ type. This did not cause any issues before, but is more correct and will help with the existing custom views work.

### DIFF
--- a/static/app/views/issueList/queries/useFetchGroupSearchViews.tsx
+++ b/static/app/views/issueList/queries/useFetchGroupSearchViews.tsx
@@ -1,12 +1,10 @@
 import type {UseApiQueryOptions} from 'sentry/utils/queryClient';
 import {useApiQuery} from 'sentry/utils/queryClient';
-import type {GroupSearchView} from 'sentry/views/issueList/types';
+import type {GroupSearchViewResponse} from 'sentry/views/issueList/types';
 
 type FetchGroupSearchViewsParameters = {
   orgSlug: string;
 };
-
-type FetchGroupSearchViewsResponse = GroupSearchView[];
 
 export const makeFetchGroupSearchViewsKey = ({
   orgSlug,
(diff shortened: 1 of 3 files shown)


static/app/views/issueList/types.tsx
```diff
@@ -21,3 +21,11 @@ export type GroupSearchView = {
...
   id?: string;
 };
+
+// Id is always returned by the response
+export type GroupSearchViewResponse = {
+  id: string;
+  name: string;
+  query: string;
+  querySort: IssueSortOptions;
+};
```
scttcper, line 31: we tend to not put "Response" in our type names, even though i definitely do on my personal projects.

scttcper: oh i see you've got two types and trying to move them over?

scttcper: maybe do `Required<GroupSearchView>`

MichaelSun48: That is exactly what i was looking for but couldn't find, thanks!

static/app/views/issueList/queries/useFetchGroupSearchViews.tsx
```diff
@@ -6,22 +6,17 @@ type FetchGroupSearchViewsParameters = {
...
+  options: Partial<UseApiQueryOptions<GroupSearchView[]>> = {}
 ) => {
-  return useApiQuery<FetchGroupSearchViewsResponse>(
-    makeFetchGroupSearchViewsKey({orgSlug}),
-    {
-      staleTime: Infinity,
-      ...options,
-    }
-  );
+  return useApiQuery<GroupSearchView[]>(makeFetchGroupSearchViewsKey({orgSlug}), {
```
MichaelSun48, line 18: I left the useApiQuery cache type as just GroupSearchView[] instead of Required<GroupSearchView[]>, even though that cache should always have an id. Feels cleaner to make it the base interface, especially since they're compatible.

malwilley: Hmm I think that I would rather be more explicit and define two separate types. `GroupSearchView` should have a required ID and perfectly match what is returned from this endpoint. The update endpoint is the only thing that doesn't require an ID, so I'd make a new type in that file like:

```
interface UpdateGroupSearchViewPayload extends Omit<GroupSearchView, 'id'> {
  id?: string;
}
```

malwilley (approved):

PR #75991, opened 2024-08-12
ref(alerts): Add anomaly detection metrics

ceorourke:
Update the metric alert rule fired and resolved metric to have the action in the name and the tag is now the detection type. There don't seem to be any other dashboards using this metric (besides the one I just created). 

Closes https://getsentry.atlassian.net/browse/ALRT-148

### DIFF
--- a/src/sentry/incidents/subscription_processor.py
+++ b/src/sentry/incidents/subscription_processor.py
@@ -542,10 +542,8 @@ class SubscriptionProcessor:
         with transaction.atomic(router.db_for_write(AlertRule)):
             # Triggers is the threshold - NOT an instance of a trigger
             for trigger in self.triggers:
-                if (
-                    self.has_anomaly_detection
-                    and trigger.alert_rule.detection_type == AlertRuleDetectionType.DYNAMIC
-                ):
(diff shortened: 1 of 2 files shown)


schew2381 (approved):
lgtm for the new tag, but was wondering if the original alert/resolve tag was useful for us at all to calculate how many alerts are triggering vs resolving?

src/sentry/incidents/subscription_processor.py
```diff
@@ -562,7 +560,10 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
                         if self.has_anomaly(
                             potential_anomaly, trigger.label
                         ) and not self.check_trigger_matches_status(trigger, TriggerStatus.ACTIVE):
-                            metrics.incr("incidents.alert_rules.threshold", tags={"type": "alert"})
+                            metrics.incr(
+                                "incidents.alert_rules.threshold.alert",
+                                tags={"type": detection_type},
```
schew2381, line 565: nit: might be good to specify the type so it's more clear in DD
```suggestion
                                tags={"detection_type": detection_type},
```

PR #75992, opened 2024-08-12
feat(dashboards): Preserve query between errors and transactions

narsaynorath:
Since we aren't guaranteed that all discover splits will be correct, we need to allow the user to switch between the errors and transactions datasets without having to set the entire widget back up.

This PR persists the settings and injects the field options because they may not be allowed in the other dataset (e.g. p99 in transactions -> errors).

This PR does _not_ handle indicating incompatible functions to the users, nor does it handle gracefully erroring when an incompatible function is selected.


shruthilayaj (approved):
lgtm

PR #68396, opened 2024-04-06
fix(replay): Let the Issue>Replay and Transaction>Replay tab counts look back 90d instead of 14

ryan953:
SCR-20240406-jkia

Fixes https://github.com/getsentry/sentry/issues/68057

PR #75993, opened 2024-08-12
fix(replay): replay queries should respect location statsPeriod

### LINKED ISSUE #75973: Replay: Tab in transaction overview changes date selector to 90 days
Sentry Feedback: [JAVASCRIPT-2V69](https://sentry.sentry.io/feedback/?referrer=github_integration&feedbackSlug=javascript%3A5688704104&project=11276)

When switching to replays tab, my date selector automatically gets set to 90 days and stays that way after switching tabs

The Replays tab in Transaction Summary manually forces the date range to be 90d without updating the URL. When the other tabs are clicked after Replays load, they all have the 90d duration.

michellewzhang:
fixes https://github.com/getsentry/sentry/issues/75973

### before: 24h -> replay tab -> 90d

https://github.com/user-attachments/assets/5f430390-2bd1-4e52-aebf-f26f9152c747

### after: 1h -> replay tab -> still 1h

https://github.com/user-attachments/assets/b0e35acb-9760-4b0d-8d60-ba63f87c139f

also fixed an error where the issue details replay tab query wasn't respecting the location `statsPeriod` either.

### before: tab says `6` replays but there are way more than 6 replays:
SCR-20240812-kpbj

SCR-20240812-kovl

### after: tab says `17` replays which is the correct amount:
SCR-20240812-kozx
SCR-20240812-koxv

also fixed a bug where the transactions summary tab count didn't match the replay list in the tab. fetching the replay list didn't respect the location query; it seems to always default to 14d. the default `eventView` was hardcoding 14d:

a14048c7-3193-4d2d-8db3-2831378316f0

https://github.com/getsentry/sentry/blob/d3c2b01801a5c28be02caaca75677cf658e617ab/static/app/views/performance/transactionSummary/transactionReplays/useReplaysFromTransaction.tsx#L64-L69

### before: 
de461924-406a-4485-bd4c-b718ca3f216a

### after:
SCR-20240812-lfmn

### DIFF
--- a/static/app/utils/replayCount/useReplayCountForIssues.tsx
+++ b/static/app/utils/replayCount/useReplayCountForIssues.tsx
@@ -1,7 +1,9 @@
 import {useMemo} from 'react';
 
 import {IssueCategory} from 'sentry/types/group';
+import {decodeScalar} from 'sentry/utils/queryString';
 import useReplayCount from 'sentry/utils/replayCount/useReplayCount';
+import {useLocation} from 'sentry/utils/useLocation';
 import useOrganization from 'sentry/utils/useOrganization';
 
 interface Props {
@@ -14,9 +16,10 @@ interface Props {
  */
 export default function useReplayCountForIssues({
   bufferLimit = 25,
-  statsPeriod = '14d',
+  statsPeriod,
 }: Props = {}) {
   const organization = useOrganization();
+  const location = useLocation();
   const {
     getOne: getOneError,
     getMany: getManyError,
@@ -27,7 +30,7 @@ export default function useReplayCountForIssues({
     dataSource: 'discover',
     fieldName: 'issue.id',
     organization,
-    statsPeriod,
+    statsPeriod: statsPeriod ?? decodeScalar(location.query.statsPeriod) ?? '90d',
   });
   const {
     getOne: getOneIssue,
@@ -39,7 +42,7 @@ export default function useReplayCountForIssues({
     dataSource: 'search_issues',
     fieldName: 'issue.id',
     organization,
-    statsPeriod,
+    statsPeriod: statsPeriod ?? decodeScalar(location.query.statsPeriod) ?? '90d',
   });
 
   return useMemo(
--- a/static/app/utils/replayCount/useReplayCountForIssues.tsx
+++ b/static/app/utils/replayCount/useReplayCountForIssues.tsx
@@ -1,9 +1,7 @@
 import {useMemo} from 'react';
 
 import {IssueCategory} from 'sentry/types/group';
-import {decodeScalar} from 'sentry/utils/queryString';
 import useReplayCount from 'sentry/utils/replayCount/useReplayCount';
-import {useLocation} from 'sentry/utils/useLocation';
 import useOrganization from 'sentry/utils/useOrganization';
 
 interface Props {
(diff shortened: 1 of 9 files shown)


static/app/views/issueDetails/groupReplays/useReplaysFromIssue.tsx
```diff
@@ -50,7 +50,7 @@ export default function useReplaysFromIssue({
       Sentry.captureException(error);
       setFetchError(error);
     }
-  }, [api, organization.slug, group.id, dataSource, location.query.environment]);
+  }, [api, organization.slug, group.id, dataSource, location]);
```
c298lee, line 53:
```suggestion
  }, [api, organization.slug, group.id, dataSource, location.query]);
```

static/app/views/performance/transactionSummary/transactionReplays/transactionReplays.tsx
```diff
@@ -39,7 +39,6 @@ function TransactionReplays() {
         ...location,
```
c298lee, line 39: could this just be `location`?

static/app/utils/replayCount/useReplayCountForIssues.tsx
```diff
@@ -14,9 +16,10 @@ interface Props {
  */
 export default function useReplayCountForIssues({
   bufferLimit = 25,
-  statsPeriod = '14d',
```
michellewzhang, line 17: removed this bc of logic from this pr: https://github.com/getsentry/sentry/pull/68396

static/app/utils/replayCount/useReplayCountForTransactions.tsx
```diff
@@ -13,15 +15,16 @@ interface Props {
  */
 export default function useReplayCountForTransactions({
   bufferLimit = 25,
-  statsPeriod = '14d',
```
michellewzhang, line 16: same comment as above

static/app/utils/replayCount/useReplayCountForTransactions.tsx
```diff
@@ -13,15 +15,16 @@ interface Props {
...
 }: Props = {}) {
   const organization = useOrganization();
+  const location = useLocation();
   const {getOne, getMany, hasOne, hasMany} = useReplayCount({
     bufferLimit,
     dataSource: 'discover',
     fieldName: 'transaction',
     organization,
-    statsPeriod,
+    statsPeriod: statsPeriod ?? decodeScalar(location.query.statsPeriod) ?? '90d',
```
ryan953, line 27: The callsite is also doing `decodeScalar(location.query.statsPeriod)` so why does it need to be inside here as a default/fallback as well? i think we should keep this method simpler and straight forward like before.

michellewzhang: good point, i had this here before i realized the callsites weren't passing in `statsPeriod` at all. can remove this now

ryan953 (changes requested):
We should think this through a little more. Reading the query params without having a way to a) show the user the current value b) change it; seems like it'll lead to mysterious situations for people.

static/app/utils/replayCount/useReplayCountForIssues.tsx
ryan953: We probably don't need to change defaults, especially to longer values. if something doesn't like the default it should be overridden, or even better is to have no default at all and force call sites to put in what they want.

static/app/views/performance/transactionSummary/transactionReplays/transactionReplays.tsx
```diff
@@ -35,13 +36,7 @@ function TransactionReplays() {
 
   return (
     <PageLayout
-      location={{
-        ...location,
-        query: {
-          ...location.query,
-          statsPeriod: '90d',
-        },
-      }}
```
ryan953, line 44: This seems like the root problem from the ticket: we're overriding whatever was there before with 90d.

PR #75996, opened 2024-08-12
feat(query-builder): Add FormattedQuery component for rendering a formatted query 

malwilley:
Renders a query similar to how it appears in the searchbar, but without any interactivity. This will be used in the recent searches UI and perhaps in other locations as well.

CleanShot 2024-08-12 at 12 31 48

### DIFF
--- a/static/app/components/searchQueryBuilder/formattedQuery.tsx
+++ b/static/app/components/searchQueryBuilder/formattedQuery.tsx
@@ -25,7 +25,7 @@ type TokenProps = {
   token: ParseResultToken;
 };
 
-const EMPTY_TAGS_COLLECTION = {};
+const EMPTY_FILTER_KEYS: TagCollection = {};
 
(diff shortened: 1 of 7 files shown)


static/app/components/searchQueryBuilder/formattedQuery.tsx
```diff
@@ -0,0 +1,125 @@
...
+  query: string;
+  fieldDefinitionGetter?: FieldDefinitionGetter;
+  filterKeys?: TagCollection;
+};
+
+type TokenProps = {
+  token: ParseResultToken;
+};
+
+const EMPTY_TAGS_COLLECTION = {};
```
scttcper, line 28: needs a type otherwise is `{}`
```suggestion
const EMPTY_TAGS_COLLECTION: TagCollection = {};
```

scttcper (approved):

PR #75999, opened 2024-08-12
feat(buffers): Assign dedicated Celery queue to a model for RedisBuffers

dashed:
Fixes https://getsentry.atlassian.net/browse/RV-1753

The `process_incr` task and the default assigned queue for it is shared among multiple models (i.e. `counters-0` queue).

If any backlogs or slowdowns occur when incrementing counts for any specific model, other models will be affected. To alleviate this, we can assign a dedicated queue for any given model. If a dedicated queue is assigned, the `process_incr` task will be processed in the assigned queue. On the other hand, if no dedicated queue is assigned, the `process_incr` task will be processed in the default queue (i.e. `counters-0` queue).

The queue to be used for the `process_incr` task is determined in the following order of precedence:

1. The queue argument passed to `process_incr.apply_async()` (we pass in any dedicated Celery queue for a given Django model in the `queue` keyword argument in this step)
2. The queue defined on the `process_incr` task (set to `counters-0`)
3. Any defined routes in `CELERY_ROUTES` (currently un-used)

See: https://docs.celeryq.dev/en/latest/userguide/routing.html#specifying-task-destination

Hence, we override the default queue of the `process_incr` task by passing in any assigned queue to the `process_incr.apply_async()` call.

An example of how this will be used to assign a Celery queue to a Django model is in: https://github.com/getsentry/getsentry/pull/14897

### DIFF
--- a/src/sentry/buffer/redis.py
+++ b/src/sentry/buffer/redis.py
@@ -82,11 +82,16 @@ class BufferHookRegistry:
 redis_buffer_registry = BufferHookRegistry()
 
 
+# Callable to get the queue name for the given model_key.
+# May return None to not assign a queue for the given model_key.
+ChooseQueueFunction = Callable[[str], str | None]
+
+
 @dataclass
 class PendingBufferValue:
     model_key: str | None
     pending_buffer: PendingBuffer
-    queue: str | None
+    generate_queue: ChooseQueueFunction | None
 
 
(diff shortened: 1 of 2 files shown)


vbro (approved):
LGTM
The queue has to be ready before enabling for a model.
https://github.com/getsentry/getsentry/pull/14897

PR #76003, opened 2024-08-12
ref(similarity): Rollout similarity delete record to 20% EA

jangjodi:
Project option `sentry:similarity_backfill_completed` is `True` for all EA projects
Rollout delete record by hash feature to 20% of these projects

### DIFF
--- a/src/sentry/tasks/delete_seer_grouping_records.py
+++ b/src/sentry/tasks/delete_seer_grouping_records.py
@@ -1,4 +1,5 @@
 import logging
+from random import randint
 from typing import Any
 
 from sentry import features, options
@@ -13,6 +14,7 @@ from sentry.silo.base import SiloMode
 from sentry.tasks.base import instrumented_task
 
 logger = logging.getLogger(__name__)
+EA_ROLLOUT_PERCENTAGE = 20
 
 
 @instrumented_task(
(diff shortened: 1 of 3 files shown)


src/sentry/tasks/delete_seer_grouping_records.py
```diff
@@ -13,6 +14,7 @@
 from sentry.tasks.base import instrumented_task
 
 logger = logging.getLogger(__name__)
+EA_ROLLOUT_PERCENTAGE = 20
```
mrduncan, line 17: Can we use an option here?

jangjodi: good catch, thanks!

src/sentry/tasks/delete_seer_grouping_records.py
```diff
@@ -56,7 +57,13 @@ def call_delete_seer_grouping_records_by_hash(
         project = group.project if group else None
     if (
         project
-        and features.has("projects:similarity-embeddings-grouping", project)
+        and (
+            features.has("projects:similarity-embeddings-grouping", project)
+            or (
+                project.get_option("sentry:similarity_backfill_completed")
+                and randint(1, 100) <= options.get("similarity.delete_task_EA_rollout_percentage")
```
mrduncan, line 64: Just to confirm, we don't want any stickyness on the project/org and it's ok if it's a randomly distributed percentage

mrduncan (approved):

src/sentry/tasks/delete_seer_grouping_records.py
jangjodi: Correct, we just want to make sure our task and seer can handle the load slowly. I don't think this slower rollout is strictly necessary, but wanted to be cautious.

PR #76004, opened 2024-08-12
feat(uptime): Add uptime to issue.category predefined values

evanpurkhiser:


gaprl (approved):

malwilley (approved):
This will update it in the new searchbar which is in EA, but if you want to add it for all users you'll also need to update this here: https://github.com/getsentry/sentry/blob/17f34ca4dcf9d1ab59a169d7794edd9aca9ee79e/static/app/stores/tagStore.tsx#L121-L124

PR #76005, opened 2024-08-12
fix(replay/issues): check `issueTypeConfig` before showing onboarding panel

michellewzhang:
SCR-20240812-mcaj

over slack, we decided to hide the entire `EventReplay` section if `issueTypeConfig.replays.enabled` is false.

### DIFF
--- a/static/app/components/events/eventReplay/index.tsx
+++ b/static/app/components/events/eventReplay/index.tsx
@@ -6,6 +6,7 @@ import LazyLoad from 'sentry/components/lazyLoad';
 import type {Event} from 'sentry/types/event';
 import type {Group} from 'sentry/types/group';
 import useEventCanShowReplayUpsell from 'sentry/utils/event/useEventCanShowReplayUpsell';
+import {getConfigForIssueType} from 'sentry/utils/issueTypeConfig';
 import {getReplayIdFromEvent} from 'sentry/utils/replays/getReplayIdFromEvent';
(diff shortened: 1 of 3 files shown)


static/app/components/events/eventReplay/index.tsx
```diff
@@ -39,7 +41,7 @@ export default function EventReplay({event, group, projectSlug}: Props) {
     return <ReplayClipSection event={event} replayId={replayId} group={group} />;
   }
 
-  if (canShowUpsell && !hasSentOneReplay) {
+  if (canShowUpsell && !hasSentOneReplay && issueTypeConfig.replays.enabled) {
```
ryan953, line 44: why is `<EventReplay>` being included if `replays.enabled == false`?

michellewzhang: we're currently discussing in a slack thread whether this check should be here or not. the check seems mostly related to whether a certain tab should show up in issue details but this panel shows inline on the main event page

michellewzhang: tagged you in the thread

malwilley: I think what Ryan means is that we could add this check to `GroupEventDetailsContent` instead and skip rendering this component at all if the config returns false. I think I have a slight preference to that since it means no extra requests get sent off if we know we don't need to render it.

michellewzhang: ahh gotcha, i'll update the callsite

malwilley (approved):

PR #76008, opened 2024-08-12
fix(similarity): Use count instead of length in backfill query

jangjodi:
Use `.count()` instead of `len` in postgres query

### DIFF
--- a/src/sentry/tasks/embeddings_grouping/utils.py
+++ b/src/sentry/tasks/embeddings_grouping/utils.py
@@ -172,7 +172,7 @@ def get_current_batch_groups_from_postgres(
         group_id_filter, project.id, batch_size
     )
 
-    total_groups_to_backfill_length = len(groups_to_backfill_batch_raw)
+    total_groups_to_backfill_length = groups_to_backfill_batch_raw.count()
     batch_end_group_id = (
         groups_to_backfill_batch_raw[total_groups_to_backfill_length - 1][0]
         if total_groups_to_backfill_length
@@ -192,7 +192,7 @@ def get_current_batch_groups_from_postgres(
         "backfill_seer_grouping_records.batch",
         extra={
             "project_id": project.id,
-            "batch_len": len(groups_to_backfill_batch),
+            "batch_len": total_groups_to_backfill_length,
             "last_processed_group_id": batch_end_group_id,
         },
     )


roggenkemper (approved):

src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -172,7 +172,7 @@ def get_current_batch_groups_from_postgres(
         group_id_filter, project.id, batch_size
     )
 
-    total_groups_to_backfill_length = len(groups_to_backfill_batch_raw)
+    total_groups_to_backfill_length = groups_to_backfill_batch_raw.count()
     batch_end_group_id = (
         groups_to_backfill_batch_raw[total_groups_to_backfill_length - 1][0]
         if total_groups_to_backfill_length
```
wedamija, line 178: You're now making two queries here - one to get the count, and one to get the end group id for the batch, in addition to fetching the batch on 183.

Is there any reason we can't just perform this logic after 183?

Any reason we can't move this logic after we filter to `groups_to_backfill_batch`?

src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -172,7 +172,7 @@ def get_current_batch_groups_from_postgres(
         group_id_filter, project.id, batch_size
```
wedamija, line 172: `_make_postgres_call_with_retry` doesn't work properly afaict. This just constructs the queryset https://github.com/getsentry/sentry/blob/23768500359ae9535da4cd50ecb6c2a58f2d9ef6/src/sentry/tasks/embeddings_grouping/utils.py#L124-L133. But since we don't convert it to a list, or attempt to iterate over it, then it never runs the query.

That means you'll never enter your retry loop here: https://github.com/getsentry/sentry/blob/23768500359ae9535da4cd50ecb6c2a58f2d9ef6/src/sentry/tasks/embeddings_grouping/utils.py#L143-L149

You should coerce to a list in the return value. That also saves you the trouping of using `.count` here since you won't have a queryset.

jangjodi: Thanks for the review @wedamija! I've opened up a follow up [PR](https://github.com/getsentry/sentry/pull/76081) to address these

PR #76011, opened 2024-08-12
feat(anomaly detection): add new anomaly detection alert fields to create/update form

mifu67:
If the anomaly detection alert type is selected, new fields pop up for sensitivity and detection type. Also add the functionality for anomaly detection alert create/update via the frontend.

Fixes https://getsentry.atlassian.net/browse/ALRT-143
Create process:

https://github.com/user-attachments/assets/618787ef-a56a-4a49-9643-05cabb4563a8

Created rule:
Screenshot 2024-08-14 at 11 56 42 AM

### DIFF
--- a/static/app/views/alerts/rules/metric/ruleForm.tsx
+++ b/static/app/views/alerts/rules/metric/ruleForm.tsx
@@ -143,6 +143,8 @@ type State = {
   project: Project;
   query: string;
   resolveThreshold: UnsavedMetricRule['resolveThreshold'];
+  sensitivity: UnsavedMetricRule['sensitivity'];
+  // because it's not passed for now
   thresholdPeriod: UnsavedMetricRule['thresholdPeriod'];
   thresholdType: UnsavedMetricRule['thresholdType'];
   timeWindow: number;
@@ -152,6 +154,7 @@ type State = {
   comparisonDelta?: number;
   isExtrapolatedChartData?: boolean;
   monitorType?: MonitorType;
+  seasonality?: string;
 } & DeprecatedAsyncComponent['state'];
 
 const isEmpty = (str: unknown): boolean => str === '' || !defined(str);
@@ -456,6 +459,14 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
     changedTriggerIndex?: number
   ) {
     const {comparisonType} = this.state;
+    // If we have an anomaly detection alert, then we don't need to validate the thresholds, but we do need to set them to 0
+    if (comparisonType === AlertRuleComparisonType.DYNAMIC) {
+      triggers.forEach(trigger => {
+        trigger.alertThreshold = 0;
+      });
+      const triggerErrors = new Map();
+      return triggerErrors; // return an empty map
+    }
     const triggerErrors = new Map();
 
     const requiredFields = ['label', 'alertThreshold'];
@@ -727,6 +738,9 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
       eventTypes,
       monitorType,
       activationCondition,
+      sensitivity,
+      seasonality,
+      comparisonType,
     } = this.state;
     // Remove empty warning trigger
     const sanitizedTriggers = triggers.filter(
@@ -763,7 +777,12 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
             activationCondition,
           };
         }
-
+        const detectionTypes = new Map([
+          [AlertRuleComparisonType.COUNT, 'static'],
+          [AlertRuleComparisonType.CHANGE, 'percent'],
+          [AlertRuleComparisonType.DYNAMIC, 'dynamic'],
+        ]);
+        const detectionType = detectionTypes.get(comparisonType) ?? '';
         const dataset = this.determinePerformanceDataset();
         this.setState({loading: true});
         // Add or update is just the PUT/POST to the org alert-rules api
@@ -787,6 +806,9 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
             eventTypes: isCrashFreeAlert(rule.dataset) ? undefined : eventTypes,
             dataset,
             queryType: DatasetMEPAlertQueryTypes[dataset],
+            sensitivity: sensitivity ?? null,
+            seasonality: seasonality ?? null,
+            detectionType: detectionType,
           },
           {
             duplicateRule: this.isDuplicateRule ? 'true' : 'false',
@@ -889,13 +911,22 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
 
   handleComparisonTypeChange = (value: AlertRuleComparisonType) => {
     const comparisonDelta =
-      value === AlertRuleComparisonType.COUNT
-        ? undefined
-        : this.state.comparisonDelta ?? DEFAULT_CHANGE_COMP_DELTA;
+      value === AlertRuleComparisonType.CHANGE
+        ? this.state.comparisonDelta ?? DEFAULT_CHANGE_COMP_DELTA
+        : undefined;
     const timeWindow = this.state.comparisonDelta
       ? DEFAULT_COUNT_TIME_WINDOW
       : DEFAULT_CHANGE_TIME_WINDOW;
-    this.setState({comparisonType: value, comparisonDelta, timeWindow});
+    const sensitivity =
+      value === AlertRuleComparisonType.DYNAMIC ? this.state.sensitivity : undefined;
+    const seasonality = value === AlertRuleComparisonType.DYNAMIC ? 'auto' : undefined; // TODO: replace "auto" with the correct constant
+    this.setState({
+      comparisonType: value,
+      comparisonDelta,
+      timeWindow,
+      sensitivity,
+      seasonality,
+    });
   };
 
   handleDeleteRule = async () => {
--- a/static/app/views/alerts/rules/metric/ruleForm.tsx
+++ b/static/app/views/alerts/rules/metric/ruleForm.tsx
@@ -85,12 +85,13 @@ import {
   DEFAULT_COUNT_TIME_WINDOW,
 } from './constants';
 import RuleConditionsForm from './ruleConditionsForm';
-import type {
-  EventTypes,
-  MetricActionTemplate,
-  MetricRule,
-  Trigger,
-  UnsavedMetricRule,
+import {
+  AlertRuleSensitivity,
+  type EventTypes,
+  type MetricActionTemplate,
+  type MetricRule,
+  type Trigger,
+  type UnsavedMetricRule,
 } from './types';
 import {
   AlertRuleComparisonType,
@@ -233,6 +234,7 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
       metricExtractionRules: null,
       triggers: triggersClone,
       resolveThreshold: rule.resolveThreshold,
+      sensitivity: rule.sensitivity || AlertRuleSensitivity.MEDIUM,
       thresholdType: rule.thresholdType,
       thresholdPeriod: rule.thresholdPeriod ?? 1,
       comparisonDelta: rule.comparisonDelta ?? undefined,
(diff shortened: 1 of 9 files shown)


static/app/views/alerts/rules/metric/ruleForm.tsx
```diff
@@ -454,6 +459,21 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
...
+      // NOTE: we don't support warning triggers for anomaly detection alerts yet
+      // once we do, uncomment this code and delete the next five uncommented lines:
+      // triggers.forEach(trigger => {
+      //   trigger.alertThreshold = 0;
+      // });
+      const criticalTriggerIndex = triggers.findIndex(
+        ({label}) => label === AlertRuleTriggerType.CRITICAL
+      );
+      const criticalTrigger = triggers[criticalTriggerIndex];
+      criticalTrigger.alertThreshold = 0;
```
saponifi3d, line 473: Can you explain this bit of code a little? (It looks like we're taking the `triggers` from the state, then mutating the variable to trigger state changes in the component)

static/app/views/alerts/rules/metric/ruleForm.tsx
```diff
@@ -454,6 +459,21 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
...
+      const triggerErrors = new Map();
```
saponifi3d, line 474: nit: i'd move this to the top of the function and then you don't need to have it defined on 477 and 474, and can just return `triggerErrors`. I think that will also help the code read a little more cleanly as an "this is an overrde for the Dynamic alert types"

static/app/views/alerts/rules/metric/triggers/anomalyAlertsForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const handleSensitivityChange = ({value}) => {
+    onSensitivityChange(value);
+  };
+
+  return (
+    <StyledField
+      label={<div>{'Sensitivity'}</div>}
+      help={
+        <div>
+          {
```
saponifi3d, line 49: nit: could you just pass in the string or do you need the display block css setting from the div?

static/app/views/alerts/rules/metric/triggers/anomalyAlertsForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const handleThresholdTypeChange = ({value}) => {
+    onThresholdTypeChange(value);
+  };
+
+  return (
+    <StyledField
+      label={<div>{'Direction'}</div>}
+      help={
+        <div>
+          {
```
saponifi3d, line 94: re: nit.

static/app/views/alerts/rules/metric/triggers/anomalyAlertsForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+const StyledField = styled(FieldGroup)`
+  & > label > div:first-child > span {
+    display: flex;
+    flex-direction: row;
+  }
+`;
+
+const SelectContainer = styled('div')`
+  flex: 1;
```
saponifi3d, line 154: nit, there's a shared component for this in: `static/app/components/container/flex.tsx`. By reusing the shared components, we can reduce our javascript bundle size which really helps performance.

saponifi3d (commented):
mostly curious about the `validateTrigger` functions, i wonder if there's a way we can improve the logic here. we should also make sure any state changes are going through the react state system so updates are triggered correctly.

static/app/views/alerts/rules/metric/ruleForm.tsx
mifu67: Dynamic alerts don't have trigger thresholds, so we don't have any fields where the user can modify the threshold. On submitting a dynamic alert, the `alertThreshold` will be empty, its default state (because the user never entered a threshold value). Because the trigger needs to have an alert threshold, we set it to 0 here.

static/app/views/alerts/rules/metric/triggers/anomalyAlertsForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+      </SelectContainer>
+    </StyledField>
+  );
+}
+
+function DirectionFormItem({
+  thresholdType,
+  onThresholdTypeChange,
+}: DirectionFormItemProps) {
+  const handleThresholdTypeChange = ({value}) => {
```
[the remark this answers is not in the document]
nhsiehgit: Also - we can probably combine these two components into a single component and generalize the props

static/app/views/alerts/rules/metric/triggers/anomalyAlertsForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+}
+
+class AnomalyDetectionForm extends Component<Props> {
+  render() {
+    const {sensitivity, onSensitivityChange, thresholdType, onThresholdTypeChange} =
+      this.props;
+
+    return (
+      <Fragment>
+        <SensitivityFormItem
```
nhsiehgit, line 133: commented above

but we can probably just have a generic component for both of these and pass in the options

static/app/views/alerts/rules/metric/triggers/index.tsx
```diff
@@ -118,7 +124,13 @@ class Triggers extends Component<Props> {
         <Panel>
           <PanelBody>
             {comparisonType === AlertRuleComparisonType.DYNAMIC ? (
-              <div>{'This is where the anomaly detection alert field choices go'}</div>
+              <AnomalyDetectionForm
```
nhsiehgit, line 127: oic
the other components named `Form` as well 😭 
still think we can call this `FormField` though

nhsiehgit (commented):
comments!

static/app/views/alerts/rules/metric/triggers/anomalyAlertsForm.tsx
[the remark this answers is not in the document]
mifu67: Josh told me to make two components

static/app/views/alerts/rules/metric/ruleForm.tsx
```diff
@@ -819,7 +850,13 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
             ? err?.responseJSON
             : Object.values(err?.responseJSON)
           : [];
-        const apiErrors = errors.length > 0 ? `: ${errors.join(', ')}` : '';
+        let apiErrors = '';
+        if (typeof errors[0] === 'object') {
```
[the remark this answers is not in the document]
mifu67: an `object` type means that the first element of the error list is a map. This would work for all HTTP errors that have this format

static/app/views/alerts/rules/metric/triggers/anomalyAlertsForm.tsx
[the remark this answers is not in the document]
nhsiehgit: hmmmm

@saponifi3d 
Thoughts? If these 2 components are identical in structure with differing content - thoughts on making these a generic select form field component?

static/app/views/alerts/rules/metric/ruleForm.tsx
```diff
@@ -785,6 +813,9 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
             eventTypes: isCrashFreeAlert(rule.dataset) ? undefined : eventTypes,
             dataset,
             queryType: DatasetMEPAlertQueryTypes[dataset],
+            sensitivity: sensitivity ?? null,
```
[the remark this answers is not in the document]
mifu67: I think that if there's no sensitivity, the field is `undefined` but the API expects `null`

static/app/views/alerts/rules/metric/types.tsx
```diff
@@ -110,6 +118,8 @@ export type UnsavedMetricRule = {
   monitorWindow?: number | null;
   owner?: string | null;
   queryType?: MEPAlertsQueryType | null;
+  seasonality?: string | null;
+  sensitivity?: string | null;
```
saponifi3d, line 122: rather than using a `string` type here, could we use `AlertRuleSensitivity`? that way we can enforce it's only one of those strings, and it needs to use the enum. that helps a lot if we need to add new levels of sensitivity.

static/app/views/alerts/rules/metric/types.tsx
```diff
@@ -95,6 +102,7 @@ export type Trigger = Partial<SavedTrigger> & UnsavedTrigger;
 export type UnsavedMetricRule = {
   aggregate: string;
   dataset: Dataset;
+  detectionType: string;
```
saponifi3d, line 105: can we use the `AlertRuleComparisonType` enum for this instead of a string? (or if that's already available, do we need to add this?)

static/app/views/alerts/rules/metric/types.tsx
```diff
@@ -110,6 +118,8 @@ export type UnsavedMetricRule = {
   monitorWindow?: number | null;
   owner?: string | null;
   queryType?: MEPAlertsQueryType | null;
+  seasonality?: string | null;
```
saponifi3d, line 121: could this become an enum with `auto` and any other types? (i'm guessing we can only support a limited number of types of seasonality)

static/app/views/alerts/rules/metric/triggers/index.tsx
```diff
@@ -33,14 +35,16 @@ type Props = {
...
   ) => void;
+  onSensitivityChange: (sensitivity: AlertRuleSensitivity) => void;
   onThresholdPeriodChange: (value: number) => void;
   onThresholdTypeChange: (thresholdType: AlertRuleThresholdType) => void;
   organization: Organization;
   projects: Project[];
   resolveThreshold: UnsavedMetricRule['resolveThreshold'];
 
-  thresholdPeriod: UnsavedMetricRule['thresholdPeriod'];
+  sensitivity: UnsavedMetricRule['sensitivity'];
```
saponifi3d, line 45: 💯 nice use of types!

static/app/views/alerts/rules/metric/triggers/anomalyAlertsForm.tsx
```diff
@@ -0,0 +1,146 @@
...
+      required
+    >
+      <SelectContainer>
+        <SelectControl
+          name="sensitivity"
+          value={thresholdType}
+          options={[
+            {
+              value: AlertRuleThresholdType.ABOVE_AND_BELOW,
+              label: 'Above and below bounds',
```
saponifi3d, line 96: nit: `and` reads a little weird to me here (in code and in the product). when i read this i'm thinking it would need to be both above and below to trigger, but i think this is a logical `or` -- if we're above or below the trigger.

static/app/views/alerts/rules/metric/ruleForm.tsx
```diff
@@ -151,6 +154,7 @@ type State = {
   comparisonDelta?: number;
   isExtrapolatedChartData?: boolean;
   monitorType?: MonitorType;
+  seasonality?: string;
```
saponifi3d, line 157:
```suggestion
  seasonality?: UnsavedMetricRule['seasonality'];
```

saponifi3d (approved):
lgtm, recommend addressing the typing stuff really quickly before merging to see if there are any other issues that may be uncovered.

static/app/views/alerts/rules/metric/types.tsx
mifu67: detectionType needs to be a string because this is what's passed into the API

mifu67: Or, the strings for the model are slightly different from the `AlertRuleComparisonType` strings, so I have a map that translates between the two and sets detectionType to be a string

static/app/views/alerts/rules/metric/triggers/anomalyAlertsForm.tsx
mifu67: hmm I agree... I might change it in a different PR

Issue #75775, opened 2024-08-07
Include user comments and screenshots in Azure DevOps Issues created from Sentry User Feedback


kerenkhatiwada:
### Problem Statement

When creating a new issue from within Sentry's User Feedback to a connected Azure DevOps instance, the new Azure DevOps item does not include the comment the user sent into the description box, just a link back to the event in Sentry. While it's possible to manually add the user comment when creating the Azure DevOps issues manually, it would be great to have the Azure DevOps issue include the user feedback's comment and include screenshots by default.

### Product Area

User Feedback

bruno-garcia:
Do you create the issue manually?

If so, a template shows up like this:


![Image](https://github.com/user-attachments/assets/2f9e9707-b8bd-4b72-9f6a-da4d19eb4f14)


Do you see the message there? Perhaps under `<!--`?

johnwc:
@bruno-garcia You are referring to a sentry issue, we are referring to a User Feedback where a user has entered a comment and potentially added a screenshot from the browser.

bruno-garcia:
Right, but since we use the same component, and with other providers, I was asking if you see the comment in it.

I have a GitHub integration with my project and I see the comment, eg:


![Image](https://github.com/user-attachments/assets/c36ca0a9-2909-4910-89fa-3778188c4417)

johnwc:
I created the request because we see nothing in the dialog's Description you are screenshotting. Nor do we see any screenshots added to the issue in Azure DevOps. We also do not see any of those extra options like Assignee.

![Image](https://github.com/user-attachments/assets/4a7c8030-e0df-4695-b444-a993f9d45bf1)

michellewzhang:
@bruno-garcia @johnwc 's screenshot is accurate -- we use a different template for non-github/jira linked issues/feedback. the problem is that this template is not azure devops specific -- so changing the template would affect all other integrations sentry offers, which could be problematic. we were able to make this change easily for github/jira linked tickets because they have specific templates, so there wasn't a huge risk in updating those default descriptions. 

@johnwc we don't currently support having screenshots in any of the linked integration forms unfortunately, this is probably not something we can support anytime soon due to privacy concerns.

bruno-garcia:
> we use a different template for non-github/jira linked issues/feedback. the problem is that this template is not azure devops specific

Could we create a new, azure specific template based on the current one. And modify it only for user-feedback so that we add the comment in it?

michellewzhang:
[...]

PR #61697, opened 2023-12-13
ref(feedback): update github & jira pre-filled ticket for feedback linked issues

michellewzhang:
For both GH and Jira linked issues created from feedback:
- move the feedback message out of the table
- change the issue title to be "Sentry Feedback" rather than "Sentry Issue"

closes https://github.com/getsentry/sentry/issues/61631

PR #76016, opened 2024-08-12
ref(feedback): add feedback message to default description in remaining integrations

michellewzhang:
related to https://github.com/getsentry/sentry/issues/75775

for feedback issues, add the feedback message + all other evidence (contact email, name, etc) to the default message when creating a linked issue. affects all integrations besides github, jira, and all others where we have a custom template.

the result should look similar to this:
image

### DIFF
--- a/src/sentry/integrations/mixins/issues.py
+++ b/src/sentry/integrations/mixins/issues.py
@@ -121,10 +121,14 @@ class IssueBasicMixin:
 
     def get_group_description(self, group, event, **kwargs):
         output = self.get_group_link(group, **kwargs)
-        if event and isinstance(event, GroupEvent) and event.occurrence is not None:
+        if (
+            event
+            and isinstance(event, GroupEvent)
+            and event.occurrence is not None
+            and group.issue_category == GroupCategory.FEEDBACK
+        ):
             body = ""
-            if group.issue_category == GroupCategory.FEEDBACK:
-                body = self.get_feedback_issue_body(event)
+            body = self.get_feedback_issue_body(event)
             output.extend([body])
         else:
             body = self.get_group_body(group, event)
--- a/src/sentry/integrations/mixins/issues.py
+++ b/src/sentry/integrations/mixins/issues.py
@@ -5,8 +5,10 @@ import logging
 from collections import defaultdict
 from collections.abc import Mapping, Sequence
 from copy import deepcopy
+from operator import attrgetter
 from typing import Any, ClassVar
 
+from sentry.eventstore.models import GroupEvent
 from sentry.integrations.models.external_issue import ExternalIssue
 from sentry.integrations.services.integration import integration_service
 from sentry.integrations.tasks.sync_status_inbound import (
@@ -79,6 +81,26 @@ class IssueBasicMixin:
                 result.append(output)
         return "\n\n".join(result)
 
+    def get_feedback_issue_body(self, event: GroupEvent) -> str:
+        messages = [
+            evidence for evidence in event.occurrence.evidence_display if evidence.name == "message"
+        ]
+        others = [
+            evidence for evidence in event.occurrence.evidence_display if evidence.name != "message"
+        ]
+
+        body = ""
+        for message in messages:
+            body += message.value
+            body += "\n\n"
+
+        body += "|  |  |\n"
+        body += "| ------------- | --------------- |\n"
+        for evidence in sorted(others, key=attrgetter("important"), reverse=True):
+            body += f"| **{evidence.name}** | {evidence.value} |\n"
+
+        return body.rstrip("\n")  # remove the last new line
+
     def get_group_link(self, group, **kwargs):
         params = {}
         if kwargs.get("link_referrer"):
@@ -99,9 +121,15 @@ class IssueBasicMixin:
 
     def get_group_description(self, group, event, **kwargs):
         output = self.get_group_link(group, **kwargs)
-        body = self.get_group_body(group, event)
-        if body:
-            output.extend(["", "```", body, "```"])
+        if isinstance(event, GroupEvent) and event.occurrence is not None:
+            body = ""
+            if group.issue_category == GroupCategory.FEEDBACK:
+                body = self.get_feedback_issue_body(event)
+            output.extend([body])
+        else:
+            body = self.get_group_body(group, event)
+            if body:
+                output.extend(["", "```", body, "```"])
         return "\n".join(output)
 
     @all_silo_function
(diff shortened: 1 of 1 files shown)


src/sentry/integrations/mixins/issues.py
```diff
@@ -99,9 +121,15 @@ def get_group_link(self, group, **kwargs):
 
     def get_group_description(self, group, event, **kwargs):
         output = self.get_group_link(group, **kwargs)
-        body = self.get_group_body(group, event)
-        if body:
-            output.extend(["", "```", body, "```"])
+        if isinstance(event, GroupEvent) and event.occurrence is not None:
```
aliu39, line 124: I think event could be None in `get_create_issue_config`

aliu39:
Where is the "Sentry Feedback" change made?

src/sentry/integrations/mixins/issues.py
```diff
@@ -99,9 +121,15 @@ def get_group_link(self, group, **kwargs):
...
+            body = ""
+            if group.issue_category == GroupCategory.FEEDBACK:
```
aliu39, line 126: Do we want to output nothing if the category != Feedback?

michellewzhang:
> Where is the "Sentry Feedback" change made?

it already exists in the code, the change here is adding the feedback message to the description

src/sentry/integrations/mixins/issues.py
```diff
@@ -79,6 +81,26 @@ def get_group_body(self, group, event, **kwargs):
...
+        messages = [
+            evidence for evidence in event.occurrence.evidence_display if evidence.name == "message"
+        ]
+        others = [
+            evidence for evidence in event.occurrence.evidence_display if evidence.name != "message"
+        ]
+
+        body = ""
+        for message in messages:
+            body += message.value
```
aliu39, line 94:
```suggestion
            body += "Message from the user: "
            body += message.value
```

src/sentry/integrations/mixins/issues.py
michellewzhang: if the category isn't feedback it'll return the default text before the table, which is what it always has returned

michellewzhang: SCR-20240812-nlxr

src/sentry/integrations/mixins/issues.py
```diff
@@ -79,6 +81,26 @@ def get_group_body(self, group, event, **kwargs):
...
+            body += "\n\n"
+
+        body += "|  |  |\n"
+        body += "| ------------- | --------------- |\n"
+        for evidence in sorted(others, key=attrgetter("important"), reverse=True):
```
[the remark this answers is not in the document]
michellewzhang: no, but we've included them in the past for gh/jira just as an fyi. the user can easily choose to delete whatever they don't see as necessary for the ticket

src/sentry/integrations/mixins/issues.py
aliu39: Is there no need to call `get_group_body` for this case?

aliu39:
> > Where is the "Sentry Feedback" change made?
> 
> it already exists in the code, the change here is adding the feedback message to the description -
> 
> https://github.com/getsentry/sentry/blob/dfcfaa7efa692bb161df9157ece589f98dacada1/src/sentry/integrations/mixins/issues.py#L111-L113
> 
> pr that added this was here: #61697, along with the changes to github & jira. i basically copied and pasted the format we use for github/jira into here @aliu39

Thanks for the added context! 🔥

src/sentry/integrations/mixins/issues.py
michellewzhang: SCR-20240812-npuf

it looks like currently for feedback, `get_group_body` just returns nothing. probably this line doesn't execute because `event.interfaces.values()` doesn't exist

https://github.com/getsentry/sentry/blob/dfcfaa7efa692bb161df9157ece589f98dacada1/src/sentry/integrations/mixins/issues.py#L78

so not calling it won't have an effect on feedback descriptions

src/sentry/integrations/mixins/issues.py
```diff
@@ -79,6 +81,26 @@ def get_group_body(self, group, event, **kwargs):
                 result.append(output)
         return "\n\n".join(result)
 
+    def get_feedback_issue_body(self, event: GroupEvent) -> str:
+        messages = [
+            evidence for evidence in event.occurrence.evidence_display if evidence.name == "message"
```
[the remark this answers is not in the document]
michellewzhang: hmm i don't think so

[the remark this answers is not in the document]
michellewzhang: this seems like the easiest way to filter for the evidence `message` though -- do you have another way to parse for just one `message`?

src/sentry/integrations/mixins/issues.py
michellewzhang: added an extra check that it exists

src/sentry/integrations/mixins/issues.py
aliu39: I meant for non-Feedbacks, so if 
`isinstance(event, GroupEvent) and event.occurrence is not None and group.issue_category != GroupCategory.Feedback`

aliu39: > if the category isn't feedback it'll return the default text before the table

What/where is the default text? Sorry if I'm kinda behind here

src/sentry/integrations/mixins/issues.py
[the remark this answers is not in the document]
aliu39: Gotcha, this way should be fine, I was just curious!

src/sentry/integrations/mixins/issues.py
michellewzhang: ahh i see what you mean -- i'll add the feedback category check into the main if statement, so if the feedback isn't issues it'll still default to the `else` case

src/sentry/integrations/mixins/issues.py
```diff
@@ -121,10 +121,14 @@ def get_group_link(self, group, **kwargs):
...
     def get_group_description(self, group, event, **kwargs):
         output = self.get_group_link(group, **kwargs)
-        if isinstance(event, GroupEvent) and event.occurrence is not None:
+        if (
+            event
+            and isinstance(event, GroupEvent)
+            and event.occurrence is not None
+            and group.issue_category == GroupCategory.FEEDBACK
+        ):
             body = ""
```
aliu39, line 130:
```suggestion
```

aliu39 (approved):

src/sentry/integrations/mixins/issues.py
michellewzhang: Flake8 gives me `undefined name 'body'` :(

aliu39: oh that's so weird

PR #76019, opened 2024-08-12
feat(perf): Add metrics warning trailing items to new Transaction Summary search

0Calories:
Adds back the warning icon action bar elements to the new `SearchQueryBuilder` component on the transaction summary page, using the new `trailingItems` prop

image

### DIFF
--- a/static/app/views/performance/transactionSummary/transactionOverview/content.tsx
+++ b/static/app/views/performance/transactionSummary/transactionOverview/content.tsx
@@ -222,6 +222,26 @@ function SummaryContent({
     return items;
   }
 
+  const generateTrailingItems = useCallback(() => {
+    if (!canUseTransactionMetricsData(organization, mepDataContext)) {
+      return (
+        <Tooltip
+          title={t(
+            'Based on your search criteria and sample rate, the events available may be limited.'
+          )}
+        >
+          <StyledIconWarning
+            data-test-id="search-metrics-fallback-warning"
+            size="sm"
+            color="warningText"
+          />
+        </Tooltip>
+      );
+    }
+
+    return null;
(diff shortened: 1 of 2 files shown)


static/app/views/performance/transactionSummary/transactionOverview/content.tsx
```diff
@@ -363,6 +359,7 @@ function SummaryContent({
           searchSource="transaction_summary"
           disableLoadingTags // already loaded by the parent component
           filterKeyMenuWidth={420}
+          trailingItems={generateTrailingItems()}
```
Zylphrex, line 362: This use of `useCallback` doesn't really make sense as you're not passing the memorized callback anywhere. It could make more sense if you used `useMemo` to memorize the component but even then, you could do something as simple as to achieve the same effect.
```suggestion
          trailingItems={canUseTransactionMetricsData(organization, mepDataContext) ? null : <MetricsWarningIcon />}
```
Not sure about the implementation of `canUseTransactionMetricsData` but if that's expensive, you can cache that and do something like
```tsx
  const ableToUseTransactionMetricsData = canUseTransactionMetricsData(organization, mepDataContext);
  return (
        <TransactionSearchQueryBuilder
          ...
          trailingItems={ableToUseTransactionMetricsData ? null : <MetricsWarningIcon />}
```

0Calories: ah yes good catch, I'll memoize the trailing items instead. It's not that `canUseTransactionMetricsData`, but I'm trying to avoid unnecessary rerenders on the `SearchQueryBuilder` component, as I've been warned that rerenders are expensive for it.

Zylphrex (approved):

PR #76028, opened 2024-08-12
ref(ui): Add LinkButton to tests

evanpurkhiser:


static/app/components/button.spec.tsx
```diff
@@ -39,3 +27,17 @@ describe('Button', function () {
...
+  it('renders react-router link', function () {
+    render(<LinkButton to="/some/route">Router Link</LinkButton>);
+  });
+
+  it('renders normal link', function () {
+    render(<LinkButton href="/some/relative/url">Normal Link</LinkButton>);
+  });
+
+  it('renders disabled normal link', function () {
+    render(<LinkButton href="/some/relative/url">Normal Link</LinkButton>);
```
michellewzhang, line 41: should this have the `disabled` prop?

priscilawebdev: agree with @michellewzhang ... this should have the disabled prop and it would be nice to test if the button is unclickable like here https://github.com/getsentry/sentry/blob/af68d66484840ce61c48059bdf4764eb6b5efb52/static/app/components/button.spec.tsx#L30-L39

static/app/components/button.spec.tsx
```diff
@@ -39,3 +27,17 @@ describe('Button', function () {
...
   });
 });
+
+describe('LinkButton', function () {
+  it('renders react-router link', function () {
+    render(<LinkButton to="/some/route">Router Link</LinkButton>);
+  });
+
+  it('renders normal link', function () {
+    render(<LinkButton href="/some/relative/url">Normal Link</LinkButton>);
```
priscilawebdev, line 37: this could render a text. shouldn't we verify if the  _button_ contains a _href_ and it is _active_?

priscilawebdev: example:

`expect(screen.getByRole('button', {name: 'Normal Link'})).toHaveAttribute('href', '/some/relative/url');`

evanpurkhiser: loooks like it didn't before lol

evanpurkhiser: Well in this case it should be a `a` element, not a button 😉

priscilawebdev: this tricked me - yes we should check for an `a` element 😄

evanpurkhiser: it is a double trick because we set `role="button"` on this lol

michellewzhang (approved):

PR #76031, opened 2024-08-12
fix(weekly-reports): dupe on report timestamp instead of batch

JoshFerge:
- It has been observed that a portion of our duplicate weekly emails come from the schedule celery task being killed and restarted, which will cause duplicates for any orgs that were queued in the first batch. 
- this also generates a new batch_id, which causes our de-duplication logic to not work, as batch_id is included in the redis key that we use.
- [x] This PR modifies our logic so that the dupe checker instead uses the timestamp for which the weekly report is generated on. this should be a safe way to ensure no duplicates occur across batches.
- [x] Updates the test.

### DIFF
--- a/src/sentry/tasks/summaries/weekly_reports.py
+++ b/src/sentry/tasks/summaries/weekly_reports.py
@@ -5,7 +5,7 @@ import logging
 import uuid
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
-from datetime import timedelta
+from datetime import datetime, timedelta
 from functools import partial
 from typing import Any, cast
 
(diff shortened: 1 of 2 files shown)


mrduncan (approved):
Definitely not necessary here but I wonder if we could just use the date as the batch id instead of a uuid, it seems like it's pretty much only used for logging from what I could tell at a glance

JoshFerge:
> Definitely not necessary here but I wonder if we could just use the date as the batch id instead of a uuid, it seems like it's pretty much only used for logging from what I could tell at a glance

yeah, but i do think it is helpful debugging context for the future, as it would be harder to tell what logs came from what batch without a unique id per schedule celery task

PR #54433, opened 2023-08-08
feat(ui): carousel component

cathteng:
Left-justified carousel.

Example:

https://github.com/getsentry/sentry/assets/70817427/3611fe76-5591-4df3-bf92-ad5d52e8d727

Mobile (ignore how ugly the cards look)

https://github.com/getsentry/sentry/assets/70817427/41055d7d-f35e-4bc1-82ed-4a867588656c

Dark mode

Screenshot 2023-08-11 at 09 25 04

PR #76033, opened 2024-08-13
feat(issues): Add scrollCarousel component

scttcper:
Adds a ScrollCarousel component that is very similar to the existing Carousel (added in https://github.com/getsentry/sentry/pull/54433) with some key differences:
- Browser scrollbars are hidden
- There's a white mask indicating there is more to scroll
- The buttons are different
- It jumps more than one item left/right on arrow click

https://github.com/user-attachments/assets/10b4916d-0fa0-4f2b-bd09-a52148724015

what it would look like on issue details

https://github.com/user-attachments/assets/82ea7da8-d647-46b6-8006-920ecd28c1d7

scroll on issue tracking
image


static/app/components/group/externalIssuesList/streamlinedExternalIssueList.tsx
```diff
@@ -78,8 +79,9 @@ export function StreamlinedExternalIssueList({
               </Tooltip>
             </ErrorBoundary>
           ))}
-          {integrations.length
-            ? integrations.map(integration => {
+          {integrations.length ? (
+            <ScrollCarousel>
```
malwilley, line 83: Are you sure you want to use the carousel for this? IMO the buttons seem better wrapping to multiple lines

static/app/utils/useRefChildrenVisibility.tsx
```diff
@@ -0,0 +1,60 @@
...
+  visibleRatio,
+}: UseRefChildrenVisibilityOptions) {
+  // The visibility match up to the elements list. Visibility of elements is
+  // true if visible in the scroll container, false if outside.
+  const [childrenEls, setChildrenEls] = useState<HTMLElement[]>([]);
+  const [visibility, setVisibility] = useState<boolean[]>([]);
+
+  // Update list of children element
+  useEffect(
+    () => setChildrenEls(Array.from(ref.current?.children ?? []) as HTMLElement[]),
```
malwilley, line 22: Seems like this component won't work if the children change after first render, is that right?

static/app/components/scrollCarousel.tsx
```diff
@@ -0,0 +1,148 @@
...
+    </ScrollCarouselWrapper>
+  );
+}
+
+const ScrollCarouselWrapper = styled('div')`
+  position: relative;
+  overflow: hidden;
+`;
+
+const ScrollContainer = styled('div')`
```
malwilley, line 101: Feels pretty good as is, but do you think it's worth adding `scroll-snap-type`?

static/app/utils/useRefChildrenVisibility.tsx
scttcper: added back children

malwilley (approved):

static/app/components/scrollCarousel.tsx
scttcper: i'm still playing around with that

malwilley:
@scttcper the arrow button seems a bit small and kinda ovular?

image

static/app/components/scrollCarousel.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+import storyBook from 'sentry/stories/storyBook';
+import {space} from 'sentry/styles/space';
+
+import {ScrollCarousel} from './scrollCarousel';
+
+export default storyBook(ScrollCarousel, story => {
+  story('Default', () => (
+    <Fragment>
+      <p>
+        <JSXNode name="SegmentedControl" /> will detect if the content overflows and show
```
malwilley, line 15: Wrong name!

PR #76036, opened 2024-08-13
feat(django-onboarding): Move filename into snippet header

ArthurKnaus:
#### Before
Screenshot 2024-08-13 at 07 11 57

#### After
Screenshot 2024-08-13 at 07 11 01

Relates to #75231


obostjancic (approved):

priscilawebdev (approved):
Thanks for doing that! It would be great if we do the same for all platforms or was django the only one missing? 🙂

PR #76038, opened 2024-08-13
feat(remix-onboarding): Remove next steps

ArthurKnaus:
The remix onboarding uses the sentry wizard, which already sets up performance monitoring and session replay.
Displaying those as next steps only creates confusion.


obostjancic (approved):

priscilawebdev (approved):
good catch here! It's interesting that, unlike the other platforms, we don't show the manual instructions in the onboarding

PR #76041, opened 2024-08-13
ref(ts): Remove external from CommonButtonProps

evanpurkhiser:


static/app/components/button.tsx
```diff
@@ -162,6 +164,10 @@ interface HrefLinkButtonProps extends BaseLinkButtonProps {
    * See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download
    */
   download?: HTMLAnchorElement['download'];
+  /**
+   * The button is an external link. Similar to the `Link` `external` property.
+   */
+  external?: boolean;
```
priscilawebdev, line 170: do we want to have the same prop 2x?

priscilawebdev (approved):

PR #76051, opened 2024-08-13
feat(webpack): Add Sentry Webpack Plugin

lforst:
Adds the Sentry webpack plugin:
- To strip away debug statements in prod
- To mark errors from third party sources with a tag so we can filter them out in Sentry


0Calories (approved):
LGTM!

PR #76053, opened 2024-08-13
feat(tooltip): Create tooltip context

### LINKED ISSUE #75624: Create a Tooltip Context to set the tooltip's position
Work on some kind of tooltip context instead of making the tooltip aware of modals. Basically the context would tell the tooltip where to render. So by default without a context it would render into the body. But then if a tooltip is rendered within the tooltip context then it would render somewhere else. Basically the global modal could be wrapped in that context, so tooltips rendered inside modals would know to render somewhere else

priscilawebdev:
We do not want the tooltip component to be aware of the global modals state, so this PR adds a tooltip context that we use inside the global modal

closes https://github.com/getsentry/sentry/issues/75624


static/app/components/tooltip.tsx
```diff
@@ -65,14 +69,13 @@ function Tooltip({
...
-  // Therefore, we check if the global modal is active. If it is, the tooltip should be rendered within the same node to ensure interactivity.
-  const container = modalIsVisible ? getModalPortal() : document.body;
-
   return (
     <Fragment>
       {wrapTrigger(children)}
-      {createPortal(<AnimatePresence>{tooltipContent}</AnimatePresence>, container)}
+      {createPortal(
+        <AnimatePresence>{tooltipContent}</AnimatePresence>,
+        context.container ?? document.body
```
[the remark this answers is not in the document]
priscilawebdev: ok we can do that too

static/app/components/tooltip.tsx
```diff
@@ -1,17 +1,21 @@
...
 
-import {useGlobalModal} from 'sentry/components/globalModal/useGlobalModal';
 import {Overlay, PositionWrapper} from 'sentry/components/overlay';
 import {space} from 'sentry/styles/space';
-import getModalPortal from 'sentry/utils/getModalPortal';
 import type {UseHoverOverlayProps} from 'sentry/utils/useHoverOverlay';
 import {useHoverOverlay} from 'sentry/utils/useHoverOverlay';
 
+interface TooltipContextProps {
+  container: Parameters<typeof createPortal>[1] | undefined;
```
evanpurkhiser, line 14: Could we add a doc comment here? something about this allowing the tooltip to be portaled to a different place and useful for focus traps

static/app/components/tooltip.tsx
```diff
@@ -35,8 +39,8 @@ function Tooltip({
   disabled = false,
   ...hoverOverlayProps
 }: TooltipProps) {
+  const context = useContext(TooltipContext);
   const theme = useTheme();
-  const {visible: modalIsVisible} = useGlobalModal();
```
evanpurkhiser, line 39: yay!

static/app/components/tooltip.tsx
```diff
@@ -1,17 +1,21 @@
-import {Fragment, useEffect} from 'react';
+import {createContext, Fragment, useContext, useEffect} from 'react';
 import {createPortal} from 'react-dom';
 import type {SerializedStyles} from '@emotion/react';
 import {useTheme} from '@emotion/react';
 import styled from '@emotion/styled';
 import {AnimatePresence} from 'framer-motion';
 
-import {useGlobalModal} from 'sentry/components/globalModal/useGlobalModal';
```
evanpurkhiser, line 8: Are we able to delete this?

priscilawebdev: it is being used in a few places

static/app/components/globalModal/index.spec.tsx
```diff
@@ -7,6 +7,8 @@ import {
 } from 'sentry-test/reactTestingLibrary';
 
 import {closeModal, openModal} from 'sentry/actionCreators/modal';
+import {Button} from 'sentry/components/button';
```
evanpurkhiser, line 10: No need to import our button here for the test, just use a regular `button`

evanpurkhiser (approved):
left a couple comments but great work!

static/app/components/tooltip.tsx
evanpurkhiser: ah ok

PR #76055, opened 2024-08-13
typing: Group deletion defaults

armenzg:


src/sentry/deletions/defaults/group.py
```diff
@@ -52,12 +55,12 @@ class EventDataDeletionTask(BaseDeletionTask):
     # Number of events fetched from eventstore per chunk() call.
     DEFAULT_CHUNK_SIZE = 10000
 
-    def __init__(self, manager, groups, **kwargs):
+    def __init__(self, manager: Any, groups: Sequence[Group], **kwargs: Any) -> None:
```
asottile-sentry, line 58: manager is `DeletionTaskManager`

src/sentry/deletions/defaults/group.py
```diff
@@ -148,17 +151,15 @@ def delete_bulk(self, instance_list):
...
 
         if not self.skip_models or similarity not in self.skip_models:
             similarity.delete(None, instance)
 
         return super().delete_instance(instance)
 
-    def mark_deletion_in_progress(self, instance_list):
-        from sentry.models.group import Group, GroupStatus
-
+    def mark_deletion_in_progress(self, instance_list: Sequence[Group]) -> None:
```
asottile-sentry, line 162: these are likely `Iterable[Group]` instead -- since a `QuerySet[Group]` is not a `Sequence[Group]`

src/sentry/deletions/defaults/group.py
```diff
@@ -84,9 +90,9 @@ def chunk(self):
             limit=self.DEFAULT_CHUNK_SIZE,
             referrer="deletions.group",
             orderby=["-timestamp", "-event_id"],
-            tenant_ids={"organization_id": self.groups[0].project.organization_id}
-            if self.groups
-            else None,
+            tenant_ids=(
+                {"organization_id": self.groups[0].project.organization_id} if self.groups else None
```
armenzg, line 94: This is now failing with: `error: Value of type "Iterable[Group]" is not indexable`

asottile-sentry (approved):

PR #76060, opened 2024-08-13
chore(performance): Use search query builder for spans pages

Zylphrex:
This ports the spans pages within performance to use the new search query builder UX.


static/app/components/performance/spanSearchQueryBuilder.tsx
```diff
@@ -110,7 +110,7 @@ export function SpanSearchQueryBuilder({
       getTagValues={getSpanFilterTagValues}
       disallowFreeText
       disallowUnsupportedFilters
-      recentSearches={SavedSearchType.EVENT}
+      recentSearches={SavedSearchType.SPAN}
```
Zylphrex, line 113: This ensures we dont pollute recent searches for events with span queries.

JonasBa (approved):

PR #76068, opened 2024-08-13
chore(hybridcloud) Add sentryapp RPC for use in ratelimiting

markstory:
Within ratelimiting we need to resolve the organization id from the installation's api_token.id. Currently this involves fetching the installation, app, and api token across RPC. From this result we only use the `organization_id`. Having a dedicated RPC method means we can trim out db queries to load the sentryapp, apiapplication, and serialize a much smaller result.

By fetching and serializing less data, I'm hoping to see an improvement in this trace segment:

Screenshot 2024-08-13 at 12 37 29 PM

### DIFF
--- a/src/sentry/sentry_apps/services/app/impl.py
+++ b/src/sentry/sentry_apps/services/app/impl.py
@@ -76,6 +76,17 @@ class DatabaseBackedAppService(AppService):
         except SentryAppInstallation.DoesNotExist:
             return None
 
+    def get_installation_org_id_by_token_id(self, token_id: int) -> int | None:
+        filters: SentryAppInstallationFilterArgs = {
+            "status": SentryAppInstallationStatus.INSTALLED,
+            "api_installation_token_id": str(token_id),
+        }
+        queryset = self._FQ.apply_filters(SentryAppInstallation.objects.all(), filters)
+        install = queryset.first()
+        if not install:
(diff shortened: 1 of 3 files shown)


src/sentry/sentry_apps/services/app/impl.py
```diff
@@ -76,6 +76,17 @@ def get_installation_by_id(self, *, id: int) -> RpcSentryAppInstallation | None:
...
+    def get_installation_org_id_by_token_id(self, token_id: int) -> int | None:
+        filters: SentryAppInstallationFilterArgs = {
+            "status": SentryAppInstallationStatus.INSTALLED,
+            "api_installation_token_id": str(token_id),
+        }
+        queryset = self._FQ.apply_filters(SentryAppInstallation.objects.all(), filters)
+        install = queryset.first()
+        if not install:
+            return None
+        return install.organization_id
```
markstory, line 88: This logic mirrors the current logic used in ratelimiting.

https://github.com/getsentry/sentry/blob/77f283353930f0b7abe05fbbfce1f78127c9a7d6/src/sentry/ratelimits/utils.py#L137-L154

I'll replace this logic with the new RPC call in a separate pull request because of deploy boundary issues.

mrduncan (approved):

PR #76072, opened 2024-08-13
ref(control_silo): Move LostPasswordHash to users module

Christinarlong:
Part of moving control silo user related resources into the users module. Includes adding of types for functions.

Apart of (#73856)


src/sentry/users/models/lostpasswordhash.py
```diff
@@ -0,0 +1,111 @@
...
+        cls._send_email("recover_password", user, hash, extra)
+
+    @classmethod
+    def send_relocate_account_email(
+        cls, user: User | RpcUser, hash: str, orgs: Iterable[str]
+    ) -> None:
+        cls._send_email("relocate_account", user, hash, {"orgs": orgs})
+
+    @classmethod
+    def _send_email(cls, mode: str, user: User | RpcUser, hash: str, extra: dict[str, Any]) -> None:
```
Christinarlong, line 62: the extra is a bit odd since it's basically **kwargs

src/sentry/users/models/lostpasswordhash.py
```diff
@@ -0,0 +1,111 @@
...
+        }
+        cls._send_email("recover_password", user, hash, extra)
+
+    @classmethod
+    def send_relocate_account_email(
+        cls, user: User | RpcUser, hash: str, orgs: Iterable[str]
+    ) -> None:
+        cls._send_email("relocate_account", user, hash, {"orgs": orgs})
+
+    @classmethod
```
Christinarlong, line 61: mainly added typing to these funcs

src/sentry/users/models/lostpasswordhash.py
Christinarlong, line 1: Added some typing here - mainly lines 48-62

src/sentry/tasks/relocation.py
```diff
@@ -63,6 +62,7 @@
 from sentry.silo.base import SiloMode
 from sentry.tasks.base import instrumented_task
 from sentry.types.region import get_local_region
+from sentry.users.models.lostpasswordhash import LostPasswordHash as LostPasswordHash
```
markstory, line 65:
```suggestion
from sentry.users.models.lostpasswordhash import LostPasswordHash
```
We don't need to alias this to the same name.

src/sentry/users/models/lostpasswordhash.py
markstory: Yeah, we don't need to change it though. We have a similar pattern in a few other email builder classes that take a mapping of data to pass into templates that are being rendered.

markstory (approved):
Looks great :clap:

PR #76073, opened 2024-08-13
fix(insights): fix sample panel padding

KevinL10:
Fixes spacing on sample panel:
image


static/app/views/insights/common/views/spanSummaryPage/sampleList/index.tsx
```diff
@@ -278,4 +278,5 @@ const Title = styled('h4')`
 
 const StyledSearchBar = styled(SearchBar)`
   margin-top: ${space(2)};
+  margin-bottom: ${space(2)};
```
DominikB2014, line 281: You can replace both of these lines with a single line
```
margin: {space(2)} 0
```

DominikB2014 (approved):
Lgtm, just one nit

PR #76078, opened 2024-08-13
ref(issue-details): Remove <GroupEventEntry />

leeandher:
This PR removes `GroupEventEntry` from the issue details page, and renames its component `EventDetailsContent` rather than `DefaultGroupEventDetailsContent`.

The reasons for this are a few:
- I think we should get rid of `EventEntry` wholesale. It is a component which can render almost 13 different components based on its props which makes it difficult to work with as you have to assume a lot of tenuously linked components have similar styles and layouts.
- `EventEntry` is used in `EventEntries` and `Entries` and another newer `Entries`, all components that I think should be removed as well. They seem to have the goal of displaying all the event data that you may find on issue details, but they are similar at best. It is not in parity with issue details and updates do not always make their way to issue details or vice versa. 

I think the next steps are to replace the above 3 components with direct calls to `EventDetailsContent` and maybe add some config for disabling sections that the caller does not want rendered, allowing each usage to be explicit about what gets shown. Currently, if you were to add to `EventEntry`, the changes would be propagated to discover, trace view, shared issues and not issue details, since this system is half adopted.


static/app/views/issueDetails/groupEventDetails/groupEventDetailsContent.tsx
```diff
@@ -74,63 +85,39 @@ const LLMMonitoringSection = lazy(
...
   const hasNewTimelineUI = useHasNewTimelineUI();
   const hasStreamlinedUI = useHasStreamlinedUI();
   const tagsRef = useRef<HTMLDivElement>(null);
+  const eventEntries: {[key in EntryType]?: Entry} = useMemo(() => {
+    const {entries = []} = event;
+    return entries.reduce((entryMap, entry) => {
+      entryMap[entry.type] = entry;
+      return entryMap;
+    }, {});
+  }, [event]);
```
scttcper, line 110: move the type inside the reduce otherwise the reduce is adding items to `{}` and its being cast after
```suggestion
  const eventEntries = useMemo(() => {
    const {entries = []} = event;
    return entries.reduce<{[key in EntryType]?: Entry}>((entryMap, entry) => {
      entryMap[entry.type] = entry;
      return entryMap;
    }, {});
  }, [event]);
```

scttcper (approved):

malwilley:
Totally agree with getting rid of GroupEventEntry 🎉

static/app/views/issueDetails/groupEventDetails/groupEventDetailsContent.tsx
```diff
@@ -455,7 +477,7 @@ function ProfilingDurationRegressionIssueDetailsContent({
               frame with the largest increase in call stack population likely
               contributed to the cause for the duration regression.`)}
             </p>
-
```
leeandher, line 458: h

static/app/views/issueDetails/groupEventDetails/groupEventDetailsContent.tsx
```diff
@@ -455,7 +477,7 @@ function ProfilingDurationRegressionIssueDetailsContent({
...
+            h
```
leeandher, line 480: h

PR #76080, opened 2024-08-13
ref(integrations): convert issue mixins to ABCs

cathteng:
Old issues mixin organization:
Screenshot 2024-08-15 at 11 38 17

New issues abstract base class organization:
Screenshot 2024-08-15 at 11 37 10
Notes about the new organization:
- `BaseRepositoryIntegration` includes the `get_repositories` function, which is present in `SourceCodeIssueIntegration` AND `RepositoryIntegration`.
- The new `SourceCodeIssueIntegration` ABC includes functionality only present in the SCM integrations, which includes searching for repos to create the issue in
- Some of the issues specs are only inherited by their respective integration, which suggests that they could be combined with the integration (e.g. `BitbucketIssuesSpec` + `BitbucketIntegration`). This could be done later, but I personally like that the issue functionality is separated since it's typically a lot of code and the integration classes already include a lot of methods.

The end goal should be that `SourceCodeIssueIntegration` includes all of the features of `IssueSyncIntegration`!

### DIFF
--- a/src/sentry/integrations/bitbucket/integration.py
+++ b/src/sentry/integrations/bitbucket/integration.py
@@ -16,6 +16,7 @@ from sentry.integrations.base import (
 )
 from sentry.integrations.models.integration import Integration
 from sentry.integrations.services.repository import RpcRepository, repository_service
+from sentry.integrations.source_code_management.repository import RepositoryIntegration
 from sentry.integrations.tasks.migrate_repo import migrate_repo
 from sentry.integrations.utils import AtlassianConnectValidationError, get_integration_from_request
 from sentry.models.repository import Repository
@@ -25,7 +26,7 @@ from sentry.shared_integrations.exceptions import ApiError
 from sentry.utils.http import absolute_uri
 
 from .client import BitbucketApiClient
-from .issues import BitbucketSCMSpec
+from .issues import BitbucketIssuesSpec
 from .repository import BitbucketRepositoryProvider
 
 DESCRIPTION = """
@@ -81,7 +82,7 @@ metadata = IntegrationMetadata(
 scopes = ("issue:write", "pullrequest", "webhook", "repository")
 
 
-class BitbucketIntegration(BitbucketSCMSpec):
+class BitbucketIntegration(RepositoryIntegration, BitbucketIssuesSpec):
     @property
     def integration_name(self) -> str:
         return "bitbucket"
--- a/src/sentry/integrations/bitbucket/integration.py
+++ b/src/sentry/integrations/bitbucket/integration.py
@@ -16,7 +16,6 @@ from sentry.integrations.base import (
 )
 from sentry.integrations.models.integration import Integration
 from sentry.integrations.services.repository import RpcRepository, repository_service
-from sentry.integrations.source_code_management.repository import RepositoryIntegration
 from sentry.integrations.tasks.migrate_repo import migrate_repo
 from sentry.integrations.utils import AtlassianConnectValidationError, get_integration_from_request
 from sentry.models.repository import Repository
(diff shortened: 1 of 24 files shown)


src/sentry/integrations/bitbucket/integration.py
```diff
@@ -26,7 +26,7 @@
 from sentry.utils.http import absolute_uri
 
 from .client import BitbucketApiClient
-from .issues import BitbucketIssueBasicMixin
+from .issues import BitbucketIssuesSpec
```
cathteng, line 29: i'm not satisfied with the names of these things 🤔

RyanSkonnord: `BitbucketIssueBasicMixin`/`BitbucketIssuesSpec` exists only to be inherited by `BitbucketIntegration`, right? Could we just delete it and move all of its contents directly into `BitbucketIntegration`? Then we wouldn't need to name it at all.

The main argument against doing that would be that it makes `BitbucketIntegration` too long. There is some value in keeping all the issue-related methods grouped together, I suppose.

But the difficulty in thinking of a name for the class can be a signal that the class doesn't represent something sensible and therefore shouldn't exist. 😉

cathteng: i'll keep this for now, but it's easily separable

src/sentry/integrations/source_code_management/repository.py
```diff
@@ -19,7 +19,28 @@
...
+        return [{
+            'name': display_name,
+            'identifier': external_repo_id,
+        }]
+
+        The shape of the `identifier` should match the data
+        returned by the integration's
+        IntegrationRepositoryProvider.repository_external_slug()
+        """
+        raise NotImplementedError
```
GabeVillalobos, line 40: Do you anticipate us moving more of the RepositoryIntegration methods into this new base class? Seems silly to have a new base class with a single new method, but I get why this was needed to handle the diamond inheritance issue.

RyanSkonnord: I don't really mind having the common base class between `SourceCodeIssueIntegration` and `RepositoryIntegration` even if it is just this.

I don't know if we're relying on the polymorphism anywhere -- i.e., code that calls `get_repositories` an object without knowing whether it's a `SourceCodeIssueIntegration` or a`RepositoryIntegration`. I doubt it. By that logic, you could argue that it would be fine for `SourceCodeIssueIntegration` and `RepositoryIntegration` each to have their own, separate `get_repositories` method. IMHO it comes down to communicating intent. If each class declares its own `get_repositories` method, that communicates that they're semantically unrelated, and that it should be fine to change the signature or return type on one of them without affecting the other. But if both inherit the abstract method from the same base class, it communicates to humans that `get_repositories` does the same thing with the same rules on both classes, even if the software doesn't care.

Considering the amount of information communicated in the docstring, I fall heavily on the latter side in this case. (Put more simply: we don't want to copy-and-paste that long docstring, do we?) In fact, the implicitly typed dict as the return value is begging to be refactored as it is, so all the better if we refactor it from a single base class.

RyanSkonnord (approved):

src/sentry/integrations/source_code_management/issues.py
```diff
@@ -0,0 +1,57 @@
...
+
+        return default_repo, repo_choices
+
+    # TODO(saif): Make private and move all usages over to `get_defaults`
+    def get_project_defaults(self, project_id):
+        if not self.org_integration:
+            return {}
+
+        return self.org_integration.config.get("project_issue_defaults", {}).get(
+            str(project_id), {}
```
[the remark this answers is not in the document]
cathteng: this is only used in source code issue integrations at the moment. i can move it out later if it turns out it's used in Jira

src/sentry/integrations/source_code_management/repository.py
GabeVillalobos: >IMHO it comes down to communicating intent. If each class declares its own get_repositories method, that communicates that they're semantically unrelated, and that it should be fine to change the signature or return type on one of them without affecting the other.

I agree, and this is the code smelly part of this implementation, which makes me wonder if we're overloading this method for multiple distinct use cases; although I reckon the implementation will be identical for now. Not a big deal, but ideally we'd reduce the number of classes we unnecessarily inherit from where we can.

PR #76081, opened 2024-08-13
fix(similarity): Fix backfill query retry logic

jangjodi:
Move iteration over query result into `_make_postgres_call_with_filter` function so that the query actually runs in this function
This will allow the retry logic to actually work and for us to remove unneeded count call
Return batch end group id from `_make_postgres_call_with_filter` since it could be one of the groups that we filtered out

### DIFF
--- a/src/sentry/tasks/embeddings_grouping/utils.py
+++ b/src/sentry/tasks/embeddings_grouping/utils.py
@@ -133,22 +133,26 @@ def _make_postgres_call_with_filter(group_id_filter: Q, project_id: int, batch_s
     )
 
     # Filter out groups that are pending deletion in memory so postgres won't make a bad query plan
-    groups_to_backfill_batch = [
-        (group[0], group[1])
-        for group in groups_to_backfill_batch_raw
-        if group[2] not in [GroupStatus.PENDING_DELETION, GroupStatus.DELETION_IN_PROGRESS]
-        and group[3] > datetime.now(UTC) - timedelta(days=90)
-    ]
+    # Get the last queried group id while we are iterating; even if it's not valid to be backfilled
+    # we want to keep the value to be used as an filter for the next batch
+    groups_to_backfill_batch, batch_end_group_id = [], None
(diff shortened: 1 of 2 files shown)


wedamija (approved):

wedamija:
> ## Suspect Issues
> 
> * ‼️ **OperationalError: QueryCanceled('canceling statement due to user request\n')** `sentry.tasks.backfill_seer_grouping_records` [View Issue](https://sentry.sentry.io/issues/5709883699/?referrer=github-pr-bot)
> 

@jangjodi  I don't think the retry will help, here's the query plan (I can send you the query on slack if you want to run it yourself)
```

Limit (cost=0.70..194.54 rows=45 width=567) (actual time=5814.282..82962.850 rows=26 loops=1) 
-> Index Scan Backward using sentry_grou_project_41a5ce_idx on sentry_groupedmessage (cost=0.70..20918.00 rows=4856 width=567) (actual time=5814.280..82962.816 rows=26 loops=1) |  
Index Cond: (project_id = '<snipped>'::bigint) 
Filter: ((times_seen > 1) AND (type = 1)) 
Rows Removed by Filter: 177544  
Planning Time: 9.216 ms 
Execution Time: 82962.982 ms


```

You can see that the filter (ie times_seen and type=1) removes 177k rows. That's the reason this is timing out. I would recommend moving these filters into memory. One side effect of doing this is that you will sometimes end up with empty batches, which will cause you to exit early here: https://github.com/getsentry/sentry/blob/10847d44ccf3c72ec11feb915595ed60e7bf64c3/src/sentry/tasks/embeddings_grouping/utils.py#L199-L214

Instead, you should probably check for `batch_end_group_id > last_processed_group_id` or something along those lines.

jangjodi:
Thank you @wedamija! I've opened a [PR](https://github.com/getsentry/sentry/pull/76188) to address the changes. I only added the manual filter for times_seen, as running explain analyze with only the type filter was quick, but please let me know if you think I should filter that out manually as well.

Issue #76076, opened 2024-08-13
Backfill UNRESOLVED issues to set substatus


snigdhas:

snigdhas:
Migration ran successfully - https://redash.getsentry.net/queries/6888

PR #76082, opened 2024-08-13
chore(issues): Backfill UNRESOLVED groups with missing substatuses

snigdhas:
We need to run a backfill to fix the missing substatuses for UNRESOLVED groups. This [redash query](https://redash.getsentry.net/queries/6888) shows we have ~500 unresolved groups that have no substatus. We can backfill the groups by 
- setting the substatus to `NEW` if the group is first seen in the last 7 days
- setting the substatus to `REGRESSED` if there is a matching GroupHistory row in the past 7 days
- setting the substatus to `ONGOING` if no other info is present. In this case, we're assuming any status changes were older than 7 days in which case the auto_transition tasks would mark this group as `ONGOING`. 

Marking this as a post-deploy migration for safety, but we'll only be updating 500ish groups.

https://github.com/getsentry/sentry/issues/76076


wedamija (approved):
Migration lgtm - there are not many groups with this status and so the query is pretty quick and no need for the range wrapper

PR #76084, opened 2024-08-13
feat(grouping): Add `grouphashmetadata` table

lobsterkatie:
This adds a stub of the new `grouphashmetadata` table, which will allow us to track metadata about the grouping process - which config version generated the hash, whether it was based on stacktrace/message/etc, whether we used Seer as part of the process, and if so, which version of the model was used, and so on and so forth.

In this initial incarnation, all the table tracks is which `grouphash` record it's associated with, and when it was created.


src/sentry/models/grouphashmetadata.py
```diff
@@ -0,0 +1,18 @@
...
+from sentry.db.models import FlexibleForeignKey, Model, region_silo_model
+
+
+@region_silo_model
+class GroupHashMetadata(Model):
+    __relocation_scope__ = RelocationScope.Excluded
+
+    # GENERAL
+    grouphash = FlexibleForeignKey("sentry.GroupHash", unique=True, related_name="metadata")
+    date_added = models.DateTimeField(default=timezone.now)
```
wedamija, line 14: If you want, you can use `DefaultFieldsModel` and you'll get `date_updated` as well

wedamija (approved):
I don't have a strong opinion either way, but just asking - any reason we wouldn't store this data directly on the `GroupHash` table?

Migration lgtm

lobsterkatie:
> any reason we wouldn't store this data directly on the `GroupHash` table?

If it were just the timestamp, we probably would. It's definitely the simpler option. But in the end it's going to be a bunch of data, and there was concern about performance impacts and being able to wall it off from the main `GroupHash` table, which is so critical to ingestion. You can check out https://github.com/getsentry/sentry/pull/75980 to get a better sense of what all will eventually end up in the table. (There are more potential uses, but that's what's planned for this first version.)

wedamija:
> > any reason we wouldn't store this data directly on the `GroupHash` table?
> 
> If it were just the timestamp, we probably would. It's definitely the simpler option. But in the end it's going to be a bunch of data, and there was concern about performance impacts and being able to wall it off from the main `GroupHash` table, which is so critical to ingestion. You can check out #75980 to get a better sense of what all will eventually end up in the table. (There are more potential uses, but that's what's planned for this first version.)

Makes sense - Keep in mind that this table will become just as critical to ingestion, since I assume we will write to it in the same pipeline. But it's probably a good idea to keep it separate

src/sentry/models/grouphashmetadata.py
```diff
@@ -0,0 +1,18 @@
+from django.db import models
+from django.utils import timezone
+
+from sentry.backup.scopes import RelocationScope
+from sentry.db.models import FlexibleForeignKey, Model, region_silo_model
+
+
+@region_silo_model
+class GroupHashMetadata(Model):
+    __relocation_scope__ = RelocationScope.Excluded
```
armenzg, line 10: Do we want this?

armenzg (approved):
Looks good 👍🏻

src/sentry/models/grouphashmetadata.py
wedamija: Yeah, the problem is that this was introduced by someone that wanted to retrofit existing models, so they made the column nullable. We could make a new version that has them not null and is for use with new models only.

src/sentry/models/grouphashmetadata.py
lobsterkatie: Yes, I believe so. My understanding of the migration system is that while things like org members and projects and teams travel with you, your event/group/etc data does not. Also, the `GroupHash` table [is excluded](https://github.com/getsentry/sentry/blob/a0a0832ef9fab0a962a2276339ab09f346317a07/src/sentry/models/grouphash.py#L14-L15), so it stands to reason the associated metadata must be as well.

src/sentry/models/grouphashmetadata.py
lobsterkatie: Thanks for the suggestion, and TIL, but most of the data here will be added at creation time, and any updates will likely be tied to things which have their own timestamps (merges, for example), so I think we're good with just the `date_added` field.

PR #76090, opened 2024-08-13
feat(ai-issue-summary): Backend endpoint to generate and store summary

jennmueng:
Guarded behind a feature flag: `organizations:ai-summary`

1. If an issue summary exists, return that
2. If it doesn't, call seer to generate one and store it in the group data

Screenshot 2024-08-13 at 2 11 13 PM
Screenshot 2024-08-13 at 2 10 53 PM

### DIFF
--- a/src/sentry/api/endpoints/group_ai_summary.py
+++ b/src/sentry/api/endpoints/group_ai_summary.py
@@ -25,8 +25,6 @@ logger = logging.getLogger(__name__)
 
 from rest_framework.request import Request
 
-TIMEOUT_SECONDS = 60 * 30  # 30 minutes
-
 
 class SummarizeIssueResponse(BaseModel):
     group_id: str
@@ -40,7 +38,6 @@ class GroupAiSummaryEndpoint(GroupEndpoint):
         "GET": ApiPublishStatus.EXPERIMENTAL,
     }
     owner = ApiOwner.ML_AI
-    # go away
     private = True
     enforce_rate_limit = True
     rate_limits = {
--- a/src/sentry/api/endpoints/group_ai_summary.py
+++ b/src/sentry/api/endpoints/group_ai_summary.py
@@ -1,0 +1,127 @@
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+import orjson
+import requests
+from django.conf import settings
+from django.contrib.auth.models import AbstractBaseUser, AnonymousUser
+from pydantic import BaseModel
+from rest_framework.response import Response
+
+from sentry import eventstore, features
+from sentry.api.api_owners import ApiOwner
+from sentry.api.api_publish_status import ApiPublishStatus
+from sentry.api.base import region_silo_endpoint
+from sentry.api.bases.group import GroupEndpoint
+from sentry.api.serializers import EventSerializer, serialize
+from sentry.models.group import Group
+from sentry.seer.signed_seer_api import sign_with_seer_secret
+from sentry.types.ratelimit import RateLimit, RateLimitCategory
+
+logger = logging.getLogger(__name__)
+
+from rest_framework.request import Request
+
+TIMEOUT_SECONDS = 60 * 30  # 30 minutes
+
+
+class SummarizeIssueResponse(BaseModel):
+    group_id: str
+    summary: str
+    impact: str
+
+
+@region_silo_endpoint
+class GroupAiSummaryEndpoint(GroupEndpoint):
+    publish_status = {
+        "GET": ApiPublishStatus.EXPERIMENTAL,
+    }
+    owner = ApiOwner.ML_AI
+    # go away
+    private = True
+    enforce_rate_limit = True
+    rate_limits = {
+        "GET": {
+            RateLimitCategory.IP: RateLimit(limit=10, window=60),
+            RateLimitCategory.USER: RateLimit(limit=10, window=60),
+            RateLimitCategory.ORGANIZATION: RateLimit(
+                limit=30, window=60
+            ),  # TODO: Raise this limit when we move out of internal preview
+        }
+    }
+
+    def _get_event(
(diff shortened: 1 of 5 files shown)


src/sentry/api/endpoints/group_ai_summary.py
```diff
@@ -0,0 +1,132 @@
...
+from sentry.api.serializers.rest_framework.base import convert_dict_key_case, snake_to_camel_case
+from sentry.models.group import Group
+from sentry.seer.signed_seer_api import sign_with_seer_secret
+from sentry.types.ratelimit import RateLimit, RateLimitCategory
+
+logger = logging.getLogger(__name__)
+
+from rest_framework.request import Request
+
+TIMEOUT_SECONDS = 60 * 30  # 30 minutes
```
malwilley, line 28: Is this used anywhere?

src/sentry/api/endpoints/group_ai_summary.py
```diff
@@ -0,0 +1,132 @@
...
+    impact: str
+
+
+@region_silo_endpoint
+class GroupAiSummaryEndpoint(GroupEndpoint):
+    publish_status = {
+        "GET": ApiPublishStatus.EXPERIMENTAL,
+    }
+    owner = ApiOwner.ML_AI
+    # go away
```
malwilley, line 43: Seems a bit rude 🙃

src/sentry/api/endpoints/group_ai_summary.py
```diff
@@ -0,0 +1,132 @@
...
+                    "id": group.id,
+                    "title": group.title,
+                    "short_id": group.qualified_short_id,
+                    "events": [serialized_event],
+                },
+            },
+            option=orjson.OPT_NON_STR_KEYS,
+        )
+        response = requests.post(
+            f"{settings.SEER_AUTOFIX_URL}{path}",
```
malwilley, line 96: How long does this endpoint take to respond?

src/sentry/api/endpoints/group_ai_summary.py
```diff
@@ -0,0 +1,132 @@
...
+            )
+
+        serialized_event = self._get_event(group, request.user)
+
+        if not serialized_event:
+            return Response({"detail": "Could not find an event for the issue"}, status=400)
+
+        issue_summary = self._call_seer(group, serialized_event)
+
+        group.data.update({"issue_summary": issue_summary.dict()})
```
malwilley, line 127: I don't like that a GET endpoint causes a side effect. Might make sense to make this a POST instead? I assume we are doing this for testing only and that in the future we would be generating these summaries in a task elsewhere?

src/sentry/features/temporary.py
```diff
@@ -44,6 +44,8 @@ def register_temporary_features(manager: FeatureManager):
 
     # Enables activated alert rules
     manager.add("organizations:activated-alert-rules", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=True)
+    # Enable AI Issue Summary feture on the Issue Details page.
+    manager.add("organizations:ai-summary", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=True)
```
malwilley, line 48: nit: probably want to add `issue` prefix to make it more descriptive

src/sentry/api/endpoints/group_ai_summary.py
jennmueng: Good catch

src/sentry/api/endpoints/group_ai_summary.py
jennmueng: Haha it was copied over from some other class a long time ago

src/sentry/api/endpoints/group_ai_summary.py
jennmueng: 2-5 seconds

src/sentry/api/endpoints/group_ai_summary.py
jennmueng: I agree, and yes we will definitely be moving this off somewhere else, I just think this get makes the most sense for now given `useApiQuery` makes a get request

src/sentry/features/temporary.py
jennmueng: I omitted `issue` because I think it would be simpler to just use the same flag for all summaries together, such as session replay summaries etc as well.

malwilley: But would you be rolling out issue and replay summaries at the same time? If so this is good but if we want to roll them out at different times, would be better to have multiple

jennmueng: Hmm, given that any sub component summary like replay summaries will feed into the main issue summary, this one is like the main flag. I don't see a world where we would have sub component summaries but not the main issue summary for example, but I can see one where we have issue summary but not component summaries.

What do you think about the approach where this one controls all summaries in general and in the case that we do want to roll out separately, we have a separate `organizations:ai-replay-summary`

src/sentry/api/endpoints/group_ai_summary.py
malwilley: You actually can modify the method if you pass `method` in the query key object if you so choose

src/sentry/features/temporary.py
malwilley: That seems fine to me

src/sentry/api/endpoints/group_ai_summary.py
jennmueng: Oh I see, we can make it a POST then

jennmueng: @malwilley  Made it a POST on both frontend and backend. I know we typically don't like combining FE/BE changes in one PR but the FE is gated by the feature flag anyways, so the call won't be made until that flag is enabled.

malwilley (approved):

PR #76091, opened 2024-08-13
fix(ecosystem): Adds special handling for team custom field in Jira

GabeVillalobos:
Adds special handling for Jira's `team` custom field, which was previously passing invalid data to the issue create endpoint.

This is mostly a quick-fix to allow users to provide a Team ID, where previously, nothing would work. Ideally this should be turned into a validated field which queries available teams, but this work will be handled in a follow-up PR.


cathteng (approved):
for reference:
https://developer.atlassian.com/platform/teams/components/team-field-in-jira-rest-api/
Screenshot 2024-08-13 at 16 53 56

PR #76030, opened 2024-08-12
feat(alerts): Update Slack metric alert rule builder for dynamic alerts

ceorourke:
Update the metric alert rule message builder for Slack to show the detection type and to include the detection type in the referrer.

Mockup from block kit builder:
Screenshot 2024-08-12 at 3 54 02 PM

See https://github.com/getsentry/sentry/pull/76009 for email update
Partially closes https://getsentry.atlassian.net/browse/ALRT-147 (other integrations to follow)

PR #76093, opened 2024-08-13
feat(alerts): Update Discord and Slack unfurl message builder for dynamic alerts

ceorourke:
Follow up to https://github.com/getsentry/sentry/pull/76030 to update Discord and the Slack unfurl metric alerts message builder (they share the same code) for dynamic alerts. The main change is adding the threshold to the message and the detection type to the url params for tracking the clickthrough rate. 

Screenshot coming soon ~

### DIFF
--- a/src/sentry/integrations/metric_alerts.py
+++ b/src/sentry/integrations/metric_alerts.py
@@ -5,6 +5,7 @@ from django.db.models import Max
 from django.urls import reverse
 from django.utils.translation import gettext as _
 
+from sentry import features
 from sentry.constants import CRASH_RATE_ALERT_AGGREGATE_ALIAS
 from sentry.incidents.logic import get_incident_aggregates
 from sentry.incidents.models.alert_rule import AlertRule, AlertRuleThresholdType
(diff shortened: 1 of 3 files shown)


src/sentry/integrations/metric_alerts.py
```diff
@@ -173,9 +174,10 @@ def metric_alert_attachment_info(
     else:
         status = INCIDENT_STATUS[IncidentStatus.CLOSED]
 
-    url_query = None
+    url_query = {"detection_type": alert_rule.detection_type}
```
saponifi3d, line 177: 🤔 - could we lookup the detection type on this page based on the alert id / notification uuid? (not super important just curious 😅 - i could also see using this for click thru rates and stuff like that)

saponifi3d (approved):

src/sentry/integrations/metric_alerts.py
ceorourke: We could look it up with the alert rule id but why would we do that if we already have it here? And this is to be used for clickthrough rates

PR #76101, opened 2024-08-13
ref(replay): update canvas banner depending on sdk version

### LINKED ISSUE #76097: Replay: Canvas banner mentions SDK version when already supported version
[Example Replay](https://sentry.sentry.io/replays/1afdf1e8aefe430197059edb9223a192/?project=11276&t_main=tags)

The banner added for `canvas` mentions the supported SDK version. That can be confusing for customers already on a supported version, such as this replay, from SDK version 8 ([8.24.0](https://sentry.sentry.io/replays/?query=sdk.version%3A%228.24.0%22))

Image

## Suggested fix

Have a copy for unsupported SDK and one for supported SDK:

Supported:
> This Replay contains a `canvas` element. You can enable canvas recording, [learn how on our docs](docs)

Not supported:
> This Replay contains a `canvas` element. Please update the SDK to the latest version to enable canvas recording. [Learn more on our docs](docs)

michellewzhang:
fixes https://github.com/getsentry/sentry/issues/76097

sdk needs update:

SCR-20240813-mrym

sdk is updated:
SCR-20240813-lgme

### DIFF
--- a/static/app/components/replays/canvasSupportNotice.tsx
+++ b/static/app/components/replays/canvasSupportNotice.tsx
@@ -17,10 +17,11 @@ const LOCAL_STORAGE_KEY = 'replay-canvas-supported';
 /**
  * Displays a notice about canvas support if a <canvas> element is detected in the replay, but not recorded
  */
-export function CanvasSupportNotice({projectId}: {projectId: string | undefined}) {
+export function CanvasSupportNotice() {
   const organization = useOrganization();
   const {dismiss, isDismissed} = useDismissAlert({key: LOCAL_STORAGE_KEY});
   const {isFetching, replay} = useReplayContext();
+  const projectId = replay?.getReplay().project_id;
   const {needsUpdate} = useProjectSdkNeedsUpdate({
     minVersion: '7.98.0',
     organization,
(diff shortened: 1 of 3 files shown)


c298lee (approved):

jas-kas:
for this case:
> sdk needs update:
are we sure we don't want to include the minimum SDK version that Canvas is on? asking folks to all upgrade to the latest AKA v8.X might be high effort before they need to bump major versions (going from v7 to v8), go thru our migration guide etc

static/app/components/replays/canvasSupportNotice.tsx
```diff
@@ -10,16 +10,22 @@ import {space} from 'sentry/styles/space';
...
 import useOrganization from 'sentry/utils/useOrganization';
+import useProjectSdkNeedsUpdate from 'sentry/utils/useProjectSdkNeedsUpdate';
 
 const LOCAL_STORAGE_KEY = 'replay-canvas-supported';
 
 /**
  * Displays a notice about canvas support if a <canvas> element is detected in the replay, but not recorded
  */
-export function CanvasSupportNotice() {
+export function CanvasSupportNotice({projectId}: {projectId: string | undefined}) {
```
c298lee, line 20: we can also get the project in this function since we already use `useReplayContext` to get the replay!

michellewzhang: ooh good catch!

michellewzhang:
> for this case:
> 
> > sdk needs update:
> 
> are we sure we don't want to include the minimum SDK version that Canvas is on? asking folks to all upgrade to the latest AKA v8.X might be high effort before they need to bump major versions (going from v7 to v8), go thru our migration guide etc
> 
> https://docs.sentry.io/platforms/javascript/migration/v7-to-v8/

@jas-kas i updated the copy to include the min version! lmk what you think

c298lee (approved):

jas-kas (approved):

Issue #75636, opened 2024-08-06
[DevToolbar] Iterate on UI design


ryan953:
@Jesse-Box put together some tweaks to the UI elements to make things look cleaner and consistent. We should iterate on each panel to bring them in-line and improve the overall look

https://www.figma.com/design/G8sBIfdPBh7wwpKdEMjqcD/Specs%3A-Dev-Toolbar-v2?node-id=2005-2843&t=Rt0ZD4spLndPPPaN-0

aliu39:
https://www.figma.com/design/G8sBIfdPBh7wwpKdEMjqcD/Specs%3A-Dev-Toolbar-v2?node-id=0-1&t=P94FnkKNAfGyrmJt-0

aliu39:
Mini tasks:


- [ ] Ensure list item styles (height, font, layout) are consistent for every panel
- [ ] Update panel headers style to match the design (left-aligned, project badge)
- [ ] Feedback panel: change header button from "+" to "Report a bug" (see design)
- [ ] FF Panel: move clear overrides button from above search bar to the bottom of the "Overrides". Use x icon and "Remove all"
- [ ] FF Panel: make a "+ Add a Flag" button in header. 
   - does this replace the "Add override button" we currently have at the bottom?
- [ ] FF Panel: make toggles larger(?)
   - Contrary to the figma I think we should keep the true/false for clarity.
- [ ] FF Panel: remove the "Name Value" header

Do last:
- [ ] Replay/Web Vital/newer panels: brush up styles and ask Jesse for design thoughts

billyvg:
Is this list still up to date @aliu39?

aliu39:
@billyvg no I don't think so. @ryan953 and @c298lee would know best, this is for the internal poc toolbar so I'd guess it's low prio

ryan953:
I think we made the PoC look as good as it's going to get. Closing this.

PR #76102, opened 2024-08-13
style(toolbar): update feature flag panel css

michellewzhang:
- relates to https://github.com/getsentry/sentry/issues/75636
- modify feature flag css according to the figma. main changes:
    - move "clear overrides" button from above search bar to the bottom of the "overrides" tab. 
    - Use x icon and "remove all" for the clear overrides button
    - make a "+ add flag" button in header
    - remove the gray "name value" header

before:
SCR-20240813-mmtm
SCR-20240813-mmul

after:

https://github.com/user-attachments/assets/acc7ede1-3f03-4999-969c-973ac410dd8c


aliu39:
I know it's different from the figma but I like "+ Override" instead of "+ Add Flag" for the button.

Maybe the `+` icon could toggle to an up chevron `^` when the "Flag name to override" box is shown, to emphasize clicking again will hide the drop-down.

Nit: can disable + button when nothing is entered in the override box.

aliu39:
A design idea: what if clicking the + Add flag button didn't make a drop-down, but instead replaced the regular search bar:
Screenshot 2024-08-14 at 11 55 59 PM

with a special override search bar, button, and toggle?
Screenshot 2024-08-14 at 11 56 13 PM

The override input would filter the FF list too (excuse the bad mock-up):
Screenshot 2024-08-14 at 11 56 54 PM
Screenshot 2024-08-14 at 11 57 08 PM

Clicking the button in the header would toggle the search bar back to normal.  
This lets me see the current status of the flag I'm about to override, and whether it exists. It also cleans up the panel.

aliu39 (approved):
Aside from the nit all my feedback was design ideas/potential follow-ups, we can merge this

michellewzhang:
> what if clicking the + Add flag button didn't make a drop-down, but instead replaced the regular search bar with the override bar?

@aliu39 i like this idea! we should bring it up with @Jesse-Box and do in a followup if he agrees

especially this point: agree that it would be useful

> This lets me see the current status of the flag I'm about to override, and whether it exists.

michellewzhang:
@aliu39 
added the chevron solution for now: 
https://github.com/user-attachments/assets/cf19eff0-1a36-4007-a0fb-b46b340065eb

PR #76103, opened 2024-08-13
fix(weekly-report): user who has no projects should not see all projects

JoshFerge:
- If a user is not a member of an organization, or has no available projects, they should not receive the weekly report email.
- i assume the first case only happens during debugging
- the second can happen if the user is assigned only to a team that has no projects.

Fixes https://github.com/getsentry/sentry/pull/76103


nhsiehgit (approved):
🙏 bless

PR #76105, opened 2024-08-13
chore: add test for snuba deletion example

kylemumma:
This change adds a test of snuba deletion to the sentry codebase. It serves as an example, and instructions, for how to actually do a snuba delete from the sentry codebase. With this merged it represents product team having the ability to self-serve deletes.

* `test_deletes` is the actual test (entry point)
* the 3 requirements.txt files is updating the SDK to the latest version that supports Delete requests
* the `snuba.py` changes make the "send request to snuba" pipeline work with the new DeleteQuery sdk object. basically snuba.py is what actually sends the HTTP request to snuba.

Heres instructions for how to get the test to pass locally (it will likely fail by default)
1. `cd` to sentry repo, and do `sentry devservices down` followed by `sentry devservices up --exclude=snuba`
2. `cd` to snuba repo and do `snuba api --debug` leave this running
3. open a new terminal tab and do `snuba admin --debug` leave this running  (note: it only works when both are run in debug mode dont ask me why)
4. go to http://localhost:1219/#config
5. click "add new", give it the key `storage_deletes_enabled` and the value `1`, click save
6. now you should be able to run the sentry test and have it pass


src/sentry/testutils/cases.py
```diff
@@ -1385,6 +1385,15 @@ def store_spans(self, spans):
...
 
+    def store_issues(self, issues):
+        assert (
+            requests.post(
+                settings.SENTRY_SNUBA + "/tests/entities/search_issues/insert",
+                data=json.dumps(issues),
+            ).status_code
+            == 200
+        )
+
```
kylemumma, line 1396: this is a helper used by `test_deletes` to insert into the`search_issues` table

tests/snuba/test_deletes.py
```diff
@@ -0,0 +1,77 @@
...
+        ) == {self.project.id: 1}
+
+        # delete it
+        req = Request(
+            dataset="search_issues",
+            app_id="my_app",
+            query=DeleteQuery(
+                "search_issues", {"project_id": [self.project.id], "occurrence_id": [str(id)]}
+            ),
+        )
```
MeredithAnya, line 61: any reason you aren't adding the tenant_ids as well? I guess it's super important for the tests but since this test helps document what the delete query should look like might as well add them

tests/snuba/test_deletes.py
```diff
@@ -0,0 +1,77 @@
...
+            end=now + timedelta(days=1),
+            groupby=["project_id"],
+            filter_keys={"project_id": [self.project.id]},
+            referrer="testing.test",
+            tenant_ids={"referrer": "testing.test", "organization_id": 1},
+        ) == {self.project.id: 1}
+
+        # delete it
+        req = Request(
+            dataset="search_issues",
```
MeredithAnya, line 56: should we be using `dataset=Dataset.IssuePlatform` for the dataset and also adding search issues to the storage key in https://github.com/getsentry/sentry/blob/405923ba174dba4f71e4369796418f0b26d5f94b/src/sentry/snuba/dataset.py#L83
?

src/sentry/utils/snuba.py
```diff
@@ -1221,6 +1228,29 @@ def _snuba_query(
...
+) -> urllib3.response.HTTPResponse:
+    query = request.query
+    if not isinstance(query, DeleteQuery):
+        raise ValueError(
+            f"Expected request to contain a DeleteQuery but it was of type {type(request.query)}"
+        )
+
+    # Enter hub such that http spans are properly nested
+    with timer("delete_query"):
+        referrer = headers.get("referer", "unknown")
```
onkar, line 1242: Do headers have the key "referer" or "referrer" with an extra r? If it's the former, then would it be a big change to make referrer consistent everywhere?

onkar (commented):
LGTM 👍

tests/snuba/test_deletes.py
kylemumma: I added tenant ids to the test.

that makes sense, id assume it not being enforced yet is also the reason its not failing yet. but im not sure where the allocation policy enforcement actually happens in the code.

tests/snuba/test_deletes.py
kylemumma: done, thank you

src/sentry/utils/snuba.py
kylemumma: i looked into this a bit and its bc of someones design of a headers object. it may be possible to change but it wouldnt be trivial and would potentially be error prone. so im inclined to leave it how it is. (its like this in a few other places at least)

MeredithAnya (approved):

tests/snuba/test_deletes.py
```diff
@@ -0,0 +1,79 @@
...
+import pytest
+from snuba_sdk import DeleteQuery, Request
+
+from sentry.snuba.dataset import Dataset, StorageKey
+from sentry.testutils.cases import SnubaTestCase, TestCase
+from sentry.utils import snuba
+
+
+class SnubaTest(TestCase, SnubaTestCase):
+    @pytest.mark.xfail
```
armenzg, line 14: Why was this test marked as expected to fail?

tests/snuba/test_deletes.py
```diff
@@ -0,0 +1,79 @@
...
+            query=DeleteQuery(
+                StorageKey.SearchIssues.value,
+                {"project_id": [self.project.id], "occurrence_id": [str(id)]},
+            ),
+            tenant_ids={"referrer": "testing.test", "organization_id": 1},
+        )
+        snuba.raw_snql_query(req)
+
+        # make sure its gone
+        time.sleep(5)  # test will currently fail without the sleep (maybe it take time to delete?)
```
armenzg, line 67: This is not good practice. What can it be done to improve this?

src/sentry/snuba/dataset.py
```diff
@@ -82,3 +82,4 @@ class EntityKey(Enum):
 @unique
 class StorageKey(Enum):
     ProfileChunks = "profile_chunks"
+    SearchIssues = "search_issues"
```
armenzg, line 85: `search_issues` is already defined in Dataset and EntityKey. Do you know if we need it also defined as a storage key?

PR #76110, opened 2024-08-13
feat(profiling) add function metrics table

JonasBa:
First pass at adding function metrics table. This just renders a table with the function metrics and is missing time series and better linking to examples


static/app/views/profiling/content.tsx
```diff
@@ -478,7 +479,8 @@ function ProfilingTransactionsContent(props: ProfilingTabContentProps) {
         <ProfilingOnboardingCTA />
       ) : (
         <Fragment>
-          {organization.features.includes('profiling-global-suspect-functions') ? (
+          {organization.features.includes('profiling-global-suspect-functions') &&
```
Zylphrex, line 482: nit: this would be a little cleaner if you swap the ordering to
```tsx
{organization.features.includes('continuous-profiling-compat') ? (
  ...
) : organization.features.includes('profiling-global-suspect-functions') ? (
  ...
) : (
  ...
)}
```

static/app/views/profiling/landing/slowestFunctionsTable.tsx
```diff
@@ -0,0 +1,297 @@
...
+        </Tooltip>
+      </SlowestFunctionCell>
+      <SlowestFunctionCell>
+        <SlowestFunctionsProjectBadge
+          examples={props.function.examples}
+          projectsLookupTable={props.projectsLookupTable}
+        />{' '}
+      </SlowestFunctionCell>
+      <SlowestFunctionCell>
+        {formatAbbreviatedNumber(props.function.count)}
```
Zylphrex, line 198: I'm still on the fence about exposing this because increasing the date range will **not** actually load more samples.

Zylphrex (approved):

static/app/views/profiling/landing/slowestFunctionsTable.tsx
JonasBa: I think we can remove it before exposing it to users, right now its just useful to look at from a dev pov.

static/app/utils/profiling/hooks/useAggregateFlamegraphQuery.ts
```diff
@@ -69,12 +71,26 @@ export function useAggregateFlamegraphQuery(
         ...normalizeDateTimeParams(datetime ?? selection.datetime),
         dataSource,
         fingerprint,
+        expand: metrics ? 'metrics' : undefined,
```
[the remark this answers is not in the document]
JonasBa: Actually, if I do that, then it complains bc ts tries to be strict about the type, but let me change the index signature here

static/app/views/profiling/landing/slowestFunctionsTable.tsx
JonasBa: I'm going to add a todo

PR #76111, opened 2024-08-13
ref(querybuilder): Remove params from snuba modules

wmak:
- Removes all the params from snuba modules for snuba_params instead


wmak:
closing for https://github.com/getsentry/sentry/pull/76198

PR #76112, opened 2024-08-13
ref: fix typing for project_artifact_bundle_file_details

asottile-sentry:

### DIFF
--- a/src/sentry/api/endpoints/project_artifact_bundle_file_details.py
+++ b/src/sentry/api/endpoints/project_artifact_bundle_file_details.py
@@ -3,7 +3,7 @@ import binascii
 import posixpath
 
 import sentry_sdk
-from django.http.response import FileResponse
+from django.http.response import FileResponse, HttpResponseBase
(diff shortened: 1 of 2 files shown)


src/sentry/api/endpoints/project_artifact_bundle_file_details.py
```diff
@@ -36,7 +36,8 @@ def download_file_from_artifact_bundle(
             ClosesDependentFiles(fp, archive),
             content_type=headers.get("content-type", "application/octet-stream"),
         )
-        response["Content-Length"] = file_info.file_size if file_info is not None else None
+        if file_info is not None:
+            response["Content-Length"] = file_info.file_size
```
asottile-sentry, line 40: another win for mypy, this was producing  `Content-Length: None\r\n` in the response 🙃

joshuarli: noiceee

joshuarli (approved):

PR #76114, opened 2024-08-13
feat(discover): Update events and events-stats endpoints to skip metrics layer on insights based queries

edwardgou-sentry:
Sets `insights_metrics_override_metric_layer` to true on discover `metrics_performance` events and events-stats endpoints which allows skipping the metrics layer when an unsupported query (insights) is detected


edwardgou-sentry:
Closing because I don't think this is actually needed at the moment

PR #76085, opened 2024-08-13
feat(onboarding): frontend for disableMemberInvite flag

ameliahsu:
add toggle for org admin to allow/disallow members from inviting teammates to their org

Admin view:
admin_view

Toggle is disabled for non-admin members and organizations that are not on a business plan:
member_view

PR #76116, opened 2024-08-13
feat(onboarding): migration to add `disable_member_invite` bit flag

ameliahsu:
add `disable_member_invite` flag for org admin to allow/disallow members from inviting teammates to their org

frontend: https://github.com/getsentry/sentry/pull/76085


src/sentry/models/organization.py
```diff
@@ -195,6 +195,9 @@ class flags(TypedClassBitField):
         # Enable codecov integration.
         codecov_access: bool
 
+        # Disable org-members from inviting members
+        disable_member_invite: bool
```
iamrajjoshi, line 199: we need to make sure a new flag in the bitfield is at the bottom of the field.

src/sentry/models/organization.py
```diff
@@ -201,6 +201,9 @@ class flags(TypedClassBitField):
         # Prevent superuser access to an organization
         prevent_superuser_access: bool
 
+        # Disable org-members from inviting members
+        disable_member_invite: bool
```
cathteng, line 205: do we want disable to be true or false by default? the default for a bitfield is false

ameliahsu: @cathteng by default we want members to invite other members, so we want this flag to be false by default

sentaur-athena: yes, default should be false.

sentaur-athena: We have a feature flag for launching it gradually but default is false unless they turn it to true to disable.

wedamija (commented):
migration lgtm

iamrajjoshi (approved):

PR #76134, opened 2024-08-13
ref(control_silo): Move UserEmail model to users module

Christinarlong:
Part of moving control silo user related resources into the users module. Includes adding any missing types for the model.

Apart of (#73856)


src/sentry/users/models/useremail.py
```diff
@@ -0,0 +1,166 @@
...
+        ]
+
+    def set_hash(self) -> None:
+        self.date_hash_added = timezone.now()
+        self.validation_hash = get_secure_token()
+
+    def hash_is_valid(self) -> bool:
+        return bool(
+            self.validation_hash and self.date_hash_added > timezone.now() - timedelta(hours=48)
+        )
```
Christinarlong, line 98: coerced into bool to get return type to be only `bool` instead of `bool | str`

markstory (approved):

PR #76139, opened 2024-08-13
fix(self-hosted): SOURCE_COMMIT

joshuarli:
This env var is read by `sentry.get_revision`. There are at least 3 self-hosted issue reports of people being confused by the frontend showing somethign like "24.3.0 unknown". Let's make that `24.3.0 self hosted`.

As far as I can tell the sentry version is cosmetic (it isn't validated by anything / doesn't crash the cli):

```
SENTRY_BUILD="self hosted" sentry --version
sentry, version 24.8.0.dev0+self hosted
```


joshuarli:
oh actually this might be coming from `__version__ = VERSION` (which also defaults to unknown), one sec i need to actually take a closer look

aldy505:
To be honest, this isnt a big deal at all, you don't have to fix this...

asottile-sentry:
the actual fix is in this line, funny typo on my part https://github.com/getsentry/sentry/blob/7e9e36ac462726aad43a0d7cc633828a910b0656/.github/workflows/self-hosted.yml#L88

joshuarli:
> the actual fix is in this line, funny typo on my part
> 
> https://github.com/getsentry/sentry/blob/7e9e36ac462726aad43a0d7cc633828a910b0656/.github/workflows/self-hosted.yml#L88

lol https://github.com/moby/buildkit/issues/2128

hubertdeng123 (approved):

aldy505 (approved):

PR #76140, opened 2024-08-13
feat(uptime): Mark uptime monitors as failed after two consecutive checks

wedamija:
This builds the concept of consecutive checks into the uptime results consumer. We'll only mark the monitor failed after we've had two consecutive failed checks. This also adds the ability for consecutive recovery checks into the consumer, but for now we'll just leave this at 1 to avoid making it look like the url is down longer than it really is.


evanpurkhiser (approved):
Looks good

PR #76142, opened 2024-08-13
fix(devservices): explicitly use repo-local devenv

joshuarli:
we should be explicitly referencing the repo-local devenv here, otherwise there's a chance it could refer to the global devenv (which generally lags behind in being updated)

### DIFF
--- a/src/sentry/runner/commands/devservices.py
+++ b/src/sentry/runner/commands/devservices.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import contextlib
+import functools
 import http
 import json  # noqa
 import os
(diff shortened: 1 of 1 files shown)


src/sentry/runner/commands/devservices.py
```diff
@@ -51,6 +52,22 @@
...
+def _gitroot() -> str:
+    from os.path import abspath
+    from subprocess import CalledProcessError, run
+
+    try:
+        proc = run(("git", "rev-parse", "--show-cdup"), check=True, capture_output=True)
+        root = abspath(proc.stdout.decode().strip())
+    except CalledProcessError:
+        raise SystemExit(
+            "git failed. Is it installed, and are you in a Git repository " "directory?",
```
nirajdewani, line 66:
```suggestion
            "git failed. Is it installed, and are you in a Git repository directory?",
```

nirajdewani (approved):

src/sentry/runner/commands/devservices.py
```diff
@@ -65,9 +82,11 @@ def _client() -> ContextManager[docker.DockerClient]:
             if DARWIN:
                 if USE_COLIMA:
                     click.echo("Attempting to start colima...")
+                    gitroot = _gitroot()
                     subprocess.check_call(
                         (
-                            "devenv",
+                            # explicitly use repo-local devenv, not the global one
+                            f"{gitroot}/.venv/bin/devenv",
```
asottile-sentry, line 89: better would be `sys.executable -m devenv` -- but we never finished the migration to `.devenv` or whatever for the configuration

PR #76146, opened 2024-08-14
feat(alerts): Update metric alert rule template for dynamic alerts

ceorourke:
Follow up to https://github.com/getsentry/sentry/pull/76030/ which moves the code into a shared location that updates Slack, MSTeams, OpsGenie, and Pagerduty. I had it the other way before to control the order of the content in the Slack notification, but it's easier this way. This updates all the aforementioned metric alert notifications to include the threshold text.

### DIFF
--- a/tests/sentry/incidents/action_handlers/test_msteams.py
+++ b/tests/sentry/incidents/action_handlers/test_msteams.py
@@ -2,13 +2,21 @@ import time
 from unittest.mock import patch
 
 import responses
+from urllib3.response import HTTPResponse
 
-from sentry.incidents.models.alert_rule import AlertRuleTriggerAction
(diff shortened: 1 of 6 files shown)


tests/sentry/incidents/action_handlers/test_msteams.py
```diff
@@ -74,6 +82,76 @@ def run_test(self, incident, method):
             incident, IncidentStatus(incident.status), metric_value
         )
 
+    @responses.activate
+    def test_build_incident_attachment(self):
```
ceorourke, line 86: This was just missing entirely which is what I based the dynamic alert tests on in other test files so I created this

nhsiehgit (approved):

PR #76147, opened 2024-08-14
feat(uptime): Change interval_seconds for active autodetected monitors to 60

wedamija:
This migration updates the interval_seconds for all active autodetected monitors to 60. Since we set `status` to CREATING, these rows will be picked up by the `subscription_checker` within 20 minutes and the updated config will be pushed to the uptime checker.


tests/sentry/migrations/test_0007_update_detected_subscription_interval.py
```diff
@@ -0,0 +1,44 @@
...
+    def setup_before_migration(self, apps):
+        self.uptime_subscription = self.create_uptime_subscription(
+            url="http://sintry.io", interval_seconds=300
+        )
+        self.create_project_uptime_subscription(
+            uptime_subscription=self.uptime_subscription,
+            mode=ProjectUptimeSubscriptionMode.AUTO_DETECTED_ACTIVE,
+        )
+
+        self.uptime_subscription_unchanged = self.create_uptime_subscription(
```
evanpurkhiser, line 19:
```suggestion
        # Will not change because it is in onboarding mode
        self.uptime_subscription_unchanged = self.create_uptime_subscription(
```

evanpurkhiser (approved):

tests/sentry/migrations/test_0007_update_detected_subscription_interval.py
wedamija: I'm just going to leave as-is so I can merge, usually these migration tests are disabled a short while after merging since they break anyway

PR #76148, opened 2024-08-14
fix(breadcrumbs): Use subgrid to align log levels

scttcper:
attempts to fix this but i think it might drive the virtual scroll crazy https://www.notion.so/sentry/Breadcrumb-level-horizontal-alignment-f7e13defa53d40d7a0af8aa6c73ea085
image


leeandher (approved):
Thanks!

PR #76151, opened 2024-08-14
fix: fix bug with IndexError

### LINKED ISSUE #75667: Intermittent Internal Error when resolving issues through API
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

1. Use REST API to resolve an issue with  request `PUT /api/0/organizations/{organization_id_or_slug}/issues/{issue_id}/` and payload:

```json
{ "status": "resolvedInNextRelease" }
```
2. The REST API sometimes return an internal error response:

```json
"detail": "Internal Error",
"errorId": "0a1a6e14bffc46d182507727900c5f3c"
 ```

### Expected Result

The API should return a 200 result in case of successful issue update, or a proper error code in case of invalid input.

### Actual Result

The API returns an Internal Error. Support has it that this error has been triggered by a `list index out of range` error on https://github.com/getsentry/sentry/blob/957585118dfd25132769bccfdb8d9828ab7eb718/src/sentry/api/helpers/group_index/update.py#L275

### Product Area

APIs

### Link

_No response_

### DSN

_No response_

### Version

_No response_

scefali:
This fixes a bug trying to `resolveInNextRelease` when no release has been created yet. In this case, we just mark the issue as resolved without a `GroupResolution`. The alternative would be to throw a 400 error saying there are no releases yet but I think this is probably more graceful.

Fixes https://github.com/getsentry/sentry/issues/75667


armenzg (approved):
Thank you @scefali !

scefali:
@armenzg thank you for the review 🚀

PR #76150, opened 2024-08-14
fix(ts): Codemod all `sentry/types` warnings

evanpurkhiser:
Wrote this codemod: https://gist.github.com/evanpurkhiser/83f2f0098f7b215894342903bd3b02d6

PR #76154, opened 2024-08-14
ref(js): Drop types/index.tsx barrel file

evanpurkhiser:


priscilawebdev:
why are we doing this? Is it something we no longer need?

biome.json
```diff
@@ -40,9 +40,7 @@
         "noRestrictedImports": {
           "level": "warn",
           "options": {
-            "paths": {
-              "sentry/types": "Please import directly. For example: import type {Organization} from 'sentry/types/organization'"
-            }
+            "paths": {}
           }
         }
```
scttcper, line 45: i think its no longer used

scttcper (approved):

evanpurkhiser:
> why are we doing this? Is it something we no longer need?

A while back @anonrig found that avoiding this barrel file improved typescript performance (?) and started moving everything to imports from their sepcific type files. I finished that in #76150 so we can delete this file

biome.json
evanpurkhiser: was just gonna leave it if we wanted to easily add more, idk

PR #76166, opened 2024-08-14
fix(form-model): Add validation for the type boolean

priscilawebdev:
**Before**

https://github.com/user-attachments/assets/1253796a-b9bc-47e5-87db-2426e503644a

**After**

https://github.com/user-attachments/assets/7feb9cb9-cee9-4b13-8798-1fa512aa2818


static/app/components/forms/model.tsx
```diff
@@ -382,6 +382,10 @@ class FormModel {
       return value.length > 0;
     }
 
+    if (typeof value === 'boolean') {
+      return value === true;
```
priscilawebdev, line 386: I am wondering if we should check for `value === "true"` here too 🤔

michellewzhang (approved):

PR #76174, opened 2024-08-14
fix(stats): Correct displayed names

priscilawebdev:
It is better to return the `apiName` elsewhere for cases where spans should be referred to as either 'span' or 'indexed_spans'. The reasoning behind this decision is that we use plural forms and display names in the UI, and we prefer not to expose whether a span is indexed or not.

before:

image

after:

image

### DIFF
--- a/static/app/views/organizationStats/usageStatsPerMin.tsx
+++ b/static/app/views/organizationStats/usageStatsPerMin.tsx
@@ -11,6 +11,7 @@ import {formatUsageWithUnits, getFormatUsageOptions} from './utils';
 
 type Props = {
   dataCategory: DataCategoryInfo['plural'];
+  dataCategoryApiName: DataCategoryInfo['apiName'];
   organization: Organization;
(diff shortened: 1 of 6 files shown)


static/app/views/organizationStats/usageStatsPerMin.tsx
```diff
@@ -52,17 +58,21 @@ function UsageStatsPerMin({dataCategory, organization, projectIds}: Props) {
     return null;
   }
 
+  const category =
+    dataCategoryApiName === 'span_indexed' ? dataCategoryApiName : dataCategory;
+
```
priscilawebdev, line 63: also this shall be fine as we are filtering for 'accepted' only but just in case this corrects the category name.

scttcper (approved):
i think this was only used on the stats page

priscilawebdev:
@scttcper yes it is only used on the stats page! I added it today :)

PR #76175, opened 2024-08-14
feat(search): Using new search query builder in discover.

Abdkhan14:
### Keys
- The filter keys are specific to the dataset, if selected. 
### Recent Searches
- The recent searches are specific to the dataset, if selected. Here's the stacked PR for this logic: https://github.com/getsentry/sentry/pull/76186
### Aggregation
- The aggregation filters work and default parameters and value types are assigned.
### Grouping
- I've added some grouping to keep filter key section counts within `5`, aligning with the issues team. We'll update the grouping soon. 

Note: Not sure why we have `discover/fields.tsx` and `discover/fields/index.ts`. I feel like they could just be one file, but that's out of the scope of this PR. I've adapted the parameters from `discover/fields.tsx/AGGREGATIONS` in`discover/fields/index.ts/AGGREGATION_FIELDS` since that's where the search query builder looks for parameters and the types of parameters differ between the two consts.

Screenshot 2024-08-14 at 12 00 15 PM


static/app/utils/fields/index.ts
```diff
@@ -308,128 +310,443 @@ export interface FieldDefinition {
...
   },
   [AggregationKey.COUNT_IF]: {
     desc: t('Count of events matching the parameter conditions'),
     kind: FieldKind.FUNCTION,
     valueType: FieldValueType.NUMBER,
+    parameters: [
+      {
+        name: 'column',
+        kind: 'column',
+        columnTypes: validateAndDenyListColumns(
```
malwilley, line 407: Does this function need to filter out aggregate keys, or should we modify SearchQueryBuilder to automatically do that?

CleanShot 2024-08-14 at 11 36 14

static/app/utils/fields/index.ts
```diff
@@ -308,128 +310,443 @@ export interface FieldDefinition {
...
+        ]),
+        defaultValue: 'transaction.duration',
+        required: true,
+      },
+    ],
   },
   [AggregationKey.MAX]: {
     desc: t('Returns maximum value of the selected field'),
     kind: FieldKind.FUNCTION,
     valueType: null,
```
malwilley, line 508: Should the valueType of this (and other similar functions) be number? I'm seeing a problem when I try to use it (probably because it defaults to string?):

CleanShot 2024-08-14 at 11 41 03

static/app/views/discover/resultsSearchQueryBuilder.tsx
```diff
@@ -0,0 +1,297 @@
...
+  }, [
+    measurements,
+    dataset,
+    customMeasurements,
+    functionTags,
+    filteredTags,
+    organization.features,
+  ]);
+
+  const filterKeySections = useMemo(() => {
```
malwilley, line 258: This can be sorted out later, but I wonder if we should have a "Functions" section for things like `count()` and such

static/app/views/discover/resultsSearchQueryBuilder.tsx
```diff
@@ -0,0 +1,297 @@
...
+      organization,
+      selection.datetime,
+      projectIds,
+      includeTransactions,
+      includeSessionTagsValues,
+      dataset,
+    ]
+  );
+
+  const getTagList: TagCollection = useMemo(() => {
```
malwilley, line 218: nit: `getTagList` sounds like a function but this is a object. Probably just call `filterKeys`?

static/app/utils/fields/index.ts
Abdkhan14: Yup this makes sense. Lets set valuetype to number.

static/app/views/discover/resultsSearchQueryBuilder.tsx
Abdkhan14: Good idea, will suggest this to my teammates.

static/app/utils/fields/index.ts
Abdkhan14: @malwilley can you elaborate a bit on how we can `modify SearchQueryBuilder to automatically` to ensure the valueType of the column is in a certain list and exclude some columns with in a certain list.

static/app/utils/fields/index.ts
Abdkhan14: Actually @malwilley, it's a bit odd to be setting valueType to Number here:
Screenshot 2024-08-14 at 4 16 03 PM

If we were able to provide an array of possible valueTypes and show all the options, or get valueType from the column selected as the parameter it would be ideal.

Abdkhan14: The only option now is to pass valueType as `null` and it defaults to string.

malwilley: Ah, I see. So the value should be dependent on the selected column?

I will follow up by adding support for this. I think I'll need to accept something like:

```
dynamicValueType: (parameters: string[]) => ValueType
```

If I add the above, would it be sufficient?

static/app/utils/fields/index.ts
malwilley: Yeah! So the two options are:

1. `validateAndDenyListColumns` could return null for things like `count`, `count_if`, etc
2. `SearchQueryBuilder` could filter out all aggregate filter keys before passing to the `columnTypes` function

Option 1 is more flexible, but option 2 would be good if we _never_ want aggregate keys to be suggested as parameter values.

Abdkhan14: Ahh I see @malwilley , I misunderstood the screen shot. Definitely go for option 2. We never expect aggregates as parameters.

malwilley: Great, I will also follow up with that change!

malwilley (approved):
Looks great, thank you!

static/app/utils/fields/index.ts
Abdkhan14: Yeah I think something like that can work or maybe a `isDynamicValueType` boolean that looks through columns provided in parameters and infers the type for the value.

static/app/components/searchQueryBuilder/index.spec.tsx
```diff
@@ -396,6 +396,36 @@ describe('SearchQueryBuilder', function () {
...
 
+      it('does not display recent filters that are not valid filter keys', async function () {
+        MockApiClient.addMockResponse({
+          url: '/organizations/org-slug/recent-searches/',
+          body: [
+            {query: 'assigned:me'},
+            {query: 'assigned:me browser.name:firefox'},
+            {query: 'assigned:me browser.name:firefox is:unresolved'},
+            // Level is not a valid filter key
+            {query: 'assigned:me browser.name:firefox is:unresolved level:error'},
```
malwilley, line 407: For this test, can you remove the first few queries and simplify this one to `assigned:me level:error`? We only return the top 3 filters now, so while the test passes I don't think it would fail if you took out the new logic new logic.

Abdkhan14: Got it, will make the change.

PR #76176, opened 2024-08-14
ref(issue-details): Manage event details state better

leeandher:
This PR refactors quite a bit. The goal was to create a system for rendering the navigation buttons that are conditional on the contents of the page, and to add the ability for sections to open when navigated to. The buttons will now only appear if a corresponding data section on the page is rendered.

To accomplish this a few things have been changed:
- There is a context provider at the top of the page for passing data about the current state of the page around. 
- Since it was getting a bit complicated I decided to use a reducer for simplifying the calls to mutate the sections.
- For clean up, the section keys are now called `SectionKey` instead of mentioning `Fold` which is just their styling, and moved to the context file.

<details>
<summary> Outdated Notes </summary>
A few notes:
- I'm not sure about the performance of this approach. Since every section is now a consumer they all update the moment any of the shared state changes, which is from navigating or opening/closing a section.
- It also seems to rerender the children of these sections (I believe) when the updates are made, which can make expensive renders on varying parts of the page repeat when unrelated sections are interacted with. 

If anyone has tips for the above, please let me know ^
</details>

https://github.com/user-attachments/assets/743d610c-a98e-4e37-814a-e3e7178ae2c5


malwilley:
@leeandher here's a couple options that could help with the performance:

1. Use `useSyncedLocalStorageState` instead of storing the state in context. This hook was made for this purpose, where the state lives in localStorage, but multiple components can use and update the state and stay in sync. If this works it could be a much simpler solution?

2. If the above isn't possible, I'd extract out the FoldSection contents to a new component wrapped in `memo` so that it only rerenders when the props actually change. That should prevent all the components from rerendering when the context changes.

static/app/views/issueDetails/streamline/context.tsx
```diff
@@ -0,0 +1,72 @@
...
+  key: SectionKey;
+  isOpen?: boolean;
+}
+
+export interface EventDetailsContextType extends EventDetailsState {
+  dispatch: Dispatch<EventDetailsActions>;
+}
+
+export const EventDetailsContext = createContext<EventDetailsContextType>({
+  searchQuery: '',
```
malwilley, line 65: Looks like this already existed, so sorry for commenting on something not related to this PR! But I'm wondering why this is in context? Seems like we should be storing the query in the URL params?

static/app/views/issueDetails/streamline/useEventDetailsReducer.tsx
```diff
@@ -0,0 +1,84 @@
...
+        case 'UPDATE_SEARCH_QUERY':
+          return {...state, searchQuery: action.searchQuery};
+        default:
+          return state;
+      }
+    },
+    []
+  );
+
+  const [state, dispatch] = useReducer(reducer, initialState);
```
scttcper, line 78: these could probably use better names

static/app/views/issueDetails/streamline/useEventDetailsReducer.tsx
```diff
@@ -0,0 +1,84 @@
+import {type Reducer, useCallback, useReducer} from 'react';
+
+import type {
+  SectionConfig,
+  SectionKey,
+} from 'sentry/views/issueDetails/streamline/context';
```
scttcper, line 6: is it possible to put this stuff inside this file. I don't like that i need to look in two files just to understand what is inside the EventDetailsState

static/app/views/issueDetails/streamline/useEventDetailsReducer.tsx
```diff
@@ -0,0 +1,84 @@
...
+  key: SectionKey;
+  type: 'UPDATE_SECTION_CONFIG';
+};
+
+function updateSectionConfig(
+  state: EventDetailsState,
+  sectionKey: SectionKey,
+  updatedConfig: Partial<SectionConfig>
+): EventDetailsState {
+  const nextState = {
```
scttcper, line 31:
```suggestion
  const nextState: EventDetailsState = {
```

static/app/views/issueDetails/streamline/useEventDetailsReducer.tsx
```diff
@@ -0,0 +1,84 @@
...
+    ...state,
+    sectionData: {
+      ...state.sectionData,
+    },
+  };
```
scttcper, line 36: its called nextState but doesn't seem to include the changes, i think it might make sense to create the new state here

static/app/views/issueDetails/streamline/foldSection.tsx
```diff
@@ -91,18 +43,37 @@ export const FoldSection = forwardRef<HTMLElement, FoldSectionProps>(function Fo
...
+    dispatch({
+      type: 'UPDATE_SECTION_CONFIG',
+      key: sectionKey,
+      config: {
+        ...config,
+        isOpen: isInitiallyCollapsed,
+      },
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
scttcper, line 75: might be able to get away without a useeffect here since this is also where it gets registered?
```suggestion
 if (!sectionData.hasOwnProperty(sectionKey)) {
    dispatch({
      type: 'UPDATE_SECTION_CONFIG',
      key: sectionKey,
      config: {
        ...config,
        isOpen: isInitiallyCollapsed,
      },
    });
}
```

static/app/views/issueDetails/streamline/context.tsx
leeandher: Hmm, bit of an oversight, you're right. It makes sense to keep it in the URL so filtered pages can be shared 👍

static/app/views/issueDetails/streamline/useEventDetailsReducer.tsx
leeandher: I just called it that because I mutate the object right after, so by the end `return nextState` makes sense.

scttcper (approved):

PR #76177, opened 2024-08-14
fix(features) Collect random samples of error from feature flag failures

markstory:
When feature flag handler fail we can end up creating a ton of errors due to how frequently features are checked.

Random sampling should still give us enough signal to know feature flagging is having problems but reduce the chances of causing additional strain on our systems.

### DIFF
--- a/src/sentry/features/manager.py
+++ b/src/sentry/features/manager.py
@@ -12,13 +12,14 @@ from typing import TYPE_CHECKING, Any
 import sentry_sdk
 from django.conf import settings
 
-from sentry.options import FLAG_AUTOMATOR_MODIFIABLE, register
+from sentry import options
 from sentry.users.services.user.model import RpcUser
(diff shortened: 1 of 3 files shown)


src/sentry/features/manager.py
```diff
@@ -102,30 +103,34 @@ def has_for_batch(
...
+                            remaining.remove(obj)
+                            result[obj] = flag
+                    span.set_data("Flags Found", batch_size - len(remaining))
+
+            default_flag = settings.SENTRY_FEATURES.get(name, False)
+            for obj in remaining:
+                result[obj] = default_flag
+        except Exception as e:
+            if in_random_rollout("features.error.capture_rate"):
+                sentry_sdk.capture_exception(e)
```
markstory, line 133: Failures in `has_for_batch` shouldn't leak into application logic.

wedamija (approved):

PR #76180, opened 2024-08-14
feat(explore): Add hooks to get the explore query options

Zylphrex:
These hooks will fetch the current state of the user specified options.


wmak (approved):
lgtm

PR #76185, opened 2024-08-14
feat(insights): implement geo region selector in web vitals and assets

DominikB2014:
Work toward #75230 

Adds geo selector in the web vitals and resource module and ensure all queries gets the filter applied

NOTE: not all the metrics are properly tagged yet, so web vitals doesn't work entierely yet, and the resource module doesn't have geo data for resource size metrics. Its all behind a feature flag, so this not an issue that customers can see.

### DIFF
--- a/static/app/views/insights/browser/resources/components/resourceView.tsx
+++ b/static/app/views/insights/browser/resources/components/resourceView.tsx
@@ -36,6 +36,7 @@ const {
   SPAN_DOMAIN,
   TRANSACTION,
   RESOURCE_RENDER_BLOCKING_STATUS,
+  USER_GEO_SUBREGION,
 } = BrowserStarfishFields;
 
 type Option = {
@@ -52,7 +53,12 @@ function ResourceView() {
     ...(filters[SPAN_DOMAIN] ? {[SPAN_DOMAIN]: filters[SPAN_DOMAIN]} : {}),
   };
 
-  const extraQuery = getResourceTypeFilter(undefined, DEFAULT_RESOURCE_TYPES);
+  const extraQuery = [
+    ...getResourceTypeFilter(undefined, DEFAULT_RESOURCE_TYPES),
+    ...(filters[USER_GEO_SUBREGION]
+      ? [`user.geo.subregion:[${filters[USER_GEO_SUBREGION]}]`]
+      : []),
+  ];
 
   return (
     <Fragment>
--- a/static/app/views/insights/browser/resources/components/resourceView.tsx
+++ b/static/app/views/insights/browser/resources/components/resourceView.tsx
@@ -56,7 +56,7 @@ function ResourceView() {
   const extraQuery = [
     ...getResourceTypeFilter(undefined, DEFAULT_RESOURCE_TYPES),
     ...(filters[USER_GEO_SUBREGION]
-      ? [`user.geo.subregion:[${filters[USER_GEO_SUBREGION]}]`]
+      ? [`user.geo.subregion:[${filters[USER_GEO_SUBREGION].join(',')}]`]
       : []),
   ];
 
--- a/static/app/views/insights/browser/resources/utils/useResourceFilters.ts
+++ b/static/app/views/insights/browser/resources/utils/useResourceFilters.ts
@@ -1,8 +1,11 @@
 import pick from 'lodash/pick';
 
+import {decodeList} from 'sentry/utils/queryString';
 import {useLocation} from 'sentry/utils/useLocation';
 import type {ResourceSpanOps} from 'sentry/views/insights/browser/resources/types';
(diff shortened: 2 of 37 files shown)


static/app/views/insights/browser/resources/components/resourceView.tsx
```diff
@@ -52,7 +53,12 @@ function ResourceView() {
     ...(filters[SPAN_DOMAIN] ? {[SPAN_DOMAIN]: filters[SPAN_DOMAIN]} : {}),
   };
 
-  const extraQuery = getResourceTypeFilter(undefined, DEFAULT_RESOURCE_TYPES);
+  const extraQuery = [
+    ...getResourceTypeFilter(undefined, DEFAULT_RESOURCE_TYPES),
+    ...(filters[USER_GEO_SUBREGION]
+      ? [`user.geo.subregion:[${filters[USER_GEO_SUBREGION]}]`]
```
KevinL10, line 59: Is this missing a `.join(',')`?

static/app/views/insights/browser/resources/utils/useResourceFilters.ts
```diff
@@ -22,17 +26,27 @@ export type ModuleFilters = {
   [BrowserStarfishFields.SPAN_OP]?: ResourceSpanOps;
   [BrowserStarfishFields.TRANSACTION]?: string;
   [BrowserStarfishFields.SPAN_DOMAIN]?: string;
+  [SpanMetricsField.USER_GEO_SUBREGION]?: SubregionCode[];
```
mjq, line 29: Nit: Since `BrowserStarfishFields.USER_GEO_SUBREGION` was added, it feels like it would be more consistent to use that here. If the whole thing should be replaced with `SpanMetricsField` instead, that can be done in one shot later. Just a bit easier to follow IMO.

static/app/views/insights/browser/resources/utils/useResourceFilters.ts
```diff
@@ -22,17 +26,27 @@ export type ModuleFilters = {
...
     BrowserStarfishFields.DESCRIPTION,
     BrowserStarfishFields.RESOURCE_RENDER_BLOCKING_STATUS,
   ]);
+
+  const subregions = decodeList(
+    location.query[SpanMetricsField.USER_GEO_SUBREGION]
+  ) as SubregionCode[];
+  if (subregions.length) {
+    filters[SpanMetricsField.USER_GEO_SUBREGION] = subregions;
+  }
```
mjq, line 49: ditto re: `BrowserStarfishFields.USER_GEO_SUBREGION` use

static/app/views/insights/browser/webVitals/components/tables/pagePerformanceTable.tsx
```diff
@@ -67,6 +71,9 @@ export function PagePerformanceTable() {
 
   const query = decodeScalar(location.query.query, '');
   const browserTypes = decodeBrowserTypes(location.query[SpanIndexedField.BROWSER_NAME]);
+  const subregions = decodeList(
```
KevinL10, line 74: I think it would make sense to use a custom query decoder for all of these places we're decoding the subregions -- similar to `decodeBrowserTypes` -- so that we can guarantee the regions are valid

KevinL10 (approved):
looks good, would suggest a custom decoder for the regions

mjq (approved):
Tiny, judgment call nit aside (and Kevin's notes above), LGTM!

static/app/views/insights/browser/webVitals/components/tables/pagePerformanceTable.tsx
DominikB2014: Hmm, When would this occur?

static/app/views/insights/browser/resources/utils/useResourceFilters.ts
DominikB2014: Yeah I agree!

static/app/views/insights/browser/webVitals/components/tables/pagePerformanceTable.tsx
DominikB2014: I'm going to address this in another PR, i'm still not 100% sure on the benefit so i'll give it some thought, there are 2 cases I can see that would produce an invalid regions.

1. Our code somehow produces an invalid subregion. In this case, i feel as though adding decodeSubregion might actually mask any problem we have in code that shouldn't be occurring anyways. I think we would want to see if this problem occurs. 
2. The user manually enters a subregion in the url bar. In this case the query should succeed here anyways, unless they enter a string that happens to match some other query syntax. Which in that case i think it might be better to show a failed query in our UI, so that the user knows what they entered in the url bar doesn't make sense. Either way I like the idea of showing the user the results of their own changes in the url bar.

PR #76186, opened 2024-08-14
feat(search): Recent searches should be valid filter keys.

Abdkhan14:
Example:
- In discover the filters keys are dataset aware, so we want the recent searches to follow.
- Introducing this change allows us to keep the filterKeys prop as the sole controller for such logic.

### DIFF
--- a/static/app/components/searchQueryBuilder/tokens/filterKeyListBox/useRecentSearchFilters.tsx
+++ b/static/app/components/searchQueryBuilder/tokens/filterKeyListBox/useRecentSearchFilters.tsx
@@ -6,6 +6,7 @@ import type {FieldDefinitionGetter} from 'sentry/components/searchQueryBuilder/t
 import {parseQueryBuilderValue} from 'sentry/components/searchQueryBuilder/utils';
 import {type ParseResult, Token} from 'sentry/components/searchSyntax/parser';
 import type {RecentSearch, TagCollection} from 'sentry/types/group';
+import {isAggregateFieldOrEquation} from 'sentry/utils/discover/fields';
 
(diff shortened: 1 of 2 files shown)


static/app/components/searchQueryBuilder/tokens/filterKeyListBox/useRecentSearchFilters.tsx
```diff
@@ -64,7 +65,14 @@ function getFiltersFromRecentSearches(
     .flatMap(search =>
       getFiltersFromQuery({query: search.query, getFieldDefinition, filterKeys})
     )
-    .filter(filter => !filtersInCurrentQuery.includes(filter))
+    .filter(filter => {
+      if (isAggregateFieldOrEquation(filter)) {
```
malwilley, line 69: Is this necessary? We probably want to treat aggregates the same as other filter keys. If we support that filter key we should show it, otherwise we should not.

I think part of the reason is that `getFiltersFromQuery` doesn't handle aggregate filters correctly. Try changing that function to return `getKeyName(token.key)` instead of `token.key.text`

malwilley (commented):
Great change! Thanks for adding this check, makes things much more usable.

Could you add a quick test for this new behavior? There are some already written in `SearchQueryBuilder -> filter key menu -> recent filter keys` that you can either modify or add a new test for. Thanks!

static/app/components/searchQueryBuilder/tokens/filterKeyListBox/useRecentSearchFilters.tsx
Abdkhan14: This works @malwilley , I added a test too.

PR #76187, opened 2024-08-14
Nhsieh/navigationv2

nhsiehgit:
https://getsentry.atlassian.net/browse/ALRT-210

Updates to the Navigation sidebar UI

https://github.com/user-attachments/assets/c3e02699-5213-4789-a3e2-0d89075730bc

TODO: tests

### DIFF
--- a/static/app/components/sidebar/index.tsx
+++ b/static/app/components/sidebar/index.tsx
@@ -124,7 +124,7 @@ function Sidebar() {
   const organization = useOrganization({allowNull: true});
   const {shouldAccordionFloat} = useContext(ExpandedContext);
   const hasNewNav = useMemo(() => {
-    return organization?.features.includes('organizations:navigation-sidebar-v2');
+    return organization?.features.includes('navigation-sidebar-v2');
   }, [organization]);
   const hasOrganization = !!organization;
   const isSelfHostedErrorsOnly = ConfigStore.get('isSelfHostedErrorsOnly');
--- a/static/app/components/sidebar/index.tsx
+++ b/static/app/components/sidebar/index.tsx
@@ -644,7 +644,7 @@ function Sidebar() {
         <SidebarSectionGroupPrimary>
           <DropdownSidebarSection
             isSuperuser={showSuperuserWarning() && !isExcludedOrg()}
-            noMargin={hasNewNav}
+            hasNewNav={hasNewNav}
           >
             <SidebarDropdown
               orientation={orientation}
@@ -704,7 +704,7 @@ function Sidebar() {
         </SidebarSectionGroupPrimary>
 
         {hasOrganization && (
-          <SidebarSectionGroup>
+          <SidebarSectionGroup hasNewNav={hasNewNav}>
             {/* What are the onboarding sidebars? */}
             <PerformanceOnboardingSidebar
               currentPanel={activePanel}
@@ -846,11 +846,11 @@ export const SidebarWrapper = styled('nav')<{collapsed: boolean; hasNewNav?: boo
   }
 `;
 
-const SidebarSectionGroup = styled('div')`
+const SidebarSectionGroup = styled('div')<{hasNewNav?: boolean}>`
   ${responsiveFlex};
   flex-shrink: 0; /* prevents shrinking on Safari */
   gap: 1px;
-  align-items: center;
+  ${p => p.hasNewNav && `align-items: center;`}
 `;
 
 const SidebarSectionGroupPrimary = styled('div')`
@@ -934,6 +934,7 @@ const SidebarSection = styled(SidebarSectionGroup)<{
 `;
 
 const DropdownSidebarSection = styled(SidebarSection)<{
+  hasNewNav?: boolean;
   isSuperuser?: boolean;
 }>`
   position: relative;
@@ -951,6 +952,7 @@ const DropdownSidebarSection = styled(SidebarSection)<{
         background: ${p.theme.superuserSidebar};
       }
     `}
+  ${p => p.hasNewNav && `align-items: center;`}
 `;
 
 const SidebarCollapseItem = styled(SidebarItem)`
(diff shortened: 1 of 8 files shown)


static/app/components/sidebar/index.tsx
```diff
@@ -586,13 +615,21 @@ function Sidebar() {
   );
 
   return (
-    <SidebarWrapper aria-label={t('Primary Navigation')} collapsed={collapsed}>
+    <SidebarWrapper
+      aria-label={t('Primary Navigation')}
+      collapsed={collapsed}
+      hasNewNav={hasNewNav}
```
saponifi3d, line 619: Rather than passing this property to each component, could we use a context to share the state between each component?

static/app/components/sidebar/index.tsx
```diff
@@ -176,16 +196,18 @@ function Sidebar() {
...
-    organization,
-  };
+    if (hasNewNav) {
+      bcl.add('hasNewNav');
+    } else {
+      bcl.remove('hasNewNav');
+    }
+
+    return () => bcl.remove('hasNewNav');
+  }, [hasNewNav]);
```
saponifi3d, line 208: thoughts on using a data property for this rather than classlist? The data property can also be used in css so your selector could check to see if a parent has the data property set.

static/app/components/sidebar/sidebarItem.tsx
```diff
@@ -364,23 +386,32 @@ const StyledSidebarItem = styled(Link, {
...
-    left: calc(-${space(2)} - 1px);
-    bottom: 6px;
-    width: 5px;
-    border-radius: 0 3px 3px 0;
-    background-color: transparent;
-    transition: 0.15s background-color linear;
-  }
+  margin: ${p => (p.hasNewNav ? space(2) : space(1))} 0;
+  ${p => {
+    if (!p.hasNewNav) {
```
saponifi3d, line 395: you can probs do all of this with a css selector, something like `[data-attr-name]="true" > parent`

Then this could all be driven from a single state change property on the navigation.

static/app/components/sidebar/index.tsx
nhsiehgit: since this is just a feature flag i'd rather keep it simple and just pass the state rather than create a context
thankfully we're only going 1 layer deep here so i think context might be a little over kill

static/app/components/sidebar/index.tsx
nhsiehgit: 🤷 

since we're already utilizing this property pattern, i'd rather keep it to a single pattern than try to introduce a new one.

This instance in particular is a really odd one.
This is actually for a global style to update the margins of the body - which we cannot set on the actual global body tag until we can actually pull the flag from the org state...

static/app/components/sidebar/sidebarItem.tsx
nhsiehgit: Ah ehh
I would rather not try to introduce a new pattern for a potentially temporary feature flag :/

static/app/components/sidebar/index.tsx
```diff
@@ -417,7 +438,11 @@ function Sidebar() {
       <SidebarItem
         {...sidebarItemProps}
         icon={<IconLightning />}
-        label={<GuideAnchor target="performance">{t('Performance')}</GuideAnchor>}
+        label={
+          <GuideAnchor target="performance">
+            {hasNewNav ? truncateString(t('Performance'), 4) : t('Performance')}
```
scttcper, line 443: you can't really truncate translated text like this, either removing translations or using another truncation method like css ellipsis to truncate might be better. Should also consider the aria label once the string is truncated

static/app/components/sidebar/index.tsx
```diff
@@ -119,11 +123,25 @@ function Sidebar() {
   const activePanel = useLegacyStore(SidebarPanelStore);
   const organization = useOrganization({allowNull: true});
   const {shouldAccordionFloat} = useContext(ExpandedContext);
+  const hasNewNav = useMemo(() => {
+    return organization?.features.includes('navigation-sidebar-v2');
+  }, [organization]);
```
scttcper, line 128: not really worth memoing since the boolean already works well with Object.is comparisons downstream and includes is very fast
```suggestion
  const hasNewNav = organization?.features.includes('navigation-sidebar-v2');
```

static/app/components/sidebar/index.tsx
nhsiehgit: hmmm good point 🤔 
let me ask Vu how we want to address this

static/app/components/sidebar/sidebarItem.tsx
nhsiehgit: So - how might you suggest doing this for elements that are not all nested at the same level?

Is there any way to select for the top level rather than just the parent?

gggritso:
Maybe too early to raise this, but watch out for sidebar upsells! On lower tier plans (e.g., AM2 Free or AM3 Team) some sidebar items (the Insights accordion and/or the Insights sub-pages) have special upsell wrappers that might not work or look right with this new layout

static/app/components/sidebar/index.tsx
```diff
@@ -417,7 +436,11 @@ function Sidebar() {
       <SidebarItem
         {...sidebarItemProps}
         icon={<IconLightning />}
-        label={<GuideAnchor target="performance">{t('Performance')}</GuideAnchor>}
+        label={
+          <GuideAnchor target="performance">
+            {hasNewNav ? truncateString('Performance', 4) : t('Performance')}
```
scttcper, line 441: you could just hardcode it to Perf

nhsiehgit: 😅 ya tru.
might as well just do that then lol

static/app/components/sidebar/sidebarItem.tsx
saponifi3d: you can just have the `[data-attr-name="true'] <child-selector> {` and that will select the corresponding children from elements that have that data attribute set (ideally the data-attr selector also has a class selector first so you don't have to check every element in the dom)

saponifi3d: chatted in slack - this is likely more difficult than i was anticipating with css-in-js

scttcper (approved):

PR #76188, opened 2024-08-14
ref(similarity): Manually filter times seen in backfill query

jangjodi:
Manually filter times seen in backfill query to prevent query from timing out
Only end the backfill task for the project if the length of the unfiltered batch is 0
Remove retry on OperationalError


JoshFerge (approved):
makes sense to me!

wedamija (approved):

PR #76189, opened 2024-08-14
style(toolbar): adjust alert indicator css

michellewzhang:
- relates to https://github.com/getsentry/sentry/issues/75636
- new styling for the alert badge -- no count, just a red dot (see [figma](https://www.figma.com/design/G8sBIfdPBh7wwpKdEMjqcD/Specs%3A-Dev-Toolbar-v2?node-id=2087-3091&t=KlTCSSkhLqDtjp8y-0))

before:
SCR-20240814-jyjf

after:
SCR-20240814-jyfs


static/app/components/devtoolbar/components/alerts/alertCountBadge.tsx
michellewzhang, line 1: this file was renamed to `alertBadge.tsx`

static/app/components/devtoolbar/components/indicatorBadge.tsx
```diff
@@ -11,12 +10,11 @@ const variants = {
...
 
 interface Props {
-  icon: React.ReactNode;
   variant: keyof typeof variants;
 }
 
-export default function IndicatorBadge({icon, variant}: Props) {
-  return <div css={[smallCss, counterCss, variants[variant]]}>{icon}</div>;
+export default function IndicatorBadge({variant}: Props) {
+  return <div css={[smallCss, counterCss, variants[variant]]} />;
```
aliu39, line 17:
```suggestion
  return <div css={[smallCss, badgeCss, variants[variant]]} />;
```
rename?

static/app/components/devtoolbar/components/indicatorBadge.tsx
```diff
@@ -25,12 +23,12 @@ const counterCss = css`
...
   box-sizing: content-box;
   color: var(--gray100);
-  height: 1rem;
+  height: 0.55rem;
   line-height: 1rem;
   position: absolute;
-  right: -6px;
-  top: -6px;
-  width: 1rem;
+  right: 2px;
```
aliu39, line 29: Css learning question: Is right and top used to position it relative to the panel icon? How does position:absolute work here?

aliu39 (approved):

static/app/components/devtoolbar/components/indicatorBadge.tsx
michellewzhang: @aliu39 
> Is right and top used to position it relative to the panel icon? 

yes!

> How does position:absolute work here?

it works so that our `top` and `left` are relative to the icon itself, since we want the dot on top of the icon.


image

relative:

SCR-20240815-izsg

absolute:


SCR-20240815-izoq

aliu39: 🔥 🔥 thanks for the diagrams!

PR #76190, opened 2024-08-14
ref(control_silo): Move UserAvatar model to users module

Christinarlong:
Part of moving control silo user related resources into the users module, this PR does not include an old UserRole import shim as it's not used in getsentry. Includes adding of types for functions.

Apart of (#73856)


tests/sentry/models/test_organization_avatar.py
Christinarlong, line 1: Broke `test_avatar` into these two files
- `test_organization_avatar`
- `test_user_avatar`

src/sentry/users/models/user_avatar.py
```diff
@@ -63,7 +63,7 @@ def outboxes_for_update(self, shard_identifier: int | None = None) -> list[Contr
         )
 
     @contextlib.contextmanager
-    def _maybe_prepare_outboxes(self, *, outbox_before_super: bool):
+    def _maybe_prepare_outboxes(self, *, outbox_before_super: bool) -> Generator[None]:
```
Christinarlong, line 66: Not rlly familiar with the convention here, should this be `Generator[None, None, None] from following [this](https://docs.python.org/3/library/typing.html#typing.Generator), or does it not matter ?

markstory: `Generator[None]` is the simpler form and one we should use. Recently all `Generator[none, none, none]` were updated. See 16b75635ac031bb5369673ace4ecdf016961a394

markstory (approved):
Looks good to me.

PR #76197, opened 2024-08-14
fix(discord): Fix Bot Command Timeout

iamrajjoshi:
Fixes https://sentry.sentry.io/issues/5131244670/events/a169bd4760e9408f86f40822285b804f/ & Discord Bot Commands that are currently failing:
image

What went wrong?

Discord expects us to respond to a webhook within 3 seconds, which is difficult to do when we are making RPC calls between region and control.

Currently the way handle the latency induced by RPC calls is to send a discord a response to a webhook, which tells it that we are processing. this buys us 15 minutes to respond with the new data, which is more enough.

However, we didn't follow the Discord API spec correctly, and out response was not formatted & we sent the incorrect status code. 
> When responding to an interaction received via webhook, your server can simply respond to the received POST request. You'll want to respond with a 200 status code (if everything went well), as well as specifying a type and data....
(https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-messages)

What this meant is in reality, we only had 3 seconds to respond and when the RPC call didn't finish in time, Discord believed we didn't respond. In this pr, i actually send the correct response, which ultimately gives us the full 15 minutes.

In the video below, i exaggerated how long it would take to respond to show the functionality. Notice how it says, the bot is thinking, which shows we told Discord that we are working on a response.

https://github.com/user-attachments/assets/ae7ad9bb-8ff2-4c34-bb01-30d8d22c1190

Note: We run through this async logic only when running on silo mode, which means testing this locally on monolith won't surface this issue.

### DIFF
--- a/src/sentry/middleware/integrations/parsers/discord.py
+++ b/src/sentry/middleware/integrations/parsers/discord.py
@@ -9,10 +9,12 @@ from rest_framework import status
 from rest_framework.request import Request
 
 from sentry.hybridcloud.outbox.category import WebhookProviderIdentifier
+from sentry.integrations.discord.message_builder.base.flags import EPHEMERAL_FLAG
 from sentry.integrations.discord.requests.base import DiscordRequest, DiscordRequestError
 from sentry.integrations.discord.views.link_identity import DiscordLinkIdentityView
 from sentry.integrations.discord.views.unlink_identity import DiscordUnlinkIdentityView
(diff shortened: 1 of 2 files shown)


cathteng (approved):
🚢

GabeVillalobos (approved):

sentaur-athena:
Good find. Just FYI, the correct response for an async request is in fact 202 though. If discord only expects 200 and accepts that as the only good response I think it's kind of bug on their side.

PR #76198, opened 2024-08-14
ref(querybuilder): Remove params from snuba modules 

wmak:
- Removes all the params from snuba modules for snuba_params instead


src/sentry/api/serializers/models/project.py
```diff
@@ -391,26 +392,26 @@ def measure_span(op_tag):
                     serialized["options"] = options[project.id]
         return result
 
-    def get_stats(self, project_ids, query):
+    def get_stats(self, projects, query):
```
wmak, line 395: Actual changes here, changed the param to a list of param objects so we don't need to hit the db again

src/sentry/data_export/processors/discover.py
```diff
@@ -21,20 +21,20 @@ class DiscoverProcessor:
     Processor for exports of discover data based on a provided query
     """
 
-    def __init__(self, organization_id, discover_query):
-        self.projects = self.get_projects(organization_id, discover_query)
-        self.environments = self.get_environments(organization_id, discover_query)
+    def __init__(self, organization, discover_query):
```
wmak, line 24: Actual changes here, had to change the param from organization_id to the organization object.

And changed `get_environments` to return a list of objects instead of names

src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -156,13 +157,12 @@ def fetch_historical_data(
     historical_data = dataset.timeseries_query(
         selected_columns=[snuba_query.aggregate],
         query=snuba_query.query,
-        params={
-            "organization_id": alert_rule.organization.id,
-            "project_id": [project.id],
-            "granularity": granularity,
```
wmak, line 162: Granularity never did anything? not sure why it was passed here

src/sentry/snuba/discover.py
```diff
@@ -360,13 +344,16 @@ def timeseries_query(
...
             if len(base_builder.aggregates) != 1:
                 raise InvalidSearchQuery("Only one column can be selected for comparison queries")
-            comp_query_params = deepcopy(params)
-            comp_query_params["start"] -= comparison_delta
-            comp_query_params["end"] -= comparison_delta
+            comp_query_params = snuba_params.copy()
+            assert comp_query_params.start is not None, "start is required"
+            assert comp_query_params.end is not None, "end is required"
+            comp_query_params.start -= comparison_delta
+            comp_query_params.end -= comparison_delta
```
wmak, line 351: I sorta hate this block 🤔 after the follow up PR to remove params entirely might add a utility to clean this up

wmak (commented):
Commented on any actual changes, everything else should just be straightforward removing of params

shruthilayaj (approved):
Okay, like you said, most of the changes were either not passing param anymore or passing that empty param dict to the query builders, so that was straightforward. I also looked at the function signatures you explicitly called out, those looked good to me too.
🚀

PR #76200, opened 2024-08-14
fix(settings): disable allowMemberProjectCreation if no biz plan

cathteng:
Screenshot 2024-08-14 at 11 40 30


JoshFerge (approved):
should we remove the feature check here? https://github.com/getsentry/sentry/blob/f08a425e6c82ab17d1f6e2cd91bfbb89b723ea46/static/app/components/projects/canCreateProject.tsx#L13

PR #76201, opened 2024-08-14
feat(discover): Split Discover Dataset task

shruthilayaj:
Task that will split Discover saved queries
This PR just introduces the task (doesn't actually run it)


src/sentry/options/defaults.py
```diff
@@ -2563,6 +2563,18 @@
...
+    "discover.saved-query-dataset-split.enable",
+    default=False,
+    flags=FLAG_PRIORITIZE_DISK | FLAG_AUTOMATOR_MODIFIABLE,
+)
+register(
+    "discover.saved-query-dataset-split.organization-allowlist",
+    type=Sequence,
+    default=[],
+    flags=FLAG_PRIORITIZE_DISK | FLAG_AUTOMATOR_MODIFIABLE,
+)
```
narsaynorath, line 2576: Is there a benefit to us setting them as options instead of Flagpole? iirc Flagpole works for ST now

src/sentry/discover/dataset_split.py
```diff
@@ -0,0 +1,384 @@
...
+                saved_query,
+                DiscoverSavedQueryTypes.TRANSACTION_LIKE,
+                DatasetSourcesTypes.INFERRED.value,
+            )
+
+        return DiscoverSavedQueryTypes.TRANSACTION_LIKE, True
+
+    if dry_run:
+        logger.info(
+            "Split decision for %s: %s", saved_query.id, DiscoverSavedQueryTypes.TRANSACTION_LIKE
```
narsaynorath, line 375:
```suggestion
            "Split decision for %s: %s", saved_query.id, DiscoverSavedQueryTypes.ERROR_EVENTS
```
Should we also record that the decision was forced?

src/sentry/tasks/split_discover_query_dataset.py
```diff
@@ -0,0 +1,178 @@
...
+        logger.info(
+            "sentry.tasks.split_discover_query_dataset - no more queries to process",
+            extra={
+                "transaction_dataset_count": transaction_dataset_count,
+                "error_dataset_count": error_dataset_count,
+                "errored_query_count": errored_query_count,
+                "inferred_with_query": inferred_with_query,
+                "inferred_without_query": inferred_without_query,
+            },
+        )
```
narsaynorath, line 64: I don't think we're logging this at the end unless we reach an error condition

src/sentry/discover/dataset_split.py
```diff
@@ -0,0 +1,384 @@
...
+    return None
+
+
+def get_field_list(fields: list[str]) -> list[str]:
+    return [field for field in fields[:] if not is_equation(field)]
+
+
+def get_equation_list(fields: list[str]) -> list[str]:
+    """equations have a prefix so that they can be easily included alongside our existing fields"""
+    return [strip_equation(field) for field in fields[:] if is_equation(field)]
```
narsaynorath, line 222: Do we need the copy for `fields[:]`?

src/sentry/options/defaults.py
shruthilayaj: The first one is a kill switch - just bypasses the whole task or forces us to break out of the task, this should just be an on/off not tied to anything so we can quickly shut it down in case it causes an incident

The second one, I kinda went back and forth on a little, my thinking was to initially just limit the queryset to the allowlisted orgs https://github.com/getsentry/sentry/pull/76201/files#diff-179f670f3f1385148441a5abb704eec14d23b90a6fbace8ca071996985c17815R54
instead of looping through the whole queryset and checking individual orgs. Then once we're done testing, just remove that option and run the task on dataset selector enabled orgs. I don't have strong opinions on this one though

src/sentry/discover/dataset_split.py
```diff
@@ -0,0 +1,386 @@
...
+    "exception_frames.function",
+    "exception_frames.in_app",
+    "exception_frames.lineno",
+    "exception_frames.module",
+    "exception_frames.package",
+    "exception_frames.stack_level",
+]
+
+
+def save_split_decision_for_query(
```
wmak, line 74: Lets prefix these with `_` and add a comment explaining that these functions should only be used for the split migration so other people in the org don't try to use 'em

src/sentry/discover/dataset_split.py
```diff
@@ -0,0 +1,386 @@
...
+
+        return check_column_matches_dataset(lhs, dataset)
+
+    if isinstance(lhs, Function) or isinstance(lhs, CurriedFunction):
+        return check_function_parameter_matches_dataset(lhs, dataset)
+
+    return False
+
+
+def check_top_level_conditions_match_dataset(
```
narsaynorath, line 154: The terminology you used here, `top_level_conditions`, are there conditions that are not top level?

src/sentry/options/defaults.py
```diff
@@ -2555,6 +2555,18 @@
     flags=FLAG_PRIORITIZE_DISK | FLAG_AUTOMATOR_MODIFIABLE,
 )
 
+register(
+    "discover.saved-query-dataset-split.enable",
+    default=False,
+    flags=FLAG_PRIORITIZE_DISK | FLAG_AUTOMATOR_MODIFIABLE,
+)
+register(
+    "discover.saved-query-dataset-split.organization-allowlist",
```
wmak, line 2564: nit: can we call this
```suggestion
    "discover.saved-query-dataset-split.organization-id-allowlist",
```

src/sentry/tasks/split_discover_query_dataset.py
```diff
@@ -0,0 +1,172 @@
...
+from sentry.tasks.base import instrumented_task
+from sentry.utils import metrics, snuba
+from sentry.utils.query import RangeQuerySetWrapper
+
+logger = logging.getLogger("sentry.tasks.split_discover_query_dataset")
+
+
+SLEEP_FOR = 5 * 60  # 5 minutes
+MAX_NOOP_ATTEMPTS = 10
+RATE_LIMIT_CACHE: MutableMapping[str | int, int] = LRUCache(maxsize=1000)
```
wmak, line 21: Do we have a rough estimate on the runtime per org yet? (could check snuba admin if not)

src/sentry/tasks/split_discover_query_dataset.py
```diff
@@ -0,0 +1,172 @@
...
+
+class EmptyQueryset(Exception):
+    pass
+
+
+class TaskKilled(Exception):
+    pass
+
+
+def _split_discover_query_dataset(dry_run):
```
wmak, line 36: Should we add some sentry span instrumentation to this task?

src/sentry/tasks/split_discover_query_dataset.py
```diff
@@ -0,0 +1,172 @@
...
+        except SoftTimeLimitExceeded:
+            errored_query_count += 1
+            logger.warning(
+                "SoftTimeLimitExceeded",
+                extra={
+                    "transaction_dataset_count": transaction_dataset_count,
+                    "error_dataset_count": error_dataset_count,
+                    "errored_query_count": errored_query_count,
+                    "inferred_with_query": inferred_with_query,
+                    "inferred_without_query": inferred_without_query,
```
wmak, line 117: Should we include saved query id here too

wmak (approved):
Mostly small comments and nits, lgtm otherwise

One thing to consider would be to unit-test the functions in dataset_split, but they're e2e tested which is fine

narsaynorath (approved):
Looks good to me, good job on this. It's very thorough and it looks like you've distilled the queries down to their most minimal forms for querying to evaluate the split

src/sentry/tasks/split_discover_query_dataset.py
shruthilayaj: Ran a redash query actually, we don't have enough distinct org ids with discover saved queries that we need cache with an upper limit - I just made it a regular dict

shruthilayaj:
> One thing to consider would be to unit-test the functions in dataset_split, but they're e2e tested which is fine

I added some unit tests for the logic that splits datasets without querying snuba!
https://github.com/getsentry/sentry/pull/76201/commits/38240918891fd3fbf09fb9b24e57f0bfc9849484#diff-1f53d05795f5e5a6afb8672dc304c7a1ca0403f167e76d62d659cb00e09b5e8dR544

src/sentry/discover/dataset_split.py
shruthilayaj: We basically don't bother checking nested conditions like (foo AND (bar OR baz)) will only check foo condition

PR #76202, opened 2024-08-14
ref(control_silo): Move UserIP model to users module

Christinarlong:
Part of moving control silo user related resources into the users module, this PR does not include an old UserIP import shim as the UserIP model is not used in getsentry. Includes adding of types for functions.

Apart of (https://github.com/getsentry/sentry/issues/73856)


markstory (approved):
🎉

PR #76208, opened 2024-08-14
feat(eap): Add a method that powers metrics explorer with EAP

colin-sentry:
Look, this is pretty gross. It short-circuit returns some things, doesn't implement all of MQL, but it's a way to at least query EAP in production. We can start implementing the rest with net-new RPCs for things like tag discovery later.

### DIFF
--- a/src/sentry/api/endpoints/organization_metrics_query.py
+++ b/src/sentry/api/endpoints/organization_metrics_query.py
@@ -182,7 +182,7 @@ class OrganizationMetricsQueryEndpoint(OrganizationEndpoint):
                             interval,
                             organization,
                             projects,
-                            Referrer.API_ORGANIZATION_METRICS_QUERY.value,
+                            Referrer.API_ORGANIZATION_METRICS_EAP_QUERY.value,
                         )
                         return Response(status=200, data=results)
 
--- a/src/sentry/api/endpoints/organization_metrics_query.py
+++ b/src/sentry/api/endpoints/organization_metrics_query.py
@@ -175,7 +175,7 @@ class OrganizationMetricsQueryEndpoint(OrganizationEndpoint):
                 if len(mql_queries) == 1 and "a" in mql_queries[0].sub_queries:
                     subquery = mql_queries[0].sub_queries["a"]
                     if "d:eap/" in subquery.mql:
-                        results = mql_eap_bridge.make_eap_request(
+                        res_data = mql_eap_bridge.make_eap_request(
                             subquery.mql,
                             start,
                             end,
@@ -184,7 +184,7 @@ class OrganizationMetricsQueryEndpoint(OrganizationEndpoint):
                             projects,
                             Referrer.API_ORGANIZATION_METRICS_EAP_QUERY.value,
                         )
-                        return Response(status=200, data=results)
+                        return Response(status=200, data=res_data)
 
             results = run_queries(
                 mql_queries=mql_queries,
--- a/src/sentry/api/endpoints/organization_metrics_query.py
+++ b/src/sentry/api/endpoints/organization_metrics_query.py
@@ -4,7 +4,7 @@ from datetime import datetime, timedelta, timezone
 from rest_framework.request import Request
 from rest_framework.response import Response
 
-from sentry import options
+from sentry import features, options
 from sentry.api.api_owners import ApiOwner
 from sentry.api.api_publish_status import ApiPublishStatus
 from sentry.api.base import region_silo_endpoint
@@ -17,6 +17,7 @@ from sentry.sentry_metrics.querying.data import (
     MQLQuery,
     run_queries,
 )
+from sentry.sentry_metrics.querying.eap import mql_eap_bridge
 from sentry.sentry_metrics.querying.errors import (
     InvalidMetricsQueryError,
     LatestReleaseNotFoundError,
@@ -156,6 +157,7 @@ class OrganizationMetricsQueryEndpoint(OrganizationEndpoint):
             start, end = get_date_range_from_params(request.GET)
             interval = self._interval_from_request(request)
             mql_queries = self._mql_queries_from_request(request)
+            projects = self.get_projects(request, organization)
 
             metrics.incr(
                 key="ddm.metrics_api.query",
@@ -166,13 +168,31 @@ class OrganizationMetricsQueryEndpoint(OrganizationEndpoint):
                 },
             )
 
+            if all(
+                features.has("projects:use-eap-spans-for-metrics-explorer", project)
+                for project in projects
+            ):
+                if len(mql_queries) == 1 and "a" in mql_queries[0].sub_queries:
+                    subquery = mql_queries[0].sub_queries["a"]
+                    if "d:eap/" in subquery.mql:
+                        results = mql_eap_bridge.make_eap_request(
+                            subquery.mql,
+                            start,
+                            end,
+                            interval,
+                            organization,
+                            projects,
+                            Referrer.API_ORGANIZATION_METRICS_QUERY.value,
+                        )
+                        return Response(status=200, data=results)
+
             results = run_queries(
                 mql_queries=mql_queries,
                 start=start,
                 end=end,
                 interval=interval,
                 organization=organization,
-                projects=self.get_projects(request, organization),
+                projects=projects,
                 environments=self.get_environments(request, organization),
                 referrer=Referrer.API_ORGANIZATION_METRICS_QUERY.value,
                 query_type=self._get_query_type_from_request(request),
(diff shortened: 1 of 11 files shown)


volokluev (approved):

src/sentry/sentry_metrics/querying/eap/protobufs/AggregateBucket_pb2.pyi
```diff
@@ -0,0 +1,91 @@
+"""
```
asottile-sentry, line 1: we should not be checking in generated files

asottile-sentry (dismissed):
please separate out the package bumps from this patch to minimize risk

src/sentry/sentry_metrics/querying/eap/protobufs/AggregateBucket_pb2.pyi
asottile-sentry: I would recommend making a python package client library and adding that as a dependency rather than vendoring this in like this

colin-sentry: this is a short term thing, we will migrate to the protos repo when things are more stable there

asottile-sentry: let's do the right thing first rather than hacking this together -- no need to clutter the repo and be risky

colin-sentry: the tooling doesn't exist yet, it's a chicken and egg problem right now because nobody uses protobuf, so there is no tooling, so we can't use protobuf 😿

asottile-sentry: so that should tell you that protobuf is not a viable technology to be used at sentry _yet_ and you should use a supported approach rather than taking a risky one

markstory: The tooling will exist *soon*. I'm actively working on it and hope to have something later this week. I agree with @asottile-sentry about not making a mess for ourselves in the short-term.

dbennett-sentry: Thanks, @markstory. So this whole topic will resolve itself this week?

src/sentry/api/endpoints/organization_metrics_query.py
```diff
@@ -166,13 +168,31 @@ def post(self, request: Request, organization: Organization) -> Response:
...
+                    subquery = mql_queries[0].sub_queries["a"]
+                    if "d:eap/" in subquery.mql:
+                        results = mql_eap_bridge.make_eap_request(
+                            subquery.mql,
+                            start,
+                            end,
+                            interval,
+                            organization,
+                            projects,
+                            Referrer.API_ORGANIZATION_METRICS_QUERY.value,
```
evanh, line 185: I would make this a separate referrer, to make it a little easier to track independently in Snuba.

src/sentry/sentry_metrics/querying/eap/mql_eap_bridge.py
```diff
@@ -0,0 +1,131 @@
...
+    referrer: str,
+) -> dict:
+    start_time_proto = ProtobufTimestamp()
+    start_time_proto.FromDatetime(start)
+    end_time_proto = ProtobufTimestamp()
+    end_time_proto.FromDatetime(end)
+
+    ts: Timeseries = snuba_sdk.mql.mql.parse_mql(query_mql)
+
+    aggregate_map = {
```
evanh, line 64: Could this be a global variable?

src/sentry/api/endpoints/organization_metrics_tag_details.py
```diff
@@ -35,6 +36,28 @@ def get(self, request: Request, organization: Organization, tag_name: str) -> Re
                 {"detail": "You must supply at least one project to see its metrics"}, status=404
             )
 
+        if all(
+            features.has("projects:use-eap-spans-for-metrics-explorer", project)
+            for project in projects
+        ):
+            if len(metric_names) == 1 and metric_names[0].startswith("d:eap"):
```
evanh, line 43: Couldn't this be fetched from the metrics platform still? At least then we'd get (more) realistic data.

src/sentry/api/endpoints/organization_metrics_tags.py
```diff
@@ -52,6 +53,14 @@ def get(self, request: Request, organization: Organization) -> Response:
         if not is_mri(metric_name):
             raise BadRequest(message="Please provide a valid MRI to query a metric's tags.")
 
+        if all(
+            features.has("projects:use-eap-spans-for-metrics-explorer", project)
+            for project in projects
+        ):
+            if metric_name.startswith("d:eap"):
+                # TODO hack for EAP, return a fixed list
+                return Response([Tag(key="color"), Tag(key="location")])
```
evanh, line 62: Similar to the other comment, could this not still go to the metrics platform for now?

src/sentry/sentry_metrics/querying/eap/mql_eap_bridge.py
evanh, line 64: What about `count_unique`?

src/sentry/api/endpoints/organization_metrics_query.py
```diff
@@ -166,13 +168,31 @@ def post(self, request: Request, organization: Organization) -> Response:
...
             )
 
+            if all(
+                features.has("projects:use-eap-spans-for-metrics-explorer", project)
+                for project in projects
+            ):
+                if len(mql_queries) == 1 and "a" in mql_queries[0].sub_queries:
+                    subquery = mql_queries[0].sub_queries["a"]
+                    if "d:eap/" in subquery.mql:
+                        results = mql_eap_bridge.make_eap_request(
```
evanh, line 178: Do you want to handle this with a try...except so we can fail over to the main pipeline?

src/sentry/sentry_metrics/querying/eap/mql_eap_bridge.py
```diff
@@ -0,0 +1,131 @@
...
+        "P99": AggregateBucketRequest.P99,
+        "count": AggregateBucketRequest.COUNT,
+    }
+
+    rpc_filters = None
+    if ts.filters is not None:
+        rpc_filters = TraceItemFilter(
+            and_filter=AndFilter(filters=list(parse_mql_filters(ts.filters)))
+        )
+    req = AggregateBucketRequest(
```
evanh, line 78: How is this determining what the interval buckets should be when building the timeseries?

src/sentry/sentry_metrics/querying/eap/mql_eap_bridge.py
```diff
@@ -0,0 +1,131 @@
...
+            )
+        elif isinstance(cond, MQLCondition):
+            if cond.op == MQLOp.EQ:
+                yield TraceItemFilter(
+                    string_comparison=StringFilter(key=cond.lhs.name, value=cond.rhs)
+                )
+        # TODO: maybe we want to implement other stuff
+
+
+def make_eap_request(
```
evanh, line 48: I noticed that there are a few features of MQL queries that this code doesn't work for:
- Formula aggregates
- Group By clauses
- With totals flag
- Order by

Looks like you are catching the first one already, but it would be good to check for the other and revert back to the original query stream in those cases.

I think this check will catch the first one, but not the other ones

src/sentry/api/endpoints/organization_metrics_tags.py
colin-sentry: sure, but this PR is already pretty big and I'd like to add that logic incrementally

src/sentry/sentry_metrics/querying/eap/mql_eap_bridge.py
colin-sentry: not implemented yet

src/sentry/sentry_metrics/querying/eap/mql_eap_bridge.py
colin-sentry: this is only for net-new metrics in the virtual "eap" namespace, falling through to the old logic wouldn't work because there wouldn't be any data for them

src/sentry/sentry_metrics/querying/eap/mql_eap_bridge.py
colin-sentry: hard coded for now

src/sentry/api/endpoints/organization_metrics_query.py
colin-sentry: done

src/sentry/api/endpoints/organization_metrics_query.py
colin-sentry: main pipeline would always fail with the 'eap' namespace, doesn't really add anything

PR #76209, opened 2024-08-14
feat(dashboards): Auto-size Big Number widget

### LINKED ISSUE #75730: Big number visualization doesn't scale with widget width
**Problem**: We have an issue with big number visualisations - by default they are cutoff and resizing them takes up an unnecessary amount of space on the dashboard. They should resize to fit the size of the widget.

Image
Image

- https://github.com/getsentry/sentry/pull/76209
- https://github.com/getsentry/sentry/pull/76669
- https://github.com/getsentry/sentry/pull/76955
- https://github.com/getsentry/sentry/pull/76995
- https://github.com/getsentry/sentry/pull/77097

gggritso:
**Before:**
Screenshot 2024-08-13 at 3 16 05 PM

**After:**
Screenshot 2024-08-13 at 3 16 17 PM

**Bonus:**
Screenshot 2024-08-14 at 3 53 42 PM

## Explanation

The "Big Number" widget is a Dashboards feature where a user can make a query like `count()` and get a big honkin' number that shows how many transactions they have. The widget has some fancy CSS to help size it big (but not too big), but CSS alone isn't enough. In many cases, the big number gets truncated! That's not very nice or very useful.

This PR changes the approach. Instead of CSS, it adds a new utility component called `AutoSizedText`. This component mimics the height and width of its parent, and then iteratively updates the font size of its contents until they fit nicely inside the parent. e.g.,

- The parent is 300px by 300x, and contains the text "OH HELLO"
- The component splits the difference and renders the text at a font size of 110px ((200px + 20px) / 2). It sets the known bounds to 20px minimum and 200px maximum
- The component determines the child is too big because it has bigger dimensions than its parent. It splits the difference between the current and minimum font size, and sets it to 165px ((110px + 200px) / 2). It updates the known bounds to 20px minimum and 200px maximum (since 200px is known to be too big) and renders again
- The child is still too big. The component narrows the font size search field again (note: this works kind of like a binary search) by splitting the difference between the current and minimum font size ((165 + 20) / 2) to 92.5. It updates the known bounds to minimum 20px and maximum 165px.
- This process continues until the child is either within 5% of the parent dimensions _and_ fits inside the parent _or_ it runs out of calculations (maximum of 5 per component)

## Other Methods Considered

1. SVG. SVGs can auto-scale to a parent, and can contain `<text>` elements. I didn't like this approach because font scaling by percentage is _not at all_ the same as changing the font size. Modern typefaces do a lot of work to look legible at different sizes at the font level, and the only way to respect this is to set the correct font size
2. CSS `transform` has the same problem as SVG scaling
3. Container Queries. This doesn't work because container queries make it easier to set the font size as a proportion of the container, but our case is very complicated. For one thing, we'd need to clamp and scale on both width and height, but the _width_ of the content is fully variable based on which characters are used. There's no known ratio that would work. For another, that approach doesn't work at all for multi-line text? Etc.
4. Canvas. There's a way (I hear) to render text to a canvas and use `measureText` to get the dimensions, but that's fairly complicated

## FAQs

I think the most obvious question is what's the performance like. In short, it's not great, but I did what I could. The most important thing is wrapping the updates in `useTransition` which will mark them as lower priority. This will allow other, higher-priority renders (like interactions with other dropdowns) to interrupt the resizing logic. I also limited the iteration count to 5, and debounced the resize listener. This seems to work _well enough_ even with CPU throttling set pretty high.

Overall I'm not too worried because resizing is not a common operation, and it's still pretty fast compared to how long a Dashboard resize _already_ takes (a long time).

The second is testing. Short of literal snapshots (which I don't think we do anymore) I couldn't come up with anything that felt more worthwhile than manual QA. You tell me, though.

Fixes [#75730](https://github.com/getsentry/sentry/issues/75730)


static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,177 @@
...
+  useResizeObserver({
+    ref: parentRef,
+    onResize,
+  });
+
+  useLayoutEffect(() => {
+    if (calculationCount > calculationCountLimit) {
+      // Exceeded the iteration count. This should be unlikely! If it happens, abandon
+      return;
+    }
```
narsaynorath, line 57: It would be interesting to collect some telemetry here in case this _does_ happen, so we can check out what the conditions are.

narsaynorath (approved):
Just a comment around the error-ish condition where we reach the maximum depth. I think we should be tracking something here so we can look back and see if there are issues, otherwise it's failing silently. This looks great, I owe you a hot chocolate 😁

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,177 @@
...
+  height: number;
+  width: number;
+};
+
+function calculateDimensionDisparity(
+  a: Dimensions,
+  b: Dimensions,
+  dimension: 'height' | 'width'
+): number {
+  return 1 - Math.abs(a[dimension] / b[dimension]);
```
edwardgou-sentry, line 163:
```suggestion
  return Math.abs(1 - (a[dimension] / b[dimension]));
```

Should the `Math.abs` be wrapped around the subtraction? Otherwise this looks like it can produce disparities < 0. Seems like resizing still works anyways because we do a check to make sure child dimensions are less than or equal to the parent.

edwardgou-sentry (approved):
Pretty cool solution! Thanks George!

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,177 @@
...
+  const [fontSize, setFontSize] = useState<number>((maxFontSize + minFontSize) / 2);
+  const [fontSizeLowerBound, setFontSizeLowerBound] = useState<number>(minFontSize);
+  const [fontSizeUpperBound, setFontSizeUpperBound] = useState<number>(maxFontSize);
+
+  const [calculationCount, setCalculationCount] = useState<number>(0);
+
+  const parentRef = useRef<HTMLDivElement>(null);
+  const childRef = useRef<HTMLDivElement>(null);
+
+  const onResize = debounce(() => {
```
JonasBa, line 33: I cant find it in the spec, but iirc resize observers might already be reasonably throttled by the browser. In any case, I think the approach here is suboptimal because you are re-querying the width and height of the element inside the callback, which causes layout thrashing. The resize observer callback arg should provide the content and bounding box rects which you can just read without having to re-query the dom (this way you avoid layout thrashing which is the most expensive part of the calculation). 

Imo the react aria useResizeObserver hook is poorly implemented and we shouldnt use it for this exact reason

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,177 @@
...
+  const [_isPending, startTransition] = useTransition();
+
+  const [parentHeight, setParentHeight] = useState<number | null>(null);
+  const [parentWidth, setParentWidth] = useState<number | null>(null);
+
+  const [fontSize, setFontSize] = useState<number>((maxFontSize + minFontSize) / 2);
+  const [fontSizeLowerBound, setFontSizeLowerBound] = useState<number>(minFontSize);
+  const [fontSizeUpperBound, setFontSizeUpperBound] = useState<number>(maxFontSize);
+
+  const [calculationCount, setCalculationCount] = useState<number>(0);
```
JonasBa, line 28: Have you considered using a reducer for this logic? I see on L36t that there seems to be some dependency between these values, a reducer might help you express them more clearly

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,177 @@
...
+  position: relative;
+`;
+
+const SizedChild = styled('div')`
+  display: inline-block;
+`;
+
+const RESIZE_DEBOUNCE = 200; // ms
+const DEFAULT_CALCULATION_COUNT_LIMIT = 5;
+const MAXIMUM_DISPARITY = 0.05;
```
JonasBa, line 151: I would probably move these to top of the file since they seem related to the component logic

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,177 @@
...
+      childDimensions.width <= parentDimensions.width &&
+      childDimensions.height <= parentDimensions.height;
+
+    if (
+      childFitsIntoParent &&
+      (widthDisparity <= MAXIMUM_DISPARITY || heightDisparity <= MAXIMUM_DISPARITY)
+    ) {
+      // The child fits completely into the parent _and_ at least one dimension is very similar to the parent size (i.e., it fits nicely in the parent). Abandon, we're done!
+      return;
+    }
```
JonasBa, line 90: Is this disparity calc required because of the recalc constraint? It feels like we should just be able to set the value on each resize event and not worry about this

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,177 @@
+import {useLayoutEffect, useRef, useState, useTransition} from 'react';
+import styled from '@emotion/styled';
+import {useResizeObserver} from '@react-aria/utils';
+import debounce from 'lodash/debounce';
+
+interface Props {
+  children: React.ReactNode;
+  maxFontSize: number;
+  minFontSize: number;
+  calculationCountLimit?: number;
```
JonasBa, line 10: As a developer, how do I know what to set these to? Can this be simplified to only require children so that folks can only worry about the min container width and have the text autosize to it?

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,177 @@
...
+    fontSizeUpperBound,
+    calculationCount,
+    calculationCountLimit,
+    parentHeight,
+    parentWidth,
+  ]);
+
+  return (
+    <ParentSizeMimic ref={parentRef}>
+      <SizedChild ref={childRef} style={{fontSize}}>
```
JonasBa, line 131: There is constraint here, which I think could be avoided. Since the autosize component does not implement the HTMLElement interface, it means the callers cannot override anything about this intermediary wrapper - we have something that can no longer be controller (I dont know if this will create problems in practice, but I'd rather have more flexibility in case we need it and defer this to whoever is using the component). 

There are two ways to solve it, one is to either implement the interface and spread the props to our ParentSizeMimic component, this way we get full access and can override the wrapper, but the second and I think more interesting option is to **not** render the wrapper at all and instead rely on childRef.parentElement reference to do all the calculations (this way you can even remove the parent ref logic and things like if(!parentRef ||! childRef) logic

gggritso:
@edwardgou-sentry @narsaynorath @JonasBa thanks for the reviews! I made some _major_ changes to the component, so I'm re-requesting the review and marking comments as stale, since so much has changed!

@narsaynorath I added instrumentation! I was going to, I swear, you caught me being lazy. The condition you asked about doesn't exist anymore, so I'm just logging the iteration count
@edwardgou-sentry the code changed a lot, and I'm not calculating disparity ratios anymore, much simpler to just use pixel thresholds

@JonasBa I've addressed, I think a lot of your concerns, thank you for those suggestions:

1. No more React state _at all_, everything is kept in refs
2. Only one resize observer, and only one ref to keep track of the child
3. The iteration runs in a loop and blocks the UI thread to prevent jitters
4. Fewer props and simpler logic! No need to set min and max anymore
5. No more parent mimic component

I did keep _some_ of the old stuff:

- `useLayoutEffect` is still there. I like it too much. It's a handy way to express side effects and cleanup, and I do like to re-mount when the iteration parameter changes. Thoughts?

The overall approach is still the same. I also updated the docs, while I was at it, and added more inline code comments.

Thank you for your reviews, please look again when you have a chance 🙏🏻

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,143 @@
+import {useLayoutEffect, useRef} from 'react';
+import styled from '@emotion/styled';
+import * as Sentry from '@sentry/react';
+
+interface Props {
+  children: React.ReactNode;
+  maximumDifference?: number;
```
JonasBa, line 7: If I were to use a component like this, this isnt something I'd want to have to think about. I think we should remove it, and revisit the component design if this ever becomes a requirement

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+export function AutoSizedText({
+  children,
+  maximumDifference = DEFAULT_MAXIMUM_DIFFERENCE,
+}: Props) {
+  const childRef = useRef<HTMLDivElement>(null);
+
+  const fontSize = useRef<number>(0);
+  const fontSizeLowerBound = useRef<number>(0);
+  const fontSizeUpperBound = useRef<number>(0);
```
JonasBa, line 18: Nit, but with multiple refs, it can more easily happen that you end up not updating some values, especially since they seem related and I see on L42,L43 you assign them as the same time. 

I would have probably opted for a single ref to store the state
```suggestion
  const font = useRef<{current: 0, range: [0, 0]}>({current: 0, range: [0, 0]}); // or separate fields for lower/upper, whatever you prefer 
  
  // this then becomes a single line assignment
  font.current.range = [min, max];
```

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,143 @@
...
+          Math.abs(widthDifference) <= maximumDifference;
+        const childIsWithinHeightTolerance =
+          Math.abs(heightDifference) <= maximumDifference;
+
+        if (
+          childFitsIntoParent &&
+          (childIsWithinWidthTolerance || childIsWithinHeightTolerance)
+        ) {
+          // Stop the iteration, we've found a fit!
+          return;
```
JonasBa, line 72: You need to stop the span, else the setAttribute and end will never be called :)

JonasBa, line 72:
```suggestion
          break;
```

gggritso: No longer a concern now that it's a `break`, right?

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const fontSize = useRef<number>(0);
+  const fontSizeLowerBound = useRef<number>(0);
+  const fontSizeUpperBound = useRef<number>(0);
+
+  useLayoutEffect(() => {
+    const childElement = childRef.current; // This is `SizedChild`
+    const parentElement = childRef.current?.parentElement; // This is the parent of `AutoSizedText`
+
+    if (!childElement || !parentElement) {
+      return () => {};
```
JonasBa, line 25: iirc that consistent return rule doesnt actually care about the type
```suggestion
      return undefined;
```

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,143 @@
...
+      let iterationCount = 0;
+
+      const span = Sentry.startInactiveSpan({
+        op: 'function',
+        name: 'AutoSizedText.iterate',
+      });
+
+      // Run the resize iteration in a loop. This blocks the main UI thread and prevents
+      // visible layout jitter. If this was done through a `ResizeObserver` or React State
+      // each step in the resize iteration would be visible to the user
```
JonasBa, line 54: 🏅 for comments like these

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,143 @@
...
+          // Stop the iteration, we've found a fit!
+          break;
+        }
+
+        adjustFontSize(childDimensions, parentDimensions);
+
+        iterationCount += 1;
+      }
+
+      span.setAttribute('iterationCount', iterationCount);
```
JonasBa, line 80: Wdyt about adding the final disparity as the attribute? This way, if you change the logic in the future, you can also evaluate how accurate the sizing algorithm is vs the iteration count
```suggestion
      span.setAttribute('disparity', final disparity);
```

gggritso: Makes sense! Added into the loop, alongside the iteration count

static/app/views/dashboards/widgetCard/autoSizedText.tsx
gggritso: 🤔 I guess that makes sense but you _don't_ have to think about it, it's optional! I _immediately_ thought that I'd want to change this if I use this in a component where fitment is important (e.g., in a headline or some other important visual element)

JonasBa:
Nice work @gggritso, I think you found a cleaner implementation that is less opinionated than before with the code being easier to follow!

static/app/views/dashboards/widgetCard/autoSizedText.tsx
gggritso: I see what you're saying! Conversely though, I also sometimes update just the lower bound and the font, or just the upper bound and the font. Grouping the bounds together didn't feel great either

static/app/views/dashboards/widgetCard/autoSizedText.tsx
JonasBa: How would you go about guiding folks to set a good value here? What I'm afraid of is that everyone would just set it to 0 because they'd want it to be the exact size (which is kind of the goal already?)

static/app/views/dashboards/widgetCard/autoSizedText.tsx
JonasBa: Nope, all good now. We might want to lint for that actually, now I wonder if there are places in the codebase that have bugs because we allow it 🤔

static/app/views/dashboards/widgetCard/autoSizedText.tsx
gggritso: I think that's okay, actually! In fact, maybe what I should do is set the default to 0 (or 1px just to be safe) and set it to 5px just in Dashboards where there might be many of these numbers and I don't wanna halt the UI thread too long. Thoughts?

JonasBa: How would you evaluate if that is a good config? Does that mean we'd need to care about how many big numbers we render and update this over time?

static/app/views/dashboards/widgetCard/autoSizedText.tsx
```diff
@@ -0,0 +1,145 @@
...
+    ) {
+      // The element is bigger than the parent, scale down
+      newFontSize = (fontSizeLowerBound.current + fontSize.current) / 2;
+      fontSizeUpperBound.current = fontSize.current;
+    } else if (
+      childDimensions.width < parentDimensions.width ||
+      childDimensions.height < parentDimensions.height
+    ) {
+      // The element is smaller than the parent, scale up
+      newFontSize = (fontSizeUpperBound.current + fontSize.current) / 2;
```
JonasBa, line 114: Thoughts on trimming or rounding the floating point values here? I dont think this is an issue tbh, but it might generate some really long floats here :D

gggritso: I could! Ostensibly modern fonts handle sub-pixel rendering gracefully enough. Are you worried about DOM bloat?

static/app/views/dashboards/widgetCard/autoSizedText.tsx
gggritso: Alright! Sold! Removed it, and added a note about iteration counts and performance related to the threshold 👍🏻

static/app/views/dashboards/widgetCard/autoSizedText.tsx
JonasBa: I _think_ the browsers will handle it fine (though different platforms render text quite differently). I think its more likely just going to be an instance where someone inevitably sees a high precision float in the DOM and starts asking around or making up some wrong ideas, which is a conversation you can spare yourself from having :D

JonasBa (approved):
Really nice work, A+ for implementation, stories and code comments!

PR #76211, opened 2024-08-14
feat(recent-search): Add errors and transaction datasets

narsaynorath:
We want to keep the recent search for errors and transactions datasets separate as we split the discover dataset

### DIFF
--- a/src/sentry/models/search_common.py
+++ b/src/sentry/models/search_common.py
@@ -8,3 +8,5 @@ class SearchType(IntEnum):
     REPLAY = 3
     METRIC = 4
     SPAN = 5
+    ERROR = 6
+    TRANSACTION = 7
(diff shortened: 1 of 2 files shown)


narsaynorath:
cc: @Abdkhan14 because I think you're working on splitting the errors and transactions datasets for events search bars. This change lets us split the recent search history

src/sentry/models/search_common.py
```diff
@@ -8,3 +8,5 @@ class SearchType(IntEnum):
     REPLAY = 3
     METRIC = 4
     SPAN = 5
+    ERROR = 6
```
shruthilayaj, line 11: Dang I see we use SearchType.EVENT in places that combine errors and transactions, otherwise we could have reused that

shruthilayaj (approved):

PR #76212, opened 2024-08-14
chore(derived_code_mappings): Report errors with warning level

armenzg:
This reports the exceptions as warning-level errors rather than error-level.

Deployments do not get paused if the error is [warning or lower](https://github.com/getsentry/getsentry/blob/b08907d9da3e8856c6cbedbd86d4a722c6f89dd9/gocd/templates/bash/backend/check-sentry-new-errors.sh#L12):
```
--additional-query="issue.type:error !level:info !level:warning"
```

Rendering of a warning-level exception:
image

### DIFF
--- a/src/sentry/integrations/github/client.py
+++ b/src/sentry/integrations/github/client.py
@@ -8,6 +8,7 @@ from typing import Any, cast
 import orjson
 import sentry_sdk
 from requests import PreparedRequest
+from sentry_sdk import capture_exception, capture_message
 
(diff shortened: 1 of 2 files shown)


JoshFerge (commented):
i'm not sure if this works? given the test failure. i think my idea was to use sentry APIs directly.

armenzg (commented):
There are different execution paths depending on the method used. The logging integration does not allow for it.

leeandher (approved):

PR #76214, opened 2024-08-14
ref(messaging): Extract ABC for linking and unlinking teams

RyanSkonnord:
Rough sketch of how we would generify Slack's team-linking and -unlinking views for other messaging integrations. A few Slack-specific details remain (around the HTML templates and such) but these are the broad strokes.

A working use case for MSTeams and Discord would require some details on how to generify interactions with Slack channels.

Note that this PR only covers the execution of the linking or unlinking operation and the view that is rendered when it is done. The in-client command and the webhook that the client calls are out of scope.


sentaur-athena:
The structure and abstractions look good! Please ping again when it's ready to review.

sentaur-athena (approved):
I reviewed the structure change but didn't dig deep into nitty gritty what functions do because I think they're all the same as before.

Also current tests passing says it works as before. 

Please correct me if I'm wrong and I should pay more attention.

PR #76219, opened 2024-08-14
chore(deps): Bump protobuf to V5

colin-sentry:
We want to start using protobuf for some experimental endpoints, snuba uses v5 so we need to standardize everywhere.

Split off from https://github.com/getsentry/sentry/pull/76208


asottile-sentry:
this is still way too much to upgrade in a single patch -- take it slow and safe and split up the upgrades to minimize impact

asottile-sentry (approved):

PR #76220, opened 2024-08-14
ref(explore): Use sort type from discover

Zylphrex:
Discover already defines a `Sort` type, so let's try to reuse that instead of redefining our own.


wmak (approved):
lgtm

PR #76223, opened 2024-08-14
fix(grouping): Fix `GroupHash.metadata` property

lobsterkatie:
As originally created, the `GroupHashMetadata` table's `grouphash` property is a `FlexibleForeignKey`, which turns out to make the reverse lookup, `GroupHash.metadata`, a `RelatedManager` rather than a direct link to the `GroupHashMetadata` object. This fixes that by changing the type of the `GroupHashMetadata.grouphash` field to a `OneToOneField`. 

I had originally avoided using a `OneToOneField` because I didn't realize `GroupHash.metadata` was allowed to be null with that setup. That turns out to be fine, except that by default, `GroupHash` records without an associated `GroupHashMetadata` error out when you try to get their metadata. This PR therefore also fixes that, by changing the `related_name` on the `grouphash` field to `_metadata` and adding a `metadata` property to the `GroupHash` class which returns `None` rather than crashing if the `GroupHash` has no metadata.

Note: I don't know if under the hood this would prompt any data migration, but it doesn't matter because the table doesn't yet have any records in it.


src/sentry/models/grouphash.py
```diff
@@ -35,3 +42,10 @@ class Meta:
         app_label = "sentry"
         db_table = "sentry_grouphash"
         unique_together = (("project", "hash"),)
+
+    @property
+    def metadata(self) -> GroupHashMetadata | None:
+        try:
+            return self._metadata
+        except Exception:
```
JoshFerge, line 50: is there a more specific `Exception` we can use?

src/sentry/models/grouphashmetadata.py
```diff
@@ -2,15 +2,17 @@
...
 
 
 @region_silo_model
 class GroupHashMetadata(Model):
     __relocation_scope__ = RelocationScope.Excluded
 
     # GENERAL
-    grouphash = FlexibleForeignKey("sentry.GroupHash", unique=True, related_name="metadata")
+    grouphash = models.OneToOneField(
+        "sentry.GroupHash", unique=True, related_name="_metadata", on_delete=models.CASCADE
```
JoshFerge, line 14: is `unique=True` still necessary on a OneToOne field?

JoshFerge (commented):
looks good. no concerns w/ migration as grouphashmetadata has no data in it. couple of small questions/comments

src/sentry/models/grouphashmetadata.py
wedamija: I think it's implicit and we can remove it

wedamija (commented):
migration lgtm

We could probably no-op the migration if we wanted to, since it looks like it removes and re-adds the same constraint. Since this table is empty we don't need to do so - more of a jfyi if you're doing something like this in the future.

src/sentry/models/grouphash.py
wedamija: Yeah, we should use `RelatedObjectDoesNotExist` here

lobsterkatie: (Wrote this before Dan posted)

The type of the error thrown is `sentry.models.grouphash.GroupHash._metadata.RelatedObjectDoesNotExist`, which felt like a real mouthful (and `RelatedObjectDoesNotExist` doesn't seem to be importable on its own, as it's a [dynamically-created type](https://github.com/django/django/blob/a57596e443ecb67140e1a9fc0f0e87446b2d0174/django/db/models/fields/related_descriptors.py#L135-L150)), which is why I went with `Exception`. But it turns out `RelatedObjectDoesNotExist` errors are just `AttributeErrors` (which makes a lot of sense), so I can catch that instead.

src/sentry/models/grouphashmetadata.py
lobsterkatie: Yup, you're both right - turns out it's not necessary, because it's [set by default](https://github.com/django/django/blob/a57596e443ecb67140e1a9fc0f0e87446b2d0174/django/db/models/fields/related.py#L1235). Good catch! I'll take it out.

lobsterkatie:
> We could probably no-op the migration if we wanted to, since it looks like it removes and re-adds the same constraint. Since this table is empty we don't need to do so - more of a jfyi if you're doing something like this in the future.

Makes sense. How do we do that (or would we do that), while still telling Django about the change?

JoshFerge (approved):
looks good.

wedamija:
> > We could probably no-op the migration if we wanted to, since it looks like it removes and re-adds the same constraint. Since this table is empty we don't need to do so - more of a jfyi if you're doing something like this in the future.
> 
> Makes sense. How do we do that (or would we do that), while still telling Django about the change?

You'd use `SeparateDatabaseAndState` and move all of the operations just into the state, so that Django is aware of the changes, but doesn't do anything to the db. No need to do so here unless you just want to see how it works

PR #76227, opened 2024-08-14
fix(uptime): Always pass span_id from check result to occurrence

evanpurkhiser:


wedamija (approved):
Do we need to use `result.get` since this is guaranteed?

evanpurkhiser:
> Do we need to use result.get since this is guaranteed?

I need to bump the kafka schemas package

PR #76231, opened 2024-08-14
ref(control_silo): Move UserOption model to users module

Christinarlong:
Part of moving control silo user related resources into the users module. Includes adding import shim for any getsentry refs.

Apart of (#73856)


markstory (approved):
:tada:

PR #76234, opened 2024-08-14
fix(issues): Allow dropdowns in section actions, use divs

scttcper:
- Move title styles to a title wrapper to avoid text styles to be applied to the action menus
- Switch from details/summary to just divs because the submenu would need to prevent default in order to stop it from collapsing


roggenkemper (approved):

static/app/views/issueDetails/streamline/foldSection.tsx
```diff
@@ -160,23 +160,25 @@ const Content = styled('div')`
   padding: ${space(0.5)} ${space(0.75)};
 `;
 
-const Summary = styled('summary')<{preventCollapse: boolean}>`
+const Summary = styled('div')<{preventCollapse: boolean}>`
```
leeandher, line 163: nit: Since it's not a summary anymore would you mine renaming this to something more accurate? I only named it that because it was a component version of `summary`

leeandher (approved):

PR #76236, opened 2024-08-14
fix(alert): Correct url

schew2381:
URL was incorrectly changed in https://github.com/getsentry/sentry/pull/76204

Will add test in future PR 🫥


ceorourke (approved):

evanpurkhiser:
Can we update the test for this as well?

schew2381:
> Can we update the test for this as well?

Was in a rush to go to something so I didn't have time to add it, will do later today 👍

PR #76239, opened 2024-08-14
chore(slack): Clean Utils `__init__.py`

iamrajjoshi:
cleaned up Slack's utils `__init__.py` file.

biggest change was removing the logger from there and defining it individually in each file. 

i also updated some logs to exceptions since it will capture the exception for us. i also had to update the logger key because formatted strings were failing flake8 tests.


nhsiehgit (approved):

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -842,10 +843,9 @@ def handle_member_approval(self, slack_request: SlackActionRequest, action: str)
                 member.approve_member_invitation(identity_user, referrer="slack")
             else:
                 member.reject_member_invitation(identity_user)
-        except Exception as err:
+        except Exception:
             # shouldn't error but if it does, respond to the user
-            logger.error(
-                err,
+            _logger.exception(
```
asottile-sentry, line 848: this breaks the call here:

```pycon
>>> logging.getLogger(__name__).exception(extra={})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Logger.exception() missing 1 required positional argument: 'msg'
```

asottile-sentry: mypy could have caught this but it's disabled on that file:
```
src/sentry/integrations/slack/webhooks/action.py:905: error: Missing positional argument "msg" in call to "exception" of "Logger"  [call-arg]
```

PR #76240, opened 2024-08-14
nit(scm): group functions by class they come from in main integration class

cathteng:
The `__Integration` classes (e.g. `GitHubIntegration`) inherit from several classes (ABCs), and they each implement functions from those parent classes. This PR moves functions such that they are grouped together via the ABC they come from, e.g. `IntegrationInstallation` functions or `RepositoryIntegration` functions.


iamrajjoshi (approved):
:shipit:

PR #76245, opened 2024-08-15
feat(grouping): Store metadata for new grouphashes

lobsterkatie:
As part of the rollout of the new grouphash metadata table, this creates a `GroupHashMetadata` row for every new `GroupHash` we create. It's gated by a killswitch and a feature flag, and doesn't do much yet - the only data in the table so far is a creation timestamp - but this should give us a sense of whether or not there's going to be any hit to ingest DB performance.

### DIFF
--- a/src/sentry/grouping/ingest/hashing.py
+++ b/src/sentry/grouping/ingest/hashing.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING
 
 import sentry_sdk
 
+from sentry import features, options
 from sentry.exceptions import HashDiscarded
 from sentry.features.rollout import in_random_rollout
 from sentry.grouping.api import (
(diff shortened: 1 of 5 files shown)


src/sentry/grouping/ingest/hashing.py
```diff
@@ -340,3 +342,22 @@ def get_hash_values(
...
+    project: Project, calculated_hashes: CalculatedHashes
+) -> list[GroupHash]:
+    grouphashes = []
+
+    for hash_value in extract_hashes(calculated_hashes):
+        grouphash, created = GroupHash.objects.get_or_create(project=project, hash=hash_value)
+
+        # TODO: Do we want to expand this to backfill metadata for existing grouphashes? If we do,
+        # we'll have to override the metadata creation date for them.
+        if created and options.get("grouping.grouphash_metadata.ingestion_writes_enabled"):
```
jangjodi, line 357: do we want to roll this out more slowly if we think there's going to be a hit in ingest DB performance (ie. in s4s first)?

lobsterkatie: Josh and I agreed that as long as there's a killswitch we're fine. Keep in mind that for now this is just new hashes, which only come either during a config transition or with new groups (which happen in < 1% of events).

JoshFerge (commented):
talked during meeting, should ramp up writes slowly just to be safe

lobsterkatie:
> talked during meeting, should ramp up writes slowly just to be safe

Yup - you beat me by about two minutes. 🙂 Just pushed it again, including a feature flag.

JoshFerge (approved):

PR #76249, opened 2024-08-15
feat(alerts): Allow 'fake' anomalies for testing

ceorourke:
In order to do end to end testing in production (and because I am not able to get ngrok working to validate notification content changes) we'll allow for a 'fake' anomaly to come through if the flag is turned on so we can see these rules trigger.

### DIFF
--- a/src/sentry/incidents/subscription_processor.py
+++ b/src/sentry/incidents/subscription_processor.py
@@ -514,6 +514,9 @@ class SubscriptionProcessor:
         self.has_anomaly_detection = features.has(
             "organizations:anomaly-detection-alerts", self.subscription.project.organization
         )
+        fake_anomalies = features.has(
+            "organizations:fake-anomaly-detection", self.subscription.project.organization
+        )
 
         if self.has_anomaly_detection:
             potential_anomalies = self.get_anomaly_data_from_seer(aggregation_value)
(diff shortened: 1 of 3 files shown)


src/sentry/incidents/subscription_processor.py
```diff
@@ -573,7 +576,7 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
                             self.trigger_alert_counts[trigger.id] = 0
 
                         if (
-                            not self.has_anomaly(potential_anomaly, trigger.label)
+                            not self.has_anomaly(potential_anomaly, trigger.label, fake_anomalies)
```
scefali, line 579: @ceorourke why not just just do an OR here:
```
self.has_anomaly(potential_anomaly, trigger.label) or fake_anomalies
```
Seems a bit strange to pass it to this function and have it bypass the the whole function

src/sentry/incidents/subscription_processor.py
```diff
@@ -514,6 +514,9 @@ def process_update(self, subscription_update: QuerySubscriptionUpdate) -> None:
         self.has_anomaly_detection = features.has(
             "organizations:anomaly-detection-alerts", self.subscription.project.organization
         )
+        fake_anomalies = features.has(
```
saponifi3d, line 517: nit: `has_fake_anomalies` - i just got a bit confused later when i was reading code thinking that these were a list of fake anomalies rather than a boolean flag.

saponifi3d (approved):

PR #76250, opened 2024-08-15
feat(nodestore): A file system-based node storage backend, with added support for S3 and GCS storage

klboke:
A file system-based node storage backend, with added support for S3 and GCS storage.

My starting point is as follows:

- Using PostgreSQL storage by default can lead to rapid storage expansion due to PostgreSQL's own table cleaning mechanisms, and it cannot perform automatic cleanup. For example: https://develop.sentry.dev/self-hosted/troubleshooting/#postgres 

### Legal Boilerplate


PMExtra:
Great job! But I worry a little about the performance while reading multiple objects as a batch. (I didn't dig it, just an intuitive guess)

PMExtra:
I'm not sure how frequently the node storage is accessed or how large the stored files are.

I don't think object storage services (such as S3 or GCS) are designed for frequent reads and writes of very small files.

This may lead to performance issues or higher costs.

klboke:
@PMExtra 

> I'm not sure how frequently the node storage is accessed or how large the stored files are.

Based on our recorded data, each node data is approximately between 15~76KB in size, with the majority being around 15KB

> This may lead to performance issues or higher costs.

I investigated the places where nodestore is written to and read from and found that there are only two scenarios that trigger these actions: 
- 1. Writing to nodestore occurs in the Sentry worker when processing Kafka event messages. 
- 2. Reading from nodestore happens in sentry-web when someone views detailed information. 

Writing is essentially offline data stream processing, so a slower speed is acceptable. The QPS (queries per second) during reading is very low, so it is not a significant issue either. Additionally, OSS storage costs are definitely the lowest, which is why many projects in recent years (like Paimon and OpenObserve) support the separation of storage (such as S3 and other object storage) and computation.

PMExtra:
@klboke Awesome. I am concerned that reading a large number of tiny files might incur higher costs. However, according to your analysis, the files are not that small, and the access frequency is not that high. Therefore, my concern may be unnecessary.

PMExtra:
@klboke 

Another point worth noting is that Sentry stores a large amount of highly repetitive ASCII text, and table store services can achieve high compression rates for this, thereby reducing storage costs.

However, object storage services usually do not support transparent compression. While each piece of data can be compressed separately, the additional computational overhead will be incurred, and the compression rate may be difficult to match compared to table store services.

Anyway, having one more option is always good. I’m just pointing out some potential drawbacks for discussion, not to undermine your work. Thank you for your contribution.

klboke:
@PMExtra  
> However, object storage services usually do not support transparent compression. While each piece of data can be compressed separately, the additional computational overhead will be incurred, and the compression rate may be difficult to match compared to table store services.

Thank you very much for your reminder and suggestions. If we switch to S3, considering our 3TB nodestore (which actually is more than most people have), cost won't be an issue at all. I would prefer storing plain JSON files directly, as it would be more convenient for observing and debugging some issues.

src/sentry/nodestore/filesystem/backend.py
```diff
@@ -1,53 +1,58 @@
...
-            os.mkdir(self.path)
-        except FileExistsError:
-            pass
-
-    def node_path(self, id: str) -> str:
-        return os.path.join(self.path, f"{id}.json")
+    def __init__(self, prefix_path=None):
+        self.prefix_path: str = "nodestore/"
+        if not settings.DEBUG and options_store.get("filestore.backend") == "filesystem":
+            raise ValueError("Local fileSystem should only be used in development!")
```
aldy505, line 27: This would break existing self-hosted users that don't use (or don't have access to) S3 compatible services. This should just log a warning instead of throwing a ValueError.

PMExtra: I don't think so. The default `NodeStore` is `sentry.nodestore.django.DjangoNodeStorage` rather than this. So it won't break any existing users. If you want to configure this `FileSystemNodeStorage` in non-debug environment, you must configure `FileStore` also.

rophy: I'd like to see this merged. 
Can we settle with being more backward compatible?

i.e. do not make the assumption that "If you want to configure this FileSystemNodeStorage in non-debug environment, you must configure FileStore also."

rophy: Ping to bump this PR.

I would really like to see the built-in support for S3 Node Store (even though performance is not endorsed).

I think the concern here is:   `FileSystemNodeStorage` by title is supposed to be using "File System". It is not intuitive that by enabling "S3 File Store", FileSystemNodeStore will push things to S3.

I think better way is to explicitly create a new S3NodeStore class with those logic.

rophy: See if #82125 makes sense to everyone.

cleptric:
I reached out internally and looks like there is no appetite to move forward with this, hence I'm closing it.

PR #76254, opened 2024-08-15
test(alerts): Write test for deleting metric alert rule

schew2381:
Writing test followup to https://github.com/getsentry/sentry/pull/76236


static/app/views/alerts/list/rules/alertRulesList.spec.tsx
```diff
@@ -179,35 +179,71 @@ describe('AlertRulesList', () => {
     expect(screen.getByText('Duplicate')).toBeInTheDocument();
   });
 
-  it('deletes a rule', async () => {
-    const {router, organization} = initializeOrg({
-      organization: defaultOrg,
+  it('deletes an issue rule', async () => {
```
schew2381, line 182: cleaned up this test as well

nhsiehgit (approved):

PR #76255, opened 2024-08-15
chore: Move function out of receivers

schew2381:
Need to move more imports out of receivers in order for getsentry imports to work with upcoming default metric alert changes

### DIFF
--- a/src/sentry/utils/db.py
+++ b/src/sentry/utils/db.py
@@ -1,9 +1,28 @@
+import logging
 from collections.abc import Sequence
 from contextlib import ExitStack
+from functools import wraps
 
 import sentry_sdk
 from django.db import DEFAULT_DB_ALIAS, connections, transaction
+from django.db.utils import OperationalError, ProgrammingError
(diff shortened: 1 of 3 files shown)


src/sentry/utils/db.py
```diff
@@ -1,11 +1,30 @@
...
+from django.db.utils import OperationalError, ProgrammingError
 from sentry_sdk.integrations import Integration
 
 
+def handle_db_failure(func, using: str = "default", wrap_in_transaction: bool = True):
+    @wraps(func)
+    def wrapped(*args, **kwargs):
+        try:
+            if wrap_in_transaction:
+                with transaction.atomic(using):
```
schew2381, line 17: I'm unable to use `router.db_for_write(Organization)` b/c this causes another import loop and there's really no other place to put this.

It's not an issue b/c we never pass in `using` so we were already defaulting to `router.db_for_write(Organization)` which just returns `default`.

markstory (approved):
Looks good to me.

PR #76259, opened 2024-08-15
feat(profiling)fetch function metrics when expanding the function in the slowest function list

JonasBa:
Add the data fetching hook and expand/close functionality

### DIFF
--- a/static/app/utils/profiling/hooks/useProfilingFunctionMetrics.tsx
+++ b/static/app/utils/profiling/hooks/useProfilingFunctionMetrics.tsx
@@ -1,0 +1,46 @@
+import {useMemo} from 'react';
+
+import {normalizeDateTimeParams} from 'sentry/components/organizations/pageFilters/parse';
+import {useApiQuery} from 'sentry/utils/queryClient';
+import useOrganization from 'sentry/utils/useOrganization';
+import usePageFilters from 'sentry/utils/usePageFilters';
+
+interface UseProfilingFunctionMetricsProps {
+  fingerprint: Profiling.FunctionMetric['fingerprint'];
+  projects: number[];
+  query?: string;
+}
+
+export function useProfilingFunctionMetrics(props: UseProfilingFunctionMetricsProps) {
+  const organization = useOrganization();
+  const {selection} = usePageFilters();
+
+  const path = `/organizations/${organization.slug}/events-stats/`;
+  const endpointOptions = useMemo(() => {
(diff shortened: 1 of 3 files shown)


static/app/utils/profiling/hooks/useProfilingFunctionMetrics.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const path = `/organizations/${organization.slug}/events-stats/`;
+  const endpointOptions = useMemo(() => {
+    const params: {
+      query: Record<string, any>;
+    } = {
+      query: {
+        project: props.projects,
+        environment: selection.environments ?? selection.environments,
+        ...normalizeDateTimeParams(selection.datetime ?? selection.datetime),
+        dataSource: 'profilesMetrics',
```
Zylphrex, line 27: I believe this is called `profileFunctionsMetrics`

static/app/utils/profiling/hooks/useProfilingFunctionMetrics.tsx
```diff
@@ -0,0 +1,46 @@
...
+        fingerprint: props.fingerprint,
+        query: props.query,
```
Zylphrex, line 29: I dont think it takes a fingerprint as a query param. Instead, you should be passing the fingerprint inside the query like
```suggestion
        query: `${props.fingerprint}`,
```

static/app/views/profiling/landing/slowestFunctionsTable.tsx
```diff
@@ -238,6 +259,29 @@ function SlowestFunctionsProjectBadge(props: SlowestFunctionsProjectBadgeProps)
...
+
+function SlowestFunctionTimeSeries(props: SlowestFunctionTimeSeriesProps) {
+  const projects = useMemo(() => {
+    return props.function.examples
+      .map(f => {
+        if (typeof f !== 'string' && 'project_id' in f) return f.project_id;
+        return null;
+      })
+      .filter(n => n !== null);
+  }, [props.function]);
```
Zylphrex, line 274: Nit: you may want to dedupe project ids from this list

Zylphrex (approved):

static/app/views/profiling/landing/slowestFunctionsTable.tsx
JonasBa: Ah yes, right. Let me do that

PR #76261, opened 2024-08-15
chore: Add another span to ratelimit middleware

markstory:
Still trying to get to the bottom of mystery latency in ratelimiting that shows up in internal RPC traffic.


wedamija (approved):
Are we able to run profiles here at all? If we are it might be easier than playing whack-a-mole with spans

markstory:
I haven't checked profiling yet, but if this span doesn't help with attributing the phantom latency, I'll check that next.

PR #76265, opened 2024-08-15
feat(dashboards): Split recent searches for dataset split

narsaynorath:
The widget builder has distinct datasets for errors and transactions, so the recent search history should also be distinct. Adds the enum entries for the saved search type and passes it along in the respective configs.


shruthilayaj (approved):

shruthilayaj:
PS @narsaynorath This only applies to the widget builder - can we also add it to Discover?

PR #76268, opened 2024-08-15
feat(explore): Basic explore span table rendering

Zylphrex:
This queries the spans table in explore and does some basic table rendering with no styling yet.


Abdkhan14 (approved):
🚢

PR #76272, opened 2024-08-15
ref(control_silo):  Move Identitiy model to users module

Christinarlong:
Part of moving control silo user related resources into the users module. Includes adding import shim for any getsentry refs. Also adds minor typing to some identity functions (will point out).

Apart of (https://github.com/getsentry/sentry/issues/73856)


src/sentry/users/models/identity.py
```diff
@@ -0,0 +1,216 @@
...
+
+    class Meta:
+        app_label = "sentry"
+        db_table = "sentry_identityprovider"
+        unique_together = (("type", "external_id"),)
+
+    def get_provider(self) -> Identity:
+        from sentry.identity import get
+
+        return get(self.type)
```
Christinarlong, line 68: I interpreted the `get()` function returning a `cls("Identity")` which would just be an Identity class... (???)

markstory: Wouldn't this be `IdentityProvider`. The `get()` will return an instance of the registered types (Providers) based on the string name provided.

Providers are registered here https://github.com/getsentry/sentry/blob/4fb67fb784549608feffa3798db2dadd2a4fb6a2/src/sentry/identity/__init__.py#L25-L34

The `Slack` provider is found https://github.com/getsentry/sentry/blob/6ef91129b58b3627e9705d0ad71c1bf458567919/src/sentry/identity/slack/provider.py#L5

and the class tree for those ends in `Provider` https://github.com/getsentry/sentry/blob/6ef91129b58b3627e9705d0ad71c1bf458567919/src/sentry/identity/base.py#L7

Christinarlong: Ohhhh, gotchas. I completely missed the whole provider registration and that the `self.values` in `IdentityManager` was populated with Providers from the registration.

markstory (approved):

PR #76274, opened 2024-08-15
chore: Use new import

schew2381:
Followup to https://github.com/getsentry/sentry/pull/76273 where I actually use the new import in sentry


schew2381:
> @schew2381 the paths would need to be updated in `getsentry` before this pull request can be merged

yup got it this is just the third step

dashed (approved):

PR #76275, opened 2024-08-15
feat(onboarding): migration to add `disable_member_invite` bit flag to hybrid cloud services

ameliahsu:
recently merged a [migration](https://github.com/getsentry/sentry/pull/76116) to add `disable_member_invite` to the organization bitflag and now need to add it to hybrid cloud services to sync


wedamija (commented):
Migration lgtm

iamrajjoshi (approved):
lgtm

markstory (approved):
Looks good to me. We deploy regions first, so when control is deployed it will already be receiving the new flag field and upsert() will continue to behave correctly.

GabeVillalobos (approved):
Looks good! For future reference, it might be good to split migration changes from the implementation logic just to be safe.

PR #76276, opened 2024-08-15
feat(replay): implement search bar key sections and fetch tags from IP instead of discover

aliu39:
Follow up to https://github.com/getsentry/sentry/pull/75552. Implements sections to make the typeahead easier to navigate.
[Screen recording](https://drive.google.com/file/d/1HxV87JgZWaHHuA74EEdIOdbL1sDxNTzd/view?usp=sharing)

"Suggested" and "Click Fields" combined is = to our documented search properties: https://docs.sentry.io/concepts/search/searchable-properties/session-replay. We always want to show these.

"Tags" are tags found in the Issue Platform dataset, sorted by times seen in the dataset (popularity). Previously we were fetching these from the TagStore which queries Discover. IP has a narrower set of more relevant tags. Some more context is at https://github.com/getsentry/sentry/pull/75878

### DIFF
--- a/static/app/views/replays/list/replaySearchBar.tsx
+++ b/static/app/views/replays/list/replaySearchBar.tsx
@@ -54,9 +54,11 @@ const REPLAY_FIELDS_AS_TAGS = fieldDefinitionsToTagCollection(REPLAY_FIELDS);
 const REPLAY_CLICK_FIELDS_AS_TAGS = fieldDefinitionsToTagCollection(REPLAY_CLICK_FIELDS);
 
 /**
- * Merges a list of supported tags and replay search fields into one collection.
+ * Merges a list of supported tags and replay search properties
+ * (https://docs.sentry.io/concepts/search/searchable-properties/session-replay/)
+ * into one collection.
  */
-function getReplaySearchTags(supportedTags: TagCollection): TagCollection {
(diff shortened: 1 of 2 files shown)


static/app/views/replays/list/replaySearchBar.tsx
```diff
@@ -50,33 +52,69 @@ function fieldDefinitionsToTagCollection(fieldKeys: string[]): TagCollection {
 
 const REPLAY_FIELDS_AS_TAGS = fieldDefinitionsToTagCollection(REPLAY_FIELDS);
 const REPLAY_CLICK_FIELDS_AS_TAGS = fieldDefinitionsToTagCollection(REPLAY_CLICK_FIELDS);
+const EXCLUDED_TAGS = ['browser', 'device', 'os', 'user'];
```
michellewzhang, line 55: why are these excluded?

aliu39: we already have browser.name, device.name, etc. They're effectively the same

aliu39: maybe user is actually user.id, but either way they're confusing

michellewzhang: ohh gotcha, could we add a comment to clarify that? just for posterity

michellewzhang (approved):

PR #76277, opened 2024-08-15
feat(custom-views): Add hook for fetching issue counts

MichaelSun48:
Adds a useAPIQuery hook for the `organization/.../issues-count` endpoint. This is going to be used by the custom views component to fetch the query counts for new tabs. 

Update 1/6: We've decided to re-add this functionality, which is why this PR has been reopened


static/app/views/issueList/queries/useFetchIssueCounts.tsx
```diff
@@ -0,0 +1,30 @@
...
+    },
+  },
+];
+
+export const useFetchGroupSearchViews = (
+  params: FetchIssueCountsParameters,
+  options: Partial<UseApiQueryOptions<GroupSearchView[]>> = {}
+) => {
+  return useApiQuery<GroupSearchView[]>(makeFetchIssueCounts(params), {
+    staleTime: Infinity,
```
malwilley, line 27: Counts seem likely to change, so I'd make this `staleTime: 0` or something small like 10 seconds

static/app/views/issueList/overview.tsx
```diff
@@ -139,7 +139,7 @@ interface EndpointParams extends Partial<PageFilters['datetime']> {
   useGroupSnubaDataset?: boolean;
 }
 
-type CountsEndpointParams = Omit<EndpointParams, 'cursor' | 'page' | 'query'> & {
+export type CountsEndpointParams = Omit<EndpointParams, 'cursor' | 'page' | 'query'> & {
```
malwilley, line 142: I'd either duplicate this type in `useFetchIssueCounts` or add it there and export here. Generally we don't want utils importing from components (should be the other way around)

malwilley (approved):

PR #76281, opened 2024-08-15
feat(insights): add geo region filter to http module

DominikB2014:
work towards https://github.com/getsentry/sentry/issues/75230

add geo region selector to http module, we may or may not want to show this for backend projects, but we can hide it in another PR if necessary

image

### DIFF
--- a/static/app/views/insights/http/views/httpLandingPage.tsx
+++ b/static/app/views/insights/http/views/httpLandingPage.tsx
@@ -9,7 +9,7 @@ import SearchBar from 'sentry/components/searchBar';
 import {t} from 'sentry/locale';
 import {trackAnalytics} from 'sentry/utils/analytics';
 import {browserHistory} from 'sentry/utils/browserHistory';
-import {decodeScalar, decodeSorts} from 'sentry/utils/queryString';
+import {decodeList, decodeScalar, decodeSorts} from 'sentry/utils/queryString';
 import {MutableSearch} from 'sentry/utils/tokenizeSearch';
 import useLocationQuery from 'sentry/utils/url/useLocationQuery';
 import {useLocation} from 'sentry/utils/useLocation';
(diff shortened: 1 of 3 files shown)


static/app/views/insights/http/views/httpLandingPage.tsx
```diff
@@ -51,15 +53,26 @@ export function HTTPLandingPage() {
...
     },
   });
 
+  const ADDITONAL_FILTERS = {
+    ...(query[SpanMetricsField.USER_GEO_SUBREGION].length > 0
+      ? {
+          [SpanMetricsField.USER_GEO_SUBREGION]: `[${query[SpanMetricsField.USER_GEO_SUBREGION].join(',')}]`,
+        }
+      : {}),
+  };
```
0Calories, line 66: this block of code is pretty tough to read despite it doing something fairly simple, can we just use an if statement instead? i.e.


```suggestion

  const ADDITIONAL_FILTERS = {};
  
  if (query[SpanMetricsField.USER_GEO_SUBREGION].length > 0) {
    ADDITIONAL_FILTERS[SpanMetricsField.USER_GEO_SUBREGION] = `[${query[SpanMetricsField.USER_GEO_SUBREGION].join(',')}]`
  }
```

0Calories (approved):

static/app/views/insights/http/views/httpLandingPage.tsx
DominikB2014: @0Calories yeah that looks a lot cleaner! I'll keep your suggestion

I think another way to clean it up would be to use a variables, I feel like all the extra spacing and newlines from the formatter makes it harder to read.  For example
```
  const {USER_GEO_SUBREGION} = SpanMetricsField;
  const subregions = query[SpanMetricsField.USER_GEO_SUBREGION]
  
  const ADDITONAL_FILTERS = {
    ...(subregions.length > 0 ? { [USER_GEO_SUBREGION]: `[${subregions.join(',')}]` }: {}),
  };
```

PR #76283, opened 2024-08-15
feat(issue-details): Enable a toggle for setting the new UI as preferred

leeandher:
Requires https://github.com/getsentry/sentry/pull/76282 for saving user prefs.

This PR adds back the 'New Experience' header button to the UI if:
- `organization` has the `issue-details-streamline` set to `true`

The new UI will be used if any of the following are true:
- `streamline` query param is set to `1`
- User viewing the issue has `prefersIssueDetailsStreamlinedUI` option set to `true`

Also adds a feedback button if available:
image

https://github.com/user-attachments/assets/ffee1105-6199-4593-a737-5bf05855be83

**todo**
- [x] Add test for conditions to make button appear
- [x] Add tests for `useHasStreamlinedUI` changes
- [x] Add test for navigation on click
- [x] Add analytics for toggles


static/app/views/issueDetails/actions/index.tsx
```diff
@@ -427,6 +427,7 @@ export function Actions(props: Props) {
             />
           </Fragment>
         ))}
+      {hasStreamlinedUI && <NewIssueExperienceButton />}
```
roggenkemper, line 430: do we need this `hasStreamlinedUI` check if we also check for it in the Button?

static/app/views/issueDetails/actions/newIssueExperienceButton.tsx
```diff
@@ -1,69 +1,75 @@
...
-      });
-    },
-    onMutate: (options: UpdateUserOptionsVariables) => {
-      ConfigStore.set('user', merge({}, user, {options}));
-    },
-    onError: () => {
-      addErrorMessage('Failed to save new issue experience preference');
-    },
-  });
+  const streamlineButton = openForm ? (
```
roggenkemper, line 27: can we change the name to mention feedback

static/app/types/user.tsx
```diff
@@ -46,11 +46,11 @@ export interface User extends Omit<AvatarUser, 'options'> {
     avatarType: Avatar['avatarType'];
     clock24Hours: boolean;
     defaultIssueEvent: 'recommended' | 'latest' | 'oldest';
-    issueDetailsNewExperienceQ42023: boolean;
     language: string;
     stacktraceOrder: number;
     theme: 'system' | 'light' | 'dark';
     timezone: string;
+    prefersIssueDetailsStreamlinedUI?: boolean;
```
roggenkemper, line 53: is there a reason why it's optional? won't we return 'false' in all the cases where it isn't set

static/app/views/issueDetails/actions/index.tsx
scttcper: i think it makes sense

scttcper: nvm see what you're saying

static/app/views/issueDetails/utils.tsx
```diff
@@ -273,13 +274,13 @@ export function getGroupEventDetailsQueryData({
...
-  const organization = useOrganization();
+  const user = useUser();
   if (location.query.streamline === '0') {
     return false;
   }
+
   return (
-    location.query.streamline === '1' ||
-    organization.features.includes('issue-details-streamline')
+    location.query.streamline === '1' || !!user?.options?.prefersIssueDetailsStreamlinedUI
```
scttcper, line 283: could probably remove the query param stuff if there's a button

static/app/views/issueDetails/actions/index.tsx
leeandher: It's so that it doesn't appear twice for the non streamlined header. It just has to do with how this component is laid out since the other render of `<NewIssueExperienceButton />` requires `hasStreamlinedUI` to be false.

static/app/views/issueDetails/utils.tsx
leeandher: That makes sense, though I'd prefer to leave it in for now because of one case.

For testing on issues from orgs without the flag, to access the streamlined UI, you'd have to open an issue from an org with it, and enable the UI to set your user preference, then refresh the issue you were looking at, rather than just changing the UI.

That and it's useful for sharing links to the pages while its in development.

leeandher: I guess one way around this is for us to have the button conditional on if the user has set the user option at all, not it's value, since that implies they've accessed the new UI before, but for now I think it's okay to keep the query param.

static/app/views/issueDetails/actions/index.tsx
roggenkemper: this might not work but could we have the button be controlled by the feature flag ? if they have the flag, they see the button , if not they don't. we could keep the query param so people could look at a single issue with the new design if they wanted to. i feel like im getting slightly confused at all the different pieces here and centralizing it around that might help

leeandher: That makes sense, it is getting complicated. Lemme try to simplify it then.

leeandher: Changed it and tests, the button is now only controlled by the organization flag. The actual UI is controlled by user preferences and the query param.

roggenkemper (approved):

scttcper (approved):

PR #76285, opened 2024-08-15
fix(query-builder): Properly position absolute date picker overlay

malwilley:
Forgot to fix the absolute date picker when merging the changes in https://github.com/getsentry/sentry/pull/76226

It was still usable, but rendered inline which causes some terrible styling issues.


static/app/components/searchQueryBuilder/tokens/filter/specificDatePicker.tsx
```diff
@@ -102,109 +104,111 @@ function SpecificDatePicker({
...
-                }}
-              />
-              {t('Include time')}
-            </CheckboxLabel>
-            <TimeUtcWrapper>
-              <TimeInput
-                disabled={!hasTime}
-                time={time ?? DEFAULT_DAY_START_TIME}
-                setTime={newTime => {
+    <StyledPositionWrapper {...overlayProps} visible={isOpen}>
```
malwilley, line 107: Hide whitespace and you'll need that I am just wrapping this component in the position wrapper div

roggenkemper (approved):

PR #76288, opened 2024-08-15
feat(issue-details): Add support for multiple suspect commits in streamlined design

roggenkemper:
this pr updates suspect commits to support multiple suspect commits in the updated design. to make the scrolling work, a few changes are made to the overall style of the page, but they mostly shouldn't be visible (expect for the box shadow). 
Screenshot 2024-08-15 at 1 11 18 PM


scttcper (approved):
they seem a bit wide on wider screens

PR #76290, opened 2024-08-15
ref: fix typing of sentry.integrations.message_builder

asottile-sentry:

### DIFF
--- a/src/sentry/integrations/message_builder.py
+++ b/src/sentry/integrations/message_builder.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
-from collections.abc import Callable, Iterable, Mapping, Sequence
-from typing import Any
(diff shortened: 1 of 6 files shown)


src/sentry/integrations/message_builder.py
```diff
@@ -19,37 +19,50 @@
 from sentry.utils.http import absolute_uri
 
 
-def format_actor_options(
-    actors: Iterable[Team | RpcUser], is_slack: bool = False
```
asottile-sentry, line 23: it's technically possible to preserve the old functions here with `@overload` but I figured it's best to just kill the [boolean trap](https://youtu.be/CnRkXO_a5mI) rather than encourage it

joshuarli (approved):

PR #76292, opened 2024-08-15
log(replay): capture message if mobile replay has mismatching touch events

michellewzhang:
log the sdk version & name if we have mismatching touch events (e.g. start without an end or vice versa) to help us debug https://github.com/getsentry/team-replay/issues/466

### DIFF
--- a/static/app/components/replays/replayContext.tsx
+++ b/static/app/components/replays/replayContext.tsx
@@ -431,6 +431,10 @@ export function Provider({
         events: events ?? [],
         // common to both
         root,
+        context: {
+          sdkName: replay?.getReplay().sdk.name,
+          sdkVersion: replay?.getReplay().sdk.version,
+        },
       });
       // `.current` is marked as readonly, but it's safe to set the value from
       // inside a `useEffect` hook.
--- a/static/app/components/replays/videoReplayerWithInteractions.tsx
+++ b/static/app/components/replays/videoReplayerWithInteractions.tsx
@@ -137,7 +137,7 @@ export class VideoReplayerWithInteractions {
             tags: {
               sdk_name: context.sdkName,
               sdk_version: context.sdkVersion,
-              type: typeof t,
+              touch_event_type: typeof t,
             },
           }
         );
--- a/static/app/components/replays/videoReplayerWithInteractions.tsx
+++ b/static/app/components/replays/videoReplayerWithInteractions.tsx
@@ -134,7 +134,11 @@ export class VideoReplayerWithInteractions {
         Sentry.captureMessage(
           'Mobile replay has mismatching touch start and end events',
           {
-            tags: {sdk_name: context.sdkName, sdk_version: context.sdkVersion},
+            tags: {
+              sdk_name: context.sdkName,
+              sdk_version: context.sdkVersion,
+              type: typeof t,
+            },
           }
         );
       }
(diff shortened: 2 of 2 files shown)


static/app/components/replays/replayContext.tsx
```diff
@@ -431,6 +431,10 @@ export function Provider({
         events: events ?? [],
         // common to both
         root,
+        context: {
+          sdkName: replay?.getReplay().sdk.name,
+          sdkVersion: replay?.getReplay().sdk.version,
```
bruno-garcia, line 436: `getReplay` this is going throgh some list to find the integration right?
No easy way in TS to (pseudo-C#):

```
if replay?.getReplay is replayIntegration {
  sdkName: replayIntegration.sdk.name,
  sdkVersion: replayIntegration.sdk.version,
```
and avoid the double lookup?

michellewzhang: `getReplay()` just returns `this._replayRecord` -- no integration lookup!

michellewzhang: https://github.com/getsentry/sentry/blob/b965109ad3fd7da0af9e824a7338b22213dc0263/static/app/utils/replays/replayReader.tsx#L518-L520

static/app/components/replays/videoReplayerWithInteractions.tsx
```diff
@@ -121,6 +124,22 @@ export class VideoReplayerWithInteractions {
       }
     });
 
+    // log instances where we have a pointer touchStart without a touchEnd
+    const touchEvents = eventsWithSnapshots.filter(
+      (e: eventWithTime) => isTouchEnd(e) || isTouchStart(e)
+    );
+    const grouped = Object.groupBy(touchEvents, (t: any) => t.data.pointerId);
+    Object.values(grouped).forEach(t => {
+      if (t?.length !== 2) {
```
bruno-garcia, line 133: Prob good to log the value of `t` too. And if a number, the value. 
JS rusty but it could be `null` `undefined`, `1`, `3` or more right?

michellewzhang: so `t` is something like this:

SCR-20240815-mkvy

but i'm not sure any of this information is useful from a debugging perspective. we can also only log primitives as tags, not objects :(

the value of `grouped` is something like this: an object of arrays, the arrays contain items similar to `t`

SCR-20240815-mksp

the arrays will always be at least length 1 because that's what we filter for up above!

bruno-garcia: You could dump t in extra. `Sentry.withScope(s => { 
s.setExtra(t);
Sentry.captureMessage(..);
});

That's 'cheap' since it's not indexed/searchable.
But if you're confident the whole thing isn't helpful, you might want to just log the `type` since it's useful to know if we're getting `null` at least, no? That's what I read when I see the `?` in there.

michellewzhang: ahh i see. added a tag for the type of `t`

static/app/components/replays/videoReplayerWithInteractions.tsx
```diff
@@ -121,6 +124,26 @@ export class VideoReplayerWithInteractions {
...
+        Sentry.captureMessage(
+          'Mobile replay has mismatching touch start and end events',
+          {
+            tags: {
+              sdk_name: context.sdkName,
+              sdk_version: context.sdkVersion,
+              type: typeof t,
```
bruno-garcia, line 140: `type` is quite generic, might conflict with something, how about:

```suggestion
              event_type: typeof t,
```

bruno-garcia (approved):

billyvg (approved):

PR #76293, opened 2024-08-15
feat(discover): adds gauge support for discover metrics avg function

edwardgou-sentry:
Adds a `snql_gauge` to the discover metrics `avg` function to support gauge metrics. Also fixes an issue where MRIs as args weren't being detected as span metrics properly


narsaynorath (approved):
👍

PR #76294, opened 2024-08-15
ref: fix typing in frontend debug views

asottile-sentry:


src/sentry/web/frontend/debug/debug_codeowners_auto_sync_failure_email.py
```diff
@@ -14,7 +15,7 @@ class DebugCodeOwnersAutoSyncFailureView(View):
     def get(self, request: HttpRequest) -> HttpResponse:
         org = Organization(id=1, slug="petal", name="Petal")
         project = Project(id=1, slug="nodejs", name="Node.js", organization=org)
-        user = User(name="Nisanthan")
+        user = Actor.from_orm_user(User(name="Nisanthan"))
```
wedamija, line 18: Was this broken before? I'm confused about how we could start passing an actor and have this still work

asottile-sentry: I _suspect_ the type annotation here is incorrect (the underlying function it calls accepts `User | Team | Actor`) and that it works fine with any of them.  but when I went to fix that annotation it exposed a variance issue due to the way the parameters are passed (it ~essentially has `def f[U: User | Team | Actor](..., user: U, by_user: dict[U, ...] | None = ...):` which I couldn't find a succinct way to make it sound for the other callers.  so yeah this was the "easiest" way to fix it.  the view itself only really cares about `.id` so they end up being equivalent

armenzg (approved):

PR #76300, opened 2024-08-15
remove releases feedback banner

nhsiehgit:
per: https://getsentry.atlassian.net/browse/ALRT-216

This feature is on ice.

We should not be pushing people towards it


cathteng (approved):
bye bye

PR #76301, opened 2024-08-15
ref(issue-details): remove replay id from replay context

michellewzhang:
remove replay ID from the replay context on trace & issue details (removed from both since they use the same component, and we expect the behavior to be the same for both situations).

reasoning for removing
- we have the replay preview immediately below, and replay ID doesn't add much useful context
- sometimes the replay ID link leads to an invalid replay ("replay not found" error) because it was dropped, deleted, etc. and this component doesn't take that into account before rendering the link

will follow up with subsequent removals of replay ID on other parts of issues & trace.

### trace before:

SCR-20240815-myks

### trace after:

SCR-20240815-myju

### issues before:

SCR-20240815-mzem

### issues after:

SCR-20240815-mzbl

### DIFF
--- a/static/app/components/events/contexts/replay/index.tsx
+++ b/static/app/components/events/contexts/replay/index.tsx
@@ -1,6 +1,5 @@
 import ErrorBoundary from 'sentry/components/errorBoundary';
 import KeyValueList from 'sentry/components/events/interfaces/keyValueList';
-import {t} from 'sentry/locale';
 import {type Event, type ReplayContext, ReplayContextKey} from 'sentry/types/event';
 import type {Organization} from 'sentry/types/organization';
 import type {Project} from 'sentry/types/project';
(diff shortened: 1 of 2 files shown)


static/app/components/events/contexts/replay/index.tsx
```diff
@@ -83,18 +80,7 @@ function getReplayKnownDataDetails({
...
-      const link = `/organizations/${organization.slug}/replays/${encodeURIComponent(replayId)}/`;
-      return {
-        subject: t('Replay ID'),
-        value: replayId,
-        action: {link},
-      };
-    }
+    case ReplayContextKey.REPLAY_ID:
     default:
       return undefined;
```
scttcper, line 85: could probably get rid of the whole function since it returns undefined

scttcper (approved):

jas-kas (approved):

leeandher (approved):

PR #76305, opened 2024-08-15
ref(issues/trace): rm replay id from tag distribution

michellewzhang:
remove replay id from the tag distribution chart in trace & issue details. 

see also
- https://github.com/getsentry/sentry/pull/76302
- https://github.com/getsentry/sentry/pull/76301

reasoning for removing
- we have the replay preview immediately below, and replay ID or its distribution doesn't add much useful context
will follow up with subsequent removals of replay ID on other parts of issues & trace.

### before
SCR-20240815-noqc

### after
SCR-20240815-noui

### DIFF
--- a/static/app/components/group/tagFacets/index.tsx
+++ b/static/app/components/group/tagFacets/index.tsx
@@ -174,7 +174,7 @@ export default function TagFacets({
   }, [data, tagFormatter, isStatisticalDetector]);
 
   // filter out replayId since we no longer want to
-  // display this on issue or trace details
+  // display this on issue details
   const topTagKeys = tagKeys.filter(
     tagKey => Object.keys(tagsData).includes(tagKey) && tagKey !== 'replayId'
   );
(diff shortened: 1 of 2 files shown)


static/app/components/group/tagFacets/index.tsx
```diff
@@ -173,9 +173,14 @@ export default function TagFacets({
     return formatted as Record<string, GroupTagResponseItem>;
   }, [data, tagFormatter, isStatisticalDetector]);
 
-  const topTagKeys = tagKeys.filter(tagKey => Object.keys(tagsData).includes(tagKey));
+  // filter out replayId since we no longer want to
+  // display this on issue or trace details
+  const topTagKeys = tagKeys
+    .filter(tagKey => Object.keys(tagsData).includes(tagKey))
+    .filter(f => f !== 'replayId');
```
JoshFerge, line 180: nit: doing two filters is somewhat inefficient as you have to loop through it twice. you should be able to do it in the line above.

michellewzhang: ah good point, kept them separate as i thought it'd be more readable/understandable but i can refactor

JoshFerge (approved):

scttcper (approved):

jas-kas (approved):

PR #76310, opened 2024-08-15
ref(grouping): Add data to `GroupHash` and `GroupHashMetadata` reprs

lobsterkatie:
To make debugging easier, this adds group id and hash to the `GroupHash` repr and group id, hash, and grouphash id to the `GroupHashMetadata` repr.

### DIFF
--- a/src/sentry/models/grouphash.py
+++ b/src/sentry/models/grouphash.py
@@ -12,6 +12,7 @@ from sentry.db.models import (
     Model,
     region_silo_model,
 )
+from sentry.db.models.base import sane_repr
 
 if TYPE_CHECKING:
     from sentry.models.grouphashmetadata import GroupHashMetadata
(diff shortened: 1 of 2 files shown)


src/sentry/models/grouphash.py
```diff
@@ -35,3 +36,5 @@ class Meta:
         app_label = "sentry"
         db_table = "sentry_grouphash"
         unique_together = (("project", "hash"),)
+
+    __repr__ = sane_repr("group_id")
```
JoshFerge, line 40: wouldn't we want the group hash in the repr as well?

JoshFerge (approved):

src/sentry/models/grouphash.py
lobsterkatie: Yeah, I suppose it depends on what kind of debugging you’re doing. In the context of tests (which is how I came to make the change), it's much easier to check the 1- or 2-digit ids against each other than the 32-character hashes. But you're right, if this were in prod (where the ids are also long and hard to compare at a glance), the hashes would probably be the better thing to use. I can add it.

PR #76311, opened 2024-08-15
ref(integrations): add abstract search_issues method for issue integrations

cathteng:
Define an abstract `search_issues` method in `IssueBasicIntegration`. This is used across Jira and SCM integrations -- am currently unsure of the return type, but this unifies the expectation of having issue search functionality across these integrations.

There are also `search_issues` functions in the integration API clients, but we should have the entry point be in `installation.search_issues()`.

### DIFF
--- a/src/sentry/integrations/github/integration.py
+++ b/src/sentry/integrations/github/integration.py
@@ -296,8 +296,10 @@ class GitHubIntegration(RepositoryIntegration, GitHubIssuesSpec, CommitContextIn
 
         return trees
 
-    def search_issues(self, query: str | None, **kwargs):
-        return self.get_client().search_issues(query)
+    def search_issues(self, query: str | None, **kwargs) -> dict[str, Any]:
+        resp = self.get_client().search_issues(query)
+        assert isinstance(resp, dict)
+        return resp
 
 
 class GitHubIntegrationProvider(IntegrationProvider):
--- a/src/sentry/integrations/github/integration.py
+++ b/src/sentry/integrations/github/integration.py
@@ -28,7 +28,6 @@ from sentry.integrations.models.integration import Integration
 from sentry.integrations.models.organization_integration import OrganizationIntegration
 from sentry.integrations.services.repository import RpcRepository, repository_service
 from sentry.integrations.source_code_management.commit_context import CommitContextIntegration
-from sentry.integrations.source_code_management.repository import RepositoryIntegration
 from sentry.integrations.tasks.migrate_repo import migrate_repo
 from sentry.integrations.utils.code_mapping import RepoTree
 from sentry.models.repository import Repository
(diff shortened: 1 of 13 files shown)


Naveen-C-Ramachandrappa (approved):

src/sentry/integrations/mixins/issues.py
```diff
@@ -303,6 +303,10 @@ def get_issue(self, issue_id, **kwargs):
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def search_issues(self, query: str | None, **kwargs):
```
cathteng, line 307: this is the abstract method declaration

src/sentry/integrations/github/integration.py
```diff
@@ -291,8 +291,7 @@ def get_trees_for_org(self, cache_seconds: int = 3600 * 24) -> dict[str, RepoTre
 
         return trees
 
-    # TODO(cathy): define in issue ABC
-    def search_issues(self, query: str) -> Mapping[str, Sequence[Mapping[str, Any]]]:
+    def search_issues(self, query: str | None, **kwargs):
```
GabeVillalobos, line 294: Can we add the return type to this somehow? I'm assuming it's just a list of issues, or a serialized form of it?

src/sentry/integrations/gitlab/integration.py
```diff
@@ -172,8 +172,10 @@ def search_projects(self, query):
         return client.search_projects(group_id, query)
 
     # TODO(cathy): define in issue ABC
-    def search_issues(self, project_id, query, iids):
+    def search_issues(self, query: str | None, **kwargs):
         client = self.get_client()
+        project_id = kwargs["project_id"]
+        iids = kwargs["iids"]
```
GabeVillalobos, line 178: Do we want guards here againt KeyErrors?

GabeVillalobos (approved):
👏🏼 Nice work!

src/sentry/integrations/github/integration.py
cathteng: it's actually just `BaseApiResponseX` because all we do is call a client function 😬 
https://github.com/getsentry/sentry/blob/673e60633fcbedc08ed772aab29edc42e55fe6d7/src/sentry/shared_integrations/client/base.py#L30-L31

src/sentry/integrations/gitlab/integration.py
cathteng: i think we want these to behave like actual parameters so we should surface `KeyErrors` -- these are required in the client function

src/sentry/integrations/github/integration.py
cathteng: i was able to narrow it down to a list of dicts or a dict. i don't think we're typing a lot of these places correctly (these are still in the allowlist in `pyproject.toml`) so i put that return type for now, we can fix it later

src/sentry/integrations/vsts/search.py
```diff
@@ -23,7 +24,7 @@ def get(
         self, request: Request, organization: RpcOrganization, integration_id: int, **kwds: Any
     ) -> Response:
         try:
-            integration = Integration.objects.get(
+            integration: Integration = Integration.objects.get(
```
asottile-sentry, line 27: this shouldn't need an annotation here

PR #76313, opened 2024-08-15
feat(issue-search): support IN for semver release search

### LINKED ISSUE #76286: Selecting multiple releases in the Issues search is failing to load
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

1. search release.version in the Issues page search bar
2. select multiple (checkbox) values for the version

### Expected Result

- load the search results

### Actual Result

**Error**: Invalid operation 'IN' for semantic version filter

Image

### Product Area

Issues

### Link

_No response_

### DSN

_No response_

### Version

_No response_

JoshFerge:
- [x] adds support for the IN Operator for semver releases.
- [ ] need to support the NOT IN operator as well, as well as any negation conditions
- [x] tests added

Fixes https://github.com/getsentry/sentry/issues/76286
Fixes SENTRY-3HAA


src/sentry/search/events/filter.py
```diff
@@ -385,52 +385,48 @@ def _semver_filter_converter(
...
-    final_operator = "IN"
-    if len(versions) == MAX_SEARCH_RELEASES:
-        # We want to limit how many versions we pass through to Snuba. If we've hit
-        # the limit, make an extra query and see whether the inverse has fewer ids.
-        # If so, we can do a NOT IN query with these ids instead. Otherwise, we just
-        # do our best.
-        operator = OPERATOR_NEGATION_MAP[operator]
-        # Note that the `order_by` here is important for index usage. Postgres seems
-        # to seq scan with this query if the `order_by` isn't included, so we
-        # include it even though we don't really care about order for this query
```
wedamija, line 416: Why are we removing all these comments?

src/sentry/search/events/filter.py
```diff
@@ -385,52 +385,48 @@ def _semver_filter_converter(
...
+            versions = [SEMVER_EMPTY_RELEASE]
+        return final_operator, versions
+
+    if isinstance(version, str):
+        final_operator, versions = get_versions(version, operator)
+        return ["release", final_operator, versions]
+    else:
+        semver_filters = []
+        for v in version:
+            _, versions = get_versions(v, operator)
```
wedamija, line 427: I'm not really sure `get_versions` is necessary for `IN`. Since in will just be doing exact matches, all the sorting junk that we do isn't really relevant.

JoshFerge:
will be working on this, came up in triage duty again

src/sentry/search/events/datasets/filter_aliases.py
```diff
@@ -187,46 +187,61 @@ def semver_filter_converter(
...
-        exclude_versions = list(qs_flipped)
-        if exclude_versions and len(exclude_versions) < len(versions):
-            # Do a negative search instead
-            final_operator = Op.NOT_IN
-            versions = exclude_versions
+            exclude_versions = list(qs_flipped)
+            if exclude_versions and len(exclude_versions) < len(versions):
+                # Do a negative search instead
+                final_operator = Op.NOT_IN
+                versions = exclude_versions
```
JoshFerge, line 230: these lines are a codepath which i haven't changed, just the indentation makes it a diff, and codecov reports no coverage on these which causes the codecov to be low

src/sentry/models/releases/util.py
```diff
@@ -80,16 +80,49 @@ def filter_by_semver_build(
                     "release_id", flat=True
                 )
             )
+        if isinstance(build, str):
```
wedamija, line 83: Wondering if it would make sense to just do

```
if isinstance(build, str):
    build = [build]
```

Then go down the IN path

src/sentry/search/events/datasets/filter_aliases.py
```diff
@@ -187,46 +187,61 @@ def semver_filter_converter(
...
+        for v in version:
+            qs = (
+                Release.objects.filter_by_semver(
+                    organization_id,
+                    parse_semver(v, operator),
+                    project_ids=builder.params.project_ids,
+                )
+                .values_list("version", flat=True)
+                .order_by(*order_by)[: constants.MAX_SEARCH_RELEASES]
+            )
```
wedamija, line 243: Since we're doing exact match here I guess we don't care about extra filtering logic in the single version branch?

wedamija (approved):

thisisthekap:
Don't close this PR while remaining unmerged! We are eagerly awaiting this improvement! 😊

src/sentry/search/events/datasets/filter_aliases.py
JoshFerge: yeah, i don't think its super necessary. i'll add a comment and monitor as we release, but i think because we're only doing matches, the extra filtering logic isn't as important.

thisisthekap:
Another comment to keep this open. We are eagerly waiting for this feature! :)

triplef:
We’d also highly appreciate the ability to search for multiple releases. 🙏

PR #69467, opened 2024-04-22
ref(rules): Don't catch snuba query rate limit exceeded errors for now

ceorourke:
We're hitting the abuse limit because this error is happening so frequently so for now don't raise it, but emit a metric we can track while the work to address the issue is underway.

PR #76315, opened 2024-08-16
ref(alerts): Remove try/catch and metric for INC-666

ceorourke:
Follow up to https://github.com/getsentry/sentry/pull/69467 to remove a metric we don't need anymore


saponifi3d (approved):
yay!

PR #76316, opened 2024-08-16
ref(issues): rm replay id from event highlights

michellewzhang:
remove replay id from the issue details event highlights edit modal. it also removes replay id (both tag and context options) from the view if it was previously pinned by any user.

see also
- https://github.com/getsentry/sentry/pull/76302
- https://github.com/getsentry/sentry/pull/76301
- https://github.com/getsentry/sentry/pull/76305
- https://github.com/getsentry/sentry/pull/76307

reasoning for removing
- we have the replay preview in the issue details if it exists, and replay ID doesn't add much useful context
- sometimes the replay ID link leads to an invalid replay ("replay not found" error) because it was dropped, deleted, etc. and this component doesn't take that into account before rendering the link

### before
SCR-20240815-ozuw

### after
SCR-20240815-ozsh


static/app/components/events/highlights/highlightsDataSection.tsx
```diff
@@ -165,7 +170,11 @@ function HighlightsData({
     []
   );
 
-  const highlightTagItems = getHighlightTagData({event, highlightTags});
+  const highlightTagItems = getHighlightTagData({
```
michellewzhang, line 173: this removes it from the UI if the user has previously pinned `replayId`

static/app/components/events/highlights/highlightsDataSection.tsx
```diff
@@ -140,11 +140,16 @@ function HighlightsData({
...
     Object.values(highlightContext).flat().length === 0 && highlightTags.length === 0;
 
+  const {replay, ...rest} = highlightContext;
+
   const highlightContextDataItems = getHighlightContextData({
     event,
     project,
     organization,
-    highlightContext,
+    highlightContext: {
```
michellewzhang, line 149: this removes it from the UI if the user has previously pinned `replay_id`

static/app/components/events/highlights/editHighlightsModal.tsx
```diff
@@ -331,9 +338,14 @@ export default function EditHighlightsModal({
   project,
   closeModal,
 }: EditHighlightsModalProps) {
-  const [highlightContext, setHighlightContext] =
-    useState<HighlightContext>(prevHighlightContext);
-  const [highlightTags, setHighlightTags] = useState<HighlightTags>(prevHighlightTags);
+  const {replay, ...rest} = prevHighlightContext;
+  const [highlightContext, setHighlightContext] = useState<HighlightContext>({
```
michellewzhang, line 342: this removes `replay_id` from the preview UI up top when in the highlight modal if the user has previously pinned it

static/app/components/events/highlights/editHighlightsModal.tsx
```diff
@@ -331,9 +338,14 @@ export default function EditHighlightsModal({
...
+    ...(replay && {replay: replay?.filter(k => k !== 'replay_id')}),
+    ...rest,
+  });
+  const [highlightTags, setHighlightTags] = useState<HighlightTags>(
```
michellewzhang, line 346: this removes `replayId` from the preview UI up top when in the highlight modal, if the user has previously pinned it

static/app/components/events/highlights/editHighlightsModal.tsx
```diff
@@ -249,12 +251,17 @@ function EditContextHighlightSection({
   const ctxItems = Object.entries(ctxData);
   const filteredCtxItems = ctxItems
     .map<[string, string[]]>(([contextType, contextKeys]) => {
+      // filter out replayId because we don't want to expose
+      // this on issue details anymore
       const filteredContextKeys = contextKeys.filter(
-        contextKey => contextKey.includes(ctxFilter) || contextType.includes(ctxFilter)
+        contextKey =>
```
michellewzhang, line 257: this removes `replay_id` as an addable option

static/app/components/events/highlights/editHighlightsModal.tsx
```diff
@@ -155,8 +155,10 @@ function EditTagHighlightSection({
   ...props
 }: EditTagHighlightSectionProps) {
   const [tagFilter, setTagFilter] = useState('');
+  // filter out replayId because we don't want to expose
+  // this on issue details anymore
   const tagData = event.tags
-    .filter(tag => tag.key?.includes(tagFilter))
+    .filter(tag => tag.key?.includes(tagFilter) && tag.key !== 'replayId')
```
michellewzhang, line 161: this removes `replayId` as an addable option

jas-kas (approved):

static/app/components/events/highlights/highlightsDataSection.tsx
leeandher: I this approach works to avoid people clicking dead links to replays but these highlights are set by users which means we'll be going against their wishes for displaying the link at the top of the page. Looking at the DB, there's about 26 projects in SaaS which have set `replayId` or `replay_id` to appear in their highlights.

Is there any harm in retaining these links for highlights? I think it might make more sense to figure out a way to prevent linking to missing replays and instead display some mention that the replay wasn't found rather than omitting it entirely.

michellewzhang: cc @jas-kas

jas-kas: > Is there any harm in retaining these links for highlights? I think it might make more sense to figure out a way to prevent linking to missing replays and instead display some mention that the replay wasn't found rather than omitting it entirely.

I'm okay with retaining these links in Issue Details highlights. Are you suggesting to keep them there _**if**_ the user has pinned `replay_id` or in general for Issue Details highlights (for all users)? I'm good with either case but want to confirm the behaviour.

>  I think it might make more sense to figure out a way to prevent linking to missing replays **and instead display some mention that the replay wasn't found rather than omitting it entirely.**

This would be ideal (in the world where a user has explicitly pinned `replay_id`) -- that's what we do in the Replay section inside Issue Details as you know. :)

JoshFerge (approved):

static/app/components/events/highlights/highlightsDataSection.tsx
michellewzhang: addressed in https://github.com/getsentry/sentry/pull/76619

PR #76323, opened 2024-08-16
feat(query-builder): Add recent search queries when searchbar is empty

malwilley:
Adds a "Recents" tab when the full-width menu is active and some number of recent searches are available.

CleanShot 2024-08-15 at 16 49 35@2x

### DIFF
--- a/static/app/components/searchQueryBuilder/tokenizedQueryGrid.tsx
+++ b/static/app/components/searchQueryBuilder/tokenizedQueryGrid.tsx
@@ -37,10 +37,15 @@ function useApplyFocusOverride(state: ListState<ParseResultToken>) {
   useLayoutEffect(() => {
     if (focusOverride && !focusOverride.part) {
       state.selectionManager.setFocused(true);
-      state.selectionManager.setFocusedKey(focusOverride.itemKey);
+
+      if (focusOverride.itemKey === 'end') {
+        state.selectionManager.setFocusedKey(state.collection.getLastKey());
+      } else {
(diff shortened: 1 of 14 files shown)


static/app/components/searchQueryBuilder/tokens/combobox.tsx
```diff
@@ -56,9 +56,9 @@ type SearchQueryBuilderComboboxProps<T extends SelectOptionOrSectionWithKey<stri
   onCustomValueCommitted: (value: string) => void;
   /**
    * Called when the user selects an option from the dropdown.
-   * Passes the value of the selected item.
+   * Passes the selected option.
    */
-  onOptionSelected: (value: string) => void;
```
malwilley, line 61: Needed to pass the full option object here so that we could do different things when a query option is selected vs a filter key.

static/app/components/searchQueryBuilder/tokenizedQueryGrid.tsx
```diff
@@ -37,10 +37,15 @@ function useApplyFocusOverride(state: ListState<ParseResultToken>) {
   useLayoutEffect(() => {
     if (focusOverride && !focusOverride.part) {
       state.selectionManager.setFocused(true);
-      state.selectionManager.setFocusedKey(focusOverride.itemKey);
+
+      if (focusOverride.itemKey === 'end') {
```
malwilley, line 41: This adds support for putting the focus at the end of the query after a state update, which is what we need to do when selecting a recent search

scttcper (approved):

PR #76336, opened 2024-08-16
feat(explore): Add column editor modal

Zylphrex:
This adds a basic column editor to explore that allows for adding/removing/changing/reordering of columns.


wmak (approved):
lgtm

PR #76337, opened 2024-08-16
chore(similarity): Add project option for all new projects

jangjodi:
Remove EA organization check and option region check before adding project option to enable seer grouping

### DIFF
--- a/src/sentry/api/endpoints/team_projects.py
+++ b/src/sentry/api/endpoints/team_projects.py
@@ -6,7 +6,7 @@ from rest_framework import serializers, status
 from rest_framework.request import Request
 from rest_framework.response import Response
 
-from sentry import audit_log, options
+from sentry import audit_log
 from sentry.api.api_owners import ApiOwner
 from sentry.api.api_publish_status import ApiPublishStatus
 from sentry.api.base import EnvironmentMixin, region_silo_endpoint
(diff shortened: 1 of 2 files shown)


src/sentry/api/endpoints/team_projects.py
```diff
@@ -218,14 +218,8 @@ def post(self, request: Request, team: Team) -> Response:
...
-            # Create project option to turn on ML similarity feature for new EA projects
-            is_seer_eligible_platform = project.platform in SEER_ELIGIBLE_PLATFORMS
-            if (
-                hasattr(project.organization, "flags")
-                and project.organization.flags.early_adopter
-                and is_seer_eligible_platform
-                and options.get("similarity.new_project_seer_grouping.enabled")
-            ):
+            # Create project option to turn on ML similarity feature for new projects
+            if project.platform in SEER_ELIGIBLE_PLATFORMS:
```
mrduncan, line 222: Do we still need the options check here for single-tenants?

jangjodi: good point, thanks! do you happen to know which [regions](https://github.com/getsentry/sentry-options-automator/tree/10f3e82690af383fd113af8e00164e5f87356dc9/options/regions) I'd need to set the option to false for in options automator?

PR #76340, opened 2024-08-16
chore(search): Adopting the new parameterDependentValueType functionality

Abdkhan14:
https://github.com/user-attachments/assets/90f8359a-f26d-4a81-8777-c7624d242de4


static/app/utils/fields/index.ts
```diff
@@ -341,6 +341,14 @@ function validateForNumericAggregate(
   };
 }
 
+function getDynamicFieldValueType() {
```
malwilley, line 344: Does this need to be wrapped in a function?

Abdkhan14: Woops @malwilley, I fixed it

malwilley (approved):

PR #76346, opened 2024-08-16
fix: sanitize org auth token name

mdtro:
Attempt to prevent mail clients from auto-linking organization auth token names that look or contain a domain name by inserting the word _joiner_ unicode character before each period.


jeff-a-holland-codecov (approved):
LGTM

PR #76352, opened 2024-08-16
chore(slack): Removed Expire Url Error Check

iamrajjoshi:
metric to capture the expired url error has existed for two weeks, but we have never hit that error. here, i am cleaning up the error util and some of the exception capturing since we don't need to worry about it.

also decided to try out the `capture_message` from our sdk to notify us via a sentry issue if there is a new slack error that we haven't handled.


tests/sentry/integrations/slack/test_link_identity.py
```diff
@@ -141,27 +141,6 @@ def test_basic_flow_with_web_client_error(self, mock_logger):
         assert mock_logger.exception.call_count == 1
         assert mock_logger.exception.call_args.args[0] == "%serror"
 
-    @patch("sentry.integrations.slack.utils.notifications._logger")
```
iamrajjoshi, line 144: this test isn't needed because this error does not occur

RyanSkonnord (approved):

PR #76353, opened 2024-08-16
feat(profiling) render time series for fn metrics

JonasBa:
Render a line chart for occurence counts of a function below the entry point in the table


static/app/views/profiling/landing/slowestFunctionsTable.tsx
```diff
@@ -22,11 +26,31 @@ import useOrganization from 'sentry/utils/useOrganization';
...
   return b.sum - a.sum;
 }
+function makeProfileLinkFromExample(
+  organization: Organization,
+  f: Profiling.FunctionMetric,
+  example: Profiling.FunctionMetric['examples'][0],
+  projectsLookupTable: Record<string, Project>
+) {
+  if (typeof example === 'string') {
+    return example;
```
Zylphrex, line 41: Is this right? Should if the `example` is a string, then it should be a profile id to link to right?

Zylphrex (commented):
Just the comment about example being a string but otherwise LGTM

static/app/views/profiling/landing/slowestFunctionsTable.tsx
JonasBa: I need to exclude this, it's referencing the type that mashes the old and the new references

Zylphrex (approved):

PR #76355, opened 2024-08-16
chore(integrations): Update integrations codeowners after consolidation

Christinarlong:
After integrations related resources have been moved to the integrations module, we should update codeowners to get rid of old paths & add new ones. Mainly get rid of redundant paths for this PR.


.github/CODEOWNERS
Christinarlong, line 1: All of these deleted lines fall under so just need to delete refs to places that don't exist anymore
```
/src/sentry/integrations/                    @getsentry/product-owners-settings-integrations @getsentry/ecosystem
```

markstory (approved):

PR #76358, opened 2024-08-16
fix(custom-views): Custom views bug fixes

MichaelSun48:
Fixes three error that were ocurring with custom views:

1. **Custom views were not saving for new users**: This was because new users were receiving default tabs, and the backend does not attach an Id to these tabs. The frontend attached the id "default" when it was received, but then it sent the "default" id back to the backend, which the backend tried to query the database with. Solution here is to filter out the id of any tabs that have "default" as the id, similar to temp tabs. 
2. "Received 'false' for non boolean attribute active" DOMWarning: This was caused by a styled component using `active` as a prop, which is already used for something else. Solution here is to change the prop name 
3. Header border is showing up as dashed in unexpected situations: Certain actions (like clicking issues in the side bar when already in issues) will trigger a change in the url query, without any other effect. This would trigger the header border to switch to 'dashed' as if it was a temp view. Ideally we'd like to rely on the url query, but as a stop gap solution we can replace it with extra state.

### DIFF
--- a/static/app/components/draggableTabs/draggableTabList.tsx
+++ b/static/app/components/draggableTabs/draggableTabList.tsx
@@ -136,7 +136,7 @@ function BaseDraggableTabList({
               {t('Add View')}
             </AddViewButton>
           </MotionWrapper>
-          <TabDivider layout active />
+          <TabDivider layout isVisible />
           <MotionWrapper layout>
             {tempTab && (
               <Tab
--- a/static/app/components/draggableTabs/draggableTabList.tsx
+++ b/static/app/components/draggableTabs/draggableTabList.tsx
@@ -71,7 +71,7 @@ function BaseDraggableTabList({
   useEffect(() => {
     setTabListState(state);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [state.selectedItem, state.selectedKey, props.children]);
+  }, [state.selectedKey]);
 
   // Detect tabs that overflow from the wrapper and put them in an overflow menu
   const tabItemsRef = useRef<Record<string | number, HTMLLIElement | null>>({});
--- a/static/app/components/draggableTabs/draggableTabList.tsx
+++ b/static/app/components/draggableTabs/draggableTabList.tsx
@@ -120,7 +120,7 @@ function BaseDraggableTabList({
               </Reorder.Item>
               <TabDivider
                 layout
-                active={
+                isVisible={
                   state.selectedKey === 'temporary-tab' ||
                   (state.selectedKey !== item.key &&
                     state.collection.getKeyAfter(item.key) !== state.selectedKey)
@@ -206,9 +206,9 @@ export function DraggableTabList({items, onAddView, ...props}: DraggableTabListP
 
 DraggableTabList.Item = Item;
 
-const TabDivider = styled(motion.div)<{active: boolean}>`
+const TabDivider = styled(motion.div)<{isVisible: boolean}>`
   ${p =>
-    p.active &&
+    p.isVisible &&
     `
     background-color: ${p.theme.gray200};
     height: 50%;
--- a/static/app/components/draggableTabs/draggableTabList.tsx
+++ b/static/app/components/draggableTabs/draggableTabList.tsx
@@ -206,6 +206,10 @@ export function DraggableTabList({items, onAddView, ...props}: DraggableTabListP
 
 DraggableTabList.Item = Item;
 
+/**
+ * TabDividers are only visible around NON-selected tabs. They are not visible around the selected tab,
+ * but they still create some space and act as a gap between tabs.
+ */
 const TabDivider = styled(motion.div)<{isVisible: boolean}>`
   ${p =>
     p.isVisible &&
--- a/static/app/components/tabs/tab.tsx
+++ b/static/app/components/tabs/tab.tsx
@@ -101,7 +101,7 @@ export const BaseTab = forwardRef(
           ref={ref}
         >
           <VariantStyledInteractionStateLayer hasSelectedBackground={false} />
-          <VariantFocusLayer />
+          <FilledFocusLayer />
           {props.children}
         </FilledTabWrap>
       );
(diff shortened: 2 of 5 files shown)


static/app/views/issueList/customViewsHeader.tsx
```diff
@@ -48,12 +49,12 @@ function CustomViewsIssueListHeader({
     orgSlug: props.organization.slug,
   });
 
+  const [headerBorderStyle, setHeaderBorderStyle] = useState<'dashed' | 'solid'>('solid');
```
malwilley, line 52: It seems like a better way of fixing this would be to update the `useEffect` in this file that reacts to the missing viewId/query/sort. Right now it only runs on first render because of the empty deps array, but it should probably run whenever the URL is updated. 

I think I'd rather not add this state now. It adds to the complexity and might make it harder to find the correct solution later

static/app/views/issueList/customViewsHeader.tsx
```diff
@@ -146,7 +152,7 @@ function CustomViewsIssueListHeaderTabsContent({
             orgSlug: organization.slug,
             groupSearchViews: newTabs.map(tab => ({
               // Do not send over an ID if it's a temporary id
-              ...(tab.id[0] !== '_' ? {id: tab.id} : {}),
+              ...(tab.id[0] !== '_' && !tab.id.startsWith('default') ? {id: tab.id} : {}),
```
malwilley, line 155: nit: would be nice to update the comment here to explain why we don't send default either

static/app/views/issueList/customViewsHeader.tsx
MichaelSun48: Good idea, didn't think of that!

static/app/components/tabs/tab.tsx
```diff
@@ -101,7 +101,7 @@ export const BaseTab = forwardRef(
           ref={ref}
         >
           <VariantStyledInteractionStateLayer hasSelectedBackground={false} />
-          <VariantFocusLayer />
+          <FilledFocusLayer />
```
MichaelSun48, line 104: This is a visual change that was bothering me about the old tab:

Before:
image

After:
image

I personally like this more, but I think it could go either way. Will check with Vu on what he prefers.

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -71,7 +71,7 @@ function BaseDraggableTabList({
   useEffect(() => {
     setTabListState(state);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [state.selectedItem, state.selectedKey, props.children]);
+  }, [state.selectedKey]);
```
MichaelSun48, line 74: This is what was causing infinite rerenders when attempting to use `useContext(TabsContext)`. `state.selectedItem` is the content inside the tab, and since there are three dynamic elements in each tab, this is likely what was triggering the useEffect over and over again. For our use case, we only need to update the state when the selected key has changed.

static/app/views/issueList/customViewsHeader.tsx
```diff
@@ -128,8 +128,7 @@ function CustomViewsIssueListHeaderTabsContent({
     return draggableTabs[0].key;
   };
 
-  // TODO: infer selected tab key state from URL params
-  const [selectedTabKey, setSelectedTabKey] = useState<string>(getInitialTabKey());
```
MichaelSun48, line 126: 🙌

static/app/views/issueList/customViewsHeader.tsx
```diff
@@ -175,39 +176,81 @@ function CustomViewsIssueListHeaderTabsContent({
...
-          )
-        );
-      } else if (!selectedTab) {
+        const isCurrentQuerySortDifferentFromExistingUnsavedChanges =
+          selectedTab.unsavedChanges &&
+          (selectedTab.unsavedChanges[0] !== query ||
+            selectedTab.unsavedChanges[1] !== sort);
+
+        const isCurrentQuerySortDifferentFromSelectedTabQuerySort =
+          query !== selectedTab.query || sort !== selectedTab.querySort;
```
MichaelSun48, line 193: I personally think verbose, XXL variable names are the lesser evil compared to outrageous boolean algebra, but I'd accept if we'd prefer to remove these.

static/app/views/issueList/customViewsHeader.tsx
```diff
@@ -175,39 +176,81 @@ function CustomViewsIssueListHeaderTabsContent({
...
+                    unsavedChanges: [query, sort],
+                  }
+                : tab
+            )
+          );
+        }
+        tabListState?.setSelectedKey(selectedTab.key);
+        return;
+      }
+      if (selectedTab && !query) {
```
MichaelSun48, line 213: This is the case when _only_ a local viewId is passed in. We need to update the query to reflect the view's query and sort, or else it will be  the default query (the prioritized query)

MichaelSun48:
Note to self: tab menu options are insanely borked, work on fixing that next

static/app/views/issueList/groupSearchViewTabs/draggableTabBar.tsx
```diff
@@ -188,17 +192,24 @@ export function DraggableTabBar({
...
+      navigate({
+        query: {
+          ...queryParams,
+          viewId: tempId,
+        },
+        pathname: `/organizations/${orgSlug}/issues/`,
+      });
       setTabs(newTabs);
-      setSelectedTabKey(tempId);
+      tabListState?.setSelectedKey(tempId);
```
MichaelSun48, line 203: I would have expected `tabListState?.setSelectedKey(tempId);` to do the navigation automatically, but it does not for some reason. I suspect this might be something to do with the fact that the new tab has not been registered into tablist state yet. For now, this is a good workaround that achieves the expected behavior.

malwilley (approved):
Good changes! Glad to be rid of the extra selected state!

PR #76364, opened 2024-08-16
feat(onboarding): create `member:invite` scope

ameliahsu:
created `member:invite` scope to allow members without `member:write` access to add new members to an organization without being able to edit them


src/sentry/api/endpoints/organization_member/__init__.py
```diff
@@ -105,8 +105,6 @@ def get_allowed_org_roles(
 
     if is_active_superuser(request):
         return roles.get_all()
-    if not request.access.has_scope("member:admin"):
```
sentaur-athena, line 108: I think this should still return but with `request.access.has_scope(member:admin) or (member:invite)`. no?

src/sentry/api/bases/organizationmember.py
```diff
@@ -10,18 +10,38 @@
...
 class MemberPermission(OrganizationPermission):
     scope_map = {
         "GET": ["member:read", "member:write", "member:admin"],
-        "POST": ["member:write", "member:admin"],
+        "POST": ["member:write", "member:admin", "member:invite"],
         "PUT": ["member:write", "member:admin"],
         "DELETE": ["member:admin"],
     }
 
+    def has_object_permission(
```
sentaur-athena, line 29: What is this doing?

tests/sentry/api/endpoints/test_organization_member_index.py
```diff
@@ -498,7 +498,41 @@ def test_cannot_invite_retired_role_with_flag(self):
...
+        self.organization.flags.disable_member_invite = True
+        self.organization.save()
+
+        owner_user = self.create_user("owner@localhost")
+        self.owner = self.create_member(
+            user=owner_user, organization=self.organization, role="owner"
+        )
+        self.login_as(user=owner_user)
+
+        data = {"email": "eric1@localhost", "role": "owner", "teams": [self.team.slug]}
```
sentaur-athena, line 511: Lol! RIP Eric

src/sentry/conf/server.py
```diff
@@ -1907,6 +1909,7 @@ def custom_parameter_sort(parameter: dict) -> tuple[str, int]:
     (
         ("member:admin", "Read, write, and admin access to organization members."),
         ("member:write", "Read and write access to organization members."),
+        ("member:invite", "Member invite access to organization members."),
         ("member:read", "Read access to organization members."),
```
schew2381, line 1913: nit: Could we order the member scopes in this PR so it goes invite, read, write, admin?

tests/sentry/api/endpoints/test_organization_member_index.py
schew2381, line 511: nit: We can move all these constants into a helper that generates a random email and takes in the role.

If you wanna be fancy I think you could somehow move these all into a loop and use OrganizationMember.get_allowed_org_roles_to_invite() to know the error code and pass it into `self.get_response`

src/sentry/conf/server.py
schew2381: sorry should be other way around: admin, write, read, invite

src/sentry/conf/server.py
```diff
@@ -1878,8 +1879,9 @@ def custom_parameter_sort(parameter: dict) -> tuple[str, int]:
     "org:integrations": {"org:integrations"},
     "org:ci": {"org:ci"},
     "member:read": {"member:read"},
-    "member:write": {"member:read", "member:write"},
-    "member:admin": {"member:read", "member:write", "member:admin"},
+    "member:invite": {"member:read", "member:invite"},
```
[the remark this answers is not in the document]
sentaur-athena: hm, I think this is kind of write and it doesn't make sense to have invite without read. It would mean you can invite someone but you can never know if they accepted the invite, cuz you can't read the members :/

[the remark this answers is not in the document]
schew2381: Ok that makes sense, fine to leave as is

src/sentry/api/bases/organizationmember.py
schew2381: has_object_permission runs inside `convert_args`, and does the endpoint scope checks for us. We need extra logic tho to check for the org bitflag on the POST.

src/sentry/api/endpoints/organization_member/__init__.py
schew2381: We use this function in places where we do want to gate it to `member:admin` only, so unfortunately the only workaround is a new arg

src/sentry/api/endpoints/organization_member/__init__.py
```diff
@@ -94,18 +94,21 @@ def get_allowed_org_roles(
...
+    creating_org_invite: bool = False,
 ) -> Collection[Role]:
     """
     Get the set of org-level roles that the request is allowed to manage.
 
     In order to change another member's role, the returned set must include both
     the starting role and the new role. That is, the set contains the roles that
     the request is allowed to promote someone to and to demote someone from.
+
+    If the request is to invite a new member, the request is allowed by non-admin members.
```
schew2381, line 106:
```suggestion
    If the request is to invite a new member, the member:admin scope is not required.
```

src/sentry/conf/server.py
```diff
@@ -1841,6 +1841,7 @@ def custom_parameter_sort(parameter: dict) -> tuple[str, int]:
     # "org:superuser",  Do not use for any type of superuser permission/access checks
     # Assigned to active SU sessions in src/sentry/auth/access.py to enable UI elements
     "member:read",
+    "member:invite",
```
schew2381, line 1844: nit: keep read, write admin ordering
```suggestion
    "member:invite",
    "member:read",
```

src/sentry/conf/server.py
schew2381, line 1882: nit: put invite above read

src/sentry/conf/server.py
```diff
@@ -1960,6 +1963,7 @@ def custom_parameter_sort(parameter: dict) -> tuple[str, int]:
             "project:read",
             "org:read",
             "member:read",
+            "member:invite",
```
schew2381, line 1966: nit: ordering for all org roles
```suggestion
            "member:invite",
            "member:read",
```

tests/sentry/api/endpoints/test_organization_member_index.py
```diff
@@ -498,49 +498,56 @@ def test_cannot_invite_retired_role_with_flag(self):
...
-        manager_user = self.create_user("manager@localhost")
-        self.manager = self.create_member(
-            user=manager_user, organization=self.organization, role="manager"
-        )
-        self.login_as(user=manager_user)
-
-        data = {"email": "eric1@localhost", "role": "owner", "teams": [self.team.slug]}
-        self.get_error_response(self.organization.slug, **data, status_code=400)
+    def invite_all_helper(self, role):
+        roles = ["owner", "manager", "member"]
```
schew2381, line 502: nit: I think it could be good to specify this variable to not get it confused with the `role` arg, maybe something like `invite_roles`.
Oh you also called them `invite_role` individually in the loop lol

src/sentry/conf/server.py
schew2381: 🐱

schew2381 (approved):
lgtm except some small nits

src/sentry/api/endpoints/organization_member/__init__.py
sentaur-athena: @ameliahsu can you please add a comment here? I think it can be confusing for future readers.

PR #76366, opened 2024-08-16
feat(uptime): Add url_domain and url_suffix columns to the uptime subscription model

wedamija:
This adds `url_domain` and `url_suffix` to the uptime subscription model. The plan is to store the domainof each url here for use in limiting total number of checks per domain

We'll use `url_suffix` to understand how many customers are using a specific host, which can be useful for analytics and potentially for large downtime detection.


src/sentry/uptime/models.py
```diff
@@ -24,6 +24,10 @@ class UptimeSubscription(BaseRemoteSubscription, DefaultFieldsModel):
 
     # The url to check
     url = models.CharField(max_length=255)
+    # The domain of the url, extracted via TLDExtract
+    url_domain = models.CharField(max_length=255, db_index=True, default="")
+    # The suffix of the url, extracted via TLDExtract
+    url_suffix = models.CharField(max_length=255, db_index=True, default="")
```
evanpurkhiser, line 30: What exactly is suffix here? Maybe we should give an example in the comment?

evanpurkhiser (approved):

PR #76369, opened 2024-08-16
fix(ecosystem): Propagate test alert errors to the UI, behind a flag

GabeVillalobos:
Adds error propagation for test alert rules, behind 2 flags:
1. `projects:verbose-test-alert-reporting` - Which opts projects/orgs into verbose error reporting
2. `ecosystem:enable_integration_form_error_raise` - Option check for whether or not to raise errors from `IntegrationFormErrors` when creating tickets.

When both of these flags are enabled, the user will receive explicit error messages in the UI if a test alert rule generates an `IntegrationError`, which typically happens when an API call to the provider fails.

Here are a couple of examples from our Jira integration of failed alert tests, first with an invalid team string:
image

And second, with an invalid Date property from another bug with the Jira form:
image

The `customfield_<id>` error is some additional tech debt due to us only surfacing `IntegrationFormErrors` with Jira field ID instead of the display name they provide us, and will be handled in yet another follow-up PR.

### DIFF
--- a/src/sentry/api/endpoints/project_rule_actions.py
+++ b/src/sentry/api/endpoints/project_rule_actions.py
@@ -91,10 +91,7 @@ class ProjectRuleActionsEndpoint(ProjectEndpoint):
             try:
                 callback(test_event, futures)
             except Exception as exc:
-                if hasattr(callback, "im_class"):
-                    cls = callback.im_class
-                else:
-                    cls = callback.__class__
+                cls = callback.__class__
                 callback_name = getattr(callback, "__name__", str(callback))
                 cls_name = cls.__name__
                 logger = logging.getLogger(f"sentry.test_rule.{cls_name.lower()}")
--- a/src/sentry/api/endpoints/project_rule_actions.py
+++ b/src/sentry/api/endpoints/project_rule_actions.py
@@ -1,14 +1,19 @@
+import logging
+
 from rest_framework.exceptions import ValidationError
 from rest_framework.request import Request
 from rest_framework.response import Response
 
+from sentry import features
 from sentry.api.api_owners import ApiOwner
 from sentry.api.api_publish_status import ApiPublishStatus
 from sentry.api.base import region_silo_endpoint
 from sentry.api.bases import ProjectAlertRulePermission, ProjectEndpoint
 from sentry.api.serializers.rest_framework import RuleActionSerializer
+from sentry.eventstore.models import Event
 from sentry.models.rule import Rule
 from sentry.rules.processing.processor import activate_downstream_actions
+from sentry.shared_integrations.exceptions import IntegrationFormError
 from sentry.utils.safe import safe_execute
 from sentry.utils.samples import create_sample_event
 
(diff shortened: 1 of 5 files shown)


src/sentry/api/endpoints/project_rule_actions.py
```diff
@@ -60,7 +65,66 @@ def post(self, request: Request, project) -> Response:
...
+        This is used in our Alert Rule UI to display errors to the user.
+        """
+        action_exceptions = []
         for callback, futures in activate_downstream_actions(rule, test_event).values():
-            safe_execute(callback, test_event, futures)
+            try:
+                callback(test_event, futures)
+            except Exception as exc:
+                if hasattr(callback, "im_class"):
+                    cls = callback.im_class
```
corps, line 95: is this in practice needed?  I thought this was removed in python 3 I believe.

src/sentry/api/endpoints/project_rule_actions.py
```diff
@@ -60,7 +65,66 @@ def post(self, request: Request, project) -> Response:
...
+                else:
+                    cls = callback.__class__
+                callback_name = getattr(callback, "__name__", str(callback))
+                cls_name = cls.__name__
+                logger = logging.getLogger(f"sentry.test_rule.{cls_name.lower()}")
```
corps, line 100: 😊

src/sentry/api/endpoints/project_rule_actions.py
GabeVillalobos: Oh I'm totally in favor of simplifying this if possible. So is it acceptable to just use `__class__.__name__` now?

corps: Yes, I feel.

markstory: Yeah, you can remove the `im_class` bits that isn't a thing in py3. `callback.__class__` is what you want.

src/sentry/api/endpoints/project_rule_actions.py
```diff
@@ -60,7 +65,66 @@ def post(self, request: Request, project) -> Response:
...
+                # noisy sentry issues, so log with a warning instead.
+                if isinstance(exc, IntegrationFormError):
+                    logger.warning(
+                        "%s.test_alert.integration_form_error",
+                        callback_name,
+                        extra={"exception": exc},
+                    )
+
+                    # IntegrationFormErrors should be safe to propagate via the API
+                    action_exceptions.append(str(exc))
```
markstory, line 112: Do we want to expose `exc.field_errors`? The GroupIntegrationDetails endpoint exposes this property in API responses, and doing the same here could help with consistent error responses from integrations.

GabeVillalobos: That's true, but I'd need to update the UI to handle this whereas this pattern is already supported. I'm also debating whether we should be exposing general `IntegrationError`  instead, but I'm not sure how safe it would be to do this.

src/sentry/api/endpoints/project_rule_actions.py
```diff
@@ -60,7 +65,59 @@ def post(self, request: Request, project) -> Response:
...
+            return self.execute_future_on_test_event(test_event, rule)
+        else:
+            # Old existing behavior to wrap this in a handler which buries exceptions
+            for callback, futures in activate_downstream_actions(rule, test_event).values():
+                safe_execute(callback, test_event, futures)
+            return Response()
+
+    def execute_future_on_test_event(
+        self,
+        test_event: GroupEvent,
```
GabeVillalobos, line 80: This type is completely wrong, (`create_sample_event` return `Event | None`, neither of which are compatible with `activate_downstream_actions`. I'll see if I can update `activate_downstream_actions` to take a `BaseEvent` in the future, since I don't think it makes sense to make this change now.

src/sentry/api/endpoints/project_rule_actions.py
```diff
@@ -60,7 +65,58 @@ def post(self, request: Request, project) -> Response:
...
+            except Exception as exc:
+                callback_name = getattr(callback, "__name__", str(callback))
+                cls_name = callback.__class__.__name__
+                logger = logging.getLogger(f"sentry.test_rule.{cls_name.lower()}")
+
+                # safe_execute logs these as exceptions, which can result in
+                # noisy sentry issues, so log with a warning instead.
+                if isinstance(exc, IntegrationError):
+                    logger.warning(
+                        "%s.test_alert.integration_error", callback_name, extra={"exc": exc}
```
cathteng, line 102: should we also check for `ApiError`? in code i often see us catch `ApiError` and subsequently raise `IntegrationError`, but i'm not sure if that's the case for every location `ApiError` could possible be raised for external API calls

src/sentry/api/endpoints/project_rule_actions.py
```diff
@@ -60,7 +65,58 @@ def post(self, request: Request, project) -> Response:
...
 
+        if features.has(
+            "projects:verbose-test-alert-reporting", project=project, user=request.user
+        ):
+            return self.execute_future_on_test_event(test_event, rule)
+        else:
+            # Old existing behavior to wrap this in a handler which buries exceptions
+            for callback, futures in activate_downstream_actions(rule, test_event).values():
+                safe_execute(callback, test_event, futures)
+            return Response()
```
cathteng, line 76: lol i guess we never used this toaster then 🤔  https://github.com/getsentry/sentry/blob/84e40a838212ae8bef95334c0e93169b27c5693c/static/app/views/alerts/rules/issue/index.tsx#L466

GabeVillalobos: We do display it when we generate validation messages with this code:
```python
serializer = RuleActionSerializer(
            context={"project": project, "organization": project.organization}, data=request.data
        )

        if not serializer.is_valid():
            raise ValidationError(serializer.errors)
```

Misconfigured slack alerts show this toast message because we validate the slack alerting config earlier in the pipeline than our Jira integrations, though there's not really a good way to do this with Jira's API afaict.

src/sentry/api/endpoints/project_rule_actions.py
GabeVillalobos: I'd rather restrict it to `IntegrationError` for now vs showing every API error our clients encounter. I'm already a bit wary of the amount of info we'll be propagating up to the UI with this 😅

cathteng (approved):

PR #76370, opened 2024-08-16
feat(settings): Convert org members list to FC

scttcper:
this component was using AsyncComponent's `renderSearchInput`

combines the organizationMembersWrapper into the organizationMembers component instead of using the router to pass in some loaded data from above.


static/app/views/settings/organizationMembers/inviteRequestRow.tsx
```diff
@@ -75,6 +75,7 @@ function InviteRequestRow({
           onChange={r => onUpdate({role: r.value})}
           value={inviteRequest.role}
           roles={allRoles}
+          aria-label={`${t('Role')}: ${role?.name}`}
```
priscilawebdev, line 78:
```suggestion
          aria-label={role ? t('Role: %s', role.name) : undefined}
```

static/app/views/settings/organizationMembers/organizationMembersList.tsx
```diff
@@ -272,114 +248,143 @@ class OrganizationMembersList extends DeprecatedAsyncView<Props, State> {
...
-
-    // Find out if current user is the only owner
-    const isOnlyOwner = !members.find(
-      ({role, email, pending}) =>
-        role === 'owner' && email !== currentUser.email && !pending
-    );
-
-    // Only admins/owners can remove members
-    const requireLink = !!this.state.authProvider && this.state.authProvider.require_link;
+  const handleQueryChange = (query: string) => {
```
priscilawebdev, line 251: do we want to use `useCallback` for these functions?

malwilley (approved):
Didn't test everything, but searching and filtering members feels so much better now (page doesn't completely disappear) 🎉

PR #76374, opened 2024-08-17
feat(eap): Make EAP requests more rugged (A vs a, for example)

colin-sentry:
In production, the formula name is $A instead of $a (for some reason)

This just ignores the name of the formula and assumes there's only one


priscilawebdev:
@colin-sentry yes, in the new UI, both lowercase and uppercase characters will work. Your changes look good, but I will leave it to the metrics backend experts.

shellmayr (approved):
LGTM - I suppose only enabling a single query is on purpose due to implementation complexity?

colin-sentry:
> LGTM - I suppose only enabling a single query is on purpose due to implementation complexity?

Yes, this is mostly just to see what the UX is like if we hit the EAP cluster

PR #76395, opened 2024-08-19
fix(issue-summary): Check for genai consent

jennmueng:
Make sure AI Issue summary checks for genai consent first before showing it. If not it will not render the cards. We're piggybacking off of the autofix setup check for now.

When genai consent is false:
Screenshot 2024-08-19 at 11 18 18 AM

When true, it will flicker the header when loading, but IMO this is better than showing the loading state and hiding it if they didn't have it enabled:
https://github.com/user-attachments/assets/36f35a32-e646-4d50-a3be-5e41e3e9b2a0

### DIFF
--- a/static/app/components/group/groupSummary.tsx
+++ b/static/app/components/group/groupSummary.tsx
@@ -2,6 +2,7 @@ import styled from '@emotion/styled';
 
 import FeatureBadge from 'sentry/components/badge/featureBadge';
 import {Button} from 'sentry/components/button';
+import {useAutofixSetup} from 'sentry/components/events/autofix/useAutofixSetup';
 import LoadingIndicator from 'sentry/components/loadingIndicator';
 import Panel from 'sentry/components/panels/panel';
 import Placeholder from 'sentry/components/placeholder';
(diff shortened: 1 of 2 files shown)


static/app/components/group/groupSummary.tsx
```diff
@@ -36,13 +37,29 @@ const makeGroupSummaryQueryKey = (
...
       staleTime: Infinity, // Cache the result indefinitely as it's unlikely to change if it's already computed
+      enabled: autofixSetupData?.genAIConsent.ok ?? false,
     }
   );
+
+  return {
+    ...queryData,
+    isLoading:
+      // isFetchedAfterMount is here so we don't have the split second where genAIConsent is true but the query isn't loading yet.
+      isAutofixSetupLoading || queryData.isLoading || !queryData.isFetchedAfterMount,
```
malwilley, line 59: Hmm are you sure that this is necessary? Seems like `isLoading` shouldn't switch to false when it gets enabled

jennmueng: Huh, I just tried it: so `isLoading` is true even when a query is disabled? that seems weird.

malwilley: Yeah `isLoading` just means that the query hasn't loaded. This actually changed in v5 but we haven't upgraded yet

malwilley (approved):

PR #76398, opened 2024-08-19
fix(issues): Avoid issue list cache if projects not updated

### LINKED ISSUE #76387: Project selection doesn't immediately apply to errors
Came up through [zendesk](https://sentry.zendesk.com/agent/tickets/129984):

> The past few months, Sentry seems to have a lot of weird caching issues, where I click on “errors” in a particular project, and then in the list view, its showing errors from other projects, then when I hit Enter again on the search criteria, it shows the right lists. Seems to be more of this happening all around the app and it didn’t used to happen before. Just seems like what used to work flawlessly is a little buggy now...

It also came up on Twitter recently: https://x.com/949630562091917312/status/1824872929622122809

> Please this to the roadmap:
> Fix faulty caching of issue list when switching between projects. 
> It’s annoying as hell. You gotta reload the entire site to get the correct issues for the project switched to.

scttcper:
fixes an issue where we load the wrong issues cache because the PageFiltersStore does not reflect the project in the url until after componentDidMount has run.

I did try getting the pageFiltersStore to update sooner but it broke other pages.

to reproduce the issue:
- Go to projects list
- Click "errors: 123" under a project name and load the issues
- Return to projects list and click another "errors: 456" link
- See that issues list is populated with the wrong issues

fixes https://github.com/getsentry/sentry/issues/76387

### DIFF
--- a/static/app/utils/withSavedSearches.tsx
+++ b/static/app/utils/withSavedSearches.tsx
@@ -39,8 +39,9 @@ function withSavedSearches<P extends InjectedSavedSearchesProps>(
         {...(props as P)}
         savedSearches={props.savedSearches ?? savedSearches}
         savedSearchLoading={
-          !organization.features.includes('issue-stream-custom-views') &&
-          (props.savedSearchLoading ?? isLoading)
+          organization.features.includes('issue-stream-custom-views')
+            ? false
+            : props.savedSearchLoading ?? isLoading
         }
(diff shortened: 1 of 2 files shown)


malwilley:
One big bug I found is that loading `/issues` with no query params doesn't end up loading anything! Maybe instead of skipping the early fetch we can just take the project IDs in the query params instead of the page filters store?

scttcper:
this should be fixed w/ issue stream views

PR #76407, opened 2024-08-19
feat(ui): Backport react-query v5 changes

scttcper:
isLoading -> isPending:
In v5 isLoading is renamed to isPending, this allows us to switch to isPending before upgrading the library

There is only one style of calling useQuery in v5 useQuery({ queryKey, etc })

depends on https://github.com/getsentry/sentry/pull/76404
depends on https://github.com/getsentry/sentry/pull/76431


malwilley (approved):
Great idea! Thanks for doing this, will make the upgrade so much easier

static/app/utils/queryClient.tsx
```diff
@@ -125,11 +143,18 @@ export function useApiQuery<TResponseData, TError = RequestError>(
   const api = useApi({persistInFlight: PERSIST_IN_FLIGHT});
   const queryFn = fetchDataQuery(api);
 
-  const {data, ...rest} = useQuery(queryKey, queryFn, options);
+  const {data, ...rest} = useQuery({
```
scefali, line 146: @scttcper @evanpurkhiser I think this change might be causing problems

PR #76410, opened 2024-08-20
feat(postprocess): INC-855 allow to split post_process_* queues

beezz:
Add possibility to configure additional routing for the `post_process_group` task. During INC-855 we reached the limits of single queue processing and this aims to be a temporary solution that will allow to split single queue into multiple ones.

An example of a `SENTRY_POST_PROCESS_QUEUE_SPLIT_ROUTER`

```
import itertools
queues = ["post_process_transactions-0", "post_process_transactions-1", "post_process_transactions-2"]
queues_gen = itertools.cycle(queues)

SENTRY_POST_PROCESS_QUEUE_SPLIT_ROUTER = {
    "post_process_transactions": lambda: next(queues_gen),
}
```

NB: Celery allows to configure routers for tasks. See ~ https://docs.celeryq.dev/en/stable/userguide/routing.html#routers
To be able to use the Celery routers code must change to not pass the `queue` argument directly to the `task.apply_async` call as that takes priority over the configured routers.

### DIFF
--- a/src/sentry/eventstream/base.py
+++ b/src/sentry/eventstream/base.py
@@ -6,6 +6,8 @@ from datetime import datetime
 from enum import Enum
 from typing import TYPE_CHECKING, Any, Optional, TypedDict, cast
 
+from django.conf import settings
+
 from sentry.issues.issue_occurrence import IssueOccurrence
 from sentry.tasks.post_process import post_process_group
 from sentry.utils.cache import cache_key_for_event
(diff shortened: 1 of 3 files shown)


tests/sentry/eventstream/test_eventstream.py
```diff
@@ -336,6 +339,47 @@ def test_transaction_queue(self, mock_eventstream_insert):
         assert "occurrence_id" not in dict(headers)
         assert body["queue"] == "post_process_transactions"
 
+    @override_settings()
+    @patch("sentry.eventstream.backend.insert", autospec=True)
+    def test_queue_split_router(self, mock_eventstream_insert):
```
mwarkentin, line 344: Do we want a test for default queue assignment?

src/sentry/eventstream/base.py
```diff
@@ -100,11 +102,15 @@ def _dispatch_post_process_group_task(
...
         event_type = self._get_event_type(event)
         if event_type == EventStreamEventType.Transaction:
-            return "post_process_transactions"
+            default_queue = "post_process_transactions"
         elif event_type == EventStreamEventType.Generic:
-            return "post_process_issue_platform"
+            default_queue = "post_process_issue_platform"
         else:
-            return "post_process_errors"
+            default_queue = "post_process_errors"
```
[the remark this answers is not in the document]
beezz: I'm trying to not change the default logic here so I'm staying with `post_process_errors` as the fallback queue.

tests/sentry/eventstream/test_eventstream.py
beezz: Added one for errors.

mwarkentin (approved):
LGTM

PR #76426, opened 2024-08-20
feat(new-trace): Passing the correct timestamp to trace meta endpoint.

Abdkhan14:
There is a mismatch between the number of events reflected by the `events-trace` endpoint and the `events-trace-meta` endpoint. The latter reports less events. This is a bigger issue now since we use the response from the `events-trace-meta` endpoint to determine the `canFetch` property of a txn in the trace waterfall (i.e. to determine if there are embedded span children and if we should render the '+' icon). 

We query for the `traceId`and `min(timestamp)` of the traces associated with a replay. We weren't passing  `timestamp: min(timestamp)` as a param to the events-trace-meta endpoint from replay trace tab.  

This doesn't solve the problem, the mismatch still exists, but this is the right thing to do and lets us narrow down the issue to the Backend.

### DIFF
--- a/static/app/views/performance/newTraceDetails/traceApi/useReplayTraceMeta.tsx
+++ b/static/app/views/performance/newTraceDetails/traceApi/useReplayTraceMeta.tsx
@@ -1,7 +1,7 @@
 import {useMemo} from 'react';
 import type {Location} from 'history';
 
-import {getUtcDateString} from 'sentry/utils/dates';
+import {getTimeStampFromTableDateField, getUtcDateString} from 'sentry/utils/dates';
 import type {TableDataRow} from 'sentry/utils/discover/discoverQuery';
 import EventView from 'sentry/utils/discover/eventView';
 import {useApiQuery} from 'sentry/utils/queryClient';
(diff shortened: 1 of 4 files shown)


narsaynorath (approved):

static/app/views/performance/newTraceDetails/traceApi/useReplayTraceMeta.tsx
```diff
@@ -66,11 +67,22 @@ export function useReplayTraceMeta(
...
+  const replayTraces = useMemo(() => {
+    const traces: ReplayTrace[] = [];
+
+    for (const row of eventsData?.data ?? []) {
+      if (row.trace) {
+        traces.push({
+          traceSlug: String(row.trace),
+          timestamp: getTimeStampFromTableDateField(row['min(timestamp)']),
+        });
+      }
```
edwardgou-sentry, line 79: nit: Maybe we can use `reduce` or some combination of `filter` and `map`?

edwardgou-sentry (approved):

PR #76427, opened 2024-08-20
fix(tooltip): Default tooltip portal to document.body at runtime

evanpurkhiser:
Refs: [JAVASCRIPT-2NAQ](https://sentry.sentry.io/issues/4322983594/)

Sometimes we're seeing `Target container is not a DOM element.`.
Hopefully this change ensures we always have a DOM element.


roggenkemper (approved):

priscilawebdev (approved):
thanks for fixing this

evanpurkhiser:
not 100% positive this will actually fix it, it's more of a guess 😬

PR #76428, opened 2024-08-20
ref(modal): Move TooltipContext up in the GlobalModal

evanpurkhiser:
We can just always render it. No need to only render it when the modal is visible.

Possibly related to https://github.com/getsentry/sentry/pull/76427


roggenkemper (approved):

scttcper (approved):

priscilawebdev (approved):
the container's default value in the context should've worked either way 🤔

PR #76431, opened 2024-08-20
fix(alerts): Switch metric alert title to useEffect

scttcper:
onError and onSuccess functions are gone in react-query v5. switching them to useEffect.


nhsiehgit (approved):
🙌 🙏 🎉

PR #76450, opened 2024-08-21
ref(js): Make httpSamplesPanel compatible with react-router 6

evanpurkhiser:


static/app/views/insights/http/components/httpSamplesPanel.spec.tsx
```diff
@@ -470,10 +470,21 @@ describe('HTTPSamplesPanel', () => {
...
-        'href',
-        '/organizations/org-slug/performance/javascript:11c910c9c10b3ec4ecf8f209b8c6ce48/?domain=%2A.sentry.dev&panel=duration&statsPeriod=10d&transactionMethod=GET'
-      );
+
+      // TODO(__SENTRY_USING_REACT_ROUTER_SIX): This first variant can be removed
+      // once react-router 3 has been removed.
+      try {
+        expect(screen.getByRole('link', {name: 'b1bf1acde131623a'})).toHaveAttribute(
+          'href',
+          '/organizations/org-slug/performance/javascript:11c910c9c10b3ec4ecf8f209b8c6ce48/?domain=%2A.sentry.dev&panel=duration&statsPeriod=10d&transactionMethod=GET'
```
evanpurkhiser, line 479: @gggritso it kind of looks like react router 3 was't actually picking up the `hash` that  get's generated here https://github.com/getsentry/sentry/blob/9edbb4461bf6fb660c39c1702e5e3f93d8275b11/static/app/utils/performance/urls.ts#L33

This works fine in 6 but not in 3 🤔 would you be able to confirm this legitimately works in prod?

gggritso (approved):
This code path is not in use in production, but it's safe to test this way. Once the feature flag it's behind is fully rolled out the tests will naturally fail, which IMO is fine

PR #76456, opened 2024-08-21
fix(alerts): Simplify form rendering conditions

evanpurkhiser:
We can just render these based on the issue types. This way we will never render two at once.

Fixes a case where uptime alerts AND issue alerts edit form were rendering together like this

image

### DIFF
--- a/static/app/views/alerts/edit.tsx
+++ b/static/app/views/alerts/edit.tsx
@@ -32,7 +32,7 @@ type Props = RouteComponentProps<RouteParams, {}> & {
 };
 
 function ProjectAlertsEditor(props: Props) {
-  const {hasMetricAlerts, hasUptimeAlerts, members, organization, project} = props;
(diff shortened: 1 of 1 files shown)


nhsiehgit (commented):
🤔 do we restrict the possible AlertType though?
or will the `METRIC` alert type only ever be available if _they have_ metric alerts?

evanpurkhiser:
I mean I assume if they've made a metric alert rule they can edit it right?

I wouldn't expect the rule to show up in their list if they have the feature turned off right?

nhsiehgit:
What if they downgraded their account?
(i'm not actually sure if this is a valid use case though or what we do to those alerts in these cases lol 😅 )

evanpurkhiser:
> What if they downgraded their account?

I would assume they just don't render in the list then right? Either way what this would do before would be render a blank page

scttcper (approved):

PR #76458, opened 2024-08-21
hackweek(Escalation policies): Escalation event action

RyanSkonnord:


src/sentry/constants.py
```diff
@@ -263,6 +263,7 @@ def get_all_languages() -> list[str]:
     "sentry.rules.actions.notify_event.NotifyEventAction",
     "sentry.rules.actions.notify_event_service.NotifyEventServiceAction",
     "sentry.rules.actions.sentry_apps.notify_event.NotifyEventSentryAppAction",
+    "sentry.rules.actions.escalation.NotifyEscalationAction",
```
RyanSkonnord, line 266: AFAICT, adding this is all we need to get the new action type to show up in the alert config menus. The form class does the rest of the heavy lifting.

PR #76459, opened 2024-08-21
feat(ui): Switch react-query isLoading to isPending

scttcper:
In react-query v5 isLoading is renamed to isPending. isLoading will continue to exist but the functionality is different from before. 

If you want to continue to have the type safety and functionality that isLoading provided before, you should now use isPending. Example below:
```ts
const {data, isPending} = useQuery();

if (isPending) {
  // assume data is undefined
}

// assume data is defined
```

[read their docs to see how isLoading might be used with disabled queries in v5](https://tanstack.com/query/v5/docs/framework/react/guides/disabling-queries#isloading-previously-isinitialloading)

part of getsentry/frontend-tsc#52


aliu39 (approved):

priscilawebdev:
Is it ok to merge these changes before we update the package?

scttcper:
@priscilawebdev isPending was added to useApiQuery here https://github.com/getsentry/sentry/pull/76407/files#diff-4cba118f6d7803ce164c3e15178c3ea7c41f4d11ba561eb396ebda41a7594c05R157

PR #76460, opened 2024-08-21
fix(integrations): Bring back type="submit" in awsLambdaFunctionSelect

evanpurkhiser:
This regressed in GH-76118

### DIFF
--- a/config/tsconfig.base.json
+++ b/config/tsconfig.base.json
@@ -19,7 +19,11 @@
     "moduleResolution": "node",
 
     // We add esnext to lib to pull in types for all newer ECMAScript features
-    "lib": ["esnext", "dom"],
+    "lib": [
+      "esnext",
+      "dom",
+      "dom.iterable"
+    ],
 
     // Skip type checking of all declaration files
     "skipLibCheck": true,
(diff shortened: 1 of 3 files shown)


config/tsconfig.base.json
```diff
@@ -19,7 +19,11 @@
     "moduleResolution": "node",
 
     // We add esnext to lib to pull in types for all newer ECMAScript features
-    "lib": ["esnext", "dom"],
+    "lib": [
+      "esnext",
+      "dom",
+      "dom.iterable"
```
evanpurkhiser, line 25: Added this to allow `.entries` to work on FormData

static/app/views/integrationPipeline/awsLambdaFunctionSelect.spec.tsx
```diff
@@ -16,6 +16,30 @@ describe('AwsLambdaFunctionSelect', () => {
     );
     expect(screen.getByRole('checkbox', {name: 'lambdaB'})).toBeInTheDocument();
     expect(screen.getByRole('button', {name: 'Finish Setup'})).toBeInTheDocument();
-    // TODO: add assertion for form post
+
+    // Uncheck lambda A
+    await userEvent.click(screen.getByRole('checkbox', {name: 'lambdaA'}));
+
+    const {promise, resolve} = Promise.withResolvers<FormData>();
```
scttcper, line 23: would avoid until babel can compile it into something older https://caniuse.com/mdn-javascript_builtins_promise_withresolvers

gggritso (approved):
Unsure if you need to unregister the window event listener at the end of the test, but LGTM otherwise!

evanpurkhiser:
> Unsure if you need to unregister the window event listener at the end of the test, but LGTM otherwise!

Let's do it to be safe!

PR #76465, opened 2024-08-21
chore(ui): Re-enable restricted import lint rules

scttcper:
The rule added in https://github.com/getsentry/sentry/pull/75030 overrode the existing rules that help point people in the right direction especially with RTL imports and lodash

there are two more that are being fixed in #76434

depends on https://github.com/getsentry/sentry/pull/76434


.eslintrc.js
```diff
@@ -35,6 +35,61 @@ module.exports = {
...
+            message:
+              "Use 'useLocation', 'useParams', 'useNavigate', 'useRoutes' from sentry/utils instead.",
+          },
+          {
+            name: 'sentry/utils/withSentryRouter',
+            importNames: ['withSentryRouter'],
+            message:
+              "Use 'useLocation', 'useParams', 'useNavigate', 'useRoutes' from sentry/utils instead.",
+          },
+        ],
```
evanpurkhiser, line 92: We have these over in the eslint repo too right? 

- Should we remove them from there?
- Should we move these there?
- Should we get rid of the eslint repo?

scttcper: i'm in favor of moving these (and more) into the sentry repo since they're super specific to developing in this project. The paths won't be the same in other projects.

evanpurkhiser: Right.

Though in the eslint repo, the package IS specific to sentry

evanpurkhiser (approved):

ryan953:
d'oh! thanks for the fixes

PR #76470, opened 2024-08-21
feat(rr6): Tie react-router 6 to it's feature flag

evanpurkhiser:
This will allow us to slow roll react router 6

### DIFF
--- a/static/app/index.tsx
+++ b/static/app/index.tsx
@@ -67,6 +67,20 @@
 //
 // [1]: https://sentry.io/careers/
 
+// TODO(__SENTRY_USING_REACT_ROUTER_SIX): Very early on check if we're running
+// using the react-router 6 faeture flag so we can enable ths beefore the app
+// boots.
+//
+try {
+  // @ts-expect-error features is an array at this point. It is unfortuantely
+  // typed incorrectly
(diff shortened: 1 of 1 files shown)


static/app/index.tsx
```diff
@@ -67,6 +67,20 @@
 //
 // [1]: https://sentry.io/careers/
 
+// TODO(__SENTRY_USING_REACT_ROUTER_SIX): Very early on check if we're running
+// using the react-router 6 faeture flag so we can enable ths beefore the app
```
priscilawebdev, line 71:
```suggestion
// using the react-router 6 feature flag so we can enable this before the app
```

static/app/index.tsx
```diff
@@ -67,6 +67,20 @@
...
+// boots.
+//
+try {
+  // @ts-expect-error features is an array at this point. It is unfortuantely
```
priscilawebdev, line 75:
```suggestion
  // @ts-expect-error features is an array at this point. It is unfortunately
```

priscilawebdev (approved):

static/app/index.tsx
evanpurkhiser: will cleanup later 👍

PR #76476, opened 2024-08-21
fix(email): don't escape strings in plaintext emails

### LINKED ISSUE #76475: Plaintext emails rendered from templates
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

1. Send any email using a text (not HTML) template, where one of the context variables being rendered contains special characters (&,<,>,',etc)

### Expected Result

1. Email renders as straightforward plaintext without any form of escaping

### Actual Result

Email is rendered with inappropriate escaping that's not needed

### Product Area

Other

### Link

_No response_

### DSN

_No response_

### Version

_No response_

keeakita:
Fixes #76475

### DIFF
--- a/src/sentry/templates/sentry/emails/access-approved.txt
+++ b/src/sentry/templates/sentry/emails/access-approved.txt
@@ -1,1 +1,3 @@
+{% autoescape off %}
 You have been granted access to the {{ organization.slug }} / #{{ team.slug }} team on Sentry.
+{% endautoescape %}
--- a/src/sentry/templates/sentry/emails/activity/release.txt
+++ b/src/sentry/templates/sentry/emails/activity/release.txt
@@ -1,5 +1,7 @@
+{% autoescape off %}
 Version {{ version_parsed }} was deployed to {{ environment }} on {{ deploy.date_finished }}
 
 {% for project, release_link, resolved_issue_count in projects %}
     {{ release_link }}
 {% endfor %}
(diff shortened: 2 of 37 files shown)


keeakita:
Ugh, yeah, this was what I was afraid of. This breaks getsentry because of how django.contrib.humanize is loaded.

I'm just going to go through all the plaintext email templates and turn off autoescape.

keeakita:
Ah, so there's a core issue with our email acceptance tests!

Here's what's up:
- My changes turn autoescaping off for rendering plaintext emails. This makes sense because plaintext emails are, well, plaintext! We shouldn't escape anything.
- However, we have some demo endpoints that render these emails inside an HTML page for debugging and acceptance tests. Handy! Except since these are HTML pages, I need to escape all the stuff I just unescaped! So I fixed that, but
- Our acceptance tests are coded to check for the unescaped versions! Which makes sense, because we want to make sure the emails are rending without escaping. But it loads it from an HTML page, which *should* be escaped!

I think the real answer here is that instead of rendering an HTML page, we should render a text/plain response and use that. This will change our chromedriver code though. I will take a stab at this today.

natemoo-re:
I'm going through the PR backlog to cut down on noise. PRs and drafts opened more than 6 months ago without any recent activity are being closed.

Seems like this is important, so please feel free to reopen if this is still something you're hoping to land.

PR #39634, opened 2022-10-04
ref(celery): Remove partitioned queues

udameli:
Partitioned queues no more. Queues for `counters` and `triggers` weren't really partitioned because `create_partitioned_queues` was looping over `range(1)` so there were always only `counters-0` and `triggers-0` (confirmed in [metrics](https://app.datadoghq.com/dashboard/j6i-rfz-67y/queue-status?fullscreen_end_ts=1664992112394&fullscreen_paused=false&fullscreen_section=overview&fullscreen_start_ts=1664977712394&fullscreen_widget=87139323&from_ts=1664975117626&to_ts=1664989517626&live=true)). This PR removes the partitioning logic.

PR #76494, opened 2024-08-22
Reintroduce split queues to reduce load on a single Celery queue

fpacifici:
https://github.com/getsentry/sentry/pull/76410 introduces a way to deliver 
messages to a set of split queues rather than relying on a single queue for
post process.

This was used to address an incident as we were saturating Rabbit resources
for a single queue (which is single threaded). Splitting messages across multiple
queues solves that problem.

This PR reintroduce the split queue support for  tasks where we set the queue in the task definition.

It introduces a router  `SplitQueueTaskRouter`. This is a router used directly by Celery and it maps a task name to a queue. The routing policy is the same as the one employed by `SplitQueueRouter`.  A list of split queues is defined in settings `CELERY_SPLIT_QUEUE_TASK_ROUTES`. 

Rollout procedure.

Phase 1: Verify `SplitQueueTaskRouter` does not break anything.
1. Set up the router in S4S via the prod configuration
2. If everything is fine turn it on everywhere. There will be no configuration so this should be a noop.

Phase 2 (s4s first): Set up split queues for save_event_transactions in prod.
1. In one PR: remove the queue name from the task and add the task to `CELERY_SPLIT_QUEUE_TASK_ROUTES` with an empty config. This cannot be feature flagged because the removal of the queue from the task is a code change.  The router would be mostly a noop as the rollout of the split queue is done via option
2. Create the queues for the task in `CELERY_SPLIT_QUEUE_TASK_ROUTES` in the affected environments.
3. Change the config of the workers in prod to listen to all queues. 
4. Rollout the task via the `celery_split_queue_task_rollout` option. Only at this point the producer starts routing messages to the split queues

Rollout procedure for a new split queue:
1. Sentry PR: Declare the split queues, add the queues in the setting together with the default queue (non split), remove the queue name from the task definition. This is a noop change.
6. Change to the production configuration to make the workers consume the split queues as well.
7. option automator PR: rollout the router on the split queue.


src/sentry/options/defaults.py
```diff
@@ -2718,3 +2718,18 @@
...
+register(
+    "celery_split_queue_rollout",
+    default={"post_process_transactions": 1.0},
+    flags=FLAG_AUTOMATOR_MODIFIABLE,
+)
+register(
+    "celery_split_queue_task_rollout",
+    default={},
+    flags=FLAG_AUTOMATOR_MODIFIABLE,
+)
```
fpacifici, line 2735: This per queue rollout option allows us to do a backward compatible rollout of a new queue without having to issue and deploy 2 sentry PRs.
We could have relied entirely on the settings to enable the split queue rollout, that would have required this procedure:
1. PR1: (sentry)declare the queues in sentry settings
2. PR2: change the arguments of the workers to listen to the new queues 
3. PR3: (sentry) change sentry settings adding the split config configuration.

This way we can replace PR3 with a simple options autoamtor change, which is much quicker and can be rolled out per region.

src/sentry/queue/routers.py
```diff
@@ -0,0 +1,103 @@
...
+
+from sentry import options
+from sentry.celery import app
+
+
+def _get_known_queues() -> set[str]:
+    return {c_queue.name for c_queue in app.conf.CELERY_QUEUES}
+
+
+def _validate_destiantions(destinations: Sequence[str]) -> None:
```
jjbayer, line 16:
```suggestion
def _validate_destinations(destinations: Sequence[str]) -> None:
```

src/sentry/options/defaults.py
```diff
@@ -2718,3 +2718,18 @@
...
     flags=FLAG_AUTOMATOR_MODIFIABLE,
 )
+register(
+    "celery_split_queue_legacy_mode",
+    default=["post_process_transactions"],
+    flags=FLAG_AUTOMATOR_MODIFIABLE,
+)
+register(
+    "celery_split_queue_rollout",
+    default={"post_process_transactions": 1.0},
```
jjbayer, line 2728: Tasks will still be routed to the default queue after this PR has been deployed, because `celery_split_queue_task_rollout` is empty, right?

jjbayer (dismissed):

src/sentry/conf/server.py
```diff
@@ -889,6 +889,9 @@ def SOCIAL_AUTH_DEFAULT_USERNAME() -> str:
     Queue("events.save_event", routing_key="events.save_event"),
     Queue("events.save_event_highcpu", routing_key="events.save_event_highcpu"),
     Queue("events.save_event_transaction", routing_key="events.save_event_transaction"),
+    Queue("events.save_event_transaction_1", routing_key="events.save_event_transaction"),
+    Queue("events.save_event_transaction_2", routing_key="events.save_event_transaction"),
+    Queue("events.save_event_transaction_3", routing_key="events.save_event_transaction"),
```
markstory, line 894: Should the queues have different routing keys? My understanding of AMQP queues is that queues are bound to an exchange with a routing key. Any [message published on an exchange with a matching routing key is delivered to all queues with the same routing key](https://www.rabbitmq.com/tutorials/tutorial-four-python#multiple-bindings). If all of these queues have the same routing key, each of them would receive a copy of messages published with `events.save_event_transaction`

src/sentry/queue/routers.py
```diff
@@ -0,0 +1,103 @@
...
+    WARNING: Do not forget to configure your workers to listen to the
+    queues appropriately before you start routing messages.
+    """
+
+    def __init__(self) -> None:
+        known_queues = _get_known_queues()
+        self.__queue_routers = {}
+        for source, destinations in settings.CELERY_SPLIT_QUEUE_ROUTES.items():
+            assert source in known_queues, f"Queue {source} in split queue config is not declared."
+            _validate_destiantions(destinations)
```
markstory, line 40:
```suggestion
            _validate_destinations(destinations)
```
To go with Joris' suggestion.

src/sentry/queue/routers.py
```diff
@@ -0,0 +1,103 @@
...
+            self.__queue_routers[source] = cycle(destinations)
+
+    def route_for_queue(self, queue: str) -> str:
+        rollout_rate = options.get("celery_split_queue_rollout").get(queue, 0.0)
+        if random.random() >= rollout_rate:
```
markstory, line 45: Generally we do `random.random() < option_val`  So that a 0.1 config option value only moves 10% of traffic instead of 90%.

src/sentry/queue/routers.py
```diff
@@ -0,0 +1,103 @@
...
+    def __init__(self) -> None:
+        known_queues = _get_known_queues()
+
+        self.__task_routers = {}
+        for source, destinations in settings.CELERY_SPLIT_QUEUE_TASK_ROUTES.items():
+            default_destination = destinations[1]
+            assert (
+                default_destination in known_queues
+            ), f"Queue {default_destination} in split queue config is not declared."
+            _validate_destiantions(destinations[0])
```
markstory, line 90:
```suggestion
            _validate_destinations(destinations[0])
```

src/sentry/queue/routers.py
```diff
@@ -0,0 +1,103 @@
...
+
+            self.__task_routers[source] = TaskRoute(default_destination, cycle(destinations[0]))
+
+    def route_for_task(self, task: str, *args: Any, **kwargs: Any) -> Mapping[str, str] | None:
+        route = self.__task_routers.get(task)
+        if route is None:
+            return None
+
+        rollout_rate = options.get("celery_split_queue_task_rollout").get(task, 0.0)
+        if random.random() >= rollout_rate:
```
markstory, line 100:
```suggestion
        if random.random() < rollout_rate:
```

To go with my other comment about rollout rates.

src/sentry/queue/routers.py
fpacifici: I thought I had installed the spell checker in VSCode.

src/sentry/queue/routers.py
jjbayer: IIUC the condition is inverted because it's an early return. I.e. for a 0.1 rollout rate, 90% fall back to the old behavior, and 10% continue to the new behavior below.

src/sentry/options/defaults.py
fpacifici: So you would think, and you will be right. 
sadly (or fortunately as that was in response to an incident) there is an override in the ops repo that added that setting, and it is set in both US and DE.
https://github.com/getsentry/ops/blob/master/k8s/services/getsentry/templates/k8s_regionsilo_ops.py.j2#L20-L26

Tasks on post process forwarder have already been routed to the split queues for two weeks.

src/sentry/conf/server.py
fpacifici: That's what I thought initially, though we already have the the post process transactions queue configured this way in prod https://github.com/getsentry/ops/blob/master/k8s/services/getsentry/templates/k8s_regionsilo_ops.py.j2#L13-L17
with the same routing_key 
```
    Queue("post_process_transactions_1", routing_key="post_process_transactions"),
    Queue("post_process_transactions_2", routing_key="post_process_transactions"),
    Queue("post_process_transactions_3", routing_key="post_process_transactions"),
```

They are definitively not getting duplicated messages as they all get 1/3 of the original load. 
I try to run this locally and I observed the load being routed in round robin. 
I think it may be for two reasons:
1. in post process queue we do not specify the queue in the task but in the `apply_async` call. 
2. For the save event, this may be working because of the router implementation `SplitQueueRouter`. 

@wedamija  I see the previous split queue implementation did not even specify the routing key on the split queues. https://github.com/getsentry/sentry/pull/39634/files#diff-866fbc2da71a688b0a929236b2c864dfa8935d4a6423321d2d97a28337c6e6e8L693 Do you know if there was a reason for that?

src/sentry/queue/routers.py
fpacifici: What @jjbayer  said. Anyway if this is confusing to read please let me know and I will refactor.

src/sentry/queue/routers.py
fpacifici: As above this is the condition to return early with the "not rolled out" behavior so that 0.1 will return early 90% of the times.

src/sentry/queue/routers.py
```diff
@@ -0,0 +1,103 @@
...
+                default_destination in known_queues
+            ), f"Queue {default_destination} in split queue config is not declared."
+            _validate_destiantions(destinations[0])
+
+            self.__task_routers[source] = TaskRoute(default_destination, cycle(destinations[0]))
+
+    def route_for_task(self, task: str, *args: Any, **kwargs: Any) -> Mapping[str, str] | None:
+        route = self.__task_routers.get(task)
+        if route is None:
+            return None
```
fpacifici, line 97: @wedamija  I think this behavior is equivalent to not having a router. Does this seem right to you regarding the old split queue? As the existence of the router cannot be easily rolled out gradually (it is a setting)it would be good to be certain about this behavior. 
If it is unclear I can roll this entire behavior in s4s first by changing the config only there.

src/sentry/conf/server.py
wedamija: @fpacifici All I really know about this is that we removed the split queues because they were totally unused.

I think that the reason this works is because there's a custom router here: https://github.com/getsentry/ops/blob/c9ff1d47d20a07a004e31907e10b71146fa28e97/k8s/services/getsentry/templates/k8s_regionsilo_ops.py.j2#L20-L26.

That's used here: https://github.com/getsentry/sentry/blob/8c7b54ca2bfa9383ba5f969bfe0dd86485ff67c7/src/sentry/eventstream/base.py#L111-L113

I think this router must override the routing key - although that kind of makes me wonder if the routing key is necessary at all here. 

I'm also not sure why we need to add these 3 queues to the default config - shouldn't this be configured specifically in production configs?

src/sentry/queue/routers.py
wedamija: I don't know this behaviour well, but from some reading online it seems that if `route_task` returns None then it will fall back to the default routing behaviour

src/sentry/conf/server.py
fpacifici: @wedamija  what about doing like the previous implementation where we provide a parameter and the queues are automatically generated ?

fpacifici: I figured out why it worked (thanks @beezz ).
```
root@rabbitmq-default-0.c.internal-sentry.internal:~# rabbitmqctl list_bindings --vhost /getsentry | grep post_process_tra
        exchange        post_process_transactions_1     queue   post_process_transactions_1     []
        exchange        post_process_transactions       queue   post_process_transactions       []
        exchange        post_process_transactions_2     queue   post_process_transactions_2     []
        exchange        post_process_transactions_3     queue   post_process_transactions_3     []
```

This queue uses the default exchange, not the direct exchange. So the queue is bound to a binding_key with the same name as the queue when it is created.  The direct exchange would indeed fan the message out to all queues.

In fact I tried to produce a message with `binding_key=post_process_transactions` rather than by specifying the queue and the message went nowhere. That, to my understanding, is because the excahnge we use does not have a binding_key = `post_process_transactions`. 

Anyway I will change the binding key to avoid disasters should we change the exchange.

src/sentry/queue/routers.py
fpacifici: Ok, no pb. I will test the change in s4s first.

src/sentry/conf/server.py
fpacifici: > I'm also not sure why we need to add these 3 queues to the default config - shouldn't this be configured specifically in production configs?

I changed the approach. We do not declare all the queues one by one anymore. We have an additional config setting that generates the queues we need. This will be set up in prod only.

fpacifici:
Some changes:
- I documented the whole rollout process in the PR description
- Stopped hardcoding all the split queues individually. There are now two settings `CELERY_SPLIT_QUEUE_ROUTES` and `CELERY_SPLIT_QUEUE_TASK_ROUTES` that instruct the system to generate the queues it needs and allow us to increase or decrease the number of queues in a backward compatible way.
- I set everything up to test all changes in s4s first.

fpacifici:
This was becoming too large. The standard split queue became totally independent from the task split queue. So I split them.
This PR only contains the task split queue.
The simple split queue (the cleanup for the incident) is https://github.com/getsentry/sentry/pull/78274

src/sentry/utils/celery.py
```diff
@@ -1,9 +1,63 @@
...
+        name = f"{base_name}_{index + 1}"
+        ret.append(name)
+    return ret
+
+
+def _build_queues(base: str, quantity: int) -> Sequence[Queue]:
+    return [Queue(name=name, routing_key=name) for name in build_queue_names(base, quantity)]
+
+
+def make_split_task_queues(config: Mapping[str, SplitQueueTaskRoute]) -> Sequence[Queue]:
```
markstory, line 28: Will we be using these methods to generate the queue list during application startup/config?

src/sentry/queue/routers.py
markstory: From reading through celery's [Router](https://github.com/celery/celery/blob/118e004856e231f3c40f9badd75155d0dc0de38e/celery/app/routes.py#L66-L111) having a queue/route of none seems like it will be ok. I wasn't able to track down how the queue decision makes it way to amqp publishing but I would assume that the queue will not be replaced when routers don't make a decision.

src/sentry/conf/server.py
```diff
@@ -821,6 +822,18 @@ def SOCIAL_AUTH_DEFAULT_USERNAME() -> str:
     "sentry.integrations.tasks",
 )
 
+# tmp(michal): Default configuration for post_process* queueus split
```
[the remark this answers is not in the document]
beezz: ... ueueueue .... 🎶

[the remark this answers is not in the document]
fpacifici: so is it supposed to be `qeueueus` right ?

src/sentry/utils/celery.py
fpacifici: We would use them to generate them in the settings file. 
The alternatives were:
- explicitly define each queues. I am pretty sure we would make mistakes there when duplicating them.
- During the Celery app initialization. That may work but I did not try that. I did not go that way because I am not confident that nobody uses CELERY_QUEUES before the Celery app is initialized. I think we should move this logic there after testing thoroughly it is ok.

lynnagara (approved):
lgtm

PR #76495, opened 2024-08-22
ref(js): Move RouteAnalyticsContext to App

evanpurkhiser:
Instead of placing this inside the renderRouter we can connect it to the
router hooks to provide the values. This allows it to work with react
router 3 and 6.


ArthurKnaus (approved):
🙌

evanpurkhiser:
Verified this still is sending analytic events

PR #76506, opened 2024-08-23
ref: squash index_together operation into feedback initial migration

asottile-sentry:


cmanallen (approved):
> we will need a hard stop for self-hosted

What does this mean?

asottile-sentry:
> > we will need a hard stop for self-hosted
> 
> What does this mean?

https://develop.sentry.dev/self-hosted/releases/#hard-stops

markstory (approved):

PR #76507, opened 2024-08-23
ref: squash index_together operation into replays initial migration

asottile-sentry:


markstory (approved):
Seems reasonable to me.

PR #76524, opened 2024-08-25
feat(new-trace): Shouldn't display missing inst. span for react-nativee traces.

Abdkhan14:
There is nothing really missing it’s just empty space between HTTP requests, similar reason as on the web.


Dhrumil-Sentry (approved):

krystofwoldrich (approved):
Make sense. Thank you.

PR #76527, opened 2024-08-26
ref(stats): Improve x-axis labels visibility

### LINKED ISSUE #74338: X-axis annotations (Jul 10, Jul 11) on the stats page are not aligned with the actual timestamps shown

priscilawebdev:
**Problem**

On the stats page, the x-axis labels and ticks are designed to prevent the chart from looking overcrowded. See the function https://github.com/getsentry/sentry/blob/28ba9c33fc6e0aa12d45ca80451931f5f00e1503/static/app/views/organizationStats/usageChart/utils.tsx#L90.

The issue with the current logic is that labels and ticks are set at fixed intervals without ensuring they align correctly with previous values. This can lead to confusion, where users might see a label like "July 10" and assume the interval starts there, while the previous interval may also include "July 10".

**Solution**

This PR introduces a new function, `getXAxisLabelVisibility`, which determines the visibility of x-axis labels based on specific conditions:

- Labels are shown only if the defined date is more than 1 day apart.
- If there are 10 or fewer unique labels, all of them are displayed.
- If there are more than 10 unique labels, the total number of labels is divided by 10 to determine the intervals for displaying the labels, but this time starting from the correct interval 

**Note**
Displaying max and min labels does not look bad. For instance, the chart does not make use of the zoom functionality too. Should we re-enable the visibility of them?

**Previews**

before:
image

after:
image

closes https://github.com/getsentry/sentry/issues/74338


leedongwei (approved):
Hah, sorry that you had to deal with my old code. 

Changes look reasonable to me. Can you put screenshots of before and after in the PR description?

static/app/views/organizationStats/usageChart/utils.tsx
```diff
@@ -83,58 +83,52 @@ export function getTooltipFormatter(dataCategory: DataCategoryInfo['plural']) {
...
+  const uniqueLabels: string[] = [];
+  const labelToPositionMap: Record<number, boolean> = {};
+  const labelVisibility: boolean[] = new Array(intervals.length).fill(false);
+
+  // Collect unique labels and their positions
+  intervals.forEach((label, index) => {
+    if (index === 0 || label.slice(0, 6) !== intervals[index - 1].slice(0, 6)) {
+      uniqueLabels.push(label);
+      labelToPositionMap[label] = index;
+    }
```
ArthurKnaus, line 108: nit, imho this would be easier to understand if `uniqueLabels` was a set:
```ts
const slicedLabel = label.slice(0, 6)
if (uniqueLabels.has(slicedLabel)) {
  return
}
uniqueLabels.add(slicedLabel)
labelToPositionMap[slicedLabel] = index
```

ArthurKnaus: Alternatively, you could also only store 	`labelToPositionMap` and iterate over the keys later on.

ArthurKnaus (approved):
lgtm 👍

static/app/views/organizationStats/usageChart/utils.tsx
```diff
@@ -83,58 +83,52 @@ export function getTooltipFormatter(dataCategory: DataCategoryInfo['plural']) {
...
-  for (let i = 0; i < daysBetweenLabels.length && numLabels > MAX_NUMBER_OF_LABELS; i++) {
-    numLabels = numTicks / daysBetweenLabels[i];
-    numTicks = days / daysBetweenTicks[i];
+  const uniqueLabels: string[] = [];
+  const labelToPositionMap: Record<number, boolean> = {};
+  const labelVisibility: boolean[] = new Array(intervals.length).fill(false);
+
+  // Collect unique labels and their positions
+  intervals.forEach((label, index) => {
+    if (index === 0 || label.slice(0, 6) !== intervals[index - 1].slice(0, 6)) {
```
scttcper, line 105: slice returns a new array and will never return true
`label.slice(0, 6) !== intervals[index - 1].slice(0, 6)`

scttcper:
14 days looks quite a bit better, i'd check the 1 hour and 1 day graphs they don't seem to have labels

image

but they also don't seem to in prod

priscilawebdev:
@scttcper yes we are not displaying labels for <= 24h

static/app/views/organizationStats/usageChart/utils.tsx
priscilawebdev: I understand your concern, but in this case we are comparing 2x strings. When comparing strings, JavaScript compares their content directly, not their references. If two strings have the same content, they are considered equal. Does that make sense?

priscilawebdev: the result of the code below for example is `false`

```
const label = 'Aug 6'
const secondLabel = 'Aug 6'

console.log(label.slice(0, 6) !== secondLabel.slice(0, 6))
```

scttcper: Oh it's a string lol

scttcper (approved):

PR #76532, opened 2024-08-26
feat(nestjs): Update nestjs onboarding with new error monitoring setup

nicohrubec:
Users now need to explicitly either add the `SentryGlobalFilter` or decorate existing global filters with `WithSentry()` if they have any.

Note: Should be merged after javascript SDK version `8.27` was released.

### DIFF
--- a/static/app/gettingStartedDocs/node/nestjs.tsx
+++ b/static/app/gettingStartedDocs/node/nestjs.tsx
@@ -64,6 +64,7 @@ getError() {
 
 const getDecoratedGlobalFilter = () => `import { Catch, ExceptionFilter } from '@nestjs/common';
 import { WithSentry } from '@sentry/nestjs';
+
 @Catch()
 export class YourCatchAllExceptionFilter implements ExceptionFilter {
   @WithSentry()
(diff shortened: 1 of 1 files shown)


static/app/gettingStartedDocs/node/nestjs.tsx
```diff
@@ -136,6 +165,44 @@ const onboarding: OnboardingConfig = {
             },
           ],
         },
+        {
+          description: tct(
+            'In case you are using a global catch-all exception filter (which is either a filter registered with [code1:app.useGlobalFilters()] or a filter registered in your app module providers annotated with a [code2:@Catch()] decorator without arguments), add a [code3:@WithSentry()] decorator to the [code4:catch()] method of this global error filter. This decorator will report all unexpected errors that are received by your global error filter to Sentry:',
```
priscilawebdev, line 170:
```suggestion
            'In case you are using a global catch-all exception filter (which is either a filter registered with [code1:app.useGlobalFilters()] or a filter registered in your app module providers annotated with a [code2:@Catch()] decorator without arguments, add a [code3:@WithSentry()] decorator to the [code4:catch()] method of this global error filter. This decorator will report all unexpected errors that are received by your global error filter to Sentry:',
```

priscilawebdev (approved):
lgtm

PR #76539, opened 2024-08-26
fix(onboarding): Fix integration issues

### LINKED ISSUE #76160: Fix actions rendered in the onboarding's instructions for AWS Lambda (Python)
https://github.com/user-attachments/assets/4f00852b-8388-4e9b-bc5f-6cb21644df90

priscilawebdev:
**Before**

https://github.com/user-attachments/assets/3f7a9da8-e61e-45ac-a7ea-af9e49dd406c

https://github.com/user-attachments/assets/6434f0c4-a24c-46b0-ae60-6765b7ae66a8

**After**

https://github.com/user-attachments/assets/7b77676e-3d84-4fed-9be0-75873b30f737

https://github.com/user-attachments/assets/9ecf6446-ec6f-45fb-88b1-6cde5ef4bd21

closes https://github.com/getsentry/sentry/issues/76160


static/app/views/projectInstall/platformOrIntegration.tsx
```diff
@@ -1,28 +1,72 @@
...
+const allPlatforms: PlatformIntegration[] = [
+  ...platforms,
+  {
+    id: 'other',
+    name: t('Other'),
+    link: 'https://docs.sentry.io/platforms/',
+    type: 'language',
+    language: 'other',
+  },
+];
```
ArthurKnaus, line 25: `platforms` already includes the `other` platform.

static/app/views/projectInstall/platformOrIntegration.tsx
```diff
@@ -1,28 +1,72 @@
...
+
+type Props = RouteComponentProps<{projectId: string}, {}>;
+
+function PlatformOrIntegration({params}: Props) {
+  const organization = useOrganization();
+  const [integrationUseManualSetup, setIntegrationUseManualSetup] = useState(false);
```
ArthurKnaus, line 31: Shouldn't we store this state in the URL, like before?

ArthurKnaus (approved):
🙌

static/app/views/projectInstall/platformOrIntegration.tsx
priscilawebdev: I decided to go with component state, just as we do when users install our JavaScript platform using 'yarn' or 'npm' 

https://github.com/user-attachments/assets/6b86d692-e31b-448e-bab5-a27964158e68

+ Currently we have components state + URL param, so I just removed the URL part. See https://github.com/getsentry/sentry/blob/9ded21a2fcf5db1240b93355823f996b3389349d/static/app/views/onboarding/setupDocs.tsx#L75-L76

static/app/views/projectInstall/platformOrIntegration.tsx
priscilawebdev: good catch!

static/app/views/projectInstall/platformOrIntegration.tsx
ArthurKnaus: hmm, k. Personally I would not expect this state to be lost on reload. However in both case it is maybe an advantage as there is no way to go back otherwise... (we should fix that at some point)

priscilawebdev: yes! I would prefer to have the states in the URL too, but if we update this, then we update for both at once

PR #76542, opened 2024-08-26
feat(alerts): Adds insights metrics as an alert type

edwardgou-sentry:
image


edwardgou-sentry:
closed this pr so i can break it down to smaller ones

PR #76550, opened 2024-08-26
fix(discover): Split count_if default parameters by dataset

narsaynorath:
Updates the default parameters for the `count_if` function in Discover and Dashboards for the errors and transactions datasets.

The current default of `transaction.duration` errors out for the Errors dataset, so switch it to something more relevant to errors. I'm open to a different default. I wanted to use `error.handled` but it doesn't look like we expose that in dashboard. I could hold off on this PR to introduce `error.handled` if we think that it's valuable enough to do so.


shruthilayaj (approved):
looks good

PR #58005, opened 2023-10-12
fix(discord): Handle channel id none error

17hogeju:
[...]

PR #76553, opened 2024-08-26
fix(incidents): Require non-null target_value for Discord

RyanSkonnord:
Raise an InvalidTriggerActionError if the target_value is None.

Modify test_discord_channel_id_none to assert for this error. (It appears that the test case's original intent was to test that the serializer doesn't crash, but the case where the target_value is None would inevitably cause an error anyway when we call `DiscordClient.get_channel`.)

### DIFF
--- a/src/sentry/incidents/endpoints/serializers/alert_rule_trigger_action.py
+++ b/src/sentry/incidents/endpoints/serializers/alert_rule_trigger_action.py
@@ -42,7 +42,7 @@ class AlertRuleTriggerActionSerializer(Serializer):
                     "discord.action.description.no.channel",
                     extra={"target_identifier": action.target_identifier},
                 )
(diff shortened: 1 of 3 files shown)


RyanSkonnord:
See https://github.com/getsentry/sentry/pull/75541 for context.

Original changes in https://github.com/getsentry/sentry/pull/58005

asottile-sentry (approved):

asottile-sentry:
oh, I would just merge this into upstream directly -- that way we minimize the impact if this isn't correct

RyanSkonnord:
> oh, I would just merge this into upstream directly -- that way we minimize the impact if this isn't correct

👍 Done.

asottile-sentry:
by upstream I mean integrate this to `master` -- not your feature branch

PR #76557, opened 2024-08-26
fix(feedback): fix query params

michellewzhang:
- remove expand `owners` - we don't need it
- remove expand `stats` - it's on by default
- remove collapse `inbox` - doesn't exist
- remove expand `inbox` - we don't need it


scttcper (approved):
lookin good

PR #76559, opened 2024-08-26
fix(issue-summary): Hide issue summary on non-error issues

roaga:
Hides issue summary header and card if the issue is not an error issue (i.e. it's a performance, replay, or user feedback issue).

### DIFF
--- a/static/app/components/group/groupSummary.spec.tsx
+++ b/static/app/components/group/groupSummary.spec.tsx
@@ -2,6 +2,7 @@ import {render, screen, waitFor} from 'sentry-test/reactTestingLibrary';
 
 import {
   GroupSummary,
+  GroupSummaryHeader,
   makeGroupSummaryQueryKey,
 } from 'sentry/components/group/groupSummary';
 import {IssueCategory} from 'sentry/types/group';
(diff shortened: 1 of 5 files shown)


static/app/components/group/groupSummary.tsx
```diff
@@ -72,13 +75,24 @@ function GroupSummaryFeatureBadge() {
...
   const {data, isLoading, isError, hasGenAIConsent} = useGroupSummary(groupId);
   const isStreamlined = useHasStreamlinedUI();
 
-  if (isError || !hasGenAIConsent || (!isLoading && !data?.headline)) {
+  if (
+    isError ||
+    !hasGenAIConsent ||
+    (!isLoading && !data?.headline) ||
+    groupCategory !== IssueCategory.ERROR ||
+    groupTitle === 'User Feedback'
```
jennmueng, line 91: Interesting, do we not have a category for user feedback and it's an error?
I'l defer to @malwilley to review this one

roaga: Yeah some User Feedback issues were classified as ERROR, probably because you can get user feedback when an error occurs in addition to the standalone widget.

I think we still want to hide issue summaries for those, right? So I put this extra check.

malwilley: I think we would want to show issue summaries for those since it still has a stacktrace attached?

roaga: That's a good point, sometimes it does have a stacktrace. But there's also cases where it doesn't and instead links to a trace connected issue that does have the error.

Screenshot 2024-08-26 at 11 54 51 AM

static/app/components/group/groupSummary.spec.tsx
```diff
@@ -86,7 +94,119 @@ describe('GroupSummary', function () {
...
+        groupTitle={'Uh oh.'}
+      />
+    );
+
+    await waitFor(
+      () => {
+        expect(setupCall).toHaveBeenCalled();
+      },
+      {timeout: 5000}
+    );
```
scttcper, line 163: We should be disabling the api query if we can determine that we aren't going to use the result ahead of time.

static/app/components/group/groupSummary.tsx
malwilley: In this case I believe it should always be `issueCategory: 'feedback'`, which should be handled by the line above

static/app/components/group/groupSummary.tsx
```diff
@@ -72,13 +75,24 @@ function GroupSummaryFeatureBadge() {
   );
 }
 
-export function GroupSummaryHeader({groupId}: GroupSummaryProps) {
+export function GroupSummaryHeader({
+  groupId,
+  groupCategory,
+  groupTitle,
+}: GroupSummaryProps) {
   const {data, isLoading, isError, hasGenAIConsent} = useGroupSummary(groupId);
```
scttcper, line 83: useGroupSummary({groupId, enabled: groupCategory === IssueCategory.ERROR etc});

static/app/components/group/groupSummary.tsx
roaga: There is no feedback category, so it comes back as "error" in my testing. @malwilley 

Screenshot 2024-08-26 at 12 03 43 PM

malwilley: Hmm if you're seeing that in prod can you link it? I don't think that's expected.  [This for example](https://sentry.sentry.io/issues/5736361003/) has `issueCategory: 'feedback'` (coming from the backend, the frontend enum may need to be updated to include that value).

Either way not that big of a deal, just would be nice to remove logic checking for a specific title

roaga: Hmm well I don't see it in prod, removing the title check then

static/app/components/group/groupSummary.spec.tsx
roaga: Good point, done in next commit

static/app/components/group/groupSummary.spec.tsx
```diff
@@ -100,4 +102,174 @@ describe('GroupSummary', function () {
...
+          ],
+        },
+      },
+    });
+
+    render(<GroupSummaryHeader groupId={groupId} groupCategory={IssueCategory.ERROR} />);
+
+    await waitFor(() => {
+      expect(screen.getByText('Test headline')).toBeInTheDocument();
+    });
```
scttcper, line 189: nit but findBy does waitFor + getBy for you
```suggestion
      expect(await screen.findByText('Test headline')).toBeInTheDocument();
```

static/app/components/group/groupSummary.spec.tsx
```diff
@@ -100,4 +102,174 @@ describe('GroupSummary', function () {
...
+
+    render(<GroupSummaryHeader groupId={groupId} groupCategory={IssueCategory.ERROR} />);
+
+    await waitFor(
+      () => {
+        expect(setupCall).toHaveBeenCalled();
+      },
+      {timeout: 5000}
+    );
```
scttcper, line 232: another tiny nit but i would hope this call happens instantly not within 5 seconds
```suggestion
    await waitFor(
      () => {
        expect(setupCall).toHaveBeenCalled();
      }
    );
```

scttcper (approved):

static/app/components/group/groupSummary.spec.tsx
roaga: Thanks for the tips!

PR #76563, opened 2024-08-26
nit(anomaly detection): snake case to camel case for consistency

mifu67:
The serializer field for detection type should be in camel case, not snake case.


schew2381 (approved):
fastest stamp in the west

ceorourke (approved):

PR #76564, opened 2024-08-26
feat(anomaly detection): update alert rule details page for anomaly detection

mifu67:
Screenshot 2024-08-26 at 12 08 26 PM

Screenshot 2024-08-26 at 3 20 53 PM

### DIFF
--- a/static/app/views/alerts/list/rules/alertRuleStatus.tsx
+++ b/static/app/views/alerts/list/rules/alertRuleStatus.tsx
@@ -89,10 +89,10 @@ export default function AlertRuleStatus({rule}: Props) {
 
   return (
     <FlexCenter>
-      {rule.detectionType !== 'dynamic' ? (
+      {rule.detectionType !== AlertRuleComparisonType.DYNAMIC ?? (
         <IconArrow color={iconColor} direction={iconDirection} />
-      ) : null}
-      {rule.detectionType !== 'dynamic' ? (
+      )}
+      {rule.detectionType !== AlertRuleComparisonType.DYNAMIC ? (
         <TriggerText>
           {`${thresholdTypeText} ${
             rule.latestIncident || (!rule.latestIncident && !resolvedTrigger)
--- a/static/app/views/alerts/list/rules/alertRuleStatus.tsx
+++ b/static/app/views/alerts/list/rules/alertRuleStatus.tsx
@@ -89,7 +89,7 @@ export default function AlertRuleStatus({rule}: Props) {
 
   return (
     <FlexCenter>
-      {rule.detectionType !== AlertRuleComparisonType.DYNAMIC ?? (
+      {rule.detectionType !== AlertRuleComparisonType.DYNAMIC && (
         <IconArrow color={iconColor} direction={iconDirection} />
       )}
       {rule.detectionType !== AlertRuleComparisonType.DYNAMIC ? (
--- a/static/app/views/alerts/list/rules/alertRuleStatus.tsx
+++ b/static/app/views/alerts/list/rules/alertRuleStatus.tsx
@@ -69,6 +69,9 @@ export default function AlertRuleStatus({rule}: Props) {
       ? t('Above')
       : t('Below');
 
+  // Anomaly detection alerts have different labels
+  const statusLabel = activeIncident ? t('Bad') : t('Good');
+
   if (activeIncident) {
     iconColor =
       trigger?.label === AlertRuleTriggerType.CRITICAL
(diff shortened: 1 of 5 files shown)


static/app/views/alerts/list/rules/alertRuleStatus.tsx
```diff
@@ -86,20 +89,26 @@ export default function AlertRuleStatus({rule}: Props) {
...
-            : resolvedTrigger?.toLocaleString()
-        }`}
-        {getThresholdUnits(
-          rule.aggregate,
-          rule.comparisonDelta
-            ? AlertRuleComparisonType.CHANGE
-            : AlertRuleComparisonType.COUNT
-        )}
-      </TriggerText>
+      {rule.detectionType !== 'dynamic' ? (
```
mifu67, line 92: The explicit 'dynamic' string here is because the AlertRuleComparisonType options don't exactly match with the detection type strings. Is this okay, or how should I deal with this?

static/app/views/alerts/rules/metric/details/sidebar.tsx
```diff
@@ -260,6 +263,40 @@ export function MetricDetailsSidebar({
...
+          {rule.detectionType === 'dynamic' && (
+            <KeyValueTableRow
+              keyName={t('Seasonality')}
+              value={
+                rule.seasonality
+                  ? rule.seasonality.charAt(0).toUpperCase() + rule.seasonality.slice(1)
+                  : ''
+              } // NOTE: if the rule is dynamic, then there must be a seasonality
+            />
+          )}
```
ceorourke, line 285: Do we still need this given we're not exposing seasonality anymore?

static/app/views/alerts/rules/metric/details/sidebar.tsx
```diff
@@ -76,6 +76,7 @@ function TriggerDescription({
     AlertWizardAlertNames[getAlertTypeFromAggregateDataset(rule)]
   );
 
+  // TODO: update this text if rule type is dynamic
```
ceorourke, line 79: Was this addressed or is this a future TODO?

ceorourke (commented):
Left a couple comments but will let a front endier person approve

static/app/views/alerts/rules/metric/details/sidebar.tsx
mifu67: This was addressed, removing

static/app/views/alerts/rules/metric/details/sidebar.tsx
mifu67: Oops

static/app/views/alerts/list/rules/alertRuleStatus.tsx
nhsiehgit: 🤔 Does it not?
I'm seeing 
```
export enum AlertRuleComparisonType {
  COUNT = 'count',
  CHANGE = 'change',
  PERCENT = 'percent',
  DYNAMIC = 'dynamic',
}
```
as the exported type?

static/app/views/alerts/list/rules/alertRuleStatus.tsx
```diff
@@ -86,20 +89,26 @@ export default function AlertRuleStatus({rule}: Props) {
...
+        <IconArrow color={iconColor} direction={iconDirection} />
+      ) : null}
```
nhsiehgit, line 94: rather than `: null` you can use `&&` rather than the ternary operator

static/app/views/alerts/rules/metric/constants.tsx
```diff
@@ -172,7 +172,7 @@ export function createDefaultRule(
     environment: null,
     resolveThreshold: '',
     thresholdType: AlertRuleThresholdType.ABOVE,
-    detectionType: AlertRuleComparisonType.COUNT,
+    detectionType: 'static',
```
nhsiehgit, line 175: can we add `static` to the enum?

static/app/views/alerts/rules/metric/details/sidebar.tsx
```diff
@@ -260,6 +262,30 @@ export function MetricDetailsSidebar({
...
+                  : ''
+              } // NOTE: if the rule is dynamic, then there must be a sensitivity
+            />
+          )}
+          {rule.detectionType === 'dynamic' && (
+            <KeyValueTableRow
+              keyName={t('Direction')}
+              value={
+                rule.thresholdType === AlertRuleThresholdType.ABOVE ? (
+                  <OverflowTableValue>{'Above threshold'}</OverflowTableValue>
```
nhsiehgit, line 280: nit: I guess the `<OverflowTableValue>` can probably go on the outside of the ternaries?
then we could maybe have a switch case for the strings?

nhsiehgit (commented):
mostly nit comments

although the `'dynamic'` strings feels like they should be just fine as enums?
can you expand on what you're seeing?

mifu67:
> mostly nit comments
> 
> although the `'dynamic'` strings feels like they should be just fine as enums? can you expand on what you're seeing?

@nhsiehgit what's different is "count" instead of "static" and "change" instead of "percent"

static/app/views/alerts/list/rules/alertRuleStatus.tsx
```diff
@@ -86,20 +89,26 @@ export default function AlertRuleStatus({rule}: Props) {
...
-            : resolvedTrigger?.toLocaleString()
-        }`}
-        {getThresholdUnits(
-          rule.aggregate,
-          rule.comparisonDelta
-            ? AlertRuleComparisonType.CHANGE
-            : AlertRuleComparisonType.COUNT
-        )}
-      </TriggerText>
+      {rule.detectionType !== AlertRuleComparisonType.DYNAMIC ?? (
```
nhsiehgit, line 92: nit: `??` (nullish coalescing) i think is mainly used when we want to default a `null` or undefined value.
I _think_ `&&` is prefered when you want to conditionally determine whether to render something or not.

nhsiehgit (approved):
👍

static/app/views/alerts/list/rules/alertRuleStatus.tsx
mifu67: oops

PR #76566, opened 2024-08-26
feat(alerts): Update metricSearchBar to support insights metric specific filter keys

edwardgou-sentry:
Insights metric alerts also support the `has` filter, and `span.module` field alias filter. Neither of these come from the `tags` or `meta` endpoints so we need to hard code them as a special case here.

Also adds a `isInsightsMetricAlert` helper function, which will be used in later prs

### DIFF
--- a/static/app/components/metrics/metricSearchBar.spec.tsx
+++ b/static/app/components/metrics/metricSearchBar.spec.tsx
@@ -3,7 +3,9 @@ import {render, screen, userEvent} from 'sentry-test/reactTestingLibrary';
 import {MetricSearchBar} from 'sentry/components/metrics/metricSearchBar';
 
 describe('metricSearchBar', function () {
+  const onChange = jest.fn();
   beforeEach(() => {
+    onChange.mockReset();
     MockApiClient.addMockResponse({
       url: '/organizations/org-slug/metrics/meta/',
       body: [],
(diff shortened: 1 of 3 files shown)


static/app/components/metrics/metricSearchBar.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+      url: '/organizations/org-slug/metrics/tags/span.module/',
+      body: [],
+    });
+    render(
+      <MetricSearchBar onChange={onChange} mri="d:spans/exclusive_time@millisecond" />
+    );
+    await screen.findByPlaceholderText('Filter by tags');
+    await userEvent.type(screen.getByPlaceholderText('Filter by tags'), 'span.module:db');
+    await userEvent.keyboard('{enter}');
+    expect(onChange).toHaveBeenCalledWith('span.module:"db"');
```
priscilawebdev, line 49: Thanks for adding tests for the metricsSearchBar 👏🚀

They look great, but I have a thought. In the last test, should we also check if the message "The field span.module isn't supported here." is not being rendered? Right now, we assume that if `onChange` is called, the message isn't shown. The same goes for the other tests -  should we also verify that `onChange` is not called?

priscilawebdev (approved):
left a few questions but otherwise it lgtm!

static/app/components/metrics/metricSearchBar.spec.tsx
edwardgou-sentry: Updated the tests! I also realized that `MetricsSearchBar` uses either `SmartSearchBar` or `SearchQueryBuilder` depending on a feature flag, so I've updated to test both components!

PR #76572, opened 2024-08-26
fix(grouping): Fix project-deletion-triggered `GroupHash` deletion

lobsterkatie:
This is a follow-up to https://github.com/getsentry/sentry/pull/76312, fixing another bug with grouphash deletion. Currently, when a project is deleted, its associated `GroupHash` records are deleted directly by the project deletion task, using the `BulkModelDeletionTask`. Unfortunately, that task doesn't take into account any dependent tables, meaning that the corresponding `GroupHashMetadata` records aren't deleted as they should be.

This fixes that by allowing `GroupHash` deletions to cascade from group deletions (which themselves are set off by project deletion), rather than having the project deletion task delete the grouphashes directly. (Group deletions already handle `GroupHash`/`GroupHashMetadata` deletion correctly, using the `GroupHash`-specific deletion task [registered in the deletion module's `__init__.py`](https://github.com/getsentry/sentry/blob/9155480cddb7454bba99ae39ea4caac91261ca26/src/sentry/deletions/__init__.py#L127).)

Note that while the results of this change aren't tested in this PR, they are implicitly tested in https://github.com/getsentry/sentry/pull/76245, which will directly follow this one and which for the first time will actually create `GroupHashMetadata` records. Without the change in this PR, the presence of that PR's new `GroupHashMetadata` records causes the project deletion test to fail, but with this change, they pass.

### DIFF
--- a/src/sentry/deletions/defaults/project.py
+++ b/src/sentry/deletions/defaults/project.py
@@ -20,7 +20,6 @@ class ProjectDeletionTask(ModelDeletionTask):
         from sentry.models.groupassignee import GroupAssignee
         from sentry.models.groupbookmark import GroupBookmark
         from sentry.models.groupemailthread import GroupEmailThread
-        from sentry.models.grouphash import GroupHash
         from sentry.models.grouprelease import GroupRelease
         from sentry.models.grouprulestatus import GroupRuleStatus
         from sentry.models.groupseen import GroupSeen
@@ -54,7 +53,6 @@ class ProjectDeletionTask(ModelDeletionTask):
             GroupAssignee,
             GroupBookmark,
             GroupEmailThread,
-            GroupHash,
             GroupRelease,
             GroupRuleStatus,
             GroupSeen,


jangjodi:
is the deletion of `GroupHash` still being done in bulk, and if not, is this a concern? I see that `GroupHashDeletionTask`  extends `ModelDeletionTask`

lobsterkatie:
> is the deletion of `GroupHash` still being done in bulk, and if not, is this a concern? I see that `GroupHashDeletionTask` extends `ModelDeletionTask`

So there are two ways a `GroupHash` deletion can be triggered - by deleting a group or by deleting a project. When it's a group that's deleted, it's already using the grouphash-specific deletion task (which, as you say, isn't bulk). And deleting a project deletes all the groups, so another option here would be to just remove `GroupHash` from the project deletion task, and let the `GroupHash` and `GroupHashMetadata` deletes cascade from the `Group` deletion.

 In any case, no, I don't think it's an issue, because if you look at that `__init__` file where all the deletion tasks are registered (that I linked in the PR description), lots of models have bespoke deletion tasks, and they all inherit from `ModelDeletionTask`.  Also, we have no choice here, because (as is the case with many of the models which have custom tasks), the delete has to cascade, and the way that happens is by subclassing and overriding the `get_child_relations` method (which is called by `ModelDeletionTask` but not by `BulkModelDeletionClass`).

JoshFerge:
are there appropriate tests to ensure that groups still get deleted when a project is?

lobsterkatie:
> are there appropriate tests to ensure that groups still get deleted when a project is?

That _groups_ get deleted? Yeah, we test that [here](https://github.com/getsentry/sentry/blob/9d84e4ab7ba4f9a467a0e02fa59fc407f7f9bf7b/tests/sentry/deletions/test_project.py#L146). But why would changing what triggers `GroupHash` deletion change the deletion of groups?

JoshFerge:
> > are there appropriate tests to ensure that groups still get deleted when a project is?
> 
> That _groups_ get deleted? Yeah, we test that [here](https://github.com/getsentry/sentry/blob/9d84e4ab7ba4f9a467a0e02fa59fc407f7f9bf7b/tests/sentry/deletions/test_project.py#L146). But why would changing what triggers `GroupHash` deletion change the deletion of groups?

sorry, meant `Group`

lobsterkatie:
> sorry, meant `Group`

Yeah, no, I know. That's what you said originally. And we do test for it, in the spot I linked above. I'm just confused why that's a question here.

JoshFerge (approved):

PR #76574, opened 2024-08-26
fix(codeowners): Fix bug with multiple codeowners 

### LINKED ISSUE #73021: CODOWNERS evaluation flow was not followed - last matching rule did not define assignment
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

I have a CODEOWNERS file linked to Sentry with the following rule `codeowners:/app/processing_unit.rb #processing-team` on line **1237** and a rule `codeowners:*payment* #payments` on line **759**. I have the following stacktrace:

File "/app/payment_service.rb", line 13, in block in call
File "/app/processing_unit.rb", line 63, in charge!
File "/app/processing_unit.rb", line 174, in call!
  
[Customer Case](https://www.notion.so/sentry/CODEOWNERS-evaluation-flow-not-respected-d1666ac8767e4b96b263184890338e62?pvs=4)

### Expected Result

I would expect that based on the [evaluation flow](https://docs.sentry.io/product/issues/ownership-rules/#evaluation-flow), the `#processing-team` would have been assigned to the issue since this rule is lower than the `#payments` rule

### Actual Result

The issue was assigned to the `#payments` team

### Product Area

Issues

### Link

_No response_

### DSN

_No response_

### Version

_No response_

roggenkemper:
this pr fixes a bug with codeowners where the wrong codeowner was selected. this is because `get_issue_owners` returns 2 owners but when auto-assignment happened, it only filtered by type. In the case that there were 2 owners of the same type, it didn't specify which one to pick. Since they are created in the correct order,  we can sort by `date_created` to find the one that's created first which should be the correct one. 

fixes https://github.com/getsentry/sentry/issues/73021
Closes https://github.com/getsentry/sentry/issues/73021

### DIFF
--- a/src/sentry/models/groupowner.py
+++ b/src/sentry/models/groupowner.py
@@ -111,7 +111,7 @@ class GroupOwner(Model):
                 group_id=group_id, project_id=project_id, type__in=autoassignment_types
             )
             .exclude(user_id__isnull=True, team_id__isnull=True)
-            .order_by("type")
+            .order_by("type", "date_added")
             .first()
         )
         # should return False if no owner
--- a/src/sentry/models/groupowner.py
+++ b/src/sentry/models/groupowner.py
@@ -106,6 +106,9 @@ class GroupOwner(Model):
         """
         Non-cached read access to find the autoassigned GroupOwner.
         """
+
+        # Ordered by date_added as well to ensure that the first GroupOwner is returned
+        # Multiple GroupOwners can be created but they are created in the correct evaluation order, so the first one takes precedence
         issue_owner = (
             cls.objects.filter(
                 group_id=group_id, project_id=project_id, type__in=autoassignment_types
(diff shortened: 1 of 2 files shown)


src/sentry/models/groupowner.py
```diff
@@ -111,7 +111,7 @@ def get_autoassigned_owner(cls, group_id, project_id, autoassignment_types):
                 group_id=group_id, project_id=project_id, type__in=autoassignment_types
             )
             .exclude(user_id__isnull=True, team_id__isnull=True)
-            .order_by("type")
+            .order_by("type", "date_added")
```
leeandher, line 114: Question: In theory, wouldn't we want to assign the _newest_ group owner? That would be the one most recently created by an ownership rule, and since those are processed in order and the bottom-most rule takes precedent?

leeandher: that is wild, great catch 😖 

Might be worth calling out in a comment as well.

leeandher (approved):

JoshFerge (approved):
🤩 bravo!!

PR #76578, opened 2024-08-26
nit(anomaly detection): edit copy to fit spec

mifu67:
Screenshot 2024-08-26 at 3 41 20 PM


ceorourke:
Thanks! I know we discussed linking to external docs (when they exist) - do we have an idea of where that link would go?

ceorourke (approved):

mifu67:
> Thanks! I know we discussed linking to external docs (when they exist) - do we have an idea of where that link would go?

Probably a link after "Anomaly: whenever values are outside of expected bounds"

PR #76579, opened 2024-08-26
fix(issues): Restyle non member in suspect commit

scttcper:
- Removes "not a member" and replaces it with a question tooltip thing.
- Fixes long emails overflowing

https://www.notion.so/sentry/Suspect-commit-underlines-mean-different-things-7d4bfb01840945e4b1bd913f539f337b?pvs=4

before - ambiguous link vs hover area:
image

after:
image

hoverstate
image


malwilley (approved):
Noticed one thing which probably doesn't matter - hovering over someone's name that is a member still has the hover state even though there isn't a tooltip

PR #76582, opened 2024-08-27
fix(issues): Adjust resolved gradient for dark mode

scttcper:
Attempt to make it less bright in dark mode

before:
image

after:
image

before (light):
image

after (light):
image


leeandher (approved):
good catch 🌔

PR #76585, opened 2024-08-27
ref(models): Remove with_projects param from TeamManager.get_for_user

RyanSkonnord:
The parameter appeared to be unused. Removing it allows us to drop extra querying code, and to simplify the return type and remove several overload signatures.

### DIFF
--- a/src/sentry/models/team.py
+++ b/src/sentry/models/team.py
@@ -1,8 +1,7 @@
 from __future__ import annotations
 
-from collections import defaultdict
 from collections.abc import Sequence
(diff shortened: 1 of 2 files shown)


RyanSkonnord:
I intended to split this into two methods (with different return type annotations) but couldn't find any cases where `with_projects` was in use. Running CI to make sure I haven't missed anything, but I think it's just dead code.

Let me know if there's any relevant history. I vaguely remember this being connected to the permissions system.

asottile-sentry (approved):

PR #76594, opened 2024-08-27
ref(relay): Only update internal state when it changed

Dav1dde:
Removes the `is_internal` update from the `relay_from_id` function, to move the check to the endpoint (where it already was anyways). All additional uses of `relay_from_id` don't seem to care about the internal state, except authentication. 
So that seems fine, also this behaviour does seem more consistent.

### DIFF
--- a/src/sentry/api/endpoints/relay/register_response.py
+++ b/src/sentry/api/endpoints/relay/register_response.py
@@ -90,7 +90,7 @@ class RelayRegisterResponseEndpoint(Endpoint):
                 relay = Relay.objects.create(
                     relay_id=relay_id, public_key=public_key, is_internal=is_internal
                 )
-            elif not relay.is_internal == is_internal:
+            elif relay.is_internal != is_internal:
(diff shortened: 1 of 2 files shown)


beezz (approved):

src/sentry/api/endpoints/relay/register_response.py
```diff
@@ -90,7 +90,7 @@ def post(self, request: Request) -> Response:
                 relay = Relay.objects.create(
                     relay_id=relay_id, public_key=public_key, is_internal=is_internal
                 )
-            else:
+            elif not relay.is_internal == is_internal:
```
jjbayer, line 93:
```suggestion
            elif relay.is_internal != is_internal:
```

jjbayer (approved):

src/sentry/api/endpoints/relay/register_response.py
Dav1dde: 3519 times `!=` and 305 for `not.*==`

PR #76596, opened 2024-08-27
ref(platformIntegrationSetup): Convert class to func

priscilawebdev:


static/app/views/projectInstall/platformIntegrationSetup.tsx
```diff
@@ -1,172 +1,137 @@
...
+  const [installed, setInstalled] = useState(false);
+  const navigate = useNavigate();
+
+  const {data: integrations} = useApiQuery<{providers: IntegrationProvider[]}>(
+    [
+      `/organizations/${organization.slug}/config/integrations/?provider_key=${integrationSlug}`,
+    ],
+    {
+      enabled: !!integrationSlug,
+      staleTime: Infinity,
```
priscilawebdev, line 45:
```suggestion
      staleTime: 0,
```

static/app/views/projectInstall/platformIntegrationSetup.tsx
```diff
@@ -1,172 +1,137 @@
...
+    window.scrollTo(0, 0);
+    // redirect if platform is not known.
+    if (!platform || platform.id === 'other') {
+      redirectToNeutralDocs();
     }
+  }, [redirectToNeutralDocs, platform]);
+
+  if (!provider || !platform || !project) {
+    return null;
+  }
```
scttcper, line 71: I think there needs to be a loading state, the AsyncComponent before was handling this

priscilawebdev: oh yes sure! always when I convert a component that was previously using `AsyncComponent` I forget about the loading state. Thanks 🙏

static/app/views/projectInstall/platformIntegrationSetup.tsx
```diff
@@ -1,172 +1,150 @@
...
+      `/organizations/${organization.slug}/config/integrations/?provider_key=${integrationSlug}`,
+    ],
+    {
+      enabled: !!integrationSlug,
+      staleTime: 0,
+    }
+  );
 
+  useEffect(() => {
+    window.scrollTo(0, 0);
```
ArthurKnaus, line 59: I would move the `scrollTo` into a separate `useEffect` to make sure it only runs on mount.

ArthurKnaus (approved):

PR #76610, opened 2024-08-27
feat(query-builder): Add easy multi-select while holding ctrl/cmd

malwilley:
This adds the "Hold Ctrl/⌘ to select multiple" text and behavior that was in the Figma designs. You could already click the checkbox for this behavior, this makes it possible to do with the keyboard now as well.

CleanShot 2024-08-27 at 10 32 11@2x

https://github.com/user-attachments/assets/22fce3ec-089e-45a2-b040-1781f9867839

### DIFF
--- a/static/app/components/searchQueryBuilder/tokens/filter/valueListBox.tsx
+++ b/static/app/components/searchQueryBuilder/tokens/filter/valueListBox.tsx
@@ -1,0 +1,85 @@
+import styled from '@emotion/styled';
+import {isMac} from '@react-aria/utils';
+
+import {ListBox} from 'sentry/components/compactSelect/listBox';
+import type {SelectOptionOrSectionWithKey} from 'sentry/components/compactSelect/types';
+import {Overlay} from 'sentry/components/overlay';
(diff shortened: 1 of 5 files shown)


static/app/components/searchQueryBuilder/tokens/filter/valueListBox.tsx
```diff
@@ -0,0 +1,85 @@
...
+import {itemIsSection} from 'sentry/components/searchQueryBuilder/tokens/utils';
+import {t} from 'sentry/locale';
+import {space} from 'sentry/styles/space';
+
+interface ValueListBoxProps<T> extends CustomComboboxMenuProps<T> {
+  isMultiSelect: boolean;
+  items: T[];
+}
+
+export function ValueListBox<T extends SelectOptionOrSectionWithKey<string>>({
```
malwilley, line 17: I've been using the `customMenu` prop for other customized menus, so I did the same thing here for the value menu. It's mostly the same as the default combobox behavior but I didn't want to add more bloat to that one

scttcper (approved):

Issue #76289, opened 2024-08-15
Search for Session Replays are rate limited in the UI


masinette:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

1. Search for session replays (with many results)

_specific example linked in shadow ticket_

### Expected Result

Search loads a list

### Actual Result

![Image](https://github.com/user-attachments/assets/03d8cdc9-de6f-4a01-b3e9-ab11ee93dc5d)

![Image](https://github.com/user-attachments/assets/1bbd28cc-89bd-42ce-b90c-dde3b133fdf5)

### Product Area

Replays

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2255) by [Unito](https://www.unito.io)

bruno-garcia:
We seem to have a quite short rate limit going here.
When I tried to repro, I couldn't for a few tries. Data loaded properly, I believe, 3x full page refresh. Then I got rate limited.

Thanks for raising, we'll dig into what's going on.

rafalzv7labs:
Hi Bruno, do you have any updates? This is still affecting our whole engineering and support teams, as we're rarely able to search for session replays.

I'm happy to record a loom video reproducing the issue, if that helps. We're keen to have this issue resolved, as it heavily affects our productivity. Thanks.

bruno-garcia:
Thanks for writing in @rafalzv7labs. This is a p0 for us and we'll be looking into to it asap.

bruno-garcia:
We identified the rate limit is coming from Snuba: https://github.com/getsentry/sentry/blob/97a906cd8b69a52dd2c56928438e897e033a55ad/src/sentry/utils/snuba.py#L338

Still looking at values and other conditions

bruno-garcia:
We increased limits on our end and I couldn't repro being rate limited after a few refreshes.
Could you verify things work as you expected?

If not, please give us more information about the usage patterns. Is there a script calling into our APIs on your end?
Does this happen every N hours? Some times of the day? etc

cmanallen:
For the tag querying issue we need to apply an index to the tags field and update our query optimizer to use the index when possible.  The way we'll index the field requires a new materialized column and will take 90 days to back fill.  Though we may use it opportunistically before then by introspecting the time range of the query.

cmanallen:
[...]

PR #76611, opened 2024-08-27
chore(replays): Capture rate-limit exceptions

cmanallen:
Force the capture of rate limit exceptions and explicitly tag them with replay metadata.

### DIFF
--- a/src/sentry/replays/usecases/query/__init__.py
+++ b/src/sentry/replays/usecases/query/__init__.py
@@ -19,6 +19,7 @@ from collections.abc import Mapping, Sequence
 from datetime import datetime, timedelta
 from typing import Any, Literal, cast
 
+import sentry_sdk
 from rest_framework.exceptions import ParseError
 from snuba_sdk import (
     And,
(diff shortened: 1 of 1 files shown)


aliu39 (approved):
~~could we add project as an extra too?~~

aliu39:
Ref https://github.com/getsentry/sentry/issues/76289

src/sentry/replays/usecases/query/__init__.py
```diff
@@ -450,15 +451,22 @@ def _select_from_fields() -> list[Column | Function]:
...
+                dataset="replays",
+                app_id="replay-backend-web",
+                query=query,
+                tenant_ids=tenant_id,
+            ),
+            referrer,
+        )
+    except RateLimitExceeded as exc:
+        sentry_sdk.set_tag("replay-rate-limit-exceeded", True)
+        sentry_sdk.set_extra("org_id", tenant_id.get("organization_id"))
```
bruno-garcia, line 466: we don't have `org_id` already instrumented at a different level, so all events have it set?

Also: should it be a tag instead?

```suggestion
        sentry_sdk.set_tag("org_id", tenant_id.get("organization_id"))
```

bruno-garcia (approved):

PR #76616, opened 2024-08-27
chore(derived_code_mappings): Report known issues as non-errors

armenzg:
This reports the exceptions as warning-level errors rather than error-level.

Deployments do not get paused if the error is [warning or lower](https://github.com/getsentry/getsentry/blob/b08907d9da3e8856c6cbedbd86d4a722c6f89dd9/gocd/templates/bash/backend/check-sentry-new-errors.sh#L12):
```
--additional-query="issue.type:error !level:info !level:warning"
```

Rendering of a warning-level exception:
image

PS: This is a redo of #76212 but without passing a level, thus, making it an info-level message.

### DIFF
--- a/src/sentry/integrations/github/client.py
+++ b/src/sentry/integrations/github/client.py
@@ -8,6 +8,7 @@ from typing import Any, cast
 import orjson
 import sentry_sdk
 from requests import PreparedRequest
+from sentry_sdk import capture_exception, capture_message
 
 from sentry.constants import ObjectStatus
 from sentry.integrations.github.blame import (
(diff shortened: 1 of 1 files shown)


src/sentry/integrations/github/client.py
```diff
@@ -427,7 +428,7 @@ def _populate_trees_process_error(self, error: ApiError, extra: dict[str, str])
         else:
             # We do not raise the exception so we can keep iterating through the repos.
             # Nevertheless, investigate the error to determine if we should abort the processing
-            logger.error("Continuing execution. Investigate: %s", error_message, extra=extra)
+            capture_message("Continuing execution. Investigate: %s", error_message, extra=extra)
```
armenzg, line 431: In the original PR it used to be this:
image

I'm removing `level="warning"` since it is not accepted.

snigdhas (approved):

PR #76619, opened 2024-08-27
fix(issues/replay): if fetchError, don't render replayId in highlights

michellewzhang:
redo of https://github.com/getsentry/sentry/pull/76316: 

instead of completely removing the replay ID from the event highlights, we now do an extra check to see if there was a `fetchError` with this replay ID (e.g. whether this ID is actually valid or not). if it's not valid, then don't render the replay ID to avoid linking to an invalid replay. the link will take you to a page like this:

SCR-20240827-kqxv

before: issue details replay processing error (blue banner shows up) but the event highlights still contains the replay ID
SCR-20240827-krda
SCR-20240827-kqac

after: replayId is shown as `--` and no link
SCR-20240827-kpwz

### DIFF
--- a/static/app/components/events/highlights/highlightsDataSection.tsx
+++ b/static/app/components/events/highlights/highlightsDataSection.tsx
@@ -154,16 +154,16 @@ function HighlightsData({
   const contextReplayItem = highlightContextDataItems.find(
     e => e.data[0].key === 'replay_id'
   );
-  const contextReplayId = contextReplayItem?.value ?? '';
+  const contextReplayId = contextReplayItem?.value ?? EMPTY_HIGHLIGHT_DEFAULT;
 
   const tagReplayItem = highlightTagItems.find(e => e.originalTag.key === 'replayId');
-  const tagReplayId = tagReplayItem?.value ?? '';
+  const tagReplayId = tagReplayItem?.value ?? EMPTY_HIGHLIGHT_DEFAULT;
 
   // if the id doesn't exist for either tag or context, it's rendered as '--'
   const replayId =
-    contextReplayId.length > 2
+    contextReplayId !== EMPTY_HIGHLIGHT_DEFAULT
       ? contextReplayId
-      : tagReplayId.length > 2
+      : tagReplayId !== EMPTY_HIGHLIGHT_DEFAULT
         ? tagReplayId
         : undefined;
 
--- a/static/app/components/events/highlights/highlightsDataSection.tsx
+++ b/static/app/components/events/highlights/highlightsDataSection.tsx
@@ -169,16 +169,16 @@ function HighlightsData({
 
   const {fetchError: replayFetchError} = useReplayData({
     orgSlug: organization.slug,
-    replayId: replayId,
+    replayId,
   });
 
-  // if fetchError, replace the replay id so we don't link to an invalid replay
+  // if fetchError, replace the replayId so we don't link to an invalid replay
   if (contextReplayItem && replayFetchError) {
-    contextReplayItem.value = '--';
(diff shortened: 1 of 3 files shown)


static/app/components/events/highlights/highlightsDataSection.tsx
```diff
@@ -147,6 +148,39 @@ function HighlightsData({
...
+  // find the replayId from either context or tags, if it exists
+  const contextReplayItem = highlightContextDataItems.find(
+    e => e.data[0].key === 'replay_id'
+  );
+  const contextReplayId = contextReplayItem?.value ?? '';
+
+  const tagReplayItem = highlightTagItems.find(e => e.originalTag.key === 'replayId');
+  const tagReplayId = tagReplayItem?.value ?? '';
+
+  // if the id doesn't exist for either tag or context, it's rendered as '--'
```
c298lee, line 162: to clarify, the contextReplayItem.value and tagReplayItem.value could be '--'?

michellewzhang: yup, if there isn't a value for it, then that's the default value returned

static/app/components/events/highlights/highlightsDataSection.tsx
```diff
@@ -147,6 +148,39 @@ function HighlightsData({
...
+  const replayId =
+    contextReplayId.length > 2
```
c298lee, line 164: nit: can we change contextReplayId and tagReplayId to default to the default value returned too (`--`) and check directly for the default value instead of using the length? so we can make this more specific

michellewzhang: yup, updated

static/app/components/events/highlights/highlightsDataSection.tsx
```diff
@@ -147,6 +148,39 @@ function HighlightsData({
...
+  const contextReplayId = contextReplayItem?.value ?? EMPTY_HIGHLIGHT_DEFAULT;
+
+  const tagReplayItem = highlightTagItems.find(e => e.originalTag.key === 'replayId');
+  const tagReplayId = tagReplayItem?.value ?? EMPTY_HIGHLIGHT_DEFAULT;
+
+  // if the id doesn't exist for either tag or context, it's rendered as '--'
+  const replayId =
+    contextReplayId === EMPTY_HIGHLIGHT_DEFAULT
+      ? contextReplayId
+      : tagReplayId === EMPTY_HIGHLIGHT_DEFAULT
```
c298lee, line 166:
```suggestion
    contextReplayId !== EMPTY_HIGHLIGHT_DEFAULT
      ? contextReplayId
      : tagReplayId !== EMPTY_HIGHLIGHT_DEFAULT
```

c298lee (approved):

static/app/views/issueDetails/groupEventDetails/groupEventDetails.spec.tsx
```diff
@@ -198,13 +198,19 @@ const mockGroupApis = (
   project: Project,
   group: Group,
   event: Event,
+  replayId: string | undefined,
```
leeandher, line 201: nit: Could we make this optional instead of explicitly `undefined`? That way you don't have to add `undefined` to all the callers, only when you want to specify the trace I guess.

```js
replayId?: string;
```

leeandher (approved):
Thanks for doing this! Even if it was only for a portion of users I think this makes a better experience :+1:

static/app/views/issueDetails/groupEventDetails/groupEventDetails.spec.tsx
michellewzhang: yup updated!

PR #76623, opened 2024-08-27
feat(alerts): Updates snuba subscription functions to support discover spm() function

edwardgou-sentry:
The discover `spm()` function takes a `start` and `end` from the query parameters time to compute a result.

Since the create, update, and delete snuba subscription functions call `entity_subscription.build_query_builder(...)` without providing a `start` and `end` parameter, the query builder fails for `spm()` aggregates.

This updates `entity_subscription.build_query_builder(...)` calls to pass in a start and end time to avoid failing.
Since the `start` and `end` times aren't actually used for anything else in these code paths, it should be safe to just pass in any time range, similar to how we're doing it here:
https://github.com/getsentry/sentry/blob/9941c4b3b6121e4e42ac28978c661b6c61e2bfb3/src/sentry/incidents/serializers/alert_rule.py#L345

### DIFF
--- a/src/sentry/snuba/entity_subscription.py
+++ b/src/sentry/snuba/entity_subscription.py
@@ -4,8 +4,10 @@ import re
 from abc import ABC, abstractmethod
 from collections.abc import Mapping, MutableMapping, Sequence
 from dataclasses import dataclass
+from datetime import timedelta
 from typing import Any, TypedDict, Union
 
+from django.utils import timezone
 from snuba_sdk import Column, Condition, Entity, Join, Op, Request
 
 from sentry import features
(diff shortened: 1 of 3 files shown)


edwardgou-sentry:
> Since the start and end times aren't actually used for anything else in these code paths, it should be safe to just pass in any time range

I'm pretty sure this is true, these functions look like they're only used for managing subscriptions, and not running the actual queries.

nhsiehgit:
> I'm pretty sure this is true, these functions look like they're only used for managing subscriptions, and not running the actual queries

🤔 hmmm... i'm not entirely sure...
since they're already included in the paramtype, i'm a bit cautious about this.

rather than requiring the start/end to be passed for your specific use case (else be unused), can you just modify the spm function to handle not having it gracefully?

edwardgou-sentry:
Hey @nhsiehgit , I took a look to see if there was a way to disable the start/end validation on spm() for this use case, and it doesn't seem like there's a quick an easy way. 

Maybe @wedamija knows if it would be safe to update these calls to include dummy start and end values? I noticed Dan has a lot of edits in these code areas.

shruthilayaj:
I see there's a `skip_time_condition` flag in the config, I wonder if we could somehow use that to bypass the start/end validation on the spm function
https://github.com/getsentry/sentry/blob/65b1eb13a426d0f3cf50daa7bf329e3be7ce3ac3/src/sentry/snuba/entity_subscription.py#L198

edwardgou-sentry:
I figured out a better way to handle this by extending the `skip_time_conditions` option to also guard the `spm` function:
https://github.com/getsentry/sentry/pull/76759

Closing this PR in favour of the above PR!

PR #76626, opened 2024-08-27
feat(dashboards): Allow cancelling a zoom selection

### LINKED ISSUE #44723: Allow escaping out of window selection on charts
### Problem Statement

When dragging a chart you cannot escape out of the dateime window selection, which forces a url transition.

image

### Solution Brainstorm

Being able to hit escape here to cancel the selection would be ideal

gggritso:
Closes https://github.com/getsentry/sentry/issues/44723. Pressing the "Escape" key while making a drag selection on a chart will cancel the selection! Makes it possible to change your mind halfway through.

**e.g.,**

https://github.com/user-attachments/assets/f4d52ddd-dec7-41d9-aaa9-c8b1dc10f6b2

## Context

The code's a little wacky looking, so here are a few of the important circumstances:

- I don't want to fire a `"restore"` event to all charts, just the one that was selected! It's probably safe to fire `"restore"` everywhere, but it feels sloppy
- I chose a janky `querySelector` way to get the chart DOM element. It's hard to do any other way because `ChartZoom` renders an arbitrary child, so it's hard for it to get at the chart's `ref`
- ECharts does not have any `mousedown` or `mouseup` handlers _for the chart area_, only for the actual data lines. There's a feature request to add those handlers, but they don't exist. Therefore I have to add them manually
- Even if ECharts _had_ those events, `ChartZoom` would have to hijack them, and there wouldn't be a nice way to pass them down to the `BaseChart` so that approach is moot
- Some dashboards have a lot of charts. I didn't want to register an event handler for every chart. Instead, only the `"mousedown"` handler is registered at mount. The other handlers are registered in `"mousedown"`!
- The `"restore"` handler currently jumps to the _earliest_ point in the chart zoom history. Why does it do that, and not the most recent one? I don't know! I chose to avoid that history logic altogether and not run any `"restore"` handlers at all if it's for the sake of cancelling a zoom

### DIFF
--- a/static/app/components/charts/chartZoom.tsx
+++ b/static/app/components/charts/chartZoom.tsx
@@ -102,8 +102,8 @@ class ChartZoom extends Component<Props> {
   }
 
   componentWillUnmount(): void {
-    document.removeEventListener('keydown', this.handleKeyDown);
-    document.removeEventListener('mouseup', this.handleMouseUp);
+    document.body.removeEventListener('keydown', this.handleKeyDown);
+    document.body.removeEventListener('mouseup', this.handleMouseUp);
     this.$chart?.removeEventListener('mousedown', this.handleMouseDown);
   }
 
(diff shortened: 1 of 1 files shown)


0Calories (approved):
This is great, thanks George! One thing to note though; The span samples panels that are present in various Insights pages are also listening to the keydown event for ESC, so when you press ESC while cancelling a zoom on one of these panels, it ends up closing the panel.

Note a huge deal or blocking, but the behaviour feels a bit weird in this instance

0Calories:
Might be worthwhile to write a test for this too

gggritso:
@0Calories great catch on the panel behaviour, thanks! Fixed now, I changed how I register the `"keydown"` handler.

I'll think about tests, too

PR #76627, opened 2024-08-27
ref(scm): search endpoint abstraction

cathteng:
Search endpoints do two things: search issues and search repositories (optional; not used in VSTS).

The shared logic for all SCMs is the following:

1. Find the integration (also using a specific `provider` string for everything except GH/GH Enterprise)
2. Validate the incoming payload
    1. `field` (str)
    2. `query` (str)
3. Get integration installation, assert the type
4. If `field == “externalIssue”`, search for the repo field in `request.GET`
    1. Search issues for the installation with the `query` and the `repo`
    2. Note: there is some variation in this due to the response from the specific integration
5. If `field` == the repo field (usually `repo`, but Gitlab uses `project`)
    1. Search repos for the installation with the `query`
    2. Note: there is some variation in this due to the response from the specific integration

### DIFF
--- a/src/sentry/integrations/bitbucket/search.py
+++ b/src/sentry/integrations/bitbucket/search.py
@@ -6,7 +6,6 @@ from sentry.api.api_owners import ApiOwner
 from sentry.api.api_publish_status import ApiPublishStatus
 from sentry.api.base import control_silo_endpoint
 from sentry.integrations.bitbucket.integration import BitbucketIntegration
-from sentry.integrations.mixins.issues import IssueBasicIntegration
 from sentry.integrations.models.integration import Integration
 from sentry.integrations.source_code_management.search import SourceCodeSearchEndpoint
 from sentry.shared_integrations.exceptions import ApiError
@@ -33,12 +32,12 @@ class BitbucketSearchEndpoint(SourceCodeSearchEndpoint):
     def installation_class(self):
         return BitbucketIntegration
 
-    def handle_search_issues(
-        self, installation: IssueBasicIntegration, query: str, repo: str
-    ) -> Response:
+    def handle_search_issues(self, installation, query: str, repo: str) -> Response:
+        assert isinstance(installation, self.installation_class)
+
         full_query = f'title~"{query}"'
         try:
-            resp = installation.search_issues(query=full_query, repo=repo)
+            response = installation.search_issues(query=full_query, repo=repo)
         except ApiError as e:
             if "no issue tracker" in str(e):
                 logger.info(
@@ -49,10 +48,12 @@ class BitbucketSearchEndpoint(SourceCodeSearchEndpoint):
                     {"detail": "Bitbucket Repository has no issue tracker."}, status=400
                 )
             raise
+
+        assert isinstance(response, dict)
         return Response(
             [
                 {"label": "#{} {}".format(i["id"], i["title"]), "value": i["id"]}
-                for i in resp.get("values", [])
+                for i in response.get("values", [])
             ]
         )
 
@@ -60,5 +61,6 @@ class BitbucketSearchEndpoint(SourceCodeSearchEndpoint):
     def handle_search_repositories(
         self, integration: Integration, installation, query: str
     ) -> Response:
+        assert isinstance(installation, self.installation_class)
         result = installation.get_repositories(query)
         return Response([{"label": i["name"], "value": i["name"]} for i in result])
--- a/src/sentry/integrations/bitbucket/search.py
+++ b/src/sentry/integrations/bitbucket/search.py
@@ -1,4 +1,5 @@
 import logging
+from typing import TypeVar
 
 from rest_framework.response import Response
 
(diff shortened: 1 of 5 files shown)


src/sentry/integrations/source_code_management/search.py
```diff
@@ -0,0 +1,110 @@
...
+        self,
+    ) -> (
+        type[IssueBasicIntegration | RepositoryIntegration]
+        | tuple[type[IssueBasicIntegration | RepositoryIntegration]]
+    ):
+        raise NotImplementedError
+
+    @abstractmethod
+    def handle_search_issues(
+        self, installation: IntegrationInstallation, query: str, repo: str
```
cathteng, line 59: i would love to be able to type `installation` as type `self.installation_class`, but i haven't been able to successfully do so

src/sentry/integrations/gitlab/search.py
```diff
@@ -1,82 +1,64 @@
...
-        if field == "externalIssue":
-            project = request.GET.get("project")
-            if project is None:
-                return Response({"detail": "project is a required parameter"}, status=400)
-            try:
-                iids = [int(query)]
-                query = None
-            except ValueError:
-                iids = None
+    def handle_search_issues(self, installation, query: str, repo: str) -> Response:
```
cathteng, line 24: for example, installation here would ideally be GitlabIntegration (or at least IssueBasicIntegration) so that the IDE can recognize the return type of `installation.search_issues`...

src/sentry/integrations/source_code_management/search.py
RyanSkonnord: Good thinking. Something roughly like this should work:

```python
I = TypeVar("I", bound=IntegrationInstallation)

class SourceCodeSearchEndpoint(IntegrationEndpoint, Generic[I], ABC):
    @property
    @abstractmethod
    def installation_class(self) -> type[I]: ...

    @abstractmethod
    def handle_search_issues(self, installation:I, query: str, repo: str): ...
```

src/sentry/integrations/github/search.py
```diff
@@ -1,84 +1,67 @@
...
 
-        field = request.GET.get("field")
-        query = request.GET.get("query")
-        if field is None:
-            return Response({"detail": "field is a required parameter"}, status=400)
-        if not query:
-            return Response({"detail": "query is a required parameter"}, status=400)
+    @property
+    def installation_class(self):
+        return (GitHubIntegration, GitHubEnterpriseIntegration)
```
RyanSkonnord, line 25: It would simplify things to remove the tuple return type from the parent class. We could do that there were a base class that `GitHubIntegration` and `GitHubEnterpriseIntegration` shared (and only them). Could you create a small one, just for this purpose? (Maybe "GenericGitHubIntegration", unless you want to rename "GitHubIntegration" and use that as the name for the base class.)

RyanSkonnord:
Looks good so far! I think your instincts are correct that parametrizing the `IntegrationInstallation` type would help and would encourage you to make a pass. Feel free to reach out for help if there's trouble with that. Otherwise I think this looks good to go.

src/sentry/integrations/source_code_management/search.py
cathteng: that did it! thank you

src/sentry/integrations/source_code_management/search.py
```diff
@@ -0,0 +1,103 @@
...
+        if not serializer.is_valid():
+            return self.respond(serializer.errors, status=400)
+
+        field = serializer.validated_data["field"]
+        query = serializer.validated_data["query"]
+
+        installation = integration.get_installation(organization.id)
+        if not isinstance(installation, self.installation_class):
+            integration_provider = (
+                self.integration_provider if self.integration_provider else "github"
```
GabeVillalobos, line 88: How could this scenario happen? Seems weird that we'd want to default this to github in the base endpoint class.

src/sentry/integrations/source_code_management/search.py
```diff
@@ -0,0 +1,103 @@
...
+            if repo is None:
+                return Response(
+                    {"detail": f"{self.repository_field} is a required parameter"}, status=400
+                )
+            return self.handle_search_issues(installation, query, repo)
+
+        if self.repository_field and field == self.repository_field:
+            return self.handle_search_repositories(integration, installation, query)
+
+        return Response(status=400)
```
GabeVillalobos, line 103: It might be good to respond with some additional info here about `field` being invalid.

src/sentry/integrations/source_code_management/search.py
cathteng: ah i have self.installation_class as an abstract property anyway, so it must be defined. let me remove this

cathteng: i think i wanted `self.integration_provider` to be defined to raise the `NotFound`, but it would be better to error that the abstract property doesn't exist before doing that. thanks!

src/sentry/integrations/bitbucket/search.py
```diff
@@ -1,73 +1,67 @@
...
+        assert isinstance(response, dict)
+        return Response(
+            [
+                {"label": "#{} {}".format(i["id"], i["title"]), "value": i["id"]}
+                for i in response.get("values", [])
+            ]
+        )
 
-        return Response(status=400)
+    # TODO: somehow type installation with installation_class
```
GabeVillalobos, line 62: Do you still need this comment, or did the generic change cover it?

GabeVillalobos (approved):
Looks good!

src/sentry/integrations/bitbucket/search.py
cathteng: oopsie

PR #76635, opened 2024-08-27
fix(custom-views): Stop tab width from animating when selected tab changes

malwilley:
While working on overflowing tabs, discovered this setting:

>  If `layout` is set to `"position"`, the size of the component will change instantly and  only its position will animate. If `layout` is set to `"size"`, the position of the component will change instantly but its size will animate.

### DIFF
--- a/static/app/components/draggableTabs/draggableTabList.tsx
+++ b/static/app/components/draggableTabs/draggableTabList.tsx
@@ -106,7 +106,7 @@ function BaseDraggableTabList({
                 dragConstraints={tabListRef} // Sets the container that the tabs can be dragged within
                 dragElastic={0} // Prevents tabs from being dragged outside of the tab bar
                 dragTransition={{bounceStiffness: 400, bounceDamping: 40}} // Recovers spring behavior thats lost when using dragElastic
-                layout
+                layout="position"
               >
                 <Tab
                   key={item.key}
@@ -119,7 +119,7 @@ function BaseDraggableTabList({
                 />
               </Reorder.Item>
               <TabDivider
-                layout
+                layout="position"
                 isVisible={
                   state.selectedKey === 'temporary-tab' ||
                   (state.selectedKey !== item.key &&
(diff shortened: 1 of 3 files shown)


MichaelSun48:
Looks good! Currently just trying to solve this weird z-index behavior where the title of an adjacent frame renders over a tab in the middle of an animation. For some reason this doesn't happen with the first tab though?? 

https://github.com/user-attachments/assets/95493b95-8d8e-4067-9611-ea8ddab11231

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -98,28 +98,33 @@ function BaseDraggableTabList({
...
-                value={item}
-                style={{display: 'flex', flexDirection: 'row'}}
-                as="div"
-                dragConstraints={tabListRef} // Sets the container that the tabs can be dragged within
-                dragElastic={0} // Prevents tabs from being dragged outside of the tab bar
-                dragTransition={{bounceStiffness: 400, bounceDamping: 40}} // Recovers spring behavior thats lost when using dragElastic
-                layout="position"
-              >
-                <Tab
+              <TabItemWrap isSelected={state.selectedKey === item.key}>
```
MichaelSun48, line 101: Ideally we could just apply the z-index styles directly to the `<Reorder.Item>`, but doing this results in the z-index set to `unset`, which fails to solve the bug. This is intentional, and is known to cause this bug: https://github.com/framer/motion/pull/1367

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -98,28 +98,33 @@ function BaseDraggableTabList({
...
+                    orientation={orientation}
+                    overflowing={false}
+                    ref={element => (tabItemsRef.current[item.key] = element)}
+                    variant={tabVariant}
+                  />
+                </Reorder.Item>
+              </TabItemWrap>
               <TabDivider
-                layout="position"
+                layout
```
MichaelSun48, line 127: Before:

https://github.com/user-attachments/assets/0e4d0a87-ad0e-4d9d-8281-44254f903c1e

After

https://github.com/user-attachments/assets/41434fb6-1fbf-4f90-be54-2f9de2f7c80e

Super nit picky, but I personally like the second one more. The layout=position is less necessary for the tab divider since it's size remains the same.

MichaelSun48 (approved):
@malwilley mind taking a look at that commit I pushed? Otherwise looks good!

PR #76637, opened 2024-08-27
feat(queries): filter charts with command/table dropdowns

mjq:
Currently, the Command and Table dropdowns only filter the query table. It would be more useful if they filtered the throughput and duration charts too.

Move the filter dropdowns up above the charts and then filter the charts by their selected values.

Screenshot 2024-08-27 at 4 53 18 PM

**Note**: Will conflict with #75942, so that one should go in first.


0Calories (approved):
looks good!

mjq:
Screenshot 2024-08-29 at 2 03 26 PM

Here it is with https://github.com/getsentry/sentry/pull/75942 integrated.

PR #76643, opened 2024-08-27
fix(anomaly detection): surface validation error to user on trigger create/update

mifu67:
Missed a spot. When creating/updating triggers, catch form validation errors and surface them as serializer validation errors.

### DIFF
--- a/tests/sentry/incidents/endpoints/test_serializers.py
+++ b/tests/sentry/incidents/endpoints/test_serializers.py
@@ -21,6 +21,8 @@ from sentry.incidents.logic import (
 from sentry.incidents.models.alert_rule import (
     AlertRule,
     AlertRuleDetectionType,
+    AlertRuleSeasonality,
+    AlertRuleSensitivity,
     AlertRuleThresholdType,
     AlertRuleTriggerAction,
 )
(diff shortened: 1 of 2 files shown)


nhsiehgit (approved):
Can we add a test?

mifu67:
> Can we add a test?

@nhsiehgit there should already be a test but I don't know where it is, so I'm waiting for it to fail (lol). If nothing fails, then I'll write new tests.

tests/sentry/incidents/endpoints/test_serializers.py
```diff
@@ -464,6 +467,22 @@ def test_boundary_off_by_one(self):
...
+        """
+        params = self.valid_params.copy()
+        params["detection_type"] = AlertRuleDetectionType.DYNAMIC
+        params["seasonality"] = AlertRuleSeasonality.AUTO
+        params["sensitivity"] = AlertRuleSensitivity.MEDIUM
+        params["time_window"] = 15
+        serializer = AlertRuleSerializer(context=self.context, data=params)
+        assert serializer.is_valid()
+
+        with pytest.raises(serializers.ValidationError):
```
ceorourke, line 483: Can you add the error message like another test in this file 

```
            with pytest.raises(
                serializers.ValidationError,
                match="Percentage-based alerts require a comparison delta",
            ):
                serializer.save()
```

ceorourke (approved):

PR #76644, opened 2024-08-27
WIP: get anomalies proxy api

nhsiehgit:
proxy api for us to fetch alert rule anomalies

ticket: https://getsentry.atlassian.net/browse/ALRT-238

### DIFF
--- a/src/sentry/incidents/endpoints/organization_alert_rule_anomalies.py
+++ b/src/sentry/incidents/endpoints/organization_alert_rule_anomalies.py
@@ -1,0 +1,94 @@
+from datetime import datetime
+
+import orjson
+import requests
+from django.conf import settings
+from drf_spectacular.utils import extend_schema
+from pydantic import BaseModel
+from rest_framework.request import Request
+from rest_framework.response import Response
+
+from sentry.api.api_owners import ApiOwner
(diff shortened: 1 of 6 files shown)


src/sentry/seer/anomaly_detection/detect_anomalies.py
```diff
@@ -0,0 +1,101 @@
+import logging
```
nhsiehgit, line 1: this is the _start_ of pulling the `get_anomaly_data_from_seer` out into a separate helper that we can utilize in both the subscription_processor as well as the proxy api.

this is incomplete atm

src/sentry/incidents/endpoints/organization_alert_rule_anomalies.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@region_silo_endpoint
+class OrganizationAlertRuleAnomaliesEndpoint(OrganizationAlertRuleEndpoint):
+    owner = ApiOwner.ISSUES
+    publish_status = {
+        "GET": ApiPublishStatus.PUBLIC,
+    }
+
+    def _call_seer(
```
nhsiehgit, line 37: This should be replaced with the `make_signed_seer_request` method.
can be scrapped

PR #76649, opened 2024-08-28
fix(onboarding): Persist project filter on redirect

ArthurKnaus:
### Problem

Simply setting the query param for project works fine for the immediate redirect. Any further navigation though will reset the selected project.

https://github.com/user-attachments/assets/d1fa4315-5dcc-44fe-855d-032eef3c949e

### Solution

Persist the page filters in local storage and mark them as pinned in the PageFilterStore, so the page filter logic picks them up again.

https://github.com/user-attachments/assets/e080f983-4ff0-4a8c-9f49-48db249b5886

### DIFF
--- a/static/app/views/projectInstall/platform.tsx
+++ b/static/app/views/projectInstall/platform.tsx
@@ -4,13 +4,14 @@ import omit from 'lodash/omit';
 
 import Feature from 'sentry/components/acl/feature';
 import {Alert} from 'sentry/components/alert';
-import {LinkButton} from 'sentry/components/button';
+import {Button} from 'sentry/components/button';
 import ButtonBar from 'sentry/components/buttonBar';
 import NotFound from 'sentry/components/errors/notFound';
 import HookOrDefault from 'sentry/components/hookOrDefault';
 import {SdkDocumentation} from 'sentry/components/onboarding/gettingStartedDoc/sdkDocumentation';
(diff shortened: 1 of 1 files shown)


vgrozdanic (approved):

static/app/views/projectInstall/platform.tsx
```diff
@@ -140,6 +144,28 @@ export function ProjectInstallPlatform({
     });
   }, [organization, currentPlatform, project?.id]);
 
+  const redirectWithProjectSelection = useCallback(
+    to => {
```
priscilawebdev, line 148: should we type this parameter?

static/app/views/projectInstall/platform.tsx
```diff
@@ -140,6 +144,28 @@ export function ProjectInstallPlatform({
...
+      if (!project?.id) {
+        return;
```
priscilawebdev, line 150: is the project's id a string? I am wondering if we should use the function `defined` here

ArthurKnaus: If is a string. If it is an empty string it would also be bad. So this is a bit more defensive than `defined` I would say.

priscilawebdev: in the types, we have `number` too, but if comes as a string it shall be fine

priscilawebdev (approved):
🙌

PR #76662, opened 2024-08-28
ref(tsc): Convert AccountSecurityDetails to FC

ArthurKnaus:
Convert `<AccountSecurityDetails />` to FC.

- part of https://github.com/getsentry/frontend-tsc/issues/2


static/app/views/settings/account/accountSecurity/accountSecurityDetails.tsx
```diff
@@ -43,160 +49,165 @@ function AuthenticatorDate({label, date}: AuthenticatorDateProps) {
...
-        )}
-
-        <RecoveryCodes
-          onRegenerateBackupCodes={onRegenerateBackupCodes}
-          isEnrolled={authenticator.isEnrolled}
-          codes={authenticator.codes}
-        />
-      </Fragment>
-    );
+  if (!authenticator || isError) {
```
priscilawebdev, line 139: shouldn't we check only for `isError` here?

static/app/views/settings/account/accountSecurity/accountSecurityDetails.tsx
```diff
@@ -43,160 +49,165 @@ function AuthenticatorDate({label, date}: AuthenticatorDateProps) {
...
+    return <LoadingError onRetry={refetch} />;
   }
+
+  return (
+    <SentryDocumentTitle title={PAGE_TITLE}>
```
priscilawebdev, line 144: hm why not inline the page title?

static/app/views/settings/account/accountSecurity/accountSecurityDetails.tsx
priscilawebdev: and if an error happens in the mutations?

ArthurKnaus: The previous code handled the response as if the endpoint could return `null`.
As far as I see from the python code that should not be possible, will adjust the types.

priscilawebdev (approved):
🙏

PR #76670, opened 2024-08-28
fix(slack): Render Loading Modal for Slack

iamrajjoshi:
fixes [SENTRY-3BWJ](https://sentry.sentry.io/issues/5590278616/)

the issue is we receive the request in control then send it over to region, which then sends a response to slack. this is a problem b/c we only have 3 seconds to respond.

here, i add a loading modal state, which we send immediately from control, eventually updating the model from region. this method gives us 30 minutes to respond to slack from the initial loading modal render.

i wrote the logic here more extensible, so if we want contractors in the future to implement something similar for the individual options (which are also might be timing out), we can have them do that.

here is a video of the changes (i exaggerated the latency)

https://github.com/user-attachments/assets/80238e35-fd99-4ef0-8c1a-38aa98d0c3ad

### DIFF
--- a/src/sentry/integrations/slack/webhooks/action.py
+++ b/src/sentry/integrations/slack/webhooks/action.py
@@ -393,6 +393,20 @@ class SlackActionEndpoint(Endpoint):
             # If the external_id is not found, Slack we send `not_found` error
             # https://api.slack.com/methods/views.update
             if unpack_slack_api_error(e) == MODAL_NOT_FOUND:
+                metrics.incr(
+                    SLACK_WEBHOOK_GROUP_ACTIONS_FAILURE_DATADOG_METRIC,
+                    sample_rate=1.0,
+                    tags={"type": "update_modal"},
+                )
+                _logger.exception(
+                    "slack.action.update-modal-not-found",
+                    extra={
+                        "organization_id": slack_request.organization.id,
+                        "integration_id": slack_request.integration.id,
+                        "trigger_id": slack_request.data["trigger_id"],
+                        "dialog": "resolve",
+                    },
+                )
                 # The modal was not found, so we need to open a new one
                 self._open_view(slack_client, modal_payload, slack_request)
             else:
--- a/src/sentry/integrations/slack/webhooks/action.py
+++ b/src/sentry/integrations/slack/webhooks/action.py
@@ -377,7 +377,7 @@ class SlackActionEndpoint(Endpoint):
             metadata=metadata,
         )
 
-    def _update_view(
+    def _update_modal(
         self,
         slack_client: SlackSdkClient,
         external_id: str,
@@ -438,7 +438,7 @@ class SlackActionEndpoint(Endpoint):
             external_id = slack_request.data.get("actions")[0].get("action_ts")
 
             if options.get("send-slack-response-from-control-silo"):
-                self._update_view(slack_client, external_id, modal_payload, slack_request)
+                self._update_modal(slack_client, external_id, modal_payload, slack_request)
             else:
                 self._open_view(slack_client, modal_payload, slack_request)
 
@@ -490,7 +490,7 @@ class SlackActionEndpoint(Endpoint):
             external_id = slack_request.data.get("actions")[0].get("action_ts")
 
             if options.get("send-slack-response-from-control-silo"):
-                self._update_view(slack_client, external_id, modal_payload, slack_request)
+                self._update_modal(slack_client, external_id, modal_payload, slack_request)
             else:
                 self._open_view(slack_client, modal_payload, slack_request)
 
--- a/src/sentry/integrations/slack/webhooks/action.py
+++ b/src/sentry/integrations/slack/webhooks/action.py
@@ -15,7 +15,7 @@ from slack_sdk.errors import SlackApiError
 from slack_sdk.models.views import View
 from slack_sdk.webhook import WebhookClient
 
-from sentry import analytics
+from sentry import analytics, options
 from sentry.api import client
 from sentry.api.api_owners import ApiOwner
 from sentry.api.api_publish_status import ApiPublishStatus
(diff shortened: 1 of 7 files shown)


src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -376,6 +377,36 @@ def build_archive_modal_payload(self, callback_id: str, metadata: str) -> View:
             metadata=metadata,
         )
 
+    def _update_view(
```
cathteng, line 380: nit: should we call this `_update_modal`?

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -376,6 +377,36 @@ def build_archive_modal_payload(self, callback_id: str, metadata: str) -> View:
...
+            slack_client.views_update(
+                external_id=external_id,
+                view=modal_payload,
+            )
+        except SlackApiError as e:
+            # If the external_id is not found, Slack we send `not_found` error
+            # https://api.slack.com/methods/views.update
+            if unpack_slack_api_error(e) == MODAL_NOT_FOUND:
+                # The modal was not found, so we need to open a new one
+                self._open_view(slack_client, modal_payload, slack_request)
```
cathteng, line 397: is this something we'll always need to account for?

src/sentry/middleware/integrations/parsers/slack.py
```diff
@@ -82,19 +92,101 @@ class SlackRequestParser(BaseRequestParser):
...
+    def handle_dialog(self, request, action, title):
+        decoded_body = parse_qs(request.body.decode(encoding="utf-8"))
+        if (payload_dict := decoded_body.get("payload")) is None:
+            raise ValueError("request format is incorrect")
+
+        payload = json.loads(payload_dict[0])
+
+        # we need to grab the action_ts to use as the external_id for the loading modal
+        # https://api.slack.com/reference/interaction-payloads/block-actions
+        action_ts = payload.get("actions")[0].get("action_ts")
```
cathteng, line 118: does this HAVE to be the `action_ts`, or are we just using it because it's basically guaranteed to be unique?

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -402,10 +433,15 @@ def open_resolve_dialog(self, slack_request: SlackActionRequest, group: Group) -
         modal_payload = self.build_resolve_modal_payload(callback_id, metadata=metadata)
         slack_client = SlackSdkClient(integration_id=slack_request.integration.id)
         try:
-            slack_client.views_open(
-                trigger_id=slack_request.data["trigger_id"],
-                view=modal_payload,
-            )
+            # We need to use the action_ts as the external_id to update the modal
+            # We passed this in control when we sent the loading modal to beat the 3 second timeout
+            external_id = slack_request.data.get("actions")[0].get("action_ts")
```
cathteng, line 438: i think you might want to set some defaults in case `actions` doesn't exist? is this guaranteed to be in the first entry of the `actions` array?

src/sentry/middleware/integrations/parsers/slack.py
```diff
@@ -82,19 +92,101 @@ class SlackRequestParser(BaseRequestParser):
...
+    def build_loading_modal(self, external_id: str, title: str):
+        return {
+            "type": "modal",
+            "external_id": external_id,
+            "title": {"type": "plain_text", "text": title},
+            "blocks": [
+                {
+                    "type": "section",
+                    "block_id": "loading_block",
+                    "text": {"type": "mrkdwn", "text": "Loading..."},
```
cathteng, line 104: nit: i wonder if it's possible to have a loading indicator, that would be ideal lMO

src/sentry/middleware/integrations/parsers/slack.py
```diff
@@ -82,19 +92,101 @@ class SlackRequestParser(BaseRequestParser):
...
             kwargs={
                 "region_names": [r.name for r in regions],
                 "payload": create_async_request_payload(self.request),
                 "response_url": self.response_url,
             }
         )
-        # We may want to enrich this with a waiting message
-        return HttpResponse(status=status.HTTP_202_ACCEPTED)
+
+        return HttpResponse(status=status.HTTP_200_OK)
```
cathteng, line 189: could we leave a note here as to why we respond with 200 OK even though we already rendered the modal?

cathteng (commented):
just a bunch of questions c:

src/sentry/integrations/slack/webhooks/action.py
RyanSkonnord: +1 to this. I'd like to factor out the `[0]`'s in general, maybe make a helper function for unpacking stuff from the request data. Type checks would also be helpful, so that we get a good exception if something isn't a sequence or mapping when we expect.

When we expect a sequence to have exactly one element, it's better to communicate/assert that with `(thing,) = sequence` rather than `sequence[0]`. If we're expecting multiple elements but only care about the first, we could clarify that as well.

src/sentry/integrations/slack/webhooks/action.py
iamrajjoshi: yes, this will help if the first message we send from control fails (ratelimited for example), this will give us another shot to still successfully send the message.

src/sentry/integrations/slack/webhooks/action.py
iamrajjoshi: i factored out the parsing and adding more type checking and assertions

src/sentry/middleware/integrations/parsers/slack.py
```diff
@@ -82,19 +92,126 @@ class SlackRequestParser(BaseRequestParser):
...
+                action_ts = action.get("action_ts")
+                if action_ts is None:
+                    raise ValueError(
+                        "Error parsing Slack payload: 'action_ts' not found in 'actions'"
+                    )
+                return payload, action_ts
+            else:
+                raise ValueError(
+                    "Error parsing Slack payload: 'actions' not found or not a list of length 1"
+                )
```
GabeVillalobos, line 136: It might be good to differentiate these two cases with separate messages when they occur.

GabeVillalobos (approved):
Pending Cathy's feedback, looks good to me

src/sentry/middleware/integrations/parsers/slack.py
iamrajjoshi: i looked for some built in block, there doesn't seem to be one. what we can do is have a static image that we render 👀

cathteng: nw this is just a nice to have

src/sentry/integrations/slack/webhooks/action.py
cathteng: hmm should we emit a metric from this? is the first message failing indicative of anything

cathteng (approved):

PR #76673, opened 2024-08-28
feat(explore): Adding chart.

Abdkhan14:
Screenshot 2024-08-28 at 5 02 38 PM


static/app/views/explore/charts/index.tsx
```diff
@@ -1,5 +1,134 @@
...
+import {t} from 'sentry/locale';
+import {space} from 'sentry/styles/space';
+import {tooltipFormatter} from 'sentry/utils/discover/charts';
+import {MutableSearch} from 'sentry/utils/tokenizeSearch';
+import usePageFilters from 'sentry/utils/usePageFilters';
+import Chart, {ChartType} from 'sentry/views/insights/common/components/chart';
+import ChartPanel from 'sentry/views/insights/common/components/chartPanel';
+import {useSpanIndexedSeries} from 'sentry/views/insights/common/queries/useDiscoverSeries';
+import {CHART_HEIGHT} from 'sentry/views/insights/database/settings';
+import {useMetricsIntervalParam} from 'sentry/views/metrics/utils/useMetricsIntervalParam';
```
Zylphrex, line 16: We may want to write our own version of this as this implementation seems to make a few assumptions around the mri/widgets on the metrics page that do not apply here.

static/app/views/explore/charts/index.tsx
```diff
@@ -1,5 +1,134 @@
...
+];
+
+// TODO: Update to support aggregate mode and multiple queries / visualizations
+export function ExploreCharts({query}: ExploreChartsProps) {
+  const [visualize] = useVisualize();
+  const [chartType, setChartType] = useChartType();
+  const {interval, setInterval, currentIntervalOptions} = useMetricsIntervalParam();
+
+  const pageFilters = usePageFilters();
+  const period = pageFilters.selection.datetime.period;
```
Zylphrex, line 48: This breaks when the user selects an absolute date range. If possible, try to use the `datetime` object directly.

static/app/views/explore/charts/index.tsx
Abdkhan14: Hey yup, I'll get it done @Zylphrex

static/app/views/explore/charts/index.tsx
Abdkhan14: Removed it

static/app/views/explore/charts/index.tsx
```diff
@@ -1,5 +1,126 @@
...
+  const pageFilters = usePageFilters();
+  const [visualize] = useVisualize();
+  const [chartType, setChartType] = useChartType();
+  const [interval, setInterval, intervalOptions] = useChartInterval();
+
+  const series = useSpanIndexedSeries(
+    {
+      search: new MutableSearch(query ?? ''),
+      yAxis: [visualize],
+      interval: interval ?? getInterval(pageFilters.selection.datetime, 'metrics'),
```
Zylphrex, line 50: Is this using the right signature for `getInterval`? It takes a fidelity as a second parameter but you're passing `'metrics'`

Abdkhan14: Heyo @Zylphrex, I took this default from the [tracesChart.tsx](https://github.com/getsentry/sentry/blob/master/static/app/views/traces/tracesChart.tsx#L155). "metrics" is a valid option.

Zylphrex (approved):

PR #76675, opened 2024-08-28
chore(apitoken): remove options

mdtro:
Removing options that are no longer needed. These were used to get us to an improved state where tokens are hashed in the database.


src/sentry/api/authentication.py
```diff
@@ -349,18 +337,11 @@ def _find_or_update_token_by_hash(self, token_str: str) -> ApiToken | ApiTokenRe
 
         hashed_token = hashlib.sha256(token_str.encode()).hexdigest()
 
-        rate = options.get("apitoken.use-and-update-hash-rate")
```
sentaur-athena, line 352: What is rate in this context?

src/sentry/models/apitoken.py
```diff
@@ -207,38 +207,34 @@ def plaintext_refresh_token(self) -> str:
...
+        else:
+            # The backup tests create a token with a refresh_token and then clear it out.
+            # So if the refresh_token is None, wipe out any hashed value that may exist too.
+            # https://github.com/getsentry/sentry/blob/1fc699564e79c62bff6cc3c168a49bfceadcac52/tests/sentry/backup/test_imports.py#L1306
+            self.hashed_refresh_token = None
 
-        if options.get("apitoken.auto-add-last-chars"):
-            token_last_characters = self.token[-4:]
-            self.token_last_characters = token_last_characters
+        token_last_characters = self.token[-4:]
```
sentaur-athena, line 220: What is the significance of 4?

sentaur-athena (approved):
The PR looks good. I use the opportunity to learn more about auth 😅

src/sentry/models/apitoken.py
mdtro: This grabs the last four characters of the token and sets them for the `token_last_characters` column. 
They are then used in the UI to help users identify their tokens. 

image

src/sentry/api/authentication.py
mdtro: This was an option I introduced to update old tokens as they were used for authentication with their hash value. The value was between 0.0 and 1.0. 

The hash updating logic was applied to a small percentage of traffic at first and slowly moved to 1.0 to apply to 100% of the traffic. See [these PRs where the option was updated in steps.](https://github.com/getsentry/sentry-options-automator/pulls?q=is%3Apr+author%3Amdtro+is%3Aclosed+%22rollout+hashed+token+auth%22)

By removing this, we are effectively setting it to 100%. In production, this will have no impact as all tokens have hash values, but I'll keep it in for one release cycle for self-hosted users.

After some time, I created a [migration that backfills all API tokens](https://github.com/getsentry/sentry/pull/71728) with their hash values that was released in 24.6.0. This covered any tokens that were not used for authentication. The number of tokens we had to perform this backfill operation on was significantly less since we "updated them on use" over time with the option above.

All that to say, is that this rate option was used to avoid unnecessary load on the database and to avoid introducing latency in the hybrid cloud replication services. :)

Jeffreyhung (approved):

PR #76680, opened 2024-08-28
chore(tests): Move tests/snuba/api/endpoint tests to tests/sentry

armenzg:
Without this change, it could sometimes be hard to find the tests since they would not match the mental mapping of `src/sentry` to `tests/sentry`.


tests/sentry/api/endpoints/test_group_details.py
```diff
@@ -788,3 +793,304 @@ def test_collapse_release(self):
         assert response.data["firstRelease"] is None
         response = self.client.get(url, {"collapse": ["release"]})
         assert "firstRelease" not in response.data
+
+
+class GroupDetailsSnubaTest(APITestCase, SnubaTestCase):
```
armenzg, line 798: I've moved this from `tests/snuba/.../test_group_details.py`

tests/sentry/api/endpoints/test_group_event_details.py
```diff
@@ -389,3 +389,118 @@ def test_query_issue_platform_title(self):
         assert response.data["id"] == str(occurrence.event_id)
         assert response.data["previousEventID"] is None
         assert response.data["nextEventID"] is None
+
+
+class GroupEventDetailsTest(APITestCase, SnubaTestCase, PerformanceIssueTestCase):
```
armenzg, line 394: I've moved this from `tests/snuba/.../test_group_event_details.py`

shellmayr:
@armenzg any specific reason why this was not done after all?

asottile-sentry:
my guess is that sns intentionally segregated these tests so make it easier to check integration in the snuba repository 🤷

PR #76685, opened 2024-08-28
style(settings): align checkbox with org name

michellewzhang:
before
SCR-20240828-kbvc

https://github.com/user-attachments/assets/354385f0-5e56-45e7-a9d5-a44bd9473a13

after 

https://github.com/user-attachments/assets/08ade250-6130-4c39-a956-85dac9fd6a88

### DIFF
--- a/static/app/views/settings/account/accountClose.tsx
+++ b/static/app/views/settings/account/accountClose.tsx
@@ -7,6 +7,7 @@ import {openModal} from 'sentry/actionCreators/modal';
 import {fetchOrganizations} from 'sentry/actionCreators/organizations';
 import {Alert} from 'sentry/components/alert';
 import {LinkButton} from 'sentry/components/button';
+import Checkbox from 'sentry/components/checkbox';
 import HookOrDefault from 'sentry/components/hookOrDefault';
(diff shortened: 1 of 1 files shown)


evanpurkhiser (approved):
Do we have a more appropriate component we can use here even?

michellewzhang:
> Do we have a more appropriate component we can use here even?

hm we might be able to use the `Checkbox` component, let me look into it

static/app/views/settings/account/accountClose.tsx
```diff
@@ -165,18 +166,20 @@ function AccountClose() {
...
+                  alignItems: 'center',
+                }}
+                name="organizations"
+                checked={orgsToRemove.has(organization.slug)}
+                disabled={singleOwner}
+                value={organization.slug}
+                onChange={evt => handleChange(organization, singleOwner, evt)}
+                size="sm"
+              />
+              {organization.slug}
```
scttcper, line 182: i think this would mean that you can no longer click on the org slug to check the box? Wrapping in a label element might fix this

michellewzhang: ah good catch, just fixed

scttcper (approved):

PR #76686, opened 2024-08-28
style(settings): add gap to identity tags

michellewzhang:
before
SCR-20240828-keuh

after 
SCR-20240828-kerz

### DIFF
--- a/static/app/views/settings/account/accountIdentities.tsx
+++ b/static/app/views/settings/account/accountIdentities.tsx
@@ -64,7 +64,7 @@ function IdentityItem({identity, onDisconnect}: IdentityItemProps) {
         </IdentityText>
       </InternalContainer>
       <InternalContainer>
-        <TagWrapper>
+        <TagWrapper css={{gap: space(0.75)}}>
           {identity.category === UserIdentityCategory.SOCIAL_IDENTITY && (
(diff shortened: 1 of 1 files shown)


static/app/views/settings/account/accountIdentities.tsx
```diff
@@ -64,7 +64,7 @@ function IdentityItem({identity, onDisconnect}: IdentityItemProps) {
         </IdentityText>
       </InternalContainer>
       <InternalContainer>
-        <TagWrapper>
+        <TagWrapper css={{gap: space(0.75)}}>
```
evanpurkhiser, line 67: Why not add the property to the `TagWrapper` styled component?

evanpurkhiser:
@michellewzhang small nit with your PR descriptions.

I would actually just prefer to have the before and after as two paragraphs. so the images expand to the full width of the comment. Makes it easier to see the difference without having to click them

static/app/views/settings/account/accountIdentities.tsx
michellewzhang: oops i totally missed that it's a styled component, just moved it

priscilawebdev (approved):
lgtm

PR #76691, opened 2024-08-28
fix(eventstore): apply date filters in get_events nodestore case

aliu39:
Fixes [SENTRY-163P](https://sentry.sentry.io/issues/4502227101/)

Today this celery task [update_user_reports.py](https://github.com/getsentry/sentry/blob/65f94afa7b7d860bc54a3b703484d8b286a66722/src/sentry/tasks/update_user_reports.py#L62-L62) has been querying eventstore with old event IDs, outside retention range (dated 7/17). We expected the time range passed in to apply, but it's not for this branch (nodestore optimization).


src/sentry/eventstore/snuba/backend.py
```diff
@@ -249,6 +249,14 @@ def __get_events(
                 start = min(event.datetime for event in nodestore_events)
                 end = max(event.datetime for event in nodestore_events) + timedelta(seconds=1)
 
+                if filter.start or filter.end:
+                    if filter.start:
+                        start = max(start, filter.start)
+                    if filter.end:
+                        end = min(end, filter.end)
+                    if start > end:
+                        return []
```
onkar, line 258: These cascading `if`s and the reliance of their ordering is error prone. This can probably be written like:
```
start = max(start, filter.start) if filter.start else start
end = min(end, filter.end) if filter.end else end
if start > end:
    return []
```

tests/sentry/tasks/test_update_user_reports.py
```diff
@@ -37,29 +38,35 @@ def test_simple(self):
         assert report3.group_id is None
         assert report3.environment_id is None
 
-    def test_start_end(self):
+    def test_start_end_reports(self):
```
onkar, line 41: I see that `test_start_end_reports()` and `test_start_end_events()` have a lot of common and repetitive code. Can that common code be moved to a function?

onkar (changes requested):

src/sentry/eventstore/snuba/backend.py
aliu39: Got it, the reason I wrote it like this is because I wanted to preserve the behavior of the `start > end` case further downstream. In case users of this function are expecting specific errors to be thrown. I.e. I only want the early termination to happen in this special case

onewland, line 258: Based on the linked Sentry issue, it seems like you're querying outside the range of 90 days. Would more bulletproof logic be setting start to `max(90 days ago, start)` and finish to `max(90 days ago, finish)`? I don't know who or what sets filter in this case.

onewland (commented):
I'm not opposed to this change, but I'll just note that in general that Search and Storage  doesn't need to review your queries. Feel free to tag us in, but we'll generally be more helpful if you have a specific performance problem or you're getting an error or data in your results that is unexpected.

src/sentry/eventstore/snuba/backend.py
aliu39: `filter` is provided by callers of `get_events`.

aliu39: `filter` is provided by callers of `get_events`. I don't think we need the 90 days cap because in the case the time range is out of retention, we want to raise an error to alert users. The problem here is the time range in `filter` isn't respected, which is what our team was expecting when making this query.

aliu39: > I'm not opposed to this change, but I'll just note that in general that Search and Storage doesn't need to review your queries.

Ok, thanks for letting me know! By "queries" do you generally mean sentry code that uses snuba sdk? I requested owners-snuba because I wasn't sure who owns this eventstore code, and my team is not familiar with it. Is there a better GH group to request next time?

aliu39: Rewrote it so these conditions and early term are no longer necessary!

onkar (approved):
LGTM, thanks for addressing the comments.

PR #76697, opened 2024-08-28
fix(feedback): catch snuba errors in update_user_reports task

aliu39:
Related to, but doesn't fix [SENTRY-163P](https://sentry.sentry.io/issues/4502227101/)

If one query fails in this loop, the whole task fails. We're backlogged to ~16k reports today because of the error in the issue above. https://cloudlogging.app.goo.gl/RSiCipKDsuSRWo2m6

~~Depends on https://github.com/getsentry/sentry/pull/76718~~

### DIFF
--- a/src/sentry/tasks/update_user_reports.py
+++ b/src/sentry/tasks/update_user_reports.py
@@ -65,13 +65,12 @@ def update_user_reports(**kwargs: Any) -> None:
                     filter=snuba_filter, referrer="tasks.update_user_reports"
                 )
                 events.extend(events_chunk)
-            except Exception as e:
+            except Exception:
+                sentry_sdk.set_tag("update_user_reports.eventstore_query_failed", True)
                 logger.exception(
                     "update_user_reports.eventstore_query_failed",
                     extra={"project_id": project_id, "start": start, "end": end},
-                )
-                sentry_sdk.set_tag("update_user_reports.eventstore_query_failed", True)
-                sentry_sdk.capture_exception(e)
+                )  # will also send exc to Sentry
 
         for event in events:
             report = report_by_event.get(event.event_id)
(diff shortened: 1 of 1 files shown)


src/sentry/tasks/update_user_reports.py
```diff
@@ -59,10 +60,18 @@ def update_user_reports(**kwargs: Any) -> None:
...
+                events_chunk = eventstore.backend.get_events(
+                    filter=snuba_filter, referrer="tasks.update_user_reports"
+                )
+                events.extend(events_chunk)
+            except Exception as e:
+                logger.exception(
+                    "update_user_reports.eventstore_query_failed",
+                    extra={"project_id": project_id, "start": start, "end": end},
+                )
+                sentry_sdk.set_tag("update_user_reports.eventstore_query_failed", True)
```
cmanallen, line 73: Move this above logger.exception.

cmanallen (changes requested):

cmanallen:
Let's add a limit that is some multiple of the ingest rate over a 15 minute period.

aliu39:
> Let's add a limit that is some multiple of the ingest rate over a 15 minute period.

I don't think we should limit this query. Aside from the backlog from this issue, we usually don't expect many reports with missing events to make it past post process. The query is already limited by time range (last 24hr).

For clearing the backlog: https://sentry.slack.com/archives/CUHS29QJ0/p1724881027856289
Issue platform is totally fine to handle a burst of ~20k, which is worst case. It's possible tomorrow morning this backlog will be smaller too.
The problem is the rate limit of 1k/hour, per project. We can temporarily raise that by a generous amount. Under normal conditions, I think it's ok for this task to get rate limited.

cmanallen (approved):

aliu39:
This actually isn't a backlog.. https://cloudlogging.app.goo.gl/CepGv7zXk2jKpCieA is from before the error and we always process roughly this amount (16k) of reports. My bad for the mis-diagnosis

PR #76699, opened 2024-08-28
chore (anomaly detection): hardcode seasonality in serializer

mifu67:
Seasonality is not a supported field for EA, so we should hardcode it (but leave the parameter for future use)

### DIFF
--- a/tests/sentry/incidents/endpoints/test_organization_alert_rule_details.py
+++ b/tests/sentry/incidents/endpoints/test_organization_alert_rule_details.py
@@ -243,7 +243,7 @@ class AlertRuleDetailsGetEndpointTest(AlertRuleDetailsBase):
 
         with pytest.raises(
             ValidationError,
-            match="Sensitivity and seasonality are not valid fields for this alert type",
+            match="Sensitivity is not a valid field for this alert type",
         ):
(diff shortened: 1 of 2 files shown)


tests/sentry/incidents/endpoints/test_organization_alert_rule_details.py
```diff
@@ -317,27 +317,24 @@ def test_dynamic_detection_type(self, mock_seer_request):
...
-            self.create_alert_rule(
-                sensitivity=AlertRuleSensitivity.MEDIUM,
-                detection_type=AlertRuleDetectionType.DYNAMIC,
-                time_window=30,
-            )  # Require both seasonality and sensitivity
-
-        with pytest.raises(
-            ValidationError, match="Dynamic alerts require both sensitivity and seasonality"
-        ):
+        # NOTE: uncomment this test when seasonality becomes a supported field
```
nhsiehgit, line 327: nit: s/NOTE/TODO

nhsiehgit (approved):

tests/sentry/incidents/endpoints/test_organization_alert_rule_details.py
ceorourke: You can keep it commented in but add a `pytest.skip` decorator

PR #76701, opened 2024-08-28
feat(issues): Add ability to query issues by hash 

### LINKED ISSUE #72753: Create API to query issues by project and `GroupHash.hash` value
### Problem Statement

Enable querying issues by fingerprint.

We currently have a public API endpoint which returns a [list of fingerprints for an issue id](https://docs.sentry.io/api/events/list-an-issues-hashes/) but we don't have any way of retrieving the inverse (issue id for a fingerprint).

### Solution Brainstorm

Add a new (experimental, for now) API endpoint which returns issue details for a fingerprint (grouphash).

### Product Area

Issues

Notes: 

- This will need to be a project endpoint, since hash values aren't unique across projects/orgs, but are unique within a project.

- The initial usecase for this is to be able to copy issue data across regions (as an org is migrating to the EU datastore, for example). Because they live in separate databases, the old and new copies of a given issue won't necessarily have the same group id, but they will have the same hash. This endpoint is therefore a way to find the new group id. Given that:
    - Since we'll need to do this for all of a project's issues, it'd be nice to be able to send a bunch of hash values at once and get back groups for all of them.
    - We might consider whether we want to have a mode where all it returns is a dictionary of hash value to group id, rather than hash value to full group.
    - Another option would be to just add the ability to pass a hash (or, better, a set of hashes) to the [project-issues endpoint](https://docs.sentry.io/api/events/list-a-projects-issues/).

- There are more places than you'd immediately think of where our endpoints are listed (and where any new endpoint would need to be added). See https://github.com/getsentry/sentry/pull/72552 for an example.

roggenkemper:
this pr adds the ability to query by hash in the `project_group_index` endpoint. If a user passes in any number of hashes, the endpoint will only look for those hashes and return matching groups it finds for those hashes. 

closes https://github.com/getsentry/sentry/issues/72753


src/sentry/api/endpoints/project_group_index.py
```diff
@@ -99,6 +101,20 @@ def get(self, request: Request, project) -> Response:
             stats_period=stats_period,
         )
 
+        hashes = request.GET.getlist("hash", [])
+        if hashes:
```
mrduncan, line 105: Do we want to limit how many hashes we support here?

src/sentry/api/endpoints/project_group_index.py
```diff
@@ -77,6 +78,7 @@ def get(self, request: Request, project) -> Response:
                                    ``"is:unresolved"`` is assumed.)
         :qparam string environment: this restricts the issues to ones containing
                                     events from this environment
+        :qparam list hash: hashes of groups to return, overrides every other query parameter, only returning list of groups found from hashes
```
mrduncan, line 81: In this case, should we return an error if a user tries to use this in combination with any other params just to make sure they don't accidentally hit this without realizing it?

src/sentry/api/endpoints/project_group_index.py
roggenkemper: yeah, we can add a limit in. will be easy to change in the future if we ever want to increase/decrease the limit

api-docs/paths/events/project-issues.json
```diff
@@ -46,6 +46,14 @@
           "type": "string"
         }
       },
+      {
+        "name": "hash",
+        "in": "query",
+        "description": "A list of hashes of groups to return. Is not compatible with 'query' parameter as it will only return the request groups with corresponding hashes. The maximum number of hashes that can be sent is 100. If more are sent, only the first 100 will be used.",
```
lobsterkatie, line 52:
```suggestion
        "description": "A list of hashes of groups to return. Is not compatible with 'query' parameter. The maximum number of hashes that can be sent is 100. If more are sent, only the first 100 will be used.",
```

Given that it just errors out if you pass both, I think it's enough to say they're incompatible. My worry would be that saying "it only returns the groups that match the hashes" might lead people to think that we mean "of the groups the query would give you back, it only returns the groups that match the hashes," which is obv not true.

Might also be worth mentioning that a given group can have multiple hashes, so the number of groups returned may be smaller than the number of hashes sent (if two or more of the requested hashes point to the same group).

src/sentry/api/endpoints/project_group_index.py
```diff
@@ -99,7 +102,27 @@ def get(self, request: Request, project) -> Response:
...
+        if hashes:
+            if query:
+                return Response({"detail": ERR_HASHES_AND_OTHER_QUERY}, status=400)
+
+            # limit to 100 hashes
+            hashes = hashes[:100]
+            groups_from_hashes = GroupHash.objects.filter(hash__in=hashes).values_list(
+                "group_id", flat=True
+            )
+            groups = list(Group.objects.filter(id__in=groups_from_hashes, project_id=project.id))
```
lobsterkatie, line 117: Is there a reason to not filter on project in the original query? I'm not amazing with the Django ORM, but I think that might let you do it all in one query.

tests/snuba/api/endpoints/test_project_group_index.py
```diff
@@ -352,6 +352,36 @@ def test_filter_not_unresolved(self):
         assert response.status_code == 200
         assert [int(r["id"]) for r in response.data] == [event.group.id]
 
+    def test_group_by_hash(self):
```
lobsterkatie, line 355: Nit:

```suggestion
    def test_single_group_by_hash(self):
```

I'd do `test_multiple_groups_by_hashes` below, too - just so it doesn't sound like "group" is a verb and like you're testing it doing grouping.

tests/snuba/api/endpoints/test_project_group_index.py
```diff
@@ -352,6 +352,36 @@ def test_filter_not_unresolved(self):
...
+        event = self.store_event(
+            data={"timestamp": iso_format(before_now(seconds=500)), "fingerprint": ["group-1"]},
+            project_id=self.project.id,
+        )
+
+        self.login_as(user=self.user)
+
+        response = self.client.get(f"{self.path}?hash={event.get_primary_hash()}")
+        assert response.status_code == 200
+        assert len(response.data) == 1
```
lobsterkatie, line 365: Maybe we should test that it sends back the correct group here also. Don't have to assert on the whole serialized response, but you could just grab the id and test against that. WDYT?

lobsterkatie (approved):
A few small comments, but overall LGTM!

src/sentry/api/endpoints/project_group_index.py
roggenkemper: `project` can be null on GroupHash, but not on Group I could only add the project filter on the groups after we got them

tests/snuba/api/endpoints/test_project_group_index.py
roggenkemper: yeah, that makes sense!

src/sentry/api/endpoints/project_group_index.py
lobsterkatie: But if the project on the grouphash is null, we don't want that record anyway. 

(It will never be the case that a grouphash has a group id but not a project id. I don't even know why `project` is nullable, since the only two places we create grouphashes ([here](https://github.com/getsentry/sentry/blob/15fcf7e673f7d7f98a6fbc3c5d430b8b1676e4a6/src/sentry/event_manager.py#L1418) and [here](https://github.com/getsentry/sentry/blob/15fcf7e673f7d7f98a6fbc3c5d430b8b1676e4a6/src/sentry/event_manager.py#L1786)) we create them with a project.)

PR #76704, opened 2024-08-28
feat(alerts): Add InsightsMetricField component to alert builder

edwardgou-sentry:
Adds an `InsightsMetricField` component to the Alerts builder. This allows users to select the Insights MRI and aggregate function they want to alert on for Insights based metrics. MRI options are obtained from the `metrics/meta` endpoint, and then filtered through a hardcoded list of Insights MRIs. Aggregate functions are currently also hardcoded.
image


narsaynorath (approved):

static/app/views/alerts/rules/metric/insightsMetricField.tsx
```diff
@@ -0,0 +1,221 @@
...
+  const metaArr = useMemo(() => {
+    return meta
+      .map(
+        metric =>
+          ({
+            ...metric,
+            ...parseMRI(metric.mri),
+          }) as ParsedMRI & MetricMeta
+      )
+      .filter(metric => INSIGHTS_METRICS.includes(metric.mri));
```
shruthilayaj, line 74: `INSIGHTS_METRICS` includes `d:spans/exclusive_time@millisecond` - should we drop it from the list - if i recall correctly, we've changed everything over to duration? especially because most of the insights metrics are leaf spans anyway

static/app/views/alerts/rules/metric/insightsMetricField.tsx
```diff
@@ -0,0 +1,221 @@
...
+];
+
+function aggregateRequiresArgs(aggregation?: string) {
+  return aggregation !== 'spm';
+}
+
+function InsightsMetricField({aggregate, project, onChange}: Props) {
+  const {data: meta, isLoading} = useVirtualizedMetricsMeta(
+    {projects: [parseInt(project.id, 10)]},
+    ['spans']
```
shruthilayaj, line 62: I see that `spans/webvital.inp`, `spans/mobile.frames_delay` etc are a part of the spans namespace but other webvitals and mobile metrics (ttid/ttfd etc) are still a part of "insights" even though they are on the transactions namespace. Are we planning to include those in the insights alerts workflow too? imo, we should, otherwise it seems arbitrary to the user that they can alert on some metrics here but not others

shruthilayaj (approved):
The code looks good to me, just a couple questions about the metrics we're allowing to alert on

static/app/views/alerts/rules/metric/insightsMetricField.tsx
edwardgou-sentry: Yup, I'm going to add the other insights metrics from the transaction namespace to the list in another PR! Just need to do some testing on those MRIs

static/app/views/alerts/rules/metric/insightsMetricField.tsx
edwardgou-sentry: Hmm I still see `span.self_time`/`d:spans/exclusive_time@millisecond` being used in some insights module code, so I think we'll have to keep it here for now. I think we can let this sit for a bit and drop this mri once we have time to confirm we're completely no longer using it

obostjancic (approved):

PR #76706, opened 2024-08-28
feat(performance): Set API statuses on last few performance endpoints

shruthilayaj:


src/sentry/api/endpoints/organization_events_stats.py
```diff
@@ -125,7 +125,7 @@
 @region_silo_endpoint
 class OrganizationEventsStatsEndpoint(OrganizationEventsV2EndpointBase):
     publish_status = {
-        "GET": ApiPublishStatus.UNKNOWN,
+        "GET": ApiPublishStatus.EXPERIMENTAL,
```
shruthilayaj, line 128: this is nowhere near ready to be publicized - a quarter's worth of work

src/sentry/api/endpoints/organization_events_trends_v2.py
```diff
@@ -49,7 +49,7 @@
 @region_silo_endpoint
 class OrganizationEventsNewTrendsStatsEndpoint(OrganizationEventsV2EndpointBase):
     publish_status = {
-        "GET": ApiPublishStatus.UNKNOWN,
+        "GET": ApiPublishStatus.PRIVATE,
```
shruthilayaj, line 52: only used in trends UI

src/sentry/api/endpoints/organization_measurements_meta.py
```diff
@@ -15,7 +15,7 @@
 @region_silo_endpoint
 class OrganizationMeasurementsMeta(OrganizationEventsEndpointBase):
     publish_status = {
-        "GET": ApiPublishStatus.UNKNOWN,
+        "GET": ApiPublishStatus.EXPERIMENTAL,
```
shruthilayaj, line 18: custom measurements is a little up in the air, so marking it as experimental too

edwardgou-sentry (approved):

src/sentry/api/endpoints/organization_measurements_meta.py
shruthilayaj, line 18: This is meant for custom measurements (like, the metrics kind), that product area is kind of in flux right now, so marking as experimental

PR #76707, opened 2024-08-28
chore (anomaly detection): get anomalies proxy API

mifu67:
Create a proxy for us to fetch anomalies from the frontend


JoshFerge (commented):
would it be possible to make this PR a bit smaller / split into multiple?

could start with just the endpoint definition w/ placeholders, then the seer logic as a follow-on.

additionally, if this is a new endpoint, i think it is fine to have it start as `EXPERIMENTAL`

mifu67:
> would it be possible to make this PR a bit smaller / split into multiple?
> 
> could start with just the endpoint definition w/ placeholders, then the seer logic as a follow-on.
> 
> additionally, if this is a new endpoint, i think it is fine to have it start as `EXPERIMENTAL`

Sure, I can break it up

nhsiehgit:
closed in favor of:
https://github.com/getsentry/sentry/pull/76934

PR #76708, opened 2024-08-28
nit(api): Added Camel Case Support for Waive Data Secrecy Endpoint

iamrajjoshi:
i don't want to do any conversions in the frontend, so handling camel case to snake case conversion in the backend.


src/sentry/data_secrecy/api/waive_data_secrecy.py
```diff
@@ -28,7 +29,7 @@ class WaiveDataSecrecyPermission(OrganizationPermission):
     }
 
 
-class DataSecrecyWaiverSerializer(serializers.Serializer, Serializer):
+class DataSecrecyWaiverSerializer(CamelSnakeSerializer, serializers.Serializer, Serializer):
```
cathteng, line 32: nice

cathteng (approved):
this is right, i also think i missed this last time

PR #76711, opened 2024-08-28
ref(replay): Use formatDuration directly, or through `<Duration>` in replay

ryan953:
This replaces `formatReplayDuration` with calls to the more general `formatDuration` and creates a new helper `<Duration/>` to make things easier inside a react tree.

Basically before, we had a weird function signature that wasn't as obvious to use. Now we have the `duration: Duration, precision: Unit` params that can be passed through the `<Duration/>` helper, or passed through ` <TimestampButton>` so it's obvious at the callsite what the output will be.

For formatting we don't need to check user preferences. The only time-related preferences are: 24-hour time  & local-vs-UTC time, neither apply to this.

The `components/duration/duration.tsx` helper is, imo, improved over the previous `components/duration.tsx` because the props are much clearer to use. Removing the old one is going to be a next step.

List of places to check/changed:
- Replay Timeline 'current time' and 'duration', no UI change (sec precision)
- Hover over the Timeline and the timestamp now shows ms precision, before was second precision
- Replay Breadcrumbs tab, no UI change (sec precision)
- Replay Console tab, right column, no UI change (sec precision)
- Replay Network tab, right column, no UI change (ms precision)
- Replay Error tab, right column, no UI change (sec precision)
- Replay A11y tab, "Results as of ..." & "Run Validation ...", no UI change (sec precision)
- Replay Memory tab, chart axes, no UI change (sec precision)
- Replay Memory tab, hover updated, shows ms precision, before was second precision
- Replay List page, duration column, no UI change (sec precision)


static/app/components/duration/duration.tsx
```diff
@@ -0,0 +1,37 @@
...
+        font-variant-numeric: tabular-nums;
+      `}
+      dateTime={formatDuration({duration, precision: 'ms', style: 'ISO8601'})}
+      title={formatDuration({duration, precision: 'ms', style: 'hh:mm:ss.sss'})}
+      {...props}
+    >
+      {formatDuration({duration, precision, style})}
+    </time>
+  );
+}
```
evanpurkhiser, line 37: We don't have something like this already?

static/app/components/duration/duration.tsx
```diff
@@ -0,0 +1,37 @@
...
+export default function Duration({duration, precision, ...props}: Props) {
+  // Style and precision should match, otherwise style will pad out missing or
+  // truncated values which we don't want in this component.
+  const style = precision === 'ms' ? 'hh:mm:ss.sss' : 'hh:mm:ss';
+
+  return (
+    <time
+      css={css`
+        font-variant-numeric: tabular-nums;
+      `}
```
evanpurkhiser, line 29: Why not use a styled component here?

If we start having two ways to do styling it's going to end up confusing people unfamiliar with the codebase. When do I use what?

static/app/components/duration/duration.tsx
```diff
@@ -0,0 +1,37 @@
...
+  duration: TDuration;
+
+  /**
+   * How granular to render the value. You can pass in something that has `ms`
+   * precision but only show the total number of seconds (ala Math.floor)
+   */
+  precision: Unit;
+}
+
+export default function Duration({duration, precision, ...props}: Props) {
```
scttcper, line 20: props here has a massive custom type
image

might be another way to avoid this

scttcper: or Pick just the properties you need

static/app/utils/duration/formatDuration.tsx
evanpurkhiser, line 1: Seems like the changes in this file should be separate from this PR no?

Looks like this is adding more features to `formatDuration`, while a large part of the rest of this PR is updating usage of duration in replay components?

static/app/utils/duration/formatDuration.spec.tsx
```diff
@@ -77,34 +63,96 @@ describe('formatDuration', () => {
...
+      {duration: [500_012, 'ms'], precision: 'sec', expected: '8:20.000'},
+      {duration: [500, 'sec'], precision: 'ms', expected: '8:20.000'},
+      {duration: [500, 'sec'], precision: 'sec', expected: '8:20.000'},
+      {duration: [500, 'sec'], precision: 'ms', expected: '8:20.000'},
+      {duration: [500, 'sec'], precision: 'sec', expected: '8:20.000'},
+      {duration: [500.012, 'sec'], precision: 'ms', expected: '8:20.012'},
+      {duration: [500.012, 'sec'], precision: 'sec', expected: '8:20.000'},
+      {duration: [500.012, 'sec'], precision: 'ms', expected: '8:20.012'},
+      {duration: [500.012, 'sec'], precision: 'sec', expected: '8:20.000'},
+    ] as Array<{duration: Duration; expected: string; precision: Unit}>)(
```
scttcper, line 79: i think there's a generic here instead of casting
```suggestion
    it.each<{duration: Duration; expected: string; precision: Unit}>([
      {duration: [500_000, 'ms'], precision: 'ms', expected: '8:20.000'},
      {duration: [500_000, 'ms'], precision: 'sec', expected: '8:20.000'},
      {duration: [500_012, 'ms'], precision: 'ms', expected: '8:20.012'},
      {duration: [500_012, 'ms'], precision: 'sec', expected: '8:20.000'},
      {duration: [500, 'sec'], precision: 'ms', expected: '8:20.000'},
      {duration: [500, 'sec'], precision: 'sec', expected: '8:20.000'},
      {duration: [500, 'sec'], precision: 'ms', expected: '8:20.000'},
      {duration: [500, 'sec'], precision: 'sec', expected: '8:20.000'},
      {duration: [500.012, 'sec'], precision: 'ms', expected: '8:20.012'},
      {duration: [500.012, 'sec'], precision: 'sec', expected: '8:20.000'},
      {duration: [500.012, 'sec'], precision: 'ms', expected: '8:20.012'},
      {duration: [500.012, 'sec'], precision: 'sec', expected: '8:20.000'},
    ])(
```

static/app/components/duration/duration.tsx
evanpurkhiser: I'm not super worried about this, we already mostly extend HTML types. It's only when we start having large intersections that it becomes a big problem

static/app/components/duration/duration.tsx
ryan953: There's `components/duration.tsx` which i mentioned in the desc.

evanpurkhiser: Shouldn't we keep using that one so we only have one component and we can migrate it to have good props?

static/app/utils/duration/formatDuration.tsx
ryan953: The new format here is used by the `dateTime` attr of the `<time>` tag tho.

static/app/components/duration/duration.tsx
ryan953: No. I'm making a nice new one and wull remove the old one.

evanpurkhiser: How much usage of the old one is there?

static/app/components/duration/duration.tsx
scttcper: this does get type checked when merged back into the time element and will take quite a bit of time

static/app/components/duration/duration.tsx
ryan953: 19 instances

static/app/components/duration/duration.tsx
ryan953: changed it up!

static/app/utils/duration/formatDuration.tsx
```diff
@@ -75,13 +79,62 @@ export default function formatDuration({
...
-      const str = formatSecondsToClock(valueInSec, {
-        padAll: style.startsWith('hh:mm:ss'),
-      });
+      const str = formatSecondsToClock(valueInSec, {padAll});
       const [head, tail] = str.split('.');
-      return includeMs ? [head, tail ?? '000'].join('.') : String(head);
+      return includeMs
+        ? [head, precision === 'ms' ? tail ?? '000' : '000'].join('.')
+        : String(head);
+    case 'ISO8601':
```
ryan953, line 95: Threw up https://github.com/getsentry/sentry/pull/76823 to handle this new case/format.

static/app/utils/duration/formatDuration.tsx
ryan953: Moved into https://github.com/getsentry/sentry/pull/76904

static/app/components/duration/duration.tsx
```diff
@@ -0,0 +1,45 @@
+import type {HTMLAttributes} from 'react';
```
priscilawebdev, line 1: shouldn't we name this file `index.tsx`?

static/app/components/duration/duration.tsx
```diff
@@ -0,0 +1,45 @@
...
+      {...props}
+    >
+      {formatDuration({duration, precision, style})}
+    </time>
+  );
+})`
+  font-variant-numeric: tabular-nums;
+`;
+
+export default Duration;
```
priscilawebdev, line 45: shall we make this a named export?

static/app/components/duration/duration.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+interface Props extends HTMLAttributes<HTMLTimeElement> {
+  /**
+   * The Duration that you want to render
+   */
+  duration: TDuration;
+
+  /**
+   * How granular to render the value. You can pass in something that has `ms`
+   * precision but only show the total number of seconds (ala Math.floor)
```
priscilawebdev, line 15:
```suggestion
   * precision but only show the total number of seconds (aka Math.floor)
```

static/app/components/duration/duration.stories.tsx
```diff
@@ -0,0 +1,27 @@
+import Duration from 'sentry/components/duration/duration';
+import storyBook from 'sentry/stories/storyBook';
+
+export default storyBook(Duration, story => {
```
priscilawebdev, line 4: it will be nice to have tests too 🙂

ryan953: added

static/app/components/duration/duration.tsx
ryan953: Can't do that now, it'll collide with the existing `components/duration.tsx` file which i plan to replace in the near future.

static/app/components/duration/duration.tsx
ryan953: ah, i lost a fix for this. updated again.

static/app/components/duration/duration.tsx
ryan953: The develop docs don't say we're doing that, and i think it's a bad pattern because it encourages people to put multiple things inside a file which goes against this part of the develop docs: "Name the file to match the component or function that is being exported."

But we shouldn't solve that problem in this PR

priscilawebdev: That's interesting. Although it's not in the docs, I recall us discussing a preference for named exports in a past TSC meeting. Maybe we should bring that up again

static/app/components/duration/duration.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+    const time = screen.getByText('01:23');
+    expect(time).toBeInTheDocument();
+  });
+
+  it('should render the duration in the specified format', () => {
+    render(<Duration duration={[83_456, 'ms']} precision="ms" format="h:mm:ss.sss" />);
+
+    const time = screen.getByText('1:23.456');
+    expect(time).toBeInTheDocument();
```
priscilawebdev, line 17:
```suggestion
    expect(screen.getByText('1:23.456')).toBeInTheDocument();
```

static/app/components/duration/duration.spec.tsx
```diff
@@ -0,0 +1,29 @@
+import {render, screen} from 'sentry-test/reactTestingLibrary';
+
+import Duration from 'sentry/components/duration/duration';
+
+describe('Duration', () => {
+  it('should render the duration in the default format', () => {
+    render(<Duration duration={[83, 'sec']} precision="sec" />);
+
+    const time = screen.getByText('01:23');
+    expect(time).toBeInTheDocument();
```
priscilawebdev, line 10:
```suggestion
    expect(screen.getByText('01:23')).toBeInTheDocument();
```

static/app/components/duration/duration.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  it('should render the duration in the specified format', () => {
+    render(<Duration duration={[83_456, 'ms']} precision="ms" format="h:mm:ss.sss" />);
+
+    const time = screen.getByText('1:23.456');
+    expect(time).toBeInTheDocument();
+  });
+
+  it('should include `dateTime` & `title` attributes for accessibility', () => {
```
priscilawebdev, line 20: nitpick: maybe we can have this check in one of the previous tests?

priscilawebdev (approved):
left a few suggestions but it looks good to me

static/app/components/duration/duration.tsx
ryan953: I know how evan feels about it, which is easy to conflate with having this resvoled. But it's been in the tsc todo task list forever, it's on [github here](https://github.com/getsentry/frontend-tsc/issues/33) because i ported the topic from notion.

PR #76716, opened 2024-08-29
fix(custom-views): Fix rename functionality for custom tabs, update growingInput to use MutationObserver

MichaelSun48:
This PR fixes the following issues with the rename function of tabs: 
1. Renaming worked maybe 10% of the time in safari 
2. Renaming would lose focus randomly sometimes 
3. Renaming would slightly shift the title and create a stutter animation 

Edit:
* Added support for double click to rename 
* There's still a bug where if you attempt to rename a tab before the issues stream has loaded, it will lose focus when the issue stream loads

### DIFF
--- a/static/app/components/growingInput.tsx
+++ b/static/app/components/growingInput.tsx
@@ -47,27 +47,10 @@ export const GrowingInput = forwardRef<HTMLInputElement, InputProps>(
 
     // If the input is controlled we resize it when the value prop changes
     useLayoutEffect(() => {
-      // This mutationObserver is used to resize the input if the inputRef's style changes
-      // It's necessary because the refs do not trigger useEffects
-      const observer = new MutationObserver(() => {
-        inputRef.current && resize(inputRef.current);
-      });
-
       if (isControlled && inputRef.current) {
         resize(inputRef.current);
       }
-
-      if (inputRef.current) {
-        observer.observe(inputRef.current, {
-          attributes: true,
-          attributeFilter: ['style'],
-        });
-      }
-
-      return () => {
-        observer.disconnect();
-      };
-    }, [props.value, props.placeholder, isControlled]);
+    }, [props.value, props.placeholder, isControlled, props.className, props.style]);
 
     // If the input is uncontrolled we resize it when the user types
     const handleChange = useCallback(
--- a/static/app/components/growingInput.tsx
+++ b/static/app/components/growingInput.tsx
@@ -47,9 +47,26 @@ export const GrowingInput = forwardRef<HTMLInputElement, InputProps>(
 
     // If the input is controlled we resize it when the value prop changes
     useLayoutEffect(() => {
+      // This mutationObserver is used to resize the input if the inputRef's style changes
+      // It's necessary because the refs do not trigger useEffects
+      const observer = new MutationObserver(() => {
+        inputRef.current && resize(inputRef.current);
+      });
+
       if (isControlled && inputRef.current) {
         resize(inputRef.current);
       }
+
+      if (inputRef.current) {
+        observer.observe(inputRef.current, {
+          attributes: true,
+          attributeFilter: ['style'],
+        });
+      }
+
+      return () => {
+        observer.disconnect();
+      };
     }, [props.value, props.placeholder, isControlled]);
 
     // If the input is uncontrolled we resize it when the user types
--- a/static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
+++ b/static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
@@ -1,25 +1,33 @@
 import {useEffect, useRef, useState} from 'react';
+import {useTheme} from '@emotion/react';
 import styled from '@emotion/styled';
+import {FocusScope} from '@react-aria/focus';
 
 import {GrowingInput} from 'sentry/components/growingInput';
+
+interface EditableTabTitleProps {
+  isEditing: boolean;
+  isSelected: boolean;
+  label: string;
+  onChange: (newLabel: string) => void;
+  setIsEditing: (isEditing: boolean) => void;
+}
 
 function EditableTabTitle({
   label,
   onChange,
   isEditing,
+  isSelected,
   setIsEditing,
-}: {
-  isEditing: boolean;
-  label: string;
-  onChange: (newLabel: string) => void;
-  setIsEditing: (isEditing: boolean) => void;
-}) {
+}: EditableTabTitleProps) {
   const [inputValue, setInputValue] = useState(label);
 
+  const theme = useTheme();
   const inputRef = useRef<HTMLInputElement>(null);
   const isEmpty = !inputValue.trim();
 
-  const handleOnBlur = () => {
+  const handleOnBlur = (e: React.FocusEvent<HTMLInputElement, Element>) => {
+    e.preventDefault();
     const trimmedInputValue = inputValue.trim();
     if (!isEditing) {
       return;
@@ -37,9 +45,6 @@ function EditableTabTitle({
   };
 
   const handleOnKeyDown = (e: React.KeyboardEvent) => {
-    if (e.key === 'Enter') {
-      handleOnBlur();
-    }
     if (e.key === 'Escape') {
       setInputValue(label.trim());
       setIsEditing(false);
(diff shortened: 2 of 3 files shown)


static/app/components/growingInput.tsx
```diff
@@ -47,9 +47,26 @@ export const GrowingInput = forwardRef<HTMLInputElement, InputProps>(
...
+      if (inputRef.current) {
+        observer.observe(inputRef.current, {
+          attributes: true,
+          attributeFilter: ['style'],
+        });
+      }
+
+      return () => {
+        observer.disconnect();
+      };
```
MichaelSun48, line 69: This pretty much just monitors the ref for style changes. This is necessary because `resize()` takes into account the fontWeight, fontSize, and fontFamily styles to generate the correct width for the input. This previously caused issues because the tab title changes its fontWeight depending on if it's selected or not. This caused an annoying bug where tabs that were not initially selected, once selected, would see their tab title get cutoff slightly, since the tab title previously had lower weight, and the growingInput container did not resize the container accordingly. 

useEffect also does not trigger when refs change, so adding `inputRef.current?.style` to the dependency array did not achieve the desired behavior.

static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
```diff
@@ -51,26 +56,27 @@ function EditableTabTitle({
 
   useEffect(() => {
     if (isEditing) {
-      setTimeout(() => {
-        inputRef?.current?.focus();
-      }, 0);
+      requestAnimationFrame(() => {
+        inputRef.current?.focus();
+      });
```
MichaelSun48, line 61: ngl i have no idea what this does, but this makes renaming work in Safari

static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
```diff
@@ -51,26 +56,27 @@ function EditableTabTitle({
...
-    <div style={{height: '20px'}}> {label}</div>
+  return (
+    <FocusScope>
+      <StyledGrowingInput
+        value={inputValue}
+        onChange={handleOnChange}
+        onKeyDown={handleOnKeyDown}
+        onBlur={handleOnBlur}
+        ref={inputRef}
+        style={{fontWeight: isSelected ? theme.fontWeightBold : theme.fontWeightNormal}}
```
MichaelSun48, line 77: This is redundant with the fontweight set by the `<Tab>` component, but it's necessary to attach this style directly to the growing input to trigger the MutationObserver. Adding it to the emotion styles does not work either.

static/app/components/growingInput.tsx
```diff
@@ -47,9 +47,26 @@ export const GrowingInput = forwardRef<HTMLInputElement, InputProps>(
...
+      });
+
       if (isControlled && inputRef.current) {
         resize(inputRef.current);
       }
+
+      if (inputRef.current) {
+        observer.observe(inputRef.current, {
+          attributes: true,
+          attributeFilter: ['style'],
```
malwilley, line 63: This seems like it would probably result in the width being calculated multiple times, since the resize function will modify the `width` style

static/app/components/growingInput.tsx
malwilley: It doesn't trigger when refs change, but it would if you did `props.style` or `props.className`. I actually think it might be simplest if you just modified the class and added `props.className` to the deps array.

MichaelSun48: oml im an idiot for not thinking of this before 🤦‍♂️

MichaelSun48: Going to add both classname and style to the dependency array since it looks like updating style rather than classname results in less jitter 

Updating classname:


https://github.com/user-attachments/assets/9152130d-bd26-403b-b056-fba8676e2944



Updating styles:

https://github.com/user-attachments/assets/7f9aa47c-2ae2-49e8-8287-c8d63cc12bf0

static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
```diff
@@ -51,40 +57,44 @@ function EditableTabTitle({
...
-      value={inputValue}
-      onChange={handleOnChange}
-      onKeyDown={handleOnKeyDown}
-      onBlur={handleOnBlur}
-      ref={inputRef}
-    />
-  ) : (
-    <div style={{height: '20px'}}> {label}</div>
+  return (
+    <FocusScope>
```
malwilley, line 71: Is this FocusScope necessary?

static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
```diff
@@ -51,40 +57,44 @@ function EditableTabTitle({
...
+      <StyledGrowingInput
```
malwilley, line 72: Seems a little wasteful to always render a GrowingInput - maybe we could render just the div if it's not selected?

static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
```diff
@@ -1,29 +1,38 @@
 import {useEffect, useRef, useState} from 'react';
+import {useTheme} from '@emotion/react';
 import styled from '@emotion/styled';
+import {FocusScope} from '@react-aria/focus';
 
 import {GrowingInput} from 'sentry/components/growingInput';
 
+interface EditableTabTitleProps {
+  isEditing: boolean;
```
malwilley, line 9: What does `isEditing` do now that the input is always being rendered? Seems like it should disable the input or something, but I don't see it doing much.

static/app/components/growingInput.tsx
malwilley: Something to note about passing `style`, it looks like you are passing the object directly which will cause this resize to happen every time the component rerenders, event when `style` doesn't actually change. You'll need to wrap it in a `useMemo` to avoid that

static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
MichaelSun48: Its a way for the parent component to control when the inputRef gets focused. I can try to refactor this so that the parent component creates the ref instead and focuses it when the rename button gets clicked.

static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
MichaelSun48: This created a little visual jitter when hitting rename because the div and growingInput didn't render the text in precisely the same position. 

https://github.com/user-attachments/assets/2978dac6-56d0-4dfe-b94f-b85997c2cf38
(This is in the issue stream now) 

Definitely a very small improvement, happy to add the div back if you think the minor change isn't worth rendering the GrowingInput

static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
MichaelSun48: Nope, this was a part of a previous solution I tried, removing it in the next commit.

static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
malwilley: I think it's good to render the input when the tab is selected, but I don't think we need to render the input when the tab isn't even selected (and thus cannot be edited), right?

The difference in styling may be due to the line-height

static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
malwilley: You could call `inputRef.current.focus()` on the double click event, right? Then you wouldn't need to do it in a useEffect

MichaelSun48: Yup! but the parent component still needs some way to trigger the focus when the user clicks on the rename button from the tab's ellipsis menu

static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
MichaelSun48: That's true, okay I'll render just the div when its unselected in that case

static/app/views/issueList/groupSearchViewTabs/editableTabTitle.tsx
malwilley: Ohhhh I see now. Gotcha, keep it as is then!

MichaelSun48:
Going to fix that bug with issue stream stealing focus to a separate PR - this PR fixes other big issues and should be merged in asap

malwilley (approved):

PR #76721, opened 2024-08-29
feat(translations): Allow reusing elements

ArthurKnaus:
As of limitations in our parsing algorithm, it is not possible to re-use elements inside rich-text translations.
E.g. if I have multiple code tags in a translation, I need to define a group and element for each of them.

```tsx
tct(
  'Add the Sentry SDK as a dependency using [codeNpm:npm] or [codeYarn:yarn]. You need a minimum version 7.27.0 of [code:@sentry/react] in order to use Session Replay. You do not need to install any additional packages.',
  {
    code: <code />,
    codeYarn: <code />,
    codeNpm: <code />,
  }
)
```

This implementation identifies groups with position based IDs instead of  using only the group name, enabling us to reuse elements:
```tsx
tct(
  'Add the Sentry SDK as a dependency using [code:npm] or [code:yarn]. You need a minimum version 7.27.0 of [code:@sentry/react] in order to use Session Replay. You do not need to install any additional packages.',
  {code: <code />}
);
```


priscilawebdev:
I tested it locally and it works like a charm 💅 thank you for working on it @ArthurKnaus 👏

static/app/locale.tsx
```diff
@@ -208,9 +209,14 @@ export function parseComponentTemplate(template: string): ParsedTemplate {
...
       } else {
-        pos = regex.lastIndex = process(regex.lastIndex, groupName, true);
+        pos = regex.lastIndex = process(
+          regex.lastIndex,
+          groupName + groupId.toString(),
+          true
+        );
       }
-      buf.push({group: groupName});
+      buf.push({group: groupName, id: groupName + groupId.toString()});
```
priscilawebdev, line 218: nit: we could use only the `id` here 
```suggestion
      buf.push({group: groupName, id: groupId.toString()});
```

ArthurKnaus: True, I added the name as it was easier to debug when logging the parsed result.

static/app/locale.spec.tsx
```diff
@@ -49,4 +49,17 @@ describe('locale.gettextComponentTemplate', () => {
...
+    const RenderChildren = ({children}: {children?: React.ReactNode}) => children;
+    render(
+      <div>
+        {tct('[render:one] [render:two] [render:three]', {
+          render: <RenderChildren />,
+        })}
+      </div>
+    );
+
+    expect(screen.getByText(textWithMarkupMatcher('one two three'))).toBeInTheDocument();
```
ryan953, line 63: Can you write a test where renderChuldren is a hyml node or react node too, so we can assert something like `<b>one<\b> <b>two<\b>` as well.

ArthurKnaus: Sure can do 👍  Though the current one is already a react node. So it should not behave differently.

ryan953: it shouldn't i know. but i'm worried about `<Fragment>` getting rendering instead of the thing that's been passed in. Something that could break one day

static/app/locale.spec.tsx
```diff
@@ -49,4 +49,48 @@ describe('locale.gettextComponentTemplate', () => {
...
+          link: <a href="/link" />,
+        })}
+      </div>
+    );
+
+    expect(
+      screen.getByText(textWithMarkupMatcher('text with another group'))
+    ).toBeInTheDocument();
+    expect(container.container.innerHTML).toEqual(
+      '<div><b>text with <a href="/link">another</a> group</b></div>'
```
ryan953, line 93: cool i didn't know this was possible!

ryan953 (approved):

evanpurkhiser (approved):
Awesome work!

scttcper (approved):
makes sense

Issue #60097, opened 2023-11-16
Linear Ticket creation is janky due to dependent fields being auto populated


Dhrumil-Sentry:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

1. Create a Linear ticket from an issue
2. See that the Team is auto populated
3. But the dependent fields of Assignee, Label cannot be edited
4. Need to update the Team selection to make the dependent fields editable

### Expected Result

The dependent fields should be editable from the start

### Actual Result

<img width="765" alt="image" src="https://github.com/getsentry/sentry/assets/109983869/92a7f81b-4964-4954-b7b6-7452e83e9510">

### Product Area

Settings - Integrations

skbogner:
@Dhrumil-Sentry notice that we only auto populate the team field if there's a single team. So bear that in mind for repro.

natikgadzhi:
I'm happy to tinker with this if you need a pair of hands — I have time tomorrow. But, looks like Linear is not on the list of integrations available in local dev environment, and that does not seem to be the feature flag thing — I don't see the code for it at all. Is it in sentry closed-source for now?

skbogner:
> I'm happy to tinker with this if you need a pair of hands — I have time tomorrow. But, looks like Linear is not on the list of integrations available in local dev environment, and that does not seem to be the feature flag thing — I don't see the code for it at all. Is it in sentry closed-source for now?

The Linear integration is owned by us (Linear), so probably not there no, but this should be easily reproducible without it. You just need a field which gets a default value and one that depends on it. The dependent field is not editable straight away.

skbogner:
@Dhrumil-Sentry can we move this forward?

Dhrumil-Sentry:
@skbogner I've asked the eng team to look into this issue, We'll comment here as we have any updates

scefali:
@skbogner @Dhrumil-Sentry  Is this bug still happening? What determines if the team gets autopopulated? I am not able to replicate the auto-population of the team for the Linear integration. I also set up my own integration locally that populates a default value through the endpoint and the fields are unlocked correctly:




![Image](https://github.com/user-attachments/assets/91ac8493-7523-47c7-bb88-f1bbcc310bf2)

skbogner:
Hi, it's still the case although it is behaving slightly differently. 

We only default the team when there's a single team, but if I just force default some team I repro the bug.

I'll send a video over email.

PR #76723, opened 2024-08-29
feat(issue-platform): handle edge case where api populates a default value

scefali:
This fixes a bug where when we have a default value we don't load dependent fields on page load. To get the values to appear, the user has to re-select the field. Before:

https://github.com/user-attachments/assets/f8f6f863-c102-4ec5-b51e-ef0d1529b40a

After:

https://github.com/user-attachments/assets/6a717613-067f-45df-9f06-b43d42ed44d8

For Linear, this matters when there is only one team and sets that as a default.

Resolves issue: https://github.com/getsentry/sentry/issues/60097

### DIFF
--- a/static/app/views/alerts/rules/issue/sentryAppRuleModal.spec.tsx
+++ b/static/app/views/alerts/rules/issue/sentryAppRuleModal.spec.tsx
@@ -176,5 +176,67 @@ describe('SentryAppRuleModal', function () {
 
       await submitSuccess();
     });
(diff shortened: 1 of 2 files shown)


sentaur-athena (commented):
SentryAppExternalForm is used in some other places as well. What does this change mean for them?

scefali:
@sentaur-athena It should have the same fix in other places, it's rather agnostic to where it is used.

sentaur-athena (approved):

PR #76728, opened 2024-08-29
fix: capture_message does not accept extra

armenzg:
Fixes issue introduced in #76616.

### DIFF
--- a/src/sentry/integrations/github/client.py
+++ b/src/sentry/integrations/github/client.py
@@ -428,7 +428,7 @@ class GitHubBaseClient(GithubProxyClient, RepositoryClient, CommitContextClient)
         else:
             # We do not raise the exception so we can keep iterating through the repos.
             # Nevertheless, investigate the error to determine if we should abort the processing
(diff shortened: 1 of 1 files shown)


src/sentry/integrations/github/client.py
```diff
@@ -428,7 +428,7 @@ def _populate_trees_process_error(self, error: ApiError, extra: dict[str, str])
         else:
             # We do not raise the exception so we can keep iterating through the repos.
             # Nevertheless, investigate the error to determine if we should abort the processing
-            capture_message("Continuing execution. Investigate: %s", error_message, extra=extra)
+            capture_message("Continuing execution. Investigate: %s", error_message)
```
leeandher, line 431: From [the docs](https://docs.sentry.io/platforms/python/usage/#capturing-messages) it looks it doesn't look like capture_message takes a second arg, You can probably swap it for an f-string here. Also could we add the extra back as context with `sentry_sdk.set_context`? In case it has something important.

leeandher (approved):

PR #76734, opened 2024-08-29
feat(onboarding): backend to allow members to delete sent invites

ameliahsu:
If members have permission to invite members to an organization (the org's `disable_member_invite` flag is `false`), we also want them to be able to remove invitations that they sent.


sentaur-athena (approved):

iamrajjoshi:
we should also make a note to update the documentation once we GA

PR #76740, opened 2024-08-29
ref(jira): Add data model for Jira create issue metadata

GabeVillalobos:
Adds Jira [Create Issue Metadata](https://developer.atlassian.com/server/jira/platform/jira-rest-api-example-discovering-meta-data-for-creating-issues-6291669/) specific classes to make our Jira integration code more type safe. This is a precursor to a refactor of our `create_issue` method, which is currently frail and often raises errors when attempting to create new issues due to incorrect data transformations for some custom fields, such as `Date`, `Development`, and `Team`.

I've verified this works on a full Jira configuration I've pulled from a test Jira instance with some of the custom fields described above.

## Next Steps:
- [ ] Wrap our `create_issue` flows in metrics to gather current failure rates
- [ ] Implement a new transform strategy that simplifies our current one
- [ ] Wrap the new `create_issue` transform implementation in a feature flag and monitor the metrics for failures when toggling it on

### DIFF
--- a/fixtures/integrations/jira/stubs/createmeta_response.json
+++ b/fixtures/integrations/jira/stubs/createmeta_response.json
@@ -47,10 +47,12 @@
               "name": "Issue Type",
               "key": "issuetype",
               "hasDefaultValue": false,
-              "operations": ["set"]
+              "operations": ["set"],
+              "schema": { "type": "issuetype", "system": "issuetype" }
             },
             "labels": {
               "required": false,
+              "operations": ["add", "set", "remove"],
               "schema": {
                 "type": "array",
                 "items": "string",
@@ -89,6 +91,7 @@
             "customfield_10200": {
               "operations": ["set"],
               "required": false,
+              "key": "customfield_10200",
               "schema": {
                 "type": "option",
                 "custom": "com.codebarrel.jira.iconselectlist:icon-select-cf",
@@ -125,6 +128,7 @@
                 "customId": 10202
               },
               "name": "Feature",
+              "key": "customfield_10300",
               "hasDefaultValue": false,
               "operations": ["add", "set", "remove"],
               "allowedValues": [
--- a/src/sentry/integrations/jira/integration.py
+++ b/src/sentry/integrations/jira/integration.py
@@ -6,6 +6,7 @@ from collections.abc import Mapping, Sequence
 from operator import attrgetter
 from typing import Any
 
+import sentry_sdk
 from django.conf import settings
 from django.urls import reverse
 from django.utils.functional import classproperty
@@ -19,6 +20,7 @@ from sentry.integrations.base import (
     IntegrationMetadata,
     IntegrationProvider,
 )
+from sentry.integrations.jira.models.create_issue_metadata import JiraIssueTypeMetadata
 from sentry.integrations.jira.tasks import migrate_issues
 from sentry.integrations.mixins.issues import MAX_CHAR, IssueSyncIntegration, ResolveSyncAction
 from sentry.integrations.models.external_issue import ExternalIssue
@@ -570,6 +572,7 @@ class JiraIntegration(IssueSyncIntegration):
         return fkwargs
 
     def get_issue_type_meta(self, issue_type, meta):
+        self.parse_jira_issue_metadata(meta)
         issue_types = meta["issuetypes"]
         issue_type_meta = None
         if issue_type:
@@ -1038,6 +1041,13 @@ class JiraIntegration(IssueSyncIntegration):
                 "organization_id": self.organization_id,
             }
         )
+
+    def parse_jira_issue_metadata(self, meta: dict[str, Any]) -> list[JiraIssueTypeMetadata] | None:
+        try:
+            return JiraIssueTypeMetadata.from_jira_meta_config(meta)
+        except Exception as e:
+            sentry_sdk.capture_exception(e)
+            return None
 
 
 class JiraIntegrationProvider(IntegrationProvider):
--- a/src/sentry/integrations/jira/models/create_issue_metadata.py
+++ b/src/sentry/integrations/jira/models/create_issue_metadata.py
@@ -19,7 +19,7 @@ class JiraSchemaTypes(str, Enum):
 
 @dataclass(frozen=True)
 class JiraSchema:
-    type: str
+    schema_type: str
     """
     The Field type. Possible types include:
     - string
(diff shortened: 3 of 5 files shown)


src/sentry/integrations/jira/models/schema_model.py
```diff
@@ -0,0 +1,174 @@
...
+            for old_name, new_name in cls.__jira_schema_parameter_map
+            if old_name in data
+        }
+
+        return cls(**mapped_field_params)
+
+    @classmethod
+    def from_dict_list(cls, data: list[dict[str, Any]]) -> list["JiraSchema"]:
+        return [cls.from_dict(item) for item in data]
+
```
GabeVillalobos, line 77: I'm not super sold on this approach, but figured it'd be nice to have a simplified means of converting from camelCase to snake_case. It also provides some parameter filtering, since we don't use all of the fields Jira provides us on these schemas, especially some of the optional custom fields.

src/sentry/integrations/jira/integration.py
```diff
@@ -570,6 +572,7 @@ def build_dynamic_field(self, field_meta, group=None):
         return fkwargs
 
     def get_issue_type_meta(self, issue_type, meta):
+        self.parse_jira_issue_metadata(meta)
```
GabeVillalobos, line 575: Adding this here will allow us to gather some error reporting around customer configs that we may not have accounted for. I figure leaving this in for a few days should be enough.

src/sentry/integrations/jira/models/create_issue_metadata.py
```diff
@@ -0,0 +1,161 @@
...
+    project = "project"
+    date = "date"
+    team = "team"
+    number = "number"
+    any = "any"
+
+
+@dataclass(frozen=True)
+class JiraSchema:
+    schema_type: str
```
cathteng, line 22: can this be typed with `JiraSchemaTypes`?

cathteng (commented):
did we ever consider using the Jira SDK? (not maintained by Jira) https://jira.readthedocs.io/

they have typing already that we could utilize / reference https://github.com/pycontribs/jira/blob/main/jira/resources.py but i know we've been wary of directly using unofficial sdks

GabeVillalobos:
> did we ever consider using the Jira SDK? (not maintained by Jira) https://jira.readthedocs.io/

@cathteng We discussed it briefly, but I have the feeling this would be a hard sell for security. Let me ask around a bit to see what folks think though.

fixtures/integrations/jira/stubs/createmeta_response.json
```diff
@@ -125,6 +128,7 @@
                 "customId": 10202
```
sentaur-athena, line 128: Can you educate me on what the fields in the custom fields are? Just by looking at the previous ones it seems like customId is the same number that comes after the field which is `10300` in this case. But I have no idea how these are used.

src/sentry/integrations/jira/integration.py
```diff
@@ -1039,6 +1042,13 @@ def migrate_issues(self):
             }
         )
 
+    def parse_jira_issue_metadata(self, meta: dict[str, Any]) -> list[JiraIssueTypeMetadata] | None:
+        try:
+            return JiraIssueTypeMetadata.from_jira_meta_config(meta)
+        except Exception as e:
+            sentry_sdk.capture_exception(e)
+            return None
```
sentaur-athena, line 1050: Eventually this would throw integration exception right?

src/sentry/integrations/jira/models/create_issue_metadata.py
```diff
@@ -0,0 +1,161 @@
...
+    id: str
+    description: str
+    name: str
+    subtask: bool
+    icon_url: str
+    url: str
+    fields: dict[str, JiraField]
+
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> "JiraIssueTypeMetadata":
```
sentaur-athena, line 135: Using this as a learning opportunity. Do type references have other benefits other than saying this type is only used here?

src/sentry/integrations/jira/integration.py
GabeVillalobos: Yes, once we're actively using the model in our code, this will raise if it fails to parse the jira meta config.

fixtures/integrations/jira/stubs/createmeta_response.json
GabeVillalobos: So I'm not 100% sure about the full schema, but these are the fields we care about:
- `schema.custom` - some namespaced identifier for the source of the custom field. Many of these seem to be plugin/add-on namespaces such as:
    - `com.atlassian.jira.plugin.system.customfieldtypes:datepicker` - date field
    - `com.pyxis.greenhopper.jira:gh-lexo-rank` - ranking field
-  `key` - We use this when creating the issue payload to send to Jira, which acts as the property name more or less.

Now for the confusing one:
`schema.customId` - Appears to be some incrementing ID field on the issue that's generated when you add it to the template? Luckily, I haven't found a need to reference this.

It's very similar to `key`, and in my own testing seems to align with the key's`customfield_<number>` identifier, although I don't think there's any guarantee on this as Jira's own API examples don't align: https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-fields/#api-rest-api-3-field-search-get.

src/sentry/integrations/jira/models/create_issue_metadata.py
GabeVillalobos: Discussed this via slack, but as far as I know, this is just the older way of specifying forward refs pre python 3.7. Importing `__future__.annotations` and removing the quotes also works the same.

fixtures/integrations/jira/stubs/createmeta_response.json
sentaur-athena: Thanks for explaining. Yeah wanted to make sure it doesn't have to align.

tests/sentry/integrations/jira/models/test_jira_schema.py
```diff
@@ -0,0 +1,13 @@
+from fixtures.integrations.jira.stub_client import StubJiraApiClient
+from sentry.integrations.jira.models.create_issue_metadata import JiraIssueTypeMetadata
+from sentry.testutils.cases import TestCase
+
+
+class TestJiraSchema(TestCase):
```
sentaur-athena, line 6: Mentioned this before but to have it on the PR too: we need better testing for these, something that looks like real life config that what stub currently have. As you monitor these can you please update the stub meta data as well?

sentaur-athena (approved):

tests/sentry/integrations/jira/models/test_jira_schema.py
GabeVillalobos: Yes, agreed! I'm trying to think of a way to decompose the schema in a way to make individual parts testable vs just having a massive schema acting as a superset of everything, so I'll circle back around to this shortly.

PR #76742, opened 2024-08-29
chore(feedback): add extra source tags to create_feedback_issue metrics

aliu39:
- `referrer`: identifies the pipeline feedback was ingested/shimmed from. See the enum + architecture doc
- `client_source`: "source" field optionally sent in new feedback envelopes. This is user/SDK-defined, current examples are "widget" vs "api".

### DIFF
--- a/src/sentry/feedback/usecases/create_feedback.py
+++ b/src/sentry/feedback/usecases/create_feedback.py
@@ -166,7 +166,6 @@ def should_filter_feedback(event, project_id, source: FeedbackCreationSource):
             tags={
                 "reason": "missing_context",
                 "referrer": source.value,
-                "client_source": event["contexts"]["feedback"].get("source"),
             },
         )
         return True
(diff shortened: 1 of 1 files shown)


src/sentry/feedback/usecases/create_feedback.py
```diff
@@ -200,7 +221,11 @@ def create_feedback_issue(event, project_id: int, source: FeedbackCreationSource
             logger.exception("Error checking if message is spam")
         metrics.incr(
             "feedback.create_feedback_issue.spam_detection",
-            tags={"is_spam": is_message_spam},
+            tags={
+                "is_spam": is_message_spam,
+                "referrer": source.value,
+                "client_source": event["contexts"]["feedback"].get("source"),
```
cmanallen, line 227: What is the cardinality of `feedback.source`?

michellewzhang (commented):
what's the diff between logging just `source.value` and the entire `source`?

aliu39:
> what's the diff between logging just `source.value` and the entire `source`?
 
Instances of a python enum are their own objects, .value gets the value of each variant (a string)

michellewzhang (approved):

PR #76745, opened 2024-08-29
feat(onboarding): frontend to allow members to delete sent invites

ameliahsu:
If members have permission to invite members to an organization (the org's `disable_member_invite` flag is false), we also want them to be able to remove invitations that they sent.

Enabled the Remove button for invites sent by the current user and  updated the tooltip message for invites that were sent by another user:
Screenshot 2024-08-29 at 11 37 04 AM

### DIFF
--- a/static/app/views/settings/organizationMembers/organizationMemberRow.tsx
+++ b/static/app/views/settings/organizationMembers/organizationMemberRow.tsx
@@ -104,7 +104,8 @@ export default class OrganizationMemberRow extends PureComponent<Props, State> {
       canAddMembers,
     } = this.props;
 
-    const {id, flags, email, name, pending, user} = member;
+    const {id, flags, email, name, pending, user, inviterName} = member;
+    const {access} = organization;
 
     // if member is not the only owner, they can leave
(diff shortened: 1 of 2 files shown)


static/app/views/settings/organizationMembers/organizationMemberRow.tsx
```diff
@@ -113,8 +114,14 @@ export default class OrganizationMemberRow extends PureComponent<Props, State> {
     const isCurrentUser = currentUser.email === email;
     const showRemoveButton = !isCurrentUser;
     const showLeaveButton = isCurrentUser;
+    const isInvite = inviterName !== null;
+    const canRemoveInvite =
```
sentaur-athena, line 118: Should this include isInvite?

ameliahsu: No, we only use `isInvite` to determine what disabled message to show the user (line 214). In `canRemoveInvite`, we already check if this is an invite through `inviterName === currentUser.name`.

sentaur-athena (approved):

Issue #75540, opened 2024-08-02
Set substatus to NEW when creating new issues


snigdhas:
Fix the substatus for New issues by explicitly setting the substatus when we create a new group ([code](https://github.com/getsentry/sentry/blob/cb5bb9734c2a4f2004130076850d5339c7c8eed0/src/sentry/event_manager.py#L1538)).

PR #76749, opened 2024-08-29
chore(substatus): Remove logs and substatus changes

### LINKED ISSUE #76754: Remove new substatus override in post_process

snigdhas:
Remove logs for substatus changes in post_process. 

Also removing any logic to set the substatus explicitly to NEW. We should not be overriding the substatus to NEW here since the group [now](https://github.com/getsentry/sentry/issues/75540) gets created with the correct substatus. 

Fixes https://github.com/getsentry/sentry/issues/76754

### DIFF
--- a/src/sentry/tasks/post_process.py
+++ b/src/sentry/tasks/post_process.py
@@ -789,7 +789,7 @@ def update_event_group(event: Event, group_state: GroupState) -> GroupEvent:
 
 
 def process_inbox_adds(job: PostProcessJob) -> None:
-    from sentry.models.group import Group, GroupStatus
+    from sentry.models.group import GroupStatus
     from sentry.types.group import GroupSubStatus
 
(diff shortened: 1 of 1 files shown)


JoshFerge (approved):

src/sentry/tasks/post_process.py
```diff
@@ -815,20 +815,6 @@ def process_inbox_adds(job: PostProcessJob) -> None:
             not is_reprocessed and not has_reappeared
         ):  # If true, we added the .ONGOING reason already
             if is_new:
```
JoshFerge, line 817: are there any tests that need to be updated from this?

snigdhas: nope - any tests that (incorrectly) assumed a new issue had the wrong substatus were updated in https://github.com/getsentry/sentry/pull/75471 where we explicitly set the substatus to NEW. the tests working as is is (hopefully) proof that these lines aren't needed.

PR #76760, opened 2024-08-29
ref(alerts): Add support for dynamic crash rate alert

ceorourke:
Add support for a crash rate alert using anomaly detection. This requires a different snuba query.

Closes https://getsentry.atlassian.net/browse/ALRT-240

### DIFF
--- a/src/sentry/seer/anomaly_detection/store_data.py
+++ b/src/sentry/seer/anomaly_detection/store_data.py
@@ -102,7 +102,6 @@ def send_historical_data_to_seer(alert_rule: AlertRule, project: Project) -> Ale
     if not historical_data:
         raise ValidationError("No historical data available.")
 
-    assert dataset
     formatted_data = format_historical_data(historical_data, dataset)
     if not formatted_data:
         raise ValidationError("Unable to get historical data for this alert.")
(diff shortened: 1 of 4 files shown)


src/sentry/seer/anomaly_detection/store_data.py
```diff
@@ -154,17 +179,54 @@ def fetch_historical_data(
...
-            organization=alert_rule.organization,
-            projects=[project],
-            start=start,
-            end=end,
-        ),
-        rollup=granularity,
-        referrer=Referrer.ANOMALY_DETECTION_HISTORICAL_DATA_QUERY.value,
-        zerofill_results=True,
-    )
+    if dataset == metrics_performance:
```
ceorourke, line 182: This was all pulled from https://github.com/getsentry/sentry/blob/master/src/sentry/api/endpoints/organization_sessions.py

nhsiehgit: would be great if we could separate concerns here a bit and pull this into some helpers?

Maybe add a note or create a ticket to do this as part of https://getsentry.atlassian.net/browse/ALRT-187?

nhsiehgit (approved):
Would prefer if we could find a way to decouple these concerns and dedup the logic 😭 

but if this passes, then we can do those as follow up items and get these in to unblock

src/sentry/seer/anomaly_detection/store_data.py
ceorourke: updated to add `get_crash_free_historical_data()`

PR #76762, opened 2024-08-29
fix(issues): Avoid animating tags on initial load

scttcper:
On page load, the tags animate in and all are rendered regardless of if they're expanded or not. Using AnimatePresence allows us to avoid the initial render animation and skip rendering collapsed tags and their mouse listeners and tooltips.

example of route navigation causing a rerender that causes the tags to animate again even thought the data is unchanged.

https://github.com/user-attachments/assets/cfcda14a-732c-4990-92ef-13d782308e15


edwardgou-sentry (approved):
nice!

PR #76764, opened 2024-08-29
test(ui): Use built-in test providers

scttcper:
just removes duplicate test providers that are included in our `render` function


ryan953 (approved):

priscilawebdev (approved):
👏

PR #76766, opened 2024-08-30
ops(celery): set hard limit for all celery tasks to 3 hours

JoshFerge:
- In #inc-865 we suspect a runaway long celery task created an extremely long database transaction (over an hour). This particular suspect task's duration was nearly 6 hours. 
- Safe to say, if a celery task has ran this long, something has gone very wrong.
- Celery tasks are already killed by deployments, so in normal working hours, our celery tasks are getting killed all the time generally.
- We can likely set lower than 3 hours, to maybe at the lowest 30 minutes, but i figured this would be a good place to start, as we do observe some outlier tasks taking this long, so would want to go down slowly to see if this causes any issues.


mwarkentin:
Does celery give you a way to trigger an event when it times out a task? Could give us a way to monitor for tasks that are hitting timeouts.

wedamija:
> Does celery give you a way to trigger an event when it times out a task? Could give us a way to monitor for tasks that are hitting timeouts.

We have a custom decorator around all task runs, so even if celery doesn't have an event like this, we could catch a timeout here and add whatever kind of monitoring https://github.com/getsentry/sentry/blob/bf507f903240d4ec8534ccac2eba4208782518f4/src/sentry/tasks/base.py#L124-L128

src/sentry/tasks/base.py
```diff
@@ -129,6 +129,13 @@ def _wrapped(*args, **kwargs):
 
             return result
 
+        if kwargs.get("soft_time_limit") and not kwargs.get("time_limit"):
+            raise ValueError("time_limit must be set if soft_time_limit is set")
+
+        if kwargs.get("time_limit") is None and kwargs.get("soft_time_limit") is None:
+            kwargs["soft_time_limit"] = int(timedelta(hours=3).total_seconds())
+            kwargs["time_limit"] = int(timedelta(hours=3, seconds=15).total_seconds())
```
wedamija, line 137: Should this be a config value instead?

wedamija (commented):
Rather than set this in the decorator, I think we can just use the django celery config options to do it.

https://celeryproject.readthedocs.io/zh-cn/latest/userguide/configuration.html#task-time-limit

So I think `CELERY_TASK_TIME_LIMIT` and `CELERY_TASK_SOFT_TIME_LIMIT` in server.py would work here

src/sentry/tasks/base.py
JoshFerge: got it, i wasn't sure about using those. I'll look into briefly and make sure they work, but can certainly use the config values.

JoshFerge:
> > Does celery give you a way to trigger an event when it times out a task? Could give us a way to monitor for tasks that are hitting timeouts.
> 
> We have a custom decorator around all task runs, so even if celery doesn't have an event like this, we could catch a timeout here and add whatever kind of monitoring
> 
> https://github.com/getsentry/sentry/blob/bf507f903240d4ec8534ccac2eba4208782518f4/src/sentry/tasks/base.py#L124-L128

i'll add a catch here and add a timeout sentry warning / metric.

src/sentry/tasks/base.py
```diff
@@ -125,7 +126,11 @@ def _wrapped(*args, **kwargs):
                 metrics.timer(key, instance=instance),
                 track_memory_usage("jobs.memory_change", instance=instance),
             ):
-                result = func(*args, **kwargs)
+                try:
+                    result = func(*args, **kwargs)
+                except SoftTimeLimitExceeded:
+                    capture_exception(level="warning")
```
wedamija, line 132: I think this error will already be caught when we raise, so isn't this double sending now? Would it be more helpful to have a stat instead?

JoshFerge: ah good point. i didn't think to look in Sentry. we indeed already have [them](https://sentry.sentry.io/issues/?project=1&query=SoftTimeLimit&referrer=issue-list&sort=date&statsPeriod=24h&viewId=50632).  i'd rather not introduce another high cardinality metric if possible. i can set up some sentry alerts for this if that works.

wedamija (approved):

PR #76787, opened 2024-08-30
ref(alerts): Remove span metrics alerts

ArthurKnaus:
Remove support for the alert type `span_metrics`.


vgrozdanic:
Is there a reason why: `static/app/views/alerts/rules/metric/spanMetricsField.tsx` is also not removed as a part of this PR?

ArthurKnaus:
> Is there a reason why: `static/app/views/alerts/rules/metric/spanMetricsField.tsx` is also not removed as a part of this PR?

Good catch! We can remove it in this PR 👍

vgrozdanic (approved):

PR #76790, opened 2024-08-30
ref(metrics): Remove span based metrics from dashboard and metrics page

ArthurKnaus:
Remove span based metrics functionality form dashboard and metrics page.
Will remove the `VirtualMetricsContextProvider` + usage in another PR. (As it is a lot)

### DIFF
--- a/static/app/components/metrics/mriSelect/index.tsx
+++ b/static/app/components/metrics/mriSelect/index.tsx
@@ -18,10 +18,7 @@ import {
   isTransactionMeasurement,
 } from 'sentry/utils/metrics';
 import {emptyMetricsQueryWidget} from 'sentry/utils/metrics/constants';
-import {
-  hasCustomMetricsExtractionRules,
-  hasMetricsNewInputs,
-} from 'sentry/utils/metrics/features';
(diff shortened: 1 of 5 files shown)


static/app/components/metrics/mriSelect/index.tsx
```diff
@@ -248,7 +244,7 @@ export const MRISelect = memo(function MRISelect({
         const parsedMRI = parseMRI(metric.mri);
         const isDuplicateWithDifferentUnit = metricsWithDuplicateNames.has(metric.mri);
         const isUnresolvedExtractedMetric = isExtractedCustomMetric(metric);
-        const showProjectBadge = hasExtractionRules && selectedProjects.length > 1;
+        const showProjectBadge = selectedProjects.length > 1;
```
priscilawebdev, line 247: Are you sure we still want to display the project badge? It appears it was only displayed when the feature flag was enabled

ArthurKnaus: The UX is much better with it. I tried the page without it, but it just feels wrong.

priscilawebdev (approved):

PR #76797, opened 2024-09-02
ref(metrics): Remove virtual metrics provider

ArthurKnaus:
Remove `<VirtualMetricsProvider />`
Remove virtual metrics handling from `<QueryBuilder />`


static/app/utils/metrics/virtualMetricsContext.tsx
```diff
@@ -61,304 +52,7 @@ const Context = createContext<ContextType>({
   isLoading: false,
 });
 
+// TODO: Remove this when all dependet code was cleaned up
 export function useVirtualMetricsContext() {
```
ArthurKnaus, line 56: There is still a lot of code accessing the context.
Though this will simply disable all span metric features, if the provider is not available.

obostjancic (approved):

PR #76806, opened 2024-09-02
fix(onboarding) conditionally show minimum version for profiling notice for all python packages

### LINKED ISSUE #76804: [Django] minimum profiling version notice shouldn't be there
https://sentry.rewatch.com/video/k2caoj8sv0v077uj-django-onboarding-walkthrough?t=236

Image

vgrozdanic:
Conditionally show information about minimal version of SDK needed for profiling at the beginning of the onboarding. information should be shown only if the user has opened onboarding docs inside a profile page, while it should be hidden on issues page.

We are showing that info to the user before they could even realise what is profiling, and since they have just installed the latest version, there is no need to mention info about minimum version at this place.

Example how it used to look for one of the python's framworks:
image

Closes https://github.com/getsentry/sentry/issues/76804

### DIFF
--- a/static/app/components/onboarding/gettingStartedDoc/types.ts
+++ b/static/app/components/onboarding/gettingStartedDoc/types.ts
@@ -34,6 +34,11 @@ export type SelectedPlatformOptions<
   [key in keyof PlatformOptions]: PlatformOptions[key]['items'][number]['value'];
 };
 
+export enum DocsPageLocation {
+  ISSUES_PAGE = 0,
+  PROFILING_PAGE = 1,
+}
+
 export interface DocsParams<
   PlatformOptions extends BasePlatformOptions = BasePlatformOptions,
 > {
@@ -54,6 +59,10 @@ export interface DocsParams<
     screenshot?: boolean;
   };
   newOrg?: boolean;
+  /**
+   * The page where the docs are being displayed
+   */
+  page?: DocsPageLocation;
   replayOptions?: {
     block?: boolean;
     mask?: boolean;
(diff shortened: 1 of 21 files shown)


ArthurKnaus:
The onboarding is re-used for the profiling onboarding sidebar, where it makes sense to show the SDK requirements.

vgrozdanic:
@ArthurKnaus this is now fixed, the profiling version text is shown conditionally only on the Profiling page

static/app/components/onboarding/gettingStartedDoc/types.ts
```diff
@@ -54,6 +59,10 @@ export interface DocsParams<
     screenshot?: boolean;
   };
   newOrg?: boolean;
+  /**
+   * The page where the docs are being displayed
+   */
+  page?: DocsPageLocation;
```
ArthurKnaus, line 65: hmm, would choose something more generic than `page`. As e.g. a sidebar may be opened on different pages.
Maybe `docsLocation`?

static/app/components/onboarding/gettingStartedDoc/types.ts
```diff
@@ -34,6 +34,11 @@ export type SelectedPlatformOptions<
   [key in keyof PlatformOptions]: PlatformOptions[key]['items'][number]['value'];
 };
 
+export enum DocsPageLocation {
+  ISSUES_PAGE = 0,
+  PROFILING_PAGE = 1,
+}
```
ArthurKnaus, line 40: Why only issues and profiling?
For now I would restrict it to `PROFILING_SIDEBAR` and add new ones if needed.

ArthurKnaus (approved):
lgtm 👍 
In the future we should find a different mechanism for it that only shows this notice if we detect an older SDK is in use for the project.

static/app/components/onboarding/gettingStartedDoc/types.ts
vgrozdanic: Those are the only 2 pages where it is being used, if it will be added to some other page, this could easily be extended.

I didn't want to make it an optional parameter, since this way is more explicit how it is being used

PR #76811, opened 2024-09-02
ref(onboarding): Add hook to deal with urls param

priscilawebdev:
This PR adds the `useOnboardingQueryParams` hook to manage all URL parameters for the onboarding flow, allowing us to consolidate and simplify the existing code.

**Next**

- Implement a method for users to navigate back from 'manual' to 'integrations'.
- Refactor the code by moving all non-ProductSelection-related functionality to a separate component.

requires https://github.com/getsentry/getsentry/pull/14996


static/app/utils/url/normalizeUrl.tsx
```diff
@@ -18,7 +18,10 @@ const NORMALIZE_PATTERNS: Array<[pattern: RegExp, replacement: string]> = [
     '/settings/$1',
   ],
   [/^\/?join-request\/[^\/]+\/?.*/, '/join-request/'],
-  [/^\/?onboarding\/[^\/]+\/(.*)/, '/onboarding/$1'],
+  [
+    /^\/?onboarding\/(?!setup-docs|select-platform|welcome)[^\/]+\/(.*)/,
+    '/onboarding/$1',
+  ],
```
priscilawebdev, line 24: fixed an issue with the regex that caused the router to redirect to `onboarding/` when the parameter was `setup-docs`, `select-platform`, or `welcome`.

static/app/views/onboarding/components/useOnboardingQueryParams.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const values = ONBOARDING_QUERY_PARAMS.reduce((acc, key) => {
+      const value = location.query[key];
+      acc[key] =
+        value === 'true'
+          ? true
+          : value === 'false'
+            ? false
+            : Array.isArray(value)
+              ? decodeList(value)
+              : value;
```
priscilawebdev, line 46: I couldn't find it, but maybe there is a function to handle this

vgrozdanic: IMHO this is hard to read, if there is no function to handle this, this would be good use case to add one, at least in this file, and we would avoid the use of triple nested ternary operator 😅

priscilawebdev: yes, I don't like this either and was hoping someone could point me to the magical function.

ArthurKnaus: You will probably want to use `useLocationQuery` here.
Also I would just hardcode reading the 3 params. No need to make it more complicated by iterating over them.

priscilawebdev: thanks @ArthurKnaus! I can do that, but considering that we may have more parameters in the future, I would prefer to keep the logic as it is. The list isn't too long at the moment

priscilawebdev: ok but by using `useLocationQuery` it is better as you suggested :)

static/app/views/onboarding/components/useOnboardingQueryParams.tsx
```diff
@@ -0,0 +1,57 @@
...
+   */
+  product?: string[];
+  /**
+   * Used to show the loader script for when the platform is javascript
+   */
+  showLoader?: boolean;
+  /**
+   * Used to show or not the integration onboarding for certain platforms, e.g. AWS (python)
+   */
+  showManualSetup?: boolean;
```
ArthurKnaus, line 20: As there is always a fallback value they can be required, right?
```suggestion
  /**
   * Used to show product selection (error monitoring, tracing, profiling and session replay) for certain platforms, e.g. javascript-react
   */
  product: string[];
  /**
   * Used to show the loader script for when the platform is javascript
   */
  showLoader: boolean;
  /**
   * Used to show or not the integration onboarding for certain platforms, e.g. AWS (python)
   */
  showManualSetup: boolean;
```

ArthurKnaus (approved):

ArthurKnaus:
Great improvement! 🙌

static/app/views/onboarding/components/useOnboardingQueryParams.tsx
priscilawebdev: hm yes for the booleans but for the product there is not yet a default

priscilawebdev: oh yes... good eyes 👀🙂

PR #76818, opened 2024-09-02
ref(whats-new): Revamp "Whats New"

### LINKED ISSUE #77055: Revamp "Whats New"
- Polish the already created frontend code implemented for the Hack Week project:
  - https://github.com/getsentry/sentry/tree/hackweek/whats-new
  - https://github.com/getsentry/getsentry/tree/hackweek/whats-new
- Introduce new styles based on the figma file https://www.figma.com/design/ozvRB7BfnycKOX6odRpqbg/Whats-New-Content-Blocks?node-id=1-5&node-type=FRAME&t=z6OOnxgp31RIu4BJ-0

priscilawebdev:
**Before**
image
image

**After**

Screenshot 2024-09-05 at 22 34 58

Screenshot 2024-09-05 at 22 34 26

closes https://github.com/getsentry/sentry/issues/77055


static/app/types/system.tsx
```diff
@@ -228,17 +228,37 @@ export type PipelineInitialData = {
...
   id: string;
   isActive: boolean;
+  /**
+   * The URL to use for the CTA
+   */
   link: string;
+  /**
+   * Image url
+   */
+  mediaUrl: string;
```
priscilawebdev, line 251: personal note: @priscilawebdev recheck the required types

static/app/components/sidebar/broadcasts.tsx
```diff
@@ -27,6 +29,7 @@ type Props = CommonSidebarProps & {
 
 type State = {
   broadcasts: Broadcast[];
+
```
roggenkemper, line 32: extra new line

static/app/components/sidebar/sidebarPanelItem.tsx
```diff
@@ -3,58 +3,33 @@ import styled from '@emotion/styled';
...
-  title?: string;
-  /**
-   * Actions to the right of the title
-   */
-  titleAction?: React.ReactNode;
-};
+  ctaText?: string;
+}
 
 function SidebarPanelItem({
```
roggenkemper, line 21: is it worth making any changes to `SidebarPanelItem` since that's the old experience?

priscilawebdev: There was a minor improvement in the types, as I noticed we could reuse the ones defined for Broadcast.

As for this component, even though it won't be used in the broadcast experience, it is still being used in forhttps://github.com/getsentry/sentry/blob/3634c69f7ff8b55114f8b69d0ced5b4b96b1077d/static/app/components/sidebar/serviceIncidents.tsx#L60

static/app/components/sidebar/broadcasts.tsx
priscilawebdev, line 32:
```suggestion
```

static/app/components/sidebar/utils.tsx
```diff
@@ -1,8 +1,13 @@
...
 
 export const isDone = (task: OnboardingTaskStatus) =>
   task.status === 'complete' || task.status === 'skipped';
 
 // To be passed as the `source` parameter in router navigation state
 // e.g. {pathname: '/issues/', state: {source: `sidebar`}}
 export const SIDEBAR_NAVIGATION_SOURCE = 'sidebar';
+
+export function hasWhatIsNewRevampFeature(organization: Organization) {
+  return organization.features.includes('what-is-new-revamp');
```
roggenkemper, line 12: two nits here - would prefer we call this `whats-new-revamp` if the flag hasn't been created yet. second, do we really need this util function ? i feel like since it is just a feature flag check we could have it be     `const hasWhatsNewRevamp = organization.features.includes()...;`

static/app/types/system.tsx
```diff
@@ -228,17 +228,38 @@ export type PipelineInitialData = {
...
   link: string;
+  /**
+   * A message with muted styling which appears above the children content
+   */
   message: string;
   title: string;
-};
+  /**
+   * Category of the broadcast.
+   * Synced with https://github.com/getsentry/sentry/blob/923a65508912c3e181e1c70cbdf076b7b956aa90/src/sentry/models/broadcast.py#L14
```
roggenkemper, line 251: can we change the branch here to be master: https://github.com/getsentry/sentry/blob/master/src/sentry/models/broadcast.py

priscilawebdev: I tried that a couple of times, but always when I clicked on the copy button I got a `blob/code`

static/app/components/sidebar/utils.tsx
priscilawebdev: yes initially I thought we would use it in more different places but it makes sense

priscilawebdev: oh and the feature flag is already created

matejminar (approved):

PR #76820, opened 2024-09-02
feat(whats-new): extend broadcast model and serializer

shellmayr:
Adds fields to the existing Broadcast model in Django and the corresponding serializers

Changes are pulled from https://github.com/getsentry/sentry/compare/master...hackweek/whats-new


src/sentry/models/broadcast.py
```diff
@@ -16,13 +16,15 @@ class Broadcast(Model):
...
-    title = models.CharField(max_length=32)
+    title = models.CharField(max_length=64)
     message = models.CharField(max_length=256)
     link = models.URLField(null=True, blank=True)
     is_active = models.BooleanField(default=True, db_index=True)
     date_expires = models.DateTimeField(default=default_expiration, null=True, blank=True)
     date_added = models.DateTimeField(default=timezone.now)
     cta = models.CharField(max_length=256, null=True, blank=True)
+    media_url = models.URLField(null=True, blank=True)
+    category = models.CharField(max_length=32, null=True, blank=True)
```
priscilawebdev, line 27: I don't think this should be `null` but either one of the defined categories

src/sentry/api/validators/broadcast.py
```diff
@@ -12,3 +12,5 @@ class AdminBroadcastValidator(BroadcastValidator):
     isActive = serializers.BooleanField(required=False)
     dateExpires = serializers.DateTimeField(required=False, allow_null=True)
     cta = serializers.CharField(max_length=256, required=True)
+    mediaUrl = serializers.URLField(required=False)
```
roggenkemper, line 15: title should be updated to 64 in this validator too

src/sentry/models/broadcast.py
wedamija: Are we able to backfill existing rows to have a category if we make this not null?

wedamija (commented):
Migration lgtm

src/sentry/models/broadcast.py
shellmayr: Yeah, I would not add a not null constraint here tbh - we can enforce the usage of categories in the frontend and during validation for new entries, adding this on a DB level just makes migration harder for no reason & I don't think we can programmatically backfill this

src/sentry/api/validators/broadcast.py
shellmayr: 👍 good catch, changed

src/sentry/api/serializers/models/broadcast.py
vgrozdanic, line 1: We could use the same serializer for both serialization and validation - the one that was used as validator. In that case there is no need to keep the fields in sync on 2 places, but we would have single source of truth.

There is `source` keyword in DRF serializers that allows us to switch between camelCase responses, and snake_case fields in models

src/sentry/api/endpoints/broadcast_details.py
```diff
@@ -80,6 +80,11 @@ def put(self, request: Request, broadcast_id) -> Response:
             update_kwargs["date_expires"] = result["dateExpires"]
         if result.get("cta"):
             update_kwargs["cta"] = result["cta"]
+        if result.get("mediaUrl"):
+            update_kwargs["media_url"] = result["mediaUrl"]
+        if result.get("category"):
+            update_kwargs["category"] = result["category"]
+
```
vgrozdanic, line 87: If we would use `source` field on the serializer, there would be no need to create our own `update_kwargs` object

src/sentry/api/serializers/models/broadcast.py
shellmayr: Agreed - but would defer this to a later PR- this should use the existing code and just extend it slightly.

src/sentry/api/endpoints/broadcast_details.py
shellmayr: Agreed, but let's defer (see above)

vgrozdanic (approved):

priscilawebdev (approved):
lgtm

PR #76822, opened 2024-09-03
ref(users): move user serializers to new abode

Christinarlong:
Moves all user serializes and tests to the user's directory :0) Fixes any additional typing along the journey ~

ref(issue # here)


Christinarlong:
underestimated the amount of typing so I'm going to separate this out

PR #76823, opened 2024-09-03
feat: Add ISO8601 format to formatDuration

ryan953:
This is a slice from https://github.com/getsentry/sentry/pull/76711 in order to start to shrink that down into specific parts.

### DIFF
--- a/static/app/utils/duration/formatDuration.tsx
+++ b/static/app/utils/duration/formatDuration.tsx
@@ -13,7 +13,11 @@ type Format =
   // example: `01:00:00.000
   | 'hh:mm:ss.sss'
   // example: `01:00:00`
-  | 'hh:mm:ss';
+  | 'hh:mm:ss'
+  // example: `PT4H18M3S`
+  // See https://en.wikipedia.org/wiki/ISO_8601#Durations
+  // See https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#durations
+  | 'ISO8601';
 
 type Args = {
   /**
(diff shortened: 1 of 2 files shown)


static/app/utils/duration/formatDuration.tsx
```diff
@@ -82,6 +86,49 @@ export default function formatDuration({
...
       const [head, tail] = str.split('.');
       return includeMs ? [head, tail ?? '000'].join('.') : String(head);
+    case 'ISO8601':
+      const output = ['P'];
+
+      let incr = 0;
+      const weeks = Math.floor(msToPrecision(ms - incr, 'week'));
+      output.push(weeks ? weeks + 'W' : '');
+      if (precision !== 'week') {
+        incr += weeks * PRECISION_FACTORS.week;
```
scttcper, line 96: wonder how much precision matters here since it isn't really formatted for humans

scttcper (approved):

static/app/utils/duration/formatDuration.tsx
ryan953: Ya, it's mostly for completeness, and the logic is the same as the magnitude goes up so not really any harm imo.

I also think the interface of asking for precision is scalable. For context our existing `getDuration()` and `getExactDuration()` functions are next to merge into here. I think they would both transform into a new format `format="human-full"` and the difference between exact and non-exact will be a matter of asking for the right precision (a helper util will have a place here).
Also, to account for the `abbreviation=true` param another `format="human-abbr"` (might not be the final names, but that idea) can be put in.

PR #76838, opened 2024-09-03
feat(php-onboarding): Add stack trace args warning

ArthurKnaus:
Screenshot 2024-09-03 at 12 25 59


cleptric (approved):

static/app/gettingStartedDocs/php/symfony.tsx
```diff
@@ -27,11 +28,9 @@ const onboarding: OnboardingConfig = {
       configurations: [
         {
           language: 'bash',
-          description: (
-            <p>
```
ArthurKnaus, line 31: Removed all the `<p>` as they were adding additional spacing in-between blocks.

matejminar (approved):

PR #76851, opened 2024-09-03
feat(platform): Add javascript-nuxt platform in BE

s1gr1d:


ArthurKnaus (commented):
You will also want to add it here:
https://github.com/getsentry/sentry/blob/master/src/sentry/utils/platform_categories.py

priscilawebdev (approved):

ArthurKnaus (approved):

PR #76372, opened 2024-08-16
feat(feedback): show issue details' TraceDataSection in feedback details

aliu39:
Fixes https://github.com/getsentry/sentry/issues/68481

For the screenshots, I used dev-ui to override the feedback's trace ID. Before merging, we need to test this by finding/creating some real-world, prod examples.

1. [Trace w/26 issues:](https://sentry.sentry.io/performance/trace/e0078a14f2e34eab85af36dc36fbfc1c/?eventId=e7f2609d56af4b478dda76748f44ebf6&feedbackSlug=javascript%3A5721692549&groupId=5721692549&pageEnd=2024-08-17T10%3A29%3A42.585&pageStart=2024-08-16T10%3A29%3A42.585&query=&referrer=feedback_list_page&source=issue_details&timestamp=1723847311)

trace-ex1

_You can try interacting with the same timeline in [this issue](https://sentry.sentry.io/issues/5545076208/?query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D+invalidsearchquery&referrer=issue-stream&statsPeriod=14d&stream_index=2)_.

2. [Trace w/1 other issue:](https://sentry.sentry.io/performance/trace/a867ab58e5014df5a8e3168549d3fcdc/?eventId=e7f2609d56af4b478dda76748f44ebf6&feedbackSlug=javascript%3A5721692549&groupId=5721692549&pageEnd=2024-08-17T10%3A29%3A42.585&pageStart=2024-08-16T10%3A29%3A42.585&referrer=feedback_list_page&source=issue_details&timestamp=1723847311)

trace-ex2

3. If missing trace, or trace has 0 other issues, _the section is not rendered at all_

PR #76858, opened 2024-09-03
feat(feedback): add trace section to details, with analytics and FF

### LINKED ISSUE #68481: [User Feedback] - Issue Trace Timeline
tl;dr: Add the new trace timeline to User Feedback details

A new component was built for the Issue Details page called [Issue Trace Timeline](https://sentry.io/changelog/issues-trace-timeline/). Previously known as Trace Navigator

image

With the JS SDK v8, each new page load or navigation will create a new trace. That means that if a user submits a User Feedback through the widget, that feedback will be using the same trace-id as other relevant data during that last page load. Including client side errors, as well as calls to the backend that might have be the culprit of that error.

With the JS SDK v7 though, there is a global trace id that wasn't used during page load requests that resulted in HTTP calls to the backend. This means that User Feedbacks from the widget are not likely to be connected to any other telemetry.

Additionally, feedback received from the legacy API (that only takes name/email/comments) doesn't have a trace id. They are connected to an error though (through `event_id`) so we potentially leverage that trace-id. We can do so by coping the `trace_id` from the [event linked in the shim code](https://github.com/getsentry/sentry/blob/1d5adc7369f53bf2e9396b1aba8a0e21f67cd5f1/src/sentry/feedback/usecases/create_feedback.py#L290-L302).

To verify the data, we can use `?dataset=issuePlatform` on Discover. For example [this query](https://sentry.sentry.io/discover/homepage/?dataset=issuePlatform&field=user.display&field=timestamp&field=trace&name=All+Events&project=11276&query=title%3A%22User+Feedback%22&sort=-timestamp&statsPeriod=14d&yAxis=count%28%29). Once v8 is installed in Sentry we'll be able to confirm what data would be visible in the component.

Also note the component might need to be adjusted to show User Feedback. It was designed for Issue Details so it shows other error events only (no spans).

<details>
<summary>Additional context</summary>

## Spans on trace navigator

image

## Trace-id behavior in JS SDK v7 vs v8

image

image
image

</details>

aliu39:
Branched from https://github.com/getsentry/sentry/pull/76372 (at first I thought this branch would be for analytics only, but now it replaces this PR)

Depends on https://github.com/getsentry/sentry/pull/76867

Closes https://github.com/getsentry/sentry/issues/68481


static/app/components/feedback/feedbackItem/feedbackItem.tsx
```diff
@@ -122,6 +133,49 @@ export default function FeedbackItem({feedbackItem, eventData, tags}: Props) {
...
+    if (isError) {
+      metrics.increment('feedback.trace_section.error');
+    } else {
+      metrics.increment('feedback.trace_section.events_loaded', traceEvents.length - 1);
+      if (hasProject && !!crashReportId && oneOtherIssueEvent?.id === crashReportId) {
+        metrics.increment('feedback.trace_section.crash_report_dup');
+      }
+    }
+    setSentMetric(true);
+  }
```
ryan953, line 169: This should be put into a `useEffect(() => { ... }, [])` so it only runs once on mount.

static/app/components/feedback/feedbackItem/feedbackItem.tsx
```diff
@@ -122,6 +133,49 @@ export default function FeedbackItem({feedbackItem, eventData, tags}: Props) {
...
+  crashReportId: string | undefined;
+  eventData: Event;
+  hasProject: boolean;
+}) {
+  // If there's a linked error from a crash report and only one other issue, showing both could be redundant.
+  // TODO: we could add a jest test .spec for this ^
+  const {oneOtherIssueEvent, traceEvents, isLoading, isError} = useTraceTimelineEvents({
+    event: eventData,
+  });
+  const hide: boolean =
```
ryan953, line 150: we don't usually annotate variables like this, it'll be a boolean, or truthy, it's fine.

```suggestion
  const hide =
```

static/app/components/feedback/feedbackItem/feedbackItem.tsx
aliu39: oh yes!

aliu39: I can prob remove the sentMetric flag then?

static/app/components/feedback/feedbackItem/feedbackItem.tsx
ryan953: It might be a good idea to rephrase this as `const show = ...` and flip the logic around. That way the variable is more proactive sounding and the return ternary would be a little easier as `return show && ffEnabled ? <Section>...</Section> : null;`

phrasing conditions like `if (yes && yes) ...` is easier to understand than `if (no || no) ... else ...` if that makes any sense.

static/app/components/feedback/feedbackItem/feedbackItem.tsx
aliu39: Actually think we still need it since the useEffect has a lot of dependencies

ryan953: ya, useEffect will replace all that.

ryan953: > Actually think we still need it since the useEffect has a lot of dependencies

the dependencies should be ok unless the data refreshes very often. otherwise they should be stable each time this renders which means that the effect will only run once when new values appear

michellewzhang (approved):

PR #76861, opened 2024-09-03
deprecate(grouping): Deprecate mobile grouping config

armenzg:
This removes `mobile:2021-02-12` from the list of grouping configs.

Removing such configurations includes the following:
* Removal of the [mobile:2021-04-02.txt](https://github.com/getsentry/sentry/blob/master/src/sentry/grouping/enhancer/enhancement-configs/mobile%402021-04-02.txt) enhancer file since it's not in use
* The removal of all the snapshots associated with it
* The removal of these features (only used in such a config):
  * hierarchical_grouping
  * discard_native_filename
  * use_package_fallback
  * native_fuzzing
  * detect_synthetic_exception_types


armenzg (commented):
Some initial comments.

PR #76864, opened 2024-09-03
fix(issues): Add upcoming release check

roggenkemper:
this pr adds the check for in_upcoming_release so that we don't hit the NotImplemented, will mimic the behavior of in_next_release for now.

### DIFF
--- a/src/sentry/models/groupresolution.py
+++ b/src/sentry/models/groupresolution.py
@@ -144,7 +144,8 @@ class GroupResolution(Model):
         # We still fallback to the older model if either current_release_version was not set (
         # i.e. In all resolved cases except for Resolved in Next Release) or if for whatever
         # reason the semver/date checks fail (which should not happen!)
-        if res_type in (None, cls.Type.in_next_release):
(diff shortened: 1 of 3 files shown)


wedamija (approved):

malwilley (approved):

src/sentry/models/groupresolution.py
```diff
@@ -144,7 +144,8 @@ def compare_release_dates_for_in_next_release(res_release, res_release_datetime,
         # We still fallback to the older model if either current_release_version was not set (
         # i.e. In all resolved cases except for Resolved in Next Release) or if for whatever
         # reason the semver/date checks fail (which should not happen!)
-        if res_type in (None, cls.Type.in_next_release):
+        # todo(roggenkemper): remove upcoming_release check after we know that no group resolutions have it
+        if res_type in (None, cls.Type.in_next_release, cls.Type.in_upcoming_release):
```
JoshFerge, line 148: is there any test we can add for this (as a follow up).

PR #76577, opened 2024-08-26
WIP: Nhsieh/sidenavbarv2

nhsiehgit:
Current state: 

https://github.com/user-attachments/assets/2176ffd2-053e-4f0f-a41e-3c09ee11e1a9

PR #76868, opened 2024-09-03
Sidebar design update

natemoo-re:
This PR updates the sidebar design across the Sentry frontend, behind the `navigation-sidebar-v2` flag. This PR supercedes #76577.

### TODO
- [x] handle alpha / beta badges
- [x] add subnav panel
- [ ] regression tests
- [ ] mobile (maintain previous UI)

### Follow-up
- [ ] overlays (what's new, onboarding) as modal dialogs
- [ ] new mobile designs


static/app/components/sidebar/index.tsx
```diff
@@ -220,17 +223,82 @@ function Sidebar() {
...
       label={<GuideAnchor target="issues">{t('Issues')}</GuideAnchor>}
       to={`/organizations/${organization.slug}/issues/`}
       search="?referrer=sidebar"
       id="issues"
-      hasNewNav={hasNewNav}
     />
   );
+  if (hasNewNav && hasOrganization) {
+    issues = (
+      <SubnavMenu
```
natemoo-re, line 238: Most of the functional markup changes are handled by a new `SubnavMenu` component, which projects children into the secondary sidebar panel as needed

static/app/components/sidebar/sidebarItem.tsx
```diff
@@ -311,9 +319,28 @@ function SidebarItem({
...
+    if (itemQuery && query) {
+      let match = false;
+      for (const key of itemQuery?.split(' ')) {
+        match = query.includes(key);
+        if (!match) break;
+      }
+      return match;
+    }
+    return !itemQuery;
+  }
```
natemoo-re, line 343: This is new "active" matching logic for the Issues subnav which required special consideration of search params

static/app/components/sidebar/subnav.tsx
```diff
@@ -0,0 +1,63 @@
...
+export function SubnavContainer({children}) {
+  const containerRef = useRef<HTMLDivElement>(null);
+  useLayoutEffect(() => {
+    if (!containerRef.current) return;
+    if (containerRef.current.children.length === 0) {
+      document.body.dataset.navLevel = '1';
+    } else {
+      document.body.dataset.navLevel = '2';
+    }
+  });
```
natemoo-re, line 32: Not the cleanest pattern, but since the sidebar uses `position: fixed`, we need an effect that surfaces how many navigation levels are active so we can adjust the static app layout from inside the render tree.

static/app/utils/theme.tsx
```diff
@@ -808,6 +808,8 @@ const commonTheme = {
     collapsedWidth: '70px',
     semiCollapsedWidth: '100px',
     expandedWidth: '220px',
+    v2_width: '80px',
+    v2_panelWidth: '160px',
```
natemoo-re, line 812: Slight adjustments from the spec'd designs to accommodate font rendering differences

static/app/views/issueList/utils.tsx
```diff
@@ -217,3 +217,32 @@ export const FOR_REVIEW_QUERIES: string[] = [Query.FOR_REVIEW];
...
+
+const IssueGroupFilter: Record<IssueGroup, string> = {
+  [IssueGroup.ALL]: '',
+  [IssueGroup.ERRORS_OUTAGES]: 'issue.category:[error,cron,uptime]',
+  [IssueGroup.TRENDS]:
+    'issue.type:[profile_function_regression,performance_p95_endpoint_regression,performance_n_plus_one_db_queries]',
+  [IssueGroup.CRAFTSMANSHIP]:
+    'issue.category:replay issue.type:[performance_n_plus_one_db_queries,performance_n_plus_one_api_calls,performance_consecutive_db_queries,performance_render_blocking_asset_span,performance_uncompressed_assets,profile_file_io_main_thread,profile_image_decode_main_thread,profile_json_decode_main_thread,profile_regex_main_thread]',
+  [IssueGroup.SECURITY]: 'event.type:[nel,csp]',
+};
```
natemoo-re, line 237: Worth discussing if these should be surfaced in the URL as raw search params under `query` or handled as a new construct (with a proper `group` param, maybe?). If these are intended to be hidden from the user search bar, a new construct would provide a more seamless user experience.

natemoo-re: After discussing in person, we decided to consider this version as a prototype of the new experience that might possibly roll into custom view updates in the future.

static/app/components/sidebar/index.tsx
```diff
@@ -628,167 +697,187 @@ function Sidebar() {
...
-          </DropdownSidebarSection>
-
-          <PrimaryItems>
-            {hasOrganization && (
-              <Fragment>
-                <SidebarSection hasNewNav={hasNewNav}>
-                  {issues}
-                  {projects}
-                </SidebarSection>
+    <SidebarWrapper>
```
nhsiehgit, line 704: Ah i see
iiuc, we're using the `SubnavContainer` component here as the main panel? (then in the children determines, whether to render a secondary panel OR the accordion?)

Do we need this wrapper anymore then or can we move the styles here to the container/primary wrapper?

Also - wonder if we should rename the `SubnavContainer` to just be `NavContainer` 😅

static/app/components/sidebar/index.tsx
```diff
@@ -628,167 +697,187 @@ function Sidebar() {
...
-                      {discover2}
-                      {dashboards}
-                      {releases}
-                      {userFeedback}
-                    </SidebarSection>
-                  </Fragment>
-                )}
+            {hasOrganization && (
+              <SidebarSectionGroup hasNewNav={hasNewNav}>
+                {/* What are the onboarding sidebars? */}
```
nhsiehgit, line 783: oops - was this a comment i left?? 😅 
can probably remove

static/app/components/sidebar/index.tsx
```diff
@@ -840,11 +939,20 @@ export const SidebarWrapper = styled('nav')<{collapsed: boolean; hasNewNav?: boo
...
+  width: ${p => p.theme.sidebar.v2_panelWidth};
+  height: 100%;
+`;
+
 const SidebarSectionGroup = styled('div')<{hasNewNav?: boolean}>`
   ${responsiveFlex};
   flex-shrink: 0; /* prevents shrinking on Safari */
-  gap: 1px;
-  ${p => p.hasNewNav && `align-items: center;`}
+  gap: 8px;
```
nhsiehgit, line 954: do we have any consts for this? maybe `space(1)` or something?

static/app/components/sidebar/sidebarItem.tsx
```diff
@@ -311,9 +319,28 @@ function SidebarItem({
 }
 
 export function isItemActive(
-  item: Pick<SidebarItemProps, 'to' | 'label'>,
+  item: Pick<SidebarItemProps, 'to' | 'label' | 'search'>,
   exact?: boolean
 ): boolean {
+  // issue subnav matches against issue groups in search params
+  if (location.pathname.startsWith('/issues/') && item?.to?.includes('/issues/')) {
```
nhsiehgit, line 326: Will this only be relevant for issues subnav?
Will we ever need to implement logic for other routes?

wondering about scalability and how we might structure this or break this out to be more composable 🤷

static/app/components/sidebar/subnav.tsx
nhsiehgit: heh
would there ever be a world where we have more layers than just 2?

nhsiehgit (approved):
buncha nit comments!

static/app/components/sidebar/subnav.tsx
natemoo-re: Good question! I don't imagine there would be, so this could be treated as a `boolean` attribute or class name.

static/app/components/sidebar/index.spec.tsx
```diff
@@ -436,4 +420,39 @@ describe('Sidebar', function () {
...
     });
   });
+
+  describe('Stacked navigation with navigation-sidebar-v2 flag', () => {
+    it('renders secondary subnav for issues', async function () {
+      renderSidebar({
+        organization: {...organization, features: ['navigation-sidebar-v2']},
+      });
+
+      expect(await screen.findByText('Issues')).toBeInTheDocument();
```
evanpurkhiser, line 430: Should it be `findByRole('link', {name: 'Issues'})`? Or is it a button `findByRole('button', {name: 'Issues'})`

static/app/views/replays/detail/layout/utils.tsx
ryan953, line 1: The profiling flamechart might do something similar to this. I remember they had a full-screen view first then replay did something similar afterwards. 
I'm not confident that old code survived or not, but that'd be a view to double-check.

natemoo-re:
Closing in favor of #77665. The new PR should be much easier to review, since it introduces entirely new components for the navigation.

natemoo-re:
Closing in favor of #77665. The new PR should be much easier to review, since it introduces entirely new components for the navigation.

PR #76876, opened 2024-09-03
ref(onboarding): kotlin language changes

michellewzhang:
prism supports `kotlin`: https://github.com/PrismJS/prism/blob/gh-pages/components/prism-kotlin.js


romtsn (approved):

priscilawebdev (approved):
tested the changes and it looks good. good catch @michellewzhang 👏

PR #76879, opened 2024-09-03
feat(explore): Add group by to explore toolbar

Zylphrex:
This adds a basic way for users to add group bys by selecting span attributes.


Abdkhan14 (approved):
lgtm 🚢

PR #76880, opened 2024-09-03
fix(integrations): Add support for CODEOWNERS sections

### LINKED ISSUE #9565: CODEOWNERS file section not handled
### Core or SDK?

Platform/SDK

### Which part? Which one?

SAAS (sentry.io)

### Description

According to the documentation on [ownership-rules](https://docs.sentry.io/product/issues/ownership-rules//#code-owners), there's no exceptions concerning codeowners [sections](https://docs.gitlab.com/ee/user/project/codeowners/reference.html#sections). So I expected it to and assign the rules accordingly.

However, after going through with the implementation, only the lines where the gitlab handle was mentioned were transformed into rules.

Here's a quick breakdown of what my CODEOWNERS file looks like :

```markdown
[section-name] @my-team-handle
not_implemented/
implemented/ @my-team-handle
```

 

### Suggested Solution

After looking at the [source code](https://github.com/getsentry/sentry/blob/master/src/sentry/ownership/grammar.py#L351), it seems this is an expected behaviour.
Would you please update the documentation accordingly

LinardsLaugalis:
Add option to use GitLab CODEOWNERS section syntax - by having one/multiple section owners.

Fixes https://github.com/getsentry/sentry-docs/issues/9565

### Legal Boilerplate

### DIFF
--- a/src/sentry/ownership/grammar.py
+++ b/src/sentry/ownership/grammar.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import re
-from collections import namedtuple
+from collections import OrderedDict, namedtuple
 from collections.abc import Callable, Iterable, Mapping, Sequence
 from typing import Any, NamedTuple
 
@@ -13,6 +13,7 @@ from rest_framework.serializers import ValidationError
 from sentry.eventstore.models import EventSubjectTemplateData
 from sentry.integrations.models.repository_project_path_config import RepositoryProjectPathConfig
 from sentry.models.organizationmember import OrganizationMember
+from sentry.ownership.section_line import SectionLine
 from sentry.types.actor import Actor, ActorType
 from sentry.users.services.user.service import user_service
 from sentry.utils.codeowners import codeowners_match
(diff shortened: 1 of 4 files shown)


wo0dyn:
Thanks for the PR @LinardsLaugalis! 🙌🏼 — I'm looking forward to using it soon. 🤞🏼

tests/sentry/ownership/test_grammar.py
```diff
@@ -920,20 +929,39 @@ def test_convert_codeowners_syntax():
...
-                "nisanthan.nanthakumar@sentry.io": "nisanthan.nanthakumar@sentry.io",
-            },
+            associations,
             code_mapping,
         )
         == "\n# cool stuff comment\ncodeowners:*.js front-sentry nisanthan.nanthakumar@sentry.io\n# good comment\n\n\ncodeowners:webpack://docs/* docs-sentry ecosystem\ncodeowners:src/sentry/* anotheruser@sentry.io\ncodeowners:api/* nisanthan.nanthakumar@sentry.io\n"
     )
 
 
+def test_convert_codeowners_multiple_sections_with_overrides():
```
cvxluo, line 939: more test cases would be great, such as:
- test handling inline comments
- test malformed sections (e.g. `[Some Section`)
- order preserving (fileA defined multiple times in a file should have the same order in the result)

src/sentry/ownership/section_line.py
cvxluo, line 1: why define this as a custom class? it seems like we can break these out as 3 utility functions (e.g. get_dict_key only needs `path`, `has_valid_path`, `original_line`, and doesn't need `is_preserved_comment`)

src/sentry/ownership/grammar.py
```diff
@@ -408,29 +409,49 @@ def convert_codeowners_syntax(
...
-        # Check if path has '#' not as first character
-        # Check if path contains '!'
-        # Check if path has a '[' followed by a ']'
-        if re.search(r"(\[([^]^\s]*)\])|[\s!#]", path):
+        section_line = SectionLine(rule, path, list(code_owners), section_owners)
+        if section_line.should_skip():
             continue
 
-        sentry_assignees = []
+        section_lines[section_line.get_dict_key()] = section_line
```
cvxluo, line 437: edge case, but would we lose duplicated paths with this, e.g. 
```
[Section]
fileA @team1
fileA @team2
```

will we be defining 2 codeowners lines for this:
```
codeowners:fileA team1
codeowners:fileA team2
```
?

cvxluo (commented):
apologies for the delay in review — left a few comments but generally looks good to support sections. 

however, it seems like there are quite a few differences between Github and Gitlab supported syntax for CODEOWNERS (e.g. Github doesn't support exclusions, but Gitlab does, Github supports inline comments, Gitlab doesn't). it makes this harder to merge, since it makes end-user experience more confusing — we'd be supporting a hybrid between the two syntaxes. going to discuss with the team on how to move forward.

cvxluo:
discussed with team and we're going to close this for now — we'd likely need a completely distinct implementation to support GitLab syntax, since there are a number of differences between GitLab and GitHub CODEOWNERS syntax support. I'll make the docs change to close https://github.com/getsentry/sentry-docs/issues/9565, to specify that we only support GitHub syntax right now.

PR #76887, opened 2024-09-03
ref(replay): don't show mobile onboarding if org can't ingest video replays

michellewzhang:
if the org can't ingest mobile replays, just show this banner for mobile platforms:

SCR-20240903-mhxb

### DIFF
--- a/static/app/components/replaysOnboarding/sidebar.tsx
+++ b/static/app/components/replaysOnboarding/sidebar.tsx
@@ -10,6 +10,7 @@ import {CompactSelect} from 'sentry/components/compactSelect';
 import RadioGroup from 'sentry/components/forms/controls/radioGroup';
 import IdBadge from 'sentry/components/idBadge';
 import LoadingIndicator from 'sentry/components/loadingIndicator';
+import {MobileBetaBanner} from 'sentry/components/onboarding/gettingStartedDoc/utils';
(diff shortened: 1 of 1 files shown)


c298lee (approved):

static/app/components/replaysOnboarding/sidebar.tsx
```diff
@@ -350,6 +351,25 @@ function OnboardingContent({
     );
   }
 
+  // if the org cannot ingest mobile replay events, don't show the onboarding
+  // TODO: remove once we GA mobile replay
+  if (organization.features.includes('session-replay-video-disabled')) {
```
cmanallen, line 356: `false` means the org can ingest mobile replays.

cmanallen (approved):

static/app/components/replaysOnboarding/sidebar.tsx
michellewzhang: yup that's the intended result! if the flag is true we want to show these banners

cmanallen: Nevermind. Its done correctly.

c298lee:
what do we show if the language isn't apple, android or react native?

michellewzhang:
> what do we show if the language isn't apple, android or react native?

then it will be not allowed to be selected in the dropdown list

bruno-garcia (approved):

ryan953 (approved):

PR #76890, opened 2024-09-03
chore(grouping): Use optimized grouping logic for 80% of non-transitioning projects

lobsterkatie:
This is a follow-up to https://github.com/getsentry/sentry/pull/76614, https://github.com/getsentry/sentry/pull/76679, and https://github.com/getsentry/sentry/pull/76758, enabling the updated grouping transition logic for another 20% of non-transitioning projects.

### DIFF
--- a/src/sentry/grouping/ingest/config.py
+++ b/src/sentry/grouping/ingest/config.py
@@ -99,5 +99,5 @@ def project_uses_optimized_grouping(project: Project) -> bool:
             project.organization,
         )
         or (is_in_transition(project))
-        or project.id % 5 < 3  # 60% of all non-transition projects
+        or project.id % 5 < 4  # 80% of all non-transition projects
     )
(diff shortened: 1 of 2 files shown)


src/sentry/grouping/ingest/config.py
```diff
@@ -99,5 +99,5 @@ def project_uses_optimized_grouping(project: Project) -> bool:
             project.organization,
         )
         or (is_in_transition(project))
-        or project.id % 5 < 3  # 60% of all non-transition projects
+        or project.id % 5 < 4  # 60% of all non-transition projects
```
jangjodi, line 102: small nit: change the comment to 80%

jangjodi (approved):

src/sentry/grouping/ingest/config.py
lobsterkatie: Yup, I realized that as soon as I came back to look at it after pasting the link in slack. Fixed.

PR #76896, opened 2024-09-03
ref(users): move user & user_identity_config & auth serializers to new abode

Christinarlong:
Moves 
- User
- User_Identity_Config
- Authenticator
and their tests to the user's directory :0) Fixes any additional typing along the journey (mainly serializer response dicts) ~

split of (#76822)
ref(issue # here)


src/sentry/api/serializers/models/user.py
Christinarlong, line 1: Replaced file with import shim for any getsentry refs

src/sentry/users/api/serializers/authenticator.py
```diff
@@ -1,15 +1,55 @@
...
+
 from sentry.api.serializers import Serializer, register
 from sentry.auth.authenticators.base import AuthenticatorInterface
 from sentry.auth.authenticators.recovery_code import RecoveryCodeInterface
 from sentry.auth.authenticators.sms import SmsInterface
 from sentry.auth.authenticators.totp import TotpInterface
 from sentry.auth.authenticators.u2f import U2fInterface
 
+if TYPE_CHECKING:
+    from django.utils.functional import _StrOrPromise
```
Christinarlong, line 15: From: https://github.com/typeddjango/django-stubs/issues/1804
is it ok if we potentially return/serialize a promise?

src/sentry/users/api/serializers/user.py
```diff
@@ -0,0 +1,401 @@
...
+from sentry.users.models.userpermission import UserPermission
+from sentry.users.models.userrole import UserRoleUser
+from sentry.users.services.user import RpcUser
+from sentry.utils.avatar import get_gravatar_url
+
+T = TypeVar("T", bound=Model)
+
+
+def manytoone_to_dict(
+    queryset: QuerySet[T], key: str, filter_func: Callable[[Any], bool] | None = None
```
Christinarlong, line 39: made this generic-ish

src/sentry/users/api/serializers/user.py
```diff
@@ -0,0 +1,401 @@
...
+    username: str
+    email: str
+    avatarUrl: str
+    isActive: bool
+    hasPasswordAuth: bool
+    isManaged: bool
+    dateJoined: datetime
+    lastLogin: datetime | None
+    has2fa: bool
+    lastActive: datetime | None
```
Christinarlong, line 107: The model says that `lastLogin` and `lastActive` are nullable fields but the openAPI spec says these are required fields, not sure if there's a source of truth/who to follow for this response

src/sentry/users/api/serializers/user.py
```diff
@@ -0,0 +1,401 @@
...
+                        "name": i.auth_provider.get_provider().name,
+                    },
+                    "dateSynced": i.last_synced,
+                    "dateVerified": i.last_verified,
+                }
+                for i in attrs["identities"]
+            ]
+
+        return d
+
```
Christinarlong, line 257: I got rid of the `DetailedUserSerializerResponse`  [here](https://github.com/getsentry/sentry/blob/891194fbacb613283c972a8e64b1d1500c1c25e9/src/sentry/api/serializers/models/user.py#L252-L254) as it wasn't really adding anything new

src/sentry/users/api/serializers/user.py
Christinarlong: I definitely interpreted the OpenApi failing job wrong, as I didn't realize the OpenApi spec is generated from the model serializer which I guess didn't exist until now (?????)

markstory: That job fails anytime there is a delta which isn't great. I took a look at the output and because we've added more accurate types schema generation has better metadata to use and the schema has changed. The properties were always nullable, but our schema didn't reflect that because of limited information.

src/sentry/users/api/serializers/authenticator.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+for interface in RecoveryCodeInterface, TotpInterface, U2fInterface:
+    register(interface)(AuthenticatorInterfaceSerializer)
+
+
+def get_interface_serializer(interface: AuthenticatorInterface) -> AuthenticatorInterfaceSerializer:
+    if isinstance(interface, SmsInterface):
+        return SmsInterfaceSerializer()
+    return AuthenticatorInterfaceSerializer()
```
Christinarlong, line 100: In the endpoints that are given an `ApplicationInterface`, interface, to serialize, this 'getter' is used to make sure we're using the right serializer. It's a weird scenario because registering the serializer in line 76 and 94 should solve this for us, but because we also have that bug of the file not being imported and therefore the serializers not being registered, this redundant function is here.....

Maybe other alternatives could be having barrel file for just the authenticator ?

src/sentry/users/api/serializers/authenticator.py
asottile-sentry: this is actually fine -- `_StrOrPromise` is ~kinda public and typing-only

src/sentry/users/api/serializers/authenticator.py
```diff
@@ -0,0 +1,100 @@
...
+from sentry.auth.authenticators.sms import SmsInterface
+from sentry.auth.authenticators.totp import TotpInterface
+from sentry.auth.authenticators.u2f import U2fInterface
+from sentry.users.models.user import User
+
+
+class EnrolledAuthenticatorInterfaceSerializerResponse(TypedDict, total=False):
+    authId: str
+    createdAt: datetime
+    lastUsedAt: datetime | None
```
markstory, line 17: Another way to solve this in future changes would be to use `typing.NotRequired` to mark these keys as optional.

src/sentry/users/api/serializers/authenticator.py
markstory: I don't hate the factory function. In the longer term, I think we should consider removing the `@register()` calls as they contribute to import tangle and import side-effects create hidden dependencies between modules.

> It's a weird scenario because registering the serializer in line 76 and 94 should solve this for us, but because we also have that bug of the file not being imported and therefore the serializers not being registered, this redundant function is here.....

If the endpoint imports the `sentry.users.api.serializers.authenticators` module I would expect that the `register()` calls would behave correctly. Is that not the case?

src/sentry/users/api/serializers/user.py
```diff
@@ -0,0 +1,388 @@
...
+
+class UserSerializerResponseOptional(TypedDict, total=False):
+    # NOTE: There is a bug here where trying to move these fields to
+    # UserSerializerResponse and using NotRequired. "identities" is marked as
+    # required for places where UserSerializerResponse is used as a field (e.g
+    # OrganizationMemberResponse).
+    identities: list[_Identity]
+    avatar: SerializedAvatarFields
+    authenticators: list[Any]  # TODO: find out what type this is
```
markstory, line 78: Interesting that we have a special serialization format for authenticators here that is different from the representation that authenticator endpoints use. 🤔

src/sentry/utils/serializers.py
```diff
@@ -0,0 +1,18 @@
+from collections import defaultdict
+from collections.abc import Callable, MutableMapping
+from typing import Any, TypeVar
+
+from django.db.models import Model, QuerySet
+
+T = TypeVar("T", bound=Model)
+
+
+def manytoone_to_dict(
```
markstory, line 10: :clap:

markstory (approved):
Looks good to me.

src/sentry/users/api/serializers/authenticator.py
asottile-sentry: actually that's a good point -- is this actually optional and nullable? (rarely something is)

Christinarlong: @asottile-sentry like are the "enrolled" fields optional & nullable ? I believe so, because they are some test cases like `tests/sentry/users/api/endpoints/test_user_authenticator_enroll.py` where the enrollment in an authenticator interface failed, so the returned user doesn't have any of these fields. Or was something else nullable ? ...

asottile-sentry: as written now there's these possibilities:
```
{}
{"lastUsedAt": some_datetime}
{"lastUsedAt": None}
```

which is usually a bug

src/sentry/users/api/serializers/user.py
Christinarlong: yeahh..... the `user_authenticator_enroll.py` endpoint also has it's own serializer /format for each of the interfaces that's different. AuthenticatorInterface serialization in general is like p. inconsistent ;-;

src/sentry/users/api/serializers/authenticator.py
Christinarlong: yeah I dunno, from what I'm seeing from our tests
`{}` -> is a real scenario [it seems like this can happen when enrollment fails], so Optional seems to be valid
`{"lastUsedAt": some_datetime}` -> also is a real scenario when we mark an authenticator as used.
`{"lastUsedAt": None}` -> also is real when a user is newly enrolled

It looks like the three scenarios all happen so probably not a bug, albeit also somewhat confusing to deal with

PR #76898, opened 2024-09-03
fix(crons): Correct cursor behavior upon filtering on monitor overview page

### LINKED ISSUE #76862: Crons cursor isn't clearing/sticking properly when applying date or ownership filters
Our `cursor` query param is not working as expected when applying the ownership filter or the date filter. Problems are:

- Navigating to a second page and then applying the ownership filter causes the `cursor` to stick around, returning an incorrect empty list of monitors. The `cursor` should clear when updating the ownership filter.
- Navigating to a second page and then applying a date filter causes the `cursor` to clear, navigating the user back to the first page. This is bad given I could be applying the date filters in the timeline itself. We should keep the `cursor` filter.

### Steps to reproduce

1. Go to Crons list page
2. Make sure no ownership filter is applied
3. Paginate to second page (query param `cursor=` will be added)
4. Apply an ownership filter to a team that you know has Crons
5. No results will be returned, even though the team does have monitors

It's similar to the date filter issue, but in reverse.

### Proposed Solution

We should map this behavior for our filters:

- Updating ownership filter: clear out `cursor`
- Updating environment filter: clear out `cursor`
- Updating project filter: clear out `cursor`
- Updating date filter: keep `cursor`

davidenwang:
Corrects two behaviors to reach the following state:

1. When changing the owner, reset the cursor (since changing the owner filter affects the list of monitors returned)
2. When changing the time field, does NOT reset the cursor (since changing the date time should not affect the quantity of results returned)

Fixes: https://github.com/getsentry/sentry/issues/76862

### DIFF
--- a/static/app/views/monitors/overview.spec.tsx
+++ b/static/app/views/monitors/overview.spec.tsx
@@ -1,0 +1,76 @@
+import {TeamFixture} from 'sentry-fixture/team';
+
+import {initializeOrg} from 'sentry-test/initializeOrg';
+import {render, screen, userEvent} from 'sentry-test/reactTestingLibrary';
+
+import OrganizationStore from 'sentry/stores/organizationStore';
+import {useNavigate} from 'sentry/utils/useNavigate';
+import Monitors from 'sentry/views/monitors/overview';
+
+jest.mock('sentry/utils/useNavigate', () => ({
+  useNavigate: jest.fn(),
+}));
+
+const mockUseNavigate = jest.mocked(useNavigate);
+const mockNavigate = jest.fn();
+mockUseNavigate.mockReturnValue(mockNavigate);
+
+describe('Monitors Overview', function () {
+  const team = TeamFixture();
+
+  beforeEach(function () {
+    OrganizationStore.init();
+    MockApiClient.clearMockResponses();
+    MockApiClient.addMockResponse({
+      url: '/organizations/org-slug/monitors/',
+      body: [],
+    });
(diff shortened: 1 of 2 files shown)


static/app/views/monitors/overview.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+  it('correctly filters on owner', async function () {
+    const {organization, router} = initializeOrg({
+      router: {location: {query: {cursor: 'test-cursor'}}},
+    });
+    OrganizationStore.onUpdate(organization);
+
+    render(<Monitors />, {organization, router});
+    expect(await screen.findByText('Cron Monitors')).toBeInTheDocument();
+
+    await userEvent.click(screen.getByText('Filter Owners'));
```
evanpurkhiser, line 63: Possible to use a more accessible selector here?

static/app/views/monitors/overview.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+    await userEvent.click(screen.getByText('#team-slug'));
```
evanpurkhiser, line 64: I think there's like a `radiomenuitem` or some role like that for this?

evanpurkhiser (approved):

static/app/views/monitors/overview.spec.tsx
davidenwang: good catches, fixing it up

PR #76907, opened 2024-09-04
feat(issues): Add trace view component

scttcper:
Adds a trace view component for all events, not just performance issues

Motivation:
People keep asking to see the trace view within issue details and this will allow us experiment internally with having the trace preview available.

image

### DIFF
--- a/tests/js/setupFramework.ts
+++ b/tests/js/setupFramework.ts
@@ -31,6 +31,11 @@ failOnConsole({
       return true;
     }
 
+    // TODO: Remove this after updating jsdom, currently it cannot handle @container queries
+    if (/Error: Could not parse CSS stylesheet/.test(errorMessage)) {
+      return true;
+    }
+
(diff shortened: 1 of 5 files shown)


tests/js/setupFramework.ts
```diff
@@ -31,6 +31,11 @@ failOnConsole({
       return true;
     }
 
+    // TODO: Remove this after updating jsdom, currently it cannot handle @container queries
+    if (/Error: Could not parse CSS stylesheet/.test(errorMessage)) {
+      return true;
+    }
```
scttcper, line 37: avoiding mocking console.error. but it does appear jsdom throws an error because of container queries.

https://github.com/getsentry/sentry/blob/master/static/app/views/performance/newTraceDetails/trace.spec.tsx#L478-L479

roggenkemper (approved):

Issue #76888, opened 2024-09-03
Ensure updating to IGNORED sets substatus in `.update` 


snigdhas:
https://github.com/getsentry/sentry/blob/master/src/sentry/api/helpers/group_index/update.py#L604

PR #76908, opened 2024-09-04
chore(substatus): Infer substatus for ignored groups

snigdhas:
The `.update()` method handles all sorts of updates to a Group, including status updates. Since this is used when handling API requests, we could possibly be receiving requests to update the status to IGNORED without a specified substatus. This change will allow us to infer the substatus based on the other information provided to avoid a breaking change for customers. 

https://github.com/getsentry/sentry/issues/76888


src/sentry/api/helpers/group_index/update.py
```diff
@@ -584,8 +591,22 @@ def update_groups(
...
             SUBSTATUS_UPDATE_CHOICES[result.get("substatus")] if result.get("substatus") else None
         )
-        if new_substatus is None and new_status == GroupStatus.UNRESOLVED:
+        if new_status == GroupStatus.IGNORED and new_substatus is None:
+            if status_details.get("untilEscalating"):
+                new_substatus = GroupSubStatus.UNTIL_ESCALATING
+            elif any(status_details.get(key) is not None for key in IGNORED_CONDITION_FIELDS):
+                new_substatus = GroupSubStatus.UNTIL_CONDITION_MET
+            else:
+                new_substatus = GroupSubStatus.FOREVER
```
mrduncan, line 600: I may have missed them, but can you make sure we have tests for these `status_details` cases (`untilEscalating` plus the 6 `IGNORED_CONDITION_FIELDS`).

mrduncan (approved):
Logic LGTM

Left one note on tests - pulling this out into a separate method might make this a bit more readable/independently testable as well, ex something along the lines of: `new_status, new_substatus = _get_new_status(result, status_details)`

PR #76914, opened 2024-09-04
fix(onboarding): Additional info about feedback integration for remix

vgrozdanic:
Added additional info about where to add sentry's feedback integration - it should be added only to a client side.

image

### DIFF
--- a/static/app/gettingStartedDocs/javascript/remix.tsx
+++ b/static/app/gettingStartedDocs/javascript/remix.tsx
@@ -202,9 +202,20 @@ const feedbackOnboarding: OnboardingConfig = {
           ],
         },
       ],
-      additionalInfo: crashReportCallout({
-        link: 'https://docs.sentry.io/platforms/javascript/guides/remix/user-feedback/#user-feedback-api',
-      }),
+      additionalInfo: (
+        <Fragment>
+          <p>
+            {tct(
+              'Note: The Feedback integration only needs to be added to your [entryClient:entry.client.tsx] file.',
+              {entryClient: <code />, sentryServer: <code />}
+            )}
+          </p>
+
+          {crashReportCallout({
+            link: 'https://docs.sentry.io/platforms/javascript/guides/remix/user-feedback/#user-feedback-api',
+          })}
+        </Fragment>
+      ),
     },
   ],
   verify: () => [],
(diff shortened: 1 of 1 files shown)


ArthurKnaus:
Wouldn't it be enough to set the filename as header in the code snippet?
Screenshot 2024-09-04 at 13 18 22

static/app/gettingStartedDocs/javascript/remix.tsx
```diff
@@ -202,9 +202,20 @@ const feedbackOnboarding: OnboardingConfig = {
...
       ],
-      additionalInfo: crashReportCallout({
-        link: 'https://docs.sentry.io/platforms/javascript/guides/remix/user-feedback/#user-feedback-api',
-      }),
+      additionalInfo: (
+        <Fragment>
+          <p>
+            {tct(
+              'Note: The Feedback integration only needs to be added to your [entryClient:entry.client.tsx] file.',
+              {entryClient: <code />, sentryServer: <code />}
```
priscilawebdev, line 210: also we no longer need 2x `<code />` thanks to @ArthurKnaus https://github.com/getsentry/sentry/pull/76721 :)

vgrozdanic: Cool, didn't know that was already merged

vgrozdanic:
@ArthurKnaus for some users it would, but I have went through this onboarding and for a users like me it would help if there is a note that you should only do it on the client side, and not on the server side. We have similar note for a lot of other frameworks - e.g. NextJS

static/app/gettingStartedDocs/javascript/remix.tsx
```diff
@@ -136,7 +136,7 @@ const replayOnboarding: OnboardingConfig = {
         {
           code: [
             {
-              label: 'JavaScript',
+              label: 'entry.client.tsx',
```
vgrozdanic, line 139: Unrelated to feedback, but since we have similar note, I have also changed the name of the file/label to point to the client file

priscilawebdev (approved):
looks good to me

ArthurKnaus (approved):
Lgtm 👍 
Maybe we can also streamline the wording with the other platforms that display the notice?

PR #76931, opened 2024-09-04
ref: squash index_together operation for ProjectDebugFile model

asottile-sentry:

we will need a hard stop for self-hosted (already done!)


joshuarli (approved):
love 2 django

wedamija (approved):

PR #76932, opened 2024-09-04
ref: squash index_together operation for GroupRelease model

asottile-sentry:


joshuarli (approved):

src/sentry/migrations/0001_squashed_0484_break_org_member_user_fk.py
```diff
@@ -6762,26 +6770,6 @@ class Migration(CheckedMigration):
...
         ),
-        migrations.SeparateDatabaseAndState(
-            database_operations=[
-                migrations.RunSQL(
-                    sql="\n                    CREATE INDEX CONCURRENTLY IF NOT EXISTS sentry_grouprelease_group_id_first_seen_53fc35ds\n                    ON sentry_grouprelease USING btree (group_id, first_seen);\n                    ",
-                    reverse_sql="\n                    DROP INDEX CONCURRENTLY IF EXISTS sentry_grouprelease_group_id_first_seen_53fc35ds;\n                    ",
-                    hints={"tables": ["sentry_grouprelease"]},
-                ),
-                migrations.RunSQL(
-                    sql="\n                    CREATE INDEX CONCURRENTLY IF NOT EXISTS sentry_grouprelease_group_id_last_seen_g8v2sk7c\n                    ON sentry_grouprelease USING btree (group_id, last_seen DESC);\n                    ",
```
asottile-sentry, line 6773: actually I think there's something ever-so-subtle here -- the `last_seen` field should be descending but that's not represented in django state 🤔

asottile-sentry: ok I fixed this, and now there's a new migration -- hopefully this is ok but would appreciate another look from @getsentry/owners-migrations

src/sentry/migrations/0754_grouprelease_represented_in_django.py
```diff
@@ -0,0 +1,38 @@
...
+        migrations.RemoveIndex(
+            model_name="grouprelease",
+            name="sentry_grou_group_i_f10abe_idx",
+        ),
+        migrations.AddIndex(
+            model_name="grouprelease",
+            index=models.Index(
+                fields=["group_id", "-last_seen"], name="sentry_grou_group_i_b6e502_idx"
+            ),
+        ),
```
wedamija, line 37: Let's switch the order of these operations around so that we're never missing index coverage. Especially important since it'll take a while to create this index.

src/sentry/migrations/0754_grouprelease_represented_in_django.py
```diff
@@ -0,0 +1,38 @@
...
+    # of a table.
+    # Here are some things that make sense to mark as post deployment:
+    # - Large data migrations. Typically we want these to be run manually so that they can be
+    #   monitored and not block the deploy for a long period of time while they run.
+    # - Adding indexes to large tables. Since this can take a long time, we'd generally prefer to
+    #   run this outside deployments so that we don't block them. Note that while adding an index
+    #   is a schema change, it's completely safe to run the operation after the code has deployed.
+    # Once deployed, run these manually via: https://develop.sentry.dev/database-migrations/#migration-deployment
+
+    is_post_deployment = False
```
wedamija, line 21: The `GroupRelease` table has nearly 700m rows, so you'll need to make this a post deployment migration.

Alternatively, you could make this mostly a state migration, and just rename the existing index instead of drop/create.

src/sentry/migrations/0754_grouprelease_represented_in_django.py
asottile-sentry: sounds good!

src/sentry/migrations/0754_grouprelease_represented_in_django.py
asottile-sentry: ehhh I could do the state dance -- but my hope was to eliminate the split-brain as much as possible now that django can represent this properly

wedamija (approved):

PR #76939, opened 2024-09-04
feat(data-secrecy): Render Data Secrecy Enabled Page for Superusers

iamrajjoshi:
currently when data secrecy is enabled, and we try to do a superuser into an org, we see
image

when you check the network tab, we see that we are getting some 500s as well as some 401s. it is because the exception we raise in `get_superuser_scopes` is not caught anywhere and keeps falling through.

For context: 
* 401: All API requests will return a 401 b/c the data secrecy error is raised (it is captured and returned for the response)
* 500: For Django FE Views, the error is NOT handled, and it falls all the way through, which ends up rendering the Interval Server Error. this above also causes [SENTRY-3EA4](https://sentry.sentry.io/issues/5792418520/). the stack trace in that issue reflects the error falling all the way through.

With this PR, for FE Views, we will instead see
image
instead of the internal error, which makes it apparent why the superuser can't access.


cathteng (commented):
the 500s should cause it to render the Internal Server Error, not the 401s. where are the 500s coming from?

schew2381 (commented):
🚩🚩🚩🚩🚩
TEST
🚩🚩🚩🚩🚩

iamrajjoshi:
> 🚩🚩🚩🚩🚩 TEST 🚩🚩🚩🚩🚩

i want to first see the appetite for this or if we should be doing something else

schew2381:
> > 🚩🚩🚩🚩🚩 TEST 🚩🚩🚩🚩🚩
> 
> i want to first see the appetite for this or if we should be doing something else

Oh the approach lgtm, pretty clean

src/sentry/templates/sentry/data-secrecy.html
```diff
@@ -0,0 +1,19 @@
...
+
+{% block main %}
+  <section class="body">
+    <div class="alert alert-block alert-error">
+      <div style="font-size:24px;margin-bottom:10px">
+        <span class="icon-exclamation" style="font-size:20px;margin-right:10px;"></span>
+        <span>Data Secrecy Enabled</span>
+      </div>
+      <p>{% trans "This organization has Data Secrecy Enabled." %}</p>
+      <p>{% trans "An organization that has data secrecy enabled cannot be accesses via superuser access." %}</p>
```
leedongwei, line 15:
```suggestion
      <p>{% trans "An organization that has data secrecy enabled cannot be accessed via superuser." %}</p>
```

leedongwei (approved):
Seems reasonable to me.

PR #76943, opened 2024-09-04
ref: this test does not need APITestCase

asottile-sentry:
this was causing some breakages when adding a "drop index" migration as it forces a transaction over `tearDown` (?)


IanWoodard (approved):

ceorourke (approved):

asottile-sentry:
hmmm very strange... the test passes on its own so now I need to figure out why some other test is polluting this ugh

asottile-sentry:
found a test pollution with `detect-test-pollution` with `tests/sentry/migrations/test_0726_apitoken_backfill_hashes.py::TestBackfillApiTokenHashesMigration::test_for_hashed_value`  -- not entirely sure what causes it but avoiding the whole HC bits in this test seems to work (by not running the factories themselves)

PR #76944, opened 2024-09-04
feat(dashboards): Add release `SearchQueryBuilder`

gggritso:
For the release dataset, if the `'search-query-builder-releases'` is on, show a `SearchQueryBuilder` instead of a regular search bar. Note: This does _not_ use `'search-query-builder-dashboards'` flag by design. This follows how the logic works for `'search-query-builder-issues'`, which, if enabled, changes the issue search bar _everywhere_ including dashboards.

**Before:**
Screenshot 2024-09-04 at 2 13 11 PM

**After:**  
Screenshot 2024-09-04 at 2 24 43 PM

### DIFF
--- a/static/app/views/dashboards/widgetBuilder/buildSteps/filterResultsStep/releaseSearchBar.tsx
+++ b/static/app/views/dashboards/widgetBuilder/buildSteps/filterResultsStep/releaseSearchBar.tsx
@@ -32,6 +32,13 @@ const supportedTags = Object.values(SESSIONS_FILTER_TAGS).reduce((acc, key) => {
   return acc;
 }, {});
 
+const invalidMessages = {
+  [InvalidReason.WILDCARD_NOT_ALLOWED]: t("Release queries don't support wildcards."),
+  [InvalidReason.FREE_TEXT_NOT_ALLOWED]: t(
+    "Release queries don't support free text search."
+  ),
+};
+
 const SEARCH_SPECIAL_CHARS_REGEXP = new RegExp(
   `^${NEGATION_OPERATOR}|\\${SEARCH_WILDCARD}`,
   'g'
@@ -98,6 +105,7 @@ export function ReleaseSearchBar({
       disallowWildcard
       disallowUnsupportedFilters
       disallowFreeText
+      invalidMessages={invalidMessages}
       recentSearches={SavedSearchType.SESSION}
     />
   ) : (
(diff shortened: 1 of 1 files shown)


static/app/views/dashboards/widgetBuilder/buildSteps/filterResultsStep/releaseSearchBar.tsx
```diff
@@ -75,7 +81,26 @@ export function ReleaseSearchBar({
     return acc;
   }, {});
 
-  return (
+  return organization.features.includes('search-query-builder-releases') ? (
+    <SearchQueryBuilder
+      initialQuery={widgetQuery.conditions}
+      filterKeySections={filterKeySections}
+      filterKeys={supportedTags}
```
malwilley, line 88: Could you extract `supportedTags` outside of the component so that the reference doesn't change on rerender? The new component is pretty heavy so we want to avoid it where possible

static/app/views/dashboards/widgetBuilder/buildSteps/filterResultsStep/releaseSearchBar.tsx
```diff
@@ -75,7 +81,26 @@ export function ReleaseSearchBar({
...
+      placeholder={t('Search for release version, session status, and more')}
+      onChange={(query, state) => {
+        onClose?.(query, {validSearch: state.queryIsValid});
+      }}
+      searchSource="widget_builder"
+      disallowWildcard
+      disallowUnsupportedFilters
+      disallowFreeText
+      recentSearches={SavedSearchType.SESSION}
+    />
```
malwilley, line 102: Do you need to pass the `invalidMessages` here? The previous version had that and we still support that prop

static/app/views/dashboards/widgetBuilder/buildSteps/filterResultsStep/releaseSearchBar.tsx
```diff
@@ -75,7 +81,26 @@ export function ReleaseSearchBar({
     return acc;
   }, {});
 
-  return (
+  return organization.features.includes('search-query-builder-releases') ? (
+    <SearchQueryBuilder
+      initialQuery={widgetQuery.conditions}
+      filterKeySections={filterKeySections}
+      filterKeys={supportedTags}
+      getTagValues={memoize(
```
malwilley, line 89: No need to wrap in `memoize` since the component will deal with caching for you now

gggritso: I thought the default messages were pretty good, but I added explicit ones 👍🏻

static/app/views/dashboards/widgetBuilder/buildSteps/filterResultsStep/releaseSearchBar.tsx
gggritso: I thought the default messages were pretty good, but I added explicit ones 👍🏻

malwilley (approved):

PR #76949, opened 2024-09-04
fix(substatus): Set substatus to regressed

snigdhas:
Similar to https://github.com/getsentry/sentry/pull/76908, we need to check and set the substatus for regressions if the substatus is not provided.


JoshFerge (approved):
is regressed both a substatus and status?

snigdhas:
It's only a substatus, but we hide that implementation detail for our users 😅

PR #76950, opened 2024-09-04
feat(insights): adds create alert buttons to db and queue modules

edwardgou-sentry:
Adds Create Alert menu items to Queue and Database insights module charts that takes users to a prepopulate Create Alert Rule page
image

### DIFF
--- a/static/app/views/insights/common/components/chartPanel.tsx
+++ b/static/app/views/insights/common/components/chartPanel.tsx
@@ -1,21 +1,68 @@
+import {useMemo} from 'react';
 import styled from '@emotion/styled';
 
 import {openInsightChartModal} from 'sentry/actionCreators/modal';
 import {Button} from 'sentry/components/button';
+import {DropdownMenu} from 'sentry/components/dropdownMenu';
 import Panel from 'sentry/components/panels/panel';
-import {IconExpand} from 'sentry/icons';
+import {IconEllipsis, IconExpand} from 'sentry/icons';
 import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
+import useOrganization from 'sentry/utils/useOrganization';
+import usePageFilters from 'sentry/utils/usePageFilters';
+import useProjects from 'sentry/utils/useProjects';
+import {getAlertsUrl} from 'sentry/views/insights/common/utils/getAlertsUrl';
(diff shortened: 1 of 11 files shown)


static/app/views/insights/common/components/chartPanel.tsx
```diff
@@ -1,21 +1,74 @@
...
+          : undefined;
+      const name = alertConfig.name ?? 'Create Alert';
+      const disabled = !singleProject;
+      return {
+        key: name,
+        label: name,
+        to: alertsUrl,
+        disabled,
+        tooltip: disabled
+          ? t('Alerts are only available for a single project')
```
DominikB2014, line 65: Should we also mention what they should do to enable alerts? I.e "please select only one project", or do we think this message is enough to imply that?

static/app/views/insights/common/utils/getAlertsUrl.tsx
```diff
@@ -0,0 +1,23 @@
...
+  query?: string;
+}) {
+  const queryParams = {
+    aggregate: aggregate,
+    dataset: Dataset.GENERIC_METRICS,
+    project: project.slug,
+    eventTypes: 'transaction',
+    query,
+  };
+  return `/alerts/new/metric/?${qs.stringify(queryParams)}`;
```
DominikB2014, line 22: Should we be using `normalizeUrl` here?

DominikB2014 (approved):
code lgtm to me overall!
Comment on the screenshot: in the context menu, is a entry per type of alert (create published alert and create processed alert) the longterm solution here? I feel like if there is a busy chart, there may be a lot of "create alert" options. Perhaps a single "create alert" button with a modal might be a more consistent and scalable approach in the future.

static/app/views/insights/common/components/chartPanel.tsx
edwardgou-sentry: Updated the tooltip message!

edwardgou-sentry:
> code lgtm to me overall! Comment on the screenshot: in the context menu, is a entry per type of alert (create published alert and create processed alert) the longterm solution here? I feel like if there is a busy chart, there may be a lot of "create alert" options. Perhaps a single "create alert" button with a modal might be a more consistent and scalable approach in the future.

Yeah the menu dropdown can get lengthy if there's a lot of series. I'm going to leave this for now since it was the easiest way to implement and we can get some design feedback from the team on how we want to change this

PR #27403, opened 2021-07-14
fix(grouping): write detected synthetic

jjbayer:
For mobile grouping, set the `synthetic` flag in event data if a synthetic exception was detected on the server side.

Fix https://getsentry.atlassian.net/browse/INGEST-170

PR #76951, opened 2024-09-04
drop(grouping): Remove mobile synthetic exception detection

armenzg:
This reverts the work in #27403

No customer on SaaS is using the mobile grouping config, thus, we can remove this now.

We can always bring it back if we find customers who need this.


tests/sentry/event_manager/test_event_manager.py
```diff
@@ -2193,37 +2193,6 @@ def test_category_match_group(self) -> None:
...
-            ),
-            project=self.project,
-        )
-        manager.normalize()
-        event = manager.save(self.project.id)
-
-        mechanism = event.interfaces["exception"].values[0].mechanism
-        assert mechanism is not None
-        assert mechanism.synthetic is True
-        assert event.title == "foo"
```
armenzg, line 2225: Currently, customers will be seeing `SIGABRT` as the title.

jjbayer (approved):

PR #76956, opened 2024-09-04
feat(insights): Set label for domain selector in DB module to `Collection` for MongoDB

0Calories:
Dynamically determine the label for the domain selector. If the user has selected MongoDB for the DB system, then the dropdown label will say `Collection`. Otherwise, it will say `Table`.

image


mjq (approved):
Nice!

PR #76962, opened 2024-09-04
ref(jira): Adds field transformer logic for issue creation

GabeVillalobos:
## Jira Transformers Refactor
The existing logic for creating a new issue via our Jira integration (`/api/0/projects/{org_slug}/{proj_slug}/rule-actions/`) is a bit of a mess, with a lot of duplicated checks and seemingly unnecessary custom field handling.

This PR attempts to clean up the transform logic by using the explicit types passed to us via Jira's Create Issue Metadata endpoint ([jira docs for ref](https://developer.atlassian.com/server/jira/platform/jira-rest-api-example-discovering-meta-data-for-creating-issues-6291669/)).

The new transformer logic is guarded behind a new feature flag `organizations:new-jira-transformers`, with the following deployment plan:
1. LA to Sentry + test organizations, watch failure metrics
2. EA any early adopters, watch failure metrics
3. GA and forward patch remaining bugs
4. Remove the feature flag and officially release the refactors.

## Why bother with this refactor?
I find it easier to reason about a small collection of (mainly) type-based transform functions. In addition, these are individually unit-testable without having to provide a full Jira metadata JSON config like our old testing, as my previous data model changes have moved that responsibility to s different set of Jira integration methods.

## Notable exception
- The `sprint` field in Jira has `type: 'array'`, and `items: 'json'`; however, this is completely different than the expected type for this field and can't cleanly fit into this parsing logic, so it's been special cased. 

I anticipate there are more special cases like this with custom fields, but this is the only issue I managed to encounter during testing.

## Next Steps
Overhaul the Jira Issue Link UI:
-  A number of fields don't currently display properly. 
- There are a couple of field types which are cumbersome, but can be queried/autofilled (team for example). 
- Reopening an issue link dialog doesn't correctly fill every field in the dialog, but retains the state resulting in unusual validation bugs when changing project/task types.

### DIFF
--- a/src/sentry/integrations/jira/models/create_issue_metadata.py
+++ b/src/sentry/integrations/jira/models/create_issue_metadata.py
@@ -16,6 +16,7 @@ class JiraSchemaTypes(str, Enum):
     date = "date"
     team = "team"
     number = "number"
+    json = "json"
     any = "any"
 
 
@@ -156,8 +157,8 @@ class JiraIssueTypeMetadata:
         )
 
     @classmethod
-    def from_jira_meta_config(cls, meta_config: dict[str, Any]) -> list[JiraIssueTypeMetadata]:
+    def from_jira_meta_config(cls, meta_config: dict[str, Any]) -> dict[str, JiraIssueTypeMetadata]:
         issue_types_list = meta_config.get("issuetypes", {})
         issue_configs = [cls.from_dict(it) for it in issue_types_list]
 
-        return issue_configs
+        return {it.id: it for it in issue_configs}
--- a/src/sentry/integrations/jira/models/create_issue_metadata.py
+++ b/src/sentry/integrations/jira/models/create_issue_metadata.py
@@ -4,6 +4,19 @@ from dataclasses import dataclass
 from enum import Enum
 from typing import Any
 
+# A list of common builtin custom field types for Jira for easy reference.
+JIRA_CUSTOM_FIELD_TYPES = {
+    "select": "com.atlassian.jira.plugin.system.customfieldtypes:select",
+    "textarea": "com.atlassian.jira.plugin.system.customfieldtypes:textarea",
+    "multiuserpicker": "com.atlassian.jira.plugin.system.customfieldtypes:multiuserpicker",
+    "tempo_account": "com.tempoplugin.tempo-accounts:accounts.customfield",
+    "sprint": "com.pyxis.greenhopper.jira:gh-sprint",
+    "epic": "com.pyxis.greenhopper.jira:gh-epic-link",
+    "team": "com.atlassian.jira.plugin.system.customfieldtypes:atlassian-team",
+    "rank": "com.pyxis.greenhopper.jira:gh-lexo-rank",
+    "development": "com.atlassian.jira.plugins.jira-development-integration-plugin:devsummarycf",
+}
+
 
 class JiraSchemaTypes(str, Enum):
     string = "string"
--- a/src/sentry/integrations/jira/models/create_issue_metadata.py
+++ b/src/sentry/integrations/jira/models/create_issue_metadata.py
@@ -40,7 +40,7 @@ class JiraSchema:
     The very long custom field name corresponding to some namespace, plugin,
     and custom field name.
     """
-    custom_id: str | None = None
+    custom_id: int | None = None
     """
     A unique identifier for a field on an issue, in the form of 'customfield_<int>'
     """
--- a/src/sentry/integrations/jira/utils/create_issue_schema_transformers.py
+++ b/src/sentry/integrations/jira/utils/create_issue_schema_transformers.py
@@ -1,16 +1,34 @@
-from collections.abc import Callable, Mapping
+from collections.abc import Callable, Iterable, Mapping
 from typing import Any, TypeVar
 
-from sentry.integrations.jira.integration import JIRA_CUSTOM_FIELD_TYPES
+# A list of common builtin custom field types for Jira for easy reference.
+JIRA_CUSTOM_FIELD_TYPES = {
+    "select": "com.atlassian.jira.plugin.system.customfieldtypes:select",
+    "textarea": "com.atlassian.jira.plugin.system.customfieldtypes:textarea",
+    "multiuserpicker": "com.atlassian.jira.plugin.system.customfieldtypes:multiuserpicker",
+    "tempo_account": "com.tempoplugin.tempo-accounts:accounts.customfield",
+    "sprint": "com.pyxis.greenhopper.jira:gh-sprint",
+    "epic": "com.pyxis.greenhopper.jira:gh-epic-link",
+    "team": "com.atlassian.jira.plugin.system.customfieldtypes:atlassian-team",
+    "rank": "com.pyxis.greenhopper.jira:gh-lexo-rank",
+    "development": "com.atlassian.jira.plugins.jira-development-integration-plugin:devsummarycf",
+}
 from sentry.integrations.jira.models.create_issue_metadata import JiraField, JiraSchemaTypes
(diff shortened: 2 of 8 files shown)


src/sentry/integrations/jira/models/create_issue_metadata.py
```diff
@@ -156,8 +157,8 @@ def from_dict(cls, data: dict[str, Any]) -> JiraIssueTypeMetadata:
         )
 
     @classmethod
-    def from_jira_meta_config(cls, meta_config: dict[str, Any]) -> list[JiraIssueTypeMetadata]:
+    def from_jira_meta_config(cls, meta_config: dict[str, Any]) -> dict[str, JiraIssueTypeMetadata]:
         issue_types_list = meta_config.get("issuetypes", {})
         issue_configs = [cls.from_dict(it) for it in issue_types_list]
 
-        return issue_configs
+        return {it.id: it for it in issue_configs}
```
GabeVillalobos, line 164: This makes parsing a single large config and finding an issue type a bit easier.

src/sentry/integrations/jira/utils/create_issue_schema_transformers.py
```diff
@@ -0,0 +1,137 @@
...
+    "select": "com.atlassian.jira.plugin.system.customfieldtypes:select",
+    "textarea": "com.atlassian.jira.plugin.system.customfieldtypes:textarea",
+    "multiuserpicker": "com.atlassian.jira.plugin.system.customfieldtypes:multiuserpicker",
+    "tempo_account": "com.tempoplugin.tempo-accounts:accounts.customfield",
+    "sprint": "com.pyxis.greenhopper.jira:gh-sprint",
+    "epic": "com.pyxis.greenhopper.jira:gh-epic-link",
+    "team": "com.atlassian.jira.plugin.system.customfieldtypes:atlassian-team",
+    "rank": "com.pyxis.greenhopper.jira:gh-lexo-rank",
+    "development": "com.atlassian.jira.plugins.jira-development-integration-plugin:devsummarycf",
+}
```
GabeVillalobos, line 15: I've temporarily moved this here to avoid cycles in our imports, but this is probably going into the create_issue_metadata model file.

src/sentry/integrations/jira/utils/create_issue_schema_transformers.py
```diff
@@ -0,0 +1,137 @@
...
+from sentry.shared_integrations.exceptions import IntegrationFormError
+
+
+class JiraSchemaParseError(Exception):
+    pass
+
+
+def parse_number_field(num_str: str) -> int | float:
+    try:
+        if "." in num_str:
```
corps, line 26: ya love to see it!!!

src/sentry/integrations/jira/utils/create_issue_schema_transformers.py
```diff
@@ -0,0 +1,141 @@
...
+            type_transformers, custom_field_transformers, field
+        )
+
+        try:
+            # We need to handle array types, but for whatever reason `Sprint`
+            # fields have a type of array/json, but expect a single value, so we
+            # have to special case them.
+            if (
+                field.schema.schema_type.lower() == JiraSchemaTypes.array
+                and field.schema.custom != JIRA_CUSTOM_FIELD_TYPES["sprint"]
```
GabeVillalobos, line 122: This is a really annoying edge case. The input type is different than the schema type. It expects a single sprint ID as an int, but returns a list of sprint JSON data, meaning we just have to special case this. I imagine there are more custom fields like this as well unfortunately :/

src/sentry/integrations/jira/utils/create_issue_schema_transformers.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def parse_number_field(num_str: Any) -> int | float:
+    try:
+        if isinstance(num_str, str) and "." in num_str:
+            return float(num_str)
+
+        return int(num_str)
+    except ValueError:
+        raise JiraSchemaParseError(f"Invalid number value provided for field: '{num_str}'")
```
GabeVillalobos, line 23: This is only used here for now, but I imagine we can start adding stricter validation with explicit error raises when we want to tighten up our inputs a bit.

src/sentry/integrations/jira/utils/create_issue_schema_transformers.py
```diff
@@ -0,0 +1,133 @@
...
+        JiraSchemaTypes.project.value: id_obj_transformer,
+        JiraSchemaTypes.number.value: parse_number_field,
+    }
+
+    return transformers
+
+
+def get_custom_field_transformer_mappings() -> TransformerType:
+    transformers = {
+        # JIRA_CUSTOM_FIELD_TYPES["select"]: identity_transformer,
```
cathteng, line 54: is this yet to be created? is there any downside to not having it in this PR too?

src/sentry/integrations/jira/utils/create_issue_schema_transformers.py
```diff
@@ -0,0 +1,133 @@
...
+    if labels := data.get("labels"):
+        data["labels"] = [label.strip() for label in labels.split(",") if label.strip()]
+
+    type_transformers = get_type_transformer_mappings(user_id_field)
+    custom_field_transformers = get_custom_field_transformer_mappings()
+
+    for field in jira_fields:
+        field_data = data.get(field.key)
+
+        # We don't have a mapping for this field, so it's probably extraneous
```
cathteng, line 100: will people complain if a field is not ported over when they try to create an issue?

src/sentry/integrations/jira/utils/create_issue_schema_transformers.py
cathteng: i don't think i'm grokking this properly, we have special case logic if the schema type is `array` but the field is not `sprint`?

GabeVillalobos: Kind of. We need to iterate over any field with type `array` and run the sub-transformers on each item if we're passed an iterable. `sprint` custom fields are funny in that Jira errors if we pass them a list, which is expected by every other `array` type I've tested, so by excluding it from the iteration logic here, it falls through to the custom type handler on L56.

GabeVillalobos: As an alternative to make this a little less messy, I could move this logic in to an array type transformer that calls into the valid sub-type transformer 🤔 It just becomes a recursive loop if we ever receive a schema with type: `array`, items: `array`.

src/sentry/integrations/jira/utils/create_issue_schema_transformers.py
GabeVillalobos: Probably! But this seems like a less terrible experience than mismapping a field, failing to create an issue, and not alerting them that something is broken. Might be worth collecting all of these unmapped fields and reporting them to the UI, though I'm not sure how spammy that will get.

src/sentry/integrations/jira/utils/create_issue_schema_transformers.py
cathteng: ohhh we're ignoring the sprint field here. got it

src/sentry/integrations/jira/utils/create_issue_schema_transformers.py
GabeVillalobos: Yep! The logic that displays `select` in the UI is currently busted, so I need to patch that before this field can be useful. I'll add TODO comments on these to make this less ambiguous.

As for `epic`, that's another display issue that I need to dig further into, but I can't get anything with type `epic` to appear in the issue-link dropdown 🤔

cathteng (approved):

PR #76965, opened 2024-09-04
fix(alerts): Pass time_period as an int

ceorourke:
We're still seeing a 401 from Seer when hitting this endpoint in prod and it may be because we're passing `time_period` as a float rather than an int. This rewrites it slightly to use the types and adds the `DetectAnomaliesRequest` to match what's defined in [Seer](https://github.com/getsentry/seer/blob/main/src/seer/anomaly_detection/models/external.py#L64-L71).

### DIFF
--- a/src/sentry/incidents/subscription_processor.py
+++ b/src/sentry/incidents/subscription_processor.py
@@ -48,7 +48,13 @@ from sentry.incidents.tasks import handle_trigger_action
 from sentry.incidents.utils.types import QuerySubscriptionUpdate
 from sentry.models.project import Project
 from sentry.net.http import connection_from_url
-from sentry.seer.anomaly_detection.types import AnomalyType
+from sentry.seer.anomaly_detection.types import (
+    AlertInSeer,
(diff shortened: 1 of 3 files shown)


src/sentry/incidents/subscription_processor.py
```diff
@@ -673,32 +679,29 @@ def anomaly_has_confidence(self, anomaly) -> bool:
         return anomaly_type != AnomalyType.NO_DATA.value
 
     def get_anomaly_data_from_seer(self, aggregation_value: float | None):
+        anomaly_detection_config = AnomalyDetectionConfig(
+            time_period=int(self.alert_rule.snuba_query.time_window / 60),
```
nhsiehgit, line 683: maybe could just do ` // 60` instead?

nhsiehgit (approved):
1 nit

src/sentry/incidents/subscription_processor.py
ceorourke: what's the difference? does that automatically make it an integer?

PR #76966, opened 2024-09-04
feat(custom-views): Add feedback button to tab menu button

MichaelSun48:
Adds a feedback button to the bottom of the tab menu button (only appears for the Sentry org). definitely not the most graceful, open to design criticism. 

image

### DIFF
--- a/static/app/views/issueList/groupSearchViewTabs/draggableTabMenuButton.tsx
+++ b/static/app/views/issueList/groupSearchViewTabs/draggableTabMenuButton.tsx
@@ -60,7 +60,7 @@ function FeedbackFooter() {
           openForm({
             messagePlaceholder: t('How can we make custom views better for you?'),
             tags: {
-              source: 'custom_views',
+              ['feedback.source']: 'custom_views',
             },
           })
         }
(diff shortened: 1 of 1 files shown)


static/app/views/issueList/groupSearchViewTabs/draggableTabMenuButton.tsx
```diff
@@ -33,12 +37,49 @@ export function DraggableTabMenuButton({
...
+  return (
+    <SectionedOverlayFooter>
+      <Button
+        size="xs"
+        icon={<IconMegaphone />}
+        onClick={() =>
+          openForm({
+            messagePlaceholder: t('How can we make custom views better for you?'),
+            tags: {
+              source: 'custom_views',
```
malwilley, line 63: We've been using this convention for the recent feedback buttons, can you use that here too?

```
['feedback.source']: '...',
['feedback.owner']: 'issues',
```

malwilley (approved):
One other way you could add a feedback button here is to give the dropdown menu a title like "Actions" or something and make it look like the priority dropdown. But code looks good!

CleanShot 2024-09-04 at 15 02 26

static/app/views/issueList/groupSearchViewTabs/draggableTabMenuButton.tsx
MichaelSun48: yup!

PR #76970, opened 2024-09-04
chore(custom-views): Add real time button back to custom views header 

MichaelSun48:
As per our custom views sync on Aug 29, we are not sure if we want to delete real time issues as a feature just yet, so this PR adds the button back to the issue stream.


scttcper (changes requested):
i object, your honor

static/app/views/issueList/customViewsHeader.tsx
```diff
@@ -67,7 +78,18 @@ function CustomViewsIssueListHeader({
...
+          <Button
+            size="sm"
+            data-test-id="real-time"
+            title={realtimeTitle}
+            aria-label={realtimeTitle}
+            icon={realtimeActive ? <IconPause /> : <IconPlay />}
+            onClick={() => onRealtimeChange(!realtimeActive)}
+          />
+        </ButtonBar>
+      </Layout.HeaderActions>
```
scttcper, line 92: you should delete this

scttcper (approved):

PR #76972, opened 2024-09-04
ref(users): move other users related serializers to new abode

Christinarlong:
Moves rest of user's serializers to the user's directory 
- UserIP
- UserRole
- Identity
- IdentitiyProvider
- UserEmail
Adds typing - mainly serializer response dicts to the serializers

split of (https://github.com/getsentry/sentry/pull/76822)
ref(issue # here)


src/sentry/users/api/endpoints/user_emails.py
```diff
@@ -78,7 +79,7 @@ def get(self, request: Request, user: User) -> Response:
 
         emails = user.emails.all()
 
-        return self.respond(serialize(list(emails), user=user))
+        return self.respond(serialize(list(emails), user=user, serializer=UserEmailSerializer()))
```
Christinarlong, line 82: Extremely weird bug and this change is one workaround.

For some reason none of the users/moved `serializers` were being registered in the registry despite having the register decorator. This also happened for the `authenticator` serializer when **only** it was moved. 

Some workarounds I found, 
- Import at least one users serializer in `sentry.api.serializers` `__init__.py` for some reason, being imported here will also cause the register to fire
- Import at least one users serializer in an endpoint (I assume it doesn't need to be an endpoint but just any file that is initialized ? I thought the `sentry.users.api.serializers` init.py would work, but I guess not?)
- The `UserSerializer` will always be registered for some reason, so if the other users serializers are also imported in the same `__init__.py`, they will all be registered. This is what happened for `authenticator` serializer in https://github.com/getsentry/sentry/pull/76896

Christinarlong: Note 2: I haven't tried recreating this in CI tests, but it was p. reproducible in local tests. E.g run this branch without `UserEmailSerializer` being imported/revert all changes in `User_Email` endpoint. Or just migrate any serializer (not UserSerializer) into users directory and run tests

markstory: > For some reason none of the users/moved serializers were being registered in the registry despite having the register decorator. This also happened for the authenticator serializer when only it was moved.

This is likely because the new locations aren't being imported at all, so the registration calls never happen. We're likely relying on application startup importing `sentry.api.serializers` which loads all of the other serializer modules.

> On a separate note: it may be a good idea to explicitly pass in the specific serializer more, it helps prevent the above bug and is clearer

I agree. Being explicit and more obvious about which serializer an endpoint uses is clearer and helps us minimize the number of modules that are imported during application startup as well.

src/sentry/users/api/serializers/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .identity import *  # noqa: F401,F403
+from .identityprovider import *  # noqa: F401,F403
+from .useremail import *  # noqa: F401,F403
+from .userip import *  # noqa: F401,F403
+from .userrole import *  # noqa: F401,F403
```
markstory, line 5: If we don't need this barrel file to satisfy import side-effects we could skip it.

markstory (approved):

src/sentry/users/api/endpoints/user_emails.py
Christinarlong: Gotchas that makes sense. I think something that confused me was that `UserSerializer` would always get imported despite the move and I was struggling/didn't know how to trace backwards to figure out where it was getting imported from. 

I added some print statements yesterday and saw that the `__init__.py` for `sentry.users.api.serializers` was getting "run" (?) during the django setup but not registered (?)

src/sentry/users/api/serializers/__init__.py
Christinarlong: Yeah, currently I think it's needed so that all the serializers will be registered, but we can just do explicit serializers in the endpoints instead to knock 2 birds with 1 stone

markstory (approved):

src/sentry/users/api/endpoints/user_emails.py
Christinarlong: Okay, followup to this for closure, I think `UserSerializer` was always imported because `DetailedSelfUserSerializer` was being explicitly passed already in the `user_details` endpt

PR #76974, opened 2024-09-04
feat(auth): add scopes and org selection to ApiApplication

sentaur-athena:
We want to be able to provides scopes for each application and store it in the model. Today the application requests the scope which is too loose.

The second change is to be able to limit the application scope to one org at a time. Basically if `requires_org_level_access` is true the user has to give the app permission twice to give the app access to 2 orgs.


wedamija (approved):

markstory:
> We want to be able to provides scopes for each application and store it in the model. Today the application requests the scope which is too loose.

What will happen if an apiapplication has scopes stored in the database and a different set of scopes is provided during the authorization flow?

sentaur-athena:
@markstory we respect data base if it's tighter than what the app is asking for. For now though it's nullable and if it's null we don't change behavior at all.

PR #76976, opened 2024-09-04
fix(grouping): Fix flaky hash comparison metric test

lobsterkatie:
This fixes the `test_records_hash_comparison` test in `test_event_manager_grouping.py` to be compatible with the logic in `_save_aggregate_new`. In the latter, secondary hashes are only calculated if primary hashes aren't found. The fact that I faked parameterization (in order to get the conveniences of class-based tests) meant that the database wasn't getting wiped between cases, which in turn meant the primary hash _was_ found in the second case, so secondary hashes weren't calculated and the test broke.

This fixes that by using real parameterization, so each test case starts with a clean slate.

(The reason this wasn't discovered earlier is that the test was flakily working, as I rolled out use of `_save_aggregate_new`. The old version, `_save_aggregate`, always calculates secondary hashes, whether it needs to or not, which hid the problem.)

### DIFF
--- a/tests/sentry/event_manager/test_event_manager_grouping.py
+++ b/tests/sentry/event_manager/test_event_manager_grouping.py
@@ -16,6 +16,7 @@ from sentry.models.auditlogentry import AuditLogEntry
 from sentry.models.group import Group
 from sentry.models.grouphash import GroupHash
 from sentry.models.grouphashmetadata import GroupHashMetadata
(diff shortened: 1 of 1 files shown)


JoshFerge (approved):
🙏

armenzg (approved):

armenzg (approved):

PR #76977, opened 2024-09-04
chore(anomaly detection): replace placeholder get_historical_data with seer call

mifu67:
Implement the seer call logic in `get_historical_anomaly_data_from_seer`

### DIFF
--- a/src/sentry/seer/anomaly_detection/get_historical_anomalies.py
+++ b/src/sentry/seer/anomaly_detection/get_historical_anomalies.py
@@ -1,24 +1,164 @@
+import logging
+from datetime import datetime
+
+from django.conf import settings
+from django.core.exceptions import ValidationError
+from urllib3.exceptions import MaxRetryError, TimeoutError
+
+from sentry.conf.server import SEER_ANOMALY_DETECTION_ENDPOINT_URL
 from sentry.incidents.models.alert_rule import AlertRule, AlertRuleStatus
 from sentry.models.project import Project
-from sentry.seer.anomaly_detection.types import AnomalyType
+from sentry.net.http import connection_from_url
+from sentry.search.events.types import SnubaParams
+from sentry.seer.anomaly_detection.types import AnomalyDetectionConfig
+from sentry.seer.anomaly_detection.utils import format_historical_data, translate_direction
+from sentry.seer.signed_seer_api import make_signed_seer_api_request
+from sentry.snuba.models import SnubaQuery
+from sentry.snuba.referrer import Referrer
+from sentry.snuba.utils import get_dataset
+from sentry.utils import json
+from sentry.utils.json import JSONDecodeError
+from sentry.utils.snuba import SnubaTSResult
+
+logger = logging.getLogger(__name__)
+
+seer_anomaly_detection_connection_pool = connection_from_url(
+    settings.SEER_ANOMALY_DETECTION_URL,
+    timeout=settings.SEER_ANOMALY_DETECTION_TIMEOUT,
+)
 
 
 def get_historical_anomaly_data_from_seer(
     alert_rule: AlertRule, project: Project, start_string: str, end_string: str
 ) -> list | None:
     """
-    Send time series data to Seer and return anomaly detection response (PLACEHOLDER).
+    Send time series data to Seer and return anomaly detection response.
     """
     if alert_rule.status == AlertRuleStatus.NOT_ENOUGH_DATA.value:
         return []
+    # don't think this can happen but mypy is yelling
+    if not alert_rule.snuba_query:
+        raise ValidationError("No snuba query associated with alert rule.")
+    subscription = alert_rule.snuba_query.subscriptions.first()
+    # same deal as above
+    if not subscription:
+        raise ValidationError("No subscription associated with alert rule.")
+    snuba_query = SnubaQuery.objects.get(id=alert_rule.snuba_query_id)
+    dataset = get_dataset(snuba_query.dataset)
+    window_min = int(snuba_query.time_window / 60)
+    start = datetime.fromisoformat(start_string)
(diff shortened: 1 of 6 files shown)


src/sentry/seer/anomaly_detection/get_historical_anomalies.py
```diff
@@ -1,24 +1,164 @@
...
+                "ad_config": anomaly_detection_config,
+                "context": formatted_data,
+                "response_data": response.data,
+                "reponse_code": response.status,
+            },
+        )
+        return None
+
+
+def fetch_historical_data(
```
nhsiehgit, line 137: This method is _different_ than the `fetch_historical_data` in the `store_data` file right?
This one is _actually_ just fetching the historical data?

slightly hesitant about naming it the same, but hopefully we can eventually get around to [this ticket](https://getsentry.atlassian.net/browse/ALRT-187?atlOrigin=eyJpIjoiNGU4NDU2MjgyYzdhNGI3M2JmZTEwNzM2MmMzZGYyZjgiLCJwIjoiaiJ9) and rename the store_data.py methods

src/sentry/seer/anomaly_detection/get_historical_anomalies.py
```diff
@@ -1,24 +1,164 @@
...
 
-    return [
-        {
-            "timestamp": 0.1,
-            "value": 100.0,
-            "anomaly": {
-                "anomaly_type": AnomalyType.HIGH_CONFIDENCE.value,
-                "anomaly_value": 100,
+    if not historical_data:
+        raise ValidationError("No historical data available.")
```
nhsiehgit, line 56: do we want to raise here?
If there's no historical data, we should jsut return `None` right? (bubbles up to a 400?)

OR - does this mean that there was an issue in the fetch to seer, and the message should actually be some error occured while fetching?

src/sentry/seer/anomaly_detection/get_historical_anomalies.py
```diff
@@ -1,24 +1,164 @@
...
+    formatted_data = format_historical_data(historical_data, dataset)
+    if (
+        not alert_rule.sensitivity
+        or not alert_rule.seasonality
+        or alert_rule.threshold_type is None
+        or alert_rule.organization is None
+    ):
+        # this won't happen because we've already gone through the serializer, but mypy insists
```
nhsiehgit, line 64: that darn stickler

src/sentry/seer/anomaly_detection/get_historical_anomalies.py
```diff
@@ -1,24 +1,164 @@
...
+                {
+                    "organization_id": alert_rule.organization.id,
+                    "project_id": project.id,
+                    "config": anomaly_detection_config,
+                    "context": formatted_data,
+                }
+            ).encode("utf-8"),
+        )
+    except (TimeoutError, MaxRetryError):
+        logger.warning(
```
nhsiehgit, line 87: should this be a warning or error?

nhsiehgit (approved):
non blocking nits

src/sentry/seer/anomaly_detection/get_historical_anomalies.py
```diff
@@ -1,24 +1,164 @@
...
+        sensitivity=alert_rule.sensitivity,
+        direction=translate_direction(alert_rule.threshold_type),
+        expected_seasonality=alert_rule.seasonality,
+    )
+    try:
+        response = make_signed_seer_api_request(
+            seer_anomaly_detection_connection_pool,
+            SEER_ANOMALY_DETECTION_ENDPOINT_URL,
+            json.dumps(
+                {
```
ceorourke, line 78: You can use the `DetectAnomaliesRequest` typeddict from https://github.com/getsentry/sentry/blob/master/src/sentry/seer/anomaly_detection/types.py#L35

src/sentry/seer/anomaly_detection/get_historical_anomalies.py
```diff
@@ -1,24 +1,164 @@
...
+
+    if response.status != 200:
+        logger.error(
+            f"Received {response.status} when calling Seer endpoint {SEER_ANOMALY_DETECTION_ENDPOINT_URL}.",  # noqa
+            extra={"response_data": response.data},
+        )
+        return None
+
+    try:
+        results = json.loads(response.data.decode("utf-8")).get("anomalies")
```
ceorourke, line 107: The [response type](https://github.com/getsentry/seer/blob/main/src/seer/anomaly_detection/models/external.py#L74-L75) says this is a list of `TimeSeriesPoints` which has singular `anomaly` so to get them all you'd want to iterate through the list and pop off the `anomaly`

mifu67: @nhsiehgit do you need the anomaly itself? Or is sending the timeseriespoints acceptable? What are you expecting in the response?

src/sentry/seer/anomaly_detection/get_historical_anomalies.py
ceorourke, line 137: Could we reuse this https://github.com/getsentry/sentry/blob/master/src/sentry/seer/anomaly_detection/store_data.py#L127-L127 instead of writing a new one? I think the only difference is the start, end, and referrer

tests/sentry/incidents/endpoints/test_organization_alert_rule_anomalies.py
```diff
@@ -68,7 +73,28 @@ def test_simple(self, mock_seer_store_request):
 
         self.login_as(self.user)
 
+        seer_return_value = {
```
ceorourke, line 76: I don't think this return value is correct based on https://github.com/getsentry/seer/blob/main/src/seer/anomaly_detection/models/external.py#L74-L74

PR #76979, opened 2024-09-05
fix(api): Rename to avoid confusion

sentaur-athena:
There is constant confusion about this workflow action, when it fails people wonder how they can fix it. This is only to display a diff and no action is required.

### DIFF
--- a/.github/workflows/openapi-diff.yml
+++ b/.github/workflows/openapi-diff.yml
@@ -14,7 +14,7 @@ env:
 
 jobs:
   check-diff:
-    name: Fails to show the difference between the prod schema and dev schema (optional)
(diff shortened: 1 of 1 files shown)


JoshFerge (approved):

.github/workflows/openapi-diff.yml
```diff
@@ -14,7 +14,7 @@ env:
 
 jobs:
   check-diff:
-    name: build api
+    name: Fails to show the difference between the prod schema and dev schema (optional)
```
schew2381, line 17: Maybe a slightly shorter name so you can read it in the GH actions + put optional at the front?
```suggestion
    name: (Optional) Shows the difference between the prod and dev schema
```

PR #76980, opened 2024-09-05
chore(api): make SlugsUpdateEndpoint private

michellewzhang:
image

fixed the typo, screenshot is from pre-fix


src/sentry/api/endpoints/organization_slugs.py
```diff
@@ -1,34 +1,53 @@
...
 @region_silo_endpoint
 class SlugsUpdateEndpoint(OrganizationEndpoint):
+    owner = ApiOwner.UNOWNED
     publish_status = {
-        "PUT": ApiPublishStatus.UNKNOWN,
+        "PUT": ApiPublishStatus.PUBLIC,
     }
 
+    @extend_schema(
+        operation_id="Update an Origanization's Project Slugs",
```
isabellaenriquez:
```suggestion
        operation_id="Update an Organization's Project Slugs",
```

JoshFerge, line 31:
```suggestion
        operation_id="Update an Organization's Project Slugs",
```

src/sentry/api/endpoints/organization_slugs.py
```diff
@@ -1,34 +1,53 @@
...
+        parameters=[GlobalParams.ORG_ID_OR_SLUG],
+        request=inline_serializer(
+            name="UpdateOrgProjectSlugs",
+            fields={
+                "slugs": serializers.DictField(
+                    help_text="a dictionary of project IDs to their intended slugs.", required=False
```
isabellaenriquez, line 37:
```suggestion
                    help_text="A dictionary of project IDs to their intended slugs.", required=False
```

could be helpful to include an example as well

JoshFerge (commented):
are we sure we want to publish this one? i feel like the project details PUT might be sufficient for doing something like this https://docs.sentry.io/api/projects/update-a-project/

src/sentry/api/endpoints/organization_slugs.py
```diff
@@ -1,34 +1,53 @@
...
+                ),
+            },
+        ),
+        responses={
+            200: inline_sentry_response_serializer("SlugsUpdateResponse", list[str]),
```
isabellaenriquez, line 42: i think the response type should be dict?

michellewzhang: oh yes good catch!

michellewzhang:
> are we sure we want to publish this one? i feel like the project details PUT might be sufficient for doing something like this https://docs.sentry.io/api/projects/update-a-project/

hmm you could be right @JoshFerge  - cc @sentaur-athena, what do you think?

JoshFerge:
> > are we sure we want to publish this one? i feel like the project details PUT might be sufficient for doing something like this https://docs.sentry.io/api/projects/update-a-project/
> 
> hmm you could be right @JoshFerge - cc @sentaur-athena, what do you think?

yeah i feel strongly that we shouldn't publish this one -- this seems like an endpoint that is useful for the frontend but doesn't seem necessary for a list of backend APIs as it feels redundant and confusing compared to the project PUT update which can update slugs

isabellaenriquez (approved):

PR #76982, opened 2024-09-05
fix(alerts): Update AlertMetricsQueryBuilder to accept time_window and update spm to use time_window instead of granularity

edwardgou-sentry:
Updates the AlertMetricsQueryBuilder (and MetricsQueryBuilder) to accept a `time_window` arg. Also updates `spm` in the metrics data set to fallback to `time_window` from the builder when a `start` and `end` are not provided. We were using `granularity` as the fallback before, but this is not correct because `granularity` does not match the actual time range that a Snuba subscription query will aggregate over.

### DIFF
--- a/src/sentry/search/events/builder/metrics.py
+++ b/src/sentry/search/events/builder/metrics.py
@@ -105,7 +105,8 @@ class MetricsQueryBuilder(BaseQueryBuilder):
         # Dataset.PerformanceMetrics is MEP. TODO: rename Dataset.Metrics to Dataset.ReleaseMetrics or similar
         dataset: Dataset | None = None,
         granularity: int | None = None,
-        time_window: int | None = None,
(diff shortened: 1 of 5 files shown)


src/sentry/search/events/builder/metrics.py
```diff
@@ -105,6 +105,7 @@ def __init__(
         # Dataset.PerformanceMetrics is MEP. TODO: rename Dataset.Metrics to Dataset.ReleaseMetrics or similar
         dataset: Dataset | None = None,
         granularity: int | None = None,
+        time_window: int | None = None,
```
shruthilayaj, line 108: let's rename this to `time_range_window` or something like that to indicate it's only ever a relative range. and add some doc strings to explain why we added it and validation (like, if start and end is specified, time_range_window probably shouldn't be etc) better to be restrictive to begin with.

shruthilayaj (approved):

PR #76984, opened 2024-09-05
fix(search): hide feedback from issues search by default

### LINKED ISSUE #60697: Bug: Issue only search doesn't correctly filter out disabled issue types
When querying in issue search we have a path where we only search in postgres:

https://github.com/getsentry/sentry/blob/678abe885a811005f5d81678238da75f202da3dd/src/sentry/search/snuba/executors.py#L808-L812

This path isn't correctly filtering out issue types that haven't been released. We need to make sure this filtering occurs here.

aliu39:
Follow up to https://github.com/getsentry/sentry/pull/59505 to filter feedbacks from the default search when querying Postgres. As before, filtering on issue.category:feedback or a negation (ex !issue.category:error) will still return feedback issues.

Fixes https://github.com/getsentry/sentry/issues/60697
Fixes https://github.com/getsentry/team-replay/issues/470


tests/sentry/issues/endpoints/test_organization_group_index.py
```diff
@@ -3917,6 +3920,58 @@ def test_snuba_heavy_error_handled_boolean(self, _: MagicMock) -> None:
...
+            create_feedback_issue(
+                feedback_event, self.project.id, FeedbackCreationSource.NEW_FEEDBACK_ENVELOPE
+            )
+            self.login_as(user=self.user)
+            res = self.get_success_response()
+
+        assert len(res.data) == 1
+        issue = res.data[0]
+        feedback_group = Group.objects.get(type=FeedbackGroup.type_id)
+        assert int(issue["id"]) != feedback_group.id
```
michellewzhang, line 3946: could we add a comment here explaining what's being tested? (for clarity)
something simple like 
`# test that the issue returned is not the feedback issue`

tests/sentry/issues/endpoints/test_organization_group_index.py
```diff
@@ -3917,6 +3920,58 @@ def test_snuba_heavy_error_handled_boolean(self, _: MagicMock) -> None:
...
+        assert issue["issueCategory"] != "feedback"
+
+    def test_feedback_filter(self, _):
```
michellewzhang, line 3949:
```suggestion
    def test_feedback_returned_from_filter(self, _):
```

tests/sentry/issues/endpoints/test_organization_group_index.py
```diff
@@ -3917,6 +3920,58 @@ def test_snuba_heavy_error_handled_boolean(self, _: MagicMock) -> None:
...
+            create_feedback_issue(
+                feedback_event, self.project.id, FeedbackCreationSource.NEW_FEEDBACK_ENVELOPE
+            )
+            self.login_as(user=self.user)
+            res = self.get_success_response(query="issue.category:feedback")
+
+        assert len(res.data) == 1
+        issue = res.data[0]
+        feedback_group = Group.objects.get(type=FeedbackGroup.type_id)
+        assert int(issue["id"]) == feedback_group.id
```
michellewzhang, line 3972: another clarifying comment here would also be good, something like 
`# test that the issue returned is the feedback issue`

michellewzhang (approved):

src/sentry/issues/search.py
```diff
@@ -8,14 +8,21 @@
...
+    get_group_type_by_type_id,
+)
 from sentry.models.environment import Environment
 from sentry.models.organization import Organization
 from sentry.search.events.filter import convert_search_filter_to_snuba_query
 from sentry.snuba.dataset import Dataset
 from sentry.utils import snuba
 from sentry.utils.snuba import SnubaQueryParams
 
+HIDDEN_FROM_DEFAULT_SEARCH_GROUP_TYPES = (FeedbackGroup,)
```
wedamija, line 24: I'm thinking that rather than including these as a hardcoded list, we should actually have it be a property on the `GroupType`. That way you don't need to hunt through the codebase to figure out how to exclude these from issue search

aliu39: Sounds good!

src/sentry/search/snuba/executors.py
```diff
@@ -1837,6 +1843,22 @@ def query(
                     where_conditions.append(
                         Condition(Column("occurrence_type_id", joined_entity), Op.IN, group_types)
                     )
+                else:
+                    all_group_type_objs: list[GroupType] = [
+                        gt_registry.get_by_type_id(id)
+                        for id in gt_registry.get_all_group_type_ids()
```
wedamija, line 1849: I think we should do this like this, on line 1828
```
if not is_errors:
    visible_group_types = group_types_from(search_filters)
    if visible_group_types is None:
        # When no group types are specified, default to all default group types
        visible_group_types = [gt for gt in all_group_type_objs if not gt.in_default_search]
    # remove `raw_group_types is not None` check, since it'll never be none now, just potentially empty
    # Then just run the same logic we were running before
```

This keeps the logic shared which should reduce the chance of bugs. Since in your new branch you missed `grouptype.registry.get_visible`, and so we'd have started showing unreleased issue types to all users.

Reading this code, this might actually be a bug already... if no issue types are specified, we're not filtering to visible issue types only. So doing this will fix the bug

aliu39: cool, this line should be like this right?
[gt for gt in all_group_type_objs if ~~not~~ gt.in_default_search]

aliu39: the visible filter seems to happen in the if not None case

aliu39: Moved the logic to the base case of `group_types_from`. For some reason this causes test_snuba_perf_issue to fail, not sure why, it worked once locally so could be flakey.

aliu39: For the test, it's because the GroupType created by storing a transaction, 2001 ProfileFileIOGroupType, is not registered. 

Not sure how GroupTypes are registered in prod, but this change makes it so issues with unregistered GroupTypes are never listed (before we wouldn't filter on type).

aliu39: > this change makes it so by default, issues with unregistered GroupTypes are never listed (before, we would just skip the filter).

Do we expect all group types to be registered in prod? Is it ok to manually register the 2001 type to pass this test?

wedamija: Not totally sure what you mean by it not being registered? It might just be because the group type hasn't been marked as visible?

You can do that like https://github.com/getsentry/sentry/blob/9d2507844acf478444d3b2bc9df1c66a3ae8ca5f/tests/snuba/search/test_backend.py#L3583

aliu39: ohh got it, that was it, thanks!

wedamija (approved):
Nice, thanks for fixing this!

PR #76990, opened 2024-09-05
Add router config in Vue onboarding

smeubank:
@AbhiPrasad can you confirm this is correct, i just noticed it while testing Nuxt and Vue apps


priscilawebdev (approved):
good catch @smeubank 👏 according to the [docs](https://docs.sentry.io/platforms/javascript/guides/vue/) `router` shall be there

PR #76997, opened 2024-09-05
chore: Preserve timestamp of empty transactions in SDK setup

gggritso:
If the transaction had no spans (e.g., a custom transaction), the filtered span logic assigns it a `"timestamp"` of `-Infinity` because it iterates an empty array. Instead, only apply the timestamp adjustment logic iff (sic) any spans were filtered out. If not, do nothing. Otherwise, `-Infinite` is normalized into `null`, and Relay appends an incorrect `"timestamp"` at ingestion time.

### DIFF
--- a/static/app/bootstrap/initializeSdk.tsx
+++ b/static/app/bootstrap/initializeSdk.tsx
@@ -132,14 +132,15 @@ export function initializeSdk(config: Config, {routes}: {routes?: Function} = {}
       addExtraMeasurements(event);
       addUIElementTag(event);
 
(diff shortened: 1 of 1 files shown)


mydea (approved):

k-fish (approved):

k-fish:
The actual solution to the "removing spans" thing is doing `beforeSpanEnd` which I believe has landed, which should stop the spans from getting committed and removes the need for this hack

gggritso:
@k-fish ah nice, yeah! I can do that in a follow-up

PR #76994, opened 2024-09-05
drop(grouping): Deprecate hierarchical code

armenzg:
This partially reverts #23861

No projects in SaaS are using the mobile grouping anymore (this is the only one using the hierarchical code).

Some code has already been reverted in:
* https://github.com/getsentry/sentry/pull/76951
* https://github.com/getsentry/sentry/pull/76991
* https://github.com/getsentry/sentry/pull/76992
* https://github.com/getsentry/sentry/pull/76993

More code needs to be removed but this is the minimal removal which passes all tests.

PR #76998, opened 2024-09-05
drop(grouping): Remove hierarchical code - part 2

armenzg:
This is a follow-up to https://github.com/getsentry/sentry/pull/76994. In that pull request, we disable the feature from the mobile grouping config, thus, reducing the number of changes in this pull request.

More code will be removed in a follow-up pull request.


jjbayer (approved):

src/sentry/eventstore/models.py
```diff
@@ -335,32 +335,17 @@ def get_grouping_config(self) -> GroupingConfig:
...
-        some other things).
-
-        The returned hashes already take SDK fingerprints and checksums into
-        consideration.
+        Returns _all_ information that is necessary to group an event into
+        issues: An event should be sorted into a group X, if there is a GroupHash
+        matching *any* of the hashes. Hashes that do not yet have a GroupHash model get
+        one and are associated with the same group (unless they already belong to another group).
 
+        The returned hashes already take SDK fingerprints and checksums into consideration.
```
lobsterkatie, line 347: Nit: 

This is already in the first paragraph of your revised docstring.

```suggestion
```

src/sentry/eventstore/models.py
```diff
@@ -373,21 +358,16 @@ def get_hashes(self, force_config: StrategyConfiguration | None = None) -> Calcu
         from sentry.grouping.api import sort_grouping_variants
 
         variants = self.get_grouping_variants(force_config)
-        flat_variants, hierarchical_variants = sort_grouping_variants(variants)
+        flat_variants = sort_grouping_variants(variants)
```
lobsterkatie, line 361: I support not doing it in this PR, which is big enough already, but we could further clean this up by getting rid of the `flat_` business, now that they're the only ones.

src/sentry/eventstore/models.py
```diff
@@ -476,9 +456,6 @@ def get_grouping_variants(
     def get_primary_hash(self) -> str | None:
```
lobsterkatie, line 456: Another follow-up thing to think about: Can this ever actually return `None`? It would save a lot of noise in tests if we didn't have to assert its existence all the time.

src/sentry/grouping/api.py
```diff
@@ -374,25 +373,16 @@ def get_grouping_variants_for_event(
     return rv
 
 
-def sort_grouping_variants(variants: dict[str, BaseVariant]) -> tuple[KeyedVariants, KeyedVariants]:
-    """Sort a sequence of variants into flat and hierarchical variants"""
+def sort_grouping_variants(variants: dict[str, BaseVariant]) -> KeyedVariants:
+    """Sort a sequence of variants into flat variants"""
```
lobsterkatie, line 377:
```suggestion
    """Sort variants in order of precedence"""
```

src/sentry/grouping/api.py
```diff
@@ -374,25 +373,16 @@ def get_grouping_variants_for_event(
...
 
     flat_variants = []
-    hierarchical_variants = []
 
     for name, variant in variants.items():
-        if name in HIERARCHICAL_VARIANTS:
-            hierarchical_variants.append((name, variant))
-        else:
-            flat_variants.append((name, variant))
+        flat_variants.append((name, variant))
```
lobsterkatie, line 382: Maybe you're saving simplifications for a follow-up PR, but do we even need to do this? Doesn't this just get us a (materialized) copy of `variants.items()`?

src/sentry/grouping/component.py
```diff
@@ -41,8 +41,6 @@ def __init__(
         contributes: bool | None = None,
         values: Sequence[str | GroupingComponent] | None = None,
         variant_provider: bool = False,
-        is_prefix_frame: bool = False,
-        is_sentinel_frame: bool = False,
```
lobsterkatie, line 45: The changes up until is point are obvious - they're just removing all the hierarchical stuff. What is this prefix/sentinel frame business about? Could you maybe explain in a sentence or two in the PR description why it's going?

src/sentry/grouping/grouping_info.py
```diff
@@ -62,22 +60,21 @@ def _check_for_mismatched_hashes(
     project: Project,
     grouping_info: dict[str, dict[str, Any]],
     hashes: list[str],
-    hierarchical_hashes: list[str],
 ) -> None:
     """
-    Given a dictionary of variant data, check each variant's hash value to make sure it is one of
-    the known values from either `hashes` or `hierarchical_hashes`.
+    Given a dictionary of variant data, check each variant's hash value to make sure it is `hashes`.
```
lobsterkatie, line 65:
```suggestion
    Given a dictionary of variant data, check each variant's hash value to make sure it is in `hashes`.
```

src/sentry/grouping/ingest/config.py
```diff
@@ -86,13 +86,6 @@ def is_in_transition(project: Project) -> bool:
 
 
 def project_uses_optimized_grouping(project: Project) -> bool:
-    primary_grouping_config = project.get_option("sentry:grouping_config")
-    secondary_grouping_config = project.get_option("sentry:secondary_grouping_config")
-    has_mobile_config = "mobile:2021-02-12" in [primary_grouping_config, secondary_grouping_config]
-
-    if has_mobile_config or options.get("grouping.config_transition.killswitch_enabled"):
-        return False
```
lobsterkatie, line 94: Until we have a few days of being 100% on the optimized transition codepath, can we please keep the killswitch check here?

src/sentry/grouping/strategies/newstyle.py
```diff
@@ -541,10 +541,7 @@ def single_exception(
     rv = {}
 
     for variant, stacktrace_component in stacktrace_variants.items():
-        values = [
-            stacktrace_component,
-            system_type_component if variant == "system" else type_component,
-        ]
```
lobsterkatie, line 547: Why/how can the system component go away?

lobsterkatie (commented):
🚀

src/sentry/grouping/component.py
armenzg: This all related to hierarchical grouping. I'm going to comment further in the main PR so we have it for posterity.

src/sentry/grouping/strategies/newstyle.py
armenzg: Good catch. I will fix it.

src/sentry/eventstore/models.py
armenzg: The original code was never None:
https://github.com/getsentry/sentry/blob/492e78e0acb7fa508c53ffeea5e9c622db8202a5/src/sentry/eventstore/models.py#L385-L387

lobsterkatie: Ah, is that from before hierarchical grouping? In that case, yeah, not in this set of PRs, but it'll be great to pull that `None` from the return type and get rid of all of the non-None assertions in tests.

PR #77007, opened 2024-09-05
feat(insights): Adds create alert buttons to resource module charts

edwardgou-sentry:
Adds Create Alert buttons to resource module charts


static/app/views/insights/browser/resources/alerts.ts
```diff
@@ -0,0 +1,22 @@
+import {t} from 'sentry/locale';
+
+export const ALERTS = {
```
DominikB2014, line 3: I think it would be a good idea to type this

static/app/views/insights/browser/resources/alerts.ts
```diff
@@ -0,0 +1,22 @@
...
+  spm: {
+    aggregate: 'spm()',
+  },
+  duration: {
+    aggregate: 'avg(d:spans/duration@millisecond)',
+  },
+  decodedSize: {
+    aggregate: 'avg(d:spans/http.decoded_response_content_length@byte)',
```
DominikB2014, line 11: I'm not too sure about this idea, but maybe we should use SpanMetricsField here
```suggestion
    aggregate: `avg(d:spans/${SpanMetricsField.HTTP_DECODED_RESPONSE_CONTENT_LENGTH}@byte)`,
```

The one thing that is stopping me from being confident, is that SpanMetricsField is for discover fields, which is usually the same as the MRI, but not always.

DominikB2014 (approved):

PR #77015, opened 2024-09-05
feat(insights): add alerts to caches module

edwardgou-sentry:
Adds Create Alert buttons to caches module charts


static/app/views/insights/cache/components/charts/transactionDurationChart.tsx
```diff
@@ -83,7 +84,12 @@ export function TransactionDurationChart({
   };
 
   return (
-    <ChartPanel title={DataTitles['transaction.duration']}>
+    <ChartPanel
+      title={DataTitles['transaction.duration']}
+      alertConfigs={[
+        {...ALERTS.duration, query: MutableSearch.fromQueryObject(search).formatString()},
+      ]}
+    >
```
DominikB2014, line 92: Is there potential here for this to be misleading? This is the sample duration chart, where we plot samples transaction samples, over the transaction duration. If we have an option to create an alert on this chart, i think some would assume the alert might relate to the samples, but in reality the alert is solely based on the transaction right?

DominikB2014 (approved):

static/app/views/insights/cache/components/charts/transactionDurationChart.tsx
edwardgou-sentry: hmm, good point, I changed the name of the button to `Create Average Duration Alert` to make it more clear. Lets get some feedback and we can think of other changes if it's still misleading.

DominikB2014: Sgtm!

PR #77018, opened 2024-09-05
fix(replay): Mobile replay missing pause time

c298lee:
When pausing, if the video's current time is not passed in, it defaults to 0. This resulted in a bug where the replayer switched to the first frame after unpausing the video.

Correct play through:

https://github.com/user-attachments/assets/b21798fb-f0f0-4e27-b960-b7ed3b025db0

Before (with bug):

https://github.com/user-attachments/assets/b2ad918b-7f9a-49ed-b7df-8cc21b42517c

After (with fix):

https://github.com/user-attachments/assets/a57753a6-43bf-438d-9368-7503eec814dc

Fixes https://github.com/getsentry/team-replay/issues/465

### DIFF
--- a/static/app/components/replays/replayContext.tsx
+++ b/static/app/components/replays/replayContext.tsx
@@ -468,7 +468,7 @@ export function Provider({
       return;
     }
     if (isPlaying) {
-      replayer.pause();
+      replayer.pause(getCurrentPlayerTime());
       replayer.setConfig({speed: prefs.playbackSpeed});
       replayer.play(getCurrentPlayerTime());
     } else {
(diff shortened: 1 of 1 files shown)


ryan953:
Does this create any chance for non-mobile replays?

static/app/components/replays/replayContext.tsx
```diff
@@ -468,7 +468,7 @@ export function Provider({
       return;
     }
     if (isPlaying) {
-      replayer.pause();
+      replayer.pause(getCurrentPlayerTime());
```
michellewzhang, line 471: can we get a comment here explaining the need for this (for posterity)?

michellewzhang (approved):
do we have a example replay/screenrecording to reference the before/after?

c298lee:
> Does this create any chance for non-mobile replays?

nope!!

c298lee:
> do we have a example replay/screenrecording to reference the before/after?

updated description with example videos!

billyvg (approved):

PR #61675, opened 2023-12-13
fix(grouping): Do not consider _dispatch_client_callout as threadbase

armenzg:
When investigating issues for a customer we noticed that marking _dispatch_client_callout as threadbase would remove all frames before as not considered for grouping.

Removing the rule can cause new groups to be generated if any of the frames discarded are now considered for grouping.

PR #77019, opened 2024-09-05
drop(hierarchical): Drop hierarchical code - part 4

armenzg:
This is a follow-up to #77017

The tests get updated because I removed a rule back in #61675


tests/sentry/grouping/snapshots/test_categorization/test_categorization/block_invoke.pysnap
armenzg, line 1: Jump to the last file in this PR.

PR #77026, opened 2024-09-05
fix(grouping): Fix getting group metadata in `_save_aggregate_new`

lobsterkatie:
When we gather metadata to use when either creating or updating a group, one of the pieces of data we include is the group's title. Because this title can be affected by the grouping process (either directly, via a custom fingerprint rule, or indirectly, as a result of grouping setting `main_exception_id`), the gathering of the metadata must happen after grouping.

This ordering is correct in `_save_aggregate`, but not in `_save_aggregate_new`. This fixes that by gathering the metadata directly in `_save_aggregate_new`'s helpers, rather than calculating it ahead of time and passing it in to them.

(This problem was discovered when a number of tests involving the group's title started flaking as we've been rolling out `_save_aggregate_new`. This therefore also removes the skips on those tests, since they should now work regardless of which of `_save_aggregate` and `_save_aggregate_new` is running.)

### DIFF
--- a/src/sentry/event_manager.py
+++ b/src/sentry/event_manager.py
@@ -933,6 +933,10 @@ def _get_group_processing_kwargs(job: Job) -> dict[str, Any]:
     """
     Pull together all the metadata used when creating a group or updating a group's metadata based
     on a new event.
+
+    Note: Must be called *after* grouping has run, because the grouping process can affect the title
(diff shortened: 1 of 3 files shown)


armenzg (approved):
Amazing find!

JoshFerge (approved):
💯

PR #77030, opened 2024-09-05
fix(slack): Send linking message in bot <> user DM if bot doesn't have access to channel

iamrajjoshi:
fixes [SENTRY-3BTE](https://sentry.sentry.io/issues/5585631410/)

the bot was attempting to send an ephemeral message in a channel that it doesn't have access to, which is why slack was sending us `channel_not_found`. 

the solution here is if we can't send the ephemeral message, we instead send the user a dm to link their identity. because sending them a link to their message would be another slack api call, i instead settled on sending them the channel that they decided to send the link in, so they can jump to it.

### DIFF
--- a/src/sentry/integrations/slack/message_builder/prompt.py
+++ b/src/sentry/integrations/slack/message_builder/prompt.py
@@ -4,19 +4,30 @@ from sentry.integrations.slack.message_builder.types import SlackBody
 
 from .base.block import BlockSlackMessageBuilder
 
-LINK_IDENTITY_MESSAGE = "Link your Slack identity to Sentry to unfurl Discover charts."
+EPHEMERAL_LINK_IDENTITY_MESSAGE = "Link your Slack identity to Sentry to unfurl Discover charts."
+NON_EPHEMERAL_LINK_IDENTITY_MESSAGE = "You sent an unfurlable link in <#{}>. Please link your Slack identity to Sentry to unfurl Discover charts."
 
 
 class SlackPromptLinkMessageBuilder(BlockSlackMessageBuilder):
-    def __init__(self, url: str) -> None:
+    def __init__(self, url: str, ephemeral: bool = True, channel_name: str | None = None) -> None:
         super().__init__()
         self.url = url
+        # message_channel is the channel the user sent the unfurl link in
+        # if passed, we will add this additional context to the message
+        self.ephemeral = ephemeral
+        self.channel_name = channel_name
 
     def build(self) -> SlackBody:
+        message = (
+            EPHEMERAL_LINK_IDENTITY_MESSAGE
+            if self.ephemeral
+            else NON_EPHEMERAL_LINK_IDENTITY_MESSAGE.format(self.channel_name)
+        )
+
         return {
             "blocks": orjson.dumps(
                 [
-                    self.get_markdown_block(LINK_IDENTITY_MESSAGE),
+                    self.get_markdown_block(message),
                     self.get_action_block([("Link", self.url, "link"), ("Cancel", None, "ignore")]),
                 ],
             ).decode()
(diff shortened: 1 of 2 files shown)


src/sentry/integrations/slack/message_builder/prompt.py
```diff
@@ -4,19 +4,30 @@
 
 from .base.block import BlockSlackMessageBuilder
 
-LINK_IDENTITY_MESSAGE = "Link your Slack identity to Sentry to unfurl Discover charts."
+EPHEMERAL_LINK_IDENTITY_MESSAGE = "Link your Slack identity to Sentry to unfurl Discover charts."
+NON_EPHEMERAL_LINK_IDENTITY_MESSAGE = "You sent an unfurlable link in <#{}>. Please link your Slack identity to Sentry to unfurl Discover charts."
```
sentaur-athena, line 8: nit: kind of personal preference comment but I prefer using lambda and "f" for formatting and including a variable name instead of <#{}> . I think it's more clear that way.

src/sentry/integrations/slack/webhooks/event.py
```diff
@@ -116,7 +118,25 @@ def prompt_link(self, slack_request: SlackDMRequest) -> None:
                 text=payload["text"],
                 **SlackPromptLinkMessageBuilder(associate_url).as_payload(),
             )
-        except SlackApiError:
+        except SlackApiError as e:
+            # If the channel is not found, it could be because the bot doesn't have access to the channel to send an ephemeral message
+            # In this case, we should send the message to the user in the bot <> user DM
+            if unpack_slack_api_error(e) == CHANNEL_NOT_FOUND:
```
sentaur-athena, line 124: Do other messages other than link your account go through this logic? If so, what if the channel is actually not found because it's deleted or something? Would it still make sense to get a DM?

GabeVillalobos:
Should we consider implementing some sort of debounce or opt out for this? I could see this becoming obnoxious for a user who doesn't want to link their slack identity.

src/sentry/integrations/slack/webhooks/event.py
iamrajjoshi: link your account is the only flow that goes through this logic. the only other cases we would get a CHANNEL_NOT_FOUND error is if the channel was deleted between us getting the unfurl webhook and sending the message and the channel id slack sent to us being valid, both extremely unlikely.

src/sentry/integrations/slack/message_builder/prompt.py
iamrajjoshi: the <# is so slack formats the channel id into the real channel name. i wanted to keep this a constant outside, so i call .format() on the message to inject the channel_id.

i also forgot to link the video recording, which i did just now.

src/sentry/integrations/slack/message_builder/prompt.py
```diff
@@ -4,19 +4,30 @@
...
 
 
 class SlackPromptLinkMessageBuilder(BlockSlackMessageBuilder):
-    def __init__(self, url: str) -> None:
+    def __init__(self, url: str, ephemeral: bool = True, channel_name: str | None = None) -> None:
         super().__init__()
         self.url = url
+        # message_channel is the channel the user sent the unfurl link in
```
cathteng, line 15: should this comment be updated?

src/sentry/integrations/slack/webhooks/event.py
```diff
@@ -116,7 +118,25 @@ def prompt_link(self, slack_request: SlackDMRequest) -> None:
...
+                service = SlackService.default()
+                service.send_message_to_slack_channel(
```
cathteng, line 126: should this be using `chat.postMessage`? from my understanding, `send_message_to_slack_channel` is used for activity or alert rule notifications. although i'm not sure what the difference is since it calls `chat.postMessage `anyway. maybe `send_message_to_slack_channel` can be retired because `SlackSdkClient` doesn't use the integration proxy?

https://github.com/getsentry/sentry/blob/ad982c43006909dbec5afc0e256c011b73755f86/src/sentry/integrations/slack/service.py#L336-L337

PR #77032, opened 2024-09-05
fix(router): Redirect to pathname instead

Zylphrex:
React Router 3 was able to handle the redirect to an absolute path but React Router 6 assumes the redirect is either a pathname or a relative path. So the absolute url was being treated as a relative path and was concatenated to the end of the current route. To get around this, parse the url and only pass the pathname component to the redirect.

### DIFF
--- a/static/app/views/projectEventRedirect.tsx
+++ b/static/app/views/projectEventRedirect.tsx
@@ -51,15 +51,14 @@ function ProjectEventRedirect({router}: Props) {
       // after following any redirects. It _should_ be the page we're trying
       // to reach; use the router to go there.
       //
-      // Here we assume that the responseURL is on the same domain as the
-      // as the window.location.origin, so we simply take the pathname.
-      //
       // Use `replace` so that hitting the browser back button will skip all
       // this redirect business.
       const url = new URL(xhr.responseURL);
       if (url.origin === window.location.origin) {
         router.replace(url.pathname);
       } else {
+        // If the origin has changed, we cannot do a simple replace with the router.
+        // Instead, we opt to do a full redirect.
         window.location.replace(xhr.responseURL);
       }
     };
--- a/static/app/views/projectEventRedirect.tsx
+++ b/static/app/views/projectEventRedirect.tsx
@@ -57,7 +57,11 @@ function ProjectEventRedirect({router}: Props) {
       // Use `replace` so that hitting the browser back button will skip all
       // this redirect business.
       const url = new URL(xhr.responseURL);
-      router.replace(url.pathname);
+      if (url.origin === window.location.origin) {
+        router.replace(url.pathname);
+      } else {
+        window.location.replace(xhr.responseURL);
+      }
     };
     xhr.onerror = () => {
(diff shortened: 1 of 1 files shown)


static/app/views/projectEventRedirect.tsx
```diff
@@ -50,10 +50,14 @@ function ProjectEventRedirect({router}: Props) {
...
-
+      //
+      // Here we assume that the responseURL is on the same domain as the
+      // as the window.location.origin, so we simply take the pathname.
+      //
       // Use `replace` so that hitting the browser back button will skip all
       // this redirect business.
-      router.replace(xhr.responseURL);
+      const url = new URL(xhr.responseURL);
+      router.replace(url.pathname);
```
evanpurkhiser, line 60: This loses search parameters, are we OK with that?

Zylphrex: Good point, it looks like it is possible there are query params and query hash in the responseURL I'll add it back

Zylphrex: nvm, looking at the backend some more, it's guaranteed to only not contain query params or query hash.

static/app/views/projectEventRedirect.tsx
```diff
@@ -50,10 +50,17 @@ function ProjectEventRedirect({router}: Props) {
...
       // Use `replace` so that hitting the browser back button will skip all
       // this redirect business.
-      router.replace(xhr.responseURL);
+      const url = new URL(xhr.responseURL);
+      if (url.origin === window.location.origin) {
+        router.replace(url.pathname);
+      } else {
+        // If the origin has changed, we cannot do a simple replace with the router.
+        // Instead, we opt to do a full redirect.
+        window.location.replace(xhr.responseURL);
```
Zylphrex, line 62: This doesn't work yet until the right CORs headers are set

evanpurkhiser (approved):

static/app/views/projectEventRedirect.tsx
markstory: What isn't correct with cors headers currently?

Zylphrex: After debugging this more closely, this isn't where the CORs error is happening. The only way I was able to reproduce this CORs error is by using webpack dev server to reroute the 302 to go to a different subdomain.

e.g.
1. start on `org1.sentry.io/org1/proj1/events/abcdef`
2. force the response to the HEAD request to redirect to `org2.sentry.io/issues/123/`
3. the HEAD request will follow the redirect and make a preflight request where the OPTIONS request is missing all the CORS headers which i believe is due to [this](https://github.com/getsentry/sentry/blob/f24f9d53e0bd1cd94a8850c9086a848fa618bef8/src/sentry/api/base.py#L116-L118) which skips all the CORS headers when it's an OPTIONS request

I'm not sure this is an actual case in practice.

Zylphrex:
Going to merge this as it at least fixes the in product links and I'll do some more testing to verify the CORS errors.

PR #77033, opened 2024-09-05
feat(saml2): Return SLO URL to the frontend

leedongwei:
Our original implementation would call the IdP's SLO endpoint, but it seems like some IdPs do not invalidate user sessions and we'd need to redirect the user's browser on the frontend back to the IdP so they can clear their browser cookies.

This PR returns the SLO URL to the frontend, so the frontend can redirect the user to the correct page.


saponifi3d (approved):
lgtm.

PR #77034, opened 2024-09-05
chore(similarity): Turn on AI grouping for all new projects

jangjodi:
Remove EA org check before adding project option on project creation
The project option will enable the ai similarity embedding feature to be used for grouping new groups


JoshFerge (approved):
🚀

PR #77036, opened 2024-09-05
feat(saml2): Redirect to SLO URL on the frontend

leedongwei:
Related to https://github.com/getsentry/sentry/pull/77033

If a SLO URL is returned by the logout API, we will redirect the browser to that URL. This is needed for Single-Logout to work, because we need the browser to clear the cookies that are attached to the IdP's domain.

### DIFF
--- a/static/app/actionCreators/account.tsx
+++ b/static/app/actionCreators/account.tsx
@@ -50,7 +50,7 @@ export async function logout(api: Client, redirectUrl = '/auth/login/') {
   const data = await api.requestPromise('/auth/', {method: 'DELETE'});
 
   // If there's a URL for SAML Single-logout, redirect back to IdP
-  window.location.replace(data?.sloUrl || redirectUrl);
+  window.location.assign(data?.sloUrl || redirectUrl);
 }
 
 export function removeAuthenticator(api: Client, userId: string, authId: string) {
--- a/static/app/actionCreators/account.tsx
+++ b/static/app/actionCreators/account.tsx
@@ -46,8 +46,11 @@ export function updateUser(user: User | ChangeAvatarUser) {
   ConfigStore.set('user', {...previousUser, ...user, options});
 }
 
-export function logout(api: Client) {
-  return api.requestPromise('/auth/', {method: 'DELETE'});
+export async function logout(api: Client, redirectUrl = '/auth/login/') {
+  const data = await api.requestPromise('/auth/', {method: 'DELETE'});
+
+  // If there's a URL for SAML Single-logout, redirect back to IdP
+  window.location.replace(data?.sloUrl || redirectUrl);
 }
 
 export function removeAuthenticator(api: Client, userId: string, authId: string) {
--- a/static/app/actionCreators/account.tsx
+++ b/static/app/actionCreators/account.tsx
@@ -1,3 +1,5 @@
+import {redirect} from 'react-router-dom';
+
 import {addErrorMessage, addSuccessMessage} from 'sentry/actionCreators/indicator';
 import {Client} from 'sentry/api';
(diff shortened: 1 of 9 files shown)


static/app/views/acceptOrganizationInvite/index.spec.tsx
```diff
@@ -245,11 +245,8 @@ describe('AcceptOrganizationInvite', function () {
     );
 
     expect(screen.getByTestId('existing-member')).toBeInTheDocument();
-
     await userEvent.click(screen.getByTestId('existing-member-link'));
-
-    expect(logout).toHaveBeenCalled();
-    await waitFor(() => expect(window.location.replace).toHaveBeenCalled());
```
leedongwei, line 252: Removing this line because redirect was moved into `logout` function, which was mocked.

static/app/views/acceptOrganizationInvite/index.spec.tsx
```diff
@@ -294,9 +291,7 @@ describe('AcceptOrganizationInvite', function () {
 
     expect(screen.getByTestId('existing-member')).toBeInTheDocument();
     await userEvent.click(screen.getByTestId('existing-member-link'));
-
-    expect(logout).toHaveBeenCalled();
-    await waitFor(() => expect(window.location.replace).toHaveBeenCalled());
```
leedongwei, line 299: Removing this line because redirect was moved into `logout` function, which was mocked.

static/app/actionCreators/account.tsx
```diff
@@ -46,8 +46,11 @@ export function updateUser(user: User | ChangeAvatarUser) {
   ConfigStore.set('user', {...previousUser, ...user, options});
 }
 
-export function logout(api: Client) {
-  return api.requestPromise('/auth/', {method: 'DELETE'});
+export async function logout(api: Client, redirectUrl = '/auth/login/') {
+  const data = await api.requestPromise('/auth/', {method: 'DELETE'});
+
+  // If there's a URL for SAML Single-logout, redirect back to IdP
+  window.location.assign(data?.sloUrl || redirectUrl);
```
saponifi3d, line 53: should we use react-router here for consistency?

```jsx
import { redirect } from "react-router-dom";

redirect(data?.sloUrl || redirectUrl))
```

static/app/components/superuserStaffAccessForm.tsx
```diff
@@ -163,21 +163,15 @@ class SuperuserStaffAccessForm extends Component<Props, State> {
...
-    const authLoginPath = `/auth/login/?next=${encodeURIComponent(window.location.href)}`;
+  handleLogout = () => {
+    let nextUrl = `/auth/login/?next=${encodeURIComponent(window.location.href)}`;
+
     const {superuserUrl} = window.__initialData.links;
     if (window.__initialData?.customerDomain && superuserUrl) {
-      const redirectURL = `${trimEnd(superuserUrl, '/')}${authLoginPath}`;
-      window.location.assign(redirectURL);
-      return;
+      nextUrl = `${trimEnd(superuserUrl, '/')}${nextUrl}`;
```
saponifi3d, line 171: nit: might be safer to use the `URL` constructor rather than build a string for the url. here's a little blog on how to use the `URL` constructor and the differences between the two approaches: https://www.builder.io/blog/new-url

saponifi3d (approved):
lgtm

just a nit for the `onClick={() = {}}` functions in handlers -- that will be a little harder to debug if there's an issue (errors will just have `anonymous`) and the browsers js engine won't be able to create a hot cache of the anon functions for the react re-render cycles.

saponifi3d (approved):
:shipit:

PR #77039, opened 2024-09-05
fix(favicons):logo in dark and light mode

doralchan:
**Before:**
Screenshot 2024-09-06 at 8 53 08 AM

**After:**
Screenshot 2024-09-06 at 12 19 20 PM

In order, Light Mode, Dark Mode and Dev Mode:
Screenshot 2024-09-06 at 12 17 55 PM


evanpurkhiser:
Could you add some screenshots of what this looks like in the browser tab?

evanpurkhiser:
I wonder if we should be providing various icon sizes that are each optimized for their own sizes?

static/app/utils/useColorscheme.tsx
```diff
@@ -21,7 +21,7 @@ function setFaviconTheme(theme: 'dark' | 'light'): void {
   }
 
   const path = faviconNode.href.split('/sentry/')[0];
-  const iconName = theme === 'dark' ? 'favicon-dark' : 'favicon';
+  const iconName = theme === 'dark' ? 'favicon-dev' : 'favicon';
```
evanpurkhiser, line 24: I'm not a fan of this. Now there's no easy way to differentiate between production sentry in dark mode and dev mode sentry.

doralchan: Oh that's a good point... ok, I'll get another color

scttcper (approved):

scttcper:
no dash 😢

PR #77042, opened 2024-09-05
feat(workflows): Set up Django app for `workflow_engine` project

wedamija:
Set up a basic skeleton for the workflow engine project and configure the app to load


saponifi3d (approved):

evanpurkhiser (approved):

schew2381:
🐴🥺

wedamija:
> 🐴🥺
We decided it's probably safer to go with the boring name since the two were tied unfortunately!

PR #77047, opened 2024-09-05
ref(ui): Use object style arguments with react query

scttcper:
passing queryKey or the mutation function as the first argument and options as the 2nd is gone in react query v5


michellewzhang (approved):

priscilawebdev (approved):
lgtm

PR #77051, opened 2024-09-05
ref(releases): move commit author and repo creation out of db transaction

JoshFerge:
split out the author and repo creation from the main transaction of setting release commits. this means we can remove hybrid cloud calls from inside the database transaction, and get rid of the use of `in_test_hide_transaction_boundary` which is good.

### DIFF
--- a/src/sentry/models/release.py
+++ b/src/sentry/models/release.py
@@ -1,11 +1,8 @@
 # Sort commit list in reverse order
 from __future__ import annotations
 
-import itertools
 import logging
-import re
 from collections.abc import Mapping, Sequence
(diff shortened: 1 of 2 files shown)


mrduncan (approved):
lgtm

If you're making more changes to it in the future, `create_repositories` can likely now use `get_or_create` to save a query since that it's no longer in this transaction

JoshFerge:
note that the source of the hybrid cloud calls was not this, but a receiver on release commit. so we cannot remove the ignore transaction boundary decorator. will attempt to remove, but may be more difficult.

PR #77052, opened 2024-09-06
ref(issue-stream): Pull IssueListBody into its own component

MichaelSun48:
This PR pulls the Issue Stream's body into a separate component, `<IssueListBody/>`. This is necessary to allow a context provider to wrap the issue stream's body component, which is will be needed to support custom view's add view flow feature. Since this refactors the second highest traffic page in Sentry, I think it's better that this change is made in its own PR.  

Reviewers: please triple check that I have passed the correct props to this new component. AFAICT, the issue stream appears to work perfectly fine after this change, but I definitely want to be absolutely sure that's the case.

### DIFF
--- a/static/app/views/issueList/issueListBody.tsx
+++ b/static/app/views/issueList/issueListBody.tsx
@@ -1,4 +1,9 @@
+import {Fragment} from 'react';
+import styled from '@emotion/styled';
+
 import type {IndexedMembersByProject} from 'sentry/actionCreators/members';
+import type {CursorHandler} from 'sentry/components/pagination';
+import Pagination from 'sentry/components/pagination';
 import Panel from 'sentry/components/panels/panel';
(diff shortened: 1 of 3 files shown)


malwilley (approved):

static/app/views/issueList/issueListBody.tsx
```diff
@@ -0,0 +1,115 @@
...
+            onSortChange={onSortChange}
+          />
+        )}
+        <PanelBody>
+          <VisuallyCompleteWithData
+            hasData={groupIds.length > 0}
+            id="IssueList-Body"
+            isLoading={issuesLoading}
+          >
+            <GroupListBody
```
malwilley, line 86: Oh just noticed that this is named very similarly. Maybe we could rename this one to IssueListTable or something?

PR #77056, opened 2024-09-06
feat(mobile-onboarding): Add replay verify step

ArthurKnaus:
Add a verify step for the replay onboarding of mobile platforms.

### iOS & Android
Screenshot 2024-09-09 at 09 20 56

### React-Native & Capacitor
Screenshot 2024-09-09 at 09 20 45

### DIFF
--- a/static/app/gettingStartedDocs/android/android.tsx
+++ b/static/app/gettingStartedDocs/android/android.tsx
@@ -431,7 +431,33 @@ const replayOnboarding: OnboardingConfig<PlatformOptions> = {
       ],
     },
   ],
-  verify: () => [],
+  verify: () => [
+    {
+      type: StepType.VERIFY,
+      description: t(
+        'To verify your Replay setup, trigger an error in your app and watch Sentry capture the event along with a recording of the user interaction.'
+      ),
+      configurations: [
+        {
+          description: tct(
+            "You can simulate an error by adding the following code to your app's [code: MainActivity]:",
+            {code: <code />}
+          ),
+          code: [
+            {
+              label: 'Kotlin',
+              value: 'kotlin',
+              language: 'kotlin',
+              code: getVerifySnippet(),
+            },
+          ],
+          additionalInfo: t(
+            'After clicking the button, wait a few moments, and you\'ll see a new session appear on the "Replays" page.'
+          ),
+        },
+      ],
+    },
+  ],
   nextSteps: () => [],
 };
 
--- a/static/app/gettingStartedDocs/android/android.tsx
+++ b/static/app/gettingStartedDocs/android/android.tsx
@@ -12,7 +12,10 @@ import type {
 } from 'sentry/components/onboarding/gettingStartedDoc/types';
 import {MobileBetaBanner} from 'sentry/components/onboarding/gettingStartedDoc/utils';
 import {getAndroidMetricsOnboarding} from 'sentry/components/onboarding/gettingStartedDoc/utils/metricsOnboarding';
-import {getReplayMobileConfigureDescription} from 'sentry/components/onboarding/gettingStartedDoc/utils/replayOnboarding';
+import {
+  getReplayMobileConfigureDescription,
+  getReplayVerifyStep,
+} from 'sentry/components/onboarding/gettingStartedDoc/utils/replayOnboarding';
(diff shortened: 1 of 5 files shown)


matejminar (approved):
Code looks good. Will leave the content review for @kahest 🙏

static/app/gettingStartedDocs/capacitor/capacitor.tsx
```diff
@@ -63,6 +63,12 @@ const platformOptions: Record<PlatformOptionKey, PlatformOption> = {
 type PlatformOptions = typeof platformOptions;
 type Params = DocsParams<PlatformOptions>;
 
+const getVerifyReplaySnippet = () => `
+setTimeout(() => {
+  throw new Error("Sentry Test Error");
+});
```
krystofwoldrich, line 69: I think button snippet would provide better first replay than `setTimeout`.

The same as we have for iOS and Android.

ArthurKnaus: I took this one from sentry-docs...
I guess we would need 3 different snippets. One for `Angular`, `React` and `Vue`?

krystofwoldrich: Looks like a wrong link. Yeah I see, it would make sense to add for Angular, React and Vue, but only if we already have some tab UI ready for these situations.

Otherwise I would just `throw new Error("Sentry Test Error");` and in desc tell users to add it to onClick.

ArthurKnaus: The link was added by our GitHub integration 😅 (as there is a project called `sentry-docs`)
There is some UI for it and in those 3 we will soon have docs that add a button. So I can re-use them 👍

krystofwoldrich:
Just a nit. RN as the only guide doesn't have a code snippet, was the intentional?

All the mobile replay integrations can be verified just by opening the app and seeing the session replay, so we could omit the code snippet in all.

OR

If the goal is to show the connection of Replayt and Errors I would add the code snippet to RN too.

(from `@sentry/wizard`)

```jsx
<Button title='Try!' onPress={ () => { Sentry.captureException(new Error('First error')) }}/>
```

ArthurKnaus:
@krystofwoldrich 
> Just a nit. RN as the only guide doesn't have a code snippet, was the intentional?

Again, I took the content from the docs page. Would prefer to have a code snippet there.
Ideally, we can showcase the replay <-> error connection.

I am not experienced with react native. Where should they place the code snippet? I assume simply in `one of your components`?

krystofwoldrich:
> I assume simply in one of your components?

Yes, exactly, the same as Web React.

static/app/gettingStartedDocs/android/android.tsx
```diff
@@ -431,7 +431,33 @@ const replayOnboarding: OnboardingConfig<PlatformOptions> = {
       ],
     },
   ],
-  verify: () => [],
+  verify: () => [
+    {
+      type: StepType.VERIFY,
+      description: t(
+        'To verify your Replay setup, trigger an error in your app and watch Sentry capture the event along with a recording of the user interaction.'
```
krystofwoldrich, line 438: Since we suggest `session_sample_rate = 1` we can just say open the app and try to add error, as an `optional`.

Something like this:

`To verify your Replay setup, open your app and wait for the session replay to be uploaded to Sentry.`

`To see how errors are sessions are connected, try triggering an error by adding the following code:`

`'After clicking the button, wait a few moments, and you\'ll see the error linked to the session recording.'`

ArthurKnaus:
@krystofwoldrich I updated the verify steps to what was agreed on with @bruno-garcia

static/app/gettingStartedDocs/apple/ios.tsx
```diff
@@ -713,7 +716,12 @@ const replayOnboarding: OnboardingConfig<PlatformOptions> = {
       ],
     },
   ],
-  verify: () => [],
+  verify: getReplayVerifyStep({
+    replayOnErrorSampleRateName:
+      'options\u200b.experimental\u200b.sessionReplay\u200b.onErrorSampleRate',
```
ArthurKnaus, line 721: Added zero width spaces here to allow wrapping lines at each segment. Otherwise it messes up the layout.

krystofwoldrich (approved):
Thank you! 🚀

Looks good.

PR #77060, opened 2024-09-06
chore(metrics): remove span attribute extraction models from django state

shellmayr:
Removes the models SpanAttributeExtractionRuleConfig and SpanAttributeExtractionRuleCondition and all testing related artifacts from the codebase.


markstory (approved):
Migration changes look good.

PR #77074, opened 2024-09-06
fix(feedback): limit user report related queries by retention period

### LINKED ISSUE #76985: Limit time range of user report endpoint queries
[SENTRY-18JF](https://sentry.sentry.io/issues/4620140170/). QueryOutsideRetentionError (handled)
[group_user_reports.py](https://github.com/getsentry/sentry/blob/b586e5fad7a76755323e6b03432e85ddc00fca86/src/sentry/api/endpoints/group_user_reports.py#L14-L14)
[project_user_reports.py](https://github.com/getsentry/sentry/blob/b586e5fad7a76755323e6b03432e85ddc00fca86/src/sentry/api/endpoints/project_user_reports.py#L64-L64)

Serializing user reports leads to a [query](https://github.com/getsentry/sentry/blob/52153f58c2591089b8c4d5aa30c74df5ab6682c6/src/sentry/api/serializers/models/userreport.py#L18-L18) for their associated events. Sometimes an event is expired past our retention period, causing the endpoints to error out. There's no way for users to specify a time range, so we should automatically filter out reports with expired events.

Helper: [outside_retention_with_modified_start](https://github.com/getsentry/sentry/blob/b586e5fad7a76755323e6b03432e85ddc00fca86/src/sentry/utils/dates.py#L169-L169)

aliu39:
Fixes https://github.com/getsentry/sentry/issues/76985

Combined with https://github.com/getsentry/sentry/pull/76691 will fix all time range errors we've been getting for user reports. Tldr; we should explicitly respect event retention for all UserReport-related queries. **Reports older than the retention period won't be returned by endpoints.**

In the rare case where a report is in retention but its associated event is not, the serialized report will exclude the user object, because we can't query the event.

I think these are all resolved but linking them for reference:
Fixes [SENTRY-18JF](https://sentry.sentry.io/issues/4620140170/)
Fixes [SENTRY-163P](https://sentry.sentry.io/issues/4502227101/)
Fixes [SENTRY-3CAB](https://sentry.sentry.io/issues/5635234171/)
Fixes [SENTRY-189J](https://sentry.sentry.io/issues/4613821445/)


JoshFerge:
should we wait to merge the other PR for more test coverage on this? but looks good.

JoshFerge (approved):

PR #61475, opened 2023-12-08
build: replace webpack with rspack

anonrig:
## Summary

This pull-request removes babel and Webpack in favor of Rspack and SWC. This change comes with a significant performance boost and improves our productivity. Currently, it takes 14 seconds to build Sentry. Prior to this PR, it took 58 seconds.

## Before

We are using Babel and Webpack for building our frontend application. It took almost 58 seconds on my M2 machine to run `yarn build-production` command.

```
webpack 5.87.0 compiled with 1 warning in 56351 ms
✨  Done in 57.91s.
```

## After

We removed 14 dependencies and added 3 new dependencies. Depending on Babel and Jest, we can remove more. Currently it takes 14 seconds to build, minify both JS and CSS files.

```
Rspack compiled successfully in 12.90 s
✨  Done in 14.00s.
```

## TODOs:

- [ ] In order to not break the Vercel previews we need the following change of Rspack.
  - https://github.com/web-infra-dev/rspack/issues/4977
- [ ] Rspack doesn't support lazy compilation, but [it is in their roadmap.](https://www.rspack.dev/misc/roadmap.html#lazy-compilation-support)
- [ ] Research why we need `build-js-loader.ts` script since it brings an unnecessary dependency: `terser`
- [ ] Research the relation of Babel and Jest and remove unnecessary configs from `babel.config`.
- [ ] Make `tsc` child process DX a little bit better.
- [ ] Fix error from `downsample` npm library in production build
  - https://github.com/web-infra-dev/rspack/issues/4980 

Thanks @hardfist for helping with this PR!

PR #77077, opened 2024-09-06
build(ui): Add rspack as webpack alternative

scttcper:
Building off the work done in #61475 I think the plan would be to use this as a dev-ui replacement until we're more confident and can setup the getsentry repo as well (if you aren't familiar we have another repo with billing pages etc).

Potential adoption roadmap
1. merge rspack as optional dev only command yarn dev-ui-rspack
1. make rspack default for dev-ui
1. use rspack for acceptance tests
1. build production with rspack
1. remove webpack

## Summary

|                  | Webpack     | Rspack      |
| ---------------- | ----------- | ----------- |
| HMR Time         | 4s   | 607ms       |
| Dev Memory Usage | 3.85 GB     | 2.8 GB      |
| Build Production | 131.53s | 39.42s |

## RSPACK + SENTRY TODO (updated Oct 7th, 2024)

- [x] Investigate emotion autoLabel in development mode. Currently set to `autoLabel: 'always',` Could just set label setting based on node_env.
- [x] Add back webpack in order to merge (getsentry repo will block until it can accept a rspack file instead of webpack)
- [x] Figure out a path forward for translation string extraction that use webpack + babel's AST `make build-js-po`
- [x] Chartcuterie is failing for some reason

Thanks @hardfist for helping with this PR!


scttcper:
Just realized its working for me locally because i fixed [platformicons](https://github.com/getsentry/platformicons) locally will need to publish a fix to the svg paths there

evanpurkhiser (approved):
slam it

rspack.config.ts
```diff
@@ -0,0 +1,786 @@
...
+  'sentry',
+  'locale',
+  'catalogs.json'
+);
+
+type LocaleCatalog = {
+  supported_locales: string[];
+};
+
+const localeCatalog: LocaleCatalog = JSON.parse(
```
JonasBa, line 131: Outside of the scope of this PR, but it would be interesting if we could time this locale procedure to see how much it contributes to the startup cost.

JonasBa (approved):
Those HMR numbers look 🔥

PR #77079, opened 2024-09-06
ref(scm): search endpoint abstraction round 2

cathteng:
Fixed version of https://github.com/getsentry/sentry/pull/76627

### DIFF
--- a/src/sentry/integrations/vsts/search.py
+++ b/src/sentry/integrations/vsts/search.py
@@ -1,60 +1,31 @@
-from typing import Any
-
-from rest_framework.request import Request
 from rest_framework.response import Response
 
-from sentry.api.api_owners import ApiOwner
-from sentry.api.api_publish_status import ApiPublishStatus
 from sentry.api.base import control_silo_endpoint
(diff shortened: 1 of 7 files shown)


src/sentry/integrations/vsts/search.py
```diff
@@ -1,60 +1,37 @@
...
-            return Response({"detail": "query is a required parameter"}, status=400)
-
-        installation = integration.get_installation(organization.id)
-        assert isinstance(installation, VstsIntegration), installation
-
-        if field == "externalIssue":
-            if not query:
-                return Response([])
-
-            resp = installation.search_issues(query=query)
```
cathteng, line 49: vsts doesn't query issues with repository, needed to account for that. there isn't a test for VSTS, may need to add that?

sentaur-athena (approved):

src/sentry/integrations/vsts/search.py
sentaur-athena: Yeah makes sense. Please at least add barebone tests.

cathteng: done!

PR #77081, opened 2024-09-06
feat(toolbar): initial views for auth on other sites

aliu39:
Fixes https://github.com/getsentry/sentry-toolbar/issues/2
Fixes https://github.com/getsentry/sentry-toolbar/issues/3

Working auth flow PoC for installing the devtoolbar on 3rd party sites! Forgoes oauth.
(My understanding of how this works:) Load, from Sentry, a login pop-up and an iframe to make API requests. Both have origin=sentry.io to bypass CORS. Once logged in, the session cookie is sufficient to authorize the requests, no need for tokens.


src/sentry/templates/sentry/toolbar/iframe.html
ryan953, line 1: this is the wrong file content. This is the file content for the popup success view.

iframe content, to test with, should be this `fetch()` call or something like that: https://github.com/getsentry/sentry-toolbar/issues/3

ryan953: we need both views to be able to test things the way i'm thinking

aliu39:
To test:
1. Run the dev server with this branch checked out on port 8000.

2. Create a new file on your computer called test-site.html (or really, called anything) and copy+paste this html+js content into it:
```
<html>
  <head>
    <title>Toolbar iFrame test</title>
  </head>
  <body>

    <iframe src="http://dev.getsentry.net:8000/toolbar/iframe"></iframe>

    <div>
      <button id="login" onClick="trigger_auth()">Authorize Toolbar</button>
    </div>

    <script>
      const SENTRY_HOST = 'http://dev.getsentry.net:8000';
      const ORG_NAME = 'sentry';

      function trigger_auth() {
        const authPath = `/organizations/${ORG_NAME}/toolbar/login-success/`;
        open(`${SENTRY_HOST}${authPath}`, 'sentry-toolbar-auth-popup', 'popup=true');
      }
    </script>
  </body>
</html>
```

4. Run a quick server to see the test html file. first cd into the folder where that html file is, then run: `python3 -m http.server 8080`. Load `http://dev.getsentry.net:8080/test-site.html` in your browser.

5. Click the "Authorize toolbar" button, we expect to see the popup present the login screen if you are not logged in. Once complete it should redirect to a success page which you can close. 

6. Reload test-site.html and verify that the API call to organizations/?member=1 (called from iframe.html) works w/code 200 success!

src/sentry/web/urls.py
```diff
@@ -146,6 +148,16 @@
         ErrorPageEmbedView.as_view(),
         name="sentry-error-page-embed",
     ),
+    re_path(
+        r"^toolbar/iframe$",
```
ryan953, line 152: Lets also nest this under `/organizations/<org_slug>`

and we should followup and make sure we're writing it the correct way so when subdomains are enabled (like on sentry.io) things are still working, or at least not auto-redirecting into the subdomain.

src/sentry/toolbar/iframe_view.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class IframeView(View):
+    @method_decorator(csrf_exempt)
+    def dispatch(self, request: HttpRequest) -> HttpResponse:
+        if request.method != "GET":
+            return HttpResponse(status=405)
+
+        response = render_to_response("sentry/toolbar/iframe.html")
+        response["X-Frame-Options"] = "ALLOWALL"
```
markstory, line 16: This could be a good place to hook in the organization's allowed domain list.

src/sentry/web/urls.py
markstory: :+1: for having it organization scoped.

src/sentry/toolbar/iframe_view.py
ryan953: Yes! We're going to be able to use the referer header to know the origin that requested the iframe, and then will check that against an allowlist, also make the header more specific in case it's cached, and there'll be a postMessage call in the template also scoped to the allowed domain.

src/sentry/toolbar/iframe_view.py
```diff
@@ -0,0 +1,12 @@
+from django.utils.decorators import method_decorator
+from django.views.decorators.csrf import csrf_exempt
+
+from sentry.web.frontend.base import BaseView
+
+
+class IframeView(BaseView):
+    @method_decorator(csrf_exempt)
+    def get(self, request, organization_slug):
```
ryan953, line 9: Do we need `csrf_exempt` for a `get` handler?

src/sentry/web/urls.py
```diff
@@ -146,6 +148,16 @@
...
+    re_path(
+        r"^organizations/(?P<organization_slug>[^/]+)/toolbar/iframe$",
+        IframeView.as_view(),
+        name="sentry-toolbar-iframe",
+    ),
+    re_path(
+        r"^organizations/(?P<organization_slug>[^/]+)/toolbar/login-success/$",
+        LoginSuccessView.as_view(),
+        name="sentry-toolbar-login-success",
+    ),
```
ryan953, line 160: These could go down near line 888 where it defines the `organizations` prefix:
https://github.com/getsentry/sentry/blob/6e85c069edd4881d52171a39ead697bea538fb1d/src/sentry/web/urls.py#L887-L889

And the `name` prop could be updated to match the format that's in there, `sentry-organization-foo-bar`

src/sentry/web/urls.py
```diff
@@ -146,6 +148,16 @@
         ErrorPageEmbedView.as_view(),
         name="sentry-error-page-embed",
     ),
+    re_path(
+        r"^organizations/(?P<organization_slug>[^/]+)/toolbar/iframe$",
```
ryan953, line 152: needs a trailing slash in the url

```suggestion
        r"^organizations/(?P<organization_slug>[^/]+)/toolbar/iframe/$",
```

src/sentry/web/urls.py
```diff
@@ -1025,6 +1027,17 @@
...
+                re_path(
+                    r"^(?P<organization_slug>[^/]+)/toolbar/iframe/$",
+                    IframeView.as_view(),
+                    name="sentry-organization-toolbar-iframe",
+                ),
+                re_path(
+                    r"^(?P<organization_slug>[^/]+)/toolbar/login-success/$",
+                    LoginSuccessView.as_view(),
+                    name="sentry-organization-toolbar-login-success",
+                ),
```
ryan953, line 1040: Actually, lets change these again!

Lets do `/toolbar/<org>/<project>/iframe/`  & `/toolbar/<org>/<project>/login-success/`  so we have:
- our own namespace
- both the vars in there (better with own namespace)
- no confusion around whether we need to pair them up like: `/organizations/<org>/toolbar/*/` and `/toolbar/*/`

src/sentry/toolbar/iframe_view.py
markstory: GET requests don't check CSRF tokens.

ryan953 (approved):
Woo, looks ready!

ryan953:
This is ready to be merged. There's no need to mess with the js and postMessage calls right now.

ryan953:
@aliu39 can we peel this back to stick more closely to what's in the tickets?

The tickets describe stepping stones that are obviously not the final views and templates that we want to endup with. They also have a test strategy written down and clear decisions/objectives that we can achieve before moving on to solve the next problems. For example, merging what we had on monday helps to lock-in the url patterns so we can document them and rely on them in other places.

src/sentry/templates/sentry/toolbar/iframe.html
```diff
@@ -0,0 +1,17 @@
...
+<html>
+<head>
+  <title></title>
+</head>
+<body>
+  {% script %}
+  <script>
+  (function() {
+    fetch('http://dev.getsentry.net:8000/api/0/organizations/?member=1');
+  })();
```
ryan953, line 13: We don't need to commit references to `dev.getsentry.net` even for testing.

```suggestion
```

src/sentry/toolbar/iframe_view.py
```diff
@@ -0,0 +1,40 @@
...
+        response = super().respond(template, context=context, status=status)
+        for header, val in IframeView.security_headers.items():
+            response[header] = val
+        return response
+
+    def handle_auth_required(self, request: HttpRequest, *args, **kwargs):
+        # Override redirects to login
+        if request.method == "GET":
+            self.default_context = {}
+            return self.respond("sentry/toolbar/iframe-no-auth.html", status=401)
```
ryan953, line 29: lets return the same `iframe.html` template in all cases for now. We have the python code here to render different templates which is great, and when we have different content we can make the new files.

One thing that might happen is... we have html content that's very similar in each case, and instead of having a new template each time we can use template variables to tweak the shared template.

src/sentry/toolbar/login_success_view.py
```diff
@@ -1,14 +1,13 @@
...
 
 
 @region_silo_view
-class LoginSuccessView(ProjectView):
+class LoginSuccessView(OrganizationView):
     def handle_permission_required(self, request: HttpRequest, organization, *args, **kwargs):
         # If already authenticated, don't redirect to login.
-        self.default_context = {}
-        return self.respond("sentry/toolbar/missing-perms.html", status=403)
+        return HttpResponse(status=403)
```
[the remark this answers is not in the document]
ryan953: ya, in the this case we DO want to show the login page if the user is not already logged into this org.

This is the url that will he shown inside the popup to users.
if they are not logged in, they must do that
after logging in they will endup at a page that renders the `login-success.html` template.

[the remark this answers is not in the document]
aliu39: permission_required isn't the same as checking for login. This only happens in the case we add a required_scope later (ex org:read, org:integrations), a user logs in, and doesn't have the required permissions. In this case logging in again won't help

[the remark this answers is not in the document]
ryan953: what is the default scope?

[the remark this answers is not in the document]
ryan953: all users in the org will need permissions so they can load the toolbar. Owner, Manager, Admin, Member will all be equally able to login.. the specific actions that they can take inside an org are different but already handled by the different api endpoints (for example: issue-assign, integration-delete, etc. etc)

src/sentry/toolbar/iframe_view.py
```diff
@@ -0,0 +1,40 @@
...
+
+@region_silo_view
+class IframeView(OrganizationView):
+    # TODO: For perms check. This is just an example and we might not need it.
+    # required_scope = "org:read,org:integrations"
+
+    security_headers = {
+        "X-Frame-Options": "ALLOWALL",
+        # TODO: # Not working, seems to get overridden by a BaseView wrapper
+        # "Content-Security-Policy": "frame-ancestors http://dev.getsentry.net/ http: https:",
```
ryan953, line 16: As long as one of these works and we can render the `iframe.html` template into an `<iframe>` node that's all we need in this PR.

I know `X-Frame-Options` was working before, but now we've got `OrganizationView` as a base class (which we want to keep!) so we just need to resolve this todo with one header or the other so it keeps working.

ryan953 (approved):

PR #26907, opened 2021-06-25
fix(activity): Add index to fix slow activity query

wedamija:
https://getsentry.atlassian.net/browse/OPS-580

We're seeing this query take up a lot of cumulative time, this index should resolve it. There's not
really a good index for this query to use, and the sort ends up being pretty slow, so we add
`datetime` here so that we can hopefully order in a reasonable amount of time.

Note that this table is 135mm rows, so this might take a while to add

With the query written as-is, Postgres is unable to use the index. If I materialize the project ids 
and pass them through instead, it successfully uses the index and the slow query is now much 
faster.

PR #77086, opened 2024-09-06
ref: squash index_together operation for Activity model

asottile-sentry:

hard-stop is already in place for self-hosted

### DIFF
--- a/src/sentry/migrations/0001_squashed_0484_break_org_member_user_fk.py
+++ b/src/sentry/migrations/0001_squashed_0484_break_org_member_user_fk.py
@@ -7369,28 +7369,11 @@ class Migration(CheckedMigration):
                 ),
             ],
         ),
-        migrations.SeparateDatabaseAndState(
-            database_operations=[
-                migrations.RunSQL(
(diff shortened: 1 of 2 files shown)


src/sentry/migrations/0001_squashed_0484_break_org_member_user_fk.py
```diff
@@ -7369,28 +7369,11 @@ class Migration(CheckedMigration):
...
-                    reverse_sql="\n                DROP INDEX CONCURRENTLY IF EXISTS sentry_activity_project_id_datetime_c00585e4_idx;\n                ",
-                    hints={"tables": ["sentry_activity"]},
-                ),
-            ],
-            state_operations=[
-                migrations.AlterIndexTogether(
-                    name="activity",
-                    index_together={("project", "type", "datetime")},
-                ),
-            ],
```
wedamija, line 7385: This is weird... Just a mistake in the old migration?

wedamija (approved):

src/sentry/migrations/0001_squashed_0484_break_org_member_user_fk.py
asottile-sentry: yeah seems a little weird -- looks like we didn't have the same PR tests back then to prevent it and a mistake was made in #26907 and then fixed in #27042

PR #77090, opened 2024-09-06
ref: squash index_together operation for Rule model

asottile-sentry:

hard-stop is already in place for self-hosted


asottile-sentry:
hmm this one seems not quite right -- need to look into this more closely (owner_team vs owner_user_id?)

asottile-sentry:
> hmm this one seems not quite right -- need to look into this more closely (owner_team vs owner_user_id?)

ah I understand now -- this index was removed / reworked later on so I need the pre-deletion name and fields here

markstory (approved):

IanWoodard (approved):

PR #77092, opened 2024-09-06
feat(flags): Track feature flag evaluations and submit on error

### LINKED ISSUE #77154: Attach flags in Sentry's BE to events

cmanallen:
Collects feature flag evaluations and adds them on the context of an error event.

There are four components I want to call out:

1. The `flag.py` module.  The flag manager is a circular buffer of flags.  This keeps our memory usage bounded.  The functions defined at the top of the module are what scope the flags to a particular thread or task.  See `test_flag.py` for examples on how this works.
2. The HTTP middleware initializes the context variable for that particular thread.
3. The features module uses the context variable and appends the flag evaluation to the thread-local manager.
4. The SDK has a before_send hook which looks at the thread-local context variable and copies the contents to the error event.

Flags are capped to 10 evaluations per thread while we're testing this feature.  We will raise gradually and observe any impact this might have on Sentry or the SDK.

Closes: https://github.com/getsentry/sentry/issues/77154


src/sentry/utils/flag.py
```diff
@@ -0,0 +1,80 @@
+# Thread safe flag tracking wrapper.
+from contextvars import ContextVar
+
+flag_manager = ContextVar("flag_manager")
+
+
+def initialize_flag_manager(capacity: int = 10) -> None:
```
bruno-garcia, line 7: capacity here means we'll keep up to 10 flags in memory at any time?

src/sentry/utils/flag.py
```diff
@@ -0,0 +1,80 @@
...
+    flag_manager.set(FlagManager(capacity=capacity))
+
+
+# NOTE: If not properly initialized this function is a no-op.
+def process_flag_result(flag: str, result: bool) -> None:
+    try:
+        _flag_manager = flag_manager.get()
+        _flag_manager.insert(flag, result)
+    except LookupError:
```
bruno-garcia, line 16: What can throw LookupError? `flag_manager.get()`?
would we want a debug level log to figure out if things were not properly initialized?

src/sentry/utils/flag.py
cmanallen: Yes.

src/sentry/utils/flag.py
cmanallen: flag_manager.get() is the only location that throws this error.  I'm not logging at the moment because I know there are gaps and I know this will be noisy.

src/sentry/utils/flag.py
bruno-garcia: Starting conservatively? How about 100 😬 ?

src/sentry/utils/flag.py
bruno-garcia: What are the known gaps? If that's known/obvious to everyone feel free to ignore this since I'm catching up

src/sentry/utils/flag.py
cmanallen: I know but I'd rather the experiment to be low risk.  Its 10 flags * the number of tasks.  I'm not sure how many that is at the moment.

src/sentry/utils/flag.py
cmanallen: Known only to me.  I'm initializing the variable for the HTTP server but not the tasks or consumers.

src/sentry/utils/flag.py
cmanallen: fwiw uwsgi is probably only running 16 threads so I'm being overly safe.

bruno-garcia: Worth documenting the thinking that led to 10. I believe we'll often under-report flags with this value so if the idea is to start at a lower value to reduce risk, then increase eventually that makes sense to me. If that's the case please clarify as a comment on the code or the PR (or even on this thread)

src/sentry/utils/flag.py
bruno-garcia: Is that worth writing in the PR description or as comment somewhere?

src/sentry/middleware/flag.py
```diff
@@ -0,0 +1,15 @@
+from collections.abc import Callable
+
+from django.http.request import HttpRequest
+from django.http.response import HttpResponseBase
+
+from sentry.utils.flag import initialize_flag_manager
+
+
+class FlagMiddleware:
```
ryan953, line 9: should this class extend something, or maybe it would be easier to make it a func:

```
def access_log_middleware(
    get_response: Callable[[Request], Response]
) -> Callable[[Request], Response]:
    def middleware(request: Request) -> Response:
        initialize_flag_manager()
        return get_response(request)
```

src/sentry/utils/sdk.py
```diff
@@ -236,6 +237,12 @@ def before_send(event: Event, _: Hint) -> Event | None:
             event["tags"]["silo_mode"] = str(settings.SILO_MODE)
         if settings.SENTRY_REGION:
             event["tags"]["sentry_region"] = settings.SENTRY_REGION
+
+    if "contexts" in event:
+        event["contexts"]["flags"] = get_flags_serialized()
+    else:
+        event["contexts"] = {"flags": get_flags_serialized()}
```
billyvg, line 244: logaf l: idk if this is more pythonic or not but I prefer to init `event["contexts"]` to a dict first and then be able to updates `flags` in the same fashion aftwards

c298lee (approved):

ryan953 (approved):

src/sentry/utils/flag.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def insert(self, flag: str, result: bool) -> None:
+        flag_ = Flag(flag, result)
+
+        if self.ip >= self.capacity:
+            self.buffer[self.index] = flag_
+        else:
+            self.buffer.append(flag_)
+
+        self.ip += 1
```
ryan953, line 73: note to self: `self.ip % self.capacity` means that the buffer will not have the earliest flag evaluation at the start. It's a ring buffer, not a FIFO pipe.

michellewzhang: was this the intended behavior? on the FE we are assuming the array we get from the `flags` array is sorted by eval order by default

ryan953: it was intended to be a circular buffer per the PR message, i don't think anyone really thought it through until the latest mocks and seeing it in the UI recently.

michellewzhang: yeah makes sense, maybe we need to revisit the data structure🤔

tests/sentry/utils/test_flag.py
```diff
@@ -0,0 +1,129 @@
...
+
+    flag.process_flag_result("e", False)
+    flag.process_flag_result("f", False)
+    flags = flag.get_flags_serialized()
+    assert len(flags) == 3
+    assert flags == [
+        {"flag": "d", "result": False},
+        {"flag": "e", "result": False},
+        {"flag": "f", "result": False},
+    ]
```
michellewzhang, line 50: actually @ryan953 check out this test case: it looks like the behavior is as intended

src/sentry/utils/flag.py
cmanallen: It's sort by evaluation method.  Look at the serialize method.  It accounts for the circular buffer.

ryan953: ahh, sliced it up. cool!

PR #77093, opened 2024-09-06
fix(alerts): Prevent muting user alerts

schew2381:


src/sentry/api/endpoints/user_notification_settings_options_detail.py
```diff
@@ -19,13 +21,29 @@ class UserNotificationSettingsOptionsDetailEndpoint(UserEndpoint):
...
+    def convert_args(
+        self,
+        request: Request,
+        notification_option_id: int,
+        user_id: int | str | None = None,
+        *args,
+        **kwargs,
+    ):
+        args, kwargs = super().convert_args(request, user_id, *args, **kwargs)
+        fetched_user = kwargs["user"]
```
mdtro, line 34: Rather than deriving the user from the URL, can we derive it from `request.user` ? That way we're restricted to just the logged in user and limits the chance of cross user queries.

schew2381: ok I tried doing that but it doesn't work due to some hybrid cloud stuff, where request.user is a lazy rpc thingy

geoffg-sentry (approved):
In absense of our ability to derive the user from the request, this looks solid

PR #77094, opened 2024-09-06
ref(integrations): Remove indirect imports from messaging integrations

RyanSkonnord:
Delete star-imports with "noqa: F401,F403" and update related import statements. Move declarations from `__init__.py` to other files.


src/sentry/integrations/slack/service.py
```diff
@@ -18,7 +18,9 @@
     IssueAlertNotificationMessageRepository,
 )
 from sentry.integrations.slack.message_builder.base.block import BlockSlackMessageBuilder
-from sentry.integrations.slack.message_builder.notifications import get_message_builder
```
iamrajjoshi, line 21: we should audit if we even need this service

iamrajjoshi (approved):
🙌🏽

cathteng (approved):

PR #77095, opened 2024-09-06
(seer client): Add warning if secret is not configured

ram-senth:
Adds check and logging for the secret key used to sign Seer requests

### DIFF
--- a/src/sentry/seer/signed_seer_api.py
+++ b/src/sentry/seer/signed_seer_api.py
@@ -1,5 +1,6 @@
 import hashlib
 import hmac
+import logging
 from random import random
 from typing import Any
 from urllib.parse import urlparse
@@ -9,6 +10,8 @@ from urllib3 import BaseHTTPResponse, HTTPConnectionPool
 
 from sentry import options
 from sentry.utils import metrics
+
+logger = logging.getLogger(__name__)
 
 
 def make_signed_seer_api_request(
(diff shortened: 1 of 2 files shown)


src/sentry/seer/signed_seer_api.py
```diff
@@ -40,10 +43,15 @@ def make_signed_seer_api_request(
...
-        auth_headers["Authorization"] = f"Rpcsignature rpc0:{signature}"
+    if random() < options.get("seer.api.use-shared-secret"):
+        if settings.SEER_API_SHARED_SECRET:
+            signature_input = b"%s:%s" % (url.encode("utf8"), body)
+            signature = hmac.new(
+                settings.SEER_API_SHARED_SECRET.encode("utf-8"), signature_input, hashlib.sha256
+            ).hexdigest()
+            auth_headers["Authorization"] = f"Rpcsignature rpc0:{signature}"
+        else:
+            logger.warning(
```
trillville, line 54: should we just throw an exception?

trillville: i guess warning is fine

ram-senth: Yes I would prefer keeping it as a warning as the call may or may not fail based on the Seer side configuration.

trillville (approved):

PR #77098, opened 2024-09-06
feat(dashboards): Add y-axis pointer

gggritso:
Borrows a feature from Insights! On hover, shows a y-axis pointer line, that shows where the cursor intersects with the y-axis. This is really useful for things like:

- comparing multiple peaks to see if they're the same
- eyeballing whether a line is going up or down
- etc.

**e.g.,**

https://github.com/user-attachments/assets/6e6541ca-9381-44af-acbb-5d95b124898f

### DIFF
--- a/static/app/views/dashboards/widgetCard/chart.tsx
+++ b/static/app/views/dashboards/widgetCard/chart.tsx
@@ -424,6 +424,9 @@ class WidgetCardChart extends Component<WidgetCardChartProps, State> {
       },
       tooltip: {
         trigger: 'axis',
+        axisPointer: {
+          type: 'cross',
+        },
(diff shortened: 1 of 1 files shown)


edwardgou-sentry (approved):
Question, seems like this linestyle is solid, whereas the insights one is a dotted line. Should we make them match?

gggritso:
@edwardgou-sentry good question, I was torn on that! I thought it looked weird to have the horizontal and vertical line be different, even though that's what Insights does

edwardgou-sentry:
@gggritso I like the solid lines over the dotted lines anyways, so no objections from me! Maybe we can update the insights crosshairs later

PR #77099, opened 2024-09-06
feat(explore-toolbar): Adding search to groubBy and delete button to rows

Abdkhan14:
Screenshot 2024-09-06 at 3 18 58 PM

https://github.com/user-attachments/assets/fcbc0fb9-22d1-4f61-91db-f3a18af4d5b9

### DIFF
--- a/static/app/views/explore/toolbar/styles.tsx
+++ b/static/app/views/explore/toolbar/styles.tsx
@@ -54,18 +54,17 @@ export function ToolbarRow<T>({
   return (
     <ToolbarRowWrapper>
       {children}
-      {rows.length > 1 ? (
-        <Button
-          borderless
-          icon={<IconDelete />}
-          size="zero"
-          onClick={e => {
-            e.preventDefault();
-            removeRow(index);
-          }}
-          aria-label={t('Remove')}
-        />
-      ) : null}
+      <Button
+        borderless
+        icon={<IconDelete />}
+        size="zero"
+        disabled={rows.length <= 1}
+        onClick={e => {
+          e.preventDefault();
+          removeRow(index);
+        }}
+        aria-label={t('Remove')}
+      />
     </ToolbarRowWrapper>
   );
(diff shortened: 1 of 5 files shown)


static/app/views/explore/toolbar/styles.tsx
```diff
@@ -28,3 +31,47 @@ export const ToolbarHeading = styled('h6')<{disabled?: boolean}>`
...
+      const newRow = rows.filter((_, rowIndex) => rowIndex !== i);
+      setRows?.(newRow);
+    },
+    [setRows, rows]
+  );
+
+  return (
+    <ToolbarRowWrapper>
+      {children}
+      {rows.length > 1 ? (
```
Zylphrex, line 57: Instead of hiding, we should disable the button with `disable={rows.length <= 1}`

static/app/views/explore/toolbar/toolbarVisualize.tsx
```diff
@@ -79,7 +79,7 @@ export function ToolbarVisualize({}: ToolbarVisualizeProps) {
         </ToolbarHeaderButton>
       </ToolbarHeader>
       {parsedVisualizes.map((parsedVisualize, index) => (
-        <VisualizeOption key={index}>
+        <ToolbarRow rows={visualizes} setRows={setVisualizes} index={index} key={index}>
```
Zylphrex, line 82: Can you undo the changes to this file here? We're discussing some changes that would be incompatible with this helper component.

Abdkhan14: Okay so will just keep the delete to groupBy for now

static/app/views/explore/toolbar/styles.tsx
Abdkhan14: Sure yup

Zylphrex (approved):

PR #77103, opened 2024-09-06
ref(similarity): Allow backfill to continue after Gateway Timeout

jangjodi:
If Seer post request fails due to a Gateway Timeout, it is because the stacktrace strings we are sending to encode are too long, and Seer times out.
Currently, we stop the entire backfill.
We want to change this behavior, since we always just check the logs and restart the backfill anyways.
Now, only when the error is a Gateway Timeout, skip the project and call the next backfill


src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -381,20 +381,13 @@ def get_events_from_nodestore(
...
 ) -> BulkCreateGroupingRecordsResponse | None:
-    try:
-        seer_response = _retry_operation(
-            post_bulk_grouping_records,
-            create_grouping_records_request,
-            retries=3,
-            delay=2,
-            exceptions=Exception,
-        )
-    except Exception as e:
```
jangjodi, line 392: removing this, as `post_bulk_grouping_records` function already handles exceptions and returns the failure reason back

JoshFerge (approved):

PR #77105, opened 2024-09-06
chore(anomaly detection): no anomaly detection for apdex alerts

mifu67:
Hide the anomaly detection option if the alert type is apdex/llm.

https://github.com/user-attachments/assets/d869c902-8e74-41e4-b8bc-26cd90424706

### DIFF
--- a/static/app/views/alerts/rules/metric/thresholdTypeForm.tsx
+++ b/static/app/views/alerts/rules/metric/thresholdTypeForm.tsx
@@ -37,6 +37,19 @@ function ThresholdTypeForm({
   if (isCrashFreeAlert(dataset)) {
     return null;
   }
+  const validAnomalyDetectionAlertTypes = new Set([
+    'num_errors',
+    'users_experiencing_errors',
+    'throughput',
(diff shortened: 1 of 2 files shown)


nhsiehgit (approved):

static/app/views/alerts/rules/metric/thresholdTypeForm.tsx
```diff
@@ -78,7 +81,10 @@ function ThresholdTypeForm({
     ],
   ];
 
-  if (hasAnomalyDetection) {
+  if (
+    hasAnomalyDetection &&
+    !(alertType === 'apdex' || alertType === 'llm_tokens' || alertType === 'llm_cost')
```
ceorourke, line 86: Can we do the inverse? So only show if it the alert types are the ones we know work (everything except these)?

mifu67: Sure, I was just concerned that it would be messy. Do you think it's okay?

ceorourke (approved):

PR #77107, opened 2024-09-06
added warning label for dynamic alerts that don't have enough data

nhsiehgit:
Adds the warning label if an alert rule has a status of not enough data
Screenshot 2024-09-06 at 1 42 11 PM


mifu67 (approved):
LGTM. Can we ask Rachel about the banner copy?

PR #77109, opened 2024-09-06
test(alerts): Debug anomaly detection

ceorourke:
This is not actually code that will be merged, but since we can't test metric alerts locally fully I pulled out the code we are having trouble with in production to be able to test it locally and make sure everything is working.

### DIFF
--- a/src/sentry/seer/anomaly_detection/get_anomaly_data.py
+++ b/src/sentry/seer/anomaly_detection/get_anomaly_data.py
@@ -1,0 +1,137 @@
+import datetime
+import logging
+
+from django.conf import settings
+from urllib3.exceptions import MaxRetryError, TimeoutError
+
+from sentry.conf.server import SEER_ANOMALY_DETECTION_ENDPOINT_URL
+from sentry.incidents.logic import WARNING_TRIGGER_LABEL
+from sentry.incidents.models.alert_rule import AlertRule
+from sentry.net.http import connection_from_url
+from sentry.seer.anomaly_detection.types import (
+    AlertInSeer,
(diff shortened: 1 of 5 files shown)


src/sentry/seer/anomaly_detection/get_anomaly_data.py
```diff
@@ -0,0 +1,137 @@
...
+    alert_rule: AlertRule, subscription: QuerySubscription, aggregation_value: float | None
+) -> None:
+    anomaly_detection_config = AnomalyDetectionConfig(
+        time_period=int(alert_rule.snuba_query.time_window / 60),
+        sensitivity=alert_rule.sensitivity,
+        direction=translate_direction(alert_rule.threshold_type),
+        expected_seasonality=alert_rule.seasonality,
+    )
+    ############################
+    now = datetime.datetime.now()
```
ceorourke, line 69: Since I don't have access to `self.last_update` from the subscription processor, I am using the current timestamp

src/sentry/seer/anomaly_detection/utils.py
```diff
@@ -103,7 +104,10 @@ def format_historical_data(data: SnubaTSResult, dataset: Any) -> list[TimeSeries
             formatted_data.append(ts_point)
     else:
         for datum in nested_data:
-            ts_point = TimeSeriesPoint(timestamp=datum.get("time"), value=datum.get("count", 0))
+            #########################
+            count = randrange(1, 100)
```
ceorourke, line 108: There is no historical data locally so I populate this with random numbers. Seer wasn't happy if I did it all the same number

PR #77115, opened 2024-09-06
fix(issue-platform): remove +1 on limit

JoshFerge:
Fixes SENTRY-3D61

its unclear to me why we'd add a +1 onto our LIMIT. we should be able to handle limit=10000 like our other query strategies, but adding plus 1 makes it 10001 which causes it to error.

let's see if any tests break because of this

### DIFF
--- a/src/sentry/search/snuba/executors.py
+++ b/src/sentry/search/snuba/executors.py
@@ -1867,7 +1867,7 @@ class GroupAttributesPostgresSnubaQueryExecutor(PostgresSnubaQueryExecutor):
                 groupby=groupby,
                 having=having,
                 orderby=[OrderBy(sort_func, direction=Direction.DESC)],
-                limit=Limit(limit + 1),
+                limit=Limit(limit),
             )
             dataset = Dataset.Events.value if is_errors else Dataset.IssuePlatform.value
             request = Request(


snigdhas (approved):

wedamija (commented):
This is usually for pagination, so that we can see if there's another page to fetch. Are you sure removing this won't break anything?

JoshFerge:
> This is usually for pagination, so that we can see if there's another page to fetch. Are you sure removing this won't break anything?

seems like it did break a few tests. was just trying to see. that makes sense. its hard as it seems like the limit in our API is 10,000, but the snuba limit is also 10,000, so the +1 here is causing lots of errors from clients. guess we need to make the limit 9,999 instead?

wedamija:
> > This is usually for pagination, so that we can see if there's another page to fetch. Are you sure removing this won't break anything?
> 
> seems like it did break a few tests. was just trying to see. that makes sense. its hard as it seems like the limit in our API is 10,000, but the snuba limit is also 10,000, so the +1 here is causing lots of errors from clients. guess we need to make the limit 9,999 instead?

Yeah, that seems fine to me

JoshFerge:
https://sentry.sentry.io/issues/5683689835/ is still spiking. worth trying to merge a fix for this

JoshFerge:
closing as I don't think this is the correct way to fix this problem. we probably want to increase the snuba API limit to 10,001 or return a ValidationError

PR #77118, opened 2024-09-06
fix(flagpole): Default `rollout` to 100

lobsterkatie:
This fixes the default value of the `rollout` property of Flagpole segments to be 100, which matches what's described in the [docs](https://develop.sentry.dev/application/feature-flags/flagpole/#segments).

### DIFF
--- a/src/flagpole/conditions.py
+++ b/src/flagpole/conditions.py
@@ -216,7 +216,7 @@ class Segment:
     conditions: list[ConditionBase] = dataclasses.field(default_factory=list)
     "The list of conditions that the segment must be matched in order for this segment to be active"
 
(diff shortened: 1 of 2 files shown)


GabeVillalobos (approved):

markstory (approved):
I looked through existing feature flags and there weren't any other feature flags with undefined `rollout` attributes so we shouldn't enable any features that were previously disabled via rollout conditions with this change.

lobsterkatie:
> I looked through existing feature flags and there weren't any other feature flags with undefined `rollout` attributes so we shouldn't enable any features that were previously disabled via rollout conditions with this change.

Good thinking. Thanks, @markstory!

PR #77127, opened 2024-09-08
feat(custom-views): Implement "Add View" flow to custom views

MichaelSun48:
This PR implements an add view flow for custom views. Previously, as a stop gap solution, clicking "Add View" would simply duplicate the query and sort of the currently selected tab to the new tab. Now, clicking "Add View" creates a tab with a blank query, and a new page containing recommended searches, and any saved searches the user created before custom-views. 

While this feature's design mostly sticks to the mockups, several important changes have been made. These are:

1. **There is no option to delete Saved Searches.** This was done because it feels unnecessary to allow the user to permanently delete their saved searches when the feature itself will be deprecated soon anyways. 
2. **The banner at the top of the page is no longer dismissible.** This was done because it is the only place that explains that views are user-scoped. This will likely change post-LA.
3. **The button within the banner is now a feedback button, rather than a "Read Docs" button.** This was done because, well, there are no docs for custom views yet since it's not even EA'd yet. 
4. ~**Added more left-padding to the search suggestion title.** In the mockups, the row container bled outside of the parent container's left border. I added a little bit more left padding as a replacement for this effect.~ Reverted after chatting with Vu

Other considerations that came up as I was creating this: 

* Saved searches currently do not show saved searches created by the user that are organization scoped (visibility=“organization”). This is because the endpoint does not appear to send the owner_id of a saved search back to the frontend currently. Do we want to show those as well? 
* I copied over the "Recommended Searches" from the designs, but perhaps there are other searches we could recommend as a discovery function? (possibly consult Josh Ferge for this) 
* Saved searches also store a sort. Do we want to display that anywhere?

image

Todo:
- [x] Add sparkles ✨
- [x] Move interactionstatelayer inset -1


static/app/components/infoTooltip.tsx
MichaelSun48, line 1: This component is largely the same as the `<QuestionTooltip/>` component, except with a different icon. Ideally there would be a `<IconTooltip/>` component, but I couldn't find one and creating it is out of the scope of this feature.

static/app/views/issueList/customViewsHeader.tsx
```diff
@@ -257,20 +259,49 @@ function CustomViewsIssueListHeaderTabsContent({
...
+            {
+              ...location,
+              query: {
+                ...queryParams,
+                viewId: tab.id,
+              },
             },
-          });
+            {replace: true}
+          );
```
MichaelSun48, line 271: Unrelated change: this makes it so when we replace a temporary viewId with the persistent Id, nothing is added to the navigation history, which is the expected behavior.

static/app/views/issueList/customViewsHeader.tsx
```diff
@@ -257,20 +259,49 @@ function CustomViewsIssueListHeaderTabsContent({
...
+        setDraggableTabs(updatedTabs);
+        debounceUpdateViews(updatedTabs);
+      } else {
+        setNewViewActive(true);
+      }
+    } else {
+      setNewViewActive(false);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [viewId, query]);
```
MichaelSun48, line 303: This ensures that the add view flow only appears on new views that have not been saved to the backend yet. It also ensures that if a user decides to skip the add view flow and write their own query, the query will be saved to the tab afterwards.

static/app/components/infoTooltip.tsx
```diff
@@ -0,0 +1,66 @@
...
+import type {SVGIconProps} from 'sentry/icons/svgIcon';
+import type {IconSize} from 'sentry/utils/theme';
+
+interface InfoTooltipProps
+  extends Partial<
+    Pick<
+      TooltipProps,
+      'containerDisplayMode' | 'isHoverable' | 'overlayStyle' | 'position'
+    >
+  > {
```
scttcper, line 15: instead of extending tooltip props, i'd prefer to see a prop specific to tooltipProps instead of several being available at the root.  

similar to this one https://github.com/getsentry/sentry/blob/master/static/app/components/button.tsx#L81-L84

static/app/components/infoTooltip.tsx
scttcper, line 1: nevermind i see what is happening here, can we just extend question tooltip to show an info icon

static/app/views/issueList/addViewPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+          {/*
+            Saved searches have an average length of approximately 16 characters
+            This container fits 16 'a's comfortably, and 20 'a's before overflowing.
+          */}
+          <StyledOverflowEllipsisTextContainer>
+            {suggestion.label}
+          </StyledOverflowEllipsisTextContainer>
+          <QueryWrapper>
+            <FormattedQuery query={suggestion.query} />
+            {
```
scttcper, line 108: unused brackets

static/app/views/issueList/addViewPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+  &:hover ${ActionsWrapper} {
+    visibility: visible;
+  }
+`;
+
+const SuggestionList = styled('ul')`
+  display: flex;
+  flex-direction: column;
+  padding: 0;
+`;
```
scttcper, line 230: once you've taken the title out of the list, it should make more sense to bring the grid up to the parent instead of applying an inline grid to each element and a margin left on StyledOverflowEllipsisTextContainer

scttcper:
the stars for me are covering the text at some widths
image

scttcper:
not sure whats going on at this width but it's not taking the full width of the page
image

static/app/views/issueList/addViewPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+const StyledPanel = styled(Panel)`
```
scttcper, line 232: what styles are you using from Panel? If it's only one i'd just style a div

static/app/views/issueList/addViewPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+  display: inline-flex;
```
scttcper, line 233:
```suggestion
  display: flex;
```

static/app/views/issueList/addViewPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+`;
+const Title = styled('div')`
+  font-size: ${p => p.theme.fontSizeMedium};
+  font-weight: ${p => p.theme.fontWeightBold};
+`;
+
+const SubTitle = styled('div')`
+  font-weight: ${p => p.theme.fontWeightNormal};
+  font-size: ${p => p.theme.fontSizeMedium};
+  margin: 0;
```
scttcper, line 254: shouldn't need no margin here
```suggestion
```

static/app/views/issueList/addViewPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+`;
+
+const AddViewWrapper = styled('div')`
+  display: flex;
+  flex-direction: column;
+  border: 0;
```
scttcper, line 260: i don't think border 0 is doing anything
```suggestion
```

MichaelSun48:
> not sure whats going on at this width but it's not taking the full width of the page

fixed in most recent commit!

MichaelSun48:
> the stars for me are covering the text at some widths

This is a kinda funny consequence of the feedback button being absent from the Vercel previews. When the feedback button is present, it won't cover the text: 

image

static/app/components/infoTooltip.tsx
malwilley: Yeah @MichaelSun48 I agree we should just use `QuestionTooltip` here. I'm not sure if the `i` vs `?` is a semantic difference or if we should just pick one everywhere so let's make our job easy and just stick with `?` for now since that's what we have been doing already

malwilley:
> This is a kinda funny consequence of the feedback button being absent from the Vercel previews. When the feedback button is present, it won't cover the text:

For self-hosted we still want a good experience, so something else you could do is hide the stars at small widths

malwilley (approved):

PR #77147, opened 2024-09-09
feat(whats-new): add method to RpcOrganization to retrieve aggregate flags of projects

shellmayr:
Adds a new method to RpcProject to enable access to the project flags via an RPC call from the control silo.

Contributes to https://github.com/getsentry/sentry/issues/77058

### DIFF
--- a/src/sentry/organizations/services/organization/impl.py
+++ b/src/sentry/organizations/services/organization/impl.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from collections.abc import Mapping
 from typing import Any
 
+from django.contrib.postgres.aggregates import BitOr
 from django.db import models, router, transaction
 from django.db.models.expressions import CombinedExpression, F
 from django.dispatch import Signal
@@ -30,6 +31,7 @@ from sentry.models.organizationaccessrequest import OrganizationAccessRequest
 from sentry.models.organizationmapping import OrganizationMapping
 from sentry.models.organizationmember import InviteStatus, OrganizationMember
 from sentry.models.organizationmemberteam import OrganizationMemberTeam
+from sentry.models.project import Project
 from sentry.models.projectbookmark import ProjectBookmark
 from sentry.models.recentsearch import RecentSearch
 from sentry.models.rule import Rule, RuleActivity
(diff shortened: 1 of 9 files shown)


src/sentry/projects/services/project/model.py
```diff
@@ -54,6 +84,11 @@ def delete_option(self, key: str) -> None:
 
         project_service.delete_option(self, key)
 
+    def get_flags(self, organization_id: int, project_id: int) -> RpcProjectFlags:
+        from sentry.projects.services.project import project_service
+
+        return project_service.get_flags(organization_id=organization_id, project_id=project_id)
```
markstory, line 90: Should flags be part of `RpcProject` all the time? There are relatively few read operations to projects over RPC so adding to the response payload should have minimal impact. If you wanted to be even more cautious, you could add parameters to the existing get methods to include flags in the results.

shellmayr: As you suggested in getsentry, it may actually be more useful to move this to RpcOrganization instead and combine flags in that call - and that obviates the need for a permanent representation, especially if we allow filtering by a subset of the project ids in the method.

src/sentry/organizations/services/organization/impl.py
```diff
@@ -349,6 +352,31 @@ def update_flags(self, *, organization_id: int, flags: RpcOrganizationFlagsUpdat
...
+        2. Pad the string with the number of leading zeros so the length of the binary representation lines up with the
+           number of bits of MAX_BIGINT / the BitField
+        3. Reverse the binary representation to correctly assign flags based on the order
+        4. Serialize as an RpcProjectFlags object
+        """
+        flag_keys = cast(list[str], Project.flags)
+
+        org = Organization.objects.filter(id=organization_id).get()
+        projects = org.project_set
+        if org.project_set.count() > 0:
```
markstory, line 368: You could skip the read on `Organization` and use `Project` to create a queryset `Project.objects.filter(organization_id=organization_id)` would give you the organization's project list without needing to load the org record.

tests/sentry/hybridcloud/test_organization.py
```diff
@@ -360,3 +360,20 @@ def test_send_sso_unlink_emails() -> None:
...
+def test_get_aggregate_project_flags() -> None:
+    org = Factories.create_organization()
+    project1 = Factories.create_project(organization_id=org.id, name="test-project-1")
+    Factories.create_project(organization_id=org.id, name="test-project-2")
+    flags = organization_service.get_aggregate_project_flags(organization_id=org.id)
+    assert flags.has_insights_http is False
+
+    with assume_test_silo_mode_of(Project):
+        project1.flags.has_insights_http = True
+        project1.update(flags=F("flags").bitor(Project.flags.has_insights_http))
```
markstory, line 376: It would be good set a different flag on the second project you create, then you could assert that the results contains flags from both projects.

markstory (approved):

PR #77149, opened 2024-09-09
drop(grouping): Delete more hierarchical related code

armenzg:
Follow-up to deleting hierarchical code (#76994).

### DIFF
--- a/src/sentry/api/endpoints/event_ai_suggested_fix.py
+++ b/src/sentry/api/endpoints/event_ai_suggested_fix.py
@@ -354,7 +354,7 @@ class EventAiSuggestedFixEndpoint(ProjectEndpoint):
 
         # Cache the suggestion for a certain amount by primary hash, so even when new events
         # come into the group, we are sharing the same response.
-        cache_key = "ai:" + event.get_primary_hash()  # type: ignore[operator]  # XXX: get_primary_hash probably typed wrong
+        cache_key = "ai:" + event.get_primary_hash()
         suggestion = cache.get(cache_key)
         if suggestion is None:
             try:
--- a/src/sentry/api/helpers/group_index/delete.py
+++ b/src/sentry/api/helpers/group_index/delete.py
@@ -11,6 +11,7 @@ from sentry import audit_log, eventstream
 from sentry.api.base import audit_logger
 from sentry.issues.grouptype import GroupCategory
 from sentry.models.group import Group, GroupStatus
+from sentry.models.grouphash import GroupHash
 from sentry.models.groupinbox import GroupInbox
 from sentry.models.project import Project
 from sentry.signals import issue_deleted
@@ -47,6 +48,8 @@ def delete_group_list(
 
     # Tell seer to delete grouping records for these groups
     call_delete_seer_grouping_records_by_hash(group_ids)
+
+    GroupHash.objects.filter(project_id=project.id, group__id__in=group_ids).delete()
 
     # We remove `GroupInbox` rows here so that they don't end up influencing queries for
     # `Group` instances that are pending deletion
--- a/src/sentry/api/helpers/group_index/delete.py
+++ b/src/sentry/api/helpers/group_index/delete.py
@@ -11,7 +11,6 @@ from sentry import audit_log, eventstream
 from sentry.api.base import audit_logger
 from sentry.issues.grouptype import GroupCategory
 from sentry.models.group import Group, GroupStatus
-from sentry.models.grouphash import GroupHash
 from sentry.models.groupinbox import GroupInbox
 from sentry.models.project import Project
 from sentry.signals import issue_deleted
@@ -48,14 +47,6 @@ def delete_group_list(
 
     # Tell seer to delete grouping records for these groups
     call_delete_seer_grouping_records_by_hash(group_ids)
-
-    # We do not want to delete split hashes as they are necessary for keeping groups... split.
-    GroupHash.objects.filter(
-        project_id=project.id, group__id__in=group_ids, state=GroupHash.State.SPLIT
-    ).update(group=None)
-    GroupHash.objects.filter(project_id=project.id, group__id__in=group_ids).exclude(
-        state=GroupHash.State.SPLIT
-    ).delete()
 
     # We remove `GroupInbox` rows here so that they don't end up influencing queries for
     # `Group` instances that are pending deletion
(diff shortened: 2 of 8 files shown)


src/sentry/api/helpers/group_index/delete.py
```diff
@@ -49,14 +48,6 @@ def delete_group_list(
     # Tell seer to delete grouping records for these groups
     call_delete_seer_grouping_records_by_hash(group_ids)
 
-    # We do not want to delete split hashes as they are necessary for keeping groups... split.
-    GroupHash.objects.filter(
-        project_id=project.id, group__id__in=group_ids, state=GroupHash.State.SPLIT
-    ).update(group=None)
-    GroupHash.objects.filter(project_id=project.id, group__id__in=group_ids).exclude(
-        state=GroupHash.State.SPLIT
-    ).delete()
```
lobsterkatie, line 58: I was about to say that we still need this overall delete, just not the `exclude`, but actually deleting the group [deletes the `GroupHash` records](https://github.com/getsentry/sentry/blob/afe8517a0b2438558f08d926656b408f0dc72c51/tests/sentry/tasks/deletion/test_groups.py#L63), so I think getting rid of this entire statement is fine.

src/sentry/eventstore/models.py
```diff
@@ -356,17 +356,13 @@ def get_hashes(self, force_config: StrategyConfiguration | None = None) -> Calcu
...
+        _variants = self.get_grouping_variants(force_config)
+        variants = sort_grouping_variants(_variants)
+        hashes = self._hashes_from_sorted_grouping_variants(variants)
+        hashes_values = [hash_ for _, hash_ in hashes]
 
         return CalculatedHashes(
-            hashes=flat_hashes_values,
+            hashes=hashes_values,
             variants=variants,
         )
```
lobsterkatie, line 367: Perhaps better suited to a follow-up PR to this one, but `_hashes_from_sorted_grouping_variants` is only ever used here AFAICT, and we ignore the first item in each of the tuples it returns - meaning we could simplify it to just return the hash values, and then this

```python
        hashes = self._hashes_from_sorted_grouping_variants(variants)
        hashes_values = [hash_ for _, hash_ in hashes]

        return CalculatedHashes(
            hashes=hashes_values,
            variants=variants,
        )
```

could be replaced with just

```python
        return CalculatedHashes(
            hashes=self._hashes_from_sorted_grouping_variants(variants),
            variants=variants,
        )
```

or (if I'm reading it right) even just

```python
        return CalculatedHashes(
            hashes={ variant.get_hash() for variant in variants.values() } - { None },
            variants=variants,
        )
```

(eliminating `_hashes_from_sorted_grouping_variants` entirely).

src/sentry/grouping/api.py
```diff
@@ -373,16 +373,16 @@ def get_grouping_variants_for_event(
...
+    for name, variant in variants_dict.items():
+        variants.append((name, variant))
 
     # Sort system variant to the back of the list to resolve ambiguities when
     # choosing primary_hash for Snuba
-    flat_variants.sort(key=lambda name_and_variant: 1 if name_and_variant[0] == "system" else 0)
+    variants.sort(key=lambda name_and_variant: 1 if name_and_variant[0] == "system" else 0)
 
-    return flat_variants
+    return variants
```
lobsterkatie, line 388: Maybe I'm missing a nuance here, but couldn't this entire function be turned into just 
`return variants_dict.items().sort(.......)`? 

(Maybe you need a `list` in there to materialize it, IDK, but now that this function isn't sorting the variants into flat variants and hierarchical variants, the `for` loop is just recreating `.items()` AFAICT.)

src/sentry/grouping/enhancer/actions.py
```diff
@@ -142,7 +141,6 @@ class VarAction(Action):
     _VALUE_PARSERS: dict[str, Callable[[Any], Any]] = {
         "max-frames": int,
         "min-frames": int,
-        "invert-stacktrace": get_rule_bool,
```
lobsterkatie, line 145: Are we sure there are no customer rules using this?

src/sentry/models/grouphash.py
```diff
@@ -26,14 +26,10 @@ class State:
...
         LOCKED_IN_MIGRATION = 1
 
-        # This hierarchical grouphash should be ignored/skipped for finding the group.
-        SPLIT = 2
-
     project = FlexibleForeignKey("sentry.Project", null=True)
     hash = models.CharField(max_length=32)
     group = FlexibleForeignKey("sentry.Group", null=True)
 
-    # not-null => the event should be discarded
```
lobsterkatie, line 36: Why get rid of this comment? I think it's useful (and besides, it has nothing to do with hierarchical grouping).

src/sentry/api/endpoints/event_ai_suggested_fix.py
```diff
@@ -354,7 +354,7 @@ def get(self, request: Request, project, event_id) -> HttpResponse | StreamingHt
 
         # Cache the suggestion for a certain amount by primary hash, so even when new events
         # come into the group, we are sharing the same response.
-        cache_key = "ai:" + event.get_primary_hash()  # type: ignore[operator]  # XXX: get_primary_hash probably typed wrong
+        cache_key = "ai:" + event.get_primary_hash()
```
armenzg, line 357: I fixed the typing in the function, thus, it is not needed anymore.

src/sentry/eventstore/models.py
```diff
@@ -452,24 +448,8 @@ def get_grouping_variants(
...
-            extra={
-                "event_id": self.event_id,
-                "event_type": type(self),
-                "group_id": getattr(self, "group_id", None),
-                "project_id": self.project_id,
-            },
-        )
-        return None
+    def get_primary_hash(self) -> str:
+        return self.get_hashes().hashes[0]
```
armenzg, line 450: The original code before hierarchical grouping was basically this.

src/sentry/grouping/enhancer/actions.py
armenzg: I removed the rules from the enhancement-config file.
If any customer uses it we can remove it from the stack trace rules once it gets deployed (I just did it from projects using `sentinel` or `prefix` parser actions).

src/sentry/grouping/enhancer/parser.py
```diff
@@ -29,7 +29,7 @@
 actions          = action+
 action           = flag_action / var_action
 var_action       = _ var_name _ "=" _ ident
-var_name         = "max-frames" / "min-frames" / "invert-stacktrace" / "category"
```
armenzg, line 32: If any stack trace rules have it we can remove it after deployment.

The only way to invert the stack trace was via the hierarchical code which has already been removed:
https://github.com/getsentry/sentry/pull/77235/files#diff-765101e1faf45dddc42ee634ea872031228b3072cd6a28d41c533c3aba7414d0L491

PR #77168, opened 2024-09-09
feat(explore-search): Supporting span aggregates.

Abdkhan14:
https://github.com/user-attachments/assets/f9daf241-d498-4d4d-95c0-de7ac11f3bc2


static/app/components/performance/spanSearchQueryBuilder.tsx
```diff
@@ -104,6 +157,7 @@ export function SpanSearchQueryBuilder({
       placeholder={placeholderText}
       filterKeys={filterTags}
       initialQuery={initialQuery}
+      fieldDefinitionGetter={(key: string) => getFieldDefinition(key, 'span')}
```
malwilley, line 160: can you extract this out to its own function so the reference stays the same on rerenders?

static/app/components/performance/spanSearchQueryBuilder.tsx
```diff
@@ -54,8 +107,8 @@ export function SpanSearchQueryBuilder({
   });
 
   const filterTags: TagCollection = useMemo(() => {
-    return {...supportedTags};
-  }, [supportedTags]);
+    return Object.assign({}, functionTags, supportedTags);
```
malwilley, line 110: nit: it's the same thing but generally we just do `{...functionTags, ...supportedTags}`

static/app/components/performance/spanSearchQueryBuilder.tsx
Abdkhan14: Done

malwilley (approved):

static/app/components/performance/spanSearchQueryBuilder.tsx
```diff
@@ -40,6 +84,19 @@ export function SpanSearchQueryBuilder({
   const api = useApi();
   const organization = useOrganization();
   const {selection} = usePageFilters();
+  const [groupBys] = useGroupBys();
+  const [visualizes] = useVisualizes();
+  const [resultsMode] = useResultMode();
```
Zylphrex, line 89: You cannot use these hooks in this component directly as this is a generic component shared by other pages while these hooks were made specifically for the explore page.

Zylphrex (changes requested):

static/app/utils/fields/index.ts
```diff
@@ -789,6 +788,80 @@ export const AGGREGATION_FIELDS: Record<AggregationKey, FieldDefinition> = {
...
+  SpanIndexedField.SPAN_DURATION,
+  SpanIndexedField.SPAN_SELF_TIME,
+];
+
+export const ALLOWED_EXPLORE_VISUALIZE_AGGREGATES: AggregationKey[] = [
+  AggregationKey.COUNT,
+  AggregationKey.MIN,
+  AggregationKey.MAX,
+  AggregationKey.AVG,
+];
```
Zylphrex, line 802: This is fine for now, but we may run into complications as we add support for numeric span attributes which will be fetched dynamically.

static/app/components/performance/spanSearchQueryBuilder.tsx
Abdkhan14: Right @Zylphrex, I'll just pass them down as props.

static/app/views/explore/content.tsx
```diff
@@ -34,6 +37,15 @@ export function ExploreContent({}: ExploreContentProps) {
...
   const {selection} = usePageFilters();
+  const [groupBys] = useGroupBys();
+  const [visualizes] = useVisualizes();
+  const [resultsMode] = useResultMode();
+
+  const fields = useMemo(() => {
+    return [...groupBys, ...visualizes.flatMap(visualize => visualize.yAxes)].filter(
+      Boolean
+    );
+  }, [groupBys, visualizes]);
```
Zylphrex, line 48: Is there a reason `groupBys` needs to be included here? The `groupBy` is always span attribute, not an aggregation, thus should always be a subset of the fields already available in the autocompletion.

static/app/components/performance/spanSearchQueryBuilder.tsx
```diff
@@ -8,39 +8,88 @@ import {t} from 'sentry/locale';
...
+
 export function SpanSearchQueryBuilder({
   initialQuery,
   searchSource,
   datetime,
   onSearch,
   placeholder,
   projects,
+  fields = [],
+  allowAggregateFunctions,
```
Zylphrex, line 83: I think you can simplify this further and just pass in a list of supported aggregation functions.

So the parent component can simply choose to pass in the list instead of having it here

Zylphrex (approved):

PR #77169, opened 2024-09-09
ref(integrations): Introduce common dispatcher for webhook commands

RyanSkonnord:
Introduce `MessagingIntegrationCommand`, an abstraction for the global set of supported chat commands with consistent strings to invoke them.

Combine code in various places for parsing command text into `MessagingIntegrationCommandDispatcher`.


RyanSkonnord:
The purpose of the new abstraction is to have a central source of truth for the interface exposed by chat commands, rather than having magic strings like `"link"` and `"unlink"` floating around in various packages without any obvious connection between them. This should help to prevent people from introducing inconsistencies, and reduce the friction for supporting new commands in the future.

It's also nice (in my opinion) to abstract out the various if-else blocks that do approximately the same task in the three messaging integrations (even if yielding callables from a `command_handlers` property is a bit more opaque). If we ever want to have metrics or SLOs on this dispatching step, we can now add it in one place.

Any one of those factors alone probably wouldn't have been enough to justify that level of complexity. But, all together, "resolving a chat command into a handler" is a common enough piece of functionality, in my opinion, to justify abstracting it out properly.

src/sentry/integrations/discord/webhooks/command.py
```diff
@@ -37,25 +41,35 @@ class DiscordCommandHandler(DiscordInteractionHandler):
 
     def handle(self) -> Response:
         command_name = self.request.get_command_name()
-        logging_data = self.request.logging_data
+        cmd_input = CommandInput(command_name)
+        dispatcher = MsTeamsCommandDispatcher(self.request)
```
sentaur-athena, line 45: I'm a bit confused as why the class and file are discord but dispatcher is MsTeams 🤔

sentaur-athena: Discussed in team meeting. It's a typo, so please update.

src/sentry/integrations/discord/webhooks/command.py
```diff
@@ -37,25 +41,35 @@ class DiscordCommandHandler(DiscordInteractionHandler):
...
+        return self.send_message(message)
 
-        logger.warning(
-            "discord.interaction.command.unknown", extra={"command": command_name, **logging_data}
-        )
-        return self.help()
 
-    def link_user(self) -> Response:
+@dataclass(frozen=True)
+class MsTeamsCommandDispatcher(MessagingIntegrationCommandDispatcher[str]):
```
sentaur-athena, line 59: Here as well

sentaur-athena (commented):
Changes make sense. Just had a few comments.
Can you please point to where this code is being tested?

RyanSkonnord:
> Can you please point to where this code is being tested?

* Slack is tested by cases in `tests/sentry/integrations/slack/webhooks/` including `commands/test_link_*.py` and `events/test_message_im.py`.
* MSTeams is tested by `tests/sentry/integrations/msteams/test_webhook.py::MsTeamsWebhookTest`.
* Discord is tested by `tests/sentry/integrations/discord/webhooks/test_command.py::DiscordCommandInteractionTest`.

src/sentry/integrations/discord/webhooks/command.py
```diff
@@ -37,25 +41,35 @@ class DiscordCommandHandler(DiscordInteractionHandler):
 
     def handle(self) -> Response:
         command_name = self.request.get_command_name()
```
cathteng, line 43: nit: should we update `get_command_name` to check the lowercase version of the command? i believe we are already doing it for slack and MSTeams

cathteng (approved):
🧠

src/sentry/integrations/discord/webhooks/command.py
RyanSkonnord: Actually, the check is always done case-insensitively (thanks to the `.casefold()` calls in the `CommandSlug` methods), so that shouldn't be necessary.

The downcasing on MSTeams is actually unnecessary, and I suppose maybe we shouldn't do that in case we ever want to preserve case for non-slug arguments in the future. So I took it out (d5374d72e4b74e745472833118aae2cfa7f64bba) instead. Thanks for the catch.

PR #77172, opened 2024-09-09
feat(workflow_engine): Add Workflows

saponifi3d:
## Description
Base DB table for the `Workflows`

Design: https://www.notion.so/sentry/Alerts-Create-Issues-9114a498098143178839d584c293fe75?pvs=4#40649cfdb98f46b49b2121732327d185


src/sentry/workflow_engine/apps.py
```diff
@@ -5,4 +5,4 @@ class Config(AppConfig):
     name = "sentry.workflow_engine"
 
     def ready(self):
-        pass
+        import sentry.workflow_engine.models.workflow  # noqa
```
wedamija, line 8: Why do you import the model here?

tests/sentry/workflow_engine/models/test_workflow.py
```diff
@@ -0,0 +1,19 @@
...
+        self.org.delete()
+
+    def test_create_simple_workflow(self):
+        workflow = Workflow.objects.create(name="test_workflow", organization=self.org)
+
+        workflow.save()
+        workflow.refresh_from_db()
+
+        assert workflow.name == "test_workflow"
+        assert workflow.organization == self.org
```
wedamija, line 19: imo you don't really need to test the orm, but up to you

saponifi3d: yeah, agree, i just wanted a placeholder test with an import / example creation so when i go to start work on the next bit, it's halfway done for me. (i also wanted to confirm that i had setup all the stuff correctly 🎉)

wedamija (approved):

PR #77173, opened 2024-09-09
fix(onboarding): disable invite removal button for members when flag is off

ameliahsu:
We should disable the Remove button for organization members when "Let Members Invite Others" is turned off (AKA when `allowMemberInvite` is false).

### "Let Members Invite Others" turned OFF
All disabled buttons have the same message ("You do not have access to remove members")

Screenshot 2024-09-09 at 11 55 58 AM

### "Let Members Invite Others" turned ON
Disabled invite buttons have a different message ("Your role cannot modify this invite")

Screenshot 2024-09-09 at 11 28 26 AM


sentaur-athena (approved):
This is one of those things that is clear now but in a year none of us will remember what all these booleans mean 😆  Can you please add a bit of comment explaining the logic for memebers?

PR #77174, opened 2024-09-09
fix(suspect-commits): get active integrations

cathteng:
Filter for active integrations for suspect commits. If we get any integration and it happens to be inactive, any subsequent outgoing request will fail in the integration proxy because the integration is not active.

https://github.com/getsentry/sentry/blob/630ae72a47bbab18df1633567198481536203e3c/src/sentry/integrations/api/endpoints/integration_proxy.py#L117-L122

Fixes SENTRY-2CY4


JoshFerge (approved):
🤩

PR #77185, opened 2024-09-09
ref: abstract class properties aren't a thing

asottile-sentry:

### DIFF
--- a/src/sentry/integrations/middleware/hybrid_cloud/parser.py
+++ b/src/sentry/integrations/middleware/hybrid_cloud/parser.py
@@ -1,10 +1,9 @@
 from __future__ import annotations
 
-import abc
 import logging
 from collections.abc import Mapping, Sequence
 from concurrent.futures import ThreadPoolExecutor, as_completed
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, ClassVar
 
 from django.core.cache import cache
 from django.http import HttpRequest, HttpResponse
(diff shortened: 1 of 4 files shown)


src/sentry/integrations/msteams/webhook.py
```diff
@@ -143,8 +143,6 @@ def verify_signature(request):
 
 
 class MsTeamsWebhookMixin:
-    provider = "msteams"
```
asottile-sentry, line 146: every subclass also sets `provider = "msteams"` so this is redundant

src/sentry/integrations/middleware/hybrid_cloud/parser.py
```diff
@@ -53,22 +52,16 @@ def __init__(
...
-class BaseRequestParser(abc.ABC):
+class BaseRequestParser:
     """Base Class for Integration Request Parsers"""
 
     _METRIC_SUCCESS_KEY = "integrations.middleware.request_parser.success"
     _METRIC_FAILURE_KEY = "integrations.middleware.request_parser.failure"
     _METRICS_INFO_KEY = "integrations.middleware.request_parser.info"
 
-    @property
-    def provider(self) -> str:
```
asottile-sentry, line 64: mypy (correctly) sees this as an instance property when the intent was classlevel based on usage and implementation

src/sentry/integrations/middleware/hybrid_cloud/parser.py
```diff
@@ -53,22 +52,16 @@ def __init__(
         self.error = error
 
 
-class BaseRequestParser(abc.ABC):
```
asottile-sentry, line 56: no `@abstractmethod`s so `ABC` doesn't make sense (and just adds runtime overhead)

src/sentry/integrations/middleware/hybrid_cloud/parser.py
```diff
@@ -53,22 +52,16 @@ def __init__(
...
     _METRIC_SUCCESS_KEY = "integrations.middleware.request_parser.success"
     _METRIC_FAILURE_KEY = "integrations.middleware.request_parser.failure"
     _METRICS_INFO_KEY = "integrations.middleware.request_parser.info"
 
-    @property
-    def provider(self) -> str:
-        raise NotImplementedError("'provider' property is required by IntegrationClassification")
-
-    @property
-    def webhook_identifier(self) -> WebhookProviderIdentifier:
```
evanpurkhiser, line 68: Are we losing doc comments here?

asottile-sentry: I didn't really feel like they added any information that they type annotations didn't add 🤷

asottile-sentry: (or the resulting stacktrace if they were missing)

joshuarli (approved):

PR #77188, opened 2024-09-09
feat(utils): Add image load function

priscilawebdev:
Add useful image data fetch function, which is utilized in this [PR](https://github.com/getsentry/getsentry/pull/15067)


static/app/utils/url/fetchImageData.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Fetches and provides an HTMLImageElement from a given URL.
+ * It validates the URL format before attempting to load the image.
+ *
+ * @param {string} url - The URL of the image to fetch and inspect.
+ * @returns {Promise<HTMLImageElement>}
+ * A promise that resolves to the HTMLImageElement object representing the loaded image.
+ *
+ * If the URL is invalid or the image fails to load, the promise is rejected with an error.
+ */
+export function fetchImageData(url: string): Promise<{height: number; width: number}> {
```
vgrozdanic, line 11: Maybe I don't understand this, but is this correct return type: `Promise<{height: number; width: number}>`?

It seems to me that we never resolve object like this, but rather we resolve image object

static/app/utils/url/isImageUrl.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Checks if a URL points to an image file based on common image extensions.
+ * The URL is trimmed and checked for extensions listed in the imageExtensionsPattern.
+ *
+ * @param url - The URL to check.
+ * @returns True if the URL ends with an image extension; otherwise, false.
+ */
+export function isImageUrl(url: string): boolean {
+  const imageExtensionsPattern = /\.(jpg|jpeg|png|gif|bmp|webp|svg)(\?.*)?$/i;
+  return imageExtensionsPattern.test(url.trim());
+}
```
ArthurKnaus, line 11: I think instead of checking extensions you could simply try to fetch the image data.
1. If it is not an image, the browser will abort fetching after receiving the `content-type` header. So there is not a big network overhead.
2. The extension in the URL is no proper indicator for the actual resource that is being served.

static/app/utils/url/fetchImageData.ts
ArthurKnaus: I would also rather return `Promise<HTMLImageElement>`

priscilawebdev: oh yes...I updated the types and missed this one...

priscilawebdev, line 11:
```suggestion
export function fetchImageData(url: string): Promise<HTMLImageElement> {
```

static/app/utils/url/isImageUrl.ts
vgrozdanic: Agree, there could be files that are images without extension

vgrozdanic (approved):

static/app/utils/url/fetchImageData.ts
```diff
@@ -0,0 +1,36 @@
...
+      }
+    };
+
+    img.src = url;
+
+    return () => {
+      isCanceled = true;
+      img.onload = null;
+      img.onerror = null;
+    };
```
vgrozdanic, line 34: Do we even need this?

priscilawebdev: probably not, but I like to clean up :)

PR #77192, opened 2024-09-09
chore(anomaly detection): surface generic caught exeption 

mifu67:
Surface info about generic caught exception in the logs


src/sentry/incidents/serializers/alert_rule.py
```diff
@@ -505,7 +505,11 @@ def create(self, validated_data):
             except forms.ValidationError as e:
                 # if we fail in create_metric_alert, then only one message is ever returned
                 raise serializers.ValidationError(e.error_list[0].message)
-            except Exception:
+            except Exception as e:
+                logger.exception(
+                    "Error when creating alert rule",
+                    extra={"details": str(e)},
```
mifu67, line 511: I know exceptions have a string but I would prefer to wrap `e` in `str()` for safety

nhsiehgit (approved):

PR #77204, opened 2024-09-09
fix(commits): early return if no groups

JoshFerge:
in the receiver for `resolved_in_commit`, we do not early return if there are no groups. Which means we end up doing a lot of unecessary hybrid cloud calls to get the commit author, even if we don't need to. 

This should reduce the number of get_user calls here by a lot.

```
select
  count(*),
  linked_type
from
  sentry_grouplink
group by
  linked_type
  ```
shows we don't have too many "resolved in commits" in our db compared to the number of commits.


mrduncan (approved):
awesome find!

PR #77206, opened 2024-09-09
feat(hydration error): Improve rendering of hydration error diff modal

ryan953:
This updates the Hydration Error diff modal to be taller, and so that the Slider tool scales inside the available space

SCR-20240909-ngvx

### DIFF
--- a/static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
+++ b/static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
@@ -70,7 +70,7 @@ export default function ReplayComparisonModal({
           {isSameTimestamp ? (
             <Alert type="warning" showIcon>
               {t(
-                "Sentry wasn't able to identify the correct event to display a diff for this hydration error."
+                "Cannot display diff for this hydration error. Sentry wasn't able to identify the correct event."
               )}
             </Alert>
           ) : (
--- a/static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
+++ b/static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
@@ -1,11 +1,12 @@
 import styled from '@emotion/styled';
 
 import type {ModalRenderProps} from 'sentry/actionCreators/modal';
+import Alert from 'sentry/components/alert';
 import FeatureBadge from 'sentry/components/badge/featureBadge';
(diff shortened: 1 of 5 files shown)


static/app/components/replays/diff/replaySliderDiff.tsx
```diff
@@ -19,15 +19,22 @@ import {useResizableDrawer} from 'sentry/utils/useResizableDrawer';
 
 interface Props {
   leftOffsetMs: number;
+  minHeight: `${number}px` | undefined;
```
ryan953, line 22: This is not optional, (ie: not `minHeight?:`) because i think it should be set each time.  When this is set to `undefined` its possible for it to look like this and be frusterating:
SCR-20240909-niqw

billyvg (approved):

static/app/components/replays/diff/replaySliderDiff.tsx
billyvg: Why allow `undefined` then?

ryan953: if the parent node is setup properly, then undefined works to give it an unconstrained height growing and shinking

ryan953: i guess, i could see what `0px` looks like and update the defaults

static/app/components/replays/breadcrumbs/replayComparisonModal.tsx
```diff
@@ -53,20 +56,33 @@ export default function ReplayComparisonModal({
...
+                boldBefore: <strong />,
+                boldAfter: <strong />,
+              }
+            )}
+          </StyledParagraph>
+
+          {isSameTimestamp ? (
+            <Alert type="warning" showIcon>
+              {t(
+                "Sentry wasn't able to identify the correct event to display a diff for this hydration error."
```
michellewzhang, line 73:
```suggestion
                "Sentry wasn't able to identify the correct event for which to display a diff for this hydration error."
```

michellewzhang: something about this sentence reads weird, trying to think of a better way to phrase it

michellewzhang (approved):

PR #77217, opened 2024-09-09
feat(workflow_engine): Add DataSource model

wedamija:
This adds in the basic `DataSource` model for the workflow_engine project.


src/sentry/workflow_engine/models/data_source.py
```diff
@@ -0,0 +1,23 @@
...
+
+@region_silo_model
+class DataSource(DefaultFieldsModel):
+    __relocation_scope__ = RelocationScope.Organization
+
+    class Type(models.IntegerChoices):
+        SNUBA_QUERY_SUBSCRIPTION = 1
+
+    organization = FlexibleForeignKey("sentry.Organization")
+    query_id = BoundedBigIntegerField()
```
wedamija, line 20: I wasn't sure if we actually needed the other `query` column, anymore, since I think we were planning on avoiding json?

src/sentry/workflow_engine/models/data_source.py
```diff
@@ -0,0 +1,23 @@
...
+    region_silo_model,
+)
+
+
+@region_silo_model
+class DataSource(DefaultFieldsModel):
+    __relocation_scope__ = RelocationScope.Organization
+
+    class Type(models.IntegerChoices):
+        SNUBA_QUERY_SUBSCRIPTION = 1
```
wedamija, line 17: I assume this is what we're using type for here?

saponifi3d, line 17: just wanted to double check this will grow?

saponifi3d (approved):
🎉 :shipit:

src/sentry/workflow_engine/models/data_source.py
saponifi3d: i think we can figure that out when we get to issue alerts, the `query` here was meant to support the class id structure.

src/sentry/workflow_engine/models/data_source.py
saponifi3d: 👍 - i do think we'll have more types, but we can add as needed.

wedamija: Yeah, we can just add more types here over time. I assume the idea is that this is an enum, and not against some dynamic db source?

saponifi3d: yep!

src/sentry/workflow_engine/models/data_source.py
```diff
@@ -0,0 +1,23 @@
...
+
+    organization = FlexibleForeignKey("sentry.Organization")
+    query_id = BoundedBigIntegerField()
+    type = models.SmallIntegerField(choices=Type.choices)
+
+    indexes = [models.Index(fields=("type", "query_id"))]
```
markstory, line 23: Do you plan on querying by organization + type/query_id? You might need additional indexes.

wedamija: To be honest I'm not sure how exactly we'll end up querying this, but probably having an org based index here will help

PR #77219, opened 2024-09-09
feat(similarity): Allow similarity backfill to create project cohorts

jangjodi:
Create project cohorts based on project_id % numer_of_threads == worker_number to uniquely assign project ids to worker threads

This will automate the backfilling of grouping records


src/sentry/tasks/embeddings_grouping/utils.py
```diff
@@ -96,6 +96,26 @@ def filter_snuba_results(snuba_results, groups_to_backfill_with_no_embedding, pr
...
 
+def create_project_cohort(worker_number: int, last_processed_project_id: int | None) -> list[int]:
+    """
+    Create project cohort by the following calculation: project_id % threads == worker_number
+    to assign projects uniquely to available threads
+    """
+    project_id_filter = Q()
+    if last_processed_project_id is not None:
+        project_id_filter = Q(id__gt=last_processed_project_id)
+    thread_number = options.get("similarity.backfill_seer_threads")
```
JoshFerge, line 107: this option already existed and represents the number of concurrent requests we'd make to seer, but this is something different no? shouldn't thread number be passed into the backfill?

JoshFerge (commented):
really like how you did this! looks good other than question on seer threads option

src/sentry/tasks/embeddings_grouping/utils.py
JoshFerge: right, but `thread_number` is already used and means something different, no?

src/sentry/seer/similarity/utils.py
```diff
@@ -265,3 +266,16 @@ def _is_snipped_context_line(context_line: str) -> bool:
...
+
+
+def project_is_seer_eligible(project: Project) -> bool:
+    """
+    Return True if the project hasn't already been backfilled, is a Seer-eligible platform, and
+    the feature is enabled in the region.
+    """
+    # TODO: use this check in team_projects
+    is_backfill_completed = project.get_option("sentry:similarity_backfill_completed")
+    is_seer_eligible_platform = project.platform in SEER_ELIGIBLE_PLATFORMS
```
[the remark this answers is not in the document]
JoshFerge: got it. yeah event platform != project platform. good to have separate lists.

src/sentry/tasks/embeddings_grouping/utils.py
jangjodi: ah I thought they were the same thing, I'll make it an option instead of another parameter to pass in, as it should be constant

JoshFerge (approved):

PR #77223, opened 2024-09-10
style(hydration errors): consistent before/after style in modal

### LINKED ISSUE #77210: [Hydration Errors] Update Modal so all references to "Before" and "After" are bold, red&green, not say "Before Hydration"

michellewzhang:
closes https://github.com/getsentry/sentry/issues/77210

https://github.com/user-attachments/assets/b3b7c0ad-068f-4155-b82e-6bddec0fbd57


static/app/components/replays/diff/replaySideBySideImageDiff.tsx
ryan953, line 1: One other thing I think would be helpful is to throw some borders around the children of `<ReplayGrid>`. Then the images that are side-by-side will have a border between them.

ryan953 (approved):

static/app/components/replays/diff/replaySideBySideImageDiff.tsx
michellewzhang: like a green / red border?

ryan953: Ya

michellewzhang: yes you made a [ticket](https://github.com/getsentry/sentry/issues/77209) for it! i'll do it in a followup pr

ryan953: goldfish memory

billyvg (commented):
Should these all styled components instead of using inline `css`/`style` prop?

michellewzhang:
> Should these all styled components instead of using inline `css`/`style` prop?

i can clean this up in a followup @billyvg

PR #77224, opened 2024-09-10
Salt seer secrets with nonce based on option

corps:
Working to improve tests

### DIFF
--- a/src/sentry/api/endpoints/group_ai_summary.py
+++ b/src/sentry/api/endpoints/group_ai_summary.py
@@ -22,7 +22,7 @@ from sentry.constants import ObjectStatus
 from sentry.eventstore.models import GroupEvent
 from sentry.models.group import Group
 from sentry.models.project import Project
-from sentry.seer.signed_seer_api import sign_with_seer_secret
+from sentry.seer.signed_seer_api import get_seer_salted_url, sign_with_seer_secret
 from sentry.types.ratelimit import RateLimit, RateLimitCategory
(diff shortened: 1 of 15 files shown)


src/sentry/api/endpoints/group_ai_summary.py
```diff
@@ -114,13 +114,14 @@ def _call_seer(
             option=orjson.OPT_NON_STR_KEYS,
         )
 
+        url, salt = get_seer_salted_url(f"{settings.SEER_AUTOFIX_URL}{path}")
```
ram-senth, line 117: minor nit - should we call this `salted_url` instead of salt?

corps: In the old code path, the salt was the url, but in the new code path, the salt is just the nonce itself, minus url context.  So that's why I kept the name more ambiguous.  It does return a url, which is the url with the salt (nonce) added separately from the signature.

ram-senth: Actually my bad. It should be just salt.

corps:
Matched test signatures here https://github.com/getsentry/seer/commit/ec794e115d886d23bfc84997bc1a2825254b1466#diff-9fea354b4f6dc2b8289ecde38e43fc8000a8516a66e2d91de9cf615ffbd08b37R225

Ensures this works e2e

ram-senth (approved):

PR #77230, opened 2024-09-10
fix(source-maps): Fix artifact bundle link

matejminar:
This likely got broken with the recent react router 6 bump.


ArthurKnaus (approved):
🙌

PR #77232, opened 2024-09-10
chore: add convert_args for ProjectPluginDetailsEndpoint

oioki:


src/sentry/api/endpoints/project_plugin_details.py
```diff
@@ -81,7 +92,7 @@ def post(self, request: Request, project, plugin_id) -> Response:
                 organization=project.organization,
                 target_object=project.id,
                 event=audit_log.get_event_id("INTEGRATION_EDIT"),
-                data={"integration": plugin_id, "project": project.slug},
+                data={"integration": plugin.slug, "project": project.slug},
```
iamrajjoshi, line 95: i believe `plugin_id` previously was the actually the plugin_slug (if you take a look at our audit log). we probably want to show the plugin slug in the log because its is a common name and i am not sure if we surface the plugin id in our UI anywhere

iamrajjoshi (commented):
thanks for this! just quick comment on the audit log update

src/sentry/api/endpoints/project_plugin_details.py
oioki: New `plugin` is `plugin = plugins.get(plugin_id)`. The plugin is looked up on the `slug` field here:
https://github.com/getsentry/sentry/blob/f0833a312bb6ee0038d21bfcd70f15789129a36f/src/sentry/plugins/base/manager.py#L73-L76
Therefore `plugin_id` is the same as `plugins.get(plugin_id).slug`.

iamrajjoshi (approved):

PR #77233, opened 2024-09-10
chore(metrics): Remove custom manager for dashboard widgets

vgrozdanic:
Removes custom manager for dashboard widgets, since it was added for span based attribute metrics.


shellmayr (approved):
LGTM

PR #77234, opened 2024-09-10
chore(metrics): Remove custom array lookup

vgrozdanic:
Removes custom array lookup created for custom lookups related to span attribute metrics.

This is no longer used anywhere and there is no need to keep it in a codebase.


shellmayr (approved):
This was a good one 😭 LGTM

PR #23861, opened 2021-02-12
feat: Hierarchical grouping hashes

untitaker:
Implement ability to generate multiple grouping hashes from within every strategy. Previously https://github.com/getsentry/sentry/pull/23829 failed. This PR instead refactors every grouping strategy to deal with potentially generating multiple variants.

See https://www.notion.so/sentry/Hierarchical-Grouping-477043c8aed940d2a1d78386288c6e97 for broader context.

Next steps:

- [x] decide which hashes to actually produce for stacktraces
- [x] potentially add system hash as highest-granularity hash? -- no, we instead have two types of hashes -- hierarchical ones and flat ones
- [x] do we want to keep 0frames variant? -- no
- [x] reviewing unapproved snapshots for mobile strategy
- [x] code cleanup
- [x] force all strategies to return variant dict
- [ ] remove get_grouping_component dispatcher -- not doing that, not a great idea

**unresolved questions**

- do we want to produce hierarchies if none of the frames are in-app? answer: punt on it for now and see if this actually happens after latest fixes to in-app normalization
- do we want to merge this into master before iterating on the strategy for mobile? likely yes just to avoid merge conflicts

PR #77235, opened 2024-09-10
drop(grouping): Deprecate hierarchical code (#76994)

armenzg:
This is merging #76994 again but manually fixing the issues discovered during that deployment.

This partially reverts #23861

No projects in SaaS are using the mobile grouping anymore (this is the only one using the hierarchical code).

Some code has already been reverted in:
* https://github.com/getsentry/sentry/pull/76951
* https://github.com/getsentry/sentry/pull/76991
* https://github.com/getsentry/sentry/pull/76992
* https://github.com/getsentry/sentry/pull/76993

More code needs to be removed but this is the minimal removal which passes all tests.


JoshFerge (approved):
bravo!

PR #77249, opened 2024-09-10
fix(integrations): get active integrations

cathteng:
For any integration that makes a call to an external API which passes through the integration proxy, we need to make sure we're always fetching active integrations, otherwise the request will be rejected by the integration proxy. This PR addresses a few features for SCMs, namely the PR comments features, that are not currently doing this.

While Slack no longer uses the integration proxy, we should still check for active integrations as requests made by inactive integrations will still technically be rejected by the integration proxy -- applying a similar approach there.

This is NOT all the places that need to be fixed, but just some easy wins that I'm sure about.

### DIFF
--- a/src/sentry/integrations/slack/sdk_client.py
+++ b/src/sentry/integrations/slack/sdk_client.py
@@ -6,6 +6,7 @@ from slack_sdk import WebClient
 from slack_sdk.errors import SlackApiError
 from slack_sdk.web import SlackResponse
 
+from sentry.constants import ObjectStatus
 from sentry.db.postgres.transactions import in_test_hide_transaction_boundary
 from sentry.integrations.base import disable_integration, is_response_error, is_response_success
 from sentry.integrations.models import Integration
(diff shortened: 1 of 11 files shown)


src/sentry/integrations/slack/sdk_client.py
```diff
@@ -119,9 +120,13 @@ def __init__(self, integration_id: int):
             # this is a read operation.
             """
             with in_test_hide_transaction_boundary():
```
GabeVillalobos, line 122: oh this is gross. We may want to revisit this transaction boundary in the future 😬

tests/sentry/integrations/github/tasks/test_open_pr_comment.py
```diff
@@ -1178,9 +1181,9 @@ def test_comment_workflow_missing_integration(
         mock_reverse_codemappings,
         mock_pr_filenames,
     ):
-        # invalid integration id
-        self.gh_repo.integration_id = 0
-        self.gh_repo.save()
+        # inactive integration
+        with assume_test_silo_mode(SiloMode.CONTROL):
```
GabeVillalobos, line 1185: This is sort of an outdated way to do this. You can use `assume_test_silo_mode_of(Integration)` to infer which silo a specific model (or multiple models) should be using instead which is a bit cleaner.

GabeVillalobos (approved):
Minor cleanup comment, otherwise looks good! Nice work

tests/sentry/integrations/github/tasks/test_open_pr_comment.py
cathteng: fixed!

PR #77251, opened 2024-09-10
feat(replay): Support absolute time

### LINKED ISSUE #52682: Support Absolute Time for Replay Details
### Problem Statement

Sometimes when troubleshooting it's nice to be able to compare timestamps of the replay session with other logs or bug reports.  FullStory has the ability to toggle the timeline between session time and absolute time:

https://help.fullstory.com/hc/en-us/articles/360020623614-Understanding-session-time-and-how-to-switch-to-absolute-time

### Solution Brainstorm

It would be useful to support the ability to toggle between session time and absolute time.

### Product Area

Replays

c298lee:
Adds support for absolute time.

All timestamps on the replay details page and breadcrumbs list support absolute time
image

The timestamp tooltip has changes to show the absolute time and the time within replay in milliseconds
image

Closes: https://github.com/getsentry/sentry/issues/52682


static/app/components/replays/timeAndScrubberGrid.tsx
```diff
@@ -58,7 +59,8 @@ export default function TimeAndScrubberGrid({
   isCompact = false,
   showZoom = false,
 }: TimeAndScrubberGridProps) {
-  const {currentTime, replay} = useReplayContext();
+  const {currentTime, replay, timestampType} = useReplayContext();
```
ryan953, line 62: I wanna suggest using these lines instead of adding stuff to `useReplayContext`:

```
import {useReplayPrefs} from 'sentry/utils/replays/playback/providers/replayPreferencesContext';

const [prefs] = useReplayPrefs();
const timestampType = prefs.timestamp;

// OR, if you wanna compress it:
const [{timestamp}] = useReplayPrefs();
```

nbd either way, but i have a longer term vision to refactor useReplayContext into some smaller parts

static/app/views/replays/detail/timestampButton.tsx
```diff
@@ -22,18 +26,52 @@ export default function TimestampButton({
   startTimestampMs,
   timestampMs,
 }: Props) {
+  const {timestampType} = useReplayContext();
+  const timestampMsAsTime = new Date(timestampMs).getTime();
```
ryan953, line 30: This statement should be unneccessary. `getTime()` will convert a `Date` into number of milliseconds since the epoch, but `timestampMs` should already have that same value.

static/app/components/replays/preferences/replayPreferences.tsx
```diff
@@ -5,16 +5,19 @@ const LOCAL_STORAGE_KEY = 'replay-config';
 export type ReplayPrefs = {
   isSkippingInactive: boolean;
   playbackSpeed: number;
+  timestamp: 'relative' | 'absolute';
```
ryan953, line 8: nit: maybe this should be named `timestampType` because that's how it's used in the components `timestampType === 'absolute' ? ... : ...`

ryan953 (approved):

ryan953:
Looking around another spot we can get in the next PR is the tooltip inside the timeline. There might be other spots too, i havn't made a big list, but we can go around and find all the little corners after this base is out!

billyvg (commented):
I haven't looked at this too carefully yet, but one thing to note with absolute timestamps is that we have two different user preferences that can affect how you render the timestamps: 24 vs 12 hr and timezones

Something to double check

static/app/components/replays/player/scrubber.tsx
```diff
@@ -63,11 +65,21 @@ function Scrubber({className, showZoomIndicators = false}: Props) {
...
-                precision: 'ms',
-                style: 'hh:mm:ss.sss',
-              })}
+              labelText={
+                timestampType === 'absolute'
+                  ? getFormattedDate(
+                      startTimestamp + currentHoverTime,
+                      shouldUse24Hours() ? 'HH:mm:ss.SSS' : 'hh:mm:ss.SSS',
+                      {
+                        local: true,
```
aliu39, line 74: Is startTimeStamp already in the right timezone?

static/app/components/replays/player/scrubber.tsx
```diff
@@ -63,11 +65,21 @@ function Scrubber({className, showZoomIndicators = false}: Props) {
...
-              labelText={formatDuration({
-                duration: [currentHoverTime, 'ms'],
-                precision: 'ms',
-                style: 'hh:mm:ss.sss',
-              })}
+              labelText={
+                timestampType === 'absolute'
+                  ? getFormattedDate(
+                      startTimestamp + currentHoverTime,
+                      shouldUse24Hours() ? 'HH:mm:ss.SSS' : 'hh:mm:ss.SSS',
```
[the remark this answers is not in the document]
c298lee: I don't think we need the AM/PM in the scrubber since the timeline has AM/PM, but if you think it'll be better to have in the scrubber too, we can add it in as well!

static/app/components/replays/player/scrubber.tsx
c298lee: all the timestamps we are getting are in milliseconds since epoch, eg. 1725993813532 so we only have to worry about timezones when we are formatting the times

c298lee:
> I haven't looked at this too carefully yet, but one thing to note with absolute timestamps is that we have two different user preferences that can affect how you render the timestamps: 24 vs 12 hr and timezones
> 
> Something to double check

@billyvg  Yes, I made sure we render all of those properly, however if you notice any gaps, please let me know! 
I made some decisions about when we show things like AM/PM, the timezone, and the second vs millisecond precision, but I'm open to changing those as well!
|location|what we show|
|------|--------|
|timeline| time in second precision + AM/PM|
|scrubber| time in millisecond precision|
|timestamp|time in second precision + AM/PM|

In the timestamp tooltip, we show both the absolute time and relative time. 
The absolute time part contains the date, time in seconds precision, AM/PM and timezone. 
The relative time part contains the relative time in milliseconds precision.
The milliseconds for both the absolute time and relative time are the same, so I only decided to include it in relative time since we already display a lot of information for the absolute time.

Super open to changing any of these! This just made the most sense to me.

c298lee:
> Looking around another spot we can get in the next PR is the tooltip inside the timeline. There might be other spots too, i havn't made a big list, but we can go around and find all the little corners after this base is out!

@ryan953 what do you mean by the tooltip inside the timeline? The formatting is a little bit different since we show ms precision, but I can add in the AM/PM if that will help with consistency
image

static/app/components/replays/player/scrubber.tsx
aliu39: Yeah, so should we always use local: true here or switch it based on the timezone?

static/app/components/replays/player/scrubber.tsx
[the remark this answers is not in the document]
aliu39: Oo didn't notice that, in that case I'm fine either way!

billyvg (commented):
image
This is a bit hard to read, would be better if it were a grid so that it was more table-like

* I think AM/PM should be hidden, it adds noise (especially in the tabs like network/console) and doesn't provide too much value, especially since it's unlikely to have two different values (except obviously when the day changes). This being said, we can include them in tooltips so you can find them if needed. [[[[[also 24 hr > 12 hr]]]]]
  * Should we add tooltips over the start/end timestamps in the scrubber? (no strong opinion here)

* A topic to follow-up on, I would love for all timestamps in the tabs (console/network/etc) to be in millisec precision

ryan953:
> > Looking around another spot we can get in the next PR is the tooltip inside the timeline. There might be other spots too, i havn't made a big list, but we can go around and find all the little corners after this base is out!
> 
> @ryan953 what do you mean by the tooltip inside the timeline? The formatting is a little bit different since we show ms precision to align with the behaviour of relative time, but I can add in the AM/PM if that will help with consistency image


Ooh, nice screen shot with two things on display. I was just thinking about that this one with the blue arrow pointing to it 
366179139-954d89dd-9ea4-4e76-b04c-39d9831a15d2

could look like this one with the two formats displayed at the same time:
SCR-20240910-lnbf


It was a thought, there's lots of spots each to consider!

ryan953:
@billyvg my gut reaction is that I like having the AM/PM in there, especially when it's a 12 hour time because it might be easy to confuse with a duration if the hour is low. For example 30 mins after lunch, and `1:30PM`, that could look like a 90 minute replay to someone not paying attention.

But there's other ways to disambiguate if it becomes a problem... although the other ways would all add noise... the tooltips help to clarify things because both units are there at the same time so maybe i'm just imagining things. Ok nevermind me!

billyvg:
> @billyvg my gut reaction is that I like having the AM/PM in there, especially when it's a 12 hour time because it might be easy to confuse with a duration if the hour is low. For example 30 mins after lunch, and `1:30PM`, that could look like a 90 minute replay to someone not paying attention.
> 
> But there's other ways to disambiguate if it becomes a problem... although the other ways would all add noise... the tooltips help to clarify things because both units are there at the same time so maybe i'm just imagining things. Ok nevermind me!

@ryan953 I could definitely see that being the case in the scrubber. another pt to consider is that the user currently needs to opt into absolute time

static/app/components/replays/player/scrubber.tsx
c298lee: local is whether or not we use the timezone or not (uses UTC if false). `local` isn't a great name for that imo

PR #77265, opened 2024-09-10
style(replay): rm some inline styles

michellewzhang:
- clean up some inline styles in hydration error files + other replay files i happened to find
- added some margins to the headers as well.

https://github.com/user-attachments/assets/6d062fd4-ab42-4324-84f9-fa0e4eda677f


billyvg (approved):
🙏

PR #77267, opened 2024-09-10
feat(insights): Adds insights alerts for http module

edwardgou-sentry:
Adds create alert buttons to http module charts.
Includes changes to `insightsMetricField` to support functions with non MRI args.
Also organizes some constants better.

### DIFF
--- a/static/app/views/alerts/rules/metric/insightsMetricField.tsx
+++ b/static/app/views/alerts/rules/metric/insightsMetricField.tsx
@@ -5,8 +5,9 @@ import Tag from 'sentry/components/badge/tag';
 import SelectControl from 'sentry/components/forms/controls/selectControl';
 import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
-import type {MetricMeta, ParsedMRI} from 'sentry/types/metrics';
+import type {MetricAggregation, MetricMeta, ParsedMRI} from 'sentry/types/metrics';
 import type {Project} from 'sentry/types/project';
+import {parseFunction} from 'sentry/utils/discover/fields';
 import {getDefaultAggregation} from 'sentry/utils/metrics';
 import {getReadableMetricType} from 'sentry/utils/metrics/formatters';
 import {
@@ -15,7 +16,6 @@ import {
   formatMRI,
   isMRI,
   MRIToField,
-  parseField,
   parseMRI,
 } from 'sentry/utils/metrics/mri';
 import {useVirtualizedMetricsMeta} from 'sentry/utils/metrics/useMetricsMeta';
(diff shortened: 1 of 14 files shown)


edwardgou-sentry:
Slightly lengthy pr 😅 Most of the changes are just to update the chart components to propagate the query filters to `ChartPanel`

static/app/views/insights/http/alerts.ts
```diff
@@ -0,0 +1,27 @@
+import type {AlertConfig} from 'sentry/views/insights/common/components/chartPanel';
+
+export const ALERTS: Record<string, AlertConfig> = {
+  spm: {
+    aggregate: 'spm()',
+    query: 'span.module:http span.op:http.client',
```
DominikB2014, line 6: This query looks the same across all alerts, I think it would be good if we create a shared variable for it.

static/app/views/insights/http/views/httpDomainSummaryPage.tsx
```diff
@@ -292,6 +293,7 @@ export function HTTPDomainSummaryPage() {
                 series={[durationData[`avg(${SpanMetricsField.SPAN_SELF_TIME})`]]}
                 isLoading={isDurationDataLoading}
                 error={durationError}
+                filters={filters}
```
DominikB2014, line 296: Long term, it would be cool if we didn't have to pass in filters into every single chart, and instead determine the filters based on the module + url.

DominikB2014 (approved):

static/app/views/alerts/rules/metric/insightsMetricField.tsx
```diff
@@ -159,8 +193,8 @@ function InsightsMetricField({aggregate, project, onChange}: Props) {
   );
 
   // When using the async variant of SelectControl, we need to pass in an option object instead of just the value
-  const selectedMriOption = selectedValues?.mri && {
-    label: formatMRI(selectedValues.mri),
+  const selectedOption = selectedValues?.mri && {
+    label: isMRI(selectedValues.mri) ? formatMRI(selectedValues.mri) : selectedValues.mri,
```
mjq, line 197: I'm curious why the `isMRI` check is necessary? Given that `selectedValues.mri` already has type `MRI` (which is the type predicate that `isMRI` is checking, and what `formatMRI` expects) 🤔

edwardgou-sentry: Hey, good question! So `selectedValues` comes from a helper `parseField` function which just naively pulls out the argument(`mri`) using regex, and then casts the type as `MRI` regardless of whether it is actually true or not.

Insights functions like `http_response_rate(3)` have arguments that are not `MRI`s which is why we now have to run into this type check scenario.

I'm going to update to not use `parseField` so we don't make any assumptions on type!

PR #77269, opened 2024-09-10
fix(new-trace): Fixing hook rendering mismatch caused by drawer cards

Abdkhan14:
Previously we tried to filter out components that returned null from within:
```
//<Component1/> --> returns a <Card/>
// {null}
// <Component2/> --> returns null
// Gives us back [<Component1/>]
```
This leads to `Rendered more hooks than during the previous render` error as we try to conditionally render the children (with hooks) to see if they return null: [example](https://sentry.sentry.io/issues/5800979093/?project=11276&query=rendered%20more%20hooks&referrer=issue-stream&sort=date&statsPeriod=7d&stream_index=6)

Now we moved the rendering logic to the parent of Component, and just filter out null values, without having to render the children:
```
//<Component1/> --> returns a <Card/>
// {null}
// { condition ? <Component2/> : null} where Component2 returns a Card
// Gives us back [<Component1/>]
```

### DIFF
--- a/static/app/components/keyValueData/index.tsx
+++ b/static/app/components/keyValueData/index.tsx
@@ -1,4 +1,4 @@
-import {Children, isValidElement, type ReactNode, useRef, useState} from 'react';
+import {Children, type ReactNode, useRef, useState} from 'react';
 import React from 'react';
 import {css} from '@emotion/react';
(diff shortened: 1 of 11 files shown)


edwardgou-sentry (approved):
some nits, feel free to address!

PR #77274, opened 2024-09-10
fix(slack): Remove Org from Logging Payload

iamrajjoshi:
quick fix for [SENTRY-3EMV](https://sentry.sentry.io/issues/5827836858/), org doesn't exist in slack request


cathteng (approved):
doesn't exist on this particular slack request 👍

PR #77280, opened 2024-09-10
fix(datadog): make disable_buffering a param instead of setting it on statsd object

JoshFerge:
set `disable_buffering` as a kwarg that we pass to `initialize`. this makes it so it only gets set once (right now it gets set in `initialize` and afterward here.

we're seeing slow profiles of this codepath in control silo. want to see if this helps.

slow profile example: https://sentry.sentry.io/profiling/profile/sentry/ad932d5ba6f244c4bb7abfc29667da09/flamegraph/?colorCoding=by%20system%20vs%20application%20frame&fov=0%2C87%2C7048375808%2C14&query=&sorting=call%20order&tid=135938038548160&view=top%20down

docs: https://datadogpy.readthedocs.io/en/stable/#datadog.initialize

### DIFF
--- a/src/sentry/metrics/dogstatsd.py
+++ b/src/sentry/metrics/dogstatsd.py
@@ -24,7 +24,7 @@ class DogStatsdMetricsBackend(MetricsBackend):
     def __init__(self, prefix: str | None = None, **kwargs: Any) -> None:
         # TODO(dcramer): it'd be nice if the initialize call wasn't a global
         self.tags = kwargs.pop("tags", None)
-        kwargs["disable_buffering"] = False
+        kwargs["statsd_disable_buffering"] = False
 
         initialize(**kwargs)
         statsd.disable_telemetry()
--- a/src/sentry/metrics/dogstatsd.py
+++ b/src/sentry/metrics/dogstatsd.py
@@ -24,9 +24,10 @@ class DogStatsdMetricsBackend(MetricsBackend):
     def __init__(self, prefix: str | None = None, **kwargs: Any) -> None:
         # TODO(dcramer): it'd be nice if the initialize call wasn't a global
         self.tags = kwargs.pop("tags", None)
+        kwargs["disable_buffering"] = False
+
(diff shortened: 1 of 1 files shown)


GabeVillalobos (approved):

GabeVillalobos:
Some context, here's the setter in question in the python dd client code: https://github.com/DataDog/datadogpy/blob/7b721e331d4287acd5bf86a5e06cb2d97c4f2a6d/datadog/dogstatsd/base.py#L613

src/sentry/metrics/dogstatsd.py
```diff
@@ -24,9 +24,10 @@ class DogStatsdMetricsBackend(MetricsBackend):
     def __init__(self, prefix: str | None = None, **kwargs: Any) -> None:
         # TODO(dcramer): it'd be nice if the initialize call wasn't a global
         self.tags = kwargs.pop("tags", None)
+        kwargs["disable_buffering"] = False
```
markstory, line 27: Should this be `statsd_disable_buffering`? That parameter name is more [aligned with dogstats](https://github.com/DataDog/datadogpy/blob/7b721e331d4287acd5bf86a5e06cb2d97c4f2a6d/datadog/__init__.py#L41-L146)

JoshFerge: great catch, thank you.

GabeVillalobos: Wait, doesn't this effectively make it identical to what we have now? https://github.com/DataDog/datadogpy/blob/7b721e331d4287acd5bf86a5e06cb2d97c4f2a6d/datadog/__init__.py#L143

JoshFerge: my hypothesis is as follows: there could be period before the buffering is enabled where it tries to flush metrics synchronously. by doing this on initialization, maybe this occurs less often / not at all. additionally, ts only calling this function once instead of twice

markstory (approved):

PR #77285, opened 2024-09-10
fix(rr6): Create routes after getsentry registerHooks

evanpurkhiser:
Before this change we were building the route tree at module load time for the routes.tsx file. This was **before** getsentry will have registered it's hooks, and thus the getsentry routes would never be registered.

### DIFF
--- a/static/app/main.tsx
+++ b/static/app/main.tsx
@@ -1,3 +1,4 @@
+import {useState} from 'react';
 import {Router, RouterContext} from 'react-router';
 import {createBrowserRouter, RouterProvider} from 'react-router-dom';
 import {wrapCreateBrowserRouter} from '@sentry/react';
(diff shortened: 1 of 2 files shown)


scttcper (approved):

static/app/main.tsx
```diff
@@ -34,21 +37,27 @@ function renderRouter(props: any) {
...
+    DANGEROUS_SET_REACT_ROUTER_6_HISTORY(router);
+  }
 
-if (window.__SENTRY_USING_REACT_ROUTER_SIX) {
-  DANGEROUS_SET_REACT_ROUTER_6_HISTORY(router);
+  return router;
 }
 
 function Main() {
+  const [router6] = useState(createReactRouter6Routes);
```
scttcper, line 52: could be a useRef but state or memo is about the same at top level

PR #77286, opened 2024-09-10
ref(replay): try 5s rage/dead click timeout for sentry orgs

aliu39:
See slack thread https://sentry.slack.com/archives/C03USURCFBJ/p1725645076151569. This flag is only enabled for internal orgs

Ref https://github.com/getsentry/sentry/pull/77284 and https://github.com/getsentry/sentry-options-automator/pull/2279

### DIFF
--- a/src/sentry/replays/usecases/ingest/dom_index.py
+++ b/src/sentry/replays/usecases/ingest/dom_index.py
@@ -396,6 +396,15 @@ def _handle_breadcrumb(
 ) -> ReplayActionsEventPayloadClick | None:
 
     click = None
+    project = Project.objects.get(id=project_id)
+    dead_click_timeout = (
+        5000
+        if features.has(
+            "organizations:session-replay-dead-click-reduced-timeout",
+            project.organization,
+        )
+        else 7000
+    )
 
     payload = event["data"].get("payload", {})
     if not isinstance(payload, dict):
@@ -412,7 +421,8 @@ def _handle_breadcrumb(
         timeout = payload["data"].get("timeAfterClickMs", 0) or payload["data"].get(
             "timeafterclickms", 0
         )
-        if is_timeout_reason and is_target_tagname and timeout >= 7000:
+
+        if is_timeout_reason and is_target_tagname and timeout >= dead_click_timeout:
             is_rage = (
                 payload["data"].get("clickCount", 0) or payload["data"].get("clickcount", 0)
             ) >= 5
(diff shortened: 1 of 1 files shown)


src/sentry/replays/usecases/ingest/dom_index.py
```diff
@@ -396,6 +396,15 @@ def _handle_breadcrumb(
 ) -> ReplayActionsEventPayloadClick | None:
 
     click = None
+    project = Project.objects.get(id=project_id)
+    dead_click_timeout = (
+        5000
```
JoshFerge, line 401: you could also make this an option if you wanted to have it be more controllable. but this makes sense!

michellewzhang (approved):

src/sentry/replays/usecases/ingest/dom_index.py
```diff
@@ -396,6 +396,15 @@ def _handle_breadcrumb(
...
+        if features.has(
+            "organizations:session-replay-dead-click-reduced-timeout",
+            project.organization,
+        )
+        else 7000
```
bruno-garcia, line 406: Ideally the option would hold the value instead. So you don't need to branch based on a bool flag and have the values hard coded here.

Will make it easier to tweak the value later (say we go from 5 to 3 seconds for example)

src/sentry/replays/usecases/ingest/dom_index.py
```diff
@@ -396,6 +396,15 @@ def _handle_breadcrumb(
 ) -> ReplayActionsEventPayloadClick | None:
 
     click = None
+    project = Project.objects.get(id=project_id)
```
cmanallen, line 399: This will blow up the database.  You're running the query for every segment ingested (2k RPS) plus every breadcrumb in the recording data.  This is going to query the database anywhere from 20k times per second to 200k times per second.  Or more realistically, we won't query at that rate and will begin to backlog as the database becomes the primary choke point.  This approach won't work.

src/sentry/replays/usecases/ingest/dom_index.py
```diff
@@ -396,6 +396,15 @@ def _handle_breadcrumb(
...
+    dead_click_timeout = (
+        5000
+        if features.has(
+            "organizations:session-replay-dead-click-reduced-timeout",
+            project.organization,
```
cmanallen, line 404: Let's not use a feature flag here.  Let's set a list-type option.  In options-automator you can specify the org-ids/project-ids you want to allow to use the new timeout.  You can do the same for the timeout itself as mentioned by Josh.

cmanallen (changes requested):

src/sentry/replays/usecases/ingest/dom_index.py
cmanallen: You need two options because flag pole only returns booleans.

PR #77289, opened 2024-09-10
chore(deletions) Move RuleFireHistory to model deletions

markstory:
Ordering NotificationMessage by id when doing a bulk delete doesn't work because a message could be past retention, but also be a 'parent' for a message that is within retention.

Similarily rulefirehistory records cannot be deleted by date alone as there are foreign keys in notification message. By moving rulefirehistory deletion to model deletes, we'll be able to traverse relations and remove all notification messages related to a rulefirehistory at once.

This does create the opportunity for the parent message to be removed before the thread is 'complete'. Based on my interpretation of rule fire history a new parent message will be created on the next trigger.

Refs SENTRY-3BW5
Refs SENTRY-3E5X

### DIFF
--- a/src/sentry/deletions/__init__.py
+++ b/src/sentry/deletions/__init__.py
@@ -98,6 +98,7 @@ def load_defaults():
         RepositoryProjectPathConfig,
     )
     from sentry.models.commitfilechange import CommitFileChange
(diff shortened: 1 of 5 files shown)


ceorourke (approved):
thanks for taking care of this!

PR #77300, opened 2024-09-11
feat(workflow_engine): Add many to many relationship between `Detector` and `DataSource`

wedamija:
As per the tech spec, these models should be related to each other.


src/sentry/workflow_engine/migrations/0004_data_source_detector.py
```diff
@@ -0,0 +1,76 @@
...
+        migrations.AddField(
+            model_name="datasource",
+            name="detectors",
+            field=models.ManyToManyField(
+                through="workflow_engine.DataSourceDetector", to="workflow_engine.datasource"
+            ),
+        ),
+        migrations.AddField(
+            model_name="detector",
+            name="data_sources",
```
markstory, line 65: Why `data_sources` in one table but `datasourcedetector` in the other? Should both tables use the same inflection for `datasource`?

src/sentry/workflow_engine/models/data_source_detector.py
```diff
@@ -0,0 +1,20 @@
...
+
+from sentry.backup.scopes import RelocationScope
+from sentry.db.models import DefaultFieldsModel, FlexibleForeignKey, region_silo_model
+
+
+@region_silo_model
+class DataSourceDetector(DefaultFieldsModel):
+    __relocation_scope__ = RelocationScope.Organization
+
+    data_source = FlexibleForeignKey("workflow_engine.Datasource")
```
markstory, line 11: Sometimes you use `Datasource` and in other references it is `DataSource`. Could we be consistent with the inflection?

fixtures/backup/model_dependencies/detailed.json
```diff
@@ -6218,6 +6218,34 @@
...
+        "model": "workflow_engine.datasource",
+        "nullable": false
+      },
+      "detector": {
+        "kind": "FlexibleForeignKey",
+        "model": "workflow_engine.detector",
+        "nullable": false
+      }
+    },
+    "model": "workflow_engine.datasourcedetector",
```
saponifi3d, line 6235: do we have any patterns or nomenclature around lookup tables / is that something we should start doing? (i've seen stuff like `datasource2detector`, `datasource_detector`, and the verbose `datasource_to_detector`). not strong thoughts / opinions on this, just curious.

src/sentry/workflow_engine/migrations/0004_data_source_detector.py
saponifi3d: yeah, I like the idea of that all being consistent as well - maybe everything is `datasource` for the tables and `DataSource` for the classes?

src/sentry/workflow_engine/models/data_source_detector.py
wedamija: That's just a mistake on my part... I'm surprised this fk actually works

src/sentry/workflow_engine/migrations/0004_data_source_detector.py
wedamija: My thought was that since it's `DataSource`, the capitalization implies two words, so `data_source`. Otherwise, it should probably be `Datasource` and `datasource`

fixtures/backup/model_dependencies/detailed.json
wedamija: The usual pattern in the code base is to just mash the model names together. So `ProjectTeam` for example. https://github.com/getsentry/sentry/blob/2bfa30d041babf147c8fa28f68b737bbf13e766b/src/sentry/models/project.py#L238

src/sentry/workflow_engine/migrations/0004_data_source_detector.py
saponifi3d: 👍 makes sense. i'd be down for either one of those as well tbh, maybe `DataSource` and `data_source`? (just because of the DataPacket terminology too)

markstory: `Datasource` and `datasource` would be my vote.

wedamija: I did some googling and usually in programming contexts it's `DataSource`, although there's not a clear consensus. I'd prefer to go with the more common option here

markstory (approved):
Looks good to me.

PR #77302, opened 2024-09-11
feat(angular-onboarding): Add standalone app docs

### LINKED ISSUE #77134: Update onboarding instructions in the product to reflect the latest Angular changes
part of https://github.com/getsentry/sentry-docs/issues/8342

ArthurKnaus:
Add platform option to `javascript-angular` that allows to switch between standalone / module app initialization.

https://github.com/user-attachments/assets/99c2504e-8917-432f-950c-e4598bc7f4c4

Closes https://github.com/getsentry/sentry/issues/77134


static/app/gettingStartedDocs/javascript/angular.tsx
```diff
@@ -31,71 +32,176 @@ import {
...
+    items: [
+      {
+        label: 'Standalone',
+        value: AppType.STANDALONE,
+      },
+      {
+        label: 'Module',
+        value: AppType.MODULE,
+      },
+    ],
```
ArthurKnaus, line 53: @Lms24 what would be the proper label for it?
Can we add some postfix to `Standalone` and `Module` to make it clearer?

static/app/gettingStartedDocs/javascript/angular.tsx
```diff
@@ -31,71 +32,176 @@ import {
...
+import { bootstrapApplication } from '@angular/platform-browser';
+import * as Sentry from "@sentry/angular";
+
+import { appConfig } from './app/app.config';
+import { AppComponent } from './app/app.component';
+  `;
+
+  const appInit = isModuleApp(params)
+    ? `
+enableProdMode();
```
ArthurKnaus, line 83: @Lms24 Why are we suggesting to enable prod mode? 
According to the [Angular docs](https://v17.angular.io/api/core/enableProdMode) this is discouraged. Maybe this deserves a comment?

priscilawebdev (approved):
code looks good to me

static/app/gettingStartedDocs/javascript/angular.tsx
Lms24: I think we can just remove this from the snippets. Looks like it was recommended until [NG14](https://v14.angular.io/api/core/enableProdMode#description), given the discouragement note was added in [NG15](https://v15.angular.io/api/core/enableProdMode#description) docs. We just didn't notice that

static/app/gettingStartedDocs/javascript/angular.tsx
Lms24: With "postfix", do you mean a tootlip-like element? Then I'd add something like

- standalone: "Use this if you have an `app.config.ts` file"
- module: "Use this if you have an `app.module.ts` file"

Some other ideas for the button names:

1. App Config / App Module
2. Standalone Config / Module-based Config
3. Standalone Config / NGModule Config
4. App Config / NGModule Config
5. Standalone Conponent Config / NGModule-based Config

After thinking a bit about it, I kinda like 3 and 4. Makes it really clear that we're talking about Angular's `NGModule`s. WDYT?

Lms24: Or can we do something like

"I have an [app.config.ts]|[app.module.ts] file", so that the selection more or less becomes part of a sentence? Hope this makes sense 😅

ArthurKnaus: Was talking about changing the button label 👍 
Will go for 4 then. that is the clearest for me as someone who is not super familiar with Angular.

Lms24 (approved):
Nice, thanks a lot Arthur! LGTM (and answered your questions).
I'll follow up with making the changes in docs once this is merged.

PR #77309, opened 2024-09-11
feat(insights): Better styling for Queries landing page filters

0Calories:
Condenses the additional page filters for the queries module into a single bar

### Before
image

### After
image


DominikB2014 (approved):
lgtm, are we going to make this more generic and do this with the other modules, perhaps a components that takes in the selectors as children and condense into a single bar?

0Calories:
@DominikB2014 Hmm I didn't plan for that since I have been focused on just the queries module, but I can see this being extended it something more generic. It would probably be useful for the additional dropdown selectors in the other modules, but the queries module needed it the most because it was a whopping 3 additional page filters with a lot of spacing in between!

PR #77310, opened 2024-09-11
ref(onboarding): Move loader code to javascript getStartedDoc file

### LINKED ISSUE #77537: Move "Install using Script Loader/npm/yarn" button to the Javascript onboarding file

priscilawebdev:
This PR consolidates the onboarding buttons for both automation and manual installation. It also moves the loader script instructions to javascript.tsx, as they are only needed for the javascript platform.

**Before**

https://github.com/user-attachments/assets/0cc77254-4e81-40ce-b586-b0087f6b9f40

**After**

https://github.com/user-attachments/assets/531a6370-13e4-4655-8e28-aaca6ecb060c

requires https://github.com/getsentry/getsentry/pull/15097

closes https://github.com/getsentry/sentry/issues/77537


static/app/gettingStartedDocs/javascript/javascript.tsx
```diff
@@ -25,10 +27,41 @@ import {
...
+  AUTO = 'auto',
+  MANUAL = 'manual',
+}
+
+const platformOptions = {
+  installationMode: {
+    label: t('Installation Mode'),
+    items: [
+      {
+        label: t('Loader Script'),
```
priscilawebdev, line 48: I am wondering if we should be consistent and have `auto/manual`?

static/app/gettingStartedDocs/javascript/javascript.tsx
```diff
@@ -25,10 +27,41 @@ import {
...
+import {ProductSolution} from 'sentry/components/onboarding/productSelection';
 import replayOnboardingJsLoader from 'sentry/gettingStartedDocs/javascript/jsLoader/jsLoader';
 import {t, tct} from 'sentry/locale';
+import type {Organization} from 'sentry/types/organization';
+import type {Project, ProjectKey} from 'sentry/types/project';
+import {trackAnalytics} from 'sentry/utils/analytics';
+import {handleXhrErrorResponse} from 'sentry/utils/handleXhrErrorResponse';
 
-type Params = DocsParams;
+export enum InstallationMode {
```
priscilawebdev, line 38: let's do it step-by-step but since we are going to use this a lot, we might want to move this elsewhere

static/app/utils/useHasMounted.tsx
```diff
@@ -0,0 +1,14 @@
+import {useEffect, useRef} from 'react';
+/**
+ * Provides a boolean indicating if the component has completed its initial mount.
+ * @returns true if the component has mounted, false otherwise.
+ */
+export function useHasMounted() {
```
priscilawebdev, line 6: I decided to introduce this hook because of issues I had in tests but I am happy to push this code in a separate PR if needed

priscilawebdev:
just noticed that I have to fix the introduction message too:

Screenshot 2024-09-13 at 08 32 02

static/app/utils/useHasMounted.tsx
```diff
@@ -0,0 +1,14 @@
...
+  const hasMounted = useRef(false);
+
+  useEffect(() => {
+    hasMounted.current = true;
+  }, []);
```
vgrozdanic, line 11: Should here also be a cleanup function that, otherwise on re-renders component might appear to be mounted, when it's not:

```
return () => { hasMounted.current = false }
```

priscilawebdev: https://github.com/getsentry/sentry/pull/77469#discussion_r1758800968

static/app/gettingStartedDocs/javascript/javascript.tsx
ArthurKnaus: I would not do it in this case. For me the loader script is not an automatic setup as you still need to change your code. 🤔

static/app/components/onboarding/gettingStartedDoc/onboardingLayout.tsx
```diff
@@ -116,6 +121,36 @@ export function OnboardingLayout({
...
+    onProductSelectionChange,
+    organization,
+    projectSlug,
+    projectId,
+    platformKey,
+    projectKeyId,
+    activeProductSelection,
+    selectedOptions,
+    api,
+  ]);
```
ArthurKnaus, line 152: I've got 2 issues with this one.
1. It is called `onProductSelectionChange` but it is triggered for product selection and platform options.
2. Using a hook here is kinda destined to break at some point (see your comment), a change handler on the respective inputs would be nicer.

My suggestion:
Create 2 callbacks `onProductSelectionChange` and `onPlatformOptionsChange` and add a change handlers to the product selection and platform option controls.

static/app/components/onboarding/gettingStartedDoc/types.ts
```diff
@@ -95,6 +97,16 @@ export interface Docs<PlatformOptions extends BasePlatformOptions = BasePlatform
   customMetricsOnboarding?: OnboardingConfig<PlatformOptions>;
   feedbackOnboardingCrashApi?: OnboardingConfig<PlatformOptions>;
   feedbackOnboardingNpm?: OnboardingConfig<PlatformOptions>;
+  onProductSelectionChange?: (params: {
```
ArthurKnaus, line 100: Would move this into the `OnboardingConfig` type. Change handlers that we implement for the general onboarding will most likely not be relevant / wanted in e.g. a sidebar doc.

static/app/gettingStartedDocs/javascript/javascript.tsx
```diff
@@ -111,43 +150,151 @@ const getInstallConfig = () => [
...
+              'In this quick guide you’ll use [strong:npm] or [strong:yarn] to set up:',
+              {
+                strong: <strong />,
+              }
+            )}
+      </TextBlock>
+    </div>
+  ),
+  install: params =>
+    isAutoInstall(params)
```
ArthurKnaus, line 234: I think we could improve a bit on the general structure here.
e.g. have two 3 configuration objects.
```ts
const loaderScriptOnboarding: {
  // ...
}

const packageManagerOnboarding: {
  // ...
}

const onboarding: {
  install:  isAutoInstall(params)
      ? loaderScriptOnboarding.install.(params)
      : packageManagerOnboarding.install.(params)
  // ...
}
```

static/app/utils/useHasMounted.tsx
ArthurKnaus: I think we can delete this in favour of `useIsMountedRef`?

ArthurKnaus (commented):
There is a lot going on in this PR. Can we split it into two?
e.g.:
1. Introduce `onChange` callbacks for product selection and platform options.
2. Merge loader script docs into `javascript.tsx`

priscilawebdev:
@ArthurKnaus the hook was deleted yesterday https://github.com/getsentry/sentry/pull/77310/commits/d30c1a00cbed14def8e15b1af83674d87e240c0e

static/app/gettingStartedDocs/javascript/javascript.tsx
```diff
@@ -149,6 +354,70 @@ const onboarding: OnboardingConfig = {
       guideLink: 'https://docs.sentry.io/platforms/javascript/sourcemaps/',
     }),
   ],
+  verify: getVerifyConfig,
```
priscilawebdev, line 357: at the moment the verify codes for the loader and package manager are different, but we can have the same for both.

static/app/gettingStartedDocs/javascript/javascript.tsx
```diff
@@ -112,8 +146,179 @@ const getInstallConfig = () => [
...
+      ),
+      link: 'https://docs.sentry.io/platforms/javascript/tracing/',
+    },
+    {
+      id: 'session-replay',
+      name: t('Session Replay'),
+      description: t(
+        'Get to the root cause of an error or latency issue faster by seeing all the technical details related to that issue in one visual replay on your web application.'
+      ),
+      link: 'https://docs.sentry.io/platforms/javascript/session-replay/',
```
priscilawebdev, line 269: In the past, we only displayed the next step for 'tracing' if this product was not selected. However, we're no longer following this approach across many platforms. Now, we can simplify the process by displaying all available options. I would even suggest unifying the next steps for both the loader and package manager. 

Feedback welcome.

static/app/gettingStartedDocs/javascript/jsLoader/updateDynamicSdkLoaderOptions.ts
```diff
@@ -0,0 +1,50 @@
+import type {Client} from 'sentry/api';
+import {ProductSolution} from 'sentry/components/onboarding/productSelection';
+import {t} from 'sentry/locale';
+import type {Organization} from 'sentry/types/organization';
+import type {Project, ProjectKey} from 'sentry/types/project';
+import {handleXhrErrorResponse} from 'sentry/utils/handleXhrErrorResponse';
+
+export async function updateDynamicSdkLoaderOptions({
```
priscilawebdev, line 8: I would love to make this a hook but we are using it conditionally

static/app/gettingStartedDocs/javascript/javascript.tsx
priscilawebdev: that is a good idea! I implemented that 🙂

priscilawebdev (commented):
.

static/app/gettingStartedDocs/javascript/javascript.tsx
matejminar: I think some next steps only make sense if the product was not selected.

I would be curious, in general, how often these are clicked. Would it be possible to add some instrumentation in a follow-up PR? Then we can decide if these are actually useful or just a distraction at that point in time.

static/app/components/onboarding/productSelection.tsx
```diff
@@ -405,22 +387,19 @@ export function ProductSelection({
   // until we improve multi snippet suppport
   const showPackageManagerInfo =
     (platform?.indexOf('javascript') === 0 || platform?.indexOf('node') === 0) &&
-    platform !== 'javascript-astro';
+    platform !== 'javascript-astro' &&
+    platform !== 'javascript';
```
mydea, line 391: just wondering, why are we excluding base JS here? 🤔

ArthurKnaus: It is now defined in `onboarding.introduction` inside `javascript.tsx`.
@priscilawebdev I think we should do this also for the rest of the javascript platforms and delete the logic here as a follow-up. wdyt? I can create a PR once this one is merged.

priscilawebdev: yes I agree... I started with [Astro](https://github.com/getsentry/sentry/pull/77702) and was waiting to get feedback there before doing this for all js platforms

mydea:
My 2 cents: this is a good change, the new UI is better than before 👏 only small feedback, IMHO the active state of the toggle is not very clear, it's a rather subtle difference. other than this (or even with this), :shipit:

static/app/views/onboarding/setupDocs.tsx
```diff
@@ -66,13 +65,6 @@ function SetupDocs({location, recentCreatedProject: project}: StepProps) {
                   projectSlug={project.slug}
                   platform={currentPlatform.name}
                 />
-              ) : params.showLoader ? (
```
ArthurKnaus, line 69: We can now remove `showLoader` from `useOnboardingQueryParams`, right?

ArthurKnaus (approved):
Great refactoring! 🙌

static/app/views/onboarding/setupDocs.tsx
priscilawebdev: thank you! yes it is done here https://github.com/getsentry/sentry/pull/77699

static/app/components/onboarding/gettingStartedDoc/onboardingLayout.tsx
```diff
@@ -73,6 +73,7 @@ export function OnboardingLayout({
     nextSteps,
     onPlatformOptionsChange,
     onProductSelectionChange,
+    onProductSelectionLoad,
```
priscilawebdev, line 76:
```suggestion
```

PR #77316, opened 2024-09-11
fix(replay): Node highlight not being removed

c298lee:
The node highlight wasn't being removed when there are multiple of the same`nodeId`. When adding the highlight, we end up drawing multiple of the same canvases. To fix this, a set is used to return unique values, so we only highlight each node once.

Before:

https://github.com/user-attachments/assets/896e3889-31ab-41d5-909f-8b75f0e83444

After:

https://github.com/user-attachments/assets/7e84e27e-9179-4be9-b675-77d82f8f5d40


ryan953 (approved):
nice!

PR #77317, opened 2024-09-11
feat(custom-views): Add tabs to a menu when they overflow

malwilley:
When there are too many tabs to display in the tab bar, removes them and appends to an overflow menu.

This does not yet implement pulling the selected tab out of the overflow menu, which can be done in a separate PR. For now, selected tabs will stay in the overflow menu.

Unfortunately these changes are very dependent on layout, so there isn't a good way to write tests for them without mocking everything.

https://github.com/user-attachments/assets/02d32beb-81d5-49d7-8947-d33a8131c3f5

### DIFF
--- a/static/app/components/draggableTabs/draggableTabList.tsx
+++ b/static/app/components/draggableTabs/draggableTabList.tsx
@@ -1,27 +1,237 @@
-import {Fragment, useContext, useEffect, useMemo, useRef} from 'react';
+import {
+  type Dispatch,
+  Fragment,
+  type SetStateAction,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 import styled from '@emotion/styled';
 import type {AriaTabListOptions} from '@react-aria/tabs';
 import {useTabList} from '@react-aria/tabs';
 import {useCollection} from '@react-stately/collections';
 import {ListCollection} from '@react-stately/list';
-import type {TabListStateOptions} from '@react-stately/tabs';
+import type {TabListState, TabListStateOptions} from '@react-stately/tabs';
 import {useTabListState} from '@react-stately/tabs';
 import type {Node} from '@react-types/shared';
 import {motion, Reorder} from 'framer-motion';
 
 import {Button} from 'sentry/components/button';
+import {CompactSelect} from 'sentry/components/compactSelect';
+import DropdownButton from 'sentry/components/dropdownButton';
 import {TabsContext} from 'sentry/components/tabs';
 import {type BaseTabProps, Tab} from 'sentry/components/tabs/tab';
-import {IconAdd} from 'sentry/icons';
+import {IconAdd, IconEllipsis} from 'sentry/icons';
 import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
+import {defined} from 'sentry/utils';
 import {browserHistory} from 'sentry/utils/browserHistory';
+import {useDimensions} from 'sentry/utils/useDimensions';
+import {useDimensionsMultiple} from 'sentry/utils/useDimensionsMultiple';
 
 import type {DraggableTabListItemProps} from './item';
 import {Item} from './item';
 
 interface BaseDraggableTabListProps extends DraggableTabListProps {
   items: DraggableTabListItemProps[];
+}
+
+function useIsDragging(ref: React.RefObject<HTMLElement>) {
+  const [isDragging, setIsDragging] = useState(false);
+
+  useEffect(() => {
+    const element = ref.current;
+
+    if (!element) {
+      return () => {};
+    }
+
+    const handlePointerDown = () => {
+      setIsDragging(true);
+    };
+    const handlePointerUp = () => {
+      setIsDragging(false);
+    };
+    element.addEventListener('pointerdown', handlePointerDown);
+    document.addEventListener('pointerup', handlePointerUp);
+
+    return () => {
+      element.removeEventListener('pointerdown', handlePointerDown);
+      document.removeEventListener('pointerup', handlePointerUp);
+    };
+  }, [ref]);
+
+  return isDragging;
+}
+
+function useOverflowingTabs({state}: {state: TabListState<DraggableTabListItemProps>}) {
+  const persistentTabs = [...state.collection].filter(
+    item => item.key !== 'temporary-tab'
+  );
+  const outerRef = useRef<HTMLDivElement>(null);
+  const addViewTempTabRef = useRef<HTMLDivElement>(null);
+  const [tabElements, setTabElements] = useState<Array<HTMLDivElement | null>>([]);
+  const {width: outerWidth} = useDimensions({elementRef: outerRef});
+  const {width: addViewTempTabWidth} = useDimensions({elementRef: addViewTempTabRef});
+  const tabsDimensions = useDimensionsMultiple({elements: tabElements});
+
+  const overflowingTabs = useMemo(() => {
+    const availableWidth = outerWidth - addViewTempTabWidth;
+
+    let totalWidth = 0;
+    const overflowing: Node<DraggableTabListItemProps>[] = [];
+
+    for (let i = 0; i < tabsDimensions.length; i++) {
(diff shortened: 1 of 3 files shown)


static/app/utils/useDimensionsMultiple.tsx
```diff
@@ -0,0 +1,45 @@
+import {useLayoutEffect, useRef, useState} from 'react';
+
+interface Props<Element extends HTMLElement> {
+  elements: Array<Element | null>;
+}
+
+/**
+ * Measures the dimensions of multiple elements at once.
+ */
+export function useDimensionsMultiple<Element extends HTMLElement>({
```
malwilley, line 10: Added this hook because we want to keep track of the width of each tab

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -24,6 +38,202 @@ interface BaseDraggableTabListProps extends DraggableTabListProps {
...
+    return () => {
+      element.removeEventListener('pointerdown', handlePointerDown);
+      document.removeEventListener('pointerup', handlePointerUp);
+    };
+  }, [ref]);
+
+  return isDragging;
+}
+
+function useOverflowingTabs({state}: {state: TabListState<DraggableTabListItemProps>}) {
```
malwilley, line 69: This hook determines which tabs are overflowing. We can't use `IntersectionObserver` like the current Tabs component, because we will need to move the tabs around based on which one is selected. It's also different because the "Add View" button is to the right of the tabs, so we can't used a static container for the IntersectionObserver.

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -24,6 +38,202 @@ interface BaseDraggableTabListProps extends DraggableTabListProps {
...
+  const isDragging = useIsDragging(tabListRef);
+
+  // Only apply this while dragging, because it causes tabs to stay within the container
+  // which we do not want (we hide tabs once they overflow
+  const dragConstraints = isDragging ? tabListRef : undefined;
+
+  return (
+    <TabListWrap {...tabListProps} className={className} ref={tabListRef}>
+      <ReorderGroup axis="x" values={values} onReorder={onReorder} as="div">
+        {tabs.map((item, i) => (
```
MichaelSun48, line 183: I noticed that tab's that are calculated as overflowing have opacity=0 applied to them, which might be the reason why the `+ Add View` button isn't snapping to the new edge when the tab disappears

image

Would removing tabs from the DOM if they're overflowing solve this?

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -24,6 +38,202 @@ interface BaseDraggableTabListProps extends DraggableTabListProps {
...
+  onChange?: (key: string | number) => void;
+  tabVariant?: BaseTabProps['variant'];
+  value?: string | number;
+}) {
+  const tabListRef = useRef<HTMLUListElement>(null);
+  const {tabListProps} = useTabList({orientation, ...ariaProps}, state, tabListRef);
+
+  const values = useMemo(() => [...state.collection], [state.collection]);
+
+  const isDragging = useIsDragging(tabListRef);
```
MichaelSun48, line 174: nit: rather than creating this hook, I think you could use Framer motion's `onDragStart` and `onDragEnd` on each tab to update an `isDragging` state

static/app/components/draggableTabs/draggableTabList.tsx
malwilley: The problem is that we need to render the tabs to know how big they are. Otherwise we won't know when they are able to fit in the available space. I'm sure there is a way to get this behavior, but it would certainly complicate the solution

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -24,6 +38,202 @@ interface BaseDraggableTabListProps extends DraggableTabListProps {
...
+              dragElastic={0} // Prevents the tab from being dragged outside of the dragConstraints (w/o this you can drag it outside but it'll spring back)
+              dragTransition={{bounceStiffness: 400, bounceDamping: 40}} // Recovers spring behavior thats lost when using dragElastic=0
+              transition={{delay: -0.1}} // Skips the first few frames of the animation that make the tab appear to shrink before growing
+              layout
+            >
+              <div
+                key={item.key}
+                style={
+                  overflowingTabs.some(tab => tab.key === item.key) ? {opacity: 0} : {}
+                }
```
MichaelSun48, line 211: Just curious - why can't we apply this style directly to the child Tab component?

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -24,6 +38,202 @@ interface BaseDraggableTabListProps extends DraggableTabListProps {
...
+}: {
+  overflowTabs: Node<DraggableTabListItemProps>[];
+  state: TabListState<any>;
+}) {
+  const options = useMemo(() => {
+    return overflowTabs.map(tab => {
+      return {
+        value: tab.key,
+        label: tab.textValue,
+        textValue: tab.textValue,
```
MichaelSun48, line 117: Should this also include the tab menu button for selected, overflowed tabs? Right now there's no way to access that functionality for overflowing tabs other than by increasing screen width.

Side note: perhaps we should talk to Vu about what this should look like? I don't think there are any mockups for this

static/app/components/draggableTabs/draggableTabList.tsx
malwilley: You're totally right! That's much better

static/app/components/draggableTabs/draggableTabList.tsx
malwilley: Ah good catch, this style wasn't necessary. Left from when I was testing something earlier.

static/app/components/draggableTabs/draggableTabList.tsx
malwilley: I don't have it implemented in this PR, but I thought the intended behavior was that selecting a tab in the overflow menu would pop it out into the tab bar. That way we won't need to add those controls to the overflow. But yeah we never got any designs on the overflow menu specifically

MichaelSun48: Gotcha, my concern with that would be that the tab ordering might get a little confusing. Would the tab replace the tab in the right-most tab that's not overflowing? I'll check in with Vu tomorrow to see what his thoughts are, but also agreed that this can be part of a different PR.

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -178,12 +178,7 @@ function Tabs({
               onDrag={() => setIsDragging(true)}
               onDragEnd={() => setIsDragging(false)}
             >
-              <div
-                key={item.key}
-                style={
-                  overflowingTabs.some(tab => tab.key === item.key) ? {opacity: 0} : {}
-                }
-              >
+              <div key={item.key}>
```
MichaelSun48, line 181: Could we get rid of this div at this point?

MichaelSun48 (approved):
I owe you big time 🫡

MichaelSun48:
Something I discovered while playing around – if you set `position: absolute` on the tab when it's overflowing (in addition to `opacity: 0`, [here](https://github.com/getsentry/sentry/blob/3e0482c167a02a103b64f3313abf9a1cc58e8d15/static/app/components/tabs/tab.tsx#L249)), it looks like the Add View button does want to snap to the left, but this also triggers some weird infinite rendering loop. Maybe if we could find a way to terminate the loop it would achieve the desired behavior? 

https://github.com/user-attachments/assets/7fd0fda4-719b-43ec-94f8-0e303ec9d456

malwilley:
@MichaelSun48 Hmm you might be on to something. Going to merge this for now, but we can follow up on trying to get that to work eventually. Though IMO the add view button sticking to the right when there is overflow isn't that bad (could even be considered expected behavior) so I wouldn't spend _too_ much time trying to fix it

PR #77322, opened 2024-09-11
:recycle: ref(slack): fix slack webhook actions typing

iamrajjoshi:
fixing the typing for Slack webhook actions. i also removed some old non slack blockkit logic since we have GA'ed for a while and won't have to deal with interactions with the old messages.


src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -700,18 +727,20 @@ def _handle_group_actions(
             return self.respond()
 
         response_url = slack_request.data["response_url"]
-        json_blocks = orjson.dumps(response.get("blocks")).decode()
         webhook_client = WebhookClient(response_url)
         try:
             webhook_client.send(
-                blocks=json_blocks,
+                blocks=response.get("blocks"),
```
iamrajjoshi, line 733: don't need to do this, originally that orjson line only converted from single quotes to double quotes (i am assume this was because in pre-slack_sdk world, we needed to. the slack_sdk handles this for us, i verified this with my local setup

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -758,44 +787,42 @@ def get_action_option(cls, slack_request: SlackActionRequest) -> str | None:
         return action_option
 
     @classmethod
-    def get_action_list(cls, slack_request: SlackActionRequest) -> list[MessageAction]:
+    def get_action_list(cls, slack_request: SlackActionRequest) -> list[BlockKitMessageAction]:
         action_data = slack_request.data.get("actions")
-        if action_data:
```
iamrajjoshi, line 763: don't need this since we have ga'ed blockkit for months now and won't be dealing with non-blockkit interactions.

src/sentry/integrations/slack/message_builder/issues.py
```diff
@@ -421,7 +423,7 @@ def __init__(
         event: Event | GroupEvent | None = None,
         tags: set[str] | None = None,
         identity: RpcIdentity | None = None,
-        actions: Sequence[MessageAction] | None = None,
+        actions: Sequence[MessageAction | BlockKitMessageAction] | None = None,
```
cathteng, line 426: what is the difference between `MessageAction` and `BlockKitMessageAction`? should we always be using one over the other if we don't support non-block kit anymore?

iamrajjoshi: ideally we should delete `MessageAction`, but i saw some places in our logic where we create a `MessageAction` payload to send to slack, so i want to slowly remove it so make sure we don't break changes since ff this is hard

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -240,11 +247,20 @@ def validation_error(
             },
         )
 
-        text: str = list(*error.detail.values())[0]
+        text: str = ""
+        if isinstance(error.detail, dict):
+            text = list(*error.detail.values())[0]
+        else:
+            text = str(error.detail[0])
```
RyanSkonnord, line 254: Technically the type definition for `error.detail` allows lists and/or dictionaries to be nested recursively. It might be more correct to do something like the helper function I sketched out [here](https://github.com/getsentry/sentry/pull/77444/files#diff-30aac95a95c4c9b782c0eb9a15b9d363724bc3c7a0179e1f7b65be80fbe7c7c4R246-R255).

src/sentry/integrations/slack/message_builder/issues.py
RyanSkonnord: Yeah, as I mentioned [here](https://github.com/getsentry/sentry/pull/77444#discussion_r1757607830) it seems like `BlockKitMessageAction` ought to extend `MessageAction` (assuming we need both).

The "TODO" comment on `MessageAction` suggests that its bottom four attributes (including `block_id`, which is confusing) ought to belong to some other subclass. But I agree with delaying that sort of refactoring for later.

I'm okay with using `MessageAction | BlockKitMessageAction` for type hints temporarily, but would consider changing `BlockKitMessageAction` to extend `MessageAction` instead.

RyanSkonnord (approved):
👏

PR #77323, opened 2024-09-11
ref(issues): Pass replay reader result around as a prop

scttcper:
Removes an unused component.

Passes the result of useReplayReader as a specific prop instead of spreading the various properties into props


ryan953 (approved):
tx!

PR #77325, opened 2024-09-11
ref(replay): try 5s rage/dead click timeout for select orgs, using options

aliu39:
Replaces https://github.com/getsentry/sentry/pull/77286 and https://github.com/getsentry/sentry/pull/77284

Depends on https://github.com/getsentry/sentry/pull/77365


JoshFerge (approved):
didn't mean to say that you should have replaced the feature flag -- that part was totally fine. just meant to say you could use an option to control the rage/deadclick timeout

aliu39:
> didn't mean to say that you should have replaced the feature flag -- that part was totally fine. just meant to say you could use an option to control the rage/deadclick timeout

True! This is all backend tho so we can just do 1 PR w/2 options

src/sentry/replays/usecases/ingest/dom_index.py
```diff
@@ -412,28 +415,38 @@ def _handle_breadcrumb(
...
-            is_rage = (
-                payload["data"].get("clickCount", 0) or payload["data"].get("clickcount", 0)
-            ) >= 5
-            click = create_click_event(
-                payload, replay_id, is_dead=True, is_rage=is_rage, project_id=project_id
+        if is_timeout_reason and is_target_tagname:
+            organization = Project.objects.get(id=project_id).organization
+            dead_click_timeout = (
+                reduced_dead_click_timeout
+                if organization.slug in options.get("feedback.dead-click.reduced-timeout-org-list")
```
cmanallen, line 422: Let's not use the org slug because we need to query for it.  We have the org_id and the project_id.  We could use those with the option directly so you don't need to query anything.

cmanallen (changes requested):

PR #77331, opened 2024-09-11
update incident activity string for dynamic alerts

nhsiehgit:
NOTE:

Comparison detection types are labeled "percent" currently
Crash Free rates are 'percent' (or may become 'percent' soon) because the values they are comparing are _percents_ - however we are _not_ comparing a _percent change_. instead Crash free rates are comparing static values that happen to be "percent values"

so for now - we check whether there's a comparison delta and whether the type is dynamic or not

Screenshot 2024-09-12 at 4 22 28 PM

### DIFF
--- a/static/app/views/alerts/rules/metric/details/metricActivity.tsx
+++ b/static/app/views/alerts/rules/metric/details/metricActivity.tsx
@@ -116,21 +116,25 @@ function MetricAlertActivity({organization, incident}: MetricAlertActivityProps)
         </Link>
       </Cell>
       <Cell>
-        {incident.alertRule.comparisonDelta ? (
-          <Fragment>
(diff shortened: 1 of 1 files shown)


mifu67:
Do you have a screenshot?

ceorourke:
Does this rely on https://getsentry.atlassian.net/browse/ALRT-262 ?

nhsiehgit:
@ceorourke Nope!

I updated the if statements to refer to the comparison rate AND type == 'dynamic' rather than solely relying on type here.

IF we determine that we want to rebrand "percent" to "percent_change" or something, then we can update this to reference _only_ the detectionType

mifu67 (approved):

PR #77332, opened 2024-09-11
fix(insights): move pagination for web vitals

### LINKED ISSUE #76065: Move web vitals page controls
The controls to change pages on web vitals landing page should be under the table so that it's consistent with every other area of our application. Currently they live at the top, to right of the search bar.

**Current State**
Image

**Future state**
Image

DominikB2014:
closes https://github.com/getsentry/sentry/issues/76065

1. Move pagination to bottom of page within in web vitals module, this is consistent with every other table in insights (and perhaps sentry as a whole)
2. Also makes the spacing above/below the search bar, consistent with queries and requests module.

Before
image

After
image


edwardgou-sentry (approved):
Thanks!

PR #77335, opened 2024-09-11
fix(alerts): properly expose validation errors

mifu67:
The logic as it was was leading to some validation errors being exposed as undefined. Open to better solutions to check for timeout errors (which are something like `[{"message": "timeout occurred", "status": 408}]`.

### DIFF
--- a/static/app/views/alerts/rules/metric/ruleForm.tsx
+++ b/static/app/views/alerts/rules/metric/ruleForm.tsx
@@ -842,7 +842,7 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
             : Object.values(err?.responseJSON)
           : [];
         let apiErrors = '';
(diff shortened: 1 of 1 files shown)


nhsiehgit (approved):

static/app/views/alerts/rules/metric/ruleForm.tsx
```diff
@@ -842,7 +842,7 @@ class RuleFormContainer extends DeprecatedAsyncComponent<Props, State> {
             : Object.values(err?.responseJSON)
           : [];
         let apiErrors = '';
-        if (typeof errors[0] === 'object') {
+        if (typeof errors[0] === 'object' && !Array.isArray(errors[0])) {
```
nhsiehgit, line 845: Oh

maybe `isinstance(errors[0], list)`?

PR #77336, opened 2024-09-11
fix(dashboards): Adjust decimal points for percentage on dashboard line graph widgets

### LINKED ISSUE #57352: Dashboard Widget - Percentage does not show on Y-axis when below 1.0%
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

Dashboard widgets that work with values lower than 1.0% like for instance 0.15 don't cause the Y-axis to rescale so it stays on 0%
Screenshot 2023-10-03 at 14 21 02

### Expected Result

I would expect the Y axis to scale automatically

### Actual Result

Screenshot 2023-10-03 at 14 21 02

### Product Area

Dashboards

### Link

_No response_

### DSN

_No response_

### Version

_No response_

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2069) by [Unito](https://www.unito.io)

nikkikapadia:
Fixes [this issue](https://github.com/getsentry/sentry/issues/57352) where percentage does not show on line charts y axes when the max percentage is <1.0%. This was fixed by increasing the amount of decimal places we are able to show for percentage values on the y axis of line charts.

Before: 
image

After:
image


static/app/utils/discover/charts.tsx
```diff
@@ -85,14 +87,15 @@ export function axisLabelFormatterUsingAggregateOutputType(
   type: string,
   abbreviation: boolean = false,
   durationUnit?: number,
-  rateUnit?: RateUnit
+  rateUnit?: RateUnit,
+  decimalPlaces?: number
```
edwardgou-sentry, line 91:
```suggestion
  decimalPlaces: number = 0
```
nit: You can default this parameter to 0 so you don't have to check for undefined below

edwardgou-sentry (approved):

PR #77340, opened 2024-09-11
ref(issue-details): Move Issues tabs to data sections

leeandher:
Converts the tabs to folding data sections. They are rendered as is, and we can introduce new designs if we have more time later on, but for now, styling and text sizes are the same.

image

image

image

**todo**
- [x] Add tests for their new location
- [ ] Figure out subrouting so directly linking to `/merged` and `/similar` does something expected

### DIFF
--- a/static/app/views/issueDetails/streamline/eventDetails.tsx
+++ b/static/app/views/issueDetails/streamline/eventDetails.tsx
@@ -28,8 +28,8 @@ import {
   EventSearch,
   useEventQuery,
 } from 'sentry/views/issueDetails/streamline/eventSearch';
-import {useFetchEventStats} from 'sentry/views/issueDetails/streamline/useFetchEventStats';
 import {IssueContent} from 'sentry/views/issueDetails/streamline/issueContent';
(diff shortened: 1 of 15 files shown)


static/app/views/issueDetails/streamline/eventDetails.tsx
```diff
@@ -153,14 +159,16 @@ const GraphAlert = styled(Alert)`
...
-const GroupContent = styled('div')<{navHeight?: number}>`
+const ExtraContent = styled('div')`
   border: 1px solid ${p => p.theme.translucentBorder};
   background: ${p => p.theme.background};
   border-radius: ${p => p.theme.borderRadius};
-  position: relative;
 `;
 
-const GroupContentPadding = styled('div')`
+const GroupContent = styled(ExtraContent)<{navHeight?: number}>`
```
scttcper, line 168: doesn't look used here
```suggestion
const GroupContent = styled(ExtraContent)`
```

scttcper (approved):

PR #77341, opened 2024-09-11
fix(replay): Move timeline grid-column to style tag

scttcper:
Since each dot is in its own column, this change prevents inserting an emotion style for each dot. I saw this when i was creating the issues timeline and never got around to fixing it.


michellewzhang (approved):

ryan953 (approved):

ryan953:
Thanks! I had wanted to keep going around looking for dynamic styled instances, we could probably be systematic and kill them all if we identified it as a problem area 🤔

PR #77343, opened 2024-09-11
fix(insights): Fixes alerts serializer failing for some insights functions with special arguments

edwardgou-sentry:
Some Insights functions have use special arguments that do not pass the existing column argument validation. For example `http_response_rate(3)` fails because `3` is not a valid measurement, supported column, or mri, etc.

This change adds a mapping of Insights functions and their arguments for the alerts serializer to check against, so that these Insights functions no longer fail when valid.


DominikB2014 (approved):
lgtm!

PR #77349, opened 2024-09-11
fix(insights): Queries module page filters did not include `span.system`

0Calories:
Fixes an issue where the action (command) and domain (table) selectors in the DB module were not including the currently selected database system. With this change, the options for these page filters will only show DB commands and tables that are valid for the currently selected database system.

### DIFF
--- a/static/app/views/insights/common/views/spans/selectors/actionSelector.tsx
+++ b/static/app/views/insights/common/views/spans/selectors/actionSelector.tsx
@@ -21,16 +21,23 @@ const {SPAN_ACTION} = SpanMetricsField;
 
 type Props = {
   moduleName: ModuleName;
+  filters?: Record<string, string>;
   spanCategory?: string;
   value?: string;
 };
 
-export function ActionSelector({value = '', moduleName, spanCategory}: Props) {
+export function ActionSelector({value = '', moduleName, spanCategory, filters}: Props) {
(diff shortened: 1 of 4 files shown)


static/app/views/insights/common/views/spans/selectors/actionSelector.tsx
```diff
@@ -21,17 +21,24 @@ const {SPAN_ACTION} = SpanMetricsField;
...
   // TODO: This only returns the top 25 actions. It should either load them all, or paginate, or allow searching
   //
   const location = useLocation();
   const organization = useOrganization();
   const eventView = getEventView(location, moduleName, spanCategory);
 
+  if (filters) {
+    Object.keys(filters).forEach(key =>
+      eventView.additionalConditions.addFilterValue(key, filters[key])
+    );
```
mjq, line 39: nit:
```suggestion
    Object.entries(filters).forEach(([key, val]) =>
      eventView.additionalConditions.addFilterValue(key, val)
    );
```

mjq (approved):
Nice!

static/app/views/insights/common/views/spans/selectors/actionSelector.tsx
0Calories: Agreed!

PR #77352, opened 2024-09-11
ref(feedback): add user.email tag to issue platform event

### LINKED ISSUE #77333: [Alerts Configuration] Condition to show multiple tags does not work as expected
## Problem 

For a User Feedback Alert (Backed on Issues Platform), I have set up a Slack alert to render both 'url' and 'user.email' tags.
Both of these tags exist on feedback events. However, only 'url' shows up. I have noticed that the `user.email` tag does not show up in the Tags list on User Feedback Details despite being indexed/searchable. This could be part of the reason why it does not render on the Slack alert. 

[Specific Alerts Configuration Page](https://sentry.sentry.io/alerts/rules/javascript/14756941/)

michellewzhang:
fixes https://github.com/getsentry/sentry/issues/77333

- adds `user.email` as a tag on the `event`, since it wasn't there before. context for this change is we want to be able to display `user.email` on the feedback UI as a tag, as well as be able to write alert conditions on it


JoshFerge (approved):
i'm curious why we need to do this, as i would have guessed that the context user would have populated everything we need.

michellewzhang:
> i'm curious why we need to do this, as i would have guessed that the context user would have populated everything we need.

we do have access to the user email through various other means, but i think the main reason we want to add it to tags specifically is because @jas-kas wanted to create alert rules using a tag called `user.email` (which exists on issues but not feedbacks for some reason).

aliu39:
Yup, the alert rules are based on custom tags, not contexts. Also, this will make user.email show up in the tags section of feedback or issue details.

src/sentry/feedback/usecases/create_feedback.py
```diff
@@ -142,8 +142,14 @@ def fix_for_issue_platform(event_data):
...
     feedback_obj = event_data.get("contexts", {}).get("feedback", {})
     contact_email = feedback_obj.get("contact_email")
-    if not ret_event["user"].get("email", ""):
-        ret_event["user"]["email"] = contact_email
+
+    if contact_email:
+        if not ret_event["user"].get("email", ""):
+            ret_event["user"]["email"] = contact_email
+
+        # Set the user.email tag
```
aliu39, line 150: can we mention in the comment why we're doing this manually

michellewzhang: yup

JoshFerge:
got it. i wonder if in the snuba processor we convert the contexts to tags, but maybe in the issues platform processer we don't. this change makes sense!

src/sentry/feedback/usecases/create_feedback.py
```diff
@@ -364,9 +371,12 @@ def shim_to_feedback(
             feedback_event["platform"] = event.platform
             feedback_event["level"] = event.data["level"]
             feedback_event["environment"] = event.get_environment().name
-            feedback_event["tags"] = [list(item) for item in event.tags]
 
-        else:
+            tags = {}
```
michellewzhang, line 375: convert the tags to a `dict` because it was causing issues in the shim pipeline

src/sentry/feedback/usecases/create_feedback.py
```diff
@@ -360,22 +367,22 @@ def shim_to_feedback(
...
+
             metrics.incr(
                 "feedback.user_report.missing_event",
                 sample_rate=1.0,
                 tags={"referrer": source.value},
             )
 
             feedback_event["timestamp"] = datetime.utcnow().timestamp()
-            feedback_event["platform"] = "other"
-            feedback_event["level"] = report.get("level", "info")
```
michellewzhang, line 378: `level` was being set 3 times

src/sentry/feedback/usecases/create_feedback.py
```diff
@@ -360,22 +367,22 @@ def shim_to_feedback(
...
             feedback_event["timestamp"] = event.datetime.timestamp()
-            feedback_event["level"] = event.data["level"]
             feedback_event["platform"] = event.platform
             feedback_event["level"] = event.data["level"]
             feedback_event["environment"] = event.get_environment().name
-            feedback_event["tags"] = [list(item) for item in event.tags]
 
-        else:
+            tags = {}
+            for [k, v] in event.tags:
```
aliu39, line 375: Should check if event.tags is a list or a dict. I think the two formats are
[[k, v], [k, v]]
and 
{k:v, k:v}

aliu39: Actually I don't think we need this conversion here, just in `fix_for_issue_platform`.

michellewzhang: i think we only need to convert the tags from list to dict in the shim part, i think tags are always dicts in the fix_for_issue_platform

aliu39: Here: https://github.com/getsentry/sentry/blob/2b87f2a655b22fd8a22722b13905f195183264dc/src/sentry/feedback/usecases/create_feedback.py#L253-L253

src/sentry/feedback/usecases/create_feedback.py
aliu39: lets do the level and platform changes in a different PR

src/sentry/feedback/usecases/create_feedback.py
aliu39: We don't have a guarantee the event passed to `fix_for_issue_platform` has a tag dict, or a tag list. So I think it'd definitely help to force it into a dict in that fx. In `shim_to_feedback` it's kinda optional (can tweak some tests if needed).

src/sentry/feedback/usecases/create_feedback.py
```diff
@@ -142,8 +142,15 @@ def fix_for_issue_platform(event_data):
     # If no user email was provided specify the contact-email as the user-email.
     feedback_obj = event_data.get("contexts", {}).get("feedback", {})
     contact_email = feedback_obj.get("contact_email")
-    if not ret_event["user"].get("email", ""):
-        ret_event["user"]["email"] = contact_email
+
+    if contact_email:
```
aliu39, line 146:
```suggestion
```
Sorry I know I suggested this condition but I think we should remove it, None is allowed here and is potentially useful for the frontend. For ex if we wanted to filter on/search for feedbacks w/no email.

src/sentry/feedback/usecases/create_feedback.py
michellewzhang: the level change i have to make now bc it broke a test, is it worth separating out the platform one if it's just one line?

src/sentry/feedback/usecases/create_feedback.py
michellewzhang: should we still keep the condition for the tag though? doesn't make sense to have `None` as a tag

src/sentry/feedback/usecases/create_feedback.py
michellewzhang: if we don't know if it's even a list, how can we turn it into a dict?

src/sentry/feedback/usecases/create_feedback.py
aliu39: event.schema.json says null values are allowed

src/sentry/feedback/usecases/create_feedback.py
aliu39: You can use `isinstance(tags, list)`

src/sentry/feedback/usecases/create_feedback.py
michellewzhang: okay removed the check completely

src/sentry/feedback/usecases/create_feedback.py
michellewzhang: my question is, if a string or something garbage is passed in how are we supposed to turn that into a dict?

michellewzhang: or are you saying the two possible types are _only_ list or dict?

aliu39: yeah, those are the 2 possibilities

michellewzhang: okay it should be modified!

src/sentry/feedback/usecases/create_feedback.py
aliu39: We don't to force this conversion anymore after doing it in create_feedback_issue. The shim tests will pass since they check the kafka call, not the shim fx itself.

aliu39:
Looks good!

src/sentry/feedback/usecases/create_feedback.py
```diff
@@ -251,10 +257,20 @@ def create_feedback_issue(event, project_id: int, source: FeedbackCreationSource
     )
     now = datetime.now()
 
+    # Force `tags` to be a dict if it's initially a list,
```
aliu39, line 260: can we move this to inside `fix_for_issue_platform` before L149, so it's colocated w/ the code that requires a tags dict?

Also the callstack is shim -> create_feedback -> fix_for_.. so if we do it at the end, all 3 of these functions can accept either tag format.

src/sentry/feedback/usecases/create_feedback.py
aliu39: Oh ok, no they look fine to keep then

aliu39:
Looks good other than suggestions above! Sorry for all the nits, just wanted to be sure we safely handle both tag formats in all fxs

src/sentry/feedback/usecases/create_feedback.py
```diff
@@ -360,22 +377,18 @@ def shim_to_feedback(
                 feedback_event["contexts"]["trace"] = event.data["contexts"]["trace"]
 
             feedback_event["timestamp"] = event.datetime.timestamp()
-            feedback_event["level"] = event.data["level"]
             feedback_event["platform"] = event.platform
             feedback_event["level"] = event.data["level"]
             feedback_event["environment"] = event.get_environment().name
             feedback_event["tags"] = [list(item) for item in event.tags]
 
-        else:
```
aliu39, line 369: Wait why are we deleting this else condition? This relates to the platform and level lines below too

michellewzhang: oops that was unintentional. GOOD CATCH

aliu39 (approved):

PR #77355, opened 2024-09-11
chore(anomaly detection): re-enable for insights metrics

mifu67:
This is pretty silly but insights metrics are not being killed, so re-enable them for anomaly detection alerts.


nhsiehgit (approved):

iamrajjoshi (commented):
lgtm

PR #77364, opened 2024-09-11
fix(anomaly detection): set resolution == time_window

mifu67:
Set update frequency to be the same as time window for anomaly detection alerts, as requested by the Seer team.

### DIFF
--- a/src/sentry/incidents/logic.py
+++ b/src/sentry/incidents/logic.py
@@ -902,7 +902,9 @@ def update_alert_rule(
             updated_fields["seasonality"] = None
         elif detection_type == AlertRuleDetectionType.DYNAMIC:
             # NOTE: we set seasonality for EA
-            updated_query_fields["resolution"] = timedelta(minutes=time_window)
+            updated_query_fields["resolution"] = timedelta(
+                minutes=time_window if time_window is not None else snuba_query.time_window
+            )
             updated_fields["seasonality"] = AlertRuleSeasonality.AUTO
             updated_fields["comparison_delta"] = None
             if (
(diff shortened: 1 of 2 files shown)


nhsiehgit (changes requested):
Lets hold on this a bit while we discuss....

this doesn't seem quite right 🤔

src/sentry/incidents/logic.py
```diff
@@ -566,8 +566,10 @@ def create_alert_rule(
     """
     if monitor_type == AlertRuleMonitorTypeInt.ACTIVATED and not activation_condition:
         raise ValidationError("Activation condition required for activated alert rule")
-
-    resolution = get_alert_resolution(time_window, organization)
+    if detection_type == AlertRuleDetectionType.DYNAMIC:
+        resolution = time_window
```
nhsiehgit, line 570: If the concern is that we're sending _too_ much data too frequently

then rather than sending _too much_ data less frequently, Maybe we should be sending less data more frequently?

trillville: Yes, right now the smallest granularity we have is 15 minutes, and introducing a 5 minute granularity should be technically feasible and help address your concern

trillville:
> Lets hold on this a bit while we discuss....
> 
> this doesn't seem quite right 🤔

For the algorithm, it's a requirement that the granularity of the historical data is identical to the granularity of the data that is being evaluated. 

The reason why: We are using [matrix profiles](https://stumpy.readthedocs.io/en/latest/Tutorial_The_Matrix_Profile.html) to do this anomaly detection, which uses pattern matching to identify anomalies. If we introduce a second format of aggregation, so our historical data is 15 minute granularity time buckets and our "evaluation" data point represents an aggregation over a _moving 15 minute window_ - the comparison becomes apples to oranges. 

Per your other comment, the best path forward here I think is to simply introduce a more granular timeframe (e.g. 5 minutes) that we can use consistently

nhsiehgit (approved):
ok - i'mma just approve this for now so i'm not the blocker here 😅 

my opinion wont' matter soon so just wanted to get these concerns down just in case 😂

ram-senth (approved):

PR #77365, opened 2024-09-11
perf(replay): reduce postgres queries in recording consumer using get_from_cache

aliu39:
@cmanallen pointed out the queries we make in _handle_breadcrumb, used to check feature flags and project options, could overload Postgres as we scale. To fix this we can do the query at the top-level and use `get_from_cache`


src/sentry/replays/consumers/recording_buffered.py
```diff
@@ -294,8 +295,9 @@ def process_message(buffer: RecordingBuffer, message: bytes) -> None:
                 else None
             )
 
+        project = Project.objects.get_from_cache(decoded_message["project_id"])
```
aliu39, line 298: note this consumer isn't actually used in prod

src/sentry/replays/usecases/ingest/__init__.py
```diff
@@ -275,9 +275,10 @@ def recording_post_processor(
             op="replays.usecases.ingest.parse_and_emit_replay_actions",
             description="parse_and_emit_replay_actions",
         ):
+            project = Project.objects.get_from_cache(id=message.project_id)
```
aliu39, line 278: the code ran in prod.

JoshFerge (approved):

src/sentry/replays/usecases/ingest/dom_index.py
```diff
@@ -300,27 +299,16 @@ def _should_report_hydration_error_issue(project_id: int) -> bool:
...
-    return all([_project_has_feature_enabled(), _project_has_option_enabled()])
+def _should_report_rage_click_issue(project: Project) -> bool:
+    """
+    The feature is controlled by Sentry admins for release of the feature,
+    while the project option is controlled by the project owner, and is a
+    permanent setting
+    """
+    return features.has(
+        "organizations:session-replay-rage-click-issue-creation",
+        project.organization,
```
cmanallen, line 310: What is the behavior of this relationship lookup?  Does it query the database?  Can we extract this out higher up so we're not querying in a loop?

cmanallen (changes requested):

src/sentry/replays/usecases/ingest/dom_index.py
aliu39: The handler queries options.. yes we can

src/sentry/replays/consumers/recording_buffered.py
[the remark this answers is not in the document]
aliu39: For the cache? I don't know. Got the inspiration from [occurrence_consumer.py](https://github.com/getsentry/sentry/blob/2b87f2a655b22fd8a22722b13905f195183264dc/src/sentry/issues/occurrence_consumer.py#L304-L304)

src/sentry/replays/usecases/ingest/dom_index.py
cmanallen: I'm talking about this dot access `project.organization`

aliu39: Ah I see. I think it's a DB lookup. I refactored it to be at the top level of the loop(s).

src/sentry/replays/consumers/recording_buffered.py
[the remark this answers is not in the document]
JoshFerge: we already are usually caching the project for other places in ingest, so it probably won't be very much additional if any

aliu39: My bad, I think it is^.

src/sentry/replays/usecases/ingest/dom_index.py
```diff
@@ -300,27 +310,16 @@ def _should_report_hydration_error_issue(project_id: int) -> bool:
...
```
cmanallen, line 321: Organization is still not cached.

src/sentry/replays/usecases/ingest/dom_index.py
```diff
@@ -300,27 +310,16 @@ def _should_report_hydration_error_issue(project_id: int) -> bool:
...
-
-    return all([_project_has_feature_enabled(), _project_has_option_enabled()])
+def _should_report_rage_click_issue(project: Project) -> bool:
+    """
+    The feature is controlled by Sentry admins for release of the feature,
+    while the project option is controlled by the project owner, and is a
+    permanent setting
+    """
+    return features.has(
+        "organizations:session-replay-rage-click-issue-creation",
```
cmanallen, line 320: This feature is 100% rolled out for GA, EA, and LA.  Can we remove it?

cmanallen (changes requested):

src/sentry/replays/usecases/ingest/dom_index.py
aliu39: Can we do in a separate PR?

src/sentry/replays/usecases/ingest/dom_index.py
aliu39: It looks like Django caches foreign key queries by default: https://docs.djangoproject.com/en/dev/topics/db/queries/#one-to-many-relationships ("Forward access to one-to-many..").

To do it all in one query (like a JOIN), I added `select_related`

src/sentry/replays/usecases/ingest/dom_index.py
aliu39: ah wait, I guess this would remove the need for this function entirely. lemme check w/the team

aliu39: https://github.com/getsentry/sentry/pull/77806
Can merge w/master after this is merged^

cmanallen (approved):

PR #77367, opened 2024-09-12
feat(eap): Use RPC to fetch tag keys

Zylphrex:
This uses RPC calls to fetch tag keys from eap spans.

Depends on getsentry/snuba#6301

### DIFF
--- a/src/sentry/api/endpoints/organization_spans_fields.py
+++ b/src/sentry/api/endpoints/organization_spans_fields.py
@@ -2,8 +2,12 @@ import sentry_sdk
 from google.protobuf.timestamp_pb2 import Timestamp
 from rest_framework.request import Request
 from rest_framework.response import Response
-from sentry_protos.snuba.v1alpha.endpoint_tags_list_pb2 import TagsListRequest, TagsListResponse
-from sentry_protos.snuba.v1alpha.request_common_pb2 import RequestMeta
+from sentry_protos.snuba.v1alpha.endpoint_tags_list_pb2 import (
+    TraceItemAttributesRequest,
+    TraceItemAttributesResponse,
+)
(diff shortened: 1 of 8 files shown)


src/sentry/api/endpoints/organization_spans_fields.py
```diff
@@ -47,6 +55,50 @@ def get(self, request: Request, organization) -> Response:
...
+            "organizations:visibility-explore-dataset", organization, actor=request.user
+        ):
+            start_timestamp = Timestamp()
+            start_timestamp.FromDatetime(
+                snuba_params.start_date.replace(hour=0, minute=0, second=0, microsecond=0)
+            )
+
+            end_timestamp = Timestamp()
+            end_timestamp.FromDatetime(
+                snuba_params.end_date.replace(hour=0, minute=0, second=0, microsecond=0)
```
colin-sentry, line 68: don't you want like "tomorrow" so that you have at least a 1 day interval?

src/sentry/api/endpoints/organization_spans_fields.py
```diff
@@ -45,8 +69,58 @@ def get(self, request: Request, organization) -> Response:
...
+                + timedelta(days=1)
+            )
+
+            rpc_request = TraceItemAttributesRequest(
+                meta=RequestMeta(
+                    organization_id=organization.id,
+                    cogs_category="performance",
+                    referrer=Referrer.API_SPANS_TAG_KEYS.value,
```
wmak, line 97: nit: same nit as the other pr, probably should do a diff Referrer so we can tell them apart easily

wmak (approved):
lgtm, probably should've reviewed this before https://github.com/getsentry/sentry/pull/77510/files
but basically the same comments here as there, nothing major stands out to me

wmak:
Merging for Tony as he's on vacation, will be monitoring it

PR #77368, opened 2024-09-12
feat(dashboards): Add split task for dashboards

shruthilayaj:
The split function for dashboards is very similar to discover, calling a lot of the functions in dataset_split.py.
1. Checks if explicit error or transaction query based on query fields
2. If org has dynamic sampling feature and it's a metrics compatible query, set to transactions (this is different from discover) this is so any queries that display metrics data from MEP stay that way
3. If errors data, set to errors
4. If no errors data but has transaction data, set to transaction
5. If no errors or transactions, set to errors


tests/sentry/discover/test_dataset_split.py
```diff
@@ -3,11 +3,12 @@
 import pytest
```
shruthilayaj, line 3: changes to this file is just due to some refactor

wmak (approved):
nitpicks that you can ignore, lgtm

PR #77371, opened 2024-09-12
feat(query-builder): Increase area between trigger and overlay

malwilley:
Tiny change to give the menu a few pixels of separation from the search box:

CleanShot 2024-09-11 at 17 56 59@2x

### DIFF
--- a/static/app/components/searchQueryBuilder/tokens/filter/filterKeyOperator.tsx
+++ b/static/app/components/searchQueryBuilder/tokens/filter/filterKeyOperator.tsx
@@ -57,7 +57,7 @@ interface OperatorProps {
   middle?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


static/app/components/searchQueryBuilder/tokens/filter/filterKeyOperator.tsx
```diff
@@ -57,6 +57,8 @@ interface OperatorProps {
   middle?: boolean;
 }
 
+const MENU_OFFSET = [0, 12] as [number, number];
```
scttcper, line 60: no casting
```suggestion
const MENU_OFFSET: [number, number] = [0, 12];
```

scttcper (approved):

PR #77377, opened 2024-09-12
feat(onboarding): remove 'take me to...' buttons

### LINKED ISSUE #77138: Friction when finalizing project creation
Having to choose between going to  issues / performance / replays is potential friction for the user.
We should be more opinionated here and just forward them to what makes sense.

Important: Do not forward to performance / replays if the verify step did not create any data for it.

https://app.amplitude.com/analytics/sentry/chart/new/sgnhne1u?source=copy+url

obostjancic:
After instrumenting the final onboarding step in 
- https://github.com/getsentry/sentry/pull/76834
it turned out that the "Take me to Performance" and "Take me to session replay" buttons are no sufficiently used.

- closes #77138

Before: 
image
After:
image

### DIFF
--- a/static/app/utils/analytics/onboardingAnalyticsEvents.tsx
+++ b/static/app/utils/analytics/onboardingAnalyticsEvents.tsx
@@ -68,14 +68,6 @@ export type OnboardingEventParameters = {
     platform: string;
     project_id: string;
   };
-  'onboarding.take_me_to_performance_clicked': {
-    platform: string;
-    project_id: string;
-  };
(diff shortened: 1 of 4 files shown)


ArthurKnaus (approved):

ale-cota:
As discussed with Dhrumil, Ben and Jasmin, we are removing the secondary CTAs in order to funnel our users to the issues page after onboarding. 
- We are already able to surface associated replays directly on the issues page today when available or display a CTAs to set it up.
- Work is also underway to make traces more prominent in the issue details page or include a CTA to set it up

PR #25481, opened 2021-04-21
ref(save_event): Consolidate save_aggregate

untitaker:
Delete the old save_aggregate, rename _save_aggregate2 to
_save_aggregate and port over the changes required for hierarchical
grouping to continue working.

Also implement the required changes from #25394 for group splitting to
work. If we do those step-by-step I don't think we reduce risk, we just
increase amount of iterations + deploys to watch.

risks:

* _find_existing_group_id can be more expensive than _find_group_id, which increases the time spent in transaction. However, it is only more expensive for projects that have hierarchical grouping turned on, which is one project right now.

PR #77387, opened 2024-09-12
drop(grouping): Drop split hashes

armenzg:
This is part of undoing hierarchical grouping code.


src/sentry/models/grouphash.py
```diff
@@ -26,9 +26,6 @@ class State:
         UNLOCKED = None
         LOCKED_IN_MIGRATION = 1
 
-        # This hierarchical grouphash should be ignored/skipped for finding the group.
-        SPLIT = 2
```
armenzg, line 30: This got added in https://github.com/getsentry/sentry/pull/25481

src/sentry/api/helpers/group_index/delete.py
```diff
@@ -49,13 +49,7 @@ def delete_group_list(
     # Tell seer to delete grouping records for these groups
     call_delete_seer_grouping_records_by_hash(group_ids)
 
-    # We do not want to delete split hashes as they are necessary for keeping groups... split.
-    GroupHash.objects.filter(
-        project_id=project.id, group__id__in=group_ids, state=GroupHash.State.SPLIT
-    ).update(group=None)
-    GroupHash.objects.filter(project_id=project.id, group__id__in=group_ids).exclude(
-        state=GroupHash.State.SPLIT
-    ).delete()
```
armenzg, line 58: This got added in #28475

src/sentry/models/grouphash.py
snigdhas: Are there any rows still using this value? I've usually seen the constant kept around with a deprecated comment for compatibilitiy.

armenzg: This query returns that 1133 rows are in the LOCKED_IN_MIGRATION and none on SPLIT.
```
select
  state,
  count(*)
from
  getsentry.sentry_grouphash as gh
where
  gh.state is not null
group by gh.state;
```

snigdhas: thanks for checking -- if this isn't incompatible for self hosted, then lgtm!

snigdhas (approved):

PR #24346, opened 2021-03-09
feat: Second iteration of hierarchical grouping

untitaker:
outstanding items:

* performance improvements
* regex support in matchers

#sync-getsentry

PR #77388, opened 2024-09-12
drop(grouping): Remove invert-stacktrace as a valid stack trace rule

armenzg:
The code to support this syntax was only used in the mobile grouping and the code was removed in #76994

### DIFF
--- a/src/sentry/grouping/enhancer/actions.py
+++ b/src/sentry/grouping/enhancer/actions.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from collections.abc import Callable, Sequence
 from typing import Any
 
-from sentry.grouping.utils import get_rule_bool
 from sentry.utils.safe import get_path, set_path
 
 from .exceptions import InvalidEnhancerConfig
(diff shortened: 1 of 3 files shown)


src/sentry/grouping/enhancer/enhancement-configs/newstyle@2023-01-11.txt
```diff
@@ -68,18 +68,6 @@ family:native function:"__*[[]Sentry*"                            -app -group
 # own grouping enhancers.
 family:native function:"?[[]Sentry*"                              -app -group
 
-# Android ANR: Exception stack is a snapshot of the UI/main thread. The
```
armenzg, line 71: Originally added in #24346.

src/sentry/grouping/enhancer/actions.py
```diff
@@ -142,7 +141,6 @@ class VarAction(Action):
     _VALUE_PARSERS: dict[str, Callable[[Any], Any]] = {
         "max-frames": int,
         "min-frames": int,
-        "invert-stacktrace": get_rule_bool,
```
armenzg, line 145: Originally added in https://github.com/getsentry/sentry/pull/24346/files#diff-9f2f24a63332eb6faec958923e3483274496c40c97528f0da431fb663d2f7d58R375

snigdhas (approved):

PR #77393, opened 2024-09-12
fix(dashboards): Use custom cell renderer in `DiscoverTable`

### LINKED ISSUE #76731: Dashboard - ID column link stops working when opening Widget in Full screen mode
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

- Create widget
- Select Column "Id" 
- Save widget
- Open widget in full mode
- See that Id column is not clickable

### Expected Result

Id column needs to be clickable

### Actual Result

You can see that in the background the columns are clickable, but in full screen mode not.

Image

### Product Area

Dashboards

### Link

https://demo.sentry.io/dashboard/104975/widget/720694/?project=5808623&query=%5Bobject+Object%5D&statsPeriod=30d

### DSN

_No response_

### Version

_No response_

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2269) by [Unito](https://www.unito.io)

gggritso:
Closes https://github.com/getsentry/sentry/issues/76731. The issue is caused by using different renderer strategies. The renderer for the _small_ table (the widget's normal view) is using `SimpleTableChart` which correctly delegates the rendering of special fields to the relevant dataset's `getCustomFieldRenderer` configuration. The full-screen modal uses the `DiscoverTable` component, which sets up the custom renderers differently. There, the custom renderers do _not_ use the linkifying logic from the table.

This PR ensures that Discover and Discover-ish datasets import and use the same field renderers. Longer-term the solution would be to move this logic to the _default_ field renderers or otherwise unify the renderers used in Dashboards but for now this fixes the issue.

### DIFF
--- a/static/app/components/modals/widgetViewerModal/widgetViewerTableCell.tsx
+++ b/static/app/components/modals/widgetViewerModal/widgetViewerTableCell.tsx
@@ -34,6 +34,7 @@ import {
   generateEventSlug,
 } from 'sentry/utils/discover/urls';
 import {formatMRIField, parseField} from 'sentry/utils/metrics/mri';
+import {getCustomEventsFieldRenderer} from 'sentry/views/dashboards/datasetConfig/errorsAndTransactions';
 import type {Widget} from 'sentry/views/dashboards/types';
 import {DisplayType, WidgetType} from 'sentry/views/dashboards/types';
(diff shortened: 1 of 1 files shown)


edwardgou-sentry (approved):
nice

nikkikapadia (approved):
lgtm :)

Issue #69881, opened 2024-04-29
[Epic] Replay x Web Vitals


ryan953:
Project Details; https://www.notion.so/sentry/Web-Vitals-x-Session-Replay-Discussion-f6cde2c7d54846749eefa83d4dd9882b
1st KTTR: https://www.notion.so/sentry/Replay-Web-Vitals-Breadcrumbs-0038416a37d8462793ad47a0cbce6ff2
2nd KTTR: https://www.notion.so/sentry/Replay-Web-Vitals-Insights-ff1230303b66485f9fac7441cc34ca46

```[tasklist]
### @sentry/replay SDK - Capture Web Vitals
- [ ] https://github.com/getsentry/sentry-javascript/issues/11639
- [ ] https://github.com/getsentry/sentry-javascript/issues/12470
- [ ] Add TTFB web vitals https://github.com/getsentry/sentry-javascript/pull/12993
```

```[tasklist]
### Replay Details - Render Vitals as Breadcrumbs
- [ ] TODO: How & where will we render the breadcrumb information? https://chromewebstore.google.com/detail/web-vitals/ahfhijdlegdabablpippeagghigmibma has some prior art to reference
```

```[tasklist]
### Replay List Page - show aggregate vitals
- [ ] TODO: Update table render (see figma https://www.figma.com/file/7ejXv6UyWd9ObWYd3i8B4b/Replay-x-Web-Vitals?type=whiteboard&node-id=3-21&t=lHvmKWYMSkWbh6Pj-0)
- [ ] TODO: Do we link to the existing /performance pages?
```

```[tasklist]
### Replay Details - New Vitals Tab/Panel
- [ ] TODO: Iterate on figma https://www.figma.com/file/7ejXv6UyWd9ObWYd3i8B4b/Replay-x-Web-Vitals?type=whiteboard&node-id=3-21&t=lHvmKWYMSkWbh6Pj-0
- [ ] TODO: implement figma
```

PR #77397, opened 2024-09-12
fix(replay): CLS score is not in ms

c298lee:
If the web vital is CLS, don't include `ms`. CLS is the only web vital in replays that isn't measured in milliseconds.

Relates to https://github.com/getsentry/sentry/issues/69881

### DIFF
--- a/static/app/utils/replays/getFrameDetails.tsx
+++ b/static/app/utils/replays/getFrameDetails.tsx
@@ -44,6 +44,7 @@ import type {
 } from 'sentry/utils/replays/types';
 import {
   getFrameOpOrCategory,
+  isCLSFrame,
   isDeadClick,
   isDeadRageClick,
   isRageClick,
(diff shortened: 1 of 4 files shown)


k-fish (approved):

static/app/utils/replays/getFrameDetails.tsx
```diff
@@ -286,8 +286,9 @@ const MAPPER_FOR_FRAME: Record<string, (frame) => Details> = {
       case 'good':
         return {
           color: 'green300',
-          description: tct('[value]ms (Good)', {
+          description: tct('[value][ms] (Good)', {
             value: frame.data.value.toFixed(2),
+            ms: frame.description === 'cumulative-layout-shift' ? '' : 'ms',
```
michellewzhang, line 291: small nit: could move this into a constant to avoid having to type this ternary out 3 times

michellewzhang (approved):

michellewzhang (approved):

PR #77405, opened 2024-09-12
chore(sentry-apps): add instrumentation for sentry-app-components endpoint

cathteng:
Fix "Missing Instrumentation" when viewing the trace for this endpoint.

### DIFF
--- a/src/sentry/api/endpoints/integrations/sentry_apps/components.py
+++ b/src/sentry/api/endpoints/integrations/sentry_apps/components.py
@@ -1,3 +1,4 @@
+import sentry_sdk
 from rest_framework.request import Request
 from rest_framework.response import Response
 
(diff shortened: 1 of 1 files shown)


sentaur-athena (approved):

src/sentry/api/endpoints/integrations/sentry_apps/components.py
```diff
@@ -54,23 +55,29 @@ def get(
         components = []
         errors = []
 
-        for install in SentryAppInstallation.objects.get_installed_for_organization(
-            organization.id
-        ).order_by("pk"):
-            _components = SentryAppComponent.objects.filter(
-                sentry_app_id=install.sentry_app_id
-            ).order_by("pk")
+        with sentry_sdk.start_transaction(name="sentry.api.sentry_app_components.get"):
```
GabeVillalobos, line 58: Is just adding a span here sufficient? The docs on performance lead me to believe we already have a top-level transaction via the SDK for Django: https://docs.sentry.io/platforms/python/tracing/instrumentation/automatic-instrumentation/

GabeVillalobos (approved):

src/sentry/api/endpoints/integrations/sentry_apps/components.py
cathteng: Screenshot 2024-09-12 at 10 35 58

The in app instructions are telling me to add a transaction too

GabeVillalobos: Ah missed that, thanks!

PR #77407, opened 2024-09-12
typing(models): Type event store models

armenzg:


src/sentry/eventstore/models.py
```diff
@@ -61,12 +61,12 @@ def __init__(
...
         event_id: str,
         snuba_data: Mapping[str, Any] | None = None,
-    ):
+    ) -> None:
         self.project_id = project_id
         self.event_id = event_id
         self._snuba_data = snuba_data or {}
 
-    def __getstate__(self) -> Mapping[str, Any]:
+    def __getstate__(self) -> dict[str, Any]:
```
armenzg, line 69: This will allow for some `.pop()` that is done later on.

src/sentry/eventstore/models.py
```diff
@@ -99,7 +99,7 @@ def platform(self) -> str | None:
         column = self._get_column_name(Columns.PLATFORM)
         if column in self._snuba_data:
             return cast(str, self._snuba_data[column])
-        return cast(str, self.data.get("platform", None))
+        return self.data.get("platform")
```
armenzg, line 102: Not needed.

src/sentry/eventstore/models.py
```diff
@@ -499,47 +496,6 @@ def get_raw_data(self, for_stream: bool = False) -> Mapping[str, Any]:
     def size(self) -> int:
         return len(orjson.dumps(dict(self.data)).decode())
 
-    def get_email_subject(self) -> str:
-        template = self.project.get_option("mail:subject_template")
-        if template:
-            template = EventSubjectTemplate(template)
-        elif self.group.issue_category == GroupCategory.PERFORMANCE:
```
armenzg, line 506: self.group is not available in `BaseEvent`. Moving this function to `EventGroup`.

src/sentry/eventstore/models.py
```diff
@@ -499,47 +496,6 @@ def get_raw_data(self, for_stream: bool = False) -> Mapping[str, Any]:
...
-            if k in data:
-                continue
-            if k == "sdk":
-                v = {v_k: v_v for v_k, v_v in v.items() if v_k != "client_ip"}
-            data[k] = v
-
-        # for a long time culprit was not persisted.  In those cases put
-        # the culprit in from the group.
-        if data.get("culprit") is None and self.group_id and self.group:
-            data["culprit"] = self.group.culprit
```
armenzg, line 535: self.group is not available in `BaseEvent`. Moving this function to `EventGroup`.

src/sentry/eventstore/models.py
```diff
@@ -591,10 +547,10 @@ def __init__(
     ):
         super().__init__(project_id, event_id, snuba_data=snuba_data)
         self.group_id = group_id
-        self.groups = groups
+        self.groups = groups or []
         self.data = data
```
armenzg, line 551: @asottile-sentry @wedamija I need help here. I can't figure out how to fix this error.

```
mypy src/sentry/eventstore/models.py
src/sentry/eventstore/models.py:549: error: Incompatible types in assignment (expression has type "Mapping[str, Any] | None", variable has type "NodeData")  [assignment]
Found 1 error in 1 file (checked 1 source file)
```

asottile-sentry: mypy doesn't understand disparate setters unfortunately -- you can work around this by making the property setter call some private method and have this call that same private method

wedamija: The good news is that `groups` is deprecated, so you could just look into removing it instead of fixing mypy

asottile-sentry: nice!  and actually latest mypy now understands disparate setters so this part might "just work" already

PR #77412, opened 2024-09-12
fix(jira): Ensures metrics/logs are captured for all issue creation exceptions

GabeVillalobos:
Currently, we only capture exceptions for `IntegrationFormErrors` which likely covers the vast majority of exception types; however, including all exceptions in our logging and metrics will give us a better basis for setting ticketing SLOs. This also removes the feature flag usage for conditionally raising.


cathteng (approved):
👀

PR #26740, opened 2021-06-21
feat(grouping): Hierarchical titles

untitaker:
Right now our existing issue titles for native are not good at distinguishing groups from each other when viewed in the issuestream, as they just use the top frame's function name.

This will become worse in the case that issue splitting lands, in which case we need to show more verbose titles to distinguish subissues further.

There are unaddressed questions about what happens with title search in discover. Right now the answer is to set the title of the most specific grouping level such that wildcard search shows good enough results. But that is something to revisit in the future.

Fixes [INGEST-75]

TODO:

- [x] test that the actual group title (and not latestEvent title) is actually used everywhere correctly in UI

----

<details><summary>SCREENSHOTS</summary>

---- 
Assuming the following stacktrace:

Screenshot 2021-06-22 at 12 30 08

Breakdown at max level:

Screenshot 2021-06-22 at 12 30 19

Title gets shorter if there is less frames being grouped by:

Screenshot 2021-06-22 at 12 30 22

Screenshot 2021-06-22 at 12 30 26

</details>

[INGEST-75]: https://getsentry.atlassian.net/browse/INGEST-75

PR #77413, opened 2024-09-12
ref(grouping): Remove CalculatedHashes

armenzg:
Code originally added in #26740


src/sentry/grouping/ingest/hashing.py
```diff
@@ -85,7 +84,7 @@ def _calculate_event_grouping(
                 event.data["grouping_config"] = get_grouping_config_dict_for_project(project)
                 hashes = event.get_hashes()
 
-        hashes.write_to_event(event.data)
```
armenzg, line 88: The only usage of this function from CalculatedHashes.

src/sentry/grouping/result.py
```diff
@@ -1,25 +0,0 @@
...
-
-@dataclass(frozen=True)
-class CalculatedHashes:
-    hashes: list[str]
-    # `variants` will never be `None` when the `CalculatedHashes` instance is created as part of
-    # event grouping, but it has to be typed including `None` because we use the `CalculatedHashes`
-    # container in other places where we don't have the variants data
-    variants: dict[str, BaseVariant] = field(default_factory=dict)
-
-    def write_to_event(self, event_data: NodeData) -> None:
```
armenzg, line 16: Only used in a single location in the codebase and I changed it.
image

src/sentry/grouping/result.py
```diff
@@ -1,25 +0,0 @@
...
-        event_data["hashes"] = self.hashes
-
-    @classmethod
-    def from_event(cls, event_data: NodeData) -> Optional["CalculatedHashes"]:
```
armenzg, line 20: I can't figure out anywhere we use it.

PR #77414, opened 2024-09-12
chore(issues): Remove feedback button from highlights

mrduncan:
This feature has been GA for a while now and we're occasionally seeing users confuse it with issue activity thinking it'll be visible to their other team members.


JoshFerge (approved):

MichaelSun48 (approved):

scttcper (approved):
yes

PR #77423, opened 2024-09-12
copy(activity): Improve copy for linked Azure Devops issue

### LINKED ISSUE #77410: Improve activity tab message when linking a Sentry issue to an exsisting Azure issue
### Problem Statement

Currently, when [a Sentry issue](https://org2-w7.sentry.io/issues/5793208489/activity/?project=4507704420139008&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=7) is linked to an Azure issue, the activity tab message is: 
Image
This isn't too clear and has caused confusion with azure issue assignment. 

### Solution Brainstorm

It would be helpful if the message made it clear that the user linked the Sentry issue to an existing Azure issue instead of suggesting a new Azure issue was created. 

Something like the below:
User linked this issue to Azure DevOps issue titled [Error: new next test](https://dev.azure.com/int4sentry/_workitems/edit/5)

### Product Area

Issues

MichaelSun48:
resolves #77410

Changes the copy of the activity log entry when a sentry issue is linked to an Azure DevOps ticket to be more clear that the two issues are linked. 

Before: 

> \<author> created an issue on Azure Devops titled <devops_issue_title>

After:

>  \<author> **linked this** issue **to** Azure Devops **issue** titled <devops_issue_title>

### DIFF
--- a/static/app/views/issueDetails/groupActivityItem.tsx
+++ b/static/app/views/issueDetails/groupActivityItem.tsx
@@ -474,6 +474,15 @@ function GroupActivityItem({
       }
       case GroupActivityType.CREATE_ISSUE: {
         const {data} = activity;
+
+        if (data.provider === 'Azure DevOps') {
+          return tct('[author] linked this issue to [provider] issue titled [title]', {
+            author,
+            provider: data.provider,
+            title: <ExternalLink href={data.location}>{data.title}</ExternalLink>,
+          });
+        }
+
         return tct('[author] created an issue on [provider] titled [title]', {
           author,
           provider: data.provider,


malwilley:
Is there some behavior that's specific to Azure here? Or is this the case for any created issue?

MichaelSun48:
> Is there some behavior that's specific to Azure here? Or is this the case for any created issue?

@malwilley I'm not sure, I just didn't want to blanket change the copy since the ticket in question only concerns Azure DevOps. I wouldn't be against changing it for all "linked ticket" actions, but I'm guessing we'd want to have a bigger discussion about that.

malwilley:
Hmm, if there is no functional difference between providers I don't think we should special case Azure. That would only introduce more confusion. Might want to ask around and see if this change makes sense or not for all create issue activity entries.

PR #77426, opened 2024-09-12
ref(grouping): Remove CalculatedHashes

armenzg:
Code originally added in https://github.com/getsentry/sentry/pull/26740


src/sentry/grouping/ingest/hashing.py
```diff
@@ -85,7 +83,6 @@ def _calculate_event_grouping(
                 event.data["grouping_config"] = get_grouping_config_dict_for_project(project)
                 hashes = event.get_hashes()
 
-        hashes.write_to_event(event.data)
```
armenzg, line 88: `get_hashes()` now takes care of writing to the event:
image

src/sentry/grouping/ingest/hashing.py
```diff
@@ -239,22 +235,17 @@ def get_hash_values(
         secondary_hashes,
     )
 
-    all_hashes = CalculatedHashes(
-        hashes=extract_hashes(primary_hashes) + extract_hashes(secondary_hashes),
-        # We don't set a combo `variants` value here because one set of variants would/could
-        # partially or fully overwrite the other (it's a dictionary), and having variants from two
-        # different configs all mixed in together makes no sense.
+    return (
```
JoshFerge, line 238: is it necessary to return the third combined item here? can the callers just concatenate them?

JoshFerge (approved):
makes sense to me. might be good for @lobsterkatie to take a look as well.

src/sentry/grouping/ingest/hashing.py
armenzg: Good point. I will look into it.

armenzg:
> makes sense to me. might be good for @lobsterkatie to take a look as well.

@JoshFerge ugh. Want me to revert? Or wait to hear back?

JoshFerge:
@armenzg i think you're fine.

PR #77435, opened 2024-09-12
feat(uptime): Add option to restrict issue creation via host id

### LINKED ISSUE #76875: Add a new options automator feature to disable issue creation for a specific providers

evanpurkhiser:
This will allow us to disable issue creation for specific hosting
providers using the `host_provider_id` field of the UptimeSubscription.

Resolves https://github.com/getsentry/sentry/issues/76875


src/sentry/options/defaults.py
```diff
@@ -2713,3 +2713,16 @@
...
+
+# Restrict uptime issue creation for specific host provider identifiers. Items
+# in this list map to the `host_provider_id` column in the UptimeSubscription
+# table.
+#
+# This may be used to stop issue creation in the event that a network / hosting
+# provider blocks the uptime checker causing false positives.
+register(
+    "uptime.restrict-issue-creation-by-hosting-provider-id",
+    type=Sequence,
```
wedamija, line 2725: Are there any options for sets here? Probably not a huge deal since this will be small anyway

src/sentry/testutils/fixtures.py
```diff
@@ -651,17 +651,22 @@ def create_uptime_subscription(
...
         url="http://sentry.io/",
+        host_provider_id: str | None = None,
         interval_seconds=60,
         timeout_ms=100,
         date_updated: None | datetime = None,
     ) -> UptimeSubscription:
         if date_updated is None:
             date_updated = timezone.now()
+        if host_provider_id is None:
+            host_provider_id = "TEST"
```
wedamija, line 662: Just noting that this makes it hard to create one of these with an actual `None` value here

src/sentry/uptime/consumers/results_consumer.py
```diff
@@ -258,9 +258,24 @@ def handle_result_for_project_active_mode(
...
+            restricted_host_provider_ids = options.get(
+                "uptime.restrict-issue-creation-by-hosting-provider-id"
+            )
+            issue_creation_restricted_by_provider = (
+                project_subscription.uptime_subscription.host_provider_id
+                in restricted_host_provider_ids
+            )
+
+            if issue_creation_restricted_by_provider:
+                metrics.incr("uptime.result_processor.restricted_by_provider", sample_rate=1.0)
```
wedamija, line 276: Wondering if we should include the provider here or not... I can't imagine this having more than 10 or so values. Or maybe a log with extra info if we don't want it in datadog?

wedamija (approved):

src/sentry/options/defaults.py
evanpurkhiser: I was looking for that too but couldn't find it

src/sentry/uptime/consumers/results_consumer.py
evanpurkhiser: Ah yeah true. That's probably pretty low cardinality. I'll toss it in

evanpurkhiser:
Will cleanup these two things after the fact.

PR #77439, opened 2024-09-12
fix(substatus): Add check constraint for Group.substatus

### LINKED ISSUE #76936: Replace pre-save with database constraint

snigdhas:
We've finally got issues to the point where the substatus-status pairings are all valid and set correctly through various update methods. We can now add a database constraint on this column to ensure the rows never get updated in an inconsistent manner. 

Fixes https://github.com/getsentry/sentry/issues/76936


wedamija (approved):
This looks good to me. Make sure you're around to monitor things when applying this constraint, just in case it causes some unexpected issue with group creation/mutation

PR #77441, opened 2024-09-12
ref(profiling) move source location to frame class

JonasBa:
Move source location to frame class and change it so that we always display lineno and colno even if we dont have file or path


Zylphrex (approved):

JonasBa:
@viglia it's not very useful, but not showing it is arguably even less useful because it makes debugging these scenarios ennoying

PR #77444, opened 2024-09-12
ref(integrations): Add type annotations to Slack webhooks package

RyanSkonnord:
Remove "sentry.integrations.slack.webhooks.*" from the Mypy exemptions list. Adjust related annotations and refactor small functions as needed.


src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -240,9 +240,20 @@ def validation_error(
...
+    @staticmethod
+    def _unpack_error_text(validation_error: serializers.ValidationError) -> str:
+        detail = validation_error.detail
+        while True:
+            if isinstance(detail, dict):
+                detail = list(detail.values())
+            element = detail[0]
+            if isinstance(element, str):
+                return element
+            detail = element
```
RyanSkonnord, line 255: The `rest_framework` library defines the "detail" type as

```python
_Detail: TypeAlias = ErrorDetail | list[_Detail] | dict[str, _Detail]
```

Because of that recursively nested structure, this kind of loop is necessary to guarantee that we unpack a string (`ErrorDetail` extends `str`) without making assumptions about the structure as the original code (`list(*error.detail.values())[0]`) does. This satisfies type-checking and should be more robust.

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -728,6 +728,17 @@ def _handle_group_actions(
...
 
+    def _send_blocks(self, response: SlackBlock, response_url: str, text: str | None = None) -> str:
+        json_blocks = orjson.dumps(response.get("blocks")).decode()
+        webhook_client = WebhookClient(response_url)
+        webhook_client.send(
+            blocks=json_blocks,  # type: ignore[arg-type]
+            text=text,
+            delete_original=False,
+            replace_original=True,
+        )
```
RyanSkonnord, line 739: Sending a `str` as the `blocks` argument just plain contradicts the library's type annotation (`Optional[Sequence[Union[Dict[str, Any], Block]]]`). I'm guessing that we're taking advantage of an (undocumented?) feature where a string value of valid JSON gets parsed into the equivalent `Block`s.

It sure looks like we could just send `response.get("blocks")` as the value of `blocks`, but I hesitate to change that without knowing why it was dumped to a string in the first place (principle of Chesterson's Fence). Was it solely so it could be passed to the logger on line 714?

I'd like to attempt that in a follow-up PR (and I'd be comfortable shipping it if no tests break), but for now this is refactoring is behaviorally equivalent (I only extracted the new helper method so that we would only need the type-ignore comment in one place).

src/sentry/notifications/utils/actions.py
```diff
@@ -33,12 +33,5 @@ class MessageAction:
...
 
 @dataclass
-class BlockKitMessageAction:
-    name: str
-    label: str
-    type: Literal["button", "select"] = "button"
-    url: str | None = None
-    value: str | None = None
-    action_id: str | None = None
+class BlockKitMessageAction(MessageAction):
```
RyanSkonnord, line 36: Other code was written in a way that suggests `BlockKitMessageAction` is supposed to have an is-a relationship to `MessageAction`. This change causes `BlockKitMessageAction` to gain a few attributes by inheritance, but only ones with default values, meaning no other code needs to change. (It's possible that some of those attributes aren't sensible, but if that's true then they should be moved from `MessageAction` into a different subclass).

RyanSkonnord (commented):
I've commented on the nontrivial parts.

PR #77448, opened 2024-09-12
feat(workflow_engine): Add DataConditions

saponifi3d:
# Description
[📜 Technical spec]( https://www.notion.so/sentry/Alerts-Create-Issues-9114a498098143178839d584c293fe75?pvs=4#2910558fcd2941b9af88d86fa37fa390)

### PR Changes
- Adds DataConditions
- Connects `DataCondition` with a 1 to many relationship for `DataConditionGroups` the Group model allows us to define `ANY` and `ALL`, and group the conditions for evaluation.


markstory (approved):
Schema looks good to me.

src/sentry/workflow_engine/models/data_condition.py
```diff
@@ -0,0 +1,26 @@
...
+
+    __relocation_scope__ = RelocationScope.Organization
+    __repr__ = sane_repr("type", "condition")
+
+    condition = models.CharField(max_length=200)
+    threshold = models.FloatField()
+    condition_result = models.JSONField()
+    type = models.CharField(max_length=200)
+
+    group = models.ForeignKey(
```
saponifi3d, line 23: not a huge fan of `group` here, any other recommendations / suggestions? (I'm nervous about name collisions with `Group` for issues.

markstory: What about `condition_group` or `condition_clause` 🤷

saponifi3d: 👍 i'll go with `condition_group` i feel like that's different enough from `issues` but close to this. thanks!

src/sentry/workflow_engine/models/data_condition_group.py
```diff
@@ -0,0 +1,28 @@
...
+from sentry.db.models import DefaultFieldsModel, region_silo_model, sane_repr
+
+from .detector import Detector
+from .workflow_action import WorkflowAction
+
+
+@region_silo_model
+class DataConditionGroup(DefaultFieldsModel):
+    """
+    A data condition is a way to specify a condition that must be met for a workflow to execute.
```
wedamija, line 13: `A data group is a way to specify a group of conditions that must be met for a workflow action to execute`

src/sentry/workflow_engine/models/data_condition_group.py
```diff
@@ -0,0 +1,28 @@
...
+    class Type(models.TextChoices):
+        ANY = "any"
+        ALL = "all"
+
+    logic_type = models.CharField(max_length=200, choices=Type.choices, default=Type.ANY)
+
+    detector = models.ForeignKey(Detector, on_delete=models.CASCADE, blank=True, null=True)
+    workflow_action = models.ForeignKey(
+        WorkflowAction, on_delete=models.CASCADE, blank=True, null=True
+    )
```
wedamija, line 28: Shouldn't we allow multiple workflow actions to be linked here? So this group of conditions fires this group of actions?

src/sentry/workflow_engine/models/data_condition.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    A data condition is a way to specify a condition that must be met for a workflow to execute.
+    """
+
+    __relocation_scope__ = RelocationScope.Organization
+    __repr__ = sane_repr("type", "condition")
+
+    condition = models.CharField(max_length=200)
+    threshold = models.FloatField()
+    condition_result = models.JSONField()
```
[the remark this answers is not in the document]
saponifi3d: This schema would just be a primitive value; either a string for the detector or boolean for the workflow actions. i can imagine some other scenarios that having this be flexible as a json primitive - but it should never be a compex object / array etc.

is there a good way to enforce that? (off the top of my head, we could have a pre-save signal that validates it's a primitive or throws an exception? 🤔)

src/sentry/backup/comparators.py
```diff
@@ -865,6 +865,7 @@ def get_default_comparators() -> dict[str, list[JSONScrubbingComparator]]:
             ],
             "sentry.userrole": [DateUpdatedComparator("date_updated")],
             "sentry.userroleuser": [DateUpdatedComparator("date_updated")],
+            "workflow_engine.datacondition": [DateUpdatedComparator("date_updated", "date_added")],
```
wedamija, line 868: You'll need to add all the new models here

src/sentry/workflow_engine/models/data_condition.py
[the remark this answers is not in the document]
wedamija: I think we need a json schema to validate it

src/sentry/workflow_engine/models/workflow.py
```diff
@@ -15,6 +17,9 @@ class Workflow(DefaultFieldsModel):
     name = models.CharField(max_length=200)
     organization = FlexibleForeignKey("sentry.Organization")
 
+    # Required as the 'when' condition for the workflow
+    condition_group = FlexibleForeignKey(DataConditionGroup, blank=True, null=True)
```
wedamija, line 21: Wondering if we should name it as `when_condition_group` or something

wedamija (commented):
Looks good although I'd like to see the generated sql after ci is fixed

src/sentry/workflow_engine/models/workflow.py
saponifi3d: we def can, one thing i liked about keeping this generic is that we can match the same "required" condition group with detectors if we don't have `when`. I thought it was kinda nice to have those two terms linked because this will be responding to what the detectors emit.

src/sentry/backup/comparators.py
saponifi3d: 👍 thanks - yeah, i was avoiding doing all those changes initially just cause of all the things. updated! (hopefully the tests are green now too 🤞)

src/sentry/workflow_engine/models/action.py
```diff
@@ -0,0 +1,24 @@
+from django.db import models
+
+from sentry.backup.scopes import RelocationScope
+from sentry.db.models import DefaultFieldsModel, region_silo_model, sane_repr
+
+
+@region_silo_model
+class Action(DefaultFieldsModel):
+    """
+    Actions are actions that can be taken if the conditions of a DataConditionGrou are satisfied.
```
wedamija, line 10: Nit: `DataConditionGroup`

src/sentry/workflow_engine/models/action.py
```diff
@@ -0,0 +1,24 @@
...
+
+    __relocation_scope__ = RelocationScope.Organization
+    __repr__ = sane_repr("workflow_id", "type")
+
+    # TODO (@saponifi3d): Don't hardcode these values
+    class Type(models.TextChoices):
+        Notification = "SendNotificationAction"
+        TriggerWorkflow = "TriggerWorkflowAction"
+
+    required = models.BooleanField(default=False)
```
wedamija, line 22: I've completely forgotten what `required` means. Might be worth a comment

src/sentry/workflow_engine/models/action.py
```diff
@@ -0,0 +1,24 @@
...
+    Examples include: detectors emitting events, sending notifications, creating an issue in the Issue Platform, etc.
+    """
+
+    __relocation_scope__ = RelocationScope.Organization
+    __repr__ = sane_repr("workflow_id", "type")
+
+    # TODO (@saponifi3d): Don't hardcode these values
+    class Type(models.TextChoices):
+        Notification = "SendNotificationAction"
+        TriggerWorkflow = "TriggerWorkflowAction"
```
wedamija, line 20: Will these be full paths to the actions in practice, similar to issue rules?

src/sentry/workflow_engine/models/data_condition.py
```diff
@@ -0,0 +1,28 @@
...
+    A data condition is a way to specify a logic condition, if the condition is met, the condition_result is returned.
+    """
+
+    __relocation_scope__ = RelocationScope.Organization
+    __repr__ = sane_repr("type", "condition")
+
+    condition = models.CharField(max_length=200)
+    threshold = models.FloatField(blank=True, null=True)
+    comparison = models.CharField(max_length=200, blank=True, null=True)
+
```
wedamija, line 21: Wondering if it's worthwhile to have comments here explaining what the values look like.

src/sentry/workflow_engine/models/data_condition.py
```diff
@@ -0,0 +1,28 @@
...
+class DataCondition(DefaultFieldsModel):
+    """
+    A data condition is a way to specify a logic condition, if the condition is met, the condition_result is returned.
+    """
+
+    __relocation_scope__ = RelocationScope.Organization
+    __repr__ = sane_repr("type", "condition")
+
+    condition = models.CharField(max_length=200)
+    threshold = models.FloatField(blank=True, null=True)
```
wedamija, line 19: An example from the doc that `threshold` could be a string value
```
{
	"threshold": "['tag-a', 'tag-b']",
	"condition": "one_of",
	"condition_result": IssuePriority.LOW
	"type": WorkflowActionTagFilter
}
```

Although I'm not sure if it's up to date, since it doesn't include the `comparison` value either.

src/sentry/workflow_engine/models/detector.py
```diff
@@ -15,6 +15,13 @@ class Detector(DefaultFieldsModel, OwnerModel):
     name = models.CharField(max_length=200)
     data_sources = models.ManyToManyField("workflow_engine.DataSource", through=DataSourceDetector)
 
+    condition_group = FlexibleForeignKey(
```
wedamija, line 18: Wondering if this should be named like `workflow_condition_group` or maybe have a comment explaining how it's used?

src/sentry/workflow_engine/models/action.py
saponifi3d: That part is a bit unclear to me, i think it'd be cool if we could derive the information from an `init` file or something and automate it as much as possible. if we did it that way w/o the full path, i think we'd need a registry underneath it to link this type to the class itself.

If we did the full id, could likely omit that registry? (i believe we could just get the module by the module id, just want to confirm)

src/sentry/workflow_engine/models/data_condition.py
saponifi3d: Yeah - i added `comparison` to be the string representation and `threshold` to be the value. i found that when i was adding it to the organization and i wanted to setup a detector - i found i needed to have a string comparison, but only had an float.

alternative could be that we store this in 1 field, but we'd have to parse that into the different types based on the condition type.

any recommendations? (i'm missing JS where i can just set the type to be what i want 🤣)

src/sentry/workflow_engine/models/action.py
wedamija: Yeah, either way can work out ok. I guess one thing that's nice about a registry is that it's less susceptible to accidental breakage when people move things around.

src/sentry/workflow_engine/models/data_condition.py
wedamija: I think either option is fine. I guess that one field that is json with an enforced schema is more consistent, since you're going to need to use json to encode the comparison field either way. But I don't really have a strong opinion on this.

src/sentry/workflow_engine/models/data_condition.py
```diff
@@ -0,0 +1,41 @@
...
+    condition = models.CharField(max_length=200)
+
+    # The threshold is the value that the condition is compared to for numeric expressions
+    threshold = models.FloatField(blank=True, null=True)
+
+    # The comparison is the value that the condition is compared to for string expressions
+    comparison = models.CharField(max_length=200, blank=True, null=True)
+
+    # The condition_result is the value that is returned if the condition is met, this must be a primitive value
+    condition_result = models.JSONField(validators=[validate_primitive_value])
```
wedamija, line 33: Unfortunately validators aren't run on model save, and they only work with Django forms, which we don't use.

You'd need to have some way to trigger this on save. It might be possible to do this on presave signal, maybe you can manually trigger the validation logic that way. If we wanted that, we could build it into the base class. Possibly opt in to the behaviour via a class variable or on the `Meta` class.

I think this pr is pretty big though, so maybe we're better off figuring out that part in a separate pr.

wedamija (approved):

saponifi3d:
going to take @wedamija 's +1 from yesterday - the changes have just been to fix the relocation scope and get the PR green.

PR #77453, opened 2024-09-12
fix(releases): use get_or_create instead of sub transaction

JoshFerge:
reduce the amount of sub transactions by using get_or_create instead of catching integrity errors. also removes the ff flag for now. will keep the ff flag defined incase i make any more radical changes

### DIFF
--- a/src/sentry/models/releases/set_commits.py
+++ b/src/sentry/models/releases/set_commits.py
@@ -7,7 +7,6 @@ from typing import TypedDict
 
 from django.db import IntegrityError, router
 
-from sentry import features
 from sentry.constants import ObjectStatus
 from sentry.db.postgres.transactions import in_test_hide_transaction_boundary
 from sentry.locks import locks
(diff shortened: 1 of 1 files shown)


snigdhas (approved):

src/sentry/models/releases/set_commits.py
```diff
@@ -58,19 +57,14 @@ def set_commits(release, commit_list):
         # the same release rapidly for different projects.
         raise ReleaseCommitError
 
-    if features.has("organizations:set-commits-updated", release.organization):
-        create_repositories(commit_list, release)
-        create_commit_authors(commit_list, release)
+    create_repositories(commit_list, release)
+    create_commit_authors(commit_list, release)
```
wedamija, line 61: Are you sure we don't still want these inside of the lock, but not the transaction?

JoshFerge: added inside of lock, but not transaction.

PR #63925, opened 2024-01-26
ref(slack): Make Archive action a modal

isabellaenriquez:
This PR changes the archive Slack action to be a modal instead of a button, allowing users to archive issues until escalating, x event happens, or forever. This PR also condenses the archive action tests a bit to reduce repeated code (will follow up in a separate PR for resolve and assign).

image

PR #63623, opened 2024-01-22
ref(slack): Add more issue context and format notification

ceorourke:
Adds issue context to Slack alert notifications and attempts to match the [designs](https://www.figma.com/file/aqQIFNXUURbHZKHEO0oYxf/Ecosystem-Notifications?type=design&node-id=318-39553&mode=design&t=6THjvg2mTmQzWI1K-0). This adds: 
* project slug (linkified)
* event count
* user count
* substate
* first seen
* error level as an emoji if the issue type is error
* issue category as an emoji if the issue type is not error
* Adds `level` and `release` tags by default, does not duplicate if explicitly requested

This also strips any empty spaces before the error text, since we were seeing some notifications come in like that.

**Issue alert**
Screenshot 2024-01-23 at 2 55 45 PM

**After Archiving**
Screenshot 2024-01-23 at 2 55 56 PM

**Workflow notification post resolution**
Screenshot 2024-01-23 at 3 12 16 PM

Closes #62902  and #62904 

Requires https://github.com/getsentry/getsentry/pull/12690

PR #77454, opened 2024-09-12
ref(integrations): Extract dialog helper class from SlackActionEndpoint

RyanSkonnord:
Factor out duplication between `open_resolve_dialog` and `open_archive_dialog` by introducing a private helper class. Group related class into the helper class to make `SlackActionEndpoint` smaller.


RyanSkonnord:
The changes might be easier to review if you focus on [the first commit](https://github.com/getsentry/sentry/pull/77454/commits/d9a66f458c0b7916ea885286433a249cf8083add) and set the "Hide whitespace" option.

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -942,3 +700,222 @@ def handle_member_approval(self, slack_request: SlackActionRequest, action: str)
...
+        # but seems like there's no other solutions [1]:
+        #
+        # [1]: https://stackoverflow.com/questions/46629852/update-a-bot-message-after-responding-to-a-slack-dialog#comment80795670_46629852
+        org = group.project.organization
+
+        callback_id_dict = {
+            "issue": group.id,
+            "orig_response_url": slack_request.data["response_url"],
+            "is_message": _is_message(slack_request.data),
+            "rule": slack_request.callback_data.get("rule"),
```
RyanSkonnord, line 827: This line gave me a surprising amount of trouble, trying to determine whether it was worth keeping it under the "if...channel" condition below, as is [the case](https://github.com/getsentry/sentry/pull/77454/files#diff-30aac95a95c4c9b782c0eb9a15b9d363724bc3c7a0179e1f7b65be80fbe7c7c4L438) in the original `open_archive_dialog`.

I spelunked into the Git history a bit and found that it was [originally written that way](https://github.com/getsentry/sentry/pull/63623/files#diff-30aac95a95c4c9b782c0eb9a15b9d363724bc3c7a0179e1f7b65be80fbe7c7c4R340) in #63623 by @ceorourke, and that it [was written](https://github.com/getsentry/sentry/pull/63925/files#diff-30aac95a95c4c9b782c0eb9a15b9d363724bc3c7a0179e1f7b65be80fbe7c7c4R435) to be unconditional in `open_archive_dialog` in #63925 by @isabellaenriquez. If either of you remember any significance to the distinction, let me know.

Otherwise, I'm leaning on the unconditional side, on the assumption that it's harmless for the `callback_id` to have an extra `"rule": None` entry.

isabellaenriquez: @RyanSkonnord I believe this was specific to the block kit feature flag not being GA'd at the time. The resolve dialog function was used in both our legacy alerts and the block kit alerts, but the archive dialog was only used for block kit alerts, so there wasn't a need for this conditional.

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -942,3 +700,222 @@ def handle_member_approval(self, slack_request: SlackActionRequest, action: str)
...
+            )
+            _logger.exception(
+                "slack.action.response-error",
+                extra={
+                    "organization_id": org.id,
+                    "integration_id": slack_request.integration.id,
+                    "trigger_id": slack_request.data["trigger_id"],
+                    "dialog": self.dialog_type,
+                },
+            )
```
GabeVillalobos, line 871: Is there a plan to have our other integrations implement a similar Modal flow as slack? I could see it being useful to wrap and raise these exceptions, letting the caller choose the response to make metric and logging decisions a bit more generic.

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -942,3 +700,222 @@ def handle_member_approval(self, slack_request: SlackActionRequest, action: str)
...
+
+    def get_modal_payload(self, callback_id: str, metadata: str) -> View:
+        return self.build_modal_payload(
+            title="Archive",
+            action_text="Archive",
+            options=ARCHIVE_OPTIONS,
+            initial_option_text="Until escalating",
+            initial_option_value="ignored:archived_until_escalating",
+            callback_id=callback_id,
+            metadata=metadata,
```
GabeVillalobos, line 920: Nice, this seems a lot cleaner

src/sentry/integrations/slack/webhooks/action.py
RyanSkonnord: Thanks! I'll leave it as unconditional.

src/sentry/integrations/slack/webhooks/action.py
RyanSkonnord: No plans as yet, but that sounds like something we'll probably want to do as part of unifying our SLOs.

In another dev branch that's open right now, I'm already running into the problem of wanting to open an interaction recordk, but then not knowing whether the interaction succeeded or failed because exceptions are being caught and swallowed further down in the call stack.

isabellaenriquez (commented):
thanks so much for cleaning this up 🙏🏽  realizing my two comments are from things we had not fixed when the archive dialog was introduced 😅

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -942,3 +700,222 @@ def handle_member_approval(self, slack_request: SlackActionRequest, action: str)
...
+            metrics.incr(
+                SLACK_WEBHOOK_GROUP_ACTIONS_SUCCESS_DATADOG_METRIC,
+                sample_rate=1.0,
+                tags={"type": f"{self.dialog_type}_modal_open"},
+            )
+        except SlackApiError:
+            metrics.incr(
+                SLACK_WEBHOOK_GROUP_ACTIONS_FAILURE_DATADOG_METRIC,
+                sample_rate=1.0,
+                tags={"type": "resolve_modal_open"},
```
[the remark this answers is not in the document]
RyanSkonnord: Yes, thanks for the catch.

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -942,3 +700,222 @@ def handle_member_approval(self, slack_request: SlackActionRequest, action: str)
...
+                    sample_rate=1.0,
+                    tags={"type": "update_modal"},
+                )
+                logging_data = slack_request.get_logging_data()
+                _logger.exception(
+                    "slack.action.update-modal-not-found",
+                    extra={
+                        **logging_data,
+                        "trigger_id": slack_request.data["trigger_id"],
+                        "dialog": "resolve",
```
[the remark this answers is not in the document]
RyanSkonnord: Yes.

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -942,3 +700,222 @@ def handle_member_approval(self, slack_request: SlackActionRequest, action: str)
...
+        # XXX(CEO): the second you make a selection (without hitting Submit) it sends a slightly different request
+        modal_payload = self.get_modal_payload(callback_id, metadata=metadata)
+        slack_client = SlackSdkClient(integration_id=slack_request.integration.id)
+        try:
+            # We need to use the action_ts as the external_id to update the modal
+            # We passed this in control when we sent the loading modal to beat the 3 second timeout
+            external_id = slack_request.get_action_ts()
+
+            if options.get("send-slack-response-from-control-silo"):
+                self._update_modal(slack_client, external_id, modal_payload, slack_request)
```
cathteng, line 848: @iamrajjoshi is this option true by default now?

src/sentry/integrations/slack/webhooks/action.py
```diff
@@ -942,3 +700,222 @@ def handle_member_approval(self, slack_request: SlackActionRequest, action: str)
...
+                "slack.action.response-error",
+                extra={
+                    "organization_id": org.id,
+                    "integration_id": slack_request.integration.id,
+                    "trigger_id": slack_request.data["trigger_id"],
+                    "dialog": self.dialog_type,
+                },
+            )
+
+    def _construct_reply(self, attachment: SlackBody, is_message: bool = False) -> SlackBody:
```
cathteng, line 873: this isn't being used!

cathteng (approved):
beautiful refactoring 🤌

src/sentry/integrations/slack/webhooks/action.py
iamrajjoshi: yes, still need to cleanup the FF

PR #77464, opened 2024-09-13
ref(whats-new): Move the logic to update internal seen state

### LINKED ISSUE #77462: Update the logic to set the 'hasSeen' property in What's new

priscilawebdev:
closes https://github.com/getsentry/sentry/issues/77462

Preview of the new behavior as a video/image can tell more than 1.000 words:

https://github.com/user-attachments/assets/6df5ee62-4ec3-4f53-9704-b5683ee9233a

### DIFF
--- a/static/app/components/sidebar/broadcasts.tsx
+++ b/static/app/components/sidebar/broadcasts.tsx
@@ -1,4 +1,4 @@
-import {Component, Fragment} from 'react';
+import {Component, Fragment, useEffect} from 'react';
 
 import {getAllBroadcasts, markBroadcastsAsSeen} from 'sentry/actionCreators/broadcasts';
 import type {Client} from 'sentry/api';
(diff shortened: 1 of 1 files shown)


static/app/components/sidebar/broadcasts.tsx
```diff
@@ -28,15 +28,13 @@ type Props = CommonSidebarProps & {
 
 type State = {
   broadcasts: Broadcast[];
-  error: boolean;
```
priscilawebdev, line 31: noticed that `error` is not being used, so I removed this too. please let me know if that is ok in this PR

static/app/components/sidebar/broadcasts.tsx
```diff
@@ -28,15 +28,77 @@ type Props = CommonSidebarProps & {
...
+  onResetCounter: () => void;
+  whatIsNewRevampFeature: boolean;
+} & Pick<CommonSidebarProps, 'orientation' | 'collapsed' | 'hidePanel'>) {
+  useEffect(() => {
+    return () => {
+      if (whatIsNewRevampFeature) {
+        onResetCounter();
+      }
+    };
+  }, [onResetCounter, whatIsNewRevampFeature]);
```
priscilawebdev, line 54: this is the only new code! basically if the user has access to the feature flag `organizations:what-is-new`, the `hasSeen` state will be reset on unmount. That is the change. The rest is the same

ArthurKnaus, line 54: Nit: Why extract all this into a separate component and not just do it in `componentWillUnmount`?

ArthurKnaus (approved):

static/app/components/sidebar/broadcasts.tsx
ArthurKnaus: fine for me, but imho we should render a fallback in this case.

static/app/components/sidebar/broadcasts.tsx
ArthurKnaus: nvm, just figured it out. 🙈 We should migrate the class component to a functional one...

priscilawebdev: yes... i have a code prepared 🙂

PR #77469, opened 2024-09-13
feat(hook): Introduce useHasMounted hook

priscilawebdev:
This PR introduces a new hook that we can use in the following codes:

https://github.com/getsentry/sentry/blob/5344837ccba67d641e38ef54ba1254015bf1b7e5/static/app/views/performance/landing/index.tsx#L116
https://github.com/getsentry/sentry/blob/5344837ccba67d641e38ef54ba1254015bf1b7e5/static/app/utils/useNavigate.tsx#L62
and in the yet-to-be-merged PR https://github.com/getsentry/sentry/blob/dbf2b494f7cd7b872ed0b0f7a52052f1556f565c/static/app/components/onboarding/gettingStartedDoc/onboardingLayout.tsx#L65

### DIFF
--- a/static/app/utils/useHasMounted.tsx
+++ b/static/app/utils/useHasMounted.tsx
@@ -1,0 +1,14 @@
+import {useEffect, useRef} from 'react';
+/**
+ * Provides a boolean indicating if the component has completed its initial mount.
+ * @returns true if the component has mounted, false otherwise.
+ */
+export function useHasMounted() {
+  const hasMounted = useRef(false);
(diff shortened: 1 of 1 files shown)


ryan953:
Looks like we already have https://github.com/getsentry/sentry/blob/master/static/app/utils/useIsMountedRef.tsx which is doing the same job.

priscilawebdev:
@ryan953 I was not aware of this hook. thank you 🙏

PR #77470, opened 2024-09-13
feat(link-utils): Add validate link exists function

priscilawebdev:
Contributes to https://github.com/getsentry/sentry/issues/77466

This cannot be a hook as I need to use it inside of a `handleSubmit` function , similar to how we use  [fetchImageData](https://github.com/getsentry/sentry/blob/5344837ccba67d641e38ef54ba1254015bf1b7e5/static/app/utils/url/fetchImageData.ts#L11)


static/app/utils/url/validateLinkExists.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export async function validateLinkExists(url: string): Promise<boolean> {
+  try {
+    const response = await fetch(url, {method: 'HEAD'}); // Using HEAD method to just check if the resource exists
+    if (response.ok) {
+      return true;
+    }
+    return false;
+  } catch (error) {
+    const message = t('Unable to validate ig the link exists');
```
ryan953, line 18:
```suggestion
    const message = t('Unable to validate if the link exists');
```

static/app/utils/url/validateLinkExists.ts
```diff
@@ -0,0 +1,23 @@
...
+    handleXhrErrorResponse(message, error);
+    addErrorMessage(message);
```
ryan953, line 20: We should remove `addErrorMessage` from here so that it can be called only in cases where it's needed.

You mentioned that the first use-case for this util is inside a `handleSubmit` callback, but the hook could also be useful in a situation where, for example, we validate that an off-site link exists before the user clicks into it. In a case like that we wouldn't want to have addErrorMessage called because there's no user interaction yet.

static/app/utils/url/validateLinkExists.ts
```diff
@@ -0,0 +1,20 @@
...
+ * @param url - The URL to check.
+ * @returns A promise that resolves to `true` if the URL exists, or `false` otherwise.
+ */
+export async function validateLinkExists(url: string): Promise<boolean> {
+  try {
+    const response = await fetch(url, {method: 'HEAD'}); // Using HEAD method to just check if the resource exists
+    if (response.ok) {
+      return true;
+    }
+    return false;
```
vgrozdanic, line 15: nit: We can just do:
```
return response.ok
```

vgrozdanic (approved):

static/app/utils/url/validateLinkExists.ts
priscilawebdev: yes I know but I wanted to make it consistent with the return we have inside the catch function :)

static/app/utils/url/validateLinkExists.ts
```diff
@@ -0,0 +1,20 @@
...
+  } catch (error) {
+    handleXhrErrorResponse(t('Unable to validate if the link exists'), error);
```
ArthurKnaus, line 17: Do we really want to create an exception in Sentry in this case?
I would leave this to the calling code.

PR #77476, opened 2024-09-13
ref: fix test pollution integrity errors due to cache leaks between tests

asottile-sentry:
this results in about a 15% failure rate:

```pytest
$ pytest --count 10 tests/sentry/event_manager/test_event_manager_grouping.py -k test_records_hash_comparison_metric
...
```

failing with:

```
E   django.db.utils.IntegrityError: ForeignKeyViolation('insert or update on table "sentry_environmentproject" violates foreign key constraint "sentry_environmentpr_environment_id_f5d02227_fk_sentry_en"\nDETAIL:  Key (environment_id)=(18) is not present in table "sentry_environment".\n')
E   SQL: SET CONSTRAINTS ALL IMMEDIATE
```

I traced this down to a cache leaking from one test to another -- after applying this patch I had no failures in 100x runs of the tests

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,6 +3,7 @@ from collections.abc import MutableMapping
 import psutil
 import pytest
 import responses
+from django.core.cache import cache
 from django.db import connections
 
 from sentry.silo.base import SiloMode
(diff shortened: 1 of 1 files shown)


asottile-sentry:
also: the reason this is flaky is that sometimes the ids collide and sometimes they do not due to something non-deterministic in event classification -- haven't quite figured that part out yet

joshuarli (approved):

PR #77479, opened 2024-09-13
chore: Move deletion tasks to be co-located with the rest of deletions

markstory:
I've moved files and imports only with this change. Celery task names will be aligned with import paths separately.


src/sentry/tasks/files.py
```diff
@@ -5,9 +5,10 @@
 
 from sentry.silo.base import SiloMode
 from sentry.tasks.base import instrumented_task, retry
-from sentry.tasks.deletion.scheduled import MAX_RETRIES
 from sentry.utils.db import atomic_transaction
 
+MAX_RETRIES = 5
```
markstory, line 10: Importing this from the new location was creating import cycles. This was the simplest solution to resolve the cycles.

corps:
👏  My god you're actually doing it.  It's actually happening.

evanpurkhiser (approved):

PR #77482, opened 2024-09-13
feat(rr6): Add reactRouterVersion to 'Route Not Found' errors

evanpurkhiser:


bruno-garcia (approved):
This helps, but actually could we (maybe another PR) change fingerprinting?

How about we finterprint like: `"route-not-found" + scope.transaction` (where transactions should be the parameterized URL)?

We can have alerts then for different teams to handle their own?

PR #77490, opened 2024-09-13
typing(grouping): Add more modules into stronger typing

armenzg:


src/sentry/grouping/strategies/__init__.py
```diff
@@ -1,4 +1,4 @@
-def _import_all():
+def _import_all() -> None:
     # The import order here is important due to cross dependencies
```
asottile-sentry, line 2: terrifying!

asottile-sentry (approved):

PR #77491, opened 2024-09-13
feat(auth): add org level api access

sentaur-athena:
Need organization in these models to be able to create an org level application auth. Here's the spec for more info: https://www.notion.so/sentry/Laravel-Partnership-For-Existing-Orgs-371b6bd1d16c447ea666e533c32e503e?pvs=4#bf4adb9100ca4a33a880279c9c666d89


src/sentry/models/apiapplication.py
```diff
@@ -66,6 +67,20 @@ class ApiApplication(Model):
...
     terms_url = models.URLField(null=True)
 
+    # API applications can be tied to an untrusted organization.
+    #
+    # This represents several security risks. A phishing risk and cross organization access for
+    # the integrator. Before creating an access_token for a user, verify the user has access to
+    # the organization defined on the application. Restrict the access_token to only access the
+    # organization defined on the application. Cross-organization access is not tolerable
+    # regardless of the token's scope.
+    organization_id = HybridCloudForeignKey(
```
markstory, line 77: How will partners create an ApiApplication for each organization? This same schema change was recently proposed for another project, and we found an alternative solution through both sentryapps (which have organization level installations already).

sentaur-athena: @markstory this was a mistake. I didn't mean to add it to application, already done the model changes in api application that I needed to in https://github.com/getsentry/sentry/pull/76974  🤦‍♀️ here's the spec showing the flow and model changes: https://www.notion.so/sentry/Laravel-Partnership-For-Existing-Orgs-371b6bd1d16c447ea666e533c32e503e

Let me update and re-request review.

src/sentry/migrations/0760_add_org_to_api_auth.py
```diff
@@ -0,0 +1,50 @@
...
+        migrations.AddField(
+            model_name="apigrant",
+            name="organization_id",
+            field=sentry.db.models.fields.hybrid_cloud_foreign_key.HybridCloudForeignKey(
+                "sentry.Organization", db_index=True, null=True, on_delete="CASCADE"
+            ),
+        ),
+        migrations.AddField(
+            model_name="apitoken",
+            name="scoping_organization_id",
```
markstory, line 45: You'll also need a similar field on `ApiTokenReplica` so that authentication code will work in regions. Be careful though, the `ApiTokenReplica` already has an `organization` field on it, but that field is used for sentry app flows currently.

sentaur-athena: ok that's very strange. Do you have more context? Why does a replica has a field that the original model doesn't have 😕 

I didn't call it `organization` specifically because ApiToken has a method `organization_id` that refers the organization from the other fields.

sentaur-athena: If I'm understanding correctly, actually this will already set the `organization` on replica so I don't need to add it there: https://github.com/getsentry/sentry/blob/6bd307fc0a5263654d89605eca14343897468d78/src/sentry/auth/services/auth/serial.py#L82

markstory: > ok that's very strange. Do you have more context? Why does a replica has a field that the original model doesn't have 😕

In order for `ApiTokenReplica` to also offer an `organization_id` attribute we synchronize the sentryapp derived organization state into the replica model so that we don't need to use RPC.

My previous comment was that we'll also need a `scoping_organization_id` on `ApiTokenReplica` so that endpoint access can be scoped in the region endpoints.

src/sentry/models/apitoken.py
```diff
@@ -356,6 +365,8 @@ def organization_id(self) -> int | None:
             SentryAppInstallationToken,
         )
 
+        if self.scoping_organization_id:
```
sentaur-athena, line 368: here I'm setting `organization_id` to `scoping_organization_id`

src/sentry/migrations/0760_add_org_to_api_auth.py
sentaur-athena: In region endpoint when want to authenticate based on token, we look at `ApiTokenReplica.organization` and that would be the organization with ID `scoping_organization_id` because of this line: https://github.com/getsentry/sentry/pull/77491/files#r1771716626

I don't want to create another organization_id or scoping_organization_id field because if they become inconsistence we can't decide which one is the source of truth.

markstory (approved):

PR #77496, opened 2024-09-13
drop(grouping): A few more post-hierarchical code removals

armenzg:


src/sentry/grouping/strategies/newstyle.py
```diff
@@ -111,18 +111,6 @@ def get_basename(string: str) -> str:
     return _basename_re.split(string)[-1]
 
 
-def get_package_component(package: str, platform: str | None) -> GroupingComponent:
```
armenzg, line 114: Unused.

Added here:
https://github.com/getsentry/sentry/pull/24346/files#diff-765101e1faf45dddc42ee634ea872031228b3072cd6a28d41c533c3aba7414d0R63

src/sentry/interfaces/stacktrace.py
```diff
@@ -154,7 +154,6 @@ def to_python(cls, data, **kwargs):
             "symbol_addr",
             "trust",
             "vars",
-            "snapshot",
```
armenzg, line 157: This was added to reverse stacktraces which we drop the support for.

Added here:
https://github.com/getsentry/sentry/pull/24346/files#diff-4cc5f7f16f745756dc19e69e113a46b7ec27e615c4323e74914e23a09cd83b37R155

armenzg: CC @jjbayer @untitaker

JoshFerge (approved):

PR #77498, opened 2024-09-13
fix(anomaly detection): get aggregation key from snuba data

mifu67:
We can't assume that the aggregation key of a snuba timeseries data point will be `"count"`, so we should set the aggregation key when we see a data point with an aggregation value.


src/sentry/seer/anomaly_detection/utils.py
```diff
@@ -102,8 +102,21 @@ def format_historical_data(data: SnubaTSResult, dataset: Any) -> list[TimeSeries
             ts_point = TimeSeriesPoint(timestamp=date.timestamp(), value=count)
             formatted_data.append(ts_point)
     else:
+        # we don't know what the aggregation key of the query is
+        # so we should see it when we see a data point that has a value
+        agg_key = ""
```
ceorourke, line 107: I think this looks good, but can we add some tests?

ceorourke (approved):

PR #77501, opened 2024-09-13
chore(users): Move User base endpoint file to users directory

Christinarlong:
Whoops missed this one 🌵 

issue ref(#73856)


src/sentry/api/bases/user.py
Christinarlong, line 1: shim for getsentry refs

markstory (approved):
:clap:

PR #77503, opened 2024-09-13
chore(apidocs): Add example for Retrieve an Organization Member

isabellaenriquez:
https://docs.sentry.io/api/organizations/retrieve-an-organization-member/ is missing an example, this PR adds that + some cleanup


src/sentry/api/endpoints/organization_member/details.py
```diff
@@ -134,7 +135,7 @@ def get(
         """
         Retrieve an organization member's details.
 
-        Will return a pending invite as long as it's already approved.
+        Response will be a pending invite if it has been approved by organization admins but is waiting to be accepted by the invitee.
```
iamrajjoshi, line 138: i am always bad with permissions, can organization managers also approve?

iamrajjoshi (approved):
Bear Lets Merge

src/sentry/api/endpoints/organization_member/details.py
isabellaenriquez: fixed

PR #77509, opened 2024-09-13
feat(replays): Pack video and recording data into the same payload

cmanallen:
Cost cutting optimization required before GA.

Web payloads are not packed.  Video payloads are packed in the following format:

1. One byte identifier (`1`).
2. 32-bit integer denoting the videos size in bytes.
3. The video.
4. The recording data.

This is a contiguous sequence of uncompressed bytes.  The bytes are compressed after packing and before storage.

Note:

- An endpoint changed ever so slightly because the return type of a function changed.
- We stopped emitting quite a few spans.
    - I couldn't find them in Sentry so I assume they're broken.
    - No one looks at them any way so there's no point in debugging.
- The packing algorithm is somewhat inconsequential but can be found in `pack.py`.
- All of the API endpoints use the `reader.py` file that's where the read changes can be found.
- Both consumers (buffered and not) had open heart surgery.  The diff is not easy to read but the changes are relatively small.

Closes: https://github.com/getsentry/team-replay/issues/476


src/sentry/replays/usecases/pack.py
```diff
@@ -0,0 +1,53 @@
...
+from enum import Enum
+
+
+class Encoding(Enum):
+    RRWEB = 0
+    VIDEO = 1
+
+
+def pack(rrweb: bytes, video: bytes | None) -> bytes:
+    def _to_u64_bytes(length):
```
bruno-garcia, line 28: max value of a 32 bit would be 2.4 GB video, right? In that case, would that be enough instead of 64?

cmanallen: Yeah we could.  Could probably get away with a 24-bit integer too.

bruno-garcia: if we're going unsigned integer (makes sense) and it's the max size of a single segment, indeed could be quite small. Anyway no relevant gains, it's just a single int 😅

cmanallen:
I'm thinking about splitting the reader and writer changes.  Deploying the read first to test my assumption is correct that unpacking is safe for old events.  And finally deploying the write changes to take advantage of the optimization.

JoshFerge:
> I'm thinking about splitting the reader and writer changes. Deploying the read first to test my assumption is correct that unpacking is safe for old events. And finally deploying the write changes to take advantage of the optimization.

this makes sense to me. you could always make an option to deploy with and revert if anything goes awry.

cmanallen:
@JoshFerge That's the better take.  This needs to be behind an option regardless of the number of PRs.  I'll update.

JoshFerge (approved):

PR #77510, opened 2024-09-13
feat(eap): Use RPC to fetch tag values

Zylphrex:
This uses RPC calls to fetch tag values from eap spans.

depends on getsentry/snuba#6304


src/sentry/api/endpoints/organization_spans_fields.py
```diff
@@ -181,6 +183,67 @@ def get(self, request: Request, organization, key: str) -> Response:
...
+            end_timestamp.FromDatetime(
+                snuba_params.end_date.replace(hour=0, minute=0, second=0, microsecond=0)
+                + timedelta(days=1)
+            )
+
+            rpc_request = AttributeValuesRequest(
+                meta=RequestMeta(
+                    organization_id=organization.id,
+                    cogs_category="performance",
+                    referrer=Referrer.API_SPANS_TAG_KEYS.value,
```
wmak, line 209: nit: should we use a different referrer than the usual query so its easier to differentiate

src/sentry/api/endpoints/organization_spans_fields.py
```diff
@@ -181,6 +183,67 @@ def get(self, request: Request, organization, key: str) -> Response:
...
+
+            paginator = ChainPaginator(
+                [
+                    [
+                        TagValue(
+                            key=key,
+                            value=tag_value,
+                            times_seen=None,
+                            first_seen=None,
+                            last_seen=None,
```
wmak, line 230: Are these all None cause its not something we can query yet? if so could we add a TODO comment

wmak (approved):
nit: around referrer name, comments, lgtm otherwise

PR #77511, opened 2024-09-13
ref: add warmup to wsgi

asottile-sentry:
this less-lazily instantiates the django app such that the first request doesn't

### DIFF
--- a/src/sentry/wsgi.py
+++ b/src/sentry/wsgi.py
@@ -1,3 +1,4 @@
+import io
 import os.path
 import sys
 
@@ -16,3 +17,16 @@ from django.core.handlers.wsgi import WSGIHandler
 
 # Run WSGI handler for the application
 application = WSGIHandler()
+
+# trigger a warmup of the application
+application(
+    {
+        "PATH_INFO": "/_health/",
+        "REQUEST_METHOD": "GET",
+        "SERVER_NAME": "127.0.0.1",
+        "SERVER_PORT": "9001",
+        "wsgi.input": io.BytesIO(),
+        "wsgi.url_scheme": "https",
+    },
+    lambda status, response_headers, exc_info=None: lambda bts: None,
+)


src/sentry/wsgi.py
```diff
@@ -16,3 +17,16 @@
 
 # Run WSGI handler for the application
 application = WSGIHandler()
+
+# trigger a warmup of the application
+application(
+    {
+        "PATH_INFO": "/_health/",
```
asottile-sentry, line 24: ideally this would be a specific `/_warmup` view or something (which at minimum gets ~4 more modules warmed) but this is a good start at least.  with a dedicated warmup view we could probably tune this better (and do more stuff like set up memcached connections / whatever else)

joshuarli (approved):

JoshFerge (approved):
🧑‍🍳

src/sentry/wsgi.py
asottile-sentry: fwiw these are the 4 additional modules that get imported by running a "real" view (crossdomain.xml):
```diff
$ diff -u <(jq . mods2.json) <(jq . mods.json)
--- /dev/fd/63	2024-09-13 16:41:26
+++ /dev/fd/62	2024-09-13 16:41:26
@@ -655,6 +655,9 @@
   "django.contrib.messages.middleware",
   "django.contrib.messages.storage",
   "django.contrib.messages.storage.base",
+  "django.contrib.messages.storage.cookie",
+  "django.contrib.messages.storage.fallback",
+  "django.contrib.messages.storage.session",
   "django.contrib.messages.utils",
   "django.contrib.postgres",
   "django.contrib.postgres.constraints",
@@ -682,6 +685,7 @@
   "django.contrib.sessions.exceptions",
   "django.contrib.sessions.middleware",
   "django.contrib.sessions.models",
+  "django.contrib.sessions.serializers",
   "django.contrib.sites",
   "django.contrib.sites.apps",
   "django.contrib.sites.checks",
```

PR #77515, opened 2024-09-13
ref(feedback): disallow shimming reports w/missing event

aliu39:
~~Discovered there's an unreachable condition in shim_to_feedback (event=None). We currently never call this function w/event=None, but the code supports it and I think it's useful to keep.~~
~~Also adds unit tests for shim_to_feedback.~~

Discovered there's an unreachable condition in shim_to_feedback (event=None). We currently never call this function w/event=None. Adds unit tests to assert this case fails to create feedback.


JoshFerge (commented):
we used to use this but found when you create user reports with no environment, etc. it does not work well. i'd advocate for simply removing this codepath

aliu39:
> when you create user reports with no environment, etc. it does not work well.
Do you mean "feedbacks with no environment"?

aliu39:
>  i'd advocate for simply removing this codepath

Ok, fine by me

JoshFerge:
> > when you create user reports with no environment, etc. it does not work well.
> 
> Do you mean "feedbacks with no environment"?

yeah, so if there's no event attached to the user report we end up losing the environment

tests/sentry/feedback/usecases/test_create_feedback.py
```diff
@@ -768,3 +770,47 @@ def test_create_feedback_spam_detection_set_status_ignored(
...
         assert group.substatus == GroupSubStatus.FOREVER
+
+
+# Unit tests for shim_to_feedback error cases. The typical behavior of this function is tested in
+# test_project_user_reports, test_post_process, and test_update_user_reports.
+
+
+@django_db_all
+def test_shim_to_feedback_missing_event(default_project, monkeypatch):
+    # Not allowing this since creating feedbacks with no environment doesn't work well.
```
michellewzhang, line 781:
```suggestion
    # Not allowing this since creating feedbacks with no event doesn't work well.
```
did you mean this?

tests/sentry/feedback/usecases/test_create_feedback.py
```diff
@@ -768,3 +770,47 @@ def test_create_feedback_spam_detection_set_status_ignored(
...
+        "sentry.feedback.usecases.create_feedback.create_feedback_issue", mock_create_feedback_issue
+    )
+    report_dict = {
+        "name": "andrew",
+        "email": "aliu@example.com",
+        "comments": "Shim this",
+        "event_id": "a" * 32,
+        "level": "error",
+    }
+    shim_to_feedback(
```
michellewzhang, line 793: looks like the tests caught this but i think the `shim_to_feedback` function requires an `event` to be passed in -- might need to fix the typing?

aliu39: Hm weird it's warning about type but not in the test below. I guess we could remove both, since they both test params different from the annotated types.

aliu39: What do you think? @michellewzhang

tests/sentry/feedback/usecases/test_create_feedback.py
aliu39: Updated the comment

tests/sentry/feedback/usecases/test_create_feedback.py
michellewzhang: discussed offline - we're going to try to ignore `mypy`, since we want to keep the tests to assert that the behavior is as expected when the stuff passed in is junk

michellewzhang (approved):

PR #77518, opened 2024-09-13
feat(uptime): Add fetch_subscription_rdap_info task

evanpurkhiser:
Adds the task that will actually fetch rdap details for the subscription

Will retry 5x on exception, this should help us avoid flakey API calls.


src/sentry/uptime/rdap/query.py
evanpurkhiser, line 1: Moved these into it's own module. Didn't feel right to add the task into the detector tasks.

wedamija (approved):

Issue #73857, opened 2024-07-04
App Modules - Consolidate SentryApps resources


markstory:
SentryApps and its direct relations are control_silo scoped resources. Currently the components that comprise the control silo aspects of sentryapps are spread throughout the repository. Consolidating those classes and functions into `sentry.sentry_apps` will help clarify the application's design and make running subsets of tests easier and faster.

- Replace usage of `Mediator` with dataclasses that implement the same approximate API as the mediators do today. Mediators duplicate the role of serializers and don't have enough adoption to justify continued inclusion in the repository.
- Move endpoints, serializers, models, and tasks to be co-located.

We'll need to provide import shims for any symbols used in getsentry until those imports can be updated separately.

cc @Christinarlong

markstory:
Closing this as we've finished the control silo consolidation work for sentryapps.

PR #77521, opened 2024-09-13
ref(sentry_apps): move sentryapp model to sentry_apps dir

Christinarlong:
Let's start migrating Sentry App resources starting witht he SentryApp model & test !

ref(#73857)
relies on (#77614) & (https://github.com/getsentry/getsentry/pull/15146)


markstory (approved):
Looking good.

src/sentry/sentry_apps/apps.py
Christinarlong, line 1: delete apps.py shim so that we won't get circular import errors from django 🌵

corps:
Literally hero level work, thank you.

markstory (approved):

PR #77524, opened 2024-09-13
feat(onboarding): Avoid `new URL` for onboarding app routing

scttcper:
using `new URL` adds the host to the path which isn't router friendly since the router expects `/org/example/` and not `https://sentry.io/org/example/`

Adds the ability for navigateTo to accept path as an object

### DIFF
--- a/static/app/actionCreators/navigation.spec.tsx
+++ b/static/app/actionCreators/navigation.spec.tsx
@@ -34,19 +34,17 @@ describe('navigation ActionCreator', () => {
 
   it('should get project from query parameters', () => {
     router.location.query.project = '2';
-    expect(navigateTo('/settings/:projectId/alert', router)).toBe(undefined);
+    navigateTo('/settings/:projectId/alert', router);
     expect(openModal).not.toHaveBeenCalled();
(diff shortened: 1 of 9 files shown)


static/app/actionCreators/navigation.spec.tsx
```diff
@@ -120,4 +120,22 @@ describe('navigation ActionCreator', () => {
...
+    router.location.query.project = '2';
+    expect(
+      navigateTo(
+        {
+          pathname: '/settings/:projectId/alert?project=:project#performance-sidequest',
+          query: {referrer: 'onboarding_task'},
+        },
+        router
+      )
+    ).toBe(undefined);
```
priscilawebdev, line 134: `navigateTo` is a `void` function, so it is always `undefined`. Do we need to have these checks?

priscilawebdev (approved):
code looks good to me

static/app/actionCreators/navigation.spec.tsx
scttcper: i'm not sure why that's there i did copy and paste this from another test

PR #77525, opened 2024-09-13
feat(custom-views): Don't show tab dividers next to hovering tabs 

MichaelSun48:
Tab dividers around the hovering tab now disappear. This mimics the behavior of our dropdown menus. Unfortunately we do have to add extra state for this. 

https://github.com/user-attachments/assets/6d1cb50f-99e6-458a-9364-531d60640e63

### DIFF
--- a/static/app/components/draggableTabs/draggableTabList.tsx
+++ b/static/app/components/draggableTabs/draggableTabList.tsx
@@ -1,6 +1,7 @@
 import {
   type Dispatch,
   Fragment,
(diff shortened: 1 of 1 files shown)


malwilley (approved):
You should be able to do this in CSS too:

```css
&:hover + div {
  opacity: 0;
}
```

Edit: Oh I see you also need  the previous sibling, which is  a little more work. But that should also be technically doable with css

PR #77528, opened 2024-09-13
feat(invite-members): new invite modal design

ameliahsu:
changes to note:
- added ability to copy/paste lists of emails
- removed the "Add another" button
- new submission flow (successfully sent emails are removed and the inputs are no longer disabled after submission, so users can continue to input emails without needing to select "Send more invites")
- slight change to the success/error messages
- refactored invite modal code

next change will be to stop overflow of email address input (I want to do that in a separate PR since the `SelectControl` component is widely used in other places)

**before:**

https://github.com/user-attachments/assets/daf4a7bb-1e5d-4d0b-bfb7-3fbf3469cf3a

**after:**

https://github.com/user-attachments/assets/0d037342-9d11-45a8-abd8-4146599269ca


static/app/components/modals/inviteMembersModal/inviteMembersModalNew.tsx
```diff
@@ -0,0 +1,168 @@
...
+            emails={[...emails]}
+            role={role}
+            teams={[...teams]}
+            roleOptions={member?.orgRoleList ?? ORG_ROLES}
+            roleDisabledUnallowed={willInvite}
+            inviteStatus={inviteStatus}
+            onRemove={reset}
+            onChangeEmails={opts => setEmails(opts?.map(v => v.value) ?? [], 0)}
+            onChangeRole={value => setRole(value?.value, 0)}
+            onChangeTeams={opts => setTeams(opts ? opts.map(v => v.value) : [], 0)}
```
ameliahsu, line 104: I eventually want to update `setEmails`, `setRole`, and `setTeams` and their corresponding states so that they're not a list, but they were written this way because the current invite modal allows you to add additional invite rows.

static/app/components/modals/inviteMembersModal/inviteMembersModalNew.tsx
```diff
@@ -0,0 +1,168 @@
...
+          />
+        </Rows>
+
+        <Footer>
+          <FooterContent>
+            <div>
+              <InviteStatusMessage
+                complete={complete}
+                hasDuplicateEmails={false}
```
ameliahsu, line 113: We can remove this `hasDuplicateEmails` prop once we stop using the old invite modal, since there will no longer be multiple invite rows and thus no possibility for duplicate emails.

static/app/components/modals/inviteMembersModal/index.tsx
```diff
@@ -70,7 +73,28 @@ function InviteMembersModal({
...
+              member={memberResult.data}
+              pendingInvites={pendingInvites}
+              reset={reset}
+              sendingInvites={sendingInvites}
+              sendInvites={sendInvites}
+              setEmails={setEmails}
+              setRole={setRole}
+              setTeams={setTeams}
+              willInvite={willInvite}
+              error={error}
```
saponifi3d, line 95: Rather than passing everything into this component as a property, could we create a context to store all of these functions and variables? That way any component inside of the `InviteMembersModalNew` could access them without passing it from this level all the way down (this is an anti-pattern called [Prop Drilling](https://www.freecodecamp.org/news/prop-drilling-in-react-explained-with-examples/)).

If we used a react context to store these functions / states, the interface to this modal could be limited to what the modal cares about.

Next, we could look at the properties of a modal, for example `Header` and `Footer`. Those items in particular are more like children of the "Modal" component, and can be composable items that are namespaced to the modal. (Here's an example from a popular ui library: https://www.radix-ui.com/primitives/docs/components/dropdown-menu and here's a blog post that dives into what a namespaced component is: https://medium.com/@kunukn_95852/react-components-with-namespace-f3d169feaf91) 

Another way we could design this interface is:
```jsx
<InviteModalContext.Provider
  value={{
    canSend={canSend}
    invites={invites}
    inviteStatus={inviteStatus}
    {/* copy all the props */}
  }}
>
  <InviteMembersModalNew>
     <InviteMembersModalNew.Header>{Header}</InviteMembersModalNew.Header>
        {/*
            what would be extra cool is to pull the
            content of the modal to here, passing it as the
            'children' variable to the component. so you can
            see all the rows etc, if you did that you
            may not even need to use a context
            provider 🤔
         */}
     <InviteMembersModalNew.Footer>{Footer}</InviteMembersModalNew.Footer>
  </InviteMembersModalNew>
</InviteModalContext.Provider>
```
^^ it's looking a lot more like HTML! that makes it a lot easier for us to understand what's happening in the component at a visual level. "I have a modal, with this header, this content, and this footer" - then inside of the modal, it will understand how to open / close (for now we can leave the invite stuff to the internals of the modal since it's specifically the invite members modal).

----

Thiiiiiisss is also a fairly large refactor. so only take on this feedback if you have the time -- happy to help with anything if you do decide to take on this endeavor though!

static/app/components/modals/inviteMembersModal/inviteRowControlNew.spec.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+describe('InviteRowControlNew', function () {
+  const mockOnChangeEmails = jest.fn();
+  const mockOnChangeRole = jest.fn();
+  const mockOnChangeTeams = jest.fn();
+
+  const teamData = [
+    {
+      id: '1',
+      slug: 'moo-deng',
```
saponifi3d, line 17: 🦛

static/app/components/modals/inviteMembersModal/inviteRowControlNew.spec.tsx
```diff
@@ -0,0 +1,103 @@
...
+  it('updates email addresses when new emails are inputted', async function () {
+    render(<InviteRowControlNew {...rowControlProps} />);
+    const emailInput = screen.getByLabelText('Email Addresses');
+    await userEvent.type(emailInput, 'test-space@example.com ');
+    await userEvent.type(emailInput, 'test-comma@example.com,');
+    await userEvent.type(emailInput, 'test-newline@example.com{enter}');
+    await userEvent.type(emailInput, 'test-unfocus@example.com');
+    await userEvent.tab();
+    expect(mockOnChangeEmails).toHaveBeenCalledTimes(4);
+  });
```
saponifi3d, line 78: In the future this test could be a bit hard to debug why it broke. the reason for that is because we are testing multiple cases in 1 expectation. (for example, if the `{enter}` stopped working for some reason, and we had to debug why - this test would fail. but we wouldn't know which case caused the error - `,` `{enter}`, `tab` and so on).

So a better way to test this would be to have something reaallly verbose like these:
```
 it('updates email addresses when new emails are inputted with a space', async function () {
    render(<InviteRowControlNew {...rowControlProps} />);
    const emailInput = screen.getByLabelText('Email Addresses');
    await userEvent.type(emailInput, 'test-space@example.com ');
    expect(mockOnChangeEmails).toHaveBeenCalledTimes(1);
    
    await userEvent.type(emailInput, 'test-space@example.com ');
    expect(mockOnChangeEmails).toHaveBeenCalledTimes(2);
  });
  
   it('updates email addresses when new emails are inputted with a enter', async function () {
    render(<InviteRowControlNew {...rowControlProps} />);
    const emailInput = screen.getByLabelText('Email Addresses');
    await userEvent.type(emailInput, 'test-space@example.com{enter}');
    expect(mockOnChangeEmails).toHaveBeenCalledTimes(1);
    
    await userEvent.type(emailInput, 'test-space@example.com{enter}');
    expect(mockOnChangeEmails).toHaveBeenCalledTimes(2);
  });
```

But that's a lot to maintain, especially for all the cases to cover. so jest introduced [`each`](https://jestjs.io/docs/api#describeeachtablename-fn-timeout) which helps write these very repetitive tests. 
```js
describe('email input tests', () => {
  describe.each([
    {email: 'test-space@example.com', delimiter: ' '},
    {email: 'test-comma@example.com', delimiter: ','},
    {email: 'test-newline@example.com', delimiter: '{enter}'},
    /* note the tab case doesn't work as it's a special case from the others */
  ])('updates email addresses when new emails are inputted with '$delimiter', ({email, delimiter}) => {
    it(`invokes the mock correctly with one`, () => {
      const emailInput = screen.getByLabelText('Email Addresses');
      await userEvent.type(emailInput, `${email}${delimiter}`);
      expect(mockOnChangeEmails).toHaveBeenCalledTimes(1);
    });
  
    it(`invokes the mock correctly with many`, () => {
      const emailInput = screen.getByLabelText('Email Addresses');
      await userEvent.type(emailInput, `${email}${delimiter}`);
      await userEvent.type(emailInput, `${email}${delimiter}`);
      await userEvent.type(emailInput, `${email}${delimiter}`);
      expect(mockOnChangeEmails).toHaveBeenCalledTimes(3);
    });
  });
});
```
This final way allows you to have the coverage of all of those other tests being written out, but the simplicity of writing it once. (Ironically, i think you'll still want to keep something like the original case to ensure a mix of cases can happen... for that you could even build out the above example more to include features like `tab` etc and built into the order of an array of delimiters.

saponifi3d (commented):
lemme know if you want to dive into that larger refactor or not. if not, i can delete that comment and we can dive into a little more specific of a review -- but i figured it'd at least be nice to see where you could change it up to make it a bit more composable. 🎉

static/app/components/modals/inviteMembersModal/inviteStatusMessage.tsx
```diff
@@ -39,12 +43,45 @@ export default function InviteStatusMessage({
...
+      const invites = (
+        <strong data-test-id="sent-invites">
+          {tn('%s invite', '%s invites', sentCount)}
+        </strong>
+      );
+      const failedInvites = (
+        <strong data-test-id="failed-invites">
+          {tn('%s invite', '%s invites', errorCount)}
+        </strong>
+      );
```
saponifi3d, line 55: Nit: recommend pulling this out as a `InviteCount` component with 2 props `isError` and `count`. Also, we may want to use CSS rather than the `strong` tag -- can just set the font weight instead.

static/app/components/modals/inviteMembersModal/inviteStatusMessage.tsx
```diff
@@ -57,12 +94,45 @@ export default function InviteStatusMessage({
...
-      <strong>{tn('%s invite request', '%s invite requests', sentCount)}</strong>
+      <strong data-test-id="sent-invite-requests">
+        {tn('%s invite request', '%s invite requests', sentCount)}
+      </strong>
+    );
+    const failedInviteRequests = (
+      <strong data-test-id="failed-invite-requests">
+        {tn('%s invite request', '%s invite requests', errorCount)}
+      </strong>
     );
```
saponifi3d, line 105: can replace these with teh `InviteCount` component as well.

static/app/components/modals/inviteMembersModal/inviteStatusMessage.tsx
```diff
@@ -57,12 +94,45 @@ export default function InviteStatusMessage({
...
     const tctComponents = {
       inviteRequests,
       failed: errorCount,
```
saponifi3d, line 108: if you made the shared component, it'd look something like:
```
failed: <InviteCount isError count={errorCount} />
```

static/app/components/modals/inviteMembersModal/inviteStatusMessage.tsx
```diff
@@ -57,12 +94,45 @@ export default function InviteStatusMessage({
...
+      failedInviteRequests,
```
saponifi3d, line 109: 🤔 is there a difference other than the test id between this and the `failedInvites` component?

static/app/components/modals/inviteMembersModal/inviteStatusMessage.tsx
```diff
@@ -39,12 +43,45 @@ export default function InviteStatusMessage({
...
+          {tn('%s invite', '%s invites', errorCount)}
+        </strong>
+      );
       const tctComponents = {
         invites,
         failed: errorCount,
+        failedInvites,
       };
 
+      if (isNewInviteModal) {
```
saponifi3d, line 62: a god rule of thumb in react code, is that if you're returning some JSX, that it can likely be broken down into a component. so tihs the status, etc could likely be pulled out.

static/app/components/modals/inviteMembersModal/inviteStatusMessage.tsx
```diff
@@ -57,12 +94,45 @@ export default function InviteStatusMessage({
...
     }
     const inviteRequests = (
-      <strong>{tn('%s invite request', '%s invite requests', sentCount)}</strong>
+      <strong data-test-id="sent-invite-requests">
+        {tn('%s invite request', '%s invite requests', sentCount)}
+      </strong>
+    );
+    const failedInviteRequests = (
+      <strong data-test-id="failed-invite-requests">
+        {tn('%s invite request', '%s invite requests', errorCount)}
```
saponifi3d, line 103: how important is the `request` copy? it seems like that's adding quite a bit of replicated code, i wonder if we could either set the `'%s invite requests'` as a prop or just remove `requests` and re-use the same `tctComponents` variable.

static/app/components/modals/inviteMembersModal/inviteStatusMessage.tsx
ameliahsu: no 😞 fixed

static/app/components/modals/inviteMembersModal/renderEmailValue.tsx
```diff
@@ -25,7 +27,9 @@ function renderEmailValue<Option extends OptionTypeBase>(
           {children}
           {!status.sent && !status.error && <SendingIndicator />}
           {status.error && <IconWarning legacySize="10px" />}
-          {status.sent && <IconCheckmark legacySize="10px" color="success" />}
+          {status.sent && !organization.features.includes('invite-members-new-modal') && (
+            <IconCheckmark legacySize="10px" color="success" />
```
ameliahsu, line 31: this is for an edge case where a user successfully invites and email and then inputs the same email again: we don't want a checkmark to appear suggesting that the email has successfully been sent, because when the user clicks "send" the second time the invite will fail to send (since the member already exists)

static/app/components/modals/inviteMembersModal/inviteRowControlNew.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const isTeamRolesAllowed = isTeamRolesAllowedForRole(role);
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLElement>) => {
+    switch (event.key) {
+      case 'Enter':
+      case ',':
+      case ' ':
+        handleInput(inputValue);
+        setInputValue('');
+        event.preventDefault();
```
saponifi3d, line 75: nit: recommend `event` being `e` or `evt` (window.event is a reserved keyword in js, it's the latest event triggered) - ES6 scoping should correctly encapsulate the variable so it should be okay.

Oh, and it's generally recommended to have `preventDefault` as the first thing in a handler if you want to prevent it. (the idea being if there's an error in `setInputValue`, then the default handler would still be triggered)

static/app/components/modals/inviteMembersModal/renderEmailValue.tsx
saponifi3d: 🤔 i wonder if there's something we should be looking at like request state here. something like: "If it's still pending form submission, then ..."

saponifi3d (approved):
this looks awesome!! 🎉 thanks so much for the refactoring, cleanups, and willingness to dive into this!

static/app/components/modals/inviteMembersModal/inviteMembersFooter.tsx
```diff
@@ -0,0 +1,80 @@
...
+          willInvite={willInvite}
+        />
+      </div>
+      <ButtonBar gap={1}>
+        <Fragment>
+          <InviteButton
+            invites={invites}
+            willInvite={willInvite}
+            size="sm"
+            data-test-id="send-invites"
```
evanpurkhiser, line 59: You want to use `data-test-id` as a last resort. Here your test selector should be `getByRole('button', {name: 'Send (50) Invites'})` (or whatever the label is).

The reason you want to write tests like this is that it's testing what the user **is actually seeing and doing**. vs when you have data test ids, the user doesn't click buttons based on those. So the label could totally break and the user wouldn't be able to use this anymore, but the test would still pass.

More reading about this: https://testing-library.com/docs/dom-testing-library/api-accessibility/

static/app/components/modals/inviteMembersModal/inviteMembersFooter.tsx
```diff
@@ -0,0 +1,80 @@
...
+          data-test-id="invite-status-message"
+          complete={complete}
+          hasDuplicateEmails={false}
+          inviteStatus={inviteStatus}
+          sendingInvites={sendingInvites}
+          willInvite={willInvite}
+        />
+      </div>
+      <ButtonBar gap={1}>
+        <Fragment>
```
evanpurkhiser, line 54: This fragment is extranious

static/app/components/modals/inviteMembersModal/inviteMembersFooter.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <InviteStatusMessage
+          data-test-id="invite-status-message"
+          complete={complete}
+          hasDuplicateEmails={false}
+          inviteStatus={inviteStatus}
+          sendingInvites={sendingInvites}
+          willInvite={willInvite}
+        />
+      </div>
+      <ButtonBar gap={1}>
```
evanpurkhiser, line 53: I don't think you need this button bar. There's only one button in here haha.

`ButtonBar` good for when there are multiple

static/app/components/modals/inviteMembersModal/inviteMembersFooter.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <InviteButton
+            invites={invites}
+            willInvite={willInvite}
+            size="sm"
+            data-test-id="send-invites"
+            priority="primary"
+            disabled={!canSend || !isValidInvites}
+            onClick={() => {
+              organization.features.includes('invite-members-new-modal') &&
+                removeSentInvites();
```
evanpurkhiser, line 64: minor nit, and maybe this was copy, but this might be a bit easier to read as a normal if statement

```
if (organization.features.includes('invite-members-new-modal')) {
  removeSentInvites();
}
```

static/app/components/modals/inviteMembersModal/inviteRowControlNew.spec.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  describe.each([
+    {email: 'test-space@example.com', delimiter: ' '},
+    {email: 'test-comma@example.com', delimiter: ','},
+    {email: 'test-newline@example.com', delimiter: '{enter}'},
+  ])('updates email addresses when new emails are inputted', ({email, delimiter}) => {
+    it(`invokes the mock correctly with one using delimiter "${delimiter}"`, async () => {
+      const mockSetEmails = jest.fn();
+      render(getComponent({...defaultInviteProps, setEmails: mockSetEmails}));
+      const emailInput = screen.getByLabelText('Email Addresses');
```
evanpurkhiser, line 71: you wanted `getByRole('textarea', {name: 'Email Addresses')` since that is scoping it explicitly to only selecting input boxes

static/app/components/modals/inviteMembersModal/inviteRowControlNew.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const [inputValue, setInputValue] = useState('');
+
+  const theme = useTheme();
+
+  const isTeamRolesAllowedForRole = useCallback<(roleId: string) => boolean>(
+    roleId => {
+      const roleOptionsMap = roleOptions.reduce(
+        (rolesMap, roleOption) => ({...rolesMap, [roleOption.id]: roleOption}),
+        {}
+      );
```
evanpurkhiser, line 61: The typing here is probably going to come out as `{}` the "object" type (which is almost never what you want).

You would want to actually type this like

```
const roleOptionsMap = roleOptions.reduce<ExpectedOutputType>(
```

static/app/components/modals/inviteMembersModal/inviteRowControlNew.tsx
```diff
@@ -0,0 +1,227 @@
...
+            creatable
+            clearable
+            onClear={reset}
+            menuIsOpen={false}
+          />
+        </EmailWrapper>
+      </div>
+      <RoleTeamWrapper>
+        <div>
+          <Heading>Role</Heading>
```
evanpurkhiser, line 125: Missing translation

static/app/components/modals/inviteMembersModal/inviteRowControlNew.tsx
```diff
@@ -0,0 +1,227 @@
...
+        ...mapToOptions(emails),
+        ...newEmails.map(email => ({label: email, value: email})),
+      ]);
+    }
+  };
+
+  return (
+    <RowWrapper>
+      <div>
+        <Heading>Email addresses</Heading>
```
evanpurkhiser, line 95: Missing translation

static/app/components/modals/inviteMembersModal/inviteRowControlNew.tsx
```diff
@@ -0,0 +1,227 @@
...
+        <EmailWrapper>
+          <SelectControl
+            aria-label={t('Email Addresses')}
+            data-test-id="select-emails"
```
evanpurkhiser, line 99: Avoid `data-test-id` for pretty much all form controls. You can see how to select these in this test https://github.com/getsentry/sentry/blob/master/static/app/components/forms/fields/accessibility.spec.tsx

static/app/components/modals/inviteMembersModal/inviteRowControlNew.tsx
evanpurkhiser, line 95: These heading elements should be `label`'s instead of divs, and then you can use `htmlFor` to connect the label to the control. Then the control wouldn't need the `aria-label` anymore.

https://medium.com/@amirakhaled2027/decoding-the-mystery-of-htmlfor-label-association-db775a61f13e

static/app/components/modals/inviteMembersModal/inviteMembersFooter.tsx
ameliahsu: Here I am testing to make sure that the correct status message is shown, so this text doesn't have a role. Would `getByText` the best way to check this, since there is no role?

static/app/components/modals/inviteMembersModal/inviteRowControlNew.tsx
ameliahsu: I think we still need `aria-label` in order to write tests, for example
`screen.getByRole('textbox', {name: 'Email Addresses'})`

If I leave out the `aria-label` it looks like the `textbox` doesn't have a name? Not sure if I'm missing something

PR #77530, opened 2024-09-13
feat(issues): Add inline preview to attachments, refactor attachments

### LINKED ISSUE #77044: Display _preview_ on the Attachments tab on issues
Came up during a customer call*:

We have attachment preview on the issue detail for some types, such as images:

Image

But unfortunately we don't have that at the issue level:

Image

It has a preview for Screenshots, but that's based on the same heuristics we use to show the screenshot preview. Which is the file name:

Image

So if a user has more than 1 attachment with different names, they can't preview them.

scttcper:
- Adds an inline preview to the attachments table
- Remove old deleted state where it would strikeout the deleted item, now just removes it.
- Reorder attachment buttons and add tooltips if it is an icon without text
- Moves attachment layout to parent component since we'll be using this component in a drawer soon.

fixes #77044

before:

Screenshot 2024-09-16 at 11 29 01 AM

after w/ preview open:
Screenshot 2024-09-16 at 11 28 48 AM

other little fixes like no tooltips on icon buttons
image


static/app/components/events/eventAttachments.tsx
```diff
@@ -30,73 +26,13 @@ type EventAttachmentsProps = {
 
 type AttachmentPreviewOpenMap = Record<string, boolean>;
 
-interface InlineAttachmentsProps
-  extends Pick<EventAttachmentsProps, 'event' | 'projectSlug'> {
-  attachment: IssueAttachment;
-  attachmentPreviews: AttachmentPreviewOpenMap;
-}
-
-const getInlineAttachmentRenderer = (attachment: IssueAttachment) => {
```
scttcper, line 39: moved to another file to export

malwilley (approved):
Beautiful

PR #77531, opened 2024-09-13
feat(analytics): add a debounced trackAnalytics util/hook

aliu39:
Sample use case would be analytics in hydration error diff modal. Or in a useEffect dependent on async query results


aliu39:
Closing since we don't have an immediate usecase rn

PR #77532, opened 2024-09-13
feat(flags): attach flag evaluations in sentry's frontend to events

### LINKED ISSUE #77446: Attach FF in Sentry's FE to events using `setContexts("flags")`
- [ ] Collect feature flag evaluations into an in-memory data structure.
- [ ] On error put the feature flags onto the context object.
    - `{"context": "flags": {"values": [{...flags...}]}}`

Don't stress too much about the data-structure.  This is all temporary code.  I used a circular buffer.  On new key I would look for it in a hashmap before inserting into the buffer.  If it existed do nothing.  If it did not exist add it and delete the old entry at the buffer's current index position.

You can opt for a simpler approach.  If you have ready access to an LRU cache that would be a perfectly acceptable outcome.

Prior art: https://github.com/getsentry/sentry/pull/77092

michellewzhang:
- set the `event.contexts.flags` value to be the list of recently evaluated feature flags on sentry JS frontend.
- default buffer size is 10; we can increase this at any time to match the python SDK
- data structure used is a FIFO queue -- when new flags are added, they are appended to the end of the queue, and the oldest one gets kicked out.
- we are removing duplicate flags automatically, so the queue is a unique set of flags.
- closes https://github.com/getsentry/sentry/issues/77446

### DIFF
--- a/static/app/utils/featureFlagOverrides.ts
+++ b/static/app/utils/featureFlagOverrides.ts
@@ -1,3 +1,4 @@
+import type {Flags} from 'sentry/types/event';
 import type {Organization} from 'sentry/types/organization';
 import localStorageWrapper from 'sentry/utils/localStorage';
 
@@ -11,6 +12,11 @@ const LOCALSTORAGE_KEY = 'feature-flag-overrides';
 
 let __SINGLETON: FeatureFlagOverrides | null = null;
 
+const BUFFER_SIZE = 10;
+
+// do we need to initialize the array with empty objects?
+const FEATURE_FLAGS: Flags = {values: new Array(10)};
+let buffer_idx = 0;
 export default class FeatureFlagOverrides {
   /**
    * Return the same instance of FeatureFlagOverrides in each part of the app.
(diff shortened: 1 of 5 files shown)


static/app/utils/featureFlagOverrides.ts
```diff
@@ -109,10 +115,38 @@ export default class FeatureFlagOverrides {
...
   public loadOrg(organization: Organization) {
     organization.features = this.getEnabledFeatureFlagList(organization);
+    // Track names of features that are passed into the .includes() function.
+    const handler = {
+      apply: function (target, orgFeatures, flagName) {
+        // Evaluate the result of .includes()
+        const flagResult = target.apply(orgFeatures, flagName);
+
+        // Check that the flag is not already in the buffer
+        if (!FEATURE_FLAGS.values.some(f => f.flag === flagName)) {
```
michellewzhang, line 137: if we are returning a buffer based on evaluation order, need to think about how we handle the same flag being evaluated. do we want to keep duplicates? or if no dups, shouldn't we delete the item from the buffer and re-insert it later?

ryan953: I would check and see how the python version is doing it today. The answer to this is something that we should 100% write down, because all sdk's should be consistent imo.

static/app/utils/featureFlagOverrides.ts
```diff
@@ -109,10 +115,38 @@ export default class FeatureFlagOverrides {
...
+          // Store the flag and its result in the buffer
+          FEATURE_FLAGS.values[buffer_idx] = {
+            flag: flagName,
+            result: flagResult,
+          };
+          // Increment the buffer index for next time
+          buffer_idx = (buffer_idx + 1) % BUFFER_SIZE;
```
ryan953, line 144: with this modulo flags won't be stored in-order. or at least, the start of the 'order' will be unknown.

Consider a case where the buffer size is 10, and 10 items are added. These would all be in-order, with the 1st item in the buffer matching the first flag that was checked.
Now add an 11th item to the buffer, it replaces the item as position 0. So now the earliest flag that was checked is stored in position 1 in the buffer. Listing the items in the buffer from 0 to 9 (10 items total) would list the most-recent flag first, followed by the oldest-flag second.

static/app/utils/featureFlagOverrides.ts
michellewzhang: discussed offline: we should be removing duplicates, they're only temporarily on the python SDK to do some testing. will remove dups from this

billyvg (commented):
Is `features.includes` the only way we access the feature flags?

Do we still do weird stuff like turn it into a set first? Are there instances where we use `indexOf`?

michellewzhang:
> Is `features.includes` the only way we access the feature flags?
> 
> Do we still do weird stuff like turn it into a set first? Are there instances where we use `indexOf`?

i can't find any instances of `indexOf` but not sure about the set🤔

ryan953 (approved):

c298lee (approved):
great tests!!

static/app/utils/featureObserver.ts
```diff
@@ -0,0 +1,72 @@
...
+    bufferSize = DEFAULT_BUFFER_SIZE,
+  }: {
+    organization: Organization;
+    bufferSize?: number;
+  }) {
+    const FLAGS = this.FEATURE_FLAGS;
+
+    if (!FLAGS) {
+      return;
+    }
```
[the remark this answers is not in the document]
michellewzhang: ooh good catch, this was leftover from an old thing

cmanallen (approved):

PR #77533, opened 2024-09-14
feat(uptime): Fetch rdap after subscription creation

evanpurkhiser:
Depends on GH-77518.

We're going to want to keep an eye on what errors this task might produce. Right now it will retry on errors, but we'll have to see what that actually looks like.

### DIFF
--- a/src/sentry/uptime/subscriptions/subscriptions.py
+++ b/src/sentry/uptime/subscriptions/subscriptions.py
@@ -9,6 +9,7 @@ from sentry.uptime.models import (
     ProjectUptimeSubscriptionMode,
     UptimeSubscription,
 )
+from sentry.uptime.rdap.tasks import fetch_subscription_rdap_info
 from sentry.uptime.subscriptions.tasks import (
     create_remote_uptime_subscription,
     delete_remote_uptime_subscription,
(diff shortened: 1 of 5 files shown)


src/sentry/uptime/subscriptions/subscriptions.py
```diff
@@ -53,6 +54,7 @@ def create_uptime_subscription(
 
     if created:
         create_remote_uptime_subscription.delay(subscription.id)
+        fetch_subscription_rdap_info.delay(subscription.id)
```
evanpurkhiser, line 57: I would like for there to be some way for this functionality to actually be mocked out typically in tests. Since this does talk to external APIs.

Right now I think tests actually go and talk to the RDAP RIR services, which is defintiely not what we want.

Is there a way for me to simply auto-mock this in most cases, and turn it off in the cases where we don't want mocks?

Unfortunately the `create_uptime_subscription` function is called in quite a lot of tests, so not sure if the best path forward is to mock it in every place it's used.

wedamija: Tasks don't run unless `with self.tasks` is used in the test.

That said, maybe we can just mock this by default in all of our uptime tests

wedamija (approved):

PR #77550, opened 2024-09-16
ref(onboarding): Add useful callbacks 

priscilawebdev:
these changes contribute to https://github.com/getsentry/sentry/issues/77537

requires https://github.com/getsentry/getsentry/pull/15157


static/app/components/onboarding/gettingStartedDoc/onboardingLayout.tsx
```diff
@@ -98,6 +105,9 @@ export function OnboardingLayout({
         ...doc.verify(docParams),
       ],
       nextSteps: doc.nextSteps?.(docParams) || [],
+      onPlatformOptionsClick: doc.onPlatformOptionsClick?.(docParams),
+      onProductSelectionClick: doc.onProductSelectionClick?.(docParams),
+      onPageLoad: doc.onPageLoad?.(docParams),
```
ArthurKnaus, line 110: You don't want to call them at this stage.

```suggestion
      onPlatformOptionsClick: doc.onPlatformOptionsClick,
      onProductSelectionClick: doc.onProductSelectionClick,
      onPageLoad: doc.onPageLoad,
```

I guess we need `onPageLoad` as we need too initially sync the loader settings with the selected products?

static/app/components/onboarding/gettingStartedDoc/onboardingLayout.tsx
```diff
@@ -126,10 +140,14 @@ export function OnboardingLayout({
...
               platform={platformKey}
               projectId={projectId}
+              onClick={onProductSelectionClick}
             />
           )}
           {platformOptions && !['customMetricsOnboarding'].includes(configType) ? (
-            <PlatformOptionsControl platformOptions={platformOptions} />
+            <PlatformOptionsControl
+              platformOptions={platformOptions}
+              onClick={onPlatformOptionsClick}
```
ArthurKnaus, line 149: super nit: Why `onClick` vs `onChange`?
I see it like an input and there we would implement `onChange`, right? 🤔

priscilawebdev: sure because we have an `input` element...in this case the `onChange` makes more sense

static/app/components/onboarding/gettingStartedDoc/onboardingLayout.tsx
priscilawebdev: so if `onLoad` is optional and is useful  to fire an analytic event when the page loads

static/app/components/onboarding/platformOptionsControl.tsx
```diff
@@ -70,13 +70,21 @@ export type PlatformOptionsControlProps = {
...
-export function PlatformOptionsControl({platformOptions}: PlatformOptionsControlProps) {
+export function PlatformOptionsControl({
+  platformOptions,
+  onChange,
+}: PlatformOptionsControlProps) {
   const router = useRouter();
   const urlOptionValues = useUrlPlatformOptions(platformOptions);
 
   const handleChange = (key: string, value: string) => {
+    onChange?.();
```
ArthurKnaus, line 87: Don't you need to pass the new value here?
`docParams`, which is available to the callback, will still hold the old value, right?
Same for product selection.

priscilawebdev: I actually tested the changes in the [PR](https://github.com/getsentry/sentry/pull/77310) and everything was correct

ArthurKnaus: But in the implementation of that PR you pass the new values to the callback?
Or am I missing something?

priscilawebdev: or yes...you're right, let me update this code

ArthurKnaus (approved):

PR #77551, opened 2024-09-16
feat(whats-new): add created by id to broadcasts

shellmayr:


priscilawebdev (approved):
in the get request we will have to return the `id` too so we can display this info in the broadcast details view. 

It looks good to me.

markstory (approved):
Schema changes look good.

wedamija (approved):

PR #77552, opened 2024-09-16
fix(insights): Prevent DB system selector from having unsupported systems

0Calories:
Some database systems that we do not support yet were sneaking by as options in the system selector, this prevents that:

### Before
image

### After
image


DominikB2014 (approved):
Lgtm

PR #77561, opened 2024-09-16
:bug: fix(slack): Update Discover Chart Interval Validation

iamrajjoshi:
Fixes [SENTRY-3A0Z](https://sentry.sentry.io/issues/5444045247/)

`validate_interval` expects an int for `top_events`, but the value in param is stringified, so just converting it here.


cathteng (approved):
what did we change that broke this? 🤔

PR #77562, opened 2024-09-16
fix(insights): JSON fails to parse some span descriptions

0Calories:
Sometimes a MongoDB span description can be truncated, or simply be invalid when attempting to parse. However, we can access the full description via `sentry_tags`. This PR adds an exhaustive check to find the most representative string for displaying the MongoDB query in the full span description, and prevents crashes.


DominikB2014 (approved):
Nice catch! Lgtm

PR #77565, opened 2024-09-16
feat(issue_platform): Support Issue Platform deletions

armenzg:
This brings parity with error issues.


armenzg:
I will be using #77794 to add this support.

PR #77566, opened 2024-09-16
feat(explore-charts): Displaying topN events in chart and table.

Abdkhan14:
Dev-ui link to test it out: [link](https://sentry.dev.getsentry.net:7999/traces/?field=project&field=id&field=span.op&field=span.description&field=span.duration&field=timestamp&field=organization_slug&field=project.organization.id&groupBy=deviceMemory&interval=1m&mode=aggregate&project=-1&query=&sort=organization_slug&statsPeriod=1h&utc=true&visualize=%7B%22yAxes%22%3A%5B%22avg%28span.duration%29%22%5D%2C%22chartType%22%3A1%7D)
Screenshot 2024-09-16 at 1 36 15 PM


static/app/views/explore/charts/index.tsx
```diff
@@ -44,14 +52,36 @@ export function ExploreCharts({query}: ExploreChartsProps) {
...
+  }, [resultMode, groupBys, visualizes]);
+  const [sorts] = useSorts({fields});
+
+  const orderby: string | string[] | undefined = useMemo(() => {
+    if (!sorts.length) {
+      return undefined;
+    }
+
+    return sorts.map(formatSort);
+  }, [sorts]);
```
wmak, line 76: should you include `resultMode` in the memo? Actually is this that expensive a function we need to memoize in the first place?

wmak (approved):
lgtm from testing and a skim

static/app/views/explore/charts/index.tsx
Abdkhan14: Hey, `resultMode` doesn't have to be a dependency since  `orderBy` is just a parsed `sort`. Also, `sort` is a string[], hence the memoization.

static/app/views/explore/charts/index.tsx
```diff
@@ -44,14 +52,36 @@ export function ExploreCharts({query}: ExploreChartsProps) {
...
 
   const yAxes = useMemo(() => {
     const deduped = dedupeArray(visualizes.flatMap(visualize => visualize.yAxes));
     deduped.sort();
     return deduped;
   }, [visualizes]);
 
-  const series = useSpanIndexedSeries(
+  const singleSeriesResult = useSpanIndexedSeries(
```
edwardgou-sentry, line 84: I _think_ you can disable this query if `topEvents === undefined`?

edwardgou-sentry (approved):
lgtm

static/app/views/explore/charts/index.tsx
Abdkhan14: Ahh yes, good catch!

PR #77571, opened 2024-09-16
feat(query-builder): Add navigation shortcuts with cmd/option/ctrl + arrow keys

malwilley:
Adds support for cmd+left/right to skip to start/end end and ctrl/option+left/right to skip over tokens.

https://github.com/user-attachments/assets/109658e5-7a4a-4314-a8ec-2f176906a703


static/app/components/searchQueryBuilder/hooks/useQueryBuilderGridItem.tsx
```diff
@@ -13,6 +15,69 @@ function isInputElement(target: EventTarget): target is HTMLInputElement {
 
 const noop = () => {};
 
+function focusNextGridCell({
```
malwilley, line 18: These functions at the top are existing logic that I extracted into functions for better readability

roggenkemper (approved):

PR #77579, opened 2024-09-16
chore(uptime): Make sure all tests use the `UptimeTestCase` base class

wedamija:
We want to mock some expensive functions in all uptime tests, so making sure we share a base class


davidenwang (approved):
nice

PR #77583, opened 2024-09-16
chore(similarity): Add longer snuba delay in backfill

jangjodi:
Increase delay between snuba calls in similarity backfill to help with snuba ratelimit being hit
Add worker number to log so we can more easily query which worker numbers are running in backfill


armenzg (approved):

JoshFerge (approved):
let's increase retry too?

PR #77584, opened 2024-09-16
feat(flag): FF for releases overlay on dashboard widgets

nikkikapadia:
Adding feature flag for the releases overlay on dashboard widgets


narsaynorath (approved):
:+1:

PR #77585, opened 2024-09-16
fix(anomaly detection): delete rule data in Seer on alert deletion

mifu67:
If we delete an anomaly detection alert or update it to no longer be dynamic, we need to delete its data in Seer.


src/sentry/incidents/logic.py
```diff
@@ -953,6 +958,14 @@ def update_alert_rule(
...
                     raise ValidationError("Failed to send data to Seer - cannot update alert rule.")
         else:
+            # if this was a dynamic rule, delete the data in Seer
+            if alert_rule.detection_type == AlertRuleDetectionType.DYNAMIC:
+                success = delete_rule_in_seer(
+                    alert_rule=alert_rule,
+                    project=projects[0] if projects else alert_rule.projects.get(),
+                )
+                if not success:
+                    raise SeerFailureError()
```
mifu67, line 968: Not sure if this needs to be a ValidationError for us to surface it to the user?

src/sentry/conf/server.py
```diff
@@ -3146,12 +3146,16 @@ def custom_parameter_sort(parameter: dict) -> tuple[str, int]:
 
 SEER_ANOMALY_DETECTION_MODEL_VERSION = "v1"
 SEER_ANOMALY_DETECTION_URL = SEER_DEFAULT_URL  # for local development, these share a URL
-SEER_ANOMALY_DETECTION_TIMEOUT = 15
+SEER_ANOMALY_DETECTION_TIMEOUT = 5
```
ceorourke, line 3149: why are we going back to 5 seconds for the timeout?

src/sentry/seer/anomaly_detection/delete_rule.py
```diff
@@ -0,0 +1,86 @@
...
+
+seer_anomaly_detection_connection_pool = connection_from_url(
+    settings.SEER_ANOMALY_DETECTION_URL, timeout=settings.SEER_DEFAULT_TIMEOUT
+)
+
+if TYPE_CHECKING:
+    from sentry.incidents.models.alert_rule import AlertRule
+
+
+def delete_rule_in_seer(alert_rule: "AlertRule") -> bool:
```
ceorourke, line 25: Why can't we just import it normally?

src/sentry/seer/anomaly_detection/delete_rule.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    Send a request to delete an alert rule from Seer. Returns True if the request was successful.
+    """
+    body = DeleteRuleRequest(
+        organization_id=cast(Organization, alert_rule.organization).id,
```
ceorourke, line 30: Why do we need to cast this to be an `Organization` when we're just using the id?

src/sentry/seer/anomaly_detection/types.py
```diff
@@ -45,6 +45,11 @@ class DetectAnomaliesRequest(TypedDict):
     context: AlertInSeer | list[TimeSeriesPoint]
 
 
+class DeleteRuleRequest(TypedDict):
+    organization_id: int
+    alert: dict
```
ceorourke, line 50: Can we make this match more closely: https://github.com/getsentry/seer/blob/main/src/seer/anomaly_detection/models/external.py#L93-L96
It seems like we're missing the project ID too

src/sentry/seer/anomaly_detection/delete_rule.py
mifu67: Circular imports (this function is used in alertrule.py, from which we import the AlertRule model)

src/sentry/conf/server.py
mifu67: Ram resolved the timeouts and told me to roll it back. But I can take the change out of this PR.

src/sentry/seer/anomaly_detection/delete_rule.py
mifu67: @nhsiehgit you wrote this code so I'll leave this to you.

src/sentry/seer/anomaly_detection/types.py
mifu67: We can't pass the project ID because a deleted rule has already been unbound from all its projects. Talked to Ram, who said he'll update the API to make project an optional field.

src/sentry/seer/anomaly_detection/delete_rule.py
nhsiehgit: Oh lol

Ya - cuz otherwise linting gets mad since `organization` is a nullable field on the `AlertRule` model 🤔 

but.... it feels like it shouldn't be nullable.... but it is... so 🤷

src/sentry/seer/anomaly_detection/delete_rule.py
nhsiehgit: ya - FWIW we should _probably_ be doing this any time we're using an import only for types? 🤷

ceorourke (approved):

PR #77586, opened 2024-09-16
ref(integrations): Type annotations on msteams/webhook.py

RyanSkonnord:
Add type annotations and fix incorrect ones.

Simplify by removing `MsTeamsWebhookMixin` and extracting its methods into a new `parsing` module.


src/sentry/integrations/msteams/webhook.py
```diff
@@ -243,25 +246,25 @@ class MsTeamsWebhookEndpoint(Endpoint, MsTeamsWebhookMixin):
...
+        self._event_handlers: dict[MsTeamsEvents, Callable[[Request], Response]] = {
+            MsTeamsEvents.MESSAGE: self._handle_message_event,
+            MsTeamsEvents.CONVERSATION_UPDATE: self._handle_conversation_update_event,
+            MsTeamsEvents.INSTALLATION_UPDATE: self._handle_installation_update_event,
+            MsTeamsEvents.UNKNOWN: self._handle_unknown_event,
         }
 
     @csrf_exempt
-    def dispatch(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
+    def dispatch(self, request: Request, *args, **kwargs) -> Response:  # type: ignore[override]
```
RyanSkonnord, line 257: I could use some help understanding the error that this "type:ignore" is suppressing, which is:

```
error: Argument 1 of "dispatch" is incompatible with supertype "APIView"; supertype defines the argument type as "HttpRequest"  [override]
```

`sentry.api.base.Endpoint.dispatch` has the parameter type `rest_framework.requests.Request`, which is also what this class gets at runtime and consistent with the rest of the class wanting to read `request.data` everywhere. The error message suggests that it's the base `Endpoint` class that's incorrectly overriding `APIView`, but AFAICT that class is covered by type-checking and produces no error.

Furthermore, `APIView` itself doesn't have any type annotations, at least not in the source code that's showing up in my venv. And it's a `rest_framework` class too, so I don't know why it wouldn't use the `rest_framework` request class.

For context, `HttpRequest` has no `data` attribute and is wrong to use in the rest of this class's handler methods. The problem does not appear to be unique to this class; I'm seeing `def dispatch(self, request: HttpRequest, *args, **kwargs) -> HttpResponse` in a few other places, but not consistently.

asottile-sentry: our view class hierarchy is a disaster -- it's kinda impossible to know what is and isn't a rest framework view at the moment

at some point these need to be fixed from the bottom up -- and perhaps without inheritance -- but it's such a large problem that I haven't even started thinking about how to approach it

technically even for `rest_framework` `def dispatch` still takes the vanilla django request -- rest_frakework then converts it into its special class.  so the type checker is actually correct here that this should be `(HttpRequest, *object, **object) -> HttpResponseBase`

RyanSkonnord: Glad to know it's not just me. 😅

> technically even for `rest_framework` `def dispatch` still takes the vanilla django request -- rest_frakework then converts it into its special class. so the type checker is actually correct here that this should be `(HttpRequest, *object, **object) -> HttpResponseBase`

That makes sense. I took another pass and it turns out that Mypy is cool if I leave this as `dispatch(self, request: HttpRequest, *args, **kwargs) -> HttpResponse` and change `post` and everything else to use `rest_framework.Request`. I had thought it would complain about the `super().dispatch` call (PyCharm even highlights it with a warning) but I guess not.

src/sentry/integrations/msteams/webhook.py
```diff
@@ -143,6 +145,8 @@ def verify_signature(request):
 
 
 class MsTeamsWebhookMixin:
+    provider: ClassVar[str]
```
asottile-sentry, line 148: so... mixins aren't typesafe -- I looked at this previously and I think this can just be converted into free functions (since nothing depends on actual `self` state and `provider` is a constant `'msteams'`)

src/sentry/integrations/msteams/webhook.py
```diff
@@ -103,7 +105,7 @@ def verify_signature(request):
     # get the open id config and jwks
     client = MsTeamsJwtClient()
     open_id_config = client.get_open_id_config()
-    jwks = client.get_cached(open_id_config["jwks_uri"])
+    jwks = cast(Mapping[str, Any], client.get_cached(open_id_config["jwks_uri"]))
```
asottile-sentry, line 108: this cast is annoying -- but it's because our api clients aren't typesafe

RyanSkonnord: Yep. `get_cached` is actually typed as `Union[BaseApiResponse, Mapping[str, Any], Response]`, so that's fun.

src/sentry/integrations/msteams/webhook.py
RyanSkonnord: I also dislike the mixin pattern. I don't hate the idea of having these methods grouped semantically as a class, but it could easily be injected (or just instantiated on demand) rather than inherited.

Good point about the provider being constant as well.

Is that a change you'd want to see in this PR? Notwithstanding the above, I'm not sure exactly what you mean about the mixin not being typesafe. The `ClassVar` annotation is doing its job AFAICT.

asottile-sentry: it is and it isn't -- it ensures internal type safety but isn't checked against the full mro when mixed in so it doesn't actually guarantee any runtime safety

ideally we'd kill this mixin while fixing this file

RyanSkonnord: Done: https://github.com/getsentry/sentry/pull/77586/commits/ee139e11efb49434936eb3e150c872af7f7920ee

I did want to preserve the logical grouping of related methods, so they're in a new module now.

src/sentry/integrations/msteams/webhook.py
```diff
@@ -147,79 +148,6 @@ def verify_signature(request) -> bool:
 class MsTeamsWebhookMixin:
     provider: ClassVar[str]
```
asottile-sentry: awesome!  last little bit is to nuke this dead class I think (doesn't look like it's used in getsentry fortunately!)

src/sentry/integrations/msteams/parsing.py
```diff
@@ -0,0 +1,84 @@
+import logging
+from collections.abc import Mapping
+from typing import Any
+
+from sentry.integrations.msteams.spec import PROVIDER
+from sentry.integrations.services.integration import integration_service
+from sentry.integrations.services.integration.model import RpcIntegration
+
+logger = logging.getLogger("sentry.integrations.msteams.parsing")
```
asottile-sentry, line 9:
```suggestion
logger = logging.getLogger(__name__)
```

asottile-sentry (commented):
nice!

src/sentry/integrations/msteams/webhook.py
RyanSkonnord: That was just an editing mistake. It's already gone.

asottile-sentry: oic I reviewed too quickly!

asottile-sentry (approved):

src/sentry/integrations/msteams/card_builder/issues.py
```diff
@@ -53,7 +53,11 @@
 
 class MSTeamsIssueMessageBuilder(MSTeamsMessageBuilder):
     def __init__(
-        self, group: Group, event: Event, rules: Sequence[Rule], integration: RpcIntegration
+        self,
+        group: Group,
+        event: Event | GroupEvent,
```
GabeVillalobos, line 58: Any chance this could be `BaseEvent` instead?

src/sentry/integrations/msteams/webhook.py
```diff
@@ -302,7 +230,7 @@ def _get_team_installation_request_data(cls, data: dict[str, Any]) -> dict:
         }
         return params
 
-    def handle_installation_update_event(self, request: HttpRequest) -> HttpResponse:
+    def _handle_installation_update_event(self, request: Request) -> Response:
```
GabeVillalobos, line 233: Was `HttpRequest/Response` not accurate before? I'm mainly curious why this change was necessary.

GabeVillalobos (approved):
👏🏼 Nice work! Couple of minor questions, but nothing that should prevent merging.

src/sentry/integrations/msteams/webhook.py
RyanSkonnord: Basically yes. Anywhere you see `request.data`, you're relying on it being a `rest_framework.requests.Request`. But the inheritance structure of our `Endpoint` class is kind of messed up per [this comment thread](https://github.com/getsentry/sentry/pull/77586#discussion_r1761799185).

src/sentry/integrations/msteams/card_builder/issues.py
RyanSkonnord: Yep, good call. https://github.com/getsentry/sentry/pull/77586/commits/81b5cc16de3ff2f0df8ecdc8c7043368c2f14c30

GabeVillalobos: Looks like this broke typing 🙈 Weird that both subclasses have a `group` attribute, but the base class doesn't.

PR #77588, opened 2024-09-16
chore(issues): Remove unnecessary type mapping

mrduncan:
This dict is only used in two places, removing it makes the usages clearer and aligns them better with all other detectors.

It is also not used within getsentry.


roggenkemper (approved):

mjq (approved):
👍 Looks great, thanks!

PR #77589, opened 2024-09-16
fix(issues): Hide traces with only orphan errors

scttcper:
Hides the trace preview when it contains no transactions


JoshFerge:
we'll still show related issues right? i sometimes like this view 🤔

armenzg (approved):
It would still show the trace related issue if there are two distinct issues.

PR #77596, opened 2024-09-16
fix(insights): Bug when sometimes parsing MongoDB queries in summary

0Calories:
Fixes a couple of things:

- Sometimes queries would fail to parse in the span summary of mongoDB queries because there wasn't an exhaustive check being applied
- The fully unparameterized query would get used in the tooltip hover due to ordering of the if statements, but we need the collection to be unparameterized


static/app/views/insights/common/components/fullSpanDescription.tsx
```diff
@@ -84,10 +84,6 @@ export function FullSpanDescription({
 
 const LINE_LENGTH = 60;
 
-function prettyPrintJsonString(json: string) {
-  return JSON.stringify(JSON.parse(json), null, 4);
-}
-
```
0Calories, line 90: moved this to utils

DominikB2014 (approved):

PR #77597, opened 2024-09-16
chore(dashboards): Reduce instrumentation for `AutoSizedText`

gggritso:
Rollout's basically done, this is the last thing. I don't want to create transactions for every resize, that's too much noise.

1. Remove some unnecessary tags (e.g., iteration exceeded) since that condition has mostly disappeared. The iteration count makes the same point
2. Do not create a transaction. In fact, do the opposite: attach a span to an existing transaction if it exists, otherwise just drop it, it's okay
3. Remove custom duration measurements. This has done its job, and it wasn't nearly as relevant for span durations as transaction durations anyway


narsaynorath (approved):
👍

PR #77598, opened 2024-09-16
fix(invite-members): enable Resend button + display correct disabled message for Remove button

ameliahsu:
was displaying incorrect disabled message:
Screenshot 2024-09-16 at 2 13 28 PM

fixed by checking if the member has the `pending` property instead of checking for `inviterName`

also enabled Resend Invite button for invites sent by the current user
Screenshot 2024-09-16 at 4 06 03 PM

### DIFF
--- a/static/app/views/settings/organizationMembers/organizationMemberRow.tsx
+++ b/static/app/views/settings/organizationMembers/organizationMemberRow.tsx
@@ -114,9 +114,9 @@ export default class OrganizationMemberRow extends PureComponent<Props, State> {
     const isCurrentUser = currentUser.email === email;
     const showRemoveButton = !isCurrentUser;
     const showLeaveButton = isCurrentUser;
-    // inviterName is null if this is not a pending invite
-    // i.e. invite has been accepted
-    const isInvite = inviterName !== null;
+    // member has a `user` property if they are registered with sentry
+    // i.e. has accepted an invite to join org
+    const isInvite = !user;
     const isInviteFromCurrentUser = inviterName === currentUser.name;
     const canRemoveInvites =
       organization.features?.includes('members-invite-teammates') &&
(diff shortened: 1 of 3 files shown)


static/app/views/settings/organizationMembers/organizationMemberRow.tsx
```diff
@@ -216,7 +213,7 @@ export default class OrganizationMemberRow extends PureComponent<Props, State> {
                     : isPartnershipUser
                       ? t('You cannot make changes to this partner-provisioned user.')
                       : // only show this message if member can remove invites but invite was not sent by them
-                        isInvite && canRemoveInvites && !isInviteFromCurrentUser
+                        pending && canInvite && !isInviteFromCurrentUser
```
cathteng, line 216: does `pending` refer to whether the member has a `user` property?

ameliahsu: Sorry I updated the PR description, originally I was checking if `user` was null but then I realized there was a `pending` property which is true if the invite is still pending

static/app/views/settings/organizationMembers/organizationMemberRow.tsx
```diff
@@ -216,7 +213,7 @@ export default class OrganizationMemberRow extends PureComponent<Props, State> {
...
                         ? t('Your role cannot modify this invite.')
```
cathteng, line 217: would it be better to say "You cannot modify this invite" bc `canInvite` is `True`?

static/app/views/settings/organizationMembers/organizationMemberRow.tsx
```diff
@@ -162,7 +159,7 @@ export default class OrganizationMemberRow extends PureComponent<Props, State> {
               {isInviteSuccessful && <span>{t('Sent!')}</span>}
               {!isInviting && !isInviteSuccessful && (
                 <Button
-                  disabled={!canAddMembers}
+                  disabled={!canAddMembers && !(canInvite && isInviteFromCurrentUser)}
```
cathteng, line 162: this is a bit hard to grok, would writing it as `!canAddMembers || !canInvite || !isInviteFromCurrentUser` still work?

cathteng (approved):
nitpicky comments

static/app/views/settings/organizationMembers/organizationMemberRow.tsx
ameliahsu: for me it's easier to think about when the button should be enabled, which is when either
- `canAddMembers` is true
- `canInvite` is true AND `isInviteFromCurrentUser` is true

but you're right this logic is hard to follow so I'll make a new boolean or something for readability 🫡

PR #77599, opened 2024-09-16
chore(azure-devops): setup new Azure DevOps integration

cathteng:
Missing setup flows, but it exists on the integrations page.

Screenshot 2024-09-16 at 14 42 56


cathteng:
going with an approach to migrate the integration in place

PR #77601, opened 2024-09-16
fix(custom-views): Remove temp tab on save, fix not being able to go back to temp tab

MichaelSun48:
This PR fixes a couple of regressed issues that have occurred with temp tabs, along with a minor cosmetic improvement: 

1. **Temp tabs would not close after saving**: Previously, clicking "Save as View" on a temporary tab would correctly save the view, but it would leave the temp tab in the tab bar. This PR makes it so the temp tab is removed as well 
2. **Unable to go back to temp tab once visiting another tab**: Previously if you started at a temp tab, then went to a persistent tab, then went back, the persistent tab would be erroneously populated with the temp tab's query, and the border style would not change to dashed. This PR fixes this behavior
3. The text in the add view button has now been aligned with the tabs


static/app/views/issueList/groupSearchViewTabs/draggableTabBar.tsx
```diff
@@ -226,6 +226,7 @@ export function DraggableTabBar({
       };
       const newTabs = [...tabs, newTab];
       setTabs(newTabs);
+      setTempTab(undefined);
```
MichaelSun48, line 229: Remove temp tab on save

static/app/views/issueList/groupSearchViewTabs/draggableTabBar.tsx
```diff
@@ -346,7 +347,7 @@ export function DraggableTabBar({
               ...queryParams,
               query: tab.unsavedChanges?.[0] ?? tab.query,
               sort: tab.unsavedChanges?.[1] ?? tab.querySort,
-              ...(tab.id !== 'temporary-tab' ? {viewId: tab.id} : {}),
+              viewId: tab.id !== 'temporary-tab' ? tab.id : undefined,
```
MichaelSun48, line 350: explicitly remove viewId from search params if temp tab is clicked. Previously, this would just persist whatever view's id you were on last.

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -422,6 +423,8 @@ const AddViewButton = styled(Button)`
   font-weight: normal;
   padding: ${space(0.5)};
   margin-right: ${space(0.5)};
+  border: none;
+  bottom: -1px;
```
MichaelSun48, line 427: the temp tab was 1 pixel over the bottom border of the tab bar previously:

Screenshot 2024-09-16 at 3 12 34 PM

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -282,7 +282,7 @@ function BaseDraggableTabList({
         <MotionWrapper>
           {tempTab && (
             <Tab
-              key={tempTab.key}
+              key={'temporary-tab'}
```
malwilley, line 285: Probably best to reference a constant for this since multiple files are using the same string

static/app/components/draggableTabs/draggableTabList.tsx
```diff
@@ -273,7 +273,7 @@ function BaseDraggableTabList({
         overflowingTabs={overflowingTabs}
       />
       <AddViewTempTabWrap ref={addViewTempTabRef}>
-        <MotionWrapper>
+        <MotionWrapper style={{marginTop: 2}}>
```
malwilley, line 276: Can you add this to the CSS instead of the style?

malwilley (approved):

PR #77607, opened 2024-09-16
chore(deps): Upgrade react aria deps

malwilley:
Fixes JAVASCRIPT-2VVF

In order to fix the above issue, I need this change: https://github.com/adobe/react-spectrum/blame/main/packages/%40react-stately/list/src/useListState.ts#L85


scttcper (approved):
weird test failing

static/app/components/dropdownMenu/item.tsx
```diff
@@ -52,7 +51,7 @@ export interface MenuItemProps extends MenuListItemProps {
    * Function to call when user selects/clicks/taps on the menu item. The
    * item's key is passed as an argument.
    */
-  onAction?: (key: MenuItemProps['key']) => void;
```
malwilley, line 55: It does look like there was a breaking change which let to removing this argument.

`useMenuItem()` now prefers the `onAction` passed in to the item props: https://github.com/adobe/react-spectrum/blob/c904e066c34160803a0aa081e13821d8f67e3e84/packages/%40react-aria/menu/src/useMenuItem.ts#L134-L138

When it uses the `onAction()` from the item props, it does not pass `key`. It doesn't seem necessary to me since the items will already know what their keys are, so I removed it from here and in the areas that passed it.

static/app/components/dropdownMenu/index.tsx
```diff
@@ -200,7 +200,7 @@ function DropdownMenu({
 
   function renderTrigger() {
     if (trigger) {
-      return trigger({...overlayTriggerProps, ...buttonProps}, isOpen);
+      return trigger({...buttonProps, ...overlayTriggerProps}, isOpen);
```
malwilley, line 203: This ordering matches the ordering below for the default dropdown button. With the old order, toggling a dropdown menu by clicking its trigger was not working.

static/app/components/dropdownMenu/index.spec.tsx
```diff
@@ -166,7 +166,6 @@ describe('DropdownMenu', function () {
 
     const parentItem = screen.getByRole('menuitemradio', {name: 'Item'});
 
-    expect(parentItem).toHaveAttribute('aria-haspopup', 'true');
```
malwilley, line 169: Removed this check. Looks `useMenuItem()` no longer adds `aria-haspopup`. But it still has `aria-expanded` and the other behavior works fine still.

PR #77611, opened 2024-09-17
feat(uptime): Add method, headers and body columns to `UptimeSubscription`

wedamija:
This adds columns that allow users to customize which method their checks are sent with, as well as custom headers and body.


wedamija:
> This PR has a migration; here is the generated SQL for `src/sentry/uptime/migrations/0012_uptime_subscription_request_fields.py` ()
> 
> ```sql
> --
> -- Custom state/database change combination
> --
> ALTER TABLE "uptime_uptimesubscription" ADD COLUMN "body" text NULL;
> ALTER TABLE "uptime_uptimesubscription" ADD COLUMN "headers" jsonb DEFAULT '{}'::jsonb NOT NULL;
> ALTER TABLE "uptime_uptimesubscription" ADD COLUMN "method" varchar(20) DEFAULT 'GET' NOT NULL;
> ```

On second thought, I should use a text field so that we control how the json is serialized. I'm not sure we'll be guaranteed the ordering we provided

davidenwang (approved):

PR #77613, opened 2024-09-17
feat(uptime): Add new unique index to `UptimeSubscription` that takes new fields into account

wedamija:
This adds a new unique constraint that takes the new columns into account. We use `md5` on both the `headers` and `body` column to keep them at a consistent size, and we can rely on the uniqueness here since we use `sort_keys` when storing values to keep the keys in a consistent order.


tests/sentry/uptime/test_models.py
```diff
@@ -15,3 +19,56 @@ def test(self):
...
+
+
+class UniqueMonitorTest(UptimeTestCase):
+    def test(self):
+        self.create_uptime_subscription(
+            url="https://santry.io",
+            interval_seconds=60,
+            method="GET",
+        )
+        with pytest.raises(IntegrityError), transaction.atomic(
```
wedamija, line 31: Using the transaction here helps us avoid the problem where the connection ends up in a bad state after a unique constraint violation

davidenwang (approved):

PR #77614, opened 2024-09-17
ref(sentry_apps): Rename apps.py and shim for getsentry

Christinarlong:
We need to rename `apps.py` so django doesn't confuse it for an `AppConfig` file when we later add `sentry_apps` to `INSTALLED APPS.` However, `apps.py` is also used in `getsentry` so we'll need to create a shim so that refs don't break. 

ref(#73857)


src/sentry/sentry_apps/apps.py
Christinarlong, line 1: Shim

src/sentry/sentry_apps/logic.py
```diff
@@ -0,0 +1,463 @@
...
+    def __post_init__(self) -> None:
+        if self.is_internal:
+            assert (
+                not self.verify_install
+            ), "Internal apps should not require installation verification"
+
+    def run(
+        self,
+        *,
+        user: User,
```
Christinarlong, line 310: This was changed from User | RpcUser to just User. Though I wasn't 100% on this one. Like SentryApp creation and updating all happens in the CS so can we assume all 'users'  won't be RpcUsers

src/sentry/integrations/vercel/integration.py
```diff
@@ -27,11 +27,11 @@
 from sentry.organizations.services.organization import RpcOrganizationSummary
 from sentry.pipeline import NestedPipelineView
 from sentry.projects.services.project_key import project_key_service
+from sentry.sentry_apps.logic import SentryAppCreator
 from sentry.shared_integrations.exceptions import ApiError, IntegrationError
 from sentry.users.models.user import User
 from sentry.utils.http import absolute_uri
 
-from ...sentry_apps.apps import SentryAppCreator
```
markstory, line 34: :bow: Thanks for fixing this relative import.

src/sentry/sentry_apps/logic.py
markstory: >  Like SentryApp creation and updating all happens in the CS so can we assume all 'users' won't be RpcUsers ?

We can likely use the narrower type. The RpcUser could show up in a future refactoring, but we can deal with that then.

markstory (approved):

PR #77623, opened 2024-09-17
ref(onboarding): Add analytics about selected products when 'Take me to issues' is clicked

vgrozdanic:

### DIFF
--- a/static/app/views/projectInstall/platform.tsx
+++ b/static/app/views/projectInstall/platform.tsx
@@ -217,6 +217,7 @@ export function ProjectInstallPlatform({
                 organization,
                 platform: platform.name ?? 'unknown',
                 project_id: project.id,
+                products,
               });
               redirectWithProjectSelection({
                 pathname: issueStreamLink,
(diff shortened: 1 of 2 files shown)


obostjancic (approved):

ArthurKnaus (approved):

static/app/views/projectInstall/platform.tsx
```diff
@@ -217,6 +217,7 @@ export function ProjectInstallPlatform({
                 organization,
                 platform: platform.name ?? 'unknown',
                 project_id: project.id,
+                products,
```
matejminar, line 220: Do we know how amplitude works when you send it an array of strings? E.g is it smart enough to be able to answer how many take me to issues where clicked with replay selected as one of the products? Or do we need to explode it into multiple booleans?

matejminar: + does different order in the array increase cardinality?

matejminar: 👍

matejminar (approved):

PR #77624, opened 2024-09-17
drop(grouping): Use new ophio parser

armenzg:
It drops requiring `is_prefix` and `is_sentinel` which has already been deprecated.

See changes here: https://github.com/getsentry/ophio/pull/63

### DIFF
--- a/requirements-base.txt
+++ b/requirements-base.txt
@@ -66,7 +66,7 @@ rfc3986-validator>=0.1.1
 # [end] jsonschema format validators
 sentry-arroyo>=2.16.5
 sentry-kafka-schemas>=0.1.109
-sentry-ophio==0.2.7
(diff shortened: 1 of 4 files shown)


requirements-base.txt
```diff
@@ -66,7 +66,7 @@ rfc3986-validator>=0.1.1
 # [end] jsonschema format validators
 sentry-arroyo>=2.16.5
 sentry-kafka-schemas>=0.1.109
-sentry-ophio==0.2.7
+sentry-ophio==0.2.9
```
armenzg, line 69: We had to do two releases:
https://github.com/getsentry/ophio/pull/63
https://github.com/getsentry/ophio/pull/64

asottile-sentry (approved):

JoshFerge (approved):

PR #77628, opened 2024-09-17
fix(onboarding-analytics): Do not fire modal closed event on configure or skip click

vgrozdanic:
With previous implementation, `onClose` function was called on every modal closing event, including clicking on "Skip" or "Configure SDK" button. Since those 2 events are already tracked by analytics, we wanted to fire `project_creation.select_framework_modal_close_button_clicked` only when close button was clicked, so the `closeModal` callback call was removed from the functions that handle "configure sdk" and "Skip" button clicks

### DIFF
--- a/static/app/views/projectInstall/createProject.tsx
+++ b/static/app/views/projectInstall/createProject.tsx
@@ -230,6 +230,12 @@ function CreateProject() {
       ),
       {
         modalCss,
+        onClose: () => {
+          trackAnalytics('project_creation.select_framework_modal_close_button_clicked', {
+            platform: selectedPlatform.key,
+            organization,
+          });
+        },
       }
     );
   }, [platform, createProject, organization]);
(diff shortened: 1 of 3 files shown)


static/app/views/projectInstall/createProject.tsx
```diff
@@ -230,12 +230,6 @@ function CreateProject() {
       ),
       {
         modalCss,
-        onClose: () => {
```
priscilawebdev, line 233: don't we want to fire the analytics event when users click on the backdrop of a modal, closing it?

vgrozdanic: This was really good feedback. I have fixed this by removing calling the `onClose` callback when "Skip" and "Configure SDK" buttons are called

priscilawebdev (approved):
great improvement 👏 tested everything and it works as expected

PR #77630, opened 2024-09-17
fix(onboarding-analytics) Do not fire `project_creation.data_removal_modal_dismissed` on modal confirm 

vgrozdanic:
`onClose` is being called whenever a modal is closed - even on modal confirm.

There is a callback `onCancel` which can be used to track `project_creation.data_removal_modal_dismissed` events. `onCancel` is called only when the modal is dismissed.

### DIFF
--- a/static/app/views/projectInstall/platformDocHeader.tsx
+++ b/static/app/views/projectInstall/platformDocHeader.tsx
@@ -104,7 +104,7 @@ export function PlatformDocHeader({platform, projectSlug, title}: Props) {
           priority="danger"
           confirmText={t("Yes I'm sure")}
           onConfirm={handleGoBack}
-          onClose={() => {
+          onCancel={() => {
             if (!recentCreatedProject) {
               return;
             }


priscilawebdev:
this solution works if the user clicks on the `cancel` button, but what if they click on the backdrop of the modal, dismissing it?

obostjancic (approved):

vgrozdanic:
@priscilawebdev 
> but what if they click on the backdrop of the modal, dismissing it?

We will not be aware of this event... Also, when `escape` key is pressed, we do not fire an event (we haven't even done it before this change that I have made). Unfortunately there is no nice way of doing this, without changing the `ConfirmModal` component, and potentially breaking a lot of other analytics instrumentation, and code that is using this component.

I would argue that this event is not very significant, and I am not sure how useful analytics for this really is, and if somebody is even querying for this event, so I would be ok with this being less precise, than firing this event on "wrong" occasions

Issue #77572, opened 2024-09-16
Divide up insights into Mobile / Web / Platform / AI


DominikB2014:
_This project is being fleshed out, broadly:_

- We want landing pages for each domain which would include a list of the specific root spans for that domain, e.g., screens for mobile, pages for web, routes for backend.
- We want to keep the existing insights modules, but they may be tabs on these pages.

See: https://www.notion.so/sentry/Proposal-Interim-Insights-Cleanup-a48686b74353458c86eae018aafdb88b?pvs=4

**Notes:**

* As part of this work we should remove the alternate Web Vitals UI from the Transaction Summary page.

```[tasklist]
### Tasks
- [x] Scope down domain views by transaction op
- [x] ensure performance pages contain correct headers/breadcrumbs
- [x] Bring back performance in sidebar and add domain views under insights
- [x] Name http module "Network Requests" in the frontend tab and "Outbound API Requests" in the backend
- [x] Add app starts, pageload and screen recording modules
- [x] Fix breadcrumbs not appending in module subpages
- [x] Fix header title not updating within module subpages
- [x] Module Tabs show "you do not have access to this feature" on AM1, we should make this nicer or hide the tabs
- [x] Fix give feedback button incorrectly place in header
- [x] Link web vitals widget in frontend overview page to new web vitals
- [x] make table exactly as they are now, just reorder columns operation next to transaction
- [x] Make bottom 3 charts the same for all views
- [x] add missing subpages for mobile screens
- [x] missing view trends button
- [ ] Create banner on transaction web vitals tab to notify that this view is going away (related to #78060)
- [ ] (nice to have) Sorting domain view table
- [ ] (nice to have) figure out transaction vs route vs pages vs something else for column header
- [ ] (nice to have): remove project column with one project
```

bcoe:
Ben's QAing:

## Frontend:

- [ ] breadcrumbs not updated when traversing into Pages, Queries, Domains, from Insights view:

![Image](https://github.com/user-attachments/assets/bc44accf-8282-4a9a-9af0-b85deae4aadd)

☝ Note the breadcrumb missing the "page".

PR #77632, opened 2024-09-17
feat(insights): init domain view

DominikB2014:
Work for https://github.com/getsentry/sentry/issues/77572

1. Adds sidebar entry for new domain specific view (under a flag)
2. Creates blank view for platform, mobile and ai
3. Scaffold basic view for web

Still a lot to do here,  but doing it in stages! There's also a lot of code to share between domain views and modules (if you see anything that's shareable between views then feel free to leave a comment).

If you're testing this, the links between module pages also need to be updated (under the flag in another PR) so that they don't go back to the insights tabs

image

### DIFF
--- a/static/app/views/insights/pages/frontendLandingPage.tsx
+++ b/static/app/views/insights/pages/frontendLandingPage.tsx
@@ -31,7 +31,7 @@ function WebLandingPage() {
       preservePageFilters: true,
     },
     {
-      label: WEB_LANDING_TITLE,
+      label: FRONTEND_LANDING_TITLE,
       to: '/performance/web',
       preservePageFilters: true,
     },
@@ -63,7 +63,7 @@ function WebLandingPage() {
           <Layout.HeaderContent>
             <Breadcrumbs crumbs={crumbs} />
 
-            <Layout.Title>{WEB_LANDING_TITLE}</Layout.Title>
+            <Layout.Title>{FRONTEND_LANDING_TITLE}</Layout.Title>
           </Layout.HeaderContent>
           <Layout.HeaderActions>
             <ButtonBar gap={1}>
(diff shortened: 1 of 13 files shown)


static/app/views/insights/pages/frontendLandingPage.tsx
```diff
@@ -0,0 +1,111 @@
...
+              </TabPanels.Item>
+              <TabPanels.Item key={ModuleName.RESOURCE}>
+                <ResourcesLandingPage {...landingPageProps} />
+              </TabPanels.Item>
+              <TabPanels.Item key={ModuleName.VITAL}>
+                <WebVitalsLandingPage {...landingPageProps} />
+              </TabPanels.Item>
+            </TabPanels>
+          </Layout.Main>
+        </Layout.Body>
```
0Calories, line 102: Let's delete the `Layout.Body` wrapper component here because it already exists in all the subpages, which is resulting in excessive padding

```suggestion
        <Layout.Main fullWidth>
          <PageAlert />
          <TabPanels>
            <TabPanels.Item key="overview">{'overview page'}</TabPanels.Item>
            <TabPanels.Item key={ModuleName.HTTP}>
              <HTTPLandingPage {...landingPageProps} />
            </TabPanels.Item>
            <TabPanels.Item key={ModuleName.RESOURCE}>
              <ResourcesLandingPage {...landingPageProps} />
            </TabPanels.Item>
            <TabPanels.Item key={ModuleName.VITAL}>
              <WebVitalsLandingPage {...landingPageProps} />
            </TabPanels.Item>
          </TabPanels>
        </Layout.Main>
```

0Calories (approved):
Looks good to me so far aside from some minor things, we'll probably have to do some restructuring and make each domain its own folder instead of keeping them all in the `pages` folder, but we can worry about that in future PRs

static/app/views/insights/pages/frontendLandingPage.tsx
DominikB2014: Good call!

PR #77635, opened 2024-09-17
ref(releases): remove top level db transaction on set_commits

JoshFerge:
removes the top level transaction for set_commits. we'll instead do a transaction per commit, which will ensure that groupresolutions etc. are all set together with the commit.

in theory if the task fails partway through, it could lead to degraded commit info on a release, but i believe this wouldn't happen very often, and in the previous case, we'd just set no commits on the release, so i don't think this is that big of a change in behavior.

There is still a redis lock on the release, so we can ensure conflicting updates won't happen at the same time. deploying under a feature flag and will test with the sentry org first to ensure no issues can occur.


src/sentry/models/releases/set_commits.py
```diff
@@ -60,22 +60,23 @@ def set_commits(release, commit_list):
         # the same release rapidly for different projects.
         raise ReleaseCommitError
 
-    if features.has("organizations:set-commits-updated", release.organization):
+    with TimedRetryPolicy(10)(lock.acquire):
         create_repositories(commit_list, release)
```
JoshFerge, line 64: moving these to be inside the lock

src/sentry/models/releases/set_commits.py
```diff
@@ -88,6 +94,45 @@ def set_commits_on_release(release, commit_list):
...
+            str(a_id)
+            for a_id in ReleaseCommit.objects.filter(
+                release=release, commit__author_id__isnull=False
+            )
+            .values_list("commit__author_id", flat=True)
+            .distinct()
+        ],
+        last_commit_id=latest_commit.id if latest_commit else None,
+    )
+    return head_commit_by_repo, commit_author_by_commit
```
armenzg, line 116: For anyone looking at this PR after me, the function is `set_commits_on_release` is unchanged regardless of what the diff may seem to indicate.

If `set_commits_on_release_with_transaction_per_commit` was further down in the module the diff would show that `set_commits_on_release` stands untouched.

src/sentry/models/releases/set_commits.py
```diff
@@ -88,6 +94,45 @@ def set_commits_on_release(release, commit_list):
...
+
+    commit_author_by_commit = {}
+    head_commit_by_repo: dict[int, int] = {}
+
+    latest_commit = None
+    for idx, data in enumerate(commit_list):
+        with (
+            atomic_transaction(router.db_for_write(type(release))),
+            in_test_hide_transaction_boundary(),
+        ):
```
armenzg, line 133: The main difference with the original function is that this with block is added.

https://github.com/getsentry/sentry/blame/4639d43292bfe72866d7edbdcaa2bd28c7bdf914/src/sentry/models/releases/set_commits.py#L89-L92
image

armenzg (approved):
lgtm!

Do you need to update any tests?

Ignore my comments. They're only to help anyone reading this PR at a later point.

JoshFerge:
gonna close this PR for now -- the changes to the redis locks to make them longer seemed to have fixed the database errors we were getting previously, and i think this code seems healthy for now without needing to do this splitting transaction stuff.

PR #77643, opened 2024-09-17
fix(rules): Fix serializer when rule snooze owner is None

ceorourke:
If the `RuleSnooze` object `owner_id` is `None` we were hitting an error calling `user_service.get_user()` on it which caused serialization to fail and `None` entries were in the `serialized_issue_rules` in the combined rule serializer.

Closes https://getsentry.atlassian.net/browse/ALRT-285
Fixes SENTRY-3A73
Fixes SENTRY-3D67

### DIFF
--- a/src/sentry/api/serializers/models/rule.py
+++ b/src/sentry/api/serializers/models/rule.py
@@ -258,7 +258,7 @@ class RuleSerializer(Serializer):
             created_by = None
             if user.id == snooze.get("owner_id"):
                 created_by = "You"
-            else:
+            elif snooze.get("owner_id"):
                 creator = user_service.get_user(snooze.get("owner_id"))
                 if creator:
                     created_by = creator.get_display_name()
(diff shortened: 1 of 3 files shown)


src/sentry/api/serializers/models/rule.py
```diff
@@ -258,7 +258,7 @@ def serialize(self, obj, attrs, user, **kwargs) -> RuleSerializerResponse:
             created_by = None
             if user.id == snooze.get("owner_id"):
                 created_by = "You"
-            else:
+            elif snooze.get("owner_id"):
                 creator = user_service.get_user(snooze.get("owner_id"))
```
schew2381, line 262: Can use a walrus here cuz it looks cool 😎
```suggestion
            elif owner_id := snooze.get("owner_id"):
                creator = user_service.get_user(owner_id)
```

schew2381 (approved):

PR #77648, opened 2024-09-17
fix(insights): Handle truncated MongoDB queries

0Calories:
DB span descriptions in the metrics dataset can be truncated if they are too long. This results in broken unparseable JSON which messes up the hover tooltip content, and the formatting of the query in the row. 

This PR introduces the `jsonrepair` module to handle truncated JSON by repairing it into something the parser can handle and format appropriately.

### Example:
This is a truncated query, after this PR, it is correctly formatted with spacing and the last entry gets assigned a value
image

The tooltip also gets formatted correctly:
image

In a followup PR I may include some sort of indication to the user that the query is truncated, since it is not obvious from looking at it here

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
     "jest-fetch-mock": "^3.0.3",
     "js-beautify": "^1.15.1",
     "js-cookie": "3.0.1",
+    "jsonrepair": "^3.8.0",
     "less": "^4.1.3",
     "less-loader": "^12.2.0",
     "lightningcss": "^1.26.0",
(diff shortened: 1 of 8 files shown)


DominikB2014 (approved):
This is a pretty neat idea! You mentioned this in the PR, but I do think we should indicate somehow that it's being truncated.

scttcper:
did look at bundle impact, seems small 6kb https://bundlejs.com/?q=jsonrepair&treeshake=%5B*%5D&text=%22jsonrepair%28%29%22

scttcper (approved):

0Calories:
thanks for verifying!

PR #77649, opened 2024-09-17
ref(sentry_apps): Move SentryAppInstallation model to sentry_apps

Christinarlong:
Cool potatoes, let's move SentryAppInstallation (SAI) to new home :D

relies on/branched off of (#77521)


src/sentry/sentry_apps/models/sentry_app_installation.py
```diff
@@ -56,48 +55,6 @@ def get_projects(self, token: ApiToken | AuthenticatedToken) -> QuerySet[Project
 
         return Project.objects.filter(organization_id=token.organization_id)
 
-    def get_related_sentry_app_components(
```
markstory, line 59: :clap: Nice to see this leave.

markstory (approved):

PR #77653, opened 2024-09-17
tests(issue_platform): Test that we can delete Issue Platform issues

armenzg:
The endpoint does not permit deleting Issue Platform issues, however, the task allows for it.

I will follow-up with a different PR to add the support for deleting the occurrence from Snuba.

### DIFF
--- a/tests/sentry/deletions/test_group.py
+++ b/tests/sentry/deletions/test_group.py
@@ -22,8 +22,10 @@ from sentry.testutils.helpers.datetime import before_now, iso_format
 class DeleteGroupTest(TestCase, SnubaTestCase):
     def setUp(self):
         super().setUp()
-        group1_data = {"timestamp": iso_format(before_now(minutes=1)), "fingerprint": ["group1"]}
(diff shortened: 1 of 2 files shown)


tests/sentry/deletions/test_group.py
```diff
@@ -17,14 +17,15 @@
...
 from sentry.testutils.helpers.datetime import before_now, iso_format
+from tests.sentry.issues.test_utils import OccurrenceTestMixin
 
 
 class DeleteGroupTest(TestCase, SnubaTestCase):
     def setUp(self):
         super().setUp()
-        group1_data = {"timestamp": iso_format(before_now(minutes=1)), "fingerprint": ["group1"]}
-        group2_data = {"timestamp": iso_format(before_now(minutes=1)), "fingerprint": ["group2"]}
-        self.project = self.create_project()
```
armenzg, line 27: It's not necessary since `TestCase` handles it.

JoshFerge (approved):

PR #77658, opened 2024-09-17
prototype(worker-tasks): Create pending task store abstraction

enochtangg:


wedamija (commented):
This is too large to review as is. I'd recommend breaking it up, and adding tests to each section as you go

john-z-yang:
> This is too large to review as is. I'd recommend breaking it up, and adding tests to each section as you go

This is not meant to target prod, it's for prototyping only. Apologies for the confusion!

src/sentry/taskworker/pending_tasks.py
```diff
@@ -0,0 +1,23 @@
...
+    PENDING = "pending"
+    PROCESSING = "processing"
+    COMPLETE = "complete"
+    FAILURE = "failure"
+    RETRY = "retry"
+
+
+@dataclass
+class PendingTask:
+    proto_task: PendingTaskProto
```
markstory, line 19: :+1: Decorating with a 'domain object' is a good path to take. Copying all the values out of the protobuf will be expensive in aggregate.

We might want to add methods/properties to this object that proxy to the `proto_task` and have the `proto_task` be a private property.

src/sentry/taskworker/pending_tasks.py
```diff
@@ -0,0 +1,23 @@
+from dataclasses import dataclass
+from enum import Enum
+
+from sentry_protos.hackweek_team_no_celery_pls.v1alpha.pending_task_pb2 import (
```
markstory, line 4: We'll have to come up with a better name for this :smile:

src/sentry/taskworker/processors/strategy_factory.py
```diff
@@ -224,3 +120,4 @@ def do_upkeep(
 
     def shutdown(self):
         self.pool.close()
+        self.pool.close()
```
markstory, line 123: Why twice?

john-z-yang: This is a mistake, it's fixed now

PR #77659, opened 2024-09-17
fix(substatus): Unset substatus when merging groups

snigdhas:
The merged groups have the status set to pending_merge, but retain the old substatus which is inconsistent. This field should be unset, and the groups will take the status/substatus of the primary group once the merge is completed.


JoshFerge (commented):
can we add a test?

JoshFerge (approved):

PR #77665, opened 2024-09-17
Stacked Nav

natemoo-re:
Update side navigation to new Stacked UI, behind the `navigation-sidebar-v2` flag.

Rather than the approach taken in https://github.com/getsentry/sentry/pull/76868, this PR implements the new navigation as an entirely new set of components in the `components/nav/` directory. This new experience is rendered from the `static/app/views/organizationLayout/index.tsx` file. The intention is that the existing `components/sidebar/` directory would be removed when this feature flag reaches GA.

## Todo
- [x] Basic Tests
- [x] Basic Mobile nav


static/app/components/nav/config.tsx
```diff
@@ -0,0 +1,192 @@
...
+    body: SidebarItem[];
+    footer: SidebarItem[];
+  };
+  secondary: {
+    body: SubmenuItem[];
+    footer: SubmenuItem[];
+  };
+}
+
+export function useNavItems(): NavItemsResult {
```
natemoo-re, line 44: The `useNavItems` hook acts as a single source of truth for the entire nav, including nested navigation and feature flag checks.

static/app/views/organizationLayout/index.tsx
```diff
@@ -33,20 +37,57 @@ function OrganizationLayout({children}: Props) {
   // oganization is loaded before rendering children. Organization may not be
   // loaded yet when this first renders.
   const organization = useOrganization({allowNull: true});
+  const hasNavigationV2 = organization?.features.includes('navigation-sidebar-v2');
+  const App = hasNavigationV2 ? AppLayout : LegacyAppLayout;
```
natemoo-re, line 41: Something to watch out for: this note calls out that **"Organization may not be loaded yet when this first renders"**. 

For this experience to exist behind a feature flag, the organization has to be loaded. This could potentially cause some layout thrashing once the org loads.

static/app/views/organizationLayout/index.tsx
```diff
@@ -33,20 +37,57 @@ function OrganizationLayout({children}: Props) {
...
+  return (
+    <AppContainer className="app">
+      <Nav />
+      <BodyContainer>
+        {organization && <OrganizationHeader organization={organization} />}
+        {organization && <DevToolInit />}
+        <Body>{children}</Body>
+        <Footer />
+      </BodyContainer>
+    </AppContainer>
```
natemoo-re, line 66: Notable: adding a new wrapper for the main content allows us to remove the `position: fixed` CSS for the sidebar. This means that we no longer have to hardcode sidebar width and manually adjust padding for the content offset.

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,165 @@
...
+    }
+    const [pathname, search] = to.split('?');
+
+    return {
+      pathname,
+      search: search ? `?${search}` : undefined,
+      state: {source: SIDEBAR_NAVIGATION_SOURCE},
+    };
+  }, [to]);
+}
```
JonasBa, line 165: I dont think this needs to be a hook, the naming looks like something we would expose from react router and it's really doing a conversion from a to obj.  How about something like makeLocationDescriptorFromTo or something similar and just expose that, wrapping it inside a useMemo is trivial for consumers to add

static/app/types/hooks.tsx
```diff
@@ -58,6 +58,9 @@ export interface Hooks
 
 export type HookName = keyof Hooks;
 
+export type WithoutPrefix<Hook extends HookName> =
```
JonasBa, line 61:
```suggestion
export type HookWithoutPrefix<Hook extends HookName> =
```

static/app/components/nav/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+          ))}
+        </Sidebar.Footer>
+      </Sidebar>
+      {nav.secondary.body.length > 0 && (
+        <Submenu role="navigation" aria-label="Secondary Navigation">
+          <Submenu.Body>
+            {nav.secondary.body.map(item => (
+              <Submenu.Item key={item.label} {...item} />
+            ))}
+          </Submenu.Body>
```
JonasBa, line 47: Would it be possible to insert a <Sidebar.Divider/> or equiv inside rendering decl instead of handle it inside the config?

static/app/components/nav/config.tsx
```diff
@@ -0,0 +1,192 @@
...
+  ) ?? {
+    submenu: [],
+  };
+  const secondary = splitAtDivider(submenu);
+
+  return {
+    primary,
+    secondary,
+  };
+}
```
JonasBa, line 192: Is there a way we could simplify this logic entirely and remove the NAV_DIVIDER as well? I think it would have a lot of nice downstream effects on simplifying the type definitions and rendering

Right now the the splitting procedure assumes that there is only a single divider, but that rule is not enforced, so someone could think that adding a divider inside the config would just render a divider, but it would actually result in a wrong split between the groups and not rendering some items (or attempt to render a symbol and crash)

static/app/components/nav/config.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const sidebar = items
+    .filter(item => !!item)
+    .map(item => (typeof item === 'object' ? resolveSidebarItem(item) : item));
+  const primary = splitAtDivider(sidebar);
+  const {submenu = []} = primary.body.find(
+    item => getActiveStatus(item, context.location) !== ActiveStatus.INACTIVE
+  ) ?? {
+    submenu: [],
+  };
+  const secondary = splitAtDivider(submenu);
```
JonasBa, line 186: Why is this needed?

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,165 @@
...
+interface SidebarItemTopLevel extends SidebarItemBase {
+  to: string;
+  submenu?: never;
+}
+
+interface SidebarItemWithChildren extends SidebarItemBase {
+  submenu: (SubmenuItem | false)[];
+  to?: never;
+}
```
JonasBa, line 38: Nice 👌🏼

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,165 @@
...
+export function resolveSidebarItem(
+  item: SidebarItemTopLevel | SidebarItemWithChildren
+): SidebarItem {
+  if (item.submenu) {
+    const submenu = item.submenu.filter(subitem => !!subitem) as SubmenuItem[];
+    const [{to}] = submenu;
+    return {...item, submenu, to};
+  }
+  return item;
+}
```
JonasBa, line 60: Why do we need this?

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,165 @@
...
+export interface SubmenuItem extends NavItem {
+  to: string;
+}
+
+export type NavItemRaw =
+  | SidebarItemTopLevel
+  | SidebarItemWithChildren
+  | typeof NAV_DIVIDER
+  | undefined
+  | false;
```
JonasBa, line 49:
```suggestion
export type NavItemRaw =
  | SidebarItemTopLevel
  | SidebarItemWithChildren
```

Can we simplify this to something like this?

static/app/components/nav/config.tsx
```diff
@@ -0,0 +1,192 @@
...
+        check: {
+          features: ['discover', 'discover-query', 'dashboards-basic', 'dashboards-edit'],
+          hook: 'dashboards-sidebar-item',
+        },
+      },
+      {label: t('Alerts'), to: `/${prefix}/alerts/rules/`, icon: <IconSiren />},
+      NAV_DIVIDER,
+      // {label: t('Help'), to: '', icon: <IconQuestion />},
+      // {label: t('New'), to: '', icon: <IconBroadcast />},
+      // {label: t('Stats'), to: '', icon: <IconStats />},
```
JonasBa, line 159:
```suggestion
```

static/app/components/nav/sidebar.tsx
```diff
@@ -0,0 +1,187 @@
...
+    gap: ${space(1)};
+  }
+`;
+
+function Item({
+  to,
+  label,
+  icon,
+  submenu,
+  check,
```
JonasBa, line 67: Is there a way we can use the same signature as the Feature component, since that is a signature people folks are already familiar with?

static/app/components/nav/index.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  describe('sidebar links', () => {
+    beforeEach(function () {
+      ConfigStore.init();
+      ConfigStore.set('features', new Set([]));
+      ConfigStore.set('user', user);
+
+      mockUseLocation.mockReturnValue({...LocationFixture()});
+    });
+
```
JonasBa, line 75:
```suggestion
afterEach(() => ConfigStore.reset())
```
I think you can also reset the store

static/app/components/nav/config.tsx
```diff
@@ -0,0 +1,192 @@
...
+export interface NavItemsResult {
+  primary: {
+    body: SidebarItem[];
+    footer: SidebarItem[];
+  };
+  secondary: {
+    body: SubmenuItem[];
+    footer: SubmenuItem[];
+  };
+}
```
JonasBa, line 39:
```suggestion
export interface NavItemsResult {
  primary: {
    body: ReadonlyArray<SidebarItem>;
    footer: ReadonlyArray<SidebarItem>;
  };
  secondary: {
    body: ReadonlyArray<SidebarItem>;
    footer: ReadonlyArray<SidebarItem>;
  };
}
```
I think you want to make sure these are not mutated in any component as it would impact the rendering

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,165 @@
...
+export const NAV_DIVIDER = Symbol('divider');
+
+interface FeatureCheck {
+  features: string | string[];
+  hook?: WithoutPrefix<keyof FeatureDisabledHooks>;
+}
+
+interface NavItem {
+  label: string;
+  check?: FeatureCheck;
```
JonasBa, line 19: same comment about reusing the feature check signature if possible

static/app/components/nav/useIndicator.tsx
```diff
@@ -0,0 +1,45 @@
+import type {HTMLAttributes, PointerEvent} from 'react';
+import {useRef, useState} from 'react';
+import {useFocusWithin, useHover} from '@react-aria/interactions';
+
+export interface IndicatorResult {
+  containerProps: HTMLAttributes<HTMLUListElement>;
+  indicatorProps: HTMLAttributes<HTMLSpanElement>;
+}
+
+export function useIndicator() {
```
JonasBa, line 10:
```suggestion
export function useNavIndicator() {
```

static/app/views/organizationLayout/index.tsx
```diff
@@ -33,20 +37,59 @@ function OrganizationLayout({children}: Props) {
   // oganization is loaded before rendering children. Organization may not be
   // loaded yet when this first renders.
   const organization = useOrganization({allowNull: true});
+  const hasNavigationV2 = organization?.features.includes('navigation-sidebar-v2');
+  const App = hasNavigationV2 ? AppLayout : LegacyAppLayout;
```
JonasBa, line 41: Thank you for not using V2 to name the layout as well :D

static/app/components/nav/utils.tsx
natemoo-re: Good catch—no need for this to be a hook! Refactored and renamed in [3bb7468](https://github.com/getsentry/sentry/pull/77665/commits/3bb74680d614a2e6f6f17cbdfbf2af5dc4c08a83).

static/app/components/nav/useIndicator.tsx
```diff
@@ -0,0 +1,45 @@
...
+      animation.current = {duration: 0.01};
+      return;
+    }
+    if (hover.current === a) return;
+    animation.current = {type: 'spring', mass: 0.1, damping: 10, stiffness: 150};
+    hover.current = a;
+    const newY = a?.getBoundingClientRect()?.y ?? 0;
+    const offset = list.getBoundingClientRect()?.y ?? 0;
+    setY(newY - offset);
+  };
```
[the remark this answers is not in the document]
natemoo-re: Yep, will spend some time to rethink this. It should be fairly simple to hoist the `getBoundingClientRect` calls out of the event handler and use a [`MotionValue`](https://www.framer.com/motion/motionvalue/) to track the pointer position outside of the `state` lifecycle.

MichaelSun48:
Couple minor bugs I noticed: 

- The `< Collapse` button that was at the bottom of the sidebar is gone, so I can't uncollapse the side bar if I turn `hasNewNav` on with my sidebar already collapsed.  

- The user feedback item is overflowing out of the button's hover state


Screenshot 2024-09-23 at 9 11 38 AM

natemoo-re:
@MichaelSun48 Correct, this new design removes the collapse feature in favor of a stacked item layout!

Based on your screenshot, you might be testing an older iteration of this feature (the flag was originally introduced in https://github.com/getsentry/sentry/pull/76187).

static/app/components/nav/index.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+import {render, screen} from 'sentry-test/reactTestingLibrary';
+
+import Nav from 'sentry/components/nav';
+import ConfigStore from 'sentry/stores/configStore';
+import type {Organization} from 'sentry/types/organization';
+import {useLocation} from 'sentry/utils/useLocation';
+
+jest.mock('sentry/actionCreators/account');
+jest.mock('sentry/utils/useServiceIncidents');
+jest.mock('sentry/utils/useLocation');
```
scttcper, line 14: i'm not sure you need to mock useLocation if you aren't setting any params. Would prefer people mock location via the router context in our render function

https://github.com/getsentry/sentry/blob/master/tests/js/sentry-test/reactTestingLibrary.tsx#L63-L71

scttcper: here's a really small example

https://github.com/getsentry/sentry/blob/master/static/app/components/modals/commandPalette.spec.tsx#L91-L93

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const normalizedCurrent = normalizeUrl(location.pathname);
+  if (normalizedTo === normalizedCurrent) return 'active';
+  if (isItemActive({to: item.to, label: item.label})) return 'active';
+  if (
+    'submenu' in item &&
+    item.submenu?.find(
+      subitem =>
+        typeof subitem === 'object' &&
+        getNavigationItemStatus(subitem, location) !== 'inactive'
+    ) !== undefined
```
scttcper, line 84: nit could be .some
```suggestion
    item.submenu?.some(
      subitem =>
        typeof subitem === 'object' &&
        getNavigationItemStatus(subitem, location) !== 'inactive'
    )
```

static/app/components/nav/config.tsx
```diff
@@ -0,0 +1,191 @@
...
+        ],
+      },
+      {label: t('Projects'), to: `/${prefix}/projects/`, icon: <IconProject />},
+      {
+        label: t('Explore'),
+        icon: <IconSearch />,
+        submenu: [
+          {
+            label: t('Traces'),
+            to: `/traces/`,
```
evanpurkhiser, line 82: Why doesn't this have the prefix?

evanpurkhiser:
Would love to see more comments and documentation of util functions and components in this PR.

Since this is a foundational part of the app it would be really nice to see it well documented. I think it would also make it much easier to review and understand how all the pieces fit together with some comment guidance.

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,158 @@
...
+}
+
+interface SidebarItemWithChildren extends SidebarItemBase {
+  submenu: (SubmenuItem | false)[];
+  to?: never;
+}
+
+export interface SubmenuItem extends NavItem {
+  to: string;
+}
```
evanpurkhiser, line 42: let's add jsdoc comments to all these interfaces and their fields

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,158 @@
...
+  | SidebarItemWithChildren
+  | typeof NAV_DIVIDER
+  | undefined
+  | false;
+
+export function resolveSidebarItem(
+  item: SidebarItemTopLevel | SidebarItemWithChildren
+): SidebarItem {
+  if (item.submenu) {
+    const submenu = item.submenu.filter(subitem => !!subitem) as SubmenuItem[];
```
evanpurkhiser, line 55: Would be nice to use a guard function instead of a cast here

static/app/components/nav/mobile.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const body = document.body;
+    const app = document.querySelector('.app > :not(nav)');
+    if (view !== 'closed') {
+      app?.setAttribute('inert', '');
+      body.style.setProperty('overflow', 'hidden');
+    } else {
+      app?.removeAttribute('inert');
+      body.style.removeProperty('overflow');
+    }
+  }, [view]);
```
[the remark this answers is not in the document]
natemoo-re: Addressed in [29f21b2](https://github.com/getsentry/sentry/pull/77665/commits/29f21b2f3c09f2a66c985151df2cfe9589e8de08)

static/app/components/nav/useIndicator.tsx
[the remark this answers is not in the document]
natemoo-re: Refactored to remove `useState` and `getBoundingClientRect()` calls in [123bcca](https://github.com/getsentry/sentry/pull/77665/commits/123bccac6925c997cc1f3f1a80219256e86522af). Hook is now much easier to read and respects `prefers-reduced-motion`.

static/app/components/nav/utils.tsx
natemoo-re: Fixed in [a9c0f1a](https://github.com/getsentry/sentry/pull/77665/commits/a9c0f1aac5788b60053556463c6669a4bbcf4fe7)

static/app/components/nav/sidebar.tsx
natemoo-re: Resolved in [a9c0f1a](https://github.com/getsentry/sentry/pull/77665/commits/a9c0f1aac5788b60053556463c6669a4bbcf4fe7)

static/app/components/nav/config.tsx
natemoo-re: Great callout, better to use the standard `Feature` interface! Resolved in [a9c0f1a](https://github.com/getsentry/sentry/pull/77665/commits/a9c0f1aac5788b60053556463c6669a4bbcf4fe7)

static/app/components/nav/utils.tsx
natemoo-re: Nice catch! Addressed as part of [92d70b5](https://github.com/getsentry/sentry/pull/77665/commits/92d70b59ab5a518fa704b5483e958590d2fbe50c).

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,165 @@
...
+    item.submenu?.find(
+      subitem =>
+        typeof subitem === 'object' &&
+        getActiveStatus(subitem, location) !== ActiveStatus.INACTIVE
+    ) !== undefined
+  ) {
+    return ActiveStatus.ACTIVE_PARENT;
+  }
+  return ActiveStatus.INACTIVE;
+}
```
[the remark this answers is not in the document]
natemoo-re: Simplified this significantly as part of [92d70b5](https://github.com/getsentry/sentry/pull/77665/commits/92d70b59ab5a518fa704b5483e958590d2fbe50c).

static/app/components/nav/config.tsx
natemoo-re: Great feedback—this has been restructured as part of [92d70b5](https://github.com/getsentry/sentry/pull/77665/commits/92d70b59ab5a518fa704b5483e958590d2fbe50c).

static/app/components/nav/config.tsx
natemoo-re: Removed in [92d70b5](https://github.com/getsentry/sentry/pull/77665/commits/92d70b59ab5a518fa704b5483e958590d2fbe50c).

static/app/components/nav/utils.tsx
natemoo-re: Removed in [92d70b5](https://github.com/getsentry/sentry/pull/77665/commits/92d70b59ab5a518fa704b5483e958590d2fbe50c).

static/app/components/nav/utils.tsx
natemoo-re: Removed in [92d70b5](https://github.com/getsentry/sentry/pull/77665/commits/92d70b59ab5a518fa704b5483e958590d2fbe50c).

static/app/components/nav/config.tsx
natemoo-re: Removed in [92d70b5](https://github.com/getsentry/sentry/pull/77665/commits/92d70b59ab5a518fa704b5483e958590d2fbe50c).

static/app/components/nav/context.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const submenu = useMemo<NavContext['submenu']>(() => {
+    for (const item of config.main) {
+      if (isNavItemActive(item, location) || isSubmenuItemActive(item, location)) {
+        if (!item.submenu) {
+          return undefined;
+        }
+        if (Array.isArray(item.submenu)) {
+          return {main: item.submenu};
+        }
+        return item.submenu;
```
saponifi3d, line 41: Could we simplify this to? (assuming if statement is doing some amount of nullish coalesce)
```suggestion
        if (Array.isArray(item.submenu)) {
          return {main: item.submenu};
        }
        return item.submenu ?? undefined;
```

static/app/components/nav/context.tsx
```diff
@@ -0,0 +1,61 @@
...
+    if (config.footer) {
+      for (const item of config.footer) {
+        if (isNavItemActive(item, location) || isSubmenuItemActive(item, location)) {
+          if (!item.submenu) {
+            return undefined;
+          }
+          if (Array.isArray(item.submenu)) {
+            return {main: item.submenu};
+          }
+          return item.submenu;
```
saponifi3d, line 53: nit - can DRY this into a function that takes the `item.submenu` - then you could move the nullasce coalesce to undefined as a default prop.

```jsx
const getSubmenu = (submenu: NavItem[] | undefined = undefined) => {
  if (Array.isArray(item.submenu) return {main: submenu}
  return submenu;
}
```
(also a guess at the types 😅)

static/app/components/nav/index.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+  'custom-metrics',
+  'user-feedback-ui',
+  'session-replay-ui',
+  'performance-view',
+  'performance-trace-explorer',
+  'starfish-mobile-ui-module',
+  'profiling',
+];
+
+describe('Nav', function () {
```
saponifi3d, line 25: <3 these tests, super well structured and easy to find why failures occurred.

static/app/components/nav/sidebar.tsx
```diff
@@ -0,0 +1,220 @@
...
+      </SidebarWrapper>
+      <Submenu />
+    </Fragment>
+  );
+}
+
+export default Sidebar;
+
+export function SidebarItems() {
+  const nav = useNavContext();
```
saponifi3d, line 43: nit: could destruct `config` from nav since it's the only item in use
```suggestion
  const { config } = useNavContext();
```
Only good reason for this is that it improves obfuscation passes.

static/app/components/nav/sidebar.tsx
```diff
@@ -0,0 +1,220 @@
...
+            <SidebarItem key={item.label} item={item} />
+          ))}
+        </SidebarFooter>
+      )}
+    </Fragment>
+  );
+}
+
+const SidebarWrapper = styled('div')`
+  height: 40px;
```
saponifi3d, line 63: nit: would be cool if we used a `rem` or something relative to the base font-size rather than pixels.

static/app/components/nav/sidebar.tsx
```diff
@@ -0,0 +1,220 @@
...
+  width: 100vw;
+  padding: ${space(2)} 0;
+  border-right: 1px solid ${theme.translucentGray100};
+  background: #3e2648;
+  background: linear-gradient(180deg, #3e2648 0%, #442c4e 100%);
```
saponifi3d, line 68: 👃 for the hardcoded colors, idk if we have anything close to these in our theme / styleguide tho. would be cool if we could use something from there though.

static/app/components/nav/sidebar.tsx
```diff
@@ -0,0 +1,220 @@
...
+  display: flex;
+  flex-direction: column;
+  z-index: ${theme.zIndex.sidebar};
+
+  @media screen and (min-width: ${p => p.theme.breakpoints.medium}) {
+    height: unset;
+    width: 74px;
```
saponifi3d, line 75: same px nit vs rem.

static/app/components/nav/sidebar.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+const SidebarIndicator = styled(motion.span)`
+  position: absolute;
+  left: 0;
+  right: 0;
+  opacity: 0;
+  pointer-events: none;
+  margin-inline: ${space(1)};
+  width: 58px;
+  height: 53px;
```
saponifi3d, line 180: just want to confirm this is hardcoded to like the icon size? (otherise, anyway i could trick you into 54px? 🤣 - odd numbers always have a way of turning into fractions for me.)

static/app/components/nav/submenu.tsx
```diff
@@ -0,0 +1,158 @@
...
+import {useLocation} from 'sentry/utils/useLocation';
+
+function Submenu() {
+  const nav = useNavContext();
+  if (!nav.submenu) {
+    return null;
+  }
+
+  return (
+    <SubmenuWrapper role="navigation" aria-label="Secondary Navigation">
```
saponifi3d, line 26: 🎉 aria-labels!? let's go!

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,170 @@
...
+function hasMatchingQueryParam(
+  item: Required<Pick<NavSidebarItem | NavSubmenuItem, 'to' | 'label'>>,
+  location: ReturnType<typeof useLocation>
+): boolean {
+  if (location.search.length === 0) {
+    return false;
+  }
+  if (item.to.includes('?')) {
+    const search = new URLSearchParams(location.search);
+    const itemSearch = new URLSearchParams(item.to.split('?').at(-1));
```
saponifi3d, line 144: nice use of `URLSearchParams`!! 🎉

static/app/views/issueList/utils.tsx
```diff
@@ -217,3 +217,32 @@ export const FOR_REVIEW_QUERIES: string[] = [Query.FOR_REVIEW];
...
+
+function getIssueGroupFilter(group: IssueGroup): string {
+  if (!Object.hasOwn(IssueGroupFilter, group)) {
+    throw new Error(`Unknown issue group "${group}"`);
+  }
+  return IssueGroupFilter[group];
+}
+
+export function getSearchForIssueGroup(group: IssueGroup): string {
+  return `?query=is:unresolved+${getIssueGroupFilter(group)}`;
```
saponifi3d, line 247: nit: we should url encode the issue group filter - but we may want to construct this with `URLSearchParams` as well for additional safety.

saponifi3d (commented):
i pretty much just had nits. for the css - i'm curious about using hardcoded pixels vs rem (we should totally do whatever is common practice tho).

oh, side thing... it could be helpful to also have a video in the PR review to reference to see the styling / animations etc.

---

thanks for putting this all together though, 🎉 looks great. since there have been a few other reviewers on this pr, i'll hold off on +1'ing until there's more time for re-reviews.

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,170 @@
+import type {ComponentProps} from 'react';
+import type {LocationDescriptor} from 'history';
+
+import type Feature from 'sentry/components/acl/feature';
+import {isItemActive} from 'sentry/components/sidebar/sidebarItem';
```
JonasBa, line 5:
```suggestion
import {isNavItemActive} from 'sentry/components/sidebar/sidebarItem';
```
Given that most things now seem Nav namespaced

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,170 @@
...
+/** NavItems are displayed in either `main` and `footer` sections */
+export interface NavItemLayout<Item extends NavItem> {
+  main: Item[];
+  footer?: Item[];
+}
+
+/** SidebarItem is a top-level NavItem which is always displayed in the app sidebar */
+export interface NavSidebarItem extends NavItem {
+  /** The icon to render in the sidebar */
+  icon: JSX.Element;
```
JonasBa, line 27:
```suggestion
  icon: JSX.Element;
```
Question: does this need to be a JSX.Element or could it be a React.ReactNode?
If we dont want falsy icon values, it could be something like NonNullable<React.ReactNode>

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,170 @@
...
+    return item.submenu.some(subitem => isNavItemActive(subitem, location));
+  }
+  return (
+    item.submenu.main.some(subitem => isNavItemActive(subitem, location)) ||
+    item.submenu.footer?.some(subitem => isNavItemActive(subitem, location)) ||
+    false
+  );
+}
+
+/** Creates a well-formed `LocationDescriptor` from a partial URL string that may contain search params */
```
JonasBa, line 101:
```suggestion
/** Creates a `LocationDescriptor` from a URL string that may contain search params */
```

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,170 @@
...
+export function makeLocationDescriptorFromTo(to: string): LocationDescriptor {
```
JonasBa, line 102: Just to be extra safe, I think we should `if(!URL.canParse(to)) throw new Error` so we dont end up with broken links somewhere without knowing :D

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,170 @@
...
+      }
+      return match;
+    }
+  }
+  return false;
+}
+
+function isSidebarItem(item: NavSidebarItem | NavSubmenuItem): item is NavSidebarItem {
+  return Object.hasOwn(item, 'icon');
+}
```
JonasBa, line 170: 🚀 for this type guard!

JonasBa (approved):
Mostly low priority comments, but this looks much simpler than the previous version + some of the comments explaining the exceptions behind routing on issues page are just really nice to read as they give helpful context to anyone having to make changes to this part of the app

static/app/components/nav/mobileTopbar.tsx
```diff
@@ -0,0 +1,124 @@
...
+  left: 0;
+  display: flex;
+  flex-direction: column;
+  background: ${p => p.theme.surface300};
+  z-index: ${p => p.theme.zIndex.modal};
+  --color: ${p => p.theme.textColor};
+  --color-hover: ${p => p.theme.activeText};
+`;
+
+let appContainer: HTMLDivElement | null = null;
```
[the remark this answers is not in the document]
natemoo-re: Nice catch! Addressed in 393b55e4016

static/app/components/nav/utils.tsx
natemoo-re: Ah, I looked into this and unfortunately `URL.canParse()` doesn't help here—it assumes a fully formed URL with scheme, so a relative or absolute path won't pass. 

If we detect that and try to add a scheme ourselves, almost anything will be treated as valid.

JonasBa: Ahh, I see! Yeah, makes sense in that case, the unknown url route still sends an issue to Sentry so we should be covered here, thanks for the context!

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,170 @@
...
+  if (!item.submenu) {
+    return false;
+  }
+  if (isNonEmptyArray(item.submenu)) {
+    return item.submenu.some(subitem => isNavItemActive(subitem, location));
+  }
+  return (
+    item.submenu.main.some(subitem => isNavItemActive(subitem, location)) ||
+    item.submenu.footer?.some(subitem => isNavItemActive(subitem, location)) ||
+    false
```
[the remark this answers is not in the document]
JonasBa: I think this wont work, the second set will return undefined and wont align with the type, please discard the comment

static/app/components/nav/utils.tsx
```diff
@@ -0,0 +1,170 @@
...
+  label: string;
+  /** Optionally, props which should be passed to a wrapping `<Feature>` guard */
+  feature?: Omit<ComponentProps<typeof Feature>, 'children'>;
+}
+
+/** NavItems are displayed in either `main` and `footer` sections */
+export interface NavItemLayout<Item extends NavItem> {
+  main: Item[];
+  footer?: Item[];
+}
```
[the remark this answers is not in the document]
natemoo-re: I opted to keep the generic but update it to use the suggested union in 4e5864ddcbf. That seemed cleaner because it won't require guards when accessing individual elements.

static/app/components/nav/index.spec.tsx
natemoo-re: Thanks for the test feedback! These have been greatly simplified in [7760e83](https://github.com/getsentry/sentry/pull/77665/commits/7760e834027dfd6ff9998f29be7056469ca5908e).

static/app/components/nav/context.tsx
natemoo-re: Addressed in cf98926d685

static/app/components/nav/context.tsx
natemoo-re: Addressed in cf98926d685!

static/app/components/nav/sidebar.tsx
natemoo-re: Nice one. Added in c0b98b054f0!

static/app/views/issueList/utils.tsx
natemoo-re: Great catch! Addressed in 4f33ddfc0b5

static/app/components/nav/sidebar.tsx
natemoo-re: Fully agree, but don't want to break from the codebase standard. Will leave as `px` for now 😞

static/app/components/nav/sidebar.tsx
natemoo-re: Agree! We don't seem to have anything remotely close, unfortunately. I added a comment in 959e39dccc9 for future ref.

static/app/components/nav/sidebar.tsx
natemoo-re: Yep, intentionally matching the icon size. Updated to `52px` in bb9d8210aa6 (since that is a nice, round `3.25rem`)

static/app/components/nav/mobileTopbar.tsx
```diff
@@ -0,0 +1,124 @@
...
+/** When the mobile menu opens, set the main content to `inert` and disable `body` scrolling */
+function updateNavStyleAttributes(view: NavView) {
+  appContainer = appContainer ?? document.querySelector('[data-content]');
```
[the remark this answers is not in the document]
natemoo-re: Addressed in 0c5c8ad0e73 and 14ed7b9d7ed

static/app/components/nav/sidebar.tsx
```diff
@@ -0,0 +1,220 @@
...
+  position: relative;
+  display: flex;
+  justify-content: center;
+  margin-bottom: ${space(1.5)};
+`;
+
+function SidebarBody({children}) {
+  const {indicatorProps, containerProps} = useNavIndicator();
+  return (
+    <div>
```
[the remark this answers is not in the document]
natemoo-re: Great catch, added a comment in cca7b995cbf

static/app/components/nav/useNavIndicator.tsx
```diff
@@ -0,0 +1,55 @@
...
+      ? {duration: 0}
+      : {
+          mass: 0.5,
+          damping: 15,
+          stiffness: 200,
+          restDelta: 0.05,
+        }
+  );
+  const {isHovered, hoverProps} = useHover({});
+  const {focusWithinProps} = useFocusWithin({});
```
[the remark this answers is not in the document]
natemoo-re: I expected these hooks to allow not passing anything, but they unfortunately require an options `object`, but it can be empty. The overhead should be quite negligible in practice.

static/app/components/nav/useNavIndicator.tsx
```diff
@@ -0,0 +1,55 @@
...
+    }
+  };
+
+  const handlePointerEnter = (e: PointerEvent<HTMLUListElement>) => {
+    if (e.target instanceof HTMLUListElement) {
+      y.jump(e.target.querySelector('a')!.offsetTop);
+    } else if (e.target instanceof HTMLElement) {
+      y.jump(e.target.offsetTop);
+    }
+  };
```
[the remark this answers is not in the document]
natemoo-re: Yep! Nice find. Done in 86eadad9941

static/app/components/nav/sidebar.tsx
```diff
@@ -0,0 +1,220 @@
...
+export function SidebarItems() {
+  const nav = useNavContext();
+  return (
+    <Fragment>
+      <SidebarBody>
+        {nav.config.main.map(item => (
+          <SidebarItem key={item.label} item={item} />
+        ))}
+      </SidebarBody>
+      {isNonEmptyArray(nav.config.footer) && (
```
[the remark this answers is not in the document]
natemoo-re: I agree! I will come back to address this in a future PR.

saponifi3d (approved):
:shipit: looks great!

PR #77667, opened 2024-09-17
feat(invite-members): backend for Resend Invites

ameliahsu:
update API to allow members to resend invites that they sent

we error if the request contains additional body params other than `reinvite` or `regenerate`

### DIFF
--- a/src/sentry/api/endpoints/organization_member/details.py
+++ b/src/sentry/api/endpoints/organization_member/details.py
@@ -235,7 +235,7 @@ class OrganizationMemberDetailsEndpoint(OrganizationMemberEndpoint):
             if not enable_member_invite:
                 return Response({"detail": ERR_INSUFFICIENT_SCOPE}, status=400)
             if not reinvite_request_only:
-                raise ValidationError({"detail": ERR_MEMBER_REINVITE})
+                return Response({"detail": ERR_MEMBER_REINVITE}, status=400)
             if not is_invite_from_user:
                 return Response({"detail": ERR_MEMBER_INVITE}, status=400)
 
--- a/src/sentry/api/endpoints/organization_member/details.py
+++ b/src/sentry/api/endpoints/organization_member/details.py
@@ -44,7 +44,8 @@ from . import get_allowed_org_roles, save_team_assignments
 ERR_NO_AUTH = "You cannot remove this member with an unauthenticated API request."
 ERR_INSUFFICIENT_ROLE = "You cannot remove a member who has more access than you."
 ERR_INSUFFICIENT_SCOPE = "You are missing the member:admin scope."
-ERR_MEMBER_INVITE = "Your role cannot remove an invitation that was sent by someone else."
+ERR_MEMBER_INVITE = "Your cannot modify invitations sent by someone else."
+ERR_MEMBER_REINVITE = "You can only reinvite members; you cannot modify other member details."
 ERR_ONLY_OWNER = "You cannot remove the only remaining owner of the organization."
 ERR_UNINVITABLE = "You cannot send an invitation to a user who is already a full member."
 ERR_EXPIRED = "You cannot resend an expired invitation without regenerating the token."
@@ -75,7 +76,7 @@ class RelaxedMemberPermission(OrganizationPermission):
     scope_map = {
         "GET": ["member:read", "member:write", "member:admin"],
         "POST": ["member:write", "member:admin"],
-        "PUT": ["member:write", "member:admin"],
+        "PUT": ["member:write", "member:admin", "member:invite"],
         # DELETE checks for role comparison as you can either remove a member
         # with a lower access role, or yourself, without having the req. scope
         "DELETE": ["member:read", "member:write", "member:admin"],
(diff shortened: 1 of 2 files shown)


src/sentry/api/endpoints/organization_member/details.py
```diff
@@ -218,6 +219,26 @@ def put(
...
+        # Members can only resend invites
+        reinvite_request_only = set(result.keys()).issubset({"reinvite", "regenerate"})
+        # Members can only resend invites that they sent
+        is_invite_from_user = member.inviter_id == request.user.id
+
+        if is_member:
+            if not enable_member_invite:
+                return Response({"detail": ERR_INSUFFICIENT_SCOPE}, status=400)
+            if not reinvite_request_only:
+                raise ValidationError({"detail": ERR_MEMBER_REINVITE})
```
sentaur-athena, line 238: why is this a validation error vs. a 400?

src/sentry/api/endpoints/organization_member/details.py
```diff
@@ -372,7 +393,7 @@ def _handle_deletion_by_member(
         acting_member: OrganizationMember,
     ) -> Response:
         # Members can only delete invitations
-        if not member.inviter_id:
+        if not member.is_pending:
```
sentaur-athena, line 396: shouldn't this be both checks? inviter id and pending?

src/sentry/api/endpoints/organization_member/details.py
ameliahsu: not sure if there is a convention for this, but I thought a ValidationError would make it more clear that the input is not valid. I believe ValidationError still returns a 400 as well

src/sentry/api/endpoints/organization_member/details.py
ameliahsu: Updated this because I think `is_pending` is the more correct way of checking if the member is still a pending invite, rather than checking if `inviter_id` still exists

src/sentry/api/endpoints/organization_member/details.py
```diff
@@ -218,6 +219,26 @@ def put(
...
+            and not organization.flags.disable_member_invite
+        )
+        # Members can only resend invites
+        reinvite_request_only = set(result.keys()).issubset({"reinvite", "regenerate"})
+        # Members can only resend invites that they sent
+        is_invite_from_user = member.inviter_id == request.user.id
+
+        if is_member:
+            if not enable_member_invite:
+                return Response({"detail": ERR_INSUFFICIENT_SCOPE}, status=400)
```
sentaur-athena, line 236: 400 and validation error mean `Bad Request` basically like you passed something in wrong format. Here we need some other status like insufficient permission 403. I think 403 is more appropriate for all the 3

src/sentry/api/endpoints/organization_member/details.py
```diff
@@ -44,7 +44,8 @@
 ERR_NO_AUTH = "You cannot remove this member with an unauthenticated API request."
 ERR_INSUFFICIENT_ROLE = "You cannot remove a member who has more access than you."
 ERR_INSUFFICIENT_SCOPE = "You are missing the member:admin scope."
-ERR_MEMBER_INVITE = "Your role cannot remove an invitation that was sent by someone else."
+ERR_MEMBER_INVITE = "Your cannot modify invitations sent by someone else."
```
schew2381, line 47:
```suggestion
ERR_MEMBER_INVITE = "You cannot modify invitations sent by someone else."
```

src/sentry/api/endpoints/organization_member/details.py
```diff
@@ -75,7 +76,7 @@ class RelaxedMemberPermission(OrganizationPermission):
     scope_map = {
         "GET": ["member:read", "member:write", "member:admin"],
         "POST": ["member:write", "member:admin"],
-        "PUT": ["member:write", "member:admin"],
+        "PUT": ["member:write", "member:admin", "member:invite"],
```
schew2381, line 79: tiny nit: keep in order of access
```suggestion
        "PUT": ["member:invite", "member:write", "member:admin"],
```

src/sentry/api/endpoints/organization_member/details.py
```diff
@@ -218,6 +219,26 @@ def put(
...
+            and not organization.flags.disable_member_invite
+        )
+        # Members can only resend invites
+        reinvite_request_only = set(result.keys()).issubset({"reinvite", "regenerate"})
+        # Members can only resend invites that they sent
+        is_invite_from_user = member.inviter_id == request.user.id
+
+        if is_member:
+            if not enable_member_invite:
+                return Response({"detail": ERR_INSUFFICIENT_SCOPE}, status=403)
```
schew2381, line 236: The err msg `ERR_INSUFFICIENT_SCOPE` mentions requiring `member:admin` which is not true, so we can just raise a generic err
```suggestion
                raise PermissionDenied
```

tests/sentry/api/endpoints/test_organization_member_details.py
schew2381, line 1: nit: I think for these tests it'd be good to assert the msg we're getting back from the err response to make sure we're exiting at the expected place

schew2381 (approved):
overall lgtm, just some small nits

PR #77675, opened 2024-09-18
feat(toolbar): add project settings page w/allowed_origins form

### LINKED ISSUE #77213: [DevToolbar] Create a settings page for the Toolbar
The page should be behind a feature flag!

The settings should be per-project.

To start we should let people have a text area to save a list of domains (new-line separated) that are allowed to host the toolbar. For sentry itself I'd expect the list to be: 
```
*.sentry.io
dev.getsentry.net:8000
*.dev.getsentry.net:7999
*.sentry.dev
```
which represents:
- prod
- local devapp
- local dev-ui
- previews on CI

It'll be important for people to include the port as well, so maybe we'll make a note of that. protocol is not important.

aliu39:
Depends on https://github.com/getsentry/sentry/pull/77674, closes https://github.com/getsentry/sentry/issues/77213

Notice the route on top + in the sidebar. Users need to manually click "Save Changes", which is only enabled if there were changes since the last save.
Screenshot 2024-09-17 at 5 19 40 PM

TODO: format and validate before submission. Can use regex or whatever we use for similar forms, [inbound filters](https://sentry.sentry.io/settings/projects/javascript/filters/data-filters/#filters-react-hydration-errors_help) is one
Questions:
- yes or no to trailing `/`'s?
- require protocol or fill in one by default? E.g. `https://`

https://github.com/getsentry/sentry/pull/77674 sets `["*"]` (allow all) as the default. Whatever it is, we should probably show a default value in the form to be consistent.

### DIFF
--- a/static/app/views/settings/project/navigationConfiguration.tsx
+++ b/static/app/views/settings/project/navigationConfiguration.tsx
@@ -67,7 +67,7 @@ export default function getConfiguration({
         {
           path: `${pathPrefix}/toolbar/`,
           title: t('Developer Toolbar'),
-          show: () => true || !!organization?.features?.includes('dev-toolbar-ui'),
+          show: () => !!organization?.features?.includes('dev-toolbar-ui'),
         },
       ],
     },
--- a/static/app/views/settings/project/navigationConfiguration.tsx
+++ b/static/app/views/settings/project/navigationConfiguration.tsx
@@ -64,6 +64,11 @@ export default function getConfiguration({
           title: t('User Feedback'),
           show: () => !isSelfHostedErrorsOnly,
         },
+        {
+          path: `${pathPrefix}/toolbar/`,
+          title: t('Developer Toolbar'),
+          show: () => true || !!organization?.features?.includes('dev-toolbar-ui'),
+        },
       ],
     },
     {
--- a/static/app/routes.tsx
+++ b/static/app/routes.tsx
@@ -557,6 +557,11 @@ function buildRoutes() {
         name={t('Replays')}
         component={make(() => import('sentry/views/settings/project/projectReplays'))}
       />
+      <Route
+        path="toolbar/"
+        name={t('Developer Toolbar')}
+        component={make(() => import('sentry/views/settings/project/toolbar'))}
+      />
       <Route path="source-maps/" name={t('Source Maps')}>
         <IndexRoute
           component={make(() => import('sentry/views/settings/projectSourceMaps'))}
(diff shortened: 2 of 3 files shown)


ryan953:
>Questions:
> yes or no to trailing /'s?
> require protocol or fill in one by default? E.g. https://

The data will be compared to the origin of `referer` header (so like `request.headers['Referer']` in django).
The referrer could be any full path: `https://example.com/page/?q=123` is a valid referer value, and the origin of that would be `https://example.com`

[MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin) says the syntax for origin is:
```
Origin: null
Origin: <scheme>://<hostname>
Origin: <scheme>://<hostname>:<port>
```
So protocol yes, seems needed. We could probably assume `http://` if you want to get fancy & flexible with it. 
Trailing slash is not needed, but again if you want to keep it flexible for folks auto-format the textbox, or strip it out later.

---
Another way to think about this is we're accepting a list of urls into this textbox... and then later we'll compare the origin of the referrer to the origin of the url's that the user saved into this textbox.

ryan953:
A valid strategy is to make the error message as good as possible, and just force the user to type it in perfectly to start. It's possible to followup in another PR in order to get fancier and make the experience nicer for folks.

static/app/views/settings/project/navigationConfiguration.tsx
```diff
@@ -64,6 +64,11 @@ export default function getConfiguration({
           title: t('User Feedback'),
           show: () => !isSelfHostedErrorsOnly,
         },
+        {
+          path: `${pathPrefix}/toolbar/`,
+          title: t('Developer Toolbar'),
+          show: () => true || !!organization?.features?.includes('dev-toolbar-ui'),
```
ryan953, line 70: TODO: don't ship the `true ||` part when the time comes.

ryan953:
```suggestion
          show: () => organization?.features?.includes('dev-toolbar-ui'),
```

aliu39: Good catch 😅

aliu39:
About the last commit - the whole settings page will be hidden if the FF (dev-toolbar-ui) is missing. No need to check again for the allowed origins form.

aliu39:
Agreed with @ryan953 to merge this and validate/format the input in a follow-up

ryan953 (approved):

static/app/views/settings/project/navigationConfiguration.tsx
```diff
@@ -64,6 +64,11 @@ export default function getConfiguration({
           title: t('User Feedback'),
           show: () => !isSelfHostedErrorsOnly,
         },
+        {
+          path: `${pathPrefix}/toolbar/`,
+          title: t('Developer Toolbar'),
+          show: () => !!organization?.features?.includes('dev-toolbar-ui'),
```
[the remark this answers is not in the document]
aliu39: Because of the `?`'s this statement might return undefined. 
> in js !! will test the truthiness of any value and convert it to a boolean. Similar to how Boolean(var) works.

Is this still correct though?

PR #77678, opened 2024-09-18
ci: Ensure Sentry SDK is always bumped by dependabot

mydea:
I was wondering why the SDK is not automatically kept up-to-date by dependabot. After looking into it, I guess the problem is that the limit is set to 10 open PRs that are generally there (some have been open for months), so there is no capacity to open new ones.

As we'd love to have the SDK be up to date generally, this PR splits the SDK updates out to not have a limit. Hopefully that should ensure that we generally have this at latest.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,9 +18,6 @@ updates:
       babel-dependencies:
         patterns:
           - '@babel/*'
-      sentry-dependencies:
-        patterns:
-          - '@sentry/*'
       spectrum-dependencies:
         patterns:
           - '@react-stately/*'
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -63,6 +60,27 @@ updates:
       - dependency-name: 'reflux'
       - dependency-name: 'sprintf-js'
       - dependency-name: 'u2f-api'
+  # For Sentry SDK updates, we do not want to limit the open PRs
+  - package-ecosystem: npm
+    open-pull-requests-limit: 0
```
asottile-sentry, line 65: I don't think it works like this -- last we tried you can only have one `npm` section in a dependabot config

mydea: ah damn, that sucks - I thought that would work :( does that mean it is not possible to realize this? 😬 
What are thoughts about increasing the PR limit (10), or finding a different way to deal with those? As of now dependabot seems basically blocked because of open PRs with low prio 🤔

asottile-sentry: afaik for packages we care about we've just been doing them periodically on our own.  dependabot isn't very good at bumping packages anyway (lol I know right?) so it's been smoother when a person does it

mydea: Yeah I mean this is what we kind of did so far, but we always lag behind a bit with the SDK and it would be nice to be up-to-date there, as it is our premier dogfooding app as well 🤔

billyvg: Yeah doesn't seem to be supported: https://github.com/dependabot/dependabot-core/issues/1778

Though looks like [there's a workaround](https://github.com/dependabot/dependabot-core/issues/1778#issuecomment-1988140219) if you have one config set `target-branch: master`

scttcper:
i usually see it make one for sentry dependencies. there's one open here https://github.com/getsentry/sentry/pull/77287

PR #77687, opened 2024-09-18
fix(traces-spans-table): Loading state on refetch

ArthurKnaus:
#### Before

The loading state is always shown, even when cached data is being rendered.

https://github.com/user-attachments/assets/dc3b74ab-69ed-44c7-9b32-bbaddf6d48eb

#### After

The loading state is only shown on the first load

https://github.com/user-attachments/assets/3099d436-d329-4e23-a3dc-20199812b17a


ArthurKnaus:
Stumbled over this while testing different product features.
Feel free to close the PR if this is not the way you want to solve this :)

Zylphrex (approved):

PR #77698, opened 2024-09-18
fix(relocation) resolve stuck outbox

markstory:
There is an outbox in production that has been stuck for a while as the file it needs cannot be found or has been removed/purged. This patch will ack the outbox and we'll need to manually recover the relocation attempt

Fixes SENTRY-3CS1


corps:
@markstory Should the outbox logic catch the Not Found exception and /always/ discard it?  This seems like it will happen again, and it seems wise the outboxes are robust against them.

markstory:
> Should the outbox logic catch the Not Found exception and /always/ discard it? 

I'm not confident on whether this should be an always scenario. Alex is out on parental leave, and I think he would have opinions on how this should behave long term.

corps (approved):

PR #77700, opened 2024-09-18
ref(releases): move create repos/authors inside lock

JoshFerge:
Move these calls inside the redis lock but outside the database transaction


armenzg (approved):
Any tests that need updating? (e.g. related to not using the flag anymore)

JoshFerge:
> Any tests that need updating? (e.g. related to not using the flag anymore)

planning on continuing to use the flag for future refactors / quick rollbacks, so keeping those around.

PR #77702, opened 2024-09-18
ref(onboarding): Move astro package manager info code

priscilawebdev:
This code can live in the astro file 

image


static/app/gettingStartedDocs/javascript/astro.tsx
```diff
@@ -95,10 +98,27 @@ const getInstallConfig = () => [
...
+            astroLink: <ExternalLink href="https://astro.build/" />,
+          }
+        )}
+      </div>
+      <TextBlock noMargin>
+        {tct("In this quick guide you'll use the [astrocli:astro] CLI to set up:", {
+          astrocli: <strong />,
+        })}
+      </TextBlock>
+    </Fragment>
```
ArthurKnaus, line 120: I wonder if we should just style the div that wraps the introduction in `OnboardingLayout` instead of repeating the styles every time we want to show more than one paragraph 🤔

ArthurKnaus: e.g. we could do this in the docs:
```suggestion
  introduction: () => (
    <Fragment>
      <p >
        {tct(
          "Sentry's integration with [astroLink:Astro] supports Astro 3.0.0 and above.",
          {
            astroLink: <ExternalLink href="https://astro.build/" />,
          }
        )}
      </p>
      <p>
        {tct("In this quick guide you'll use the [astrocli:astro] CLI to set up:", {
          astrocli: <strong />,
        })}
      </p>
    </Fragment>
```

and in `OnboardingLayout` we could have something like:

```
const IntroductionWrapper = styled('div')`
  & > p:not(:last-child) {
    margin-bottom: spacing(1)
  }
`
```

This would also make it easier to re-use when the onboarding docs are rendered somewhere else.

wdyt?

ArthurKnaus (approved):

Lms24 (approved):

static/app/gettingStartedDocs/javascript/astro.tsx
priscilawebdev: sure. that is a good idea! thanks.

PR #77703, opened 2024-09-18
ref: set relay fixture scope to module instead of session

asottile-sentry:
this should be a reasonable balance between the current state (relay lasting for the whole session and causing flakiness) and function (setting up and tearing down the container continually)


asottile-sentry:
test output definitely seems a lot quieter after this change

jjbayer (approved):
Does it make tests any slower?

Dav1dde (approved):

PR #77714, opened 2024-09-18
fix(deletion): Fix test to delete Issue Platform issue

armenzg:
My original PR (#77653) was not correct.


tests/sentry/deletions/test_group.py
```diff
@@ -195,11 +196,30 @@ def test_delete_groups_delete_grouping_records_by_hash(
...
+            # We are using ReplayDeadClickType as a representative of Issue Platform
+            type=ReplayDeadClickType.type_id,
+            event_data={
+                "fingerprint": ["issue-platform-group"],
+                "timestamp": before_now(minutes=1).isoformat(),
+            },
+        )
+        assert group_info is not None
+        issue_platform_group = group_info.group
+        assert len(Group.objects.all()) == 2
```
wedamija, line 211: I don't think we need to check the size of the group table here. Maybe check that `event.group_id != issue_platform_group.id` instead?

wedamija (approved):

PR #77720, opened 2024-09-18
fix(insights): Incorrect comma placement in MongoDB queries

0Calories:
Fixes a bug where a comma separator would not be added when formatting MongoDB queries, when the original query JSON has the operation entry as the very last entry in the object

### Before
```
"findOne": "my_collection""first_key": "first_value", "second_key": "second_value",
```

### After
```
"findOne": "my_collection", "first_key": "first_value", "second_key": "second_value"
```


DominikB2014 (approved):
lgtm and nice tests!

PR #77721, opened 2024-09-18
chore(deletions) Move deletions code to mypy strong list

markstory:
Add missing type information to get deletions into the strong typing list. I've also removed the `dependencies` functionality from `DeletionTaskManager` as it was unused.

Refs #77479


src/sentry/deletions/base.py
```diff
@@ -125,7 +145,6 @@ def delete_bulk(self, instance_list) -> bool:
         self.mark_deletion_in_progress(instance_list)
 
         child_relations = self.get_child_relations_bulk(instance_list)
-        child_relations = self.extend_relations_bulk(child_relations, instance_list)
```
armenzg, line 128: Is this not needed anymore?

src/sentry/deletions/base.py
```diff
@@ -134,41 +153,50 @@ def delete_bulk(self, instance_list) -> bool:
...
             if child_relations:
                 has_more = self.delete_children(child_relations)
                 if has_more:
                     return has_more
 
-        return self.delete_instance_bulk(instance_list)
+        self.delete_instance_bulk(instance_list)
 
-    def delete_instance(self, instance):
+        return False
```
armenzg, line 164: What does `False` mean for this function? That it did not delete?

armenzg (approved):

src/sentry/deletions/base.py
markstory: Nope. No deletion tasks have implemented this method.

src/sentry/deletions/base.py
markstory: The `delete_bulk` method returns a boolean to indicate if there are more records to delete. This return value is mostly unused as `chunk()` uses a separate query to determine if all the data is removed.

https://github.com/getsentry/sentry/blob/b2e0b6e5ba549e28ccbeca3b78f28f7db216d701/src/sentry/deletions/base.py#L193-L214

PR #77723, opened 2024-09-18
feat(uptime): Add validator fields for method, body, header

davidenwang:
Adds new fields to the validator matching up with https://github.com/getsentry/sentry/pull/77611

### DIFF
--- a/src/sentry/uptime/endpoints/validators.py
+++ b/src/sentry/uptime/endpoints/validators.py
@@ -14,6 +14,7 @@ from sentry.uptime.subscriptions.subscriptions import (
     create_project_uptime_subscription,
     create_uptime_subscription,
 )
+from sentry.utils import json
 from sentry.utils.audit import create_audit_entry
 
 MAX_MONITORS_PER_DOMAIN = 100
@@ -28,6 +29,8 @@ for the domain `sentry.io` both the hosts `subdomain-one.sentry.io` and
 Importantly domains like `vercel.dev` are considered TLDs as defined by the
 public suffix list (PSL). See `extract_domain_parts` fo more details
 """
+SUPPORTED_HTTP_METHODS = {"GET", "POST", "HEAD", "PUT", "DELETE", "PATCH", "OPTIONS"}
+MAX_REQUEST_SIZE = 1000
 
 
 @extend_schema_serializer()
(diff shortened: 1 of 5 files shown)


src/sentry/uptime/endpoints/validators.py
```diff
@@ -28,6 +28,8 @@
 Importantly domains like `vercel.dev` are considered TLDs as defined by the
 public suffix list (PSL). See `extract_domain_parts` fo more details
 """
+SUPPORTED_HTTP_METHODS = {"GET", "POST", "HEAD", "PUT", "DELETE", "PATCH", "OPTIONS"}
+MAX_REQUEST_SIZE = 1000
```
evanpurkhiser, line 32: Can we add a unit to this?

src/sentry/uptime/endpoints/validators.py
```diff
@@ -47,6 +49,39 @@ class UptimeMonitorValidator(CamelSnakeSerializer):
         required=True, min_value=60, max_value=int(timedelta(days=1).total_seconds())
     )
     mode = serializers.IntegerField(required=False)
+    method = serializers.CharField(required=False, max_length=20)
```
evanpurkhiser, line 52: This should be a choice field no?

src/sentry/uptime/endpoints/validators.py
```diff
@@ -47,6 +49,39 @@ class UptimeMonitorValidator(CamelSnakeSerializer):
...
+            headers_dict = self.instance.uptime_subscription.headers
+            method = self.instance.uptime_subscription.method
+            body = self.instance.uptime_subscription.body or ""
+            url = self.instance.uptime_subscription.url
+        request_line_size = len(
+            f"{attrs.get('method', method)} {attrs.get('url', url)} HTTP/1.1\r\n"
+        )
+
+        headers_dict = attrs.get("headers", headers_dict)
+        headers_size = sum(len(f"{key}: {value}\r\n") for key, value in headers_dict.items())
```
evanpurkhiser, line 71: I think to get the correct length in bytes we need to encode this as UTF-8. Otherwise we're just getting the number of code points (unique UTF-8 characters) and not the actual byte length.

```suggestion
        headers_size = sum(len(f"{key}: {value}\r\n".encode('utf-8')) for key, value in headers_dict.items())
```

evanpurkhiser, line 71: IMO we should pull all of this logic out of this function and put it into a named function that we can write unit tests for, where we pass the header dict, body string and method.

src/sentry/uptime/endpoints/validators.py
```diff
@@ -47,6 +49,39 @@ class UptimeMonitorValidator(CamelSnakeSerializer):
...
+
+        body_size = len(attrs.get("body", body))
+
+        request_size = request_line_size + headers_size + len("\r\n") + body_size
+
+        if request_size > MAX_REQUEST_SIZE:
+            raise serializers.ValidationError("Request is too large")
```
evanpurkhiser, line 78: Should we specify the max size in the error message?

src/sentry/uptime/endpoints/validators.py
```diff
@@ -73,9 +108,13 @@ def validate_mode(self, mode):
             )
 
     def create(self, validated_data):
+        method_headers_body = {
+            k: v for k, v in validated_data.items() if k in {"method", "headers", "body"}
+        }
```
evanpurkhiser, line 113: Why do it like this vs just explicitly having

```py
method=validated_data["methods"],
headers=validated_data["headers"],
body=validated_data["body"],
```

wedamija: Since these are optional, this filters out things that weren't passed so that the defaults from the method we call are set.

Although maybe it'd be better to set a default on the fields and just always pass it?

src/sentry/uptime/endpoints/validators.py
wedamija: We should probably do the same for body too?

src/sentry/uptime/endpoints/validators.py
evanpurkhiser: Ah yeah makes sense. I am +1 to defaults

wedamija: I'll handle this as part of the update api. Adding defaults might have some unexpected side effects and I think it's better to get this merged

evanpurkhiser: 👍

evanpurkhiser (approved):

src/sentry/uptime/endpoints/validators.py
```diff
@@ -28,6 +28,17 @@
 Importantly domains like `vercel.dev` are considered TLDs as defined by the
 public suffix list (PSL). See `extract_domain_parts` fo more details
 """
+SUPPORTED_HTTP_METHODS = ["GET", "POST", "HEAD", "PUT", "DELETE", "PATCH", "OPTIONS"]
+MAX_REQUEST_SIZE_BYTES = 1000
```
evanpurkhiser, line 32: I wonder if we should have done 1024 to be sane. nbd

PR #77731, opened 2024-09-18
deps(ui): Upgrade biome 1.9

scttcper:
notes https://biomejs.dev/blog/biome-v1-9/

vscode biome already supports this version, maybe it will stop putting `import type type type`

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -36,7 +36,6 @@
         "noExcessiveNestedTestSuites": "error"
       },
       "nursery": {
-        "noDuplicateJsonKeys": "error",
         "noRestrictedImports": {
           "level": "warn",
           "options": {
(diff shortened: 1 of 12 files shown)


biome.json
```diff
@@ -36,7 +36,6 @@
         "noExcessiveNestedTestSuites": "error"
       },
       "nursery": {
-        "noDuplicateJsonKeys": "error",
```
scttcper, line 39: this rule has graduated nursery and is still linted via `noDuplicateObjectKeys`

michellewzhang (approved):

biome.json
```diff
@@ -100,6 +99,10 @@
       ".devenv"
     ]
   },
+  "css": {
+    "formatter": { "enabled": false },
+    "linter": { "enabled": false }
+  },
```
scttcper, line 105: disabling css linting for now, emotion not supported and linting/formatting our existing css would take a lot of changes

PR #77732, opened 2024-09-18
fix(custom-views): Re-save view if it's deleted but attempted to be updated

MichaelSun48:
resolves SENTRY-3DDC

This PR fixes a rare bug where the backend would attempt to update a groupsearchview with an id that doesn't exist anymore. This scenario would occur in rare edge cases where, for example, the user would delete a view in one browser window, then update the same view in a separate browser window. We decide to save the view back in this scenario rather than deleting it.


roggenkemper (approved):

malwilley (approved):
May want to write a quick test for this too

MichaelSun48:
@malwilley done!

malwilley (approved):

Issue #77570, opened 2024-09-16
Sentry Azure Dev ops integration will break on Sep. 30


sentaur-athena:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

Microsoft is disabling Sentry's integration for Azure Dev ops on Sep. 30 based on this email sent to customers:
```
You are receiving this email because you are listed as a top user of the following third-party Azure DevOps OAuth app. As part of the Secure First Initiative, we are locking down all first-party (1P) usage of Azure DevOps OAuth.

·       App Id: 81b764b5-6558-4435-807c-4191963da611
·       App Name: Sentry (v3)
·       App URL: https://sentry.io/
·       Your Usage (last 30 days): 385

Starting September 30, we will block all third-party apps from being able to access any first-party resources.

These third-party apps using Azure DevOps OAuth have not been signed off by our Microsoft Digital Software Licensing Service that approves and manages all Third Party Software use at Microsoft.

Stop using this app

If you are no longer using this app or do not remember authorizing this app, revoke its authorization immediately.

After September 30, all apps will be blocked when attempting to access Microsoft resources. No exceptions will be made.

Continue using this app

If you would like to continue using the named app, you are asked to:

1.       Some apps have been already been restricted and evicted from the Microsoft software suite. Confirm that your app is not in this list and meet all accepted software requirements before requesting approval.
2.       Complete a request for software approval.
3.       Reach out to the app developer and make sure they have set up an Entra app to authenticate to Azure DevOps APIs. Azure DevOps OAuth apps will no longer be accepted to access 1P resources and is on the path to deprecation for all users.

```

Looking at documentation this means we need to create a new application using Entra and migrate our customers to the new integration.

### Expected Result

N/A

### Actual Result

N/A

### Product Area

Settings - Integrations

sentaur-athena:
Posting some update on this issue. We're building a new integration for ADO and plan to migrate all ADO orgs to the new integration before Sep 30.

Right now IT is helping us get a verified microsoft partner account and Raj is planning the work needed for migrating away from the old integration to new without losing sync capabilities.

sentaur-athena:
We have a fix merging today behind a feature flag for testing. The idea is that the customers with ADO will go to settings -> integrations -> Azure  and install again. This will in the background replace the deprecated integration with a new one and migrate everything including issues and webhooks to work with the new one.

[...]

PR #77736, opened 2024-09-18
:sparkles: feat(azure_devops): Update Azure DevOps Integration

iamrajjoshi:
# Update:
i also updated the IdentityProvider logic so that it uses the new OAuth flow if the organization has the feature flag. the main difference between the existing provider and the new provider is i changed the `/authorize` and `/token` urls, the parameters we pass to each of these endpoints & updated the refresh token params.

---
Part of addressing: https://github.com/getsentry/sentry/issues/77570

In this pr, i add logic to migrate existing integrations to the new app. in the integration installation pipeline, we check for a `integration_migrated` key in metadata. If it doesn't exist, we create a new subscription and add it to the integration dict that `build_integration` returns. we will also set the `integration_migration` key to be True.

In `pipelin.py` - https://github.com/getsentry/sentry/blob/c1ebcc18b4c6a41311f45dee35482553feb0f922/src/sentry/integrations/pipeline.py#L88-L129

the data (the dict that we returned earlier) will then be passed to `ensure_integration`, which ends up saving the new user credentials as well as the migrated key:
https://github.com/getsentry/sentry/blob/c1ebcc18b4c6a41311f45dee35482553feb0f922/src/sentry/integrations/pipeline.py#L29-L41

### DIFF
--- a/src/sentry/auth/helper.py
+++ b/src/sentry/auth/helper.py
@@ -727,7 +727,7 @@ class AuthHelper(Pipeline):
         self.organization: RpcOrganization = self.organization
         self.provider: Provider = self.provider
 
-    def get_provider(self, provider_key: str | None) -> PipelineProvider:
+    def get_provider(self, provider_key: str | None, **kwargs) -> PipelineProvider:
         if self.provider_model:
             return cast(PipelineProvider, self.provider_model.get_provider())
         elif provider_key:
--- a/src/sentry/features/temporary.py
+++ b/src/sentry/features/temporary.py
@@ -191,7 +191,7 @@ def register_temporary_features(manager: FeatureManager):
     # Enables the search bar for metrics samples list
     manager.add("organizations:metrics-samples-list-search", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=True)
     # Migrate Orgs to new Azure DevOps Integration
-    manager.add("organizations:migrate-azure-devops-integration", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=False, default=True)
+    manager.add("organizations:migrate-azure-devops-integration", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=False)
     # Enable Session Stats down to a minute resolution
     manager.add("organizations:minute-resolution-sessions", OrganizationFeature, FeatureHandlerStrategy.INTERNAL, default=True, api_expose=True)
     # Display CPU and memory metrics in transactions with profiles
--- a/src/sentry/features/temporary.py
+++ b/src/sentry/features/temporary.py
@@ -190,6 +190,8 @@ def register_temporary_features(manager: FeatureManager):
     manager.add("organizations:metric-alert-threshold-period", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=True)
     # Enables the search bar for metrics samples list
     manager.add("organizations:metrics-samples-list-search", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=True)
+    # Migrate Orgs to new Azure DevOps Integration
+    manager.add("organizations:migrate-azure-devops-integration", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=False)
     # Enable Session Stats down to a minute resolution
     manager.add("organizations:minute-resolution-sessions", OrganizationFeature, FeatureHandlerStrategy.INTERNAL, default=True, api_expose=True)
     # Display CPU and memory metrics in transactions with profiles
--- a/src/sentry/features/temporary.py
+++ b/src/sentry/features/temporary.py
@@ -191,7 +191,7 @@ def register_temporary_features(manager: FeatureManager):
     # Enables the search bar for metrics samples list
     manager.add("organizations:metrics-samples-list-search", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=True)
     # Migrate Orgs to new Azure DevOps Integration
-    manager.add("organizations:migrate-azure-devops-integration", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=False)
+    manager.add("organizations:migrate-azure-devops-integration", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=True)
     # Enable Session Stats down to a minute resolution
     manager.add("organizations:minute-resolution-sessions", OrganizationFeature, FeatureHandlerStrategy.INTERNAL, default=True, api_expose=True)
     # Display CPU and memory metrics in transactions with profiles
--- a/src/sentry/features/temporary.py
+++ b/src/sentry/features/temporary.py
@@ -191,7 +191,7 @@ def register_temporary_features(manager: FeatureManager):
     # Enables the search bar for metrics samples list
     manager.add("organizations:metrics-samples-list-search", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=True)
     # Migrate Orgs to new Azure DevOps Integration
-    manager.add("organizations:migrate-azure-devops-integration", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=False)
+    manager.add("organizations:migrate-azure-devops-integration", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=False, default=True)
     # Enable Session Stats down to a minute resolution
     manager.add("organizations:minute-resolution-sessions", OrganizationFeature, FeatureHandlerStrategy.INTERNAL, default=True, api_expose=True)
     # Display CPU and memory metrics in transactions with profiles
--- a/src/sentry/identity/pipeline.py
+++ b/src/sentry/identity/pipeline.py
@@ -38,6 +38,7 @@ class IdentityProviderPipeline(Pipeline):
         # Use configured redirect_url if specified for the pipeline if available
         return self.config.get("redirect_url", associate_url)
 
+    # TODO(iamrajjoshi): Delete this after Azure DevOps migration is complete
     def get_provider(self, provider_key: str, **kwargs) -> PipelineProvider:
         if kwargs.get("organization"):
             organization: Organization | RpcOrganization = kwargs["organization"]
--- a/src/sentry/identity/pipeline.py
+++ b/src/sentry/identity/pipeline.py
@@ -41,9 +41,8 @@ class IdentityProviderPipeline(Pipeline):
     def get_provider(self, provider_key: str, **kwargs) -> PipelineProvider:
         if kwargs.get("organization"):
             organization: Organization | RpcOrganization = kwargs["organization"]
-        if (
-            features.has("organizations:migrate-azure-devops-integration", organization.id)
-            and provider_key == "vsts"
+        if provider_key == "vsts" and features.has(
+            "organizations:migrate-azure-devops-integration", organization
         ):
             provider_key = "vsts_new"
 
--- a/src/sentry/identity/pipeline.py
+++ b/src/sentry/identity/pipeline.py
@@ -5,7 +5,10 @@ from django.http import HttpResponseRedirect
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
-from sentry.pipeline import Pipeline
+from sentry import features
+from sentry.models.organization import Organization
+from sentry.organizations.services.organization.model import RpcOrganization
+from sentry.pipeline import Pipeline, PipelineProvider
 from sentry.users.models.identity import Identity, IdentityProvider
 from sentry.utils import metrics
 
@@ -34,6 +37,17 @@ class IdentityProviderPipeline(Pipeline):
 
         # Use configured redirect_url if specified for the pipeline if available
         return self.config.get("redirect_url", associate_url)
+
+    def get_provider(self, provider_key: str, **kwargs) -> PipelineProvider:
+        if kwargs.get("organization"):
+            organization: Organization | RpcOrganization = kwargs["organization"]
+        if (
+            features.has("organizations:migrate-azure-devops-integration", organization.id)
+            and provider_key == "vsts"
+        ):
+            provider_key = "vsts_new"
+
+        return super().get_provider(provider_key)
 
     def finish_pipeline(self):
         # NOTE: only reached in the case of linking a new identity
--- a/src/sentry/identity/vsts/provider.py
+++ b/src/sentry/identity/vsts/provider.py
@@ -137,6 +137,10 @@ class VSTSOAuth2CallbackView(OAuth2CallbackView):
         return orjson.loads(body)
 
 
+# TODO(iamrajjoshi): Make this the default provider
+# We created this new flow in order to quickly update the DevOps integration to use
+# the new Azure AD OAuth2 flow.
+# This is a temporary solution until we can fully migrate to the new flow once customers are migrated
 class VSTSNewIdentityProvider(OAuth2Provider):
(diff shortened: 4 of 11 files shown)


src/sentry/integrations/vsts/integration.py
```diff
@@ -463,8 +468,36 @@ def build_integration(self, state: Mapping[str, Any]) -> Mapping[str, Any]:
             integration_model = IntegrationModel.objects.get(
                 provider="vsts", external_id=account["accountId"], status=ObjectStatus.ACTIVE
             )
-            # preserve previously created subscription information
-            integration["metadata"]["subscription"] = integration_model.metadata["subscription"]
+
+            # Get Integration Metadata
+            integration_migrated = integration_model.metadata.get("integration_migrated", False)
```
GabeVillalobos, line 473: Nit: It might be good to change this to a version number + check in case we need to conditionally migrate again.

src/sentry/integrations/vsts/integration.py
```diff
@@ -463,8 +468,36 @@ def build_integration(self, state: Mapping[str, Any]) -> Mapping[str, Any]:
...
+            if (
+                features.has(
+                    "organizations:migrate-azure-devops-integration", self.pipeline.organization.id
+                )
+                and not integration_migrated
+            ):
+                subscription_id, subscription_secret = self.create_subscription(
```
GabeVillalobos, line 480: Just because I'm unfamiliar with this integration: Is this for creating a new webhook subscription, or something else?

src/sentry/integrations/vsts/integration.py
```diff
@@ -463,8 +468,36 @@ def build_integration(self, state: Mapping[str, Any]) -> Mapping[str, Any]:
...
+                }
+
+                integration["metadata"]["integration_migrated"] = True
+
+                logger.info(
+                    "vsts.build_integration.migrated",
+                    extra={
+                        "organization_id": self.pipeline.organization.id,
+                        "user_id": user["id"],
+                        "account": account,
```
GabeVillalobos, line 495: Do we want to log the subscription Id here as well?

GabeVillalobos (approved):
Nice and simple, looks good to me!

src/sentry/integrations/vsts/integration.py
```diff
@@ -422,6 +422,11 @@ def post_install(
             )
 
     def get_scopes(self) -> Sequence[str]:
+        if features.has(
+            "organizations:migrate-azure-devops-integration", self.pipeline.organization.id
+        ):
+            # This is the new way we need to pass scopes to the OAuth flow
+            return ("offline_access", "499b84ac-1321-427f-aa17-267ca6975798/.default")
```
cathteng, line 429: is the second part of the tuple okay to put in a public repo? link to docs?

src/sentry/integrations/vsts/integration.py
```diff
@@ -463,8 +468,36 @@ def build_integration(self, state: Mapping[str, Any]) -> Mapping[str, Any]:
...
                 provider="vsts", external_id=account["accountId"], status=ObjectStatus.ACTIVE
             )
-            # preserve previously created subscription information
-            integration["metadata"]["subscription"] = integration_model.metadata["subscription"]
+
+            # Get Integration Metadata
+            integration_migrated = integration_model.metadata.get("integration_migrated", False)
+            if (
+                features.has(
+                    "organizations:migrate-azure-devops-integration", self.pipeline.organization.id
```
cathteng, line 476: FYI `self.pipeline.organization` can be `None` but i think this is fine

cathteng (approved):
this looks good, i'm curious how the other parts will shape up

src/sentry/integrations/vsts/integration.py
iamrajjoshi: yeah it is, ill link docs here. tbh i want to put a link to the notion doc cause 1 link won't suffice this

src/sentry/integrations/vsts/integration.py
iamrajjoshi: what this does is setup the client and subscribe to the webhook to sync from devops -> sentry
https://github.com/getsentry/sentry/blob/1548e7a18767d982633ee648695abd3c2413208a/src/sentry/integrations/vsts/client.py#L119-L134

src/sentry/identity/vsts/provider.py
```diff
@@ -135,3 +135,108 @@ def exchange_token(self, request: Request, pipeline, code):
...
             return dict(parse_qsl(body))
         return orjson.loads(body)
+
+
+class VSTSNewIdentityProvider(OAuth2Provider):
+    key = "vsts_new"
+    name = "Azure DevOps"
+
+    oauth_access_token_url = "https://login.microsoftonline.com/common/oauth2/v2.0/token"
+    oauth_authorize_url = "https://login.microsoftonline.com/common/oauth2/v2.0/authorize"
```
iamrajjoshi, line 145: these urls are new

src/sentry/identity/vsts/provider.py
```diff
@@ -135,3 +135,108 @@ def exchange_token(self, request: Request, pipeline, code):
...
+
+    def get_oauth_client_secret(self):
+        return options.get("vsts_new.client-secret")
+
+    def get_refresh_token_url(self):
+        return self.oauth_access_token_url
+
+    def get_pipeline_views(self):
+        return [
+            VSTSOAuth2LoginView(
```
iamrajjoshi, line 158: created a new view below

src/sentry/identity/vsts/provider.py
```diff
@@ -135,3 +135,108 @@ def exchange_token(self, request: Request, pipeline, code):
...
+    def exchange_token(self, request: Request, pipeline, code):
+        from urllib.parse import parse_qsl
+
+        from sentry.http import safe_urlopen, safe_urlread
+        from sentry.utils.http import absolute_uri
+
+        req = safe_urlopen(
+            url=self.access_token_url,
+            headers={"Content-Type": "application/x-www-form-urlencoded", "Content-Length": "1322"},
+            data={
```
iamrajjoshi, line 231: main difference between existing `VSTSOAuth2CallbackView`

src/sentry/pipeline/base.py
```diff
@@ -118,7 +118,7 @@ def __init__(
         )
         self.state = self.session_store_cls(request, self.pipeline_name, ttl=PIPELINE_STATE_TTL)
         self.provider_model = provider_model
-        self.provider = self.get_provider(provider_key)
+        self.provider = self.get_provider(provider_key, organization=organization)
```
iamrajjoshi, line 121: passing this here so in the identity pipeline, we can use the org to do the feature flag check. i made the super class `get_provider` take a **kwargs, which i know is smelly but ideally we will remove the flag soon and clean this up.

src/sentry/auth/helper.py
```diff
@@ -727,7 +727,7 @@ def __init__(
         self.organization: RpcOrganization = self.organization
         self.provider: Provider = self.provider
 
-    def get_provider(self, provider_key: str | None) -> PipelineProvider:
+    def get_provider(self, provider_key: str | None, **kwargs) -> PipelineProvider:
```
iamrajjoshi, line 730: fix tests since we are passing org to `get_provider`. will remove after we fix the feature flag

src/sentry/features/temporary.py
```diff
@@ -188,6 +188,8 @@ def register_temporary_features(manager: FeatureManager):
     manager.add("organizations:metric-alert-threshold-period", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=True)
     # Enables the search bar for metrics samples list
     manager.add("organizations:metrics-samples-list-search", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=True)
+    # Migrate Orgs to new Azure DevOps Integration
+    manager.add("organizations:migrate-azure-devops-integration", OrganizationFeature, FeatureHandlerStrategy.FLAGPOLE, api_expose=True)
```
sentaur-athena, line 192: why expose?

src/sentry/identity/vsts/provider.py
```diff
@@ -135,3 +135,108 @@ def exchange_token(self, request: Request, pipeline, code):
         if req.headers["Content-Type"].startswith("application/x-www-form-urlencoded"):
             return dict(parse_qsl(body))
         return orjson.loads(body)
+
+
+class VSTSNewIdentityProvider(OAuth2Provider):
```
sentaur-athena, line 140: Please add some comments on what this is. I can imagine it'll be very confusing to future engineers if we end up not cleaning up fast

src/sentry/identity/vsts/provider.py
sentaur-athena: Why didn't the old one need a specific view and this one does?

src/sentry/features/temporary.py
iamrajjoshi: incase we want to have some ui in the frontend to show if we migrated the integration yet.

iamrajjoshi:
don't know why i can't respond to your original comment @sentaur-athena 
> Why didn't the old one need a specific view and this one does?

basically i made a simple new view which overrides the `get_param` function so i can complete the new oauth flow.

src/sentry/identity/pipeline.py
```diff
@@ -35,6 +38,16 @@ def redirect_url(self):
         # Use configured redirect_url if specified for the pipeline if available
         return self.config.get("redirect_url", associate_url)
 
+    def get_provider(self, provider_key: str, **kwargs) -> PipelineProvider:
+        if kwargs.get("organization"):
+            organization: Organization | RpcOrganization = kwargs["organization"]
+        if provider_key == "vsts" and features.has(
+            "organizations:migrate-azure-devops-integration", organization
+        ):
+            provider_key = "vsts_new"
```
RyanSkonnord, line 47: Is this another place where we need to "todo: remove" comment?

src/sentry/identity/vsts/provider.py
cathteng: the `VSTSOAuth2LoginView` overwrites some parts of the normal `OAuthLogin2View`, which it subclasses. similarly for the callback view

src/sentry/identity/vsts/provider.py
```diff
@@ -135,3 +135,114 @@ def exchange_token(self, request: Request, pipeline, code):
...
+                client_id=self.get_oauth_client_id(),
+                client_secret=self.get_oauth_client_secret(),
+            ),
+        ]
+
+    def get_refresh_token_headers(self):
+        return {"Content-Type": "application/x-www-form-urlencoded", "Content-Length": "1654"}
+
+    def get_refresh_token_params(self, refresh_token, *args, **kwargs):
+        # Note: ignoring the below from the original provider
```
sentaur-athena, line 180: Please add a todo here to fix for vsts-limited.

src/sentry/integrations/vsts/integration.py
sentaur-athena: I just read the stackoverflow post. This is wild and I want to know who approved that PR on microsoft side! Anyways, @cathteng still not sure if I understand this fully but here's the docs on Microsoft side. It's ok to be public:
https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#request-an-authorization-code
https://learn.microsoft.com/en-us/rest/api/azure/devops/tokens/?view=azure-devops-rest-7.1&tabs=powershell#personal-access-tokens-pats

src/sentry/integrations/vsts/client.py
```diff
@@ -199,9 +201,19 @@ def _refresh_auth_if_expired(self):
...
-            )
+            # TODO(iamrajjoshi): Remove this after migration
+            # Need this here because there is no way to get any identifier which would tell us which method we should use to refresh the token
+            from sentry.identity.vsts.provider import VSTSNewIdentityProvider
+
+            try:
+                self.identity.get_provider().refresh_identity(
+                    self.identity, redirect_url=self.oauth_redirect_url
+                )
+            except ApiError as e:
```
cathteng, line 212: when/how can we be confident we can always use the new identity provider?

src/sentry/integrations/vsts/client.py
```diff
@@ -199,9 +201,19 @@ def _refresh_auth_if_expired(self):
...
         if int(time_expires) <= int(time()):
-            self.identity.get_provider().refresh_identity(
-                self.identity, redirect_url=self.oauth_redirect_url
-            )
+            # TODO(iamrajjoshi): Remove this after migration
+            # Need this here because there is no way to get any identifier which would tell us which method we should use to refresh the token
+            from sentry.identity.vsts.provider import VSTSNewIdentityProvider
+
+            try:
+                self.identity.get_provider().refresh_identity(
```
cathteng, line 209: does the integration need to migrated for us to be able to use the new refresh method? when does this fail?

src/sentry/integrations/vsts/integration.py
sentaur-athena: old integration id too? Just to help debugging if someone has multiple Azure

tests/sentry/integrations/vsts/test_integration.py
```diff
@@ -8,19 +8,106 @@
...
 from sentry.shared_integrations.exceptions import IntegrationError, IntegrationProviderError
 from sentry.silo.base import SiloMode
+from sentry.testutils.helpers import with_feature
 from sentry.testutils.silo import assume_test_silo_mode, control_silo_test
 from sentry.users.models.identity import Identity
 
 FULL_SCOPES = ["vso.code", "vso.graph", "vso.serviceendpoint_manage", "vso.work_write"]
 LIMITED_SCOPES = ["vso.graph", "vso.serviceendpoint_manage", "vso.work_write"]
 
+NEW_SCOPES = ["offline_access", "499b84ac-1321-427f-aa17-267ca6975798/.default"]
```
sentaur-athena, line 27: Can you actually define this as a const in the vsts integration and reuse in both places you need?

sentaur-athena (approved):

src/sentry/integrations/vsts/client.py
iamrajjoshi: yep new integration needs the new method. this method will fail when we try to use the old refresh method on the new integration, then it will try using the new method.

src/sentry/integrations/vsts/client.py
iamrajjoshi: when all users are migrated to the new integration, we can updating the existing provider and remove this try catch

src/sentry/integrations/vsts/client.py
cathteng: could you find the `Integration` from the `org_integration_id` and check the metadata?

iamrajjoshi: hmm, thats a good point, let me try

cathteng (approved):

src/sentry/integrations/vsts/client.py
```diff
@@ -199,9 +201,24 @@ def _refresh_auth_if_expired(self):
         if time_expires is None:
             raise InvalidIdentity("VstsApiClient requires identity with specified expired time")
         if int(time_expires) <= int(time()):
-            self.identity.get_provider().refresh_identity(
-                self.identity, redirect_url=self.oauth_redirect_url
+            # TODO(iamrajjoshi): Remove this after migration
+            # Need this here because there is no way to get any identifier which would tell us which method we should use to refresh the token
+            from sentry.identity.vsts.provider import VSTSNewIdentityProvider
+            from sentry.integrations.vsts.integration import VstsIntegrationProvider
```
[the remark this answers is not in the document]
iamrajjoshi: unfortunately yes

src/sentry/integrations/vsts/client.py
cathteng: ok i guess we just track a redash query

PR #77740, opened 2024-09-18
ref(issue-details): Refactor hooks for simpler discover calls

leeandher:
Found some better hooks to be using for API calls to discover. We can standardize the queries with `EventViews` and provide routes for `event-stats`. We will  use the `events` route for the All Events tables.


static/app/views/issueDetails/streamline/useIssueDetailsDiscoverQuery.tsx
```diff
@@ -0,0 +1,84 @@
...
+    route,
+    eventView,
+    location,
+    orgSlug: organization.slug,
+    getRequestPayload: () => ({
+      ...eventView.getEventsAPIPayload(location),
+      query: eventView.query,
+      interval: eventView.interval,
+      yAxis: eventView.yAxis,
+      partial: 1,
```
leeandher, line 79: 1 small change here to add `partial: 1` so that we include the extra data point to avoid inconsistencies

PR #77741, opened 2024-09-18
fix(substatus): Add migration to fix substatuses

snigdhas:
Fixes substatus inconsistencies by removing substatus for groups that aren't unresolved or ignored and recalculating substatus for unresolved/ignored groups that have an incompatible substatus. 

From [this query](https://redash.getsentry.net/queries/6949), there should be around 3k groups to fix.


src/sentry/migrations/0764_migrate_bad_status_substatus_rows.py
```diff
@@ -0,0 +1,170 @@
...
+    "ignoreCount",
+    "ignoreDuration",
+    "ignoreUntil",
+    "ignoreUserCount",
+    "ignoreUserWindow",
+    "ignoreWindow",
+}
+
+
+def fix_substatus_for__groups(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
```
markstory, line 81:
```suggestion
def fix_substatus_for_groups(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
```

Was the double `_` intentional?

markstory (approved):
Migration/ORM usage looks good to me, but I'm not well versed in the requirements around issue substatus.

src/sentry/migrations/0764_migrate_bad_status_substatus_rows.py
snigdhas: nope, good catch!

PR #77749, opened 2024-09-18
feat: backend changes to support EAP-based LLM monitoring

colin-sentry:
We want to start migrating insights modules to EAP, this is the backend changes for LLM monitoring to move over.


wmak (approved):
lgtm

PR #77750, opened 2024-09-18
feat: add ability to power LLM monitoring with EAP

colin-sentry:
Requires https://github.com/getsentry/sentry/pull/77749 first, this implements the LLM monitoring frontend using EAP instead of metrics.


DominikB2014 (commented):
So it looks like the query syntax and fields stay the same, just the dataset param that we send to discover changes? If that’s the case, couldn’t we have put the useEAP conditional in useSpanIndexed ? That would effectively update it everywhere in the app

DominikB2014 (approved):
code lgtm, there is some inconsistency where some components will either call `useEAPSpans` or `useSpan...` and sometimes we rewrite an `EAP` version of a component. But as this is temporary, we can probably just do what makes sense/is easier

PR #77751, opened 2024-09-18
fix(tests): Fix act warnings in SearchQueryBuilder tests

malwilley:
Some tests did not await any act() functions which caused act warnings. These warnings were somehow covered up by a missing await early in the test suite (somehow).


scttcper (approved):
async

PR #77754, opened 2024-09-18
fix(ecosystem): Breaks issue sync cycles

GabeVillalobos:
## The problem:
When a Jira issue's `assignee` field is quickly modified multiple times in a row, we end up assigning and resyncing the assignees to Jira, which causes a loop. This is because _all_ inbound integration webhooks that modify an issue assignment generate an outbound update for all issue sync integrations as well.

## Solution:
This PR adds a new `AssignmentSource` dataclass, which acts as a context carrier when an external webhook triggers an issue assignment change. This is then checked on the outbound sync logic, preventing the change from being propagated to the same integration that initially triggered the change.

### DIFF
--- a/src/sentry/integrations/mixins/issues.py
+++ b/src/sentry/integrations/mixins/issues.py
@@ -12,6 +12,7 @@ from typing import Any, ClassVar
 from sentry.eventstore.models import GroupEvent
 from sentry.integrations.base import IntegrationInstallation
 from sentry.integrations.models.external_issue import ExternalIssue
+from sentry.integrations.services.assignment_source import AssignmentSource
 from sentry.integrations.services.integration import integration_service
 from sentry.integrations.tasks.sync_status_inbound import (
     sync_status_inbound as sync_status_inbound_task,
@@ -62,7 +63,7 @@ class ResolveSyncAction(enum.Enum):
 
 
 class IssueBasicIntegration(IntegrationInstallation, ABC):
-    def should_sync(self, attribute):
+    def should_sync(self, attribute, sync_source: AssignmentSource | None = None):
         return False
 
     def get_group_title(self, group, event, **kwargs):
@@ -378,10 +379,17 @@ class IssueSyncIntegration(IssueBasicIntegration, ABC):
     outbound_assignee_key: ClassVar[str | None] = None
     inbound_assignee_key: ClassVar[str | None] = None
 
-    def should_sync(self, attribute: str) -> bool:
+    def should_sync(self, attribute: str, sync_source: AssignmentSource | None = None) -> bool:
         key = getattr(self, f"{attribute}_key", None)
         if key is None or self.org_integration is None:
             return False
+
+        # Check that the assignment source isn't this same integration in order to
+        # prevent sync-cycles from occurring. This should still allow other
+        # integrations to propagate changes outward.
+        if sync_source and sync_source.integration_id == self.org_integration.integration_id:
+            return False
+
         value: bool = self.org_integration.config.get(key, False)
         return value
 
@@ -400,7 +408,14 @@ class IssueSyncIntegration(IssueBasicIntegration, ABC):
         raise NotImplementedError
 
     @abstractmethod
-    def sync_status_outbound(self, external_issue, is_resolved, project_id, **kwargs):
+    def sync_status_outbound(
+        self,
+        external_issue,
+        is_resolved,
+        project_id,
+        assignment_source: AssignmentSource | None = None,
+        **kwargs,
+    ):
         """
         Propagate a sentry issue's status to a linked issue's status.
         """
--- a/src/sentry/integrations/services/assignment_source.py
+++ b/src/sentry/integrations/services/assignment_source.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
+from dataclasses import asdict, dataclass
 from datetime import datetime
 from enum import Enum
 from typing import TYPE_CHECKING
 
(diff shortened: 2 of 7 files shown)


src/sentry/integrations/services/assignment_source.py
```diff
@@ -0,0 +1,32 @@
...
+from dataclasses import dataclass
+from datetime import datetime
+from enum import Enum
+from typing import TYPE_CHECKING
+
+from django.utils import timezone
+
+if TYPE_CHECKING:
+    from sentry.integrations.models import Integration
+    from sentry.integrations.services.integration import RpcIntegration
```
GabeVillalobos, line 12: I couldn't find a place to put this module that wouldn't cause an import loop, though I don't know if it really belongs here so suggestions welcome.

src/sentry/integrations/tasks/sync_assignee_outbound.py
```diff
@@ -24,7 +25,12 @@
         Organization.DoesNotExist,
     )
 )
-def sync_assignee_outbound(external_issue_id: int, user_id: int | None, assign: bool) -> None:
+def sync_assignee_outbound(
+    external_issue_id: int,
+    user_id: int | None,
+    assign: bool,
+    assignment_source: AssignmentSource | None = None,
```
markstory, line 32: I would be careful with dataclasses in celery tasks. Celery is configured to pickle parameters, and we could run into problems in the future with python upgrades. You could pass the source integration as a scalar and use that, or json serialize the dataclass and deserialize it in the task 🤷

src/sentry/integrations/services/assignment_source.py
markstory: Using TYPE_CHECKING for imports that are only used in type hints is a reasonable solution.

src/sentry/integrations/services/assignment_source.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class AssignmentSourceType(str, Enum):
+    integration = "integration"
+
+
+@dataclass(frozen=True)
+class AssignmentSource:
+    source_type: AssignmentSourceType
```
markstory, line 21: If this only has a single value do you need it?

GabeVillalobos: My thought was to expand it once I have more use cases, but omitting it for now seems fine.

src/sentry/integrations/tasks/sync_assignee_outbound.py
GabeVillalobos: That's fair. Funnily enough, won't JSON serialization have many of the same pitfalls as pickling?

GabeVillalobos: I'd like this to eventually become a context carrier vs just an ever-expanding list of vars personally, but I can see the argument for keeping it simple for now. Let me explore if this solves other problems (like our lack of integration info in the activities pane for JIRA) before removing it.

src/sentry/integrations/utils/sync.py
```diff
@@ -120,5 +134,10 @@ def sync_group_assignee_outbound(group: Group, user_id: int | None, assign: bool
 
     for external_issue_id in external_issue_ids:
         sync_assignee_outbound.apply_async(
-            kwargs={"external_issue_id": external_issue_id, "user_id": user_id, "assign": assign}
+            kwargs={
+                "external_issue_id": external_issue_id,
+                "user_id": user_id,
+                "assign": assign,
+                "assignment_source_json": assignment_source.json() if assignment_source else None,
```
cathteng, line 141: out of curiousity, what's the difference between dumping this as a json and then reconstructing the class from the json vs dumping it as a dictionary? is it because one of the values (`AssignmentSourceType`) is another class?

src/sentry/integrations/mixins/issues.py
```diff
@@ -378,10 +379,17 @@ class IssueSyncIntegration(IssueBasicIntegration, ABC):
...
-    def should_sync(self, attribute: str) -> bool:
+    def should_sync(self, attribute: str, sync_source: AssignmentSource | None = None) -> bool:
         key = getattr(self, f"{attribute}_key", None)
         if key is None or self.org_integration is None:
             return False
+
+        # Check that the assignment source isn't this same integration in order to
+        # prevent sync-cycles from occurring. This should still allow other
+        # integrations to propagate changes outward.
+        if sync_source and sync_source.integration_id == self.org_integration.integration_id:
```
cathteng, line 390: after this merges, do we always expect there to be a `sync_source`?

cathteng (approved):

src/sentry/integrations/utils/sync.py
markstory: Dumping as a dictionary would work as well. We should avoid passing the dataclass instance as that relies on celery using pickle for task parameters which we'd like to phase out.

markstory (approved):

src/sentry/integrations/mixins/issues.py
GabeVillalobos: At the moment, no. It's something we'd have to incrementally add since there's some post-process auto-assignment that potentially wouldn't have this 🤔

src/sentry/integrations/utils/sync.py
GabeVillalobos: Removed the nested enum type and converted this to use dictionaries instead since that's a bit cleaner. 

>  is it because one of the values (AssignmentSourceType) is another class?
Yep! Having the nested class was preventing me from doing this before.

PR #77755, opened 2024-09-18
ref(alerts): Update Snuba queries to match events-stats more closely

ceorourke:
When a user creates an anomaly detection alert we need to query snuba for 28 days worth of historical data to send to Seer to calculate the anomalies.  Originally (https://github.com/getsentry/sentry/pull/74614) I'd tried to pull out the relevant parts of the `events-stats` endpoint to mimic the data we see populated in metric alert preview charts (but for a larger time period, and it's happening after the rule is saved so I can't use any of the `request` object stuff) but I think I missed some things, so this PR aims to make that data be the same. 

Closes https://getsentry.atlassian.net/browse/ALRT-288 (hopefully)

### DIFF
--- a/src/sentry/api/bases/organization_events.py
+++ b/src/sentry/api/bases/organization_events.py
@@ -43,11 +43,12 @@ from sentry.utils.snuba import MAX_FIELDS, SnubaTSResult
 
 
 def get_query_columns(columns, rollup):
-    # Backwards compatibility for incidents which uses the old
-    # column aliases as it straddles both versions of events/discover.
-    # We will need these aliases until discover2 flags are enabled for all
-    # users.
-    # We need these rollup columns to generate correct events-stats results
+    """
+    Backwards compatibility for incidents which uses the old
+    column aliases as it straddles both versions of events/discover.
+    We will need these aliases until discover2 flags are enabled for all users.
+    We need these rollup columns to generate correct events-stats results
+    """
     column_map = {
         "user_count": "count_unique(user)",
         "event_count": "count()",
--- a/src/sentry/api/bases/organization_events.py
+++ b/src/sentry/api/bases/organization_events.py
@@ -42,6 +42,26 @@ from sentry.utils.http import absolute_uri
 from sentry.utils.snuba import MAX_FIELDS, SnubaTSResult
 
 
+def get_query_columns(columns, rollup):
+    # Backwards compatibility for incidents which uses the old
+    # column aliases as it straddles both versions of events/discover.
+    # We will need these aliases until discover2 flags are enabled for all
+    # users.
+    # We need these rollup columns to generate correct events-stats results
+    column_map = {
+        "user_count": "count_unique(user)",
+        "event_count": "count()",
+        "epm()": "epm(%d)" % rollup,
+        "eps()": "eps(%d)" % rollup,
+        "tpm()": "tpm(%d)" % rollup,
+        "tps()": "tps(%d)" % rollup,
+        "sps()": "sps(%d)" % rollup,
+        "spm()": "spm(%d)" % rollup,
+    }
+
+    return [column_map.get(column, column) for column in columns]
+
+
 def resolve_axis_column(column: str, index: int = 0) -> str:
     return get_function_alias(column) if not is_equation(column) else f"equation[{index}]"
 
@@ -438,30 +458,13 @@ class OrganizationEventsV2EndpointBase(OrganizationEventsEndpointBase):
                     date_range = snuba_params.date_range
                     stats_period = parse_stats_period(get_interval_from_range(date_range, False))
                     rollup = int(stats_period.total_seconds()) if stats_period is not None else 3600
-
                 if comparison_delta is not None:
                     retention = quotas.get_event_retention(organization=organization)
                     comparison_start = snuba_params.start_date - comparison_delta
                     if retention and comparison_start < timezone.now() - timedelta(days=retention):
                         raise ValidationError("Comparison period is outside your retention window")
 
-                # Backwards compatibility for incidents which uses the old
-                # column aliases as it straddles both versions of events/discover.
-                # We will need these aliases until discover2 flags are enabled for all
-                # users.
-                # We need these rollup columns to generate correct events-stats results
-                column_map = {
-                    "user_count": "count_unique(user)",
-                    "event_count": "count()",
-                    "epm()": "epm(%d)" % rollup,
-                    "eps()": "eps(%d)" % rollup,
-                    "tpm()": "tpm(%d)" % rollup,
-                    "tps()": "tps(%d)" % rollup,
-                    "sps()": "sps(%d)" % rollup,
-                    "spm()": "spm(%d)" % rollup,
-                }
-
-                query_columns = [column_map.get(column, column) for column in columns]
+                query_columns = get_query_columns(columns, rollup)
             with sentry_sdk.start_span(op="discover.endpoint", description="base.stats_query"):
                 result = get_event_stats(
                     query_columns, query, snuba_params, rollup, zerofill_results, comparison_delta
--- a/src/sentry/seer/anomaly_detection/get_historical_anomalies.py
+++ b/src/sentry/seer/anomaly_detection/get_historical_anomalies.py
@@ -82,7 +82,10 @@ def get_historical_anomaly_data_from_seer(
         )
         return None
     formatted_data = format_historical_data(
-        historical_data, query_columns, dataset, project.organization
+        data=historical_data,
+        query_columns=query_columns,
+        dataset=dataset,
+        organization=project.organization,
     )
     if (
         not alert_rule.sensitivity
--- a/src/sentry/seer/anomaly_detection/get_historical_anomalies.py
+++ b/src/sentry/seer/anomaly_detection/get_historical_anomalies.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from django.conf import settings
 from urllib3.exceptions import MaxRetryError, TimeoutError
 
+from sentry.api.bases.organization_events import get_query_columns
 from sentry.conf.server import SEER_ANOMALY_DETECTION_ENDPOINT_URL
 from sentry.incidents.models.alert_rule import AlertRule, AlertRuleStatus
 from sentry.models.project import Project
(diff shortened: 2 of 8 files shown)


src/sentry/api/bases/organization_events.py
```diff
@@ -42,6 +42,27 @@
 from sentry.utils.snuba import MAX_FIELDS, SnubaTSResult
 
 
+def get_query_columns(columns, rollup):
```
ceorourke, line 45: I moved this to be reused by anomaly detection

src/sentry/seer/anomaly_detection/utils.py
```diff
@@ -75,55 +78,48 @@ def get_crash_free_historical_data(
...
-        If there are no results, it's just the timestamp
-        {'time': 1719012000}, {'time': 1719018000}, {'time': 1719024000}
-
-        If there are results, the aggregate is added
-        {'time': 1721300400, 'count': 2}
-
-    For metrics_performance dataset/sessions data:
-        The count is stored separately from the timestamps, if there is no data the count is 0
     """
+    serializer = SnubaTSResultSerializer(organization=organization, lookup=None, user=None)
```
ceorourke, line 87: I'm using the same serializer the `events-stats` endpoint uses and just pulling that data off to format into a list of `TimeSeriesPoint`s for Seer's API

src/sentry/seer/anomaly_detection/utils.py
```diff
@@ -75,55 +78,48 @@ def get_crash_free_historical_data(
...
+    serialized_result = serializer.serialize(
+        data,
+        resolve_axis_column(query_columns[0]),
+        allow_partial_buckets=False,
+        zerofill_results=False,
```
ceorourke, line 92: I was getting strange results with this set to True, and for our purposes I think it doesn't matter that much since we default to sending Seer a 0 if we don't find a count anyway

ceorourke: By "strange" I mean it was hitting [this line](https://github.com/getsentry/sentry/blob/master/src/sentry/api/serializers/snuba.py#L35-L35) and overwriting data with a count I had in a test as an empty array.

src/sentry/seer/anomaly_detection/utils.py
```diff
@@ -147,33 +177,47 @@ def fetch_historical_data(
...
+
+    snuba_params = SnubaParams(
+        organization=alert_rule.organization,
+        projects=[project],
+        start=start,
+        end=end,
+        stats_period=None,
+        environments=environments,
+    )
+    snuba_query_string = get_snuba_query_string(snuba_query)
```
ceorourke, line 202: This is one of the key changes here - the front end constructs a stringified query based on `snuba_query.query` AND `snuba_query.event_types`. This adds a join to the table for things like errors count with the `is:unresolved` query, or when you're using the dropdown to select "errors", "default", or "errors OR default" event types

ceorourke:
The users experiencing errors query is selecting data as a different name but it's otherwise the same, I don't know if that makes a difference to the outcome?
`events-stats`:


`SELECT (events._snuba_events.time AS _snuba_events.time), (uniq((events._snuba_events.tags[sentry:user] AS _snuba_events.tags[sentry:user])) AS _snuba_count_unique_user)`


anomaly detection:
`SELECT (events._snuba_events.time AS _snuba_events.time), (uniq((events._snuba_events.tags[sentry:user] AS _snuba_events.tags[sentry:user])) AS _snuba_count_unique_tags_sentry_user)`

src/sentry/seer/anomaly_detection/utils.py
```diff
@@ -143,37 +173,52 @@ def fetch_historical_data(
     granularity = snuba_query.time_window
 
     dataset_label = snuba_query.dataset
+
     if dataset_label == "events":
         # DATASET_OPTIONS expects the name 'errors'
         dataset_label = "errors"
     elif dataset_label == "generic_metrics":
+        # XXX: performance alerts in prod
         dataset_label = "transactions"
```
ceorourke, line 182: not sure if I need to be using the discover dataset here, it's hard to know since it's different locally

ceorourke: I think we _do_ want to use the discover dataset - unfortunately this differs in prod vs. locally so it's confusing but locally we make this request:

`api/0/organizations/sentry/events-stats/?interval=60m&project=1&query=event.type%3Atransaction&referrer=api.organization-event-stats&statsPeriod=9998m&yAxis=count%28%29`

and in prod we make this request:

`api/0/organizations/sentry/events-stats/?dataset=metricsEnhanced&interval=60m&project=1&query=event.type%3Atransaction&referrer=api.organization-event-stats&statsPeriod=9998m&yAxis=count%28%29`

the `events-stats` endpoint gets the dataset label from the request [here](https://github.com/getsentry/sentry/blob/master/src/sentry/api/bases/organization_events.py#L83-L89) and falls back to using discover if nothing is passed (like in production). 

In our case we're not getting the dataset label from the request but rather the snuba query object, so instead of it being "discover" it's "generic_metrics" in production and "transactions" in my local db. 

All that to say for performance metrics alerts we need to use the discover dataset for both `generic_metrics` and `transaction` dataset labels

tests/sentry/seer/anomaly_detection/test_store_data.py
```diff
@@ -70,13 +74,64 @@ def test_anomaly_detection_format_historical_data_two(self):
...
+            query_columns=["count_unique_tags_sentry_user"],
+            dataset=errors,
+            organization=self.organization,
+        )
         assert result == expected_return_value
 
     def test_anomaly_detection_fetch_historical_data(self):
         alert_rule = self.create_alert_rule(organization=self.organization, projects=[self.project])
         snuba_query = SnubaQuery.objects.get(id=alert_rule.snuba_query_id)
+        # CEO: despite passing event_type to store_event, the event type is default
```
ceorourke, line 91: I need to figure this out, it's causing issues in `test_organization_alert_rule_anomalies.py`

ceorourke: Something changed in Relay because after normalization the event type is changed to "default" https://github.com/getsentry/sentry/blob/master/src/sentry/event_manager.py#L434-L435 

I had added `event_type` to `self.store_event` and it used to work, I'm surprised tests haven't broken because of this.

src/sentry/seer/anomaly_detection/get_historical_anomalies.py
```diff
@@ -74,7 +81,9 @@ def get_historical_anomaly_data_from_seer(
             },
         )
         return None
-    formatted_data = format_historical_data(historical_data, dataset)
+    formatted_data = format_historical_data(
```
saponifi3d, line 84: nit: it'd be nice to be consistent with kwargs or args for `fetch` and `format`.

src/sentry/seer/anomaly_detection/utils.py
```diff
@@ -30,7 +35,36 @@ def translate_direction(direction: int) -> str:
     return direction_map[AlertRuleThresholdType(direction)]
 
 
-NUM_DAYS = 28
+def get_snuba_query_string(snuba_query: SnubaQuery) -> str:
+    """
+    Generate a query string that matches what events-stats does
```
saponifi3d, line 40: not sure what event-stats are from this comment, might be useful to add more context to this comment block (or omit it) - since it's kinda the same as the method name.

src/sentry/seer/anomaly_detection/utils.py
```diff
@@ -30,7 +35,36 @@ def translate_direction(direction: int) -> str:
...
+
+    if len(snuba_query.event_types) > 1:
+        # e.g. (is:unresolved) AND (event.type:[error, default])
+        event_types_list = [
+            SNUBA_QUERY_EVENT_TYPE_TO_STRING[event_type] for event_type in snuba_query.event_types
+        ]
+        event_types_string = "(event.type:["
+        for event_type in event_types_list:
+            event_types_string += event_type + ","
+        event_types_string = event_types_string[:-1]  # cut off the trailing comma
```
saponifi3d, line 56: nit, this can be a 1 liner and read a little more clearly:
```suggestion
        event_types_string += ', '.join(event_types_list)
```

src/sentry/seer/anomaly_detection/utils.py
```diff
@@ -75,23 +109,16 @@ def get_crash_free_historical_data(
...
-        {'time': 1721300400, 'count': 2}
-
-    For metrics_performance dataset/sessions data:
-        The count is stored separately from the timestamps, if there is no data the count is 0
     """
     formatted_data: list[TimeSeriesPoint] = []
-    nested_data = data.data.get("data", [])
 
     if dataset == metrics_performance:
+        nested_data = data.data.get("data", [])
```
saponifi3d, line 121: so, this is... `data.data.data`? 😅 is there anywhere we have control over any of those that we could rename the variable? i think adding context like what kind of data it is would help a lot.

src/sentry/seer/anomaly_detection/utils.py
```diff
@@ -102,28 +129,31 @@ def format_historical_data(data: SnubaTSResult, dataset: Any) -> list[TimeSeries
             ts_point = TimeSeriesPoint(timestamp=date.timestamp(), value=count)
             formatted_data.append(ts_point)
     else:
```
saponifi3d, line 131: thoughts on decomposing this function a little? breaking out the contents of the `if` statements might be a good way to handle that (both just return their formatted data).

tests/sentry/incidents/endpoints/test_organization_alert_rule_anomalies.py
```diff
@@ -49,8 +48,8 @@ def test_simple(self, mock_seer_request, mock_seer_store_request):
                     "timestamp": iso_format(two_weeks_ago + timedelta(minutes=1)),
                     "fingerprint": ["group1"],
                     "tags": {"sentry:user": self.user.email},
+                    "exception": [{"value": "BadError"}],
                 },
-                event_type=EventType.ERROR,
```
saponifi3d, line 53: TIL.

saponifi3d (approved):
lgtm! I just had some nitpicks - may want to have someone with more insights into seer take another glance too.

src/sentry/seer/anomaly_detection/utils.py
ceorourke: Sadly no, this is the format of `SnubaTSResult` which is typed in the function signature.

src/sentry/seer/anomaly_detection/utils.py
ceorourke: clarified with the endpoint name

PR #77757, opened 2024-09-18
fix(spans): Span's field test 

wmak:
- Need to look into why, but looks like the default tags list is
  different in EAP?


tests/sentry/api/endpoints/test_organization_spans_fields.py
```diff
@@ -89,6 +86,38 @@ def do_request(self, query=None, features=None, **kwargs):
                 **kwargs,
             )
 
+    def test_tags_list(self):
```
colin-sentry, line 89: are there two test_tags_list ?

wmak: Yeah one is inheriting from `OrganizationSpansTagsEndpointTest` while the other is that class

colin-sentry (approved):

PR #77758, opened 2024-09-18
feat(issue-views): Add better recommended searches

MichaelSun48:
Makes the recommended searches not garbage.

### DIFF
--- a/static/app/views/issueList/addViewPage.tsx
+++ b/static/app/views/issueList/addViewPage.tsx
@@ -26,13 +26,16 @@ interface SearchSuggestionListProps {
 }
 
 const RECOMMENDED_SEARCHES: SearchSuggestion[] = [
-  {label: 'Assigned to Me', query: 'assigned:me'},
-  {label: 'My Bookmarks', query: 'bookmarks:me'},
-  {label: 'Errors Only', query: 'status:unresolved level:error'},
+  {label: 'Prioritized', query: 'is:unresolved issue.priority:[high, medium]'},
+  {label: 'Assigned to Me', query: 'assigned_or_suggested:me is:unresolved'},
   {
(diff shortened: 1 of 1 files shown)


static/app/views/issueList/addViewPage.tsx
```diff
@@ -26,13 +26,16 @@ interface SearchSuggestionListProps {
...
+  {label: 'Assigned to Me', query: 'assigned_or_suggested:me is:unresolved'},
   {
-    label: 'Unhandled',
-    query: 'status:unresolved error.unhandled:True',
+    label: 'For Review',
+    query: 'is:unresolved is:for_review assigned_or_suggested:[me, my_teams, none]',
   },
+  {label: 'Request Errors', query: 'http.status_code:5*'},
+  {label: 'High Volume Issues', query: '!is:resolved timesSeen:>100'},
+  {label: 'Recent Errors', query: 'issue.category:error firstSeen:-24h !is:resolved'},
```
mrduncan, line 37: nit: should `!is:resolved` be either first or last for both of these? Right now one is first and one is last

mrduncan: These are also 1/2 `is:unresolved` and `!is:resolved` - should we standardize these or is it intentional to show off different behaviors?

mrduncan (approved):

static/app/views/issueList/addViewPage.tsx
MichaelSun48: Good point, I think the `is:unresolved` should trailing for all the queries that include it. I didn't intentionally differentiate between `is:unresolved` and `!is:resolved`, but I kind of like it to show off the different behaviors

PR #77759, opened 2024-09-18
fix(replay): Allow replay player instances to optionally be interactable/scrollable

ryan953:
This is a followup of https://github.com/getsentry/sentry/pull/77647

What we really want is for the main player instance (and maybe other players around the app, but none are set so far) to allow interaction and be cool like that. Instances inside the diff tools (for example) should not capture mouse/pointer events because it conflicts with the diff-slider tool.

### DIFF
--- a/static/app/components/replays/player/replayPlayer.tsx
+++ b/static/app/components/replays/player/replayPlayer.tsx
@@ -79,6 +79,7 @@ function useReplayerInstance() {
 
 interface Props extends HTMLAttributes<HTMLDivElement> {
   css?: Interpolation<Theme>;
+  ['data-inspectable']?: boolean;
   offsetMs?: undefined | number;
 }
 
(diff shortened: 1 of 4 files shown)


static/app/components/replays/player/replayPlayer.tsx
```diff
@@ -79,6 +79,7 @@ function useReplayerInstance() {
 
 interface Props extends HTMLAttributes<HTMLDivElement> {
   css?: Interpolation<Theme>;
+  ['data-inspectable']?: boolean;
```
ryan953, line 82: This was allowed because we extend `HTMLAttributes<HTMLDivElement>` and any `data-` is allowed, but i wanted to be more specific about the type so it's listed, and it's boolean.

billyvg (approved):
Just noticed we have `replays/player/replayPlayer.tsx` and `replays/replayPlayer.tsx` -- are there better names for these two components?

PR #77762, opened 2024-09-19
wip: GRPC server

### LINKED ISSUE #78065: Build prototype of pull based worker
Build a prototype of the pull based worker with RPC.

- Draft protocol buffers for tasks
- Prototype of gRPC server consumer rpc
- Prototype of worker using rpc client

https://www.notion.so/sentry/Worker-Push-or-Pull-9fec384a252d43e99fdc86b89ac3d84d?pvs=4#0ff8b10e4b5d808da7cbe18f6ae09214

john-z-yang:
Turns out our grpc version is too old, bumps grpcio version to 1.65.4.
Run `devenv sync` to install the new dependency, then run `pip install -e ../sentry-protos/py` to install the code gen from our sentry protos branch (`worker-pull`).

You can start the server with
```sentry run kafka-task-grpc-server```
It doesn't do anything yet, but it's useful to check if imports are successful.

John Sept: 19:
I decided to spend the entire day today to get rid of the `PendingTask` class that the `PendingTaskStore` has.

The result of this is that `PendingTaskStore` and the consumer will use a protobuf data structure directly instead of the python dataclass that wraps a protobuf data structure.

The main reason being that in order to use grpc we cannot pass `PendingTask` between the services (because it's a python class that wraps a protobuf type, plus some additional data), and instead of creating yet another Protobuf definition, I just added the fields into the existing protobuf message.

The new protobuf messages are structured like the following (name are probably going to change):

`Work`: An abstract unit of work, which only contains information like retry policy, the task ID, and the function parameters. Theses are the messages that are on Kafka.

`Task`: The actual physical task for the worker, which contains a `Work`, but also things like status, topic, partition, recieved_at. These are the messages the consumer stores into the `PendingTaskStore`.

This leaves us with a single python data structure
`PendingTask`: The Django model, used only in `PendingTaskStore` and it's just has all the fields a `Task` has, but flattened.

Now that `PendingTaskStore` is just taking a protobuf message (`Task`), this makes implementing the grpc server relatively simple.

```python
def GetTask(self, request: GetTaskRequest, context) -> GetTaskResponse:
    # task is already a GRPC message
    task = self.pending_task_store.get_pending_task(
        request.partition if request.partition else None,
        request.topic if request.topic else None,
    )
    return GetTaskResponse(task=task)
```

TODO:
- ~~Implement the functions defined in `ConsumerServicer` (we need to make sure to use the `PendingTaskStore` abstraction here)~~
- Have the worker fetch the tasks via grpc by calling gRPC stub in `TaskService`
- Test different failure modes


src/sentry/taskworker/config.py
```diff
@@ -75,10 +75,10 @@ def wrapped(func):
         return wrapped
 
     def retry_task(self, taskdata: PendingTasks) -> None:
-        task_message = taskdata.to_message()
+        message = taskdata.to_proto()
         self.producer.produce(
             ArroyoTopic(name=self.topic),
-            KafkaPayload(key=None, value=task_message.SerializeToString(), headers=[]),
+            KafkaPayload(key=None, value=message.work.SerializeToString(), headers=[]),
         )
```
john-z-yang, line 82: We might want to move this to `pending_task_store.py` because it is using `PendingTasks` which is the Django model which should be kept in `pending_task_store.py`. Also the only thing that calls this is `pending_task_store.py`

src/sentry/taskworker/service/service.py
```diff
@@ -18,47 +20,18 @@ class TaskService:
...
-            return serialize_task(task)
-        except PendingTasks.DoesNotExist:
-            return None
+    def set_task_status(self, *, task_id: int, task_status: Status.ValueType):
+        self.pending_task_store.set_task_status(task_id=task_id, task_status=task_status)
 
-    def complete_task(self, *, task_id: int) -> RpcTask | None:
-        return self.set_task_status(task_id=task_id, task_status=PendingTasks.States.COMPLETE)
+    def complete_task(self, *, task_id: int):
+        self.set_task_status(task_id=task_id, task_status=COMPLETE)
```
john-z-yang, line 34: Removed the return value here because I'm not clear what we're trying to accomplish with the return value (nothing uses it)

src/sentry/taskworker/config.py
markstory: Having this accept a `Task` sounds like a good plan to me.

src/sentry/taskworker/models.py
```diff
@@ -6,8 +6,15 @@
 from django.db import models
 from django.utils import timezone
 from sentry_protos.hackweek_team_no_celery_pls.v1alpha.pending_task_pb2 import (
-    PendingTask as PendingTaskProto,
+    COMPLETE,
+    FAILURE,
+    PROCESSING,
+    RETRY,
```
markstory, line 12: `buf` will want us to prefix all these enum values with the name of the enum.

src/sentry/taskworker/service/client.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+
+    def __init__(self):
+        self.pending_task_store = PendingTaskStore()
+        self.host = "localhost"
+        self.server_port = 50051
+        self.channel = grpc.insecure_channel(f"{self.host}:{self.server_port}")
+        self.stub = ConsumerStub(self.channel)
+
+    def get_task(self, partition: int | None = None, topic: str | None = None) -> Task | None:
```
markstory, line 34: This is a nice way to put a simpler to operate interface on top of grpc :clap:

PR #77774, opened 2024-09-19
fix(platforms): Split up between cloudflare pages and workers

AbhiPrasad:
In https://github.com/getsentry/sentry/pull/77769 I added the cloudflare sdk as a new platform for onboarding docs.

This PR builds on top of that by reverting the change to add `javascript-cloudflare` as a platform and instead have two platforms for cloudflare workers and cloudflare pages (given they have completely different setup instructions and use cases).

In our docs pages and workers have two different configure paths: https://docs.sentry.io/platforms/javascript/guides/cloudflare/#configure


priscilawebdev (approved):
lgtm. We have to add these platforms to the frontend as well so that they are visible there

Lms24 (approved):

AbhiPrasad:
> lgtm. We have to add these platforms to the frontend as well so that they are visible there

yup will do that next!

Issue #77770, opened 2024-09-19
Org stats: project abuse limit reported as internal


obostjancic:

priscilawebdev:
This was clarified internally and is an outcome we want to expose as internal - as it is already.

PR #77775, opened 2024-09-19
fix(stats): Expose project abuse limit

priscilawebdev:
Contributes to https://github.com/getsentry/sentry/issues/77770


shellmayr (approved):

priscilawebdev:
This was clarified internally and is an outcome we want to expose as internal - as it is already.

PR #77776, opened 2024-09-19
chore(auth) Remove require email verification feature

markstory:
This behavior was added back in 2021 and has never been activated. However, the serializer code if used will trigger SiloLimit errors and not work. Instead of patching up an unused feature, removing it solves the problem and makes maintenance easier in the future.

Fixes SENTRY-3F4E


vgrozdanic (approved):

src/sentry/models/organization.py
```diff
@@ -189,7 +189,7 @@ class flags(TypedClassBitField):
         # Temporarily opt out of new visibility features and ui
         disable_new_visibility_features: bool
 
-        # Require and enforce email verification for all members.
+        # Require and enforce email verification for all members. (deprecated, not in use)
```
vgrozdanic, line 192: nit: I don't know how BItField works, and if this would even work, but does it makes sense to make this also a `@property` of a class and add a deprecation warning if someone tries to access this field?

markstory: Bitfield deserializes the storage integer into a set of flags that quack like booleans, enabling usage like `org.flags.require_email_verification` to be read/written. The `flags` instance is generated via metaclasses so adding deprecations via `@property` will be challenging 🤔

PR #77789, opened 2024-09-19
test(spans): Add sampling columns temporarily

wmak:
- This adds the sampling columns to help explore the extrapolation data better, We should not use this in the UI and its just for testing


src/sentry/utils/snuba.py
```diff
@@ -188,6 +188,9 @@ def log_snuba_info(content):
     # message also maps to span description but gets special handling
     # to support wild card searching by default
     "message": "name",
+    # These sample columns are for debugging only and shouldn't be used
+    "sampling_weight": "sampling_weight",
+    "sampling_factor": "sampling_factor",
```
k-fish, line 193: These act as hidden columns and only work if queried against like the rest of our hidden columns, yeah?

k-fish (approved):

src/sentry/utils/snuba.py
wmak: yep

PR #77794, opened 2024-09-19
feat(issue_platform): Support deleting groups and related data

armenzg:
This allows the Issue Platform to delete groups and all related occurrence events.

### DIFF
--- a/src/sentry/deletions/defaults/group.py
+++ b/src/sentry/deletions/defaults/group.py
@@ -159,6 +159,42 @@ class ErrorEventsDeletionTask(EventsBaseDeletionTask):
             eventstream.backend.end_delete_groups(eventstream_state)
 
 
+class EventIssuePlatformDeletionTask(EventsBaseDeletionTask):
+    """
+    This class helps delete Issue Platform events which use the new Clickhouse light deletes.
+    """
+
+    dataset = Dataset.IssuePlatform
+
+    def delete_events_from_nodestore(self, events: Sequence[Event]) -> None:
+        # Remove from nodestore
+        node_ids = [
+            Event.generate_node_id(event.project_id, event._snuba_data["occurrence_id"])
+            for event in events
+        ]
+        nodestore.backend.delete_multi(node_ids)
+
+    def delete_events_from_snuba(self, project_groups: Mapping[int, Sequence[int]]) -> None:
+        pass
+
+    # def delete_issue_platform_events(self, issue_platform_events: Sequence[Event]) -> None:
+    #     pass
+    # print(issue_platform_events)
+    # XXX: For now, we will only delete one group. We will change this in the future
+    # group_id = issue_platform_group_ids[0]
+    # query = DeleteQuery(
+    #     EntityKey.IssuePlatform.value,
+    #     {"project_id": [self.project.id], "occurrence_id": [occurrence_id]},
+    # )
+    # request = Request(
+    #     dataset=Dataset.IssuePlatform.value,
+    #     app_id=self.referrer,
+    #     query=query,
+    #     tenant_ids=self.tenant_ids,
+    # )
+    # results = bulk_snuba_queries([request])[0]["data"]
+
+
 class GroupDeletionTask(ModelDeletionTask[Group]):
     # Delete groups in blocks of 1000. Using 1000 aims to
     # balance the number of snuba replacements with memory limits.
@@ -194,13 +230,19 @@ class GroupDeletionTask(ModelDeletionTask[Group]):
         for model in _GROUP_RELATED_MODELS:
             child_relations.append(ModelRelation(model, {"group_id__in": group_ids}))
 
-        error_groups, _ = separate_by_group_category(instance_list)
+        error_groups, issue_platform_groups = separate_by_group_category(instance_list)
 
         # If this isn't a retention cleanup also remove event data.
         if not os.environ.get("_SENTRY_CLEANUP"):
             if error_groups:
                 params = {"groups": error_groups}
                 child_relations.append(BaseRelation(params=params, task=ErrorEventsDeletionTask))
+            # XXX: Feature gate here
+            if issue_platform_groups:
+                params = {"groups": issue_platform_groups}
+                child_relations.append(
+                    BaseRelation(params=params, task=EventIssuePlatformDeletionTask)
+                )
 
         self.delete_children(child_relations)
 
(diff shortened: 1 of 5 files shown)


src/sentry/deletions/defaults/group.py
```diff
@@ -159,6 +164,53 @@ def delete_events_from_snuba(self) -> None:
...
+            for event in events
+        ]
+        nodestore.backend.delete_multi(node_ids)
+
+    def delete_events_from_snuba(self) -> None:
+        requests = []
+        for group in self.groups:
+            query = DeleteQuery(
+                self.dataset.value,
+                column_conditions={"project_id": [group.project_id], "group_id": [group.id]},
```
MeredithAnya, line 202: Since these are all in the same project, you can make one request with multiple group ids. It will be better to have one request with multiple group ids than many requests each with one group id

JoshFerge (commented):
makes sense to me. point about batching groups together makes sense.

MeredithAnya (approved):
I think this looks good from SNS standpoint

src/sentry/deletions/defaults/group.py
```diff
@@ -159,6 +164,53 @@ def delete_events_from_snuba(self) -> None:
...
+            self.delete_events_from_snuba()
+            return False
+
+    def delete_events_from_nodestore(self, events: Sequence[Event]) -> None:
+        # We delete by the occurrence_id instead of the event_id
+        node_ids = [
+            Event.generate_node_id(event.project_id, event._snuba_data["occurrence_id"])
+            for event in events
+        ]
+        nodestore.backend.delete_multi(node_ids)
```
wedamija, line 195: This deletes the occurrence from nodestore. Does it delete the event associated with the occurrence too?

Note that we should only delete the event if it was created during occurrence ingestion. If we passed an event id, then the event should remain.

src/sentry/deletions/defaults/group.py
```diff
@@ -159,6 +164,53 @@ def delete_events_from_snuba(self) -> None:
...
+                column_conditions={"project_id": [project_id], "group_id": group_ids},
+            )
+            request = Request(
+                dataset=self.dataset.value,
+                app_id=self.referrer,
+                query=query,
+                tenant_ids=self.tenant_ids,
+            )
+            requests.append(request)
+        bulk_snuba_queries(requests)
```
wedamija, line 211: Cool, so we can actually submit deletion queries to snuba now, rather than relying on replacements?

Do these deletions run in real time, or are the enqueued?

armenzg: The rows immediately become unavailable to be queried, however, they will truly be deleted later.

src/sentry/deletions/defaults/group.py
armenzg: We only delete the occurrence event from the Issue Platform dataset (not the Events one). The error event from the Events dataset is left untouched.

See the test logic below:
```python
        # The original event and group still exist
        assert Group.objects.filter(id=event.group_id).exists()
        event_node_id = Event.generate_node_id(event.project_id, event.event_id)
        assert nodestore.backend.get(event_node_id)

        # The Issue Platform group and occurrence are deleted
        assert issue_platform_group.issue_type == FeedbackGroup
        assert not Group.objects.filter(id=issue_platform_group.id).exists()
        occurrence_node_id = Event.generate_node_id(occurrence.project_id, occurrence.id)
        assert not nodestore.backend.get(occurrence_node_id)
        # Assert that occurrence is gone
        assert self.select_issue_platform_events(self.project.id) == []
```

armenzg: Under what circumstances does the event get created during occurrence ingestion? I don't think we spend time finding the code for that.

Last time we spoke we agreed to keep it simple and let the event expire with the TTL.

wedamija: > Under what circumstances does the event get created during occurrence ingestion? I don't think we spend time finding the code for that.

It is hard to tell the difference between the two post ingestion - we don't differentiate since the occurrence just has an event associated with it. The split happens here: https://github.com/getsentry/sentry/blob/a86b9b672709bc9c4558cffb2c825965b8cee0d1/src/sentry/issues/occurrence_consumer.py#L324-L339

> Last time we spoke we agreed to keep it simple and let the event expire with the TTL.

I don't specifically remember that, but since these events can't be accessed in any reasonable way maybe that's fine, since there's nothing linking to them. I think we should probably mention that explicitly in a comment here if that's the case so that we're aware.

wedamija: > We only delete the occurrence event from the Issue Platform dataset (not the Events one). The error event from the Events dataset is left untouched.
> 
> See the test logic below:
> 
> ```python
>         # The original event and group still exist
>         assert Group.objects.filter(id=event.group_id).exists()
>         event_node_id = Event.generate_node_id(event.project_id, event.event_id)
>         assert nodestore.backend.get(event_node_id)
>         assert self.select_error_events(self.project.id) == error_expected
>         # The Issue Platform group and occurrence are deleted
>         assert issue_platform_group.issue_type == FeedbackGroup
>         assert not Group.objects.filter(id=issue_platform_group.id).exists()
>         occurrence_node_id = Event.generate_node_id(occurrence.project_id, occurrence.id)
>         assert not nodestore.backend.get(occurrence_node_id)
>         # Assert that occurrence is gone
>         assert self.select_issue_platform_events(self.project.id) is None
> ```

I'm not talking about the events dataset, just the event in Nodestore.

src/sentry/deletions/defaults/group.py
wedamija: How long does this query take? Is it dependent on the number of occurrences/events associated with the group?

src/sentry/deletions/defaults/group.py
armenzg: @wedamija We check in the test that it is not deleted from either

src/sentry/deletions/defaults/group.py
wedamija: Discussed this on slack. The deletion happens sync, so the query might time out for very large groups. We shouldn't release this on saas without handling cases like this, but I'm fine to merge for testing

wedamija (approved):

src/sentry/deletions/defaults/group.py
armenzg: Thanks, @wedamija for all your help!

I have added a comment to try to encompass what you explained.

image

PR #77795, opened 2024-09-19
ref(onboarding): Consolidate Manual button in integrations

### LINKED ISSUE #77538: Move integration installation code to onboarding files

priscilawebdev:
https://github.com/user-attachments/assets/f5e0a769-5d1a-492d-9b15-cc4627649af9

closes https://github.com/getsentry/sentry/issues/77538


static/app/gettingStartedDocs/node/awslambda.tsx
```diff
@@ -12,9 +13,33 @@ import {
 } from 'sentry/components/onboarding/gettingStartedDoc/utils/feedbackOnboarding';
 import {getJSServerMetricsOnboarding} from 'sentry/components/onboarding/gettingStartedDoc/utils/metricsOnboarding';
 import {t, tct} from 'sentry/locale';
+import {trackAnalytics} from 'sentry/utils/analytics';
 import {getInstallConfig, getSdkInitSnippet} from 'sentry/utils/gettingStartedDocs/node';
 
-type Params = DocsParams;
+export enum InstallationMode {
```
priscilawebdev, line 19: I find this repetition very verbose and have the feeling we could make it more reusable 🤔

static/app/views/onboarding/components/useOnboardingQueryParams.tsx
```diff
@@ -1,11 +1,16 @@
...
 import useLocationQuery from 'sentry/utils/url/useLocationQuery';
 import {useLocation} from 'sentry/utils/useLocation';
 import {useNavigate} from 'sentry/utils/useNavigate';
 
 type QueryValues = {
+  /**
+   * Used to show the installation mode for certain platforms, e.g. manual or auto.
+   * This is defined inside of a platform file
+   */
+  installationMode: string;
```
priscilawebdev, line 13: more refactoring is needed, I would like to get this from the hook `useLoadGettingStarted` but added this here for now.

static/app/views/onboarding/components/useOnboardingQueryParams.tsx
```diff
@@ -31,6 +36,7 @@ export function useOnboardingQueryParams(): [
       product: decodeList,
       showLoader: decodeBoolean,
       showManualSetup: decodeBoolean,
+      installationMode: decodeScalar,
```
priscilawebdev, line 39: also it would be better to return `MANUAL` or `AUTO` from an enum but I did not want to add a custom `decode` here now

static/app/views/onboarding/integrationSetup.tsx
```diff
@@ -187,8 +173,55 @@ function IntegrationSetup(props: Props) {
...
+  if (!docsConfig || !dsn || !projectKeyId) {
+    return (
+      <LoadingError
+        message={t(
+          'The getting started documentation for this platform is currently unavailable.'
+        )}
+        onRetry={refetch}
+      />
+    );
+  }
```
priscilawebdev, line 189: In the future, we might consider moving this to the parent component, as we currently have the same check within `SdkDocumentation`

static/app/views/onboarding/components/useOnboardingQueryParams.tsx
ArthurKnaus: Good idea, but I would not put it there yet as only 2 platforms are using it. let's see first if more similar use cases arise.

static/app/gettingStartedDocs/node/awslambda.tsx
ArthurKnaus: Personally I would define an enum that is tied to this integration setup use-case and import it here and the two docs files + use it to check the param value in `setupDocs`

static/app/views/onboarding/integrationSetup.tsx
```diff
@@ -187,8 +173,55 @@ function IntegrationSetup(props: Props) {
...
+      <SetupIntroduction
+        stepHeaderText={t('Automatically instrument %s SDK', platform.name)}
+        platform={platform.id}
+      />
+      {docsConfig.platformOptions && (
+        <PlatformOptionsControl
+          platformOptions={docsConfig.platformOptions}
+          onChange={docsConfig.onboarding.onPlatformOptionsChange?.(docParams)}
+        />
+      )}
```
ArthurKnaus, line 223: I'm not sure if we really want to display all platform options here. That can have unwanted side effects, when another option is added to the `awslambda` files.

ArthurKnaus (commented):
I am not 100% sure about the UX of the this solution. As a user, I would be very surprised by the big layout shift, on clicking the toggle.
@matejminar @ale-cota wdyt?

priscilawebdev:
> I am not 100% sure about the UX of this solution. As a user, I would be very surprised by the big layout shift, on clicking the toggle. @matejminar @ale-cota wdyt?

I agree with you @ArthurKnaus, but in my opinion we should improve this step by step and this PR would be the first step. I also don't believe that many users use the integrations during onboarding. It would be interesting to confirm this.

priscilawebdev:
We also wanted to provide a way to switch between manual and auto and I don't see a simple solution without a lot of refactoring.

static/app/views/onboarding/components/useOnboardingQueryParams.tsx
priscilawebdev: if we do not define it somewhere,  it won't have types when checking for it in the `setupDocs.tsx`

priscilawebdev: I understand...was trying to avoid the type check in `setupDocs` to be `any`. https://github.com/getsentry/sentry/blob/d65139251f6e157810ab1bdb2af1cef0d7c39544/static/app/views/onboarding/setupDocs.tsx#L41

matejminar:
yeah, not a huge fan of changing the content dramatically (especially content above the toggle) but I don't have a good alternative solution in mind :/

static/app/views/onboarding/integrationSetup.tsx
ale-cota: How about if we update the text, so that it's relevant to both auto and manual and leave the checkboxes there, but disabled to avoid the layout shift? I don't love it, but it's the only thing I could think of to avoid it.
If not, I'm also fine to merge it as-is and improve it later. As Pri pointed out, there aren't that many users who select this option.

ArthurKnaus (approved):

PR #77796, opened 2024-09-19
ref(dx): Detect bad test runs

evanpurkhiser:
To avoid confusion detect when people try to run jest directly and hint
for them to use `yarn test` instead of jest directly.


scttcper (approved):
test flake

scttcper:
breaks the test balancing command `yarn -s jest --listTests --json`

PR #77797, opened 2024-09-19
analytics(issue-views): Add analytics to issue views

MichaelSun48:
Adds analytics for the following events:

- User switched views
- User updated a view with a new query or sort 
- User discarded changes in a view 
- User renamed a view 
- User duplicated a view
- User deleted a view
- User reordered their views
- User clicked on `+ Add View`
- User saved a custom query from add view menu 
- User saved a deprecated saved search from add view menu 
- User saved a recommended view from add view menu 
- User opened a shared view from another user 
- User discarded a temporary view
- User saved a temporary view

### DIFF
--- a/static/app/utils/analytics/issueAnalyticsEvents.tsx
+++ b/static/app/utils/analytics/issueAnalyticsEvents.tsx
@@ -272,6 +272,7 @@ export type IssueEventParameters = {
     priority: PriorityLevel;
   };
   'issues_tab.viewed': {
+    issue_views_enabled: boolean;
     num_issues: number;
     num_new_issues: number;
     num_old_issues: number;
--- a/static/app/utils/analytics/issueAnalyticsEvents.tsx
+++ b/static/app/utils/analytics/issueAnalyticsEvents.tsx
@@ -221,6 +221,30 @@ export type IssueEventParameters = {
     search_source: string;
     search_type: string;
   };
+  'issue_views.add_view.clicked': {};
+  'issue_views.add_view.custom_query_saved': {
+    query: string;
+  };
+  'issue_views.add_view.recommended_view_saved': {
+    label: string;
+    persisted: boolean;
+    query: string;
+  };
+  'issue_views.add_view.saved_search_saved': {
+    query: string;
+  };
+  'issue_views.deleted_view': {};
+  'issue_views.discarded_changes': {};
+  'issue_views.duplicated_view': {};
+  'issue_views.renamed_view': {};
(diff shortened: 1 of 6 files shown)


static/app/views/issueList/groupSearchViewTabs/draggableTabBar.tsx
```diff
@@ -267,6 +294,9 @@ export function DraggableTabBar({
       });
       setTabs(newTabs);
       tabListState?.setSelectedKey(tempId);
+      trackAnalytics('issue_views.add_view_clicked', {
```
roggenkemper, line 297: i think some of these names are slightly different : `issue_views.add_view_clicked` vs `issue_views.add_view.clicked`

static/app/utils/analytics/issueAnalyticsEvents.tsx
```diff
@@ -349,6 +373,21 @@ export const issueEventMap: Record<IssueEventKey, string | null> = {
   'issue_error_banner.proguard_misconfigured.clicked':
     'Proguard Potentially Misconfigured Issue Error Banner Link Clicked',
   'issues_tab.viewed': 'Viewed Issues Tab',
+  'issue_views.switched_views': 'Issue Views: Switched Views',
```
roggenkemper, line 376: not sure how many of these are button clicks but if they are ,  we don't have to add an explicit call to `trackAnalytics` - https://develop.sentry.dev/development/analytics/#button-click-analytics

static/app/views/issueList/groupSearchViewTabs/draggableTabBar.tsx
MichaelSun48: ah my bad, I updated that in the definition and forgot to change it in the usages

static/app/utils/analytics/issueAnalyticsEvents.tsx
MichaelSun48: Good call - i there are definitely a couple places where we can replace the track analytics with this

static/app/views/issueList/overview.tsx
```diff
@@ -197,6 +197,11 @@ class IssueListOverview extends Component<Props, State> {
     }
     this.fetchTags();
     this.fetchMemberList();
+    this.props.setRouteAnalyticsParams?.({
+      issueViewsEnabled: this.props.organization.features.includes(
```
roggenkemper, line 201: we usually want to use snake case

roggenkemper (approved):

PR #77800, opened 2024-09-19
feat(insights): Occlude very large queries in the hover tooltip

0Calories:
Extremely long queries can be overwhelming and ugly when they appear in the tooltip preview after hovering over a query description. This PR uses the `ClippedBox` component to add a max height to queries, and provides a "View full query" button to encourage the user to inspect the query in its entirety. Clicking on this button will just jump to the summary page for that query

Applies to both SQL and MongoDB queries.

image

### DIFF
--- a/static/app/views/insights/common/components/fullSpanDescription.tsx
+++ b/static/app/views/insights/common/components/fullSpanDescription.tsx
@@ -8,11 +8,13 @@ import {IconOpen} from 'sentry/icons';
 import {t} from 'sentry/locale';
 import {space} from 'sentry/styles/space';
 import {SQLishFormatter} from 'sentry/utils/sqlish/SQLishFormatter';
+import {MutableSearch} from 'sentry/utils/tokenizeSearch';
 import {useLocation} from 'sentry/utils/useLocation';
 import {useNavigate} from 'sentry/utils/useNavigate';
+import {useSpansIndexed} from 'sentry/views/insights/common/queries/useDiscover';
 import {useFullSpanFromTrace} from 'sentry/views/insights/common/queries/useFullSpanFromTrace';
 import {prettyPrintJsonString} from 'sentry/views/insights/database/utils/jsonUtils';
-import {ModuleName} from 'sentry/views/insights/types';
+import {ModuleName, SpanIndexedField} from 'sentry/views/insights/types';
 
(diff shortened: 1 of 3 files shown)


0Calories:
Putting this back into draft, I'm making some new decisions on how to handle this behaviour as a whole

static/app/views/insights/common/components/fullSpanDescription.tsx
```diff
@@ -53,24 +76,38 @@ export function FullSpanDescription({
...
+      } else if (fullSpan?.sentry_tags?.description) {
+        result = prettyPrintJsonString(fullSpan?.sentry_tags?.description);
       } else {
         stringifiedQuery = description || fullSpan?.sentry_tags?.description || 'N/A';
       }
 
-      return <CodeSnippet language="json">{stringifiedQuery}</CodeSnippet>;
+      if (result) {
+        const {prettifiedQuery} = result;
+        stringifiedQuery = prettifiedQuery;
```
DominikB2014, line 95: we could just squeeze this in one line
```suggestion
        stringifiedQuery = result.prettifiedQuery;
```

static/app/views/insights/common/components/fullSpanDescription.tsx
```diff
@@ -29,16 +36,32 @@ export function FullSpanDescription({
...
   const {
     data: fullSpan,
     isLoading,
     isFetching,
-  } = useFullSpanFromTrace(group, [INDEXED_SPAN_SORT], Boolean(group), filters);
+  } = useFullSpanFromTrace(group, [INDEXED_SPAN_SORT], Boolean(indexedSpan), filters);
 
-  const description = fullSpan?.description ?? shortDescription;
+  const description =
+    indexedSpan?.['span.description'] ?? fullSpan?.description ?? shortDescription;
```
DominikB2014, line 61: So correct me if i'm wrong here, but were prioritizing the span description from the indexed span because were not showing large span descriptions anyways, so the truncation is okay. But if we return back a indexed span, then were still fetching the full trace to get the full span description to use as a backup?

Could we have just stayed with the original approach of only fetching the full span description, if were going to always query for it anyways? Or is there something i'm missing?

DominikB2014 (approved):
lgtm overall! just some specific questions about a portion of the code

PR #77804, opened 2024-09-19
feat(insights): ensure all links within domain views, stay within domain views

DominikB2014:
Work for https://github.com/getsentry/sentry/issues/77572

This ensures any links within a module that is within a domain view, stays within that domain view.
(this means you can now go to the http domain summary, resource summary, etc)

This is only necessary to support existing insights, and domain view work, at the same time. Something to keep in mind, this is all likely to change at somepoint when we remove the flag and do the release, i was trying to be "non-intrusive" as possible with the existing stuff so it works well in two places

This was quite tricky to do, so to make it support both at the same time, i'm using the pathname rather then query params in order to keep track of what page to render. This is because all modules use subpaths to get to their respective summary pages. This all handled in a new component called `ModuleRenderer` which looks at the current url to render the appropriate module page/summary page.

There are other approaches to this problem, 
1. Create a wrapper component for each module, the wrapper would have all the domain view controls (overview, web vitals, http, assets). The issue I found with this approach, is that it seemed like a lot of duplication. And we'll to handle the case where a module fits under many domain views.

### DIFF
--- a/static/app/views/insights/pages/moduleRenderer.tsx
+++ b/static/app/views/insights/pages/moduleRenderer.tsx
@@ -1,0 +1,55 @@
+import ResourcesLandingPage from 'sentry/views/insights/browser/resources/views/resourcesLandingPage';
+import ResourceSummaryPage from 'sentry/views/insights/browser/resources/views/resourceSummaryPage';
+import PageOverview from 'sentry/views/insights/browser/webVitals/views/pageOverview';
+import WebVitalsLandingPage from 'sentry/views/insights/browser/webVitals/views/webVitalsLandingPage';
+import CachesLandingPage from 'sentry/views/insights/cache/views/cacheLandingPage';
+import DatabaseLandingPage from 'sentry/views/insights/database/views//databaseLandingPage';
+import HTTPDomainSummaryPage from 'sentry/views/insights/http/views/httpDomainSummaryPage';
+import HTTPLandingPage from 'sentry/views/insights/http/views/httpLandingPage';
+import LLMLandingPage from 'sentry/views/insights/llmMonitoring/views/llmMonitoringLandingPage';
+import AppStartsLandingPage from 'sentry/views/insights/mobile/appStarts/views/appStartsLandingPage';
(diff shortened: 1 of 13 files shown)


static/app/views/insights/pages/moduleRenderer.tsx
```diff
@@ -0,0 +1,55 @@
...
+  [ModuleName.RESOURCE]: ResourcesLandingPage,
+  [ModuleName.VITAL]: WebVitalsLandingPage,
+  [ModuleName.MOBILE_UI]: MobileUiLandingPage,
+  [ModuleName.SCREEN_LOAD]: ScreenLoadLandingPage,
+  [ModuleName.MOBILE_SCREENS]: ScreensLandingPage,
+  [ModuleName.APP_START]: AppStartsLandingPage,
+  [ModuleName.OTHER]: () => null,
+};
+
+const ModuleSubpages: Partial<Record<ModuleName, React.ComponentType>> = {
```
DominikB2014, line 51: Todo, fill this in

0Calories:
This is fine if it's just a temporary change and to help you debug, but I'm not quite following the rationale for having a structure like this. It's understandably more tedious, but in my opinion a much cleaner approach, if we were to stick to the pattern of specifying individual routes.

The approach you have here to automatically determine the URL based on the module is smart, but it's a pattern that isn't followed anywhere else in our codebase and has the potential to be difficult to build upon and restricts us if we need to refactor in the future.

For the sake of organization, we should prioritize separation of concerns rather than minimizing some code duplication. Here's what I suggest:

- Make separate PRs for each new domain landing page
- Use a query param to determine which tab is selected, aka the module to display in the current domain
- Define separate routes for the subpages, like span summary pages for the individual domains
- You can still greatly reduce code duplication by using a pattern like [Composition & Specialization](https://legacy.reactjs.org/docs/composition-vs-inheritance.html). You can use Composition to define the general structure of the domain landing pages and then Specialization to implement the specific structures for each landing page

DominikB2014:
Closing in favour of another approach

Issue #68731, opened 2024-04-11
[Replays] Duplicating elements when viewing Replays in Ui


Fwang36:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

Reported by a user.

When viewing replays in our Ui, sometimes elements can start duplicating.  Link to replay provided by the user in linked Jira ticket.  Sometimes the elements display properly but seems to break again later.  Attaching screenshots of the behavior in Jira as well.

In linked replay - starts at around 18:44 of the replay.  Interaction with the slider causes buttons below it to start duplicating in the direction where the slider is being dragged.

[Linked Jira Ticket](https://getsentry.atlassian.net/browse/FEEDBACK-2142)

### Expected Result

Elements should not duplicate

### Actual Result

Duplication

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2142) by [Unito](https://www.unito.io)

aaronjensen:
I'm seeing this too. It's good to know that it's a problem with the viewer. It's also inconsistent. If I refresh the replay page, sometimes it doesn't happen anymore.

bruno-garcia:
Thanks for letting us known @aaronjensen. Do you have a link to an example Replay you could share?

bruno-garcia:
Seems like it's an rrweb bug: 
* https://github.com/rrweb-io/rrweb/issues/1449

aaronjensen:
I've supplied a few replays to support. With regards to it being an rrweb bug, I don't really know how to respond to that. I don't pay for rrweb, I pay for Sentry, so I kind of have to assume that the origin doesn't matter -- it's a Sentry bug to Sentry's customers, right?

billyvg:
We use GitHub for tracking development work, he's just adding context to the issue, not asking anything of you.

aaronjensen:
Sounds good, thanks!

billyvg:
Unassigning myself for now, I won't be able to take this on until mid-July at the earliest

joshua-webdev:
My team is experiencing something like this as well, in many session replays. The problem seems to get worse as you approach the end of the replay. I'm surprised there's been no movement on this since April - the feature is basically unusable if you can't see what's happening in the recording.

billyvg:
@joshua-webdev Can you email me some example replays that show this? `billy at sentry.io` -- we've made some fixes in other parts of the replay player, but we can take a closer look at this specifically.

[...]

PR #77808, opened 2024-09-19
feat(replay): Remove breadcrumb HTML + DOM node count chart

billyvg:
These two features require us to replay an entire replay to generate the HTML, as well as count DOM nodes. This is pretty CPU intensive, especially with larger replays, and seems to factor into generating incorrect HTML for the replay itself (ref https://github.com/getsentry/sentry/issues/68731).

Temporarily remove these features while we think of alternatives that perform better.


michellewzhang:
rip my replayerStepper refactor

billyvg:
Ah damn, web vitals relies on this too 😭 Going to leave this in for now.

PR #77813, opened 2024-09-19
feat(anomaly detection):preview chart proxy api endpoint

mifu67:
Build out the preview chart proxy endpoint. Calls a placeholder function.

### DIFF
--- a/src/sentry/api/endpoints/organization_events_anomalies.py
+++ b/src/sentry/api/endpoints/organization_events_anomalies.py
@@ -47,7 +47,7 @@ class OrganizationEventsAnomaliesEndpoint(OrganizationEventsV2EndpointBase):
         },
         examples=OrganizationExamples.GET_HISTORICAL_ANOMALIES,
     )
-    def _format_historical_data(self, data) -> list[TimeSeriesPoint]:
+    def _format_historical_data(self, data) -> list[TimeSeriesPoint] | None:
         """
         Format EventsStatsData into the format that the Seer API expects.
         EventsStatsData is a list of lists with this format:
@@ -84,7 +84,7 @@ class OrganizationEventsAnomaliesEndpoint(OrganizationEventsV2EndpointBase):
             )
 
         anomalies = get_historical_anomaly_data_from_seer_preview(
-            project_id, config, historical_data, current_data
+            current_data, historical_data, project_id, config
         )
         # NOTE: returns None if there's a problem with the Seer response
         if anomalies is None:
--- a/src/sentry/api/endpoints/organization_events_anomalies.py
+++ b/src/sentry/api/endpoints/organization_events_anomalies.py
@@ -1,0 +1,98 @@
+from drf_spectacular.utils import extend_schema
+from rest_framework.request import Request
+from rest_framework.response import Response
+
+from sentry import features
+from sentry.api.api_owners import ApiOwner
+from sentry.api.api_publish_status import ApiPublishStatus
+from sentry.api.base import region_silo_endpoint
+from sentry.api.bases.organization_events import OrganizationEventsV2EndpointBase
+from sentry.api.exceptions import ResourceDoesNotExist
+from sentry.api.paginator import OffsetPaginator
+from sentry.api.serializers.base import serialize
+from sentry.apidocs.constants import (
+    RESPONSE_BAD_REQUEST,
+    RESPONSE_FORBIDDEN,
+    RESPONSE_NOT_FOUND,
+    RESPONSE_UNAUTHORIZED,
+)
+from sentry.apidocs.parameters import GlobalParams
+from sentry.apidocs.utils import inline_sentry_response_serializer
+from sentry.models.organization import Organization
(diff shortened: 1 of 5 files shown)


src/sentry/api/endpoints/organization_events_anomalies.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@region_silo_endpoint
+class OrganizationEventsAnomaliesEndpoint(OrganizationEventsV2EndpointBase):
+    owner = ApiOwner.ALERTS_NOTIFICATIONS
+    publish_status = {
+        "POST": ApiPublishStatus.EXPERIMENTAL,
+    }
+
+    @extend_schema(
```
iamrajjoshi, line 36: in the future, if the api is experimental, documenting isn't explicitly required - 🥔 for doing this anyway!

iamrajjoshi (commented):
api design lgtm

tests/sentry/api/endpoints/test_organization_historical_anomalies.py
```diff
@@ -0,0 +1,52 @@
...
+    AlertRuleSeasonality,
+    AlertRuleSensitivity,
+    AlertRuleThresholdType,
+)
+from sentry.testutils.cases import APITestCase
+from sentry.testutils.helpers.features import with_feature
+from sentry.testutils.outbox import outbox_runner
+
+
+class OrganizationEventsAnomaliesEndpointTest(APITestCase):
```
natemoo-re, line 13: Can you add a test for the **less than 7 days historical data should return `[]`** case? Are there any additional error cases to cover?

src/sentry/api/endpoints/organization_events_anomalies.py
```diff
@@ -0,0 +1,98 @@
...
+            400: RESPONSE_BAD_REQUEST,
+            401: RESPONSE_UNAUTHORIZED,
+            403: RESPONSE_FORBIDDEN,
+            404: RESPONSE_NOT_FOUND,
+        },
+        examples=OrganizationExamples.GET_HISTORICAL_ANOMALIES,
+    )
+    def _format_historical_data(self, data) -> list[TimeSeriesPoint] | None:
+        """
+        Format EventsStatsData into the format that the Seer API expects.
```
ceorourke, line 52: Mostly as a note to myself, but after my [other pr](https://github.com/getsentry/sentry/pull/77755) lands we should be able to consolidate these functions and only have one. Mine serializes the data to get it into the format you're expecting here, but I can just do that before calling the function.

src/sentry/api/endpoints/organization_events_anomalies.py
```diff
@@ -0,0 +1,98 @@
...
+
+        historical_data = self._format_historical_data(request.data.get("historical_data"))
+        current_data = self._format_historical_data(request.data.get("current_data"))
+
+        config = request.data.get("config")
+        project_id = request.data.get("project_id")
+
+        if project_id is None or not config or not historical_data or not current_data:
+            return Response(
+                "Unable to get historical anomaly data: missing required argument(s) project, start, and/or end",
```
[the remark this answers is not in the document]
mifu67: No start and end for this endpoint, because we're passing in the time series data directly (check the test for an example)

src/sentry/api/endpoints/organization_events_anomalies.py
```diff
@@ -0,0 +1,98 @@
...
+                status=400,
+            )
+
+        anomalies = get_historical_anomaly_data_from_seer_preview(
+            current_data, historical_data, project_id, config
+        )
```
natemoo-re, line 88: Worth skipping this seer request if there isn't enough data to detect anomalies? Or should that be handled upstream in seer?

natemoo-re (approved):
Looks good! Consider my comments non-blocking, but a few small things to think through.

tests/sentry/api/endpoints/test_organization_historical_anomalies.py
```diff
@@ -0,0 +1,52 @@
...
+
+    method = "post"
+
+    @with_feature("organizations:anomaly-detection-alerts")
+    @with_feature("organizations:incidents")
+    def test_simple(self):
+        self.create_team(organization=self.organization, members=[self.user])
+        self.login_as(self.user)
+
+        data = {
```
ceorourke, line 24: Can you use the types here? `AnomalyDetectionConfig`, `TimeSeriesPoint`, etc

ceorourke (approved):
Left some non-blocking comments but otherwise looks good!

tests/sentry/api/endpoints/test_organization_historical_anomalies.py
mifu67: Going to add these once I implement the real Seer call!

src/sentry/api/endpoints/organization_events_anomalies.py
mifu67: We'll check in the backend code and return early there

schew2381 (approved):
for my 🟩

PR #77835, opened 2024-09-19
fix(routes): track last route in routeNotFound

michellewzhang:
### The problem
In Sentry today, tons of “route not found” events are grouped together into one issue, even if the suspect routes/transactions are not related. This poses a problem because the relevant teams are not alerted to these issues, which means 
1) we’re not able to triage & fix these issues as quickly, and 
2) we do not have proper scope of the problem areas. 

See an issue example [here](https://sentry.sentry.io/issues/496412334/events/c7dce88ad52944a095d672d0d114d7da/?project=11276&referrer=replay-errors), and notice that not all the [events](https://sentry.sentry.io/issues/496412334/events/c720b6d1807849b9ae0859f9b165a344/events/?project=11276&referrer=next-event) have the same transaction.

### The goal
Improve our grouping for “route not found” issues so that the events are split up into better issues — specifically, based on where the last known route was. That way, we can have more relevant issues grouped by (hopefully) the suspect problem route & triage more effectively.

### Next steps
Next step would be setting the fingerprinting rules in the `javascript` project to include the new tag-based grouping rule. This is in project settings:
SCR-20240919-tsyc

[notion doc](https://www.notion.so/sentry/Improve-Route-not-found-issue-grouping-1068b10e4b5d80129e5dc9b84b1b5eb5?showMoveTo=true&saveParent=true)

### DIFF
--- a/static/app/views/lastKnownRouteContextProvider.tsx
+++ b/static/app/views/lastKnownRouteContextProvider.tsx
@@ -8,7 +8,7 @@ interface Props {
   children: React.ReactNode;
 }
 
-export const LastKnownRouteContext = createContext<string>('unknown');
+export const LastKnownRouteContext = createContext<string>('<unknown>');
 
 export function useLastKnownRoute() {
   return useContext(LastKnownRouteContext);
--- a/static/app/views/lastKnownRouteContextProvider.tsx
+++ b/static/app/views/lastKnownRouteContextProvider.tsx
@@ -1,4 +1,4 @@
-import {createContext} from 'react';
+import {createContext, useContext} from 'react';
 
 import getRouteStringFromRoutes from 'sentry/utils/getRouteStringFromRoutes';
 import usePrevious from 'sentry/utils/usePrevious';
@@ -10,17 +10,19 @@ interface Props {
 
 export const LastKnownRouteContext = createContext<string>('');
 
-function useLastKnownRoute() {
-  const route = useRoutes();
-  const lastRoute = usePrevious(route);
-  return getRouteStringFromRoutes(lastRoute);
+export function useLastKnownRoute() {
+  return useContext(LastKnownRouteContext);
 }
 
+// This provider tracks the last known route that the user has navigated to.
+// This is used to better group issues when we hit "route not found" errors.
 export default function LastKnownRouteContextProvider({children}: Props) {
-  const route = useLastKnownRoute();
+  const route = useRoutes();
+  const prevRoute = usePrevious(route);
+  const lastKnownRoute = getRouteStringFromRoutes(prevRoute);
 
   return (
-    <LastKnownRouteContext.Provider value={route}>
+    <LastKnownRouteContext.Provider value={lastKnownRoute}>
       {children}
     </LastKnownRouteContext.Provider>
   );
--- a/static/app/views/lastKnownRouteContextProvider.tsx
+++ b/static/app/views/lastKnownRouteContextProvider.tsx
@@ -5,7 +5,7 @@ import usePrevious from 'sentry/utils/usePrevious';
 import {useRoutes} from 'sentry/utils/useRoutes';
 
 interface Props {
-  children?: React.ReactNode;
+  children: React.ReactNode;
 }
 
 export const LastKnownRouteContext = createContext<string>('');
--- a/static/app/views/lastKnownRouteContextProvider.tsx
+++ b/static/app/views/lastKnownRouteContextProvider.tsx
@@ -1,0 +1,27 @@
+import {createContext} from 'react';
+
+import getRouteStringFromRoutes from 'sentry/utils/getRouteStringFromRoutes';
+import usePrevious from 'sentry/utils/usePrevious';
+import {useRoutes} from 'sentry/utils/useRoutes';
+
+interface Props {
+  children?: React.ReactNode;
+}
+
+export const LastKnownRouteContext = createContext<string>('');
+
+function useLastKnownRoute() {
+  const route = useRoutes();
+  const lastRoute = usePrevious(route);
+  return getRouteStringFromRoutes(lastRoute);
+}
+
+export default function LastKnownRouteContextProvider({children}: Props) {
+  const route = useLastKnownRoute();
+
+  return (
+    <LastKnownRouteContext.Provider value={route}>
+      {children}
+    </LastKnownRouteContext.Provider>
+  );
+}
--- a/static/app/views/lastKnownRouteContextProvider.tsx
+++ b/static/app/views/lastKnownRouteContextProvider.tsx
@@ -8,14 +8,16 @@ interface Props {
   children: React.ReactNode;
 }
 
-export const LastKnownRouteContext = createContext<string>('');
+export const LastKnownRouteContext = createContext<string>('unknown');
 
 export function useLastKnownRoute() {
   return useContext(LastKnownRouteContext);
 }
 
-// This provider tracks the last known route that the user has navigated to.
-// This is used to better group issues when we hit "route not found" errors.
+/**
+ * This provider tracks the last known route that the user has navigated to.
+ * This is used to better group issues when we hit "route not found" errors.
+ */
 export default function LastKnownRouteContextProvider({children}: Props) {
   const route = useRoutes();
   const prevRoute = usePrevious(route);
--- a/static/app/views/routeNotFound.tsx
+++ b/static/app/views/routeNotFound.tsx
@@ -1,4 +1,4 @@
-import {useContext, useLayoutEffect} from 'react';
+import {useLayoutEffect} from 'react';
 import * as Sentry from '@sentry/react';
 
 import NotFound from 'sentry/components/errors/notFound';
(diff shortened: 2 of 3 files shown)


static/app/views/lastKnownRouteContextProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import getRouteStringFromRoutes from 'sentry/utils/getRouteStringFromRoutes';
+import usePrevious from 'sentry/utils/usePrevious';
+import {useRoutes} from 'sentry/utils/useRoutes';
+
+interface Props {
+  children?: React.ReactNode;
+}
+
+export const LastKnownRouteContext = createContext<string>('');
```
evanpurkhiser, line 11: Instead of exporting this let's repurpose `useLastKnownRoute` to look like this

```tsx
function useLastKnownRoute() {
	return useContext(LastKnownRouteContext);
}
```

static/app/views/lastKnownRouteContextProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+function useLastKnownRoute() {
+  const route = useRoutes();
+  const lastRoute = usePrevious(route);
+  return getRouteStringFromRoutes(lastRoute);
+}
```
evanpurkhiser, line 17: Ju

static/app/views/lastKnownRouteContextProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+interface Props {
+  children?: React.ReactNode;
+}
+
+export const LastKnownRouteContext = createContext<string>('');
+
+function useLastKnownRoute() {
+  const route = useRoutes();
+  const lastRoute = usePrevious(route);
+  return getRouteStringFromRoutes(lastRoute);
```
evanpurkhiser, line 16: Just inline this into the provider IMO

static/app/views/lastKnownRouteContextProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export default function LastKnownRouteContextProvider({children}: Props) {
```
evanpurkhiser, line 19: Let's add some comments describing what this is used for

static/app/views/routeNotFound.tsx
```diff
@@ -8,11 +8,13 @@ import SentryDocumentTitle from 'sentry/components/sentryDocumentTitle';
 import Sidebar from 'sentry/components/sidebar';
 import {t} from 'sentry/locale';
 import type {RouteComponentProps} from 'sentry/types/legacyReactRouter';
+import {LastKnownRouteContext} from 'sentry/views/lastKnownRouteContextProvider';
 
 type Props = RouteComponentProps<{}, {}>;
 
 function RouteNotFound({router, location}: Props) {
   const {pathname, search, hash} = location;
+  const lastRoute = useContext(LastKnownRouteContext);
```
evanpurkhiser, line 17: Let's change like this
```suggestion
  const lastRoute = useLastKnownRoute();
```

static/app/views/lastKnownRouteContextProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+interface Props {
+  children?: React.ReactNode;
+}
+
+export const LastKnownRouteContext = createContext<string>('');
+
+function useLastKnownRoute() {
+  const route = useRoutes();
+  const lastRoute = usePrevious(route);
```
evanpurkhiser, line 15: I wonder if we should consider excluding the route not found route, since it isn't really "known". Not that it would be a problem I think

static/app/views/lastKnownRouteContextProvider.tsx
evanpurkhiser, line 11: let's start with a default like `<unknown>`

static/app/views/lastKnownRouteContextProvider.tsx
```diff
@@ -0,0 +1,29 @@
+import {createContext, useContext} from 'react';
+
+import getRouteStringFromRoutes from 'sentry/utils/getRouteStringFromRoutes';
+import usePrevious from 'sentry/utils/usePrevious';
+import {useRoutes} from 'sentry/utils/useRoutes';
+
+interface Props {
+  children?: React.ReactNode;
```
JonasBa, line 8:
```suggestion
  children: React.ReactNode;
```
we always want the children for this provider

static/app/views/routeNotFound.tsx
```diff
@@ -8,11 +8,13 @@ import SentryDocumentTitle from 'sentry/components/sentryDocumentTitle';
 import Sidebar from 'sentry/components/sidebar';
 import {t} from 'sentry/locale';
 import type {RouteComponentProps} from 'sentry/types/legacyReactRouter';
+import {useLastKnownRoute} from 'sentry/views/lastKnownRouteContextProvider';
 
 type Props = RouteComponentProps<{}, {}>;
 
 function RouteNotFound({router, location}: Props) {
   const {pathname, search, hash} = location;
+  const lastRoute = useLastKnownRoute();
```
JonasBa, line 17:
```suggestion
  const lastKnownRoute = useLastKnownRoute();
```
naming nit

static/app/views/routeNotFound.tsx
```diff
@@ -27,13 +29,14 @@ function RouteNotFound({router, location}: Props) {
       scope.setFingerprint(['RouteNotFound']);
       scope.setTag('isMissingSlash', isMissingSlash);
       scope.setTag('pathname', pathname);
+      scope.setTag('lastKnownRoute', lastRoute);
```
JonasBa, line 32:
```suggestion
      scope.setTag('lastKnownRoute', lastKnownRoute);
```
naming nit

static/app/views/lastKnownRouteContextProvider.tsx
```diff
@@ -0,0 +1,29 @@
...
+export function useLastKnownRoute() {
+  return useContext(LastKnownRouteContext);
+}
+
+// This provider tracks the last known route that the user has navigated to.
+// This is used to better group issues when we hit "route not found" errors.
+export default function LastKnownRouteContextProvider({children}: Props) {
+  const route = useRoutes();
+  const prevRoute = usePrevious(route);
+  const lastKnownRoute = getRouteStringFromRoutes(prevRoute);
```
JonasBa, line 22: I wish we could just use the location state, but we call navigate in so many areas that it would be really annoying to update

static/app/views/lastKnownRouteContextProvider.tsx
```diff
@@ -0,0 +1,29 @@
...
+import {useRoutes} from 'sentry/utils/useRoutes';
+
+interface Props {
+  children?: React.ReactNode;
+}
+
+export const LastKnownRouteContext = createContext<string>('');
+
+export function useLastKnownRoute() {
+  return useContext(LastKnownRouteContext);
```
JonasBa, line 14: Do we care if this hook ever gets called outside of the provider?

JonasBa (approved):
Just a couple small comments. Great improvement btw, that umbrella issue is probably hiding all sorts of bad links around the dashboard

static/app/views/lastKnownRouteContextProvider.tsx
michellewzhang: would you cast the return value of the hook then?

```
export function useLastKnownRoute() {
  return useContext(LastKnownRouteContext) as string;
}

```

static/app/views/lastKnownRouteContextProvider.tsx
michellewzhang: hmm, i guess it _could_ possibly be useful elsewhere in the future, so it's not awful if it does get called outside of here. the only place we currently need it is inside `routeNotFound.tsx`

static/app/views/lastKnownRouteContextProvider.tsx
michellewzhang: what do you mean by "exclude"?

static/app/views/lastKnownRouteContextProvider.tsx
evanpurkhiser: Oh sorry I literally mean the string “<unknown>”

static/app/views/lastKnownRouteContextProvider.tsx
```diff
@@ -0,0 +1,29 @@
...
+  children: React.ReactNode;
+}
+
+export const LastKnownRouteContext = createContext<string>('');
+
+export function useLastKnownRoute() {
+  return useContext(LastKnownRouteContext);
+}
+
+// This provider tracks the last known route that the user has navigated to.
```
evanpurkhiser, line 17: This should be a jsdoc style comment. Otherwise the LSP won’t pick this up as a documentation comment

static/app/views/routeNotFound.tsx
```diff
@@ -27,13 +29,14 @@ function RouteNotFound({router, location}: Props) {
       scope.setFingerprint(['RouteNotFound']);
```
billyvg, line 29: You could alternatively set the fingerprint here, though it'd be less flexible

michellewzhang: i think i might prefer to do it in settings!

static/app/views/lastKnownRouteContextProvider.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import getRouteStringFromRoutes from 'sentry/utils/getRouteStringFromRoutes';
+import usePrevious from 'sentry/utils/usePrevious';
+import {useRoutes} from 'sentry/utils/useRoutes';
+
+interface Props {
+  children: React.ReactNode;
+}
+
+export const LastKnownRouteContext = createContext<string>('unknown');
```
evanpurkhiser, line 11: lol sorry, I was reviewing on my phone and it lost the `<`. I actually meant the string `<unknown>` so it's very easy to disambiguate it from a route named `unknown` or whatever

michellewzhang: updated again 😆

evanpurkhiser (approved):

PR #77838, opened 2024-09-20
chore(grouping): Stop calling `_save_aggregate`

lobsterkatie:
As of https://github.com/getsentry/sentry/pull/77027, all projects are using the new grouping config transition flow, because `project_uses_optimized_grouping` now always returns `True`. We can therefore start removing the old code path. 

This takes the first step in that direction by removing the call to the old version of the `_save_aggregate` function. Though this isn't strictly necessary before removing the function (since now no one should be going through the branch which calls it), this makes it extra obvious that nothing can go wrong when that happens.

### DIFF
--- a/src/sentry/event_manager.py
+++ b/src/sentry/event_manager.py
@@ -1323,20 +1323,11 @@ def get_culprit(data: Mapping[str, Any]) -> str:
 
 @sentry_sdk.tracing.trace
 def assign_event_to_group(event: Event, job: Job, metric_tags: MutableTags) -> GroupInfo | None:
-    if job["optimized_grouping"]:
-        group_info = _save_aggregate_new(
-            event=event,
-            job=job,
-            metric_tags=metric_tags,
-        )
-    else:
-        group_info = _save_aggregate(
-            event=event,
-            job=job,
-            release=job["release"],
-            received_timestamp=job["received_timestamp"],
-            metric_tags=metric_tags,
-        )
+    group_info = _save_aggregate_new(
+        event=event,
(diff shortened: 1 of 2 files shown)


src/sentry/event_manager.py
```diff
@@ -1323,20 +1323,11 @@ def get_culprit(data: Mapping[str, Any]) -> str:
...
-        )
-    else:
-        group_info = _save_aggregate(
-            event=event,
-            job=job,
-            release=job["release"],
-            received_timestamp=job["received_timestamp"],
-            metric_tags=metric_tags,
-        )
+    group_info = _save_aggregate_new(
```
armenzg, line 1326: Can we renamed it as part of this?

armenzg (changes requested):
Can you please make all changes in one shot? It does not seem to be too many changes and the PR should not be too overwhelming to review.

lobsterkatie:
> Can you please make all changes in one shot? It does not seem to be too many changes and the PR should not be too overwhelming to review.

Okay - I was going to do it in stages and PR as I went, but lemme actually do all the steps and see how big it ends up being. I'll leave this open for now but won't merge it.

lobsterkatie:
UPDATE:

I did the whole thing out, and it was pretty big, so I ended up splitting it in 4:

https://github.com/getsentry/sentry/pull/78260
https://github.com/getsentry/sentry/pull/78261
https://github.com/getsentry/sentry/pull/78262
https://github.com/getsentry/sentry/pull/78268

Closing this in favor of those.

PR #77839, opened 2024-09-20
ref(flags): search button leads to focus in drawer

michellewzhang:
Add a search icon button to the feature flag section, which opens the fly-out drawer with the search focused. This is the same behavior as the breadcrumbs section.

https://github.com/user-attachments/assets/553cfd39-f4b3-480b-aa0d-ce961e2d6361


static/app/components/events/featureFlags/featureFlagDrawer.tsx
```diff
@@ -74,6 +74,22 @@ interface FlagDrawerProps {
...
+  // This will allow us to simulate :focus-visible on the button elements.
+  const getFocusProps = useCallback(
+    (option: FlagControlOptions) => {
+      return option === focusControl
+        ? {autoFocus: true, onBlur: () => setFocusControl(undefined)}
+        : {};
+    },
+    [focusControl]
+  );
+  return {getFocusProps};
```
michellewzhang, line 92: this is the same function as in `breadcrumbsDrawer.tsx`:
https://github.com/getsentry/sentry/blob/f98fe827f109f9946aaeeca3473f6436c4f58fc1/static/app/components/events/breadcrumbs/breadcrumbsDrawer.tsx#L52-L65

i _could_ rewrite the function to be more generic so that it can be imported by both files. don't know if anyone has a super strong opinion about this

static/app/components/events/featureFlags/eventFeatureFlagList.tsx
```diff
@@ -128,10 +133,18 @@ export function EventFeatureFlagList({
   const actions = (
     <ButtonBar gap={1}>
       {feedbackButton}
+      <Button
+        aria-label={t('Open Breadcrumb Search')}
```
c298lee, line 137:
```suggestion
        aria-label={t('Open Feature Flag Search')}
```

static/app/components/events/featureFlags/featureFlagDrawer.tsx
c298lee: I think a generic function would be better, but also no strong opinion on this

c298lee (approved):

static/app/components/events/breadcrumbs/breadcrumbsDrawer.tsx
```diff
@@ -49,21 +50,6 @@ export const enum BreadcrumbControlOptions {
   SORT = 'sort',
 }
 
-function useFocusControl(initialFocusControl?: BreadcrumbControlOptions) {
```
michellewzhang, line 52: moved to `useFocusControl.tsx` to be reusable

